PR #35, opened 2026-04-13
feat(asyncio): async DB layer, budget strategies, and log writer

tbille:
## Summary
- Converts entire DB layer from sync `psycopg2` + `Session` to async `asyncpg` + `AsyncSession`. Contended DB calls now yield to the event loop instead of blocking it.
- Adds pluggable `budget_strategy` config: `for_update` (default/legacy), `cas` (lock-free, recommended), `disabled` (skip budget checks)
- Adds pluggable `log_writer_strategy` config: `single` (inline, default), `batch` (background flush for high-throughput)
- Adds DB pool tuning config: `db_pool_size`, `db_max_overflow`, `db_pool_timeout`, `db_pool_recycle`
- Adds new Prometheus metrics for log writer queue depth, batch size, flush duration, and row counts
- App startup uses FastAPI lifespan for proper async initialization and graceful shutdown
- Alembic migrations remain sync (psycopg2) for compatibility

Ported from https://github.com/mozilla-ai/any-llm/pull/1001

Co-Authored-By: Julian Bright <brightsparc@gmail.com>

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -1,0 +1,3 @@
+mode: platform
+platform:
+  base_url: http://localhost:8100/api/v1
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -10,6 +10,8 @@ requires-python = ">=3.13"
 dependencies = [
     "any-llm-sdk[all]",
     "alembic>=1.13.0",
+    "aiosqlite>=0.19.0",
+    "asyncpg>=0.29.0",
     "click>=8.1.0",
     "fastapi>=0.115.0",
     "prometheus-client>=0.20.0",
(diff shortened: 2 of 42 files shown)


brightsparc:
@tbille looks like you didn't bring over the load testing script - I think that could still be useful, should I create a PR ?

tbille:
@brightsparc good catch. Can you make a PR, I'll review it today

brightsparc:
> @brightsparc good catch. Can you make a PR, I'll review it today

Here is the PR, I also noticed a small bug when re-running, so added graceful shutdown to capture all logs
https://github.com/mozilla-ai/gateway/pull/42

PR #45, opened 2026-04-15
feat(gateway): rename auth header to AnyLLM-Key (RFC 6648)

peteski22:
## Summary

Rename the gateway's Bearer-authentication header from `X-AnyLLM-Key` to `AnyLLM-Key`, so the name follows [RFC 6648](https://datatracker.ietf.org/doc/html/rfc6648) (no `X-` prefix). The gateway has no live deployments yet, so there is no backward-compatibility path — this is a simple rename plus removal of the now-unneeded `x-api-key` compat fallback.

### What changes

- `core/config.py`: `API_KEY_HEADER` flips to `"AnyLLM-Key"`.
- `api/deps.py`: `_extract_bearer_token` accepts `AnyLLM-Key` first, then `Authorization`. The `x-api-key` Anthropic-compat fallback is removed. The three `verify_*` docstrings reference the new header name.
- `main.py`: the CORS `allow_headers` list uses the `API_KEY_HEADER` constant and drops the `x-api-key` entry.
- `tests/integration/test_client_args.py`, `tests/integration/test_provider_kwargs_override.py`: hard-coded header literals replaced with `API_KEY_HEADER`.
- `tests/integration/test_messages_endpoint.py`: `x-api-key` auth test removed (path no longer exists).
- `tests/unit/test_extract_bearer_token.py` (new): unit tests of `_extract_bearer_token` covering `AnyLLM-Key`/`Authorization` precedence, malformed (non-`Bearer`) input on each header, and the missing-credentials 401.

### History

Earlier revisions of this PR shipped a full RFC 6648 / 9745 / 8594 deprecation design (legacy header kept, `Deprecation`/`Sunset` response headers, `DeprecationHeadersMiddleware`). `@tbille` pointed out the gateway is not yet live, so there is no old name to deprecate; the branch was force-pushed to drop the deprecation machinery. A follow-up review flagged the remaining `x-api-key` compat fallback on the same grounds — that has now been removed too.

### Commits

1. `feat(gateway): rename API_KEY_HEADER from X-AnyLLM-Key to AnyLLM-Key`
2. `test(gateway): cover bearer-token extraction precedence and error paths`
3. `refactor(gateway): drop x-api-key fallback from bearer extraction`

## Test plan

- [x] `make test` — 313 passed, 9 skipped, no regressions
- [x] `make lint` — clean
- [x] `make typecheck` — clean
- [x] Unit tests cover precedence between `AnyLLM-Key` and `Authorization`, malformed-input branches on each, and the missing-credentials 401

### DIFF
--- a/src/gateway/api/deps.py
+++ b/src/gateway/api/deps.py
@@ -42,9 +42,8 @@ def reset_config() -> None:
 def _extract_bearer_token(request: Request, config: GatewayConfig) -> str:
     """Extract and validate Bearer token from request header.
 
-    Checks AnyLLM-Key first, then falls back to standard Authorization header
-    for OpenAI client compatibility, then falls back to x-api-key header
-    for Anthropic client compatibility.
+    Checks the canonical AnyLLM-Key header first, then falls back to the
+    standard Authorization header.
     """
     auth_header = request.headers.get(API_KEY_HEADER) or request.headers.get("Authorization")
 
@@ -56,11 +55,6 @@ def _extract_bearer_token(request: Request, config: GatewayConfig) -> str:
                 detail="Invalid header format. Expected 'Bearer <token>'",
             )
         return auth_header[7:]
-
-    # Fallback: x-api-key header (Anthropic client compatibility, no Bearer prefix).
-    api_key = request.headers.get("x-api-key")
-    if api_key:
-        return api_key
 
     record_auth_failure("missing_credentials")
     raise HTTPException(
--- a/src/gateway/api/deps.py
+++ b/src/gateway/api/deps.py
@@ -42,7 +42,7 @@ def reset_config() -> None:
 def _extract_bearer_token(request: Request, config: GatewayConfig) -> str:
     """Extract and validate Bearer token from request header.
 
-    Checks X-AnyLLM-Key first, then falls back to standard Authorization header
+    Checks AnyLLM-Key first, then falls back to standard Authorization header
     for OpenAI client compatibility, then falls back to x-api-key header
     for Anthropic client compatibility.
     """
@@ -57,7 +57,7 @@ def _extract_bearer_token(request: Request, config: GatewayConfig) -> str:
             )
         return auth_header[7:]
 
-    # Fallback: x-api-key header (Anthropic client compatibility, no Bearer prefix)
+    # Fallback: x-api-key header (Anthropic client compatibility, no Bearer prefix).
     api_key = request.headers.get("x-api-key")
     if api_key:
         return api_key
@@ -130,7 +130,7 @@ async def verify_api_key(
     db: Annotated[AsyncSession, Depends(get_db)],
     config: Annotated[GatewayConfig, Depends(get_config)],
 ) -> APIKey:
-    """Verify API key from X-AnyLLM-Key header.
+    """Verify API key from AnyLLM-Key header.
 
     Args:
         request: FastAPI request object
@@ -152,7 +152,7 @@ async def verify_master_key(
     request: Request,
     config: Annotated[GatewayConfig, Depends(get_config)],
 ) -> None:
-    """Verify master key from X-AnyLLM-Key header.
+    """Verify master key from AnyLLM-Key header.
 
     Args:
         request: FastAPI request object
@@ -182,7 +182,7 @@ async def verify_api_key_or_master_key(
     db: Annotated[AsyncSession, Depends(get_db)],
     config: Annotated[GatewayConfig, Depends(get_config)],
 ) -> tuple[APIKey | None, bool]:
-    """Verify either API key or master key from X-AnyLLM-Key header.
+    """Verify either API key or master key from AnyLLM-Key header.
 
     Args:
         request: FastAPI request object
--- a/src/gateway/core/config.py
+++ b/src/gateway/core/config.py
@@ -9,8 +9,6 @@ from dotenv import load_dotenv
 from pydantic import BaseModel, Field
 from pydantic_settings import BaseSettings, SettingsConfigDict
 
-# HTTP request header name for gateway Bearer authentication. Follows RFC 6648
-# (no `X-` prefix).
 API_KEY_HEADER = "AnyLLM-Key"
 
 
--- a/src/gateway/core/config.py
+++ b/src/gateway/core/config.py
@@ -9,7 +9,9 @@ from dotenv import load_dotenv
 from pydantic import BaseModel, Field
 from pydantic_settings import BaseSettings, SettingsConfigDict
 
-API_KEY_HEADER = "X-AnyLLM-Key"
+# HTTP request header name for gateway Bearer authentication. Follows RFC 6648
+# (no `X-` prefix).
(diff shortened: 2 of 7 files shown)


src/gateway/api/deps.py
```diff
@@ -9,15 +9,46 @@
...
+_LEGACY_HEADER_STATE_ATTR = "legacy_api_key_header_used"
+
+
+def flag_legacy_header(request: Request) -> None:
+    """Flag on the request state that a deprecated credential header supplied the token."""
+    setattr(request.state, _LEGACY_HEADER_STATE_ATTR, True)
+
+
+def has_legacy_header(request: Request) -> bool:
+    """Report whether the request authenticated via a deprecated credential header."""
```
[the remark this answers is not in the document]
peteski22: Confirmed intentional behaviour — malformed legacy credentials should still surface the Sunset signal so the caller learns to migrate (that's the `test_malformed_legacy_header_raises_401_but_still_flags_deprecation` case). Reworded the helper docstrings and the adjacent attribute comment in e66706a to say "used" instead of "authenticated" / "supplied the token", matching the pre-validation semantics. Also updated the `logger.warning` message ("Request used deprecated %s header").

src/gateway/main.py
```diff
@@ -137,11 +143,29 @@ async def dispatch(self, request: Request, call_next: RequestResponseEndpoint) -
...
         return response
 
 
+class DeprecationHeadersMiddleware(BaseHTTPMiddleware):
+    """Signal deprecation of the legacy ``X-AnyLLM-Key`` authentication header.
+
+    When a request authenticated via the legacy header flag is observed on
+    ``request.state``, append RFC 9745 ``Deprecation`` and RFC 8594 ``Sunset``
+    response headers so clients can detect and plan migration to the canonical
+    ``AnyLLM-Key`` header.
```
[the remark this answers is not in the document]
peteski22: Same root observation as the `deps.py` thread — flag set on header *use*, not *success*. Middleware docstring updated in e66706a to call out the malformed-credential case explicitly ("including on requests whose credential was malformed") rather than implying post-auth semantics.

src/gateway/api/deps.py
```diff
@@ -9,15 +9,48 @@
...
+
+
+def flag_legacy_header(request: Request) -> None:
+    """Flag on the request state that the deprecated credential header was used."""
+    setattr(request.state, _LEGACY_HEADER_STATE_ATTR, True)
+
+
+def has_legacy_header(request: Request) -> bool:
+    """Report whether the deprecated credential header was used on this request."""
+    return bool(getattr(request.state, _LEGACY_HEADER_STATE_ATTR, False))
```
tbille: no need for deprecation notice. This feature is not live yet.

tbille:
@peteski22 this looks great. Since the gateway with the platform is not yet live, no need to add all the deprecation code. Could you remove it?

src/gateway/api/deps.py
peteski22: I wasn't sure if this gateway was a migration of the any-llm one where the `x-` prefixed key might still be being used? If not, then I can ditch all that 😄

peteski22: Done — force-pushed the branch to drop all the legacy/deprecation machinery. PR is now just the rename + unit tests for extraction precedence.

peteski22:
Force-pushed the branch with the deprecation code removed — PR is now a straight rename from `X-AnyLLM-Key` to `AnyLLM-Key` plus the unit tests that cover header-precedence and malformed-input branches. Title and summary updated to match.

src/gateway/api/deps.py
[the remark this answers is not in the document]
peteski22: Superseded: the deprecation machinery these docstrings described has been removed entirely in the force-push — see the top-level comment for context.

src/gateway/main.py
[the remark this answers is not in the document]
peteski22: Superseded: `DeprecationHeadersMiddleware` and all legacy-header handling have been removed — see the top-level comment for context.

tests/unit/test_extract_bearer_token.py
```diff
@@ -0,0 +1,112 @@
...
+        {
+            API_KEY_HEADER: "Bearer canonical-wins",
+            "Authorization": "Bearer authorization-loses",
+        }
+    )
+
+    assert _extract_bearer_token(request, config) == "canonical-wins"
+
+
+def test_canonical_takes_precedence_over_x_api_key(config: GatewayConfig) -> None:
```
tbille: No need for this legacy test

tests/unit/test_extract_bearer_token.py
```diff
@@ -0,0 +1,112 @@
...
+        {
+            API_KEY_HEADER: "Bearer canonical-wins",
+            "x-api-key": "x-api-key-loses",
+        }
+    )
+
+    assert _extract_bearer_token(request, config) == "canonical-wins"
+
+
+def test_authorization_takes_precedence_over_x_api_key(config: GatewayConfig) -> None:
```
tbille: no need for this legacy test

src/gateway/core/config.py
```diff
@@ -9,7 +9,9 @@
 from pydantic import BaseModel, Field
 from pydantic_settings import BaseSettings, SettingsConfigDict
 
-API_KEY_HEADER = "X-AnyLLM-Key"
+# HTTP request header name for gateway Bearer authentication. Follows RFC 6648
+# (no `X-` prefix).
```
tbille: No need for this legacy comment

src/gateway/api/deps.py
```diff
@@ -57,7 +57,7 @@ def _extract_bearer_token(request: Request, config: GatewayConfig) -> str:
             )
         return auth_header[7:]
 
-    # Fallback: x-api-key header (Anthropic client compatibility, no Bearer prefix)
+    # Fallback: x-api-key header (Anthropic client compatibility, no Bearer prefix).
```
tbille: Let's get rid of the old way without having a fallback. This part is not released yet. It's ok to not support an unrelease change

tbille (commented):
There are still a few legacy behaviours that we can remove without having the full backward compatibility

src/gateway/api/deps.py
peteski22: Done. Dropped the x-api-key fallback in deps.py and removed its CORS entry in main.py. Kept the Authorization header (standard Bearer). Updated the docstring accordingly.

tests/unit/test_extract_bearer_token.py
peteski22: Removed — along with test_x_api_key_returns_raw_key and the integration test test_messages_endpoint_x_api_key_header, since the x-api-key path no longer exists.

tests/unit/test_extract_bearer_token.py
peteski22: Removed.

src/gateway/core/config.py
peteski22: Removed.

tbille (approved):
amazing thanks

PR #58, opened 2026-05-07
feat: routing-policy fallback for chat completions (streaming + non-streaming)

agpituk:
## Summary

Extends the gateway's platform-mode resolve protocol to support an ordered list of resolution attempts. When the platform peer returns multiple attempts (driven by a user-configured routing policy on the platform side), the gateway walks them in order and falls through to the next entry on retryable upstream failures. Covers **both non-streaming and streaming** requests.

Single-attempt resolves and older platform peers using the flat legacy shape continue to work unchanged.

## Why this change

A user who configures a fallback chain like `[anthropic/claude-sonnet-4-5, openai/gpt-4o]` on the platform side wants the gateway to handle a regional outage transparently — Anthropic 401 / 429 / 5xx falls through to OpenAI without the client seeing the failure. The gateway is the only place that observes upstream errors at request time, so the retry loop has to live here. Policy is configured platform-side; gateway just walks whatever ordered list it gets back.

### `feat(chat)`: non-streaming fallback walk
- New `ResolvedAttempt` / `ResolvedRoute` Pydantic models for the parsed resolve payload.
- `_parse_resolve_payload` accepts **either** the new `attempts[]` shape **or** the legacy single-attempt shape (flat `provider`/`model`/`api_key`/... + top-level `correlation_id`). Backwards compatible — older platform peers don't need to change anything.
- For each attempt, classify failures via `_classify_upstream_error` (timeouts, connection errors, 5xx, 429, 408, **and 401/403** → retryable; 400/422/404 → not). Each attempt sends a usage report so the platform attributes spend correctly.
- Empty `attempts[]` returns 502 with a clear message rather than falling through silently.
- New optional `error_class` field on usage reports (`timeout` / `conn_err` / `http_<code>` / `unknown`).
- New `X-Otari-Request-ID` response header surfaces the resolve `request_id` (groups all attempts for a single request).

### `feat(streaming)`: streaming fallback via first-chunk gate
- Streaming requests now walk `attempts[]` the same way non-streaming requests do.
- Per-attempt **first-chunk gate**: open the upstream stream, wait up to `STREAMING_FALLBACK_FIRST_CHUNK_TIMEOUT_MS` (default 2000 ms) for the first chunk. If the upstream raises before yielding or the wait times out, fall through to the next attempt. As soon as a chunk lands, commit — bytes flow to the client and any further errors propagate to the SSE channel.
- **Latency contract: zero added latency in the success case** — the first-chunk-arrival event is itself the "this attempt is alive" signal; there's no fixed buffer window to wait through. Failure-path latency is bounded by the per-attempt timeout.
- New `walk_streaming_attempts()` helper in `gateway/streaming.py`, generic over attempt type.
- `X-Correlation-ID` is now set on the `StreamingResponse.headers` directly (mutating the dependency-injected `Response` doesn't propagate to streaming responses).

### `docs`: protocol spec + roadmap
- `docs/platform-protocol.md` documents both endpoints, both response shapes (legacy + multi-attempt), retry semantics, the error-class taxonomy, the streaming first-chunk-gate, and the relevant config env vars.
- `docs/routing-roadmap.md` captures the design intent for v1.2 (policy-level timeouts), v1.3 (per-entry retries with backoff), and v1.4 (global latency budget) so future contributors have a coherent target without committing to scope here.
- README's "Start in platform mode" section links to the protocol doc.

## Why 401/403 are retryable

These are usually terminal client errors and a single-attempt request still surfaces them directly. They become retryable **only** inside a multi-attempt route — i.e. when the user has explicitly configured a fallback chain on the platform side. The whole point of the policy is that one provider's bad credentials should fall through to the next, not surface to the client. Comments at the constant declaration explain this.

## Tests

- 94 tests pass (84 existing unit + 10 platform-mode integration including 3 new tests covering: backwards-compat legacy shape, streaming fall-through on first-attempt failure, streaming all-attempts-fail returns 502).
- No behavioral regressions in standalone mode.
- OpenAPI spec freshness check (`scripts/generate_openapi.py --check`) passes — gateway's own surface didn't change; the resolve and usage shapes are in the platform peer's API.
- mypy clean.

## AI Usage

Drafted with Claude Code (Claude Opus 4.7). Reviewed and validated locally before submitting.

### DIFF
--- a/docs/platform-protocol.md
+++ b/docs/platform-protocol.md
@@ -1,0 +1,177 @@
+# Platform protocol
+
+When the gateway runs in **platform mode** (`OTARI_PLATFORM_TOKEN` is set), it
+delegates per-request authorization and provider-credential resolution to a
+peer platform service over HTTP. This document describes the wire contract
+the gateway expects from that peer.
+
+The default peer implementation is [otari](https://github.com/mozilla-ai/otari),
+but any service that implements this contract can stand in.
+
+## Endpoints
+
+The gateway calls two endpoints, both rooted at `PLATFORM_BASE_URL`:
+
+| Endpoint | Purpose |
+|---|---|
+| `POST {base}/gateway/provider-keys/resolve` | Authorize a request and return one or more provider credentials to try |
+| `POST {base}/gateway/usage`                 | Report the outcome of an attempt back to the platform |
+
+## Authentication
+
+Both endpoints require `X-Gateway-Token: <gw_...>` in the request headers. This
+proves the caller is the gateway instance configured against this platform
+deployment. The resolve endpoint additionally requires `X-User-Token: <tk_...>`,
+which is the workspace API token forwarded opaquely from the end user's
+`Authorization: Bearer ...` header.
+
+## Resolve
+
+### Request
+
+```http
+POST /gateway/provider-keys/resolve
+X-Gateway-Token: gw_...
+X-User-Token: tk_...
+Content-Type: application/json
+
+{
+  "model": "gpt-4o-mini",
+  "provider": "openai"          // optional; otherwise inferred from model prefix
+}
+```
+
+### Response — multi-attempt shape (preferred)
+
+```json
+{
+  "request_id": "01HXY...",
+  "fallback_enabled": true,
+  "attempts": [
+    {
+      "attempt_id": "01HX1...",
+      "position": 0,
+      "provider": "anthropic",
+      "model": "claude-sonnet-4-5",
+      "api_key": "sk-ant-...",
+      "api_base": null,
+      "managed": false
+    },
+    {
+      "attempt_id": "01HX2...",
+      "position": 1,
+      "provider": "openai",
+      "model": "gpt-4o",
+      "api_key": "sk-...",
+      "api_base": "https://api.openai.com/v1",
+      "managed": false
+    }
+  ]
+}
+```
+
+The gateway walks `attempts` in order. On a retryable failure it moves to the
+next entry; on success it stops. The `attempt_id` of the entry that ultimately
+succeeded (or the last one tried, on total failure) is what the gateway echoes
+back via `X-Correlation-ID` and reports through `/gateway/usage`.
+
+`request_id` groups every `attempt_id` from the same resolve call so the
+platform can attribute spend, render trace timelines, and emit fallback events.
+The gateway also surfaces it as the `X-Otari-Request-ID` response header.
+
+`fallback_enabled` is informational — set by the platform when its routing
+policy actually allows fallback (i.e. the policy has multiple enabled entries
+and `fallback_enabled = true`). The gateway uses `len(attempts) > 1` for its
+own behaviour.
+
+`attempts` MUST contain at least one entry. An empty list is treated as a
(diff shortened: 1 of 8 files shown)


docs/platform-protocol.md
```diff
@@ -0,0 +1,210 @@
...
+The gateway calls two endpoints, both rooted at `PLATFORM_BASE_URL`:
+
+| Endpoint | Purpose |
+|---|---|
+| `POST {base}/gateway/provider-keys/resolve` | Authorize a request and return one or more provider credentials to try |
+| `POST {base}/gateway/usage`                 | Report the outcome of an attempt back to the platform |
+
+## Authentication
+
+Both endpoints require `X-Gateway-Token: <gw_...>` in the request headers. This
```
tbille, line 24: To double check, I think it's not called this way.

agpituk: sorry what do you mean? https://github.com/mozilla-ai/otari/blob/cb0b47d0c34c167f2a4d3680afe089d580441666/src/gateway/api/routes/chat.py#L272

tbille:
## Manual QA: Standalone Mode Testing

Ran manual testing against this branch in **standalone mode** (no platform token set) with:
- **OpenAI**: valid API key
- **Anthropic**: intentionally invalid API key (realistic format, fake credentials)

### Test Results

| # | Test | Expected | Actual | Result |
|---|------|----------|--------|--------|
| 1 | `GET /health` | 200 | 200 | PASS |
| 2 | Chat with invalid gateway key | 401 | 401 | PASS |
| 3 | OpenAI non-streaming (valid key) | 200 with `choices` | 200 | PASS |
| 4 | OpenAI streaming (valid key) | 200 SSE with `data:` + `[DONE]` | 200 | PASS |
| 5 | Anthropic non-streaming (fake key) | 502 `"LLM provider error"` | 502 | PASS |
| 6 | Anthropic streaming (fake key) | 200 with error event in SSE | 200 with `data: {"error": ...}` | PASS |

All 6 scenarios pass. The standalone code path is not broken by the platform-mode multi-attempt changes.

### Pre-existing bug found during testing (not introduced by this PR)

**`TypeError: can't subtract offset-naive and offset-aware datetimes`** in `src/gateway/api/deps.py:112`

When running with **SQLite** (the default for local/standalone), the `last_used_at` column comes back as a naive datetime despite the model declaring `DateTime(timezone=True)`. On the **second** request using the same API key, the subtraction `datetime.now(UTC) - api_key.last_used_at` crashes because one is timezone-aware and the other is naive.

- **First request** works because `last_used_at is None` short-circuits the check.
- **Every subsequent request** hits the `TypeError` and returns 500.
- The same issue exists for `expires_at` on line 102.
- This only affects SQLite -- PostgreSQL stores and returns timezone-aware datetimes correctly.

**Suggested fix** (for a separate PR):
```python
last_used = api_key.last_used_at
if last_used is not None and last_used.tzinfo is None:
    last_used = last_used.replace(tzinfo=UTC)
```

Same pattern needed for `expires_at`. This should probably be addressed before merge if standalone + SQLite is a supported configuration, since the multi-attempt feature will increase request volume and make this more visible.

tbille (approved):
This is awesome!
LGTM

PR #63, opened 2026-05-14
feat(chat): MCP tool-use support with streaming

agpituk:
## Summary

Adds MCP tool-use support to `/v1/chat/completions`. When a request body includes `mcp_servers`, the gateway connects to each server (Streamable HTTP), exposes its tools to the model in OpenAI function-tool format, runs an internal tool-use loop (executes tool calls, feeds results back, re-streams the model's response), and returns the final assistant message to the caller. Streaming-aware.

Any model the gateway can route to — open-weight or frontier — gains MCP tool-use through this code path. The model never knows MCP exists; it sees a regular `tools[]` array.

## What's in it

**The loop itself** (`src/gateway/services/mcp_loop.py`)
- Streaming and non-streaming variants. The streaming variant yields chunks across multiple inner `acompletion` calls as a single `AsyncIterator[ChatCompletionChunk]`, so the existing SSE pipeline (`streaming_generator`) is unchanged.
- Tool execution surfaces failures as `role: "tool"` messages with `[tool error]` so the model can recover. Cancellation and programming errors propagate.
- Per-iteration cap (`max_tool_iterations`, default 10, server cap 25). Configurable per-request.
- Multi-turn tool calls in a single round are dispatched together.

**MCP client** (`src/gateway/services/mcp_client.py`)
- `MCPClientPool` is an async-context-managed pool of live MCP sessions for one request's lifetime.
- Converts MCP tools to OpenAI function-tool format. Tool name collisions across servers: first server wins, later same-named tools dropped entirely.
- Smart content-block flattening: text passes through verbatim; image and embedded-resource blocks render as informative `[image type=… ]` / `[resource uri=…]` placeholders.

**URL safety** (`src/gateway/services/url_safety.py`)
- Rejects RFC 1918 / link-local / multicast / unspecified / reserved IP ranges. DNS resolution is checked too.
- Loopback allowed by default (sidecars, dev). Override via `GATEWAY_MCP_ALLOW_LOOPBACK=false`.
- Operator-wide opt-out via `GATEWAY_MCP_ALLOW_PRIVATE_HOSTS=true`.
- Plain `http://` URLs rejected when an `authorization_token` is set — credentials don't go on the wire in cleartext.

**Request body fields**
```jsonc
{
  "mcp_servers": [
    {
      "name": "calendar",
      "url": "https://my-calendar.example.com/mcp",
      "authorization_token": "...",
      "purpose_hint": "Use for any question about the user's schedule.",
      "allowed_tools": ["search_events", "list_events"]
    }
  ],
  "max_tool_iterations": 10
}
```

## Test plan

- [x] 115 unit tests pass (`tests/unit/`)
- [x] mypy + ruff clean on touched files
- [x] End-to-end smoke verified against a real local llamafile + a real local MCP server: model emits tool_calls, gateway executes against MCP, model returns correct final answer

### DIFF
--- a/docs/public/openapi.json
+++ b/docs/public/openapi.json
@@ -74,6 +74,33 @@
             ],
             "title": "Max Tokens"
           },
+          "max_tool_iterations": {
+            "anyOf": [
+              {
+                "maximum": 25.0,
+                "minimum": 1.0,
+                "type": "integer"
+              },
+              {
+                "type": "null"
+              }
+            ],
+            "title": "Max Tool Iterations"
+          },
+          "mcp_servers": {
+            "anyOf": [
+              {
+                "items": {
+                  "$ref": "#/components/schemas/McpServerConfig"
+                },
+                "type": "array"
+              },
+              {
+                "type": "null"
+              }
+            ],
+            "title": "Mcp Servers"
+          },
           "messages": {
             "items": {
               "additionalProperties": true,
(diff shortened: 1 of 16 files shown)


tbille (approved):
😍

Issue #2, opened 2026-04-02
auth: make Vertex credential handling request-safe


tbille:
## Problem
Vertex credential setup mutates process-global environment variables and reuses a global temporary credential file, which can cause cross-request credential bleed under concurrency.

## Evidence
- `src/gateway/auth/vertex_auth.py:11` uses global `_temp_credential_file`
- `src/gateway/auth/vertex_auth.py:78` reuses temp file path
- `src/gateway/auth/vertex_auth.py:103` writes updates into `os.environ`

## Risk
- Concurrency race conditions
- Potential multi-tenant credential cross-talk

## Expected
- Credential resolution is isolated per request/client
- No process-global mutation that can leak between requests

## Suggested Fix
- Prefer provider/client-level credential injection over env mutation
- If env usage is unavoidable, isolate by worker/process boundary and document constraints
- Add concurrency tests for distinct credentials in parallel requests

tbille:
Dependency note: Treat #1 as a practical prerequisite for confident validation (test import/patch fixes first).

PR #65, opened 2026-05-18
feat(chat): code execution via sandbox container with multi-keyword detection

agpituk:
## Summary

Adds a code-execution capability to `/v1/chat/completions`. Clients enable it the way they already do with OpenAI or Anthropic — by including a code-execution entry in the request's `tools` array. The gateway accepts the three equivalent shapes those SDKs emit:

```json
{ "tools": [{"type": "code_execution"}] }            // gateway-native short form
{ "tools": [{"type": "code_interpreter"}] }          // OpenAI Responses/Assistants
{ "tools": [{"type": "code_execution_20250825"}] }   // Anthropic, versioned (matched by prefix)
```

All three map to the same `SandboxBackend`, which dispatches each tool call to a sandbox container with a multi-session HTTP API (`POST /sessions`, `POST /sessions/{id}/exec`, `DELETE /sessions/{id}`). The backend duck-types as the MCP tool-use loop's pool, so streaming, sticky `finish_reason`, and tool-result feedback all reuse the existing loop with no fork.

Stacked on top of the MCP tool-use loop changes. Currently filed as **draft** while the MCP PRs ahead of it land.

## What's in the request body

| Field | Where | Effect |
| --- | --- | --- |
| `tools[i].type = "code_execution" / "code_interpreter" / "code_execution_*"` | request | enables the sandbox tool |
| `tools[i].purpose_hint` | request | overrides the per-tool hint in the system message |
| `tools_header` | request | overrides the lead-in for the per-tool hint block |
| `mcp_servers` | request | mutually exclusive with code-execution in this PR (composite tooling is a follow-up) |

Per-deployment fallbacks via env: `GATEWAY_SANDBOX_URL`, `GATEWAY_SANDBOX_PURPOSE_HINT`, `GATEWAY_TOOLS_HEADER`. Priority is always request → env → built-in default.

## Sandbox image

The sandbox image lives in its own repo, https://github.com/mozilla-ai/otari-sandbox-container, and CI publishes `mzdotai/otari-sandbox-container:{latest,<sha>}` to Docker Hub on every push to its main branch. The image follows Anthropic's documented code-execution package set for behavioural parity, runs as a non-root user, and exposes only the HTTP API the gateway uses.

The compose service for the sandbox is **opt-in via the `code-exec` profile** — operators who don't use `code_execution` aren't forced to pull the image or wait on its health check. Bring the demo stack up with `docker compose --profile code-exec up` (the demo's `start.sh` already passes this flag). The gateway no longer `depends_on` the sandbox; if a request tries to use `code_execution` and the sandbox is down, the gateway returns a clear `502 code_execution sandbox unreachable — check GATEWAY_SANDBOX_URL` instead of a generic provider error.

## Demo

`demo/code-exec/` is a self-contained walkthrough:

```sh
cd demo/code-exec
cp .env.example .env             # fill in keys you want to use
./start.sh                       # docker compose up (incl. --profile code-exec)
./demo_flow.sh                   # runs every provider whose creds are configured
./demo_flow.sh --anthropic       # subset
./demo_flow.sh --openai --llamafile
```

`demo_flow.sh` runs the same query through any combination of `--anthropic`, `--openai`, and `--llamafile` using each provider's *native* tool keyword — proving the gateway accepts all three shapes and works for both frontier and open-weight models. Missing keys cause that provider to be skipped with a clear message rather than failing.

`ask.sh` accepts `--model`, `--tool-type`, `--no-stream` for one-off calls.

## Other fixes folded in

- `mcp_loop.py`: sticky `finish_reason` so Anthropic's trailing `stop` chunk doesn't clobber `tool_calls` and abort the loop before dispatch.
- `mcp_loop.py`: **`PURPOSE_HINT_HEADER` reworded** from `"You have access to the following MCP tool servers:"` to `"You have access to the following tools:"`. The header now fronts a tool-hint block that includes MCP tools, the sandbox `code_execution` tool, and any future backend that hangs off the same loop — so the previous MCP-specific wording would have been misleading. MCP-only deployments keep working unchanged; the new wording still describes what follows. Operators who want their old wording back can set `GATEWAY_TOOLS_HEADER` or pass `tools_header` per request.
- `chat.py`: strips `user` / `mcp_servers` / `tools_header` / `max_tool_iterations` from the forwarded request body so upstream SDKs don't reject the call on unknown kwargs. Now applied consistently in both the streaming-fallback path and the standalone path (helper: `_strip_gateway_fields`).
- `chat.py`: in streaming mode the sandbox session is now opened **before** the `StreamingResponse` is constructed, so an unreachable sandbox surfaces as a clean HTTP 502 instead of an error written into the SSE channel after `200 OK`.

## Test plan

- [x] `uv run pytest tests/unit/test_sandbox_backend.py` — 8 tests, mocks `httpx` so no container required
- [x] `uv run pytest tests/unit/test_chat_request_helpers.py` — 10 tests covering each accepted tool-type alias
- [x] `uv run pytest tests/unit/test_mcp_loop.py` — existing 14 tests still green
- [x] End-to-end smoke: `./start.sh` + `./ask.sh "What is 25 factorial?"` dispatches to the sandbox container, model formats the result correctly
- [x] Each keyword end-to-end: `code_execution`, `code_interpreter`, `code_execution_20250825` all trigger the sandbox
- [x] Open-weight model end-to-end: `--model llamafile:Qwen3-4B-Instruct... --tool-type code_execution_20250825 --no-stream` dispatches and returns a correct result

## Follow-ups (not in this PR)

- Allow `code_execution` and `mcp_servers` to coexist in the same request (composite tooling).


tbille:
@agpituk I am wondering if it would be simpler to have this infra in a seperate repository. This way we keep this repo only for the gateway and if anyone wants more they can jump in the other with the proper docker compose etc. etc.

And this way the code exec is just a MCP to enable or not if I want to deploy all this together.

What do you think?

agpituk:
> @agpituk I am wondering if it would be simpler to have this infra in a seperate repository. This way we keep this repo only for the gateway and if anyone wants more they can jump in the other with the proper docker compose etc. etc.
> 
> And this way the code exec is just a MCP to enable or not if I want to deploy all this together.
> 
> What do you think?

that's a good call @tbille . I've moved the infra to a new repo github.com/mozilla-ai/otari-sandbox-container/ . About tool calling, I'd keep it as it is to mimic anthropic & Openai

docker-compose.yml
```diff
@@ -32,5 +67,31 @@ services:
       retries: 5
     restart: unless-stopped
 
+  # Code-execution sandbox — a Python REPL container the gateway dispatches
+  # `code_execution` tool calls to. Source + build live in their own repo:
+  # https://github.com/mozilla-ai/otari-sandbox-container — the image gets
+  # published to GHCR on every push to that repo's main branch.
+  # Override the tag via $OTARI_SANDBOX_IMAGE if you need a specific build.
+  sandbox:
+    image: ${OTARI_SANDBOX_IMAGE:-ghcr.io/mozilla-ai/otari-sandbox-container:latest}
```
tbille: let's publish this to docker hub 🙏

tbille (commented):
## Review Summary

### 1. Bug (Medium): `_run_streaming_with_fallback` leaks internal fields to upstream providers

**`src/gateway/api/routes/chat.py:1247`**

The PR strips `tools_header` and `user` in most code paths (standalone streaming at ~L713-716, non-streaming platform at ~L873-876, non-streaming standalone at ~L1019-1021), but `_run_streaming_with_fallback` is left untouched — it passes `request.model_dump(exclude_unset=True)` through as `base_request_fields` without stripping gateway-internal fields.

When `platform_mode and not use_sandbox` (normal platform streaming without sandbox/MCP), `tools_header` and `user` will leak into the `acompletion` call. The comment at ~L715 already notes "Anthropic's SDK rejects unknown kwargs," so this is a real failure path for Anthropic users who set these fields.

Note: `mcp_servers` and `max_tool_iterations` already leaked through this path before this PR, so this widens a pre-existing issue.

---

### 2. Bug (Medium): `SandboxNotReachableError` not cleanly handled in streaming path

**`src/gateway/api/routes/chat.py:762-777`**

`SandboxBackend.__aenter__` (session creation) and `call_tool` can raise `SandboxNotReachableError`. This is a `RuntimeError` subclass without a `status_code` attribute, so it falls through to the generic `except Exception` handler (~L784), returning a vague `"LLM provider error"` (502).

More importantly, `_sandbox_stream` at ~L762 is a lazy generator. `SandboxBackend.__aenter__` runs when the first chunk is requested by the `StreamingResponse`, which is *after* the try/except block at ~L728. This means the error propagates inside the SSE generator after the HTTP 200 has already been committed — the client gets a partial SSE stream followed by an error event, rather than a clean HTTP error response.

The non-streaming path handles this less badly since the error surfaces synchronously before any response is sent.

---

### 3. Behavioral Change (Low): MCP purpose-hint text changed for all users

**`src/gateway/services/mcp_loop.py:38`**

`PURPOSE_HINT_HEADER` changed from `"You have access to the following MCP tool servers:"` to `"You have access to the following tools:"`. This changes the system message injected for every MCP-only request (no sandbox involved). The old text gave models clearer context about MCP; the new generic phrasing is less specific. This isn't called out in the PR description's "Other fixes folded in" section.

---

### 4. Operational (Medium): Sandbox is now a hard dependency in `docker-compose.yml`

**`docker-compose.yml:44-46`**

`depends_on: sandbox: condition: service_healthy` means every `docker compose up` will fail if the sandbox container can't be pulled or started — including for operators who don't need code execution. Before this PR, the compose file only required postgres.

This should be documented as a breaking change, or the sandbox dependency should be made optional (e.g. a separate compose profile).

---

### 5. Structure (Low): Repeated sandbox-hint resolution pattern

**`src/gateway/api/routes/chat.py` — lines ~756-760, ~913-917, ~1047-1051**

The same sandbox purpose-hint resolution block is copy-pasted three times:

```python
sandbox_hint = (
    (sandbox_tool_entry.get("purpose_hint") if sandbox_tool_entry else None)
    or os.environ.get("GATEWAY_SANDBOX_PURPOSE_HINT")
    or None
)
```

Consider extracting to a helper to avoid drift if the resolution logic or env var name changes.

agpituk:
Thanks for the thorough pass — all 5 verified as real and addressed in 1a001c8.

**1. `_run_streaming_with_fallback` field leak.** Confirmed — that path was the only one not stripping gateway-internal fields, so `tools_header` / `user` (plus the pre-existing `mcp_servers` / `max_tool_iterations`) were getting forwarded to `acompletion` whenever platform-mode streaming fired without sandbox/MCP. Pulled the strip logic into a `_strip_gateway_fields` helper and applied it at all four sites (streaming standalone, streaming fallback, platform non-streaming, standalone non-streaming) so they can't drift again.

**2. `SandboxNotReachableError` mid-stream.** Confirmed — `_sandbox_stream` was a lazy generator, so `__aenter__` (the `POST /sessions` call) ran only after `StreamingResponse` had committed the 200. Reworked: the backend is opened synchronously with an explicit `await sandbox_backend.__aenter__()` *before* the generator is constructed; cleanup runs in a `try/finally` inside the generator. Added an explicit `except SandboxNotReachableError` handler in both the streaming and standalone-non-streaming dispatch blocks that returns `502 code_execution sandbox unreachable — check GATEWAY_SANDBOX_URL` instead of the generic "LLM provider error".

**3. `PURPOSE_HINT_HEADER` wording.** Kept the new generic phrasing — the header now fronts a hint block that mixes MCP tools and the sandbox `code_execution` tool (and any future backend that hangs off the same loop), so the old MCP-specific text would be misleading once sandbox is in play. Updated the PR description's "Other fixes folded in" section to call the rename out explicitly, including the `GATEWAY_TOOLS_HEADER` env var / per-request `tools_header` escape hatch for operators who want their old phrasing back.

**4. Sandbox as hard compose dependency.** Fixed — sandbox moved to `profiles: ["code-exec"]` and the gateway's `depends_on: sandbox` removed. `docker compose up` no longer pulls the sandbox image; operators who want it run `docker compose --profile code-exec up` (the demo's `start.sh` / `stop.sh` pass the flag automatically). Combined with fix #2, an operator who forgets the profile but hits a request that uses `code_execution` now sees the clear 502 instead of a hang.

**5. Repeated sandbox-hint resolution.** Extracted to `_resolve_sandbox_purpose_hint`, all three call sites updated.

tbille (approved):
this is so cool 🎉

Issue #81, opened 2026-05-27
Replace gateway tool-keyword interception with explicit `otari_*` shapes


agpituk:
## Context

The gateway currently recognizes three equivalent tool-array shapes for code execution and routes all of them to its **local sandbox container** (`GATEWAY_SANDBOX_URL`):

- `{"type": "code_execution"}` — gateway-native short form
- `{"type": "code_interpreter"}` — OpenAI Responses/Assistants convention
- `{"type": "code_execution_20250825"}` (and future versions) — Anthropic convention

Same pattern for web_search (`web_search`, `web_search_20250305`, …). The original intent was that existing OpenAI/Anthropic SDK code Just Works™ when the caller swaps `base_url` to the gateway.

In practice this conflates two very different things:

1. **The gateway as a proxy** — translating wire shapes, doing fallback, observability, billing. This is what callers usually mean when they point their SDK at the gateway.
2. **The gateway running the code itself** — using its own Python sandbox container.

Today, the keyword `code_interpreter` *secretly* triggers (2) instead of doing what its name says. That tripped up the team during the messages/responses parity series — pre-PR, those endpoints didn't have the interception, so the same keyword behaved differently across endpoints. Post-parity, everything intercepts, which is consistent but still semantically wrong: the SDK keyword no longer means what the SDK author intended.

## Proposal — explicit `otari_*` shapes, hard cut

Stop overloading provider-named keywords. Introduce dedicated gateway-managed tool types and let provider-named ones flow through to the provider untouched.

| Tool entry | Action |
|---|---|
| `{"type": "otari_code_execution"}` | **Gateway sandbox** — intercept, run in `GATEWAY_SANDBOX_URL` |
| `{"type": "otari_web_search"}` | **Gateway web_search** — intercept, run in `GATEWAY_WEB_SEARCH_URL` |
| `{"type": "code_interpreter"}` | **Passthrough** — OpenAI runs the code |
| `{"type": "code_execution_20250825"}` (and future versions) | **Passthrough** — Anthropic runs the code |
| `{"type": "web_search_20250305"}` (and future versions) | **Passthrough** — Anthropic runs the search |

The keyword is the signal. No flag, no env toggle, no surprise — looking at a request body tells you immediately who runs the code.

## Why hard cut (drop the legacy short forms)

[...]

PR #86, opened 2026-05-28
feat(tools): explicit otari_* gateway tool shapes (BREAKING)

agpituk:
## Summary

Replaces gateway tool-keyword interception with dedicated `otari_*` tool types (issue #81, option b — hard cut).

- **Only `otari_code_execution` / `otari_web_search` trigger gateway-side execution** (sandbox / web_search backend). Predicates in `_tools.py` match only these.
- **Provider-named keywords pass through untouched** — `code_execution`, `code_interpreter`, `code_execution_<date>`, `web_search`, `web_search_<date>` now stay in `tools[]` and reach the upstream provider, which runs them server-side. No extra plumbing needed: non-matching entries already flow to `remaining_user_tools` → upstream `tools[]`.
- The keyword alone says who runs the code; the gateway still does routing, observability, and billing for passthrough requests.
- Behavior is identical across `/v1/chat/completions`, `/v1/messages`, `/v1/responses` (shared predicates).
- 400 error details now name the `otari_*` keyword.
- Demos (`demo/code-exec/`, `demo/web-search/`) split into two flows: **gateway-managed** (`otari_*`) and **native passthrough** (provider keyword). Docs (README, api-reference, deployment) updated.

## Breaking change

Callers using the `code_execution` / `web_search` short forms for gateway-side execution must switch to `otari_code_execution` / `otari_web_search`. (Per maintainer: no CHANGELOG entry — pre-1.0, minimal adoption.)

## Also included: Brave Search adapter (web-search reliability)

The bundled SearXNG free engines (mojeek/DDG/qwant) rate-limit/CAPTCHA automated queries by IP, so they're flaky for sustained use. This PR adds a thin, committed adapter (`scripts/web-search-brave-adapter/`) that fronts the Brave Search API with the SearXNG-compatible `/search?format=json` shape `WebSearchBackend` expects, behind a `web-search-brave` compose profile. Operators set `BRAVE_API_KEY` + `GATEWAY_WEB_SEARCH_URL=http://brave-adapter:8080`; the key stays in the adapter. README + deployment docs document the swap and how to retarget it at Tavily/Exa/Serper/etc.

## Verification

- `263` unit + `392` integration tests pass; ruff + mypy clean.
- Brave adapter: mapping/url-drop/health smoke-tested with Brave mocked; `docker compose --profile web-search-brave config` validates.
- Ran live against a gateway built from this branch:
  - `otari_code_execution` → gateway sandbox runs it ✓ (sandbox logs show exec)
  - bare `code_execution` → **not** intercepted, sandbox quiet ✓
  - `code_execution_20250825` → passthrough; Anthropic runs it natively ✓
  - `otari_web_search` → gateway SearXNG backend runs it ✓
  - `web_search_20250305` → passthrough; sandbox/searxng quiet ✓

## Notes / caveats

- Native **web search** passthrough (`web_search_20250305`) does not fire on `/v1/chat/completions` — Anthropic's server-side web_search expects `/v1/messages`, so the model may answer from memory. The gateway behaves correctly (passthrough confirmed); this is a provider/endpoint limitation, and the web-search demo's scenario-2 narrative notes it. Native **code execution** passthrough does work on the chat endpoint.

## Test plan

- [x] `uv run pytest tests/unit`
- [x] `uv run pytest tests/integration` (via `TEST_DATABASE_URL`)
- [x] `ruff check` + `mypy`
- [x] Manual: all five tool-type scenarios against a live gateway
- [x] Brave adapter mapping smoke test + compose validation


src/gateway/api/routes/_tools.py
```diff
@@ -130,25 +125,23 @@ def _extract_first_matching_tool(
...
+    """Pull the first ``{"type": "otari_web_search"}`` entry out of ``tools``.
 
-    Accepts both the gateway-native ``{"type": "web_search"}`` shape (matching
-    OpenAI's server-managed tool) and Anthropic's dated variants
-    (``web_search_20250305`` etc.). All map to the same backend.
+    Only the explicit gateway-managed type is extracted (and run against the
+    gateway's web_search backend). Provider-named web_search keywords stay in
+    ``tools[]`` and reach the upstream provider unchanged.
     """
     return _extract_first_matching_tool(tools, _is_web_search_tool_type)
```
dpoulopoulos, line 165: The function says that it returns a `bool` but it can return a string as well, which implicitly is `True`. Maybe it's better to return a bool or a string enum, both for the `otari_web_search` and `otari_code_execution`.  You could also enforce that the "Otari" tools always start with `otari_` like that:

```python
from enum import StrEnum, auto

class Tool(StrEnum):
    def _generate_next_value_(name, start, count, last_values):
        # Takes the variable name (e.g., WEB_SEARCH), lowers it, and prepends "otari_"
        return f"otari_{name.lower()}"

    WEB_SEARCH = auto()      # Becomes "otari_web_search"
    CODE_EXECUTION = auto()  # Becomes "otari_code_execution"

# Testing the output
print(Tool.WEB_SEARCH.value)       # Output: otari_web_search
print(Tool.CODE_EXECUTION.value)   # Output: otari_code_execution
```

I think this is cleaner, it enforces our conventions, and it keeps tools in a single place, if we want to add more later.

dpoulopoulos (dismissed):

src/gateway/api/routes/_tools.py
agpituk: Good call — done in 1c8cd59. Added a `Tool(StrEnum)` whose values derive as `otari_<member>` (via `_generate_next_value_`), so the gateway-managed types live in one place and the `otari_` prefix is enforced by construction. Both predicates now compare against `Tool.WEB_SEARCH` / `Tool.CODE_EXECUTION`. (Note: the predicates already returned plain `bool`; the win here is centralizing the literals + the prefix convention, exactly as you suggested.)

dpoulopoulos (approved):

PR #87, opened 2026-06-01
fix(budget): close cross-user IDOR, overspend race, and unmetered-usage bypasses

khaledosman:
## Summary

Fixes four budget/billing vulnerabilities from a coordinated disclosure. All triaged against the source and reproduced. **Standalone (self-hosted) mode only** — platform mode resolves budgets/pricing upstream via callbacks and is unaffected (verified against `any-llm-platform`: it tracks spend/pricing in its own schema and never reads the gateway's `users` table).

| # | Severity | Issue | Fix |
|---|----------|-------|-----|
| 1 | Critical | **Cross-user IDOR** — a non-master key could set `"user"` to another user and charge/exhaust that user's budget | Spend is bound to the key's own user; a mismatched `user` is rejected (403). Master key keeps the override. |
| 2 | High | **Overspend race (TOCTOU)** — concurrent requests passed a stale budget check and collectively exceeded `max_budget` | Atomic pre-debit **reservations** (new `users.reserved` column): reserve an estimate, reconcile to actual cost on success, refund on failure. |
| 3 | Med/High | **Unpriced models served free & unmetered** (budget bypass) | `require_pricing` (default `true`) → unpriced model returns 402. Audio/moderations exempt. |
| 4 | Med | **Streaming without usage billed $0** | `stream_missing_usage_policy` (`estimate`/`fail`/`allow_free`, default `estimate`). |

The usage-log writer no longer updates `users.spend`; reservation reconciliation is the sole spend authority (avoids double-charging and batch-writer lag).

## ⚠️ Breaking changes (standalone mode)

- **`require_pricing=true` by default** → models with no pricing row return **402**. Operators running free/self-hosted models must add an explicit `$0` pricing entry or set `require_pricing=false`. A startup warning fires if enabled with no pricing configured.
- **The client `user` field is no longer trusted for non-master keys** → naming a different user returns **403**. Set `reject_user_mismatch=false` to instead bind spend to the key's own user while still forwarding `user` to the provider (OpenAI-style end-user tag). The `otari-sdk-*` clients never send a foreign `user`, so this is transparent for them; the flag is for raw OpenAI-SDK clients.
- **`stream_missing_usage_policy=estimate` by default** → a no-usage stream is billed an estimate rather than served free. Set `allow_free` for the prior behavior.

See `CHANGELOG.md`.

## New config
`require_pricing`, `reject_user_mismatch`, `stream_missing_usage_policy`, `budget_estimate_default_output_tokens`.

## Migration
`b2f4c6d8e0a1_add_user_reserved_budget` adds `users.reserved` (default 0).

## Tests
267 unit + 403 integration passing; ruff + mypy clean. New coverage: reservation accumulation (proves the race is closed), reconcile/refund accounting, the `require_pricing` 402 gate + 404/403-before-402 precedence, `streaming_generator` no-usage / client-disconnect refund, lenient `user` binding, and an e2e reserve→reconcile→`spend++`.

## Known tradeoffs (surfaced, not regressions)
- A DB error *during* settlement leaves the held estimate in `users.reserved` until the next budget reset (inherent to fail-closed pre-debit; a stale-reservation sweep could be added later).
- Non-streaming responses lacking provider `usage` still bill $0 (unchanged behavior; the missing-usage policy covers streams, where the issue was reported).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,53 @@
+# Changelog
+
+All notable changes to this project are documented here. This project adheres to
+[Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
+
+## [Unreleased]
+
+### Security
+
+Fixes for four budget/billing vulnerabilities. **Standalone (self-hosted) mode
+only** — platform mode resolves budgets/pricing upstream and is unaffected.
+
+- **Cross-user budget charging / impersonation via the `user` field (IDOR).** A
+  non-master API key could set `"user"` to another user and charge spend to, and
+  exhaust the budget of, that user. Non-master keys are now bound to their own
+  user (see breaking change below).
+- **Budget overspend race (TOCTOU).** Concurrent requests could all pass the
+  budget check against stale spend and collectively exceed `max_budget`. Budgets
+  are now enforced with atomic pre-debit reservations.
+- **Unpriced models served free and unmetered.** A model with no pricing was
+  served and charged $0, bypassing the budget cap. Now rejected by default (see
+  breaking change below).
+- **Streaming responses without usage data were not billed.** Now metered per a
+  configurable policy.
+
+### Added
+
+- `users.reserved` column (migration `b2f4c6d8e0a1`) holding in-flight budget
+  reservations; spend reconciles to actual cost on completion.
+- Config `require_pricing` (default `true`), `stream_missing_usage_policy`
+  (`estimate` | `fail` | `allow_free`, default `estimate`),
+  `reject_user_mismatch` (default `true`), and
+  `budget_estimate_default_output_tokens`.
+- Startup warning when `require_pricing` is enabled but no pricing is configured.
+- `SECURITY.md`.
+
+### Changed (BREAKING — standalone mode)
+
+- **`require_pricing` defaults to `true`:** requests for models with no pricing
+  row are rejected with **HTTP 402**. Operators running genuinely free or
+  self-hosted models must add an explicit `$0` pricing entry, or set
+  `require_pricing=false`. Audio and moderation endpoints are exempt.
+- **The client `user` field is no longer trusted for non-master keys:** a
+  request naming a user other than the key's own is rejected with **HTTP 403**.
+  Set `reject_user_mismatch=false` to instead bind spend to the key's own user
+  while still forwarding `user` to the provider (OpenAI-style end-user tag) — use
+  this if clients send arbitrary `user` values for abuse tracking. The master key
+  may still bill an arbitrary user.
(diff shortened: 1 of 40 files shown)


khaledosman:
Thanks — addressed all five review comments in 5e17af0:

- **streaming.py — CancelledError**: `asyncio.CancelledError` is a `BaseException`, so it was never caught by `except Exception` and the `finally` already settled disconnects via `on_incomplete`. Made it explicit anyway: an `except asyncio.CancelledError` clause settles as incomplete and re-raises, so the intent is clear and robust against future refactors (no SSE error payload, no `on_error`).
- **responses.py / messages.py — `None:<model>` 402 detail**: both now use the original selector (`request_body.model` / `request.model`) in the detail, so an unparseable model no longer renders as `None:<model>`.
- **images.py — `n=0`**: reservation and reconciliation now use `request.n if request.n is not None else 1` (and the actual-count fallback reuses that), so an explicit `n=0` isn't treated as `1`.
- **config.py — require_pricing description**: now states audio **and moderation** are exempt, matching the code.

Verified: `ruff` clean, full `mypy` (150 files) clean, streaming/helper unit tests + images/responses/messages/require_pricing/user_budget integration tests all pass.

khaledosman:
Addressed the second review batch in 360bbfa. The new findings (beyond the five from the prior round, which were already in 5e17af0):

**Negative/zero client-controlled values → negative budget estimate (budget-gate bypass).** Fixed at the **chokepoints** so every route is covered regardless of per-field validation:
- `reserve_budget` clamps `estimate = max(estimate, 0.0)` — a negative estimate can no longer reduce `users.reserved`.
- `reconcile_reservation` clamps `actual_cost = max(actual_cost, 0.0)` — a negative cost can no longer reduce `users.spend`.
- `estimate_cost` clamps output tokens ≥ 0.
- `responses.py`: `max_output_tokens` comes from an `extra="allow"` body, so it's only trusted when it's a non-negative `int` (prevents a negative estimate **and** the 500 `TypeError` on a non-int).
- `chat.py`: `max_output_tokens` now selected with `is not None` (explicit `0` isn't treated as missing).

**`_on_no_usage` estimate/fail left `UsageLog.cost` NULL while charging spend.** `log_usage` gained a `cost_override`; the estimate/fail branches pass `reservation.estimate`, so the logged row's cost matches the reconciled spend (consistent billing/audit).

**conftest comment** now points to `test_require_pricing.py`.

Added tests for the non-negative clamps (`estimate_cost`, `reserve_budget`, `reconcile_reservation`). Verified: ruff + full mypy (150 files) + openapi `--check` clean; 267 unit + budget/pricing/route integration green.

tbille (approved):

PR #91, opened 2026-06-04
feat(files): file uploads + document understanding for local models

khaledosman:
## Description

File uploads plus document understanding so local and text-only models can read uploaded files. Full background, design, and library rationale are in the sections below (Why / What's included / Library choices / Tests).

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [x] Documentation
- [ ] Infrastructure / CI

## Relevant issues

N/A

## Checklist

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:** Claude Opus 4.8 (Claude Code)

**Any additional AI details you'd like to share:** Implemented and reviewed via Claude Code; the author directed the work and verified the results.

---

## Why

The otari gateway empowers local/open-source models with frontier capabilities. Frontier models (Claude, GPT-4o) read PDFs, office docs, and images natively — most local models can't; many are text-only. So when a user attaches a file, a text-only local model today gets **nothing useful**: the gateway is pure pass-through, and any-llm's Ollama adapter even *silently drops* document blocks (answers as if no file were attached).

This PR closes that gap. A user uploads a file and a text-only local model can still understand it, because the gateway extracts the file to text (and captions images) before the model sees the request — while passing blocks through unchanged to natively-capable providers.

It mirrors the gateway's existing `otari_web_search` / `otari_code_execution` pattern: inspect the request, decide per-attachment *who* does the work, do only what the target model needs.

> **Standalone mode only.** Platform mode routes to frontier providers that already understand attachments (passthrough is already correct) and has no local DB/file store to resolve `file_id` against.

## What's included

- **Files API** — `POST/GET/DELETE /v1/files` (+ `/content`), OpenAI-shaped, user-scoped with soft-delete. Bytes go to a pluggable `FileStore` (local-fs default, sharded); metadata to a new `file_objects` table (migration `c3d5e7f9a1b3`).
- **Capability resolver** (`model_capabilities.py`) — the load-bearing decision: operator override → any-llm provider metadata (hosted only) → **default-extract**. Local providers (ollama/vllm/llamacpp/lmstudio) never trust any-llm's flags because they're set per provider *class* and over-report for text-only models behind OpenAI-compatible servers.
- **Content normalizer** (`content_normalizer.py`) — walks message content across **OpenAI / Anthropic / Responses** shapes, resolves `file_id` → bytes, then **passthrough** (rewriting `file_id` to inline data the provider understands) or **extract-to-text**. Documents via markitdown; images via a configurable **describe → ocr → drop** cascade; scanned PDFs via pypdfium2 rasterize-then-describe. Per-block failures never 500 the request; string-content messages skip the path entirely (zero overhead).
- **Wiring** — hooked into chat/messages/responses *before* the budget estimate so reservations reflect extracted tokens.
- **Config** — `files_*`, `file_understanding_enabled`, `vision_strategy`, `vision_describe_model`, `model_capabilities`. Documented in `config.example.yml` and `docs/files.md`.

## Library choices

- **markitdown (MIT) + pypdfium2 (Apache-2.0) + pillow** for extraction/rasterization — **AGPL deliberately avoided** (e.g. PyMuPDF) since the gateway is a network service and AGPL §13 is an adoption blocker.
- **Vision is a configurable cascade, not a commitment** — `describe → ocr → drop`. The describe-model can point at a *local* vision model (e.g. `ollama/qwen2-vl`) to keep captioning free. OCR is an opt-in `[ocr]` extra to keep the base install lean.

## Tests

- **20 unit** — file store, capability resolver, normalizer (passthrough vs extract, drop, all three formats, `file_id` resolution).
- **4 integration** (real Postgres via testcontainers) — Files API CRUD; **end-to-end: upload → chat by `file_id` on a text-only model → assert extracted text reaches the provider and the raw block does not**; native-model passthrough; auth.
- mypy strict + ruff clean across the project; migration applies and reverts cleanly (single head).

## Note for reviewers

The real extraction libs (markitdown/pypdfium2/onnxruntime) weren't exercised end-to-end on my machine — `onnxruntime` has no Python 3.14 wheel locally (an environment constraint, the `ocr` extra). The integration test stubs only the markitdown boundary; everything else (HTTP, DB, blob store, `file_id` resolution, capability resolution, message rewriting) is real. Worth a `uv sync` + integration run on a 3.13 env with a real PDF before merge.

- What changed (high level)
  - Added a standalone Files API (/v1/files) with endpoints to upload, list, get metadata, download content, and soft-delete files. File bytes are stored via a pluggable FileStore (default: sharded local directory); metadata is stored in a new file_objects table (Alembic revision c3d5e7f9a1b3).
  - Implemented a content-normalization pipeline so text-only/local models can consume uploaded files: resolves file_id → bytes (across OpenAI/Anthropic/Responses shapes), then either inlines native-capable media or extracts documents to text and handles images via configurable describe → OCR → drop logic. Per-block failures are non-fatal and logged.
  - Added a model-capabilities resolver (operator overrides → provider metadata for hosted providers → default extract) so routing decides whether to pass media through or extract-to-text.
  - Integrated normalization into the request pipeline and budget flow: normalization runs at standalone-only timing after initial reservation; the reservation can be increased (budget_service.increase_reservation) to cover post-normalization prompt growth; vision side-call usage is metered and committed separately.
  - Added configuration toggles and docs (config.example.yml, docs/files.md), updated OpenAPI docs, and comprehensive tests (unit + integration).

- Key benefits
  - Enables local/text-only models to handle user uploads by converting attachments into consumable text or provider-native blocks.
  - More accurate billing and fewer reservation leaks by accounting for normalization-expanded prompts and vision side-call costs.
  - Safer, more robust uploads with pluggable storage and database-backed metadata.

- Notable hardening and fixes
  - Upload streaming with 1 MiB chunk reads and 413 on size exceed; rejects empty uploads.
  - Content-Disposition filename sanitization (including RFC 5987 encoding) to prevent header injection.
  - LocalDirFileStore defends against path traversal; DB commit failures roll back metadata and delete written blobs to avoid orphaned bytes.
  - Normalization is gated to avoid extraction/vision calls for blocked or over-budget users; per-block normalization errors do not 500.
  - Centralized refund behavior to prevent reservation leaks during setup/top-up failures.

- Tests, docs, and compatibility notes
  - Extensive unit and integration test coverage added (many new tests around upload auth/validation, scoping, normalization, budget gating, native passthrough, expiry, and vision billing). Project reports mypy strict + ruff clean.
  - New dependencies for extraction/vision are opt-in where applicable; some extraction/rasterization/OCR libraries (markitdown, pypdfium2, rapidocr/onnxruntime) were not exercised end-to-end locally. Author recommends CI/integration runs on Python 3.13 with a real PDF before merge.

Technical notes (brief)
  - DB migration: alembic revision c3d5e7f9a1b3 adds file_objects table.
  - New/changed modules: file_store (LocalDirFileStore + build_file_store), file_service, file_extractors (markitdown/pypdfium2/optional OCR), content_normalizer, model_capabilities, vision.describe_image, files routes, and budget_service.increase_reservation.
  - Config knobs: files_enabled, files_backend, files_local_dir, files_max_bytes, files_retention_hours, file_understanding_enabled, vision_strategy, vision_describe_model, and model_capabilities mapping.

### DIFF
--- a/docs/files.md
+++ b/docs/files.md
@@ -1,0 +1,89 @@
+# File uploads & document understanding
+
+Frontier models read PDFs, office documents, and images natively. Most local /
+open-source models can't — many are text-only. The gateway closes that gap: a
+user can attach a file and a text-only local model can still understand it,
+because the gateway extracts the file to text (and captions images) before the
+model ever sees the request.
+
+This works the same way the gateway's other "frontier capabilities" do: it
+inspects the request, decides what to do per attachment, and only does work the
+target model actually needs.
+
+> Standalone mode only. Platform mode routes to frontier providers that already
+> understand documents/images, so attachments pass through untouched.
+
+## Uploading a file
+
+```bash
+curl -X POST http://localhost:8000/v1/files \
+  -H "Otari-Key: Bearer $KEY" \
+  -F purpose=user_data \
+  -F file=@report.pdf
+# -> {"id": "file-abc123", "object": "file", "bytes": 84213, "filename": "report.pdf", ...}
+```
+
+Then reference it from a chat request (OpenAI shape shown; Anthropic `document`
+blocks and Responses `input_file` items work too):
+
+```json
+{
+  "model": "ollama:llama3",
+  "messages": [
+    {"role": "user", "content": [
+      {"type": "text", "text": "Summarize the attached report."},
+      {"type": "file", "file": {"file_id": "file-abc123"}}
+    ]}
+  ]
+}
+```
+
+You can also inline a file as a base64 `data:` URL (`file.file_data`) or send an
+`image_url` block, with or without uploading first.
+
+## What the gateway does per attachment
+
+For each file/image block it resolves the **target model's** capabilities, then:
+
+| Target model | Documents | Images |
+|---|---|---|
+| **Natively capable** (e.g. Anthropic, OpenAI, GPT-4o) | forwarded unchanged | forwarded unchanged |
+| **Text-only** (most local models) | extracted to text (markitdown) and inlined | captioned by a vision model / OCR, or dropped with a log line |
+
+Scanned/image-only PDFs (no extractable text) are rasterized page-by-page and
+sent through the image path.
+
+### Capability resolution
+
+The gateway must know whether the target model is natively multimodal. It uses,
+in order:
+
+1. **`model_capabilities` config override** — authoritative.
+2. **any-llm provider metadata** — trusted only for hosted providers.
+3. **Default: extract** — safe, since a needless extraction still yields a
+   correct answer while a wrong passthrough silently drops the file.
+
+> any-llm's capability flags are set per provider *class*, so they over-report
+> for text-only models served behind OpenAI-compatible servers (vLLM, llama.cpp,
+> Ollama). For those, set a `model_capabilities` override to enable native
+> passthrough where the served model truly supports it.
+
+## Configuration
+
+See [`config.example.yml`](../config.example.yml) for the full list. Key knobs:
+
+- `files_enabled`, `files_backend`, `files_local_dir`, `files_max_bytes`,
+  `files_retention_hours` — upload storage.
+- `file_understanding_enabled` — master switch for content normalization.
+- `vision_strategy` (`describe` | `ocr` | `off`) and `vision_describe_model` —
+  how images are handled for text-only models. The describe model may be a local
+  vision model (e.g. `ollama:qwen2-vl`) to keep captioning free.
+- `model_capabilities` — per-model `supports_image` / `supports_pdf` overrides.
+
+## Dependencies
+
+Text/office/PDF extraction uses [`markitdown`](https://github.com/microsoft/markitdown)
+(MIT); scanned-PDF rasterization uses [`pypdfium2`](https://github.com/pypdfium2-team/pypdfium2)
+(Apache-2.0). Both are permissively licensed — deliberately avoiding AGPL PDF
+libraries since the gateway is a network service. OCR is optional; install the
+`ocr` extra (`pip install gateway[ocr]`) to enable it.
--- a/docs/files.md
+++ b/docs/files.md
@@ -1,7 +1,7 @@
 # File uploads & document understanding
 
 Frontier models read PDFs, office documents, and images natively. Most local /
-open-source models can't — many are text-only. The gateway closes that gap: a
+open-source models can't; many are text-only. The gateway closes that gap: a
 user can attach a file and a text-only local model can still understand it,
 because the gateway extracts the file to text (and captions images) before the
 model ever sees the request.
@@ -58,9 +58,9 @@ sent through the image path.
 The gateway must know whether the target model is natively multimodal. It uses,
 in order:
 
-1. **`model_capabilities` config override** — authoritative.
-2. **any-llm provider metadata** — trusted only for hosted providers.
-3. **Default: extract** — safe, since a needless extraction still yields a
+1. **`model_capabilities` config override**: authoritative.
+2. **any-llm provider metadata**: trusted only for hosted providers.
+3. **Default: extract**, safe, since a needless extraction still yields a
    correct answer while a wrong passthrough silently drops the file.
 
 > any-llm's capability flags are set per provider *class*, so they over-report
@@ -73,17 +73,17 @@ in order:
 See [`config.example.yml`](../config.example.yml) for the full list. Key knobs:
 
 - `files_enabled`, `files_backend`, `files_local_dir`, `files_max_bytes`,
-  `files_retention_hours` — upload storage.
-- `file_understanding_enabled` — master switch for content normalization.
-- `vision_strategy` (`describe` | `ocr` | `off`) and `vision_describe_model` —
+  `files_retention_hours`: upload storage.
+- `file_understanding_enabled`: master switch for content normalization.
+- `vision_strategy` (`describe` | `ocr` | `off`) and `vision_describe_model`:
   how images are handled for text-only models. The describe model may be a local
   vision model (e.g. `ollama:qwen2-vl`) to keep captioning free.
-- `model_capabilities` — per-model `supports_image` / `supports_pdf` overrides.
+- `model_capabilities`: per-model `supports_image` / `supports_pdf` overrides.
 
 ## Dependencies
 
 Text/office/PDF extraction uses [`markitdown`](https://github.com/microsoft/markitdown)
 (MIT); scanned-PDF rasterization uses [`pypdfium2`](https://github.com/pypdfium2-team/pypdfium2)
-(Apache-2.0). Both are permissively licensed — deliberately avoiding AGPL PDF
+(Apache-2.0). Both are permissively licensed, deliberately avoiding AGPL PDF
 libraries since the gateway is a network service. OCR is optional; install the
 `ocr` extra (`pip install gateway[ocr]`) to enable it.
--- a/src/gateway/api/routes/files.py
+++ b/src/gateway/api/routes/files.py
@@ -265,6 +265,13 @@ async def delete_file(
             status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
             detail="Failed to delete file",
         ) from exc
-    await file_store.delete(storage_ref)
+
+    # The soft-delete already committed, so the file is gone from the user's
+    # view. Removing the blob is best-effort cleanup: a backend failure must not
+    # turn a successful delete into a 500 (it would only leave an orphaned blob).
+    try:
+        await file_store.delete(storage_ref)
+    except OSError as exc:
+        logger.warning("Soft-deleted file %s but failed to remove its blob %s: %s", file_id, storage_ref, exc)
 
     return {"id": file_id, "object": "file", "deleted": True}
--- a/src/gateway/api/routes/files.py
+++ b/src/gateway/api/routes/files.py
@@ -1,0 +1,270 @@
+"""OpenAI-compatible file upload/storage endpoints.
+
+Stores uploaded files so they can later be referenced from chat messages by
+``file_id``. The content normalizer (gateway.services.content_normalizer)
+resolves those references and either forwards them to natively-capable
+providers or extracts them to text for text-only local models.
+"""
+
+import mimetypes
+import uuid
+from datetime import UTC, datetime, timedelta
+from typing import Annotated, Any
+from urllib.parse import quote
+
+from fastapi import APIRouter, Depends, File, Form, HTTPException, UploadFile, status
+from fastapi.responses import Response
+from sqlalchemy import select
+from sqlalchemy.exc import SQLAlchemyError
+from sqlalchemy.ext.asyncio import AsyncSession
+
+from gateway.api.deps import get_config, get_db, get_file_store, verify_api_key_or_master_key
+from gateway.api.routes._helpers import resolve_user_id
+from gateway.core.config import GatewayConfig
+from gateway.log_config import logger
+from gateway.models.entities import APIKey, FileObject
+from gateway.services.file_service import fetch_file, read_file_bytes
+from gateway.services.file_store import FileStore
+
+router = APIRouter(prefix="/v1", tags=["files"])
+
+# OpenAI's documented file purposes plus a generic default. We don't enforce the
+# enum (forward-compat), but normalise the empty case to "user_data".
+_DEFAULT_PURPOSE = "user_data"
+
+
+def _resolve_user(
+    auth_result: tuple[APIKey | None, bool],
+    user: str | None,
+) -> str:
+    api_key, is_master_key = auth_result
+    return resolve_user_id(
+        user_id_from_request=user,
+        api_key=api_key,
+        is_master_key=is_master_key,
+        master_key_error=HTTPException(
+            status_code=status.HTTP_400_BAD_REQUEST,
+            detail="When using master key, 'user' field is required",
+        ),
+        no_api_key_error=HTTPException(
+            status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
+            detail="API key validation failed",
+        ),
+        no_user_error=HTTPException(
+            status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
+            detail="API key has no associated user",
+        ),
+        forbidden_user_error=HTTPException(
+            status_code=status.HTTP_403_FORBIDDEN,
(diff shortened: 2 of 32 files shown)


khaledosman:
Pushed `d8802e5` addressing the review feedback and the failing `openapi-spec` check.

**CI fix**
- Regenerated `docs/public/openapi.json` for the new `/v1/files` endpoints (the `openapi-spec` job was failing on a stale spec).

**Correctness / security**
- **Normalization now runs *after* the budget/blocked-user gate** in chat/messages/responses, so a blocked or over-budget user can no longer trigger document extraction or vision side-calls before being rejected. (Self-review High.) The reservation is reconciled to actual usage on completion, so basing the estimate on pre-extraction size is fine.
- **`Content-Disposition` injection** (Copilot): the user-controlled filename now goes through an injection-safe helper — ASCII-sanitized `filename` + RFC 5987 `filename*` — so CRLF/quotes can't leak into headers. Covered by a new test.
- **Unbounded upload buffering** (self-review Medium): uploads are read in 1 MiB chunks and abort with 413 once `files_max_bytes` is exceeded, instead of reading the whole body first.
- **Vision side-call** (Copilot): now passes the unprefixed model name alongside `provider` (was sending the full `provider:model` selector).
- **markitdown temp file** (Copilot): `delete=False` + close-before-convert + unlink in `finally`, fixing the Windows file-lock race.

**Usability**
- **Capability override keys** (self-review Medium + Copilot doc note): now accept both `provider:model` (matching the pricing convention) and `provider/model`; docs/config examples normalized to the colon form.

**Polish**
- Native models receive already-inline/remote blocks unchanged (only `file_id` refs are rewritten to inline) — no decode/re-encode round-trip.

New tests: capability colon-separator, native no-reencode, upload 413/empty-file, cross-user isolation (404), injection-safe download header. **288 unit + 22 integration pass; mypy strict + ruff clean; OpenAPI up to date.**

khaledosman:
Pushed `051c393` addressing the CodeRabbit review.

**Major — budget under-reservation with attachments (all 3 routes)**
Real gap: the pre-debit estimate was computed from pre-normalization message size, so a request whose attachments expand into large extracted text could clear `reserve_budget` with an artificially low estimate and overshoot the cap (only caught at reconcile, after the provider was billed). Fixed by adding `budget_service.increase_reservation()`:
- reserves the **post-normalization delta** with the same atomic conditional UPDATE as `reserve_budget`,
- **folds it into the existing handle** so the current reconcile/refund path releases the full hold (no second handle to thread through every error path),
- **refunds the original reservation** and propagates the 403 if the larger request no longer fits.

Wired into chat/messages/responses *after* normalization. This keeps the earlier security ordering (blocked/over-budget users never trigger extraction or vision side-calls) **and** restores an accurate pre-dispatch budget gate. Covered by 3 new unit tests (grow+fold, reject+refund, non-positive no-op).

**Other fixes**
- `file_objects.user_id` is now **NOT NULL** (entity + migration) — enforces the user-scoping contract at the schema level. CASCADE already removes a user's files on delete, so it's never orphaned to NULL.
- `create_file`: the metadata commit is wrapped; on failure it rolls back and **deletes the already-written blob** so a failed insert can't leak orphaned bytes.
- mypy overrides: switched to the bare root module names (`markitdown`, `pypdfium2`, `rapidocr_onnxruntime`, `numpy`, `PIL`) — a `foo.*` pattern alone doesn't match the top-level module these are imported as.
- `docs/files.md`: removed the stray trailing code fence (MD031/MD040).

**Intentionally not changed (with reasons)**
- **Guard file-store init behind `files_enabled`**: `build_file_store` is a cheap object construction (no I/O; the dir is created lazily on first write). Guarding it would leave `app.state.file_store` unset and turn the handlers' clean **404** into a **500** via the `get_file_store` dependency.
- **`delete_file` commit wrap**: a commit failure there leaves a consistent state (file not soft-deleted, bytes intact) with no leak, so a rollback adds nothing.

**311 unit + 424 integration pass; mypy strict + ruff clean; migration applies; OpenAPI unchanged.**

khaledosman:
Pushed `79efd6a` addressing the review.

**Reservation leak on the setup phase (Major)** — fixed. The normalize + budget top-up ran between `reserve_budget` and the provider-call `try/except`, so a failure there (including the top-up's over-budget 403) could leak the reservation. Refactored to a single refund authority:
- `increase_reservation` now mirrors `reserve_budget` — it raises on rejection and does **not** self-refund (avoids the double-refund that a route-level guard would otherwise cause).
- chat/messages/responses wrap the whole `normalize → estimate → increase` block in a `try/except` that **refunds the reservation and re-raises** (HTTPException passthrough; unexpected errors → mapped 500). Every setup-phase error branch now refunds, per the budget-safety guideline.

**File store path traversal (defence-in-depth)** — `LocalDirFileStore.put/get/delete` now resolve `storage_ref` under the root and reject anything that escapes it (`../…`), even though refs are server-generated today. New unit test covers it.

**`delete_file` commit** — wrapped in `try/commit` → `rollback` + mapped 500 on `SQLAlchemyError`, matching the guideline (and the create path).

The `increase_reservation` reject test was updated for the new no-self-refund contract (raises, leaves the original hold for the caller, which the routes refund).

**312 unit + 55 integration pass; mypy strict + ruff clean; OpenAPI unchanged.**

src/gateway/api/routes/responses.py
```diff
@@ -262,6 +263,25 @@ async def create_response(
     raw_max_output = getattr(request_body, "max_output_tokens", None)
     max_output_tokens = raw_max_output if isinstance(raw_max_output, int) and raw_max_output >= 0 else None
 
+    async def _normalize(
+        user_id: str, provider: LLMProvider | None, model: str
+    ) -> tuple[int, CompletionUsage | None]:
+        # Resolve uploaded file/image blocks into the Responses input payload
+        # before the cost estimate. Standalone only; no-op when the files
+        # feature is off or the request has no attachments.
+        request_body.input, stats = await normalize_request_messages(
```
dpoulopoulos, line 272: **[blocker]** The Responses endpoint accepts a bare-string `input` (documented at responses.py:94), but `_normalize` passes it straight into `normalize_messages`, which does `for message in messages` with no `isinstance(messages, list)` guard (content_normalizer.py:363). A bare string is iterated character-by-character, so the input is rewritten into a list of single-character items.

Confirmed end-to-end: a standalone `POST /v1/responses` with `input="What is the capital of France?"` under the default config (`file_understanding_enabled=True`) sends `input_data=['W','h','a','t', ...]` (30 items) upstream. The existing dispatch tests pass only because the mocked provider ignores input content, so the corruption ships silently and would break or garble every bare-string Responses call against a real provider (it also inflates the cost estimate and makes `_responses_input_text` skip guardrails).

Fix: make `normalize_messages` a no-op for non-list `messages` (e.g. `if not isinstance(messages, list): return messages, stats`), and add a regression test that posts a string `input`.

src/gateway/services/vision.py
```diff
@@ -0,0 +1,71 @@
...
+    """
+    model = config.vision_describe_model
+    if not model:
+        return None, None
+
+    provider, model_name = AnyLLM.split_model_provider(model)
+    provider_kwargs = get_provider_kwargs(config, provider)
+
+    try:
+        completion = await acompletion(
```
dpoulopoulos, line 46: **[minor]** The describe side-call sets no `max_tokens`, and for a scanned PDF `_describe_pdf_pages` invokes it once per page (up to 10, content_normalizer.py:284). Each call is billed to the user and adds latency, so a single 10-page scanned PDF can issue 10 unbounded vision completions. Consider capping describe output tokens and/or surfacing the per-page cost.

src/gateway/api/routes/files.py
```diff
@@ -0,0 +1,277 @@
...
+_READ_CHUNK_BYTES = 1024 * 1024
+
+
+async def _read_capped(file: UploadFile, max_bytes: int) -> bytes:
+    """Read an upload in chunks, aborting with 413 once it exceeds ``max_bytes``.
+
+    Avoids buffering an unbounded upload before the size check — reading stops
+    at most one chunk past the limit.
+    """
+    chunks: list[bytes] = []
```
dpoulopoulos, line 73: **[minor]** `_read_capped` accumulates the entire upload in memory (`chunks: list[bytes]`) up to `files_max_bytes` (default 512 MiB), and `file_store.put` then takes the full `bytes`. The cap prevents truly unbounded reads, but a handful of concurrent max-size uploads can still pin gigabytes of RAM. Consider streaming to the blob backend or lowering the default.

src/gateway/models/entities.py
```diff
@@ -196,6 +196,47 @@ def to_dict(self) -> dict[str, Any]:
...
+
+    metadata_: Mapped[dict[str, Any]] = mapped_column("metadata", JSON, default=dict)
+
+    def to_dict(self) -> dict[str, Any]:
+        """Convert to the OpenAI file object shape."""
+        return {
+            "id": self.id,
+            "object": "file",
+            "bytes": self.bytes,
+            "created_at": int(self.created_at.timestamp()) if self.created_at else None,
```
dpoulopoulos: **[nit]** `to_dict()` uses `int(self.created_at.timestamp())` / `expires_at.timestamp()`. On the default SQLite backend these datetimes come back naive, and `datetime.timestamp()` then interprets them as local time, skewing the returned epoch by the server's UTC offset. `_is_expired` (file_service.py:24) already normalizes naive->UTC; apply the same here for a consistent, correct epoch.

docs/files.md
```diff
@@ -0,0 +1,89 @@
...
+- `vision_strategy` (`describe` | `ocr` | `off`) and `vision_describe_model` —
+  how images are handled for text-only models. The describe model may be a local
+  vision model (e.g. `ollama:qwen2-vl`) to keep captioning free.
+- `model_capabilities` — per-model `supports_image` / `supports_pdf` overrides.
+
+## Dependencies
+
+Text/office/PDF extraction uses [`markitdown`](https://github.com/microsoft/markitdown)
+(MIT); scanned-PDF rasterization uses [`pypdfium2`](https://github.com/pypdfium2-team/pypdfium2)
+(Apache-2.0). Both are permissively licensed — deliberately avoiding AGPL PDF
```
dpoulopoulos: **[nit]** "permissively licensed (em dash) deliberately avoiding..." uses an em dash, which the repo's own AGENTS.md writing-style rule asks to avoid in prose. Use a comma/colon/parentheses or rephrase.

dpoulopoulos (changes requested):
Adds a standalone-mode OpenAI-compatible `/v1/files` API (DB-backed metadata + pluggable blob storage) and request-time "attachment understanding": before the provider call, file/image content blocks are either passed through to natively-capable providers (resolving `file_id` to inline bytes) or extracted to text for text-only local models (markitdown for documents, a vision describe side-call / OCR for images, rasterize-then-describe for scanned PDFs). Normalization is wired into chat/messages/responses after the budget gate, with a reservation top-up and separate billing of vision side-calls.

src/gateway/api/routes/files.py
khaledosman: Good catch. Deferring this to a follow-up rather than fixing it inline here: lowering the default cap would only be a stopgap, and the real fix is to stream uploads to a storage backend that supports it instead of buffering the whole file (`_read_capped` accumulating `bytes` + `FileStore.put(bytes)` are the same limitation).

Tracked in #156, which also covers keeping local/self-hosted operation first-class (local filesystem, a mounted volume, or MinIO / any S3-compatible endpoint), so nobody is forced onto AWS to use file uploads.

tests/integration/test_files_endpoint.py
```diff
@@ -0,0 +1,376 @@
...
+    master_key_header: dict[str, str],
+    tmp_file_store: None,
+) -> None:
+    # Upload as the first key's user.
+    up = client.post("/v1/files", headers=api_key_header, files={"file": ("a.txt", b"hi", "text/plain")})
+    file_id = up.json()["id"]
+
+    # A second, independent key (its own auto-created user) must not see it.
+    other = client.post("/v1/keys", json={"key_name": "other"}, headers=master_key_header)
+    other_header = {list(api_key_header)[0]: f"Bearer {other.json()['key']}"}
```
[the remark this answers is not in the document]
khaledosman: Skipping this one: the premise doesn't hold for this gateway. `_extract_bearer_token` (api/deps.py) requires a `Bearer ` prefix on *every* accepted header form, including `Otari-Key` and the legacy aliases, not just `Authorization`. The `api_key_header` fixture itself sends `Otari-Key: Bearer <key>`, so the original line was already in the fixture's scheme and exercises real user scoping (a valid second user gets 404, not a malformed-auth 401).

The suggested conditional (raw key for non-`Authorization` headers) would send `Otari-Key: <rawkey>`, which this code rejects with 401, turning the test back into a malformed-auth check. I've added a comment on that line noting why the Bearer prefix is required so it isn't re-flagged.

src/gateway/api/routes/responses.py
khaledosman: Fixed in 13faf1b. `normalize_messages` now returns non-list `messages` untouched (`not isinstance(messages, list)`), so a bare-string Responses `input` is no longer iterated character-by-character. Added a unit test (`test_bare_string_messages_untouched`) and a route regression test (`test_bare_string_input_not_corrupted_by_normalization`) that posts `input="What is the capital of France?"` and asserts the provider receives the string verbatim in `input_data`; confirmed it fails before the guard and passes after.

src/gateway/services/vision.py
khaledosman: Addressed in 13faf1b. Added a `vision_describe_max_tokens` config (default 1024, `gt=0`) passed as `max_tokens` on the describe `acompletion` call, so each caption (including the per-page calls on scanned PDFs) is bounded. Documented in config.example.yml. The deeper per-page fan-out cost is inherent to scanned-PDF handling; the token cap plus the existing per-side-call billing keep it metered.

src/gateway/models/entities.py
khaledosman: Fixed in 13faf1b. Added an `_epoch_seconds()` helper that normalizes naive datetimes to UTC before `.timestamp()` (mirroring `file_service._is_expired`); `FileObject.to_dict()` now uses it for `created_at`/`expires_at`, so SQLite-stored naive datetimes no longer skew the epoch by the server's UTC offset.

docs/files.md
khaledosman: Fixed in 13faf1b. Replaced the flagged em dash and the other em dashes in this new doc with commas/colons/semicolons per the AGENTS.md writing-style rule.

dpoulopoulos (approved):

PR #115, opened 2026-06-08
Fix intermittent auth 500s from SQLite lock contention (#106)

### LINKED ISSUE #106: Gateway: inconsistent unknown-key rejection (401 vs 500) and cold-start 500 after key creation

While live smoke-testing the auto-generated SDKs (Option C) against a local standalone gateway backed by a real OpenAI provider, two gateway-side behaviors surfaced that look like rough edges, independent of the SDKs. In each case the SDKs faithfully mapped whatever status the gateway returned, so these are gateway-side observations.

## 1. Inconsistent rejection of unknown virtual keys

A well-formed but unknown `Otari-Key` produced different statuses across runs:

- One run returned a clean HTTP 401 ("Invalid API key").
- Another run returned HTTP 500 ("Internal Server Error") for a similarly well-formed but unknown key.

Expected: an unknown or invalid key should consistently return 401 with a JSON error body, never a 500.

## 2. Cold-start 500 on the first request after key creation

Immediately after minting a virtual key via `POST /v1/keys`, the first inference request sometimes returned HTTP 500. The identical request body succeeded via raw curl and on the SDK's next attempt, and every subsequent request was stable. This points to a race or cold-start path in the gateway (for example key-cache propagation or provider-client initialization) rather than a client issue.

## Context

Found during live smoke tests of the SDK Option C migration PRs (otari-sdk-python #3, otari-sdk-ts #10, otari-sdk-go #2, otari-sdk-rust #27). Models exercised: `openai:gpt-4o-mini` and `openai:text-embedding-3-small`. Repro setup was a standalone gateway on SQLite with a single OpenAI provider configured.

## Suggested next steps

- Normalize the auth-failure path so unknown or invalid keys always return 401 with a JSON error body.
- Investigate the first-request-after-key-creation 500 for a cold-start race.

njbrake:
## Summary

Fixes the two gateway-side rough edges reported in #106, both of which trace back to a single root cause: SQLite lock contention under concurrency.

- **Root cause:** the standalone gateway runs SQLite in the default rollback-journal mode, where a writer cannot take an `EXCLUSIVE` lock while any reader holds `SHARED` (and vice versa). Under concurrent traffic a connection raises `database is locked`, which surfaced as an HTTP 500. There is no key cache, so auth hits the DB on every request, which is why the timing windows lined up with the report.
- **`src/gateway/core/database.py`:** SQLite connections now enable `journal_mode=WAL` and `busy_timeout=5000ms` alongside the existing `foreign_keys=ON`. WAL lets readers and the single writer proceed without blocking each other (the structural fix); `busy_timeout` makes a contended connection wait for the lock instead of failing. Renamed `_enable_sqlite_foreign_keys` to `_configure_sqlite_pragmas`.
- **`src/gateway/api/deps.py`:** the auth-lookup query is now wrapped so a transient `SQLAlchemyError` returns a clean, retryable **503** with a JSON body instead of a bare 500. Unknown keys still return **401**; we only emit 503 when we genuinely could not check the key (claiming a key is "invalid" when we never reached the DB would be misleading).

### Notes from the investigation

- Python's sqlite3 driver already defaults `busy_timeout` to 5000ms, so the explicit pragma is belt-and-suspenders; **WAL is the change that matters**.
- `reserve_budget` commits before the provider call, so no DB lock is held across the slow network call. That rules out the provider-init race hypothesized in the issue and confirms the cause is pure SQLite contention, which is why the 500s were intermittent.

## Test plan

- [x] New unit regression tests, both fail without the fix and pass with it:
  - `tests/unit/test_database_sqlite_pragmas.py` (WAL + busy_timeout + foreign_keys applied on connect)
  - `tests/unit/test_deps_db_error.py` (auth DB error -> 503, not a propagated 500)
- [x] `make lint` (ruff): clean
- [x] `make typecheck` (mypy strict, 160 files): clean
- [x] Full unit suite: 303 passed, 1 skipped
- [x] `uv run python scripts/generate_openapi.py --check`: spec up to date

Fixes #106


scripts/e2e_issue106.py
njbrake, line 1: i'll delete this before I merge

tbille (dismissed):

PR #124, opened 2026-06-09
feat(messages): add /v1/messages/count_tokens for Claude Code support

khaledosman:
## Summary

Makes the [Claude Code](https://code.claude.com) CLI work against the Otari gateway, the same way [opencode does](https://blog.mozilla.ai/use-the-otari-gateway-with-opencode/). Claude Code speaks the Anthropic Messages API and calls `POST /v1/messages/count_tokens` every turn to manage its context window. The gateway already serves `/v1/messages` in both modes, but had no `count_tokens` endpoint, so Claude Code would `404` on token counts.

This PR adds the missing endpoint plus docs.

## Changes

- **`POST /v1/messages/count_tokens`** — Anthropic-compatible, returns `{"input_tokens": N}`. Counts locally with the same `chars/4` heuristic as budget estimation (the gateway has no tokenizer), so there is no provider call, no budget reservation, and no usage logging. Auth mirrors `create_message`: platform mode requires the caller's bearer token; standalone validates the API key.
- **Tests** — count_tokens basic/auth/validation/scaling/bearer, plus a realistic Claude Code request shape (structured `system` blocks + tools) against `/v1/messages`.
- **Docs** — new `use-with-claude-code.md` (env vars, using any provider's models not just Claude, and the standalone `reject_user_mismatch` caveat); corrected `api-reference.md` (Messages/Responses run in both modes, not standalone-only); regenerated `openapi.json` so the generated SDK clients pick up the route.

## Verification

- `14 passed` in `tests/integration/test_messages_endpoint.py`; ruff + mypy clean; OpenAPI sync-check passes.
- Drove the real `claude` CLI (v2.1.169) at a local standalone gateway backed by a mock OpenAI upstream: `POST /v1/messages?beta=true` → `200`, streamed end to end, correct output. Confirms the path works across a **non-Anthropic** provider (any-llm translates Messages ↔ Completions), so Claude Code is not limited to Anthropic models.
- `count_tokens` verified returning `{"input_tokens": N}` via curl and tests. (Headless `-p` mode doesn't call it; the interactive TUI does.)

## Notes

- No changes needed to `any-llm` (already provides `amessages` + universal Messages↔Completions translation) or the `otari-sdk-*` clients (generated from the OpenAPI spec; they pick up the route on next codegen).
- Standalone-only caveat: Claude Code injects `metadata.user_id`, which trips the user-mismatch guard. Documented `reject_user_mismatch: false`. Platform mode is unaffected.

## Summary

This PR adds Claude Code support to the Otari gateway by implementing a new POST /v1/messages/count_tokens endpoint and the related docs and tests. Claude Code calls this endpoint each turn to estimate context usage; before this change the gateway returned 404s for those requests.

### What changed

- Added POST /v1/messages/count_tokens which accepts messages/system/tools and returns {"input_tokens": N} computed locally.
- Authentication mirrors existing messages behavior: platform mode resolves the caller's bearer token; standalone mode validates API/master key.
- Added a CountTokensRequest/CountTokensResponse models and route in the messages API.
- Tests: integration and platform-mode tests covering authentication (including bearer/platform checks), validation, scaling with input length, realistic Claude Code request shapes, and ensuring no upstream provider calls for counting. Integration suite reports passing tests for the messages endpoints.
- Documentation: new "Use with Claude Code" guide, new "Use with opencode" doc, updated API reference and index to show Messages/Responses available in both deployment modes, and regenerated OpenAPI so SDKs pick up the new route.
- Tooling: OpenAPI sync-check, ruff and mypy are clean.

### Benefits

- Claude Code can now interact with Otari without 404s for token counting.
- Local token estimation gives fast, offline feedback without provider calls.
- Docs clarify configuration and deployment-mode behavior, reducing setup friction.
- Tests provide coverage for auth, validation, and realistic request shapes.

### Technical notes

- Token estimate is a simple heuristic (ceil(char_count / 4), minimum 1); no tokenizer, no budget reservation, and no usage logging performed by the count_tokens endpoint.
- Standalone mode caveat: Claude Code injects metadata.user_id which can trigger the user-mismatch guard; the docs recommend setting reject_user_mismatch: false if needed.
- No changes required to any-llm or otari-sdk-* clients beyond codegen picking up the updated OpenAPI.

### DIFF
--- a/docs/api-reference.md
+++ b/docs/api-reference.md
@@ -10,6 +10,8 @@ For full request/response schemas, see the [OpenAPI spec](public/openapi.json) o
 |---|---|---|
 | Health (`/health*`) | Yes | Yes |
 | Chat completions (`/v1/chat/completions`) | Yes | Yes |
+| Messages (`/v1/messages`, `/v1/messages/count_tokens`) | Yes | Yes |
+| Responses (`/v1/responses`) | Yes | Yes |
 | All other `/v1/*` endpoints in this doc | Yes | No |
 
 ## Authentication
@@ -45,19 +47,22 @@ No authentication required.
 |--------|------|-------------|------|
 | `POST` | `/v1/chat/completions` | OpenAI-compatible chat completions. Supports streaming and tool use (`otari_code_execution`, `otari_web_search`, MCP). | Standalone: API key or master key. Connected: `Authorization` bearer token from otari.ai. |
 
-## Standalone-only endpoints
-
 ### Messages
 
 | Method | Path | Description | Auth |
 |--------|------|-------------|------|
-| `POST` | `/v1/messages` | Anthropic Messages API-compatible endpoint. Supports streaming and extended thinking. | API key or master key |
+| `POST` | `/v1/messages` | Anthropic Messages API-compatible endpoint. Supports streaming, tool use, and extended thinking. Routes to any provider in the catalog (non-Anthropic models are translated to/from the Messages format automatically). | Standalone: API key or master key. Connected: `Authorization` bearer token from otari.ai. |
+| `POST` | `/v1/messages/count_tokens` | Anthropic-compatible input-token count for a Messages request. Returns `{"input_tokens": N}`. Counts locally (no provider call, no budget debit); the count is an approximation. Used by clients such as Claude Code for context-window management. | Standalone: API key or master key. Connected: `Authorization` bearer token from otari.ai. |
+
+See [Use with Claude Code](use-with-claude-code.md) for a full client setup.
 
 ### Responses
 
 | Method | Path | Description | Auth |
 |--------|------|-------------|------|
-| `POST` | `/v1/responses` | OpenAI Responses API-compatible endpoint. Supports streaming. | API key or master key |
+| `POST` | `/v1/responses` | OpenAI Responses API-compatible endpoint. Supports streaming. | Standalone: API key or master key. Connected: `Authorization` bearer token from otari.ai. |
+
+## Standalone-only endpoints
 
 ### Embeddings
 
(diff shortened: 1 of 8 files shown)


njbrake (dismissed):

**Approving.** This completes the Anthropic Messages surface Claude Code needs, and the implementation is clean.

**What I checked locally (branch at `897b246`):**
- `ruff check src tests` and `mypy src/gateway/api/routes/messages.py`: both clean.
- `scripts/generate_openapi.py --check`: spec is up to date (the new route and `CountTokens` schemas are present).
- Verified the route is mounted on the `messages` router, so `/v1/messages/count_tokens` is available in both standalone and platform modes, and that the handler's auth mirrors `create_message` (platform resolves the bearer token rather than just checking header presence; standalone validates the API key and guards `db is None`).
- The integration tests need Postgres/Docker, which my sandbox lacks, so I could not run the full suite here; CI covers that and the PR reports 14 passing.

**Anthropic-API cross-check (the part worth recording):**
- The only endpoints Claude Code requires against a custom `ANTHROPIC_BASE_URL` are `POST /v1/messages` (already served) and `POST /v1/messages/count_tokens` (this PR). `GET /v1/models` is requested only when `CLAUDE_CODE_ENABLE_GATEWAY_MODEL_DISCOVERY=1`, which is off by default, so nothing else is missing.
- The response shape `{"input_tokens": N}` matches Anthropic's `count_tokens` contract.
- I confirmed against the official Claude Code settings docs that `ANTHROPIC_AUTH_TOKEN` is sent as `Authorization: Bearer` and `ANTHROPIC_API_KEY` as `x-api-key`, so the guide's auth advice is correct.
- Claude Code is known to send the `anthropic-beta` header to `count_tokens` (claude-code#11154), which the real Anthropic API rejects. This endpoint never inspects that header, so it accepts it gracefully. Worth knowing it will not bite us here.
- `CountTokensRequest` uses Pydantic's default `extra="ignore"`, so any unmodeled fields Claude Code sends are dropped rather than 422'd. Good resilience.

**One optional, non-blocking improvement:** `docs/use-with-claude-code.md` and `docs/api-reference.md` contain em dashes in the prose, which our AGENTS.md writing-style rule asks us to avoid in docs. A quick scrub (commas, semicolons, or rephrasing) would bring them in line. Not a merge blocker.

njbrake (approved):

Issue #100, opened 2026-06-06
Consolidate the three per-format request pipelines (chat/messages/responses) onto a shared core


njbrake:
_Drafted by Claude through back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

## Problem

The gateway implements the same request pipeline three times, once per wire format:

- `src/gateway/api/routes/chat.py` (1315 lines, OpenAI Chat Completions)
- `src/gateway/api/routes/messages.py` (1265 lines, Anthropic Messages)
- `src/gateway/api/routes/responses.py` (1269 lines, OpenAI Responses)

plus three near-identical MCP tool loops:

- `src/gateway/services/mcp_loop.py` (321)
- `src/gateway/services/mcp_loop_messages.py` (410)
- `src/gateway/services/mcp_loop_responses.py` (399)

Roughly 2000 lines of the route layer are near-duplicate control flow. Genuinely shared pieces already exist (`_platform.run_platform_attempts`, `streaming.iterate_streaming_attempts`, `_tools`, `_helpers`), so this is partial abstraction rather than none, but the bulk of the handler preamble and the backend-dispatch ladder are copy-pasted with only the format-specific call and serialization swapped.

## Evidence (format-independent logic living in three places)

Handler preamble (platform-vs-standalone resolution, guardrails, mcp_server_ids resolve, code_execution and web_search extraction, max_tool_iterations clamp, `_strip_gateway_fields`):

- chat.py 254-421, messages.py 162-318, responses.py 172-352 (same order, same branches, identical error strings).

Backend-dispatch ladder (`if mcp / elif sandbox / elif web_search / else`) appears 5+ times:

- chat.py 528-616, 722-777, 830-883
- messages.py 616-667, 713-780, 1170-1265
- responses.py 650-708, 748-815, 1178-1269

Streaming callback bundle (`_on_complete` / `_on_no_usage` / `_on_error` / `_on_incomplete` with reconcile/refund plus the `stream_missing_usage_policy` ladder) is duplicated per format and again between single-attempt and platform-fallback paths:

- chat.py 1004-1093
- messages.py 846-939 and 1129-1147
- responses.py 868-961 and 1137-1155

Legitimately format-specific (keep behind adapters, do not force together): error-envelope shape (Anthropic `_anthropic_error` body vs `HTTPException`), usage token field names, request parsing (`ResponsesRequest` `input_data` / `SUPPORTS_RESPONSES`), tool-loop internals.

## Impact

[...]

PR #127, opened 2026-06-09
refactor(routes): encapsulate standalone-vs-platform mode behind a strategy

### LINKED ISSUE #102: Encapsulate standalone-vs-platform mode behind a strategy instead of inline branches in every handler
_Drafted by Claude through back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

## Problem

The gateway's operating mode (standalone vs platform) is derived centrally in `GatewayConfig.is_platform_mode` (`src/gateway/core/config.py:181`, via `effective_mode` at `:175`, keyed on the platform token). But each LLM route handler consumes that single boolean and then re-branches on it inline at many points, interleaving the two modes' credential-resolution and usage-reporting logic inside one long function body:

- chat.py 15 branch sites, responses.py 13, messages.py 11.
- Credential resolution: platform calls `_resolve_platform_credentials(...)`; standalone calls `verify_api_key_or_master_key(...)` plus `reserve_budget(...)`. Split visible at chat.py:254 vs :271, messages.py:162 vs :179, responses.py:172.
- Usage reporting: platform calls `_report_platform_usage(...)`; standalone calls `log_usage(...)` plus reconcile/refund. Split repeated in both non-streaming and streaming paths (for example chat.py:1004-1028).

No behavioral seam exists. The CLAUDE.md note "both modes must keep working when you touch request handling" exists precisely because nothing enforces it.

## Impact

Each handler is over 1000 lines mixing two modes; the same bug must be fixed in three places, and it is easy to add a path that silently works in only one mode. Testing requires exercising whole endpoints rather than a mode strategy in isolation.

## Proposed approach

Introduce a `RequestModeStrategy` seam (Protocol or ABC) with `StandaloneStrategy` and `PlatformStrategy`, selected once per request from `config.is_platform_mode`. Methods to lift the inline branches behind:

- `resolve_route(request, raw_request, config, db) -> ResolvedRoute` (standalone builds a single-attempt route from local provider creds plus budget reservation; platform calls `_resolve_platform_credentials`).
- `report_usage(...)` / `report_error(...)` (standalone -> `log_usage` plus reconcile/refund; platform -> `_report_platform_usage`).

Keep `run_platform_attempts` as the shared runner; have the standalone strategy produce a one-attempt route so the non-streaming path converges on a single loop. Land incrementally behind the existing both-mode integration tests. This is the natural first phase of the pipeline-consolidation epic.

## Effort

L.

## Acceptance criteria

- A single mode-strategy object is selected once per request; handlers contain no inline `if platform_mode` branches for credential resolution or usage reporting (tool/stream mechanics may remain shared).
- Existing standalone and platform integration tests pass unchanged; new unit tests cover each strategy in isolation.
- No behavior change on the wire (same status codes, headers, and budget reconcile/refund semantics).

njbrake:

## Summary

Fixes #102. The chat, messages, and responses handlers each read `config.is_platform_mode` and re-branched on it inline at many sites (chat 15, responses 13, messages 11), interleaving the two modes' credential-resolution and usage-reporting logic in one long function body. A bug had to be fixed in three places and it was easy to add a path that silently worked in only one mode.

- New module `src/gateway/api/routes/_mode_strategy.py` with two seams, selected once per request:
  - `RequestModeStrategy` (`StandaloneStrategy` / `PlatformStrategy`): `resolve()` runs the platform resolve call or the auth plus budget-reservation flow and stashes the per-request state; `resolve_mcp_servers()` replaces the `mcp_server_ids` mode branch; `make_settlement()` builds the right settlement.
  - `RequestSettlement` (`StandaloneSettlement` / `PlatformSettlement`): supplies the `on_success` / `on_error` / `on_no_usage` / `on_incomplete` hooks the streaming generators already accept, replacing the inline `if platform_mode` branches in the streaming callbacks and the standalone non-streaming tail.
- The three handlers shed about 529 net lines; the top resolve block becomes `strategy.resolve(spec=...)`.
- Dispatch mechanics (the `run_platform_attempts` runner vs the standalone single call, the streaming fallback-vs-single split) stay in the handlers, as the issue permits.

### Scope notes

- Lighter seam, not full convergence onto `run_platform_attempts`. Standalone keeps its own dispatch because `get_provider_kwargs` returns richer credentials (`client_args`, `api_version`, Vertex env) than `ResolvedAttempt` models, so routing it through the attempt-based runner would silently drop them.
- No behavior change on the wire for the refactor itself: status codes, headers, and budget reconcile/refund semantics are preserved.

## Bundled bug fix: streaming pre-commit reservation leak

Investigating the seam surfaced a pre-existing standalone bug (introduced when the reservation system landed, not by this refactor). On `/v1/messages` and `/v1/responses`, a streaming request whose provider call failed before the first chunk logged the error but did not refund the budget reservation, while `/v1/chat/completions` and all non-streaming paths did. The leaked hold was never released (a budget-period reset zeroes `spend` but not `reserved`), so repeated pre-commit streaming failures accumulated `reserved` until the user hit a spurious "budget exceeded" 403 with low actual spend.

The fix routes the messages/responses pre-commit path through `settlement.on_error` (log plus refund), matching chat and the non-streaming handlers, and removes the now-dead `on_provider_error_precommit` hook. In platform mode this also reports an error outcome upstream on a pre-commit failure, consistent with how mid-stream errors already report. The HTTP status and body are unchanged; only `users.reserved` is released instead of leaked.

## Test plan

- [x] `make lint` (ruff) and `make typecheck` (mypy strict, 154 files)
- [x] New `tests/unit/test_mode_strategy.py`: strategy selection, both settlements, no-DB reject paths
- [x] New `tests/integration/test_mode_strategy_e2e.py` (11 tests): standalone reconcile/refund settlement, the now-uniform pre-commit refund across all three endpoints, `mcp_server_ids` resolution, and platform request-id/correlation headers plus upstream usage reporting
- [x] Full suite green: 749 passed, 10 env-gated skips, including the platform-mode and standalone chat/messages/responses suites unchanged
- [x] `uv run python scripts/generate_openapi.py --check` (no schema change)

### DIFF
--- a/src/gateway/api/routes/_mode_strategy.py
+++ b/src/gateway/api/routes/_mode_strategy.py
@@ -1,0 +1,493 @@
+"""Standalone-vs-platform mode seam for the LLM route handlers.
+
+The gateway's operating mode is derived once from ``config.is_platform_mode``
+(``core/config.py``). Historically each of the chat / messages / responses
+handlers consumed that single boolean and then re-branched on it inline at many
+sites, interleaving the two modes' credential-resolution and usage-reporting
+logic inside one long function body. This module lifts those two concerns
+behind a per-request strategy object so the handlers select the mode once and
+never branch on it again for resolution or settlement.
+
+Two seams live here:
+
+* :class:`RequestModeStrategy` — credential resolution. ``resolve`` runs the
+  platform resolve call (platform) or the auth + budget-reservation flow
+  (standalone) and stashes the per-request state the handler needs.
+* :class:`RequestSettlement` — usage reporting. The four async hooks match the
+  callbacks ``gateway.streaming.streaming_generator`` already accepts, plus the
+  standalone non-streaming success/error tail. Platform reports usage upstream;
+  standalone writes a ``UsageLog`` row and reconciles/refunds the reservation.
+
+Dispatch mechanics (the ``run_platform_attempts`` runner vs the standalone
+single call, and the streaming fallback-vs-single split) deliberately stay in
+the handlers — they are tool/stream plumbing, not mode-resolution.
+"""
+
+from __future__ import annotations
+
+import asyncio
+import time
+import uuid
+from abc import ABC, abstractmethod
+from collections.abc import Callable
+from dataclasses import dataclass
+
+from any_llm import AnyLLM, LLMProvider
+from any_llm.exceptions import AnyLLMError
+from any_llm.types.completion import CompletionUsage
+from fastapi import HTTPException, Request, Response
+from sqlalchemy.ext.asyncio import AsyncSession
+
+from gateway.api.deps import verify_api_key_or_master_key
+from gateway.api.routes._helpers import resolve_user_id
+from gateway.api.routes._platform import (
+    ResolvedRoute,
+    _extract_platform_user_token,
+    _report_platform_usage,
+    _resolve_platform_credentials,
+    _resolve_platform_mcp_servers,
+)
+from gateway.core.config import GatewayConfig
+from gateway.log_config import logger
+from gateway.models.mcp import McpServerConfig
+from gateway.rate_limit import RateLimitInfo, check_rate_limit
+from gateway.services.budget_service import (
+    ReservationHandle,
+    estimate_cost,
+    reconcile_reservation,
+    refund_reservation,
+    reserve_budget,
+)
+from gateway.services.log_writer import LogWriter
+from gateway.services.pricing_service import find_model_pricing, pricing_required_but_missing
+
+# ---------------------------------------------------------------------------
+# Usage-reporting seam
+# ---------------------------------------------------------------------------
+
+
+class RequestSettlement(ABC):
+    """Settles a request's usage once the provider call resolves.
+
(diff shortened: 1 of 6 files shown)


njbrake:

Closing this as superseded. Since this branch diverged, two PRs have landed on main that overtake it:

- #129 already shipped the bundled fix for the standalone streaming pre-commit budget-reservation leak on `/v1/messages` and `/v1/responses`.
- #130 (closes #100) consolidated the chat/messages/responses handlers onto a shared `_pipeline.py` core, building the streaming settlement callbacks in one place and reducing the route layer from 3927 to 1361 lines. Its commit also fixed the same budget leak by construction.

Both the headline strategy refactor here and its bundled bug fix are now redundant, and the branch conflicts irreconcilably with the new `_pipeline.py` structure (it was written against the old ~1000-line handlers that no longer exist). Rebasing would be a rewrite, not a fixup.

Issue #102 stays open. Its residual concern (lifting the remaining `platform_mode` branches behind a seam) is still valid, but it should now be reframed as a small follow-up on top of the `FormatAdapter` core, reusing the `RequestSettlement` hook shape from this branch's `_mode_strategy.py`.

Issue #140, opened 2026-06-11
Rename "platform mode" to a non-legacy term


njbrake:
_Note: this issue was drafted by Claude via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

Split out from #125 to keep that issue scoped to the "Otari Gateway" -> "Otari" rename.

In #125, @tbille noted:

> I would love to get rid also of the "platform mode" term which is a legacy term we used when we had any-llm platform.

"Platform mode" (and the related `GATEWAY_MODE=platform` / `effective_mode == "platform"` surface) is a holdover from the any-llm platform era. We should pick a clearer name for the connected-to-otari.ai runtime mode and rename it consistently across config, code, and docs.

Open question from #125 (no answer yet): what should "platform mode" be renamed to? Candidate framings already used in the docs include "connected mode" / "connected to otari.ai". Worth settling the term before doing the rename.

Scope when picked up:
- `mode` config field values and `GATEWAY_MODE` / `OTARI_MODE` env handling (`src/gateway/core/config.py`)
- `is_platform_mode` / `effective_mode` and call sites
- `_platform.py`, platform-protocol docs, and `tests/integration/test_platform_mode_*.py`
- docs (`modes.md`, `deployment.md`, `platform-protocol.md`)

Deferred from #125 because it is independent scope and the replacement term is not yet agreed.

PR #142, opened 2026-06-11
Rename "Otari Gateway" to "Otari"

### LINKED ISSUE #125: Rename "Otari Gateway" -> "Otari"
Should we clean up the docs a bit and rename "otari gateway" to be just "otari"? Based on the workweek discussion I think this is the correct action to consolidate our names to be what we agreed on.

@tbille

njbrake:
## Description

Consolidate the product name on **Otari** everywhere a user or operator sees it (CLI, env vars, OpenAPI title, Docker, CI workflows, docs), while keeping full backward compatibility. Closes the workweek decision to drop the "Otari Gateway" / "the gateway" naming.

Key points:
- CLI is now `otari serve` (the `gateway` command stays as a **legacy alias**).
- Env vars use the `OTARI_` prefix; the legacy `GATEWAY_` prefix still works as a fallback, and `OTARI_` wins when both are set. A new `core/env.py` `otari_env()` helper centralizes this; config aliasing now covers every scalar field (was a hand-listed 6).
- OpenAPI/FastAPI title (`otari`), default SQLite filename (`otari.db`), root tutorial page, web-search User-Agent.
- Dockerfile, docker-compose (service `otari`, postgres creds), Makefile, config.example, CI workflow files (`otari-*.yml`), and the demos.
- Docs: "the gateway" -> "Otari", keeping the "LLM gateway" descriptor, the `502 Bad Gateway` status, and the `/gateway/...` platform wire-protocol paths + `X-Gateway-Token` header (contract surface, not names).

**The Python import package stays `gateway` on purpose:** the `otari` distribution name on PyPI belongs to the Otari client SDK (a transitive dep via `any-llm-sdk`), so a top-level `otari` package here would collide with it. This is documented in `AGENTS.md` and `pyproject.toml`. Internal identifiers, Prometheus metric names, and the logger name are likewise left as `gateway`.

Existing `GATEWAY_*` variables and the `gateway` command keep working, so no configuration change is required.

### Reviewer notes / action items
- **Branch protection:** workflow files and display names were renamed. Job ids (`lint`, `typecheck`, `test`, `openapi`) are unchanged, so required-check contexts should still match; if your rules reference workflow-name-qualified contexts, update repo settings.
- **Local dev:** docker-compose postgres creds renamed `gateway` -> `otari`; an existing `postgres_data` volume will fail auth, so drop it locally. Operators whose host `.env` set compose-passed `GATEWAY_*` vars should switch to `OTARI_*` (demo `.env.example` files updated; the app still honors `GATEWAY_*`).
- The "platform mode" rename (raised by @tbille on #125) is split out into #140 and intentionally not in this PR.

## PR Type

- [ ] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [x] Documentation
- [x] Infrastructure / CI

## Relevant issues

Fixes #125

## Checklist

Note: `make lint`, `make typecheck`, `make openapi-check`, and the unit suite (350 passed) are green locally. Integration tests requiring PostgreSQL were not run in this environment (no Docker) and will run in CI.

## AI Usage

**AI Model/Tool used:** Claude Code (Opus 4.8), driven via back-and-forth with @njbrake. The reasoning and decisions are his; the implementation and prose are Claude's.

Scope (full rename including infra/demos, package-name decision, backward-compat strategy) was decided interactively with @njbrake.

## Summary

This PR consolidates the product name from "Otari Gateway" to simply "Otari" across user-facing surfaces while maintaining backward compatibility. 

### What Changed

**User-Facing Naming:**
- CLI command is now `otari serve` (with `gateway` kept as a legacy alias)
- Environment variables now use `OTARI_` prefix (e.g., `OTARI_API_KEY`, `OTARI_MODE`), with legacy `GATEWAY_` variables still supported as fallbacks
- Default SQLite database renamed from `otari-gateway.db` to `otari.db`
- OpenAPI specification title updated to "Otari"
- Docker image and docker-compose service naming changed to `otari`
- All documentation, tutorials, and guides updated to reference "Otari" instead of "Otari Gateway"

**Internal Improvements:**
- New `core/env.py` module centralizes environment variable handling with intelligent prefix detection (`OTARI_` takes precedence over legacy `GATEWAY_`)
- Configuration system expanded to support environment variable overrides for all scalar fields
- Updated CI workflows, Docker setup, and deployment documentation

**What Stayed the Same:**
- Python import package remains `gateway` (avoiding PyPI naming conflicts)
- Internal logger names, Prometheus metrics, and code identifiers retain `gateway` references
- Existing `GATEWAY_` environment variables continue to work with full backward compatibility

### Benefits

- **Clearer product identity**: Users encounter a single, consistent product name
- **Backward compatible**: Existing scripts, deployments, and automations continue working without changes
- **Flexible migration path**: Teams can gradually transition to `OTARI_` prefix while old settings remain functional

### Migration Notes

- Docker Compose users should drop their local `postgres_data` volume and update database credentials in configs (from `gateway:gateway` to `otari:otari`)
- Standalone deployments may need to migrate the SQLite database file from `otari-gateway.db` to `otari.db`


src/gateway/core/config.py
```diff
@@ -57,7 +59,7 @@ class GatewayConfig(BaseSettings):
     )
 
     database_url: str = Field(
-        default="sqlite:///./otari-gateway.db",
+        default="sqlite:///./otari.db",
```
dpoulopoulos, line 84: **[major]** The default `database_url` changes from `sqlite:///./otari-gateway.db` to `sqlite:///./otari.db`. For a standalone deployment on the SQLite default (no explicit `database_url`), upgrading silently points the app at a fresh, empty database: previously-issued API keys start returning 401, users/budgets/usage history are orphaned in the old file, and a new bootstrap key is generated. This contradicts the PR's "full backward compatibility / no configuration change is required" claim. Production and example configs that set `database_url` (Postgres) are unaffected, which narrows the blast radius, but the SQLite default is a documented, supported path. Please either keep the old default filename or add an explicit migration note (rename `otari-gateway.db` -> `otari.db`, or set `database_url`) to the CHANGELOG and the reviewer notes.

docker-compose.yml
```diff
@@ -69,22 +69,22 @@ services:
...
       # network at postgres:5432, so the host mapping is only for external
-      # tools (psql / TablePlus / etc.). Override with $GATEWAY_POSTGRES_PORT.
-      - "${GATEWAY_POSTGRES_PORT:-5433}:5432"
+      # tools (psql / TablePlus / etc.). Override with $OTARI_POSTGRES_PORT.
+      - "${OTARI_POSTGRES_PORT:-5433}:5432"
     environment:
-      - POSTGRES_USER=gateway
-      - POSTGRES_PASSWORD=gateway
-      - POSTGRES_DB=gateway
+      - POSTGRES_USER=otari
```
dpoulopoulos, line 75: **[minor]** Postgres is now initialized as `otari/otari/otari` (was `gateway/...`). The reviewer note tells operators to drop the `postgres_data` volume, but an existing operator's gitignored `config.yml` (copied from the old `config.example.yml`) still has `database_url: postgresql://gateway:gateway@postgres:5432/gateway`, so after dropping the volume the app fails auth until `config.yml` is also updated, and that second step isn't documented. The rename is also cosmetic (internal DB user/name); consider keeping the `gateway` credentials to avoid the migration friction, or document the `config.yml` edit alongside the volume drop. See also `config.example.yml:4`.

src/gateway/core/config.py
```diff
@@ -244,15 +246,43 @@ def _parse_bool_env(value: str) -> bool:
...
 def _apply_otari_env_overrides(config: dict[str, Any]) -> None:
-    for env_name, (field_name, caster) in OTARI_ENV_ALIASES_TO_GATEWAY.items():
-        value = os.getenv(env_name)
+    """Layer OTARI_<FIELD> env vars over the config dict for every scalar field.
+
+    Written into the init dict so they take precedence over both YAML and the
+    legacy GATEWAY_ pydantic prefix. Complex fields (lists/dicts) are left to
+    YAML and pydantic's native env handling.
+    """
+    for field_name in GatewayConfig.model_fields:
```
dpoulopoulos, line 377: **[minor]** `_apply_otari_env_overrides` only covers scalar fields; complex fields (`cors_allow_origins`, `providers`, `pricing`, `platform`) are skipped, so `OTARI_CORS_ALLOW_ORIGINS` is silently ignored while the legacy `GATEWAY_CORS_ALLOW_ORIGINS` still works via pydantic's native env prefix. It's documented in the docstring and these are usually set via YAML, so impact is low, but the asymmetry means an operator migrating an env-set list/dict from `GATEWAY_` to `OTARI_` loses it with no error. Worth a line in the configuration docs, or handling complex fields too.

demo/code-exec/ask.sh
```diff
@@ -33,13 +33,13 @@ fi
...
-GATEWAY_CONTAINER=$(cd "$GATEWAY_ROOT" && docker compose ps -q gateway 2>/dev/null | head -1 || true)
-SANDBOX_CONTAINER=$(cd "$GATEWAY_ROOT" && docker compose ps -q sandbox 2>/dev/null | head -1 || true)
+OTARI_CONTAINER=$(cd "$OTARI_ROOT" && docker compose ps -q otari 2>/dev/null | head -1 || true)
+SANDBOX_CONTAINER=$(cd "$OTARI_ROOT" && docker compose ps -q sandbox 2>/dev/null | head -1 || true)
 
-GATEWAY_PORT=${OTARI_PORT:-${GATEWAY_PORT:-8000}}
-GATEWAY_URL=${GATEWAY_URL:-http://localhost:${GATEWAY_PORT}}
-GATEWAY_KEY=${GATEWAY_KEY:-demo-master-key}
-GATEWAY_USER=${GATEWAY_USER:-demo}
+OTARI_PORT=${OTARI_PORT:-${OTARI_PORT:-8000}}
```
dpoulopoulos: **[nit]** `OTARI_PORT=${OTARI_PORT:-${OTARI_PORT:-8000}}` is a mechanical-rename artifact of the old `${OTARI_PORT:-${GATEWAY_PORT:-8000}}` and is now self-referentially redundant; collapse to `${OTARI_PORT:-8000}`. Same in `demo/guardrails/ask.sh:40`, `demo/guardrails/demo_flow.sh:32`, and `demo/web-search/ask.sh:45`.

README.md
```diff
@@ -150,9 +150,9 @@ uv run pytest tests/unit/test_gateway_cli.py -v
 
 ## Docker
 
-The gateway image is published on [Docker Hub](https://hub.docker.com/r/mzdotai/otari).
+Otari image is published on [Docker Hub](https://hub.docker.com/r/mzdotai/otari).
```
dpoulopoulos: **[nit]** Mechanical "The gateway" -> "Otari" replacement dropped the article: "Otari image is published..." (and "Otari will be available..." at line 172). Same pattern in docker-compose.yml comments ("The Otari calls it" / "The Otari reaches it", lines 191/210/106). Reads awkwardly and AGENTS.md asks for clean prose; reword to "The Otari image..." / "Otari calls it...".

dpoulopoulos (changes requested):
Renames the product from "Otari Gateway" to "Otari" across the user-facing surface (CLI, env vars, OpenAPI title, default SQLite filename, Docker/compose, CI workflow files, docs, demos) while keeping the Python import package `gateway`. Backward compatibility is preserved: a new `otari_env()` helper plus a generalized `_apply_otari_env_overrides` accept the `OTARI_` prefix with `GATEWAY_` as a legacy fallback (`OTARI_` wins), and both the `otari` and `gateway` CLI entry points are retained. Contract surfaces (`X-Gateway-Token`, `/gateway/*` wire paths, the `gateway` logger and `gateway_*` metric names, `502 Bad Gateway`) are intentionally left unchanged. The core is sound and well-tested; the concerns are a data-continuity edge on the changed SQLite default plus a few references the rename sweep missed.

### General notes
- **[minor]** `alembic/env.py:33` — This fallback default still points at the old `sqlite:///./otari-gateway.db`, now inconsistent with the app default (`otari.db` in `core/config.py`). It is not reached by app-startup auto-migrate, `otari migrate`, or `init-db` (all set `sqlalchemy.url` / `OTARI_DATABASE_URL` explicitly), but a bare `alembic upgrade head` with no `*_DATABASE_URL` set now targets a different file than the running app, which is confusing in dev. One-line fix to `otari.db` for consistency. (Not part of the diff, but a direct consequence of the default rename.)
- **[minor]** `.gitignore:19` — `.gitignore` ignores `gateway.db` and `otari-gateway.db` but not the new default `otari.db`. A local standalone run now writes `otari.db` to the repo root untracked-and-unignored, risking an accidental commit of a dev database (hashed API keys, usage). Add `otari.db`. (Not part of the diff, but a consequence of the default rename.)

src/gateway/core/config.py
njbrake: _Note: drafted by Claude via back-and-forth with @njbrake; reasoning and decisions are his, prose is Claude's._

Good catch, agreed. Added an explicit "Upgrade notes" block to the CHANGELOG (5733226) spelling out that standalone deployments on the default SQLite path must rename `otari-gateway.db` to `otari.db` or set `database_url`, otherwise the app starts on a fresh, empty DB. Also aligned `alembic/env.py`'s fallback default to `otari.db` and added `otari.db` to `.gitignore` (84e1090), since both still pointed at the old name.

docker-compose.yml
njbrake: _Note: drafted by Claude via back-and-forth with @njbrake; reasoning and decisions are his, prose is Claude's._

Documented rather than reverted. Reverting the creds to `gateway` would break fresh users, since `config.example.yml` already ships `otari:otari`, so a clean `cp config.example.yml config.yml` would then mismatch the container. The CHANGELOG upgrade note (5733226) now tells existing operators to both drop the `postgres_data` volume and update their local `config.yml` `database_url` to `postgresql://otari:otari@postgres:5432/otari`.

src/gateway/core/config.py
njbrake: _Note: drafted by Claude via back-and-forth with @njbrake; reasoning and decisions are his, prose is Claude's._

Documented in `docs/configuration.md` (5733226): `OTARI_` overrides apply to scalar fields only; list/dict fields (`cors_allow_origins`, `providers`, `pricing`, `platform`) are set via YAML, and the `platform` block has dedicated `PLATFORM_*` vars. Kept it doc-only rather than expanding the override to complex fields to keep the change surface small.

demo/code-exec/ask.sh
njbrake: _Note: drafted by Claude via back-and-forth with @njbrake; reasoning and decisions are his, prose is Claude's._

Fixed, collapsed to `${OTARI_PORT:-8000}` in all four scripts (5733226).

README.md
njbrake: _Note: drafted by Claude via back-and-forth with @njbrake; reasoning and decisions are his, prose is Claude's._

Fixed `README.md:153` ("The Otari image...") and the `docker-compose.yml` comments at 191 and 210 ("Otari calls it" / "Otari reaches it") in 5733226. Left `README.md:172` ("Otari will be available at...") and the `sandbox` comment at 106 ("The Otari container reaches it...") as is, since those read correctly with Otari as a proper noun / with "container" as the head noun.

dpoulopoulos (approved):

PR #145, opened 2026-06-11
fix(sdk-codegen): type response schemas in serialization mode so reasoning is a string

### LINKED ISSUE #143: SDK can't deserialize message.reasoning: schema types it as an object but the gateway emits a string

## Summary

Completions against reasoning-capable models fail client-side in the generated Python SDK because the schema types `message.reasoning` as an object while the gateway serializes it as a plain string.

```
1 validation error for CCReasoning
  Input should be a valid dictionary or instance of CCReasoning
  [type=model_type, input_value='User says "hi". Likely g...', input_type=str]
```

## Reproduction

Reproduces with `mzai:openai/gpt-oss-120b` and `mzai:nvidia/nemotron-3-super-120b-a12b` (anthropic models are unaffected):

```python
from otari import AsyncOtariClient
c = AsyncOtariClient(platform_token="...")
await c.completion(model="mzai:openai/gpt-oss-120b", messages=[{"role": "user", "content": "hi"}])
```

## Root cause (schema vs serializer mismatch)

The gateway serializes `message.reasoning` as a plain string. The generated `CCReasoning` docstring even states this intent: "Serializes as a plain JSON string so that responses are compatible with OpenAI-style clients that expect `message.reasoning` to be a string."

But the OpenAPI schema types it as an object: `CCChatCompletionMessage.reasoning: Optional[CCReasoning]`, where `CCReasoning.content` is a required `StrictStr`, and `from_dict` calls `CCReasoning.from_dict(obj["reasoning"])`. Since the SDK is generated from that schema, it rejects the string the server actually returns.

The fix belongs in the response schema/serializer that the generated client is built from (there is no hand-written `Reasoning` type in `src/`, so the generated `CCReasoning` is purely schema-derived). Suggested: type `reasoning` as `string | CCReasoning` in the schema, or align the serializer and the schema on one shape.

## Impact

any-llm and any OpenAI-style consumer of the Python SDK cannot use any reasoning-emitting model routed through the gateway; the call fails during response deserialization before the caller ever sees the content.

njbrake:
## Description

The SDK codegen enriches the spec with the typed response schemas from `any-llm` via `model_json_schema()`. Those calls used the default validation mode, which typed any-llm's `Reasoning` as an object (`{content: str}`). But `Reasoning` has a `model_serializer` that emits a plain string, so the gateway puts a string on the wire for `message.reasoning`. The generated client, built from the object-typed schema, then failed to deserialize the string, and every reasoning-emitting model (for example `mzai:openai/gpt-oss-120b`, `mzai:nvidia/nemotron-3-super-120b-a12b`) broke during response parsing before the caller ever saw the content.

This switches the five response schemas (`ChatCompletion`, `ChatCompletionChunk`, `MessageResponse`, `RerankResponse`, `CreateEmbeddingResponse`) to `mode="serialization"`, so the schema reflects the serialized wire format. Today that changes only the `Reasoning` shape (object to string); the other response types are byte-identical, but serialization mode keeps them aligned with what is actually sent should their serializers ever diverge from their validation shape.

A regression test asserts `CC_Reasoning` and `CCK_Reasoning` are plain strings. It was verified to fail without the fix (schema is an object) and pass with it.

Note: the linked any-llm issue (mozilla-ai/any-llm#1126) describes two further failures (tool_calls wrapper shape and non-streaming `MessageResponse`); both are characterized there as any-llm-side conversion bugs, not otari schema bugs, so they are out of scope here.

## PR Type

- [x] Bug Fix

## Relevant issues

Fixes #143

## Checklist

- [x] If the API contract changed, I regenerated the OpenAPI spec (`uv run python scripts/generate_openapi.py`). (No change: the main spec is unaffected; this only touches SDK-codegen enrichment, and `make openapi-check` passes.)

## AI Usage

Implemented via the fix-github-issue workflow under @njbrake's direction; reasoning and decisions are his.

**NOTE:**

## Overview

This PR fixes SDK client deserialization errors for AI models that return reasoning outputs. The generated SDK schemas are now aligned with the actual wire format sent by the gateway, preventing validation failures when receiving reasoning fields.

## What Changed

**SDK Code Generation (`scripts/sdk_codegen/generate.py`)**
- Modified the `enrich_spec` function to generate response schemas using Pydantic's serialization mode instead of the default validation mode
- Applies to 5 response types: ChatCompletion, ChatCompletionChunk, MessageResponse, RerankResponse, and CreateEmbeddingResponse
- The serialization-mode schemas now match the actual wire format that the gateway sends

**Tests (`tests/unit/test_sdk_codegen.py`)**
- Added a regression test (`test_enrich_types_reasoning_as_string_matching_wire_format`) to verify that reasoning fields (`CC_Reasoning` and `CCK_Reasoning`) are correctly typed as plain strings in the generated schemas

## Why This Matters

Previously, the `Reasoning` type was generated as an object (`{content: string}`) in the SDK schema, but the gateway serializes it as a plain string on the wire. This mismatch caused client-side deserialization failures for reasoning-capable models (like `mzai:openai/gpt-oss-120b` and `mzai:nvidia/nemotron-3-super-120b-a12b`). By switching to serialization-mode schema generation, the SDK now correctly expects a string, eliminating these validation errors.

## Technical Notes

The fix uses Pydantic's `model_json_schema(mode="serialization")` to generate schemas that reflect the actual serialized output rather than the validation input format. While this specifically addresses the `Reasoning` field conversion (object → string), it keeps other response types aligned with their wire format and prepares for any future serializer changes in upstream dependencies.

### DIFF
--- a/scripts/sdk_codegen/generate.py
+++ b/scripts/sdk_codegen/generate.py
@@ -264,11 +264,20 @@ def enrich_spec(spec: dict[str, Any]) -> dict[str, Any]:
             schemas[f"{prefix}_{name}"] = body
         return js
 
+    # These are response shapes, so generate the JSON Schema in ``serialization``
+    # mode: it reflects the bytes the gateway actually puts on the wire, not the
+    # (looser) validation shape. The any-llm ``Reasoning`` type is the motivating
+    # case: it accepts ``{"content": str}`` on input but a ``model_serializer``
+    # emits a plain string, so ``message.reasoning`` is a string in responses.
+    # Validation-mode schema typed it as an object, and the generated client then
+    # rejected the string the server sends. See mozilla-ai/otari#143.
     schemas["ChatCompletion"] = absorb(
-        ChatCompletion.model_json_schema(ref_template="#/components/schemas/CC_{model}"), "CC"
+        ChatCompletion.model_json_schema(mode="serialization", ref_template="#/components/schemas/CC_{model}"),
+        "CC",
     )
     schemas["ChatCompletionChunk"] = absorb(
-        ChatCompletionChunk.model_json_schema(ref_template="#/components/schemas/CCK_{model}"), "CCK"
+        ChatCompletionChunk.model_json_schema(mode="serialization", ref_template="#/components/schemas/CCK_{model}"),
+        "CCK",
     )
     msgs = absorb(
         TypeAdapter(list[ChatCompletionMessageParam]).json_schema(ref_template="#/components/schemas/MSG_{model}"),
@@ -281,14 +290,21 @@ def enrich_spec(spec: dict[str, Any]) -> dict[str, Any]:
     # inject them so the generated core returns typed models instead of ``object``.
     # Audio/images are intentionally left opaque: they return binary/file payloads
     # that do not map onto a JSON response schema.
+    # Response shapes too: serialization mode keeps the schema aligned with the
+    # serialized wire format (see the ChatCompletion note above).
     schemas["MessageResponse"] = absorb(
-        MessageResponse.model_json_schema(ref_template="#/components/schemas/MR_{model}"), "MR"
+        MessageResponse.model_json_schema(mode="serialization", ref_template="#/components/schemas/MR_{model}"),
+        "MR",
     )
     schemas["RerankResponse"] = absorb(
-        RerankResponse.model_json_schema(ref_template="#/components/schemas/RR_{model}"), "RR"
+        RerankResponse.model_json_schema(mode="serialization", ref_template="#/components/schemas/RR_{model}"),
+        "RR",
     )
     schemas["CreateEmbeddingResponse"] = absorb(
-        CreateEmbeddingResponse.model_json_schema(ref_template="#/components/schemas/EMB_{model}"), "EMB"
+        CreateEmbeddingResponse.model_json_schema(
+            mode="serialization", ref_template="#/components/schemas/EMB_{model}"
+        ),
+        "EMB",
     )
 
     def set_json_200(path: str, schema_name: str, description: str) -> None:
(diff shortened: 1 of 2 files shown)


njbrake:

A question that came up while reviewing this fix, posted here in case others wonder the same thing: the whole pipeline uses any-llm types on both ends (client-side any-llm wraps the otari SDK, and the gateway uses any-llm to call providers), so shouldn't the shapes match by default? Why did the schema and the wire format disagree at all?

The symmetry holds at the Python type level, but the otari SDK in the middle does not use those Python classes. It is generated by OpenAPI Generator from a JSON Schema snapshot of them, and that snapshot step is lossy:

1. Gateway side: any-llm builds a real `ChatCompletion` pydantic object from the provider response. When the gateway serializes it, the `Reasoning` `model_serializer` fires and `message.reasoning` goes on the wire as a plain string.
2. The middle hop: the generated SDK types (`CCReasoning` and friends) exist only in generated code, reconstructed from the schema export rather than shared with any-llm.
3. Client side: any-llm's otari provider wraps the generated SDK, so the strict generated types sit in the parse path before any-llm's own types ever see the data.

The subtle part is that `Reasoning` is an asymmetric type. In Python, the class carries both behaviors at once: a `model_validator` that accepts either a string or `{"content": str}` on the way in, and a `model_serializer` that always emits a string on the way out. One class, two wire shapes depending on direction. JSON Schema cannot express both at once, so pydantic makes you pick: `model_json_schema()` defaults to validation mode (the object shape the type accepts), while `mode="serialization"` gives the shape it emits (a string). The codegen was exporting validation mode for response bodies, so the generated client was built expecting the input shape of a type it only ever sees in its output shape.

There is a second asymmetry stacked on top: forgiveness. If the client parsed responses with any-llm's actual pydantic classes, there would be no bug even with the schema mismatch, because the `Reasoning` validator happily coerces a plain string. Generated code has none of that leniency; `from_dict` does literal schema enforcement, so the string is a hard validation error. That is also why plain OpenAI clients never hit this: they treat `reasoning` as an untyped field.

Hence the fix here: export the response schemas in serialization mode so the schema matches the direction the data flows. Today only the `Reasoning` shape changes (object to string); the other response models serialize the same shape they validate, and the new test pins the string shape so it cannot regress.

khaledosman (approved):

PR #159, opened 2026-06-15
fix(docker): keep python 3.14, bump onnxruntime instead of downgrading

njbrake:
## Description

Fixes the failing `build-and-push-image` job on `main` (run [27434669567](https://github.com/mozilla-ai/otari/actions/runs/27434669567/job/81093472187)) while keeping the Docker image on Python 3.14.

Root cause: `onnxruntime` resolved to `1.20.1`, which ships only `cp313` wheels (no `cp314`), so the build failed under Python 3.14:

```
error: Distribution `onnxruntime==1.20.1` can't be installed because it doesn't have a source distribution or wheel for the current platform
hint: You're using CPython 3.14 (`cp314`), but `onnxruntime` (v1.20.1) only has wheels with the following Python ABI tags: `cp313`, `cp313t`
```

`onnxruntime` is not a direct dependency. It is pulled in transitively by `magika` (a core dependency via `markitdown`) and by `rapidocr-onnxruntime` (the optional `ocr` extra). The version was held back by `magika` 0.6.3, which declares `onnxruntime<=1.20.1` for `sys_platform == "win32"`. `uv` resolves a single universal lock, so that win32 cap applies to every platform and pins `onnxruntime` at `1.20.1` everywhere. `magika` itself cannot be bumped: the latest `markitdown` (0.1.6) pins `magika~=0.6.1`, and only `magika` >= 1.0 gates `onnxruntime` correctly by Python version.

Fix: this is a follow-up to @tbille's review, who preferred bumping the dependency over downgrading Python. Otari is deployed as a Linux server and developed on Linux / macOS; it is never packaged for Windows. Restricting the `uv` resolution environments to `linux` and `darwin` (`[tool.uv].environments` in `pyproject.toml`) drops the win32-only cap, and `onnxruntime` then resolves to `1.26.0`, which ships `cp314` wheels for `linux/amd64` and `linux/arm64`. The Dockerfile stays on `python:3.14-slim`.

CI (tests, lint, typecheck, sdk-codegen, sdk staleness) is bumped to Python 3.14 so the suite runs on the same version we ship. `requires-python` stays `">=3.13"`, so the package still supports 3.13+.

Tradeoff: the lock no longer covers Windows, so `uv sync --frozen` will not resolve on a Windows host. That is acceptable for a Linux/macOS server project, and the rationale is documented in the `[tool.uv]` comment.

Validation (all on Python 3.14):
- `uv sync --frozen --no-dev` (the Docker build path) succeeds; `uv sync --frozen --extra ocr --no-dev` succeeds.
- `make lint` clean, `make typecheck` clean (173 files), unit tests 393 passed / 1 skipped.
- `uv lock --check` confirms the lock matches `pyproject.toml`.
- The `otari-docker.yml` workflow has no `pull_request` trigger, so this PR's own checks do not exercise the multi-arch build; a `workflow_dispatch` run on this branch is the proof.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [x] Infrastructure / CI

## Relevant issues

No tracked issue. Unblocks the failing `build-and-push-image` job on `main` (run 27434669567).

## Checklist

- [ ] I have added or updated tests that cover my change (`tests/unit`, `tests/integration`). N/A: dependency/CI change, no application code; validated by lint, typecheck, unit tests, and a Docker build on 3.14.
- [x] I ran the Definition of Done checks locally (`make lint`, `make typecheck`, and unit tests; integration needs Docker and runs in CI).
- [x] Documentation was updated where necessary (AGENTS.md CI Python note).
- [ ] If the API contract changed, I regenerated the OpenAPI spec (`uv run python scripts/generate_openapi.py`). N/A: no API change.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:** Claude Code (Opus 4.8), driven via back-and-forth with @njbrake. The reasoning and decisions are his; the implementation and prose are Claude's.

Diagnosed the dependency chain by reproducing the resolution locally, then validated the fix with `uv sync` on Python 3.14.

**NOTE:**

## Overview
This PR resolves a failing Docker build by updating the project to use Python 3.14 consistently and configuring dependency resolution to work correctly with the new Python version.

## Changes Made

**CI Workflows**: Updated five GitHub Actions workflows (`otari-lint`, `otari-tests`, `otari-typecheck`, `otari-sdk-codegen`, and `sdk-regen-staleness`) to run tests and checks against Python 3.14 instead of 3.13, aligning with the Docker image version.

**Dependency Configuration**: Added a `[tool.uv]` configuration block to `pyproject.toml` that restricts dependency resolution to Linux and macOS platforms only, preventing Windows-specific constraints from affecting the resolution process.

**Documentation**: Updated `AGENTS.md` to reflect that the CI infrastructure now uses Python 3.14, while the package continues to support Python 3.13 and later.

## Technical Notes
The root cause was an incompatibility where `onnxruntime 1.20.1` (pulled in transitively) lacks wheel distributions for Python 3.14. By restricting `uv`'s resolution environment to Linux and Darwin only, the dependency resolver can now select `onnxruntime 1.26.0`, which provides Python 3.14 wheels for both `linux/amd64` and `linux/arm64` architectures. The project is server-only software never packaged for Windows, so this platform-specific constraint exclusion is appropriate.

### DIFF
--- a/.github/workflows/otari-lint.yml
+++ b/.github/workflows/otari-lint.yml
@@ -30,7 +30,7 @@ jobs:
 
     - uses: astral-sh/setup-uv@v7
       with:
-        python-version: 3.13
+        python-version: 3.14
         activate-environment: true
 
     - name: Install dependencies
--- a/.github/workflows/otari-sdk-codegen.yml
+++ b/.github/workflows/otari-sdk-codegen.yml
@@ -59,7 +59,7 @@ jobs:
 
       - uses: astral-sh/setup-uv@v7
         with:
-          python-version: 3.13
+          python-version: 3.14
           activate-environment: true
 
       - name: Install Otari deps
--- a/.github/workflows/otari-tests.yml
+++ b/.github/workflows/otari-tests.yml
@@ -35,7 +35,7 @@ jobs:
 
     - uses: astral-sh/setup-uv@v7
       with:
-        python-version: 3.13
+        python-version: 3.14
         activate-environment: true
 
     - name: Install dependencies
@@ -68,7 +68,7 @@ jobs:
 
     - uses: astral-sh/setup-uv@v7
       with:
-        python-version: 3.13
+        python-version: 3.14
         activate-environment: true
 
     - name: Install dependencies
--- a/.github/workflows/otari-typecheck.yml
+++ b/.github/workflows/otari-typecheck.yml
@@ -30,7 +30,7 @@ jobs:
 
     - uses: astral-sh/setup-uv@v7
       with:
-        python-version: 3.13
+        python-version: 3.14
         activate-environment: true
 
     - name: Install dependencies
--- a/.github/workflows/sdk-regen-staleness.yml
+++ b/.github/workflows/sdk-regen-staleness.yml
@@ -39,7 +39,7 @@ jobs:
 
       - uses: astral-sh/setup-uv@v7
         with:
-          python-version: 3.13
+          python-version: 3.14
 
       - name: Check for stale regeneration PRs
         env:
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -149,7 +149,7 @@ ORM entities are in `src/gateway/models/entities.py` (User, APIKey, Budget, Usag
 - Prefer structured/contextual log messages with `%s` formatting placeholders.
 - Do not log secrets, tokens, or raw API keys (bootstrap exception is intentional one-time behavior).
 ## CI Rules to Mirror Locally
-- Python version for CI: 3.13 (`.github/workflows/otari-tests.yml`).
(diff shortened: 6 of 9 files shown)


tbille (changes requested):
No a big fan of downgrading to a lower version of python otari's docker image just for one library.

Can you check if upgrading the library to a more recent version (like 1.26.0) would fix the issue? https://pypi.org/project/onnxruntime/

njbrake:
@tbille works for me! In this case I updated our packages to be for py3.14 so that we are testing with py3.14 as well. Turns out the onnxruntime pin to 1.20 was because of a windows dependency. Since we're not using windows for otari, I made a tweak so that now we can have both

tbille (approved):
amazing, thanks for bumping the python version everywhere else 💙 🐍

PR #161, opened 2026-06-15
fix(chat): forward reasoning_effort to the provider

### LINKED ISSUE #150: Gateway drops reasoning_effort (missing from ChatCompletionRequest), so reasoning never reaches the provider

## Summary

The OpenAI-compatible chat completions endpoint silently drops `reasoning_effort`, so models that need it to enable reasoning (Anthropic extended thinking; OpenAI reasoning effort levels) never produce reasoning content through the gateway.

`ChatCompletionRequest` in `src/gateway/api/routes/chat.py` enumerates its accepted fields (`max_tokens`, `max_completion_tokens`, `top_p`, `tools`, `tool_choice`, `response_format`, `guardrails`, ...) but does not include `reasoning_effort`. Pydantic drops unknown fields, so the value never reaches the `acompletion(**kwargs)` call. any-llm (which the gateway uses) already maps `reasoning_effort` to Anthropic extended thinking and surfaces the result, but it never receives the parameter.

## Reproduction

```python
from otari import AsyncOtariClient
c = AsyncOtariClient(platform_token="...")
r = await c.completion(
    model="anthropic:claude-sonnet-4-6",
    messages=[{"role": "user", "content": "What is 17*23? Think step by step."}],
    reasoning_effort="high",
    max_tokens=1500,
)
print(r.choices[0].message.reasoning)  # None
```

Observed across `anthropic:claude-haiku-4-5`, `claude-sonnet-4-6`, and `claude-opus-4-8` (all return no reasoning, with or without `reasoning_effort`). `mzai:openai/gpt-oss-120b` does return reasoning, but only because it reasons by default; it is not driven by `reasoning_effort` either.

## Fix

Add `reasoning_effort` to the gateway's `ChatCompletionRequest` (and forward it through to `acompletion`), then regenerate the SDK. any-llm handles the provider-specific translation.

## Prevention

This slipped through because the gateway's OpenAI-compatible request schema is hand-maintained and can drift from the OpenAI / any-llm completion-parameter surface, while the SDK-codegen and staleness checks only verify SDK-vs-spec consistency (both omit the field consistently). A request-parameter coverage test, asserting the schema includes the completion params the gateway intends to support (for example, derived from any-llm's `CompletionParams`, which the gateway already depends on), would catch this and future omissions automatically.

Found via any-llm's otari e2e tests, where the reasoning tests failed for otari.

njbrake:
## Description

The OpenAI-compatible chat completions endpoint silently dropped `reasoning_effort`. `ChatCompletionRequest` (`src/gateway/api/routes/chat.py`) is a hand-maintained allowlist of accepted fields and did not declare `reasoning_effort`, so pydantic dropped the value before `request.model_dump(exclude_unset=True)`, and it never reached `acompletion(**kwargs)`. Models that rely on it to enable reasoning (Anthropic extended thinking, OpenAI reasoning effort levels) therefore never produced reasoning content through the gateway. any-llm already maps `reasoning_effort` to the provider-specific behavior; it just never received the parameter.

This PR:

- Adds `reasoning_effort: str | None = None` to `ChatCompletionRequest`. The loose `str | None` typing matches neighboring fields and leaves value translation/validation to any-llm. Default `None` plus `exclude_unset=True` means callers who do not set it keep any-llm's own default (`auto`), so existing traffic is unchanged.
- Removes the `reasoning_effort` entry from `KNOWN_UNSUPPORTED` in the chat-request param coverage test (the allowlist-accuracy test enforces this removal).
- Adds an integration test asserting `reasoning_effort` reaches `acompletion`.
- Regenerates the OpenAPI spec.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes #150

## Checklist

Note: the new integration test could not run locally (no Docker for the testcontainers Postgres); it will run in CI. The forwarding path was validated directly: the field is accepted, survives `_strip_gateway_fields`, forwards to `acompletion`, and is omitted when unset. Unit suite, ruff, mypy, and `openapi-check` all pass locally.

## AI Usage

- [ ] No AI was used.

Implemented via Claude Code through back-and-forth with @njbrake. The reasoning and decisions are his; the code and prose were drafted by Claude.

**NOTE:**

### DIFF
--- a/docs/public/openapi.json
+++ b/docs/public/openapi.json
@@ -335,6 +335,17 @@
             "title": "Model",
             "type": "string"
           },
+          "reasoning_effort": {
+            "anyOf": [
+              {
+                "type": "string"
+              },
+              {
+                "type": "null"
+              }
+            ],
+            "title": "Reasoning Effort"
+          },
           "response_format": {
             "anyOf": [
               {
--- a/src/gateway/api/routes/chat.py
+++ b/src/gateway/api/routes/chat.py
@@ -88,6 +88,7 @@ class ChatCompletionRequest(BaseModel):
     max_tokens: int | None = None
     max_completion_tokens: int | None = None
     top_p: float | None = None
+    reasoning_effort: str | None = None
     stream: bool = False
     stream_options: dict[str, Any] | None = None
     tools: list[dict[str, Any]] | None = None
--- a/tests/integration/test_client_args.py
+++ b/tests/integration/test_client_args.py
@@ -140,3 +140,36 @@ async def test_provider_config_without_client_args(
         )
 
     assert "client_args" not in captured_kwargs or captured_kwargs.get("client_args") is None
+
+
+@pytest.mark.asyncio
+async def test_reasoning_effort_passed_to_acompletion(
+    client: TestClient,
+    master_key_header: dict[str, str],
+    test_user: dict[str, Any],
+) -> None:
+    """`reasoning_effort` must reach acompletion (mozilla-ai/otari#150).
+
+    The schema is a hand-maintained allowlist, so a field missing from
+    ``ChatCompletionRequest`` is silently dropped by pydantic before the
+    provider call and reasoning never gets enabled.
+    """
+    captured_kwargs: dict[str, Any] = {}
+
+    async def mock_acompletion(**kwargs: Any) -> None:
+        captured_kwargs.update(kwargs)
+        raise MockCompletionError
+
+    with patch("gateway.api.routes.chat.acompletion", new=mock_acompletion):
+        client.post(
+            "/v1/chat/completions",
+            json={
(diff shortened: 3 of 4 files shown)


njbrake:

Closing in favor of #160, which fixes the same root cause (#150) structurally. Instead of hand-adding `reasoning_effort` to the allowlist, #160 derives the request schemas from any-llm's typed `*Params` models, so `reasoning_effort` plus nine other previously-dropped chat params all forward and future drift becomes impossible by construction. #160 already carries `Closes #150`, so the issue stays linked there. This PR is a narrow subset and would only create a merge conflict (it edits the coverage test that #160 deletes), so there is no reason to keep it open.

PR #164, opened 2026-06-15
feat(web-search): forward X-Gateway-Token to the search backend

agpituk:
## Description

In platform mode, forward the gateway's platform token as an `X-Gateway-Token` header on the web-search backend's `/search` call.

This lets a web-search backend authenticate the calling gateway. When the backend is reachable beyond the gateway's private network, an unauthenticated `/search` could be abused to spend provider quota; forwarding the gateway token lets the backend reject arbitrary clients.

Standalone backends (a bundled SearXNG or a self-hosted adapter) receive **no** token and ignore the header, so their behaviour is unchanged.

### How
- `WebSearchBackend` gains an optional `auth_token`; when set it is sent as `X-Gateway-Token` on the `/search` GET.
- `_build_web_search_backend` threads the token through.
- `ToolContext` carries `web_search_auth_token`, set from `config.platform_token` **only in platform mode and only when `GATEWAY_WEB_SEARCH_URL` targets the platform base URL** (it is `None` in standalone and for non-platform backends, so a high-privilege platform credential is never leaked to a SearXNG/third-party adapter).

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

n/a

## Checklist

- [x] If the API contract changed, I regenerated the OpenAPI spec (`uv run python scripts/generate_openapi.py`). — N/A: the gateway's own API is unchanged; this only adds a header the gateway *sends* to an external backend.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:** Claude Opus 4.8 (via Claude Code)

**Any additional AI details you'd like to share:** Change scoped and reviewed against the existing web-search backend; tests added for both the header-present and header-absent paths.

## Overview
This PR updates the gateway’s **platform mode** so it can safely forward its platform authentication token to the web-search backend. The token is only sent to the trusted, platform-hosted web-search endpoint, and **never** to standalone or third-party search backends.

## What Changed
- **Web-search backend:** Added optional `auth_token` support so the backend includes an `X-Gateway-Token` header on **`/search`** requests when a token is provided.
- **Gateway plumbing:** Added `web_search_auth_token` to tool context and threaded it through all web-search backend construction paths (so it’s applied consistently where applicable).
- **Secure URL scoping:** Added URL/origin + base-path boundary checks to ensure the token is forwarded **only** when the configured web-search URL truly targets the platform (exact scheme/host/port match, and only when it falls under the platform base path with a `/` boundary). This blocks spoofed hosts and path-prefix collisions.
- **Documentation:** Updated `docs/platform-protocol.md` to document the platform-mode web-search token-forwarding rules.

## Tests
- Added/updated **unit tests** for:
  - sending `X-Gateway-Token` when `auth_token` is set
  - omitting the header when no token is set
  - ensuring the header is sent **only** on `/search` and never on per-result page fetches
  - validating the URL scoping logic (including confusable/spoofed host cases and port normalization)
- Added **integration test** coverage confirming:
  - the platform token is forwarded to the platform web-search backend in platform mode
  - the token is not leaked to non-platform web-search backends
  - test stub state is reset between assertions

## Benefits
- **Security:** Prevents leakage of a high-privilege platform token to untrusted web-search destinations.
- **Safety & abuse prevention:** Enables the backend to authenticate gateway-originated search requests while reducing risk of unauthorized access and provider quota misuse.

### DIFF
--- a/src/gateway/api/routes/_pipeline.py
+++ b/src/gateway/api/routes/_pipeline.py
@@ -523,6 +523,7 @@ class ToolContext:
         use_web_search: bool,
         web_search_tool_entry: dict[str, Any] | None,
         web_search_url: str | None,
+        web_search_auth_token: str | None,
         remaining_user_tools: list[dict[str, Any]] | None,
         max_tool_iterations: int,
         tools_header: str | None,
@@ -534,6 +535,7 @@ class ToolContext:
         self.use_web_search = use_web_search
         self.web_search_tool_entry = web_search_tool_entry
         self.web_search_url = web_search_url
+        self.web_search_auth_token = web_search_auth_token
         self.remaining_user_tools = remaining_user_tools
         self.max_tool_iterations = max_tool_iterations
         self.tools_header = tools_header
@@ -603,6 +605,11 @@ async def prepare_gateway_tools(
 
         web_search_tool_entry, remaining_user_tools = _extract_web_search_tool(tools_after_sandbox)
         web_search_url: str | None = otari_env("WEB_SEARCH_URL") or None
+        # Forwarded to the search backend as `X-Gateway-Token`. Only set in
+        # platform mode, where the backend may be the platform-hosted web-search
+        # endpoint that authenticates the gateway. Standalone backends (SearXNG /
+        # self-hosted adapter) get no token and ignore the header.
+        web_search_auth_token: str | None = None
         use_web_search = False
         if web_search_tool_entry is not None:
             if web_search_url is None:
@@ -626,6 +633,7 @@ async def prepare_gateway_tools(
             # Standalone mode has no platform to consult.
             if ctx.platform_mode:
                 assert ctx.user_token is not None  # guaranteed by the platform-mode preamble
+                web_search_auth_token = ctx.config.platform_token
                 web_search_policy = await _resolve_platform_web_search(
                     config=ctx.config,
                     user_token=ctx.user_token,
@@ -656,6 +664,7 @@ async def prepare_gateway_tools(
         use_web_search=use_web_search,
         web_search_tool_entry=web_search_tool_entry,
         web_search_url=web_search_url,
+        web_search_auth_token=web_search_auth_token,
         remaining_user_tools=remaining_user_tools,
         max_tool_iterations=min(
             max_tool_iterations or DEFAULT_MAX_TOOL_ITERATIONS,
@@ -828,6 +837,7 @@ async def dispatch_non_stream(
     async with _build_web_search_backend(
         base_url=tool_ctx.web_search_url,
         tool_entry=tool_ctx.web_search_tool_entry,
+        auth_token=tool_ctx.web_search_auth_token,
     ) as web_backend:
         kwargs = adapter.inject_hints(call_kwargs, web_backend.purpose_hints(), header=tool_ctx.tools_header)
         return await adapter.run_tool_loop(kwargs, web_backend, tool_ctx.max_tool_iterations, on_first_response)
@@ -899,6 +909,7 @@ async def open_stream(
     web_search_backend = _build_web_search_backend(
         base_url=tool_ctx.web_search_url,
         tool_entry=tool_ctx.web_search_tool_entry,
+        auth_token=tool_ctx.web_search_auth_token,
     )
     await web_search_backend.__aenter__()  # may raise WebSearchNotReachableError
     return _eager_backend_stream(adapter, kwargs, web_search_backend, tool_ctx)
@@ -1196,6 +1207,7 @@ async def run_streaming_with_fallback(
                 _build_web_search_backend(
                     base_url=tool_ctx.web_search_url,
                     tool_entry=tool_ctx.web_search_tool_entry,
+                    auth_token=tool_ctx.web_search_auth_token,
                 ),
             )
     except BaseException:
--- a/src/gateway/api/routes/_pipeline.py
+++ b/src/gateway/api/routes/_pipeline.py
@@ -633,7 +633,14 @@ async def prepare_gateway_tools(
             # Standalone mode has no platform to consult.
             if ctx.platform_mode:
                 assert ctx.user_token is not None  # guaranteed by the platform-mode preamble
-                web_search_auth_token = ctx.config.platform_token
+                # Forward the platform token only when the search backend IS the
+                # platform (its URL is under the platform base URL the gateway
+                # already trusts this token with for resolve). Never leak this
+                # high-privilege credential to a bundled SearXNG or a third-party
+                # adapter that an operator happened to point GATEWAY_WEB_SEARCH_URL at.
+                platform_base = (ctx.config.platform.get("base_url") or "").rstrip("/")
+                if platform_base and web_search_url.startswith(platform_base):
+                    web_search_auth_token = ctx.config.platform_token
                 web_search_policy = await _resolve_platform_web_search(
                     config=ctx.config,
                     user_token=ctx.user_token,
--- a/src/gateway/api/routes/_pipeline.py
+++ b/src/gateway/api/routes/_pipeline.py
@@ -36,7 +36,7 @@ from contextlib import AsyncExitStack
 from datetime import UTC, datetime
 from enum import Enum, auto
 from typing import Any, Generic, Protocol, TypeVar
-from urllib.parse import urlparse
+from urllib.parse import ParseResult, urlparse
 
 from any_llm import AnyLLM, LLMProvider
 from any_llm.exceptions import AnyLLMError
@@ -155,6 +155,19 @@ class ErrorKind(Enum):
     PERMISSION = auto()
 
 
+_DEFAULT_PORTS = {"http": 80, "https": 443}
+
+
+def _normalized_origin(parsed: ParseResult) -> tuple[str, str | None, int | None]:
+    """(scheme, host, port) with the scheme's default port filled in.
+
+    So ``https://h`` and ``https://h:443`` compare equal (and ``http`` / ``:80``),
+    rather than failing on ``None != 443`` and silently not forwarding the token.
+    """
+    port = parsed.port if parsed.port is not None else _DEFAULT_PORTS.get(parsed.scheme)
+    return (parsed.scheme, parsed.hostname, port)
+
+
 def web_search_url_targets_platform(web_search_url: str, platform_base_url: str | None) -> bool:
     """True when ``web_search_url`` is the platform itself (same origin, under its base path).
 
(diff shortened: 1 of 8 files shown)


agpituk:
Thanks — both addressed in 9a044ee:

- **Token scoping (Copilot):** the platform token is now forwarded only when `GATEWAY_WEB_SEARCH_URL` is under the platform `base_url` the gateway already trusts it with for resolve. A bundled SearXNG or third-party adapter (and all standalone setups) receive no token. Added a test for the platform-URL case and one asserting a non-platform backend gets none.
- **Test state (CodeRabbit):** `_FakeWebSearchBackend.last_auth_token` is now reset before the request.

src/gateway/api/routes/_pipeline.py
```diff
@@ -626,6 +633,14 @@ async def prepare_gateway_tools(
             # Standalone mode has no platform to consult.
             if ctx.platform_mode:
                 assert ctx.user_token is not None  # guaranteed by the platform-mode preamble
+                # Forward the platform token only when the search backend IS the
+                # platform (its URL is under the platform base URL the gateway
+                # already trusts this token with for resolve). Never leak this
+                # high-privilege credential to a bundled SearXNG or a third-party
+                # adapter that an operator happened to point GATEWAY_WEB_SEARCH_URL at.
+                platform_base = (ctx.config.platform.get("base_url") or "").rstrip("/")
+                if platform_base and web_search_url.startswith(platform_base):
```
dpoulopoulos: **[major]** The token-scoping gate uses a raw string-prefix check (`web_search_url.startswith(platform_base)`) rather than a URL-origin comparison. With the default/canonical base URL (`https://api.otari.ai/api/v1`, which includes a path) the authority is pinned, so this is safe. But if an operator sets `PLATFORM_BASE_URL` to a path-less origin (e.g. `https://api.otari.ai`), a confusable web-search URL such as `https://api.otari.ai.evil.com/...` or `https://api.otari.ai@evil.com/...` satisfies `startswith` and would leak the high-privilege platform token to a foreign host, which is exactly what this commit set out to prevent. Recommend parsing both URLs with `urllib.parse.urlparse` and comparing `(scheme, hostname, port)` (and requiring the web-search path to sit under the base path with a `/` boundary). Not exploitable in the default config, so not a merge blocker, but worth hardening since it guards a credential.

tests/unit/test_web_search_backend.py
```diff
@@ -549,3 +549,38 @@ def test_max_results_clamps_above_cap_to_max_results_cap() -> None:
...
+    """A standalone backend (no auth_token) sends no X-Gateway-Token header."""
+    transport = _patched_async_client(
+        {("searxng", "/search"): httpx.Response(200, json=SEARXNG_OK_BODY)},
+        monkeypatch,
+    )
+
+    async with WebSearchBackend(base_url="http://searxng:8080", extract_content=False) as backend:
+        await backend.call_tool(WEB_SEARCH_TOOL_NAME, {"query": "x"})
+
+    assert "X-Gateway-Token" not in transport.captured[0].headers
```
dpoulopoulos, line 586: **[nit]** Consider adding a regression test with `extract_content=True` asserting `X-Gateway-Token` is present on the `/search` request but absent on the follow-up result-page fetches. The code is already correct (headers are per-request), but such a test would lock in the no-leak behavior against future refactors, as CodeRabbit suggested.

dpoulopoulos (changes requested):
Forwards the gateway's platform token to the web-search backend as an `X-Gateway-Token` header so a platform-hosted search endpoint can authenticate the gateway. The token is threaded through `ToolContext.web_search_auth_token` and `_build_web_search_backend`, and is set only in platform mode and only when the configured web-search URL is under the platform base URL. Includes unit and integration tests covering both the forwarded and not-forwarded cases.

### General notes
- **[nit]** `docs/platform-protocol.md:24` — This PR adds a new token-bearing call surface: the platform-hosted web-search query endpoint now receives `X-Gateway-Token`. The protocol doc enumerates the gateway-to-platform endpoints that carry this header (resolve/usage) but does not mention the web-search query path. A line documenting it would keep the protocol surface complete (possibly out of scope for this PR).

agpituk:
Thanks @dpoulopoulos — addressed in 3d21d35:

- **[major] URL-origin scoping:** replaced the raw `startswith` with `web_search_url_targets_platform()`, which parses both URLs and compares `(scheme, hostname, port)` exactly, then requires the search path to sit under the base path at a `/` boundary. Confusable hosts (`api.otari.ai.evil.com`, `api.otari.ai@evil.com`) and path-prefix collisions (`/api/v1beta`) are now rejected even with a path-less `PLATFORM_BASE_URL`. Added `tests/unit/test_web_search_url_scoping.py` covering those cases.
- **[nit] no-leak regression test:** added `test_auth_token_not_leaked_to_result_page_fetches` — asserts `X-Gateway-Token` is on the `/search` call but absent on the per-result page fetches.
- **[nit] protocol doc:** added a note in `docs/platform-protocol.md` documenting the web-search query call surface and that the token is only sent when the URL shares the platform origin.

src/gateway/api/routes/_pipeline.py
```diff
@@ -154,6 +155,28 @@ class ErrorKind(Enum):
...
+    a confusable URL like ``https://api.otari.ai.evil.com`` or
+    ``https://api.otari.ai@evil.com`` would satisfy ``startswith`` and leak the
+    token. So compare the parsed (scheme, host, port) origin exactly, and require
+    the search path to sit under the base path at a ``/`` boundary.
+    """
+    if not platform_base_url:
+        return False
+    base = urlparse(platform_base_url)
+    target = urlparse(web_search_url)
+    if (target.scheme, target.hostname, target.port) != (base.scheme, base.hostname, base.port):
```
dpoulopoulos: **[nit]** The origin comparison uses `target.port`/`base.port` verbatim and does not normalize default ports. If an operator sets `GATEWAY_WEB_SEARCH_URL` with an explicit `:443` (e.g. `https://api.otari.ai:443/api/v1/...`) while `PLATFORM_BASE_URL` omits it (default -> `.port` is `None`), `443 != None` fails the gate and the token is silently not forwarded. This is fail-closed (no credential leak), so it's not a security issue, but the auth feature would quietly not engage. Consider normalizing the default port per scheme if that configuration is plausible.

tests/unit/test_web_search_url_scoping.py
```diff
@@ -0,0 +1,49 @@
...
+    [
+        # Canonical base (with path) -> the platform's own web-search endpoint.
+        ("https://api.otari.ai/api/v1/gateway/web-search", "https://api.otari.ai/api/v1"),
+        # Internal service addressing (what deployments actually use).
+        ("http://backend:8000/api/v1/gateway/web-search", "http://backend:8000/api/v1"),
+        # Trailing slash on the base must not matter.
+        ("http://backend:8000/api/v1/gateway/web-search", "http://backend:8000/api/v1/"),
+    ],
+)
+def test_accepts_platform_targets(web_search_url: str, platform_base: str) -> None:
```
dpoulopoulos, line 29: **[nit]** The accept-cases all use a base URL that carries a path (`/api/v1`). There's no positive case for a path-less base (e.g. base `https://api.otari.ai`, target `https://api.otari.ai/gateway/web-search`), which exercises the `base_path == ""` branch. The function handles it correctly (the negative path-less cases confirm the origin check), but a positive assertion would lock in that branch.

src/gateway/api/routes/_pipeline.py
```diff
@@ -626,6 +656,13 @@ async def prepare_gateway_tools(
             # Standalone mode has no platform to consult.
             if ctx.platform_mode:
                 assert ctx.user_token is not None  # guaranteed by the platform-mode preamble
+                # Forward the platform token only when the search backend IS the
+                # platform (its URL is under the platform base URL the gateway
+                # already trusts this token with for resolve). Never leak this
+                # high-privilege credential to a bundled SearXNG or a third-party
+                # adapter that an operator happened to point GATEWAY_WEB_SEARCH_URL at.
+                if web_search_url_targets_platform(web_search_url, ctx.config.platform.get("base_url")):
+                    web_search_auth_token = ctx.config.platform_token
```
dpoulopoulos, line 679: **[nit]** The forwarded `X-Gateway-Token` reuses the high-privilege `config.platform_token` (the same credential used for resolve). This is acceptable given the strict same-origin gate (the token only reaches a host already trusted with it for resolve), but a dedicated, narrower-scope web-search token would limit blast radius if the search endpoint were ever compromised independently. Out of scope for this PR; noting for the record since the original Copilot comment raised the alternative.

dpoulopoulos (dismissed):
In platform mode, the gateway now forwards its platform token as an `X-Gateway-Token` header on the web-search backend's `/search` call, so a platform-hosted search endpoint can authenticate the gateway. The token is threaded through `ToolContext.web_search_auth_token` and `_build_web_search_backend`, and is forwarded **only** when the configured web-search URL targets the platform origin itself; standalone and third-party backends receive no token.

**Prior review feedback: all addressed.** The [major] raw `startswith` gate is replaced by `web_search_url_targets_platform()` (parsed origin + `/`-boundary path check); the Copilot token-leak concern is now gated behind that check; the [nit] regression test for `extract_content=True` was added; the new token-bearing call surface is documented in `docs/platform-protocol.md`; and the CodeRabbit `last_auth_token` reset is in place. Verified in source that the token rides only `/search` (never result-page fetches) and is never logged. Unit tests (59), the two platform-mode integration tests, `ruff`, and `mypy` all pass locally.

agpituk:
Thanks for the approval @dpoulopoulos — folded in the two quick nits (e645882):

- **Port normalization:** the origin gate now fills in the scheme's default port (http=80, https=443) before comparing, so `:443` explicit vs omitted no longer silently fails the gate. Added positive tests for that and for a path-less base URL.
- **Dedicated web-search token:** agreed it's a good future hardening — leaving as a follow-up per your note (out of scope here).

njbrake (approved):

Approving. Focused defense-in-depth change: forwards the platform token as `X-Gateway-Token` on the web-search backend's `/search` call, gated so it only ever reaches the platform itself (exact scheme/host/port origin, under the base path), never a bundled SearXNG or third-party adapter.

What I checked:
- Header naming matches the existing `X-Gateway-Token` used by all the `_platform.py` resolve calls.
- `web_search_url` is narrowed to `str` by the not-configured raise before the gate, so the typed helper signature holds; token is set only inside the `platform_mode` branch.
- The token is threaded into all three backend-construction sites (non-stream, stream, streaming-with-fallback).
- Strong adversarial test coverage on the scoping logic: confusable hosts (`api.otari.ai.evil.com`, userinfo `@evil.com`), path-prefix collisions, scheme mismatch, and default-port normalization. Plus a regression test that the header rides only `/search` and never the per-result page fetches.
- 62 relevant unit tests pass locally.

The one CodeRabbit nit (resetting `last_auth_token` before the merge test) is already addressed. No blocking issues.

Two non-blocking notes: `web_search_url_targets_platform` could arguably live in `url_safety.py` alongside the other origin logic, and this code omits the header when the token is unset whereas resolve sends an empty string; the omit behavior here is cleaner.

PR #190, opened 2026-06-22
feat(memory): recall and store memory around chat completions

### LINKED ISSUE #189: [FEATURE]: Persistent memory for chat completions (platform mode)
## Description & Motivation

When the gateway runs in platform mode, chat completions it routes should be able to benefit from **persistent memory**: the assistant should remember durable facts a user shared in earlier conversations and use them in later ones, instead of treating every request as a blank slate.

Today, every completion is stateless. A user can tell the assistant something meaningful in one request "I prefer metric units", "our service is written in FastAPI"), and on the very next request the gateway hands the model a clean prompt with none of that context. Users repeat themselves and assistants built on the gateway feel amnesiac.

From a caller's point of view, a completion should be able to:

- **Recall** relevant things already known about the user and bring them to bear on the answer, without the user repeating themselves.
- **Remember** durable new facts from the exchange so they're available next time.

### Requirements / constraints

- **Opt-in and off by default.** No behavior change for existing deployments unless memory is explicitly enabled.
- **Never break or noticeably slow chat.** If memory is unavailable or slow, the completion must proceed normally — memory is a best-effort enhancement, not a hard dependency.
- **Covers the normal completion paths**, both streaming and non-streaming.
- **Platform mode only.** Memory is a platform-managed capability; standalone deployments must be unaffected (no behavior change, no overhead).

## Acceptance Criteria

- [ ] In platform mode, with memory enabled for a workspace, a completion is enriched with relevant remembered context before the model answers.
- [ ] Durable facts from a completion are captured for future recall.
- [ ] Memory is opt-in and off by default; disabling it (or any memory failure) leaves completions behaving exactly as they do today.
- [ ] Both streaming and non-streaming completions are covered.
- [ ] No memory behavior or added latency in standalone mode.

dpoulopoulos:
## Description

Wires **persistent memory** into chat completions in platform mode: the gateway recalls relevant remembered context before the model answers and stores durable new facts afterward. It is **off by default** (`OTARI_MEMORY_ENABLED`), best-effort, and covers both streaming and non-streaming completions.

The point is that assistants built on the gateway feel continuous across sessions instead of stateless. The platform owns configuration, storage, privacy, and accounting (companion PR `mozilla-ai/otari-ai#1148`); the gateway just brackets each completion with recall + remember.

What's in it:

- `OTARI_MEMORY_ENABLED` master switch (default off → no behavior change for existing deployments).
- A best-effort platform memory client + prompt-injection helpers that swallow errors/timeouts, so memory never breaks or noticeably slows a chat.
- Recall before dispatch (covers every path) and a fire-and-forget remember afterward, for both non-streaming and streaming completions (the streaming path accumulates the assistant text and stores only on normal completion).

No memory behavior or added latency in standalone mode.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Closes #189

## Checklist

- [x] If the API contract changed, I regenerated the OpenAPI spec (`uv run python scripts/generate_openapi.py`). (No contract change: memory is internal, the spec is unchanged.)

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:** Claude Code (Claude Opus 4.8)

The memory integration was implemented end-to-end with Claude Code; the human author directed the work and reviews the result.

## Summary
- Added persistent “platform memory” to platform-mode chat completions: the gateway recalls relevant facts before dispatch and saves durable facts after completion.
- Integrated memory into both streaming and non-streaming platform responses, with best-effort behavior so memory issues never fail the chat.
- Memory behavior is gated to platform mode only and controlled by both gateway and platform `memory_enabled` flags (with operator override support).
- Added environment-configurable memory timeouts and updated the hybrid-mode provider-keys resolve schema to include `memory_enabled`.
- Added/expanded unit and integration tests covering injection, recall/remember failure handling, malformed timeout parsing, streaming settlement behavior, and end-to-end platform wiring.

## Technical notes
- Pre-dispatch recall: conditional call to `/gateway/memory/recall`, then recalled facts are injected as system “memory facts” content ahead of the user prompt.
- Post-dispatch remember:
  - Non-streaming: schedules a fire-and-forget `/gateway/memory/remember` write using the latest user turn + assistant reply.
  - Streaming: captures streamed assistant text, then triggers the remember write only after normal completion (not on early disconnect).
- Reliability: all recall/remember operations are best-effort; timeouts/errors are swallowed.
- Timeout configuration: `PLATFORM_MEMORY_RECALL_TIMEOUT_MS` and `PLATFORM_MEMORY_REMEMBER_TIMEOUT_MS` are parsed safely with defaults (recall defaulting to 8s; remember defaulting to 10s when malformed).

### DIFF
--- a/src/gateway/api/routes/_helpers.py
+++ b/src/gateway/api/routes/_helpers.py
@@ -124,6 +124,45 @@ def latest_user_text(messages: Sequence[Any]) -> str:
     return ""
 
 
+MEMORY_FACTS_HEADER = (
+    "Relevant memory about the user (use when helpful; do not mention that you are recalling memory):"
+)
+
+
+def inject_memory_facts(messages: list[dict[str, Any]], facts: list[str]) -> list[dict[str, Any]]:
+    """Prepend or extend the system message with recalled memory facts.
+
+    Mirrors :func:`gateway.services.mcp_loop.inject_purpose_hints`: augments an existing
+    system message, or inserts one when absent. Returns a new list; the input is not
+    mutated. A no-op when ``facts`` is empty.
+    """
+    if not facts:
+        return messages
+    block = "\n".join([MEMORY_FACTS_HEADER, *(f"- {fact}" for fact in facts)])
+    out = list(messages)
+    if out and isinstance(out[0], dict) and out[0].get("role") == "system":
+        existing = out[0].get("content") or ""
+        out[0] = {**out[0], "content": f"{existing}\n\n{block}" if existing else block}
+    else:
+        out.insert(0, {"role": "system", "content": block})
+    return out
+
+
+def build_remember_messages(messages: Sequence[Any], assistant_text: str) -> list[dict[str, str]]:
+    """Build the minimal exchange to store: the latest user turn plus the assistant reply.
+
+    Keeping it to the new turn (rather than the whole history) avoids re-extracting facts
+    the platform has already processed. Returns an empty list when there is nothing to store.
+    """
+    out: list[dict[str, str]] = []
+    user_text = latest_user_text(messages)
+    if user_text:
+        out.append({"role": "user", "content": user_text})
+    if assistant_text:
+        out.append({"role": "assistant", "content": assistant_text})
+    return out
+
+
 async def apply_input_guardrails(
     guardrails: list[GuardrailConfig] | None,
     input_text: str,
--- a/src/gateway/api/routes/_pipeline.py
+++ b/src/gateway/api/routes/_pipeline.py
@@ -31,7 +31,7 @@ from __future__ import annotations
 import asyncio
 import time
 import uuid
-from collections.abc import AsyncIterator, Awaitable, Callable
+from collections.abc import AsyncIterator, Awaitable, Callable, Coroutine
 from contextlib import AsyncExitStack
 from datetime import UTC, datetime
 from enum import Enum, auto
@@ -1484,6 +1484,8 @@ async def run_streaming_with_fallback(
     rate_limit_info: RateLimitInfo | None,
     tool_ctx: ToolContext,
     session_label: str | None = None,
+    extract_stream_text: Callable[[ChunkT], str | None] | None = None,
+    on_memory_settled: Callable[[str], Coroutine[Any, Any, None]] | None = None,
 ) -> StreamingResponse:
     """Multi-attempt streaming for hybrid-mode requests.
 
@@ -1613,6 +1615,9 @@ async def run_streaming_with_fallback(
     if pool_for_loop is not None:
         stream_to_return = _stream_with_stack_cleanup(stream, backend_stack)
 
+    if extract_stream_text is not None and on_memory_settled is not None:
+        stream_to_return = _stream_with_memory_capture(stream_to_return, extract_stream_text, on_memory_settled)
+
     return build_streaming_response(
         adapter=adapter,
         stream=stream_to_return,
@@ -1640,6 +1645,43 @@ async def _stream_with_stack_cleanup(
             yield chunk
     finally:
         await backend_stack.aclose()
+
+
+def _swallow_memory_task_exception(task: "asyncio.Task[None]") -> None:
+    """Done-callback that retrieves a detached memory-write task's exception so it
+    is logged rather than leaking as an unretrieved-task-exception warning."""
+    if task.cancelled():
+        return
+    exc = task.exception()
+    if exc is not None:
+        logger.warning("Background memory write failed: %s", exc)
+
+
+async def _stream_with_memory_capture(
+    stream: AsyncIterator[ChunkT],
+    extract_text: Callable[[ChunkT], str | None],
+    on_settled_text: Callable[[str], Coroutine[Any, Any, None]],
+) -> AsyncIterator[ChunkT]:
+    """Accumulate streamed assistant text and, on normal completion, fire a
+    best-effort memory write. The write is skipped on mid-stream cancellation
+    (client disconnect), and scheduled fire-and-forget so it never delays the
+    final bytes, mirroring the streaming usage-report pattern."""
+    parts: list[str] = []
+    completed = False
+    try:
+        async for chunk in stream:
+            piece = extract_text(chunk)
+            if piece:
+                parts.append(piece)
+            yield chunk
+        completed = True
+    finally:
+        if completed and parts:
+            task = asyncio.create_task(on_settled_text("".join(parts)))
+            # Retrieve the result so a raised memory write surfaces as a log line
+            # instead of an "Task exception was never retrieved" warning, keeping
+            # best-effort persistence silent on the response path.
+            task.add_done_callback(_swallow_memory_task_exception)
 
 
 async def run_platform_non_stream(
--- a/src/gateway/api/routes/_platform.py
+++ b/src/gateway/api/routes/_platform.py
@@ -675,6 +675,85 @@ async def _resolve_platform_code_execution(
     )
 
 
+def _coerce_timeout_ms(value: Any, default_ms: int) -> int:
+    """Parse a configured timeout (milliseconds) into a positive int, falling
+    back to ``default_ms`` for missing, non-numeric, or non-positive values so a
+    misconfigured timeout never breaks a best-effort memory call."""
+    try:
+        parsed = int(value)
+    except (TypeError, ValueError):
+        return default_ms
+    return parsed if parsed > 0 else default_ms
+
+
+async def _recall_platform_memory(
+    config: GatewayConfig,
+    user_token: str,
+    query: str,
+) -> list[str]:
+    """Recall relevant memory facts for the user-token's workspace.
+
+    Best-effort and on the hot path: any failure (misconfigured platform, timeout,
+    network error, non-200, or a malformed body) returns no facts so a slow or
+    unavailable memory service never breaks the completion. The platform also
+    returns an empty list when the workspace has memory disabled.
+    """
+    platform_base_url = config.platform.get("base_url")
+    if not platform_base_url or not query.strip():
+        return []
+
+    timeout_ms = _coerce_timeout_ms(config.platform.get("memory_recall_timeout_ms", 2000), 2000)
+    url = _platform_url(platform_base_url, "/gateway/memory/recall")
+    headers = {
+        "X-Gateway-Token": config.platform_token or "",
+        "X-User-Token": user_token,
+    }
+
+    try:
+        response = await _post_platform(
+            url=url, headers=headers, body={"query": query}, timeout_seconds=timeout_ms / 1000
+        )
+    except httpx.HTTPError:
+        return []
+
+    if response.status_code != 200:
+        return []
+    try:
+        payload = response.json()
+    except ValueError:
+        return []
+    facts = payload.get("facts") if isinstance(payload, dict) else None
+    if not isinstance(facts, list):
+        return []
+    return [fact for fact in facts if isinstance(fact, str)]
+
+
+async def _remember_platform_memory(
+    config: GatewayConfig,
+    user_token: str,
+    messages: list[dict[str, Any]],
+) -> None:
+    """Store durable facts from a completed exchange. Best-effort and fire-and-forget:
+    failures are swallowed so they never affect the user's response path."""
+    platform_base_url = config.platform.get("base_url")
+    if not platform_base_url or not messages:
+        return
+
+    timeout_ms = _coerce_timeout_ms(config.platform.get("memory_remember_timeout_ms", 10000), 10000)
+    url = _platform_url(platform_base_url, "/gateway/memory/remember")
+    headers = {
+        "X-Gateway-Token": config.platform_token or "",
+        "X-User-Token": user_token,
+    }
+
+    try:
+        await _post_platform(
+            url=url, headers=headers, body={"messages": messages}, timeout_seconds=timeout_ms / 1000
+        )
+    except httpx.HTTPError:
+        return
+
+
 async def _report_platform_usage(
     config: GatewayConfig,
     correlation_id: str,
--- a/src/gateway/api/routes/_platform.py
+++ b/src/gateway/api/routes/_platform.py
@@ -97,6 +97,11 @@ class ResolvedRoute(BaseModel):
     request_id: str
     fallback_enabled: bool
     attempts: list[ResolvedAttempt]
+    # Whether the workspace has persistent memory enabled. The platform is authoritative;
+    # the gateway uses this to decide whether to make the hot-path recall call at all.
+    # Defaults to True so a gateway pointed at an older platform (which omits the field)
+    # still attempts recall, relying on the platform's empty-when-off short-circuit.
+    memory_enabled: bool = True
 
 
 class _AttemptFailure(NamedTuple):
@@ -439,6 +444,8 @@ def _parse_resolve_payload(payload: dict[str, Any]) -> ResolvedRoute:
             request_id=str(payload["request_id"]),
             fallback_enabled=bool(payload.get("fallback_enabled", False)),
             attempts=attempts,
+            # Absent on older platforms; default True so recall still runs (see ResolvedRoute).
+            memory_enabled=bool(payload.get("memory_enabled", True)),
         )
 
     correlation_id = str(payload["correlation_id"])
@@ -456,6 +463,7 @@ def _parse_resolve_payload(payload: dict[str, Any]) -> ResolvedRoute:
                 managed=bool(payload.get("managed", False)),
             )
         ],
+        memory_enabled=bool(payload.get("memory_enabled", True)),
     )
 
 
@@ -702,7 +710,7 @@ async def _recall_platform_memory(
     if not platform_base_url or not query.strip():
         return []
 
-    timeout_ms = _coerce_timeout_ms(config.platform.get("memory_recall_timeout_ms", 2000), 2000)
+    timeout_ms = _coerce_timeout_ms(config.platform.get("memory_recall_timeout_ms", 8000), 8000)
     url = _platform_url(platform_base_url, "/gateway/memory/recall")
     headers = {
         "X-Gateway-Token": config.platform_token or "",
--- a/src/gateway/api/routes/chat.py
+++ b/src/gateway/api/routes/chat.py
@@ -1,6 +1,6 @@
 import asyncio
 import uuid
-from collections.abc import AsyncIterator, Callable
+from collections.abc import AsyncIterator, Callable, Coroutine
 from typing import Annotated, Any
 
 import httpx
@@ -17,7 +17,7 @@ from pydantic import Field, field_validator
 from sqlalchemy.ext.asyncio import AsyncSession
 
 from gateway.api.deps import get_config, get_db_if_needed, get_log_writer
-from gateway.api.routes._helpers import latest_user_text
+from gateway.api.routes._helpers import build_remember_messages, inject_memory_facts, latest_user_text
 from gateway.api.routes._normalize import normalize_request_messages
 from gateway.api.routes._pipeline import (
     ALL_PROVIDERS_FAILED_DETAIL,
@@ -39,7 +39,11 @@ from gateway.api.routes._pipeline import (
     run_standalone_non_stream,
     run_streaming_with_fallback,
 )
-from gateway.api.routes._platform import ResolvedAttempt
+from gateway.api.routes._platform import (
+    ResolvedAttempt,
+    _recall_platform_memory,
+    _remember_platform_memory,
+)
 from gateway.api.routes._schema_derive import SESSION_LABEL_DESC, SESSION_LABEL_MAX_LENGTH, derive_request_base
 from gateway.api.routes._tools import _strip_gateway_fields
 from gateway.core.config import GatewayConfig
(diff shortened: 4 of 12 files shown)


src/gateway/api/routes/_platform.py
```diff
@@ -583,6 +583,74 @@ async def _resolve_platform_web_search(
...
+    headers = {
+        "X-Gateway-Token": config.platform_token or "",
+        "X-User-Token": user_token,
+    }
+
+    try:
+        response = await _post_platform(
+            url=url, headers=headers, body={"query": query}, timeout_seconds=timeout_ms / 1000
+        )
+    except (httpx.TimeoutException, httpx.NetworkError):
```
dpoulopoulos: **[major]** Recall promises in its docstring that "any failure ... never breaks the completion", but it only catches `(httpx.TimeoutException, httpx.NetworkError)`. Other `httpx` request errors escape: `RemoteProtocolError` ("server disconnected without sending a response"), `ProxyError`, `UnsupportedProtocol`, and `InvalidURL`. The unguarded `int(config.platform.get("memory_recall_timeout_ms", 2000))` at line 602 can also raise `ValueError`/`TypeError` on a misconfigured timeout. Because the call site in `chat.py:362` (`recalled = await _recall_platform_memory(...)`) is **not** wrapped in try/except, any of these propagates and turns a chat that would have succeeded into a 500 -- the opposite of best-effort. The common cases (timeouts, connect/read errors) are handled, so this only bites on the error long-tail; but a memory-service deploy or overload that emits `RemoteProtocolError` would take down every memory-enabled chat. Suggest a blanket guard for this best-effort helper (e.g. wrap the body after the cheap guards in `try/except Exception: return []`, which also covers the `int()` parse), or wrap the call site in `chat.py`. Worth fixing before setting `OTARI_MEMORY_ENABLED=true` in production. The same narrow catch exists in `_remember_platform_memory` (line 650).

dpoulopoulos (commented):
Adds opt-in persistent memory to platform-mode chat completions: a new `memory_enabled` flag (off by default) gates a best-effort recall-before-dispatch + remember-after flow. Message helpers inject recalled facts into the system prompt and build the turn to store; two platform clients (`_recall_platform_memory`, `_remember_platform_memory`) talk to `/gateway/memory/{recall,remember}`; a streaming wrapper accumulates assistant text and fires a fire-and-forget remember on normal completion. Both streaming and non-streaming paths are wired. Standalone mode is unaffected.

src/gateway/api/routes/_pipeline.py
```diff
@@ -1434,6 +1439,29 @@ async def _stream_with_stack_cleanup(
...
+    try:
+        async for chunk in stream:
+            piece = extract_text(chunk)
+            if piece:
+                parts.append(piece)
+            yield chunk
+        completed = True
+    finally:
+        if completed and parts:
+            asyncio.create_task(on_settled_text("".join(parts)))
```
dpoulopoulos: **[minor]** `asyncio.create_task(on_settled_text(...))` is fire-and-forget with no stored reference and no done-callback. If `on_settled_text` (the `_remember` closure -> `_remember_platform_memory`) raises anything outside `(TimeoutException, NetworkError)` -- same gap as the recall finding -- it surfaces as an "Task exception was never retrieved" warning rather than being silently swallowed as the docstring promises. (The non-streaming path has the analogous risk: a raised `background_tasks` remember is logged by Starlette.) This mirrors the existing `_report_platform_usage` create_task pattern (lines 1020/1080), so it's consistent with the codebase -- but broadening the remember catch to `httpx.HTTPError` and/or adding `task.add_done_callback(...)` that swallows would make the "never affects the response path" contract actually hold. Per Python docs, keeping a reference also prevents the task being GC'd mid-flight.

src/gateway/api/routes/chat.py
```diff
@@ -303,6 +354,15 @@ async def _normalize(
         tools_header=request.tools_header,
     )
 
+    # Memory recall (platform mode, best-effort). Inject recalled facts into the system
+    # message before dispatch so every downstream path (stream/non-stream, tool/no-tool)
+    # sees them. A failure or a disabled workspace yields no facts and never blocks chat.
+    memory_user_token = ctx.user_token if (config.memory_enabled and ctx.platform_mode) else None
+    if memory_user_token:
+        recalled = await _recall_platform_memory(config, memory_user_token, latest_user_text(request.messages))
```
dpoulopoulos: **[minor]** Recall is awaited synchronously before dispatch, so when the memory service is slow it adds up to `memory_recall_timeout_ms` (default 2000ms) to time-to-first-token on every memory-enabled platform request. This is inherent to recall-before-dispatch (the facts must be in the prompt) and the timeout is configurable, so it's a reasonable trade-off -- but it does qualify the PR's "never noticeably slows a chat" claim under a degraded memory service. Worth documenting the operator-visible latency budget and perhaps a more aggressive default than 2s.

src/gateway/api/routes/_helpers.py
```diff
@@ -123,6 +123,45 @@ def latest_user_text(messages: Sequence[Any]) -> str:
...
+    system message, or inserts one when absent. Returns a new list; the input is not
+    mutated. A no-op when ``facts`` is empty.
+    """
+    if not facts:
+        return messages
+    block = "\n".join([MEMORY_FACTS_HEADER, *(f"- {fact}" for fact in facts)])
+    out = list(messages)
+    if out and isinstance(out[0], dict) and out[0].get("role") == "system":
+        existing = out[0].get("content") or ""
+        out[0] = {**out[0], "content": f"{existing}\n\n{block}" if existing else block}
```
dpoulopoulos, line 145: **[nit]** When the first message is a system message whose `content` is a list of content parts (multimodal system prompt), `f"{existing}\n\n{block}"` stringifies the list via its Python `repr`, corrupting the system content. System content is almost always a plain string and this exactly mirrors the existing `inject_purpose_hints` helper, so it's low-risk and consistent -- just flagging the shared edge case.

tests/unit/test_memory_platform.py
```diff
@@ -0,0 +1,119 @@
...
+async def test_recall_non_200_yields_no_facts(monkeypatch: pytest.MonkeyPatch) -> None:
+    async def fake_post(**kwargs: Any) -> httpx.Response:
+        return httpx.Response(500, json={"detail": "boom"})
+
+    monkeypatch.setattr(platform_module, "_post_platform", fake_post)
+    assert await _recall_platform_memory(_config(), "tk", "q") == []
+
+
+@pytest.mark.asyncio
+async def test_recall_timeout_is_swallowed(monkeypatch: pytest.MonkeyPatch) -> None:
```
dpoulopoulos, line 67: **[nit]** Tests cover timeout, non-200, and malformed-body swallowing, but not a generic non-timeout/non-network `httpx` error (e.g. `RemoteProtocolError`) or a non-numeric `memory_recall_timeout_ms`. Adding those would lock in the best-effort contract and would have surfaced the gap in the major finding (CodeRabbit raised the malformed-timeout variant too).

dpoulopoulos:
Thanks for the careful review. Pushed fixups addressing the best-effort gaps:

- Timeout parsing no longer uses a bare `int(...)`. Added a `_coerce_timeout_ms` helper that falls back to the default for missing, non-numeric, or non-positive values, used by both recall and remember.
- Broadened the recall/remember error handling from `(TimeoutException, NetworkError)` to `httpx.HTTPError`, so the long tail (`RemoteProtocolError`, `ProxyError`, `InvalidURL`, etc.) can't turn a would-succeed chat into a 500.
- Wrapped the recall call site in chat.py in a try/except that logs and proceeds without facts, as a belt-and-suspenders boundary.
- The streaming remember task now keeps its reference and attaches a done-callback that retrieves and logs any exception, so a failed background write doesn't leak a "Task exception was never retrieved" warning.
- Added regression tests: malformed timeout values (recall + remember) asserting the default fallback, a `RemoteProtocolError` swallowed on both paths, and a raising streamed settled-callback that doesn't propagate.

Two points I left as-is, with reasoning:
- Recall-before-dispatch adds up to the recall timeout to TTFT under a degraded memory service. That's inherent (facts must be in the prompt before the call) and already bounded by a configurable timeout, so no change beyond the safer parsing above.
- The multimodal-system-content `repr` edge case in `inject_memory_facts` mirrors the existing `inject_purpose_hints` helper; system content is effectively always a string, and changing only this helper would diverge from that pattern.

ruff, mypy (strict), and the full unit suite are green.

src/gateway/api/routes/chat.py
```diff
@@ -303,6 +354,18 @@ async def _normalize(
         tools_header=request.tools_header,
     )
 
+    # Memory recall (platform mode, best-effort). Inject recalled facts into the system
+    # message before dispatch so every downstream path (stream/non-stream, tool/no-tool)
+    # sees them. A failure or a disabled workspace yields no facts and never blocks chat.
+    memory_user_token = ctx.user_token if (config.memory_enabled and ctx.platform_mode) else None
+    if memory_user_token:
+        try:
+            recalled = await _recall_platform_memory(config, memory_user_token, latest_user_text(request.messages))
```
dpoulopoulos, line 378: **[minor]** Recall is awaited synchronously before dispatch, so under a slow/degraded memory service it adds up to `memory_recall_timeout_ms` (default **2000ms**) to time-to-first-token on every memory-enabled platform request. This is inherent to recall-before-dispatch (the facts must be in the prompt before the call) and is a reasonable trade-off, but the operator-facing latency budget isn't documented anywhere: `memory_recall_timeout_ms` / `memory_remember_timeout_ms` are read from the `config.platform` dict but aren't surfaced in the `memory_enabled` field description, `config.example.yml`, or the docs. Worth documenting the latency budget (and perhaps a more aggressive recall default than 2s) so operators understand the cost of enabling it. Carried over from the prior review; non-blocking.

src/gateway/api/routes/chat.py
```diff
@@ -303,6 +354,18 @@ async def _normalize(
         tools_header=request.tools_header,
     )
 
+    # Memory recall (platform mode, best-effort). Inject recalled facts into the system
+    # message before dispatch so every downstream path (stream/non-stream, tool/no-tool)
+    # sees them. A failure or a disabled workspace yields no facts and never blocks chat.
+    memory_user_token = ctx.user_token if (config.memory_enabled and ctx.platform_mode) else None
```
dpoulopoulos: **[minor]** No integration test exercises the wired memory flow in `chat_completions`. The unit tests cover the helpers, the platform client, and the streaming-capture wrapper in isolation, but the actual wiring here, gating on `config.memory_enabled and ctx.platform_mode`, recall then `inject_memory_facts`, the best-effort `try/except`, `_schedule_memory_remember` on the non-streaming path (line 474), and `on_memory_settled` on the streaming path (line 405), is untested end to end. An integration test in `tests/integration/test_platform_mode_chat.py` (memory off = no platform memory calls; memory on = recall injected and remember fired) would lock in this behavior and catch wiring regressions the unit tests can't see.

src/gateway/api/routes/_helpers.py
dpoulopoulos, line 145: **[nit]** When the first message is a system message whose `content` is a *list* of content parts (a multimodal system prompt), `f"{existing}\n\n{block}"` stringifies the list via its Python `repr`, corrupting the system content. System content is almost always a plain string, and this exactly mirrors the existing `inject_purpose_hints` helper (which has the same edge case), so it's low-risk and consistent, just flagging the shared gap. Carried over from the prior review.

src/gateway/api/routes/_pipeline.py
```diff
@@ -1434,6 +1439,43 @@ async def _stream_with_stack_cleanup(
...
+    try:
+        async for chunk in stream:
+            piece = extract_text(chunk)
+            if piece:
+                parts.append(piece)
+            yield chunk
+        completed = True
+    finally:
+        if completed and parts:
+            task = asyncio.create_task(on_settled_text("".join(parts)))
```
dpoulopoulos, line 1680: **[nit]** `asyncio.create_task(...)` keeps no strong reference; only `add_done_callback` is attached. Per the asyncio docs the event loop holds only a weak reference to tasks, so a detached task can in theory be garbage-collected before it finishes. In practice the remember runs on the next loop tick and this mirrors the existing `_report_platform_usage` create_task pattern, so it's consistent with the codebase and low-risk. The exception-leak concern from the prior review is fully resolved by `_swallow_memory_task_exception`; this is just the remaining GC subtlety. Optional: hold the task in a set and discard it in the done-callback.

dpoulopoulos (commented):
Adds opt-in, best-effort persistent memory to platform-mode chat completions, gated by a new `memory_enabled` config flag (`OTARI_MEMORY_ENABLED`, default off). Before dispatch the gateway recalls facts from the platform (`/gateway/memory/recall`) and injects them into the system prompt; after a completion it stores the new turn fire-and-forget (`/gateway/memory/remember`) for both non-streaming (background task) and streaming (a capture wrapper that accumulates assistant text and persists only on normal completion). Standalone mode and the `/messages` and `/responses` endpoints are unaffected. This is a re-review after the author addressed the previous round of feedback.

**Prior feedback status: resolved.** All actionable items from the previous review (CodeRabbit's 3 plus the prior `[major]`/`[minor]` review's 5) are addressed in the current code: the narrow `(TimeoutException, NetworkError)` catch was broadened to `httpx.HTTPError` in both recall and remember; a `_coerce_timeout_ms` helper guards timeout parsing; the recall call site in `chat.py` is wrapped in a best-effort `try/except`; the streaming `create_task` now has an exception-swallowing done-callback; and regression tests for malformed timeouts, generic httpx errors, and a raising settled-callback were added. Two low-priority items the prior review explicitly flagged as by-design/low-risk remain open (recall latency, multimodal-system-content edge case). Verified: ruff, mypy (strict), and all 32 new unit tests pass. No web UI exists for this backend feature, so Playwright verification is not applicable.

dpoulopoulos:
Pushed fixups for the second-round feedback (folded into their original commits):

- **Surfaced the memory latency budget.** The recall/remember timeouts were only readable from the `platform:` dict with no operator-visible knob. Added `PLATFORM_MEMORY_RECALL_TIMEOUT_MS` / `PLATFORM_MEMORY_REMEMBER_TIMEOUT_MS` env mappings (consistent with the existing `PLATFORM_*_TIMEOUT_MS` siblings) and expanded the `memory_enabled` field description to spell out that recall runs before dispatch (so it adds up to its timeout to TTFT) while remember is fire-and-forget, including the defaults (2000ms / 10000ms). Extended `test_load_config_platform_env_overrides` to lock in the two new mappings.
- **Added end-to-end integration coverage of the wiring.** Two tests in `test_platform_mode_chat.py`: memory off makes zero `/gateway/memory/*` calls, and memory on recalls with the latest user text, injects the fact into the system message the provider sees, then stores the user+assistant turn after the non-streaming completion.

Left as-is (both flagged low-risk/by-design in review):
- The multimodal-system-content `repr` edge case in `inject_memory_facts` mirrors the existing `inject_purpose_hints` helper; system content is effectively always a string, so changing only this helper would diverge from that pattern.
- The streaming `create_task` GC subtlety mirrors the existing `_report_platform_usage` pattern; the exception-leak concern is already handled by the done-callback, and the task runs on the next loop tick.

ruff, mypy (strict), and the memory + config test suites are green.

njbrake (changes requested):
Hi @dpoulopoulos , tagging as request changes to make sure we chat before it gets merged, I have an open question about the background for this PR in the Github Issue: https://github.com/mozilla-ai/otari/issues/189#issuecomment-4767270763 

Thanks!

src/gateway/core/config.py
```diff
@@ -227,6 +227,20 @@ class GatewayConfig(BaseSettings):
         default=True,
         description="Enable the /v1/files upload/storage endpoints (standalone mode).",
     )
+    memory_enabled: bool = Field(
```
khaledosman, line 230: The PR description says memory is "off by default (`OTARI_MEMORY_ENABLED`)" with "no behavior change for existing deployments", but this defaults to `True`. The last commit moved to the platform-authoritative design; the description was never updated. Since PRs are squash-merged and the description drives review, please update it. Note the behavior-change claim is also no longer accurate for hybrid deployments against a platform without otari-ai#1148 (see the `ResolvedRoute.memory_enabled` comment).

khaledosman (commented):
Reviewed the full diff plus enclosing code, and cross-checked the wire contract against the companion platform PR (mozilla-ai/otari-ai#1148): endpoint paths, `X-Gateway-Token`/`X-User-Token` headers, and the `{"query"}` -> `{"facts"}` / `{"messages"}` shapes all match. Streaming disconnect handling is correct (no remember on aborted or errored streams), standalone mode is untouched, and test coverage is solid.

Findings are inline. The two that most need a decision before merge: the PR description still says memory is off by default while the code defaults `memory_enabled` to on, and the absent-field default of `memory_enabled=true` makes every hybrid request against a pre-memory platform pay a wasted recall round-trip.

One doc gap that has no diff line to anchor on: `docs/configuration.md` has an otari.ai env-var table listing `PLATFORM_RESOLVE_TIMEOUT_MS` etc.; the new `PLATFORM_MEMORY_RECALL_TIMEOUT_MS` / `PLATFORM_MEMORY_REMEMBER_TIMEOUT_MS` (and `OTARI_MEMORY_ENABLED`) should be added there.

*This review was created by Claude Code.*

src/gateway/api/routes/_platform.py
```diff
@@ -97,6 +97,11 @@ class ResolvedRoute(BaseModel):
     request_id: str
     fallback_enabled: bool
     attempts: list[ResolvedAttempt]
+    # Whether the workspace has persistent memory enabled. The platform is authoritative;
+    # the gateway uses this to decide whether to make the hot-path recall call at all.
+    # Defaults to True so a gateway pointed at an older platform (which omits the field)
+    # still attempts recall, relying on the platform's empty-when-off short-circuit.
+    memory_enabled: bool = True
```
khaledosman, line 104: Defaulting to `True` when the resolve response omits `memory_enabled` makes every hybrid chat completion against a pre-memory platform POST `/gateway/memory/recall` and get a 404: a wasted hot-path round-trip per request. The stated rationale (never silently stop recalling) only protects a platform that has the memory endpoints but omits the flag, and that platform can't exist: otari-ai#1148 ships the endpoints and the resolve field together, and its response schema always includes `memory_enabled`. Default to `False` on absence; it removes the per-request call with no downside. (Applies to the doc paragraph in `docs/hybrid-mode-protocol.md` too.)

src/gateway/api/routes/_helpers.py
```diff
@@ -124,6 +124,45 @@ def latest_user_text(messages: Sequence[Any]) -> str:
...
+    Mirrors :func:`gateway.services.mcp_loop.inject_purpose_hints`: augments an existing
+    system message, or inserts one when absent. Returns a new list; the input is not
+    mutated. A no-op when ``facts`` is empty.
+    """
+    if not facts:
+        return messages
+    block = "\n".join([MEMORY_FACTS_HEADER, *(f"- {fact}" for fact in facts)])
+    out = list(messages)
+    if out and isinstance(out[0], dict) and out[0].get("role") == "system":
+        existing = out[0].get("content") or ""
```
khaledosman, line 144: OpenAI allows system `content` as a list of content parts. When `content` is a list, `f"{existing}\n\n{block}"` stringifies the Python list repr into the system prompt. Inherited from `inject_purpose_hints`, but worth fixing here: guard with `isinstance(existing, str)` and otherwise insert a separate system message (or append a text part).

src/gateway/api/routes/_platform.py
```diff
@@ -675,6 +683,85 @@ async def _resolve_platform_code_execution(
...
+    Best-effort and on the hot path: any failure (misconfigured platform, timeout,
+    network error, non-200, or a malformed body) returns no facts so a slow or
+    unavailable memory service never breaks the completion. The platform also
+    returns an empty list when the workspace has memory disabled.
+    """
+    platform_base_url = config.platform.get("base_url")
+    if not platform_base_url or not query.strip():
+        return []
+
+    timeout_ms = _coerce_timeout_ms(config.platform.get("memory_recall_timeout_ms", 8000), 8000)
```
khaledosman, line 713: 8s default on the time-to-first-token path is a lot for a best-effort feature: if the platform hangs (rather than failing fast), every request stalls up to 8s until it recovers, with no circuit breaker. The PR body promises memory never "noticeably slows" a chat; a ~2000ms default fits that much better. Operators who need more can raise `PLATFORM_MEMORY_RECALL_TIMEOUT_MS`.

src/gateway/core/config.py
```diff
@@ -551,6 +565,12 @@ def _apply_platform_env_overrides(config: dict[str, Any]) -> None:
         "PLATFORM_RESOLVE_TIMEOUT_MS": ("resolve_timeout_ms", int),
         "PLATFORM_USAGE_TIMEOUT_MS": ("usage_timeout_ms", int),
         "PLATFORM_USAGE_MAX_RETRIES": ("usage_max_retries", int),
+        # Best-effort memory timeouts (platform mode). Recall is on the hot path
+        # (added to time-to-first-token); remember is fire-and-forget after the
+        # completion. Both fall back to their defaults on a missing or invalid
```
khaledosman, line 570: Inaccurate for env values: `PLATFORM_MEMORY_RECALL_TIMEOUT_MS=abc` crashes at startup via `int(value)` in the loop below, before `_coerce_timeout_ms` ever runs. The fallback only covers YAML-sourced values. Fail-fast on a bad env var matches the other `PLATFORM_*` ints, so just scope the comment to config-file values.

src/gateway/core/config.py
```diff
@@ -551,6 +565,12 @@ def _apply_platform_env_overrides(config: dict[str, Any]) -> None:
...
+        # value (see _coerce_timeout_ms in routes/_platform.py).
+        "PLATFORM_MEMORY_RECALL_TIMEOUT_MS": ("memory_recall_timeout_ms", int),
```
khaledosman, line 572: `docs/configuration.md` documents the other `PLATFORM_*` variables in the otari.ai table; add `PLATFORM_MEMORY_RECALL_TIMEOUT_MS` and `PLATFORM_MEMORY_REMEMBER_TIMEOUT_MS` there, and document `OTARI_MEMORY_ENABLED`.

tests/integration/test_hybrid_mode_chat.py
```diff
@@ -1622,3 +1622,201 @@ async def fake_mcp_tool_loop(**kwargs: Any) -> ChatCompletion:
...
+    monkeypatch.setattr("gateway.api.routes.chat.acompletion", fake_acompletion)
+
+    with TestClient(app) as client:
+        response = client.post(
+            "/v1/chat/completions",
+            json={"model": "gpt-4o-mini", "messages": [{"role": "user", "content": "hi"}]},
+            headers={"Authorization": "Bearer user_test_token"},
+        )
+
+    reset_config()
```
khaledosman, line 1746: `reset_config()`/`reset_db()` aren't in a `try`/`finally`, so if the request raises, the global config/db override leaks into later tests. This test (and the next one) also duplicates the `platform_client` fixture but with the legacy `mode="platform"` instead of the file's `mode="hybrid"`. Extending `platform_client` to take config kwargs (or a small factory fixture with teardown after yield) fixes both.

src/gateway/api/routes/_pipeline.py
```diff
@@ -1642,6 +1647,43 @@ async def _stream_with_stack_cleanup(
...
```
khaledosman, line 1680: asyncio holds only a weak reference to tasks, so this fire-and-forget write can be garbage-collected mid-flight; the done-callback doesn't prevent that. The existing `_report_platform_usage` calls share the pattern, so low priority, but the standard fix is one line: keep tasks in a module-level set and `task.add_done_callback(_TASKS.discard)`.

src/gateway/api/routes/_pipeline.py
```diff
@@ -1642,6 +1647,43 @@ async def _stream_with_stack_cleanup(
...
+) -> AsyncIterator[ChunkT]:
+    """Accumulate streamed assistant text and, on normal completion, fire a
+    best-effort memory write. The write is skipped on mid-stream cancellation
+    (client disconnect), and scheduled fire-and-forget so it never delays the
+    final bytes, mirroring the streaming usage-report pattern."""
+    parts: list[str] = []
+    completed = False
+    try:
+        async for chunk in stream:
+            piece = extract_text(chunk)
```
khaledosman, line 1673: In tool-loop streaming this accumulates content deltas across all tool iterations, so the remembered "assistant reply" concatenates intermediate assistant turns, while the non-streaming path stores only the final message's content. Probably tolerable since the platform's fact extraction is LLM-based, but the inconsistency is worth a deliberate decision (or a note here).

PR #197, opened 2026-06-23
feat(gateway): classify upstream provider errors into specific safe responses

khaledosman:
## Description

When an upstream provider call fails, the gateway returned a blanket `502 "LLM provider error"` for everything except timeouts. This classifies the failure into a safe, more specific status + detail, so callers can tell a bad request or an unknown model apart from a genuine upstream outage, without leaking provider internals.

Surfaced while investigating mozilla-ai/otari-ai#1158 ("unusable models due to missing pricing"): the reported `502` is not a pricing problem, it is the downstream provider call failing, hidden behind a generic message.

### Classification (all details are fixed strings; the raw provider message is never included)

| Upstream | Client status | Detail |
|---|---|---|
| timeout | 504 | `LLM provider timeout` |
| 400 / 422 | 400 | `The provider rejected the request as invalid (check the model name and parameters)` |
| 404 | 404 | `The requested model was not found on the provider` |
| 401 / 403 | 502 | `The provider rejected the gateway's credentials` (gateway-config fault, never the caller's auth error) |
| 429 | 429 | `The provider rate-limited this request` (see note below) |
| 5xx / no status / unknown | 502 generic | (unchanged) |

Status is read from `exc.status_code` or `exc.response.status_code`.

### Where it applies

- **Standalone**: each format adapter's `provider_error()` consults the classifier (chat/responses use the status+detail directly; messages also maps it to the matching Anthropic `error.type`).
- **Hybrid (platform) mode** — what otari.ai runs: `run_platform_attempts` now routes **single-attempt** and **non-retryable** terminal failures through the same classifier (`_provider_failure_http_exc`), so a downstream 4xx reaches the client as 400/404/429 instead of a generic 502. A **multi-attempt fallthrough** keeps the generic `502 "All upstream providers failed"` rather than attributing one provider's status to the whole set.

### Notes / scope

- **`/v1/messages` default**: for an *unclassifiable* provider failure the messages adapter keeps its pre-existing `500` + `api_error` default (asserted by `test_messages_endpoint.py`), not the `502` the table lists for the other formats. Behavior unchanged; called out to avoid surprise.
- **429 `Retry-After`**: a provider 429 maps to a client 429 but the `(status, detail)` pair cannot carry the upstream `Retry-After`, so the exact backoff window is dropped (documented in `classify_provider_error`). A bare 429 still signals the caller to back off.
- **Streaming**: errors that occur after a stream has started surface as in-stream SSE error events, a separate path from this HTTP-status classification; left as a follow-up.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Relates to mozilla-ai/otari-ai#1158

## Checklist

- [x] I ran the Definition of Done checks locally (`ruff check`, `ruff format --check`, `mypy --strict`, and the affected unit + integration tests).
- [ ] Documentation was updated where necessary. (No docs affected.)
- [ ] If the API contract changed, I regenerated the OpenAPI spec. (No API contract change.)

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

Claude Code (Claude Opus 4.8)

Implemented and tested with Claude Code as part of investigating mozilla-ai/otari-ai#1158.

## Overview

This pull request improves error handling in the gateway API by replacing generic "502 Bad Gateway" responses with specific, caller-friendly error codes for upstream provider failures. When a provider error occurs, the gateway now identifies whether it was a timeout, invalid request, missing model, rate limit, or credential issue—and responds accordingly without exposing internal provider details.

## What Changed

**Core Classification Logic**
- Added error classification mechanism in `src/gateway/api/routes/_pipeline.py` with a new `classify_provider_error()` function that maps upstream exceptions to safe HTTP status codes (e.g., 504 for timeouts, 404 for missing models, 429 for rate limits)
- Defined fixed error messages for different failure types to ensure consistency across the API

**API Adapter Updates**
- Updated three API route adapters (`/v1/chat/completions`, `/v1/responses`, `/v1/messages`) to use the new classifier when handling provider errors
- The `/v1/messages` adapter additionally maps classified errors to Anthropic-compatible error format
- For errors that cannot be classified, adapters fall back to generic error responses (preserving backward compatibility)

**Hybrid/Platform Mode Support**
- Extended error classification to the platform execution layer (`src/gateway/api/routes/_platform.py`) so that otari.ai's multi-provider fallback system can return specific error codes for single-attempt or non-retryable failures
- Multi-attempt fallthrough scenarios continue returning the generic "All upstream providers failed" message to avoid misattribution

**Test Infrastructure**
- Added 144 lines of new integration tests covering classification across all three adapters
- Added 92 lines of unit tests for the classification function itself
- Consolidated reusable test fixtures (`responses_request_body`, `messages_request_body`) into a shared `conftest.py` file to avoid duplication

## Benefits

- **Better Diagnostics**: API consumers can now distinguish between timeout, missing model, rate limit, authentication, and unknown provider failures—enabling smarter client-side retry strategies
- **Security**: Raw upstream error messages are never exposed to callers; all responses use fixed, safe strings
- **Consistency**: Error handling is standardized across standalone and hybrid deployment modes


tests/integration/test_provider_error_classification.py
```diff
@@ -0,0 +1,144 @@
...
+    assert response.status_code == 502
+    assert response.json()["detail"] == PROVIDER_ERROR_DETAIL
+    assert "SECRET" not in response.text
+
+
+@pytest.mark.parametrize(("upstream", "expected_status", "expected_detail"), _CASES)
+def test_responses_classifies_provider_error(
+    client: TestClient,
+    master_key_header: dict[str, str],
+    responses_request_body: dict[str, Any],
```
dpoulopoulos, line 93: **[blocker]** The `responses_request_body` (line 93) and `messages_request_body` (line 124) fixtures are defined only in `test_responses_endpoint.py` / `test_messages_endpoint.py`, which are module-local; pytest does not share fixtures across test modules. As a result the 6 `test_responses_classifies_provider_error` and 4 `test_messages_classifies_provider_error` cases **error at setup** (`fixture '…_request_body' not found`) instead of running. I reproduced this locally (7 passed, 10 errors) and confirmed it on the PR's CI (the `test` job fails; the cases RERUN 5x per `pytest.ini` then ERROR). Net effect: CI is red, and the responses/messages classification paths (the bulk of the `messages.py` change, including the Anthropic `error.type` mapping) have **no** passing coverage. Fix: move both fixtures into `tests/integration/conftest.py` (or define them in this file).

src/gateway/api/routes/_pipeline.py
```diff
@@ -155,6 +163,56 @@ class ErrorKind(Enum):
...
+    """
+    status_code = getattr(exc, "status_code", None)
+    if status_code is None:
+        response = getattr(exc, "response", None)
+        if response is not None:
+            status_code = getattr(response, "status_code", None)
+    return status_code if isinstance(status_code, int) else None
+
+
+def classify_provider_error(exc: BaseException) -> ProviderErrorMapping | None:
```
dpoulopoulos, line 188: **[major]** `classify_provider_error` is only consulted by the per-format `provider_error()` handlers, which run on the standalone non-streaming path (`run_standalone_non_stream`) and the standalone/single-attempt pre-stream path (`run_single_attempt_stream`). In hybrid (platform) mode, upstream failures flow through `run_platform_attempts` / `run_streaming_with_fallback` (and the inline `except` in `chat.py:361`), which use `_classify_upstream_error` and raise a generic 502/504 — the new 400/404/429 distinctions never reach the client. Since the otari.ai-operated gateway runs in hybrid mode and the motivating issue (otari-ai#1158) is on that platform, this PR likely does not change what an otari.ai caller sees for a downstream 4xx. Either extend the platform error plumbing to consult the classifier, or explicitly scope the PR/docs to standalone mode (the mapping table currently reads as a global contract).

src/gateway/api/routes/messages.py
```diff
@@ -181,6 +193,10 @@ def error(self, status_code: int, message: str, kind: ErrorKind = ErrorKind.API)
         return _anthropic_error(_ERROR_KIND_TO_ANTHROPIC_TYPE[kind], message, status_code)
 
     def provider_error(self, exc: BaseException) -> HTTPException:
+        mapping = classify_provider_error(exc)
+        if mapping is not None:
+            error_type = _PROVIDER_STATUS_TO_ANTHROPIC_TYPE.get(mapping.status_code, _ERR_API)
+            return _anthropic_error(error_type, mapping.detail, mapping.status_code)
         return _anthropic_error(_ERR_API, _PROVIDER_ERROR, status.HTTP_500_INTERNAL_SERVER_ERROR)
```
dpoulopoulos, line 208: **[minor]** The PR description's mapping table says "5xx / no status / unknown -> 502 `LLM provider error` (unchanged generic)", but the messages adapter's generic fallback returns **500** with `api_error` / `"The request could not be completed by the provider"` (pre-existing, asserted by `test_messages_endpoint.py:178`). The behavior is unchanged and fine; the description/table should note that `/v1/messages` defaults to 500, not 502, to avoid misleading consumers.

src/gateway/api/routes/_pipeline.py
```diff
@@ -155,6 +163,56 @@ class ErrorKind(Enum):
...
+        return None
+    if status_code in (400, 422):
+        return ProviderErrorMapping(status.HTTP_400_BAD_REQUEST, PROVIDER_BAD_REQUEST_DETAIL)
+    if status_code == 404:
+        return ProviderErrorMapping(status.HTTP_404_NOT_FOUND, PROVIDER_MODEL_NOT_FOUND_DETAIL)
+    # A provider rejecting the gateway's credentials is a gateway-config fault,
+    # not the caller's: surface it as a 502, never as a client-facing 401/403.
+    if status_code in (401, 403):
+        return ProviderErrorMapping(status.HTTP_502_BAD_GATEWAY, PROVIDER_CREDENTIALS_DETAIL)
+    if status_code == 429:
```
dpoulopoulos, line 217: **[minor]** A provider 429 is mapped to a client 429, but the `(status, detail)` tuple can't carry the upstream `Retry-After`, so callers can't honor the provider's backoff window (the platform resolve path does forward `Retry-After`, see `_platform.py:506`). Also consider whether a provider rate-limiting the gateway's shared key is really the caller's fault: by the same reasoning used for 401/403->502 ("gateway-config fault, not the caller's"), a shared-quota 429 arguably maps better to 502/503 than to a bare client 429. At minimum, document the dropped `Retry-After`.

src/gateway/api/routes/_pipeline.py
```diff
@@ -155,6 +163,56 @@ class ErrorKind(Enum):
...
+    """Map an upstream provider exception to a safe, specific (status, detail).
+
+    Returns ``None`` when the failure carries no signal we can safely act on, so
+    the caller falls back to its existing generic provider-error response. Every
+    detail returned here is a fixed string: the raw provider message is never
+    included, preserving the no-leak guarantee. The mapping is intentionally
+    conservative, classifying only the cases a caller can act on and leaving
+    everything else (including provider 5xx) to the generic 502.
+    """
+    if isinstance(exc, (asyncio.TimeoutError, TimeoutError, httpx.TimeoutException)):
```
dpoulopoulos, line 198: **[nit]** `asyncio.TimeoutError` has been an alias of the builtin `TimeoutError` since Python 3.11, so listing both in the isinstance tuple is redundant. Harmless, and it mirrors the existing `_classify_upstream_error`, so fine to leave for consistency.

dpoulopoulos (changes requested):
Adds `classify_provider_error` + `ProviderErrorMapping` to `_pipeline.py`, mapping upstream provider failures (timeout, 400/422, 404, 401/403, 429) to safe, fixed-string `(status, detail)` pairs, and wires it into the `chat`, `responses`, and `messages` adapters' `provider_error()` handlers (with `messages` also producing the matching Anthropic `error.type`). Replaces the previous blanket `502 "LLM provider error"` (timeouts aside) with more specific, non-leaking responses, plus unit and integration tests.

khaledosman:
Thanks for the thorough review @dpoulopoulos — all points addressed in e7b958c (rebased onto the platform→hybrid rename):

- **[blocker] shared fixtures** — moved `responses_request_body` / `messages_request_body` into `tests/integration/conftest.py`. The 6 responses + 4 messages classification cases now run instead of erroring at setup. (You were right that the earlier green was misleading — they were erroring + rerunning, not passing.)
- **[major] hybrid mode** — extended the classifier into the hybrid path: `run_platform_attempts` now routes single-attempt and non-retryable terminal failures through `classify_provider_error` (via `_provider_failure_http_exc`, a deferred import to avoid the `_pipeline`↔`_platform` cycle), so an otari.ai caller now gets 400/404 for a downstream 4xx instead of a generic 502. Multi-attempt fallthrough keeps the aggregate `502 "All upstream providers failed"`. Updated `test_hybrid_mode_non_retryable_error_raises_immediately` to assert the classified 400.
- **[minor] messages 500 default** — documented in the PR description's mapping table (the messages adapter keeps its pre-existing 500 + `api_error` for unclassifiable failures).
- **[minor] 429 Retry-After** — documented in `classify_provider_error` that the `(status, detail)` pair drops the upstream `Retry-After`; a bare 429 still signals backoff.
- **[nit] asyncio.TimeoutError** — left for consistency with `_classify_upstream_error`, as you noted.
- **streaming** — scoped as a follow-up in the description (in-stream SSE errors are a separate path from HTTP-status classification).

99 affected tests pass locally; ruff + mypy clean.

tests/integration/test_provider_error_classification.py
dpoulopoulos, line 93: **[blocker]** The `responses_request_body` (line 93) and `messages_request_body` (line 124) fixtures are defined only in `test_responses_endpoint.py` / `test_messages_endpoint.py`, which are module-local; pytest does not share fixtures across test modules. As a result the 6 `test_responses_classifies_provider_error` and 4 `test_messages_classifies_provider_error` cases **error at setup** (`fixture '…_request_body' not found`) instead of running. I reproduced this locally (7 passed, 10 errors) and confirmed it on the PR's CI (the `test` job fails; the cases RERUN 5x per `pytest.ini` then ERROR). Net effect: CI is red, and the responses/messages classification paths (the bulk of the `messages.py` change, including the Anthropic `error.type` mapping) have **no** passing coverage. Fix: move both fixtures into `tests/integration/conftest.py` (or define them in this file).

src/gateway/api/routes/_pipeline.py
```diff
@@ -156,6 +164,61 @@ class ErrorKind(Enum):
...
+    attached ``exc.response.status_code``; everything else has neither.
+    """
+    status_code = getattr(exc, "status_code", None)
+    if status_code is None:
+        response = getattr(exc, "response", None)
+        if response is not None:
+            status_code = getattr(response, "status_code", None)
+    return status_code if isinstance(status_code, int) else None
+
+
```
dpoulopoulos, line 187: **[major]** `classify_provider_error` is only consulted by the per-format `provider_error()` handlers, which run on the standalone non-streaming path (`run_standalone_non_stream`) and the standalone/single-attempt pre-stream path (`run_single_attempt_stream`). In hybrid (platform) mode, upstream failures flow through `run_platform_attempts` / `run_streaming_with_fallback` (and the inline `except` in `chat.py:361`), which use `_classify_upstream_error` and raise a generic 502/504 — the new 400/404/429 distinctions never reach the client. Since the otari.ai-operated gateway runs in hybrid mode and the motivating issue (otari-ai#1158) is on that platform, this PR likely does not change what an otari.ai caller sees for a downstream 4xx. Either extend the platform error plumbing to consult the classifier, or explicitly scope the PR/docs to standalone mode (the mapping table currently reads as a global contract).

src/gateway/api/routes/_pipeline.py
```diff
@@ -156,6 +164,61 @@ class ErrorKind(Enum):
...
+    if status_code is None:
+        return None
+    if status_code in (400, 422):
+        return ProviderErrorMapping(status.HTTP_400_BAD_REQUEST, PROVIDER_BAD_REQUEST_DETAIL)
+    if status_code == 404:
+        return ProviderErrorMapping(status.HTTP_404_NOT_FOUND, PROVIDER_MODEL_NOT_FOUND_DETAIL)
+    # A provider rejecting the gateway's credentials is a gateway-config fault,
+    # not the caller's: surface it as a 502, never as a client-facing 401/403.
+    if status_code in (401, 403):
+        return ProviderErrorMapping(status.HTTP_502_BAD_GATEWAY, PROVIDER_CREDENTIALS_DETAIL)
```
dpoulopoulos, line 211: **[minor]** A provider 429 is mapped to a client 429, but the `(status, detail)` tuple can't carry the upstream `Retry-After`, so callers can't honor the provider's backoff window (the platform resolve path does forward `Retry-After`, see `_platform.py:506`). Also consider whether a provider rate-limiting the gateway's shared key is really the caller's fault: by the same reasoning used for 401/403->502 ("gateway-config fault, not the caller's"), a shared-quota 429 arguably maps better to 502/503 than to a bare client 429. At minimum, document the dropped `Retry-After`.

src/gateway/api/routes/_pipeline.py
```diff
@@ -156,6 +164,61 @@ class ErrorKind(Enum):
...
+def classify_provider_error(exc: BaseException) -> ProviderErrorMapping | None:
+    """Map an upstream provider exception to a safe, specific (status, detail).
+
+    Returns ``None`` when the failure carries no signal we can safely act on, so
+    the caller falls back to its existing generic provider-error response. Every
+    detail returned here is a fixed string: the raw provider message is never
+    included, preserving the no-leak guarantee. The mapping is intentionally
+    conservative, classifying only the cases a caller can act on and leaving
+    everything else (including provider 5xx) to the generic 502.
+    """
```
dpoulopoulos, line 197: **[nit]** `asyncio.TimeoutError` has been an alias of the builtin `TimeoutError` since Python 3.11, so listing both in the isinstance tuple is redundant. Harmless, and it mirrors the existing `_classify_upstream_error`, so fine to leave for consistency.

dpoulopoulos (changes requested):
Adds `classify_provider_error` + `ProviderErrorMapping` to `_pipeline.py`, mapping upstream provider failures (timeout, 400/422, 404, 401/403, 429) to safe, fixed-string `(status, detail)` pairs, and wires it into the `chat`, `responses`, and `messages` adapters' `provider_error()` handlers (with `messages` also producing the matching Anthropic `error.type`). Replaces the previous blanket `502 "LLM provider error"` (timeouts aside) with more specific, non-leaking responses, plus unit and integration tests.

### General notes
- **[minor]** `src/gateway/api/routes/messages.py:200` — The PR description's mapping table says "5xx / no status / unknown -> 502 `LLM provider error` (unchanged generic)", but the messages adapter's generic fallback returns **500** with `api_error` / `"The request could not be completed by the provider"` (pre-existing, asserted by `test_messages_endpoint.py:178`). The behavior is unchanged and fine; the description/table should note that `/v1/messages` defaults to 500, not 502, to avoid misleading consumers.

tests/integration/test_hybrid_mode_messages.py
```diff
@@ -421,7 +430,7 @@ async def fake_amessages(**kwargs: Any) -> MessageResponse:
         headers={"Authorization": "Bearer user_test_token"},
     )
 
-    assert response.status_code == 502
+    assert response.status_code == 400
```
dpoulopoulos, line 433: **[nit]** `test_hybrid_mode_non_retryable_error_raises_immediately` now asserts only `status_code == 400` and the call count; it dropped the prior response-body assertion. Consider also asserting the classified detail (e.g. `PROVIDER_BAD_REQUEST_DETAIL`) and that the raw upstream text does not leak, so the no-leak guarantee and the hybrid messages body shape are locked down by a test rather than left implicit.

src/gateway/api/routes/_pipeline.py
```diff
@@ -156,6 +164,61 @@ class ErrorKind(Enum):
...
 
 
+class ProviderErrorMapping(NamedTuple):
+    """A safe, client-facing (status, detail) for a classified provider failure."""
+
+    status_code: int
+    detail: str
+
+
+def _upstream_status_code(exc: BaseException) -> int | None:
```
dpoulopoulos, line 174: **[nit]** Latent fragility (shared with existing code): `_upstream_status_code` reads `status_code`/`response.status_code` off the raw provider SDK exception. any-llm currently re-raises those raw exceptions (so OpenAI/Anthropic `APIStatusError` work), but it emits a DeprecationWarning that it will convert to unified `AnyLLMError` (which carry neither attribute) in a future version, and does so today when `ANY_LLM_UNIFIED_EXCEPTIONS=1`. If that flips, the classifier silently returns None for everything and degrades to the generic 502. `_classify_upstream_error` has the same dependency, so this isn't new, but the PR broadens reliance on it; worth a tracking note.

dpoulopoulos (approved):
Adds `classify_provider_error` + `ProviderErrorMapping` to `_pipeline.py`, mapping upstream provider failures (timeout, 400/422, 404, 401/403, 429) to safe, fixed-string `(status, detail)` pairs that never echo the raw provider message, and wires it into the `chat`/`responses`/`messages` adapters' `provider_error()` and into hybrid-mode terminal failures via `_provider_failure_http_exc` in `_platform.py`. Replaces the previous blanket `502 "LLM provider error"` (timeouts aside) with more specific, non-leaking responses, plus unit and integration tests.

The two prior review items that needed code changes are addressed and verified: the **[blocker]** shared-fixture bug is fixed (fixtures moved to `tests/integration/conftest.py`; the 6 responses + 4 messages classification cases now run and pass), and the **[major]** hybrid-mode gap is substantially resolved for the non-streaming path (verified end-to-end via `test_hybrid_mode_non_retryable_error_raises_immediately`). The two doc-only [minor] items (messages 500 default; 429 dropping `Retry-After`) are documented. Residual items are minor/nit and largely pre-existing: the hybrid **streaming** path is still generic (documented follow-up), and hybrid `/v1/messages` errors bypass the Anthropic envelope. Unit + integration tests pass locally; ruff + mypy clean.

### General notes
- **[minor]** `src/gateway/api/routes/chat.py:372` — Streaming hybrid pre-first-byte failures stay generic. When every streaming attempt fails before its first chunk, all three formats return a generic 502/504 without consulting `classify_provider_error` (chat.py:372, responses.py:409, messages.py:406; `run_streaming_with_fallback` re-raises into these blocks). So in hybrid mode the same downstream 4xx now yields a different status purely based on `stream`: non-stream -> 400/404/429, stream -> 502. The PR description scopes "streaming" as a follow-up but frames it as "errors after a stream has started (in-stream SSE)" - that mischaracterizes this case, which is a pre-first-byte failure surfacing as a normal HTTP status, exactly the shape the classifier handles. Consider routing these terminal exceptions through the classifier (same shape as the non-stream path) or tightening the scope note.
- **[minor]** `src/gateway/api/routes/messages.py:455` — Hybrid `/v1/messages` provider errors bypass the Anthropic envelope. The hybrid non-streaming path (`run_platform_non_stream` -> `run_platform_attempts`) raises the format-agnostic `_provider_failure_http_exc`, a plain `{"detail": <str>}` HTTPException, which the route just re-raises (messages.py:465-466). So a hybrid caller does not get the `{"type":"error","error":{"type":...}}` shape - nor the new `error.type` mapping at lines 138-143/204-207 - that the standalone path produces via `_MessagesAdapter.provider_error`. This is pre-existing (the old generic 502 was also plain), but it means the PR's headline messages mapping only takes effect standalone, not on the otari.ai platform that motivated the change.

khaledosman:
will do the fixes along with covering the streaming path in a follow up PR

PR #199, opened 2026-06-24
feat(tools): add resolve_time gateway tool backed by dateparser

agpituk:
## Description

Adds an `otari_resolve_time` built-in tool: it resolves natural-language time expressions ("last 2 weeks", "yesterday", "January 1 to January 15") into deterministic ISO 8601 timestamps using `dateparser`, so any model, including open-weight ones, gets reliable relative-date handling instead of doing brittle date math in its head.

Parity with the existing `web_search` built-in tool:

- New `ResolveTimeBackend` implements the `ToolBackend` protocol and runs **locally**, with no upstream provider and no credential (secret-free).
- In **hybrid mode** the per-workspace policy (timezone mode, prefer-dates / date-order / week-start, languages, purpose hint, opaque `parser_options`) is resolved from the platform's new `POST /gateway/resolve-time/resolve` endpoint and merged with per-request overrides.
- In **standalone mode** the request tool entry drives it directly.
- Wired into all three dispatch sites (`dispatch_non_stream`, `open_stream`, platform-fallback streaming). It shares the single tool-loop pool, so it is mutually exclusive with the other gateway-run tools for now (returns 400), matching the existing sandbox / web_search / MCP behaviour.

Hardening: `dateparser.parse` runs off the event loop via `asyncio.to_thread`, the `expression` is length-capped, and `parser_options` / `languages` are clamped in standalone mode (the platform applies the same caps to stored config). `dateparser` was promoted from a transitive to an explicit dependency.

> Companion platform PR (adds the `/gateway/resolve-time/resolve` endpoint, the per-workspace config, and the Time UI): mozilla-ai/otari-ai#1197

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Implements the gateway side of the relative-time tool (cf. mozilla-ai/octonous#1301).

## Checklist

- [x] If the API contract changed, I regenerated the OpenAPI spec (`uv run python scripts/generate_openapi.py`). (Verified no-op: the tool type adds no schema change.)

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

Claude Opus 4.8, via Claude Code.

The design was human-directed; the implementation, tests, and this description were written by the agent.

**NOTE:**

## Summary
Added a new built-in time-resolution tool that can turn natural-language date and time phrases into stable ISO 8601 results. It works locally, with workspace policy support in hosted mode, and now appears in the product documentation alongside the other built-in tools.

## Technical notes
- Added `otari_resolve_time` wiring through request handling and stream/tool-loop dispatch.
- Introduced a local resolve-time backend powered by `dateparser`.
- Added platform policy lookup for hybrid mode and request-level overrides.
- Hardened parsing with input limits and background execution.
- Updated dependencies and added unit/integration coverage.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@ Otari sits between your applications and LLM providers so you can control access
 - User and budget controls (`/v1/users`, `/v1/budgets`)
 - Usage and pricing tracking (`/v1/usage`, `/v1/pricing`)
 - Health and metrics endpoints (`/health`, optional `/metrics`)
-- Built-in tools Otari runs itself, `otari_code_execution` (sandboxed Python REPL) and `otari_web_search`. See [Built-in tools](#built-in-tools).
+- Built-in tools Otari runs itself, `otari_code_execution` (sandboxed Python REPL), `otari_web_search`, and `otari_resolve_time` (natural-language date resolution). See [Built-in tools](#built-in-tools).
 - Request-level [guardrails](#guardrails) (e.g. prompt-injection detection) Otari enforces on input before calling the provider.
 
 ## Quickstart
@@ -173,11 +173,12 @@ Otari will be available at `http://localhost:8000`.
 
 ## Built-in tools
 
-Otari can run a couple of tools itself so any model, including
+Otari can run a few tools itself so any model, including
 open-weight ones, gets parity with what frontier APIs expose as managed
-tools. Both are opt-in via the request's `tools` array and run inside
-docker-compose profiles so operators who don't use them don't pull extra
-images.
+tools. They are opt-in via the request's `tools` array. The sandbox and search
+backends run inside docker-compose profiles so operators who don't use them
+don't pull extra images; `otari_resolve_time` is pure local computation and
+needs no extra service.
 
 These use dedicated `otari_*` tool types. The keyword decides who runs the
 code: an `otari_*` type means Otari runs it. Every other tool type, the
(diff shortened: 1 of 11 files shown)


njbrake (changes requested):

## On `resolve_time` as a gateway built-in

Before the implementation details, a step back on whether this belongs as a built-in tool at all. I don't think it does, at least not as framed, and I'd like to talk it through before it lands.

**The existing built-ins share a property `resolve_time` doesn't: they need something only the operator can provide.** `otari_code_execution` needs a sandbox; `otari_web_search` needs network egress, SSRF policy, and a search backend with keys. A model cannot reach the web or run code no matter how capable it is, so the operator is the right party to own that infrastructure. The README frames the category this way: parity with what frontier APIs expose as managed tools, and both OpenAI and Anthropic expose code-interpreter and web-search as server-side tools for that reason.

`resolve_time` is different in kind. It is pure local computation wrapping `dateparser`: no secret, no network, no sandbox, no trust boundary. The model is not incapable of resolving relative dates; it is merely unreliable at it. That is the distinction I'd draw:

- web_search and code_execution close a **capability gap**: something the model fundamentally cannot do without operator-owned infrastructure.
- resolve_time closes a **quality gap**: something the model can already do but does poorly.

The fix for a quality gap is a helper function, and a helper function that is a pure function of its inputs is ordinary client-side function-calling territory. The caller defines `resolve_time`, the gateway forwards it untouched (the existing pass-through path), the model emits a call, and the caller runs `dateparser`. Nothing about that needs the operator to run it, because the operator holds nothing the caller lacks.

**Why this matters beyond this one PR:** once "things the model is bad at" qualifies a tool for built-in status, there is no principled line. Unit conversion, arithmetic, JSON validation, regex testing, base64, and UUID generation are all the same shape. "Things the model physically cannot do without operator infrastructure" is a closed, defensible set; "things the model does poorly" is not.

**The mutual-exclusion constraint also lands hardest on this tool.** Sharing the single tool-loop pool means `resolve_time` cannot combine with `code_execution`, `web_search`, or MCP. For the heavy tools that is a tolerable v1 limitation. But date resolution is a supporting utility whose value is alongside a retrieval or search step (the README's own example, "summarize my notes from the last 2 weeks," implies retrieval it cannot coexist with). So in the cases where a gateway-run date helper would earn its keep, it is forbidden.

**Suggested path:** move this into the agent layer (octonous, per the referenced octonous#1301) as a client-side function tool. That delivers the open-weight-parity benefit without the gateway taking on a runtime dependency, a platform endpoint, per-workspace config plus a Time UI, a hybrid-mode round-trip, and the maintenance surface of fuzzy natural-language date parsing.

If there is a deliberate platform-product reason for a centrally managed, per-workspace time policy (the companion Time UI in otari-ai#1197 suggests someone may want that), then it is a product decision rather than an engineering one, and I would still gate it on solving multi-backend dispatch first, since a centrally managed date tool that cannot run next to search or code does not work in its primary use case.

### Secondary: the advertised examples do not parse

Independent of the placement question, the tool advertises forms it cannot resolve. `last 2 weeks` and `since Monday` appear in the tool description, the model-facing purpose hint, the parameter docs, and the README example, but `_resolve_common_period_window` only handles fixed phrases and `dateparser.parse` returns `None` for quantified or duration phrases. Reproduced against the backend at a frozen reference time:

```
'last 2 weeks'  -> Could not parse time expression: 'last 2 weeks'
'since Monday'  -> Could not parse time expression: 'since Monday'
'past week'     -> Could not parse time expression: 'past week'
'last week'     -> {"start": "...", "end": "..."}   # works
```

The purpose hint tells the model to call `resolve_time("last 2 weeks")`, and it gets a parse error back. CI does not catch it because no test exercises the quantified form. If the tool does stay, this needs either a quantified-window branch with tests, or narrowing the advertised examples to forms that resolve.

(CodeRabbit's other points: the README "platform mode" wording should be "hybrid mode" per the recent rename; the unguarded `response.json()` on a 200 is real but mirrors `_resolve_platform_web_search`, so fix both or neither; and the logging nit conflicts with the sibling backend's own pattern, so I would not block on it.)

Marking request-changes mainly to force the placement conversation, not because the code is wrong; the implementation itself faithfully mirrors the web_search pattern.

agpituk:
Closing this pr @njbrake , I still believe it could be useful somehow to give this kind of tools to the models. I'm not quite sure I agree this is a quality gap, I think resolve time is still more a capability gap and it helps a lot with the rest of the tools that we give) but probably this isn't a server tool, just a library, so it would be better fitted in something like an SDK, let's discuss it

PR #201, opened 2026-06-24
feat(pricing): add genai-prices default pricing fallback

### LINKED ISSUE #97: [GA]: Use a package for LLM pricing defaults instead of mzai maintained json
### Description & Motivation

Consider using https://github.com/pydantic/genai-prices or https://github.com/anomalyco/models.dev to pull pricing and model information from them, since they are maintained by the community.

njbrake:
## Description

_Note: this PR was authored by Claude via back-and-forth with @njbrake. The reasoning and decisions are his; the prose and code are Claude's._

Otari prices billable models from the database (the config `pricing:` block or `POST /v1/pricing`). With `require_pricing=true` (the default), an unpriced model is rejected with HTTP 402, so onboarding meant entering rates for every model before traffic could flow.

This adds an **opt-in** default-pricing fallback sourced from the community-maintained [`genai-prices`](https://github.com/pydantic/genai-prices) dataset (issue #97 proposed `genai-prices` or `models.dev`; `genai-prices` was chosen because it is a real pip package with data bundled offline, so there is no network dependency, no vendored snapshot, and built in model matching).

When enabled and the database has no entry for a model, the fallback resolves per-million rates inside `find_model_pricing`:

- **Resolution order**: database `provider:model`, then the legacy `provider/model` key, then the genai-prices default. The database is consulted first, so config and `/v1/pricing` entries always win.
- It returns a **transient, unpersisted** `ModelPricing` (a lookup result, not a stored row), so nothing is written to the database and the `effective_at` history is preserved.
- Unknown models still resolve to `None`, so `require_pricing` keeps failing closed (402) for them.
- Tiered ("cliff") prices are flattened to their base rate; input-only models (embeddings, rerank) are priced with a `0` output rate; historical lookups pass `as_of` through as the price timestamp.
- The genai-prices lookup is fail-safe: any lookup error degrades to "unpriced" (require_pricing decides) rather than erroring the request.

### Off by default

The fallback is gated by a new `default_pricing` config field that **defaults to off**. A billing gateway should generally price from rates the operator controls: community estimates can lag or differ from real provider rates, and enabling the fallback changes what `require_pricing=true` guarantees. Operators opt in with `default_pricing: true` (or `OTARI_DEFAULT_PRICING=true`). The resolved flag is held in `pricing_service` and set once at startup from config, mirroring the module-level engine pattern in `core.database`, so the toggle reaches the lookup without threading config through the budget and billing call sites.

### Verification

- `make lint`, `make typecheck`, and the OpenAPI freshness check all pass.
- Full suite green on a local PostgreSQL instance.
- Live end to end with a real OpenAI key: with `default_pricing=true` and `require_pricing=true`, `openai:gpt-4o` is priced and returns HTTP 200; with `default_pricing=false` it returns HTTP 402.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes #97

## Checklist

- [x] If the API contract changed, I regenerated the OpenAPI spec. (No API contract change; the OpenAPI freshness check confirms the spec is current.)

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:** Claude Opus 4.8 (Claude Code)

Research, implementation, and tests were produced by Claude through back and forth with @njbrake, who made the design decisions (source choice, the live fallback approach, and off by default).

## Summary
Added an opt-in default pricing fallback so Otari can use community-maintained pricing data when a model is missing from the database. The gateway now prefers database pricing first, then falls back to bundled defaults only when enabled, while still failing closed for truly unknown models.

Also updated startup/config handling, documentation, and tests to cover the new behavior and ensure pricing state does not leak between tests.

## Technical notes
- New `default_pricing` config flag, off by default and enableable via config or `OTARI_DEFAULT_PRICING=true`
- Default pricing lookups use `genai-prices` as a read-only fallback, with database pricing taking precedence
- Lookup failures degrade to “unpriced” instead of breaking requests
- Input-only models now get a zero output rate
- Added regression coverage for fallback resolution, startup warnings, and fail-closed behavior


khaledosman (commented):
## Review

Solid, well-tested, and architecturally clean. It correctly uses the **bundled offline** `calc_price` (no network dependency, so genai-prices' blocking `UpdatePrices` fetch is never triggered), returns a **transient, unpersisted** `ModelPricing` so DB pricing always wins and the `effective_at` history is preserved, and the conftest autouse fixture resets the process-global flag so it can't leak across tests. The tests correctly *disable* defaults where they would otherwise mask the 402 / DB-only paths. Standards-clean (no em-dashes / `--`; type hints present).

### Findings

**[design] Default-on softens the fail-closed billing posture** (you already flagged this)
With `default_pricing=true` by default, an unpriced model is auto-priced from *community estimates* rather than rejected. For a gateway that bills, that means charging on prices that can lag or differ from real provider rates. Issue #97 did ask for "fast default pricing", so default-on matches intent, but for a billing system opt-in is the more conservative default. Worth an explicit team decision since it changes what `require_pricing=true` guarantees.

**[correctness] Provider-agnostic fallback can mis-price ambiguous model names**
`default_model_pricing` retries with `provider_id=None` after the provider-scoped lookup. If a model *name* exists under multiple providers at different rates (or the gateway's provider id is not one genai-prices recognizes), the model-only match can resolve to a *different* provider's price, a silent mischarge. The debug log of the matched provider/model helps diagnosis but not prevention. Consider gating/warning on the provider-agnostic match, or at least documenting the risk.

**[robustness] Only `LookupError` is caught around `calc_price`**
Any other exception (a `ValueError`, or a future genai-prices API/data change) would propagate out of `find_model_pricing`, the per-request hot path, turning into a request error for that model rather than degrading to "unpriced -> require_pricing". With `_ZERO_USAGE` the known cache-overcount `ValueError` cannot fire today, but a defensive `except Exception: log + return None` would make a genai-prices hiccup fail safe instead of failing the request.

**[correctness / limitation] Tiered prices flattened to base**
Requests that cross a tier boundary (long-context / large) are billed at the base rate, an under/over-charge. It is acknowledged in the comment; worth surfacing in the docs for billing transparency.

**[perf, low] No memoization** — `calc_price` (clause/regex matching over the dataset) runs per request on a DB miss. A small `(provider, model) -> rates` cache would help sustained traffic on a genai-known-but-unconfigured model.

**[nit] HuggingFace models will not get defaults** — genai-prices models HuggingFace inference as per-backend providers (`huggingface_<backend>`), with no bare `huggingface` provider, so `huggingface:<model>` -> `LookupError` -> `None` -> 402. That is *correct* (HF auto routing is not priceable from the model id alone), but worth a docs note so operators are not surprised.

None are blocking. The design point and the provider-agnostic mis-pricing risk are the two I would most want weighed before this lands.

---
*Review generated with [Claude Code](https://claude.com/claude-code).*

njbrake:
@khaledosman thanks for the review! Yes 100% agree, the biggest decision we need to decide is how to handle default pricing and whether it should default to on or off. Personally i am fine with either way.

njbrake:
_Note: this reply was drafted by Claude via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

Thanks for the thorough review. Changes in 5755b75:

- **Default flipped to off (opt-in).** Agreed that conservative is the right call for a billing gateway; `require_pricing: true` keeps its fail-closed meaning unless an operator opts in.
- **Provider-agnostic mis-pricing:** documented the risk in `configuration.md` (prefer configuring such models explicitly), and kept the debug log of the matched provider/model for diagnosis.
- **Error handling:** broadened the catch around `calc_price` so any lookup failure degrades to "unpriced" (require_pricing decides) instead of erroring the request.
- **Tiered-to-base** and the **HuggingFace per-backend** gap are now called out in the docs.
- **Input-only models** (embeddings, rerank) are priced with a `0` output rate rather than dropped, which also addresses Copilot's note.
- Skipped the redundant provider-agnostic pass when no provider is supplied (Copilot).

I left memoization out for now: the lookup key includes the `as_of` timestamp for historical reconcile, so a naive cache could return stale historical rates. Happy to add a date-bucketed cache in a follow-up if sustained traffic on unconfigured-but-known models shows up.

khaledosman (dismissed):
forgot to pre-approve.

on opt-in vs opt-out, I think failing fast with a descriptive error when there's no configured pricing (should be the current behaviour) is the better approach because it prevents $ spend or bypassing the cost calculation/budgets which is the bigger risk

khaledosman (approved):

PR #211, opened 2026-06-25
feat(config): support full provider/pricing config via environment

### LINKED ISSUE #208: Support full provider/pricing config via environment variables (PaaS-friendly, e.g. Railway)

### Problem

Otari is almost perfectly env-configurable: `load_config` in `src/gateway/core/config.py` layers `OTARI_<FIELD>` overrides on top of any YAML for every **scalar** field (`master_key`, `database_url`, `port`, `require_pricing`, etc.). But `providers` and `pricing` are `dict` fields, and `_coerce_scalar_env` raises `_NonScalarField` for them, so they can only be set from a mounted `config.yml`.

Provider API keys still work in env-only deploys because any-llm reads `OPENAI_API_KEY` / `ANTHROPIC_API_KEY` / etc. directly. The gap is everything that needs structured provider config: custom `api_base`, `client_args`, Vertex (project/location/service-account), and per-model `pricing` overrides.

This is the config that platforms like Railway, Render, Fly.io, and Kubernetes (env/ConfigMap-driven, no convenient file mount) make awkward to supply. Mounting a YAML file on these platforms means baking a custom image or adding a wrapper entrypoint, which undercuts the "pull the published image and set env vars" deploy story.

### Proposal

Let an operator supply the structured config through the environment. Options (not mutually exclusive):

1. **`OTARI_CONFIG_YAML`** (raw YAML) and/or **`OTARI_CONFIG_B64`** (base64-encoded YAML), parsed in `load_config` and merged exactly like a config file, so the full schema is reachable with no new surface area.
2. **`OTARI_PROVIDERS`** / **`OTARI_PRICING`** as JSON strings, narrower, targeting just the two non-scalar fields most people need.

Merge precedence to define: env-structured config over file, and scalar `OTARI_<FIELD>` overrides still win on top. `OTARI_CONFIG_YAML` (option 1) is the lowest-maintenance since it reuses the existing YAML path and schema validation.

### Acceptance criteria

- [ ] Otari can run with full provider + pricing config and **no** `config.yml` present, configured entirely via env.
- [ ] Precedence is documented and tested (file < env-structured < scalar `OTARI_<FIELD>`).
- [ ] Invalid YAML/JSON fails fast with a clear startup error.
- [ ] Docs note the PaaS use case and the `require_pricing=true` interaction (unpriced models are rejected by default, so a deploy needs either pricing config or `OTARI_REQUIRE_PRICING=false`).

### Context

Motivated by building a one-click **Deploy on Railway** template. The minimal template (Otari + Postgres, provider keys via standard env vars) already works today; this issue unblocks the template growing to full provider/pricing config without a custom image.

njbrake:
## Description

Otari already layers `OTARI_<FIELD>` overrides on every scalar config field, but `providers` and `pricing` are dict fields that `_coerce_scalar_env` rejects, so they could only be set from a mounted `config.yml`. That makes PaaS deploys (Railway, Render, Fly.io, Kubernetes) awkward: reaching custom `api_base`, `client_args`, Vertex settings, or per-model pricing meant baking a custom image or adding a wrapper entrypoint, undercutting the "pull the published image and set env vars" story.

This adds two env vars that carry the entire YAML schema, reusing the existing config-file path and pydantic validation (the issue's recommended option 1):

- `OTARI_CONFIG_YAML`: the full config as raw YAML.
- `OTARI_CONFIG_B64`: the same YAML base64-encoded, for env-var UIs that mangle multiline values.

`${VAR}` references resolve exactly as in a config file. Precedence, lowest to highest: config file, env-structured config (raw YAML wins over base64), then scalar `OTARI_<FIELD>` overrides. Env-structured keys replace the matching top-level keys from the file. Invalid base64, invalid YAML, or a non-mapping top level fail fast at startup with a clear error.

Option 2 from the issue (`OTARI_PROVIDERS` / `OTARI_PRICING` JSON) was intentionally not implemented: option 1 satisfies every acceptance criterion with less surface area.

### Acceptance criteria

- [x] Runs with full provider + pricing config and no `config.yml`, configured entirely via env.
- [x] Precedence documented and tested (file < env-structured < scalar `OTARI_<FIELD>`).
- [x] Invalid YAML/base64 fails fast with a clear startup error.
- [x] Docs note the PaaS use case and the `require_pricing=true` interaction.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [x] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes #208

## Checklist

No API contract changed (config loading only); OpenAPI spec confirmed up to date. New tests live in `tests/integration/test_config_env_loading.py`.

## AI Usage

- [ ] No AI was used.

**Any additional AI details you'd like to share:** Drafted via the fix-github-issue workflow; reasoning and decisions are @njbrake's.

## Summary
- Added support for loading full gateway configuration from environment variables, including structured provider and pricing settings.
- Documented the new env-based configuration flow for PaaS deployments and clarified how pricing requirements behave when pricing is not provided.
- Added integration coverage for raw YAML, base64 YAML, config precedence, env interpolation, and fast-fail validation for invalid structured input.

## Technical notes
- New env vars: `OTARI_CONFIG_YAML` and `OTARI_CONFIG_B64`.
- Precedence is now: config file → structured env config → scalar `OTARI_<FIELD>` overrides.

### DIFF
--- a/deploy/railway/README.md
+++ b/deploy/railway/README.md
@@ -47,9 +47,10 @@ Notes:
   Railway's `DATABASE_URL` works without edits.
 - `OTARI_REQUIRE_PRICING=false` is deliberate. The image default is `true`
   (fail-closed), which rejects any model without configured pricing; that would
-  make an env-only deploy unusable until pricing is added. Configuring per-model
-  pricing via env is tracked in
-  [#208](https://github.com/mozilla-ai/otari/issues/208).
+  make an env-only deploy unusable until pricing is added. To serve priced
+  models instead, supply `pricing` (and any other structured config like custom
+  `api_base` or Vertex settings) through `OTARI_CONFIG_YAML` / `OTARI_CONFIG_B64`;
+  see [Full config via environment](../../docs/configuration.md#full-config-via-environment).
 
 ## Deploy
 
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -82,13 +82,40 @@ pricing:
 
 The following `OTARI_` variables override config file values for their matching fields. For example, `OTARI_PORT=9000` overrides `port: 8000` in the YAML.
 
-`OTARI_` overrides apply to scalar fields (strings, numbers, booleans). List and dict fields (`cors_allow_origins`, `providers`, `pricing`, and the `platform` block) are not read from `OTARI_` variables; set them in the YAML file instead. The `platform` block also has dedicated `PLATFORM_*` variables (see the otari.ai variables below).
+`OTARI_` overrides apply to scalar fields (strings, numbers, booleans). List and dict fields (`cors_allow_origins`, `providers`, `pricing`, and the `platform` block) are not read from individual `OTARI_` variables; set them in the YAML file, or supply the whole config through the environment with `OTARI_CONFIG_YAML` / `OTARI_CONFIG_B64` (see [Full config via environment](#full-config-via-environment)). The `platform` block also has dedicated `PLATFORM_*` variables (see the otari.ai variables below).
 
 The config file also supports `${ENV_VAR}` interpolation:
 
 ```yaml
 master_key: "${MY_SECRET_KEY}"
 ```
+
+### Full config via environment
+
+On PaaS platforms (Railway, Render, Fly.io, Kubernetes) where mounting a `config.yml` is awkward, you can supply the entire config, including the non-scalar `providers` and `pricing` fields, through the environment. This reaches the full schema with no file mount and no custom image.
+
+| Variable | Description |
+|----------|-------------|
+| `OTARI_CONFIG_YAML` | The full config as raw YAML, parsed exactly like a `config.yml`. |
+| `OTARI_CONFIG_B64` | The same YAML, base64-encoded, for env-var UIs that mangle multiline values. |
+
+The YAML supports the same `${ENV_VAR}` interpolation as a config file, so you can keep secrets in separate variables:
+
+```yaml
+# value of OTARI_CONFIG_YAML
+providers:
+  openai:
+    api_key: ${OPENAI_API_KEY}
+    api_base: https://my-proxy.example/v1
+pricing:
+  openai:gpt-4o:
+    input_price_per_million: 2.5
+    output_price_per_million: 10
+```
+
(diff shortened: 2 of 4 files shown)


khaledosman (commented):
## Review

Clean, well-scoped, and it reuses the existing config-file path plus pydantic validation rather than a parallel schema. The tests are thorough: env-only config, base64, newline-wrapped base64, raw-wins-over-base64, the full file < structured < scalar precedence, `${VAR}` resolution, and all three fail-fast paths (bad YAML / bad base64 / non-mapping). A few notes.

### [security, low] Error detail can echo secret-bearing config

`_load_structured_env_config` raises `f"{source} is not valid YAML: {exc}"`. A `yaml` parse error's string can include a fragment of the offending content, and this variable is *designed* to carry provider `api_key`s and pricing, so a malformed `OTARI_CONFIG_YAML` could surface a snippet of a secret in the fail-fast startup error/logs. In practice yaml errors mostly point at a position rather than echo the value, so the risk is bounded, but since this input intentionally holds secrets I would drop the raw `{exc}` (or log only line/column) to be safe. The base64 and non-mapping messages are fine: they echo only the exception class and the top-level type respectively.

### [design, confirm] Top-level replace, not deep-merge

`config_dict.update(structured_env_config)` replaces matching *top-level* keys wholesale, so file `providers.openai` and env `providers.mistral` cannot coexist: the env `providers` wins entirely. It is documented ("replace the matching top-level keys") and tested, so this is a confirm-intent note rather than a bug. A one-liner in the docs that it is key-level replace, not a merge, would set operator expectations (no partial-provider overlays).

### [nit] Silent base64 ignore when both set

If both `OTARI_CONFIG_YAML` and `OTARI_CONFIG_B64` are set, the base64 is silently dropped (documented and tested). A `logger.debug("both set; using OTARI_CONFIG_YAML")` would aid diagnosis.

### Positives

Fail-fast with clear messages, base64 whitespace/newline-wrap handling, `${VAR}` parity with config files, and docs covering the PaaS use case plus the `require_pricing` interaction.

Only the security note is worth acting on before merge; the rest are confirm/nit.

---
*Review generated with [Claude Code](https://claude.com/claude-code).*

khaledosman (approved):

PR #227, opened 2026-06-26
feat(ui): add a standalone admin dashboard for models, pricing, and settings

### LINKED ISSUE #225: Real Web Dashboard
This probably is a task for post-GA: however, not having a basic web dashboard for otari is a miss. The first thing I want to do when I have otari deployed is open up a web dashboard to create virtual keys and observe usage and traffic. Using the Postman collection is fine for a stopgap but if I could have a web dashboard that would be ideal.

njbrake:
## Description

Adds a standalone admin web dashboard for Otari, plus the gateway endpoints and services that back it. Until now the only web page was a static get-started splash, so an operator who wanted to see which models the gateway exposes, price them, or manage aliases had to reach for the API directly.

## The dashboard (frontend)

A React 19 + HeroUI v3 single-page app under `web/` (Vite, Tailwind v4, TanStack Query). It drives the standalone management API using the master key, which the operator enters on a sign-in screen and which is held only in the browser tab's session storage (a 401 anywhere drops it and returns to sign-in).

This first pass is deliberately small: two pages behind a collapsible, resizable sidebar.

- **Models**: the full catalogue of discovered and hand-added models in one table. Sortable columns (context window, release date, input and output price), defaulting to newest first and remembered across refreshes; filters (provider, pricing, source, per-model capability, context, price, release window); and a persistent detail panel showing pricing, specs, modalities, capabilities, and provider info. Prices are editable inline, any model can be aliased, and a separate Aliases tab manages runtime aliases.
- **Settings**: toggle model discovery and default (community-rate) pricing at runtime; changes persist across restarts.

## Gateway changes (backend)

This PR is not only the frontend. To back the dashboard, it adds a set of management endpoints and services to the gateway. They are standalone-only and master-key gated, and none of them change the request/completion hot path.

- **Model aliases** (`api/routes/aliases.py`, `services/alias_service.py`): create, list, and delete named aliases (for example `fast-model` to `openai:gpt-4o-mini`) that resolve to a real `provider:model` at dispatch time. New `ModelAlias` entity and the `model_aliases` table (migration `f1a3b5c7d9e2`).
- **Provider-aware model discovery** (`GET /v1/models/discoverable`, `services/model_discovery_service.py`): reports the models each configured provider's credentials can actually reach, so the UI can offer a real picker instead of free text.
- **models.dev catalog enrichment** (`GET /v1/models/metadata`, `services/model_catalog_service.py`): fetches and caches the public models.dev catalog to annotate each model with context window, release date, input/output modalities, and capability flags. `GET /v1/models` now also carries `context_window`. Two new config flags gate and tune it: `models_dev_metadata` (on/off) and `models_dev_cache_ttl_seconds`.
- **Provider metadata** (`GET /v1/providers`, `services/provider_metadata_service.py`): per-provider name, docs URL, env key, and capability flags, combined from any-llm's provider metadata and the genai-prices catalog.
- **Runtime settings** (`GET`/`PATCH /v1/settings`, `services/runtime_settings_service.py`): read and toggle model discovery and default pricing without a redeploy. Overrides are persisted in the DB and re-applied on startup. New `RuntimeSetting` entity and the `runtime_settings` table (migration `a2c4e6f8b0d1`).
- **Pricing** (`services/pricing_service.py`, `api/routes/pricing.py`): a shared genai-prices resolver, a `context_window` helper, per-model default-pricing status on the model listing, and a price-history endpoint.

Supporting plumbing: `core/config.py` gains the models.dev flags and related wiring; `core/database.py` gains session helpers used by the new services; and `models/entities.py` gains the two entities above with their Alembic migrations.

## Serving and packaging

Serving (`src/gateway/main.py`, `src/gateway/dashboard.py`): in standalone mode the gateway serves `index.html` at `/` and the hashed assets under `/assets`; the get-started tutorial moves to `/welcome`. Hybrid mode has no local management API, so `/` keeps serving the tutorial, and the root also falls back to the tutorial when the bundle has not been built. Navigation is client-side, so no server catch-all route is needed.

The built bundle is committed under `src/gateway/static/dashboard` and added to `package-data`, so the wheel and Docker image ship the dashboard with no Node build stage. The Dockerfile is unchanged. A CI workflow (`.github/workflows/otari-dashboard.yml`) type-checks, tests, and builds the frontend on `web/**` changes, and fails if the committed bundle is stale.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [x] Infrastructure / CI

## Relevant issues

Fixes #225

## Checklist

- [x] I have added or updated tests that cover my change (`tests/unit`, `tests/integration`, and the frontend Vitest suite).
- [x] I ran the Definition of Done checks locally (`make lint`, `make typecheck`, `make test`, plus the dashboard typecheck, tests, and build).
- [x] If the API contract changed, I regenerated the OpenAPI spec and Postman collection (`make openapi-check`, `make postman-check`).

## AI Usage

- [ ] No AI was used.

Implemented by Claude Code via back-and-forth with @njbrake, who made the scoping and design decisions (packaging approach, page scope, mount point, and the models-page and settings design). Frontend checks pass locally: the TypeScript typecheck, the Vitest suite, and the production build with the committed bundle rebuilt to match. The branch is merged up to date with `main`, with the OpenAPI spec and Postman collection regenerated; the new endpoints' integration tests pass locally, and CI runs the full Python unit and integration suites.

## Summary

- Added a standalone React admin dashboard for managing models, pricing, aliases, providers, and runtime settings.
- Added master-key-protected APIs for model discovery, metadata, provider details, aliases, pricing, and persisted settings.
- Added model aliases with validation, caching, target hiding, inherited pricing, and provider-aware routing.
- Added model catalogue enrichment from `models.dev`, context-window data, provider metadata, and resilient pricing resolution.
- Standalone mode now serves the dashboard at `/`; hybrid mode retains the tutorial at `/welcome` and disables dashboard management APIs.
- Packaged the built dashboard bundle for Python wheels and Docker images, with CI checks for type safety, tests, builds, and stale assets.
- Added database entities and migrations for aliases and runtime settings, plus extensive frontend, unit, and integration coverage.


khaledosman (commented):
Review by Claude Code (requested by @khaledosman).

One CI-blocking issue found, plus six inline findings below.

**CI: `docs/public/openapi.json` is stale** (the `openapi-spec` check is failing). Reproduced locally: the committed spec is missing the `security: [{"ApiKeyAuth": []}]` stamp on exactly the two new operations (`GET /v1/settings`, `POST /v1/usage/backfill`), so it was generated before those routes picked up the security stamp from `custom_openapi`. Fix: `uv run python scripts/generate_openapi.py` and commit. The Postman collection is fresh and the bundle staleness check passes.

Also: the PR description still describes the pre-slimming scope (keys/users pages, Recharts chart); worth updating alongside the doc fixes noted inline.

src/gateway/api/routes/usage.py
```diff
@@ -92,3 +96,93 @@ async def list_usage(
...
+        cost = (prompt_tokens / 1_000_000) * pricing.input_price_per_million + (
+            completion_tokens / 1_000_000
+        ) * pricing.output_price_per_million
+        row.cost = cost
+        total_cost += cost
+        if row.user_id:
+            per_user[row.user_id] = per_user.get(row.user_id, 0.0) + cost
+
+    try:
+        await db.commit()
```
khaledosman: The backfill is neither atomic nor concurrency-safe (security-review instructions §0.2/§0.3):

1. Costs are committed here, then spend is folded in via separate per-user commits below. If anything fails after this commit, rows carry a cost but spend is never charged, and a re-run is a no-op (rows no longer match `cost IS NULL`), so the gap is unrecoverable through this endpoint.
2. Two concurrent backfills of the same model both read the same `cost IS NULL` rows before either commits, and both increment spend: a double charge (the cost write is idempotent, the spend increment is not).

Fix both by doing the row-cost updates and the spend updates in one transaction with a single commit, guarded by a conditional `UPDATE ... WHERE cost IS NULL` (or `SELECT ... FOR UPDATE`) so only one writer wins.

web/README.md
```diff
@@ -0,0 +1,63 @@
+# Otari admin dashboard
+
+A React + [HeroUI v3](https://www.heroui.com) single-page app for the Otari
+gateway's standalone admin panel: create and revoke virtual API keys, manage
```
khaledosman: Stale scope: the dashboard was slimmed to Overview/Usage/Models, but this still says "create and revoke virtual API keys, manage users", lists `/v1/keys` and `/v1/users`, and names Recharts (no longer a dependency). Same stale copy in: the `AGENTS.md` Web Dashboard section, root `README.md` (Admin dashboard section), `web/index.html` meta description, `web/src/components/Login.tsx:44`, `deploy/railway/README.md` ("Pricing page" is now Models), and the `src/gateway/main.py` dashboard comment.

.github/workflows/otari-dashboard.yml
```diff
@@ -0,0 +1,57 @@
...
+
+    - name: Test
+      run: npm test
+
+    - name: Build
+      run: npm run build
+
+    - name: Verify committed bundle is up to date
+      run: |
+        if ! git diff --quiet -- ../src/gateway/static/dashboard; then
```
khaledosman: `git diff --quiet` only sees tracked files: a build that only *adds* a file (e.g. a new `web/public/` asset copied verbatim into the bundle) passes this check while the committed bundle misses it. Check `git status --porcelain -- ../src/gateway/static/dashboard` and fail if non-empty. Related: `pyproject.toml` package-data only covers `index.html`, `favicon.svg`, and `assets/*`, so a new top-level public asset would also be dropped from the wheel.

web/src/pages/UsagePage.tsx
```diff
@@ -0,0 +1,66 @@
...
+            entries.map((entry) => (
+              <Tr key={entry.id}>
+                <Td className="text-[var(--otari-muted)] whitespace-nowrap">{formatDateTime(entry.timestamp)}</Td>
+                <Td className="break-all">{entry.model}</Td>
+                <Td className="text-[var(--otari-muted)]">{entry.provider ?? "—"}</Td>
+                <Td className="font-mono text-xs break-all">{entry.user_id ?? "—"}</Td>
+                <Td className="text-right text-[var(--otari-muted)]">{formatNumber(entry.prompt_tokens)}</Td>
+                <Td className="text-right text-[var(--otari-muted)]">{formatNumber(entry.completion_tokens)}</Td>
+                <Td className="text-right">{formatNumber(entry.total_tokens)}</Td>
+                <Td className="text-right">{formatCost(entry.cost)}</Td>
```
khaledosman: `formatCost(null)` renders `$0.00`, so unpriced rows (exactly what the backfill feature exists to surface) are indistinguishable from genuinely free requests. Render `—` for null cost, as the provider/user columns do.

web/src/pages/ModelsPage.tsx
```diff
@@ -0,0 +1,456 @@
...
+    setPricing.mutate(
+      {
+        model_key: row.key,
+        input_price_per_million: Number(input),
+        output_price_per_million: Number(output),
+      },
+      {
+        onSuccess: () => {
+          setEditing(false);
+          if (row.requests > 0) {
```
khaledosman: `row.requests` is computed from only the last 500 usage rows, so a model whose unpriced traffic is older than the window never gets the backfill offer after pricing. Backfill is a safe no-op when there is nothing to do; offer it unconditionally, as `AddModelForm` does.

src/gateway/main.py
```diff
@@ -28,6 +30,8 @@
 # Public, unauthenticated static assets that are safe for shared caches and set
 # their own Cache-Control; the middleware leaves their caching headers alone.
 _CACHEABLE_PATHS = ("/favicon.svg",)
+# Hashed dashboard bundles under /assets are immutable; let them be cached too.
```
khaledosman: Nit: this comment claims the hashed bundles are cached as immutable, but nothing sets a positive `Cache-Control`; StaticFiles only sends ETag/Last-Modified, so browsers heuristically cache and revalidate. Adding `Cache-Control: public, max-age=31536000, immutable` on the `/assets` mount would realize the claimed benefit.

njbrake:
@khaledosman thanks for the review! I slimmed down the PR a whole bunch to just get an initial implementation off the ground: future PRs can then build on top and add more pages for monitoring things like the API keys, usage etc. I think it's now ready for you to take another look and then should be good to merge.

src/gateway/services/pricing_service.py
```diff
@@ -60,22 +60,45 @@ def normalize_effective_at(value: datetime | None) -> datetime:
...
+    Memoized per (provider, model, day); see ``_resolve_genai_price_uncached``
+    for the matching rules.
+    """
+    key = (provider, model, as_of.date())
+    if key in _price_cache:
+        return _price_cache[key]
+    result = _resolve_genai_price_uncached(provider, model, as_of)
+    if len(_price_cache) >= _PRICE_CACHE_MAX:
+        _price_cache.clear()
+    _price_cache[key] = result
```
khaledosman, line 108: This memoizes the generic-exception path too: `_resolve_genai_price_uncached` returns `None` from its `except Exception` branch, and that `None` is now cached under `(provider, model, day)`. On main, a genai-prices hiccup degraded a single request to "unpriced" and the next request retried; now it sticks until the date changes or the cap clears the cache. On the billing path with `default_pricing=true` and `require_pricing=false`, that is up to a day of unmetered $0 usage for the affected model (security instructions §0.4); with `require_pricing=true` it is a day of 402s. Low probability since the dataset is bundled and local, but the fix is cheap: cache successes and genuine `LookupError` misses, and skip caching when the uncached call hit the exception branch (e.g. have it return a distinct sentinel).

src/gateway/api/routes/aliases.py
```diff
@@ -0,0 +1,150 @@
...
+async def set_alias(
+    request: AliasRequest,
+    db: Annotated[AsyncSession, Depends(get_db)],
+    config: Annotated[GatewayConfig, Depends(get_config)],
+) -> AliasResponse:
+    """Create or update a stored alias."""
+    stored = await refresh_alias_cache(db)
+    _validate(config, request.name, request.target, stored)
+
+    alias = (await db.execute(select(ModelAlias).where(ModelAlias.name == request.name))).scalar_one_or_none()
```
khaledosman, line 106: Nit: two concurrent POSTs creating the same new name both pass this `scalar_one_or_none` check; the loser hits the primary-key constraint and surfaces as a generic 500 "Database error" instead of a clean retry or 409. `stage_override` in the settings route has the same shape for a new key. Single-operator surface, so fine to leave, just noting it is a known 500.

tests/integration/test_settings_toggle.py
```diff
@@ -0,0 +1,150 @@
...
+
+    async def override_get_db() -> AsyncGenerator[AsyncSession, None]:
+        async with async_session_factory() as session:
+            yield session
+
+    app.dependency_overrides[get_db] = override_get_db
+    try:
+        with TestClient(app) as client:
+            yield client
+    finally:
```
khaledosman, line 55: This `_make_client` (and its copies in `test_providers_endpoint.py` and `test_model_metadata_endpoint.py`) kept the engine-dispose half of the pattern from `test_model_discovery.py` but dropped the `Base.metadata.drop_all(bind=engine)` teardown. Every app startup commits a bootstrap `User` + `APIKey` row (`bootstrap_api_key` defaults to true), and `test_models_carry_context_window` commits a `ModelPricing` row for `openai:gpt-4o` that nothing deletes, so state leaks into the shared session-scoped Postgres and becomes an ordering hazard under `-n auto`. Restoring the `drop_all` in the three `finally` blocks matches the file this was copied from. Extracting the now four-fold duplicated helper into `tests/integration/conftest.py` would also prevent the next copy from drifting.

tests/integration/test_alias_api.py
```diff
@@ -0,0 +1,277 @@
...
+    return captured
+
+
+# ---------------------------------------------------------------------------
+# Auth and CRUD
+# ---------------------------------------------------------------------------
+
+
+def test_alias_routes_require_master_key(client: TestClient) -> None:
+    assert client.get(ALIASES).status_code == 401
```
khaledosman, line 97: These auth tests only cover the no-key case. None of the new management endpoints except `/v1/models/discoverable` is tested against a *valid non-master user key*, so swapping `verify_master_key` for `verify_api_key_or_master_key` on `POST /v1/aliases`, `PATCH /v1/settings`, `GET /v1/providers`, or `GET /v1/models/metadata` (a one-token diff that would hand any tenant key control of routing and billing toggles) would not fail a single test. `test_discoverable_requires_master_key` already shows the pattern: mint a key via `/v1/keys`, assert 401 here while it still works on `/v1/models`. One parametrized test across the five new handlers would close this; `DELETE /v1/aliases/{name}` currently has no auth test at all.

tests/integration/test_hybrid_mode_surface.py
```diff
@@ -59,6 +59,28 @@ def test_hybrid_mode_disables_local_management_endpoints(monkeypatch: pytest.Mon
     reset_db()
 
 
+def test_hybrid_mode_disables_dashboard_management_endpoints(monkeypatch: pytest.MonkeyPatch) -> None:
```
khaledosman, line 62: Good stub coverage for the four prefixes, but only via GET. The stubs declare all methods, and the higher-risk case is a state-changing `PATCH /v1/settings` or `POST /v1/aliases` in hybrid; one loop over methods would pin that. Also nothing asserts `/v1/models/metadata` and `/v1/models/discoverable` are unreachable in hybrid (re-registering `models.router` there would go unnoticed), and nothing asserts hybrid `/` serves the tutorial rather than the dashboard; the existing fallback test covers the missing-bundle branch, which is a different condition.

pyproject.toml
```diff
@@ -80,7 +80,13 @@ where = ["src"]
 include = ["gateway*"]
 
 [tool.setuptools.package-data]
-gateway = ["templates/*.html", "templates/*.svg"]
+gateway = [
+    "templates/*.html",
+    "templates/*.svg",
+    "static/dashboard/index.html",
+    "static/dashboard/favicon.svg",
+    "static/dashboard/assets/*",
```
khaledosman, line 88: Two notes on the package-data expansion: (1) it still enumerates exact paths, so a future top-level file in `web/public/` (anything beside `favicon.svg`) would be copied into the committed bundle, pass the CI staleness check, and silently not ship in the wheel; (2) no test can catch that class of bug, because `test_dashboard_build.py` resolves `importlib.resources.files("gateway")` to the source checkout, so deleting `static/dashboard/assets/*` from this list would leave every test green. A CI step that builds the wheel and asserts the bundle files are inside would cover both.

web/src/pages/ModelsPage.tsx
```diff
@@ -0,0 +1,1466 @@
...
+        title="Models"
+        description="Browse the models your providers can reach, inspect their pricing and capabilities, and manage aliases."
+        action={
+          <Button variant="primary" onPress={openAdd}>
+            {showForm ? "Hide form" : "Add"}
+          </Button>
+        }
+      />
+
+      {showForm ? <AddForm onClose={() => setShowForm(false)} initialMode={addMode} initialTarget={addTarget} /> : null}
```
khaledosman: `AddForm` copies `initialMode`/`initialTarget` into `useState` once, and this mount site has no `key`, so "Alias to this model" silently no-ops whenever the form is already open: `openAliasTo` sets the state and calls `setShowForm(true)`, but the mounted form keeps its old tab and target (open the form on "Price a model", click "Alias to this model" on a row: nothing visibly happens; or with the alias form open for model A, clicking the button on model B keeps A as the target). The existing test only covers the closed-form path. Fix: `key={`${addMode}:${addTarget}`}` on this element, or lift the tab/target state out of the form.

web/src/pages/ModelsPage.tsx
```diff
@@ -0,0 +1,1466 @@
...
+
+      <ErrorBanner error={models.error ?? pricing.error ?? aliases.error} />
```
khaledosman: Errors from `useDiscoverableModels`, `useProviders`, and `useModelMetadata` are never surfaced; only these three feed the banner. If `/v1/models/discoverable` returns 500, every discovered-only model vanishes with no banner, and with discovery off the empty state claims "No models yet. Check provider credentials in config.yml" when the real problem is a failed request. The `DiscoveredErrors` banner only covers per-provider `ok: false` inside a successful response. Including the other three query errors here (or a dedicated notice) would close the one error-vs-empty gap in an otherwise careful UI.

web/src/components/Table.tsx
```diff
@@ -0,0 +1,79 @@
...
+}: {
+  children: ReactNode;
+  className?: string;
+  // When set, the row is clickable (used for row-selection tables).
+  onClick?: () => void;
+  selected?: boolean;
+}) {
+  return (
+    <tr
+      onClick={onClick}
```
khaledosman, line 62: The clickable `<tr>` has no `tabIndex`, no key handler, and no focus styling, and row selection is the only way to reach the detail panel, which is the only place to edit or clear a price and the only "Alias to this model" entry point. A keyboard-only or screen-reader operator cannot select a row at all (WCAG 2.1.1). `aria-selected` on a plain table row is also not announced, and the sortable headers never set `aria-sort` (sort state is conveyed only by the glyph). Minimum viable fix: `tabIndex={0}`, Enter/Space handler, a visible focus ring, and `aria-sort` on the sorted `<th>`.

web/src/components/AppShell.tsx
```diff
@@ -0,0 +1,222 @@
...
+const COLLAPSED_SIDEBAR = 60;
+const SIDEBAR_WIDTH_KEY = "otari.dashboard.sidebarWidth";
+const SIDEBAR_COLLAPSED_KEY = "otari.dashboard.sidebarCollapsed";
+const SIDEBAR_STEP = 16;
+
+const clampSidebar = (width: number) => Math.min(MAX_SIDEBAR, Math.max(MIN_SIDEBAR, width));
+
+function readStoredSidebarWidth(): number {
+  if (typeof window === "undefined") return DEFAULT_SIDEBAR;
+  const raw = window.localStorage.getItem(SIDEBAR_WIDTH_KEY);
```
khaledosman: Unguarded `localStorage` access (here, lines 30/83/89, and the sort-persist effect in `ModelsPage.tsx`): in a browser where storage access throws (Chrome with "block all cookies", some private modes), this `useState` initializer throws and the whole shell white-screens right after sign-in. `AuthContext.tsx` deliberately wraps every storage call in try/catch; the same treatment (or a tiny shared `safeStorage` helper) is needed at these five sites.

web/src/lib/format.ts
```diff
@@ -0,0 +1,115 @@
+export function formatNumber(value: number | null | undefined): string {
+  if (value == null) {
+    return "0";
+  }
+  return new Intl.NumberFormat("en-US").format(value);
+}
+
+export function formatCost(value: number | null | undefined): string {
+  if (value == null) {
```
khaledosman, line 9: `formatCost(null)` still returns "$0.00", and `format.test.ts` locks that contract in. Every current call site guards null and renders a dash, so nothing user-visible is wrong today, but the helper's null branch contradicts the dash convention and the next unguarded caller reprises exactly the bug flagged last round (unpriced indistinguishable from free). Returning "—" from the helper (and from `formatNumber(null)`, same pattern) would let the call-site ternaries disappear.

web/src/components/UpdatePrompt.tsx
```diff
@@ -0,0 +1,54 @@
...
+// loaded. The comparison is against the first build this tab ever saw rather
+// than a value baked in at build time: "the app changed while you had it open"
+// is the question worth asking, and a tab opened after a deploy is already
+// current and must stay quiet.
+export function useUpdateAvailable(): boolean {
+  const { data } = useDashboardBuild();
+  const loadedBuild = useRef<string | null>(null);
+
+  if (data && loadedBuild.current === null) {
+    loadedBuild.current = data.build;
```
khaledosman, line 16: The baseline is "first successful poll", not the build the tab actually loaded: if the first `/dashboard-build.json` poll fails (`retry: false`) and a deploy lands before the next tick, the stale tab records the new build id as its own and never prompts. Baking the build id into the bundle at build time (e.g. `import.meta.env` define) would make the comparison exact. Separately, writing `loadedBuild.current` during render is the ref-write-in-render anti-pattern; a `useEffect` on `data` avoids it.

web/src/api/hooks.ts
```diff
@@ -0,0 +1,196 @@
...
+}
+
+// The pricing endpoint caps `limit` at 1000 server-side, so page through it
+// rather than truncating: a gateway with a long price history could otherwise
+// have older rows silently vanish from the models table.
+const PRICING_PAGE_SIZE = 1000;
+
+async function fetchAllPricing(): Promise<PricingResponse[]> {
+  const all: PricingResponse[] = [];
+  for (let skip = 0; ; skip += PRICING_PAGE_SIZE) {
```
khaledosman: The pagination fix from last round looks right, one robustness note: this loop terminates only when a page comes back short, so a backend or proxy that ignores/rewrites `skip` turns it into an unbounded request loop. A max-page cap or a seen-key check would bound it.

khaledosman (approved):
Review by Claude Code (requested by @khaledosman).

Approving: CI is fully green, there are no security or correctness blockers, and everything from the previous round was verified fixed (openapi spec regenerated; `git status --porcelain` now catches untracked bundle files; the pricing hook pages past the 1000-row cap; `/assets` gets a real `Cache-Control: immutable`; the docs/README/AGENTS/Login copy matches the slimmed Models + Settings scope). Master-key handling in the browser is solid: sessionStorage only, key only ever in the `Authorization` header, a 401 anywhere drops it and clears the query cache, and there is no XSS surface. The alias work is the standout on both sides: the validation matrix (shadowing, chaining, provider collisions), routing tested through the real dispatch path, and the pricing-an-alias rejection with proof nothing was written.

The inline comments below are non-blocking, ranked roughly by severity. The ones I would prioritize in a follow-up: the pricing memoization caching transient failures for a day (billing-adjacent), the three integration files that dropped the `drop_all` teardown, the missing user-key authorization tests on the new management endpoints, and the two frontend bugs (stale-prop Add form, unguarded `localStorage`).

Not inline, for the record:

- **models.dev coverage**: every test patches `_fetch`, so the real fetch body (timeout, `raise_for_status`, the 64 MB streaming cap, non-dict JSON guard) and the whole cache policy in `_read_cache` (success TTL, `ttl=0`, the 60s negative TTL, the single-flight lock) never execute in the suite. The cache half is trivially unit-testable in the same style as `test_alias_service.py`.
- **Hybrid mode**: nothing pins that `/v1/models/metadata` and `/v1/models/discoverable` are unreachable in hybrid, only GET (not PATCH) is exercised on the new stubs, and no test asserts hybrid `/` serves the tutorial rather than the dashboard (the existing fallback test covers the missing-bundle condition, which is a different branch).
- **Smaller test notes**: `test_deleted_alias_stops_routing` pins a transport-level `ValueError` where production callers get an unhandled 500; three files use loose `in (401, 403)` where sister tests assert exact 401; the alias routes' `SQLAlchemyError` branches and `run_alias_refresher` have no coverage; `PATCH /v1/settings` has no invalid-value (422) or empty-body test.
- **Smaller frontend notes**: `metadataByKey` mints a new `{}` every render until metadata resolves (invalidates the `modelRows`/`filteredModels` memos), and `releaseCutoff` uses `Date.now()` inside the memo deps; `ModelRow.aliasSource` is computed but never read; `InfoTooltip` hard-codes `id="pricing-info-tip"`, a duplicate-id bug the moment it is instantiated twice; `providerFromModelKey` returns a literal "—" that becomes a provider option in the dropdown once someone prices a colon-free key.

PR #229, opened 2026-06-26
feat(models): config-level model aliases that map to real selectors

### LINKED ISSUE #228: feat(models): config-level model aliases (expose friendly names that map to real selectors)

## Summary

Add **model aliases**: a config-level naming layer that lets an operator expose a friendly model name (e.g. `myopusmodel`) that maps to a real selector (e.g. `anthropic:claude-opus-4`). Users see and call the alias; they never need to know the underlying provider or model. This builds directly on the named provider instances added in #219 (an alias target can be any resolvable selector, including a named instance like `home_lab:qwen3`).

## Motivation

Operators often want to:

- Expose a curated, stable set of model names that stay constant even if the underlying provider/model changes.
- Hide which provider/model is actually serving a request (e.g. show `fastmodel` instead of `openai:gpt-5`).
- Present vendor-neutral names to downstream users.

Today `GET /v1/models` returns provider-prefixed keys (`anthropic:...`, `openai:...`) from discovery and pricing, and requests must use those real selectors. There is no indirection layer.

### Config

A top-level `aliases:` map (string to selector), mirroring the existing `pricing:` map style:

```yaml
aliases:
  myopusmodel: anthropic:claude-opus-4
  fastmodel: openai:gpt-5
  housemodel: home_lab:qwen3   # target may be a named instance from #219
```

Validated at startup: each target must resolve; alias names must not collide with a configured provider instance name; alias targets are not themselves aliases (no chaining for v1).

### Resolution

Add alias lookup at the top of `resolve_provider_selector` (`services/provider_kwargs.py`). This is the single chokepoint already used by every route (chat, messages, responses, embeddings, images, audio, rerank, moderations, batches, vision), so one change covers all surfaces. When the incoming model matches an alias, substitute the target selector, then proceed with the existing instance/implementation resolution.

### Decisions (confirmed)

1. **Billing key: real target.** Aliases are pure naming indirection. Usage logs, pricing, and budgets key on the resolved `instance:model` (e.g. `anthropic:claude-opus-4`). Pricing is configured once for the real model and inherited by every alias pointing at it.
2. **List visibility: additive.** Alias entries appear in `GET /v1/models` (and `GET /v1/models/{id}`) alongside discovered models, with pricing looked up from the real target. To expose *only* curated alias names, the operator sets `model_discovery: false` so the full provider catalog is not listed. No new flag needed.
3. **Response echo: rewrite to alias.** The response `model` field (non-streaming and streaming chunks) is rewritten back to the alias the caller used, so a request to `myopusmodel` also returns `myopusmodel`. The real model name never appears in the response. Behavior is unchanged for non-alias requests.

## Scope / files

- `core/config.py`: `aliases` field + startup validation.
- `services/provider_kwargs.py`: alias substitution in `resolve_provider_selector`; carry the alias name on `ResolvedProvider` for response rewriting.
- `api/routes/models.py` + `services/model_discovery_service.py`: surface alias entries in listing.
- Response `model` rewrite at the response-construction points (chat stream/non-stream, messages, responses, etc.).
- Docs: `docs/models.md`, `config.example.yml`.
- Tests: unit for resolution/validation, integration for listing and request routing.

## Out of scope (v1)

- Alias chaining (alias pointing at another alias).
- Per-alias independent pricing/tracking (would require billing on the alias key instead of the real target).
- A flag to suppress aliased real models from the list while discovery is on.

Relates to #219, #213.

njbrake:

Closes #228. Builds on named provider instances (#219).

## What

Adds a top-level `aliases` map (display name to target selector) so an operator can expose a friendly, stable model name and keep the underlying provider/model hidden:

```yaml
aliases:
  myopusmodel: anthropic:claude-opus-4
  fastmodel: openai:gpt-5
  housemodel: home_lab:qwen3     # target may be a named instance from #219
```

A request whose `model` is an alias routes to its target. The alias is what callers see in `GET /v1/models` and in the response `model` field; the real model name never appears on the wire.

## Design decisions

1. **Billing keys on the real target.** Aliases are pure naming indirection: usage logs, pricing, and budgets key on the resolved `instance:model`. Configure pricing once for the real model and every alias pointing at it inherits it.
2. **Listing is additive.** Alias entries appear in `GET /v1/models` (and `GET /v1/models/{id}`) with pricing read from the target, reported as `owned_by: otari`. Only on the unfiltered listing; a `?provider=` filter will not leak the mapping. To expose only curated alias names, set `model_discovery: false`.
3. **Response model is relabeled to the alias.** The `model` field on both non-streaming results and streaming chunks is rewritten to the alias the caller used.

Aliases are standalone-mode only, like named instances: in hybrid mode model resolution and routing are owned by the otari.ai platform.

## Implementation

- `core/config.py`: `aliases` field, `resolve_alias()`, and `validate_aliases()` (called at load). Fails fast on empty or prefix-less targets, unknown provider prefixes, names colliding with a provider instance, and alias chaining.
- `services/provider_kwargs.py`: alias substitution in `resolve_provider_selector` (the single chokepoint used by every route), with the alias carried on `ResolvedProvider` for relabeling.
- `api/routes/models.py`: alias entries in the listing and single-model endpoints, pricing looked up from the resolved target.
- `streaming.py` + `_pipeline.py`: a `relabel_model()` helper threaded as `display_model` through the standalone non-stream and stream run paths. Inert for non-alias requests.
- Docs: `docs/models.md` and `config.example.yml`.

## Out of scope (v1)

Alias chaining, per-alias independent pricing, and a flag to hide aliased real models while discovery is on.

## Testing

- 17 unit tests (resolution, validation, relabeling) and 8 integration tests (listing, routing to provider and named-instance targets, response echo, provider-filter exclusion).
- `ruff`, `mypy --strict`, and the OpenAPI freshness check pass. Full unit suite: 555 passed. The alias integration suite and related route/streaming/usage suites pass against PostgreSQL.

## Reviewer note

I chose `owned_by: "otari"` for aliased listing entries to avoid leaking the real provider. Happy to make this configurable per alias if preferred.

## PR Type

- [x] New Feature

## Checklist

- [ ] If the API contract changed, I regenerated the OpenAPI spec (no API contract change).

## AI Usage

- [ ] No AI was used.

## Summary

- Added a new “model aliases” feature that lets callers use friendly, vendor-neutral model names.
- Aliases are resolved at startup and then applied consistently across model routing and API responses (chat, embeddings, moderations), including both streaming and non-streaming.
- Updated `/v1/models` so alias models appear in listings (with inherited pricing from the real target) and `/v1/models/{alias}` returns metadata for the alias.
- Updated documentation and the example config to describe alias behavior, limitations (standalone mode only, no alias chaining), and discovery controls.
- Added unit and integration tests covering alias resolution, response relabeling, listing behavior, routing to provider instances, and validation failures.

## Technical notes

- API responses echo the caller’s alias as the `model` field, while pricing/usage/budgets/billing remain tied to the resolved underlying target.
- Aliases are validated at startup, including format/collision checks and rejecting unsupported alias target prefixes.

### DIFF
--- a/src/gateway/api/routes/embeddings.py
+++ b/src/gateway/api/routes/embeddings.py
@@ -26,6 +26,7 @@ from gateway.services.budget_service import (
 from gateway.services.log_writer import LogWriter
 from gateway.services.pricing_service import find_model_pricing, pricing_required_but_missing
 from gateway.services.provider_kwargs import resolve_provider_selector
+from gateway.streaming import relabel_model
 
 router = APIRouter(prefix="/v1", tags=["embeddings"])
 
@@ -172,4 +173,6 @@ async def create_embedding(
         for key, value in rate_limit_headers(rate_limit_info).items():
             response.headers[key] = value
 
+    if resolved.alias is not None:
+        relabel_model(result, resolved.alias)
     return result
--- a/src/gateway/api/routes/embeddings.py
+++ b/src/gateway/api/routes/embeddings.py
@@ -15,6 +15,7 @@ from gateway.api.routes._helpers import resolve_user_id
 from gateway.api.routes.chat import rate_limit_headers
 from gateway.core.config import GatewayConfig
 from gateway.log_config import logger
+from gateway.model_labeling import relabel_model
 from gateway.models.entities import APIKey, UsageLog
 from gateway.rate_limit import check_rate_limit
 from gateway.services.budget_service import (
@@ -26,7 +27,6 @@ from gateway.services.budget_service import (
 from gateway.services.log_writer import LogWriter
 from gateway.services.pricing_service import find_model_pricing, pricing_required_but_missing
 from gateway.services.provider_kwargs import resolve_provider_selector
-from gateway.streaming import relabel_model
 
 router = APIRouter(prefix="/v1", tags=["embeddings"])
 
@@ -104,7 +104,7 @@ async def create_embedding(
         await refund_reservation(db, reservation)
         raise HTTPException(
             status_code=status.HTTP_402_PAYMENT_REQUIRED,
-            detail=f"No pricing configured for model '{resolved.instance}:{model}'",
+            detail=f"No pricing configured for model '{request.model}'",
         )
 
     provider_kwargs = resolved.kwargs
--- a/src/gateway/api/routes/models.py
+++ b/src/gateway/api/routes/models.py
@@ -85,6 +85,11 @@ def _alias_model(config: GatewayConfig, alias: str, pricing_lookup: dict[str, Mo
     )
 
 
+def _alias_target_keys(config: GatewayConfig) -> set[str]:
+    """Canonical pricing keys of every configured alias target."""
+    return {normalize_pricing_key(config, target) for target in config.aliases.values()}
+
+
 async def _get_pricing_map(db: AsyncSession, provider_filter: str | None = None) -> dict[str, ModelPricing]:
     """Load latest pricing per model_key, optionally filtered by provider prefix."""
     latest_effective = (
@@ -154,9 +159,15 @@ async def list_models(
             )
 
     # Phase 2: pricing-only models (not discovered but have pricing entries).
+    # An alias target is skipped: billing keys on the real model, so aliasing one
+    # forces a pricing entry for it, and publishing that entry here would expose
+    # the very name the alias exists to hide. Whether real models are listed is
+    # governed by ``model_discovery`` (phase 1), never by pricing config.
+    alias_targets = _alias_target_keys(config)
     for model_key, pricing in pricing_map.items():
-        if model_key not in merged:
-            merged[model_key] = _model_from_pricing(pricing)
+        if model_key in merged or normalize_pricing_key(config, model_key) in alias_targets:
+            continue
+        merged[model_key] = _model_from_pricing(pricing)
 
     # Phase 3: configured aliases. An alias is a display name, not a provider, so
     # it is only listed for the unfiltered listing; a ``?provider=`` filter asks
--- a/src/gateway/api/routes/models.py
+++ b/src/gateway/api/routes/models.py
@@ -13,8 +13,14 @@ from gateway.core.config import GatewayConfig
 from gateway.log_config import logger
 from gateway.models.entities import ModelPricing
 from gateway.services.model_discovery_service import discover_all_models, get_model_cache
+from gateway.services.provider_kwargs import normalize_pricing_key
 
 router = APIRouter(prefix="/v1", tags=["models"])
+
+# ``owned_by`` reported for alias entries. Aliases intentionally hide the
+# underlying provider, so the gateway itself is named as the owner rather than
+# the real upstream.
+ALIAS_OWNED_BY = "otari"
 
 
 class ModelPricingInfo(BaseModel):
@@ -57,6 +63,28 @@ def _model_from_pricing(pricing: ModelPricing) -> ModelObject:
     )
 
 
+def _alias_model(config: GatewayConfig, alias: str, pricing_lookup: dict[str, ModelPricing]) -> ModelObject:
+    """Build a ModelObject for a configured alias.
+
+    The alias id is what the caller sees; pricing is looked up from the resolved
+    target's canonical key so an alias shows the real model's price without
+    revealing the provider/model behind it.
+    """
+    target = config.aliases[alias]
+    pricing = pricing_lookup.get(normalize_pricing_key(config, target))
+    return ModelObject(
+        id=alias,
+        created=0,
+        owned_by=ALIAS_OWNED_BY,
+        pricing=ModelPricingInfo(
+            input_price_per_million=pricing.input_price_per_million,
+            output_price_per_million=pricing.output_price_per_million,
+        )
+        if pricing
+        else None,
+    )
+
+
 async def _get_pricing_map(db: AsyncSession, provider_filter: str | None = None) -> dict[str, ModelPricing]:
     """Load latest pricing per model_key, optionally filtered by provider prefix."""
     latest_effective = (
@@ -96,6 +124,9 @@ async def list_models(
     exist in the pricing table are also included for backward compatibility.
     """
     pricing_map = await _get_pricing_map(db, provider_filter=provider)
+    # Snapshot before phase 1 mutates ``pricing_map`` (it pops matched keys), so
+    # alias pricing can still be looked up by the target's canonical key.
+    pricing_lookup = dict(pricing_map)
 
     merged: dict[str, ModelObject] = {}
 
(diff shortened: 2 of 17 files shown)


jarmbruster74:
Took you up on the offer to test this branch (context: [#228 comment](https://github.com/mozilla-ai/otari/issues/228#issuecomment-4905335303) — evaluating otari for an LLM-gateway bake-off where bare model names from real coding harnesses were the blocker).

**Setup:** this branch at 444d6a4, run from source (`uv sync` + `otari serve`), SQLite, one named provider instance (`provider_type: openai`) pointed at a local LM Studio backend, `model_discovery: false`, and three aliases mapping bare names to `lmstudio:<model>` targets — including `claude-sonnet-4-5` aliased to a llama model to simulate the Claude Code / CoWork case.

**Results — everything I was blocked on now works:**

- Bare `model: "llama-3.2-3b-instruct"` on `/v1/chat/completions` routes to the named-instance target with its credentials. Non-streaming and streaming both fine; streamed chunks carry the alias and `[DONE]` arrives.
- Bare `model: "claude-sonnet-4-5"` on `/v1/messages` returns a well-formed Anthropic message, and the response `model` echoes the alias — which matters for Anthropic clients that validate the name looks like `claude-*`. Streaming SSE is consumable end-to-end (`message_start` / `content_block_delta` / `message_stop`), alias in the stream, target never leaks.
- `GET /v1/models` lists the alias ids with `owned_by: otari`.
- Governance keys on the resolved target as described: `/v1/usage` rows show the target provider/model, and computed cost matched my configured per-million pricing exactly.

**Two observations, neither a blocker for this PR:**

1. With `model_discovery: false`, models that have explicit `pricing:` config entries still appear in the listing alongside the aliases (`lmstudio:llama-3.2-3b-instruct` next to its alias). That matches the documented "plus any models you priced explicitly" behavior, but an operator using aliases specifically to hide targets might not expect pricing config to re-expose them.
2. Separate from aliasing: `/v1/messages` rejects `x-api-key`, accepting only `Otari-Key` / `Authorization: Bearer`. Anthropic clients configured with `ANTHROPIC_API_KEY` send `x-api-key`, so they'd get a 401 even with aliases working; `ANTHROPIC_AUTH_TOKEN` (Bearer) works. Happy to file separately if useful.

From my side this closes the gap in #228. Thanks for the quick turnaround on the reopen!

njbrake:
@jarmbruster74 great, thank you for the review and for trying it out. Those are both great findings. I'll wrap the fix for your first item into this PR and I'll create a separate github issue to address your second finding. 🙏

src/gateway/api/routes/embeddings.py
```diff
@@ -172,4 +173,6 @@ async def create_embedding(
         for key, value in rate_limit_headers(rate_limit_info).items():
             response.headers[key] = value
 
+    if resolved.alias is not None:
```
khaledosman, line 176: The 402 above (line 107) still names the resolved target: `No pricing configured for model '{resolved.instance}:{model}'`. With `require_pricing` on (the default) and an unpriced alias target, a caller sending the alias gets back the real `instance:model`, the exact string aliases exist to hide. Echo the caller's `request.model` instead, matching the shared gate in `_pipeline.py` (`No pricing configured for model '{model}'`). Same pattern in `rerank.py:104` and `images.py:109`.

src/gateway/services/provider_kwargs.py
```diff
@@ -116,11 +123,18 @@ def resolve_provider_selector(config: GatewayConfig, model_selector: str) -> Res
...
+    A selector that names a configured alias is first substituted with the
+    alias target, then resolved as usual; the resulting ``ResolvedProvider``
+    carries ``alias`` so response ``model`` fields can be relabeled.
+
     Raises ``ValueError`` / ``AnyLLMError`` (from any-llm) for a selector that
     names neither a configured instance nor a known provider, mirroring the
     prior ``AnyLLM.split_model_provider`` behavior.
     """
-    split = split_selector(model_selector)
+    alias = config.resolve_alias(model_selector)
```
khaledosman, line 134: Substituting here means every caller of this chokepoint now accepts aliases: `/v1/rerank`, `/v1/images`, `/v1/audio/*`, and `/v1/batches` route alias requests too, but only chat/messages/responses/embeddings/moderations relabel. Rerank returns the provider response verbatim and some providers (Voyage, Jina) carry `model` in it, so the hidden name leaks on that surface. Add the same guarded `relabel_model` call to rerank at minimum (it is a safe no-op on objects with no `model` field), or document which surfaces support aliases.

src/gateway/api/routes/models.py
```diff
@@ -57,6 +63,33 @@ def _model_from_pricing(pricing: ModelPricing) -> ModelObject:
...
 
+def _alias_model(config: GatewayConfig, alias: str, pricing_lookup: dict[str, ModelPricing]) -> ModelObject:
+    """Build a ModelObject for a configured alias.
+
+    The alias id is what the caller sees; pricing is looked up from the resolved
+    target's canonical key so an alias shows the real model's price without
+    revealing the provider/model behind it.
+    """
+    target = config.aliases[alias]
+    pricing = pricing_lookup.get(normalize_pricing_key(config, target))
```
khaledosman, line 75: `pricing_lookup` is keyed by raw DB `model_key` values, but this looks up the normalized target. Phase 2 normalizes each DB key defensively (line 168) so a legacy-form row like `anthropic/claude-opus-4` is withheld from the listing, yet that same row won't match here, so its price shows neither on the real model nor on the alias entry. Normalize the lookup map keys the same way phase 2 does.

src/gateway/api/routes/models.py
```diff
@@ -138,6 +187,12 @@ async def get_model(
     config: Annotated[GatewayConfig, Depends(get_config)],
 ) -> ModelObject:
     """Get details for a specific model."""
+    # A configured alias resolves to its own entry, with pricing read from the
+    # resolved target so the underlying provider/model stays hidden.
+    if model_id in config.aliases:
+        pricing_lookup = await _get_pricing_map(db)
```
khaledosman: Nit: this loads the latest pricing row for every model to serve one alias. A single query filtered on `normalize_pricing_key(config, config.aliases[model_id])` would do.

khaledosman (commented):
Reviewed the alias feature end to end (chokepoint substitution, billing keys, listing phases, relabel paths). Solid overall: billing/budget/usage key on the resolved target, validation fails fast, and the `?provider=` filter doesn't leak the mapping. Four findings inline; two of them also apply to files outside this diff:

- `rerank.py:104` and `images.py:109` have the same 402 target-name leak flagged on `embeddings.py`.
- Rerank, images, audio, and batches now accept aliases via `resolve_provider_selector` but never relabel; rerank returns the provider result verbatim and some rerank providers (Voyage, Jina) include `model` in it, so the hidden name is echoed on the wire there.

🤖 Review by Claude Code, requested by @khaledosman.

khaledosman (approved):

PR #233, opened 2026-06-30
feat(hybrid): forward request session_label to platform usage report

khaledosman:
## Description

In hybrid (platform) mode, callers had **no way to attribute gateway-routed spend to a session** without standing up a full OpenTelemetry SDK + OTLP exporter just to set one attribute, which is disproportionate for a batch inference script or a simple agent loop.

This adds an optional `session_label` so a caller can tag a run / experiment / conversation, and Otari forwards it onto the `/gateway/usage` report. The platform attributes the attempt's spend to that session (companion PR below), making gateway-mode cost filterable / groupable in the Usage UI without OTel.

How:

- **Body field** `session_label` on the chat, messages, and responses request schemas (shared `SESSION_LABEL_DESC` / `_MAX_LENGTH` in `_schema_derive.py`, capped at 255). Because the schemas drive the OpenAPI spec, **generated SDKs pick the field up automatically** (no manual SDK changes).
- **Stripped before the upstream provider call** (added to `_GATEWAY_INTERNAL_FIELDS`, same path as the existing `user` / `mcp_servers` / `guardrails` gateway-internal fields), so it never leaks to the provider.
- **Threaded onto every usage-report path** in `_pipeline.py`: streaming success/error (`build_streaming_response`), single-attempt streaming, non-streaming fallback (`run_platform_non_stream`), and the inline all-failed flush (`_flush_pending_usage_reports`). `_report_platform_usage` adds it to the POST body, trimmed, with blank treated as absent. The 255 cap matches the platform's indexed session-label keyword so the value is never truncated downstream.
- **`user` no-op documented** in `docs/hybrid-mode-protocol.md`: the OpenAI-standard `user` field is not used for cost attribution in hybrid mode (it is stripped and not reported), so callers reach for `session_label` instead.
- Regenerated `docs/public/openapi.json`.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Addresses the gateway/SDK portion of mozilla-ai/otari-ai#1161. Companion to platform-side PR mozilla-ai/otari-ai#1231 (accepts `session_label` on `GatewayUsageReport` and lands it on the synthetic Usage span).

## Checklist

- [x] I ran the Definition of Done checks locally (`make lint`, `make typecheck`, `make test`). Note: `make lint` and `make typecheck` pass; for tests I ran the full `tests/unit` suite plus the hybrid-mode integration files (`601 passed`). The complete `tests/integration` suite needs Docker/testcontainers, which is unavailable in this environment, so it will be validated by CI.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

Claude Code (Opus 4.8)

Authored with Claude Code. The maintainer (repo owner) reviewed the design decisions (body field vs header, consistency with the platform-side contract) before implementation.

🤖 Created with [Claude Code](https://claude.com/claude-code)

## Summary
Added an optional `session_label` field to hybrid-mode chat, messages, and responses requests so callers can tag usage by session for billing and reporting. The gateway preserves the label for its internal usage reporting, strips it before sending requests upstream, and includes it in `/gateway/usage` so platform data can be attributed and filtered by session. Docs were updated to clarify that the standard `user` field is not used for cost attribution in hybrid mode, and that `session_label` has no effect in standalone mode.

## Technical notes
- Added `session_label` (optional, max 255 characters) to chat, messages, and responses request schemas.
- Trimmed whitespace and treated blank values as absent.
- Threaded `session_label` through hybrid streaming and non-streaming execution paths, ensuring it’s included in platform usage-report payloads.
- Stripped `session_label` before forwarding upstream provider requests.
- Regenerated OpenAPI so SDKs pick up the new field automatically.
- Updated/added tests for forwarding, trimming/omission, stripping upstream, and correct inclusion in success usage reports (including streaming-specific coverage and stricter call/report count checks).

### DIFF
--- a/src/gateway/api/routes/chat.py
+++ b/src/gateway/api/routes/chat.py
@@ -39,7 +39,7 @@ from gateway.api.routes._pipeline import (
     run_streaming_with_fallback,
 )
 from gateway.api.routes._platform import ResolvedAttempt
-from gateway.api.routes._schema_derive import derive_request_base
+from gateway.api.routes._schema_derive import SESSION_LABEL_DESC, SESSION_LABEL_MAX_LENGTH, derive_request_base
 from gateway.api.routes._tools import _strip_gateway_fields
 from gateway.core.config import GatewayConfig
 from gateway.core.usage import GatewayUsage
@@ -114,6 +114,7 @@ class ChatCompletionRequest(derive_request_base(CompletionParams)):  # type: ign
         ),
     )
     max_tool_iterations: int | None = Field(default=None, ge=1, le=MAX_TOOL_ITERATIONS_CAP)
+    session_label: str | None = Field(default=None, max_length=SESSION_LABEL_MAX_LENGTH, description=SESSION_LABEL_DESC)
 
 
 class _ChatAdapter:
@@ -344,6 +345,7 @@ async def chat_completions(
                     background_tasks=background_tasks,
                     rate_limit_info=ctx.rate_limit_info,
                     tool_ctx=tool_ctx,
+                    session_label=request.session_label,
                 )
             except HTTPException:
                 raise
@@ -421,6 +423,7 @@ async def chat_completions(
             background_tasks=background_tasks,
             config=config,
             rate_limit_info=ctx.rate_limit_info,
+            session_label=request.session_label,
         )
 
     resolved = resolve_provider_selector(config, request.model)
--- a/tests/integration/test_hybrid_mode_chat.py
+++ b/tests/integration/test_hybrid_mode_chat.py
@@ -222,6 +222,11 @@ def test_hybrid_mode_forwards_session_label_and_strips_it_upstream(
     )
 
     assert response.status_code == 200
+    # Exactly one provider call and one usage report — pin the counts so a
+    # regression that double-reports or double-dispatches this single-attempt
+    # request is caught rather than masked by indexing the first entry.
+    assert len(upstream_kwargs) == 1
+    assert len(usage_reports) == 1
     # The label rides the usage report ...
     assert usage_reports[0]["session_label"] == "my-run-personas"
     # ... but never leaks to the upstream provider call.
--- a/tests/integration/test_hybrid_mode_chat.py
+++ b/tests/integration/test_hybrid_mode_chat.py
@@ -153,6 +153,81 @@ def test_hybrid_mode_sets_correlation_id_and_reports_usage(
     ]
 
 
+def test_hybrid_mode_forwards_session_label_and_strips_it_upstream(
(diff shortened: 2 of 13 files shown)


tests/integration/test_hybrid_mode_chat.py
```diff
@@ -153,6 +153,86 @@ async def fake_acompletion(**kwargs: Any) -> ChatCompletion:
     ]
 
 
+def test_hybrid_mode_forwards_session_label_and_strips_it_upstream(
```
dpoulopoulos, line 156: **[minor]** Coverage is concentrated on the chat non-streaming path plus the shared `_report_platform_usage`/`_flush_pending_usage_reports` helpers. Not directly exercised with a `session_label`: (1) the streaming usage-report closures (`build_streaming_response._on_complete`/`_on_error`), and (2) the `run_single_attempt_stream` hybrid path that `messages.py`/`responses.py` use for tool-loop streaming (a path `chat.py` does not take). The wiring is identical and mypy-checked so risk is low, but a streaming assertion and/or a messages-path assertion would close the gap and guard against a future refactor dropping the label on those routes.

src/gateway/api/routes/chat.py
```diff
@@ -114,6 +114,7 @@ def validate_message_structure(cls, v: list[dict[str, Any]]) -> list[dict[str, A
         ),
     )
     max_tool_iterations: int | None = Field(default=None, ge=1, le=MAX_TOOL_ITERATIONS_CAP)
+    session_label: str | None = Field(default=None, max_length=SESSION_LABEL_MAX_LENGTH, description=SESSION_LABEL_DESC)
```
dpoulopoulos, line 117: **[nit]** In standalone mode `session_label` is accepted, validated, then silently stripped with no effect (there is no platform to report to). This is reasonable given the shared schema, and the description does scope the behavior to hybrid mode, but the standalone no-op is left implicit. Consider a half-sentence making it explicit (e.g. "no effect in standalone mode") so a standalone caller is not surprised the label disappears.

dpoulopoulos (approved):
Adds an optional `session_label` request-body field to the chat/messages/responses schemas so callers can attribute hybrid-mode (platform) spend to a run/experiment/session without standing up OpenTelemetry. The label is stripped before the upstream provider call and threaded onto every platform usage-report path (streaming success/error, single-attempt streaming, non-streaming fallback, and the inline all-failed flush), trimmed and omitted when blank, capped at 255 chars. Docs and the OpenAPI spec are updated; unit and integration tests cover forwarding, trimming/omission, and upstream stripping.

PR #250, opened 2026-07-13
perf: async MCP/guardrail URL checks, dedupe provider resolution

tbille:
## Description

Two targeted fixes to the hot `/v1/chat/completions`, `/v1/messages`, `/v1/responses` request path, found while auditing gateway routing performance. Both are covered by new/updated tests. No behavior changes beyond one deliberate status-code change (documented below).

### 1. MCP-server / guardrail URL SSRF check no longer blocks the event loop

`validate_mcp_url` did a blocking `socket.getaddrinfo()` call from inside a Pydantic `model_validator`, which runs synchronously during request-body parsing (before auth, before any dependency). Any request carrying `mcp_servers` or a guardrail `url` override stalled **every other concurrent request** on that worker for the duration of the DNS lookup.

- `validate_mcp_url` is now `async`, reusing the async resolver (`loop.getaddrinfo`) that `validate_outbound_fetch_url` already had.
- The check moved out of the Pydantic validators (which can't `await`) and into the async pipeline: `prepare_gateway_tools` validates every MCP server URL concurrently via `asyncio.gather` (covers both request-body and platform-resolved `mcp_server_ids` servers — both hit the same bug), and `run_input_guardrails` validates a per-guardrail `url` override before use.
- **Behavior change**: a rejected URL now returns **400** instead of Pydantic's **422**. Confirmed no existing test depended on 422 for this path (all `mcp_servers` integration tests use loopback URLs, which are allowed by default). New tests lock in the 400 behavior.

### 2. Provider selector no longer resolved twice per request

`resolve_provider_selector` ran once for the pricing/budget gate and again for dispatch. For Vertex AI instances this meant re-reading and re-parsing the service-account JSON and reconstructing RSA-backed credentials **twice per request**, synchronously, in the event loop.

- `RequestContext` now carries `resolved_provider`, populated once during the gate check.
- All 5 dispatch call sites (chat/messages/responses × stream/non-stream) now reuse it via a new `resolve_dispatch_provider()` helper instead of re-resolving.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [x] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

None.

## Checklist

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:** OpenCode (Claude Sonnet)

The agent audited the chat/messages/responses request pipeline end to end (auth, budget reservation, guardrails, MCP tool setup, provider dispatch, streaming, logging), verified each finding directly against the source before proposing fixes, and implemented the two highest-impact, lowest-risk items with the user's sign-off. All changes verified with `make lint`, `make typecheck`, and the full test suite (1070 passed) before opening this PR.

**NOTE:**

## Summary

- Moved MCP server and guardrail URL safety checks from request parsing/validation into the asynchronous request pipeline.
- Added async, concurrent URL/DNS validation for MCP server URLs and guardrail URL overrides (including guardrails that only apply to output).
- Changed rejected unsafe URLs to return **HTTP 400** instead of **HTTP 422**.
- Resolved the “provider routing” decision once per request and reused it during dispatch to reduce duplicate work.
- Removed the now-obsolete benchmark script.
- Updated docs and expanded route/service/unit tests to cover the async validation flow, provider reuse, and 400 error behavior.

## Benefits

- Improves performance on chat/message/response endpoints by avoiding blocking DNS work during parsing.
- Prevents unsafe URLs more consistently before any downstream use.
- Provides clearer, more consistent client errors for rejected unsafe URLs.


tbille:
Thanks for the review, @coderabbitai.

The one nitpick (`scripts/bench_chat_endpoint.py` — keep `TestClient`'s lifespan open through the `httpx.AsyncClient` yield) is now moot: that file was removed entirely in a follow-up commit at the user's request, so there's nothing left to fix there.

I also shortened the PR title per the "Title check" warning (was ~90 chars, now 64). The "Linked Issues" / "Out of Scope Changes" checks reference issues #1/#2 that aren't linked to this PR (there are no linked issues), so I'm leaving those as-is. Docstring coverage is a pre-existing repo-wide condition unrelated to this change's scope.

khaledosman (commented):
Reviewed the two changes. Both look correct and are well covered by tests: MCP validation runs in `prepare_gateway_tools` after platform-resolved servers are merged in (so nothing reaches a connection unvalidated), and the resolved-provider reuse keys on the same `request.model` selector used at the gate. One forward-looking note inline.

src/gateway/services/guardrails.py
```diff
@@ -165,6 +168,15 @@ async def run_input_guardrails(
...
     async with httpx.AsyncClient(timeout=_DEFAULT_TIMEOUT_S) as client:
         for cfg in input_guardrails:
+            # A caller-supplied `url` override is SSRF-checked here rather than
+            # at request-body-parse time (the check does a DNS lookup that must
+            # be awaited). Unlike GuardrailsNotReachableError below, an unsafe
+            # URL is a malformed request, not a runtime failure: it always
+            # rejects the request (mode-independent) via UnsafeURLError, which
+            # this loop deliberately does not catch — the caller
+            # (apply_input_guardrails) maps it to a 400.
+            if cfg.url is not None:
```
khaledosman: This SSRF check now only covers input-direction guardrails, since it lives inside the `"input" in g.on` filter. The removed parse-time validator checked any guardrail `url` regardless of direction. No exposure today because output guardrails aren't enforced, but when output enforcement lands this won't cover them. Add the equivalent `validate_mcp_url` call to the output path when it's built (or leave a comment here pointing at it).

khaledosman (approved):

tbille:
Addressed Copilot's review feedback in 9843b09:

- `_pipeline.py` (`_validate_mcp_server_urls`): fixed the docstring's inaccurate claim that `asyncio.gather` cancels sibling awaitables when one raises. It doesn't with the default `return_exceptions=False` — only cancelling `gather()` itself cancels its children. Kept `gather` rather than switching to `TaskGroup`, since the sibling DNS lookups that keep running in the background after the first rejection are harmless (no side effects beyond the lookup itself), and `TaskGroup` would require unwrapping an `ExceptionGroup` at the call site for no real behavioral benefit here.
- Replaced `--` used as a prose separator with a semicolon/colon in the two flagged docstrings (`tests/integration/test_url_safety_route.py`, `tests/unit/test_guardrails_service.py`) plus two more instances of the same pattern I found in `_pipeline.py` while fixing the others, per the repo's writing-style guidance in AGENTS.md.

src/gateway/services/guardrails.py
tbille: Good catch, thanks. Fixed in 55e8f8b: the SSRF check now runs over every entry in `guardrails` up front (before the `input` filter), so an output-only guardrail's `url` is validated regardless of whether it's currently evaluated. Added regression tests at both the service layer (`test_output_only_guardrail_url_is_still_ssrf_checked`) and the route layer (`test_unsafe_output_only_guardrail_url_returns_400`) so this doesn't regress again when output enforcement lands.

PR #254, opened 2026-07-15
fix: return 400 for unresolvable model selectors

### LINKED ISSUE #252: fix(routes): unparseable or unknown-provider model name returns a bare 500

## Summary

In standalone mode, a chat request whose `model` is unparseable (no `provider:` prefix, e.g. `nosuchmodel`) or names an unknown provider (e.g. `nobody:model`) returns a bare `500 Internal Server Error`. Both are client input errors and should map to a 400 (or 404) with a helpful detail, like the other request validation paths do.

Found while verifying #229 end to end. Pre-existing: reproduced on `main` as well, so it is not caused by the alias work. Aliases actually mask it for curated names, but any typo'd bare model name still hits it.

## Repro

```bash
curl -s -w " [%{http_code}]" -X POST http://127.0.0.1:8000/v1/chat/completions \
  -H 'Otari-Key: Bearer <master-key>' -H 'Content-Type: application/json' \
  -d '{"model":"nosuchmodel","messages":[{"role":"user","content":"hi"}],"user":"<user>"}'
# Internal Server Error [500]
```

Gateway log shows the uncaught exception:

```
ValueError: Invalid model format. Expected 'provider:model' or 'provider/model', got 'nosuchmodel'
```

The unknown-provider variant raises `any_llm.exceptions.UnsupportedProviderError` from the same call site.

## Cause

`resolve_request_context` (`src/gateway/api/routes/_pipeline.py`) deliberately tolerates an unresolvable selector at the budget gate (`except (ValueError, AnyLLMError)`: "a model we can't parse simply has no pricing"). But the route handlers then call `resolve_dispatch_provider` / `resolve_provider_selector` again for dispatch, and at that second call site nothing catches the `ValueError` / `AnyLLMError` documented as raised by `resolve_provider_selector`, so it escapes as a 500.

Affected standalone routes: chat, messages, and responses via `resolve_dispatch_provider`, plus embeddings, images, audio, rerank, moderations, and batches, which call `resolve_provider_selector` bare before their try blocks.

## Suggested fix

Catch `(ValueError, AnyLLMError)` at the dispatch resolution site (in `resolve_dispatch_provider`, plus the bare call sites in the non-chat routes, or a small shared wrapper) and map to `400` with a detail such as `Unknown or invalid model '<selector>'`. Keep the security posture: do not echo any-llm's full supported-provider list in the response detail.

Two secondary things worth confirming in the fix:

1. The budget reservation taken at the gate is refunded on this path (today the exception escapes between reservation and the settled dispatch path; the estimate is 0 without pricing, so there may be no practical leak, but it should be deliberate).
2. `GET /v1/models/{id}` for an unknown id still returns its current behavior once the shared mapping exists.

AloysJehwin:
## Description

`resolve_provider_selector` raises `ValueError` for a selector with no `provider:` prefix and `AnyLLMError` for an unknown provider. Every dispatch site called it without catching either — a typo like `nosuchmodel` or `nobody:model` escaped as a bare `500 Internal Server Error`. Both are client input errors and should return `400` with a useful detail string.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Closes #252

## Changes

**`src/gateway/api/routes/_pipeline.py`**
- Add `_raise_for_unresolvable_model(model_selector, exc)` helper — single place that maps `ValueError`/`AnyLLMError` → `HTTP 400` with the bad model name in the detail
- Wrap the fallback `resolve_provider_selector` call in `resolve_dispatch_provider` with `try/except` → `_raise_for_unresolvable_model` (covers chat, messages, responses routes)

**`src/gateway/api/routes/{embeddings,images,audio,rerank,moderations,batches}.py`**
- Import `_raise_for_unresolvable_model` and `AnyLLMError`
- Wrap each bare `resolve_provider_selector` call with the same guard

**`tests/unit/test_bad_model_name_returns_400.py`** — 7 new unit tests

## Checklist

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:** Claude (claude-sonnet)

**Any additional AI details you'd like to share:** Used to identify the bug, trace the call path, write the fix and unit tests.

## Summary

- Model/selector resolution failures that previously surfaced as HTTP 500 now return HTTP 400 with the invalid model name included in the response detail.
- Standard and pass-through routes were updated to apply the same client-friendly error mapping across chat, messages, responses, embeddings, images, audio, rerank, moderations, and batches.
- Added centralized handling for “unparseable/unknown” model selectors, including preserving reserved budget behavior when resolution fails in reservation-first flows.
- Expanded test coverage:
  - New unit tests for the selector/provider resolution error-to-HTTP-400 mapping.
  - Updated integration assertions to require the invalid model string in the error detail.
  - Added a regression test for billable pass-through routes to fail fast with HTTP 400 when the model selector can’t be resolved.

## Technical notes

- Introduced an internal `_raise_for_unresolvable_model` helper to convert `ValueError` and `AnyLLMError` into a standardized `HTTPException` (400) that instructs callers to use `provider:model`.
- Updated resolution and pass-through routing code paths to use this helper, including refunding reserved budget when applicable.


njbrake (dismissed):

Thanks for this. The diagnosis is right and the fix is in the right place: I read `resolve_provider_selector` and confirmed it raises `ValueError` for an unparseable selector and `AnyLLMError` for an unknown provider, so catching both and mapping to 400 is correct. Centralizing that in `_raise_for_unresolvable_model` is a good call, and there is no circular import (all six route modules import cleanly).

The blocker is that the PR does not pass the Definition of Done checks the checklist marks as run. Both `make lint` and `make typecheck` fail:

**`make lint`: 7 `I001` import-sort errors.** The new imports were prepended out of order in every modified source file. For example `from any_llm.exceptions import AnyLLMError` lands before `from any_llm import ...`, and the `_pipeline` import drops into the middle of the `gateway.services` block. `ruff check --fix` resolves all 7.

**`make typecheck`: 11 mypy errors.** The decisive one is in `_pipeline.py`: `resolve_dispatch_provider` reports `Missing return statement`. Because `_raise_for_unresolvable_model` is typed `-> None`, mypy believes the `except` branch can fall through without returning a `ResolvedProvider`. Typing the helper `-> NoReturn` (it always raises) fixes that and documents intent. The other 10 errors are the two untyped test helpers (`_make_ctx`, `_make_config`) and their call sites under mypy strict.

Both checks are enforced in CI (`otari-lint.yml`, `otari-typecheck.yml`), so this is red as submitted.

Two smaller items:

- **batches.py has dead code.** The inner catch delegates to the helper, which raises `HTTPException`, not `ValueError`, so the outer `except ValueError` is now unreachable. It can be dropped, collapsing the double `try` into one. CodeRabbit flagged the same thing.
- **audio.py reserves budget before resolving the model.** CodeRabbit noted this; it is worth mentioning but not blocking. The reservation amount is `0.0` and the ordering predates this PR, so no behavior regresses here. If you want resolution to gate reservation, that is a separate, small follow-up.

One suggestion, not blocking: the new tests exercise the helper and `resolve_dispatch_provider` in isolation but nothing asserts a route returns 400 end to end. An integration case (for example POST `/v1/embeddings` with `model="nosuchmodel"`) would lock in the wiring across all six routes.

I have the lint fix, the `NoReturn` fix, the test annotations, and the batches.py dead-code removal staged locally; with those applied, ruff is clean, mypy is clean across 195 files, and the 7 new tests pass. Happy to push them to this branch if that is easier than reworking it yourself.

AloysJehwin:
Thanks for the detailed review @njbrake — addressed all three blockers in the latest commit:

- **`-> NoReturn`** on `_raise_for_unresolvable_model` — mypy now correctly infers the `except` branch always raises, so `resolve_dispatch_provider` has no missing-return error
- **Import sort** — ran `ruff check --fix` across all 7 modified source files + the test file; all `I001` errors cleared
- **Dead `except ValueError` in `batches.py`** — collapsed the double `try` to a single block as suggested (CodeRabbit flagged the same)
- **Test helper annotations** — added `-> MagicMock` return types and typed the `resolved_provider` param; mypy strict clean across all 195 files

`make lint` and `make typecheck` both pass cleanly now.

njbrake:

Thanks for this, @AloysJehwin. The diagnosis was spot on: `resolve_provider_selector` raises `ValueError` for an unparseable selector and `AnyLLMError` for an unknown provider, and those were escaping as bare 500s. The `_raise_for_unresolvable_model` helper and the unit tests were a clean way to fix it.

Heads up on what changed: I've pushed an update to your branch to bring it up to date with `main` and re-target the fix. While this PR was open, #275 landed and consolidated the pass-through routes (audio, embeddings, images, moderations, rerank) onto a shared `run_passthrough` helper, so those routes no longer call `resolve_provider_selector` directly. Your per-route edits no longer applied, and the branch had conflicts in 6 files. That timing was just bad luck, not anything wrong with your patch.

The update keeps your `_pipeline.py` helper and unit tests as-is (they still cover chat, messages, and responses), and moves the same fix to where resolution now happens:

- `_passthrough.py`: guards both `resolve_provider_selector` calls in `run_passthrough`, so all the pass-through routes get the 400. On the audio path, the budget reservation is taken before resolution, so it now refunds that reservation before raising the 400 (this closes the budget leak CodeRabbit flagged on your original audio change; the same latent issue had moved into `_passthrough.py`).
- `batches.py`: unified onto your helper so an unknown provider (`AnyLLMError`), not just an unparseable selector, maps to 400.
- Added a test in `test_passthrough_enforcement.py` covering every billable route (including the audio reserve-first paths), and updated `test_alias_api.py`, which had been asserting the old bare-`ValueError` behavior your fix corrects.

Locally: lint, mypy, the OpenAPI check, and the full unit + integration suites pass. The PR is now mergeable with a focused 7-file diff. I also tweaked the PR title to the `fix:` prefix our conventional-commit check wants.

PR #284, opened 2026-07-15
fix(cli): honor --log-level names and reject unsupported --workers

### LINKED ISSUE #268: cli: --workers is silently ignored and --log-level only accepts integer values
> _Note: this issue was drafted by Claude via back-and-forth with @njbrake. The findings and prioritization are his; the prose is Claude's._

## Problem

Two `otari serve` options are broken:

1. **`--workers` is silently ignored.** `src/gateway/cli.py:108-113` passes an app instance to `uvicorn.run(app, ..., workers=workers)`. Uvicorn only honors `workers` (and `reload`) when given an import string; with an instance it logs a warning and runs a single worker. `--workers 4` appears to work but does not multi-process.

2. **`--log-level` only accepts integers.** `cli.py:45-49` uses `click.Choice([logging.DEBUG, logging.INFO, ...])`, i.e. the numeric constants 10/20/30/40/50. Click matches CLI input as strings, so the natural `--log-level INFO` is rejected and the user must pass `--log-level 20`.

## Suggested fix

For workers, pass an import string or app factory when `workers > 1` (or drop the option). For log level, use `click.Choice(["DEBUG", "INFO", "WARNING", "ERROR", "CRITICAL"], case_sensitive=False)` and map to the numeric level, keeping the numeric forms accepted for compatibility if desired.

njbrake:
> _Note: this PR was drafted by Claude via back-and-forth with @njbrake. The analysis and decisions are his; the prose is Claude's._

## Description

Two `otari serve` options were broken and this fixes both.

`--log-level` used `click.Choice` over the numeric `logging` constants
(10/20/30/40/50), so the natural `--log-level INFO` was rejected and users had
to pass `--log-level 20`. It now accepts the symbolic names DEBUG, INFO,
WARNING, ERROR, CRITICAL (case-insensitive) mapped to their numeric level, and
still accepts numeric strings for backward compatibility.

`--workers` was silently ignored. uvicorn only honors workers greater than 1
when given an import string, but the CLI passes a live app instance, so it
always ran a single worker while appearing to accept `--workers 4`. A factory
path is unsound here: the app is built in-process from a resolved config object,
and its startup hooks (schema init and bootstrap key creation) are not safe to
run once per worker (the bootstrap check-then-insert would race and mint
multiple keys). Instead of pretending, the CLI now fails fast with a clear,
actionable error when workers greater than 1 is requested, and the limitation is
documented in the option help text.

Tests: unit tests cover symbolic log level names (lower and upper case), the
numeric back-compat form, rejection of an invalid level, the default
single-worker start, and rejection of `--workers 4` before the server starts.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes #268

## Checklist

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:** Claude Code (Fable 5)

**Any additional AI details you'd like to share:** Directed by @njbrake against issue #268. The decision to reject multi-worker rather than add an app factory reflects his review. Unit suite, lint, and typecheck run locally; integration tests were not affected (CLI-only change).

### DIFF
--- a/src/gateway/cli.py
+++ b/src/gateway/cli.py
@@ -13,6 +13,30 @@ from gateway.core.config import load_config
 from gateway.log_config import setup_logger
 from gateway.main import create_app
 
+_LOG_LEVEL_NAMES: dict[str, int] = {
+    "DEBUG": logging.DEBUG,
+    "INFO": logging.INFO,
+    "WARNING": logging.WARNING,
+    "ERROR": logging.ERROR,
+    "CRITICAL": logging.CRITICAL,
+}
+
+
+def _parse_log_level(ctx: click.Context, param: click.Parameter, value: str | None) -> int:
+    """Map a symbolic (DEBUG/INFO/...) or numeric log level to its numeric value."""
+    if value is None:
+        return logging.INFO
+    normalized = value.strip().upper()
+    if normalized in _LOG_LEVEL_NAMES:
+        return _LOG_LEVEL_NAMES[normalized]
+    if normalized.isdigit():
+        return int(normalized)
+    choices = ", ".join(_LOG_LEVEL_NAMES)
+    raise click.BadParameter(
+        f"{value!r} is not a valid log level. Choose one of {choices} (case-insensitive) "
+        "or a numeric level such as 20."
+    )
+
 
 @click.group()
 def cli() -> None:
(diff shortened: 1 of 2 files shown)


njbrake:
> _Note: this comment was drafted by Claude via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

@dpoulopoulos @tbille heads up on a behavior change here for when this rolls into otari-ai:

- `otari serve --workers N` with N > 1 used to be silently ignored (uvicorn ran a single worker regardless). It now fails fast at startup with a clear error, because the app is built in-process and the startup hooks (schema init, bootstrap key creation) are not safe to run once per worker. If any otari-ai deployment scripts or process managers pass `--workers`, they should drop the flag or set it to 1; the supported scale-out path is multiple otari processes behind a load balancer.
- `--log-level` now accepts the symbolic names (DEBUG, INFO, WARNING, ERROR, CRITICAL, case-insensitive) in addition to numeric values, which previously were the only accepted form.

PR #297, opened 2026-07-17
feat(ui): manage providers, pricing, and aliases at runtime from the dashboard

njbrake:
## Description

Lets an admin launch otari with almost nothing configured, then add provider API keys, set model pricing, and manage aliases entirely from the web dashboard. No config-file edit or restart required.

On first run with no master key, otari generates one and prints it to the console; the dashboard sign-in screen takes that key. From there the new Providers page can add credentials, which are encrypted at rest and merged into the live provider config by a background refresher.

### Backend

- **Encryption at rest.** `services/secret_box.py` encrypts stored keys with `OTARI_SECRET_KEY` (Fernet). Keys are write-only over the API; responses expose only `last4`. Undecryptable rows (for example after the secret key changes) are flagged rather than crashing.
- **Provider store.** `services/provider_store_service.py` persists credentials and overlays them onto `config.providers` through a TTL refresher (standalone only), mirroring the existing alias refresher. The per-config baseline lives on `config._provider_baseline`.
- **First-run bootstrap.** `services/master_key_service.py` generates and prints a master key when none is set; the hash is stored in `runtime_settings`.
- **API.** `/v1/provider-credentials` CRUD, a per-instance test endpoint and an ad-hoc test-before-save endpoint, and a providers catalog endpoint. All routes are master-key gated and standalone-only (not mounted in hybrid mode).
- **Migration.** New Alembic revision for the `provider_credentials` table.
- **Bug fix.** `/v1/models` returned a 500 when an orphaned pricing key referenced a provider that had been removed. `normalize_pricing_key` now tolerates `AnyLLMError`, so the stale key is skipped instead of failing the request.

### Dashboard

- **Providers page.** Two-tab add form (known vs custom), autopopulation of provider type and base URL, test-before-save, and a clear flag when a stored key can no longer be decrypted.
- **Onboarding.** A first-run panel routes a fresh install to Providers with a short walkthrough.
- **Aliases** moved to their own page. "Make an alias" from a model detail navigates over with the target prefilled.
- **Models page** reworked around browsing plus inline pricing. Removed the add form and the Released and Features columns. Price "Reset" now has a tooltip explaining it restores defaults.
- **Routing.** Adopted `react-router-dom` (HashRouter) in place of hardcoded section switching.
- **Layout polish.** Header with logo and sign-out, main-pane-only scroll, update prompt aligned under the header, a subtle otari.ai link, and CSS-masked secret fields.

### Notes for reviewers

- The Postgres integration tests (`test_provider_store.py`, `test_provider_credentials_api.py`) run in CI under Docker; locally I exercised the same behavior against SQLite.
- Secret fields use `-webkit-text-security` for masking, so the key renders unmasked on Firefox (masked on Chromium and Safari). Called out in case we want a JS fallback later.
- Routing uses HashRouter deliberately, to avoid a server catch-all shadowing `/v1`. Happy to switch to BrowserRouter with clean URLs if preferred.

## PR Type

- [x] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## AI Usage

- [ ] No AI was used.

Implemented by Claude Code through back-and-forth with @njbrake. The reasoning and design decisions are his; the code and prose are Claude's.

## Summary

- Added dashboard-based runtime management for provider credentials, including encrypted storage, CRUD operations, credential testing, provider discovery, and live configuration refresh without restarts.
- Added first-run master-key generation and persistence with safe concurrent startup handling.
- Added provider catalog support, credential migration behavior, validation, optimistic concurrency, and improved error handling.
- Updated dashboard navigation with hash-based routing and added dedicated Providers and Aliases pages.
- Improved model pricing workflows with inline editing, reset-error handling, and clearer missing-pricing guidance.
- Added documentation, API collection updates, database migration, bundled dashboard assets, and extensive backend/frontend test coverage.

## Benefits

Operators can configure providers, aliases, and pricing from the dashboard while keeping credentials encrypted, avoiding configuration-file edits and service restarts.


src/gateway/services/master_key_service.py
```diff
@@ -0,0 +1,88 @@
...
+        existing = await _load_hash(session)
+        if existing:
+            config._master_key_hash = existing
+            return
+        token = generate_master_key()
+        session.add(RuntimeSetting(key=MASTER_KEY_HASH_KEY, value=hash_master_key(token)))
+        await session.commit()
+        config._master_key_hash = hash_master_key(token)
+        _print_banner(config, token)
+    except SQLAlchemyError:
```
khaledosman, line 75: **Concurrent first-run leaves losing workers locked out.** With multiple workers/replicas on a shared DB (the refresher comments target exactly this topology), each process loads no hash, generates a *different* token, and races to INSERT. The loser hits an IntegrityError caught here, rolls back, and returns **without setting `config._master_key_hash`** — so that worker answers every management request with 503, and N banners get printed of which only one token works. Self-heals on restart, but broken on first run.

Re-load the winner's hash in the except before giving up:
```python
except SQLAlchemyError:
    await session.rollback()
    existing = await _load_hash(session)  # another worker won the race
    if existing:
        config._master_key_hash = existing
        return
    logger.warning("Could not persist a generated master key; ...")
```

src/gateway/services/master_key_service.py
```diff
@@ -0,0 +1,88 @@
...
+        return
+    try:
+        existing = await _load_hash(session)
+        if existing:
+            config._master_key_hash = existing
+            return
+        token = generate_master_key()
+        session.add(RuntimeSetting(key=MASTER_KEY_HASH_KEY, value=hash_master_key(token)))
+        await session.commit()
+        config._master_key_hash = hash_master_key(token)
```
khaledosman: `hash_master_key(token)` is computed here and again on line 70. Compute once and reuse:
```python
hashed = hash_master_key(token)
session.add(RuntimeSetting(key=MASTER_KEY_HASH_KEY, value=hashed))
await session.commit()
config._master_key_hash = hashed
```

src/gateway/api/routes/providers.py
```diff
@@ -85,3 +121,312 @@ async def list_providers(
...
+
+@router.post("/provider-credentials", status_code=status.HTTP_201_CREATED, dependencies=[Depends(verify_master_key)])
+async def create_stored_provider(
+    request: CreateStoredProviderRequest,
+    db: Annotated[AsyncSession, Depends(get_db)],
+    config: Annotated[GatewayConfig, Depends(get_config)],
+) -> StoredProviderResponse:
+    """Add a provider at runtime. Storing a key requires OTARI_SECRET_KEY."""
+    _validate_instance(request.instance, request.provider_type)
+    if await get_credential(db, request.instance) is not None:
```
khaledosman, line 339: TOCTOU: this existence check isn't under a row lock (unlike the PATCH path), so two concurrent POSTs for the same instance both pass here, both add, and the second `_commit` raises IntegrityError → generic 500 "Database error" instead of the intended 409. Catch IntegrityError around the commit and map it to 409 to keep double-submits benign.

src/gateway/api/routes/providers.py
```diff
@@ -85,3 +121,312 @@ async def list_providers(
...
+        decrypt_secret(row.encrypted_api_key)
+    except (SecretBoxUnavailableError, SecretDecryptionError):
+        return False
+    return True
+
+
+class CreateStoredProviderRequest(BaseModel):
+    """Create a stored provider. ``api_key`` is write-only and requires OTARI_SECRET_KEY."""
+
+    instance: str = Field(description="Routing key, e.g. 'openai' or a named instance like 'home_lab'.")
```
khaledosman: `instance` has no length/blank validation. `_validate_instance` rejects `:` and `/`, but an empty or whitespace-only string passes and becomes the primary key. Add `Field(min_length=1)` and reject blank/whitespace.

src/gateway/api/routes/providers.py
```diff
@@ -85,3 +121,312 @@ async def list_providers(
...
+        PROVIDER_TYPE_ALIASES.get(request.provider_type, request.provider_type)
+        if request.provider_type
+        else (request.instance or "")
+    )
+    if not impl:
+        raise HTTPException(
+            status_code=status.HTTP_400_BAD_REQUEST,
+            detail="Provide a provider or a provider type to test.",
+        )
+    result = await test_provider_credentials(
```
khaledosman, line 312: The test endpoints issue live outbound requests to a caller-supplied `api_base`, and unlike outbound MCP URLs this path does not go through `url_safety.validate_*`. It's master-key gated (operator trust), so acceptable under the current threat model, but if the master key is ever more exposed than host-network trust this is an internal-port prober. Consider running `api_base` through the existing SSRF check here and in `/{instance}/test`.

khaledosman (commented):
Inline comments below. One correctness item (master-key race on concurrent first run); the rest are smaller actionable fixes.

khaledosman (approved):

PR #315, opened 2026-07-18
feat(auth): accept x-api-key header so Anthropic-native clients work (fixes #253)

### LINKED ISSUE #253: feat(auth): accept x-api-key so Anthropic-native clients work on /v1/messages

Reported by @jarmbruster74 while testing #229 (model aliases): https://github.com/mozilla-ai/otari/pull/229#issuecomment-4972598652

### What happens

Otari resolves the API key from `Otari-Key`, then the legacy `AnyLLM-Key` / `X-AnyLLM-Key` aliases, then `Authorization`, and in every case requires the value to be `Bearer <token>` (`src/gateway/api/deps.py:57-86`). Nothing reads `x-api-key`.

Anthropic's own clients configured with `ANTHROPIC_API_KEY` send the credential as a bare `x-api-key` header. Against Otari those requests get a 401 on `/v1/messages`, even though the endpoint is otherwise Anthropic-compatible and the request is well-formed.

### Why it matters

The premise of #228 (and the aliases work in #229) is letting real coding harnesses point at Otari with the bare model names they already send. With #229, `claude-sonnet-4-5` can be aliased and routed correctly, and the response `model` echoes back a `claude-*` name that Anthropic clients accept. Auth is then the remaining thing standing between an Anthropic-native client and a working setup, and it fails for a reason unrelated to what the operator configured.

`ANTHROPIC_API_KEY` is the canonical Anthropic variable, so this is the default path a user tries first. Failing it with a 401 puts the friction precisely at the on-ramp.

### Current workaround

Use `ANTHROPIC_AUTH_TOKEN`, which is sent as `Authorization: Bearer <token>`. This works today and is documented in `docs/use-with-claude-code.md:46-49`, which currently describes the `x-api-key` gap as intended behavior.

### Decisions to settle before implementing

1. **Scope.** Accept `x-api-key` globally, or only on the Anthropic-shaped `/v1/messages`? Global is a smaller diff, but it means a non-Otari header authenticates the key, user, budget, and pricing management routes. Scoping it to `/v1/messages` is narrower but makes auth extraction route-aware, which the shared `_extract_bearer_token` helper does not currently support.
2. **Token format.** `x-api-key` carries a raw token with no `Bearer ` prefix. `_extract_bearer_token` currently raises a 401 (`Invalid header format. Expected 'Bearer <token>'`) for any value without that prefix, so the raw form needs its own branch rather than joining the existing fallback chain.
3. **Precedence.** If both `x-api-key` and `Authorization` are present and disagree, which wins?

### Surface to touch

- `src/gateway/api/deps.py:57-86`: header extraction and format validation. Shared by `verify_api_key`, `verify_master_key`, and `verify_api_key_or_master_key`, so any change here reaches every authenticated route.
- `src/gateway/main.py:168-173`: CORS `allow_headers` would need `x-api-key` for browser clients.
- `src/gateway/main.py:128-133`: the OpenAPI `ApiKeyAuth` security scheme names a single header. A second scheme would require regenerating the spec (`make openapi-check`).
- `docs/use-with-claude-code.md:46-49`: documents the limitation and would need rewriting.

AloysJehwin:
## Summary

Anthropic SDK clients configured with `ANTHROPIC_API_KEY` send credentials as a bare `x-api-key` header (no `Bearer` prefix). Against otari these requests got a 401 on every route, making `x-api-key` the only thing standing between an Anthropic-native client and a working setup.

This PR adds `x-api-key` as a final fallback in `_extract_bearer_token`, after all existing header checks.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes #253

## Changes

- **`config.py`** — adds `X_API_KEY_HEADER = "x-api-key"` constant
- **`deps.py`** — `_extract_bearer_token` falls back to `x-api-key` (raw token, no `Bearer` prefix) before raising 401; `Authorization: Bearer` wins if both are present
- **`main.py`** — `x-api-key` added to CORS `allow_headers`; `XApiKeyAuth` OpenAPI security scheme added alongside `ApiKeyAuth`
- **`tests/unit/test_extract_bearer_token.py`** — 4 new cases: raw token extraction, `Authorization` precedence over `x-api-key`, canonical `Otari-Key` precedence, no-Bearer-prefix succeeds

## Precedence (unchanged headers win)

`Otari-Key` > `AnyLLM-Key`/`X-AnyLLM-Key` > `Authorization: Bearer` > `x-api-key`

## Decisions from issue #253

1. **Scope** — global across all routes (smaller diff, consistent behaviour)
2. **Token format** — raw token branch added separately from the `Bearer` path; no prefix stripping needed
3. **Precedence** — `Authorization` wins over `x-api-key` if both are present

## Checklist

- [x] lint (`make lint`) passes
- [x] typecheck (`make typecheck`) passes — 219 files clean
- [x] 14/14 unit tests pass (`tests/unit/test_extract_bearer_token.py`)

## AI Usage

Drafted with Claude Code assistance; logic and test cases reviewed manually.

## Summary

- Added support for authenticating with the `x-api-key` header (for Anthropic-native clients), including a new shared constant for the header name.
- Kept the existing authentication order, now with `x-api-key` used as the final fallback (after `Otari-Key`, `AnyLLM-Key`/`X-AnyLLM-Key`, and `Authorization: Bearer`).
- Updated browser/CORS settings and OpenAPI security documentation to recognize `x-api-key`.
- Added unit tests to verify correct token extraction, precedence rules, and handling of raw (non-`Bearer`-prefixed) values, using obviously fake tokens to avoid false positives.


njbrake (approved):

## Approving

The `x-api-key` fallback is correct and well scoped. It solves the real interoperability gap from #253 (Anthropic-native clients send a bare `x-api-key` and were getting a 401 on every route), with no security regression: the raw token flows through the same hash and lookup path as the Bearer credential. Precedence is `Otari-Key` then legacy `AnyLLM-Key` / `X-AnyLLM-Key` then `Authorization: Bearer` then `x-api-key`, verified against the source and the new tests.

What was folded in on top of the original change:
- Regenerated `docs/public/openapi.json` so the `XApiKeyAuth` security scheme is in the committed artifact (`make openapi-check` was failing on the stale spec).
- Added `x-api-key` to the missing-credentials 401 detail message and to the `_extract_bearer_token` docstring.
- Merged current `main` into the branch and regenerated the spec, so it now reflects both the provider-credentials endpoints from #297 and this PR's security scheme. Without that, CI (which builds the PR merged with main) failed the OpenAPI freshness check.

CI is green: lint, typecheck, tests, the OpenAPI freshness job, and SDK codegen all pass.

One housekeeping item that is not a merge blocker: the PR description is missing the `## PR Type` section, which keeps the PR Template Check red and arms the daily auto-close job. Adding that section clears the label and the pending close.

Issue #317, opened 2026-07-20
Surface failing requests so admins notice missing-billing (and other) rejections


njbrake:
> _Drafted by Claude via back-and-forth with @njbrake._

## Problem

When `require_pricing` is on and a model has no price, the gateway rejects those requests with HTTP 402. Today an admin has no way to *see* that requests are failing; they find out only when a user complains. The same blind spot applies to other failure classes (provider 5xx, auth 401, budget 403, model-access 403). The dashboard's `require_pricing` alarm (the app-shell PricingWarning) tells you the config is in a reject state, but not that live traffic is actually being dropped, or how much.

We need a way to monitor failing requests so an admin sees the consequences of missing billing info (and other misconfigurations) without waiting for a support ticket.

## What we already have

Usage is logged per request: `UsageLog` carries `status` and `error_message`, so the data to count and break down failures largely exists. It is just not surfaced anywhere an admin looks.

## Possible directions (not mutually exclusive)

- **A failing-requests view / metric on the dashboard:** recent failures with their reason (402 no pricing, 401 auth, 403 budget or model access, 502 provider) and a failure count or rate over a window. Natural fit alongside usage and analytics (#303) and the request log viewer (#304).
- **A live "N requests rejected in the last hour" signal**, so the reject state reads as an active incident rather than a static config note. Could enrich the PricingWarning banner with a real count.
- **A simple "allow unpriced models" toggle in Settings** as a one-click mitigation: a friendlier framing of `require_pricing` / `default_pricing` so an admin can stop the bleeding immediately while they set prices.

## Why it matters

Missing billing info silently turns into dropped traffic. Making failures visible, and giving a one-click way to stop rejecting, closes the gap between "the gateway is misconfigured" and "the admin knows and can act."

Part of #299. Related: #303 (usage and analytics), #304 (request log viewer).

Issue #301, opened 2026-07-17
Dashboard: API key management page


njbrake:
> _Drafted by Claude via back-and-forth with @njbrake._

Manage per-user API keys from the dashboard via the existing `keys` router.

**Why:** Operators currently mint keys only through the bootstrap flow or the raw API; the dashboard should let them issue, list, and revoke keys.

## Tasks
- [ ] List keys (masked, showing last4, owner, status, created/last-used, expiry).
- [ ] Create a key for a user; show the plaintext once and never again (the master-key banner pattern).
- [ ] Revoke/deactivate and set expiry.
- [ ] Never render a full key after creation; mirror the write-only pattern from provider credentials.
- [ ] Hooks, a `KeysPage`, tests; rebuild bundle.

**Depends on:** user management (keys belong to users).

Part of #299.

Issue #300, opened 2026-07-17
Dashboard: user management page


njbrake:
> _Drafted by Claude via back-and-forth with @njbrake._

Add a Users page to the standalone dashboard, backed by the existing `users` router.

**Why:** API keys and budgets attach to users, so user management is the foundation for API key management and budget assignment. It is also where the per-user default for model access control lives, so a user default can flow to all of that user's keys.

## Tasks
- [ ] List users with status (active/inactive) and their assigned budget, if any.
- [ ] Create a user; edit; deactivate and reactivate.
- [ ] Surface the per-user identifier used by `reject_user_mismatch` routing.
- [ ] Wire `/v1/users` into `web/src/api/hooks.ts`; add a `UsersPage`.
- [ ] Component tests; rebuild the committed bundle.

## Model access: the per-user layer (second layer)

Per-key model access control ships first with the API keys work: a per-key `allowed_models` allow-list enforced on `/v1/models` and on inference, backed by `services/model_access.py`. This issue adds the **per-user default layer** so the two-layer model is complete.

- [ ] Add `User.allowed_models` (nullable JSON) as the default allow-list every one of that user's keys inherits. `null` means unrestricted, `[]` means deny-all, a non-empty list is canonical `instance:model` entries with `instance:*` and `instance:prefix*` wildcards (same grammar as the per-key layer). Alembic migration.
- [ ] Effective-set resolution: extend `effective_allowlist(api_key, user)` in `services/model_access.py` so a key with no list of its own inherits the user default. The resolver signature was written for this, so it is an additive change, not a rewrite.
- [ ] Narrow-only override: a key must never broaden its user's grant. Enforce with **validate-subset-on-write** (the keys API rejects a key `allowed_models` that is not a subset of the user's set), not runtime intersection, so a stored key list always reflects what it truly permits.
- [ ] User-default editor on the Users page: reuse the 3-mode scope control (Any model / Only selected / Block all) and the model token input built for the per-key control.
- [ ] Tests: a user default flows to a key with no list of its own; a key list that exceeds the user's set is rejected on write; unit and integration coverage mirroring the per-key layer.

Pairs with API key management on a shared "Users and keys" section if that reads better in the UI.

Part of #299.

Issue #299, opened 2026-07-17
Epic: web dashboard to a great state


njbrake:
Tracking issue for taking the standalone admin dashboard (\`web/\`) from "configure providers, pricing, and aliases" to a complete, trustworthy operator console.

## Scope
The dashboard is standalone-only. Hybrid mode has no local management API (management lives on otari.ai), so none of this applies to hybrid. That is an explicit non-goal.

## Management parity
- [x] #300 User management page
- [x] #301 API key management page (depends on #300)
- [x] #305 Budget management & monitoring (depends on #300)
- [x] #306 Built-in tools & guardrails configuration
- [x] #307 Expand Settings: full config viewer + wider runtime-settable set

## Observability
- [x] #302 Overview / home page
- [x] #303 Usage & analytics page
- [x] #304 Activity / request log viewer
- [x] #309 Provider health monitor

## Security & operations
- [x] #308 Secret-key & master-key lifecycle
- [x] #316 Opt-in SSRF gate for provider api_base

## Dashboard as a product
- [x] #310 End-to-end tests in CI
- [x] #311 Code-split the bundle
- [x] #312 Accessibility, responsive, and UX polish
- [x] #313 Docs: dashboard user guide

## Rough ordering
Users and keys (#300, #301), then overview (#302), then usage + request logs + budgets (#303, #304, #305), then tools/guardrails and settings (#306, #307), then key lifecycle and health (#308, #309), then quality (#310, #311, #312), then docs (#313). #300, #302, and the usage/budget trio are load-bearing for "great state"; #308 and the quality items are what make it trustworthy rather than just functional.

All 14 tracked issues are now closed as COMPLETED.

PR #318, opened 2026-07-20
feat(dashboard): API key management and per-key model access control

njbrake:
https://github.com/user-attachments/assets/bb9b07e5-f149-414c-a2f0-9c65829bc4aa

## Description

Two dashboard features plus polish, on top of the runtime provider management shipped in #297.

**API key management (Keys page).** Issue, list, disable, regenerate, and delete API keys from the dashboard. The one-time secret reveal doubles as an activation moment: it shows a copy-paste first call (curl + OpenAI SDK) with the key and gateway origin injected, works on non-HTTPS origins (a selectable field when the Clipboard API is unavailable), and cannot be dismissed by accident. Adds a display-only `key_prefix` fingerprint (nullable column + migration) populated at all three key-mint sites (create, regenerate, first-run bootstrap). Disable is the primary, reversible revoke; permanent delete is offered only once a key is disabled.

**Per-key model access control.** A key can carry an `allowed_models` allow-list: `null` = any model, `[]` = deny all, or canonical `instance:model` entries with `instance:*` and `instance:prefix*` wildcards. It is enforced on both catalog visibility (`GET /v1/models`, `/v1/models/{id}`) and inference (chat, messages, responses, the pass-through routes, and batches) through one shared matcher fed the same canonical key on both surfaces, so what a key can see and what it can call never diverge. Aliases and legacy pricing ids are canonicalized before matching (no visible-but-blocked gap), enforcement fails closed, and the master key is never restricted. New `services/model_access.py`, a nullable column + migration, and a 3-mode scope control on the Keys page. The per-user default layer is tracked in #300.

**Dashboard polish.** Grouped sidebar nav (Catalog: Providers/Models/Aliases; Access: API keys; Settings standalone) with a divider, distinct provider (server-stack) and API-keys (key) icons, the update prompt reworked into a floating pill over the header, and the `require_pricing` alarm hoisted from the Providers page to a global app-shell banner so it shows wherever an operator is.

Verified with a browser smoke test (Playwright/Chromium against the running gateway): login, key creation + reveal + snippet, the model-scope control, and live enforcement (scoped `list_models`, denied inference 403, denied `get_model` 404).

## PR Type

- [x] New Feature

## Relevant issues

Part of #299 (dashboard epic). Implements #301 (API keys). The per-user model-access layer is tracked in #300. Related: #317.

## Checklist

- [x] I ran the Definition of Done checks locally (`make lint`, `make typecheck`; unit + integration test bodies. Full Postgres integration runs in CI).

## AI Usage

**Any additional AI details you'd like to share:** Planned via a multi-phase review pass, implemented and verified (unit, integration, and a Playwright browser smoke test) by the agent. The reasoning and decisions are @njbrake's; the code and prose are the agent's.

## Summary

- Added dashboard API key management: create, edit, disable, regenerate, delete, and one-time secret reveal.
- Added per-key model access settings, including unrestricted, selected-model, and deny-all modes.
- Enforced model access consistently across model listings, inference requests, and batches.
- Added key fingerprints, database migrations, API documentation updates, and comprehensive unit, integration, and UI tests.
- Refreshed dashboard navigation and presentation with grouped sections, updated icons, a floating update prompt, and a global pricing warning.
- Regenerated dashboard assets and OpenAPI-related examples.


khaledosman (commented):
Reviewed the per-key model-access core against the security checklist: the `is_model_allowed` matcher and its four enforcement sites (catalog, chat/pipeline, batches, pass-through) feed the same canonical `instance:model` key, `None`/`[]` branch on `is None` everywhere, the gate is fail-closed and runs before/refunds the budget reservation, and the migrations are a clean linear chain. Unit suite passes locally. Two items below: one wrong-key overwrite in the Keys page, one test gap.

web/src/pages/KeysPage.tsx
```diff
@@ -0,0 +1,691 @@
...
+        />
+      ) : null}
+
+      {addOpen ? (
+        <CreateKeyForm
+          onClose={() => setAddOpen(false)}
+          onCreated={(result) => setRevealed({ title: "API key created", result })}
+        />
+      ) : null}
+      {editingKey ? <EditKeyForm apiKey={editingKey} onClose={() => setEditing(null)} /> : null}
```
khaledosman: `EditKeyForm` has no `key` prop, and a row's `onClick` does `setEditing(k.id)` without closing the form first. Clicking row B while editing row A reconciles the same `EditKeyForm` instance; its state (`keyName`, `expiresAt`, `allowedModels`) and `ModelScopeControl`'s internal state are `useState(...)`-initialized once, so the form keeps A's values while `apiKey.id` is now B. Save then PATCHes key B with key A's name/expiry/allow-list, silently overwriting B's model access.

Fix: `<EditKeyForm key={editingKey.id} apiKey={editingKey} onClose={...} />` (remount cascades to `ModelScopeControl`).

tests/integration/test_model_access_api.py
```diff
@@ -0,0 +1,169 @@
...
+
+
+def test_get_model_404_for_denied(client: TestClient, master_key_header: dict[str, str]) -> None:
+    _seed_pricing(client, master_key_header, DENIED)
+    scoped = _make_key(client, master_key_header, ["openai:*"])
+    # 404 (not 403): a denied model is indistinguishable from a missing one.
+    assert client.get(f"/v1/models/{DENIED}", headers=scoped).status_code == 404
+
+
+# --- inference gate --------------------------------------------------------
```
khaledosman, line 115: The gate was added to `_passthrough.py` (audio/images/embeddings/moderations/rerank, all billable), but coverage here only exercises chat and batches. A future refactor of `run_passthrough` could drop the gate undetected. Suggest one 403-on-denied test against e.g. `/v1/embeddings`. Alias-target canonicalization at the gate is also untested (shared code path, so lower risk).

khaledosman (approved):

PR #319, opened 2026-07-20
feat(memory): expose platform memory as an agent tool

dpoulopoulos:
## Description
Exposes platform memory as a gateway-managed agent tool: `Tool.MEMORY` (`"otari_memory"`) alongside the existing `otari_web_search` / `otari_code_execution` types.

- Adds `MemoryBackend` (`src/gateway/services/memory_backend.py`), which exposes `memory_search` / `memory_store` / `memory_forget` as model-callable tools and dispatches them to the platform's `/gateway/memory/{search,store,forget}` endpoints (dual-auth: gateway token + user token).
- Adds `CompositeToolBackend` (`src/gateway/services/composite_backend.py`): memory is the one gateway tool designed to run *alongside* whichever other backend is active (MCP, web search, or sandbox) rather than being mutually exclusive with them, so this fans out the tool-loop protocol (`openai_tools`, `owns_tool`, `purpose_hints`, `call_tool`) across an ordered list of backends.
- Wires `_extract_memory_tool` / `_resolve_platform_memory` into `_pipeline.py`'s shared tool-context builder, used by all three request formats (chat completions, Anthropic messages, Responses API) — `otari_memory` is hybrid-mode only (400 if standalone) and gated on the workspace's platform-side memory config (403 if not enabled there).

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

Notes on the checklist:
- Ran `uv run ruff check src tests scripts` (clean), `uv run mypy` (clean, 199 files), and `uv run pytest tests/unit` (615 passed). `tests/integration` requires Testcontainers/Docker Postgres and was not run in this environment.
- `uv run python scripts/generate_openapi.py --check` confirms the committed spec is already up to date (the tool-type string doesn't change the request schema).
- `AGENTS.md` still only documents `otari_code_execution`/`otari_web_search` and hasn't been updated to mention `otari_memory` — flagging as a known gap rather than silently checking the docs box.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:** Claude Code (Claude Sonnet 5)

This PR (and the companion platform-side PR mozilla-ai/otari-ai#1344) were authored by Claude Code across a multi-turn session implementing and wiring up the workspace memory feature.

## Summary

This change adds a new platform-managed “memory” capability exposed as the `otari_memory` tool. It lets the gateway perform three actions—searching saved memories, storing new memories, and deleting memories—while enforcing that memory works only in hybrid mode and only when workspace policy allows it. The gateway now coordinates this memory tool alongside existing tool sources (MCP, web search, and sandbox) during both chat and Responses-style requests, including correct behavior for streaming tool execution.

It also introduces a shared “composite” tool coordinator so the gateway can route tool calls through the right backend in a consistent way, and improves dispatch/error handling (especially for streaming and hybrid fallbacks). Unit tests were added for the composite coordinator and memory components, and integration tests were updated to assert the new composite pooling behavior. `AGENTS.md` still needs updating for `otari_memory`.

## Technical notes

- Memory is resolved via a platform endpoint using dual authentication (gateway token + user token), and the resolved memory backend is injected into per-request tool context.
- Memory tools are included only after workspace/hybrid-mode policy checks pass.
- Dispatch logic was refactored so non-streaming and streaming paths use composite coordination with special care to preserve timing and ensure failures surface correctly.
- Unit tests cover tool exposure/ownership, request wiring, response formatting, and error conditions; integration tests were adjusted for composite pooling assertions.


njbrake:

@dpoulopoulos  a design question on this one before we go further, not a blocker on the implementation quality (the wiring and tests look solid).

Right now `otari_memory` lands as a first-class built-in tool with its own dispatch: `MemoryBackend` advertises `memory_search` / `memory_store` / `memory_forget`, intercepts them in the tool loop, and POSTs to the platform's `/gateway/memory/*` endpoints. That is roughly 550 lines of memory-specific code in the OSS gateway, and it is hybrid-only (a hard 400 in standalone).

The thing I keep coming back to: functionally this is an MCP server. Three tools, remote dispatch, auth headers. And the gateway already knows how to advertise MCP tools, run the loop, and forward calls to a remote server generically via `MCPClientPool`, which implements the same `openai_tools` / `owns_tool` / `purpose_hints` / `call_tool` protocol that `MemoryBackend` reimplements by hand. `McpServerConfig` even already carries every field a memory server needs (`name`, `url`, `authorization_token`, `purpose_hint`, `allowed_tools`).

So the question: why does any memory code need to live in otari at all, rather than memory being an MCP server hosted by otari-ai?

If it were an MCP server, `/gateway/mcp-servers/resolve` could hand back a memory server for enabled workspaces (with a short-lived scoped token minted at resolve time, which also removes the dual-auth token plumbing from the gateway). We could keep the `{type: "otari_memory"}` sugar as a thin flag that just folds that resolved server into the mcp list, so callers get the same one-liner opt-in. Net effect: `Tool.MEMORY`, `MemoryBackend`, and `_resolve_platform_memory` come out of the gateway, the entire memory contract lives in otari-ai and can iterate on a platform deploy instead of a lockstep release across both repos, and standalone stops being a permanent 400 (anyone could point at their own memory MCP server).

The one piece I would keep is `CompositeToolBackend`, but as a generic capability: today MCP is mutually exclusive with web search and sandbox, and letting the loop drive several backends at once is useful for everyone, not just memory.

Is there a constraint I am missing that makes the built-in-tool approach necessary? The two I can think of are (a) the ergonomic opt-in, which the flag preserves, and (b) automatic non-tool recall injection, which is a separate feature and could live on the platform side of the per-request resolve anyway. If neither is load-bearing, exposing memory as a platform MCP server seems like it keeps the OSS gateway free of a feature only one hosted backend can serve. Curious where you land.

njbrake:
Closing while we prioritize the OSS refactor of otari.ai. Work can be resumed as needed as a part of revised roadmap.

PR #320, opened 2026-07-20
test(dashboard): Playwright end-to-end smoke suite in CI

### LINKED ISSUE #310: Dashboard: end-to-end tests in CI

Add Playwright end-to-end coverage for the dashboard.

**Why:** Coverage is component-level only; the multi-page flows (onboarding, add-provider, pricing, aliases) are untested end to end.

## Tasks
- [ ] A Playwright project running against a built bundle served by the gateway.
- [ ] Smoke the core flows: first-run onboarding, add and test a provider, set pricing, create an alias.
- [ ] Wire into the dashboard CI workflow.

Part of #299.

njbrake:
## Description

Adds Playwright end-to-end coverage for the dashboard, the missing layer above the Vitest component tests.

A Playwright project boots a real gateway (`web/playwright.config.ts` `webServer` → `web/e2e/serve.sh`: fresh SQLite, migrated to head, a fixed master key, serving the built bundle in standalone mode) and drives a browser through the core multi-page flows:

- first-run onboarding (Welcome screen before any provider exists),
- adding a provider and seeing it land in the table (onboarding then clears),
- navigating the management pages (Models, Aliases, Settings, Providers),
- creating an alias.

An `e2e` job is wired into the dashboard CI workflow (`.github/workflows/otari-dashboard.yml`): it sets up Python + uv and Node, builds the bundle, installs Chromium, and runs the suite. Vitest is scoped to `src/` so it does not try to collect the browser specs. The committed bundle shifted only because the new dev dependency bumped the minifier; the app is unchanged (54 Vitest tests still pass).

Verified locally: `npx playwright test` → **4 passed** (browser against the gateway-served bundle).

## PR Type

- [x] Infrastructure / CI

## Relevant issues

Closes #310. Part of #299.

## Checklist

- [x] I ran the Definition of Done checks locally (`make lint`, `make typecheck`; web typecheck, Vitest, and the new Playwright suite).
- [x] If the API contract changed, I regenerated the OpenAPI spec (no API change in this PR).

## AI Usage

**Any additional AI details you'd like to share:** The agent built and ran the suite (4 passing browser tests against a live gateway) before opening this. Reasoning and decisions are @njbrake's; the code and prose are the agent's.

## Summary

- Added Playwright end-to-end smoke tests covering onboarding, provider creation, page navigation, and alias creation.
- Added CI automation to build the dashboard and run the browser tests against a fresh gateway and database.
- Added supporting test configuration, scripts, and dependency setup.
- Updated Vitest scope to keep browser tests separate from component tests.
- Refreshed generated dashboard assets after dependency/minifier changes.

## Benefits

- Validates key dashboard workflows across multiple pages.
- Provides repeatable browser coverage in CI.
- Keeps end-to-end runs isolated and deterministic.


khaledosman:
I would recommend testing using vitest and the react-testing-library instead of e2e mocked tests that run against a real browser with playwright due to the common downsides of e2e tests being more brittle and slower, etc.. but considering you're testing against the real backend I see there might be some benefit to it

web/package.json
```diff
@@ -10,7 +10,9 @@
     "preview": "vite preview",
     "typecheck": "tsc -b --noEmit",
     "test": "vitest run",
-    "test:watch": "vitest"
+    "test:watch": "vitest",
+    "e2e": "playwright test",
```
khaledosman: CI builds the bundle before `npm run e2e`, but this script and `serve.sh` never build. Locally the suite serves the committed (possibly stale) bundle from `src/gateway/static/dashboard`, so a dev iterating on `web/src` gets false pass/fail. Suggest `"e2e": "npm run build && playwright test"` so local matches CI.

.github/workflows/otari-dashboard.yml
```diff
@@ -57,3 +57,45 @@ jobs:
           git --no-pager status --porcelain -- ../src/gateway/static/dashboard
           exit 1
         fi
+
+  e2e:
```
khaledosman, line 61: This job boots the real Python gateway, but the workflow only triggers on `web/**` and `src/gateway/static/dashboard/**`. A `src/gateway` backend change that breaks a dashboard-served flow won't run this suite. Either add the relevant backend paths to the trigger or note the scoping is intentional.

khaledosman (commented):
E2E/CI smoke suite looks solid; config is internally consistent, ephemeral SQLite files are gitignored, and Vitest is correctly scoped to `src/`. Two inline notes below. Optional nits: the `dashboard` and `e2e` jobs each run `npm ci` + `npm run build` from scratch (could share via `needs:` + an uploaded artifact), and the `e2e:install` script is unused since CI calls `npx playwright install --with-deps chromium` directly.

.github/workflows/otari-dashboard.yml
```diff
@@ -57,3 +57,45 @@ jobs:
...
+
+    - uses: actions/setup-node@39370e3970a6d050c480ffad4ff0ed4d3fdee5af # v4.1.0
+      with:
+        node-version: 22
+        cache: npm
+        cache-dependency-path: web/package-lock.json
+
+    - name: Install web deps
+      working-directory: web
+      run: npm ci
```
khaledosman, line 94: For consistency with the otari.ai platform (which standardizes on pnpm), `web/` here should move off npm to pnpm. Beyond consistency, pnpm is faster, more disk-efficient (content-addressed store with hard links), and safer against supply-chain attacks by default: its strict, non-flat `node_modules` blocks phantom-dependency access and it does not run install lifecycle scripts unless a package is explicitly allow-listed.

This PR follows the existing npm convention correctly, so don't switch just this new job (that would make the repo internally inconsistent). Instead, do it as a dedicated migration: replace `package-lock.json` with `pnpm-lock.yaml`, update both the `dashboard` and `e2e` jobs (`cache: pnpm`, `pnpm install --frozen-lockfile`, `pnpm run ...`), and update `AGENTS.md`. Fine as a follow-up PR.

khaledosman (commented):
One more, on package-manager consistency (see inline).

khaledosman (approved):

njbrake:
> I would recommend testing using vitest and the react-testing-library instead of e2e mocked tests that run against a real browser with playwright due to the common downsides of e2e tests being more brittle and slower, etc.. but considering you're testing against the real backend I see there might be some benefit to it

@khaledosman Thanks! I'll look into it: yes the idea was that having at least a small subset of playwright driven tests should give highest confidence that things work, but the flakiness and slowness is a definite downside

PR #323, opened 2026-07-20
fix(auth): accept a raw token on the Otari-Key header, not just Bearer

njbrake:
## Description

The dashboard's copy-paste curl snippet sends `-H "Otari-Key: gw-..."` (no `Bearer` prefix), but the gateway required `Bearer` on that header and rejected it with `401 {"detail":"Invalid header format. Expected 'Bearer <token>'"}`. Requiring the `Bearer` scheme inside a header literally named `Otari-Key` is also surprising.

This makes the branded key headers (the canonical `Otari-Key`, and the legacy `AnyLLM-Key` / `X-AnyLLM-Key` aliases) carry the token directly: a `Bearer ` prefix is accepted and stripped for back-compat, but is no longer required. It matches the raw `x-api-key` path already supported for Anthropic-native clients (added in #315). The standard `Authorization` header still requires the `Bearer` scheme, and header precedence is unchanged.

Net effect: `Otari-Key: gw-...` now authenticates (what the dashboard hands out), and `Otari-Key: Bearer gw-...` keeps working.

## PR Type

- [x] Bug Fix

## Checklist

- [x] I ran the Definition of Done checks locally (`make lint`, `make typecheck`; the extract-token unit tests and a new end-to-end raw-header integration test; integration runs isolated on SQLite in the sandbox, so Postgres CI is the real signal).
- [x] Documentation was updated where necessary (no docs affected).
- [x] If the API contract changed, I regenerated the OpenAPI spec (no schema change; auth header parsing only).

## AI Usage

**Any additional AI details you'd like to share:** Diagnosed from @njbrake hitting the 401 with the dashboard's own snippet against a live gateway; he chose the fix direction (make the branded header accept a raw token rather than change the snippet to require Bearer). Reasoning and decision are his; the code and prose are the agent's.

## Summary

- Updated branded and legacy API key headers to accept raw tokens, while continuing to support `Bearer`-prefixed values.
- Kept `Authorization` validation and header precedence unchanged.
- Added unit and integration coverage for raw-token authentication.

This aligns authentication with the dashboard’s generated curl commands while preserving backward compatibility.

### DIFF
--- a/docs/api-reference.md
+++ b/docs/api-reference.md
@@ -19,7 +19,7 @@ For full request/response schemas, see the [OpenAPI spec](public/openapi.json) o
 
 ### Standalone
 
-- Preferred header: `Otari-Key: Bearer <token>`
+- Preferred header: `Otari-Key: <token>` (a `Bearer ` prefix is also accepted)
 - `Authorization: Bearer <token>` is also accepted
 
 Regular API endpoints use an API key. Management endpoints use the master key.
--- a/docs/api-reference.md
+++ b/docs/api-reference.md
@@ -19,7 +19,7 @@ For full request/response schemas, see the [OpenAPI spec](public/openapi.json) o
 
 ### Standalone
 
-- Preferred header: `Otari-Key: <token>` (a `Bearer ` prefix is also accepted)
+- Preferred header: `Otari-Key: <token>` (a `Bearer` prefix is also accepted)
 - `Authorization: Bearer <token>` is also accepted
 
 Regular API endpoints use an API key. Management endpoints use the master key.
--- a/docs/files.md
+++ b/docs/files.md
@@ -18,7 +18,7 @@ target model actually needs.
 
 ```bash
 curl -X POST http://localhost:8000/v1/files \
-  -H "Otari-Key: Bearer <your-api-key>" \
+  -H "Otari-Key: <your-api-key>" \
   -F purpose=user_data \
   -F file=@report.pdf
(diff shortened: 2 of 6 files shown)


khaledosman:
**Reviewed — LGTM, no blocking issues.**

Checked correctness, the security posture, tests, and docs:

- **Precedence preserved.** The new `(Otari-Key, *legacy)` → `Authorization` → `x-api-key` order matches the old flow, and `Authorization` still enforces the `Bearer` scheme. The precedence tests still guard it.
- **No auth regression.** Accepting a raw token only broadens the accepted input format; validation still hinges on the SHA-256 hash lookup, so garbage on a branded header now falls through to the same 401 (via `hash_key`) instead of the format 401. Nothing is left unguarded, and no credential leaks.
- **Edge cases hold.** Empty/absent headers fall through as before; a `Bearer `-only value degrades to the same empty-token behavior the `Authorization` path already had, and `gw-`-prefixed keys never mis-strip.
- **Docs fully swept.** Grepped `docs/`, `README.md`, and `web/` — no stale `Otari-Key: Bearer` snippets left behind.
- **Coverage is right.** The removed malformed-header unit tests were correctly dropped (those inputs are now valid raw tokens), and the new integration test exercises the exact dashboard curl shape plus the Bearer back-compat form end-to-end.

Optional nit (non-blocking): the helper is still named `_extract_bearer_token` though it no longer requires Bearer on the branded headers. It's private and renaming touches three call sites, so leaving it is fine.

khaledosman (approved):

PR #330, opened 2026-07-20
feat(dashboard): activity / request log viewer with per-request latency

### LINKED ISSUE #304: Dashboard: activity / request log viewer

A per-request log/audit view.

**Why:** The aggregate usage page answers "how much"; this answers "what happened", including failures and latency.

## Tasks
- [ ] Recent requests: timestamp, user, model, tokens, cost, latency, status, error.
- [ ] Filter by status/user/model; paginate.
- [ ] Decide the read path (extend the `usage` router vs. a dedicated logs endpoint).
- [ ] Tests; rebuild bundle.

Pairs with the usage & analytics page.

Part of #299.

njbrake:
## Description

Adds the **Activity** page: a per-request log of what the gateway served (tokens, cost, latency, status, errors), plus the backend read path it uses.

Stacked on #322 (`feat/dashboard-budgets`), so this targets that branch rather than `main`. It should be reviewed and merged after #322.

**Latency (`usage_logs.latency_ms`)**

- A monotonic start is captured once at the top of the shared handler preamble (`resolve_request_context`), carried on `RequestContext`, and threaded into `log_usage` through a single `_elapsed_ms` helper. That covers non-streaming success, the failure/refund path, all three streaming settlement callbacks, and the two direct `UsageLog` constructors in `_passthrough.py`.
- Batch rows keep a NULL latency on purpose: a batch is an asynchronous job, so there is no synchronous request duration to record.
- Streaming settles after the stream drains, so the recorded value is total server-side time. It is labelled "Total time" in the UI rather than a bare "latency", so it is not mistaken for time to first token. TTFT is deliberately deferred to a follow up.

**Read path (`/v1/usage`)**

- Additive `status`, `model`, and `endpoint` filters, and `latency_ms` on the response.
- The response stays a bare JSON array, so the existing billing and analytics export contract is unchanged. The paginator's total comes from a new `GET /v1/usage/count` that shares one filter helper with the list, so the count cannot drift from the rows. A regression test guards the bare array shape.
- The migration adds the nullable column plus an `(status, timestamp)` index for the primary "show errors, newest first" query. `model` is high cardinality and is left unindexed on purpose.

**Dashboard**

- Time range presets (24h default), status, endpoint, user, and model filters; failed requests surfaced in the shared red status treatment; an expandable detail row carrying the full provider error text with a copy control; a pager backed by the count endpoint; and distinct empty states for "no matches" versus "nothing recorded yet".
- Activity leads a new **Observability** sidebar section. "Access" covers who may call the gateway, whereas a request log is what it did. That section is where usage analytics and an overview dashboard will live.

**Privacy:** no request or response content is captured anywhere in this change. `UsageLog` stores counts, timing, and provider reported status only.

**Review follow-ups (second commit)**

- The endpoint filter's curated list omitted `/v1/batches` and `/v1/batches/results`, which `log_batch_usage` writes, so batch rows could not be reached by that filter. Both added, with a regression test.
- The pager derived `hasNext` from the total alone, which is zero when the count request fails; a failed count therefore disabled Next and stranded the operator on the first page. It now falls back to "a full page came back, so there is probably more", with a regression test.
- Documented why the vision describe side-call records a NULL latency: that row bills the describe model, so the enclosing request's duration would be misattributed to it.

## PR Type

- [x] New Feature

## Relevant issues

Closes #304. Part of #299.

## Checklist

- [x] I ran the Definition of Done checks locally (`make lint`, `make typecheck`; 726 unit tests; web typecheck and 97 Vitest tests; OpenAPI and Postman checks).
- [x] Documentation was updated where necessary (in-app copy; no external docs affected).
- [x] If the API contract changed, I regenerated the OpenAPI spec (and the Postman collection).

**On CI:** because this PR targets `feat/dashboard-budgets` rather than `main`, the four quality workflows (`otari-tests`, `otari-lint`, `otari-typecheck`, `otari-dashboard`) do not run automatically on it. All four are gated with `pull_request: branches: [ main ]`, so only the PR title, template, and SDK codegen checks fire here. CodeRabbit's auto review is limited to `main` for the same reason. All four workflows have been dispatched manually against branch `304` instead; their results are on the Actions tab and are the authoritative signal for this PR.

**On the integration suite:** Docker and Postgres are unavailable in the sandbox this was built in, and the shared conftest teardown uses `DROP TABLE ... CASCADE`, which SQLite rejects and which then corrupts state for the following test. The new integration tests were therefore validated locally one at a time against a fresh SQLite database (10 of 11 passing). The eleventh, `test_list_usage_response_shape`, is a pre-existing Postgres oriented test that compares an exact timestamp string; SQLite does not preserve the timezone offset, so it reports `09:30:00` against the expected `09:30:00+00:00`. The `latency_ms` value it now asserts matches. The manually dispatched Postgres run covers that file properly.

## AI Usage

**Any additional AI details you'd like to share:** Planned and built through back and forth with @njbrake, who drove the decisions: reusing `/v1/usage` rather than adding a dedicated logs endpoint, deferring TTFT to keep this change focused, building the full filter and drill-down set rather than a thinner first cut, confirming that no call content should ever be captured, and moving Activity out of "Access" into its own Observability section at the top of the sidebar. The plan was reviewed by independent strategy, architecture, and design agents before implementation; that review caught a response envelope that would have broken the export contract, and corrected where latency actually had to be threaded. The reasoning and product calls are his; the code and prose are the agent's.


njbrake:
@coderabbitai review

_Note: this comment was posted by Claude via back and forth with @njbrake. Auto review was skipped on this PR because it targets `feat/dashboard-budgets` rather than `main`, so this is a manual trigger. The branch was just rebased onto the latest #322, and all four quality workflows were dispatched manually and are passing against the current tip._

web/src/pages/ActivityPage.tsx
```diff
@@ -0,0 +1,429 @@
...
+          </FilterSelect>
+          <div className="flex flex-col gap-1">
+            <label htmlFor="filter-model" className="text-xs font-medium text-[var(--otari-muted)]">
+              Model
+            </label>
+            <input
+              id="filter-model"
+              value={modelFilter}
+              onChange={(e) => setModelFilter(e.target.value)}
+              placeholder="e.g. gpt-4o"
```
khaledosman: Model filter is exact-match (`UsageLog.model == model`), but stored model strings may carry a provider prefix, a version suffix, or be an alias. Typing the placeholder `gpt-4o` will silently return nothing if rows are stored as e.g. `openai/gpt-4o`. Consider substring match, or hint at the expected format. Low / UX.

_Claude Code_

web/src/pages/ActivityPage.tsx
```diff
@@ -0,0 +1,429 @@
...
+        <DetailField label="Cache write tokens">{formatTokens(entry.cache_write_tokens)}</DetailField>
+        <DetailField label="Total time">{formatLatency(entry.latency_ms)}</DetailField>
+        <DetailField label="Request ID">{entry.id}</DetailField>
+      </div>
+    </div>
+  );
+}
+
+// A token-styled select matching the app's other filter controls.
+function FilterSelect({
```
khaledosman, line 169: `FilterSelect` is now defined here and in `ModelsPage.tsx`. The two APIs differ (options-array + `aria-label` vs children + visible `<label>`), so it's a judgment call, but a shared primitive in `components/ui.tsx` would stop them drifting. Nit.

_Claude Code_

khaledosman (commented):
Reviewed against the repo standards (backend, frontend, security, performance instructions). This is a clean, thorough, well-tested change: migration chains correctly to the current single head with a reversible downgrade; the default newest-first view is index-served (`timestamp` already carries `index=True`) and the new `(status, timestamp)` composite serves error triage; both `/v1/usage` and the new `/v1/usage/count` are master-key gated and share one filter helper so the count can't drift from the rows; no request/response content is stored or logged; front/back types match and OpenAPI + Postman are regenerated. The two review-follow-up regressions (batch endpoints reachable, Next surviving a failed count) both have tests.

Two minor, non-blocking notes left inline. Nice work.

_Review generated by Claude Code._

khaledosman (approved):

web/src/pages/ActivityPage.tsx
```diff
@@ -0,0 +1,429 @@
...
+      ? `${rangeStart}–${rangeEnd}`
+      : "0"
+    : total === 0
+      ? "0 of 0"
+      : `${rangeStart}–${rangeEnd} of ${total.toLocaleString()}`;
+  // Prefer the exact total, but fall back to "a full page came back, so there is
+  // probably more" when the count request failed. Otherwise a failed count would
+  // strand the operator on page 1 with rows they cannot reach.
+  const hasNext = count.data ? (page + 1) * PAGE_SIZE < total : rows.length === PAGE_SIZE;
+
```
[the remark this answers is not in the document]
njbrake: Agreed, this was a real bug. Fixed in d3786efa.

You are right that `count.data` is not a test for "the current count succeeded": TanStack keeps the last successful data when a refetch errors. The same applies to `keepPreviousData`, which serves the previous filters' total while a new one loads. Either case could show a stale total and wrongly disable Next.

The pager now gates the exact total path on `count.isSuccess && !count.isPlaceholderData` and falls back to the visible row range otherwise.

_Reply drafted by Claude via back and forth with @njbrake._

web/src/api/hooks.ts
```diff
@@ -493,3 +497,46 @@ export function useDeleteUser() {
...
+      const params = usageParams(filters);
+      params.set("skip", String(page * pageSize));
+      params.set("limit", String(pageSize));
+      return apiFetch<UsageEntry[]>(`/v1/usage?${params.toString()}`);
+    },
+    placeholderData: keepPreviousData,
+    // A request log moves constantly; keep it fresh but don't refetch on every focus.
+    staleTime: 10_000,
+  });
+}
```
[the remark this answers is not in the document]
njbrake: Checked this against the installed `@tanstack/query-core` rather than reasoning about it, and I do not think it holds.

`hashKey` is `JSON.stringify` with `Object.keys(val).sort()`, so query keys are hashed **structurally**, not by object identity. A freshly constructed object with the same contents produces the same hash and resolves to the same cache entry.

Runtime check against the pinned version:

```
distinct identities: true
hash(a) === hash(b): true
reordered keys hash equal: true
different value differs: true
```

So an inline `filters` object does not churn the cache or cause an extra refetch. Serializing to a string first would add indirection without changing behavior, so leaving as is. Separately, `filters` is `useMemo`'d at the only call site.

_Reply drafted by Claude via back and forth with @njbrake._

web/src/api/hooks.ts
```diff
@@ -493,3 +497,46 @@ export function useDeleteUser() {
...
+// Total rows matching the same filters, for the paginator's "N of M". A separate
+// request so /v1/usage stays a bare array; run alongside the list.
+export function useUsageCount(filters: UsageFilters) {
+  return useQuery({
+    queryKey: [USAGE, "count", filters],
+    queryFn: () => apiFetch<UsageCount>(`/v1/usage/count?${usageParams(filters).toString()}`),
+    placeholderData: keepPreviousData,
+    staleTime: 10_000,
+  });
+}
```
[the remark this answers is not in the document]
njbrake: Same as the `useUsageLogs` thread: `hashKey` hashes query keys structurally (`JSON.stringify` with sorted keys), so a non-memoized caller passing a structurally identical `filters` object hits the same cache entry rather than churning it. Verified at runtime against the pinned `@tanstack/query-core`. Leaving as is.

_Reply drafted by Claude via back and forth with @njbrake._

web/src/pages/ActivityPage.tsx
njbrake: Good catch on the silent empty result, and the hint is fixed in d3786efa.

One correction on the shape: the model is stored as the resolved short name (`resolved.model`, for example `gemini-2.5-flash`), never provider prefixed, so `openai/gpt-4o` is not a value that occurs. The version suffix case you describe is real though: a row logged as `gpt-4o-2024-11-20` will not match a typed `gpt-4o`.

I fixed the hint rather than the matching semantics, because `model=` is now a public query param on `/v1/usage` and switching it to substring matching would change that contract for the external export consumers the endpoint already serves. The placeholder now reads `exact name, e.g. gemini-2.5-flash` with a tooltip saying it is an exact match on the logged model name.

_Reply drafted by Claude via back and forth with @njbrake._

web/src/pages/ActivityPage.tsx
njbrake: Agreed this is real duplication, and thanks for flagging it.

Leaving it for a follow up rather than folding it into this PR: the two APIs genuinely differ (options array plus `aria-label` versus children plus a visible `<label>`), so unifying them means editing `ModelsPage.tsx`, which this PR does not otherwise touch. That is the kind of adjacent refactor I would rather not smuggle into a feature branch.

Happy to do it as a separate change if you want a shared primitive in `components/ui.tsx`.

_Reply drafted by Claude via back and forth with @njbrake._

PR #334, opened 2026-07-21
refactor(dashboard): extract shared FilterSelect for Models and Activity

### LINKED ISSUE #333: Dashboard: consolidate the duplicated FilterSelect into a shared primitive

## Context

Raised by @khaledosman while reviewing #330 (activity / request log viewer). Deferred there deliberately, because fixing it means editing `ModelsPage.tsx`, which that PR did not otherwise touch. Filing so it is not lost now that the review thread is resolved.

## Problem

There are two separate `FilterSelect` components in the dashboard:

- `web/src/pages/ModelsPage.tsx` (added earlier)
- `web/src/pages/ActivityPage.tsx` (added in #330)

They diverge in both API and appearance:

| | `ModelsPage` | `ActivityPage` |
|---|---|---|
| Options | `options: {value,label}[]` prop | `children` (`<option>` elements) |
| Labelling | `aria-label`, no visible label | visible `<label htmlFor>` |
| Wrapper | none | `div.flex.flex-col.gap-1` |
| Radius | `rounded-md` | `rounded-lg` |
| Background | `bg-white` | `bg-[var(--otari-bg)]` |
| Padding | `px-2 py-1.5` | `px-3 py-2` |

So two selects in the same app render with different corner radius, background, and padding. Separately, the `ModelsPage` variant uses `bg-white`, which the frontend standards steer away from for general chrome in favour of the `--otari-*` tokens.

## Suggested approach

Extract one primitive into `web/src/components/ui.tsx` alongside the other shared pieces, and have both pages use it. Points to settle while doing it:

- Support both labelling modes, for example an optional `label` that renders a visible `<label htmlFor>` and falls back to `aria-label` when omitted, so neither call site loses its accessible name.
- Pick one options API. Children are more flexible (the Activity page groups and conditionally renders options); an `options` array is terser for static lists. Supporting children with an optional `options` convenience is one way to avoid churn at either call site.
- Settle on one set of tokens for radius, padding, and background, and drop `bg-white`.
- Keep the existing Vitest coverage for both pages green; neither page's behaviour should change.

## Notes

Purely a consistency and maintenance cleanup. No user facing behaviour change is intended beyond the two selects looking the same.

Shaurya2k06:
## Description

Extracts the duplicated page-local `FilterSelect` helpers from Models and Activity into one shared primitive in `web/src/components/ui.tsx`, so filter chrome stays consistent across the dashboard.

Raised in review on #330 and deferred there because that PR did not otherwise touch ModelsPage.

**Shared `FilterSelect`**
- Lives in `components/ui.tsx` next to the other small UI primitives.
- Optional visible `label` (+ `id`) for Activity-style labelled filters; falls back to `ariaLabel` for Models-style compact controls.
- Supports both an `options` array (Models) and `children` (Activity's conditional / empty options).
- Single token chrome: `rounded-lg`, `bg-[var(--otari-bg)]`, `px-3 py-2`, plus a brand focus border. Drops Models' `bg-white`.

**Call sites**
- `ModelsPage` and `ActivityPage` import the shared component; local copies removed.
- Dashboard static bundle rebuilt and committed.

No intentional behaviour change beyond the Models filters picking up the shared recessed styling.

## PR Type

- [x] Refactor

## Relevant issues

Follow-up from review on #330.
Fixes #333 

## Checklist

- [x] I have added or updated tests that cover my change (`tests/unit`, `tests/integration`). Existing Vitest coverage for Models and Activity filter selects kept green (29 tests); no behaviour change requiring new cases.
- [x] I ran the Definition of Done checks locally (`npm --prefix web run typecheck`; `npm --prefix web test -- --run src/pages/ModelsPage.test.tsx src/pages/ActivityPage.test.tsx`; `npm --prefix web run build`).
- [x] Documentation was updated where necessary. N/A (internal consistency cleanup).
- [x] If the API contract changed, I regenerated the OpenAPI spec (`uv run python scripts/generate_openapi.py`). N/A (dashboard-only).

## AI Usage

**AI Model/Tool used:** Cursor (Grok)

## Summary

- Consolidated the duplicated dashboard filter controls into a shared `FilterSelect` component.
- Added support for accessible labels, caller-provided or generated IDs, and both option lists and custom children.
- Standardized filter styling across the Models and Activity pages while preserving existing behavior.
- Rebuilt the committed dashboard assets.


njbrake (approved):

Approving. Clean, well-scoped consolidation that does exactly what #333 asks: one shared `FilterSelect` primitive, both call sites migrated, no behavior change beyond the Models filters adopting the shared recessed styling.

Verified locally on the pushed head:

- `npm run typecheck` clean, 29 Vitest tests green (Models 21, Activity 8).
- `npm run build` produces no bundle drift; the committed dashboard assets match source.
- Both call-site APIs are covered by the new component (`ariaLabel` + `options` for Models, `id` + `label` + `children` for Activity).

I pushed one small follow-up: a `useId()` fallback so a caller passing a visible `label` without an explicit `id` still gets a valid `<label htmlFor>` association. That was the one CodeRabbit nit with real merit; the others (Tailwind parenthesized var syntax, `outline-hidden`, HeroUI `Select`) I left alone to stay consistent with the surrounding code and keep the refactor tight.

Thanks @Shaurya2k06.

PR #340, opened 2026-07-21
feat(batches): add a batches table for idempotent accounting, spend folding, and strict ownership

### LINKED ISSUE #290: feat(batches): add a batches table for idempotent accounting, spend folding, and strict ownership
> _Note: this issue was drafted by Claude via back-and-forth with @njbrake. The findings and prioritization are his; the prose is Claude's._

## Problem

PR #273 added enforcement to `/v1/batches` (user resolution, budget gate, rate limiting, metadata-anchored ownership, and token/cost recording on results retrieval), but three gaps were deliberately deferred because the gateway persists no batch state; ownership and accounting are anchored solely on provider-side batch metadata:

1. **Result usage logging is not idempotent.** Every `GET /v1/batches/{batch_id}/results` sums the batch's full token totals and writes a fresh `UsageLog` row, so repeated retrievals (retries, polling clients) overstate usage reporting for the same spend. Flagged by CodeRabbit on PR #273.
2. **Batch cost is never folded into `users.spend`.** Because retrieval can happen any number of times, the charge cannot be deduplicated without persistence, so batch spend is visible in usage logs but does not consume the user's budget.
3. **Ownership is fail-open for unmarked batches.** Batches created before the marker existed, or via providers that do not round-trip metadata, carry no `otari_user_id` and remain accessible to any authenticated key. A strict guarantee needs a local record of who created each batch.

## Suggested fix

Add a `batches` table (batch id, provider, user id, api key id, model, created_at, results_accounted_at or similar) written at creation time. That single record enables:

- idempotent results accounting (bill/log once, on the first completed retrieval),
- folding batch cost into `users.spend` at that same point,
- strict ownership checks that do not depend on provider metadata round-tripping,
- and listing/cancel checks without an extra provider retrieve per call.

Needs an Alembic migration; batches created before the table ships can keep the current metadata-anchored fallback path.

## Related

- #258 (the enforcement gap this hardens further)
- PR #273 (where these were consciously deferred)

njbrake:
## Description

PR #273 anchored `/v1/batches` ownership and accounting solely on provider-side batch metadata, which left three gaps (all called out in #290):

1. Results accounting was not idempotent: every `GET /v1/batches/{id}/results` wrote a fresh `UsageLog`, so retries and polling clients overstated usage for the same spend.
2. Batch cost was never folded into `users.spend`, so batch spend was visible in usage logs but never consumed budget.
3. Unmarked batches (created before the marker, or via providers that do not round-trip metadata) stayed accessible to any authenticated key.

This PR adds a `batches` table written at creation time, and uses it to close all three:

- **Idempotent accounting**: the first completed results retrieval atomically claims `results_accounted_at` (NULL to now via a conditional UPDATE guarded on rowcount); later retrievals return the results without re-logging or re-billing.
- **Spend folding**: on that same first retrieval the batch cost is folded into `users.spend` once, through a new `budget_service.record_external_spend` that writes spend outside the reservation flow (via `reconcile_reservation` with an empty handle, keeping `users.spend` single-writer). It is deliberately not budget-gated, since the spend already happened upstream at the provider.
- **Strict ownership**: retrieve, cancel, results, and list now consult the local record first. Only its owner (or the master key) gets access, regardless of whether the provider round-tripped the `otari_user_id` marker. List does one bulk record lookup instead of a per-batch provider call.

Batches created before this table shipped carry no record and keep the existing metadata-anchored fallback (log every retrieval, no spend fold), as specified in the issue.

Includes a matching Alembic migration (`f3b6d8a1c4e7`, chained to the current head, reversible). The FK columns follow house conventions: `user_id` to `users` with `ondelete=CASCADE` (indexed), `api_key_id` to `api_keys` with `ondelete=SET NULL` (indexed).

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes #290

## Checklist

- [x] If the API contract changed, I regenerated the OpenAPI spec (`uv run python scripts/generate_openapi.py`). No contract change; `make openapi-check` is clean.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

Claude Code (Claude Opus 4.8), driven by @njbrake.

Implementation and tests were drafted by Claude via back-and-forth with @njbrake; the design decisions and review are his. Integration tests were run against a local PostgreSQL 18 instance (Docker/testcontainers was unavailable in the sandbox): 36 batch tests, 734 unit tests, and the budget/user/usage integration suites all pass.

**NOTE:**

## Summary

- Added persistent batch records to track ownership, provider details, models, and one-time accounting status.
- Enforced ownership for batch retrieval, cancellation, results, and listing, while preserving legacy metadata-based behavior.
- Ensured completed batch results are logged and charged only once.
- Added external spend tracking to include batch costs in user spending without budget checks.
- Added migration, supporting services, and integration tests covering accounting, pricing deferral, ownership, and legacy behavior.

## Technical notes

- Accounting claims and spend updates occur atomically.
- Master keys bypass ownership checks.

### DIFF
--- a/src/gateway/api/routes/batches.py
+++ b/src/gateway/api/routes/batches.py
@@ -574,11 +574,22 @@ async def retrieve_batch_results(
     # Recorded batches are accounted exactly once: the first completed retrieval
     # claims the accounting slot, logs usage, and folds the cost into the owner's
     # spend; later retrievals (retries, polling clients) return results without
-    # re-billing. Legacy batches with no record keep the prior behavior (log every
-    # retrieval, no spend fold), since they cannot be deduplicated without a record.
-    should_account = True
-    if record is not None:
+    # re-billing. The claim and the spend fold commit together (below), so if the
+    # spend fold fails or the process dies mid-flight the claim rolls back and a
+    # later retrieval re-accounts, rather than marking the batch accounted but
+    # never billing it.
+    #
+    # A retrieval that finds billable tokens but no pricing row (cost is None)
+    # does not claim the slot, so once pricing exists a later retrieval still
+    # folds the spend once. Legacy batches with no record keep the prior behavior
+    # (log every retrieval, no spend fold), since they cannot be deduplicated.
+    pricing_missing = record is not None and bool(total_tokens) and cost is None
+    if pricing_missing:
+        should_account = False
+    elif record is not None:
         should_account = await claim_batch_accounting(db, batch_id)
+    else:
+        should_account = True
 
     if should_account:
         await log_batch_usage(
@@ -593,8 +604,13 @@ async def retrieve_batch_results(
             total_tokens=total_tokens,
             cost=cost,
         )
-        if record is not None and cost and user_id:
-            await record_external_spend(db, user_id, cost)
+        if record is not None:
+            if cost and user_id:
+                # Folds the spend and commits the claim in one transaction.
+                await record_external_spend(db, user_id, cost)
+            else:
+                # Nothing to bill (no cost/owner); still persist the one-time claim.
+                await db.commit()
 
     return {
         "results": [
--- a/src/gateway/api/routes/batches.py
+++ b/src/gateway/api/routes/batches.py
@@ -21,10 +21,17 @@ from gateway.api.routes._pipeline import _raise_for_unresolvable_model
 from gateway.api.routes.chat import rate_limit_headers
 from gateway.core.config import GatewayConfig
 from gateway.log_config import logger
-from gateway.models.entities import APIKey, UsageLog
+from gateway.models.entities import APIKey, BatchRecord, UsageLog
 from gateway.rate_limit import check_rate_limit
+from gateway.services.batch_service import (
+    claim_batch_accounting,
+    get_batch_record,
+    get_batch_records,
+    record_batch,
+)
 from gateway.services.budget_service import (
     reconcile_reservation,
+    record_external_spend,
     refund_reservation,
     reserve_budget,
 )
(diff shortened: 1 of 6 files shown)


src/gateway/api/routes/batches.py
```diff
@@ -522,18 +571,30 @@ async def retrieve_batch_results(
...
-        cost=cost,
-    )
+    # Recorded batches are accounted exactly once: the first completed retrieval
+    # claims the accounting slot, logs usage, and folds the cost into the owner's
+    # spend; later retrievals (retries, polling clients) return results without
+    # re-billing. Legacy batches with no record keep the prior behavior (log every
+    # retrieval, no spend fold), since they cannot be deduplicated without a record.
+    should_account = True
+    if record is not None:
+        should_account = await claim_batch_accounting(db, batch_id)
```
khaledosman, line 590: Billing is now at-most-once with a crash window: `claim_batch_accounting` commits the `results_accounted_at` transition before `log_batch_usage`/`record_external_spend` run in their own transactions. If the process dies (or `record_external_spend` raises) after the claim commits, the batch is permanently marked accounted and never billed on retry. Prior behavior was at-least-once. Fine to keep, but make the trade-off explicit in the comment (or commit the claim last).

src/gateway/api/routes/batches.py
```diff
@@ -522,18 +571,30 @@ async def retrieve_batch_results(
...
+            provider=provider,
+            endpoint="/v1/batches/results",
+            user_id=user_id,
+            prompt_tokens=prompt_tokens,
+            completion_tokens=completion_tokens,
+            total_tokens=total_tokens,
+            cost=cost,
+        )
+        if record is not None and cost and user_id:
+            await record_external_spend(db, user_id, cost)
```
khaledosman: If this first (winning) retrieval computes `cost=None` (no `ModelPricing` row yet), the slot is claimed and spend is never folded; adding pricing afterward won't help since later retrievals short-circuit on `should_account=False`. Either gate the claim on `cost is not None` for recorded batches, or document that pricing must exist before the first results retrieval.

src/gateway/services/batch_service.py
```diff
@@ -0,0 +1,87 @@
...
+    double-count.
+    """
+    result = await db.execute(
+        update(BatchRecord)
+        .where(BatchRecord.id == batch_id, BatchRecord.results_accounted_at.is_(None))
+        .values(results_accounted_at=datetime.now(UTC))
+        .execution_options(synchronize_session=False)
+    )
+    await db.commit()
+    return bool(getattr(result, "rowcount", 0))
```
khaledosman, line 95: `getattr(result, "rowcount", 0)` is inconsistent with `reserve_budget`/`reconcile_reservation`, which read `result.rowcount` directly. `CursorResult.rowcount` is always present for an UPDATE on aiosqlite/asyncpg, so the guard is unnecessary; prefer `bool(result.rowcount)` to match the neighboring code.

src/gateway/models/entities.py
```diff
@@ -375,6 +375,37 @@ def to_dict(self) -> dict[str, Any]:
...
+
+    __tablename__ = "batches"
+
+    # Provider-assigned batch id (globally unique per provider), used as the
+    # lookup key on retrieve/cancel/results.
+    id: Mapped[str] = mapped_column(primary_key=True)
+    # Instance/provider name the batch was created against (echoed to clients).
+    provider: Mapped[str] = mapped_column()
+    # Billed owner, stamped from the authenticated principal at creation. CASCADE:
+    # deleting the user drops the ownership record (the user's keys are gone too,
```
khaledosman: nit: `provider` is stored but never read; lookups key on `id` alone and `_authorize_batch` doesn't validate it against the requested `?provider=`. Fine if it's purely informational; otherwise consider using it (or dropping it).

khaledosman (approved):

PR #342, opened 2026-07-21
fix(models): bound provider discovery, cache failures, single-flight

njbrake:

## Description

Deleting a provider in the dashboard appeared to hang for ~60s. The DELETE handler does no provider I/O; the stall was a model-discovery fanout hitting an unreachable `moonshot` upstream. Investigation (including reading any-llm source) surfaced four compounding issues in the discovery path:

1. **No timeout.** any-llm builds an `AsyncOpenAI` client with no timeout (`openai/base.py`), so a black-holed endpoint hangs on the SDK default (about 600s with 2 retries). The gateway never bounded it.
2. **Failures were never cached.** A provider whose `list_models` failed was dropped without caching, so it was re-dialed on every discovery, paying the full timeout each time.
3. **No single-flight.** `GET /v1/models` and `GET /v1/models/discoverable` are both mounted on the Models page and share no in-flight coalescing, so each fired its own full fanout (2x per provider on a cold cache or after any provider write).
4. **Lock held across the network.** The cache lock wrapped the entire fanout, serializing every discovery consumer behind one slow provider.

### Changes

All in `src/gateway/services/model_discovery_service.py` plus two config knobs:

- **Per-provider timeout** on every live `list_models` call via `model_discovery_timeout_seconds` (default 10s), plus a fixed 10s bound on the unsaved "test connection" path.
- **Negative caching**: a failed discovery is remembered for `model_discovery_negative_ttl_seconds` (default 30s; 0 restores retry-every-time).
- **Single-flight**: concurrent discoveries of the same provider share one upstream call through an in-flight task map, so the two Models-page listings coalesce.
- **Lock scope**: the cache lock now only guards the small cache and in-flight bookkeeping, never the network.
- `clear()` detaches any in-flight discovery, and `_run` only writes the cache while it is still the registered in-flight, so a credential change mid-discovery cannot repopulate the cache with a stale result. This keeps the post-write "test connection" a live check.
- Unified `GET /v1/models` and the discoverable listing onto one cached, single-flighted per-provider path.

The result: a broken provider goes from "60s tax on every click, doubled" to "one bounded 10s attempt per 30s window, shared across all callers."

Note on the frontend: `invalidateProviderViews` still invalidates both discovery views on a provider write. With single-flight plus positive/negative caching, a forced refetch is now a cache hit for healthy providers and a coalesced, negatively-cached miss for broken ones, so narrowing the invalidation was not needed.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [x] Documentation
- [ ] Infrastructure / CI

## Relevant issues

None filed; found while debugging a provider delete that hung on ai.natebrake.com.

## Checklist

- [x] I ran the Definition of Done checks locally (`make lint`, `make typecheck`, `make test`). Note: `make lint`, `make typecheck`, and the unit suite (740 passed) ran clean locally; the integration suite needs Docker and is left to CI.
- [x] Documentation was updated where necessary (`docs/configuration.md`).
- [x] If the API contract changed, I regenerated the OpenAPI spec. No route or schema changed here, so the spec is unaffected; `make openapi-check` confirms it is up to date.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:** Claude Code (Opus 4.8), directed by @njbrake through a debugging and review session.

**Any additional AI details you'd like to share:** Root cause, fix design, and the code were produced by Claude under @njbrake's direction; the review pass that caught the `clear()` behavior was also run by Claude and verified with isolated reproductions.


src/gateway/services/model_discovery_service.py
```diff
@@ -262,64 +390,19 @@ async def discover_all_models(
...
-                        result,
-                    )
-                    continue
-                _, models = result
-                cache.set(provider_name, models)
-                result_models.extend((provider_name, m) for m in models)
+    # Each instance goes through the shared cache + single-flight path, so a
+    # failing provider is dialed at most once per negative-TTL window and the
+    # concurrent discoverable listing reuses the same in-flight call.
+    results = await asyncio.gather(*(discover_provider_models(config, name) for name in instances))
```
khaledosman: This gather (and the identical one in `discover_models_with_status`) dropped the old `return_exceptions=True` and now relies entirely on `_discover_uncached` never raising. That holds today (it catches `Exception`), but `discover_models_with_status` feeds `list_discoverable_models`, which awaits it with no surrounding try/except, unlike this `/v1/models` path which the route guards. If a future edit ever lets something escape `_discover_uncached`, one provider would 500 the whole operator listing instead of dropping out. Cheap defense-in-depth: pass `return_exceptions=True` (and filter), or guard the discoverable route.

khaledosman (commented):
Reviewed the discovery cache changes. Strong PR, no blocking issues; CI is green. I traced both subtle paths:

- **`clear()` + in-flight race**: `_run` writes the cache only if it's still the registered in-flight, and `clear()` detaches by popping `_inflight`, so a credential change mid-discovery can't repopulate with a stale result and a late caller starts fresh. Matches what `test_stored_provider` relies on, and it's directly tested.
- **Single-flight + lock scope**: the lock guards only the cache/in-flight bookkeeping, never the `await shield(task)`, so one slow provider no longer serializes other consumers; `shield` correctly stops one caller's cancellation from killing the shared discovery.

Two low-severity, non-blocking notes inline.

_Review generated by Claude Code._

src/gateway/services/model_discovery_service.py
```diff
@@ -12,47 +29,147 @@
 from gateway.log_config import logger
 from gateway.services.provider_kwargs import get_provider_kwargs
 
+# Ad-hoc credential tests (unsaved providers) are not config-bound, so they use
+# this fixed bound rather than ``model_discovery_timeout_seconds``. It matches
+# that field's default so behaviour is consistent whether or not creds are saved.
+_ADHOC_DISCOVERY_TIMEOUT_SECONDS = 10.0
```
khaledosman, line 36: This matches `model_discovery_timeout_seconds` only at its default. An operator who raises the config timeout gets a stricter bound on the unsaved "test connection" than on the saved path, so "consistent whether or not creds are saved" is default-only. Fine as a documented trade-off; flagging the divergence.

khaledosman (approved):

PR #343, opened 2026-07-21
feat(dashboard): connection toast, resizable columns, and form polish

### LINKED ISSUE #327: Dashboard: error message should be toast or in header
Image

If the dashboard in the web ui fails to connect, this should show at app level either in bottom right toast or as a part of the header, since it means that the backend isn't responding to the web app

### LINKED ISSUE #328: Dashboard: input boxes are way too wide
Input box dropdowns for things like model selection, owner selection, etc, are way too wide. They take up the whole width, which is really annoying in fullscreen because you have to move your mouse all the way over to the right in order to click the dropown arrow.

Image

### LINKED ISSUE #336: Dashboard: columns should be resizeable
When viewing tables like the dashboard model registry or the activity pane, the column widths for each column should be resizeable. Right now they're static.

njbrake:
## Description

Dashboard polish bundling three tracked issues plus follow-on fixes found while smoke testing. All of these touch the committed web bundle (`src/gateway/static/dashboard`), so they ship as one PR to avoid conflicting rebuilds.

Tracked issues:
- **#327** Lost backend connection now surfaces as an app-level toast (`ConnectionStatus`, bottom-right) that watches the TanStack query cache for network-level failures (`apiFetch` normalizes an unreachable gateway to `ApiError` status 0). It clears itself once a request succeeds again, and excludes 401/403 (which already bounce to sign-in), instead of each page rendering its own inline banner.
- **#328** Form inputs are width-capped (`max-w-md`) so dropdowns no longer stretch the full page: the model/owner/scope comboboxes, the shared `Field`, and the provider `SecretField` / `ProviderComboBox`.
- **#336** Table columns are resizable in the shared `Table` primitive: a drag handle on each header's right edge with an always-visible divider that brightens on hover/focus, pointer drag plus arrow-key support. Tables stay auto-sized until the first resize, then pin to a fixed layout so resizing one column does not reflow the others.

Follow-on polish:
- Primary page actions (Create key / Add provider, etc.) moved from the top-right to their own left-aligned row under each page header, so they stay near the sidebar.
- The add-card `Close` moved to a bottom-left `Cancel`, matching the existing edit forms; the whole dashboard's form footers are now consistent.
- Connection-test error messages wrap instead of truncating, so a long provider error is fully readable.
- `_short_error` strips the redundant `[provider]` tag that `any-llm` prepends to provider errors (only when it names the provider being tested). Every dashboard surface that shows this is already provider-scoped, so the tag was noise.

## PR Type

- [x] New Feature
- [x] Bug Fix
- [x] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes #327, Fixes #328, Fixes #336

## Checklist

Test evidence: `make lint` clean, `make typecheck` clean (234 files), 739 unit tests pass. Web dashboard: `npm --prefix web run typecheck` clean, 108 Vitest tests pass, bundle rebuilt and committed. Integration tests (Testcontainers/Postgres) were not run in this environment; the change is a pure internal string helper plus frontend, with no API contract change, so the OpenAPI spec is unaffected.

## AI Usage

- [ ] No AI was used.

**Any additional AI details you'd like to share:** Implemented by Claude via back-and-forth with @njbrake; the decisions and direction are his, the code and prose are Claude's.


src/gateway/services/model_discovery_service.py
```diff
@@ -145,8 +145,14 @@ class ProviderDiscovery:
...
 
-def _short_error(exc: BaseException) -> str:
+def _short_error(exc: BaseException, provider: str | None = None) -> str:
     message = str(exc).strip() or exc.__class__.__name__
+    # any-llm prefixes a provider error with a "[provider]" tag (e.g.
+    # "[anthropic] No anthropic API key provided…"). Every surface that shows this
+    # is already provider-specific, so the tag is redundant noise; drop it when it
+    # names the provider being tested. Any other bracketed text is left intact.
+    if provider and message.startswith(f"[{provider}]"):
+        message = message[len(provider) + 2 :].lstrip()
```
khaledosman: The empty-message fallback (`or exc.__class__.__name__` on the line above) runs before the tag is stripped, so a message of exactly `"[anthropic]"` strips to `""` and is returned as-is. The dashboard renders `error ?? "Connection failed."`, and `""` isn't null, so the operator sees a blank error rather than the fallback.

Fix: re-apply the fallback after stripping, e.g. `message = message[len(provider) + 2 :].lstrip() or exc.__class__.__name__`.

Low severity (any-llm errors carry body text in practice); fine to skip if you'd rather not touch it.

khaledosman (commented):
Reviewed against the repo's `AGENTS.md`/`CONTRIBUTING.md`. Solid PR: verified `_short_error` passes the resolved impl type (matching any-llm's `[provider]` tag, not the instance alias); traced the Table auto→fixed pin/nudge state batching (lands 144/120 as tested) and confirmed sort lives on an inner button so the resize handle can't trigger it; ConnectionStatus's status-0 normalization + 401/403 exclusion match the tests. Hardcoded `text-red-700`/`bg-red-50` are the repo's existing error-color idiom on a light-only dashboard, so consistent. One minor optional nit inline.

khaledosman (approved):

PR #344, opened 2026-07-21
fix(providers): make API key optional when the provider env var is set

### LINKED ISSUE #335: Add Provider form: don't force an API key when the provider's env var is already set

## Summary

On the dashboard Add Provider page, if a provider's credential env var is already set on the server (for example `OPENAI_API_KEY`), the form should not force me to paste an API key. any-llm already reads the key from the environment when the stored config doesn't override it, so requiring a pasted key is redundant and confusing.

## Current behavior

Adding a known provider (say OpenAI) always requires an API key before the form will submit, even when `OPENAI_API_KEY` is present in the gateway's environment.

The catalog already carries each provider's credential env var name, but `requires_api_key` is computed as "does this provider define an env var at all", not "is that env var actually set right now":

- `src/gateway/services/provider_metadata_service.py:185` sets `requires_api_key=env_key is not None`.
- `KnownProviderSchema` exposes both `env_key` and `requires_api_key` (`src/gateway/api/routes/providers.py:126`), but nothing reports whether `env_key` is populated in the process environment.
- The form gates submission on that flag: `needsKey = selected?.requires_api_key ?? true` and `canSubmit = ... && (!needsKey || apiKey.trim() !== "")` (`web/src/pages/ProvidersPage.tsx:210`, `:216`).

So OpenAI always reads as "needs a key" regardless of the environment.

## Expected behavior

When the selected provider's `env_key` is already set on the server, the API key field should be optional. I can submit the provider with no stored key and let any-llm fall back to the env var. The field copy should also make clear the key will come from the environment (for example: "OPENAI_API_KEY is set on the server, so a key is optional here").

If I do paste a key, it should still win over the env var (unchanged behavior).

## Suggested approach

- Backend: add something like `env_key_present: bool` to `KnownProvider` / `KnownProviderSchema`, computed from `os.getenv(env_key)` (only meaningful when `env_key` is set). Keep `requires_api_key` as-is so the keyless-backend logic (Ollama, llama.cpp) is untouched.
- Frontend: relax `needsKey` to `requires_api_key && !env_key_present`, and adjust the field label/description to explain the env fallback.

## Notes

- Keyless local backends (Ollama, llama.cpp) already submit without a key via `requires_api_key=false`; this change is about the separate case of a key-based provider whose env var is already populated.
- Worth confirming the copy distinguishes the two "optional key" reasons: keyless backend vs env var already set.

njbrake:
## Description

On the dashboard Add Provider page, a key-based provider (e.g. OpenAI) always forced a pasted API key before the form would submit, even when the provider's credential env var (`OPENAI_API_KEY`) was already set on the server. any-llm reads the key from the environment when the stored config doesn't override it, so requiring a pasted key there is redundant and confusing.

The catalog only reported whether a provider defines an env var, not whether that var is populated right now. This change adds `env_key_present` and uses it to make the key optional when the env var is set.

- Backend (`provider_metadata_service.py`): add `env_key_present: bool` to `KnownProvider`, computed from `os.getenv(env_key)` (whitespace-only counts as absent). `requires_api_key` is left unchanged, so the keyless-backend path (Ollama, llama.cpp) is untouched. Uses `os.getenv` directly rather than `otari_env()`, which only reads the `OTARI_` prefix; these are provider-native vars, and this is a network-free catalog listing on form load, not a hot path.
- Backend (`providers.py`): expose `env_key_present` on `KnownProviderSchema`.
- Frontend (`ProvidersPage.tsx`): relax `needsKey` to `requires_api_key && !env_key_present`. When the env var is set, the field reads "API key (optional)" with copy explaining the env fallback ("OPENAI_API_KEY is set on the server, so a key is optional here. Paste one to override it."). A pasted key still wins over the env var. The two "optional key" reasons (keyless backend vs env var already set) get distinct copy.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes #335

## Checklist

## AI Usage

- [ ] No AI was used.

Implementation and tests were drafted by Claude via back-and-forth with @njbrake; the reasoning and decisions are his.

**NOTE:**

## Summary

- Made API keys optional in the Add Provider form when the provider’s server-side environment credential is configured.
- Added provider catalog metadata to indicate whether an environment credential is available.
- Updated labels and guidance to clarify required, optional, and unnecessary keys while preserving key override behavior.
- Added backend, frontend, integration, and unit test coverage for configured, missing, blank, and keyless credentials.
- Refreshed the dashboard asset bundle.

## Benefits

- Providers can be added without duplicating credentials already configured on the server.
- Keyless local backends retain their existing behavior.


khaledosman (approved):
Reviewed the credential-resolution path end to end: a provider stored with `api_key=null` produces an overlay entry with no `api_key`, so any-llm falls back to the provider-native env var at dispatch. The "optional key when env var is set" behavior is real, `requires_api_key` is untouched (keyless backends unaffected), `env_key_present` exposes only a bool (never the secret), and the catalog route stays master-key gated and standalone-only. Whitespace-only env values are correctly treated as absent. Tests cover env-set / unset / blank / keyless. Clean, well-scoped fix.

PR #345, opened 2026-07-21
feat(dashboard): usage & analytics page with aggregation endpoints (#303)

### LINKED ISSUE #303: Dashboard: usage & analytics page

Aggregate usage views over `UsageLog`, backed by the existing `usage` router.

**Why:** Operators need to see cost and volume by model and by user over time.

## Tasks
- [ ] Spend and token/request counts by model, by user, and over a chosen time range.
- [ ] Filters (date range, model, user) and CSV export.
- [ ] Hooks plus a `UsagePage`.
- [ ] Tests; rebuild bundle.

Pairs with the per-request log viewer (same data, different granularity).

Part of #299.

njbrake:
## What

Adds the Usage & analytics page (issue #303) and the aggregation API behind it, on the observability track of #299. Operators can now see aggregate spend, tokens, and request volume by model, by user, and over time, drill from any breakdown row into the Activity log, and export to CSV.

## Backend (`src/gateway/api/routes/usage.py`)

- `GET /v1/usage/summary` (master-key gated, JSON): totals, `by_model` / `by_user` / `by_api_key`, and a UTC-bucketed time series, all reusing the existing `_usage_filters`.
- `GET /v1/usage/summary.csv`: uncapped CSV export (finance wants every row), with a spreadsheet formula-injection guard, kept on its own route so `/summary` keeps one clean JSON response model.
- Range-bounded: default last 30 days, hard span cap, so an aggregate never scans the whole append-only log; it stays served by the `timestamp` index.
- Portable hour/day bucketing across SQLite (`strftime`) and PostgreSQL (`date_trunc`), pinned to UTC with `AT TIME ZONE 'UTC'` so buckets do not shift with the server timezone or break across DST.
- Coalesced sums, a conditional error count, and top-N breakdowns with a synthesized "other" fold that reconciles with the totals. No new index (a `model` index would not help the group-by; range-bounding does the work).

## Dashboard (`web/`)

- `UsagePage`: stat tiles with period-over-period deltas, spend-by-model and spend-by-user tables whose rows deep-link into the Activity log pre-filtered on that dimension and window, a single-metric SVG bar chart with a Cost/Tokens/Requests toggle (hand-rolled, no new chart dependency), and CSV export.
- `FilterComboBox`: a type-to-filter combobox replacing the native `<select>` (unusable at thousands of users) and the exact-match model text box. The Usage model picker is sourced from the window's own `by_model`; the Activity model picker gains the same typeahead with suggestions sourced from usage so they match what the log stores, while still accepting any typed model.
- Nav entry and route under Observability; `ActivityPage` seeds its filters from the drill-down query string.

## Testing

- `tests/integration/test_usage_summary.py`: aggregation correctness, cross-dialect UTC bucketing, top-N reconciliation, empty/null/`ZeroDivision` guards, and CSV shape plus injection. Runs on both SQLite and PostgreSQL in CI (needs Docker; verified locally on SQLite end to end).
- Vitest: 117 web tests pass, including the Usage page, the drill-down, and both typeahead pickers.
- Backend: 740 unit tests pass, ruff and mypy clean, OpenAPI regenerated and up to date, dashboard bundle rebuilt.

## Notes

- Standalone-only, hybrid has no local management API (per #299).
- Deferred with explicit triggers, not in this PR: an hourly rollup table for when a single query window holds ~1M+ rows, latency percentiles, and a Prometheus/BI export.
- `scripts/seed_usage_smoke.py` is a local smoke-test helper; happy to drop it if you would rather not ship it.

Closes #303. Part of #299.

## Summary

- Added a Usage & Analytics dashboard under Observability with spend, request, token, latency, time-series, and breakdown views.
- Added date, model, and user filters, period comparisons, typeahead selection, CSV export, and drill-down links to Activity logs.
- Added authenticated JSON and CSV aggregation endpoints with totals, top-N groupings, UTC time buckets, bounded ranges, and formula-injection protection.
- Added integration and UI tests, regenerated API definitions, rebuilt dashboard assets, and added usage-data smoke-test seeding.


src/gateway/api/routes/usage.py
```diff
@@ -174,3 +188,320 @@ async def count_usage(
...
+
+
+def _resolve_window(start_date: datetime | None, end_date: datetime | None) -> tuple[datetime, datetime]:
+    """Clamp the requested window to a bounded, forward-ordered range.
+
+    A summary must never scan an unbounded log: absent a start we look back
+    ``_DEFAULT_SUMMARY_LOOKBACK``; a span wider than ``_MAX_SUMMARY_SPAN`` has its
+    start pulled forward so the aggregates stay bounded by the timestamp index.
+    """
+    end = end_date or datetime.now(UTC)
```
khaledosman, line 267: **Medium:** `datetime.now(UTC)` is tz-aware, but a query `start_date` without an offset (e.g. `?start_date=2025-01-01T00:00:00`, a value the param description advertises as valid ISO-8601) parses to a naive datetime. With `end_date` omitted, `start > end` on the next line raises `TypeError: can't compare offset-naive and offset-aware datetimes` -> unhandled 500. The epoch form parses to aware UTC and works, so the crash is input-shape-dependent. Normalize both bounds to aware UTC at the top of `_resolve_window` (e.g. `dt.replace(tzinfo=UTC) if dt.tzinfo is None else dt`) and add a naive-input test.

web/src/pages/ActivityPage.tsx
```diff
@@ -202,6 +208,26 @@ export function ActivityPage() {
...
+  const modelSuggestFilters: UsageFilters = useMemo(
+    () => ({
+      start_date: startDate,
+      status: statusFilter || undefined,
+      endpoint: endpointFilter || undefined,
+      user_id: userFilter || undefined,
+    }),
+    [startDate, statusFilter, endpointFilter, userFilter],
+  );
+  const modelSummary = useUsageSummary(modelSuggestFilters, "day");
```
khaledosman, line 223: **Low (perf):** this calls `useUsageSummary` only to populate the model typeahead, but `usage_summary` runs five aggregate GROUP BY scans (totals + 3 breakdowns + series) and refires on every `start_date`/`status`/`endpoint`/`user_id` change. Only `by_model` keys are consumed, and `by_model` is the intentionally-unindexed dimension. Consider a lighter endpoint/param that returns just distinct in-window model names, or document the cost as a deliberate trade-off.

web/src/pages/UsagePage.tsx
```diff
@@ -0,0 +1,546 @@
...
+        preserveAspectRatio="none"
+        role="img"
+        className="w-full"
+        aria-label={`${metric} over time`}
+      >
+        <title>{`${metric} per ${bucket}`}</title>
+        {series.map((point, i) => {
+          const value = metricValue(point, metric);
+          const h = (value / max) * (height - pad);
+          const x = pad + i * slot + (slot - barW) / 2;
```
khaledosman, line 243: **Low:** bars are positioned by array index (`pad + i * slot`), but the backend series omits empty buckets (`GROUP BY`). A sparse range (e.g. usage on day 1 and day 20 of a 30-day window) renders as two adjacent bars, so the x-axis is not linear in time and the trend reads misleadingly. Zero-fill the series server-side (bounded by the span cap) or lay bars out by timestamp rather than index.

khaledosman (approved):
Solid work. Both endpoints are master-key gated and standalone-only, SQL is parameterized (the one dynamic token is a validated `Literal`), the window is hard-bounded, the top-N "other" fold reconciles, the CSV formula-injection guard is in place, and cross-dialect UTC bucketing is correct. Approving to unblock; one Medium worth fixing before/right after merge, plus two Low notes inline.

**Medium:** `_resolve_window` can 500 on a naive `start_date` (see inline). The dashboard always sends `...Z` so it is unaffected, but a direct API client passing a valid offset-less ISO datetime hits it.

PR #346, opened 2026-07-21
feat(dashboard): expand Settings with a full config viewer, wider settable set, and search

### LINKED ISSUE #307: Dashboard: expand Settings (full config viewer + wider runtime-settable set)

Today only `model_discovery` and `default_pricing` are runtime-settable; the config surface is roughly 50 fields.

**Why:** Operators expect to see, and where safe change, more of what `config.yml` controls.

## Tasks
- [ ] Read-only "full config" viewer for every effective setting, clearly marking startup-only fields (`host`, `port`, `database_url`, `mode`, `db_pool_*`, `cors_allow_origins`, `budget_strategy`, etc.).
- [ ] Backend: widen `SETTABLE_KEYS` and `apply_override` for fields safe to hot-change. Candidates: `require_pricing`, `rate_limit_rpm`, `model_cache_ttl_seconds`, `reject_user_mismatch`, `stream_missing_usage_policy`.
- [ ] Group advanced sections (files, vision / file-understanding, MCP flags) so the page stays legible.
- [ ] Tests; rebuild bundle.

**Note:** most fields are startup-only and must be display-only. The UI must not imply they hot-apply.

Part of #299.

njbrake:
## Description

Expands the dashboard Settings page from two toggles into a full, grouped view of every effective gateway setting, and widens the runtime-settable set to every field that can be safely hot-changed.

**Backend**
- Reworks `runtime_settings_service` around a per-key spec registry (bool / int / float / enum-string / nullable-string), replacing the previous bool-only path. Each key carries the same constraints its `GatewayConfig` field declares, so a dashboard write cannot push the gateway into a state a fresh config load would reject.
- Widens the settable set from 2 to 15 fields. Every added key is read fresh from config on the request path (or the next model discovery / catalog lookup), so mutating config hot-applies in this worker: `require_pricing`, `reject_user_mismatch`, `model_cache_ttl_seconds`, `stream_missing_usage_policy`, `model_discovery_timeout_seconds`, `model_discovery_negative_ttl_seconds`, `models_dev_metadata`, `models_dev_cache_ttl_seconds`, `file_understanding_enabled`, `vision_strategy`, `vision_describe_model`, `vision_describe_max_tokens`, `budget_estimate_default_output_tokens`.
- Deliberately kept read-only (documented in the service): startup-only fields (`host`, `port`, `database_url`, `mode`, `db_pool_*`, `rate_limit_rpm`, `cors_allow_origins`, `budget_strategy`, docs/metrics, log writer), and the outbound network-safety gates (`mcp_allow_*`, `web_search_allow_private_hosts`, the `*_url` fields), since opening an SSRF gate should not sit behind a dashboard toggle. `rate_limit_rpm` in particular is consumed once at startup to build `app.state.rate_limiter`, so mutating config alone would not change running behavior.
- `GET /v1/settings` now returns a grouped, read-only "full config" view of every non-secret effective setting, each marked settable or startup-only, with the field's description and (for enums) its allowed options sourced from the config schema. Secrets (`master_key`, provider credentials) and the complex catalog fields managed on their own pages (providers, pricing, aliases, model_capabilities, platform) are omitted.
- `PATCH` uses `model_fields_set`, so an explicit `null` clears a nullable field (`vision_describe_model`) while an omitted field is left unchanged; invalid keys or values are rejected before any write.
- Factors the stream-policy and vision-strategy allowed sets into shared constants used by both the field validators and the runtime-settings layer.

**Dashboard**
- Rebuilds the Settings page as grouped sections rendered from the config view. Settable fields get an interactive control (toggle, number plus Save for int/float, select for enums, text plus Save for nullable strings); startup-only fields render read-only with a "startup-only" marker, so the UI never implies they hot-apply.
- Adds a fuzzy search box (matches key, description, and group, plus a subsequence match on the key so "mctts" finds `model_cache_ttl_seconds`; "/" focuses it, Esc clears), a "settable only" filter, a visible-count line, per-group counts, and an empty state, so the ~47-field page stays navigable.

**Tests**
- Backend: validation for every type and bound; `apply_override`; the GET full-view shape; PATCH apply for bool/int/float/enum plus nullable clear-to-null; ignore of startup-only fields; 422 on bad input; and a sqlite restart test that verifies the serialize to DB then parse then apply roundtrip for the new (non-bool) types, plus `load_overrides` skipping stale/invalid rows. The Postgres integration test that already covered "override survives a restart" for the bool flags continues to run in CI; the new sqlite test mirrors that promise locally and extends it to the new types.
- Dashboard: the fuzzy `fieldMatches` helper, each control type (including float decimals and clear-to-null), the search and settable-only filters, and the empty state.

Regenerated OpenAPI spec and dashboard bundle are included.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes #307 (part of #299).

## Checklist

## AI Usage

- [ ] No AI was used.

Written by Claude Code via back-and-forth with @njbrake; the design decisions (which fields are safe to hot-change, keeping SSRF gates and startup-wired fields read-only) are his, driven by review of each field's runtime read site.

**NOTE:**

## Summary

- Expanded the Settings page into a grouped viewer for effective gateway configuration.
- Added search, settable-only filtering, group and visibility counts, and controls for toggles, numbers, choices, and text values.
- Added clear indicators for startup-only settings while keeping sensitive, catalog-managed, and network-safety fields read-only or hidden.
- Broadened runtime setting support to safely changeable booleans, numbers, choices, and nullable text values.
- Added shared validation, persistence, restart recovery, and URL credential redaction for settings updates.
- Regenerated dashboard assets, API documentation artifacts, and expanded backend and frontend test coverage.


src/gateway/api/routes/settings.py
```diff
@@ -1,47 +1,243 @@
...
+_REDACTED_URL_FIELDS = frozenset({"database_url", "sandbox_url", "guardrails_url"})
+
+
+def _redact_url_password(value: str) -> str:
+    """Mask the password in a URL's userinfo, leaving the rest intact."""
+    try:
+        parts = urlsplit(value)
+    except ValueError:
+        return value
+    if parts.password is None:
```
khaledosman: **Low:** this masks only `parts.password`. A credential in the query string (`https://sandbox.internal/?api-key=secret`) or in the username position (`https://TOKEN@host`) is returned unredacted, which contradicts the "a secret is never echoed into the response or the browser's query cache" goal for `sandbox_url`/`guardrails_url`. `database_url`'s `user:pass@` form is fine. Consider masking the full userinfo and stripping/masking known credential query params for `_REDACTED_URL_FIELDS`.

web/src/pages/SettingsPage.tsx
```diff
@@ -39,81 +66,298 @@ function Toggle({
...
+  const [draft, setDraft] = useState(String(committed));
+  const isFloat = field.type === "float";
+
+  useEffect(() => {
+    setDraft(String(committed));
+  }, [committed]);
+
+  const parsed = Number(draft);
+  const wellFormed = draft.trim() !== "" && Number.isFinite(parsed) && (isFloat || Number.isInteger(parsed));
+  const valid = wellFormed && parsed >= 0;
```
khaledosman: **Low (UX):** the Save gate is hardcoded `parsed >= 0`, but `vision_describe_max_tokens` and `model_discovery_timeout_seconds` are declared `gt=0`, so `0` enables Save then gets rejected with a 422 error banner instead of a disabled button. Thread the spec's `gt`/`ge` through `ConfigField` so the min/allowed-zero is per-field.

khaledosman (approved):
Reviewed the read-only/settable partition closely and it holds: `SETTABLE_KEYS` derives from `_SPECS`, `UpdateSettingsRequest` declares exactly those 15 fields, and the write path is gated at three layers (schema ignores extras, `key in SETTABLE_KEYS`, `validate_value`), so there is no path to PATCH a startup-only or secret key. Every SSRF gate and secret is display-only. Null-clear via `model_fields_set`, commit-then-apply ordering, and fail-safe `load_overrides` are all correct, and every settable field is read fresh at runtime (verified the discovery-TTL and models.dev cases). Approving. Two Low notes inline and one design question below.

**Design question (non-blocking):** `require_pricing` and `reject_user_mismatch` are now runtime-mutable. Both keep fail-closed defaults and are master-key gated, so this is not an exploitable vuln, but by the docstring's own SSRF-gate rationale ("a security decision should not sit behind a dashboard toggle") these are the §0.4 metering and §0.1 tenant-isolation controls. Worth a docstring note on why they differ from the gates, or a UI confirm step for `require_pricing`.

Issue #302, opened 2026-07-17
Dashboard: overview / home page


njbrake:
> _Drafted by Claude via back-and-forth with @njbrake._

An at-a-glance landing page shown right after sign-in.

**Why:** Operators want a single first screen (spend, traffic, errors, health) before drilling into a specific page.

## Tasks
- [ ] KPIs: spend today and this period, request volume, error rate, active keys/users.
- [ ] Provider health summary (reuses the provider connectivity work).
- [ ] Recent-activity preview linking to the request log.
- [ ] Make this the authenticated index route (currently Providers/Models).

**Depends on:** usage aggregates and budgets for the numbers, and the provider health monitor for the health tile.

Part of #299.

PR #347, opened 2026-07-21
feat(dashboard): provider health monitor on the Providers page

### LINKED ISSUE #309: Dashboard: provider health monitor

A dashboard-wide provider connectivity view.

**Why:** A per-provider "test" exists, but there is no at-a-glance health across all providers.

## Tasks
- [ ] Show each provider's reachability using the existing test path, with a last-checked time.
- [ ] Optional periodic/background refresh vs. purely on-demand.
- [ ] Feed the overview page summary tile.
- [ ] Tests; rebuild bundle.

Part of #299.

njbrake:
## Description

Adds a dashboard-wide provider health view. Today an operator can only check
connectivity one provider at a time (the per-row "Test connection" button), so a
provider with a dead or rotated key is easy to miss. This surfaces every
configured provider's reachability at a glance, on the existing Providers page
rather than a separate page, so the status sits right where keys are managed.

Backend: a new `GET /v1/providers/health` (standalone-only, master-key gated)
reports each provider's ok/error, model count, and a last-checked time. It
reuses the existing per-provider model-discovery test path
(`discover_provider_models`) rather than adding a second probe mechanism, so a
provider is "healthy" exactly when its credentials can list models. Results are
served from the discovery cache (cheap enough to poll); `?refresh=true` clears
the cache to force a live re-dial. The model cache now records a wall-clock
timestamp, so "last checked" honestly reflects when a provider was actually
dialed, not when the dashboard last asked. Per-provider `healthy` / `total`
counts are precomputed on the response so the planned overview summary tile
(#302) can reuse them without re-deriving.

Frontend: a Status column on the Providers table (a Reachable / Unreachable
pill, with the error and last-checked time on hover) plus a top summary banner
("N of M providers reachable" with a live "Re-check all"). Config-only
providers, which have no per-row Test button, now show a status too. The
`useProviderHealth` / `useRecheckProviderHealth` hooks are the reusable data
layer for #302.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes #309

## Checklist

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:** Claude Code (Opus 4.8), driven by @njbrake.

The reasoning and design decisions (fold health into the Providers page rather
than a standalone page; reuse the existing discovery path; keep the data
reusable for #302) are @njbrake's; the prose and code are Claude's. Local
checks run: `make lint`, `make typecheck`, the full unit suite, and the touched
integration tests against a local Postgres (the full testcontainers integration
run needs Docker, which was unavailable in this environment).

## Summary

Added a dashboard-wide provider health monitor to the Providers page. It shows reachability, model counts, errors, and last-checked times for each provider, with aggregate reachable counts and a “Re-check all” action.

## Benefits

- Makes provider connectivity visible in one place.
- Supports cached checks with optional, debounced refreshes.
- Adds reusable health hooks for future dashboard views.
- Includes API, UI, integration, and unit test coverage.
- Regenerates dashboard assets and API documentation.


src/gateway/services/provider_health_service.py
```diff
@@ -0,0 +1,95 @@
...
+    refresh: bool = False,
+) -> ProviderHealth:
+    """Report one instance's reachability via the shared model-discovery path.
+
+    ``refresh`` clears the cached listing for this instance first, forcing a live
+    re-dial (the same cache clear the stored-provider test uses), so an operator's
+    explicit "re-check now" is a fresh probe rather than a cached verdict.
+    """
+    if refresh:
+        get_model_cache().clear(instance)
```
khaledosman, line 78: **Low:** `?refresh=true` calls `clear(instance)`, which pops the in-flight task too, so N concurrent refreshes each start an independent dial per provider instead of sharing one, defeating the discovery subsystem's single-flight coalescing. It is master-key gated and each dial is timeout-bounded, so this is an admin-authenticated cost lever, not a DoS. If you want to harden: debounce refresh (ignore if dialed within the last few seconds) or let refresh ride single-flight rather than detaching the in-flight task. At minimum a one-line comment that refresh intentionally bypasses coalescing.

khaledosman (approved):
Nicely done. Auth + mode gating are structural (master-key dep + `providers` router not mounted in hybrid), so no unauthenticated/non-master caller can trigger provider dials. The `checked_at` (wall-clock, for display) vs `cached_at` (monotonic, for TTL) split is a genuinely careful touch, so "last checked" honestly reflects when a provider was last dialed. Fan-out uses `gather(return_exceptions=True)` and re-raises real cancellation, config-only providers get a status too, and the error string reuses the existing capped/sanitized `_short_error` (no new leakage surface). Tests cover happy + error + gating. Approving. One Low note inline.

PR #348, opened 2026-07-21
feat: add ability to edit alias target in dashboard

### LINKED ISSUE #341: Dashboard: add ability to shift alias to a different model
There should be an 'edit' ability to make an alias point to a different model than that which was initially configured.

njbrake:
## Description

The dashboard could create and delete stored aliases but had no way to change an alias's target model after creation. This adds an **Edit** button on stored alias rows that opens a form pre-filled with the current target, allowing the operator to shift the alias to a different model.

The backend `POST /v1/aliases` already performs an upsert (update if the name exists, create otherwise), so the frontend reuses `useCreateAlias`. The name is displayed read-only in the edit form since it is the lookup key; renaming requires delete + recreate.

## PR Type

- [x] New Feature

## Relevant issues

Fixes #341

## Checklist

- [x] I ran the Definition of DoD checks locally (`npm --prefix web run typecheck`, `npm --prefix web test`).

## AI Usage

**AI Model/Tool used:** Laguna S 2.1 from Poolside AI

## Summary

- Added dashboard support for editing an existing alias’s target model.
- Added an edit form with the alias name locked and the target pre-filled.
- Ensured create and edit forms cannot be open at the same time.
- Added regression tests for editing and form visibility.
- Rebuilt the dashboard bundle to include the updated changes.


eiso:
First open source PR from Laguna S2.1 ❤️ 😍 

Thank you @njbrake

web/src/pages/AliasesPage.tsx
```diff
@@ -121,6 +173,7 @@ export function AliasesPage() {
                 <Td className="text-right whitespace-nowrap">
                   {alias.source === "stored" ? (
                     <span className="inline-flex items-center gap-2">
+                      <Button size="sm" variant="ghost" onPress={() => setEditing(alias)}>Edit</Button>
```
khaledosman: Nit (UX): while the New-alias form is open (`adding=true`), these per-row Edit buttons stay active, so clicking one renders `NewAliasForm` and `EditAliasForm` at the same time. The header action is guarded (`adding || editing`) but the row buttons aren't. Consider closing `adding` when `setEditing` fires (or disabling row Edit while adding).

_Claude Code_

khaledosman (approved):
Clean, small, well-tested. Reusing `useCreateAlias` for the upsert and showing the name read-only are the right calls. Approving, one UX nit inline.

njbrake:
@eiso Thank you for releasing a wonderful model! I am seriously blown away with the quality of its reasoning and content, as well as the speed you can get because of the small-ish size. Keep em coming!

PR #350, opened 2026-07-21
feat(dashboard): built-in tools & guardrails configuration (#306)

### LINKED ISSUE #306: Dashboard: built-in tools & guardrails configuration

A page to configure the built-in tools and guardrails: web search, code execution/sandbox, and request guardrails.

**Why:** These are runtime-configurable services like providers, and the search/sandbox backends take API keys, so they fit the encrypt-at-rest, write-only pattern built for provider credentials.

## Tasks
- [ ] Web search: engine selection, result limits, extract toggle, and backend API keys (e.g. Tavily/Exa) stored via `secret_box`.
- [ ] Code execution/sandbox: `sandbox_url` and purpose hint.
- [ ] Guardrails: `guardrails_url` and a reachability test.
- [ ] Reuse the encrypt-at-rest plus last4 write-only pattern for any secret.
- [ ] Backend: decide which of these become runtime-settable vs. startup-only.
- [ ] Hooks, a page, tests; rebuild bundle.

Part of #299.

njbrake:

## Description

Adds a standalone-only **Tools & Guardrails** dashboard page plus a `/v1/tool-settings` management API to make the built-in service endpoints editable at runtime: the web-search, sandbox, and guardrails URLs, plus the web-search knobs (engines, max results, extract, purpose hints).

These are exactly the `*_url` and SSRF-adjacent fields that #307 deliberately keeps display-only on the Settings page ("opening an SSRF gate should not sit behind a dashboard toggle"). So #306 is the complement to that page, not a duplicate: it makes those fields editable with the safety that justifies it, a dedicated page, per-service reachability tests, and structural URL validation.

**Stacked on #307.** This branch is based on `feat/settings-expansion` and reuses its runtime-settings machinery. Please review/merge #346 first; retarget this to `main` once it lands.

### Backend
- `tool_settings_service`: a separate `_TOOL_SPECS` registry, kept out of the generic `/v1/settings` PATCH so these fields can never be flipped through it. Every field is nullable; clearing one falls back to the configured env/YAML value (not to "nothing").
- URL validation is **structural** (http/https scheme plus a host), not a deny-private SSRF gate. The operator is master-key trusted and can already point the runtime anywhere via config/env, and a deny-private gate would reject the bundled docker-compose sidecars (`http://searxng:8080`, `http://guardrails:8000`) that this page exists to configure. The result-fetch SSRF gates (`web_search_allow_private_hosts`, `mcp_allow_*`) stay display-only and are never widened from the dashboard.
- `tool_settings` route: `GET` (effective values, URL passwords masked via the existing redaction), `PATCH` (omit-vs-null semantics), `POST /{service}/test` (structural check, then an HTTP reachability probe of the typed, possibly unsaved URL). Master-key gated, standalone-only, mounted after the hybrid guard.
- Read-path normalization: the request-path read sites now read `config.<attr>` with an env fallback, so a dashboard override hot-applies on the next request. Adds `web_search_url` as a config field plus `ENV_BRIDGED_FIELDS`, carries `config` on `ToolContext`, and applies persisted tool overrides at startup. Pure-env deployments are byte-for-byte unchanged.

### Dashboard
- `ToolsGuardrailsPage`: vertical grouped sections (reusing the Settings layout, not a card mosaic), URL rows with a shared draft that Test probes and Save commits, inline per-field errors, an `aria-live` reachability result, a tri-state select for the nullable extract flag, and a save-confirmation toast. Nav entry and route added. Also adds an `aria-live` wrapper to the Providers connection test (a small a11y fix in the same area).

### Tests
- Backend: the tool-settings service (structural URL validation accepts private/loopback, rejects non-http(s) and bad bounds), the endpoint (bundled-sidecar URLs are accepted, clear-and-restart falls back to the configured value, standalone-only gating returns 404 in hybrid, `/test` reachable and unreachable), and the read-site config resolution (a dashboard override reaches `_build_web_search_backend` / guardrails, not just `config`).
- Dashboard: renders the three services, saves and clears a URL, tests reachability, shows an inline save error, and sets the tri-state extract flag.
- Regenerated OpenAPI spec and dashboard bundle.

**Note:** the Postgres-backed integration tests could not run in the authoring sandbox (no Docker/testcontainers). The sqlite `TestClient` endpoint tests and read-site unit tests cover the same paths; CI should exercise the full `tests/integration` dispatch suite.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Closes #306 (part of #299). Stacked on #307 (#346).

## Checklist

- [x] I have added or updated tests that cover my change (`tests/unit`, plus dashboard vitest).
- [x] I ran the Definition of Done checks locally (`make lint`, `make typecheck`, unit tests, `web` typecheck + tests).

## AI Usage

- [ ] No AI was used.

**Any additional AI details you'd like to share:** Planned with the /autoplan review pipeline (CEO/design/eng passes) and implemented via back-and-forth with @njbrake; the design decisions (structural-not-deny-private URL validation given a trusted operator, deferring an encrypted web-search key until a consumer exists, folding into the #307 machinery rather than a new store) are his.

## Summary

- Added a standalone-only **Tools & Guardrails** dashboard for configuring web search, sandbox, and guardrail services.
- Added master-key-protected APIs to view, update, validate, persist, and test service settings.
- Applied saved settings at startup and runtime, with environment/YAML fallback support.
- Added URL reachability testing, inline validation errors, success notifications, and accessibility improvements.
- Updated routing, API types/hooks, OpenAPI/Postman documentation, and rebuilt dashboard assets.
- Added comprehensive backend and frontend test coverage.


src/gateway/api/routes/_pipeline.py
```diff
@@ -1624,6 +1632,7 @@ async def run_streaming_with_fallback(
                     base_url=tool_ctx.web_search_url,
                     tool_entry=tool_ctx.web_search_tool_entry,
                     auth_token=tool_ctx.web_search_auth_token,
+        config=tool_ctx.config,
```
khaledosman: Nit (style): this `config=tool_ctx.config,` is indented to 8 spaces while its sibling kwargs (`base_url`/`tool_entry`/`auth_token`) are at ~20; the two other identical call sites (`dispatch_non_stream`, `open_stream`) are aligned. `make lint` runs only `ruff check`, so it won't flag this, but a `ruff format` pass would. Align it.

_Claude Code_

web/src/pages/ToolsGuardrailsPage.tsx
```diff
@@ -0,0 +1,381 @@
...
+  disabled: boolean;
+}) {
+  if (field.type === "url") {
+    return <UrlRow field={field} onSave={onSave} saveError={saveError} disabled={disabled} />;
+  }
+  if (field.type === "int") {
+    return <NumberRow field={field} onSave={onSave} saveError={saveError} disabled={disabled} />;
+  }
+  if (field.type === "bool") {
+    return <BoolRow field={field} onSave={onSave} disabled={disabled} />;
```
khaledosman: `BoolRow` never receives or renders `saveError` (unlike `UrlRow`/`TextRow`/`NumberRow`), and there's no error toast on failure, so a failed `web_search_extract` save is silent. Pass `saveError={saveError}` here and render it in `BoolRow` like the other rows.

Separately: `BoolRow` auto-saves on select change while the other rows require an explicit Save click, worth confirming that inconsistency is intentional.

_Claude Code_

khaledosman (approved):
Reviewed against the backend/frontend/security standards. Solid, thorough, well-tested. Verified the security-sensitive parts and they hold:

- Standalone-only: `register_routers` early-returns before `tool_settings.router` in hybrid mode, and the hybrid-404 test confirms it.
- Master-key gated on all three routes; URL-password redaction on GET; commit-before-apply ordering; `follow_redirects=False` on the probe; `is None` checks (no falsy-zero trap, bool `False` preserved).
- Structural-not-deny-private URL validation is correctly reasoned given the trusted operator + bundled sidecars, with the real SSRF gates staying display-only.

Approving. Two low-severity inline nits below.

Informational (no change requested): hot-apply mutates only the in-process `config`, so in a multi-worker/replica deployment other workers keep the old value until restart (they re-read via `apply_tool_overrides_from_db`). The page copy says "Changes apply immediately." This matches the existing runtime-settings machinery you reused, so it's a pre-existing characteristic, not a regression, worth at most a tooltip caveat.

web/src/pages/ToolsGuardrailsPage.tsx
```diff
@@ -0,0 +1,381 @@
...
+      <svg viewBox="0 0 24 24" fill="none" stroke="currentColor" strokeWidth="2" aria-hidden className="h-5 w-5">
+        <path d="M20 6 9 17l-5-5" strokeLinecap="round" strokeLinejoin="round" />
+      </svg>
+      {message}
+    </div>
+  );
+}
+
+const INPUT_CLASS =
+  "rounded-md border border-[var(--otari-line)] bg-white px-2 py-1 text-sm focus:border-[var(--otari-brand)] focus:outline-none disabled:opacity-50";
```
khaledosman: `bg-white` should be `bg-[var(--otari-surface)]` per design-tokens.md (no raw palette classes for chrome). No visible change today (the surface token is white and the dashboard is light-only), but the rule exists so a future theme/redesign is a one-place variable change. The `SaveToast` green classes are fine, status surfaces are the documented exception.

Note: several existing inputs (`ModelsPage.tsx:139`/`:610`, `ModelScopeControl`, `SettingsPage`) do the same, so this is a pre-existing repo pattern, flagging it on the new code; a repo-wide cleanup could be a separate pass.

_Claude Code_

PR #354, opened 2026-07-22
feat(dashboard): overview / home page (#302)

njbrake:
## Description

Adds an at-a-glance Overview as the authenticated index route. It answers "is everything OK, and what needs my attention" before an operator drills into a specific page. Part of #299.

- **System-status strip.** A calm one-liner when all clear ("All systems normal, N/N providers healthy, budgets within limit"), otherwise only the problems (unreachable providers, budgets over or near limit, high error rate), each linking to the page that fixes it.
- **KPI tiles.** Spend today, spend and request volume over the last 30 days with period-over-period deltas, error rate, budget health, providers healthy, active keys, active users. Status tiles pair color with a word (Healthy, Elevated, Over budget, Degraded), never hue alone, so they read for colorblind operators.
- **Recent-activity preview** linking to the full request log.
- **Sidebar** gets a header-less "Overview" lead item; its NavLink uses `end` so it does not stay highlighted on every route.

Design notes:

- Pure frontend. Composes existing hooks (usage summary, provider health, usage logs, budgets, keys, users). No backend change, no new endpoint, no migration.
- First-run gating stays outside the page (`OverviewIndex`), so a fresh gateway with no provider configured redirects to Providers without ever firing the overview's query fan-out.
- Budget health reuses the per-user cap math (`max_budget * user_count`), excluding unlimited and user-less budgets, matching the Budgets page.
- Date windows are frozen once (stable query keys, no refetch loop); "Today" is local midnight so the operator's wall-clock day is what the server aggregates.
- Shared formatters (`formatUsd`, `formatTokens`, `formatPct`, `deltaFraction`) and `DeltaHint` moved into `lib/format.ts` and `components/ui.tsx`; `UsagePage` now consumes them instead of local copies.

Trend sparklines on the tiles and adopting a charting library are out of scope, tracked in #353 (also retires the hand-rolled `UsageChart`).

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Closes #302

## Checklist

- [x] I understand the code I am submitting.
- [x] I have added or updated tests that cover my change (21 new frontend tests: `web/src/pages/OverviewPage.test.tsx`, `web/src/lib/overview.test.ts`).
[...]

PR #356, opened 2026-07-22
feat: price cached input tokens in standalone cost calculation

### LINKED ISSUE #198: Standalone cost does not price cached input tokens
## Summary

Follow-up from PR #196 (which captures cache token counts but does not price them in standalone mode).

`log_usage` (`src/gateway/api/routes/_pipeline.py:807-813`) now records `cache_read_tokens` / `cache_write_tokens` on the `UsageLog` row, but standalone cost is still computed purely from `prompt_tokens` and `completion_tokens`:

```python
cost = (usage_data.prompt_tokens / 1_000_000) * pricing.input_price_per_million + (
    usage_data.completion_tokens / 1_000_000
) * pricing.output_price_per_million
```

As a result, cached tokens are mispriced in standalone mode:

- **OpenAI / Gemini**: cached tokens are a subset of `prompt_tokens`, so they are billed at the **full input rate** rather than the discounted cached rate (overcharge).
- **Anthropic**: `prompt_tokens` (mapped from `input_tokens`) **excludes** cache, so `cache_read_tokens` and `cache_write_tokens` are effectively billed at **$0** (undercharge); cache writes should be billed at a premium.

This is pre-existing and was out of scope for PR #196; that PR captures the data so this can be addressed as a clean follow-up.

## Proposed change

1. Add cache pricing columns to `ModelPricing` (`src/gateway/models/entities.py`), e.g. `cache_read_price_per_million` and `cache_write_price_per_million` (nullable), plus an Alembic migration.
2. Extend the standalone cost calculation in `log_usage` to price `cache_read_tokens` / `cache_write_tokens` when cache rates are configured, handling the provider inclusion conventions (OpenAI/Gemini: cached is a subset of prompt and should be discounted, not double-counted; Anthropic: cache read/write are separate and additive).
3. Surface cache rates in the pricing config loader and the pricing API.
4. Update OpenAPI spec and docs.

## References

- Cost computation: `src/gateway/api/routes/_pipeline.py:807-813`
- Pricing entity: `src/gateway/models/entities.py` (`ModelPricing`)
- Provider conventions: documented in `docs/platform-protocol.md`
- Originating PR: #196

njbrake:
## Description

Standalone mode previously computed cost from `prompt_tokens` and `completion_tokens` only, ignoring the `cache_read_tokens` / `cache_write_tokens` that PR #196 already captures on the `UsageLog` row. This caused cached tokens to be mispriced:

- **OpenAI / Gemini**: cached tokens (a subset of `prompt_tokens`) were billed at the **full input rate** (overcharge).
- **Anthropic**: `cache_read_tokens` and `cache_write_tokens` were billed at **$0** (undercharge), since `prompt_tokens` excludes cache for Anthropic.

This PR prices cache tokens using a single convention borrowed from the genai-prices dataset this project already uses: the input/prompt token count is the grand total that *includes* cache reads and writes, and each physical token is charged once. Providers report cache tokens two different ways, so usage carries an explicit `GatewayUsage.cache_tokens_in_prompt` flag and `_compute_cost` normalizes both shapes onto that one model:

- **OpenAI / Gemini** (`cache_tokens_in_prompt=True`): cache reads are already a subset of `prompt_tokens`, so the cached portion is discounted (re-priced at `cache_read_price_per_million` instead of the full input rate) rather than double counted.
- **Anthropic** (`cache_tokens_in_prompt=False`, native `/v1/messages` path): cache reads/writes are separate from `input_tokens`, so they are folded into the total and billed additively at `cache_read_price_per_million` / `cache_write_price_per_million`. This holds on every turn, including warm-cache reads that create no new cache (`cache_write_tokens` is 0).

The flag is set at the usage-capture boundary (the Anthropic messages adapter sets it False; all OpenAI-shaped paths default True), so the cost function never has to guess the provider. When a cache rate is unset, those tokens stay in the input total and are billed at `input_price_per_million`. Cache buckets are clamped to the input total so cost can never go negative.

### Backend
- **`ModelPricing` entity + migration**: nullable `cache_read_price_per_million` / `cache_write_price_per_million` columns.
- **`GatewayUsage.cache_tokens_in_prompt`** (`core/usage.py`): explicit per-usage flag for the cache-accounting convention, propagated through the streaming accumulator.
- **`_compute_cost()`** (`_pipeline.py`): testable helper that normalizes both conventions, discounts or adds cache tokens per configured rates, falls back to the input rate when a rate is unset, and clamps so cost is never negative.
- **`PricingConfig`** (`config.py`), **pricing init service**, and **default pricing** (`pricing_service.py`, from the genai-prices `cache_read_mtok` / `cache_write_mtok` fields).
- **Pricing API** (`pricing.py`): cache fields on `SetPricingRequest` / `PricingResponse`; `set_pricing` inherits a model's stored cache rates when the client omits the fields, so a partial input/output update never silently wipes cache pricing (an explicit null still clears it).
- **Models API** (`models.py`): `/v1/models` now surfaces cache rates for configured, alias, and default-priced models.

### Dashboard (`web/`)
- Set cache read/write rates in every pricing editor; the pricing table condenses input/output and cache into two paired, clickable columns; the model detail panel is hidden until a row is selected and can be closed.
- Cache read/write totals on the Usage analytics page and a cache-reads tile on the Overview home page (per-request cache tokens were already in the Activity drawer).

### Docs / specs
- Documented cache token pricing in `configuration.md`; regenerated OpenAPI + Postman; rebuilt and committed the dashboard bundle.

## Test plan

- `uv run pytest tests/unit`: passing (`test_compute_cost.py` covers no-cache, OpenAI/Gemini discount, Anthropic additive, warm-cache-read regression, partial-rate fallback, clamp, and the convention flag).
- `npm --prefix web test` / `typecheck`: passing (editors, table/detail display, Usage/Overview tiles, hide/close panel).
- `make lint`, `make typecheck`: passing.
- OpenAPI + Postman freshness checks: up to date.
- Integration: `/v1/models` cache round-trip and `set_pricing` cache persistence / omitted-field preservation added to `test_models_endpoint.py` and `test_pricing_config.py`.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes #198. Follow-up for TTL-aware Anthropic cache-write pricing tracked in #357.

## Checklist

## AI Usage

**AI Model/Tool used:** Laguna S 2.1 (Poolside AI) for the initial draft; Claude Code (via @njbrake) for the corrected cost model, dashboard work, and review follow-ups.

**Any additional AI details you'd like to share:** The cost calculation normalizes both provider cache conventions through an explicit `cache_tokens_in_prompt` flag set at usage capture, not a token-count heuristic. Decisions were made with maintainer clarification on provider cache accounting (genai-prices' inclusive-total model).


src/gateway/api/routes/pricing.py
```diff
@@ -155,12 +165,16 @@ async def set_pricing(
     if pricing:
         pricing.input_price_per_million = request.input_price_per_million
         pricing.output_price_per_million = request.output_price_per_million
+        pricing.cache_read_price_per_million = request.cache_read_price_per_million
+        pricing.cache_write_price_per_million = request.cache_write_price_per_million
```
khaledosman: `set_pricing` overwrites both cache rates unconditionally, and both fields default to `None` on `SetPricingRequest`. Any client that updates a model's input/output price without re-sending cache fields writes `NULL` cache rates. Because an explicit DB row fully beats the (transient, unpersisted) genai-prices default, this silently disables cache pricing for that model: OpenAI/Gemini cached tokens revert to the full input rate (overcharge), Anthropic cache tokens go to $0 (undercharge). This is a live path, not theoretical: the dashboard's own `ModelsPage.tsx` setPricing calls send only input/output. Fix: use `exclude_unset` / a sentinel so omitted cache fields preserve the stored value, or make full-replace explicit and update the dashboard form to always send cache fields.

src/gateway/api/routes/_pipeline.py
```diff
@@ -957,6 +957,58 @@ async def prepare_gateway_tools(
...
+        total_input_tokens = prompt_tokens + cache_read + cache_write
+
+    read_rate = pricing.cache_read_price_per_million
+    write_rate = pricing.cache_write_price_per_million
+
+    # Only pull a cache bucket out of the uncached-input total when it has its
+    # own price; otherwise it stays billed at the input rate.
+    priced_cache_read = cache_read if read_rate is not None else 0
+    priced_cache_write = cache_write if write_rate is not None else 0
+    uncached_input_tokens = total_input_tokens - priced_cache_read - priced_cache_write
```
khaledosman, line 1013: No lower-bound clamp: on the `cache_tokens_in_prompt=True` (OpenAI/Gemini) branch, `total_input_tokens == prompt_tokens`, so if `cache_read` ever exceeds `prompt_tokens` this goes negative and subtracts from cost (undercharge). Today the invariant holds, but the streaming merge combines counts from different chunks with `or`, so a provider quirk could break it. This is money math, a `max(0, ...)` clamp is cheap defense.

khaledosman (commented):
Review by Claude Code (driven by @khaled).

Core money math is correct: `_compute_cost` normalizes both provider conventions via the explicit `cache_tokens_in_prompt` flag (cleaner than the `cache_write_tokens > 0` heuristic the PR description claims, please fix the description to match). Migration chains cleanly onto the current head, single head, reversible. Unit coverage of `_compute_cost` is strong.

One finding worth acting on before merge (set_pricing null-wipe), one defensive minor, plus:

- Testing gap: no integration coverage of the migration up/down, ORM round-trip, `set_pricing` persistence, or `default_model_pricing` seeding end to end. `test_pricing_config.py` was not run locally per the checklist. Suggest extending the existing `test_log_usage_finds_pricing_with_colon_format` pattern to price a cached request through the real DB.
- Em dashes were added in prose/doc-comments (`docs/configuration.md`, several docstrings in `test_compute_cost.py`); AGENTS.md forbids them as separators and `make lint` will not catch it.

Verdict: approve-with-nits.

src/gateway/api/routes/pricing.py
```diff
@@ -155,12 +165,16 @@ async def set_pricing(
     if pricing:
         pricing.input_price_per_million = request.input_price_per_million
         pricing.output_price_per_million = request.output_price_per_million
+        pricing.cache_read_price_per_million = request.cache_read_price_per_million
```
khaledosman: Residual from the first review, now lower severity since the dashboard sends cache fields: the API still assigns both cache rates unconditionally and they default to `None` on the request model, so a non-UI client (curl, the Postman collection) that updates input/output without re-sending cache fields silently nulls the stored cache rates. Since an explicit DB row beats the genai-prices default, that reverts cache tokens to the input rate. Optional: use `exclude_unset`/a sentinel to preserve omitted cache fields, or document `set_pricing` as an intentional full-replace. Non-blocking.

web/src/pages/ModelsPage.tsx
```diff
@@ -755,6 +818,47 @@ function InlinePriceForm({ row, onClose }: { row: ModelRow; onClose: () => void
...
+  rowKey: string;
+  primaryLabel: string;
+  secondaryLabel: string;
+  onEdit: () => void;
+}) {
+  const number = (value: number | null, label: string) => (
+    <button
+      type="button"
+      aria-label={`Edit ${label} price for ${rowKey}`}
+      className="tabular-nums hover:text-[var(--otari-brand-dark)] hover:underline"
```
khaledosman, line 844: Minor: the price numbers reveal their editability only via `hover:underline`, so on touch and for keyboard-only users the fact that these are editable isn't discoverable (clicking still works). A persistent subtle affordance (e.g. always-underlined or a small edit glyph) would surface it without hover. Non-blocking.

web/src/pages/ModelsPage.tsx
```diff
@@ -515,6 +557,14 @@ function ModelDetailPanel({
             </div>
             {metadata?.family ? <p className="text-xs text-[var(--otari-muted)]">{metadata.family}</p> : null}
           </div>
+          <button
+            type="button"
+            aria-label="Close model details"
```
khaledosman, line 562: Minor: this close control (`px-1.5 py-0.5 text-lg`) and the paired price-number buttons are small tap targets on mobile. otari's frontend standards don't specify a minimum size, so this is a usability nit rather than a violation, but a slightly larger hit area would help on touch. Non-blocking.

khaledosman (approved):
Follow-up by Claude Code (driven by @khaled). Re-reviewed after the fixes; approving.

Resolved from the earlier review:
- Negative-cost guard: `_compute_cost` now clamps the cache buckets into the input total, so the uncached remainder can't go negative (and the cache charge is capped too). Correct fix.
- The `set_pricing` cache-null-wipe: the dashboard now sends cache fields at all three `setPricing` sites, so the live regression path is closed.

Non-blocking notes below (see inline). The only one worth considering before merge is the backend `set_pricing` contract, which still full-replaces on a partial payload for non-UI clients.

PR hygiene: this branch bundles the entire #354 overview feature (the branches aren't rebased and `main` has neither yet). Worth rebasing once #354 lands so this PR's diff is just the pricing work; it keeps the squash-merge clean.

New dashboard cache-pricing UI (`ModelsPage.tsx`) is solid and consistent with the file's conventions: tokenized colors, aria-labels, `stopPropagation`, optional-price validation, and tests updated. LGTM.

PR #360, opened 2026-07-22
feat: add dashboard key lifecycle controls

### LINKED ISSUE #308: Dashboard: secret-key & master-key lifecycle

Support rotating the encryption key and the master key from the operator's point of view.

**Why:** `OTARI_SECRET_KEY` rotation is env-only today, and a rotated or lost key leaves stored provider keys "unreadable" with only a manual recovery path.

## Tasks
- [ ] Guided `OTARI_SECRET_KEY` rotation with re-encryption of stored provider credentials (MultiFernet already supports multi-key decrypt).
- [ ] Regenerate the master key from the dashboard (invalidate the old one, show the new one once).
- [ ] A clear recovery flow for the "key unreadable" state (re-enter affected keys).
- [ ] Tests; rebuild bundle.

Part of #299.

njbrake:
## Description

Adds a Credential security category to Dashboard Settings for the lifecycle of both operator keys.

- Adds dashboard regeneration for generated master keys, with an explicit confirmation and one-time, copy-first reveal. The old key is invalidated immediately, while the current tab switches to the new key without collapsing or scrolling the Settings view.
- Identifies configuration-managed master keys and directs operators to rotate `OTARI_MASTER_KEY` or `config.yml`, then restart.
- Adds provider-credential re-encryption for staged `OTARI_SECRET_KEY` rotation, including exact restart guidance and recovery guidance for unreadable provider keys.
- Exposes the non-secret master-key source in the Settings API so the dashboard can present the correct lifecycle action.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes #308  
Part of #299

## Checklist

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:** GPT-5 Codex

**Any additional AI details you would like to share:** Implemented and validated with operator-directed UX review.

## Summary

- Added a **Credential security** section to Dashboard Settings for safely managing operator master-key regeneration and provider credential encryption.
- Introduced **operator-driven master key regeneration** with explicit confirmation, one-time key reveal/copy, and immediate invalidation of the old key—while keeping the current settings screen usable.
- Added **provider credential re-encryption** support to help rotate `OTARI_SECRET_KEY`, including reporting of unreadable entries and an in-UI recovery path.
- Extended the **Settings API** to expose the master-key source (configured vs generated) and added a master-key rotation endpoint.
- Updated the frontend to add the necessary API calls and user flows, expanded/added integration and unit tests, refreshed generated dashboard assets, and regenerated the OpenAPI spec.


khaledosman (commented):
Solid PR: correct staged-then-commit ordering, no secrets logged or returned, both endpoints gated by `verify_master_key`, and thorough tests. Four findings inline (most severe first): one multi-replica correctness/security gap in rotation, and three frontend-standards deviations.

src/gateway/api/routes/settings.py
```diff
@@ -331,3 +342,29 @@ async def update_settings(
...
+            status_code=status.HTTP_409_CONFLICT,
+            detail="This gateway uses a configured master key; change OTARI_MASTER_KEY or config.yml and restart.",
+        )
+    try:
+        token, hashed = await stage_generated_master_key_rotation(db)
+        await db.commit()
+    except SQLAlchemyError:
+        await db.rollback()
+        raise HTTPException(status_code=status.HTTP_500_INTERNAL_SERVER_ERROR, detail="Database error") from None
+    config._master_key_hash = hashed
```
khaledosman, line 377: **[Medium] Rotation only takes effect on the replica that served the request.** The new hash is committed to the DB, but this updates only this process's in-memory `config._master_key_hash`, which auth (`api/deps.py:186`) reads and which is loaded once at startup and never re-read from the DB.

The documented scale-out is several processes behind a load balancer (`cli.py:96`), and `main.py:126` builds TTL refreshers so alias/provider writes converge across replicas. The master-key hash has no equivalent refresher, so after a rotation on a multi-replica deployment:
- the **old key stays valid** on every other replica until restart (rotation does not revoke fleet-wide, defeating the point when the key leaked);
- the **new key is rejected (401)** on other replicas, intermittently signing the operator out.

`test_rotate_generated_master_key_invalidates_old_key` uses a single in-process client, so it can't catch this.

Fix: re-read the hash from `runtime_settings` in the auth path (mirroring the alias/provider convergence), or document the "restart other replicas" caveat and soften the UI copy in `SettingsPage.tsx` ("This immediately invalidates the current dashboard master key. Other signed-in dashboard sessions will need the new key").

web/src/pages/SettingsPage.tsx
```diff
@@ -258,6 +265,271 @@ function ConfigRow({
...
+      event.preventDefault();
+      first.focus();
+    }
+  };
+
+  return (
+    <div className="fixed inset-0 z-50 flex items-center justify-center bg-black/40 p-4" role="presentation">
+      <div
+        ref={dialogRef}
+        role="dialog"
```
khaledosman: **[Medium] Use HeroUI v3 `Modal` instead of a hand-rolled dialog + focus trap.** This builds a raw `<div role="dialog">`, backdrop, and hand-written Tab focus trap. The frontend standards say to use HeroUI v3's compound `Modal` and to check `components/ui.tsx` before hand-rolling. HeroUI's Modal provides focus trapping, scroll-lock, portaling and ARIA for free; the hand-rolled trap has no scroll-lock and is easy to get subtly wrong.

web/src/pages/SettingsPage.tsx
```diff
@@ -258,6 +265,271 @@ function ConfigRow({
...
+  return (
+    <div className="flex flex-col gap-1">
+      <div className="flex items-center justify-between">
+        <span className="text-xs font-medium text-[var(--otari-muted)]">New master key</span>
+        <Button size="sm" variant="outline" onPress={copy}>
+          {copied ? "Copied" : "Copy"}
+        </Button>
+      </div>
+      <input
+        ref={ref}
```
khaledosman, line 300: **[Low] Raw `<input>` re-skinned via `className`.** The `border/bg/text` classes re-skin what a HeroUI `Input` (or the `Field` primitive) already styles. Standards: reserve `className` for layout/positioning, not for re-skinning a HeroUI-styled component.

web/src/pages/SettingsPage.tsx
```diff
@@ -258,6 +265,271 @@ function ConfigRow({
...
+        readOnly
+        value={value}
+        onFocus={(event) => event.currentTarget.select()}
+        className="w-full rounded-lg border border-[var(--otari-line)] bg-[var(--otari-bg)] px-3 py-2 font-mono text-xs text-[var(--otari-ink)]"
+      />
+      <span aria-live="polite" className="text-xs text-green-700">
```
khaledosman: **[Nit] `text-green-700` is outside the documented palette exception.** The raw-Tailwind-palette exception in design-tokens.md covers only the `ErrorBanner` (red) / `InfoBanner` (amber) status banners; a bare green span is chrome. Use an `--otari-*` token or an `InfoBanner`.

khaledosman (approved):
Approving. `4619796b` resolves the multi-replica gap from my earlier review, and goes further than suggested.

**Finding #1 (Medium) — resolved.** `_is_valid_master_key` now reads the hash from the DB per request (`deps.py`), so rotation converges across all replicas immediately: the old key is rejected and the new key accepted fleet-wide, closing both the revocation gap and the intermittent-401 sign-out. The lookup is correctly short-circuited behind `is_generated_master_key(token)` and the configured-key case, so regular `gw-` API keys and operator-configured deployments incur no extra query.

**Bonus hardening.** Rotation is now a compare-and-swap (`UPDATE ... WHERE value = current_hash` → 409 `MasterKeyRotationConflictError` on `rowcount != 1`), which closes the lost-update window between two concurrent rotations. Matches the atomic-CAS guidance in the security-review instructions (§0.2).

Backend LGTM. The three frontend findings from the first review remain open and are non-blocking:
- #2 (Medium) hand-rolled dialog + focus trap instead of HeroUI v3 `Modal`
- #3 (Low) raw `<input>` re-skinned via `className`
- #4 (Nit) `text-green-700` outside the documented palette exception

PR #364, opened 2026-07-22
fix: preserve Codex Responses metadata for OpenAI

### LINKED ISSUE #362: Accept or strip Codex metadata extensions in /v1/responses input items
### Problem

Otari rejects otherwise valid OpenAI Responses API requests from Codex during request validation.

Codex sends standard input items such as `type: "message"`, with `developer` and `user` roles plus `input_text` content, but includes the per-item metadata extension `internal_chat_message_metadata_passthrough` (for example, a `turn_id`). It also sends top-level `client_metadata` (for example, a `session_id`).

Otari's `ResponsesParams` Pydantic model rejects both as `extra_forbidden`, producing a very large union-validation response (`509 validation errors`) that makes the actual cause hard to identify.

### Expected behavior

Otari should accept Codex Responses requests and either:

1. Strip `internal_chat_message_metadata_passthrough` from individual input items and `client_metadata` from the request before validation/forwarding, or
2. Explicitly allow and ignore these metadata fields in the relevant request models.

The fields are metadata only and should not affect the request sent upstream.

### Actual behavior

The request fails before reaching the provider. The first useful errors are:

```text
input.list[...].0.Message.internal_chat_message_metadata_passthrough
  Extra inputs are not permitted [type=extra_forbidden]

client_metadata
  Extra inputs are not permitted [type=extra_forbidden]
```

The remaining errors are Pydantic union fallthrough after the `Message` variant is rejected.

### Reproduction

1. Configure Codex to use Otari as its OpenAI-compatible Responses endpoint.
2. Run a basic Codex request, for example:

```bash
codex exec -m work-openai:gpt-5.5 'Reply with exactly: otari-route-ok'
```

3. Otari returns `509 validation errors for ResponsesParams`.

A minimal hand-written `/v1/responses` request without Codex metadata succeeds, confirming the model route and gateway authentication are functional.

### Scope

This is related to #331 in that both affect tool-capable Codex/OpenAI traffic, but it is a separate issue: this failure happens in Otari's own `/v1/responses` input validation, before any upstream provider request.

During investigation please also investigate if this is an issue with any-llm itself, not just otari

njbrake:
## Description

Preserves Codex `client_metadata` and per-item `internal_chat_message_metadata_passthrough` on the OpenAI Responses route by carrying them through any-llm 1.22.0's typed `extra_body` handoff. Non-OpenAI routes receive the standards-compliant input without the Codex extension. Client-supplied raw provider bodies remain blocked.

The dependency release came from https://github.com/mozilla-ai/any-llm/pull/1192.

## PR Type

- [x] Bug Fix

## Relevant issues

Fixes #362

## Checklist

- [ ] I ran the Definition of Done checks locally (`make lint`, `make typecheck`, `make test`). `make lint`, `make typecheck`, and all unit tests pass; the integration suite requires Docker, which is unavailable in this environment.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:** GPT-5 via Codex

**Any additional AI details you'd like to share:** The implementation and tests were reviewed against repository conventions and verified locally.

## Summary

- Updated the Responses route to accept and preserve Codex metadata for OpenAI requests.
- Removed Codex-only metadata from standard request inputs and excluded it from unsupported providers.
- Prevented client-supplied provider-specific request bodies from being forwarded.
- Updated the any-llm dependency and added coverage for metadata forwarding and filtering behavior.

## Benefits

- Fixes Codex `/v1/responses` validation failures while preserving turn context and attribution.
- Keeps requests to other providers standards-compliant and secure.

### DIFF
--- a/src/gateway/api/routes/responses.py
+++ b/src/gateway/api/routes/responses.py
@@ -404,6 +404,10 @@ async def create_response(
         tools_extracted=tool_ctx.tools_extracted,
         remaining_user_tools=tool_ctx.remaining_user_tools,
     )
+    # This is an internal handoff populated below, never a client request field.
+    # ``ResponsesRequest`` permits extra fields for OpenAI compatibility, so
+    # remove a caller-supplied value before constructing the provider kwargs.
+    request_fields.pop(_CODEX_EXTRA_BODY_FIELD, None)
     client_metadata = request_fields.pop(_CODEX_CLIENT_METADATA_FIELD, None)
     if request_fields.get("tools"):
         request_fields["tools"] = openai_to_responses_tools(request_fields["tools"])
--- a/src/gateway/api/routes/responses.py
+++ b/src/gateway/api/routes/responses.py
@@ -53,6 +53,9 @@ router = APIRouter(prefix="/v1", tags=["responses"])
 
 _MASTER_KEY_USER_REQUIRED = "When using master key, 'user' field is required in request body"
 _USER_FORBIDDEN = "'user' field does not match the authenticated API key's user"
+_CODEX_CLIENT_METADATA_FIELD = "client_metadata"
+_CODEX_INPUT_METADATA_FIELD = "internal_chat_message_metadata_passthrough"
+_CODEX_EXTRA_BODY_FIELD = "_codex_extra_body"
 
 
 class ResponsesRequest(derive_request_base(ResponsesParams)):  # type: ignore[misc]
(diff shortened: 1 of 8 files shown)


src/gateway/api/routes/responses.py
```diff
@@ -224,9 +278,10 @@ def attempt_kwargs(
         kwargs: dict[str, Any] = {"api_key": attempt.api_key}
         if attempt.api_base:
             kwargs["api_base"] = attempt.api_base
+        request_fields = _with_codex_extra_body(base_request_fields, LLMProvider(attempt.provider))
```
khaledosman, line 285: This `_with_codex_extra_body` translation only runs on paths that go through `attempt_kwargs` — i.e. standalone (both inline at L464/L496) and hybrid **streaming** (`_pipeline.py:1710`). The hybrid **non-streaming** path does not: `run_platform_non_stream` → `run_platform_attempts` uses `default_attempt_kwargs` (`_platform.py:210`), which spreads `**base_request_fields` verbatim and never calls this override. The only hook on that path, `prepare_platform_call_kwargs` (L289), just re-splits `provider:model`.

Result: `base_request_fields[_codex_extra_body]` (set at L433) is passed straight to `aresponses` as a stray kwarg (swallowed by its `**kwargs`), so `extra_body` is never built and the Codex metadata is silently dropped in platform mode — the deployment #362 is about.

Fix in the single chokepoint for that path:

```python
def prepare_platform_call_kwargs(self, kwargs: dict[str, Any]) -> dict[str, Any]:
    merged_model = kwargs.pop("model")
    provider_str, model_str = merged_model.split(":", 1)
    kwargs["model"] = model_str
    provider = LLMProvider(provider_str)
    kwargs["provider"] = provider
    return _with_codex_extra_body(kwargs, provider)
```

tests/integration/test_responses_route_dispatch.py
```diff
@@ -101,6 +101,100 @@ async def fake_aresponses(**kwargs: Any) -> Response:
     assert captured["input_data"] == prompt
 
 
+def test_codex_metadata_is_forwarded_to_openai_provider(
```
khaledosman, line 104: Both new integration tests cover standalone only. Add a hybrid/platform non-streaming case asserting `captured["extra_body"]` is populated (and that a non-OpenAI attempt gets none). That test would fail against the current code and would have caught the dropped-metadata gap flagged in `responses.py`. The standards require new billable logic to be correct in both standalone and hybrid mode.

khaledosman (changes requested):
Reviewed the full diff plus the untouched dispatch paths in `_pipeline.py` / `_platform.py`, against the backend + security standards.

One substantive bug: the Codex `extra_body` handoff is wired into standalone (stream + non-stream) and hybrid streaming, but **not** hybrid non-streaming — the exact platform-mode deployment this PR targets. Metadata is silently dropped there. Details inline, plus a test-coverage note (no hybrid test, which is why it slipped through).

The rest is solid: `extra_body` stripping + smuggle prevention, no aliasing in `_split_codex_input_metadata`, non-OpenAI providers get standards-compliant input, and no budget/auth/tenant regressions.

khaledosman (approved):

PR #367, opened 2026-07-22
feat(pricing): add reviewable price refresh

### LINKED ISSUE #361: Dashboard/core: allow for refresh of genai-prices with latest
When a new model comes out, I think we need to upgrade the genai-prices package to pull in the prices. We need a button in the dashboard (and maybe a feature in the gateway) to update with the latest prices. 

There should be some sort of logic to confirm the changes/updates before they take effect. It should also intelligently handle prices that have been manually set.

Maybe we need some sort of highlighting/syntax in the web dashboard to indicate a custom set price vs a default price.

njbrake:
## Description
Adds a reviewable Settings workflow for updating the `genai-prices` catalog. Accepted snapshots and pending reviews are persisted in the local database, custom prices remain higher-priority overrides, and the Models page is simplified with configurable 15, 25, or 50 row pagination.

## PR Type
- [x] New Feature
- [x] Bug Fix
- [x] Documentation

## Relevant issues
Fixes #361

## Checklist
- [ ] I ran the Definition of Done checks locally (`make lint`, `make typecheck`, `make test`). Integration tests require Docker, which is unavailable in this environment.

## AI Usage

**AI Model/Tool used:** Codex CLI

**Any additional AI details you would like to share:** All code and test changes were reviewed locally.

## Summary

- Added a Settings workflow to check for GenAI pricing catalog updates and review added/changed/removed rates before applying them.
- Persisted pending and accepted pricing snapshots in the local database, with startup loading and multi-worker convergence.
- Kept manually configured/custom model prices as higher-priority overrides so refreshes don’t overwrite them.
- Updated the Models page to simplify the UI (removing provider/context rendering and the “price an unlisted model” flow) and added configurable pagination (15/25/50 rows per page).
- Updated documentation, API examples, frontend types/hooks, and regenerated OpenAPI-generated assets; added/updated backend and integration tests.

## Technical notes

- Implemented master-key-protected refresh endpoints for preview/confirm/reject, backed by a new DB table for pricing snapshots and a refresh service that computes diffs and enforces a pending/approved flow.
- Canceled background refresher task creation if log writer startup fails, and improved shutdown handling to avoid stopping an unstarted log writer.
- Definition of Done checks were not run because integration tests require Docker, which was unavailable.


src/gateway/services/pricing_refresh_service.py
```diff
@@ -0,0 +1,215 @@
...
+    else:
+        active_row.snapshot = pending_row.snapshot
+    await session.delete(pending_row)
+    try:
+        await session.commit()
+    except SQLAlchemyError as exc:
+        await session.rollback()
+        raise PricingRefreshError("Unable to save the latest genai-prices data") from exc
+
+    set_custom_snapshot(snapshot)
```
khaledosman, line 194: Multi-worker/replica divergence: this runs only in the worker that served the confirm; other workers/replicas keep the old snapshot until restart. Aliases and provider credentials have TTL refreshers for exactly this reason (documented in AGENTS.md); the accepted pricing snapshot has none, so an accepted price change applies inconsistently across a multi-worker deployment. Add a reload hook/refresher or document the restart requirement.

src/gateway/services/pricing_refresh_service.py
```diff
@@ -0,0 +1,215 @@
...
+
+
+async def confirm_price_refresh(session: AsyncSession) -> bool:
+    """Persist and activate the pending snapshot, returning false when absent."""
+
+    pending_row = (
+        await session.execute(
+            select(PricingSnapshot)
+            .where(PricingSnapshot.source == GENAI_PRICES_PENDING_SOURCE)
+            .with_for_update()
```
khaledosman, line 167: with_for_update() is a no-op on SQLite (the standalone default), so concurrent confirms/rejects aren't actually serialized there. Master-key-only admin path so low risk, but the lock gives false assurance on the default backend.

khaledosman (approved):
Approving to unblock. Solid change. All three endpoints are master-key gated (test confirms 401), the network fetch runs off the event loop via asyncio.to_thread, the DB commit precedes the in-memory set_custom_snapshot, persisted snapshots are re-validated on load, and the billable settlement path is untouched. httpx2 is a legitimate dep (it is genai-prices' own HTTP client). Migration chains correctly to head with a real downgrade().

Inline notes below. One product question: ModelsPage drops "Price a model not listed here" (PriceModelForm/ModelComboBox), so an operator can no longer price a brand-new or unconfigured-provider model from the UI (the /v1/pricing API still supports it). Reads as intentional per the description; just confirming it's the desired call.

Issue #376, opened 2026-07-22
Epic: provider-metered pricing for accurate cost accounting and budgets


njbrake:
## Summary

PR #356 corrected the first cache-pricing bug, but Otari still models a provider bill as four scalar token rates. That is not sufficient for current provider pricing, so recorded spend and budget enforcement can drift from actual charges.

This epic introduces a provider-metered pricing engine. It makes actual post-response cost reproducible and makes pre-call reservations conservative when the exact bill is not knowable yet.

## Why the current model is insufficient

- `ModelPricing` has only input, output, cache-read, and cache-write rates.
- Default pricing deliberately flattens `TieredPrices` to the base rate, so context-threshold pricing is silently lost.
- `UsageLog` stores only aggregate prompt, completion, cache-read, and cache-write token counts. It cannot preserve the provider meters or rate conditions that explain a bill.
- The reservation estimate uses a character heuristic and base input/output rates. It cannot reserve for a long-context tier, a cache write, a premium service tier, or a billable tool call.

Examples from current provider documentation:

- OpenAI: cache reads are a subset of input; GPT-5.6+ reports billable cache writes. Standard, Batch, Flex, Priority, regional processing, search, file search, and hosted containers are distinct price dimensions.
- Anthropic: 5-minute writes, 1-hour writes, and reads are distinct cache meters. Batch, Fast mode, and data residency stack with them.
- Gemini: some models change input, output, and cache rates above a context threshold; explicit cache storage is billed per token-hour.
- DeepSeek and Kimi expose cache-hit and cache-miss pricing that must be captured as separate usage values.

Provider references:

- https://developers.openai.com/api/docs/pricing
- https://developers.openai.com/api/docs/guides/prompt-caching
- https://platform.claude.com/docs/en/about-claude/pricing
- https://platform.claude.com/docs/en/build-with-claude/prompt-caching
- https://ai.google.dev/gemini-api/docs/pricing
- https://api-docs.deepseek.com/quick_start/pricing
- https://api-docs.deepseek.com/guides/kv_cache
- https://platform.kimi.ai/docs/pricing/chat-k3

## Immediate correctness tranche

Do these first. They address known misbilling without waiting for every non-token feature.

1. Define a provider-specific `BillableUsage` normalizer at the usage-capture boundary. Preserve safe raw usage metadata for reconciliation and normalize into named meters rather than forcing every provider into one overlap assumption.
2. Capture and test:
[...]

PR #379, opened 2026-07-22
feat(pricing): meter cache writes and long context

### LINKED ISSUE #357: Support TTL-aware Anthropic cache-write pricing (5-min vs 1-hour)

## Summary

Cache-token pricing (PR #356) bills Anthropic cache-**write** (creation) tokens at a single flat rate. Anthropic actually prices cache writes by **TTL**: 5-minute writes at 1.25x the input rate, 1-hour writes at 2x. We currently collapse both into one `cache_write_price_per_million`, so 1-hour cache writes are undercharged relative to Anthropic's real price.

This is a tracked follow-up to the cache-pricing work; not a regression in #356, which is correct for the single-rate model.

## Current behavior

- `ModelPricing` has one `cache_write_price_per_million` column.
- `services/pricing_service.py` populates it from genai-prices, whose `ModelPrice` exposes only a single `cache_write_mtok` (the 5-minute/base rate). There is no 1-hour variant in the dataset.
- `api/routes/messages.py` reads only `result.usage.cache_creation_input_tokens` (the total), discarding the split.
- Net: a 1-hour Anthropic cache write is billed at the 5-minute rate, i.e. undercharged by roughly 2x / 1.25x.

## What Anthropic and peers do

- Anthropic's SDK `Usage.cache_creation` is a `CacheCreation` object with `ephemeral_5m_input_tokens` and `ephemeral_1h_input_tokens`; `cache_creation_input_tokens` is their sum.
- LiteLLM and Bifrost both split these buckets and price them at distinct rates (`cache_creation_input_token_cost` vs `cache_creation_input_token_cost_above_1hr`).

## Rough scope

1. Capture the split on the native path: in `messages.py`, read `usage.cache_creation.ephemeral_5m_input_tokens` / `ephemeral_1h_input_tokens` instead of only the total. Carry the 1-hour portion through `GatewayUsage` (new field) and persist it (usage-log column + migration).
2. Add a second pricing rate: `cache_write_price_per_million_above_1h` (or similar) on `ModelPricing` (+ migration), `PricingConfig`, the `/v1/pricing` API request/response, and `to_dict`.
3. Extend `_compute_cost` in `_pipeline.py` to charge the 1-hour portion at the new rate and the remainder at the existing rate, following the same clamp-and-normalize model already in place.
4. Defaults: genai-prices has no above-1h rate, so the new column defaults to `None` and, per the existing missing-rate convention, those tokens fall back to the base cache-write rate (not the input rate, since they are still cache writes). Confirm that is the desired fallback.
5. Docs + OpenAPI/Postman regeneration.

## Constraint worth noting up front

This only works on the native `/v1/messages` endpoint. On `/v1/chat/completions`, cache-write tokens are lost entirely: any-llm normalizes provider usage to the OpenAI shape, which has no slot for cache creation, so `cache_creation` is folded into `prompt_tokens` and billed at the input rate. Full parity would first require any-llm to surface the cache-write count (and its TTL split), e.g. via the `extra_content.anthropic` channel it already uses for other Anthropic-only fields. Track that separately in any-llm if we want it.

## Acceptance

- A 1-hour Anthropic cache-write turn on `/v1/messages` is billed at the 1-hour rate when configured, and at the base cache-write rate when not.
- 5-minute writes and all existing behavior are unchanged.
- Unit tests cover the 5m-only, 1h-only, and mixed cases.

### LINKED ISSUE #372: Pricing: handle graded pricing between long and short context
OpenAI for example bills differently between long context and short context. 

https://developers.openai.com/api/docs/pricing?latest-pricing=standard

Our pricing model needs to be flexible to accomodate a variety of pricing models.

njbrake:
## Description

Adds provider-metered usage pricing for cache reads, regular cache writes, Anthropic one-hour cache writes, and request-level long-context price tiers. Usage records now retain billing meters and a pricing breakdown for auditability.

The standalone dashboard now compares base and effective rates at a selected context size, presents compact cache and tier policies, supports inline tier editing, and avoids repeating the provider instance in model labels.

## PR Type

- [x] New Feature
- [x] Bug Fix
- [ ] Refactor
- [x] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Closes #357
Closes #372
Refs #376
Refs mozilla-ai/any-llm#1195

## Checklist

Validation: `make lint`, `make typecheck`, `uv run pytest tests/unit` (857 passed, 1 skipped), `npm --prefix web run typecheck`, `npm --prefix web test` (195 passed), dashboard production build, and OpenAPI freshness check all pass. Integration tests could not run locally because Docker is unavailable.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:** Codex

**Any additional AI details you would like to share:** Implementation and validation were performed interactively with maintainer direction.

## Summary

- Added flexible, tiered pricing for long-context requests, plus separate pricing for standard cache writes and Anthropic “one-hour” cache writes.
- Extended usage logging and usage/analytics endpoints to store cache-write hour metrics and keep an auditable breakdown of how charges were computed.
- Updated pre-dispatch budgeting/estimation to reserve and price cache-write tokens correctly based on both the cache-write TTL (5m vs 1h) and the applicable context tier.
- Enhanced the dashboard to compare base vs effective rates at a selected context size, show cache and pricing-policy details more compactly, support inline editing of pricing tiers, display “1h cache writes,” and avoid repeating provider names in model labels.
- Updated pricing configuration, API schemas, and documentation; added/extended migrations and validation; regenerated dashboard assets and refreshed related frontend/backend tests and response contracts.

## Technical notes

- The “one-hour” cache-write rate falls back to the regular cache-write rate when not configured, while explicitly configured zero rates remain free.
- Pricing tiers act as whole-request thresholds (a tier applies to the entire request once its threshold is reached).
- The change also includes stricter validation for pricing tiers (e.g., no duplicate thresholds; tiers must override at least one price field).
- Unit tests, dashboard tests, lint/type checks, OpenAPI freshness, and production build passed; integration tests could not be run locally because Docker was unavailable.


njbrake:
Addressed the current CodeRabbit findings in 156e7980:

- reserve long-context and explicitly requested Anthropic cache-write charges before provider dispatch
- reject no-op pricing tiers and correct tiered-default documentation
- sanitize provider diagnostics in the dashboard activity detail
- order context thresholds numerically and update the dashboard tests to mock apiFetch

Also updated the five stale exact-response integration assertions that caused the previous CI failure. Local validation: 870 unit tests passed, dashboard tests passed, lint, type checks, OpenAPI freshness, and dashboard build passed. CodeRabbit is currently rate-limited for a follow-up review.

src/gateway/api/routes/pricing.py
```diff
@@ -25,8 +25,19 @@
...
 
+class PricingTier(BaseModel):
+    """Whole-request price cliff selected by total billable input tokens."""
+
+    min_input_tokens: int = Field(gt=0)
+    input_price_per_million: float | None = Field(default=None, ge=0)
+    output_price_per_million: float | None = Field(default=None, ge=0)
+    cache_read_price_per_million: float | None = Field(default=None, ge=0)
+    cache_write_price_per_million: float | None = Field(default=None, ge=0)
+    cache_write_1h_price_per_million: float | None = Field(default=None, ge=0)
```
khaledosman, line 36: This API tier model lacks the "at least one rate override" check that `PricingTierConfig` enforces in `core/config.py` (`validate_has_rate_override`). Via `POST /v1/pricing` a tier of just `{min_input_tokens}` validates and is stored as a silent no-op at billing time. Add the matching `model_validator` for parity so config and API reject the same input. (Master-key only, so not a security issue.)

web/src/pages/ActivityPage.tsx
```diff
@@ -144,21 +126,20 @@ function DetailField({ label, children }: { label: string; children: ReactNode }
...
           <div className="flex items-center justify-between">
             <span className="text-[11px] font-medium uppercase tracking-wide text-[var(--otari-muted)]">
-              Error (provider-reported)
+              Error
             </span>
-            <CopyButton text={entry.error_message} />
           </div>
           <pre className="max-h-48 overflow-auto rounded-lg border border-red-200 bg-red-50 p-3 text-xs whitespace-pre-wrap break-all text-red-700">
-            {entry.error_message}
+            The provider returned an error. Inspect gateway logs for details.
```
khaledosman, line 142: The provider-error redaction (dropping `CopyButton` + the verbatim `error_message`) is unrelated to pricing; CONTRIBUTING asks to keep diffs focused, so this reads more naturally as its own PR. Also note `error_message` is still returned by the `/v1/usage` API, so this hides it in the UI only, not at the boundary. Reasonable hardening either way.

src/gateway/api/routes/messages.py
```diff
@@ -359,6 +394,12 @@ async def _normalize(
             user_id_from_request=str(user_from_metadata) if user_from_metadata else None,
             estimate_prompt_chars=len(str(request.messages)) + len(str(request.system or "")),
             estimate_max_output_tokens=request.max_tokens,
+            estimate_cache_write_ttl=_requested_cache_write_ttl(
```
khaledosman, line 397: `estimate_cache_write_ttl` is wired only into the Anthropic `/v1/messages` path. Anthropic prompt caching driven through `/v1/chat/completions` won't be reflected in the up-front reservation, so those requests under-reserve (final spend still reconciles correctly). Consistent with prior behavior; flagging only as a known asymmetry.

khaledosman (approved):
Reviewed the full diff against the backend/security/performance and dashboard standards. Billing path is sound: `fresh_input` stays non-negative in both the additive (Anthropic) and subset (OpenAI) conventions, the 1h-cache-write rate fallback is correct, a configured $0 rate is preserved, and the reserve/reconcile lifecycle is unchanged. Migration is reversible with a `server_default`, OpenAPI + dashboard bundle regenerated, docs updated, and the unit/integration coverage hits the tricky cases (whole-request cliff, 1h vs 5m rate, $0 rate, conservative over-reservation).

Approving. Three low-severity notes inline, none blocking.

src/gateway/api/routes/pricing.py
njbrake: Fixed in fc10f6c5. The API now rejects no-op tiers with the same validation as config, and unit plus route coverage cover the rule.

web/src/pages/ActivityPage.tsx
njbrake: Kept this focused UI hardening because it resolves the CodeRabbit security finding: the dashboard no longer renders or copies raw provider diagnostics. Agreed that the master-key usage API remains a separate operator/API-boundary policy, so I did not broaden that change here.

src/gateway/api/routes/messages.py
njbrake: Verified before changing: ChatCompletionRequest has no cache_control field and its default Pydantic config drops unknown extras, so the OpenAI-compatible endpoint cannot request Anthropic prompt-cache creation. Native Anthropic cache controls are accepted only on /v1/messages, where the conservative reservation is applied.

src/gateway/api/routes/pricing.py
```diff
@@ -25,8 +25,19 @@
...
+
```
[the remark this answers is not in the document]
njbrake: Fixed in fc10f6c5, which adds the matching per-tier override validator to the pricing API and regression coverage.

src/gateway/services/metered_pricing.py
```diff
@@ -0,0 +1,167 @@
...
+
+    if cache_write_ttl == "1h":
+        cache_write_rate = rates["cache_write_1h_price_per_million"]
+        if cache_write_rate is None:
+            cache_write_rate = rates["cache_write_price_per_million"]
+    else:
+        cache_write_rate = rates["cache_write_price_per_million"]
+    if cache_write_rate is not None:
+        cost += cache_write_tokens * cache_write_rate / 1_000_000
+    return cost
```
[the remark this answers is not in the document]
njbrake: Fixed in e1c2d4c1. Conservative reservations now price potential cache-write tokens at the selected normal input rate when no cache-write rate is configured, matching final reconciliation. The regression test now asserts the doubled input charge.

Issue #156, opened 2026-06-12
Stream uploads to object storage (S3 / MinIO / local) instead of buffering in memory


khaledosman:
## Background

Raised in review of #91 (file uploads & document understanding).

`_read_capped` in `src/gateway/api/routes/files.py` accumulates the entire upload in memory (`chunks: list[bytes]`) up to `files_max_bytes` (default 512 MiB), and `FileStore.put(file_id, data: bytes)` then takes the full byte string. The cap prevents truly unbounded reads, but a handful of concurrent max-size uploads can still pin gigabytes of RAM.

This was deliberately deferred from #91: lowering the default cap is only a stopgap, and the real fix is tied to a streaming-capable storage backend.

## Scope

Add an object-storage `FileStore` backend and a streaming put path so uploads flow chunk-by-chunk to the backend instead of being fully buffered:

- Extend / add to the `FileStore` interface a streaming/multipart upload path (e.g. `put_stream`) so `_read_capped` can forward chunks straight through rather than accumulating them, and likewise stream on download.
- New backend selectable via `files_backend` alongside the existing `local` filesystem backend.

## Storage options (no forced cloud dependency)

Users must be able to run file uploads without any cloud service. Support, at minimum:

- **Local filesystem** (existing `local` backend) stays first-class; no object store required. This also covers pointing `files_local_dir` at a mounted volume (Docker/k8s persistent volume, NFS, etc.), which gives durable, scalable storage with no object store at all.
- **Self-hosted S3-compatible** via MinIO (and any S3-compatible endpoint), configured by endpoint URL + credentials, not hardcoded to AWS.
- **AWS S3** for users who want managed cloud storage.

The S3-compatible and AWS paths should share one implementation (S3 API with a configurable endpoint); MinIO is just an endpoint override. Nobody should be forced onto AWS to use file uploads.

## Acceptance

- A configurable storage backend that streams uploads/downloads without buffering the whole file in memory.
- Local-only operation remains fully supported (no cloud account needed).
- MinIO / S3-compatible endpoints work via configurable endpoint + credentials.

champ18ion:
Hey! I'd like to pick this up. To align before writing code:

[...]

PR #380, opened 2026-07-23
feat(files): stream uploads and downloads instead of buffering in memory

champ18ion:
## Description
`_read_capped` accumulated an entire upload into a list of chunks before handing it to `FileStore.put`, and `get_file_content` buffered the full blob before responding. A handful of concurrent max-size uploads could still pin gigabytes of RAM even with the size cap in place (see #156).

Adds `put_stream`/`get_stream` to the `FileStore` protocol so bytes flow chunk-by-chunk instead of being buffered whole. `LocalDirFileStore` streams to/from disk directly and cleans up the partial file if the chunk source fails mid-stream (e.g. the route's 413 size-cap check firing partway through). `put`/`get` are untouched for callers like the content normalizer that need the full blob in memory regardless.

This is PR 1 of 2 per the discussion on #156: local-backend streaming now, S3/MinIO backend (sync `boto3` + `asyncio.to_thread`, since `aioboto3` conflicts with the pinned `botocore`) as a follow-up.

## PR Type
- [x] New Feature

## Relevant issues
Part of #156

## Checklist
- [x] I ran the Definition of Done checks locally (`make lint`, `make typecheck`, `make test`) — note: ran via a manual pip venv, not `uv`, since Windows is excluded from your lockfile's supported platforms; integration tests (`tests/integration`) couldn't run locally without Docker, but the modified files pass lint/typecheck/unit tests and the integration test file collects cleanly.
- [x] Documentation was updated where necessary (docstrings).

## AI Usage

**Any additional AI details you'd like to share:** Design discussed and reviewed with @khaledosman on #156 before implementation; code written by Claude Code under my direction and review.

## Summary

- Added streaming upload and download for the `/files` endpoints so large files can be transferred without loading entire contents into memory.
- Updated the file storage interface to support streaming reads/writes, and implemented chunk-by-chunk disk IO in the local file store.
- Improved failure handling: partially written files and now-empty shard folders are cleaned up when uploads fail, are cancelled, or exceed the size limit.
- Adjusted the download response to fail early if the first read chunk errors, and removed `Content-Length` to avoid mismatches with the streamed body.
- Updated Postman documentation and expanded unit tests to cover streaming behavior, cancellation-safe cleanup, and download “priming” behavior.

## Technical notes

- Upload streaming keeps the existing size cap; if the stored byte count ends up as zero, the just-written data is deleted and the request is rejected.
- Downloads use a small “prime first chunk” helper so read errors happen before the response is started.


src/gateway/services/file_store.py
```diff
@@ -5,19 +5,39 @@
...
+
+
+@asynccontextmanager
+async def _open_handle(path: Path, mode: str) -> AsyncIterator[IO[bytes]]:
+    """Open ``path`` off the event loop, guaranteeing the handle is closed."""
+    handle = await asyncio.to_thread(path.open, mode)
+    try:
+        yield handle
+    finally:
+        await asyncio.to_thread(handle.close)
```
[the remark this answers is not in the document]
champ18ion: Good catch, fixed the same issue you'd have hit on line 140 too: `handle.close()` in `_open_handle`'s `finally` had the identical unshielded-await-under-cancellation gap as the unlink I fixed earlier. Shielded it the same way in 2a9e2d3.

champ18ion:
Addressing review feedback:

- CodeRabbit's file-handle context-manager nitpick: fixed in 6ee7be0, `put_stream`/`get_stream` now use an `@asynccontextmanager`.
- Copilot flagged (as a low-confidence suppressed comment, so replying here since there's no thread for it) that the partial-blob unlink on cancellation wasn't shielded: fixed in 02450d5.
- Copilot's follow-up inline comment caught that I'd missed the identical gap in `_open_handle`'s own `handle.close()`: fixed in 2a9e2d3, replied inline.

All three fixes have test coverage (including a new cancellation-simulation test) and pass lint/typecheck/unit tests locally.

src/gateway/api/routes/files.py
```diff
@@ -230,11 +233,13 @@ async def get_file_content(
...
-    data = await read_file_bytes(file_store, record)
-    return Response(
-        content=data,
+    return StreamingResponse(
+        file_store.get_stream(record.storage_ref),
         media_type=record.mime_type,
-        headers={"Content-Disposition": _content_disposition(record.filename)},
+        headers={
+            "Content-Disposition": _content_disposition(record.filename),
+            "Content-Length": str(record.bytes),
```
khaledosman: `Content-Length` is set from `record.bytes` (DB) while the body comes from `get_stream` (disk). If the two ever diverge (corruption, partial write, tampering), the response is malformed: the client hangs on missing bytes or sees a length mismatch. The old full-buffer path derived length from the actual bytes. Either `stat` the file for the header, or drop it and let the response be chunked.

src/gateway/api/routes/files.py
```diff
@@ -230,11 +233,13 @@ async def get_file_content(
     if record is None:
         raise HTTPException(status_code=status.HTTP_404_NOT_FOUND, detail="File not found")
 
-    data = await read_file_bytes(file_store, record)
-    return Response(
-        content=data,
+    return StreamingResponse(
```
khaledosman, line 283: Moving the blob read into `StreamingResponse` means a missing/unreadable blob now raises inside the generator *after* `200 OK` + headers are flushed → truncated body on a 200, where the pre-PR code returned a 500 before headers. Rare (soft-delete/expiry hit 404 first), but it's a silent-corruption-shaped failure mode — worth a line in the PR description acknowledging it.

src/gateway/api/routes/files.py
```diff
@@ -127,12 +128,14 @@ async def create_file(
 
     user_id = _resolve_user(auth_result, user, config)
 
-    data = await _read_capped(file, config.files_max_bytes)
-    if not data:
-        raise HTTPException(status_code=status.HTTP_400_BAD_REQUEST, detail="Uploaded file is empty")
-
     file_id = f"file-{uuid.uuid4().hex}"
-    storage_ref = await file_store.put(file_id, data)
+    storage_ref, size = await file_store.put_stream(file_id, _capped_chunks(file, config.files_max_bytes))
```
khaledosman, line 165: Empty-upload path now creates a zero-byte file then deletes it (and leaves the `xx/` shard dir behind), whereas the pre-PR route rejected empty before any disk write. Minor extra I/O + an empty dir per rejected upload.

tests/unit/test_file_store.py
```diff
@@ -10,13 +12,85 @@
...
+
+@pytest.mark.asyncio
+async def test_put_stream_cleans_up_partial_file_on_failure(tmp_path: Path) -> None:
+    store = LocalDirFileStore(str(tmp_path))
+
+    async def _failing_chunks() -> AsyncIterator[bytes]:
+        yield b"partial data that should not survive"
+        raise RuntimeError("simulated upstream failure mid-stream")
+
+    with pytest.raises(RuntimeError, match="simulated upstream failure"):
```
khaledosman, line 63: The partial-cleanup tests use `RuntimeError`/cancellation, but the real production trigger is `_capped_chunks` raising `HTTPException(413)` mid-stream. Add a case asserting the `HTTPException` path through `put_stream` also leaves no blob — that's the actual route interaction and currently has no direct coverage.

src/gateway/services/file_store.py
```diff
@@ -82,6 +128,45 @@ async def get(self, storage_ref: str) -> bytes:
...
+            path.unlink(missing_ok=True)
+
+        await asyncio.to_thread(_mkparent)
+        total = 0
+        try:
+            async with _open_handle(path, "wb") as handle:
+                async for chunk in chunks:
+                    total += len(chunk)
+                    await asyncio.to_thread(handle.write, chunk)
+        except BaseException:
```
khaledosman, line 151: Good and easy to regress: `try/except BaseException` wraps the `async with`, so the handle is closed before `_unlink_partial` runs (required on Windows), and the shield keeps cleanup from being cancelled. Keep this ordering.

src/gateway/services/file_store.py
```diff
@@ -31,6 +59,24 @@ async def get(self, storage_ref: str) -> bytes:
...
+        """Persist ``chunks`` for ``file_id`` without buffering them fully in memory.
+
+        Returns the opaque storage ref and the total byte count written. Callers
+        that need to cap the upload size (e.g. the ``/v1/files`` route) must
+        enforce that themselves while producing ``chunks``; this is a pure
+        storage primitive and does not know about HTTP limits.
+        """
+        ...
+
+    def get_stream(self, storage_ref: str) -> AsyncIterator[bytes]:
```
khaledosman: Correct subtlety: declaring `get_stream` as non-`async def` returning `AsyncIterator` matches the async-generator impl (an async generator function's type isn't a coroutine). Don't "fix" this to `async def`.

src/gateway/api/routes/files.py
champ18ion: oh good catch, you're right, that could go really wrong if they ever drift apart. simplest fix is just dropping the header entirely and letting it be chunked, done in 05e51ce

src/gateway/api/routes/files.py
champ18ion: yeah that's a real gap, thanks for catching it. added a small _prime helper that reads the first chunk before we even build the StreamingResponse, so a missing/corrupted blob now fails clean with a 500 instead of truncating a 200. won't catch a failure later mid-stream (kinda unavoidable with real streaming) but at least the common case is covered now

src/gateway/api/routes/files.py
champ18ion: lol yeah that's a bit wasteful, my bad. made delete() try to rmdir the parent shard dir if it just emptied it out, best effort so it doesn't blow up if other files are still in there

tests/unit/test_file_store.py
champ18ion: fair, honestly should've had this from the start, the RuntimeError/cancellation tests don't actually mirror what really happens in prod. added one that raises a real HTTPException(413) mid-stream like _capped_chunks does and checks cleanup still holds

src/gateway/services/file_store.py
champ18ion: thanks for calling this out explicitly, good to know it actually matters on windows too. will keep the ordering in mind for the S3 backend PR

src/gateway/services/file_store.py
champ18ion: haha appreciate you double checking, was a little worried someone would 'fix' it back to async def down the line without realizing why it's not. glad it reads clearly enough as intentional

src/gateway/api/routes/files.py
```diff
@@ -81,8 +83,31 @@ async def _read_capped(file: UploadFile, max_bytes: int) -> bytes:
...
+    except StopAsyncIteration:
+        first = None
+
+    async def _rest() -> AsyncIterator[bytes]:
+        if first is not None:
+            yield first
+            async for chunk in chunks:
+                yield chunk
+
+    return _rest()
```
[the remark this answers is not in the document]
champ18ion: good catch, you're right that would've leaked fds under real traffic. wrapped _rest in try/finally calling chunks.aclose(), and had to widen get_stream's return type from AsyncIterator to AsyncGenerator since aclose() isn't on the narrower protocol. added a test that simulates an early disconnect and checks the inner source actually gets closed, not just that it works when nothing goes wrong

khaledosman:
Thanks for the contribution! good to go from my side, leaving the final review/merge duties for @njbrake

njbrake (approved):

Approving. This bounds peak RAM on the `/v1/files` path by streaming uploads and downloads chunk-by-chunk instead of buffering whole blobs, which is the memory pressure tracked in #156.

What I verified locally on the PR head:

- `pytest tests/unit/test_file_store.py tests/unit/test_files_route_streaming.py` passes (19 tests).
- `ruff check` and `mypy` on the two changed source files are clean.
- `generate_openapi.py --check` reports the spec is up to date.
- Walked the risky paths: no fd leak on a first-chunk prime failure (`get_stream`'s own `_open_handle` closes the handle during unwind), empty and legacy zero-byte blobs return a clean empty 200, and a 413 mid-stream unlinks the partial blob and re-raises with no orphan left behind.

Every substantive bot and maintainer finding was addressed with a matching test: the async context manager for file handles, shielding the partial-blob cleanup and the handle close from cancellation, and closing the primed stream on early client disconnect.

Two optional, non-blocking notes for later, not for this PR:

1. `delete` now `rmdir`s an emptied shard directory, which opens a microsecond-wide race with a concurrent upload into the same shard (guarded, best-effort, and called out in the comment). Worth a mental note for the S3 follow-up.
2. The cancellation test raises `CancelledError` synchronously rather than through a real task `cancel()`, so it covers the cleanup path but not the precise second-cancel-during-unwind case that `shield` defends against. The shield logic is correct regardless.

Nice, careful work, and thorough test coverage.

Issue #331, opened 2026-07-21
Surface OpenAI's actionable error when tools + reasoning_effort are unsupported on /v1/chat/completions (e.g. gpt-5.6-sol)


tbille:
### Problem
Some newer OpenAI reasoning models (e.g. `gpt-5.6-sol`, intended primarily for the Responses API) reject requests combining `tools` with an explicit non-`'none'` `reasoning_effort` on `/v1/chat/completions`:
```
Error code: 400 - {'error': {'message': "Function tools with reasoning_effort are not
supported for gpt-5.6-sol in /v1/chat/completions. To use function tools, use
/v1/responses or set reasoning_effort to 'none'.", 'type': 'invalid_request_error',
'param': 'reasoning_effort', 'code': None}}
```
Today, `classify_provider_error()` (`src/gateway/api/routes/_pipeline.py`) maps every upstream 400/422 to the same generic, fixed `PROVIDER_BAD_REQUEST_DETAIL` string, so callers never see OpenAI's actual, actionable remedy.

Note this isn't limited to callers who pass `tools` explicitly: the gateway's tool-loop also injects real function-tool schemas from MCP servers / sandbox / web-search backends whenever `mcp_servers`, `mcp_server_ids`, `otari_code_execution`, or `otari_web_search` are used, even if the caller's own `tools` field was empty — so this can surface without the caller ever writing a `tools` array.

### Investigation notes
- Filed and got a definitive answer from any-llm upstream ([mozilla-ai/any-llm#1189](https://github.com/mozilla-ai/any-llm/issues/1189), closed): the SDK cannot detect this ahead of time (OpenAI exposes no capability metadata for it), so this isn't fixable at the any-llm layer without an unmaintainable per-model version matrix.
- litellm hit the identical bug ([BerriAI/litellm#33221](https://github.com/BerriAI/litellm/issues/33221)) and fixed it by auto-bridging affected requests to `/v1/responses` under the hood, gated by a hardcoded `gpt-5.6+` version classifier. That's a bigger, format-bridging change with an ongoing model-matrix maintenance cost; not proposed here.
- Portkey does plain param passthrough with no special-casing (same exposure as any-llm's baseline).
[...]

PR #381, opened 2026-07-23
fix(hybrid): recognize SDK-wrapped timeout/connection errors for fallback

tbille:
## Description

Hybrid mode's multi-attempt fallback (and the single-attempt terminal error
mapping) misclassified real provider timeouts and connection failures as
`unknown`/non-retryable, so a request would fail outright instead of falling
through to the next attempt in the routing policy.

`_classify_upstream_error` (`_platform.py`) and `classify_provider_error`
(`_pipeline.py`) only recognized raw `httpx.TimeoutException`/
`httpx.NetworkError` and exceptions carrying a `status_code`. In production,
any-llm calls the OpenAI/Anthropic SDKs directly rather than httpx. Both SDKs
catch `httpx.TimeoutException`/network errors internally and re-raise as their
own `APITimeoutError`/`APIConnectionError`, neither of which is an instance of
any httpx exception, and neither carries `status_code`/`response`. So a real
"provider timeout" or "provider unreachable" (DNS failure, connection refused,
TLS error) surfacing through OpenAI/Anthropic-backed providers (the majority
of any-llm's ~50 providers reuse `BaseOpenAIProvider`/`BaseAnthropicProvider`)
was misclassified as an unrecognized error and treated as terminal,
contradicting the documented fallback semantics
(`docs/hybrid-mode-protocol.md`).

Verified empirically:
```python
>>> isinstance(openai.APIConnectionError(request=req), httpx.NetworkError)
False
>>> hasattr(openai.APIConnectionError(request=req), "status_code")
False
```

Changes:
- `src/gateway/api/routes/_platform.py`: new `upstream_exception_shape()`
  helper recognizes `openai`/`anthropic` `APITimeoutError`/`APIConnectionError`
  explicitly (timeout checked before connection, since `APITimeoutError`
  subclasses `APIConnectionError` in both SDKs), plus a conservative
  duck-typed fallback (by exception class name, only when no status code is
  present) for the other any-llm provider SDKs (cohere, mistral, groq,
  bedrock, ...), mirroring the heuristic
  `any_llm.utils.exception_handler.convert_exception` already uses
  internally.
- `src/gateway/api/routes/_pipeline.py`: `classify_provider_error` and
  `raise_all_streaming_attempts_failed` now use the same shared helper, so a
  single-attempt/all-exhausted timeout gets the correct 504 instead of a
  generic 502. Removed the now-redundant `_upstream_status_code`.
- Consolidated the previously-duplicated status/timeout/connection detection
  into one shared helper so the non-streaming and streaming paths (and the
  terminal error mapping) can't drift again — this duplication is how the gap
  went unnoticed.
- `docs/hybrid-mode-protocol.md`: updated the `error_class` table to describe
  the broadened detection.
- Tests: added regression cases raising the real
  `openai.APIConnectionError`/`APITimeoutError` and Anthropic equivalents
  (unit, in both classifiers), a duck-typed-fallback case, and an integration
  test driving `run_platform_attempts` end-to-end with an
  `openai.APIConnectionError` on attempt 1 falling through to a successful
  attempt 2.

Out of scope: bespoke-SDK provider exceptions (e.g. `botocore`/Bedrock's
`EndpointConnectionError`) aren't enumerated explicitly, only covered by the
duck-typed class-name fallback. Standalone mode has no cross-provider
fallback by design (unaffected either way).

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

None.

## Checklist

- [x] If the API contract changed, I regenerated the OpenAPI spec (`uv run python scripts/generate_openapi.py`). (No contract change; verified with `--check`.)

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:** OpenCode (Claude Sonnet 4.5)

Root cause found by tracing the OpenAI/Anthropic SDK's `_base_client.py`
exception-wrapping behavior and empirically confirming the wrapped exceptions
don't satisfy the classifier's `isinstance`/`status_code` checks.

**NOTE:**

## Summary

- Fixed hybrid-mode fallback so the gateway correctly recognizes provider timeouts and connection failures even when they’re wrapped by SDK-specific exception types.
- Centralized the upstream error “shape” detection and reused it consistently for both streaming and non-streaming retry paths, ensuring fallback behavior matches the right failure class.
- Improved response mapping so terminal timeouts now return HTTP 504 (instead of a generic failure code), including for the streaming “all attempts timed out” case.
- Updated documentation and added regression tests to cover SDK-wrapped connection errors, streaming timeout behavior, unified any-llm-style exception wrapping, and safe unwrapping.

## Technical notes

- Introduced a shared helper (`upstream_exception_shape`) that classifies upstream exceptions into timeout/connection categories using SDK types, conservative name-based heuristics, and bounded `original_exception` unwrapping with cycle protection.
- Updated pipeline/route logic to rely on this shared classification for retryability and for the final “all attempts” streaming decision.

### DIFF
--- a/docs/hybrid-mode-protocol.md
+++ b/docs/hybrid-mode-protocol.md
@@ -311,10 +311,21 @@ resolve response). The platform is responsible for correlating them.
 
 | Tag | Cause |
 |---|---|
-| `timeout` | `httpx.TimeoutException`, `asyncio.TimeoutError`, `TimeoutError` |
-| `conn_err` | `httpx.NetworkError` |
+| `timeout` | `httpx.TimeoutException`, `asyncio.TimeoutError`, `TimeoutError`, or the OpenAI/Anthropic SDKs' own `APITimeoutError` |
+| `conn_err` | `httpx.NetworkError`, or the OpenAI/Anthropic SDKs' own `APIConnectionError` |
 | `http_<code>` | Provider returned an HTTP status code (e.g. `http_429`, `http_401`) |
 | `unknown` | Any other exception class |
+
+Otari calls the OpenAI/Anthropic SDKs directly rather than httpx, and both SDKs
+catch `httpx.TimeoutException`/network errors internally and re-raise as their
+own `APITimeoutError`/`APIConnectionError`; neither is an instance of any
+`httpx` exception, and neither carries `status_code`/`response`. Otari
+recognizes these wrapped types explicitly (covering the majority of providers,
+which reuse the OpenAI/Anthropic base provider classes), plus a conservative
+class-name-based fallback (`*TimeoutError` / `*ConnectionError`, only when no
+status code is present) for the other any-llm provider SDKs, so a real
+"provider unreachable" or provider-side timeout still falls through to the
+next attempt instead of being misclassified as `unknown`.
 
 The field is **omitted entirely** when Otari can't map the failure to an
 exception class; this happens with mid-stream errors surfaced via the
--- a/docs/hybrid-mode-protocol.md
+++ b/docs/hybrid-mode-protocol.md
@@ -325,7 +325,13 @@ which reuse the OpenAI/Anthropic base provider classes), plus a conservative
 class-name-based fallback (`*TimeoutError` / `*ConnectionError`, only when no
 status code is present) for the other any-llm provider SDKs, so a real
 "provider unreachable" or provider-side timeout still falls through to the
-next attempt instead of being misclassified as `unknown`.
+next attempt instead of being misclassified as `unknown`. If any-llm's unified
+exceptions (`ANY_LLM_UNIFIED_EXCEPTIONS=1`) are enabled in the future, the same
+detection still applies through `original_exception`: any-llm re-wraps a raw
+SDK timeout/connection error into a generic `any_llm.exceptions.ProviderError`
+that carries neither a recognizable class name nor a status code, but keeps
+the original SDK exception on `.original_exception`, which Otari unwraps and
+re-classifies.
 
 The field is **omitted entirely** when Otari can't map the failure to an
 exception class; this happens with mid-stream errors surfaced via the
--- a/src/gateway/api/routes/_pipeline.py
+++ b/src/gateway/api/routes/_pipeline.py
@@ -38,7 +38,6 @@ from enum import Enum, auto
 from typing import Any, Generic, Literal, NamedTuple, NoReturn, Protocol, TypeVar
 from urllib.parse import ParseResult, urlparse
 
-import httpx
 from any_llm import LLMProvider
 from any_llm.exceptions import AnyLLMError
 from any_llm.types.completion import (
@@ -69,6 +68,7 @@ from gateway.api.routes._platform import (
     _resolve_platform_mcp_servers,
     _resolve_platform_web_search,
     run_platform_attempts,
+    upstream_exception_shape,
 )
 from gateway.api.routes._platform import (
     default_attempt_kwargs as default_attempt_kwargs,  # explicit re-export for the route modules
@@ -192,20 +192,6 @@ class ProviderErrorMapping(NamedTuple):
     detail: str
 
 
-def _upstream_status_code(exc: BaseException) -> int | None:
-    """Pull an HTTP status off an upstream exception, if it carries one.
-
-    any-llm surfaces provider HTTP errors either as ``exc.status_code`` or via an
-    attached ``exc.response.status_code``; everything else has neither.
-    """
-    status_code = getattr(exc, "status_code", None)
-    if status_code is None:
-        response = getattr(exc, "response", None)
-        if response is not None:
-            status_code = getattr(response, "status_code", None)
-    return status_code if isinstance(status_code, int) else None
-
-
 def _upstream_error_param(exc: BaseException) -> str | None:
     """Pull the offending request field off an upstream exception, if it names one.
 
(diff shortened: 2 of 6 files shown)


src/gateway/api/routes/_pipeline.py
```diff
@@ -1841,7 +1832,8 @@ def raise_all_streaming_attempts_failed(
     invalid_request = mapping is not None and mapping.status_code == status.HTTP_400_BAD_REQUEST
     if invalid_request or len(route.attempts) <= 1:
         raise adapter.provider_error(exc) from exc
-    if isinstance(exc, (asyncio.TimeoutError, TimeoutError, httpx.TimeoutException)):
+    kind, _ = upstream_exception_shape(exc)
+    if kind == "timeout":
```
khaledosman, line 1894: This streaming terminal-504 path isn't covered by the new tests (they hit `classify_provider_error` and `_classify_upstream_error` only). Add a case asserting `raise_all_streaming_attempts_failed` returns 504 for an `openai.APITimeoutError` — that's the exact drift this PR is closing.

src/gateway/api/routes/_platform.py
```diff
@@ -506,23 +510,81 @@ def _parse_resolve_payload(payload: dict[str, Any]) -> ResolvedRoute:
     )
 
 
-def _classify_upstream_error(exc: BaseException) -> tuple[bool, str]:
-    """Classify an upstream provider error.
-
-    Returns ``(retryable, error_class)``. ``error_class`` is a short string used
-    for logging and reporting back to the platform. Streaming-only failures still
-    pass through this classifier; the caller decides whether to actually retry.
+def upstream_exception_shape(exc: BaseException) -> tuple[str | None, int | None]:
```
khaledosman: Return `tuple[Literal["timeout", "conn_err"] | None, int | None]` so mypy flags a typo at the `== "timeout"` comparison sites instead of silently never matching.

khaledosman (approved):

njbrake:

Thanks for the careful writeup, the diagnosis is right and I confirmed the load-bearing pieces locally: on the pinned SDKs `APITimeoutError` subclasses `APIConnectionError` (so the timeout-first ordering is required), neither wrapped type is an httpx instance or carries `status_code`, and any-llm 1.22.1 re-raises the raw SDK exception by default, so these really do reach the classifier in the shape you describe. Consolidating the two paths onto one helper is the right call.

One thing I want to resolve before merging, because it ties into #331 and any-llm#1190: this adds a dependency on raw SDK exceptions that is even tighter than the one I flagged there. #331 reads `.status_code`/`.param` off the exception; this does `isinstance` against the `openai.*` / `anthropic.*` types directly. any-llm already emits a deprecation warning that provider exceptions "will be converted to unified any-llm exceptions in a future version," and the plan in #331 is to enable `ANY_LLM_UNIFIED_EXCEPTIONS=1` once #1190 lands. When we make that flip:

- the `isinstance` checks against the SDK types go dead (those types stop arriving), and
- a converted timeout becomes a generic `any_llm.exceptions.ProviderError`, whose class name ends in neither `TimeoutError` nor `ConnectionError`, so the duck-typed fallback misses it too and it lands back in `unknown`/non-retryable, reintroducing the exact bug this PR fixes.

The cheap way to make this survive the flip is the same `original_exception` unwrap I described on #331 (works today and after the flip): have `upstream_exception_shape` also inspect `getattr(exc, "original_exception", None)` so the raw SDK type is still reachable once it is wrapped. That keeps this consistent with where #331 is heading instead of adding a second raw-exception coupling we have to unwind later.

If you would rather keep this PR narrow, I am fine landing it as-is on the condition that we add it to the #331 migration checklist so it does not silently regress when unified exceptions become the default. Either is fine with me; I just do not want the dependency to be implicit.

Minor, non-blocking, and I think @khaledosman already noted the first two inline: (1) the `Literal["timeout", "conn_err"] | None` return type so mypy catches typos at the comparison sites, and (2) a direct test for the streaming terminal path (`raise_all_streaming_attempts_failed` returning 504 on a wrapped `APITimeoutError` over a multi-attempt route), which is the one branch the new tests do not cover. I have both staged locally if it is easier for me to push them onto the branch.

docs/hybrid-mode-protocol.md
```diff
@@ -311,11 +311,22 @@ resolve response). The platform is responsible for correlating them.
...
-| `timeout` | `httpx.TimeoutException`, `asyncio.TimeoutError`, `TimeoutError` |
-| `conn_err` | `httpx.NetworkError` |
+| `timeout` | `httpx.TimeoutException`, `asyncio.TimeoutError`, `TimeoutError`, or the OpenAI/Anthropic SDKs' own `APITimeoutError` |
+| `conn_err` | `httpx.NetworkError`, or the OpenAI/Anthropic SDKs' own `APIConnectionError` |
 | `http_<code>` | Provider returned an HTTP status code (e.g. `http_429`, `http_401`) |
 | `unknown` | Any other exception class |
 
+Otari calls the OpenAI/Anthropic SDKs directly rather than httpx, and both SDKs
+catch `httpx.TimeoutException`/network errors internally and re-raise as their
+own `APITimeoutError`/`APIConnectionError` — neither is an instance of any
```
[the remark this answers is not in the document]
tbille: Fixed; replaced the em dash with a semicolon in that paragraph. Thanks for the catch.

src/gateway/api/routes/_pipeline.py
tbille: Good catch, added `test_hybrid_mode_streaming_returns_504_when_all_attempts_time_out` (tests/integration/test_hybrid_mode_chat.py): both attempts raise `openai.APITimeoutError` on a multi-attempt route, asserting the client sees 504 / "All upstream providers timed out" rather than the generic 502. That's the branch that was previously uncovered.

src/gateway/api/routes/_platform.py
tbille: Done. `upstream_exception_shape` now returns `tuple[Literal["timeout", "conn_err"] | None, int | None]` (`UpstreamErrorKind = Literal["timeout", "conn_err"]`), so mypy flags a typo at any `kind == "..."` comparison site instead of it silently never matching.

tbille:
Thanks for the thorough check, and for confirming the timeout-before-connection ordering and the raw-passthrough assumption hold on the pinned SDKs.

Went with the unwrap now rather than deferring to #331. Implemented as: `upstream_exception_shape` walks `.original_exception` (bounded by an `id()`-based seen-set so a pathological self-referential chain can't loop) whenever the current exception carries no classifiable signal of its own, before falling back to `unknown`. Also folded in your ordering point implicitly: the unwrap re-runs the full step 1-4 ladder (stdlib/httpx types, SDK types, status code, duck-typed class-name fallback) on the unwrapped exception, so it's not a narrower check than the top-level one.

Added a regression test wrapping `openai.APITimeoutError`/`APIConnectionError` (and the Anthropic equivalents) in a fake `AnyLLMError`-shaped object exposing only `original_exception`, asserting `_classify_upstream_error` still returns retryable, plus one for `classify_provider_error` mapping to 504 through the same path, plus a defensive cycle-termination test. Also picked up @khaledosman's two minor asks (`Literal["timeout", "conn_err"]` return type, and the missing streaming-504 test for `raise_all_streaming_attempts_failed`).

Left a cross-reference on #331 so the unified-exceptions migration work there has a pointer to this precedent (the `original_exception` unwrap pattern) rather than re-discovering it.

khaledosman (approved):
I already pre-approved, but it seems coderabbit dismissed it with the re-request, so im blindly re-approving

PR #382, opened 2026-07-23
fix(dashboard): eliminate provider picker dropdown lag

### LINKED ISSUE #365: Dashboard: adding a provider the provider selector dropdown has a lag
I'm guessing that the dashboard is making a request to a live endpoint or something that has some latency to compile all the info: please investigate a more robust and faster way to gather the list of providers.

njbrake:
## Description

The add-provider picker in the dashboard fetches `GET /v1/providers/catalog`. Its handler walked every `any-llm` provider class to build the catalog, which import-loads ~50 provider SDK modules and cost a few seconds on the first open per process (measured 1 to 3 seconds). That was the lag reported in #365.

This PR splits the catalog so the list never imports a provider SDK:

1. `GET /v1/providers/catalog` now returns `id` + display name only, built from the any-llm registry plus the bundled genai-prices names. No provider module is imported, so the list is about 2ms and the picker opens instantly. Display names are unchanged (verified identical to the old path for all 50 providers).
2. `GET /v1/providers/catalog/{provider_id}` is new: it returns the autofill hints (credential env var, default endpoint, whether a key is required, whether that env var is already set on the server) for a single provider, importing only that one SDK. Unknown id returns 404.

The dashboard fetches the detail lazily (`useProviderDetail`) for just the provider the operator selects, so nobody pays to import SDKs they never pick. The OpenAPI spec is regenerated and the dashboard bundle rebuilt.

Note: this changes the `/v1/providers/catalog` response shape (full objects to id + name). The only consumer is this dashboard, updated in the same commit.

## PR Type

- [x] Bug Fix

## Relevant issues

Fixes #365

## Checklist

Local checks: ruff and mypy clean, 877 backend unit tests pass, OpenAPI `--check` up to date, frontend typecheck clean and 197 dashboard tests pass, dashboard bundle rebuilt. An end-to-end TestClient smoke confirmed the list route returns 200 in about 2ms importing zero provider SDKs, the detail route returns hints (404 for unknown ids), and both are master-key gated. The provider-catalog integration tests require Docker/Postgres (Testcontainers), unavailable in my sandbox; they run in CI.

New / updated tests:
- `tests/unit/test_provider_metadata.py`: the summary list imports no provider SDK (and leaves any-llm's per-provider submodules unimported); per-provider detail carries the autofill hints and a live `env_key_present`; unknown id returns None.
- `tests/integration/test_provider_credentials_api.py`: list returns id+name only; detail returns hints, 404s on unknown ids, and is master-key gated.
- `web/src/pages/ProvidersPage.test.tsx`: the form fetches provider hints lazily, only after a provider is selected.

## AI Usage

Diagnosis, implementation, and tests were done by Claude Code via back-and-forth with @njbrake; the reasoning and decisions are his.

## Summary

- Reduced “add provider” picker loading time by making the provider catalog initially return only lightweight entries (provider id + display name), so the UI can render immediately without loading every provider’s SDK.
- Added a new on-demand lookup for a selected provider’s catalog details (including API-key requirements and whether the needed environment variables are present), returning 404 for unknown providers.
- Updated the dashboard to fetch provider details only after a user selects a provider.
- Regenerated the OpenAPI spec and rebuilt the dashboard bundle, and updated/added unit, integration, and frontend tests to cover the new lazy-loading flow and 404 behavior.

## Benefits

- Cuts initial provider-picker load time from ~1–3 seconds down to ~2ms by avoiding unnecessary SDK imports during the initial catalog request.
- Improves responsiveness and keeps provider-specific autofill info accurate by loading it only when needed.


src/gateway/services/provider_metadata_service.py
```diff
@@ -188,7 +196,35 @@ def list_known_providers() -> list[KnownProvider]:
...
+    :func:`_known_providers_static`). Only ``env_key_present`` is resolved live
+    against the current environment, so a key set after the cache warmed is still
+    reflected here.
+    """
+    return [
+        replace(
+            provider,
+            env_key_present=provider.env_key is not None and bool((os.getenv(provider.env_key) or "").strip()),
+        )
+        for provider in _known_providers_static()
```
khaledosman: The warm is best-effort, and the fallback path blocks the event loop. `provider_catalog` (`api/routes/providers.py:154`) is `async def` and consumes `list_known_providers()` synchronously, so if the catalog is hit *before* the background warm finishes (the 1-3s window right after startup), this `_known_providers_static()` call runs the ~50-module import inline on the event loop thread and stalls **all** requests for its full duration, not just the caller. That's the exact failure #365 is about, just moved to a smaller window rather than closed.

To make the guarantee real, offload the fallback too, e.g. in the route:
```python
return [_to_known_schema(p) for p in await asyncio.to_thread(list_known_providers)]
```
or make the handler `def` so FastAPI runs it in the threadpool. If you'd rather leave it, a comment here acknowledging the window would help the next reader.

src/gateway/services/provider_metadata_service.py
```diff
@@ -188,7 +196,35 @@ def list_known_providers() -> list[KnownProvider]:
...
+    ]
+
+
+def warm_known_providers_cache() -> None:
```
khaledosman: Minor: `functools.cache` releases its internal lock during the wrapped call, so a request landing during the warm can trigger a second concurrent execution of the import walk. Python's import lock keeps this safe, just redundant work. Not worth a code change; noting for completeness.

khaledosman (commented):
Clean diagnosis and well-documented. Splitting the expensive SDK-import walk (`_known_providers_static`, memoized) from the live `env_key_present` overlay is the right shape, and `test_env_key_present_is_live_after_cache_warm` guards exactly the failure mode memoization could introduce (a frozen stale key answer). Two notes inline; the first is the one I'd want addressed or explicitly acknowledged before merge.

khaledosman (approved):

PR #383, opened 2026-07-23
fix(dashboard): align input boxes on the tools & guardrails page

### LINKED ISSUE #355: Dashboard: clean up input box alignment in the tools/guardrails page
Right now the tools and guardrails page is functional but the spacing and alignment of the input boxes/settings is a bit all over the place. Please use design skills and refine this page.

njbrake:
## Description

The tools & guardrails page laid each field out as a flex row that right-aligned its control cluster. Three things made the boxes and buttons look uneven down a card:

- The **Test** button on URL rows pushed **Save** inward, so Save did not share a right edge with the other rows.
- Inputs used mismatched widths (`w-64` for text/URL, `w-28` for numbers).
- Per-row status/error text had no fixed column.

This renders every field as a grid row with three fixed-width tracks: `label | input (16rem) | actions`. Because the input and action tracks are the same width on every row, the input boxes and the Save buttons line up in columns regardless of a row's type, help text, or whether it also carries a Test button. The numeric input stays narrow but is pinned to the input column's right edge, so its edge (and its Save button) still align with the text inputs; the Test button hangs off the right of Save on URL rows. Below the `sm` breakpoint the grid collapses to a single stacked column.

Verified visually in a real browser against the built CSS: input left edges, input right edges (including the right-aligned numeric input), and the Save column all line up across every row and both cards; the mobile layout stacks cleanly.

## PR Type

- [x] Bug Fix

## Relevant issues

Fixes #355

## Checklist

- [x] I have added or updated tests that cover my change (`web/src/pages/ToolsGuardrailsPage.test.tsx`; the backend `tests/unit`/`tests/integration` suites are unaffected by this dashboard-only change).
- [x] I ran the Definition of Done checks locally (`npm --prefix web run typecheck`, `npm --prefix web test` (197 passing), `npm --prefix web run build`, and re-committed the bundle).
- [x] If the API contract changed, I regenerated the OpenAPI spec (n/a: no API change).

## AI Usage

Layout refactor and test drafted by Claude via back-and-forth with @njbrake; the reasoning and decisions are his.

## Summary

- Improved the Tools & Guardrails page layout by aligning labels, inputs, buttons, and messages consistently.
- Added responsive behavior so fields stack on smaller screens.
- Added tests covering alignment across different field types.
- Regenerated dashboard assets and updated bundled references.
- No API or backend changes.


web/src/pages/ToolsGuardrailsPage.tsx
```diff
@@ -74,9 +74,25 @@ function SaveToast({ message }: { message: string | null }) {
...
+// extra help line, or a narrower numeric input. Below `sm` the grid collapses
+// to a single column and the pieces stack.
+const ROW_CLASS = "grid gap-x-4 gap-y-1.5 py-4 sm:grid-cols-[minmax(0,1fr)_16rem_10rem] sm:items-start";
+const INPUT_CELL = `w-full sm:col-start-2 ${INPUT_CLASS}`;
+const ACTIONS_CELL = "flex items-center gap-2 sm:col-start-3";
+const MESSAGE_CELL = "flex flex-col gap-1 sm:col-span-2 sm:col-start-2";
+
+function SaveError({ message }: { message?: string }) {
+  if (!message) return null;
+  return <span className="break-words text-xs text-red-700">{message}</span>;
```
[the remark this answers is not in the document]
njbrake: _Note: this reply was drafted by Claude via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

Thanks, but keeping these as raw Tailwind status classes here. The dashboard's frontend-standards intentionally carve out one exception to the `--otari-*` tokens: semantic status surfaces use the raw `red` (error), `amber` (warning), and `emerald` (success) palette, matching `ErrorBanner`, `InfoBanner`, and the `StatCard` status pills. Moving these inline error and reachability messages onto `--otari-*` tokens would break from that convention, and there is no brand token for an error or success hue. These classes are also pre-existing; this PR only relocated them while regridding the row layout for #355.

web/src/pages/ToolsGuardrailsPage.tsx
```diff
@@ -74,9 +74,25 @@ function SaveToast({ message }: { message: string | null }) {
 const INPUT_CLASS =
   "rounded-md border border-[var(--otari-line)] bg-[var(--otari-surface)] px-2 py-1 text-sm focus:border-[var(--otari-brand)] focus:outline-none disabled:opacity-50";
 
+// Every field renders as one grid row with three fixed-width tracks:
+// label | input (16rem) | actions (10rem). Because the input and action tracks
+// have the same width on every row, the boxes and the Save buttons line up in
+// columns down a card regardless of which rows also carry a Test button, an
+// extra help line, or a narrower numeric input. Below `sm` the grid collapses
+// to a single column and the pieces stack.
+const ROW_CLASS = "grid gap-x-4 gap-y-1.5 py-4 sm:grid-cols-[minmax(0,1fr)_16rem_10rem] sm:items-start";
```
khaledosman, line 88: The actions track is `10rem` but URL rows carry both Save + Test, so this relies on Test overflowing past the track's right edge (grid tracks don't clip). You've verified it visually, but it's a fragile assumption: a longer button label or i18n would push Test further off-edge with no reflow. A one-line note here that the actions track intentionally sizes for the Save-aligned column (Test overflows by design) would save the next reader the puzzle.

khaledosman (commented):
Pure layout refactor to a 3-track grid; no logic change and the grid auto-placement is sound (label→col1, input→col2, actions→col3, message spans col2-3 onto row 2). No correctness issues. One inline note, plus two things worth a quick check:

- **Bundle churn:** ~13 chunks re-hashed for a change that touches only `ToolsGuardrailsPage.tsx`. That's the expected Vite cascade when the shared `index`/CSS hash changes, but CI's stale-bundle check verifies the bundle matches source, not that the diff is minimal, so it won't catch an unrelated dep shift riding along. Worth confirming `package-lock.json` is unchanged on this branch.
- **Message-cell asymmetry (minor):** `UrlRow` always renders `MESSAGE_CELL` (for the aria-live reachability status) while Text/Number/Bool render it only when `saveError` is set. Justified, but not obvious from the diff.

khaledosman (approved):

web/src/pages/ToolsGuardrailsPage.tsx
njbrake: _Note: this reply was drafted by Claude via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

Good call. Added a comment at the `ROW_CLASS` definition (08b9b6cf) spelling out that the actions track is sized for the Save button so Save stays column-aligned, and that Test is intentionally left to overflow the track's right edge rather than widening it (which would push Save inward).

On the two review-body notes: `web/package-lock.json` is unchanged on this branch, so the ~13 re-hashed chunks are purely the Vite cascade from the shared index/CSS hash, with no dependency shift riding along. And `UrlRow` always renders the message cell because it holds the aria-live reachability region, which needs a stable node to announce into; the other row types render it only on error by design.

PR #384, opened 2026-07-23
fix(dashboard): persist sign-in with an HttpOnly session cookie

### LINKED ISSUE #338: Dashboard: persist auth so operators don't re-paste the master key on every open

## Problem

The dashboard makes you re-paste the master key almost every time you open it: a new tab, a closed tab, or a browser restart all lose the session. That is annoying enough that it discourages using the dashboard.

## How it works today

- On sign-in the raw master key is stored in `sessionStorage` under `otari.dashboard.masterKey` (`web/src/auth/AuthContext.tsx:7`, `:62`) and seeded into the API client (`setMasterKey`).
- Every management request sends it as `Authorization: Bearer <masterKey>` (`web/src/api/client.ts:74`).
- A 401/403 anywhere drops the key and bounces to sign-in (`client.ts:87`, `AuthContext.tsx:71`).

`sessionStorage` is per-tab and does not survive a tab close or browser restart, which is the whole source of the friction. It was very likely a deliberate choice to shrink the XSS blast radius: the master key is the highest-privilege credential (full admin, provider-secret management), so keeping it out of long-lived, cross-tab storage is defensible. Any "remember me" change is therefore a conscious security tradeoff, not a pure convenience bug.

One structural note: there is a single shared master key, not per-user accounts, so a "session" here just means "proved possession of the master key recently." This is standalone-only; hybrid mode has no management API.

### A. "Remember me" → localStorage (lowest effort)
Add an opt-in checkbox; when checked, store the key in `localStorage` instead of `sessionStorage` so it survives tabs and restarts.
- Pros: tiny change, no backend work, solves the friction.
- Cons: the raw master key sits in JS-readable storage indefinitely; any XSS can exfiltrate it, and it does not expire. Off by default at best. Convenience up, security down.

### B. Server-issued session cookie (recommended)
On sign-in, POST the master key once to a new endpoint (say `POST /v1/auth/session`); the backend verifies it and sets an `HttpOnly`, `Secure`, `SameSite` cookie holding a short-lived signed or opaque session token, not the master key. `apiFetch` switches to `credentials: "include"` and the auth layer (`src/gateway/main.py` middleware / `api/deps.py`) accepts the session cookie in addition to the existing master-key header.
- Pros: the raw key is never persisted in JS-readable storage; an `HttpOnly` cookie cannot be read by XSS; the session gets a TTL and can be revoked server-side; survives tabs and restarts (the convenience win). This is both more convenient and more secure than today.
- Cons: real backend work (mint/verify/expire a session token, optional session store), CSRF protection is now required for cookie-authenticated mutations (SameSite=Strict or Lax plus a double-submit/CSRF token), server-side logout must invalidate the cookie. Same-origin is already true (dashboard is served from the API origin), which keeps this simpler.

### C. Client-held short-lived bearer session token (middle ground)
Master key mints an opaque/JWT session token with a TTL; store the token (not the master key) client-side and send it as the bearer; refresh as needed.
- Pros: no CSRF machinery (still header-based); raw master key is not persisted; token expires and is revocable, shrinking the exposure window versus option A.
- Cons: the token is still JS-readable, so XSS can use it while it is valid; needs token issue/refresh plumbing.

## Recommendation

Prefer **B** (HttpOnly session cookie): it removes the re-paste friction and reduces exposure at the same time, which is the rare case where the convenient option is also the safer one. If the cookie + CSRF machinery is more than we want right now, **C** is a reasonable lighter step that still avoids persisting the raw key. **A** only behind an explicit, defaulted-off "remember me", with the risk called out in the UI.

Whichever we pick, keep the current safety behaviors: clear the React Query cache on login/logout (`AuthContext.tsx:44`, `:59`) and drop the session on any 401/403. Consider a configurable session TTL and an idle timeout.

## Decision needed

Which strategy (B, C, or A) do we commit to before implementation? Tagging `needs-decision`.

njbrake:
> _Note: this PR was implemented and drafted by Claude via back-and-forth with @njbrake. The decisions (option B from the issue, SameSite instead of CSRF-token machinery, the 7-day default TTL) are his; the prose and code are Claude's._

## Description

The dashboard forced operators to re-paste the master key on almost every open: the raw key lived in per-tab `sessionStorage`, so a new tab, a closed tab, or a browser restart lost the session.

This implements option B from #338, the server-issued session cookie. Signing in now sends the master key once to a new `POST /v1/auth/session` endpoint, which verifies it and answers with an opaque session token in an `HttpOnly` cookie. The sign-in survives tabs and restarts, and the browser never stores a JS-readable credential, so this is more convenient and more secure than the status quo at the same time.

**Backend**
- New `dashboard_sessions` table (plus Alembic migration) storing only SHA-256 token hashes. DB-backed rather than in-memory so every worker and replica accepts a session and revocation is global, following the same reasoning as the alias/provider TTL refreshers.
- `POST /v1/auth/session` mints the session; `DELETE` signs out (idempotent, revokes only the caller's own cookie session). Standalone-only, like the rest of the management API.
- Cookie attributes: `HttpOnly`, `SameSite=Strict`, `Path=/`, `Max-Age` from the new `dashboard_session_ttl_hours` config field (default 168h). `Secure` mirrors the request scheme instead of being hard-coded, so plain-HTTP LAN and localhost deployments still receive the cookie back; that is no worse than such deployments already sending the raw key in cleartext.
- `api/deps.py`: `verify_master_key` and `verify_api_key_or_master_key` accept the cookie only when the request carries no header credentials at all, so API clients keep exactly the previous behavior, including failure modes.
- CSRF: `SameSite=Strict` is the primary control (the dashboard and API are same-origin, and the gateway origin serves no user-generated content); a `Sec-Fetch-Site` check on the cookie path adds defense in depth. `same-site` is deliberately rejected so a sibling-subdomain page cannot ride the cookie.
- Master-key rotation revokes every dashboard session in the same transaction (a session only proves possession of the now-dead key) and re-mints the caller's cookie on the rotation response, so the tab performing the rotation stays signed in.

**Frontend**
- The key is sent once to mint the session and never stored (previously it sat in `sessionStorage`). A non-secret `localStorage` marker (`otari.dashboard.hasSession`) keeps the signed-in state synchronous on load; if it is ever stale, the first 401 drops it and bounces to sign-in, the same path as any mid-session revocation.
- Sign-out revokes the session server-side before clearing local state; the React Query cache is still cleared on login/logout as before.
- `replaceMasterKey` plumbing is gone: the rotation response re-mints the cookie server-side, so the client no longer swaps credentials at all.

Verified end-to-end against a live gateway: cookie mint, cookie-only management reads, wrong-key/no-credential/cross-site rejections, server-side sign-out revocation, header auth unaffected, and a session minted before a full gateway restart still signs in (the exact friction in the issue).

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes #338

## Checklist

Note on tests: lint, typecheck, the full unit suite (881 passed, including 8 new session tests), the web checks (typecheck, 199 tests, fresh committed bundle), and `make openapi-check` all ran locally. The integration suite needs Docker/testcontainers, which the sandbox does not have, so it is covered by CI on this PR.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

Claude Code (Claude Fable 5)

Implemented via the repo's fix-github-issue workflow with @njbrake choosing the approach (option B) and reviewing before push.

## Summary

- Added persistent dashboard sign-in using server-managed sessions stored in secure, HttpOnly cookies.
- Added sign-in and sign-out endpoints, session storage, configurable expiration, and server-side revocation.
- Sessions are revoked when the master key changes, including during startup and key rotation.
- Updated the dashboard to use a non-secret local sign-in marker instead of storing the master key in browser storage.
- Preserved existing API header authentication and added protections against cross-site cookie use.
- Updated documentation, API definitions, generated dashboard assets, and backend/frontend tests.

## Benefits

- Sign-in persists across tabs and browser restarts.
- The master key is no longer stored in JavaScript-readable browser storage.
- Operators can sign out and sessions can be revoked centrally.


khaledosman (approved):
**Approved.** Solid, well-tested implementation. Correcting two findings from my initial review below.

**Correction (both withdrawn):**
- My "`Secure` dropped behind a TLS-terminating proxy" concern was wrong. `request_is_https()` (`dashboard_session_service.py`) deliberately honors `X-Forwarded-Proto` regardless of source, and `create_session` mints the cookie with `secure=request_is_https(request)`, so the PaaS/proxy case is handled, not `request.url.scheme == "https"` as I stated.
- The related "HTTPS `Secure` test gap" was also wrong: `test_https_requests_get_a_secure_cookie` and `test_forwarded_https_gets_a_secure_cookie` both cover it. The cross-PR #388 concern rested on these, so it's withdrawn too.

**Remaining points (both minor, inline):** the unauthenticated/unthrottled sign-in endpoint, and `Path=/` extending the cookie to inference routes (the code owns both tradeoffs in docstrings).

**Verified correct:** token entropy/hashing (`secrets.token_urlsafe(32)` → SHA-256, PK lookup no meaningful timing oracle); the `str | None` return change on `verify_master_key` (only rotation binds it); rotation-via-cookie CAS preserves concurrent-rotation rejection; restart survival via `revoke_sessions_on_master_key_change` + re-mint. Lifecycle coverage (header-wins-over-cookie, cross-site rejection, server-side revocation, expiry, restart) is thorough.

src/gateway/api/routes/auth_session.py
```diff
@@ -0,0 +1,98 @@
...
+@router.post("")
+async def create_session(
+    body: CreateSessionRequest,
+    request: Request,
+    response: Response,
+    db: Annotated[AsyncSession, Depends(get_db)],
+    config: Annotated[GatewayConfig, Depends(get_config)],
+) -> SessionResponse:
+    """Verify the master key and set the HttpOnly session cookie."""
+    if not await is_valid_master_key(body.master_key, config, db):
```
khaledosman, line 55: Minor / defense-in-depth: `POST /v1/auth/session` verifies the master key with no rate limiting, and each attempt does a DB read for the generated-key hash. Brute force against a high-entropy key is infeasible and header auth was already unthrottled, so this isn't a new credential exposure, just a new unauthenticated DB-touching surface. Worth a rate limit / lockout, or confirming the deployment fronts one.

src/gateway/services/dashboard_session_service.py
```diff
@@ -0,0 +1,170 @@
...
+    """
+    max_age = max(0, int((expires_at - datetime.now(UTC)).total_seconds()))
+    response.set_cookie(
+        SESSION_COOKIE_NAME,
+        token,
+        max_age=max_age,
+        httponly=True,
+        secure=secure,
+        samesite="strict",
+        path="/",
```
khaledosman, line 164: Minor: `Path=/` attaches the session cookie to every gateway request including `/v1/chat/completions`, so a same-origin browser fetch with no `Authorization` bills as the master user. Intended and cross-site-blocked by `SameSite=Strict` (and the docstring above owns the tradeoff); flagging only that scoping the cookie to the management surface would be tighter.

khaledosman (commented):
Two inline notes below (both minor). See the updated review summary for the correction to my earlier findings.

src/gateway/api/routes/auth_session.py
njbrake: _Note: this reply was drafted by Claude on @njbrake's behalf; the decision is his._

Agreed on the surface being new; tracked as a follow-up in #390 (failed-attempt rate limit on this endpoint) rather than growing this PR.

src/gateway/services/dashboard_session_service.py
njbrake: _Note: this reply was drafted by Claude on @njbrake's behalf; the decision is his._

Agreed, and keeping Path=/ as documented: the management routes live directly under /v1 beside inference, so a narrower path cannot cover them, and the cookie grants nothing beyond master authority with cross-site use blocked by SameSite=Strict.

PR #385, opened 2026-07-23
fix: surface actionable tools + reasoning_effort provider error

### LINKED ISSUE #331: Surface OpenAI's actionable error when tools + reasoning_effort are unsupported on /v1/chat/completions (e.g. gpt-5.6-sol)

njbrake:
## Description

Some newer OpenAI reasoning models (e.g. `gpt-5.6-sol`) reject a request that combines function tools with a non-`'none'` `reasoning_effort` on the chat-completions endpoint, and the upstream error is actionable: retry with `reasoning_effort='none'`, or use the responses endpoint. Today `classify_provider_error()` (`src/gateway/api/routes/_pipeline.py`) flattens every upstream 400/422 to the generic `PROVIDER_BAD_REQUEST_DETAIL`, so callers never see that remedy.

This adds a narrow classification case: when a 400/422 is flagged with `param == "reasoning_effort"` and its message names "function tools", return a new fixed `PROVIDER_REASONING_TOOLS_UNSUPPORTED_DETAIL` instead of the generic string. Any other 400/422 (including an unrelated `reasoning_effort` rejection, or a "function tools" message on a different param) still falls through to the generic detail.

Design notes:

- **No raw provider text leaked.** The new detail is a fixed string, preserving the existing no-leak guarantee (`test_error_detail_leakage`). No request mutation, no format bridging.
- **Forward-compatible with any-llm's unified-exception flip.** `param` and `message` are read off the raw exception and its `original_exception`, so the check works today against raw SDK errors and keeps working once `ANY_LLM_UNIFIED_EXCEPTIONS` becomes the default (see [any-llm#1190](https://github.com/mozilla-ai/any-llm/issues/1190)). No cleanup debt is created.
- **Covers all completion routes.** chat, messages, responses, and hybrid mode share this classifier, so all four benefit automatically.
- **Not limited to explicit `tools`.** The gateway's own tool loop injects function-tool schemas (MCP / sandbox / web_search), so this can surface even when the caller sent no `tools` array.

Why here and not in any-llm: preventing the error upstream isn't feasible ([any-llm#1189](https://github.com/mozilla-ai/any-llm/issues/1189), closed: OpenAI exposes no capability metadata, so it would need an unmaintainable per-model version matrix). Translating an upstream failure into otari's safe, fixed, caller-facing detail is otari's own error-contract responsibility, which is what `classify_provider_error()` owns.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes #331

## Checklist

- [x] Documentation was updated where necessary. (No user-facing docs change: internal error-classifier behavior only.)
- [ ] If the API contract changed, I regenerated the OpenAPI spec. (API contract unchanged; `openapi-check` confirms the spec is still up to date.)

## AI Usage

- [ ] No AI was used.

Implementation and tests were drafted by Claude via back-and-forth with @njbrake; the reasoning and decisions are his.


khaledosman (approved):
Approving. No correctness, security, or leak issues: the narrow `param == "reasoning_effort"` + "function tools" match, the `original_exception` unwrap, and the fixed-string detail are all sound, and tests cover both fall-through cases plus the wrapped-exception path. A few minor readability/robustness notes are posted as inline comments.

src/gateway/api/routes/_pipeline.py
```diff
@@ -202,6 +206,58 @@ def _upstream_status_code(exc: BaseException) -> int | None:
...
+    provider flags the offending field as ``param == "reasoning_effort"`` and
+    names "function tools" in the message. Match narrowly on both so an
+    unrelated ``reasoning_effort`` rejection still falls through to the generic
+    bad-request detail. Note the gateway's own tool loop can inject function
+    tools (MCP / sandbox / web_search) even when the caller sent no ``tools``,
+    so this can surface without an explicit ``tools`` array.
+    """
+    if _upstream_error_param(exc) != "reasoning_effort":
+        return False
+    return "function tools" in _upstream_error_message(exc).lower()
```
khaledosman, line 263: The `"function tools"` substring probe is English-text-coupled: if OpenAI rewords the message (e.g. "function calling tools", or a localized/variant phrasing) this case silently reverts to the generic detail with no test signal. That's the intended safe degradation, not a bug, but worth a comment noting the match is best-effort against current OpenAI phrasing.

src/gateway/api/routes/_pipeline.py
```diff
@@ -202,6 +206,58 @@ def _upstream_status_code(exc: BaseException) -> int | None:
...
+    fixed-string classification, never echoed back to the caller.
+    """
+    parts: list[str] = []
+    for candidate in (exc, getattr(exc, "original_exception", None)):
+        if candidate is None:
+            continue
+        message = getattr(candidate, "message", None)
+        if isinstance(message, str):
+            parts.append(message)
+        parts.append(str(candidate))
```
khaledosman, line 240: Minor: for the raw-SDK path `getattr(candidate, "message")` and `str(candidate)` are typically the same text for `openai.BadRequestError`, so the matched haystack contains the message twice. Harmless (substring match is idempotent) — just slightly redundant.

src/gateway/api/routes/_pipeline.py
```diff
@@ -202,6 +206,58 @@ def _upstream_status_code(exc: BaseException) -> int | None:
...
 
+def _upstream_error_param(exc: BaseException) -> str | None:
+    """Pull the offending request field off an upstream exception, if it names one.
+
+    OpenAI-style provider errors carry the rejected field on ``exc.param`` (from
+    the error body). any-llm currently surfaces the raw SDK error, so the
+    attribute is read directly; it also survives on ``original_exception`` once
+    any-llm's unified-exception layer becomes the default, so unwrap that too.
+    """
+    for candidate in (exc, getattr(exc, "original_exception", None)):
```
khaledosman, line 217: Precedence here is correct: a raw `exc.param` shadows `original_exception.param`, so the outer raw error is authoritative when present, and the `_WrappedError` test confirms the fallthrough when `exc` carries no `param`. No change needed.

khaledosman (commented):
Inline notes below. Classification logic is sound and tests are thorough; minor points only, nothing blocking.

PR #386, opened 2026-07-23
feat(dashboard): adopt recharts and retire hand-rolled UsageChart

### LINKED ISSUE #353: Dashboard: adopt a charting library and retire the hand-rolled UsageChart

Adopt a standard charting library for the dashboard and retire the hand-rolled SVG chart.

## Why

The dashboard renders charts by hand: `UsageChart` in `web/src/pages/UsagePage.tsx` is a bespoke SVG bar chart (private to that page), and there is no charting library in `web/package.json`. Hand-rolled visualizations accrue maintenance cost and miss the things a library gives for free (tooltips, responsive sizing, axis handling, accessibility). As the dashboard grows more views that want trend lines and sparklines, we do not want to keep extending the from-scratch pattern.

This was surfaced while planning the overview / home page (#302). The overview intentionally ships without inline charts so it stays a focused feature PR; trend sparklines on its KPI tiles are deferred to this work.

## Scope

- [ ] Pick a charting library. Default candidate: **recharts** (declarative, React-native, well-maintained, handles responsive/tooltips/axes). Lightweight alternative if bundle size is a concern: **uPlot** (~40kb, very fast for time series, more imperative). Record the choice and the bundle-size impact.
- [ ] Add it to `web/` and establish a small shared chart wrapper/component so pages do not each reimplement chart setup.
- [ ] Replace the hand-rolled `UsageChart` in `UsagePage.tsx` with the library, preserving current behavior (single-series bar chart of the selected metric, no meaning encoded by hue alone).
- [ ] Add a reusable sparkline for KPI tiles (the usage summary endpoint already returns the bucketed `series`, so the data is on the wire).
- [ ] Follow-up (can be its own PR): drop sparklines onto the overview page's spend and request-volume tiles.

## Notes

- Keep the accessibility posture the current chart already has: do not encode information by color/hue alone.
- Rebuild and commit the `web` bundle (`src/gateway/static/dashboard/`) as usual, or CI fails the staleness check.

Related: #302 (overview page, where sparklines are deferred to here), part of #299.

njbrake:
## Description

Retires the hand-rolled SVG `UsageChart` and adopts a real charting library for the dashboard.

**Library:** recharts (the issue's default candidate: declarative, React-native, tooltips/axes/responsive out of the box).

A small shared chart module lands in `web/src/components/charts.tsx` so pages compose charts instead of re-solving axes/tooltips/sizing:

- `BarTrendChart`: the single-series bar chart the Usage "Over time" panel now uses in place of `UsageChart`. Behavior is preserved (one series, one brand color so nothing is encoded by hue alone; the breakdown tables and caption stay the accessible source of truth).
- `Sparkline`: a reusable, axis-free trend line for KPI tiles. `StatCard` gained an optional `chart` slot. Two consumers ship here, both drawing off the bucketed series already on the wire:
  - Usage page: Spend / Requests / Tokens tiles.
  - Overview page: Spend and Requests (last 30 days) tiles (the #302 follow-up, folded in per discussion).

Also trims two Overview tiles per review: "Providers healthy" (a degraded state already surfaces in the attention strip, so the tile was redundant) and "Cache reads, last 30 days".

**Bundle-size impact:** recharts is ~371 kB raw, ~109 kB gzip. It is split into its own vendor chunk and loaded only with the chart-bearing route bundles, not the app shell (it is not in the shell's modulepreload set). Because the Overview is the landing route, that chunk now loads on the initial route, still lazily after the shell rather than blocking it. uPlot (~40 kB) was the lighter alternative considered; recharts was chosen for the declarative React fit and the free tooltip/axis/responsive handling, since the cost stays confined to the routes that actually chart.

Accessibility posture is unchanged: single-color series, no meaning encoded by hue alone; charts carry `role="img"` labels and the data tables remain the readable source of truth.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [x] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes #353. Folds in the sparkline follow-up from #302. Part of #299.

## Checklist

Dashboard checks run: `npm --prefix web run typecheck`, `npm --prefix web test` (199 passing), `npm --prefix web run build` (bundle rebuilt and committed). No backend/API change, so the OpenAPI spec is untouched.

## AI Usage

- [ ] No AI was used.

Implementation and tests were written with Claude Code via back-and-forth with @njbrake; the library choice, the tile trims, and the scope decisions are his.


khaledosman (approved):
Approving. Solid recharts adoption; no correctness issues in the chart wiring, the `ResizeObserver` test stub, the `hasTrend` (2+ point) guard, or the vendor-chunk split, and the removed imports are fully dropped.

Detailed points are now inline (see the follow-up review): a test whose name outpaces its assertion, an all-zero chart/caption mismatch, an untested `ChartTooltip`, reduced per-point non-visual access vs the old SVG, and the recharts chunk still loading on the landing route despite the shell split.

web/src/pages/OverviewPage.test.tsx
```diff
@@ -122,23 +122,36 @@ describe("OverviewPage", () => {
...
     });
     renderPage(<OverviewPage />);
 
-    expect(await screen.findByText("5.1M")).toBeInTheDocument();
-    expect(screen.getByText("Cache reads, last 30 days")).toBeInTheDocument();
+    expect(await screen.findByRole("img", { name: "Spend trend over the last 30 days" })).toBeInTheDocument();
+    expect(screen.getByRole("img", { name: "Request volume trend over the last 30 days" })).toBeInTheDocument();
   });
 
   it("shows a dash for error rate when there are no requests", async () => {
```
khaledosman, line 150: This test's name ("shows a dash for error rate when there are no requests") no longer asserts the dash: it only checks the `"Error rate, last 30 days"` label renders and that `"Elevated"` is absent, so it would pass even if the value rendered `0%` or `0/0` instead of `—`. Assert the actual `—` value node to keep the test honest to its name.

web/src/components/charts.tsx
```diff
@@ -0,0 +1,100 @@
...
+  return (
+    <div className="rounded-md border border-[var(--otari-line)] bg-[var(--otari-surface)] px-2 py-1 text-xs shadow-sm">
+      <div className="text-[var(--otari-muted)]">{label}</div>
+      <div className="font-medium tabular-nums text-[var(--otari-ink)]">{formatValue(value)}</div>
+    </div>
+  );
+}
+
+// A single-series bar chart of one metric over time. Responsive width, fixed
+// height; the y-axis is omitted (the caller's caption carries the peak) and the
```
khaledosman, line 47: `BarTrendChart` sets no `YAxis`/`domain`, so recharts auto-scales bar heights to the data max. The caption in `UsagePage.tsx` reports `trend.peak` floored at `1` (`Math.max(1, …)`). For an all-zero series these disagree: the caption says "1 peak" while recharts derives its own degenerate `[0,0]` domain. The retired hand-rolled chart tied bar heights to that same floored max, so this is a small behavior change in the empty/all-zero case. Consider `<YAxis hide domain={[0, (max) => Math.max(1, max)]} />` to realign, or drop the floor from the caption.

web/src/components/charts.tsx
```diff
@@ -0,0 +1,100 @@
...
+  payload,
+  formatValue,
+}: {
+  active?: boolean;
+  label?: ReactNode;
+  payload?: readonly { value?: number }[];
+  formatValue: (value: number) => string;
+}) {
+  const value = payload?.[0]?.value;
+  if (!active || value == null) {
```
khaledosman: `ChartTooltip`'s branches (active/inactive, `value == null`, the `formatValue` path) are untested; `charts.test.tsx` only asserts the chart mounts and bar/line counts. The `value == null` guard here (correctly allowing `0`) is exactly the kind of thing worth one render-with-payload test.

web/src/components/charts.tsx
```diff
@@ -0,0 +1,100 @@
...
+        <BarChart data={data} margin={{ top: 4, right: 4, left: 4, bottom: 0 }}>
+          <XAxis
+            dataKey="label"
+            tickLine={false}
+            axisLine={false}
+            interval="preserveStartEnd"
+            minTickGap={24}
+            tick={{ fontSize: 10, fill: "var(--otari-muted)" }}
+          />
+          <Tooltip cursor={{ fill: "var(--otari-line)", opacity: 0.35 }} content={<ChartTooltip formatValue={formatValue} />} />
```
khaledosman, line 73: The per-bar `<title>` the old SVG gave each bucket (keyboard/hover + AT access to individual values) is gone; this recharts `<Tooltip>` is pointer-only. Acceptable since the breakdown tables remain the accessible source of truth and the chart carries a `role="img"` label, but it is a real reduction in per-point non-visual access, worth confirming that's intended.

web/src/pages/OverviewPage.tsx
```diff
@@ -12,9 +12,10 @@ import {
   useUsers,
 } from "@/api/hooks";
 import type { UsageEntry } from "@/api/types";
+import { Sparkline } from "@/components/charts";
```
khaledosman, line 15: The split correctly keeps recharts out of the shell chunk, but `OverviewPage` (the landing route) statically imports `Sparkline` → `charts` → recharts, so the ~109 kB gzip chunk still loads on the very first route. The PR body acknowledges this; flagging only because "loaded only with chart-bearing route bundles, not the app shell" reads as if the landing paint avoids it. If landing-route weight matters, gate the sparklines behind a lazy import or drop them from Overview.

web/vite.config.ts
```diff
@@ -60,6 +60,9 @@ export default defineConfig({
           heroui: ["@heroui/react"],
           react: ["react", "react-dom", "react-dom/client", "react/jsx-runtime", "react-router-dom"],
           "tanstack-query": ["@tanstack/react-query"],
+          // recharts (and its d3 deps) is a large, self-contained vendor lib. Split
```
khaledosman, line 63: The recharts vendor chunk is correctly split out of the shell here, but note the landing Overview route pulls it in via its static `Sparkline` import (see OverviewPage.tsx), so the comment's "loads with the chart-bearing route bundles, not the shell" is accurate for the shell but the chunk still lands on first paint of the default route.

khaledosman (commented):
Follow-up with inline comments (the earlier review collapsed everything into one block). No correctness issues in the recharts wiring, ResizeObserver stub, `hasTrend` (2+ point) guard, or vendor-chunk config; removed imports (`formatTokens`, `PROVIDER_WORDS`) are fully dropped with no dangling references. Points inline below, mostly test/accessibility/bundle nuance.

PR #388, opened 2026-07-23
fix(models): list env-credentialed providers in /v1/models

### LINKED ISSUE #221: /v1/models discovery and completion routing disagree on what a "configured provider" is

## Summary

A provider that is usable for completions is not necessarily listed by `/v1/models`, because the two paths use different definitions of what counts as a "configured" provider.

## Details

- Completion routing works for any any-llm provider once its native credential env var is set. `get_provider_kwargs` returns `{}` for providers absent from `config.providers`, so any-llm reads the env var directly (`src/gateway/services/provider_kwargs.py:26`).
- `/v1/models` discovery only queries providers present in `config.providers` (`src/gateway/services/model_discovery_service.py:138`).

## Impact

Setting only `ANTHROPIC_API_KEY` (the path the Railway template README recommends for adding "another provider") makes Anthropic completions work, but Anthropic models do not appear in `/v1/models` until `anthropic` is also added to `OTARI_CONFIG_YAML`. The two surfaces silently disagree, and the documented quick path leads users straight into it.

## Repro

1. Set only `ANTHROPIC_API_KEY` (no `anthropic` entry in `OTARI_CONFIG_YAML`).
2. Call `anthropic:<model>` via `/v1/chat/completions`. It works.
3. `GET /v1/models`. No anthropic models are listed.
4. Add `anthropic` to `OTARI_CONFIG_YAML` and redeploy. The models now appear.

## Possible directions

1. Have discovery also include any-llm providers with detectable env-var credentials, so the model list matches what is actually callable.
2. If full auto-detection is undesirable, emit a startup warning when a provider is callable but absent from `config.providers` (and therefore excluded from discovery).
3. At minimum, document the distinction in `docs/configuration.md` and the Railway template README.

## Related

Related to #213, but distinct. #213 covers multiple named instances of the same provider implementation (real OpenAI plus a self-hosted OpenAI-compatible backend). This issue is about the discovery vs. completion mismatch for any single provider added by env var alone.

njbrake:
## Description

`/v1/models` discovery and completion routing disagreed on what a "configured provider" is. Completions work for any any-llm provider once its native credential env var is set (`get_provider_kwargs` returns `{}` for a provider absent from `config.providers`, so any-llm reads the env var directly), but model discovery only queried `config.providers`. Setting only `ANTHROPIC_API_KEY` therefore made Anthropic completions work while its models never appeared in `GET /v1/models`, the exact quick path the Railway template recommends.

Discovery now also picks up any-llm providers made callable by an env var alone: a provider whose credential env var is present and that supports live model listing is treated as discoverable under its implementation name (the same `provider:model` selector routing resolves), unless it is already a configured instance. This is wired into both `GET /v1/models` and `GET /v1/models/discoverable` (and the `?provider=` filter), so the two surfaces agree with what completions can reach. Provider-health and the per-instance connection test are untouched; they operate on explicit or configured instances.

Also restores `os.environ` around each test in `test_config_env_loading`, whose python-dotenv loads leaked provider credentials into the process. That was harmless before, but now that discovery is env-aware it polluted an unrelated test; a pre-existing isolation gap the change surfaced.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [x] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes #221

## Checklist

- [x] If the API contract changed, I regenerated the OpenAPI spec (no contract change; `--check` confirms the spec is up to date).

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

Claude Code (Opus 4.8), driven by @njbrake.

Implemented and tested end to end by Claude via back-and-forth with @njbrake; the reasoning and decisions are his.

**NOTE:**

### DIFF
--- a/src/gateway/services/model_discovery_service.py
+++ b/src/gateway/services/model_discovery_service.py
@@ -221,7 +221,7 @@ def _supports_list_models(provider_name: str) -> bool:
         return False
 
 
-def _env_provider_instances(configured: set[str]) -> list[str]:
+def _env_provider_instances(configured_impls: set[str]) -> list[str]:
     """any-llm provider implementations made callable by their credential env var.
 
     Completion routing works for any any-llm provider once its native credential
@@ -234,14 +234,30 @@ def _env_provider_instances(configured: set[str]) -> list[str]:
     resolves.
 
     Only providers that support live model listing are included; one that cannot
-    list models would just error out and be dropped from the catalog. Providers
-    already in ``configured`` are skipped so a named instance is neither shadowed
-    nor duplicated.
+    list models would just error out and be dropped from the catalog. A provider
+    whose implementation already backs a configured instance is skipped
+    (``configured_impls`` holds those implementation names): otherwise a custom
+    instance such as ``my-anthropic`` (``provider_type: anthropic``) alongside
+    ``ANTHROPIC_API_KEY`` would dial the same credential twice and list the same
+    models under two keys.
+
+    A keyless local provider (``ollama``/``llamacpp``/``llamafile``, whose
+    ``env_key`` is absent) has no credential signal to detect and so is not
+    surfaced here; discovering it would mean dialing a fixed localhost endpoint on
+    spec. That narrow callable-but-undiscoverable case is tracked separately.
     """
     detected: list[str] = []
-    for metadata in AnyLLM.get_all_provider_metadata():
+    try:
+        all_metadata = AnyLLM.get_all_provider_metadata()
+    except Exception:
+        # Match the module's defensive posture (see _supports_list_models and the
+        # return_exceptions fanouts): a registry hiccup must not 500 GET /v1/models
+        # or /v1/models/discoverable. Fall back to configured-only discovery.
+        logger.warning("Could not enumerate provider metadata for env-based discovery", exc_info=True)
+        return detected
+    for metadata in all_metadata:
         name = metadata.name
-        if name in configured or not metadata.list_models:
+        if name in configured_impls or not metadata.list_models:
             continue
         env_key = metadata.env_key
         if not env_key:
@@ -264,10 +280,13 @@ def _discoverable_instances(config: GatewayConfig) -> list[str]:
 
     Configured instances first, then any-llm providers made discoverable by an
     env var alone (see ``_env_provider_instances``), so GET /v1/models matches
-    what completion routing can actually reach.
+    what completion routing can actually reach. Env detection is deduplicated
+    against the *implementations* the configured instances resolve to, so a
+    custom-named instance is never shadowed by a bare env-detected duplicate.
     """
     configured = list(config.providers.keys())
-    return configured + _env_provider_instances(set(configured))
+    configured_impls = {config.provider_instance_type(instance) for instance in configured}
+    return configured + _env_provider_instances(configured_impls)
 
 
 def _declared_models(config: GatewayConfig, instance: str) -> list[Model]:
--- a/src/gateway/services/model_discovery_service.py
+++ b/src/gateway/services/model_discovery_service.py
@@ -18,6 +18,7 @@ of upstream calls:
 """
 
 import asyncio
+import os
 import time
 from collections.abc import Awaitable, Callable, Sequence
 from dataclasses import dataclass, field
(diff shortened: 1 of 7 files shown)


khaledosman (approved):
Approving. Detailed points are inline (see the follow-up review). Themes:

- `_env_provider_instances` dedups by instance name, not implementation — a configured instance plus its matching credential env var can double-discover the same provider.
- `AnyLLM.get_all_provider_metadata()` is the one unguarded external call in an otherwise defensive module; a raise there 500s `/v1/models`.
- Env edge cases (slash-alternatives, compound, `"None"`, blank) verified correct; no new capability exposed. Tests are solid bar the same-implementation duplication case.

src/gateway/services/model_discovery_service.py
```diff
@@ -220,6 +221,55 @@ def _supports_list_models(provider_name: str) -> bool:
...
+
+    Only providers that support live model listing are included; one that cannot
+    list models would just error out and be dropped from the catalog. Providers
+    already in ``configured`` are skipped so a named instance is neither shadowed
+    nor duplicated.
+    """
+    detected: list[str] = []
+    for metadata in AnyLLM.get_all_provider_metadata():
+        name = metadata.name
+        if name in configured or not metadata.list_models:
```
khaledosman: Dedup is by instance *name*, not implementation. If a configured instance `my-anthropic` (`provider_type: anthropic`) coexists with `ANTHROPIC_API_KEY`, both `my-anthropic` and `anthropic` get discovered against the same credential — a duplicate upstream `list_models` call and two catalog entries for the same models. Routable but wasteful/surprising. Consider also skipping an env provider whose implementation already backs a configured instance (compare against `{config.provider_instance_type(i) for i in configured}`). Not covered by a test either.

src/gateway/services/model_discovery_service.py
```diff
@@ -220,6 +221,55 @@ def _supports_list_models(provider_name: str) -> bool:
...
+    name, which is exactly the request selector ``<impl>:<model>`` that routing
+    resolves.
+
+    Only providers that support live model listing are included; one that cannot
+    list models would just error out and be dropped from the catalog. Providers
+    already in ``configured`` are skipped so a named instance is neither shadowed
+    nor duplicated.
+    """
+    detected: list[str] = []
+    for metadata in AnyLLM.get_all_provider_metadata():
```
khaledosman: `AnyLLM.get_all_provider_metadata()` is unguarded, but every other external touch in this module is defensive (`_supports_list_models` swallows all, discovery uses `return_exceptions=True`). A raise here 500s both `GET /v1/models` and `/v1/models/discoverable`. Wrap it to match the module's posture.

src/gateway/services/model_discovery_service.py
```diff
@@ -220,6 +221,55 @@ def _supports_list_models(provider_name: str) -> bool:
...
+
+def _discoverable_instances(config: GatewayConfig) -> list[str]:
+    """Instance names to run discovery for.
+
+    Configured instances first, then any-llm providers made discoverable by an
+    env var alone (see ``_env_provider_instances``), so GET /v1/models matches
+    what completion routing can actually reach.
+    """
+    configured = list(config.providers.keys())
+    return configured + _env_provider_instances(set(configured))
```
khaledosman: Minor perf: this runs the full ~50-provider metadata enumeration + `os.environ` scan on every request — the cache only covers per-provider `list_models`, not instance-list construction. Cheap in absolute terms, but both `/v1/models` and `/v1/models/discoverable` call it on each dashboard Models-page load.

src/gateway/services/model_discovery_service.py
```diff
@@ -220,6 +221,55 @@ def _supports_list_models(provider_name: str) -> bool:
...
+    already in ``configured`` are skipped so a named instance is neither shadowed
+    nor duplicated.
+    """
+    detected: list[str] = []
+    for metadata in AnyLLM.get_all_provider_metadata():
+        name = metadata.name
+        if name in configured or not metadata.list_models:
+            continue
+        env_key = metadata.env_key
+        if not env_key:
```
khaledosman, line 263: Residual gap in the "catalog matches what routing can reach" thesis: keyless local providers (`ollama`/`llamacpp`/`llamafile`, `env_key == "None"`) drop out here yet stay callable for completions when running locally, so they remain callable-but-undiscoverable. Fine to defer, but the PR frames itself as closing exactly this disagreement — worth a note or a follow-up issue.

src/gateway/services/model_discovery_service.py
```diff
@@ -220,6 +221,55 @@ def _supports_list_models(provider_name: str) -> bool:
...
+            continue
+        # ENV_API_KEY_NAME may name interchangeable alternatives separated by "/"
+        # (e.g. gemini's "GEMINI_API_KEY/GOOGLE_API_KEY"); any one set is enough.
+        # Compound descriptions ("AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY")
+        # and the literal "None" are not real single variable names and simply
+        # never match os.environ. A blank value (common from container
+        # templating) is treated as unset, so it does not add a provider that
+        # would only fail to authenticate.
+        candidates = (part.strip() for part in env_key.split("/"))
```
khaledosman, line 272: Verified against any-llm locally: the slash-split handles `gemini` (`GEMINI_API_KEY/GOOGLE_API_KEY`), and the compound (`AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY`) / literal `"None"` / `""` values all correctly fall through to unset. No env-var-enumeration risk — env-detected providers were already routable by any key, so no new capability is exposed.

tests/unit/test_model_discovery_env_providers.py
```diff
@@ -0,0 +1,151 @@
...
+    def test_detects_provider_with_credential_env_var(self, monkeypatch: pytest.MonkeyPatch) -> None:
+        monkeypatch.setenv("ANTHROPIC_API_KEY", "sk-ant-test")
+        assert "anthropic" in _env_provider_instances(configured=set())
+
+    def test_absent_env_var_is_not_detected(self, monkeypatch: pytest.MonkeyPatch) -> None:
+        monkeypatch.delenv("ANTHROPIC_API_KEY", raising=False)
+        assert "anthropic" not in _env_provider_instances(configured=set())
+
+    def test_blank_env_var_is_treated_as_unset(self, monkeypatch: pytest.MonkeyPatch) -> None:
+        # A blank credential (common from container templating) would only fail to
```
khaledosman, line 40: Good coverage overall (slash-alternatives, blank-as-unset, configured-dedup, filter, operator-status, plus the end-to-end `/v1/models` integration test); the autouse `_restore_environ` and `_no_env_providers` isolation fixes are the right response to now-env-aware discovery. One gap: the same-implementation duplication case (configured `my-anthropic` + `ANTHROPIC_API_KEY`) isn't exercised — add it once the dedup behavior above is decided.

khaledosman (commented):
Follow-up: converting the earlier prose review into inline comments; see threads on `model_discovery_service.py` and the test files. No blockers.

**Retraction:** my earlier cross-PR concern (that this PR's Railway/PaaS promotion collides with a `Secure`-cookie downgrade in #384) was based on a misread of #384 against `main`. #384 honors `X-Forwarded-Proto` via `request_is_https()`, so the proxy case is handled and the concern does not apply. Withdrawn.

PR #404, opened 2026-07-23
fix(dashboard): CodeRabbit follow-ups on budgets, usage, and tool tests

### LINKED ISSUE #377: Dashboard: follow-up fixes surfaced by CodeRabbit on #367 (unrelated pages)

## Context

CodeRabbit's review of #367 flagged four dashboard findings that are out of scope for that PR. They live in pages #367 never changed; they only appeared in its diff because the code-splitting from #373 now emits each page as its own bundle chunk, so CodeRabbit reviewed the regenerated chunks. They were deferred to keep #367 focused on the pricing refresh feature. Tracking them here instead.

Caveat: every finding below was read from a minified, code-split bundle, not from source. A sibling finding in the same review (a claimed keyboard-accessibility gap on the Models table) turned out to be a false positive once checked against the real `Tr` component. So verify each against the current `web/src` source before acting; some may not reproduce.

### 1. Sanitize provider diagnostics before they reach the UI (Security, Major)
`ActivityPage.tsx` and `ProvidersPage.tsx` render provider-supplied error text verbatim (`entry.error_message`, connection-test and health error fields). Consider returning a generic message plus an opaque diagnostic ID to the dashboard, keeping the raw detail in protected server-side logs.

### 2. Reject fractional budget reset periods (Budgets, Minor)
`BudgetsPage.tsx` uses `Math.round` on the reset-period field, so `1.5` silently becomes 2 days even though the field promises whole days. Validate with `Number.isInteger(...)` and reject non-integer input rather than rounding the operator's value.

### 3. Do not show a stale test result for a newly edited URL (Tools and Guardrails, Minor)
`ToolsGuardrailsPage.tsx` leaves the URL field editable while a test request is pending and renders the returned data without the URL it belongs to. If the operator edits the value before the request completes, the old result can be shown against the new URL. Track the tested URL and only render matching results, or disable the input while a test is in flight.

### 4. Preserve the other active filter during usage drill-down (Usage, Minor)
On the Usage page, drilling into a dimension drops the current `model` or `user_id` filter and keeps only the clicked one. Drilling into a model while filtering by user loses the user constraint and shows unrelated activity. Merge the existing filters with the clicked value before navigating.

## Source pointers
- `web/src/pages/ActivityPage.tsx`, `web/src/pages/ProvidersPage.tsx`
- `web/src/pages/BudgetsPage.tsx`
- `web/src/pages/ToolsGuardrailsPage.tsx`
- `web/src/pages/UsagePage.tsx`

Related: #367, #373.

njbrake:
## Description

Addresses #377 (CodeRabbit's out-of-scope review of #367). Each of the four findings was verified against current source; two did not reproduce.

| # | Finding | Verdict | Action |
|---|---------|---------|--------|
| 1 | Sanitize provider diagnostics (Security, Major) | False positive | No change |
| 2 | Reject fractional budget reset periods | Real | Fixed |
| 3 | Stale test result for edited URL | Partly real | Fixed the real part |
| 4 | Preserve other filter on usage drill-down | Real | Fixed |

**Budgets (finding 2):** `PeriodPicker` used `Math.round(n) * DAY`, so 1.5 silently became 2 days. Now uses a local draft with `Number.isInteger` validation, rejects non-integer or non-positive input, flags it inline ("Enter a whole number of days"), and leaves it unsaved rather than coercing the operator's value.

**Usage (finding 4):** Drilling from a breakdown row into the Activity log dropped the other active filter (drilling into a model while filtering by user lost the user, and vice versa). Now merges the existing filter with the clicked value before navigating.

**Tools and Guardrails (finding 3):** The described scenario (editing the URL while a test is pending) is already handled: the `onChange` `test.reset()` detaches the mutation observer, so that result is dropped. The residual gap `reset()` misses is the committed URL changing from a refetch (no keystroke), which strands a result against a new URL. Now tracks the tested URL and renders a result only while it still matches the field. The added regression test fails without this guard.

**Already handled, no change (finding 1):** `ActivityPage` renders a hardcoded generic message, never `entry.error_message`. `ProvidersPage` test and health errors are already sanitized and capped server-side (`_short_error`: 300 char cap, strips API keys, drops the provider tag) and echoed to the master-key operator by design.

Rebuilt the committed dashboard bundle. The build is deterministic and produces identical hashes under Node 22 (CI) and Node 24, so the staleness check passes.

## PR Type

- [x] Bug Fix

## Relevant issues

Fixes #377

## Checklist

- [x] I have added or updated tests that cover my change (dashboard Vitest suite under `web/src`; 215 pass, 4 added).
- [x] I ran the Definition of Done checks locally (frontend DoD: `npm --prefix web run typecheck`, `npm --prefix web test`, `npm --prefix web run build`; this PR is frontend only).
- [ ] Documentation was updated where necessary. (No user-facing behavior docs affected.)
- [ ] If the API contract changed, I regenerated the OpenAPI spec. (No API contract change.)

## AI Usage

**AI Model/Tool used:** Claude Code (Claude Opus 4.8), directed by @njbrake.

**Any additional AI details you'd like to share:** Each CodeRabbit finding was verified against current source; two were confirmed false positives and left unchanged, with the reasoning recorded above and in the commit message.

## Summary

- Improved budget period validation to reject fractional and non-positive values, preserving invalid drafts for inline feedback and preventing submission.
- Prevented stale Tools & Guardrails test results from appearing after a URL changes.
- Preserved active model and user filters when navigating from Usage breakdowns to Activity.
- Rebuilt the committed dashboard assets with updated deterministic hashes.
- Added regression coverage for budget validation, stale test results, and filter persistence.


web/src/pages/BudgetsPage.tsx
```diff
@@ -68,7 +68,21 @@ function parseLimit(raw: string): { value: number | null; valid: boolean } {
 function PeriodPicker({ value, onChange }: { value: number | null; onChange: (seconds: number | null) => void }) {
   const isPreset = PERIOD_PRESETS.some((p) => p.seconds === value);
   const [custom, setCustom] = useState(!isPreset);
-  const customDays = value !== null && value % DAY === 0 ? String(value / DAY) : "";
+  const committedDays = value !== null && value % DAY === 0 ? String(value / DAY) : "";
+  // A local draft so an in-progress, not-yet-valid entry (e.g. "1.5") stays on
+  // screen to be flagged, rather than being coerced. Reseed it when the committed
+  // value changes from outside (a preset click, or editing a different budget).
+  const [draft, setDraft] = useState(committedDays);
+  useEffect(() => {
```
khaledosman: This effect defeats the fix on the interactive and edit paths. When the operator makes a previously-valid value invalid, `onChange(null)` (L118) sets the parent `value` to `null`, so `committedDays` (L71) recomputes to `""`, and this effect then resets `draft` to `""` — wiping the invalid entry instead of holding it, so "Enter a whole number of days." never persists.

Repro: open **Edit** on a budget with a 14-day period (draft starts `"14"`), put the cursor in the field and type to make it `"14.5"` → the field blanks with no error. Same on create if you type a valid integer first, then a fractional char (e.g. `12` then `.5`).

The new test only *pastes* into an empty field, where `value` stays `null` and `committedDays` stays `""`, so this effect never fires — the one path that happens to avoid the bug.

The draft needs to be reseeded only on genuinely external value changes (preset click, switching which budget is edited), not on the `null` this component itself just emitted. Options: key `PeriodPicker` per budget and drop the effect; or track the last value this component emitted and skip the reseed when the incoming `value` matches it; or lift the draft so the invalid state isn't round-tripped through the parent.

web/src/pages/BudgetsPage.test.tsx
```diff
@@ -292,6 +292,30 @@ describe("BudgetsPage", () => {
...
 
+  it("rejects a fractional custom period instead of rounding it up", async () => {
+    const fetchMock = mockApi({ budgets: [] });
+    const user = userEvent.setup();
+    renderPage(<BudgetsPage />);
+
+    await user.click(await screen.findByRole("button", { name: "Create your first budget" }));
+    await user.click(screen.getByRole("button", { name: "Custom" }));
+    await user.click(screen.getByLabelText("Every N days"));
+    await user.paste("1.5");
```
khaledosman, line 299: Pasting into an empty field is the single input path that doesn't exercise the draft-reseed effect (`value`/`committedDays` never change from `null`/`""`), so this passes while the real defect goes uncovered. Add a case that makes a *committed* value fractional — e.g. type `.5` after an existing `14`, or open the Edit form for a budget with a period — and assert the field still shows `"14.5"` and the error. That test fails against the current code.

khaledosman (commented):
CodeRabbit follow-ups: reject fractional budget periods, carry the complementary filter on Usage drill-down, and gate a tool-test result on the URL it was run against. The triage table is well-reasoned and the two false positives are correctly left alone.

Two of the three fixes are solid; the budget one has a real defect (inline). Assessment of the parts I checked:

- **Usage drill-down (finding 4):** correct and complete — the Usage page only has model + user filters (plus the time window, which `drillTo` carries via `start_date`), so merging model↔user covers every cross-filter that exists. No api-key/status filter is dropped because none exists on this page.
- **Tool test URL gate (finding 3):** correct. `testedUrl` + `resultMatches` closes the no-keystroke refetch path that `test.reset()` misses, and the regression test genuinely fails without the guard (it drives a PATCH-triggered re-seed of the field). Good, subtle catch.
- **Budgets (finding 2):** the validation intent is right but the draft is wiped on the interactive path — see inline.

khaledosman (approved):

Issue #312, opened 2026-07-17
Dashboard: accessibility, responsive, and UX polish


njbrake:
> _Drafted by Claude via back-and-forth with @njbrake._

A tracking issue for cross-cutting dashboard polish.

## Tasks
- [ ] Accessibility pass (focus order, aria, contrast, keyboard navigation).
- [ ] Responsive/mobile layout pass.
- [ ] Consistent empty and error states across pages.
- [ ] Revisit routing: HashRouter vs. BrowserRouter with a server catch-all for clean URLs (weighed against shadowing `/v1` and the asset paths).
- [ ] Secret-field masking fallback for Firefox, which ignores `-webkit-text-security`.

Part of #299.

PR #406, opened 2026-07-23
fix(dashboard): make the dashboard mobile friendly

### LINKED ISSUE #351: Dashboard: make the dashboard mobile friendly
The Pages render great on desktop, but on mobile they look squashed

njbrake:
## Description

The dashboard rendered well on desktop but looked squashed on mobile. This makes the shell and the three worst-offending pages mobile friendly, while leaving the desktop layout intact.

**Sidebar (`AppShell`)**: below the `md` breakpoint the fixed-width sidebar (200px min, 60px collapsed) kept its footprint and crushed page content. It now renders as an off-canvas drawer, a proper accessible dialog: hidden by default, opened from a new header menu button, overlaid on a dimming backdrop, and dismissed by the backdrop, Escape, or tapping a destination. It carries dialog semantics (role, accessible name, aria-modal), moves focus in on open and restores it to the trigger on close, traps Tab while open, and is marked inert while closed so its off-canvas links leave the tab order. The resizable, collapsible rail is unchanged on desktop.

**Stat tiles (`StatCard`)**: fixed the Overview overlap (a fixed 180px min-width was wider than the grid track at two-up on mobile, so tiles overflowed into each other; switched to min-w-0) and the oversized look (HeroUI's Card already pads its content and `Card.Content` was stacking another pad on top, roughly doubling each tile's height; zeroed the card's own padding and made the content padding and value font responsive). Tile height dropped from 132px to 100px.

**Usage and Activity**: Usage tiles move from flex-wrap to the same responsive grid as Overview, so two fit per row on mobile. On both pages the status/key/user/model filter controls collapse behind a "Filters" toggle on mobile (labelled with the active-filter count) so the tiles and table sit near the top; the time-range presets stay visible and desktop shows the filters inline.

Verified with real screenshots at 390px (Overview, Usage, Activity in both filter states, and the drawer open) and with a scripted keyboard walk of the drawer (open moves focus in, Escape closes and restores focus to the toggle, nav closes on selection). Confirmed the desktop layout is unchanged.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes #351. Part of #312 (responsive/mobile layout pass) and #299.

## Checklist

- [x] I have added or updated tests that cover my change. (Colocated Vitest tests under `web/src`: drawer open/close/Escape/backdrop/inert and focus restore, stat-tile fit and padding, and the Usage/Activity filter disclosures. This is a frontend-only change, so there is nothing in `tests/unit` or `tests/integration` to touch.)
- [x] I ran the Definition of Done checks locally. (`make lint` and `make typecheck` pass; `npm --prefix web run typecheck`, `npm --prefix web test` (230 passing), and `npm --prefix web run build` pass. No Python changed, so Python `make test` is unaffected.)
- [x] Documentation was updated where necessary. (No user-facing docs needed for this layout change.)
- [x] If the API contract changed, I regenerated the OpenAPI spec. (Not applicable: no API contract change.)

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:** Claude Code (Claude Opus 4.8), directed by @njbrake.

Implemented and verified by Claude via back-and-forth with @njbrake, who directed the work and reviewed the results. The reasoning and decisions are his; the code and this description are the agent's.

## Summary

- Made the dashboard responsive for mobile while preserving the desktop layout.
- Added an accessible off-canvas navigation drawer with keyboard, focus, backdrop, and navigation dismissal behavior.
- Improved mobile stat tiles and Usage card layout to prevent overlap.
- Added collapsible mobile filters with active-filter counts for Usage and Activity.
- Added frontend tests covering navigation, responsive cards, and filter interactions.
- Updated the bundled dashboard assets.


web/src/components/AppShell.tsx
```diff
@@ -281,23 +382,51 @@ export function AppShell() {
...
+        ) : null}
         <aside
           ref={asideRef}
-          style={{ width }}
+          id="app-sidebar"
+          // On mobile the drawer is a modal dialog; give it a name and mark it
+          // modal while open. While closed it is off-canvas, so inert takes its
+          // links out of the tab order and the accessibility tree until opened.
+          role={isMobile ? "dialog" : undefined}
+          aria-modal={isMobile && mobileNavOpen ? true : undefined}
```
khaledosman, line 417: The drawer is `aria-modal` while open, but only the aside is marked `inert` (and only while closed) — the page content behind the backdrop stays reachable by an AT virtual cursor, since `aria-modal` isn't universally honored. Pointer (backdrop) and keyboard (Escape + Tab trap) are covered, so this is a partial gap, not a break. For true modality, marking `<main>` inert while `isMobile && mobileNavOpen` would close it. Fine to defer.

web/src/components/AppShell.tsx
```diff
@@ -212,9 +235,71 @@ export function AppShell() {
...
+      asideRef.current?.focus();
+    } else if (asideRef.current?.contains(document.activeElement)) {
+      toggleRef.current?.focus();
+    }
+  }, [isMobile, mobileNavOpen]);
+
+  // Keep Tab within the open drawer so focus cannot wander to the page behind the
+  // backdrop. Paired with the aside being inert while closed, this bounds keyboard
+  // focus to whichever surface is actually interactive.
+  const trapFocus = useCallback((event: ReactKeyboardEvent<HTMLElement>) => {
```
khaledosman, line 294: The trap bounds Tab within the aside, but the header's toggle/logout controls sit outside it and aren't part of the trap. In practice focus starts inside the aside and wraps within it, so it stays bounded — just noting the trap isn't the sole guarantee; it leans on focus never being placed in the header while the drawer is open.

khaledosman (commented):
Makes the shell plus Overview/Usage/Activity mobile-friendly: off-canvas drawer with dialog semantics, StatCard overlap/double-padding fixes, and collapsible filters. Clean, and the drawer accessibility work is more thorough than typical. Two minor notes inline.

Confirmed non-issues while reviewing: the `hidden` + `md:flex` filter-region toggle relies on Tailwind's responsive-variant ordering, which correctly shows inline on desktop and collapses on mobile; `inert` as a boolean prop is React 19-supported and desktop passes `undefined` for all dialog attributes, so the rail is genuinely unchanged there; and the breakpoint-change effect plus the `if (!isMobile) return` focus-restore guard correctly avoid yanking focus to the hidden toggle.

Process: the committed `src/gateway/static/dashboard/` bundle is in the diff, so CI's staleness check should pass — just confirm it was regenerated from these `web/src` changes.

khaledosman (approved):

PR #411, opened 2026-07-24
feat(files): add S3-compatible object storage backend

champ18ion:
## Description
Adds `S3FileStore`, an implementation of the `FileStore` protocol for AWS S3 or any S3-compatible endpoint (MinIO, self-hosted), completing the second half of #156's streaming work.

Uses sync `boto3` via `asyncio.to_thread` rather than `aioboto3`, which pins an `aiobotocore`/`botocore` range incompatible with the version already required for Bedrock support (confirmed earlier on #156). Streaming uploads spool to a `SpooledTemporaryFile` (bounded in-memory threshold, spills to disk past it) before handing off to `upload_fileobj`, which manages multipart upload internally and aborts an incomplete multipart upload on failure; confirmed with a test that injects a mid-upload failure and checks no multipart upload lingers afterward.

Credentials resolve through boto3's standard chain (env vars, `~/.aws/credentials`, IAM role), not app config, matching how `master_key` is already treated as a secret. New config: `files_s3_bucket`, `files_s3_endpoint_url`, `files_s3_region`. S3 support sits behind an optional `otari[s3]` extra with a lazy import inside `S3FileStore`, so local-only installs aren't forced into it.

## PR Type
- [x] New Feature

## Relevant issues
Part of #156

## Checklist
- [x] I ran the Definition of Done checks locally (`make lint`, `make typecheck`, `make test`) — note: ran via a manual pip venv, not `uv`, since Windows is excluded from your lockfile's supported platforms; integration tests (`tests/integration`) couldn't run locally without Docker, but the modified files pass lint/typecheck/unit tests.
- [x] If the API contract changed, I regenerated the OpenAPI spec — not applicable this time, no routes changed.

## AI Usage

**Any additional AI details you'd like to share:** Design discussed and confirmed on #156 before implementation (credential handling, config fields, streaming/multipart strategy); code written by Claude Code under my direction and review, including verifying the multipart-abort behavior with a test rather than assuming it.

- [ ] I am an AI Agent filling out this form

## Summary

- Added a new S3-backed file storage backend (`S3FileStore`) for AWS S3 and S3-compatible services (e.g., MinIO).
- Extended configuration to allow selecting the new `"s3"` backend and supplying S3 bucket/endpoint/region settings.
- Updated storage reference/key generation to use shared helper logic for consistent formatting across file-store implementations.
- Added an optional dependency group for S3 support (`otari[s3]`) and expanded dev dependencies to include S3 typing/mocking tooling.
- Added unit tests for S3 operations, including streaming uploads/downloads, deterministic reference formatting, delete idempotency, and cleanup behavior for failed or cancelled multipart uploads.

## Technical notes

- Wraps synchronous `boto3` S3 calls with `asyncio.to_thread` so the async API remains responsive.
- Streaming uploads buffer data in a bounded `SpooledTemporaryFile` before using S3 multipart upload via `upload_fileobj`, with tests ensuring no orphaned multipart uploads remain.

### DIFF
--- a/src/gateway/services/file_store.py
+++ b/src/gateway/services/file_store.py
@@ -19,8 +19,8 @@ from __future__ import annotations
 
 import asyncio
 import tempfile
-from collections.abc import AsyncGenerator, AsyncIterator
-from contextlib import asynccontextmanager
+from collections.abc import AsyncGenerator, AsyncIterator, Iterator
+from contextlib import asynccontextmanager, contextmanager
 from pathlib import Path
 from typing import IO, TYPE_CHECKING, Protocol, runtime_checkable
 
@@ -210,6 +210,35 @@ class LocalDirFileStore:
         await asyncio.to_thread(_unlink)
 
 
+@contextmanager
+def _translate_s3_errors(storage_ref: str) -> Iterator[None]:
+    """Re-raise boto3/botocore failures as the ``OSError`` family callers expect.
+
+    The local backend surfaces storage failures as ``OSError`` (a missing blob
+    is ``FileNotFoundError``), and the route/service callers catch ``OSError``
+    accordingly: the download route maps it to a clean 500 and the delete
+    route's best-effort cleanup swallows it so a committed soft-delete never
+    becomes a 500. Those callers live in the default local-only install and
+    cannot import ``botocore`` (it ships behind the optional ``otari[s3]``
+    extra), so translating here keeps them backend-agnostic. A missing object
+    maps to ``FileNotFoundError`` to mirror the local backend; every other S3
+    failure maps to ``OSError``.
+    """
+    from botocore.exceptions import BotoCoreError, ClientError
+
+    try:
+        yield
+    except ClientError as exc:
+        code = str(exc.response.get("Error", {}).get("Code", ""))
+        if code in {"NoSuchKey", "NoSuchBucket", "404", "NotFound"}:
+            raise FileNotFoundError(storage_ref) from exc
+        msg = f"S3 operation failed for {storage_ref!r}: {exc}"
+        raise OSError(msg) from exc
+    except BotoCoreError as exc:
+        msg = f"S3 operation failed for {storage_ref!r}: {exc}"
+        raise OSError(msg) from exc
+
+
 class S3FileStore:
     """S3-compatible object-storage :class:`FileStore` (AWS S3, MinIO, or any
     S3 API-compatible endpoint via ``endpoint_url``).
@@ -258,7 +287,8 @@ class S3FileStore:
             finally:
                 body.close()
 
-        return await asyncio.to_thread(_get)
+        with _translate_s3_errors(storage_ref):
+            return await asyncio.to_thread(_get)
 
     async def put_stream(self, file_id: str, chunks: AsyncIterator[bytes]) -> tuple[str, int]:
         key = _shard_key(file_id)
@@ -320,7 +350,8 @@ class S3FileStore:
         return key, total
 
     async def get_stream(self, storage_ref: str) -> AsyncGenerator[bytes, None]:
-        response = await asyncio.to_thread(self._client.get_object, Bucket=self._bucket, Key=storage_ref)
+        with _translate_s3_errors(storage_ref):
+            response = await asyncio.to_thread(self._client.get_object, Bucket=self._bucket, Key=storage_ref)
         body = response["Body"]
         try:
             while chunk := await asyncio.to_thread(body.read, _STREAM_CHUNK_BYTES):
@@ -332,7 +363,8 @@ class S3FileStore:
                 logger.warning("get_stream: failed to close S3 response body for %s: %s", storage_ref, close_exc)
 
     async def delete(self, storage_ref: str) -> None:
-        await asyncio.to_thread(self._client.delete_object, Bucket=self._bucket, Key=storage_ref)
+        with _translate_s3_errors(storage_ref):
+            await asyncio.to_thread(self._client.delete_object, Bucket=self._bucket, Key=storage_ref)
 
 
 def build_file_store(config: GatewayConfig) -> FileStore:
--- a/src/gateway/services/file_store.py
+++ b/src/gateway/services/file_store.py
@@ -34,7 +34,9 @@ _STREAM_CHUNK_BYTES = 1024 * 1024
 # Incoming upload chunks spool here before handing the whole thing to boto3's
 # upload_fileobj (sync boto3 has no streaming-from-async-iterator primitive).
 # Small uploads never touch disk; past this threshold SpooledTemporaryFile
-# transparently rolls over to a real temp file, so memory stays bounded.
+# transparently rolls over to a real temp file, so memory stays bounded. This
+# is a per-upload budget, not a global one: N concurrent put_stream calls can
+# hold up to N * this many bytes in memory before any of them spill to disk.
 _SPOOL_MAX_MEMORY_BYTES = 10 * 1024 * 1024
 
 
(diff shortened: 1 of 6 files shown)


src/gateway/services/file_store.py
```diff
@@ -192,10 +208,88 @@ def _unlink() -> None:
...
+        key = _shard_key(file_id)
+        await asyncio.to_thread(self._client.put_object, Bucket=self._bucket, Key=key, Body=data)
+        return key
+
+    async def get(self, storage_ref: str) -> bytes:
+        def _get() -> bytes:
+            response = self._client.get_object(Bucket=self._bucket, Key=storage_ref)
+            return response["Body"].read()
+
+        return await asyncio.to_thread(_get)
```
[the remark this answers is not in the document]
champ18ion: good catch, fixed, now closes the body in a finally, same pattern as get_stream already used

src/gateway/services/file_store.py
```diff
@@ -192,10 +208,88 @@ def _unlink() -> None:
...
+            # no manual abort_multipart_upload is needed here.
+            await asyncio.to_thread(self._client.upload_fileobj, spool, self._bucket, key)
+        finally:
+            # Shielded like _open_handle: this runs during unwind on a
+            # cancelled upload too, and an unshielded await here could be cut
+            # off by a repeated cancel() before the spool file actually closes.
+            try:
+                await asyncio.shield(asyncio.to_thread(spool.close))
+            except Exception as cleanup_exc:
+                logger.warning("put_stream: failed to close spool file for %s: %s", key, cleanup_exc)
```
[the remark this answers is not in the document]
champ18ion: yeah that's a real one, to_thread cancelling only detaches us, doesn't stop the thread. now wraps the upload in a real Task and on cancellation shields a wait on it to see what actually happened before deciding whether to clean up an orphaned upload. added a test with threading.Event gates to deterministically reproduce cancel-after-success rather than guessing at timing

src/gateway/services/file_store.py
```diff
@@ -192,10 +208,128 @@ def _unlink() -> None:
...
+        except ImportError as exc:
+            msg = "S3FileStore requires boto3. Install it with: pip install otari[s3]"
+            raise ImportError(msg) from exc
+
+        self._bucket = bucket
+        self._client: S3Client = boto3.client("s3", endpoint_url=endpoint_url, region_name=region or "us-east-1")
+
+    async def put(self, file_id: str, data: bytes) -> str:
+        key = _shard_key(file_id)
+        await asyncio.to_thread(self._client.put_object, Bucket=self._bucket, Key=key, Body=data)
```
khaledosman, line 276: `put` buffers the whole body in memory, and `get` (L243) does `body.read()` — a file at the `files_max_bytes` ceiling (512 MB) is fully resident. This matches `LocalDirFileStore` so it's not a regression, but the S3 path could stream both directions. Fine to defer; worth a comment noting the intentional symmetry with the local backend.

src/gateway/services/file_store.py
```diff
@@ -18,15 +18,37 @@
...
 
+if TYPE_CHECKING:
+    from mypy_boto3_s3 import S3Client
+
 _STREAM_CHUNK_BYTES = 1024 * 1024
+# Incoming upload chunks spool here before handing the whole thing to boto3's
+# upload_fileobj (sync boto3 has no streaming-from-async-iterator primitive).
+# Small uploads never touch disk; past this threshold SpooledTemporaryFile
+# transparently rolls over to a real temp file, so memory stays bounded.
+_SPOOL_MAX_MEMORY_BYTES = 10 * 1024 * 1024
```
khaledosman, line 40: `_SPOOL_MAX_MEMORY_BYTES` is a per-upload budget: N concurrent `put_stream` calls hold up to N×10 MiB in RAM before spilling to disk. Reasonable — just noting it's not a global cap.

src/gateway/services/file_store.py
```diff
@@ -192,10 +208,128 @@ def _unlink() -> None:
...
+                    await asyncio.to_thread(spool.write, chunk)
+                await asyncio.to_thread(spool.seek, 0)
+                # upload_fileobj runs in a worker thread and, once started,
+                # keeps running there even if *we* get cancelled while
+                # awaiting it: to_thread's cancellation only detaches us from
+                # waiting, it does not stop the thread. Wrapping it in a real
+                # Task lets us find out what actually happened even after
+                # being cancelled, instead of assuming "cancelled" means
+                # "nothing was uploaded".
+                upload_task = asyncio.ensure_future(
```
khaledosman: Nit: `asyncio.ensure_future(asyncio.to_thread(...))` → `asyncio.create_task(...)` is the current idiom for wrapping a coroutine in a task.

src/gateway/services/file_store.py
```diff
@@ -192,10 +208,128 @@ def _unlink() -> None:
...
+                    asyncio.to_thread(self._client.upload_fileobj, spool, self._bucket, key)
+                )
+                # upload_fileobj manages multipart upload internally,
+                # including aborting an incomplete multipart upload if it
+                # fails partway, so no manual abort_multipart_upload is
+                # needed here.
+                await asyncio.shield(upload_task)
```
khaledosman, line 318: Worth calling out (non-obvious and easy to get wrong): this correctly handles that `to_thread` can't interrupt the worker thread — a cancel mid-`upload_fileobj` can still land the object, and the double-`shield` + re-await notices that and deletes the orphan instead of assuming "cancelled" means nothing was uploaded. The `threading.Event`-gated test makes that race deterministic rather than timing-based. Nicely done.

khaledosman (commented):
S3FileStore over sync boto3 via `to_thread`, behind the optional `otari[s3]` extra. Well-built; the streaming/cancellation path is the tricky part and it's correct and thoroughly tested (moto). Inline notes are all minor.

Optional: objects inherit the bucket's defaults for at-rest encryption (no `ServerSideEncryption` set). Since this stores user-uploaded bytes, a one-line doc note that SSE is the bucket operator's responsibility would help operators.

Confirmed non-issues while reviewing: the `self._client: S3Client` annotation is safe because `from __future__ import annotations` is in effect (never evaluated); sharing one low-level boto3 client across `to_thread` calls is fine (low-level clients are documented thread-safe); `endpoint_url` is admin config, not user input, so no SSRF surface; and a missing `files_s3_bucket` fails fast at startup via `build_file_store` in `main.py`.

khaledosman (approved):

tests/unit/test_s3_file_store.py
```diff
@@ -0,0 +1,176 @@
...
+        collected.extend(chunk)
+    assert bytes(collected) == payload
+
+
+@pytest.mark.asyncio
+async def test_delete_removes_object(s3_store: S3FileStore) -> None:
+    ref = await s3_store.put("file-deadbeef", b"data")
+    await s3_store.delete(ref)
+    with pytest.raises(Exception, match="Not Found|NoSuchKey|404"):
+        await s3_store.get(ref)
```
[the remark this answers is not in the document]
champ18ion: good catch, now asserts ClientError specifically and checks the S3 error code instead of just matching on a string

tests/unit/test_s3_file_store.py
```diff
@@ -0,0 +1,176 @@
...
+
+    def _gated_upload_fileobj(fileobj: IO[bytes], bucket: str, key: str) -> None:
+        upload_started.set()
+        release_upload.wait(timeout=5)
+        original_upload_fileobj(fileobj, bucket, key)
+
+    monkeypatch.setattr(client, "upload_fileobj", _gated_upload_fileobj)
+
+    task = asyncio.ensure_future(s3_store.put_stream("file-orphancheck01", _iter([b"data"])))
+    await asyncio.to_thread(upload_started.wait, 5)  # don't cancel until the upload thread is truly running
```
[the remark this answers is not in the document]
champ18ion: fair, fixed, now asserts the wait actually succeeded instead of silently letting a timeout through

src/gateway/services/file_store.py
champ18ion: good point, added a comment noting it's intentional symmetry with the local backend, not an oversight

src/gateway/services/file_store.py
champ18ion: fair, added a note that it's per-upload not global

src/gateway/services/file_store.py
champ18ion: done, switched to create_task

src/gateway/services/file_store.py
champ18ion: thanks for digging into that one closely, appreciate it

champ18ion:
thanks for the thorough review! addressed all 4 inline notes plus the SSE doc note in cd87518, all minor as you said so nothing functional changed

PR #413, opened 2026-07-24
fix(compose): bind quickstart postgres to localhost

### LINKED ISSUE #412: Quickstart docker-compose binds postgres to 0.0.0.0 with default credentials
### Summary

In the quickstart `docker-compose.yml`, the `postgres` service publishes its port with the Docker short syntax `"${OTARI_POSTGRES_PORT:-5433}:5432"`, which binds to `0.0.0.0` (every host interface). Combined with the static default credentials `otari/otari`, anyone who can reach the Docker host on TCP 5433 can log in and read or write the entire gateway database: API key hashes, users, budgets, usage data, and dashboard session rows.

Every other service in the same file is deliberately bound to `127.0.0.1` with a security comment explaining why. The database is the one exception, and it is the service that holds the most sensitive data.

### Affected location

- File: `docker-compose.yml:78` (port mapping), `:80-81` (default credentials)
- Commit: `bf90e94`

### Details

The port line uses the two-part short form:

```yaml
  postgres:
    ports:
      - "${OTARI_POSTGRES_PORT:-5433}:5432"
```

Docker's two-part form binds the host side to all interfaces. On a LAN, or on a cloud VM without a host firewall, port 5433 is then reachable from outside the host, and `otari/otari` grants full access to the database.

The rest of the compose file already follows the safer pattern. For example the sandbox:

```yaml
  sandbox:
    ports:
      # Bound to localhost only: the sandbox runs arbitrary user-supplied
      # code, so an open port on the host network would be an RCE surface
      # to anyone on the LAN.
      - "127.0.0.1:8180:8080"
```

The same `127.0.0.1:` prefix is used for `searxng`, `brave-adapter`, `tavily-adapter`, `anyguardrails`, and `encoderfile`. The comment on the postgres block itself states the host mapping is "only for external tools (psql / TablePlus / etc.)", so the gateway does not need it (it reaches the database over the Docker network at `postgres:5432`). Binding it to all interfaces looks unintended rather than a design choice.

### Impact

Read/write access to the whole gateway database with the shipped default credentials: API key hashes, user records, budgets, usage/spend data, and `dashboard_sessions`. On an internet-reachable host without a firewall this is a full data exposure; on a shared LAN it exposes the database to every peer on the network.

### Reproduction

1. `docker compose up` on a host reachable by another machine (a LAN peer, or a cloud VM whose 5433 is not firewalled).
2. From that other machine: `psql -h <host> -p 5433 -U otari -d otari` with password `otari`.
3. The connection succeeds and the full schema is readable and writable.

### Suggested fix

Bind the port to localhost, matching every other service in the file:

```yaml
      - "127.0.0.1:${OTARI_POSTGRES_PORT:-5433}:5432"
```

A PR with exactly this change is attached. As a secondary hardening step you may also want to source `POSTGRES_PASSWORD` from an environment variable rather than shipping a static `otari` default, so a copy-paste deployment does not run with guessable credentials even on localhost.

---
*Reported as part of [KIberblick.de](https://kiberblick.de), AI-assisted security analysis of open-source projects.*

pcornelissen:
> _Note: @njbrake restored the required template sections (via Claude) so the `check-template` gate passes. The contributor's original description is preserved below. The Checklist and AI Usage boxes are left unchecked, as those are the contributor's own attestations to make._

## Description

Bind the quickstart `postgres` service to `127.0.0.1` instead of all host interfaces.

The postgres port used Docker's two-part short syntax (`"${OTARI_POSTGRES_PORT:-5433}:5432"`), which binds to `0.0.0.0`. With the static default credentials `otari/otari`, any peer that can reach the host on port 5433 (a LAN machine, or a cloud VM without a firewall) could read and write the whole gateway database: API key hashes, users, budgets, usage, and dashboard sessions.

Every other service in the file (`sandbox`, `searxng`, `brave-adapter`, `tavily-adapter`, `anyguardrails`, `encoderfile`) is already bound to `127.0.0.1` with a security comment. The database was the only exception, and the comment on the block already notes the host mapping is only for external tools, so the gateway itself does not need it (it uses the Docker network at `postgres:5432`).

**Change:** `docker-compose.yml` adds the `127.0.0.1:` prefix to the postgres port mapping and expands the comment to explain the localhost bind, consistent with the other services. No behavior change for the gateway (it never used the host mapping). Operators who connect external tools from the host itself are unaffected; anyone who connected from another machine should use an SSH tunnel or set `OTARI_POSTGRES_PORT` and bind explicitly.

**Test:**
- `docker compose config` still parses cleanly.
- Verified the gateway continues to reach postgres over the internal Docker network (`postgres:5432`), which the change does not touch.

*Contribution as part of [KIberblick.de](https://kiberblick.de).*

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [x] Infrastructure / CI

## Relevant issues

Closes #412.

## Checklist

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -71,11 +71,15 @@ services:
   postgres:
     image: postgres:16-alpine
     ports:
-      # Host port 5433 (container still :5432) — avoids clashing with another
-      # local postgres on the default port. Gateway connects via the Docker
-      # network at postgres:5432, so the host mapping is only for external
-      # tools (psql / TablePlus / etc.). Override with $OTARI_POSTGRES_PORT.
-      - "${OTARI_POSTGRES_PORT:-5433}:5432"
+      # Bound to localhost only. The database holds API key hashes, users,
+      # budgets, usage, and dashboard sessions, so an open port on the host
+      # network would expose all of it (default creds otari/otari) to anyone
+      # on the LAN. The Otari container reaches it via the Docker network at
+      # postgres:5432; this host mapping is only for external tools (psql /
+      # TablePlus / etc.) on the operator's own machine.
+      # Host port 5433 (container still :5432) avoids clashing with another
+      # local postgres on the default port. Override with $OTARI_POSTGRES_PORT.
+      - "127.0.0.1:${OTARI_POSTGRES_PORT:-5433}:5432"
     environment:
       - POSTGRES_USER=otari
       - POSTGRES_PASSWORD=otari


njbrake (approved):

Thanks for this, and for the clear writeup in #412. This is a well-scoped hardening fix.

I traced the history to confirm the previous `0.0.0.0` binding was not intentional: the `127.0.0.1` pattern entered this file in `88551271` ("address Copilot review feedback on PR #65"), which locked down the sandbox service and, over the same effort, the other published ports. postgres was the one service that pass missed, even though its own comment already stated the host mapping is "only for external tools" on the operator's machine. This PR finishes that job and brings the last service in line with the rest of the file.

The change is correct: Docker's `IP:HOSTPORT:CONTAINERPORT` short form restricts publication to loopback, the gateway keeps reaching the DB over the Docker network at `postgres:5432`, and host-based tooling on the operator's own machine is unaffected.

One process note, not blocking: the template bot flagged the PR body for the missing PR Type / Checklist / AI Usage sections and has a 24h auto-close timer, so it is worth restoring those sections. I'm approving now regardless.

(For reference, CodeRabbit's suggestion to align `OTARI_POSTGRES_PORT` in `.env.example` is a non-issue: that file has no such entry, so there is nothing to align.)

PR #415, opened 2026-07-24
feat(dashboard): shared table framework with filtering, pagination, selection, and bulk actions

### LINKED ISSUE #403: Dashboard: shrinking a resized table column leaves a white gap inside the card

Part of the #312 UX-polish umbrella (epic #299).

### What happens
On any dashboard table (Activity, Usage, Keys, etc.), dragging a column's resize handle to make the column **narrower** leaves an empty strip of the card's white/surface background to the right of the table. The table compacts but the card does not compress with it, so it looks unfinished.

### Why
The tables are a hand-rolled `<table>` with a custom column-resize feature in `web/src/components/Table.tsx` (HeroUI v3's react-aria Table is intentionally not used). The bordered "card" is the table's own wrapper div (`Table.tsx:65`): `overflow-x-auto rounded-xl border bg-[var(--otari-surface)]`, a full-width block.

Once a column is dragged, the table switches to `table-fixed` and is given an explicit width equal to the sum of its column widths (`Table.tsx:68`, `width: ${totalWidth}px`). When a column shrinks, `totalWidth` drops below the wrapper width, so the table becomes narrower than its bordered card and the surface background shows through on the right. Widening past the container scrolls correctly via `overflow-x-auto`; only shrinking is affected.

This is a side effect of a deliberate choice (`Table.tsx:57-60`): sizing to the exact column sum "so shrinking one never redistributes slack into the others."

### Repro
1. Open Activity (or any table page).
2. Drag a column's right-edge handle to the left to shrink it.
3. A white gap appears between the end of the table and the card's right border.

### Options
- **A. Card hugs the table.** Add `w-fit max-w-full` to the wrapper in the resized state only, so it shrinks to the table (no gap) and still scrolls when wider. Simple; downside is the card becomes narrower than the filter row above it.
- **B. Table always fills the card.** Absorb the freed width into a flexible trailing column/spacer so the card stays full width. More consistent with the rest of the dashboard, but re-introduces the "slack redistributes" behavior the current code avoids.
- **C. Drop the custom resize.** The feature is a niche, non-persisted convenience (widths are in-memory only). Removing it returns the tables to a plain auto-width layout and deletes this whole class of bug. Worth weighing against how much the resize is actually used.

### Recommendation
Lean B for visual consistency, or C if the resize feature is not carrying its weight. A is the smallest patch if a quick fix is preferred.

### LINKED ISSUE #409: feat(dashboard): shared table framework — filtering, fast pagination, selection & bulk actions across all table pages

## Summary

Build one shared table framework for the dashboard and adopt it across every table page, so filtering, pagination, and row selection work the same everywhere. Adds a composable filter bar (date range + "add filter" menu), fast pagination (jump to first/last, type-a-page, rows-per-page), and row selection with bulk actions — starting with **mass delete** and **manual price fixes** on usage rows. Supersedes and closes the hand-rolled-table resize bug (#403).

## Motivation

The dashboard tables are inconsistent and thin on interaction:

- **Filtering** is hand-rolled per page, mostly not URL-synced, and there is no general way to add a filter or filter by an arbitrary date/time range.
- **Pagination** (Activity) moves one page at a time — no jump-to-first/last, no go-to-page, no control over rows per page.
- **No selection or bulk actions** anywhere, so routine operator work (delete a batch of imported rows, correct the cost on a set of rows, revoke several keys) is impossible from the UI and currently requires direct DB access.
- The tables are a hand-rolled `<table>` with a custom column-resize feature (`web/src/components/Table.tsx`), which carries its own bug (#403).

## Proposed solution

A shared, reusable table built on HeroUI v3's react-aria `Table`, plus a small set of surrounding primitives:

- `components/Table.tsx` — rebuilt on react-aria `Table` (selection, sort, keyboard grid nav from the library).
- `<TableToolbar>` — date-range control + `+ Add filter` menu + editable/removable filter chips; all filter state URL-synced (shareable, back-button-friendly).
- `<Pagination>` — first / prev / **type-a-page** / next / last, plus a **rows-per-page** selector (25 / 50 / 100 / 250 / 500).
- `useTableSelection` + a contextual **bulk-action bar**, including a **"select all N matching this filter"** affordance so bulk ops can target the whole filtered set, not just the visible page.

### Filtering

```
1h 24h 7d [30d] All │ Custom…      model:opus-4-8 ✕   unpriced ✕   + Add filter ▾
```

- Date range is always present: presets + a "Custom…" from/to datetime range.
- `+ Add filter` opens a menu of fields (Model, Source, User, API key, Status, Priced?); picking one adds an editable chip with the right control.

### Pagination

```
Rows [100 ▾]                 1–100 of 4,231   «  ‹  Page [ 1 ] / 43  ›  »
```

### Selection & bulk actions

- Per-row checkboxes; selecting the page reveals "Select all 4,231 rows matching this filter".
- Bulk-action bar is per-page (see matrix). For usage rows the first two actions are **Delete** and **Set price**.

## Backend (two new endpoints, master-key, standalone)

Both accept **either** an explicit list of row ids (the current selection) **or** a filter (for "all matching"), so one UI drives page-scoped and set-scoped operations.

- **`DELETE /v1/usage`** — delete usage rows by `ids[]` or by filter (`source`, `model`, `user_id`, `status`, `start_date`, `end_date`, priced/unpriced). Client shows a confirm dialog.
- **`POST /v1/usage/set-price`** — set cost for rows by `ids[]` or filter, given per-1M **input / output / cache-read / cache-write** rates; recomputes `cost` / `billing_meters` / `pricing_breakdown` from each row's own token counts (reuses `services/metered_pricing.calculate_metered_cost`). Manual rates, entered in the UI — not "recompute from configured pricing."

**Safety invariant:** both operations touch **only imported rows** (`counts_toward_budget = false`). Enforced/gateway rows and the spend ledger (`users.spend`) are never affected, matching the boundary used elsewhere in usage handling.

## Rollout

Nine table pages move onto the shared components; each opts into what fits it.

| Page | Filter + date | Paginate + per-page | Select | Bulk / row actions |
|---|:--:|:--:|:--:|---|
| Activity | ✅ | ✅ | ✅ | Delete, Set price, Export |
| Usage | ✅ | ✅ | ✅ | Delete, Set price (row list under a breakdown) |
| Models | ✅ | ✅ | ✅ | Bulk set pricing, enable/disable |
| API keys | ✅ | ✅ | ✅ | Bulk revoke/delete, exempt toggle |
| Users | ✅ | ✅ | ✅ | Bulk delete, assign budget |
| Budgets | ✅ | ✅ | ✅ | Bulk delete |
| Aliases | ✅ | ✅ | ✅ | Bulk delete |
| Providers | ✅ | ✅ | — | Row test/health (no bulk) |

## Phased delivery (each phase its own PR)

1. **Framework primitives** — rebuild `Table.tsx` on react-aria; add `<TableToolbar>`, `<Pagination>`, `useTableSelection` + bulk-action bar. No page-behavior change yet. **Closes #403** (removes the hand-rolled resize/white-gap code entirely).
2. **`DELETE /v1/usage`** endpoint + tests.
3. **`POST /v1/usage/set-price`** endpoint + tests.
4. **Adopt on Activity + Usage** (filters, pagination, selection, Delete + Set price).
5. **Roll across the remaining pages** (Models, API keys, Users, Budgets, Aliases, Providers) per the matrix.

## Scope notes

- **#403** is folded into Phase 1 and closed by it (the react-aria table replaces the resizable hand-rolled one).
- **Automatic repricing is out of scope.** A filter-scoped auto-reprice endpoint (recompute from *configured* pricing) was prototyped on the `feat/usage-reprice` branch (issue #407); it is **shelved** in favor of the manual `set-price` above. #407 stays open.
- **react-aria adoption reverses a deliberate decision** — `Table.tsx` currently avoids react-aria on purpose. This issue is the buy-in for that change; bundle size is measured in Phase 1 before rolling further.

## Out of scope

- Automatic repricing from configured pricing (#407).
- Saved filter views / column customization / CSV import.
- Per-row inline cell editing (bulk "Set price" covers the cost-correction need).

njbrake:
## Description

Builds one shared table framework for the dashboard and adopts it across every table page, so filtering, pagination, and row selection work the same everywhere, plus two backend endpoints for operator cleanup of imported usage. Implements #409.

Highlights:

- **Shared framework** on HeroUI v3's react-aria Table: `DataTable` (selection, sort, keyboard grid nav, column resize), `TablePagination` (rows-per-page plus first / prev / type-a-page / next / last), `useTableSelection` + `BulkActionBar` (with a "select all N matching this filter" affordance), `ConfirmDialog`, `SetPriceDialog`, and URL-synced filter/pagination helpers.
- **All nine table pages migrated** (Activity, Usage, Models, API keys, Users, Budgets, Aliases, Providers, Overview), each opting into selection and bulk actions per the issue's matrix (bulk delete / set-price on Activity and Usage, bulk delete on Aliases and Budgets, delete plus assign-budget on Users, disable / budget-exempt / delete on API keys, bulk set-pricing on Models). The hand-rolled `components/Table.tsx` and its custom column-resize code are deleted, which closes #403 (the resize / white-gap bug lived in that code; HeroUI owns resizing now).
- **Backend (master-key, standalone):**
  - `DELETE /v1/usage`: delete imported usage rows by an explicit `ids` list or by filter.
  - `POST /v1/usage/set-price`: recompute cost / billing meters / pricing breakdown from manual per-1M input / output / cache-read / cache-write rates against each row's own token counts (reuses `services/metered_pricing.calculate_metered_cost`).
  - Both target rows two ways (`ids` or `by_filter`) and are scoped to imported rows only (`counts_toward_budget = false`), so enforced gateway rows and the spend ledger (`users.spend`) are never touched.
  - Adds `priced` and `counts_toward_budget` read filters on the usage list / count / summary so the toolbar can filter unpriced rows and back the "select all N matching" count.
- **Also included, separate from the framework, at the maintainer's request:** model discovery is now scoped to configured providers only. Previously a provider callable by a credential env var alone (for example `OPENAI_API_KEY` present in the environment) populated the Models page even with no providers configured; now an empty gateway lists no discovered models until a provider is configured in config.yml or via the Providers page.

### Design / QA notes

The pages were dogfooded in a real browser with seeded data (Playwright), which drove a couple of header-styling iterations. The final approach makes `.otari-table` the single card owner rather than wrapping HeroUI's `Table.Root` (which is itself a card, with its own surface, 32px radius, and inner padding) in a second card, so the brand-tint header sits flush inside the card corner on both sides.

### Testing

`make lint` and `make typecheck` are clean. The full unit suite (942) and the web suite (243) pass, along with every integration test covering the changed areas (159: usage admin, usage list/count/summary, external usage events, model discovery, aliases, hybrid surface, config env loading). The entire integration suite was not run end to end in one local process because the sandbox Postgres was OOM-killed under full-suite load; CI runs the complete suite.

## PR Type

- [x] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes #409
Closes #403

## Checklist

- [x] I ran the Definition of Done checks locally (`make lint`, `make typecheck`, `make test`; see the Testing note above for the one caveat on the full integration run).

## AI Usage

- [ ] No AI was used.

The code was generated by Claude Code via back-and-forth with @njbrake. The scope, product decisions (the bulk-action matrix, the discovery-scope choice), and the visual QA feedback that shaped the design are his; the implementation and prose are the AI's.

## Summary

- Introduced a shared, accessible dashboard table system with consistent filtering, URL-synced state, pagination, sorting, selection, resizing, keyboard navigation, and bulk actions.
- Migrated all dashboard table pages to the shared components and removed the legacy table and resize implementation, eliminating the white-gap issue.
- Added bulk delete and manual pricing administration for imported usage records, with gateway and spend-ledger protections.
- Added usage pricing, budget-scope, and source filters across list, count, summary, export, and dashboard views.
- Restricted model discovery to configured providers.
- Added reusable dialogs, selection utilities, pagination controls, custom date ranges, and comprehensive frontend/backend tests.


node_modules/.vite/vitest/da39a3ee5e6b4b0d3255bfef95601890afd80709/results.json
```diff
@@ -0,0 +1 @@
+{"version":"4.1.10","results":[[":web/src/pages/BudgetsPage.test.tsx",{"duration":0,"failed":true}]]}
```
khaledosman: Accidentally committed vitest cache artifact. It also records a stale failing marker (`BudgetsPage.test.tsx` `failed:true`), which contradicts the "web suite 243 pass" note. Remove it and add `node_modules/` to `.gitignore` — `git check-ignore` shows root `node_modules/` isn't currently ignored.

src/gateway/services/usage_admin_service.py
```diff
@@ -0,0 +1,241 @@
...
+# Cap on an explicit id list. Page selections drive the id path and the largest
+# rows-per-page the UI offers is 500; 1000 leaves headroom without letting a single
+# request name an unbounded set of ids.
+_MAX_IDS = 1000
+# Load matched rows for repricing in chunks under SQLite's default bound on bind
+# variables in one IN() (999), mirroring the ingest path.
+_REPRICE_CHUNK = 500
+
+
+class UsageSelection(BaseModel):
```
khaledosman, line 44: `by_filter` selection can drift from the count shown to the operator. The dashboard's "select all N matching" count uses the full `filters` object, but `UsageSelection` here omits `endpoint`, and each page hand-lists which fields it forwards into the selection body. Today they line up so there's no over-deletion, but this is a destructive op: add a filter to a page's count query without mirroring it into both `selectionBody()` and this model, and "select all N" will delete/reprice more rows than the count promised. Consider deriving the by-filter body from the same object the count uses, or asserting field parity, rather than keeping three hand-maintained lists in sync.

src/gateway/services/usage_admin_service.py
```diff
@@ -0,0 +1,241 @@
...
+        pricing_tiers=[],
+    )
+    conditions = _selection_conditions(request)
+    ids = list((await db.execute(select(UsageLog.id).where(*conditions).order_by(UsageLog.id))).scalars().all())
+    result = UsageSetPriceResult(matched=len(ids))
+    if not ids:
+        return result
+
+    try:
+        for start in range(0, len(ids), _REPRICE_CHUNK):
```
khaledosman: Set-price commits per chunk inside the loop, so a `SQLAlchemyError` on a later chunk leaves earlier chunks committed while the caller gets a 500 and a discarded result — a silent partial reprice. Delete, by contrast, is atomic (one statement + commit). Acceptable for an operator tool, but the asymmetry isn't documented; either note it in the docstring or defer the commit to after the loop.

web/src/lib/urlState.ts
```diff
@@ -0,0 +1,76 @@
...
+  /** Apply several key changes in one history entry; "" or the default drops a key. */
+  patch: (updates: Partial<Record<K, string | number>>) => void;
+}
+
+export function useUrlState<K extends string>(defaults: Record<K, string>): UrlState<K> {
+  const [params, setParams] = useSearchParams();
+
+  const get = useCallback((key: K) => params.get(key) ?? defaults[key], [params, defaults]);
+
+  const getNumber = useCallback(
```
khaledosman, line 46: `getNumber` returns 0 on unparseable input, so a hand-edited `?size=abc` yields `pageSize=0` → `limit=0` → server 422 instead of falling back to the default. Minor: clamp to the default when `NaN`.

khaledosman (approved):
Shared table framework + usage-admin endpoints. The imported-only scoping is solid (server pins `counts_toward_budget = False` as a fixed WHERE clause and the frontend intersects the react-aria selection with `selectableKeys` — defense-in-depth on a destructive path), and test coverage is thorough. Inline notes below; none blocking.

🤖 Review by [Claude Code](https://claude.com/claude-code)

njbrake:
> _Note: this comment was drafted by Claude via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

Thanks for the reviews. All the feedback is addressed (pushed in the last few commits), the branch is merged up to latest `main`, and the pages are re-verified in a browser on mobile. Rundown:

**Correctness / safety**
- **Ops could touch budget-exempt gateway rows** (Copilot): the delete / set-price selection is now scoped with `source != "gateway"` on top of `counts_toward_budget = False`, so gateway traffic on a budget-exempt key is never a target. Added a regression test.
- **set-price memory and atomicity** (Copilot, khaledosman): it now keyset-paginates the matched rows (bounded memory) and commits once at the end, so it is all-or-nothing like delete rather than committing per chunk.
- **Summary could disagree with list/count** (Copilot): `/v1/usage/summary` and the CSV export now accept `counts_toward_budget`, threaded through `_summary_context`.
- **urlState.getNumber returned 0 on a non-numeric param** (Copilot, khaledosman, CodeRabbit): it now falls back to the key's default, so a hand-edited `?size=abc` cannot send `limit=0`. Added tests.
- **Selection field drift** (khaledosman): added `endpoint` to the selection model so its filter set matches `_usage_filters`.

**Frontend**
- **Dialogs kept stale form state on reopen** (CodeRabbit): SetPriceDialog and the assign-budget picker reset when they open.
- Added ConfirmDialog tests plus bulk-action tests for Budgets and Users.

**Hygiene**
- **Committed vitest cache artifact** (Copilot, khaledosman): removed `node_modules/.vite/...` from tracking and gitignored root `node_modules/`.

Also folded in the follow-up requests: the branch is merged up to `main` (so the mobile nav drawer from #406 is back), the Activity and Usage filter controls collapse behind a mobile "Filters" toggle again, and the Usage page gained a custom from/to time range like Activity.

Resolving the threads now.

PR #416, opened 2026-07-24
fix(dashboard): stop table selection clicks from re-rendering every row

njbrake:
## Description

Clicking a row checkbox in the dashboard tables could take around a second to render as selected. Measured click-to-selected latency in a production bundle (Playwright, pointerdown to the aria-selected flip): 35 ms at 50 rows per page, ~320 ms at 500 rows, and ~1.2 s at 500 rows on a 4x-throttled CPU. Latency is linear in page size because every click re-rendered the entire table.

Root cause has two layers:

1. `DataTable` built every `Table.Row` as static children. Selection is controlled state in the page component, so each click re-rendered the page and rebuilt all row elements, forcing react-aria to reprocess the whole collection.
2. react-aria-components 1.x `Table` re-renders every mounted row internally on any selection change (table state context, no per-row subscription), costing ~0.6 ms per row per click regardless of what the app does. Still true in 1.19.0.

Changes:

- `DataTable` renders rows through react-aria's items collection, so row elements are cached per row object and a selection toggle no longer re-invokes cell renderers for the page (verified in-browser: 0 cell invocations per click, previously one per row). Callers must keep `columns` and `getRowKey` referentially stable; UsagePage's are hoisted to module scope and the DataTable docstring states the contract.
- Since the react-aria per-row cost is out of our hands, `PAGE_SIZE_OPTIONS` is capped at 100 (was up to 500). Measured at 100 rows: 64-68 ms per click, ~250 ms at 4x throttle. Bulk operations over more rows than a page already go through the "select all N matching this filter" affordance, which is set-scoped rather than id-scoped.
- ActivityPage clamps URL-supplied sizes, so an old `?size=500` bookmark renders 100 rows instead of resurrecting the lag (verified in-browser).
- Regression test: a selection change must not re-invoke cell renderers (fails on the old DataTable, passes now).

If larger pages are wanted later, the remaining path is react-aria's Virtualizer, which needs its own PR with visual QA since HeroUI's table styling was not designed for virtualized positioning.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

None filed; reported directly by the maintainer from a deployed instance.

## Checklist

- [x] I have added or updated tests that cover my change (dashboard change: regression test in `web/src/components/DataTable.test.tsx`, full web suite 279/279 passing; no Python behavior touched).
- [x] I ran the Definition of Done checks locally (`make lint`, `make typecheck` pass; `npm --prefix web run typecheck` and `npm --prefix web test` pass; the committed dashboard bundle was rebuilt).
- [x] Documentation was updated where necessary (behavior contract documented in the DataTable docstring; no user-facing docs cover page sizes).
- [x] If the API contract changed, I regenerated the OpenAPI spec (no API change).

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:** Claude Code (Claude Fable 5)

Investigation, fix, and measurements produced by Claude Code in a sandbox under njbrake's direction; the page-size cap was njbrake's decision from the presented options. Latency numbers come from scripted Playwright runs against a seeded local deployment.


web/src/components/DataTable.tsx
```diff
@@ -111,6 +112,34 @@ export function DataTable<Row>({
   const showSelection = selectionMode === "multiple";
   const Container = resizable ? Table.ResizableContainer : Table.ScrollContainer;
 
+  // Rows render through react-aria's items-collection path so each row element
+  // is cached per row object: a selection toggle re-renders only the affected
+  // row instead of the whole page of rows (which made checkbox clicks lag by
+  // whole seconds at large page sizes). The cache is invalidated when any input
+  // that changes row rendering does (the `dependencies` on Table.Body below),
+  // so callers must keep `columns` (and `rowClassName`, if used) referentially
```
khaledosman: This new stability contract is only honored by ActivityPage and UsagePage. The other five selectable tables — `ModelsPage.tsx:893/951`, `KeysPage.tsx:617/814`, `UsersPage.tsx:392/541`, `AliasesPage.tsx:159/253`, `BudgetsPage.tsx:453/649` — still build `columns` as a fresh array each render and pass inline `getRowKey`/`rowClassName` arrows. For those, `renderRow`'s identity changes every render, so `dependencies={[renderRow]}` invalidates the whole row cache on each selection click and cell renderers still re-invoke for the entire page — the exact app-level cost this PR removes.

Not a regression (matches prior behavior, and all five are paginated at ≤100 rows), but the "0 cell invocations per click" win only lands on two pages. `ModelsPage` is the most exposed given its `PriceCell`/`CachingCell` renderers. Worth either extending the stabilization to those callers or explicitly scoping this PR to Activity/Usage in the description.

web/src/components/DataTable.test.tsx
```diff
@@ -98,6 +102,40 @@ describe("DataTable", () => {
     expect(screen.getByTestId("count")).toHaveTextContent("0");
   });
 
+  it("does not re-render row cells when the selection changes", async () => {
```
khaledosman, line 105: This test uses stable `spiedColumns`/`getRowKey`, so it verifies the caching path works but would still pass if a caller passed unstable columns and got zero benefit. Given that five callers currently violate the contract, a companion test asserting that *unstable* columns *do* re-invoke cell renderers would make the fragility visible in the suite.

khaledosman (approved):
Core approach is sound and well-documented; the row-caching path and the page-size cap both make sense, and the regression test captures the invariant. Two non-blocking notes left inline. Nothing here blocks merge.

Verified as non-issues: the ActivityPage `size` clamp can't produce 0 (getNumber falls back to the default), UsagePage/ModelsPage page size is local state seeded from a ≤100 constant so needs no URL clamp, `Row extends object` is correct for react-aria's items collection, and the new test imports are all present.

PR #417, opened 2026-07-24
fix(dashboard): reopen the request detail inline under the clicked activity row

njbrake:
## Description

Regression from the shared-table migration (#415): clicking a request row on the Activity page looked like it did nothing. Before #415 the request detail expanded inline directly under the clicked row; #415 moved it to a card rendered below the whole table, so on a full page the panel opened far outside the viewport.

`DataTable` gains an inline detail affordance: when `detailKey` matches a row's key, `renderDetail(row)` renders as a full-width, selection-disabled row directly under that row. The panel is interleaved into the react-aria items collection as a sentinel item, so the per-row cache from #416 stays valid for every ordinary row; selection latency is unchanged (re-measured: 35 ms per click at 50 rows, 65 ms at 100). Activating the detail row itself does not fire `onRowAction`, so it cannot re-toggle the panel. ActivityPage renders its request detail (Close button, cache token counts, billed meters) through this instead of the below-table card.

Verified in a scripted browser session against the built bundle: the panel opens adjacent and in-viewport on row click (single `<td colspan>` row, matching the pre-#415 accordion), works on selection-disabled gateway rows, follows the newly clicked row, closes on re-click, and coexists with checkbox selection; no console errors.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

None filed; reported directly by the maintainer from a deployed instance after #416 merged (the regression itself dates to #415).

## Checklist

- [x] I have added or updated tests that cover my change (regression test pinning the adjacent placement in `web/src/components/DataTable.test.tsx`; full web suite 282/282).
- [x] I ran the Definition of Done checks locally (`npm --prefix web run typecheck` and `npm --prefix web test` pass; dashboard bundle rebuilt and committed; no Python touched).
- [x] Documentation was updated where necessary (DataTable docstring documents the new props and their stability requirement).
- [x] If the API contract changed, I regenerated the OpenAPI spec (no API change).

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:** Claude Code (Claude Fable 5)

Root-caused, implemented, and browser-verified by Claude Code under njbrake's direction, from his report that row clicks stopped visibly doing anything.

## Summary

- Fixed Activity page request details so they open inline directly beneath the selected row, with close controls and preserved request metrics.
- Improved table interactions with selection feedback, reliable row-detail activation, animated detail reveals, and reduced-motion support.
- Repositioned the bulk-action bar to avoid layout shifts while scrolling.
- Added regression tests covering inline detail placement, closing details, selection behavior, and detail-panel interactions.
- Rebuilt dashboard assets and updated their references.

## Benefits

Request details are easier to associate with their rows, table interactions feel more responsive, and bulk actions remain accessible without disrupting the page layout.


web/src/components/DataTable.tsx
```diff
@@ -108,9 +155,106 @@ export function DataTable<Row extends object>({
...
+
+  // Host <tr> management: find the target row by its data-key and insert the
+  // host right after it. react-aria commits its real rows in a second render
+  // pass, so the target may not exist yet when this effect first runs (e.g.
+  // mounting with a detailKey already set); the MutationObserver finishes the
+  // insertion as soon as the row appears. The deps re-run it (re-inserting at
+  // the right spot) whenever the row set, order, or target changes, and
+  // cleanup always removes the host, so a vanished target (filtered out, page
+  // flipped) leaves nothing behind.
+  useLayoutEffect(() => {
```
khaledosman, line 180: The detail row is `document.createElement`'d and spliced into react-aria's `<tbody>` with `target.after()`, then React portals into the hand-made `<td>`. This is verified working and React removes DOM by node reference so it's mostly safe, but interleaving a foreign node among keyed react-aria rows is a known footgun: row reordering on refetch, sort changes, concurrent rendering, or a react-aria-components upgrade could orphan/misplace the host or throw on commit. The `MutationObserver` + effect-re-run + cleanup machinery here mitigates it but also shows how much is being fought.

Not a blocker given the verification, but worth a comment noting this couples tightly to RAC's DOM internals and should be revisited if/when RAC gains native expandable-row support.

web/src/components/DataTable.tsx
```diff
@@ -108,9 +155,106 @@ export function DataTable<Row extends object>({
...
+
+  // react-aria's toggle selection behavior repurposes row clicks once the
+  // selection is non-empty: they extend the selection instead of firing the
+  // row action (useSelectableItem's hasPrimaryAction requires an empty
+  // selection manager). For these tables the checkbox owns selection and a row
+  // click must keep opening the drill-in (the Gmail convention), so while a
+  // selection exists, clicks on ordinary data cells are intercepted before the
+  // row's press handler sees them and routed to the row action instead.
+  // Checkboxes, buttons, links, inputs, and the detail panel pass through.
+  const interceptedRowKey = useCallback(
```
khaledosman, line 246: This interception is a global `DataTable` behavior change, not scoped to ActivityPage. For any consumer with `onRowAction` + a non-empty selection, row clicks now route to the row action instead of react-aria's selection-extend. Besides ActivityPage, **ModelsPage** matches (`onRowAction={onSelect}` + `selectionMode="multiple"`): while a bulk model selection is active, clicking a row now fires the pricing-comparison drill instead of toggling that row into the checkbox selection.

That's consistent with the "checkbox owns selection" convention this PR argues for, and likely fine, but it's a silent cross-page change. Worth confirming ModelsPage didn't rely on row-click-to-extend, and a one-line mention in the PR body. (Keys/Users/Aliases/Budgets pass no `onRowAction`, so the `if (!onRowAction) return null` guard leaves them untouched — verified.)

web/src/components/DataTable.tsx
```diff
@@ -1,34 +1,66 @@
...
+  useEffect(() => {
+    if (flash === null) return;
+    const timer = setTimeout(() => setFlash(null), 600);
+    return () => clearTimeout(timer);
+  }, [flash]);
+
+  const showChecked = flash ?? (isSelected || isIndeterminate);
+  return (
+    <span
+      onPointerDown={() => {
```
khaledosman, line 33: `setFlash(!isSelected)` on the shared component means clicking the indeterminate "select all" header flashes it fully-checked; if react-aria resolves that click to *clearing* the selection, `isSelected` never matches `flash` and the checkmark lingers until the 600 ms backstop. Cosmetic only, and per-row checkboxes are unaffected — noting in case it's easy to skip the optimistic flash when `isIndeterminate`.

web/src/components/BulkActionBar.tsx
```diff
@@ -35,7 +41,11 @@ export function BulkActionBar({
     : `${selectedCount.toLocaleString()} selected`;
 
   return (
-    <div className="flex flex-wrap items-center gap-3 rounded-xl border border-[var(--otari-brand)] bg-[var(--otari-brand-tint)] px-4 py-2.5">
+    <div
+      role="toolbar"
```
khaledosman, line 45: `fixed bottom-4 … z-40` floats the bar over whatever sits at the viewport bottom (the paginator, the last rows, or an open detail panel) with no reserved bottom padding while it's shown. On short viewports this can cover the pagination controls the operator needs mid-selection. Consider reserving space (bottom padding/spacer) when the bar is active. `role="toolbar"` + `aria-label` are good additions.

khaledosman (approved):
Solid regression fix with strong test coverage (row caching + its failure mode, adjacency, intercept-with-selection, page-size snapping). Findings below are non-blocking; the main one is a fragility/maintainability note on the imperative DOM insertion, not a defect.

Verified as non-issues: `CSS.escape` guards the selector, the `Math.max(0, page)` + nearest-option `reduce` clamps handle 0/negative/out-of-range sizes (tests pin them), `prefers-reduced-motion` disables both animations, the detail row's `role=presentation` keeps grid semantics correct, and capture-phase `stopPropagation` keeps `onRowAction` firing exactly once.

Issue #422, opened 2026-07-24
Remove keyless-custom-endpoint placeholder workaround once any-llm ships create_openai_compatible


njbrake:
_Note: this issue was drafted by Claude via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

### Background

#421 fixed the "Custom endpoint says the API key is optional but requires one" bug by injecting a placeholder API key (`otari-no-key-required`) whenever an instance points at a custom `api_base`, has no configured key, and the provider's native env var is unset. The logic lives in `keyless_placeholder_api_key()` (`src/gateway/services/provider_kwargs.py`) and is also applied on the "Test connection" path (`src/gateway/services/model_discovery_service.py`).

This is a workaround. any-llm mandates a non-empty key for the `openai` and `anthropic` implementations even when they point at a self-hosted, keyless backend, so we fake one.

### The proper fix

any-llm https://github.com/mozilla-ai/any-llm/pull/1198 adds a first-class custom OpenAI-compatible path, `AnyLLM.create_openai_compatible(name, api_base, api_key=None, ...)`, with keyless tolerance built in (it falls back to `OPENAI_COMPATIBLE_API_KEY`, then to a placeholder). It also reports the endpoint under its own name instead of misreporting it as `openai`. That is the intended way to reach an unlisted gateway.

### What to do here once that lands and is pinned

- Bump the pinned `any-llm-sdk` to the release that includes the custom-endpoint factory.
- Route custom endpoints (`provider_type` of `openai-compatible` or `anthropic-compatible`) through the first-class path instead of injecting `otari-no-key-required` via `get_provider_kwargs`.
- Remove `keyless_placeholder_api_key()`, `_KEYLESS_PLACEHOLDER_API_KEY`, and the call sites in `provider_kwargs.py` and `model_discovery_service.py`, plus their unit tests, if the upstream path makes them redundant.
- Confirm the anthropic-compatible dialect is covered too. PR #1198 is scoped to OpenAI-compatible, so anthropic-compatible may still need the workaround or a separate upstream path.

### Acceptance

- Keyless custom endpoints still add, test, and serve without a key.
- The `otari-no-key-required` placeholder no longer appears in the codebase, or is scoped only to dialects that upstream does not yet cover.

Follow-up to #421. Depends on mozilla-ai/any-llm#1198 (and the rest of mozilla-ai/any-llm#1197).

Triumph1701:
[...]

PR #423, opened 2026-07-24
fix(providers): honor the optional API key for keyless custom endpoints

### LINKED ISSUE #421: Dashboard: Custom Endpoint does required api key even though it says it's optional

njbrake:
## Description

The dashboard's "Custom endpoint" form labels the API key optional, but any-llm's `_verify_and_set_api_key` raises `MissingApiKeyError` for the `openai` and `anthropic` implementations when no key and no native env var are present. A keyless local backend (vLLM, llama.cpp, Ollama) added through the form therefore could neither be tested nor served, contradicting the "optional" label.

This supplies a harmless placeholder key (`otari-no-key-required`) when an instance points at a custom `api_base` and has no key anywhere: no configured key and the provider's native env var (`OPENAI_API_KEY`, `ANTHROPIC_API_KEY`, ...) both empty. The change is purely additive: it only affects the case any-llm would already reject, and it never overrides a real key or the documented env-var fallback. A local server ignores the placeholder; a real one that needs a key rejects it, which is the same failure the operator would already get.

The placeholder is injected in two places so they agree:

- `get_provider_kwargs` (`services/provider_kwargs.py`), the dispatch and stored-provider test path.
- `test_provider_credentials` (`services/model_discovery_service.py`), the pre-save "Test connection" path.

This mirrors any-llm's own keyless tolerance in mozilla-ai/any-llm#1198. When that first-class custom-endpoint path lands and is pinned, this workaround should be removed; tracked in #422.

No frontend change: the form already treats the key as optional, so the "API key (optional)" label is now honest.

## PR Type

- [x] Bug Fix

## Relevant issues

Fixes #421. Follow-up cleanup tracked in #422.

## Checklist

## AI Usage

**AI Model/Tool used:** Claude Opus 4.8, via Claude Code.

Implemented by Claude via back-and-forth with @njbrake, who directed the approach (placeholder for keyless custom endpoints, gated so it never shadows a real key or the env-var fallback) and pointed at the upstream any-llm path as the eventual replacement. Verified locally: ruff, mypy strict, and the full unit suite pass; the keyless flow was reproduced with `OPENAI_API_KEY` unset and confirmed fixed. Integration tests need Docker, which was not available locally, so they run in CI.

## Summary

- Improved custom endpoint support so dashboards can test and serve keyless local backends when the optional API key field is left empty.
- Injects a harmless placeholder API key only when a custom API base is configured but no real API key (and no provider environment-variable key) is available.
- Applies the same credential behavior to both the saved-provider dispatch path and the pre-save “test connection” flow.
- Added unit tests to verify placeholder injection, correct precedence over explicit keys/env vars, and provider-specific environment-variable suppression rules.

### DIFF
--- a/src/gateway/services/model_discovery_service.py
+++ b/src/gateway/services/model_discovery_service.py
@@ -28,7 +28,7 @@ from any_llm.types.model import Model
 
 from gateway.core.config import GatewayConfig
 from gateway.log_config import logger
-from gateway.services.provider_kwargs import get_provider_kwargs
+from gateway.services.provider_kwargs import get_provider_kwargs, keyless_placeholder_api_key
 
 # Fallback bound for ad-hoc credential tests when a caller does not pass one. The
 # route passes ``model_discovery_timeout_seconds`` so the saved and unsaved paths
@@ -397,6 +397,10 @@ async def test_provider_credentials(
             models=[],
             error=f"'{impl_name}' is not a known provider implementation.",
         )
+    # A keyless custom endpoint (api_base set, no key) would otherwise be rejected
+    # by any-llm before the connection is even attempted; supply the same
+    # placeholder the saved path uses so "Test connection" honors the optional key.
+    api_key = api_key or keyless_placeholder_api_key(provider_enum, api_base, api_key)
     try:
         # Bounded like the stored-provider path so a black-holed endpoint cannot
         # hang the test button for the SDK's ~600s default.
--- a/src/gateway/services/provider_kwargs.py
+++ b/src/gateway/services/provider_kwargs.py
@@ -9,6 +9,7 @@ any-llm provider (the common case, no ``provider_type`` declared), the two
 coincide and behavior is identical to splitting the selector directly.
 """
 
+import os
 from dataclasses import dataclass
 from typing import Any
 
@@ -22,6 +23,44 @@ from gateway.services.alias_service import resolve_effective_alias
 # Keys that describe an instance to otari but are not credentials any-llm
 # understands, so they must be stripped before the provider call.
 _INSTANCE_META_KEYS = ("provider_type", "models")
+
+# any-llm rejects a keyless call to most providers (openai, anthropic, ...) with
+# MissingApiKeyError, but a self-hosted OpenAI-/Anthropic-compatible backend
+# (vLLM, llama.cpp, Ollama) usually needs no auth. When an instance points at a
+# custom api_base and supplies no key anywhere, hand any-llm this harmless
+# placeholder so the keyless endpoint is reachable, making good on the
+# dashboard's "API key (optional)" promise for custom endpoints. A local server
+# ignores the value; a real one that does need a key rejects it, which is the
+# same failure the operator would already get. Mirrors any-llm's own keyless
+# tolerance (mozilla-ai/any-llm#1198).
+_KEYLESS_PLACEHOLDER_API_KEY = "otari-no-key-required"
+
+
+def _provider_env_key_present(provider: LLMProvider) -> bool:
+    """Whether the provider's native API-key env var (e.g. OPENAI_API_KEY) is set.
+
+    Read directly from the environment because these are any-llm's own SDK
+    variables, not otari config: any-llm falls back to them before raising, so
+    the placeholder must not shadow a key the operator supplied that way.
+    """
+    try:
+        env_name = AnyLLM.get_provider_class(provider).ENV_API_KEY_NAME
+    except (ImportError, AttributeError):
+        return False
+    return bool(env_name and os.getenv(env_name))
+
+
+def keyless_placeholder_api_key(provider: LLMProvider, api_base: Any, api_key: Any) -> str | None:
+    """Return a placeholder key for a keyless custom endpoint, else ``None``.
+
+    A custom endpoint is one with an ``api_base`` set; it is keyless when no key
+    is configured for it and the provider's native env var is unset. Only that
+    case (which any-llm would otherwise reject) gets a placeholder, so this never
+    overrides a real key or the documented env-var fallback.
+    """
+    if api_base and not api_key and not _provider_env_key_present(provider):
+        return _KEYLESS_PLACEHOLDER_API_KEY
+    return None
 
 
 def get_provider_kwargs(
@@ -68,6 +107,10 @@ def get_provider_kwargs(
             kwargs = {k: v for k, v in provider_config.items() if k != "client_args"}
             if "client_args" in provider_config:
                 kwargs["client_args"] = provider_config["client_args"]
+
+    placeholder = keyless_placeholder_api_key(provider, kwargs.get("api_base"), kwargs.get("api_key"))
+    if placeholder is not None:
+        kwargs["api_key"] = placeholder
 
     return kwargs
 
--- a/src/gateway/services/provider_kwargs.py
+++ b/src/gateway/services/provider_kwargs.py
@@ -42,12 +42,19 @@ def _provider_env_key_present(provider: LLMProvider) -> bool:
     Read directly from the environment because these are any-llm's own SDK
     variables, not otari config: any-llm falls back to them before raising, so
     the placeholder must not shadow a key the operator supplied that way.
+
+    ``ENV_API_KEY_NAME`` is not always a single variable name: some providers use
+    a ``/``-joined label listing alternatives (e.g. gemini's
+    ``"GEMINI_API_KEY/GOOGLE_API_KEY"``), so treat it as candidates and count the
+    key present when any one of them is set.
     """
     try:
         env_name = AnyLLM.get_provider_class(provider).ENV_API_KEY_NAME
     except (ImportError, AttributeError):
         return False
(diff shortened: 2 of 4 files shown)


njbrake:

@tbille tagging you on this one. It fixes #421 (keyless custom endpoints were rejected despite the "API key optional" label). CI is green and CodeRabbit had no findings. Feel free to merge once you're happy with it. Follow-up cleanup for when any-llm ships the first-class path is tracked in #422.

src/gateway/services/provider_kwargs.py
```diff
@@ -23,6 +24,44 @@
...
+
+def _provider_env_key_present(provider: LLMProvider) -> bool:
+    """Whether the provider's native API-key env var (e.g. OPENAI_API_KEY) is set.
+
+    Read directly from the environment because these are any-llm's own SDK
+    variables, not otari config: any-llm falls back to them before raising, so
+    the placeholder must not shadow a key the operator supplied that way.
+    """
+    try:
+        env_name = AnyLLM.get_provider_class(provider).ENV_API_KEY_NAME
```
khaledosman, line 52: `ENV_API_KEY_NAME` is a human-readable label for several providers, not always an env var name:

| provider | value |
|---|---|
| gemini | `"GEMINI_API_KEY/GOOGLE_API_KEY"` |
| sagemaker | `"AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY"` |
| vertexai, vertexaianthropic | `""` |
| llamacpp, llamafile | `"None"` |

Concrete regression: with `providers: {gemini: {api_base: "https://proxy/…"}}` and `GEMINI_API_KEY` set, `os.getenv("GEMINI_API_KEY/GOOGLE_API_KEY")` is always `None`, so the placeholder is injected and wins over gemini's own `_verify_and_set_api_key`, which reads `GEMINI_API_KEY` / `GOOGLE_API_KEY` separately (`any_llm/providers/gemini/gemini.py:22-28`). A working config starts returning 401 — exactly the env-var fallback the docstring below promises not to shadow.

src/gateway/services/provider_kwargs.py
```diff
@@ -23,6 +24,44 @@
...
+
+def keyless_placeholder_api_key(provider: LLMProvider, api_base: Any, api_key: Any) -> str | None:
+    """Return a placeholder key for a keyless custom endpoint, else ``None``.
+
+    A custom endpoint is one with an ``api_base`` set; it is keyless when no key
+    is configured for it and the provider's native env var is unset. Only that
+    case (which any-llm would otherwise reject) gets a placeholder, so this never
+    overrides a real key or the documented env-var fallback.
+    """
+    if api_base and not api_key and not _provider_env_key_present(provider):
```
khaledosman, line 68: The gate is much broader than "any-llm would reject this". Only providers inheriting the base `_verify_and_set_api_key` raise `MissingApiKeyError`; 14 override it, either to relax the requirement or to run their own credential resolution (SigV4 session creds, GCP ADC, Azure AD token): bedrock, sagemaker, azure, azureopenai, azureanthropic, vertexai, vertexaianthropic, gemini, ollama, lmstudio, vllm, llamacpp, llamafile, cascadia.

Gating on the real invariant is narrower and fixes the gemini case for free, because for base-impl providers `ENV_API_KEY_NAME` is by construction the var the base reads:

```python
cls = AnyLLM.get_provider_class(provider)
if cls._verify_and_set_api_key is not AnyLLM._verify_and_set_api_key:
    return None  # provider does its own keyless / credential handling
```

Worth adjusting the PR description too: `vllm`, `ollama`, `llamacpp`, and `lmstudio` already tolerate keyless calls, so the motivating case is really `provider_type: openai` / `anthropic` (including the `*-compatible` aliases in `config.py:22-27`) — which is precisely what the narrow gate keeps.

src/gateway/services/provider_kwargs.py
```diff
@@ -23,6 +24,44 @@
...
+# custom api_base and supplies no key anywhere, hand any-llm this harmless
+# placeholder so the keyless endpoint is reachable, making good on the
+# dashboard's "API key (optional)" promise for custom endpoints. A local server
+# ignores the value; a real one that does need a key rejects it, which is the
+# same failure the operator would already get. Mirrors any-llm's own keyless
+# tolerance (mozilla-ai/any-llm#1198).
+_KEYLESS_PLACEHOLDER_API_KEY = "otari-no-key-required"
+
+
+def _provider_env_key_present(provider: LLMProvider) -> bool:
```
khaledosman, line 39: This duplicates logic that already exists in `provider_metadata_service.known_provider_detail` (`:209-218`): it resolves `env_key` via `get_provider_metadata()`, trims it, treats the literal `"None"` as absent, and computes `env_key_present` as `bool((os.getenv(env_key) or "").strip())`.

Two divergent notions of "env key present" now exist — a whitespace-only value counts as present here and absent there — and the dashboard's "API key (optional)" label is driven by the metadata one. Worth factoring a single helper so the form's promise and the runtime behavior cannot drift.

src/gateway/services/provider_kwargs.py
```diff
@@ -23,6 +24,44 @@
...
+    the placeholder must not shadow a key the operator supplied that way.
+    """
+    try:
+        env_name = AnyLLM.get_provider_class(provider).ENV_API_KEY_NAME
+    except (ImportError, AttributeError):
+        return False
+    return bool(env_name and os.getenv(env_name))
+
+
+def keyless_placeholder_api_key(provider: LLMProvider, api_base: Any, api_key: Any) -> str | None:
```
khaledosman, line 60: Nit: `api_base: Any, api_key: Any` — both are only tested for truthiness, so `object` is truer and keeps mypy useful at the call sites. `Any` here silences checking for no gain.

src/gateway/services/provider_kwargs.py
```diff
@@ -69,6 +108,10 @@ def get_provider_kwargs(
             if "client_args" in provider_config:
                 kwargs["client_args"] = provider_config["client_args"]
 
+    placeholder = keyless_placeholder_api_key(provider, kwargs.get("api_base"), kwargs.get("api_key"))
```
khaledosman, line 118: Diagnosability: the gate is "has an `api_base`", which also covers a hosted provider fronted by a proxy where a missing key is a plain misconfiguration. That case used to fail with `MissingApiKeyError` naming the env var; it now fails as a provider 401 truncated by `_short_error`. A `logger.debug` when the placeholder is injected would make it recoverable from logs.

src/gateway/services/model_discovery_service.py
```diff
@@ -397,6 +397,10 @@ async def test_provider_credentials(
             models=[],
             error=f"'{impl_name}' is not a known provider implementation.",
         )
+    # A keyless custom endpoint (api_base set, no key) would otherwise be rejected
+    # by any-llm before the connection is even attempted; supply the same
+    # placeholder the saved path uses so "Test connection" honors the optional key.
+    api_key = api_key or keyless_placeholder_api_key(provider_enum, api_base, api_key)
```
khaledosman, line 403: The `or` is redundant and slightly misleading: `keyless_placeholder_api_key` already returns `None` when `api_key` is truthy, so the key is passed twice and the expression reads as if the helper could override it. Mirror the `provider_kwargs.py:111-113` shape, or `api_key = keyless_placeholder_api_key(provider_enum, api_base, api_key) or api_key`.

tests/unit/test_provider_instances.py
```diff
@@ -189,6 +191,57 @@ def test_get_provider_kwargs_strips_provider_type_and_models() -> None:
...
+    assert "api_key" not in kwargs
+
+
+def test_env_var_fallback_preserved_over_placeholder(monkeypatch: pytest.MonkeyPatch) -> None:
+    # any-llm falls back to OPENAI_API_KEY before raising, so the placeholder must
+    # not shadow a key the operator supplied that way.
+    monkeypatch.setenv("OPENAI_API_KEY", "sk-from-env")
+    config = GatewayConfig(providers={"home_lab": {"provider_type": "openai", "api_base": "http://x/v1"}})
+    kwargs = get_provider_kwargs(config, LLMProvider.OPENAI, instance="home_lab")
+    assert "api_key" not in kwargs
```
khaledosman, line 232: This asserts the invariant for `openai`, where `ENV_API_KEY_NAME` happens to be a real variable name — the one provider family where the guard cannot fail. Two cases would cover the gap:

- `gemini` + `api_base` + `GEMINI_API_KEY` set ⇒ no placeholder (fails today, see the comment on `provider_kwargs.py:47`).
- one env-auth provider (`bedrock` or `azureopenai`) + `api_base` ⇒ no placeholder.

khaledosman (changes requested):
Approach is sound and the two injection points agree, but the env-var guard reads `ENV_API_KEY_NAME` as if it were always a real variable name, which it isn't for several providers — so for `gemini` the placeholder shadows the env key the docstring promises to preserve. Details inline; narrowing the gate to providers that actually inherit any-llm's base `_verify_and_set_api_key` fixes that and shrinks the blast radius.

khaledosman (approved):

PR #434, opened 2026-07-26
feat(dashboard): bundle the operator user guide and serve it at /#/docs

### LINKED ISSUE #329: Dashboard: bundle docs with dashboard server
When a user deploys otari, they may be deploying a version of otari that deviates from the latest or from what is running on otari.ai. Because of this the dashboard should have a /docs page (behind authentication) that renders the most recent user guide for the dashboard. Then there should be a link in the bottom of the left sidebar that helps direct a user to the guide. See #313 first for development of docs.

njbrake:

## Description

Fixes #329.

A self-hosted gateway can run a version of Otari that differs from otari.ai, so the dashboard should carry the guide that matches the running build rather than sending operators to a separate docs site that may describe a different version.

What this does:

- New docs page at `/#/docs` (behind the existing master-key sign-in) that renders `docs/dashboard.md` inline, with a "User guide" link in the sidebar footer.
- The guide is imported with Vite's `?raw`, so it inlines into the lazy-loaded docs chunk at build time and stays version-matched to the bundle. The chunk loads only when the page is visited, so the shell stays small.
- Rendering uses `react-markdown` + `remark-gfm` (the guide has GFM tables and fenced code). Prose is styled with the existing `--otari-*` tokens via a scoped `.otari-markdown` stylesheet, mirroring the `.otari-table` pattern (no typography plugin added).
- Sibling doc links that are not bundled here (for example `configuration.md`) are rewritten to the rendered source on GitHub and open in a new tab; in-page anchors stay in the tab.
- The guide's own leading title is dropped so the page shows a single top-level heading ("User guide"), with the guide's intro flowing under it.

Frontend only: `/#/docs` is a client-side `HashRouter` route served by the existing `index.html`, so no backend route was needed. Because the guide lives outside `web/`, `web/vite.config.ts` grants `server.fs` access to the repo root, and the dashboard staleness workflow now also triggers on `docs/dashboard.md` so an edit that leaves the committed bundle stale is caught in CI.

## PR Type

- [x] New Feature

## Relevant issues

Fixes #329

## Checklist

- [x] I have added or updated tests that cover my change (`web/src/pages/DocsPage.test.tsx`, `web/src/components/AppShell.test.tsx`; this is a `web/` change, so the JS suite covers it rather than `tests/unit`).
- [x] I ran the Definition of Done checks locally (`npm --prefix web run typecheck`, `npm --prefix web test`, `npm --prefix web run build`; no Python changed).
- [x] Documentation was updated where necessary (AGENTS.md notes the new guide bundling and its rebuild coupling).
- [x] If the API contract changed, I regenerated the OpenAPI spec (no API contract change).

## AI Usage

**Any additional AI details you'd like to share:** Implemented via back-and-forth with @njbrake; the reasoning and decisions are his.

## Summary

- Added an authenticated `/docs` dashboard page that embeds the version-matched user guide.
- Added a sidebar link to the guide and improved accessibility for navigation elements.
- Added Markdown/GFM rendering, responsive tables, accessible code blocks, scoped styling, and GitHub links for related documentation.
- Updated build configuration and workflow checks so guide changes rebuild or flag stale dashboard assets.
- Regenerated bundled dashboard assets and added tests covering rendering, links, tables, and accessibility.

## Benefits

Users can access the deployed version’s guide directly from the dashboard without leaving the application, while related documentation remains easy to discover and the rendered content is more accessible.


web/vite.config.ts
```diff
@@ -42,6 +49,9 @@ export default defineConfig({
     },
   },
   server: {
+    // Allow reading the bundled user guide (docs/dashboard.md) from the repo
+    // root; the default only permits web/ and its workspace.
+    fs: { allow: [repoRoot] },
```
khaledosman: **Blocker: this exposes `config.yml` (and `otari.db`) over the dev server.**

An explicit `server.fs.allow` *replaces* Vite's default rather than extending it, and the default `fs.deny` only covers `.env*`, `*.{crt,pem}`, `**/.git/**`. Widening the root to the whole repo makes every gitignored file readable at `/@fs/<repoRoot>/...`.

Reproduced on this branch with a sentinel `config.yml`:

```
pr-434:  curl /@fs/<repo>/config.yml  ->  HTTP 200   master_key: sk-SENTINEL-SECRET-123
main:    curl /@fs/<repo>/config.yml  ->  HTTP 403
```

Anyone who can reach the port gets the master key and provider API keys: `npm run dev -- --host`, a devcontainer, or a forwarded port in Codespaces.

Only `docs/` needs to be readable. `web/` has to stay listed explicitly, since the default is replaced:

```ts
fs: {
  allow: [
    fileURLToPath(new URL("./", import.meta.url)),
    fileURLToPath(new URL("../docs", import.meta.url)),
  ],
},
```

web/src/pages/DocsPage.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+const markdownComponents: Components = {
+  a: ({ href, children, ...props }: AnchorHTMLAttributes<HTMLAnchorElement>) => {
+    const resolved = resolveDocHref(href);
+    // In-page anchors stay in the tab; every other link (bundled-out doc pages
+    // rewritten to GitHub, or already-external URLs) opens in a new tab so the
+    // operator does not lose the dashboard.
+    const external = href?.startsWith("#") ? false : isExternal(resolved);
+    return (
+      <a href={resolved} {...(external ? { target: "_blank", rel: "noreferrer" } : {})} {...props}>
```
khaledosman: **Every link in the guide ships an invalid `node="[object Object]"` attribute.**

react-markdown 9 passes the hast `node` to custom components. Destructuring only `href` and `children` leaves `node` in `...props`, and it gets forwarded onto the DOM element. Verified by rendering the page:

```
ANCHORS: 8   WITH_NODE_ATTR: 8
<a href="https://github.com/mozilla-ai/otari/blob/main/docs/configuration.md"
   target="_blank" rel="noreferrer" node="[object Object]">Configuration</a>
```

Typecheck does not catch it because the parameter is annotated `AnchorHTMLAttributes<HTMLAnchorElement>` rather than `Components["a"]`, so TS never sees the extra prop. That also stringifies the whole hast subtree once per link.

```ts
a: ({ node: _node, href, children, ...props }: ComponentPropsWithoutRef<"a"> & ExtraProps) => {
```

(`ExtraProps` is exported from `react-markdown`.) Worth an assertion in `DocsPage.test.tsx` that anchors carry no `node` attribute.

web/src/pages/DocsPage.tsx
```diff
@@ -0,0 +1,72 @@
...
+// happens to start with `#` is left alone.
+const guideBody = dashboardGuide.replace(/^#{1,6}[ \t]+.*\r?\n+/, "");
+
+const markdownComponents: Components = {
+  a: ({ href, children, ...props }: AnchorHTMLAttributes<HTMLAnchorElement>) => {
+    const resolved = resolveDocHref(href);
+    // In-page anchors stay in the tab; every other link (bundled-out doc pages
+    // rewritten to GitHub, or already-external URLs) opens in a new tab so the
+    // operator does not lose the dashboard.
+    const external = href?.startsWith("#") ? false : isExternal(resolved);
```
khaledosman: In-page anchors are kept in-tab here, but HashRouter owns `location.hash`, so this is the one link shape that breaks.

A `](#the-two-key-model)` link renders as `<a href="#the-two-key-model">`; clicking it rewrites `#/docs` to `#the-two-key-model`, react-router parses the pathname as `/the-two-key-model`, nothing matches, and the `path="*"` route in `App.tsx:77` redirects the operator to Overview. Reproduced against the real route table under Vitest.

Compounding it, there is no `rehype-slug`, so headings carry no `id` for such a link to target even under BrowserRouter.

Latent today (the guide has no `](#` links), but remark-gfm footnotes generate `#user-content-fn-*` anchors that hit the same path. Either rewrite `#frag` to the GitHub source like the sibling doc links, or add `rehype-slug` and `preventDefault()` + `scrollIntoView` on click.

web/src/styles/globals.css
```diff
@@ -118,6 +121,133 @@
...
+.otari-markdown hr {
+  margin: 1.5rem 0;
+  border: 0;
+  border-top: 1px solid var(--otari-line);
+}
+.otari-markdown table {
+  display: block;
+  width: max-content;
+  max-width: 100%;
+  overflow-x: auto;
```
khaledosman: **`display: block` on a `<table>` strips its table semantics.**

Overriding the display removes the table box from the layout tree, and engines that derive a11y exposure from layout (WebKit definitively) drop `role=table` along with the row/cell/columnheader relationships on descendants. The guide's two-key table (`docs/dashboard.md:19-25`) then reaches VoiceOver as an undifferentiated run of text with no header-to-cell association, on a table whose third column states that losing `OTARI_SECRET_KEY` makes every stored provider key undecryptable.

`DocsPage.test.tsx:33` cannot catch this: jsdom applies no stylesheet, so `getAllByRole("table")` resolves the implicit role from the bare `<table>` tag and passes either way.

Keep the native display and move the scroll to a wrapper via a `components.table` override in `DocsPage.tsx`:

```tsx
table: ({ node: _node, ...props }) => (
  <div className="otari-markdown-scroll" tabIndex={0} role="region" aria-label="Table">
    <table {...props} />
  </div>
),
```

with `.otari-markdown-scroll { overflow-x: auto; max-width: 100%; }`.

web/src/styles/globals.css
```diff
@@ -118,6 +121,133 @@
...
+  color: var(--otari-brand-dark);
+  padding: 0.1rem 0.35rem;
+  border-radius: 0.35rem;
+}
+.otari-markdown pre {
+  margin: 1rem 0;
+  padding: 0.85rem 1rem;
+  background-color: var(--otari-ink);
+  border-radius: 0.6rem;
+  overflow-x: auto;
```
khaledosman, line 208: Scrollable region with no keyboard access (axe `scrollable-region-focusable`, WCAG 2.1.1): this `pre` and the `table` rule below both set `overflow-x: auto` but get no `tabindex` and contain no focusable descendants, so a keyboard-only or switch user tabs straight past them and can never reach the clipped content.

The table is the unconditional case (`width: max-content` sizes the two-key table to ~1200px, then clips); the `pre` blocks only overflow on narrow viewports. Add `tabIndex={0}` plus `role="region"` and an `aria-label` on the scroll containers via `markdownComponents` overrides, alongside the table wrapper fix.

web/src/styles/globals.css
```diff
@@ -118,6 +121,133 @@
...
+  margin: 0.75rem 0;
+}
+.otari-markdown a {
+  color: var(--otari-brand-dark);
+  text-decoration: underline;
+  text-underline-offset: 2px;
+}
+.otari-markdown a:hover {
+  color: var(--otari-brand);
+}
```
khaledosman, line 176: Hover drops link contrast below AA: `--otari-brand-dark` #3c6678 is 6.2:1 on the white card, `--otari-brand` #4e8295 is **4.24:1**. `.otari-markdown` sets `font-size: 0.875rem` at normal weight, so WCAG 1.4.3 requires 4.5:1. Contrast falls below threshold in exactly the state meant to emphasise the link.

Keep `--otari-brand-dark` and signal hover another way (`text-decoration-thickness`, or a `--otari-brand-tint` background), or darken rather than lighten.

khaledosman (changes requested):
Reviewed the source diff (bundle excluded). One blocker (dev-server secret exposure), one confirmed rendering bug, and three a11y issues on the new Markdown styles. I reproduced the first two by running the code; details inline.

Also: this conflicts with #435 in `web/src/components/AppShell.tsx` and `AppShell.test.tsx` (both rewrite the sidebar footer, both insert tests at the same anchor). Whoever merges second should take this PR's footer `<div>` structure, re-apply `aria-hidden` to **both** svgs inside it (this PR's new book icon never had one, and resolving in favour of this block drops #435's addition on the otari.ai icon), keep both test blocks, run `npm ci` before rebuilding, and regenerate the bundle from scratch.

Not worth blocking on: the `DocsPage` lazy chunk is 168 KB for react-markdown + remark-gfm + the inlined guide. It is genuinely lazy, so it only costs visitors of `/#/docs`.

---

web/src/pages/DocsPage.tsx
```diff
@@ -0,0 +1,129 @@
...
+//      in) is circular here. Drop it; it stays in docs/dashboard.md for readers
+//      on GitHub and the /welcome tutorial, and its post-sign-in substance is
+//      covered by the page-by-page reference below.
+const guideBody = dropSection(
+  dashboardGuide.replace(/^#{1,6}[ \t]+.*\r?\n+/, ""),
+  "First-run walkthrough",
+);
+
+// react-markdown passes each node's hast `node` to custom components; it must be
+// destructured out of the DOM spread or it renders as node="[object Object]" on
```
khaledosman, line 79: Nit: `role="region"` with an accessible name is a **landmark**, so the guide now injects four of them into the landmark list, three identically named:

```
landmark regions: 4
pres: role=region,region,region   labels: Code,Code,Code
```

Navigating by landmark gives "Code, Code, Code, Table" with nothing to tell them apart.

`scrollable-region-focusable` only requires `tabindex`; the role is what makes them landmarks. `role="group"` keeps the name announced when focus lands without polluting landmark navigation:

```diff
-  table: ({ node: _node, ...props }: MdProps<"table">) => (
-    <div className="otari-markdown-scroll" tabIndex={0} role="region" aria-label="Table">
+  table: ({ node: _node, ...props }: MdProps<"table">) => (
+    <div className="otari-markdown-scroll" tabIndex={0} role="group" aria-label="Table">
```

Same for the `pre` override.

web/src/pages/DocsPage.test.tsx
```diff
@@ -0,0 +1,81 @@
...
+    expect(h1s[0]).toHaveTextContent("User guide");
+  });
+
+  it("renders GFM tables from the guide, keeping table semantics inside a focusable scroll region", () => {
+    render(<DocsPage />);
+
+    // The two-key model is a Markdown table; rendering it as a real <table>
+    // proves remark-gfm is wired up (plain Markdown would leave it as text).
+    const tables = screen.getAllByRole("table");
+    expect(tables.length).toBeGreaterThan(0);
```
khaledosman, line 46: Nit: the new coverage for the table region and the `node` leak is good, but two behaviours added in the fix commit are untested:

- the `#frag` → `dashboard.md#frag` rewrite, which is the case that originally broke under HashRouter
- `pre` focusability (the table wrapper is asserted, the code blocks are not)

Both extend this test naturally, e.g. asserting an `#anchor` href resolves to `https://github.com/mozilla-ai/otari/blob/main/docs/dashboard.md#…` and that every `<pre>` carries `tabindex="0"`.

khaledosman (approved):
Re-reviewed at `30340fa`. All six findings from my last pass are fixed, and I re-ran each check rather than reading the diff:

| Finding | Verification |
|---|---|
| `server.fs.allow` served repo secrets | Live dev server on this branch: `config.yml` → **403** (was 200, master key in the body), `pyproject.toml` → 403, `docs/dashboard.md?raw` → 200, app root → 200. |
| `node="[object Object]"` on every anchor | Rendered page: **0** elements carry a `node` attribute (was 8/8 anchors). |
| `display: block` stripped table semantics | `<table>` keeps its native display and renders inside `.otari-markdown-scroll`. |
| Scroll regions unreachable by keyboard | Table wrapper and all 3 `<pre>` now `tabindex="0"`. No global outline reset in the built CSS (the one `outline:none` is scoped to a toast button), so the UA focus ring shows. |
| `#frag` links broke HashRouter | Rewritten to `…/docs/dashboard.md#frag`; all 8 links now resolve to absolute GitHub URLs with `target="_blank"`. |
| Hover contrast 4.24:1 | `text-decoration-thickness` instead, keeping `--otari-brand-dark` at 6.2:1. |

`tsc -b --noEmit` clean, 291/291 tests pass, and a fresh `npm run build` leaves `src/gateway/static/dashboard/` byte-identical, so the committed bundle is current.

The `aria-hidden` added on the book icon also clears the merge hazard I flagged against #435: both svgs now carry it on either side, so resolving in favour of this PR's footer block no longer silently reverts that a11y fix.

Two nits inline, neither blocking.

---

PR #435, opened 2026-07-26
fix(dashboard): accessibility and UX polish pass

### LINKED ISSUE #312: Dashboard: accessibility, responsive, and UX polish

njbrake:
## Description

Cross-cutting accessibility, loading, and empty-state polish for the admin dashboard. Started from a survey of all 11 pages so the changes are consistent rather than one-off.

**Accessibility**
- Skip-to-content link as the first tab stop in `AppShell`, so keyboard users can jump past the nav to the page body. It moves focus to the main region programmatically (the region carries `tabIndex={-1}`) rather than via an anchor, because `HashRouter` owns the URL hash and an `href="#main-content"` would register as a route change.
- Decorative sidebar and header SVG icons marked `aria-hidden`, matching the convention already used elsewhere, so they are not announced beside their visible text labels.

**Loading states**
- Shared `PageLoading` indicator (`role="status"`) on the pages that previously rendered a bare header over blank space while their first fetch resolved: the Overview index (was returning `null`), Settings, and Tools & Guardrails.

**Empty states**
- Shared `EmptyState` component replacing four near-identical hand-rolled onboarding panels (Keys, Users, Budgets, Usage), so empty pages read the same way.

Items that needed no code change: secret fields already mask via native `type="password"` (no `-webkit-text-security` in use), and routing stays `HashRouter` (rationale documented in AGENTS.md: avoids shadowing `/v1` and the asset paths, needs no server catch-all).

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes #312

## Checklist

- [x] I have added or updated tests that cover my change. (Colocated Vitest tests under `web/src`; this PR touches no Python, so no `tests/unit` or `tests/integration` changes.)
- [x] I ran the Definition of Done checks locally. (Dashboard checks: `npm --prefix web run typecheck`, `npm --prefix web test`, `npm --prefix web run build`; no Python changed, so the `make` targets are not applicable.)
- [x] Documentation was updated where necessary. (No user-facing docs needed changes.)
- [ ] If the API contract changed, I regenerated the OpenAPI spec. (N/A: no API contract change; dashboard-only.)

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

Claude Code (Opus 4.8)

Implemented and reviewed by Claude Code via back-and-forth with @njbrake; the reasoning and decisions are his, the code and prose are Claude's.

## Summary

- Improved dashboard accessibility with a keyboard-accessible skip link and clearer treatment of decorative icons.
- Added consistent loading indicators to Overview, Settings, and Tools & Guardrails.
- Standardized empty states across Keys, Users, Budgets, and Usage, including cleaner first-time onboarding displays.
- Added focused tests covering accessibility, loading, empty states, and onboarding behavior.
- Refreshed generated dashboard assets to reflect the updated frontend build.

## Benefits

These changes provide clearer feedback during loading, more consistent guidance for empty pages, and improved keyboard and screen-reader usability across the admin dashboard.

## Technical notes

Routing remains based on `HashRouter`; native password masking, responsive layouts, API contracts, and Python code are unchanged.


web/src/components/AppShell.tsx
```diff
@@ -367,6 +384,16 @@ export function AppShell() {
...
+      {/* The first tab stop: a keyboard user can jump straight to the page body
+          instead of tabbing through the whole nav on every route. Visually hidden
+          until focused, then pinned top-left over the header (z above it). */}
+      <button
+        type="button"
+        onClick={skipToMain}
+        className="sr-only focus:not-sr-only focus:absolute focus:top-3 focus:left-3 focus:z-50 focus:rounded-lg focus:border focus:border-[var(--otari-brand)] focus:bg-[var(--otari-surface)] focus:px-4 focus:py-2 focus:text-sm focus:font-medium focus:text-[var(--otari-brand-dark)] focus:shadow-md focus:outline-none"
+      >
+        Skip to main content
+      </button>
```
khaledosman, line 400: The skip link stays live while the mobile drawer is open, but its action is a no-op there.

The button is a sibling of `<header inert={backgroundInert}>` and `<main inert={backgroundInert}>`, so it never picks up `inert` itself. With the drawer open it remains in the tab order and the accessibility tree, and it is first in DOM order, so it is the one background control an AT virtual cursor reaches ahead of the modal drawer. Activating it calls `mainRef.current?.focus()` on a `main` that is inert, and per spec `focus()` on an inert element is ignored, so nothing happens.

That contradicts the invariant the comment at 379-383 states about `inert` keeping the virtual cursor out of obscured controls.

```diff
       <button
         type="button"
+        inert={backgroundInert}
         onClick={skipToMain}
```

The new test at `AppShell.test.tsx:171` only runs `mockMatchMedia(false)`, so the drawer-open case is uncovered; worth a `mockMatchMedia(true)` case asserting the button is inert.

khaledosman (approved):
Good consistency pass. One a11y gap in the skip link (inline) that is a one-line fix, nothing blocking.

I chased and could not substantiate three things worth naming so they do not get re-raised: `EmptyState`'s `children`/`isActionDisabled` are legitimate optional API rather than dead surface; the orphaned `// ---------- onboarding ----------` banners are cosmetic; and the new `PageLoading` on Settings does not double up with the empty-state copy, since `"No settings match your search"` is guarded by `data &&`.

One coordination note: this conflicts with #434 in `web/src/components/AppShell.tsx` and `AppShell.test.tsx` (both rewrite the sidebar footer, both insert tests at the same anchor). #434's footer is structurally a superset, so resolving in its favour silently reverts the `aria-hidden` added here on the otari.ai icon, and #434's new book icon never had one. The blanket svg assertion added in this PR would catch that, but only if it survives its own conflict in the same file. Whoever merges second should take #434's footer `<div>` and re-apply `aria-hidden` to both svgs inside it.

---

PR #439, opened 2026-07-27
docs: explain how to use Otari with a ChatGPT Plus/Pro subscription

### LINKED ISSUE #436: docs: explain how to use Otari with a ChatGPT Plus/Pro subscription

## Summary

Add a `docs/` recipe for using Otari with a ChatGPT Plus/Pro subscription.

Users keep asking whether they can use Otari when they have a ChatGPT subscription rather than an OpenAI Platform API key. It works, but the setup is not obvious, so it should be documented.

## Background

A ChatGPT subscription has no standard API key; ChatGPT subscriptions and the API platform are separate billing systems. The subscription is only reachable programmatically through Codex OAuth, so Otari cannot point at it directly.

The working path: run a local Codex-OAuth proxy (for example Hermes' own proxy, `codex-proxy`, or `CLIProxyAPI`/`ai-cli-proxy-api`) that logs in with the subscription and exposes a standard OpenAI-compatible endpoint on localhost. Otari then registers it as an `openai-compatible` provider with `api_base: http://localhost:PORT/v1`, either using the proxy's downstream key or keyless (the #423 keyless-custom-endpoint fix makes this clean). Budgets, usage logging, and key management all layer on top unchanged.

## Deliverable

- [ ] A `docs/` page walking through pointing Otari at a local Codex-OAuth proxy as a custom endpoint.
- [ ] Note the caveat that this rides an unofficial ChatGPT backend (can break, subscription usage caps apply), and that an API key is the sturdier route.

njbrake:
## Description

A ChatGPT Plus/Pro subscription has no OpenAI Platform API key and is only reachable programmatically through Codex OAuth, so users keep asking whether Otari can front it. It can, via a local Codex-OAuth proxy that re-exposes the subscription as an OpenAI-compatible endpoint, but the setup was undocumented.

This adds `docs/chatgpt-subscription.md`, a recipe that walks through:

- Why the extra hop is needed (subscription vs Platform API, no `sk-` key, Codex OAuth only), with an `app -> Otari -> Codex-OAuth proxy -> Codex OAuth` diagram.
- Running a Codex-OAuth proxy (CLIProxyAPI / `ai-cli-proxy-api`, Hermes' `codex-proxy`) and sanity-checking it.
- Registering it as a named `openai-compatible` provider instance with `api_base: http://localhost:PORT/v1`, both with a downstream key and keyless (the #423 keyless-custom-endpoint fix makes the keyless path clean).
- Optional counterfactual pricing so cost analytics show API-equivalent spend.
- The `log_writer_strategy: batch` note for streaming clients (inline logging can drop usage rows when a client disconnects at SSE `[DONE]`), surfaced by @jtyun in the issue.
- Caveats: unofficial backend, subscription caps, an API key is the sturdier route, keep the proxy local.

It also links the new page from the docs index under "For operators".

## PR Type

- [ ] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [x] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes #436

## Checklist

- [ ] I have added or updated tests that cover my change (`tests/unit`, `tests/integration`). Docs-only change; no code paths to test.
- [x] I ran the Definition of Done checks locally (`make lint`, `make typecheck`, `make test`). N/A for docs-only; no docs-lint CI exists, and all internal doc links resolve.
- [ ] If the API contract changed, I regenerated the OpenAPI spec (`uv run python scripts/generate_openapi.py`). No API change.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:** Claude Code (Opus 4.8), drafted via back-and-forth with @njbrake.

**Any additional AI details you'd like to share:** Prose drafted by Claude; the structure, sources, and caveats reflect @njbrake's reasoning and the discussion on #436.

## Summary

- Added a new documentation page, `docs/chatgpt-subscription.md`, explaining how to route ChatGPT Plus/Pro model traffic through Otari using a local Codex-OAuth compatibility proxy (including provider setup, optional cost analytics, streaming/usage logging guidance, and important caveats/risks).
- Linked the new “Use with a ChatGPT subscription” guide from `docs/index.md` under “For integrators.”
- Updated `docs/configuration.md` to clarify `log_writer_strategy` behavior—especially the tradeoffs for streaming usage logging (`single` vs `batch`).

This helps users integrate Otari with an existing ChatGPT subscription while still using Otari’s usage logging and budgeting features (with clear warnings about the unofficial backend and subscription limits).

### DIFF
--- a/docs/chatgpt-subscription.md
+++ b/docs/chatgpt-subscription.md
@@ -85,8 +85,9 @@ listing.
 ## 2. Register the proxy as a provider in Otari
 
 Add the proxy as a named OpenAI-compatible instance under `providers:` in your
-`config.yml`. Give it a custom instance name and set `provider_type: openai`
-(the aliases `openai-compatible` and `openai_compatible` also work):
+`config.yml`. Give it a custom instance name and set
+`provider_type: openai-compatible` (this and `openai_compatible` are aliases for
+the underlying `openai` implementation, so `provider_type: openai` works too):
 
 ```yaml
 providers:
@@ -108,6 +109,12 @@ providers:
     api_base: "http://localhost:8317/v1"
     # no api_key: the proxy handles auth via its own Codex OAuth session
 ```
+
+> **The keyless path needs a build newer than v0.4.0.** The #423 fix landed after
+> the `v0.4.0` release, so on the `0.4.0` PyPI package or Docker image a keyless
+> instance still raises `MissingApiKeyError`. Until the next release, either run a
+> build from `main` or use the downstream-key form above (set any non-empty
+> `api_key`; a local proxy that ignores it still accepts the call).
 
 Models on this instance are addressed as `chatgpt:<model>`, so the selector
 carries the instance name rather than the raw `openai:` provider. If the proxy
@@ -147,6 +154,13 @@ Pricing is keyed on the instance name, so price `chatgpt:gpt-5`, not
 so the unpriced model is still served and logged with `cost: null`. See
 [Configuration](configuration.md) for both knobs.
 
+> **Counterfactual pricing is not analytics-only.** Configured pricing also
+> drives budget reservations and reconciliation, so a per-key budget depletes
+> against this imaginary cost and will eventually return HTTP 402 even though no
+> money was spent. If you price counterfactually, size budgets to match, or keep
+> `require_pricing: false` when you want the flat-fee reality and no budget
+> enforcement on these calls.
+
 ## 4. Route a request
 
 Use an Otari client key and the `chatgpt:<model>` selector:
@@ -162,8 +176,9 @@ resp = client.chat.completions.create(
 print(resp.choices[0].message.content)
 ```
 
-Budgets, per-key limits, and usage logging all apply exactly as they do for any
-other provider.
+Budgets, per-key limits, and usage logging apply as they do for any other
+provider (see the pricing note above for how counterfactual pricing interacts
+with budgets).
 
 ## Streaming and usage logging
 
@@ -175,16 +190,26 @@ log_writer_strategy: batch
 ```
 
 With the default inline (`single`) strategy, a streaming client can disconnect
-the moment it sees the SSE `[DONE]` marker, before Otari finishes writing the
-usage row, so some records go uncommitted. The background `batch` writer decouples
-the database write from the response stream and records usage reliably. See
-[Configuration](configuration.md) for the field.
+the moment it sees the SSE `[DONE]` marker, mid-write, so the usage row can go
+uncommitted and, because the cancellation lands before reconciliation runs, the
+budget reservation for that call is left unreconciled too. The background `batch`
+writer moves the database write off the response stream, which removes that
+disconnect race for both the usage row and budget reconciliation.
+
+`batch` is not a durability guarantee: it queues rows in memory and flushes on a
+one-second interval or in 100-row batches, so usage can be up to a second late
+and queued rows are still lost on a hard crash. It fixes the client-disconnect
+race, not process death. This behavior is provider-agnostic; see
+[Configuration](configuration.md) for the `log_writer_strategy` field.
 
 ## Caveats
 
 - **Unofficial backend.** This reaches the ChatGPT backend through Codex OAuth
   and a compatibility proxy, not the supported Platform API. The proxy can break
   when the backend changes.
+- **Account risk.** Reaching the subscription through a third-party OAuth proxy
+  may run against OpenAI's terms of service and could put the account itself at
+  risk. Weigh this before pointing anything you care about at it.
 - **Subscription caps apply.** Usage still counts against your ChatGPT
   subscription's limits; Otari's budgets sit on top and do not raise those caps.
 - **An API key is sturdier.** For anything you depend on, an OpenAI Platform API
--- a/docs/chatgpt-subscription.md
+++ b/docs/chatgpt-subscription.md
@@ -1,0 +1,179 @@
+# Use with a ChatGPT subscription
+
+You can route Otari at the models behind a ChatGPT Plus or Pro subscription,
+even though a subscription ships no API key. This page walks through the one
+extra hop that makes it work.
+
+## Why this needs a proxy
+
+A ChatGPT subscription and the OpenAI Platform API are separate products with
+separate billing. The subscription has no standard `sk-...` API key, so Otari
+cannot point at it directly. The subscription is only reachable programmatically
+through **Codex OAuth**, the same login the Codex CLI uses.
+
+The bridge is a small **Codex-OAuth proxy** you run on `localhost`. It logs in
+with your subscription and re-exposes those models as a standard
+OpenAI-compatible endpoint. Otari then treats that endpoint like any other
+self-hosted OpenAI-compatible backend, and budgets, usage logging, and key
+management layer on top unchanged.
+
+```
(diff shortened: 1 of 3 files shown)


docs/chatgpt-subscription.md
```diff
@@ -0,0 +1,179 @@
...
+```yaml
+providers:
+  chatgpt:                       # custom instance name; used in the model selector
+    provider_type: openai-compatible
+    api_base: "http://localhost:8317/v1"
+    api_key: ${CHATGPT_PROXY_KEY}   # the proxy's downstream key
+```
+
+If the proxy is keyless, drop `api_key` entirely. The
+[keyless-custom-endpoint fix (#423)](https://github.com/mozilla-ai/otari/pull/423)
```
khaledosman, line 101: This keyless path is not in any release yet. #423 landed as `5fa80d0`, after `v0.4.0` (`git tag --contains 5fa80d0` is empty), so a reader on the PyPI/Docker `0.4.0` who follows this example hits exactly the `MissingApiKeyError` the section says is fixed.

Suggest a version note ("requires Otari newer than v0.4.0, or a build from `main`"), or lead with the downstream-key variant and mark keyless as the newer path.

docs/chatgpt-subscription.md
```diff
@@ -0,0 +1,179 @@
...
+usage log writer to `batch`:
+
+```yaml
+log_writer_strategy: batch
+```
+
+With the default inline (`single`) strategy, a streaming client can disconnect
+the moment it sees the SSE `[DONE]` marker, before Otari finishes writing the
+usage row, so some records go uncommitted. The background `batch` writer decouples
+the database write from the response stream and records usage reliably. See
```
khaledosman: "records usage reliably" overstates it. `BatchLogWriter` (`src/gateway/services/log_writer.py:55-80`) queues in memory, flushes on a 1s interval or 100-row batch, and only drains on a graceful `stop()`: rows are still lost on a hard crash, and usage becomes visible up to ~1s late. "removes the disconnect race" would be the accurate claim.

Also, this guidance is provider-agnostic and `docs/configuration.md:73` only lists the field with no such note. A cross-reference there would surface it for anyone streaming, not just ChatGPT users.

docs/chatgpt-subscription.md
```diff
@@ -0,0 +1,179 @@
...
+## Streaming and usage logging
+
+If you stream responses (for example from an agent runtime like Hermes), set the
+usage log writer to `batch`:
+
+```yaml
+log_writer_strategy: batch
+```
+
+With the default inline (`single`) strategy, a streaming client can disconnect
```
khaledosman, line 192: The race is bigger than the usage row, which makes this recommendation stronger than stated. Under `single`, cancellation lands inside `log_writer.put`'s commit, so the `reconcile_reservation` call right after it never runs (`src/gateway/api/routes/_pipeline.py:1424-1428`), and `settled` is already `True` by then in `streaming.py:154-161`, so `on_incomplete` doesn't settle the reservation either. Worth one clause: `batch` also keeps budget reconciliation from being skipped.

docs/chatgpt-subscription.md
```diff
@@ -0,0 +1,179 @@
...
+getting from the subscription.
+
+```yaml
+pricing:
+  chatgpt:gpt-5:
+    input_price_per_million: 1.25
+    output_price_per_million: 10.00
+```
+
+Pricing is keyed on the instance name, so price `chatgpt:gpt-5`, not
```
khaledosman, line 152: Counterfactual pricing is framed as analytics-only here, but it feeds real budget enforcement: reservations and reconciliation use configured pricing, so per-key budgets deplete against money nobody spent and will eventually 402. Combined with ":144 Budgets, per-key limits, and usage logging all apply exactly as they do for any other provider", a reader will be surprised.

One line would cover it: if you price counterfactually, size budgets accordingly, or keep `require_pricing: false` when you only want the flat-fee reality.

docs/chatgpt-subscription.md
```diff
@@ -0,0 +1,179 @@
...
+- **Unofficial backend.** This reaches the ChatGPT backend through Codex OAuth
+  and a compatibility proxy, not the supported Platform API. The proxy can break
+  when the backend changes.
+- **Subscription caps apply.** Usage still counts against your ChatGPT
+  subscription's limits; Otari's budgets sit on top and do not raise those caps.
+- **An API key is sturdier.** For anything you depend on, an OpenAI Platform API
+  key via the [OpenAI provider guide](providers/openai.md) is the supported
+  route.
+- **Keep it local.** The proxy holds your subscription's OAuth session. Run it on
+  `localhost` and do not expose it publicly.
```
khaledosman, line 219: The caveats cover breakage and usage caps but not account risk: routing a subscription through a third-party OAuth proxy may run against OpenAI's terms and put the account itself at risk. That is arguably the caveat a reader most needs before starting, so it belongs in this list.

docs/chatgpt-subscription.md
```diff
@@ -0,0 +1,179 @@
...
+  -H "Authorization: Bearer <downstream-key-if-any>"
+```
+
+You should get back a list of models the subscription exposes (Codex serves
+GPT-5 class models; the exact ids come from the proxy).
+
+## 2. Register the proxy as a provider in Otari
+
+Add the proxy as a named OpenAI-compatible instance under `providers:` in your
+`config.yml`. Give it a custom instance name and set `provider_type: openai`
```
khaledosman: Prose says set `provider_type: openai` with the aliases as an aside, but all three YAML blocks below use `openai-compatible`. Make the prose and the examples agree.

docs/chatgpt-subscription.md
```diff
@@ -0,0 +1,179 @@
...
+
+- the local base URL it serves, for example `http://localhost:8317/v1`, and
+- whether it requires a **downstream key** (a token you set on the proxy and
+  then present to it) or serves keyless.
+
+Confirm the proxy answers before wiring Otari to it:
+
+```bash
+curl -sS http://localhost:8317/v1/models \
+  -H "Authorization: Bearer <downstream-key-if-any>"
```
khaledosman: This always sends `Authorization: Bearer <downstream-key-if-any>`; for a keyless proxy the header should be dropped entirely rather than sent with a placeholder. Worth saying so, since the keyless setup is documented as the clean path.

docs/index.md
```diff
@@ -34,6 +34,7 @@ Running and managing a gateway.
 - [Access control](access-control.md): users, API keys, and budgets, with the management endpoints that drive them.
 - [Supported models](models.md): providers, model format, and capabilities.
 - [OpenAI provider guide](providers/openai.md): configure OpenAI and route your first request through Otari.
+- [Use with a ChatGPT subscription](chatgpt-subscription.md): route Otari at ChatGPT Plus/Pro models through a local Codex-OAuth proxy.
```
khaledosman: Placement nit: this sits under "For operators" while the sibling "Use with X" recipes are under "For integrators" (lines 47-49). Defensible either way given the provider-config framing, but the title reads like the integrator set.

khaledosman (approved):
Docs-only, and the config claims check out against the code (provider_type aliases in `core/config.py:23-24`, instance-keyed pricing, `require_pricing`/`log_writer_strategy` defaults). Links and anchors all resolve. Findings inline: the main one is that the keyless path documented here is not in any released version yet. Nothing blocking.

PR #441, opened 2026-07-27
feat(providers): add opt-in SSRF gate for provider api_base

### LINKED ISSUE #316: Dashboard: opt-in SSRF gate for provider api_base

Optional, opt-in SSRF hardening for the operator-supplied provider `api_base`.

## Context
The provider test and discovery paths issue live outbound requests to a caller-supplied `api_base`:
- `POST /v1/provider-credentials/test` (ad-hoc, test-before-save)
- `POST /v1/provider-credentials/{instance}/test`
- model discovery when a stored/config provider is used at runtime

Unlike outbound MCP URLs (`services/url_safety.py`), this path does **not** run `api_base` through an SSRF check, so an operator can point the gateway at internal addresses.

## Why this is deliberately *not* a default block today
`api_base` is admin-supplied (master-key gated, standalone-only), the same trust level as a `config.yml` provider endpoint, which the codebase also does not SSRF-check. Both @khaledosman and the review agreed it is acceptable under the current threat model. Critically, the home-lab / self-hosted use case depends on private-network `api_base` values (`localhost`, RFC-1918), so applying the MCP validator as-is (which blocks private ranges by default) would break a core use case.

## Proposed approach (opt-in, default-allow)
Add an env-gated check that mirrors the existing `url_safety` toggles (`OTARI_MCP_ALLOW_PRIVATE_HOSTS`, `OTARI_WEB_SEARCH_ALLOW_PRIVATE_HOSTS`):
- Default: allow everything (no behavior change; home-lab keeps working).
- When an operator sets an opt-in flag (e.g. `OTARI_PROVIDER_ALLOW_PRIVATE_HOSTS=false`), run `api_base` through the private/link-local/reserved-range check before both the test endpoints and runtime discovery.

## Tasks
- [ ] Decide the flag name and default (default must stay allow-all).
- [ ] Validate `api_base` at the test endpoints and in `model_discovery_service` when the gate is enabled, reusing `url_safety` rather than new SSRF logic.
- [ ] Document the flag alongside the other `*_ALLOW_PRIVATE_HOSTS` toggles.
- [ ] Tests for allow (default) and block (opt-in) paths.

**Only relevant to standalone mode.** Hybrid does not mount these routes and resolves credentials on the platform side.

Part of #299.

njbrake:
## Description

Adds an opt-in, default-allow SSRF gate for the operator-supplied provider `api_base`.

The provider test and discovery paths issue outbound requests to a caller-supplied `api_base` without any SSRF check (`POST /v1/provider-credentials/test`, `POST /v1/provider-credentials/{instance}/test`, and runtime model discovery), so an operator could point the gateway at internal addresses. Unlike outbound MCP or web-search URLs, `api_base` is admin-supplied (master-key gated, standalone-only), the same trust level as a `config.yml` provider endpoint. The home-lab / self-hosted use case depends on private-network endpoints (`localhost`, RFC 1918), so a default block would break a core use case.

New `validate_provider_api_base()` in `services/url_safety.py` mirrors the existing `url_safety` toggles but defaults to allow-all. When an operator sets `OTARI_PROVIDER_ALLOW_PRIVATE_HOSTS=false`, `api_base` runs through the same private/link-local/reserved-range check (loopback blocked too) before both the connection-test endpoints and runtime discovery:

- A blocked discovery fails outright rather than falling back to the declared `models:` list.
- A blocked ad-hoc test returns a sanitized error without dialing the endpoint or echoing the key.

The flag follows the same wiring as the other gates: a `provider_allow_private_hosts` config field bridged to `OTARI_PROVIDER_ALLOW_PRIVATE_HOSTS`, surfaced display-only in the dashboard config viewer, and documented in `docs/configuration.md`.

Scoped to standalone mode only; hybrid does not mount these routes and resolves credentials on the platform side. Deliberately not applied to the chat-completion dispatch, matching the issue's scope (test endpoints + discovery).

## PR Type

- [x] New Feature

## Relevant issues

Fixes #316
Part of #299

## Checklist

- [x] If the API contract changed, I regenerated the OpenAPI spec (no contract change; `--check` passes).

## AI Usage

Implemented by Claude Code via back-and-forth with @njbrake; the reasoning and decisions are his.

**NOTE:**

## Summary

Added a new opt-in protection gate for provider “API base” URLs in standalone mode. When enabled, the system blocks attempts to reach private/loopback/link-local/reserved or otherwise unsafe addresses before running provider connection checks or discovering models. The feature is configurable via a new `provider_allow_private_hosts` setting (and `OTARI_PROVIDER_ALLOW_PRIVATE_HOSTS` environment variable), shows up in the settings dashboard, and is documented alongside existing network-access options.

The update also adds coverage to ensure the gate only applies where intended (provider API base checks/model discovery in standalone mode), preserves current default behavior, avoids fallback to declared models when discovery is blocked, and returns sanitized errors for blocked ad-hoc connection tests.

## Technical notes

- Default remains “allow” (no behavior change unless the operator explicitly sets the env var to `false`).
- The gate does not apply to chat-completion dispatch or hybrid mode.
- URL-safety logic was extended with a dedicated `validate_provider_api_base` helper used by the relevant discovery/connection paths.


src/gateway/services/model_discovery_service.py
```diff
@@ -265,6 +266,13 @@ async def _discover_for_provider(
     api_base = kwargs.pop("api_base", None)
     client_args = kwargs.pop("client_args", None)
 
+    # Opt-in SSRF gate (default allow-all). Raised before the declared-models
+    # fallback below so a blocked endpoint fails outright rather than silently
+    # serving its declared listing. Truthy check so an empty/absent api_base
+    # (the "use the SDK default endpoint" case) is not treated as a URL.
+    if api_base:
+        await validate_provider_api_base(api_base)
```
khaledosman, line 274: The gate covers the two paths that only *report* on an `api_base`, not the ones that use or persist it:

- `POST`/`PATCH /v1/provider-credentials` (`api/routes/providers.py:476`, `:514`) will still store `api_base: http://169.254.169.254` with the gate on.
- Chat dispatch dials it for real regardless — `get_provider_kwargs` → `acompletion(api_base=…)` has no check.

So with `provider_allow_private_hosts=false` you get a provider that reports "blocked" on the models/providers page but happily serves `POST /v1/chat/completions` against the internal address. An operator flipping this flag will read it as "this gateway won't reach internal endpoints."

Cheapest way to make the flag coherent: gate `save_credential` (400 on a blocked `api_base`) and `validate_provider_instances()` at startup, so a gated deployment fails fast instead of silently losing its catalogue. If chat dispatch is deliberately out of scope, worth stating that in the docs table too.

src/gateway/services/url_safety.py
```diff
@@ -179,3 +187,56 @@ async def validate_outbound_fetch_url(url: str) -> None:
                 f"fetch host {host!r} resolves to {addr} which is {reason}; "
                 "rejecting to prevent SSRF. Set OTARI_WEB_SEARCH_ALLOW_PRIVATE_HOSTS=true to override."
             )
+
+
+def _allow_provider_private_hosts() -> bool:
```
khaledosman, line 207: This parser and the one that validates the config field disagree, and the gate fails open on the disagreement. `_parse_bool_env` (`core/config.py:875`) accepts `on`/`off`; this only rejects `{"0", "false", "no"}`. So `OTARI_PROVIDER_ALLOW_PRIVATE_HOSTS=off` shows as `false` in `/v1/settings` (operator reads: gate on) while this returns `True` and allows everything.

Pre-existing in `_allow_loopback`, but this is a new security gate: either reuse `_parse_bool_env` or add `off` to the set.

src/gateway/services/url_safety.py
```diff
@@ -179,3 +187,56 @@ async def validate_outbound_fetch_url(url: str) -> None:
...
+    # Defaults to True (allow-all), the opposite of the MCP/web-search gates: an
+    # operator-supplied api_base is master-key gated and the home-lab use case
+    # depends on private endpoints, so the check is off until an operator opts in.
+    return otari_env("PROVIDER_ALLOW_PRIVATE_HOSTS", "true").lower() not in {"0", "false", "no"}
+
+
+async def validate_provider_api_base(url: str) -> None:
```
khaledosman, line 216: This is a near-verbatim copy of `validate_outbound_fetch_url` above — three copies of parse → literal-or-resolve → `_blocked_reason` loop now differ only in message strings and the override hint. Worth one private helper (`_reject_internal_address(url, *, label, override_var)`) with each public function keeping its own short-circuit semantics.

Also worth a line in the docstring: this inverts the ordering the other two use — the allow-flag short-circuit runs *before* scheme/host validation, so with the gate off a `file://` `api_base` isn't rejected here. Intentional for a default-allow gate, but surprising next to its siblings.

src/gateway/services/url_safety.py
```diff
@@ -15,6 +15,14 @@
   ``OTARI_WEB_SEARCH_ALLOW_PRIVATE_HOSTS`` for operators with unusual
   setups (private indexes etc.).
 
+* **Provider ``api_base``** (:func:`validate_provider_api_base`) — URL is
```
khaledosman: Module docstring still opens with "Two distinct call sites" (line 3) while three are now documented.

src/gateway/core/config.py
```diff
@@ -56,6 +56,7 @@
     "web_search_allow_private_hosts",
     "mcp_allow_loopback",
     "mcp_allow_private_hosts",
+    "provider_allow_private_hosts",
```
khaledosman, line 59: The bridge is the only thing that makes a YAML-set `provider_allow_private_hosts: false` reach `_allow_provider_private_hosts()`, and it's the piece most likely to regress silently — which is why `tests/integration/test_config_env_loading.py:289` (`test_yaml_ssrf_gates_round_trip_through_env_bridge`) exists for the other two gates. Worth a case there for this field, especially since the polarity is inverted (`false` *enables* the gate, so the bridge has to serialize the value the parser above rejects).

src/gateway/api/routes/settings.py
```diff
@@ -121,6 +121,7 @@
             "web_search_allow_private_hosts",
             "mcp_allow_loopback",
             "mcp_allow_private_hosts",
+            "provider_allow_private_hosts",
```
khaledosman, line 124: `tests/unit/test_settings_endpoint.py:223` asserts the network-safety gates stay `settable is False`; add this field to that list so it can't be promoted to hot-settable by accident.

docs/configuration.md
```diff
@@ -105,6 +105,7 @@ pricing:
 | `web_search_allow_private_hosts` | bool | `false` | SSRF gate: allow the web-search backend to fetch private/loopback/reserved hosts. Also settable via `OTARI_WEB_SEARCH_ALLOW_PRIVATE_HOSTS`. |
 | `mcp_allow_loopback` | bool | `true` | SSRF gate: allow MCP server URLs that resolve to loopback (same-host sidecars). Also settable via `OTARI_MCP_ALLOW_LOOPBACK`. |
 | `mcp_allow_private_hosts` | bool | `false` | SSRF gate: allow MCP server URLs that resolve to private/reserved hosts (and accept hostnames that fail to resolve at validation time). Also settable via `OTARI_MCP_ALLOW_PRIVATE_HOSTS`. |
+| `provider_allow_private_hosts` | bool | `true` | SSRF gate: allow a provider `api_base` that resolves to private/loopback/reserved hosts. On by default (unlike the other gates), because `api_base` is master-key gated and the home-lab use case depends on private endpoints. Set to `false` to make provider connection tests and model discovery refuse an internal `api_base`. Also settable via `OTARI_PROVIDER_ALLOW_PRIVATE_HOSTS`. |
```
khaledosman: "make provider connection tests and model discovery refuse an internal `api_base`" is accurate but reads narrower than an operator will interpret the flag name. Worth an explicit sentence that chat dispatch and the credential write path are *not* gated, so nobody treats this as an egress control.

khaledosman (approved):
Gate wiring (config field, `ENV_BRIDGED_FIELDS` bridge, config-viewer entry, docs) is consistent with the existing SSRF gates, and the discovery path correctly fails outright instead of falling back to the declared `models:` list. Unit tests pass locally (67).

The main thing worth a follow-up is that the gate covers the two paths that only *report* a failure, while the paths that actually dial the endpoint (chat dispatch) or persist it (`POST`/`PATCH /v1/provider-credentials`) are ungated, so a gated deployment ends up in a half-state. Details inline; nothing blocking.

---

PR #442, opened 2026-07-28
feat(web-search): support recency filtering and publish dates in Brave adapter

### LINKED ISSUE #402: [FEATURE] Support recency filtering with Brave web-search adapter. It currently ignores `time_range` and drops publish dates
### Problem or use case

The bundled Brave adapter (`scripts/web-search-brave-adapter/app.py`) translates the
gateway's SearXNG-style `/search` request into a Brave call, but discards the two
pieces of information a time-sensitive consumer needs: it never forwards a recency
filter to Brave, and it drops Brave's per-result date. As a result, `otari_web_search`
returns relevance-ranked evergreen pages regardless of the requested time window, with
no date attached — so a daily/recent-mentions use case can't filter by recency at all.

champ18ion:
## Description
The Brave adapter ignored `time_range` entirely and dropped Brave's per-result publish date, so a time-sensitive query returned relevance-ranked evergreen pages with no way to judge how current any result was.

Adds `time_range` support to the Brave adapter, mapped onto Brave's `freshness` filter (`day`/`d` -> `pd`, `week`/`w` -> `pw`, `month`/`m` -> `pm`, `year`/`y` -> `py`), matching the vocabulary the Tavily adapter already uses. Each result's `page_age` (preferred) or `age` field is passed through as `published_date`, whichever Brave supplied, unmodified.

Also updates `web_search_backend.py`'s result rendering to surface `published_date` next to the title when a backend supplies it (Brave now does; plain SearXNG never will), otherwise the model has no way to judge recency from a snippet alone, which is the actual point of requesting a `time_range` filter in the first place.

## PR Type
- [x] New Feature

## Relevant issues
Fixes #402

## Checklist
- [x] I ran the Definition of Done checks locally (`make lint`, `make typecheck`, `make test`) — note: ran via a manual pip venv, not `uv`, since Windows is excluded from your lockfile's supported platforms.
- [x] Documentation was updated where necessary (adapter README's Provider options section).
- [x] If the API contract changed, I regenerated the OpenAPI spec — not applicable, no gateway routes changed.

## AI Usage

**Any additional AI details you'd like to share:** Followed the existing Tavily adapter's pattern for time_range handling and test structure; code written by Claude Code under my direction and review.

- [ ] I am an AI Agent filling out this form

## Summary

- Added Brave recency support to the web-search adapter by translating a requested `time_range` (day/week/month/year) into Brave’s `freshness` filter so callers can get recent results.
- Extended results to carry a per-hit `published_date` from Brave (using Brave’s `page_age` when available, otherwise `age`) and display it alongside each result title when present.
- Tightened Brave integration robustness: added provider/date parameter validation, safer response parsing/shape checks, and clearer upstream error handling.
- Updated the adapter README and the result contract documentation to describe `published_date` and how provider `time_range` is forwarded.
- Improved safety when rendering `published_date` in the backend by normalizing whitespace and bounding the displayed length.
- Added/expanded unit tests for Brave recency, `published_date` mapping/rendering, input validation, error handling, and missing API-key behavior; also refactored shared test helpers used by adapter test suites.

These changes help consumers retrieve and assess time-sensitive search results rather than only relevance-ranked evergreen pages.

## Technical notes

- `time_range` is mapped to Brave freshness codes: `day|d→pd`, `week|w→pw`, `month|m→pm`, `year|y→py`.
- `published_date` is sourced from Brave’s `page_age` (preferred) or `age`, then rendered in the backend after whitespace normalization and max-length bounding.

### DIFF
--- a/scripts/web-search-brave-adapter/README.md
+++ b/scripts/web-search-brave-adapter/README.md
@@ -6,9 +6,20 @@ bundled SearXNG metasearch (whose free engines rate-limit/CAPTCHA automated
 queries by IP).
 
 It exposes `GET /search?q=…&format=json` returning
-`{"results": [{"url", "title", "content"}]}` — the exact shape
-`WebSearchBackend` expects — and translates that to/from Brave's API. The Brave
+`{"results": [{"url", "title", "content", "published_date"?}]}` — the exact
+shape `WebSearchBackend` expects, plus an optional publish/last-updated date
+when Brave supplies one — and translates that to/from Brave's API. The Brave
 key lives in this service, never in the gateway.
+
+## Provider options
+
+The gateway forwards provider-specific knobs (set via `provider_options` on the
+`otari_web_search` tool entry, or as workspace defaults in hybrid mode) as
+extra query params. This adapter recognizes one, mapped onto Brave's
+`freshness` filter; anything else is ignored:
+
+- `time_range` (`day` | `week` | `month` | `year`, or the single-letter forms
+  `d` | `w` | `m` | `y`)
 
 ## Run it via docker-compose
 
--- a/scripts/web-search-brave-adapter/README.md
+++ b/scripts/web-search-brave-adapter/README.md
@@ -6,9 +6,9 @@ bundled SearXNG metasearch (whose free engines rate-limit/CAPTCHA automated
 queries by IP).
 
 It exposes `GET /search?q=…&format=json` returning
-`{"results": [{"url", "title", "content", "published_date"?}]}` — the exact
+`{"results": [{"url", "title", "content", "published_date"?}]}`, the exact
 shape `WebSearchBackend` expects, plus an optional publish/last-updated date
-when Brave supplies one — and translates that to/from Brave's API. The Brave
+when Brave supplies one, and translates that to/from Brave's API. The Brave
 key lives in this service, never in the gateway.
 
 ## Provider options
--- a/scripts/web-search-brave-adapter/app.py
+++ b/scripts/web-search-brave-adapter/app.py
@@ -16,9 +16,10 @@ SearXNG path). Set ``extracted_content`` here instead if you want snippet-only
 behaviour and to skip the gateway's per-URL fetch.
 
 The gateway may forward provider-specific knobs as extra query params
-(``provider_options`` on the ``otari_web_search`` tool entry). This adapter
-only recognizes ``time_range``, mapped onto Brave's ``freshness`` filter;
-anything else is ignored, never forwarded to Brave.
+(``provider_options`` on the ``otari_web_search`` tool entry, set by the
+operator or workspace, never per-query by the model). This adapter only
+recognizes ``time_range``, mapped onto Brave's ``freshness`` filter; anything
+else is ignored, never forwarded to Brave.
 """
 
 from __future__ import annotations
@@ -52,9 +53,12 @@ app = FastAPI(title="otari web-search Brave adapter")
 
 
 @app.get("/health")
-async def health() -> dict[str, str]:
-    # Surfaces a misconfigured deployment (missing key) without a live query.
-    return {"status": "healthy" if BRAVE_API_KEY else "missing BRAVE_API_KEY"}
+async def health() -> JSONResponse:
+    # Fail closed when misconfigured (missing key) so orchestrators don't treat
+    # an unusable adapter as healthy.
+    if not BRAVE_API_KEY:
+        return JSONResponse(status_code=503, content={"status": "missing BRAVE_API_KEY"})
+    return JSONResponse(status_code=200, content={"status": "healthy"})
 
 
 @app.get("/search")
@@ -84,8 +88,21 @@ async def search(
         )
     except httpx.HTTPError as exc:
         return JSONResponse(status_code=502, content={"error": f"brave search failed: {exc}"})
+    except ValueError as exc:
+        # Invalid / non-JSON body from Brave. Translate to a 502 rather than
+        # letting it bubble up as a 500, so the gateway sees a consistent signal.
+        return JSONResponse(status_code=502, content={"error": f"brave search returned invalid JSON: {exc}"})
 
-    hits = (payload.get("web") or {}).get("results") or []
+    if not isinstance(payload, dict):
+        return JSONResponse(status_code=502, content={"error": "brave search returned an unexpected shape"})
+
+    web = payload.get("web")
+    hits = web.get("results") if isinstance(web, dict) else None
+    if not isinstance(hits, list):
+        # A missing/non-list `results` is an upstream contract break, not "no
+        # hits", so surface it as a 502 instead of silently returning empty.
+        return JSONResponse(status_code=502, content={"error": "brave search returned an unexpected shape"})
+
     results: list[dict[str, Any]] = []
     for h in hits:
         if not isinstance(h, dict) or not h.get("url"):
@@ -95,10 +112,11 @@ async def search(
             "title": h.get("title", ""),
             "content": h.get("description", ""),
         }
-        # Brave documents both `page_age` and `age` on a result with no
-        # documented distinction between them; prefer page_age (appears to be
-        # the more specific field) and fall back to age. Passed through as an
-        # opaque string, whatever format Brave sent, not reformatted here.
+        # page_age (ISO 8601 timestamp) and age (human-readable, e.g. "3 days
+        # ago") are different formats, not interchangeable; prefer page_age
+        # for a consistent, parseable format across results and fall back to
+        # age only when Brave didn't supply it. Passed through as an opaque
+        # string either way, not reformatted here.
         published_date = h.get("page_age") or h.get("age")
         if published_date:
             result["published_date"] = published_date
--- a/scripts/web-search-brave-adapter/app.py
+++ b/scripts/web-search-brave-adapter/app.py
@@ -14,6 +14,11 @@ Each result's ``content`` is Brave's snippet; ``extracted_content`` is left
 unset so the gateway still fetches and extracts the full page (matching the
 SearXNG path). Set ``extracted_content`` here instead if you want snippet-only
 behaviour and to skip the gateway's per-URL fetch.
+
+The gateway may forward provider-specific knobs as extra query params
+(``provider_options`` on the ``otari_web_search`` tool entry). This adapter
+only recognizes ``time_range``, mapped onto Brave's ``freshness`` filter;
+anything else is ignored, never forwarded to Brave.
 """
 
 from __future__ import annotations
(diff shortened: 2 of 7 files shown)


src/gateway/services/web_search_backend.py
```diff
@@ -423,5 +423,11 @@ def _format_results_for_model(query: str, results: list[dict[str, Any]]) -> str:
...
         if len(body) > _CONTENT_TRUNCATE_CHARS:
             body = body[:_CONTENT_TRUNCATE_CHARS].rstrip() + "…"
-        parts.append(f"[{i}] {title}\n{url}\n{body}".rstrip())
+        # Optional: only backends that supply a recency signal (e.g. the
+        # Brave adapter's provider_options time_range support) set this. The
+        # model can't judge how current a result is from the snippet alone,
+        # so surface it when present instead of silently dropping it.
+        published_date = str(r.get("published_date") or "").strip()
+        header = f"[{i}] {title}" + (f" ({published_date})" if published_date else "")
+        parts.append(f"{header}\n{url}\n{body}".rstrip())
```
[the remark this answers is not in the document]
champ18ion: good catch, published_date is untrusted backend input. applied your fix and added a test for it

scripts/web-search-brave-adapter/README.md
```diff
@@ -6,10 +6,21 @@ bundled SearXNG metasearch (whose free engines rate-limit/CAPTCHA automated
 queries by IP).
 
 It exposes `GET /search?q=…&format=json` returning
-`{"results": [{"url", "title", "content"}]}` — the exact shape
-`WebSearchBackend` expects — and translates that to/from Brave's API. The Brave
+`{"results": [{"url", "title", "content", "published_date"?}]}` — the exact
+shape `WebSearchBackend` expects, plus an optional publish/last-updated date
+when Brave supplies one — and translates that to/from Brave's API. The Brave
```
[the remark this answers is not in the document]
champ18ion: fixed

scripts/web-search-brave-adapter/app.py
```diff
@@ -63,13 +86,21 @@ async def search(q: str = Query(..., min_length=1)) -> JSONResponse:
...
+            continue
+        result: dict[str, Any] = {
             "url": h["url"],
             "title": h.get("title", ""),
             "content": h.get("description", ""),
         }
-        for h in hits
-        if isinstance(h, dict) and h.get("url")
-    ]
+        # Brave documents both `page_age` and `age` on a result with no
```
khaledosman: Brave does distinguish these two: `page_age` is an ISO 8601 timestamp from the page, `age` is a human-readable string (`"3 days ago"`, `"November 1, 2023"`). Coalescing them means one response can carry `2026-07-20T00:00:00.000Z` for result 1 and `3 days ago` for result 2 — and a relative string has no reference point once it's sitting in model context, which undercuts the stated goal of letting the model judge recency.

Either pass `page_age` only, or keep the fallback and fix the comment to name the format difference so the next reader knows the field isn't one format.

src/gateway/services/web_search_backend.py
```diff
@@ -423,5 +428,14 @@ def _format_results_for_model(query: str, results: list[dict[str, Any]]) -> str:
...
             body = body[:_CONTENT_TRUNCATE_CHARS].rstrip() + "…"
-        parts.append(f"[{i}] {title}\n{url}\n{body}".rstrip())
+        # Optional: only backends that supply a recency signal (e.g. the
+        # Brave adapter's provider_options time_range support) set this. The
+        # model can't judge how current a result is from the snippet alone,
+        # so surface it when present instead of silently dropping it.
+        # Untrusted backend-controlled input: collapse to one line and cap
+        # the length so a malicious/compromised backend can't use this field
+        # to inject newlines or an oversized string into model context.
+        published_date = " ".join(str(r.get("published_date") or "").split())[:_PUBLISHED_DATE_MAX_CHARS]
```
khaledosman, line 441: The injection rationale doesn't hold as written, because `title` — rendered on the same line, three lines up — is only `.strip()`ed. A compromised backend can already inject newlines into the model-facing block through `title` (and through `content`, unavoidably), so normalising only the new field reads as protection that isn't there.

Either collapse `title` the same way (one line, and arguably the more valuable of the two), or drop the security framing from the comment and keep the cap as plain rendering hygiene.

scripts/web-search-brave-adapter/app.py
```diff
@@ -63,13 +86,21 @@ async def search(q: str = Query(..., min_length=1)) -> JSONResponse:
         return JSONResponse(status_code=502, content={"error": f"brave search failed: {exc}"})
 
     hits = (payload.get("web") or {}).get("results") or []
```
khaledosman: This adapter still lacks the response guards its Tavily sibling has, and the PR rewrote the block right below:

- `resp.json()` raising `ValueError` isn't caught (only `httpx` errors are), so a non-JSON body from Brave becomes a 500 instead of the 502 every other failure maps to (`web-search-tavily-adapter/app.py:91`).
- `payload.get("web")` on a non-dict payload is an `AttributeError` → 500 (Tavily checks `isinstance(payload, dict)` at `:97`).
- A missing/non-list `results` is silently "no hits" here; Tavily surfaces it as a 502 at `:102`.

Pre-existing, but this PR adds the file's first test coverage, so it's the natural moment.

tests/unit/test_brave_adapter.py
```diff
@@ -0,0 +1,190 @@
...
+    return module
+
+
+@pytest.mark.asyncio
+async def test_health_reports_missing_key(monkeypatch: pytest.MonkeyPatch) -> None:
+    module = _load_adapter(monkeypatch, api_key="")
+    transport = httpx.ASGITransport(app=module.app)
+    async with httpx.AsyncClient(transport=transport, base_url="http://adapter") as client:
+        resp = await client.get("/health")
+    assert resp.json() == {"status": "missing BRAVE_API_KEY"}
```
khaledosman, line 35: This asserts the body but not the status, which locks in a health check that lies: `/health` returns **200** with `{"status": "missing BRAVE_API_KEY"}`, so a Compose or orchestrator healthcheck treats a keyless adapter as healthy. Tavily fails closed with a 503 for the same condition and its test asserts it (`test_tavily_adapter.py:37`). Worth matching, and asserting the status here.

src/gateway/services/web_search_backend.py
```diff
@@ -77,6 +77,11 @@
 # /search?format=json shape.
 _DEFAULT_ENGINES = ("duckduckgo", "mojeek", "qwant", "wikipedia")
 _CONTENT_TRUNCATE_CHARS = 1500
+# published_date is backend-controlled, untrusted input (e.g. whatever a
+# search-API-fronting adapter forwards from the provider); bound it before it
+# reaches model context so a malicious/compromised backend can't inject
+# newlines or an oversized string through this field.
+_PUBLISHED_DATE_MAX_CHARS = 128
```
khaledosman, line 85: The module docstring (lines 8-16) is where a custom-backend author learns the result contract — it names the optional `extracted_content` field explicitly. `published_date` should be listed there too, not only in the Brave README.

tests/unit/test_brave_adapter.py
```diff
@@ -0,0 +1,190 @@
...
+    transport = httpx.ASGITransport(app=module.app)
+    async with httpx.AsyncClient(transport=transport, base_url="http://adapter") as client:
+        resp = await client.get("/search", params={"q": "x"})
+    assert resp.status_code == 503
+
+
+# ----- helpers -----
+
+
+def _mock_async_client(handler: Any) -> Any:
```
khaledosman: `_mock_async_client` and `_load_adapter` are now duplicated verbatim from `test_tavily_adapter.py`. A shared helper in `tests/unit/` would make the third adapter cheap — worth doing now that there are two copies to keep in sync.

scripts/web-search-brave-adapter/README.md
```diff
@@ -6,10 +6,21 @@ bundled SearXNG metasearch (whose free engines rate-limit/CAPTCHA automated
...
 key lives in this service, never in the gateway.
 
+## Provider options
+
+The gateway forwards provider-specific knobs (set via `provider_options` on the
+`otari_web_search` tool entry, or as workspace defaults in hybrid mode) as
+extra query params. This adapter recognizes one, mapped onto Brave's
+`freshness` filter; anything else is ignored:
+
+- `time_range` (`day` | `week` | `month` | `year`, or the single-letter forms
```
khaledosman, line 21: Worth one sentence that `time_range` reaches the adapter only from operator/workspace `provider_options` (`web_search_backend.py:251`), never per-query from the model — it's a fixed window for every search, not something the model picks. Easy to read the opposite from "recency filtering".

khaledosman (approved):
`time_range` → Brave `freshness` mapping matches the Tavily adapter's vocabulary exactly, and the `Query(pattern=…)` guard means the dict lookup can't `KeyError`. Tests pass locally (48, including the Tavily adapter and web-search backend suites).

Two things worth addressing: `page_age`/`age` are not interchangeable formats (the code comment says they are), and the injection rationale for the new sanitisation doesn't hold while `title` on the same rendered line is untouched. Details inline; nothing blocking.

---

scripts/web-search-brave-adapter/app.py
champ18ion: good catch, thanks for the correction. kept the fallback but fixed the comment to actually name the format difference instead of claiming there wasn't one

src/gateway/services/web_search_backend.py
champ18ion: fair, you're right, title isn't protected so the framing overclaimed. reworded it as rendering hygiene rather than a security boundary

scripts/web-search-brave-adapter/app.py
champ18ion: good point, since this is the file's first test coverage anyway. added the same guards Tavily has (ValueError on bad JSON, dict check on payload, list check on results) plus a test for each

tests/unit/test_brave_adapter.py
champ18ion: yeah that's a real bug not just a test gap, fixed /health to fail closed with 503 matching Tavily, and the test now asserts the status code

src/gateway/services/web_search_backend.py
champ18ion: added it there too

tests/unit/test_brave_adapter.py
champ18ion: agreed, pulled both into tests/unit/_search_adapter_test_support.py and switched both test files to use it

scripts/web-search-brave-adapter/README.md
champ18ion: good clarification, added a sentence about that

champ18ion:
thanks for the thorough review, appreciate you (and Claude Code) catching all of that. addressed everything in f2fcc63: fixed the page_age/age comment, reworded the injection framing, added the missing response guards + tests to match Tavily, fixed /health to actually fail closed, documented published_date in the module docstring, extracted the shared test helpers, and clarified the README on time_range

tbille (approved):

tbille:
nice one, thanks for the contribution @champ18ion 🚀

PR #444, opened 2026-07-28
fix(providers): gate provider api_base write path behind SSRF check

### LINKED ISSUE #443: Extend provider api_base SSRF gate to the write path and chat dispatch (or decide report-only is final)

Follow-up to #316. The opt-in `provider_allow_private_hosts` SSRF gate added there deliberately covers only the paths that *report* on a provider `api_base`: the connection-test endpoints (`POST /v1/provider-credentials/test`, `POST /v1/provider-credentials/{instance}/test`) and runtime model discovery. Two paths that actually use or persist an `api_base` are still ungated, so a gated deployment ends up in a half-state.

## Ungated paths
- **Credential write path**: `POST` / `PATCH /v1/provider-credentials` still stores an internal `api_base` (e.g. `http://169.254.169.254`) with the gate on. The dashboard then shows the provider as "blocked" on the models/providers page while the credential sits persisted.
- **Chat dispatch**: `get_provider_kwargs` -> `acompletion(api_base=...)` dials the endpoint for real on every request, with no SSRF check regardless of the gate.

## Why it was scoped out of #316
The gate as shipped is not a general egress control, and #316 explicitly scoped it to the report paths (this is now documented in the config field description and `docs/configuration.md`). @khaledosman flagged the half-state in review of #441 as a non-blocking follow-up.

## Options to consider
- Gate the write path: reject a private/internal `api_base` at `POST` / `PATCH /v1/provider-credentials` when `provider_allow_private_hosts=false`, so a blocked endpoint cannot be persisted in the first place.
- Decide whether chat dispatch should be gated at all. It is the hot path (every request), the resolved credentials may come from `config.yml` (same trust level), and any check there needs a design that does not regress the home-lab use case or add per-request DNS resolution cost. This likely belongs with the broader egress-policy work in #299.
- Alternatively, accept the report-only scope as final and just make the docs even more prominent.

Part of #299.

njbrake:
## Description

Follow-up to #441 / #316. Extends the opt-in `provider_allow_private_hosts` SSRF gate to the **credential write path**. `POST` / `PATCH /v1/provider-credentials` now reject an internal `api_base` (private/link-local/reserved) when the gate is on, so a blocked endpoint cannot be persisted in the first place instead of sitting stored while the dashboard shows it as "blocked".

Chat dispatch stays ungated and is deferred as a follow-up: it is the hot path, resolved credentials may come from `config.yml` (same trust level), and gating it needs a design that does not regress the home-lab case or add per-request DNS cost. That belongs with the broader egress-policy work in #299.

### Behavior

- Default stays **allow-all**: the home-lab / self-hosted case that relies on private endpoints is unaffected unless the operator sets `OTARI_PROVIDER_ALLOW_PRIVATE_HOSTS=false`.
- With the gate on, an internal `api_base` is refused with HTTP 400 (reusing `validate_provider_api_base`); a public `api_base` still saves.
- `PATCH` only re-validates when `api_base` is present in the request body, preserving the omit-vs-null semantics.

### Changes

- `src/gateway/api/routes/providers.py`: new `_gate_api_base()` helper mapping `UnsafeURLError` to a 400, wired into create (before the existence check, so nothing persists) and update.
- `src/gateway/services/url_safety.py`, `src/gateway/core/config.py`, `docs/configuration.md`: updated the gate's documented scope (write path now covered; chat dispatch still not).
- `tests/integration/test_provider_credentials_api.py`: 4 tests covering gate-on rejection on POST (persists nothing) and PATCH (original untouched), gate-on still allowing a public base, and default allowing an internal base.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes #443 (write-path scope only). Part of #299.

## Checklist

- [x] If the API contract changed, I regenerated the OpenAPI spec (`uv run python scripts/generate_openapi.py`). (No contract change; `--check` confirmed the spec is up to date.)

## AI Usage

- [ ] No AI was used.

**Any additional AI details you'd like to share:** Implemented by Claude Code via back-and-forth with @njbrake; the reasoning and decisions are his.

**NOTE:**

## Summary

- Extended private-host protection to provider credential creation and updates.
- Prevented blocked, invalid, or unresolved `api_base` values from being saved when protection is enabled.
- Preserved the default allow-all behavior.
- Updated configuration documentation.
- Added integration tests for blocked, public, default, unresolved, and omitted `api_base` behavior.

## Technical notes

- `PATCH` validates `api_base` only when the request includes it.
- Chat dispatch remains outside this protection.


src/gateway/core/config.py
```diff
@@ -540,10 +540,10 @@ class GatewayConfig(BaseSettings):
...
             "On by default (the opposite of the other SSRF gates) because operator-supplied api_base "
             "values are master-key gated and the home-lab / self-hosted use case depends on private "
-            "endpoints. Set to false to make provider connection tests and model discovery refuse an "
-            "internal api_base. Scoped to those report paths only: chat dispatch (which dials the "
-            "endpoint) and the credential write path (which persists it) are not gated, so this is not "
-            "a general egress control. Also settable via OTARI_PROVIDER_ALLOW_PRIVATE_HOSTS."
+            "endpoints. Set to false to make provider connection tests, model discovery, and the "
+            "credential write path (POST / PATCH /v1/provider-credentials) refuse an internal api_base. "
+            "Chat dispatch (which dials the endpoint on every request) is not gated, so this is not a "
+            "general egress control. Also settable via OTARI_PROVIDER_ALLOW_PRIVATE_HOSTS."
```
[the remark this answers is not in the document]
njbrake: _Note: this reply was drafted by Claude via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

Intentional, and out of scope for this PR. `provider_allow_private_hosts` is not introduced here; it shipped in #441 as an opt-in, default-allow gate, a deliberate exception to the fail-closed guideline. The rationale: an operator-supplied `api_base` is master-key gated and standalone-only (the same trust level as a `config.yml` provider endpoint), and the home-lab / self-hosted use case depends on private-network endpoints (localhost, RFC 1918). The exception is documented in the config field description, the `validate_provider_api_base` docstring, and `docs/configuration.md`.

This PR only extends that existing gate to the credential write path; issue #443 explicitly keeps the allow-all default and defers any default change to the broader egress-policy work in #299. So the default stays as-is here by design.

tests/integration/test_provider_credentials_api.py
```diff
@@ -222,6 +222,73 @@ def test_delete_round_trip(
...
+) -> None:
+    """The gate only blocks internal hosts; a public api_base still saves."""
+    _with_key(monkeypatch)
+    monkeypatch.setenv("OTARI_PROVIDER_ALLOW_PRIVATE_HOSTS", "false")
+    resp = client.post(
+        "/v1/provider-credentials",
+        json={"instance": "openai", "api_key": "sk-1234", "api_base": "https://api.openai.com/v1"},
+        headers=master_key_header,
+    )
+    assert resp.status_code == 201, resp.text
```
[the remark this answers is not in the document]
njbrake: _Note: this reply was drafted by Claude via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

Fixed in 48c047f5: the test now uses a public IP literal (`8.8.8.8`) instead of `api.openai.com`, so the gate's address check runs without external DNS. The resolve path itself stays covered by the url_safety unit tests.

src/gateway/services/url_safety.py
```diff
@@ -231,9 +231,10 @@ async def validate_provider_api_base(url: str) -> None:
     check is skipped; scheme/host validation applies only once the gate is on.
 
     Scope: this gate covers the paths that *report* on an ``api_base`` (the
-    connection-test endpoints and model discovery). It does not gate the chat
-    dispatch that dials the endpoint for real, nor the credential write path that
-    persists it, so it is not a general egress control.
+    connection-test endpoints and model discovery) and the credential write path
+    that persists it (``POST`` / ``PATCH /v1/provider-credentials``). It does not
+    gate the chat dispatch that dials the endpoint for real on every request, so
+    it is not a general egress control.
```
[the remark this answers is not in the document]
njbrake: _Note: this reply was drafted by Claude via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

Fixed in 5d900465: the write-path text now names both routes distinctly, `POST /v1/provider-credentials` and `PATCH /v1/provider-credentials/{instance}`, in the docstring, the config field description, and the docs table.

src/gateway/core/config.py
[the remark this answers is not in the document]
njbrake: _Note: this reply was drafted by Claude via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

Fixed in 5d900465: the write-path text now names both routes distinctly, `POST /v1/provider-credentials` and `PATCH /v1/provider-credentials/{instance}`, in the docstring, the config field description, and the docs table.

docs/configuration.md
```diff
@@ -105,7 +105,7 @@ pricing:
...
+| `provider_allow_private_hosts` | bool | `true` | SSRF gate: allow a provider `api_base` that resolves to private/loopback/reserved hosts. On by default (unlike the other gates), because `api_base` is master-key gated and the home-lab use case depends on private endpoints. Set to `false` to make provider connection tests, model discovery, and the credential write path (`POST` / `PATCH /v1/provider-credentials`) refuse an internal `api_base`, so a blocked endpoint cannot be persisted. Chat dispatch (which dials the endpoint on every request) is not gated, so it is not a general egress control. Also settable via `OTARI_PROVIDER_ALLOW_PRIVATE_HOSTS`. |
 | `mode` | string | none | Operating mode (`"standalone"` or `"hybrid"`; the legacy value `"platform"` means hybrid): when unset, the mode is derived from the presence of `OTARI_AI_TOKEN`; when set explicitly it is enforced at startup, so `"hybrid"` without a token and `"standalone"` with a token both fail as conflicting configuration. |
```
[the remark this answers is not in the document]
njbrake: _Note: this reply was drafted by Claude via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

Fixed in 5d900465: the write-path text now names both routes distinctly, `POST /v1/provider-credentials` and `PATCH /v1/provider-credentials/{instance}`, in the docstring, the config field description, and the docs table.

src/gateway/api/routes/providers.py
```diff
@@ -369,6 +370,24 @@ def _validate_instance(instance: str, provider_type: str | None) -> None:
...
+    in the default allow-all state, but when the operator sets
+    ``OTARI_PROVIDER_ALLOW_PRIVATE_HOSTS=false`` a private/link-local/reserved
+    ``api_base`` is refused here, so a blocked endpoint cannot be persisted in
+    the first place (issue #443). Truthy check so an empty/absent api_base
+    (the "use the SDK default endpoint" case) is not treated as a URL.
+    """
+    if not api_base:
+        return
+    try:
+        await validate_provider_api_base(api_base)
```
khaledosman, line 386: `_reject_internal_host` rejects a host it cannot resolve (DNS-rebinding TOCTOU), so with the gate on this makes the write path strictly harder than the report path it mirrors: a **public** endpoint whose hostname is temporarily unresolvable — flaky resolver, split-horizon DNS, a compose sibling not up yet — can no longer be saved at all, where before it merely showed as "blocked" until it resolved. Each POST/PATCH also now does in-request DNS with no timeout bound. Both are defensible for an opt-in gate, but they should be a decision rather than a side effect of reusing the validator.

Same call site, unrelated nit: `from None` here where `_validate_instance` above uses `from exc`. `detail` is already `str(exc)`, so nothing extra leaks by keeping the chain for the logs.

docs/configuration.md
```diff
@@ -105,7 +105,7 @@ pricing:
...
+| `provider_allow_private_hosts` | bool | `true` | SSRF gate: allow a provider `api_base` that resolves to private/loopback/reserved hosts. On by default (unlike the other gates), because `api_base` is master-key gated and the home-lab use case depends on private endpoints. Set to `false` to make provider connection tests, model discovery, and the credential write path (`POST /v1/provider-credentials` and `PATCH /v1/provider-credentials/{instance}`) refuse an internal `api_base`, so a blocked endpoint cannot be persisted. Chat dispatch (which dials the endpoint on every request) is not gated, so it is not a general egress control. Also settable via `OTARI_PROVIDER_ALLOW_PRIVATE_HOSTS`. |
```
khaledosman: Two behaviours the gate now has on the write path that this row does not mention: the `api_base` must resolve when it is saved (an unresolvable host is refused, not just a private one), and non-http(s) schemes and hostname-less values are refused too — so `myproxy:8080` becomes a 400. Operators turning the gate on will hit the shape rules before they hit the SSRF rule.

tests/integration/test_provider_credentials_api.py
```diff
@@ -222,6 +222,78 @@ def test_delete_round_trip(
...
+    monkeypatch.setenv("OTARI_PROVIDER_ALLOW_PRIVATE_HOSTS", "false")
+    resp = client.patch(
+        "/v1/provider-credentials/openai",
+        json={"api_base": "http://169.254.169.254/latest/"},
+        headers=master_key_header,
+    )
+    assert resp.status_code == 400, resp.text
+    # The original public api_base is untouched.
+    stored = client.get("/v1/provider-credentials", headers=master_key_header).json()
+    assert stored[0]["api_base"] == "https://api.openai.com/v1"
```
khaledosman, line 294: The two cases most likely to surprise an operator are unpinned: an unresolvable hostname with the gate on (the behaviour change above), and PATCH *omitting* `api_base` leaving an existing internal base untouched — the omit-vs-null semantics the PR body claims to preserve. The rejection paths are well covered.

khaledosman (approved):
Validating before the existence check so nothing persists is the right ordering, and keeping the default allow-all preserves the home-lab case.

The main point inline: with the gate on, this makes a credential unsaveable whenever the endpoint's hostname does not resolve, which is a harder failure than the report path it mirrors — worth deciding deliberately and documenting either way. Plus a nit and a test gap.

---

src/gateway/api/routes/providers.py
njbrake: _Note: this reply was drafted by Claude via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

Deliberate decision: keeping the write path aligned with the report path. An unresolvable host stays refused (the anti-DNS-rebinding TOCTOU guard), so a temporarily-unresolvable public endpoint cannot be saved while the gate is on. Since the gate is opt-in and off by default, this stricter behavior only applies once an operator turns it on.

Documented in ecdf5108 (config reference + env-var row) and pinned by a new test (`test_create_rejects_unresolvable_host_when_gate_on`). Softening the write path to persist unresolvable hosts is deferred to the broader egress-policy work in #299.

docs/configuration.md
njbrake: _Note: this reply was drafted by Claude via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

Documented in ecdf5108. The config reference now states that with the gate on the write path requires an `http(s)` URL with a resolvable hostname, so a non-`http(s)` scheme, a hostname-less value (e.g. `myproxy:8080`), or an unresolvable host is refused with HTTP 400 before the private-range check. The env-var row got a shorter matching note.

tests/integration/test_provider_credentials_api.py
njbrake: _Note: this reply was drafted by Claude via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

Added in ecdf5108: `test_create_rejects_unresolvable_host_when_gate_on` covers the unresolvable-host behavior change, and `test_patch_omitting_api_base_keeps_existing_base_when_gate_on` pins the omit-vs-null semantics (omitting `api_base` leaves an existing internal base untouched).

PR #445, opened 2026-07-28
feat(dashboard): make the activity histogram the time-range selector

### LINKED ISSUE #440: Rework Usage page time-range selection: "All" means 30 days, and the custom picker only filters if you enter a time

Line references are against `origin/main` at `5fa80d05`.

## Summary

The Usage page's time-range control has two independent defects that both produce a plausible-looking but wrong window, plus a timezone bug. Together they mean the range shown often isn't the range requested, which reads as a cost-accounting problem rather than a filtering one. The control needs rework rather than patching.

Symptom that surfaced it: an "All" spend total that showed a bit over $6,000 one day and about $5,900 the next. The cost data was correct — the window had rolled forward.

## Defect 1: "All" returns a rolling 30 days

The preset carries a null span (`UsagePage.tsx:67`):

```ts
{ label: "All", seconds: null, bucket: "day" },
```

Selecting it clears `startDate` (`:509-513`), an undefined start is omitted from the query (`hooks.ts:637`), and the server reads omission as its *default*, not as unbounded (`usage.py:413-414`):

```python
end = end_date or datetime.now(UTC)
start = start_date if start_date is not None else end - _DEFAULT_SUMMARY_LOOKBACK
```

with `_DEFAULT_SUMMARY_LOOKBACK = timedelta(days=30)` (`usage.py:40`). So "All" is `[now - 30 days, now]`, and because it rolls with the clock the total drops whenever the day falling off the back outspent the day joining the front.

The intent is documented right above the preset list (`UsagePage.tsx:59-60`):

```ts
// Sub-day windows bucket hourly; longer ranges bucket daily. "All" has no lower
// bound, so no previous period to compare against (deltas hide).
```

"No lower bound" is what was meant, and the opposite of what happens.

Two related notes: `clearFilters()` resets to this preset (`:515-516`), so clearing filters silently scopes to 30 days; and true all-time is unreachable regardless, because `_MAX_SUMMARY_SPAN = timedelta(days=366)` (`usage.py:41`) pulls any earlier start forward (`:417-418`).

## Defect 2: the custom picker only filters if you also enter a time

Custom mode renders two bare `<input type="datetime-local">` fields (`UsagePage.tsx:584-598`). A `datetime-local` input emits a value only when **both** the date and the time-of-day are filled — pick a date alone and `event.target.value` is `""`.

That empty string is then treated as "no bound" (`:436-437`):

```ts
const winStart = customMode ? customFrom || undefined : startDate;
const winEnd = customMode ? customTo || undefined : undefined;
```

`undefined` means the param is omitted, which means the server applies its 30-day default. So a half-filled custom range doesn't fail and doesn't filter — it silently returns the last 30 days while the "Custom…" button stays visually active. That is worse than an error: it looks like it worked.

The underlying problem is that the control demands minute precision for a question that is almost always date-granular.

## Defect 3: local wall-clock times are read as UTC

`datetime-local` produces a naive string (`2026-07-27T00:00`) with no offset, and it is passed straight through. The server coerces naive input to UTC (`usage.py:409-412`):

```python
if start_date is not None and start_date.tzinfo is None:
    start_date = start_date.replace(tzinfo=UTC)
```

So a user outside UTC picking midnight gets a window shifted by their offset — at UTC-7, "today 00:00" silently means 17:00 the previous day. This is a correctness bug independent of the two above, and it also affects the drill-through into Activity, which forwards `start_date` (`:535`).

## Smaller issues in the same control

- **Custom mode always buckets daily** (`:438`, `const bucket = customMode ? "day" : preset.bucket`), so a three-hour custom range collapses to a single point.
- **No `From <= To` validation.** The server responds by setting `start = end` (`usage.py:415-416`), yielding an empty window rather than a complaint.
- **Clamping is invisible.** A custom range wider than 366 days is silently narrowed with no indication.
- **The effective window is never displayed.** `UsageSummary` already echoes the `start_date` / `end_date` the server actually aggregated over, and the page doesn't render them — which is why all of the above fail silently instead of being self-evident.

## Proposed direction

A date-range control rather than two datetime fields:

1. **Date granularity by default**, with time-of-day as an optional refinement rather than a requirement. Most cost questions are "last month" or "the 3rd through the 10th".
2. **Delete the "All" preset.** Narrowing `Preset.seconds` from `number | null` to `number` makes Defect 1 unrepresentable rather than renamed, and the compiler then finds the dead null branches (`:423-424`, `:460`, `:512`, and the `refresh()` guard). A side benefit: period-over-period deltas start working for every preset, since all become bounded.
3. **Add bounded long-range presets** in its place — `90d` and `12mo` (`366 * DAY_S`, matching `_MAX_SUMMARY_SPAN`) — so removing "All" doesn't cap the page at 30 days when the server serves a year. Calendar-relative options ("this month", "last month", "quarter to date") are what people usually reach for and are worth considering alongside.
4. **Always show the effective window**, sourced from the server's echoed range, so any default or clamp is visible.
5. **Decide the timezone contract explicitly** and state it in the UI — either send an offset from the client or label the control UTC. Silent reinterpretation is the thing to avoid.
6. **Pick the bucket from the range length**, not from whether the mode is custom.
7. **Validate the range in the client** before querying.

## Note on the Activity page

Activity has its own copy of the presets, including `{ key: "all", label: "All", seconds: null }` (`ActivityPage.tsx:82`). That one is currently truthful — it queries `GET /v1/usage`, whose `_usage_filters` (`usage.py:145-148` equivalent) applies no default and no clamp, so omission really is unbounded there. So "All" means all time on Activity and 30 days on Usage, in the same dashboard. Whatever shape the new control takes is worth sharing between the two pages so the labels can't drift apart again.

njbrake:
## Description

Reworks the Usage and Activity time-range controls around a request-volume histogram, the pattern CloudWatch Logs Insights, Kibana, and Grafana/Loki use for log exploration. The selection lives on the chart itself: full-height handles at the window edges zoom (react-aria slider thumbs, pointer-captured, 0.1-bucket steps), a slim strip along the axis band pans, regions outside the window dim in place, and −/+ buttons step the zoom. Zooming out past the extent promotes to the next larger preset (30d to 90d to 12mo), and Reset restores the full extent. Presets set the extent the chart spans; the tiles, breakdowns, and log follow the selection.

This fixes the three defects from the issue:

- "All" on Usage silently meant a rolling 30 days (an omitted start hit the server's 30-day default), so the total drifted with the clock. Usage drops "All" for bounded 90d and 12mo presets.
- The old custom picker (two `datetime-local` inputs) only filtered when both a date and a time were entered, and read local wall-clock times as UTC. The selection is now bucket-based and UTC-aligned, matching the summary endpoint's bucketing and the existing "times in UTC" captions.
- The bucket is chosen from the active preset's extent, and the effective window is always captioned, so a default or clamp is never invisible.

Other changes riding along, all part of the same control rework:

- Activity keeps a truthful "All" (its list endpoint is genuinely unbounded); both pages now share the timeline, presets, and window math (`web/src/lib/timeRange.ts`), so the labels can't drift apart again.
- Active model/user/key/status/priced filters render as removable chips with an "Add filter" toggle for the pickers.
- The Refresh button becomes an icon with an "updated Ns ago" freshness label, riding the timeline's preset row on Usage and Activity, and added to the Overview header.
- Interaction details: drag feedback is CSS-positioned shades (the recharts tree never re-renders mid-drag), the window commits once on release snapped outward to whole buckets, the histogram spans the preset extent independently of the selection so there is always context to zoom back out into, bar tooltips keep working (the pan strip overlays the axis band, not the plot), and `touch-pan-y` keeps mobile scrolling intact.

Verified end-to-end against a live gateway with seeded data (drag-zoom, pan, step zoom with preset promotion, tooltips, mobile layout, drill-through to Activity).

## PR Type

- [x] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes #440

## Checklist

Notes: this is a dashboard-only change (no gateway API contract change, so no OpenAPI regeneration needed). Frontend checks run instead: `npm --prefix web run typecheck`, `npm --prefix web test` (327 tests), and the committed bundle was rebuilt (`npm --prefix web run build`).

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

Claude Code (Opus 4.8 and Fable 5), directed by @njbrake

Design direction (histogram-as-selector, overlay handles, preset promotion on zoom-out) was decided by @njbrake across several review rounds against the running dashboard.

## Summary

- Replaced Usage and Activity time controls with a shared request-volume timeline supporting presets, zooming, panning, reset, and UTC-aligned windows.
- Updated Usage to use bounded 90-day and 12-month ranges, while preserving Activity’s unbounded “All” option.
- Made selected windows drive charts, metrics, breakdowns, and request logs, with clearer effective-window labels and improved bucket selection.
- Added removable filter chips with an “Add filter” control.
- Added refresh freshness indicators and icon-based refresh actions across dashboard pages.
- Added comprehensive tests for timeline interactions, time-range calculations, filters, refresh behavior, and page queries.
- Rebuilt the dashboard production assets and updated their references.


njbrake:

Merging without waiting for review: this change is dashboard-only (no gateway API, schema, or config surface is touched), all seven checks are green, and it reverts cleanly as a single squash commit if needed. The intent here is near-term usability; the time-range UI may still be completely overhauled later, so treat this as a usable baseline rather than a final design. @khaledosman post-merge review feedback is still very welcome.

khaledosman (commented):
Reviewing after the fact (already merged as ffad6b5), so these are follow-ups rather than blockers. Substantial work and the three defects the issue named are genuinely fixed.

Two of the notes below are the same class of bug this PR set out to kill, re-entering through the Activity page; the rest are smaller.

Worth keeping and not obvious: the capture-phase arrow-key intercept in `ActivityTimeline.tsx`. react-aria's built-in step is the 0.1-bucket pointer step, which the bucket-snapping commit rounds away, making arrows a silent no-op. The comment explaining it is the reason a future reader will not "simplify" it back into a bug.

web/src/pages/ActivityPage.tsx
```diff
@@ -318,19 +320,67 @@ export function ActivityPage() {
...
+  const winOutsideExtent = Boolean(
+    win.start && extentWin.start && new Date(win.start).getTime() < new Date(extentWin.start).getTime(),
+  );
+  const extentKey = winOutsideExtent ? CUSTOM_KEY : range;
+  const extentBucket = winOutsideExtent
+    ? bucketForWindow(win.start as string, win.end)
+    : (extentPreset?.bucket ?? "day");
+  const contextFilters: UsageFilters = useMemo(
+    () => ({
+      start_date: winOutsideExtent ? win.start : extentWin.start,
```
khaledosman, line 344: The "All" preset reintroduces the exact class of bug this PR set out to kill. With `range=all`, `extentWin.start` is undefined, so `contextFilters.start_date` is omitted and the summary endpoint applies its 30-day default (`usage.py:414`). The histogram then shows 30 days while the caption above it reads "All time", and dragging inside it cannot reach anything older than that.

The PR's own third fix is "a default or clamp is never invisible" -- this is that default, invisible. Either send an explicit `start_date` for the histogram and caption its span separately from the list window, or drop "All" from Activity the way it was dropped from Usage.

web/src/pages/ActivityPage.tsx
```diff
@@ -318,19 +320,67 @@ export function ActivityPage() {
...
+  const contextSummary = useUsageSummary(contextFilters, extentBucket);
+  const timelineSeries = (contextSummary.data?.series ?? []).map((p) => ({
+    bucketStart: p.bucket_start,
+    requests: p.requests,
+  }));
+
   const rows = usage.data ?? [];
   const totalIsExact = count.isSuccess && !count.isPlaceholderData;
   const total = totalIsExact ? (count.data?.total ?? 0) : null;
   const anyFilter = Boolean(
```
khaledosman, line 363: `anyFilter` still counts the default 24h window via `hasWindow`. UsagePage was fixed for precisely this (`timeFiltered = customMode || preset.key !== USAGE_DEFAULT_KEY`); Activity was not.

So a brand-new gateway's Activity page reads "No requests match these filters." instead of "No requests recorded yet." `ActivityPage.test.tsx` was updated to click "All" first to reach the onboarding copy, which pins the behavior rather than the intent.

web/src/lib/timeRange.ts
```diff
@@ -0,0 +1,168 @@
...
+// long-range presets (90d, a year) in place of the old rolling "All", so the
+// page still reaches the full span the server will serve without an unbounded,
+// clock-rolling window whose total silently changes day to day.
+export const USAGE_PRESETS: RangePreset[] = [
+  { key: "1h", label: "Last hour", seconds: HOUR_S, bucket: "hour" },
+  { key: "24h", label: "24h", seconds: DAY_S, bucket: "hour" },
+  { key: "7d", label: "7d", seconds: 7 * DAY_S, bucket: "day" },
+  { key: "30d", label: "30d", seconds: 30 * DAY_S, bucket: "day" },
+  { key: "90d", label: "90d", seconds: 90 * DAY_S, bucket: "day" },
+  { key: "12mo", label: "12mo", seconds: MAX_SPAN_DAYS * DAY_S, bucket: "day" },
```
khaledosman, line 42: `12mo` is exactly `_MAX_SUMMARY_SPAN`. But `start` is anchored client-side by `isoAgo` while the server resolves `end` at request time (`usage.py:413`), so `end - start` always exceeds 366 days by the round-trip delta and `usage.py:417-418` always pulls the start forward.

Harmless at millisecond scale, but it means this preset never actually spans 366 days, and a browser clock behind the server's silently shortens it further. `365 * DAY_S` leaves headroom and never trips the clamp.

web/src/components/ActivityTimeline.tsx
```diff
@@ -0,0 +1,422 @@
...
+    selRef.current = next;
+    setSelState(next);
+  };
+  const dragging = useRef(false);
+  useEffect(() => {
+    if (!dragging.current) setSel(windowToRange());
+    // starts is derived from series each render; resync on the window bounds and
+    // the extent length, not the fresh array identity.
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [windowStart, windowEnd, n]);
```
khaledosman, line 155: The resync keys on `[windowStart, windowEnd, n]`, so a series whose buckets roll forward while both bound strings and the length stay identical leaves the thumbs on stale positions. Narrow in practice, since the bounds usually move when the series does. Flagging because the `eslint-disable` above is doing real work and its comment explains the array-identity problem rather than this gap.

web/src/components/ActivityTimeline.tsx
```diff
@@ -0,0 +1,422 @@
...
+                  style={{ width: `${100 - hiPct}%` }}
+                />
+              </>
+            ) : null}
+
+            {/* Pan strip along the axis band: slides the window without resizing
+                it. Sits below the bars, so plot-area tooltips are unaffected. */}
+            <div
+              aria-hidden
+              className="absolute bottom-0 z-[1] cursor-grab touch-none rounded bg-[var(--otari-brand)]/10 hover:bg-[var(--otari-brand)]/20 active:cursor-grabbing"
```
khaledosman, line 369: The pan strip is `aria-hidden` with pointer-only handlers, so panning has no keyboard path; arrow keys resize an edge instead of sliding the window. Zoom and Reset are real buttons so the control stays operable, but a keyboard user cannot reach a mid-extent window without resizing both edges in sequence.

njbrake:
@khaledosman whoops sorry about the merge, mistake on my side 🫠 glad nothing big was wrong though and I'll handle the follow ups

PR #449, opened 2026-07-29
feat(dashboard): surface failing requests so admins see dropped traffic

### LINKED ISSUE #317: Surface failing requests so admins notice missing-billing (and other) rejections

njbrake:
## Description

An admin had no way to see that the gateway was dropping traffic. The `require_pricing`
banner told you the config was in a reject state, but not that live requests were actually
failing, or how many.

The premise in the issue was that the data already existed and only needed surfacing. It
turned out that was not quite true for the case the issue leads with. The missing-pricing
gate refunded the budget reservation and raised the 402 without writing a usage row
(`_pipeline.py`), so a gateway-side rejection left no trace anywhere. A banner counting
`status=error` would have read "0 failed" at the exact moment unpriced models were being
rejected. Every admin view was blind to it: the activity log, the overview error rate, and
the recent-activity list all showed nothing.

So this is two small changes:

**Record the rejection.** Both missing-pricing gates (the chat/messages/responses pipeline
and the pass-through routes used by embeddings, images, and rerank) now log a
`status="error"` usage row with `cost=null` before raising the 402. The reservation refund
is unchanged, so nothing is billed and nothing stays reserved; the row exists purely to
make the drop visible and countable.

**Read that count in the banner.** The pricing alarm now shows "N requests failed in the
last hour" with a link into the activity log filtered to those failures
(`/activity?status=error&range=1h`), so the reject state reads as an active incident
rather than a static config note. It is served by a new `useFailureCount` hook that polls
through TanStack Query's `refetchInterval`. The window is resolved inside `queryFn` rather
than in the query key, which keeps the key stable (a "now"-derived key would mint a new
cache entry on every render) and re-anchors on every refetch, so a tab left open keeps
reporting the last hour instead of quietly widening to everything since it was opened. The
query only runs while the alarm is up.

The count deliberately covers every failure class, not only the pricing rejections. The
operator's question in this state is "is traffic getting through", and over-reporting a
failure is safer than a banner reading "0" while requests are being dropped.

One knock-on fix: the activity detail panel's error summary said "The provider returned an
error." That was accurate when only provider failures were logged. Gateway-side rejections
land in that list now, so the copy is source-neutral ("This request failed."). Provider
diagnostics still stay server-side.

Not included: the issue also floats an "allow unpriced models" toggle in Settings. The
banner's existing "Enable default pricing" button already is that one-click mitigation, so
nothing was added for it.

### Verification

Beyond the test suites, this was smoke tested against a running gateway with
`require_pricing` on and a deliberately fake provider key (the gate rejects before
dispatch, so no real credential is needed):

| Check | Result |
| --- | --- |
| Unpriced chat requests | `402 402 402` |
| Unpriced embeddings (pass-through gate) | `402` |
| Rejections logged, `cost=null` | 4 rows, correct endpoints |
| `/v1/usage/count?status=error&start_date=…` | `{"total": 4}` |
| Budget untouched after 4 rejections | `total_spend=0.0`, `total_reserved=0.0` |
| Budget-exempt key still skips the gate | `502`, not `402` |
| "Enable default pricing" clears the 402 | `502`, past the gate |

In the browser: the banner reported the live count, the link landed on the activity log
with the 1h preset and the Status: Error filter pre-applied, and the count updated itself
from 2 to 7 in 57 seconds with no reload. The Overview page's "Needs attention" strip, the
error-rate tile, and Recent activity all populated as a side effect, having been silent
before.

## PR Type

- [x] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes #317

Part of #299. Adjacent to #303 (usage and analytics) and #304 (request log viewer): showing
a safe per-row failure reason in the detail panel needs a way to tell a gateway message
from a raw provider message, which belongs with that work rather than here.

## Checklist

Tests added: two integration cases in `tests/integration/test_require_pricing.py` (the chat
gate and the pass-through gate), both confirmed to fail without the backend change, and
five dashboard cases in `PricingWarning.test.tsx` including one that pins the re-anchoring
behavior (confirmed to fail against a frozen-window implementation). The OpenAPI spec is
unchanged: no route signature or schema moved.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

Written by Claude working from the issue, then reviewed and smoke tested with @njbrake in
the loop. The decision to add the backend logging (rather than ship a banner that would
have counted zero) and the scope boundary around the detail-panel reason were discussed
rather than assumed.

**NOTE:**

## Summary

- Record missing-pricing rejections as error usage entries while preserving budget refunds.
- Add failure visibility to the PricingWarning banner, including live counts and filtered activity-log links.
- Support source filtering in activity details and use neutral failure messaging.
- Update dashboard documentation and rebuilt dashboard assets.
- Expand integration and dashboard tests for pricing gates, budget exemptions, failure counts, polling, and filtering.

## Benefits

Administrators can identify failed gateway requests more easily, while pricing enforcement and budget accounting remain consistent.


src/gateway/api/routes/_passthrough.py
```diff
@@ -219,9 +219,28 @@ async def run_passthrough(
             and pricing_required_but_missing(pricing, require_pricing=config.require_pricing)
         ):
             await refund_reservation(db, reservation)
+            no_pricing_detail = no_pricing_error_detail(model)
+            # Record the rejection so dropped traffic is visible in the activity
+            # log and countable as an error, rather than only reaching the
+            # operator as a user complaint. cost stays null: nothing was spent.
+            await log_writer.put(
```
khaledosman, line 226: This is the third hand-built `UsageLog` literal in this file (also the success row at ~263 and the provider-error row at ~290), all eleven fields repeated. A local `_usage_row(status, **overrides)` helper would make the next field addition a one-line change instead of three sites to keep in sync.

src/gateway/api/routes/_passthrough.py
```diff
@@ -219,9 +219,28 @@ async def run_passthrough(
...
+                UsageLog(
+                    id=str(uuid.uuid4()),
+                    api_key_id=api_key_id,
+                    user_id=user_id,
+                    timestamp=datetime.now(UTC),
+                    model=resolved.model,
```
khaledosman, line 232: Logs `model=resolved.model` (bare, e.g. `text-embedding-3-small`) where `_pipeline.py` logs `model=model`, the full selector (`openai:gpt-4o`). Each matches its own file's success path, so this PR does not introduce the split, but it does extend it to the new error rows: in the Activity log a rejected embeddings row and a rejected chat row render their Model column differently, and the model filter matches one form or the other.

Not necessarily worth changing here, but the new tests do not pin it either (the chat test asserts the model string, the passthrough one does not), so asserting the expected embeddings model would at least record which form is intended.

tests/integration/test_require_pricing.py
```diff
@@ -85,6 +85,50 @@ def test_unpriced_model_rejected_with_402(strict_pricing_client: TestClient) ->
...
+    resp = c.post(
+        "/v1/embeddings",
+        json={"model": "openai:text-embedding-3-small", "input": "hi", "user": "priced-user"},
+        headers=_MASTER_HEADER,
+    )
+    assert resp.status_code == 402
+
+    rows = c.get("/v1/usage", params={"status": "error"}, headers=_MASTER_HEADER).json()
+    assert len(rows) == 1
+    assert rows[0]["endpoint"] == "/v1/embeddings"
```
khaledosman, line 139: Missing the case that guards the invariant this change depends on: a budget-exempt key skips the gate entirely, so no error row is written and no spend moves. The PR description's smoke table covers it (`502`, not `402`), but this is the one place where a future refactor of the `budget_exempt` condition would silently start writing rows with `counts_toward_budget=False`, which the dashboard then classifies as imported and offers up for bulk delete. Cheap to pin here.

web/src/api/hooks.ts
```diff
@@ -683,6 +684,40 @@ export function useUsageCount(filters: UsageFilters, enabled = true) {
...
+      };
+      return apiFetch<UsageCount>(`/v1/usage/count?${usageParams(filters).toString()}`);
+    },
+    enabled,
+    refetchInterval: FAILURE_COUNT_POLL_MS,
+    refetchOnWindowFocus: true,
+    staleTime: 0,
+    // A failed count is not worth surfacing: it sits beside its own alarm, and
+    // the next poll retries anyway.
+    retry: false,
```
khaledosman, line 724: Resolving the window inside `queryFn` rather than in the key is the right call and the re-anchoring test pins it well.

One consequence worth confirming is intended: `retry: false` plus `staleTime: 0` means a transient failure leaves `failures.data` on the last successful value until the next 60s tick, so the banner can display a stale count for up to a minute after the gateway starts erroring. Advisory number, so probably fine, but it is the opposite of failing loud.

web/src/components/PricingWarning.tsx
```diff
@@ -27,6 +45,17 @@ export function PricingWarning() {
           <span>
             Requests are rejected until pricing is set (<code>require_pricing</code> is on). Enable default pricing to
             meter new models with public rates right away.
+            {failureCount > 0 ? (
+              <>
+                {" "}
+                <strong className="font-semibold">
+                  {failureCount.toLocaleString()} {failureCount === 1 ? "request" : "requests"} failed in the last hour.
+                </strong>{" "}
+                <NavLink to="/activity?status=error&range=1h&source=gateway" className="underline underline-offset-2">
```
khaledosman: `NavLink` is `Link` plus active-state styling, and the active state can never apply here (the banner is never rendered on the activity route it points at). `Link` expresses the intent without the dead machinery.

khaledosman (approved):
The premise correction is the valuable part here: the issue assumed the data already existed and it did not. A banner over `status=error` without the backend write would have read "0 failed" at precisely the moment traffic was being dropped. Catching that before building the UI, and covering both gates rather than only the one the issue names, is the right call.

I checked the budget-safety property this change could plausibly have broken: both gates are guarded by `not budget_exempt`, so `counts_toward_budget=not budget_exempt` is always `True` on these rows. They cannot be mistaken for imported rows, which means bulk delete and set-price cannot reach them and the `source=gateway` count does not miss them. That invariant deserves a test (see inline).

Five non-blocking notes inline.

Landing order: this, #450, and #451 all edit `docs/dashboard.md` and rewrite the same hashed bundle assets, so whoever lands second needs a rebase plus `npm --prefix web run build`.

src/gateway/api/routes/_passthrough.py
njbrake: _Note: this reply was drafted by Claude via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

Agreed, and deliberately out of scope here: two of the three literals predate this PR, so the helper is a refactor of existing code rather than part of this change. Tracked in #465, along with the bare-model versus full-selector split this PR pinned but did not change.

web/src/api/hooks.ts
njbrake: _Note: this reply was drafted by Claude via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

Intended, though worth stating rather than leaving implicit. The count is advisory and sits beside an alarm that is already firing on config state, so holding the last value for up to a minute beats the banner dropping its count on a single failed poll. The alarm itself does not depend on this query, so a failed count never hides the reject state. If we would rather it fail loud, the change is to hide the count on error, not to retry.

PR #450, opened 2026-07-30
feat(dashboard): install the dashboard to a home screen with the Otari icon

njbrake:
## Description

Saving the self-hosted dashboard to a phone home screen produced a generic shortcut. The app shipped only an SVG favicon, so Android had no manifest to read a name or icon from, and iOS looked for an `apple-touch-icon` it never found.

This adds a web app manifest and PNG app icons under `web/public/pwa/`, all generated from the existing mark in `web/public/favicon.svg` (white glyph on the `--otari-brand` teal):

| File | Purpose |
| --- | --- |
| `icon-192.png`, `icon-512.png` | Android manifest icons (both sizes are needed for an install offer) |
| `icon-maskable-512.png` | Extra padding, so a launcher's icon shape crops the background instead of the glyph |
| `apple-touch-icon.png` (180px) | What iOS Safari uses; it ignores the manifest entirely |
| `manifest.webmanifest` | Name "Otari Dashboard" / short name "Otari", `display: standalone`, `start_url: /`, brand theme color, `#f6f9fa` splash background |

### Wiring

- `web/index.html`: manifest link, apple-touch-icon, `theme-color`, and the `apple-mobile-web-app-*` tags. Status bar style is `default` rather than `black-translucent`, since the app shell has no safe-area insets and would otherwise slide underneath the status bar.
- `src/gateway/main.py`: mounts `/pwa` from the built bundle next to the existing `/assets` mount, so it rides along with the standalone dashboard. Hybrid mode serves the tutorial at `/`, which is not an app worth installing, so nothing is mounted there. `/pwa/` gets the favicon's day of public caching rather than `/assets`' immutable year, because the filenames are not content-hashed.
- Rebuilt and committed the dashboard bundle. `AGENTS.md` and the bundled user guide (`docs/dashboard.md`) gain notes on the feature and on how the icons are regenerated. A stale session-storage description the review turned up was split out to #451.

### Deliberately not included

No service worker. Chrome [dropped the fetch-handler requirement](https://developer.chrome.com/blog/update-install-criteria) for menu-driven installs in 108 (mobile) and 112 (desktop), so the install works without one, and the dashboard is live API data: caching a shell offline would mostly create staleness bugs alongside the existing build-staleness poll.

One caveat worth knowing when trying this out: installability requires a secure context. Over `https://` or `http://localhost` you get "Install app"; a self-hosted gateway reached at a plain-HTTP LAN address gets a plain bookmark shortcut instead. iOS "Add to Home Screen" uses the apple-touch-icon either way.

### Verification

- Booted the gateway and loaded it in real Chromium: `Page.getAppManifest` reports zero parse errors, and all three manifest icons decode at their declared sizes.
- Measured the rendered PNGs: the maskable icon's ink fits within 79.9% of the canvas, inside the 80% safe circle a launcher mask can crop to, while the plain icons stay at the larger 70% width that suits an unmasked tile.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

None filed; this came from trying to install the self-hosted dashboard on a phone.

## Checklist

- [x] If the API contract changed, I regenerated the OpenAPI spec (`uv run python scripts/generate_openapi.py`). No API contract change; the new routes are a static mount excluded from the schema, and `make openapi-check` is clean.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

Claude Code (Opus 5)

Implementation, tests, icon generation, and this description were produced by Claude Code working from @njbrake's request, with the reasoning and decisions his. The icons were generated from the existing `favicon.svg` mark with `rsvg-convert` rather than hand-drawn, and the manifest was checked against a real Chromium via the DevTools Protocol.

## Summary

- Added installable PWA support for the Otari dashboard on Android and iOS, including branded icons, app metadata, and home-screen configuration.
- Added backend serving and caching for PWA assets, with coverage for missing or stale dashboard bundles.
- Documented installation steps, secure-context requirements, session behavior, and icon regeneration.
- Rebuilt the dashboard bundle to include the new PWA metadata and documentation page.
- No service worker or API contract changes.


src/gateway/main.py
```diff
@@ -267,6 +271,13 @@ async def favicon() -> Response:
             StaticFiles(directory=dashboard_dir / "assets"),
             name="dashboard-assets",
         )
+        # Installing the dashboard to a phone home screen needs the manifest and
+        # the PNG icons it points at (index.html links them from /pwa/). Only the
+        # standalone dashboard is an app worth installing, so this rides along
+        # with it rather than being served in hybrid mode next to the tutorial.
+        pwa_dir = dashboard_dir / "pwa"
+        if pwa_dir.is_dir():
```
khaledosman, line 279: This guard is untested. `test_pwa_assets_are_absent_without_dashboard` monkeypatches `get_dashboard_dir` to `None`, which exercises the outer branch, not this one.

The case it actually protects against is a dashboard bundle built before this PR (an older wheel, a stale Docker layer): `assets/` exists, `pwa/` does not, so `/pwa` silently 404s and the manifest link in the shipped `index.html` dangles. A `tmp_path` bundle with `assets/` but no `pwa/` would cover it in a few lines.

web/public/pwa/manifest.webmanifest
```diff
@@ -0,0 +1,32 @@
...
+  "id": "/",
+  "name": "Otari Dashboard",
+  "short_name": "Otari",
+  "description": "Otari admin dashboard: browse and price models, manage aliases, and toggle runtime settings.",
+  "start_url": "/",
+  "scope": "/",
+  "display": "standalone",
+  "orientation": "any",
+  "background_color": "#f6f9fa",
+  "theme_color": "#4e8295",
```
khaledosman, line 11: `theme_color` here and the `<meta name="theme-color">` in `web/index.html` both hardcode `--otari-brand`. A manifest cannot read CSS custom properties so this is unavoidable, but nothing links the three copies: changing the design token leaves the installed app's chrome on the old teal with no build-time signal. A comment next to the token definition naming both files would keep them in sync.

src/gateway/main.py
```diff
@@ -55,6 +55,10 @@
 # immutable; the middleware marks these public and cacheable for a year, since
 # StaticFiles does not set Cache-Control on its own.
 _CACHEABLE_PREFIXES = ("/assets/",)
+# The install manifest and home-screen icons, public like the page that links
+# them. Their names are stable rather than content-hashed, so they get a day of
+# caching like the favicon, not the immutable year /assets gets.
+_SHORT_CACHE_PREFIXES = ("/pwa/",)
```
khaledosman, line 61: Worth noting this prefix covers `manifest.webmanifest` itself, not just the PNGs. Since these filenames are not content-hashed, a manifest change (new name, changed icon set) is pinned for a day in every already-installed app. That reads like the intended trade-off given the comment, just flagging that it applies to the metadata and not only the images.

khaledosman (approved):
Clean and well-scoped. The mount rides `get_dashboard_dir()` so it stays standalone-only, the cache tier is correctly distinguished from `/assets`' immutable year, and `test_pwa_manifest_and_icons_are_served` verifying that every icon the manifest advertises actually resolves is the right test to write.

Checked and not a problem: `.webmanifest` resolves to `application/manifest+json` from Python 3.14's builtin mimetypes map, so the slim Docker image serves the right content type without a system `mime.types`.

Three non-blocking notes inline.

Landing order: this, #451, and #449 all edit `docs/dashboard.md` and rewrite the same hashed bundle assets, so whoever lands second needs a rebase plus `npm --prefix web run build`.

src/gateway/main.py
njbrake: _Note: this reply was drafted by Claude via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

Right on both counts, and yes, that is the intended trade-off. Leaving it as is, with the reasoning written down here.

The manifest is the thing most likely to change (a new icon set, a renamed app), and it is the thing pinned longest relative to its size. What softens it is that the manifest is only read at install time and on subsequent visits, not per launch of an installed app: a stale copy delays a metadata change by up to a day for someone who visits in that window, and does not affect anyone already installed, since the launcher keeps the icons it captured at install. Re-installing is what actually picks up a new icon set, and that is a deliberate user action rather than something a cache header controls.

The alternative would be splitting the manifest out to its own route with a shorter max-age, or dropping it to no-store so it revalidates. That buys faster propagation of a change we expect roughly never, at the cost of a second serving path for one file, so it did not seem worth the moving part. Worth revisiting if the manifest starts carrying anything that changes on a normal release cadence, `screenshots` or shortcuts being the likely candidates.

src/gateway/main.py
njbrake: _Note: drafted by Claude via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

Good catch, and the distinction was the part I had missed: the existing test exercises the outer branch, not this guard. Fixed in e24863f5 with `test_dashboard_without_pwa_dir_still_starts`, which builds a bundle that has `assets/` but no `pwa/`, the older-wheel/stale-layer case you describe.

I checked it actually bites rather than assuming: with the guard removed the new test fails on `RuntimeError: Directory '.../pwa' does not exist` from StaticFiles construction, and passes with it. The test also asserts the stale `index.html` still gets served and its dangling manifest link merely 404s.

web/public/pwa/manifest.webmanifest
njbrake: _Note: drafted by Claude via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

Agreed, and there was no signal linking the copies. Fixed in e24863f5 with a comment at the `--otari-brand` definition in `web/src/styles/globals.css`, naming both places that duplicate the hex where a custom property cannot reach (the `theme-color` meta tag and `theme_color`/`background_color` in the manifest) and saying outright that nothing fails the build if they drift.

PR #451, opened 2026-07-30
docs(dashboard): correct the stale session-storage description of sign-in

njbrake:
## Description

Split out of #450, where reviewing a nearby change turned this up.

The dashboard guide still described the pre-cookie sign-in, and both halves of what it said are now backwards:

| The guide said | What actually happens |
| --- | --- |
| The master key "is held only in this browser tab's session storage" | The key is POSTed once to `/v1/auth/session` and exchanged for an HttpOnly cookie. The browser never stores the key; only a non-secret `otari.dashboard.hasSession` marker lands in localStorage, so the app can render signed-in without probing the server first. |
| It "does not persist across tabs or survive closing the tab" | The cookie is persistent, with a `dashboard_session_ttl_hours` lifetime (168 hours by default), and sessions are rows in `dashboard_sessions`, so a sign-in deliberately survives closing the tab and restarting the gateway. |

Three places said it, and all three are corrected: the two-key table's "Where it lives" row, first-run step 4, and the security note. The security note is retitled "A session cookie, not a stored key" and now records the cookie's actual attributes (`HttpOnly`, `SameSite=Strict`, `Secure` when the request arrives over HTTPS) along with what sign-out revokes and the fact that rotating the master key logs other browsers out.

Checked against the implementation rather than the old prose: `dashboard_session_service.py:142` for the cookie attributes, `core/config.py:207` for the 168-hour default, `web/src/auth/AuthContext.tsx:12` for what the browser actually retains, and `api/routes/settings.py:412` for the rotation behavior. Swept the repo for the same claim elsewhere; the only other hit is `web/src/components/Login.test.tsx:49`, a test asserting the key is *not* in session storage, which agrees with the corrected text.

The guide is bundled into the dashboard (`DocsPage` imports it with `?raw`), so the committed bundle is rebuilt alongside it. That means this PR and #450 both touch the built bundle: whichever lands second will need a rebase and a rebuild.

## PR Type

- [ ] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [x] Documentation
- [ ] Infrastructure / CI

## Relevant issues

None filed. Noticed while reviewing #450.

## Checklist

- [x] I have added or updated tests that cover my change (`tests/unit`, `tests/integration`). Documentation only, so no behavior to test; the existing suites and the bundle staleness check pass.
- [x] If the API contract changed, I regenerated the OpenAPI spec (`uv run python scripts/generate_openapi.py`). No API change.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

Claude Code (Opus 5)

Claude spotted the stale text while reviewing #450, verified the current behavior against the source files cited above, and wrote the correction and this description. The reasoning and decisions are @njbrake's.

## Summary

- Updated dashboard documentation to accurately describe cookie-based sign-in, session persistence, revocation, and security guidance.
- Clarified master-key handling and browser storage behavior.
- Rebuilt the committed dashboard bundle to include the updated documentation.

## Benefits

- Gives users accurate guidance about dashboard sessions and security.
- Explains how to revoke sessions and protect access on shared machines.


docs/dashboard.md
```diff
@@ -95,11 +95,13 @@ can skip this step if all your providers live in `config.yml`.
...
 Browse to the gateway root, for example `http://localhost:8000/`. You land on a
-sign-in screen. Paste your master key and select **Sign in**. The key is held
-only in this browser tab's session storage and sent directly to this gateway; it
-is never persisted to disk by the browser and is cleared when you sign out. If
-you are on a fresh install and are not sure where your key is, the "First run?
-Where to find your key" hint on the sign-in screen points you back at the logs.
+sign-in screen. Paste your master key and select **Sign in**. The key is sent
+once to this gateway and exchanged for a session cookie; the browser never stores
+the key itself, so it cannot be read back out of the page. The sign-in lasts
+`dashboard_session_ttl_hours` (a week by default) and survives closing the tab
```
khaledosman: "survives closing the tab and restarting the gateway" holds only while the session table persists. The default containerized SQLite with no mounted volume loses `dashboard_sessions` on restart, so a Docker user following this will be surprised. Suggest qualifying it, e.g. "...and restarting the gateway (as long as its database persists)".

docs/dashboard.md
```diff
@@ -187,10 +189,13 @@ For how users, keys, and budgets fit together and the management endpoints behin
...
   so put the gateway behind HTTPS or a trusted reverse proxy before signing in
   from another host.
-- **Session storage, not local storage.** The dashboard keeps the master key in
-  the browser tab's session storage, so it does not persist across tabs or
-  survive closing the tab, and signing out clears it along with any cached admin
-  data.
+- **A session cookie, not a stored key.** The dashboard trades your master key
+  for an HttpOnly cookie (`SameSite=Strict`, and `Secure` whenever the request
+  arrives over HTTPS), so the key itself is never kept in the browser and script
+  on the page cannot read the cookie. Signing out revokes the session on the
```
khaledosman, line 217: The note correctly documents that the session now outlives the tab, but the flip side is that there is no idle timeout: a 168-hour cookie on an unattended browser stays signed in. The old (wrong) text at least implied tab-close logout. Worth one clause telling operators to sign out on shared machines, since that is no longer automatic.

khaledosman (approved):
Checked every claim against the implementation and they hold: `dashboard_session_service.py:157-163` sets `httponly=True`, `samesite="strict"`, and `secure` from the request scheme; `core/config.py:207` defaults `dashboard_session_ttl_hours` to 168. The old text was wrong in both directions and the correction is accurate.

Two optional wording notes inline; neither blocks.

Heads up on landing order: this, #450, and #449 all edit `docs/dashboard.md` and therefore rewrite the same hashed bundle assets. Whoever lands second needs a rebase plus `npm --prefix web run build` or the staleness check fails. This one is docs-only and cheapest to rebase, so it is the natural one to land first.

docs/dashboard.md
njbrake: _Note: drafted by Claude via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

You are right, and the Docker case is the one that would bite. Fixed in f87cd2ca: the text now separates surviving a tab close from surviving a restart, and says the latter holds only as long as the gateway's database does, calling out that a container on the default `sqlite:///./otari.db` with no mounted volume starts every run signed out. Confirmed the default URL in `core/config.py` and that compose mounts no volume for it.

docs/dashboard.md
njbrake: _Note: drafted by Claude via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

This is the sharper half of the review: the correction removed a security property without replacing it. Fixed in f87cd2ca with a new security note. Verified the premise rather than assuming it, `is_valid_dashboard_session` compares `expires_at` and never extends it, so there is no idle timeout and no sliding window; a session runs its full TTL. The note says to sign out on a shared machine, offers shortening `dashboard_session_ttl_hours`, and points at master-key rotation as the way to revoke a session you can no longer reach.

PR #452, opened 2026-07-30
chore(deps): refresh bundled genai-prices lock to price MiniMax-M3

octo-patch:
Reason: A fresh install locks genai-prices 0.0.66, which prices MiniMax-M2.7 but not MiniMax-M3, so MiniMax-M3 requests are rejected with HTTP 402 under the default require_pricing.

## What

Otari resolves default model pricing from the bundled `genai-prices` catalog. With `require_pricing` enabled by default, any model outside that catalog is rejected with HTTP 402. The bundled lock pins `genai-prices==0.0.66`, whose snapshot does not yet include `minimax-m3`, so a fresh installation cannot price MiniMax-M3 and returns 402 until an operator applies a newer snapshot through the runtime refresh workflow.

`minimax-m3` first appears in `genai-prices` 0.0.67 and is present in the current release 0.0.73. This change refreshes the bundled lock so a fresh installation prices MiniMax-M3 out of the box, without relying on a manual refresh.

## Changes

- `pyproject.toml`: raise the `genai-prices` floor from `>=0.0.66` to `>=0.0.67`, the first release whose catalog includes `minimax-m3`.
- `uv.lock`: relock `genai-prices` from `0.0.66` to `0.0.73` (latest release) via `uv lock --upgrade-package genai-prices`. No other packages changed.

## Verification

- `uv lock --upgrade-package genai-prices` resolved 223 packages successfully; only `genai-prices` was updated (0.0.66 -> 0.0.73).
- Confirmed the packaged `genai_prices/data.py` in 0.0.73 contains `minimax-m3` (`MiniMax M3`), so the fresh-install catalog can now price it.

## Summary

- Updated the bundled `genai-prices` dependency and lockfile to include MiniMax-M3 pricing.
- Fresh installations can now price MiniMax-M3 requests without manual catalog updates, avoiding default HTTP 402 rejections.


njbrake:
Hi @octo-patch this might help: in the dashboard and the settings tab there is an "update prices" button which refreshes the price catalogue without the need to update the package deps and rebuilding: does that help?

njbrake:
@octo-patch once you refresh the prices, the prices are stored in the database of otari so you shouldn't need to keep updating the dependency and rebuilding, you can just refresh to update the prices

PR #453, opened 2026-07-30
fix(dashboard): address timeline review follow-ups from #445

njbrake:
## Description

Picks up the review follow-ups on #445 (the histogram-as-time-range-selector work) from @khaledosman and Copilot that landed unaddressed at merge. All are dashboard-only; no gateway API contract change.

### Khaled's comments

- **Activity "All" reintroduced the invisible-default bug.** With `range=all` the histogram sent no `start_date`, so the summary endpoint applied its 30-day default: the bars silently showed a rolling month while the caption read "All time", and you could not drag older. A new `resolveExtentWindow` now gives the histogram an explicit year-long start (deterministic, and the axis shows exactly what it covers), while the list stays genuinely all-time. This is Khaled's first suggested fix (explicit start for the histogram, list window captioned separately); "All" keeps its unbounded list.
- **Empty state counted the default window as a filter.** `anyFilter` used `hasWindow`, so a brand-new gateway on the 24h default read "No requests match these filters." It now mirrors UsagePage's `timeFiltered` (the default preset is not itself a filter), so an empty first load reads "No requests recorded yet."
- **`12mo` was exactly the server cap.** `366 * DAY_S` always exceeds `_MAX_SUMMARY_SPAN` by the client/server round-trip delta and gets pulled forward, so the preset never spanned a year. It now requests `365 * DAY_S`, which leaves headroom and never trips the clamp.
- **Thumb resync missed a rolling series.** The resync keyed on `[windowStart, windowEnd, n]`; it now also keys on the extent's first and last bucket, so a series that rolls forward with identical bounds and length no longer strands the thumbs.
- **Panning had no keyboard path.** The pan strip is now a single-value slider (the window's left edge) operable with arrows, Page Up/Down, Home, and End, so a keyboard user can reach a mid-extent window without stepping both edges in turn.

### Copilot's comments

- **Overview refresh state.** `refresh()` triggers `previous.refetch()`, but `isRefreshing` omitted `previous.isFetching`; it is now included, so the button cannot re-enable (and the spinner cannot stop) while that refetch is in flight.
- **Pan strip pointer-cancel.** `onPointerCancel` now releases pointer capture (when held), mirroring `onPointerUp`.
- **Filter-chip focus ring.** The remove control gets a `focus-visible` ring (also raised by CodeRabbit).
- **Locale-sensitive tests.** The timeline caption tests asserted on `"Jul"`, but the UI formats with `toLocaleDateString(undefined, ...)`. They now assert on day numbers and the UTC marker.

### Deliberately not changed

- **`useDisplayClock` `setInterval` (Copilot).** Kept. The frontend-standards rule targets manual data *polling* (use `refetchInterval`); this is a display-only clock that fetches nothing, already pauses while the tab is hidden, and only exists to keep the "updated Ns ago" label truthful. Removing it would freeze the label between refetches, making it misleading. Happy to revisit if reviewers disagree.

## PR Type

- [x] Bug Fix

## Relevant issues

Follow-up to #445 (#440).

## Checklist

- [x] I ran the Definition of Done checks locally.

Notes: dashboard-only, so no OpenAPI change. `npm --prefix web run typecheck` and `npm --prefix web test` (340 tests) pass, and the committed bundle was rebuilt.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

Claude Code (Opus 4.8), directed by @njbrake

## Summary

- Improved dashboard activity timelines with clearer date-range behavior, synchronized rolling-series selections, keyboard navigation, pointer cancellation, and more accessible controls.
- Corrected empty states and one-year/custom range handling while keeping activity lists unbounded.
- Updated overview refresh status to include previous-period data and improved filter-chip focus styling.
- Refreshed locale-independent timeline and activity tests, with 340 web tests passing.
- Rebuilt dashboard assets and updated hashed bundle references, including the Activity, Overview, and Documentation pages.


web/src/pages/ActivityPage.tsx
```diff
@@ -360,8 +375,12 @@ export function ActivityPage() {
   const rows = usage.data ?? [];
   const totalIsExact = count.isSuccess && !count.isPlaceholderData;
   const total = totalIsExact ? (count.data?.total ?? 0) : null;
+  // The default preset is not itself a filter (mirroring UsagePage): only an
+  // explicit sub-window or a non-default preset counts as a time filter, so a
+  // brand-new gateway on its 24h default reads "never used", not "filtered empty".
+  const timeFiltered = Boolean(startParam || endParam) || range !== ACTIVITY_DEFAULT_KEY;
```
khaledosman: `range !== ACTIVITY_DEFAULT_KEY` makes the **broadest** preset count as a filter: with `range=all` there is no window at all, yet an empty result now reads "No requests match these filters." The assertion removed from `ActivityPage.test.tsx` was pinning the correct behaviour for exactly that case.

UsagePage is not a valid mirror here, because it has no unbounded preset. Excluding the unbounded case keeps both ends honest:

```ts
const preset = findPreset(ACTIVITY_PRESETS, range);
const timeFiltered =
  Boolean(startParam || endParam) || (range !== ACTIVITY_DEFAULT_KEY && preset?.seconds != null);
```

web/src/pages/ActivityPage.tsx
```diff
@@ -129,6 +130,21 @@ function resolveWindow(range: string, start: string, end: string): { start?: str
...
+// The histogram extent (what the bars span), which is *not* always the list
+// window. For bounded presets it matches `resolveWindow`. For the unbounded "All"
+// the list genuinely omits its start, but the summary endpoint would then apply a
+// hidden 30-day default: the bars would silently show a rolling month while the
+// caption reads "All time". So the histogram gets an explicit year-long start,
+// giving a deterministic, draggable span (the axis shows exactly what it covers)
+// while the list stays all-time.
+function resolveExtentWindow(range: string): { start?: string; end?: string } {
+  const win = resolveWindow(range, "", "");
+  if (win.start || range === CUSTOM_KEY) return win;
```
khaledosman: `range=custom` still returns `{}` here, which reintroduces the bug this function exists to fix. `contextFilters` only substitutes `win`'s bounds when `winOutsideExtent` is true, and that requires **both** `win.start` and `extentWin.start`; with `extentWin.start` undefined the summary request carries no `start_date` and the server applies its hidden 30-day default. Result: bars show a rolling month while the caption shows the custom window.

Only reachable by hand-editing `?range=custom` today (nothing calls `pickPreset` with it), so either give the custom case `win`'s bounds or drop the special case rather than leave a dormant instance of the same bug.

web/src/components/ActivityTimeline.tsx
```diff
@@ -256,6 +259,40 @@ export function ActivityTimeline({
...
+    const [slo, shi] = selRef.current;
+    const span = shi - slo;
+    const plo = Math.max(0, Math.min(n - span, Math.round(slo) + deltaBuckets));
+    const next: [number, number] = [plo, plo + span];
+    setSel(next);
+    commit(next);
+  };
+
+  const onPanKeyDown = (event: React.KeyboardEvent) => {
+    const span = Math.max(1, Math.round(hi - lo));
```
khaledosman: `span` is read from the render-state `hi - lo` while `panBy` (line 266) reads `selRef.current`. That is the split the ref was introduced to avoid: hold PageUp/PageDown and the repeat fires before the re-render, so the pan uses a stale span.

Deriving the span inside `panBy` from `selRef.current` and passing a multiplier (`panBy(1)` / `panBy(-1)` for Page keys) keeps both halves on one source of truth.

web/src/components/ActivityTimeline.tsx
```diff
@@ -363,11 +400,21 @@ export function ActivityTimeline({
...
             <div
-              aria-hidden
-              className="absolute bottom-0 z-[1] cursor-grab touch-none rounded bg-[var(--otari-brand)]/10 hover:bg-[var(--otari-brand)]/20 active:cursor-grabbing"
+              role="slider"
+              aria-label="Pan the selected window"
+              aria-valuemin={0}
+              aria-valuemax={n}
+              aria-valuenow={Math.round(Math.min(lo, hi))}
+              aria-valuetext={`Window starting at ${formatTick(starts[Math.min(Math.round(Math.min(lo, hi)), n - 1)] ?? starts[0], bucket)}`}
+              tabIndex={zoomed ? 0 : -1}
```
khaledosman: `tabIndex={zoomed ? 0 : -1}` drops an announced `role="slider"` out of the tab order with no stated reason, so a screen-reader user meets a control that is sometimes unreachable. Adding `aria-disabled={!zoomed}` alongside keeps it discoverable and explains why it cannot be operated at full extent.

khaledosman (approved):
Dashboard-only follow-ups to #445. Nothing blocking; four inline points, the first two are behavioural.

The `starts[0]` / `starts[n - 1]` resync deps are the right fix and the array length stays constant, so the eslint suppression stays honest. Keeping `useDisplayClock` is correct: the frontend-standards rule is about data polling, and a display clock that fetches nothing is not that.

---

PR #454, opened 2026-07-30
fix(dashboard): distinguish a missing /models endpoint from an unreachable provider

### LINKED ISSUE #447: Dashboard: if a provider has an unimplemented '/models' endpoint it may still be usable
otari.ai as of yesterday didn't have /models endpoint available. This meant that even if you added a correct otari.ai api key, the dashboard said it was unreachable, because it relies on the /models endpoint for that check. If the /model endpoint is unavailable you should still be able to use the endpoint, though there should be a warning in the dashboard that model discovery isn't working.

Feel free to ask any clarification questions.

njbrake:
## Description

The dashboard derives provider status from model discovery: a provider is "Reachable" when its credentials can list models. A backend that never implemented `/v1/models` therefore read as "Unreachable" even with a perfectly good key, which is what happened with otari.ai. That is misleading twice over: it accuses the credential, and it hides that the provider can still serve requests.

Discovery now classifies its failures. A 404, 405, or 501 from the listing call, and a provider whose any-llm implementation cannot list models at all, are reported as `discovery_unsupported` rather than as plain failures. The status is read off the exception (OpenAI-style `status_code`, httpx-style `response.status_code`, or an any-llm wrapper's `original_exception`), so 401/403, connection errors, timeouts, and 5xx all stay "unreachable". The flag rides through the provider health, discoverable models, and connection test responses, and `GET /v1/providers/health` counts those providers under a new `degraded` total.

There are now three states rather than two:

| Pill | Condition | Meaning |
|---|---|---|
| Reachable (green) | `ok: true` | listing succeeded; credentials confirmed working |
| No model discovery (amber) | `ok: false` + `discovery_unsupported: true` | the backend serves no model listing; the key is unverified, not bad |
| Unreachable (red) | `ok: false` + `discovery_unsupported: false` | auth rejected, connection failed, timeout, 5xx |

Amber is deliberately "unknown", not "healthy": the gateway never talked to that provider successfully, so `ok` stays `false` and it is not counted in `healthy`. It gets its own `degraded` count instead, and the summary reads "1 of 2 providers reachable, 1 without model discovery" rather than folding it into either side.

UI surfaces, all driven by that one flag:

- **Providers page**: an amber "No model discovery" status pill (the provider error plus "Requests to it may still work" stays in the tooltip); the summary banner appends "N without model discovery" and keeps its dot amber rather than red when discovery gaps are the only issue; both the per-row test result and the add/edit form test result say the key could not be verified instead of that the connection failed.
- **Models page**: the discovery warning banner splits, so a provider with no listing endpoint is pointed at its `models:` config key instead of at the operator's credentials.
- **Overview page**: the system status strip lists "N providers without model discovery" as its own linked problem, and no longer grades the gateway as an outage when discovery gaps are all that is wrong.

`docs/dashboard.md` documents the distinction and the committed dashboard bundle is rebuilt.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes #447

## Checklist

Test coverage added: unit cases for the status classification (including 401, 500, and no-status staying "unreachable", and the httpx / any-llm exception shapes), the health-service mapping, integration tests on `GET /v1/providers/health` and `GET /v1/models/discoverable`, a degraded case on the stored-provider test endpoint, and frontend tests for the pill, the summary, the test outcome, the models banner, and the overview grading.

Local run: ruff, mypy, `openapi-check`, `npm run typecheck`, 349 web tests, and 1832 python tests. All pass, no failures.

Verified against the live service rather than only against stubs: `api.otari.ai/v1/models` returns 404 today, and a gateway configured with it reports `discovery_unsupported: true` / `degraded: 1` while serving completions normally, which is exactly the state this PR adds and the one `main` currently mislabels as "Unreachable". The classification was also confirmed under `ANY_LLM_UNIFIED_EXCEPTIONS`, where any-llm replaces the SDK error with `ModelNotFoundError` carrying no `status_code`; the `original_exception` hop is what keeps it working, and a test now pins that.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

Implemented by Claude through back-and-forth with @njbrake: the diagnosis, the three-state design, and the decision to keep `ok: false` for the degraded case are his; the code and prose are Claude's.


src/gateway/services/model_discovery_service.py
```diff
@@ -326,6 +336,37 @@ def _short_error(exc: BaseException, provider: str | None = None) -> str:
...
+            continue
+        code = getattr(candidate, "status_code", None)
+        if code is None:
+            code = getattr(getattr(candidate, "response", None), "status_code", None)
+        if isinstance(code, int):
+            return code
+    return None
+
+
+def _is_missing_models_endpoint(exc: BaseException) -> bool:
```
khaledosman, line 382: A 404 here is not only "this backend never implemented `/v1/models`": a typo'd, missing, or duplicated `/v1` in `api_base` produces the same status. In that case the operator is now told the credential "may still work" for what is really a misconfiguration they have to fix.

The classification is still the right default (the not-implemented case is far more common and the old behaviour actively accused the key). The fix is not to narrow it, but to stop hiding the evidence in the UI: see the notes on `ProvidersPage.tsx`.

src/gateway/services/model_discovery_service.py
```diff
@@ -326,6 +336,37 @@ def _short_error(exc: BaseException, provider: str | None = None) -> str:
...
 
 
+# Statuses that mean the backend serves no model-listing endpoint at this URL,
+# as opposed to refusing the credentials (401/403) or not answering at all. An
+# OpenAI-compatible deployment that has not implemented /v1/models answers 404
+# here while still serving completions perfectly well.
+_MISSING_ENDPOINT_STATUSES = frozenset({404, 405, 501})
+
+
+def _status_code(exc: BaseException) -> int | None:
```
khaledosman, line 349: This unwraps exactly one wrapper level (`exc`, then `exc.original_exception`). A doubly-wrapped SDK error, or one chained through `raise ... from`, falls through to `None` and is silently reported as "unreachable", which is the failure mode this function exists to prevent.

Worth walking `__cause__` as well, in a small loop with a depth cap:

```python
seen: set[int] = set()
candidate: BaseException | None = exc
while candidate is not None and len(seen) < 5:
    seen.add(id(candidate))
    code = getattr(candidate, "status_code", None)
    if code is None:
        code = getattr(getattr(candidate, "response", None), "status_code", None)
    if isinstance(code, int):
        return code
    nxt = getattr(candidate, "original_exception", None) or candidate.__cause__
    candidate = None if nxt is None or id(nxt) in seen else nxt
return None
```

web/src/pages/ProvidersPage.tsx
```diff
@@ -568,6 +575,15 @@ function TestOutcome({ state }: { state: TestState | undefined }) {
       </span>
     );
   }
+  // A backend with no model-listing endpoint cannot be verified this way, but the
+  // key is not therefore wrong: say so instead of reporting a failed connection.
+  if (state.discovery_unsupported) {
```
khaledosman, line 589: This branch discards `state.error`, and `ProvidersPage.test.tsx` asserts the 404 text is *absent*. Combined with the `api_base` ambiguity noted on `model_discovery_service.py`, the operator loses the only signal that distinguishes "this backend has no listing endpoint" from "my base URL is wrong".

`HealthPill` (line 608) gets this right by keeping the provider error in the tooltip. Same treatment here: keep the sentence, add the raw error (a `title` attribute, or a second muted line).

web/src/pages/ProvidersPage.tsx
```diff
@@ -182,6 +182,13 @@ function ConnectionTest({ getPayload }: { getPayload: () => CreateStoredProvider
             <span className="text-xs font-medium text-green-700">
               Connected. {test.data.model_count} model{test.data.model_count === 1 ? "" : "s"} available.
             </span>
+          ) : test.data.discovery_unsupported ? (
```
khaledosman, line 185: Same as `TestOutcome`: the amber branch replaces `test.data.error` rather than adding to it. This is the add/edit form, where a wrong `api_base` is the *most* likely cause of a 404, since the operator just typed it. Surfacing the error alongside the reassurance would let them spot the typo instead of saving a credential against a bad URL.

web/src/pages/ModelsPage.tsx
```diff
@@ -977,14 +977,34 @@ function ModelTable({
...
+      {unreachable.length > 0 ? (
+        <span className="block">
+          Could not list {unreachable.map((provider) => provider.provider).join(", ")}. Check that provider's
+          credentials in config.yml; its models are missing from the list below.
+        </span>
+      ) : null}
+      {noDiscovery.length > 0 ? (
+        <span className="block">
+          {noDiscovery.map((provider) => provider.provider).join(", ")} does not offer model discovery, so its models
+          are missing from the list below. The provider may still serve requests; declare the model ids it serves under
```
khaledosman: Plural agreement: with two providers this reads "openai, anthropic does not offer model discovery". The `unreachable` branch just above has the mirror problem ("Check that provider's credentials" for a list of providers), which is pre-existing but cheap to fix while the banner is being split.

khaledosman (approved):
The three-state design is right, and keeping `ok: false` for the degraded case is the correct call: it keeps `healthy` meaning "the gateway actually talked to it" rather than "we assume it works." `providerHealthStatus`, the `degraded` accounting, and the pill/summary/overview surfaces stay consistent with each other.

Nothing blocking. The substantive point is that a 404 does not only mean "not implemented", and the two UI branches that report it are the ones that throw the evidence away.

---

PR #455, opened 2026-07-30
feat(aliases): scope a model alias to a single user

### LINKED ISSUE #448: Dashboard: alias page is not tied to a user
The alias page is a bit light. The aliases are applied globally, but I would think we would want them to be able to be attached to specific users or api keys?

njbrake:

## Description

The aliases page was thin because that was all an alias could be: one global map, so a display name meant the same thing to every caller. That makes a stable name like `fast` unusable for the thing operators actually want it for, which is pointing different people at different models without anyone changing the `model` they send.

Stored aliases now carry an optional `user_id`. `NULL` is global (what every existing row is, and what `config.yml` aliases always are); a `user_id` scopes the alias to that user alone.

**Precedence is most-specific-first: user-scoped > config.yml > global stored.** That ordering is load-bearing. It is why a user-scoped alias *may* override a configured name (the override genuinely works), while a global stored one still may not (config would win and the stored row would silently never resolve, which the API keeps refusing with a 400).

Consequences that follow, all covered by tests:

* **Name plus scope is the identity.** Creating `fast` for `alice` leaves the global `fast` alone, and deleting either leaves the other in place. Deletes take the scope as a query parameter (`DELETE /v1/aliases/fast?user_id=alice`), or omit it for the global one.
* **`GET /v1/models` is scoped the same way**, so a caller sees the global and configured aliases plus their own, never another user's. `GET /v1/aliases` stays the master key management view and lists every scope at once.
* **Target hiding follows the scope.** A global alias hides its target from everyone's listing; a user-scoped one hides it only from that user's, so another caller may still see the real model subject to their own model access. Operators using aliases to curate one catalogue for everybody should keep those aliases global. This is documented.
* **`user_id` must name a live user** (unknown or soft-deleted is a 404), matching how every other user-scoped route treats users.
* **The two writes that must never store an alias name as a model key** (pricing rows, model allow-list entries) check every scope at once via a new scope-blind `all_alias_names`.

Resolution has to stay synchronous, so the process-wide alias cache grows a per-user layer rather than taking a database read per request, and `resolve_provider_selector` now takes the billed user. Every request-path call site passes it. The vision describe model deliberately does not: it is operator config rather than caller input, so it is global by definition.

### Schema

The natural key is now `(name, user_id)`, which cannot be a primary key because `user_id` is nullable, so the PK moves to a surrogate `id` back-filled from the old name. Uniqueness needs two constraints, because SQLite and PostgreSQL both treat NULLs as distinct in a unique index: a composite constraint for `(name, user)`, plus a partial unique index for one global row per name.

Docker was unavailable locally, so the migration was verified by hand against a real PostgreSQL 18 instance and against SQLite: upgrade carrying legacy data, both unique constraints rejecting the right duplicates, the FK cascade, downgrade, re-upgrade, and `create_all` parity with the migrated schema.

### Scoped to users, not API keys

The issue asked "users or api keys?". This scopes to the user; a key inherits its user's aliases. The key level can layer on later exactly the way `allowed_models` layers a key over its user, if that turns out to be wanted.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes #448

## Checklist

Local results: ruff, mypy (strict), `openapi --check` and the dashboard typecheck are clean; 344 web tests pass; 1829 Python tests pass. Two integration tests fail in this sandbox (`test_error_detail_leakage`, `test_streaming_error_event`); both make real outbound calls to OpenAI and expect a 502 while the sandbox gets a genuine upstream 404. They were reproduced failing identically at the base commit `ffad6b5` with none of these changes, so they are pre-existing and unrelated.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

Written by Claude working from the issue, then reviewed by @njbrake. Worth flagging one thing the automated checks could not see: an editing slip left three raw NUL bytes in `AliasesPage.tsx`, which every check passed over because JavaScript treats `\0` as an ordinary string character. The only signal was `git diff --stat` reporting the file as `Bin`. It is fixed (the separator is now an explicit `\u0000` escape) and the file is normal text again. A related claim made during review, that the row key parsing also mis-targeted deletes for user ids containing a space, turned out to be wrong on closer inspection: the split and join used the same separator, so the round trip was correct. The parsing was still replaced with a row lookup, as a readability change rather than a fix.

**NOTE:**

## Summary

- Added user-scoped model aliases while preserving global aliases.
- Added user-aware alias creation, editing, deletion, listing, validation, caching, model visibility, pricing, and provider selection.
- Added dashboard and API support for selecting alias owners and distinguishing scopes.
- Added database migration, documentation, API updates, and comprehensive tests.
- API keys now inherit aliases from their associated users, improving per-user customization without affecting global defaults.


alembic/versions/c3f7a9d1e5b8_scope_model_aliases_to_users.py
```diff
@@ -0,0 +1,85 @@
...
+    op.add_column("model_aliases", sa.Column("id", sa.String(), nullable=True))
+    op.add_column("model_aliases", sa.Column("user_id", sa.String(), nullable=True))
+    # Existing rows are global and their names are unique (name was the primary
+    # key), so the name doubles as a back-filled surrogate id.
+    conn.execute(sa.text("UPDATE model_aliases SET id = name WHERE id IS NULL"))
+
+    with op.batch_alter_table("model_aliases") as batch_op:
+        batch_op.alter_column("id", existing_type=sa.String(), nullable=False)
+        if not is_sqlite:
+            batch_op.drop_constraint("model_aliases_pkey", type_="primary")
```
khaledosman: The SQLite path works, but only because SQLAlchemy silently reassigns the primary key during the batch recreate. Running `upgrade` on SQLite emits this twice:

```
SAWarning: Table '_alembic_tmp_model_aliases' specifies columns 'name' as primary_key=True,
not matching locally specified columns 'id'; setting the current primary key columns to 'id'.
This warning may become an exception in a future release
```

So the PK swap depends on behaviour SQLAlchemy has explicitly reserved the right to turn into an error. Dropping the constraint on the SQLite branch too (or `recreate="always"` with an explicit `copy_from`) makes the intent explicit rather than implied.

Worth weighing because there is no migration test harness in the repo at all, so this path has no regression net: the next SQLAlchemy bump would surface it as a failed upgrade on a user's database rather than in CI.

src/gateway/api/routes/aliases.py
```diff
@@ -48,49 +63,80 @@ def from_model(cls, alias: ModelAlias) -> "AliasResponse":
...
+                "Rename it, scope it to a user, or edit config.yml."
             ),
         )
-    # The chaining check has to see both sides: an alias pointing at a stored
-    # alias is just as broken as one pointing at a configured alias.
-    alias_names = set(stored) | set(config.aliases) | {name}
+    # The chaining check has to see every layer that could resolve for this
+    # alias's audience: an alias pointing at a stored alias is just as broken as
+    # one pointing at a configured alias.
+    alias_names = set(cached_aliases()) | set(config.aliases) | {name}
```
khaledosman: This hand-built union omits every other user's scoped names, so `_validate` is scope-blind for pricing rows and allow-list entries (via `all_alias_names`) but not here.

The practical impact is nil, which is worth stating plainly: resolution is single-pass, and `validate_alias` checks the target's *prefix*, so a target naming another user's alias would be rejected as an unknown provider anyway. So this is a consistency point rather than a bug. `all_alias_names(config) | {name}` covers all three layers in one line and removes the `if user_id is not None` branch below it.

src/gateway/models/entities.py
```diff
@@ -166,12 +166,37 @@ class ModelAlias(Base):
...
+            "name",
+            unique=True,
+            sqlite_where=text("user_id IS NULL"),
+            postgresql_where=text("user_id IS NULL"),
+        ),
+    )
 
-    name: Mapped[str] = mapped_column(primary_key=True)
+    id: Mapped[str] = mapped_column(primary_key=True, default=lambda: str(uuid.uuid4()))
+    name: Mapped[str] = mapped_column(index=True)
```
khaledosman: `index=True` here is redundant on PostgreSQL: `uq_model_aliases_name_user` already indexes `(name, user_id)` with `name` leading, and `uq_model_aliases_global_name` indexes `name` again (partially). That is three indexes over the same leading column on a table that is read through a 30-second-TTL process cache, so the reads it would serve mostly do not happen while every write pays for all three.

If you keep it for `create_all` / migration parity, that is a fine reason, just worth being deliberate about.

docs/models.md
```diff
@@ -218,6 +218,61 @@ Like named instances, aliases are a standalone-mode feature. In hybrid mode mode
...
+  may not (config would win, and the stored one would silently never be used, so
+  the API refuses it with a 400).
+- A name plus a scope is the identity. Creating `fast` for `alice` leaves the
+  global `fast` alone, and deleting either leaves the other in place. Deletes take
+  the scope as a query parameter: `DELETE /v1/aliases/fast?user_id=alice`, or omit
+  it for the global one.
+- `GET /v1/models` is scoped the same way, so a caller sees the global and
+  configured aliases plus their own, never another user's. `GET /v1/aliases` is
+  the master-key management view and lists every scope at once.
+- Target-hiding follows the scope. A global alias hides its target from
```
khaledosman, line 263: This bullet covers global-vs-scoped hiding, but not the one case where creating an alias *reveals* a model: when a user-scoped alias overrides a **config** alias name, the merged map for that user no longer contains the config target, so `list_models` stops withholding it and the config alias's real `provider:model` reappears in that user's catalogue (subject to their allow-list).

It is operator-initiated and master-key gated, so not a leak as such, but it is the counterintuitive direction and belongs next to the rule it inverts.

src/gateway/api/routes/models.py
```diff
@@ -341,7 +345,7 @@ async def list_models(
     pricing_lookup = _normalized_pricing_lookup(config, pricing_map)
     # Read once: phase 2 withholds these targets and phase 3 lists the names, and
     # the two must agree even if a write lands between them.
-    aliases = effective_aliases(config)
+    aliases = effective_aliases(config, caller_user_id)
```
khaledosman, line 355: The listing scopes on the API key's user while dispatch scopes on the billed user. For a non-master key those are always the same (`resolve_user_id` binds spend to the key's own user), so there is no leak.

The asymmetry shows up for a master-key caller sending `"user": "alice"`: dispatch resolves alice's scoped aliases, but `GET /v1/models` shows none of them, since `auth[0]` is `None`. That is the exact path an operator uses to test a scoped alias after creating it, so it is worth a line in the docs even if the behaviour stays as is.

khaledosman (approved):
I ran the migration against SQLite end to end rather than take the manual verification on trust, and it holds up: upgrade carries a legacy row with `id` back-filled from `name`, the composite constraint and the partial global index each reject the right duplicate (`UNIQUE constraint failed: model_aliases.name` for a second global row, `...name, model_aliases.user_id` for a duplicate scoped row), the FK rejects an unknown user, and downgrade drops scoped rows and restores the `name` primary key cleanly. The two-constraint design and the reasoning in the docstring are both correct.

The security-relevant part is sound: `resolve_user_id` binds a non-master key to its own user, so the body `user` field cannot be used to resolve someone else's aliases, and the cross-user test asserts `dispatch.assert_not_awaited()` rather than just a status code. Leaving `vision.py` and `_bill_vision_side_call` unscoped is right, and the docstring says why.

Nothing blocking. One forward-compatibility note on the migration, and four smaller points.

---

PR #466, opened 2026-07-31
fix(pricing): follow genai-prices to the live v2 price feed

HareeshBahuleyan:
## Description

genai-prices fetches price data at runtime from a URL baked into the installed version. Every release before `0.1.0` (this repo was on `0.0.66`) points at `prices/data.json`, which upstream froze in favour of `prices/new_data/v2/data.json`:

> **v1 is frozen.** `prices/data.json` and `prices/data_slim.json` are compatibility snapshots for clients released before v2. No build step writes them, and they receive no provider, model or price updates. `prices/new_data/v2/` is the live feed.

Last change to the v1 file: `c885e328`, 2026-07-25. So an operator running a pricing refresh (`prepare_price_refresh` -> `confirm_price_refresh`) got a successful fetch, a diff preview, and data that can no longer move. `gpt-5.6-luna` was cut upstream to 0.20/1.20 per Mtok (pydantic/genai-prices#531) while the frozen feed still served 1.00/6.00, which feeds budget reservations and the `require_pricing` gate.

Two things made this more than a version bump:

- `0.0.66` cannot parse the v2 feed at all: 48 validation errors on the wider extractor destinations (`cache_write_5m_tokens` and friends). Pointing the old pin at v2 is not an option, so the bump is required.
- `0.1.0` made raw-payload parsing private (`providers_schema` -> `_providers_schema` / `_providers_from_raw`). This service persists the raw payload in `PricingSnapshot` and re-parses it on confirm and on load, and `UpdatePrices.fetch` couples fetching to parsing, so it cannot be used for a stored payload. The three parse sites now go through one `_parse_snapshot` helper that calls the same code path the library's own updater calls. Worth raising upstream so this can move back onto a public API.

Also added a unit test asserting the resolved feed URL is the v2 path, so a future pin regression fails instead of quietly re-freezing prices. Relevant because `.github/dependabot.yml` here covers `github-actions` only: nothing bumps Python deps, which is how the pin drifted five releases behind. Adding the `uv` ecosystem is a maintainer call, so it is not in this PR.

The same root cause on the platform side is fixed in mozilla-ai/otari-ai#1510.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

None filed here; context is in mozilla-ai/otari-ai#1509.

## Checklist

- [x] Documentation was updated where necessary. (No user-facing behaviour change: the refresh flow is unchanged, it now reads the maintained feed.)
- [x] If the API contract changed, I regenerated the OpenAPI spec. (No API change.)

Local results: `ruff` clean, `mypy` clean over 272 files, `pytest tests/unit` 1035 passed, `pytest tests/integration` 821 passed / 9 skipped.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

The freeze was found while investigating why a known upstream price cut was not reaching the platform; the same investigation produced mozilla-ai/otari-ai#1509. The v2-parse failure under `0.0.66` and the private-API breakage were both reproduced locally against clean installs before the fix was written.

## Summary

- Updated `genai-prices` to `0.1.0`.
- Switched pricing refreshes to the maintained v2 feed.
- Centralized snapshot parsing for fetched and persisted pricing data.
- Normalized errors from malformed pricing snapshots.
- Added tests for the v2 feed URL and snapshot parsing.
- Confirmed linting, type checking, and unit/integration tests pass.

## Technical notes

- Replaced the removed private parsing dependency with a local `_parse_snapshot` helper.
- Updated tests to use the public `DataSnapshot` parser.


njbrake (approved):

Approving. Nice find, @HareeshBahuleyan. A frozen upstream feed that still answers 200 and still renders a plausible diff preview is close to the worst shape a pricing bug can take, and catching it from the other end (a known price cut not arriving) is good instinct.

I reproduced the load-bearing claims against clean installs of both versions and against the live upstream repo rather than taking the description on trust. All of them hold:

| Claim | Result |
| --- | --- |
| `0.0.66` fetches `prices/data.json` | Confirmed |
| `0.1.0` fetches `prices/new_data/v2/data.json` | Confirmed |
| v1 is frozen | Confirmed, the quote is verbatim from `prices/README.md` and the root README repeats it |
| v1's last change is `c885e328`, 2026-07-25 | Confirmed; v2 has commits on 07-30 and 07-31 since |
| `0.0.66` cannot parse v2, "48 validation errors" | Confirmed, literally 48, first one is `dest='cache_write_5m_tokens'` |
| `providers_schema` went private in `0.1.0` | Confirmed, `data.providers_schema` is gone and `types._providers_from_raw` is the replacement |
| `UpdatePrices.fetch` couples fetching to parsing | Confirmed by reading the source; it returns only a `DataSnapshot` and never surfaces the raw payload |
| The helper uses the library's own code path | Confirmed, `fetch()` itself does `from .types import _providers_from_raw` |
| gpt-5.6-luna cut to 0.20/1.20 while v1 serves 1.00/6.00 | Confirmed |

Quantifying the impact, with each feed applied as a custom snapshot and priced through `calc_price`:

```
v1 (frozen):  1M in + 1M out gpt-5.6-luna => $11.00
v2 (live):    1M in + 1M out gpt-5.6-luna =>  $2.20
```

So a refresh run today does not merely fail to improve prices, it re-affirms a 5x overcharge on budget reservations for that model.

One refinement, not a correction: the v2 dated `prices` list shape parses fine under `0.0.66`. The incompatibility is purely the widened extractor destinations, which is what the description says.

### Non-regression checks

* Full unit suite: 1036 passed, 1 skipped.
* Integration suite on Postgres, run serially: 819 passed, 9 skipped, 2 failed. Both failures (`test_provider_error_does_not_leak_details`, `test_streaming_creation_error_returns_http_error`) reproduce on this branch's parent commit, so they are pre-existing and specific to a sandbox with no outbound provider network.
* `ruff` clean, `mypy` clean over 272 files.
* Upgrade path for existing deployments: a v1-format payload already sitting in `pricing_snapshot` still parses under `0.1.0`, resolving 34 providers with correct prices. No startup warning and no silent fallback to bundled data. That was the regression I most expected to find and it is not there.
* Every other `genai_prices` import in the tree (`Usage`, `calc_price`, `PriceCalculation`, `TieredPrices`, `Tier`, `get_snapshot`) is still public in `0.1.0`.

### On the Copilot comment

Already addressed by `0ea2336a`. Worth recording that I could not construct an input where `_providers_from_raw` raises something that is not a `ValueError`: pydantic's `ValidationError` and `json.JSONDecodeError` are both `ValueError` subclasses, and dict, string, null, int, and non-list payloads all come back as `ValidationError`. So the wrapper is insurance against a private API rather than a fix for a reachable crash. Cheap, and worth keeping for that reason.

### Two follow-ups, neither blocking

1. **A parser-level regression test.** Both fixtures in the test file are v1-shaped, so nothing exercises the payload shape this PR exists to support. `test_refresh_targets_the_maintained_feed` pins the URL but not the parser, so a pin that can reach v2 without being able to read it would still go green. A minimal fixture carrying a `cache_write_5m_tokens` mapping closes that: I checked it passes on `0.1.0` and fails on `0.0.66`. Happy to push it to this branch, or it can land separately.

2. **The unbounded pin over a private API.** `genai-prices>=0.1.0` plus a module-level `from genai_prices.types import _providers_from_raw` means a future release that renames it takes down gateway startup, since `main.py` imports this service. Loud rather than silent, which is the better failure mode, but an installed wheel is where it would land. A `<0.2.0` cap would contain it at the cost of being the only capped runtime dependency in `pyproject.toml`. Raising it upstream for a public API, as the description suggests, is the better long-term fix.

Adding the `uv` ecosystem to `.github/dependabot.yml` is the actual root-cause fix for a five-release drift. Correctly left out of here, and worth its own issue.

PR #467, opened 2026-07-31
fix(dashboard): clear a provider's connection-test result when it changes

### LINKED ISSUE #464: Dashboard: a provider row keeps showing a stale connection-test result after the provider is edited
The Providers page keeps per-row connection-test results in component state (`web/src/pages/ProvidersPage.tsx`, the `tests` record). Entries are written when a test starts and when it settles, but nothing ever clears them: not editing the provider, not saving new credentials, not deleting and re-adding, and not a provider-health refetch.

So a row can show a verdict that no longer describes the provider. Fix the thing the test complained about, save, and the old red failure text stays under the Actions column until the page is reloaded, sitting directly beneath a status pill that has already updated. The two disagree, and the stale one is the more eye-catching.

This came up while debugging an otari.ai provider: the row showed the amber "No model discovery" pill next to an error from a test run against the *previous* configuration, which made the state look self-contradictory and sent us looking for a backend bug that did not exist.

### Suggested fix

Drop `tests[instance]` when that instance is updated or deleted. The edit and delete mutations already know the instance name, so it is a small change at each mutation's success handler. Worth considering whether a successful health refetch should also clear it, so the row never carries a verdict older than the pill above it.

### Severity

Low. Cosmetic, self-corrects on reload, but actively misleading while it lasts.

njbrake:
## Description

The Providers page kept per-row connection-test verdicts in component state (the `tests` record) and never dropped them. A row could therefore show a red failure from a test that ran against the previous credentials, sitting directly under a status pill that had already refreshed. The two disagreed, and the stale one was the more eye-catching, which is how a healthy provider came to look like a backend bug during debugging.

The verdict is now dropped for an instance when that instance is saved or deleted:

- `EditProviderForm` gained an `onSaved` prop, called from the update mutation's success handler with the saved instance, so a successful save retires the verdict while a cancelled form keeps it.
- The delete mutation clears the entry too, so a provider re-added under the same name does not inherit a verdict about a different one.
- A small `clearTest(instance)` helper on the page owns the state edit; it returns the previous object when there is nothing to drop, so it does not churn a render.

### The in-flight case (second commit)

Clearing on save closes the front door, but a test already in flight walked back in through the side: `runTest`'s settle handlers wrote their result unconditionally, so a late verdict restored the very staleness the clear exists to remove. The trigger is the common case rather than a contrived one, because a test against a wrong `api_base` settles only when it times out, which is precisely when the operator goes off to fix the credentials.

A `settleTest(instance, run, state)` helper now records a result only while its run is still the current one for that row, tracked as a monotonic per-instance id in a ref. Retiring a verdict bumps the id, which is what cancels a request still in flight.

An earlier version of this branch guarded on the row's `pending` marker instead, and argued the clear-then-retest case was unreachable because the Test button is disabled while a row is pending. That reasoning was wrong: the button is per-row and keys off the marker that `clearTest` has just removed, so it is re-enabled in exactly that instant. The case was in fact unreachable only because React Query v5 detaches the observer from the previous mutation and drops its per-call callbacks. That same detach was a bug in its own right, so fixing it (below) made the race reachable and the per-instance run id necessary rather than disproportionate.

### Every row settles independently

One `useMutation` observer serves every provider row, but the Test button is disabled per row, so two tests could be in flight at once. `MutationObserver.mutate` detaches the observer from the previous mutation, discarding that call's `onSuccess`/`onError`, so testing a second provider left the first row spinning on "Testing…" with its Test button disabled for the life of the page. This reproduced on `main` too, so it is not a regression from this branch, but it lives in the lines being rewritten here. Each row now resolves from its own `mutateAsync` promise, which settles regardless of the detach.

### What is deliberately not cleared

A provider-health refetch does not clear the verdict. `useProviderHealth` polls on a `refetchInterval`, so clearing there would make a result the operator just asked for vanish on its own while nothing about the provider had changed. The issue listed it only as worth considering.

### Tests

Six Vitest cases, each confirmed to fail against the unfixed page and pass with the fix:

- editing a provider after a failed test clears the error text,
- delete then re-add under the same name comes back with no verdict,
- a test still in flight when the save lands does not write its verdict back,
- the same, for a delete rather than a save,
- a retest started after a save wins even when the pre-save test answers last,
- two rows tested at once both settle, and neither Test button stays disabled.

The test file's fetch mock gained a `PATCH` branch, which previously fell through and returned the whole stored list where a single row was expected. It mirrors the backend's `model_fields_set` semantics (an omitted field is kept, a field sent as null is cleared, see `routes/providers.py`), so it cannot mislead a later test about clearing a field.

Only `web/` changed, plus the rebuilt dashboard bundle. No API, schema, or docs surface was touched, so no OpenAPI or Postman regeneration was needed.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes #464

## Checklist

Dashboard checks, which are the ones this change touches: `npm --prefix web run typecheck` clean, `npm --prefix web test` 378 passed across 35 files, and `npm --prefix web run build` with the rebuilt bundle committed. The build is reproducible here: rebuilding at unmodified `main` left `src/gateway/static/dashboard` byte-identical, so the hash churn across app chunks is this repo's normal pattern rather than environment drift. No Python surface changed, so the OpenAPI spec and Postman collection are unaffected.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

Claude Code (Opus 5)

Implemented via the repo's fix-github-issue workflow: read the issue and the referenced page, made the minimal change at each mutation's success handler, wrote tests that were confirmed to fail without the fix, then ran the dashboard checks and rebuilt the committed bundle. The in-flight race in the second commit was found by a separate review pass over the diff, not by the implementation pass.

**NOTE:**

## Summary

- Clear provider connection-test results after a successful edit or delete.
- Preserve results when an edit is cancelled or during routine health checks.
- Keep tests independent across provider rows.
- Ignore results from tests replaced by edits, deletes, or newer tests.
- Add coverage for edits, deletion and re-adding, in-flight tests, and concurrent tests.
- Rebuild the dashboard assets.

These changes prevent outdated test results after credential changes or provider replacement.


web/src/pages/ProvidersPage.tsx
```diff
@@ -787,21 +806,42 @@ export function ProvidersPage() {
...
+      delete next[instance];
+      return next;
+    });
+
+  // Only record a result while this run's "pending" marker is still there. A test
+  // against an unreachable endpoint settles slowly, which is when the operator is
+  // most likely to go fix the credentials mid-flight; without this guard the save
+  // would clear the verdict and the timed-out test would then write it straight
+  // back, restoring the very staleness clearTest exists to remove.
+  const settleTest = (instance: string, state: TestState) =>
```
khaledosman: `testProvider` is a single `useMutation` observer serving every row (`:766`), but the Test button is disabled per row only (`:919`). TanStack Query v5's `MutationObserver.mutate` does:

```js
mutate(variables, options) {
  this.#mutateOptions = options;
  this.#currentMutation?.removeObserver(this);
  ...
}
```

So a second `mutate` detaches the first mutation and discards its per-call callbacks. Testing row B while row A is in flight means A's `settleTest` never runs: A shows the "Testing…" spinner permanently and its Test button stays disabled for the life of the page.

Confirmed with a probe test (two stored providers, both tests gated on one promise, released together): exactly one row settles, the other keeps `Testing…` with a disabled button. It reproduces identically on `main`, so **not a regression from this PR** — but it lives in the lines being rewritten here, it is the same mechanism this PR reasons about, and there is no test covering it.

The clean fix also subsumes the residual race documented in the description: give each run a per-instance id and settle only when the stored id still matches, or `await testProvider.mutateAsync(instance)` inside `runTest` and settle from the awaited result so each row owns its own settle path.

web/src/pages/ProvidersPage.tsx
```diff
@@ -787,21 +806,42 @@ export function ProvidersPage() {
...
+    setTests((prev) => {
+      // Own-property check: `in` also reports inherited keys, so an instance
+      // named after one (`toString`) would read as having a verdict it does not.
+      if (!Object.hasOwn(prev, instance)) return prev;
+      const next = { ...prev };
+      delete next[instance];
+      return next;
+    });
+
+  // Only record a result while this run's "pending" marker is still there. A test
```
khaledosman: This comment and the PR description both lean on "the Test button is disabled while a row is pending" to argue the clear-then-retest race is unreachable. That guard is per-row (`:919`) and keys off the pending marker — which `clearTest` has just removed, so the button is in fact re-enabled at exactly that moment.

The conclusion still holds, but only because React Query drops the first call's callbacks (see the other comment). Worth correcting so a future reader does not build on the wrong invariant.

web/src/pages/ProvidersPage.test.tsx
```diff
@@ -137,6 +141,27 @@ function mockApi(opts: MockOpts = {}) {
         storedList = [...storedList, row];
         return jsonResponse(row, 201);
       }
+      if (method === "PATCH") {
+        const instance = decodeURIComponent(url.split("/").pop() ?? "");
+        const body = JSON.parse(String(init?.body)) as {
```
khaledosman, line 157: This branch mirrors `model_fields_set` semantics carefully for `provider_type` / `api_base` / `api_key`, but does not parse `expected_updated_at` at all. The backend does enforce it and returns 412 (`src/gateway/api/routes/providers.py:564-566`).

The mock now reads as authoritative, so a later test for the optimistic-concurrency reload path would pass against a mock the real backend would reject. Two lines closes it:

```ts
if (body.expected_updated_at && body.expected_updated_at !== existing.updated_at) {
  return jsonResponse({ detail: "Provider changed since it was loaded; reload and retry." }, 412);
}
```

web/src/pages/ProvidersPage.tsx
```diff
@@ -965,7 +1007,13 @@ export function ProvidersPage() {
           out to be unavailable, retract it so its submit can never reach the create
           mutation. The banner above explains why. */}
       {addOpen && secretKeyConfigured ? <AddProviderForm onClose={() => setAddOpen(false)} /> : null}
-      {editingProvider ? <EditProviderForm provider={editingProvider} onClose={() => setEditing(null)} /> : null}
+      {editingProvider ? (
+        <EditProviderForm
+          provider={editingProvider}
+          onClose={() => setEditing(null)}
+          onSaved={() => clearTest(editingProvider.instance)}
```
khaledosman: Nit: this re-derives an instance the child already holds as `provider.instance`. Threading it through the callback (`onSaved: (instance: string) => void`) drops the dependency on the parent closure agreeing with the child's prop.

khaledosman (approved):
Correct fix for #464, and the reasoning about what *not* to clear (the health poll) is sound. Verified locally on `bebf1c5`: `npm --prefix web run typecheck` clean, 365/365 dashboard tests pass, and `npm --prefix web run build` leaves `src/gateway/static/dashboard/` byte-identical to what is committed, so the bundle really is in sync.

Nothing blocking. Inline: one pre-existing defect in the lines this PR rewrites (a shared test mutation can strand a row in "Testing…" permanently, reproduced on `main` too), a correction to the stated reason the residual race is unreachable, and a gap in the new mock branch.

Nits: the description's Tests section says three cases but four were added (the delete in-flight one came in `bebf1c5`), and the suite is at 365, not 364.

---

PR #469, opened 2026-08-02
feat(usage): break spend down by session, endpoint, and provider

### LINKED ISSUE #425: Usage summary is missing session, endpoint, and provider breakdowns (and never renders by_source)

Related: #299 (dashboard epic).

## Summary

`GET /v1/usage/summary` breaks down by model, user, API key, and source. Three dimensions that are stored on every `UsageLog` row are missing from the aggregates, and a fourth is computed but never rendered. All four are close to free, because `_breakdown()` is already generic over a column.

## The gaps

**1. `source_label` (session) — not aggregated anywhere.**
It's stored per row, returned by `/v1/usage`, and rendered in `ActivityPage.tsx:140` as a detail field. But there is no `by_source_label`, and it is not even a filter on `list_usage`. For agent traffic this is the single most useful grouping: in a measured Claude Code workload of 3,551 requests across 50 sessions, the **top 3 sessions were 57.7% of all tokens** and one session alone was 39.5%. Without a session dimension, "spend went up" can never become "*this* task went wrong."

**2. `endpoint` — not aggregated.**
Stored, and already a supported filter on `list_usage`, but no `by_endpoint`. An operator running `/v1/messages` + `/v1/responses` + `/v1/chat/completions` + `/v1/embeddings` through one gateway cannot split spend by API surface.

**3. `provider` — not aggregated.**
Stored on every row, no `by_provider`. For a gateway whose core purpose is multi-provider routing, "which provider am I spending on, which is slow, which is failing" is a first-order question. `/v1/providers/health` covers liveness, not usage.

**4. `by_source` — computed, shipped, and dropped on the floor.**
The server computes it (`usage.py:620`) and returns it in `UsageSummary`, but it is absent from `web/src/api/types.ts` and rendered nowhere. So imported agent usage can't be compared against gateway-served usage in the UI even though the data is already on the wire.

## Proposal

- Add `by_source_label`, `by_endpoint`, and `by_provider` to `UsageSummary`, alongside the existing breakdowns.
- Add `source_label` to `_usage_filters` so the request log can be scoped to one session (see the Activity-page issue).
- Add `by_source` to the client types and render it.
- Consider whether `by_source_label` needs a higher `_BREAKDOWN_TOP_N` than the other dimensions, since session cardinality is much higher than model/user cardinality.

## Why this is worth doing first

These are the cheapest analytic wins available — one `_breakdown()` call each — and they serve both audiences: session is what agent operators need, endpoint and provider are what gateway operators need.

njbrake:
## Description

`GET /v1/usage/summary` broke spend down by model, user, API key, and source. Three dimensions stored on every `UsageLog` row had no aggregate at all, and a fourth was computed, shipped, and rendered nowhere.

- **`source_label` (session)** had no aggregate and was not even a filter, so agent traffic could not be attributed to the task that caused it. A few long-running sessions routinely carry most of a workload's tokens, so this is the grouping that turns "spend went up" into "this task went wrong".
- **`endpoint`** and **`provider`** had no aggregate, so an operator running several API surfaces or several upstreams through one gateway could not split spend by either.
- **`by_source`** was already computed and returned, but absent from the client types and rendered nowhere.

**Backend.** `UsageSummary` gains `by_source_label`, `by_endpoint`, and `by_provider`, one `_breakdown()` call each. Sessions get a deeper top-N (250 vs 100): their cardinality is an order of magnitude higher than model or user cardinality, so the shallow cap would fold away the head that matters. All three also join the CSV export, so it still mirrors the summary rather than reporting four of seven dimensions. `source_label` and `provider` join `endpoint` and `source` as filters on the list, count, and summary endpoints, so every breakdown row can drill into the request log behind it. `provider` goes slightly beyond the issue's ask: it was the only rendered dimension with no filter, and a breakdown you cannot drill into is a dead end.

**Dashboard.** The four secondary dimensions (session, endpoint, provider, source) share one panel behind a picker rather than six stacked tables; session is the default. Model and user keep their always-visible tables. The Activity page accepts `source_label`, `endpoint`, and `provider` as drill-down filters shown as clearable chips, mirroring how `source` already worked. Unlabelled gateway rows read as "(no session)" rather than "(unknown)": carrying no session label is their normal state, not missing data.

**Bug fix, worth calling out separately.** An "all matching" bulk delete or reprice dropped the `source` filter from its request body. The count that sizes "Select all N matching" is computed under every active filter, but the body the server re-derives the target set from omitted `source`, so a delete launched from a drill-down scoped to one imported source removed **every** imported row in the window instead of the N the operator was shown. On `main` this was latent: the only in-app link that set `source` scoped to gateway traffic, which bulk ops never touch, so reaching it took a hand-edited URL. The new Source, Session, Endpoint, and Provider drill-downs turn it into a one-click path, so `selectionBody` now forwards every filter the table honors. Pinned by "carries the drill-down filters into an 'all matching' delete" in `ActivityPage.test.tsx`.

**Cost, and the `dimensions` selector that answers it.** `/summary` computes seven `GROUP BY` scans where it computed four, over the same bounded, timestamp-indexed window. Review flagged that the Usage and Activity pages fire several summary calls concurrently (tiles, timeline context, model typeahead, previous period) and that most of them read only `totals` and `series`, so rather than leave the selector as a follow-up this PR adds it: `/summary` now takes a repeatable `dimensions=` parameter where each value names the `by_<value>` field it fills. Omitting the parameter computes all seven, so existing consumers are unaffected; an excluded field still returns `[]`, so the response schema never changes shape; and `none` is the explicit empty selection. The dashboard now asks only for what it renders, counted per hook: 7 breakdown passes instead of 28 on Usage, 0 instead of 21 on the overview, and 1 instead of 14 on Activity. Worth being precise about the Usage number: on an unfiltered page three of those reads already shared a query key and TanStack deduped them into one request, so the real saving there is 14 to 7. The full saving lands once a filter or a custom range makes the keys diverge, and everywhere on Overview and Activity, which never deduped. The CSV export walks the same dimension table, so an eighth grouping is one edit rather than two.

Two things left deliberately undone, both recorded rather than fixed. `source_label` and `provider` stay unindexed: a `GROUP BY` inside a timestamp-bounded window scans the range and hash-aggregates whatever indexes exist, and the new equality filters are range-bounded too, making them the same shape as the existing `model` filter that `entities.py` leaves unindexed on purpose. Only a composite `(provider, timestamp)` mirroring `ix_usage_logs_status_timestamp` would help, and that is worth sizing against a real workload before adding write cost to the hottest insert table in the schema. And `summary.csv` stays uncapped, which is its documented behavior; the memory shape is worse with sessions in it, and the larger half is that `_breakdown(limit=None)` materializes a pydantic row per group across all seven dimensions before any CSV text exists, so the fix is a streaming response with a server-side cursor rather than a quiet row ceiling.

One reading note now that #472 has landed: it writes gateway-side rejection rows with `provider = NULL`, so those rows surface in the new `by_provider` breakdown as a real null-key group rendered "(unknown)". That is correct, they are genuine rows and with `cost = NULL` they sort last, but it is a reading of that table that did not exist when the breakdown was written.

## PR Type

- [x] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes #425

## Checklist

Tests added: 9 backend integration tests (breakdown content, the deeper session cap, filter scoping on summary and on list/count, CSV dimensions, and four covering the `dimensions` selector) and 10 dashboard tests (default session render, dimension switching, session and provider drill-through, the non-drillable null row, the `source` filter in an "all matching" body, and the `aria-pressed` state on both toggle groups).

Local results, after rebasing onto `1a55d9f0`: `make lint` clean including the new architecture check, `mypy` strict clean over 276 files, 1059 unit tests pass, 853 integration tests pass, 372 dashboard tests pass, `make openapi-check` and `make postman-check` both up to date, dashboard bundle rebuilt and reproducible (a second build leaves the tree clean). Two integration tests fail (`test_provider_error_does_not_leak_details`, `test_streaming_creation_error_returns_http_error`); both fail identically on a clean `HEAD` and on both the pre- and post-rebase base, because they make unmocked provider calls and this machine has outbound network plus a real credential, so they are unrelated to this change.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

Implemented from issue #425 through the fix-github-issue workflow: the issue's proposal was followed directly, with two deliberate additions noted above (the `provider` filter, so provider rows can drill, and the new dimensions in the CSV export, so it does not diverge from the summary). Review follow-ups, including the `dimensions` selector and the `aria-pressed` fix, were addressed in a later pass. Reviewed by @njbrake before merge.


njbrake:
@coderabbitai review

njbrake:
@coderabbitai full review

web/src/pages/ActivityPage.tsx
```diff
@@ -469,6 +538,10 @@ export function ActivityPage() {
           user_id: filters.user_id,
           api_key_id: filters.api_key_id,
           status: filters.status,
+          source: filters.source,
```
khaledosman, line 548: This line is a bug fix, not part of the feature, and it deserves to be called out. Before it, an operator who drilled into an imported source, hit "select all N matching", and deleted, sent a body with no `source` — the server re-derived the set and removed **every** imported row in the window, not the ones counted for them. Worth its own line in the PR description (and it makes the case for the comment you added above `selectionBody` much more concretely than the new filters do).

src/gateway/api/routes/usage.py
```diff
@@ -652,6 +695,9 @@ async def usage_summary(
     by_user = await _breakdown(db, UsageLog.user_id, conditions, totals, limit=_BREAKDOWN_TOP_N)
     by_api_key = await _breakdown(db, UsageLog.api_key_id, conditions, totals, limit=_BREAKDOWN_TOP_N)
     by_source = await _breakdown(db, UsageLog.source, conditions, totals, limit=_BREAKDOWN_TOP_N)
+    by_source_label = await _breakdown(db, UsageLog.source_label, conditions, totals, limit=_SESSION_BREAKDOWN_TOP_N)
+    by_endpoint = await _breakdown(db, UsageLog.endpoint, conditions, totals, limit=_BREAKDOWN_TOP_N)
+    by_provider = await _breakdown(db, UsageLog.provider, conditions, totals, limit=_BREAKDOWN_TOP_N)
```
khaledosman: `/summary` goes from 4 `GROUP BY` scans to 7 here, and #470 adds an 8th. The Usage and Activity pages fire several summary calls concurrently (tiles, timeline context, model typeahead, previous period) and most of them read only `totals`/`series`. Two PRs doubling the same query in one release makes the `dimensions=` selector you name as a follow-up worth doing now — or at least letting the typeahead/context callers opt out. `source_label` and `provider` are also unindexed, so both new filters and both new groupings scan.

src/gateway/api/routes/usage.py
```diff
@@ -734,6 +789,9 @@ async def usage_summary_csv(
         ("user", await _breakdown(db, UsageLog.user_id, conditions, totals, limit=None)),
         ("api_key", await _breakdown(db, UsageLog.api_key_id, conditions, totals, limit=None)),
         ("source", await _breakdown(db, UsageLog.source, conditions, totals, limit=None)),
+        ("session", await _breakdown(db, UsageLog.source_label, conditions, totals, limit=None)),
```
khaledosman: The CSV is deliberately uncapped and now covers 7 dimensions including the highest-cardinality one, all accumulated in `io.StringIO` before the response. A wide window over agent traffic could produce a very large in-memory body. Pre-existing shape, but materially worse with sessions in it — a streaming response or a row ceiling with an explicit header would be worth a follow-up.

web/src/pages/UsagePage.tsx
```diff
@@ -741,6 +786,44 @@ export function UsagePage() {
...
+              })
+            }
+            loading={summary.isLoading}
+            action={
+              <div className="inline-flex gap-1.5">
+                {DIMENSION_TABS.map((tab) => (
+                  <Button
+                    key={tab.key}
+                    size="sm"
+                    variant={dimension === tab.key ? "primary" : "outline"}
```
khaledosman, line 834: Selected state is carried only by `variant`, so assistive tech sees four identical buttons with no indication of which dimension is active. `aria-pressed={dimension === tab.key}` is the one-line fix; a tabs/toggle-group primitive would be the idiomatic one.

khaledosman (approved):
One thing the description undersells: this also fixes a data-loss bug (`source` was missing from the "all matching" mutation body), flagged inline — that belongs in the PR body and the changelog, not just in the diff.

Otherwise: the summary-scan cost, the uncapped CSV now that the highest-cardinality dimension joined it, and an a11y nit on the dimension picker.

---

Issue #410, opened 2026-07-24
Dashboard/API: Monitoring


njbrake:
The dashboard should have a notification system? Some method to subscribe to alerts so that you as the admin can get notified when a provider goes down or if a budget hits its limit, etc. 

Right now there is no method for otari to tell the admin that something is wrong unless the admin logs in and checks.

PR #470, opened 2026-08-02
feat(usage): record a status code on usage logs so failures can be classified

### LINKED ISSUE #433: UsageLog has no status_code, so failures can be counted but not classified

Related: #317 (surface failing requests) — this is the missing data that issue's "possible directions" depend on.

## Summary

`UsageLog` records `status` (`"success"` / `"error"`) and a free-text `error_message`, but no HTTP status code. So failures can be counted but not *classified*, and the classification is the entire point: 429 rate limits, 5xx provider faults, 402 missing-pricing rejections, 403 budget denials, and context-length rejections call for completely different operator responses.

## Why this is worth its own issue

#317 proposes a failing-requests view and notes that "the data to count and break down failures largely exists" since `UsageLog` carries `status` and `error_message`. That's true for counting but not for breaking down. With only a boolean status and free text, any breakdown has to be built on substring matching over provider-specific error prose, which is brittle across providers and changes without warning.

A `status_code` column makes the breakdown a `GROUP BY` instead of a heuristic, and it's a small migration.

## What it unlocks

- **Error taxonomy** in the usage summary: failures grouped by code, per model and per provider. "Which provider is rate-limiting me" and "how much of my error rate is my own misconfiguration" become answerable.
- **Actionable alerting** — a spike in 429s means raise limits or shed load; a spike in 5xx means fail over; a spike in 402s means fix pricing config. These are different pages to wake someone for. (Relevant to #410.)
- **Error grouping in the Activity log**, so browsing failures doesn't require expanding rows one at a time (see the Activity-page issue). Today the detail panel deliberately renders a fixed string — "The provider returned an error. Inspect gateway logs for details." — so the UI can never explain a failure. A status code is safe to surface: it's a classification, not provider diagnostics or prompt content.
- **Retry-cost attribution**, alongside the attempt/fallback counts proposed in #401 — knowing *what* was retried, not just how often.

## Proposal

- Add a nullable `status_code: int | None` to `UsageLog` (nullable so historical rows stay valid).
- Populate it from the upstream response on both the `_pipeline` and `_passthrough` error paths, and from the gateway's own rejection code when otari itself rejects (402 missing pricing, 403 budget, 401 auth).
- Expose it as a filter on `/v1/usage` and as a breakdown dimension in the summary.
- Optionally normalize into coarse classes (`rate_limit`, `provider_error`, `auth`, `budget`, `pricing`, `client_error`) for display, keeping the raw code for precision.

Note: Claude Code's own `api_request` event already carries a `status_code` attribute, so the OTLP import path could populate this for imported usage too.

njbrake:
## Description

`UsageLog` carried `status` (success/error) and a free-text `error_message`, so failures could be counted but not broken down. Any breakdown had to be built on substring matching over provider-specific error prose, which differs per provider and changes without notice.

This adds a nullable `usage_logs.status_code` and populates it on every path that already writes an error row: the pipeline's non-streaming and streaming provider failures, the tool-loop cap (422), the missing-pricing rejections (402) in both the pipeline and the pass-through gate, the pass-through provider failure, and batch creation.

**The recorded code prefers the status the provider returned, which is deliberately not always the status the caller saw.** An upstream 401/403 is a provider rejecting the gateway's credentials, so the response stays a generic 502 and never says so; the log keeps the 401, which is what makes "how much of my error rate is my own misconfiguration" answerable. A provider that never answered carries no status, so the row records the gateway's own classification (504 timeout, 502 unreachable) rather than staying unclassifiable. Success rows stay NULL: there is no failure to classify, and a constant 200 would dilute every `GROUP BY` over the column.

New read surface:

- `status_code` filter on all four `/v1/usage` read endpoints (list, count, summary, summary.csv).
- `errors_by_status_code` on `/v1/usage/summary`, scoped to failures and carrying a coarse `error_class` (`pricing` / `rate_limit` / `auth` / `provider_error` / `client_error`) alongside the raw code. Counts reconcile with `totals.error_count`.

One shared-utility change: mid-stream failures settle through `streaming_generator`'s `on_error`, which received a rendered message. It now receives the exception itself so the callback can classify it; a string cannot carry a status. Without this, the streaming half of the traffic would log NULL.

### Two parts of the issue deliberately left out

- **403 budget / blocked-user and 401 auth rejections.** These write no usage-log row at all today, so there is nothing to stamp. The issue presumes those rows exist; adding them is #317's scope (which is what added the 402 row) and would put a log write on a hot rejection path. 401 is additionally out of reach: it happens before a model or user is resolved, and `usage_logs.model` is non-nullable.
- **The OTLP import path.** It hardcodes `status="success"` and drops events with no token counts, so populating a status code there would change *what gets imported*, not just classify it. The issue framed this as a "could", not part of the proposal.

No dashboard changes, so the committed bundle is not stale. The Activity-log and alerting consumers of this data are follow-ups (#317, #410).

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes #433

## Checklist

### Test notes

New `tests/integration/test_usage_status_code.py` covers: the upstream status landing per code; the 401-recorded-behind-a-502 case (the one that makes the column load-bearing); timeouts recording 504; a streaming failure recording its upstream status; success recording NULL; the filter agreeing between list and count; and the summary grouping with its coarse classes. `tests/unit/test_usage_error_class.py` pins the class mapping including garbage codes. `failure_status_code` unit tests live alongside the existing `classify_provider_error` ones.

Migration `d7b9f1a3c5e8` applies and reverses cleanly against PostgreSQL, and `alembic check` reports no drift for `usage_logs`.

Full unit suite passes (1068). Integration: 830 passed, 2 failed, both pre-existing and environmental in my sandbox (`test_streaming_error_event` and `test_error_detail_leakage::test_provider_error_does_not_leak_details` make a real provider call and get a 404 where they expect the generic 502). I confirmed both fail identically on clean `origin/main` in a separate worktree, so they are not a regression from this change; worth a sanity check that they pass in CI.

## AI Usage

- [ ] No AI was used.

Implemented by Claude Code via back-and-forth with @njbrake. The design decisions (recording the upstream status rather than the client-facing one, keeping success rows NULL, scoping out the 401/403 and OTLP paths) were reviewed by him; the prose is Claude's.

**NOTE:**

### DIFF
--- a/alembic/versions/d7b9f1a3c5e8_add_usage_logs_status_code.py
+++ b/alembic/versions/d7b9f1a3c5e8_add_usage_logs_status_code.py
@@ -1,0 +1,30 @@
+"""Add usage_logs status_code column.
+
+Revision ID: d7b9f1a3c5e8
+Revises: c3f7a9d1e5b8
+Create Date: 2026-07-31 10:00:00.000000
+
+"""
+
+from collections.abc import Sequence
+
+import sqlalchemy as sa
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision: str = "d7b9f1a3c5e8"
+down_revision: str | Sequence[str] | None = "c3f7a9d1e5b8"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+
+def upgrade() -> None:
+    """Upgrade schema."""
+    # Nullable: historical rows predate the column, a successful request has no
+    # failure to classify, and some failures carry no HTTP status at all.
+    op.add_column("usage_logs", sa.Column("status_code", sa.Integer(), nullable=True))
+
+
+def downgrade() -> None:
+    """Downgrade schema."""
+    op.drop_column("usage_logs", "status_code")
--- a/alembic/versions/d7b9f1a3c5e8_add_usage_logs_status_code.py
+++ b/alembic/versions/d7b9f1a3c5e8_add_usage_logs_status_code.py
@@ -22,6 +22,14 @@ def upgrade() -> None:
     """Upgrade schema."""
     # Nullable: historical rows predate the column, a successful request has no
     # failure to classify, and some failures carry no HTTP status at all.
+    #
+    # Deliberately unindexed. Filtering on a code implies status='error' (see
+    # ``_usage_filters``) and every aggregate over it is range-bounded, so the
+    # existing (status, timestamp) index already narrows to the error rows in the
+    # window before the code is applied. usage_logs is append-heavy on the request
+    # path, so a further (status, status_code) index would tax every write to save
+    # a scan over one window's failures; add it if error volume ever makes that
+    # scan the bottleneck.
     op.add_column("usage_logs", sa.Column("status_code", sa.Integer(), nullable=True))
 
 
--- a/src/gateway/api/routes/_pipeline.py
+++ b/src/gateway/api/routes/_pipeline.py
@@ -2335,7 +2335,7 @@ async def run_standalone_non_stream(
         # Gateway-owned cap, not an upstream provider failure. 422 lets
         # callers distinguish a runaway tool loop from a real outage.
         logger.warning("Tool loop iteration cap hit (standalone): cap=%d", tool_ctx.max_tool_iterations)
-        await _log_failure_and_refund(ctx, adapter, provider, model, str(e), 422)
+        await _log_failure_and_refund(ctx, adapter, provider, model, str(e), status.HTTP_422_UNPROCESSABLE_CONTENT)
         raise adapter.error(422, str(e), ErrorKind.INVALID_REQUEST) from e
     except SandboxNotReachableError as e:
         # Sandbox is gateway-side infra, not an LLM provider. Clearer detail
--- a/src/gateway/api/routes/_pipeline.py
+++ b/src/gateway/api/routes/_pipeline.py
@@ -297,6 +297,25 @@ def classify_provider_error(exc: BaseException) -> ProviderErrorMapping | None:
     return None
 
 
+def failure_status_code(exc: BaseException) -> int:
+    """The HTTP status to record on the usage log for an upstream failure.
+
+    Prefers the status the provider actually returned, which is deliberately not
+    always the status the caller saw: an upstream 401/403 surfaces to the caller
+    as a generic 502 (a provider rejecting the gateway's credentials is a
+    gateway-config fault, and the response must not say so), but the log keeps
+    the 401 so "how much of my error rate is my own misconfiguration" stays
+    answerable. When the provider returned no status at all (timeout,
+    unreachable), records the gateway's own classification instead, so an error
+    row still carries a code to group on.
+    """
+    _kind, status_code = upstream_exception_shape(exc)
+    if status_code is not None:
+        return status_code
+    mapping = classify_provider_error(exc)
+    return mapping.status_code if mapping is not None else status.HTTP_502_BAD_GATEWAY
+
+
 _DEFAULT_PORTS = {"http": 80, "https": 443}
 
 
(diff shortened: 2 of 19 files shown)


njbrake:
@coderabbitai review

njbrake:
@coderabbitai full review

src/gateway/api/routes/_pipeline.py
```diff
@@ -726,6 +755,7 @@ async def resolve_request_context(
                 endpoint=adapter.endpoint,
                 user_id=user_id,
                 error=no_pricing_detail,
+                status_code=402,
```
khaledosman: Bare int literal where `_passthrough.py` uses `status.HTTP_402_PAYMENT_REQUIRED` for the same value (and `_log_failure_and_refund(..., 422)` further down does the same). Repo convention is the `fastapi.status` constants; worth keeping the two scaffolds spelled identically since they are read side by side.

src/gateway/api/routes/usage.py
```diff
@@ -536,6 +577,60 @@ async def _breakdown(
     return result
 
 
+def error_class_for(status_code: int | None) -> ErrorClass:
+    """Coarse display bucket for a failure's HTTP status code.
+
+    401 and 403 read as ``auth`` rather than as a budget denial because the codes
+    that actually reach this column come from upstream: a provider rejecting the
+    gateway's credentials. Budget and blocked-user rejections are refused before
```
khaledosman: This claim stops being true the moment #472 merges — it logs exactly these rejections (blocked user and over-budget, raised inside `reserve_budget`). Once both are in, those rows exist and, without a `status_code` on the shared writer, classify as `unknown` rather than as the 403 they are. Worth updating the docstring on whichever rebase lands second.

alembic/versions/d7b9f1a3c5e8_add_usage_logs_status_code.py
```diff
@@ -0,0 +1,30 @@
...
+down_revision: str | Sequence[str] | None = "c3f7a9d1e5b8"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+
+def upgrade() -> None:
+    """Upgrade schema."""
+    # Nullable: historical rows predate the column, a successful request has no
+    # failure to classify, and some failures carry no HTTP status at all.
+    op.add_column("usage_logs", sa.Column("status_code", sa.Integer(), nullable=True))
```
khaledosman, line 33: No index covers the new column. `usage_logs` has `(user_id, timestamp)`, `(status, timestamp)` and `source`, so a `status_code=429` query that does not also pass `status=error` scans the window. The dashboard always sends both, so this is low urgency — but either say so in the filter description or plan a `(status, status_code)` index before someone builds an alert on the bare filter.

src/gateway/api/routes/batches.py
```diff
@@ -343,6 +345,7 @@ async def create_batch(
             endpoint="/v1/batches",
             user_id=user_id,
             error=str(e),
+            status_code=failure_status_code(e),
```
khaledosman, line 348: This stamp and the pass-through provider-failure one (`_passthrough.py`) are both unexercised: the new integration tests cover the chat pipeline, and `test_require_pricing.py` covers the 402 rows. A `status_code` that silently stops being written on the batch or pass-through path would not fail anything today.

khaledosman (approved):
Merge-order note first: #472 replaces every rejection-row construction this PR stamps (the 402 in `resolve_request_context` and the pass-through one) with a shared `log_gateway_rejection` that takes no `status_code`. Whichever of the two lands second has to thread the code through that writer, or all four of #472's new gates write NULL and land in `errors_by_status_code` as `unknown` — the gateway-side drops both PRs exist to surface.

The 401-behind-a-502 case is the right thing to pin, and checking `MaxToolIterationsExceeded` before the upstream shape in `failure_status_code` is a non-obvious ordering the docstring earns. Rest inline.

---

PR #471, opened 2026-08-02
fix(models): let a keyless local provider opt into discovery

### LINKED ISSUE #389: Keyless local providers (ollama/llamacpp/llamafile) are callable but never discoverable

## Summary

Keyless local providers (`ollama`, `llamacpp`, `llamafile`) are callable for completions but can never appear in `GET /v1/models`, so the discovery vs. routing consistency that #221 established does not extend to them.

## Background

#221 (PR #388) made `/v1/models` discovery match completion routing for providers whose native credential environment variable is set: a provider with, for example, `ANTHROPIC_API_KEY` present is now discovered under its implementation name even without a `providers:` entry.

That mechanism keys off `metadata.env_key`. The local providers have no credential env var (`env_key` is absent / the literal `"None"`), so there is no signal to detect them by. `_env_provider_instances` (`src/gateway/services/model_discovery_service.py`) deliberately skips them, since auto-discovering one would mean dialing a fixed localhost endpoint on spec.

## Impact

Running a local Ollama and calling `ollama:llama3` works, but the model never shows up in `GET /v1/models` (or `/v1/models/discoverable`) unless `ollama` is added to `config.providers`. This is the same class of disagreement #221 addressed, just for the keyless-provider subset.

## Possible directions

1. Let an operator opt a keyless local provider into discovery via a lightweight `providers:` entry (already works today: add `ollama` with a `models:` list or rely on its `/v1/models`), and document that this is the intended path for local backends.
2. Optionally probe a small set of well-known local providers when their default endpoint is reachable (bounded, opt-in), so a running local server is auto-listed.
3. At minimum, document the distinction so the local-provider path is clear.

## Related

Follow-up to #221 / #388.

njbrake:
## Description

Keyless local backends (ollama, llamacpp, llamafile) take no credential, so there is no environment variable that could signal one is in use, and the gateway will not dial a localhost port on spec. Discovery is scoped to the configured instances, which leaves the `providers:` entry as the way to get a local backend's models into `GET /v1/models`.

That entry was awkward to write. A keyless backend has nothing to put under its key, and the natural YAML

```yaml
providers:
  ollama:
```

parses as `None`, which the `dict[str, dict[str, Any]]` annotation rejected with a pydantic type error. Operators had to know to write `ollama: {}` instead.

This PR coerces a valueless provider entry to an empty config block, so the bare key loads and the instance becomes both routable and discoverable.

### A note on the issue's premise

The issue says #221 / #388 made discovery match routing for providers whose credential env var is set, and frames local providers as the leftover gap. That mechanism was reverted on main by #415 (commit `e6c9da9b`, "discover models only from configured providers"); `_env_provider_instances` no longer exists. Discovery today is scoped to `config.providers` for every provider, keyless or not, so the local providers are not a special case any more. The remaining local-specific gap is the config papercut above.

`docs/models.md` was still documenting #388's removed behavior ("its models also appear in `GET /v1/models`, even without a `providers` entry"), which has been wrong since #415. Corrected here, since writing the local-provider guidance next to a contradictory paragraph would not have helped anyone.

### Not done here

Probing well-known local endpoints for reachability (direction 2 in the issue). That would put back the auto-detection that scoping discovery to configured instances deliberately removed, and deserves its own decision rather than arriving as a side effect of this fix.

### Changes

- `src/gateway/core/config.py`: a `providers` before-validator normalizing a valueless entry to `{}`. No effect on any other entry shape.
- `docs/models.md`: corrected the stale env-var discovery claim, and split routing (env var suffices) from discovery (needs the entry). New "Local providers" section covering the bare entry, the default endpoints (`127.0.0.1:11434` for Ollama, which also honors `OLLAMA_HOST`; `127.0.0.1:8080/v1` for llama.cpp and llamafile), the `api_base` override, and what happens when the server is offline.
- `config.example.yml`: commented `ollama:` example.
- Tests: two unit tests in `test_provider_instances.py`, one integration test in `test_model_discovery.py` asserting `ollama:llama3` reaches `GET /v1/models` from a valueless entry.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [x] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes #389

## Checklist

### Verification

All three new tests were confirmed to fail with the validator disabled and pass with it.

`make lint`, `make typecheck` (272 files), `make openapi-check`, and `make postman-check` all pass. Full suite: 1856 passed, 2 failed, 10 skipped. The two failures (`test_error_detail_leakage.py::test_provider_error_does_not_leak_details` and `test_streaming_error_event.py::test_streaming_creation_error_returns_http_error`) both dial real OpenAI and fail on the sandbox's lack of network egress; both were confirmed to fail identically with `config.py` restored to HEAD, so they are pre-existing and unrelated to this change.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

Implemented by Claude through back-and-forth with @njbrake. The investigation that turned up the reverted #388 mechanism and the stale doc paragraph, the decision to fix the config papercut rather than add localhost probing, and the scope boundary above are all part of that discussion; the prose is Claude's.

## Summary

- Keyless local providers now support valueless configuration entries such as `ollama:`.
- Configured keyless providers now appear in `GET /v1/models`.
- Documentation now explains routing, model discovery, local-provider endpoints, offline behavior, and `api_base` overrides.
- Added configuration examples and regression tests for provider parsing and model discovery.
- Endpoint reachability probing remains outside the scope of this change.


src/gateway/core/config.py
```diff
@@ -732,6 +732,33 @@ def validate_provider_instances(self) -> None:
...
+
+        A ``providers:`` block with no entries at all gets the same treatment, so
+        commenting out every entry reads as "no providers" rather than the same
+        confusing type error.
+        """
+        if providers is None:
+            return {}
+        if not isinstance(providers, dict):
+            return providers
+        return {instance: ({} if entry is None else entry) for instance, entry in providers.items()}
```
khaledosman, line 828: This coerces *any* valueless entry, not just the keyless backends the docstring is about. `providers:\n  openai:` used to be a startup type error; now it loads as a configured OpenAI instance with no credential — routable, discoverable, and either failing at request time or silently picking up `OPENAI_API_KEY`. For ollama / llamacpp / llamafile that is exactly right; for a keyed provider a bare entry is far more likely a truncated YAML edit than intent, and the old error caught it at startup. A warning (or a check in `validate_provider_instances`) for a bare entry naming a provider that normally needs a credential would keep the papercut fixed without losing the typo detection.

khaledosman (approved):
Correcting `docs/models.md` for the #415 revert is the more valuable half of this — the note that `_env_provider_instances` no longer exists saves the next reader a wrong turn, and the routing-vs-discovery split is now stated where it was previously contradicted.

One scope question on the validator, inline.

---

PR #472, opened 2026-08-02
feat(observability): log the remaining gateway-side rejections

### LINKED ISSUE #465: Log the remaining gateway-side rejections, and de-duplicate the UsageLog literals in _passthrough.py
Follow-ups from #449, which closed #317 by logging the `require_pricing` rejection and reading it back as a live failure count in the dashboard.

## 1. The other gateway-side rejections still write no usage row

#317 asked for "missing-billing **(and other)** rejections". #449 delivered the missing-billing case and the counting infrastructure, but only the two missing-pricing gates log. Every other gateway-side rejection still raises without writing a `UsageLog` row, so the failure count reads 0 during those incidents and the activity log shows nothing:

- `_pipeline.py:680`: model not allowed for the key (403)
- `_pipeline.py:650`: forbidden user, on a user/key mismatch (403)
- budget rejections raised inside `reserve_budget` (`services/budget_service.py`): over budget (403)
- the unresolvable-selector paths that map to 400

Auth failures (401) never reach a point where a user is known, so they are a separate question: a row with no `user_id` may not be worth writing. Worth deciding explicitly rather than by omission.

The dashboard side needs nothing new. The count and the drill-down already read `status=error` scoped to `source=gateway`, so any gate that starts logging flows into both for free.

Two things to keep in mind when doing this:

- **Refund before logging.** Each of those sites already refunds its reservation; the log call goes after the refund and must not change it. The rows carry `cost=null`, so they never move spend.
- **`counts_toward_budget` must stay `True`.** The dashboard treats `counts_toward_budget=False` as imported and offers those rows to bulk delete and set-price. A gateway-written row must never land in that set. `tests/integration/test_require_pricing.py::test_budget_exempt_key_writes_no_pricing_rejection_row` pins this for the pricing gate and is worth copying per gate.

## 2. Three hand-built `UsageLog` literals in `_passthrough.py`

Raised by @khaledosman in review ([comment](https://github.com/mozilla-ai/otari/pull/449#discussion_r3681061171)). `run_passthrough` now builds the same eleven-field `UsageLog` in three places: the success row, the provider-error row, and the missing-pricing rejection row added by #449. Two of the three predate that PR, so the cleanup was left out of it rather than bundled into a feature change.

A local `_usage_row(status, **overrides)` helper would make the next field addition a one-line change instead of three sites to keep in sync.

While in there, worth deciding deliberately about a related split #449 pinned but did not change: the pass-through gate logs the bare model (`text-embedding-3-small`) with the instance in `provider`, where the chat pipeline logs the full `instance:model` selector. The two forms render differently in the Activity log's Model column and match different values of the model filter. Both forms are asserted in `test_require_pricing.py` so a change is deliberate, but the inconsistency itself is still there.

## Not included

The failure reason is deliberately not shown per row: the activity detail panel renders a fixed "This request failed. Inspect gateway logs for details." because it cannot distinguish a safe gateway message from a raw provider one, and provider diagnostics stay server-side. Surfacing a safe reason needs that distinction and belongs with #304.

njbrake:
## Description

#449 made one gateway refusal visible: the missing-pricing gate wrote an error row, and the dashboard read that back as a live failure count. Every other gate still raised without writing anything, so during those incidents the count read 0 and the activity log showed nothing while requests were being dropped. That is the situation #317 was filed to end.

**Log the remaining gateway-side rejections.** A row is now recorded for a model outside a key's allow-list, a user/key mismatch, a blocked or over-budget user (raised inside `reserve_budget`), and a selector that no longer resolves to a configured provider. Both scaffolds are covered: the chat/messages/responses pipeline and the pass-through routes behind embeddings, images, audio, moderations, and rerank.

All of them go through one writer, `log_gateway_rejection`, which fixes `counts_toward_budget=True`. That matters per gate: the dashboard classifies `counts_toward_budget=False` as imported usage and offers those rows for bulk delete and set-price, and a key flagged `exclude_from_budget` can reach these gates (unlike the pricing gate, which is guarded by `not budget_exempt`), so inheriting its flag would have put rejection rows in that set. There is no cost on these rows for the flag to gate. Reservations are untouched: each site refunds exactly as before and the row carries `cost=null`, so nothing moves spend.

**Some refusals stay unlogged, deliberately rather than by omission,** for three different reasons. A 401 is refused before any user is known; the column is nullable so a NULL-user row would insert fine, but it could not be attributed, filtered, or acted on, and writing one would let an unauthenticated caller append to the usage table. A 404 for a named nonexistent user is skipped for a harder reason: `usage_logs.user_id` is a foreign key to `users`, so that row could not be inserted at all. A 429 is skipped because a throttle is expected, self-limiting behavior rather than dropped traffic, unlike the client-driven gates that do log. All three are pinned by tests. Note this is not an exhaustive list of unlogged refusals: a guardrail block, an unreachable sandbox or web-search backend, and the responses-unsupported 400 also write no row today, all outside the scope #465 listed.

**The model/provider split is unified rather than left alone.** Rejection rows now name the resolved target (`model=gpt-4o`, `provider=openai`) instead of the request selector. #449 logged the full `instance:model` selector on the chat gate, which split a model's failures from its successes: the activity log's model filter is an exact match on that column, and usage-by-model groups on it. Both scaffolds now write the one form every success row already used. The assertions `test_require_pricing.py` pinned are updated accordingly.

**The three hand-built `UsageLog` literals in `_passthrough.py` collapse to one** (raised by @khaledosman in review of #449): `_usage_row(status, **outcome)` for the outcomes of an attempted provider call, and the shared rejection writer for the missing-pricing gate. Adding a column is now a one-line change instead of three sites to keep in sync.

`resolve_dispatch_provider` becomes async (5 call sites) so the unresolvable-selector 400 can log, and it now refunds, which it never did. I first wrote that off as a formality on the grounds that a selector we cannot resolve has no pricing, so the hold would always be 0.0. That is wrong, and review caught it. The preamble carries an unresolvable selector into the pricing lookup as the bare model with no provider, and `find_model_pricing` then keys on the model alone, which is exactly the `provider:model` form stored pricing rows use. So an instance removed from `config.yml` while its pricing row survives prices normally, reserves a real estimate, and only fails later at dispatch; before this refund the hold stayed on `users.reserved` until the next budget reset, or forever for a budget with no period. The regression test leaks 0.0102 with the refund removed. This PR therefore fixes a real reservation leak on top of the logging work.

The dashboard side needed nothing new: the count and its drill-down already read `status=error` scoped to `source=gateway`. The bundled guide (`docs/dashboard.md`) is updated to name which refusals are logged and which are not, so the dashboard bundle is rebuilt with it; the 17 renamed asset files are the usual hash cascade off the entry chunk, byte-identical apart from import filenames.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [x] Refactor
- [x] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes #465

## Checklist

### Test notes

New `tests/integration/test_gateway_rejection_logging.py` has 13 tests: one per gate on both scaffolds, the two deliberate omissions, and a regression test pinning `users.reserved` back to 0 after the unresolvable-selector 400 (verified to fail, leaking 0.0102, with the refund removed). Each rejection test asserts a single error row with no cost and `counts_toward_budget=True`.

`ruff`, `mypy --strict` (273 files), `openapi-check`, `postman-check`, dashboard `typecheck` and its 361 tests all pass. Unit: 1034 passed. Integration: 832 passed, 2 failed. The two failures are `test_error_detail_leakage.py::test_provider_error_does_not_leak_details` and `test_streaming_error_event.py::test_streaming_creation_error_returns_http_error`; both fail identically on a clean `main` worktree in the same environment, which has outbound network and so gets a real provider 404 instead of the expected failure mode. Not related to this change.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

Implemented by Claude through back-and-forth with @njbrake. The decisions this PR had to make explicitly (whether 401 and 429 log, which of the two model forms to standardize on, and pinning `counts_toward_budget` at the writer instead of per gate) were discussed and settled before implementation, not chosen by the model alone.

## Summary

- Added activity logging for gateway rejections across chat, messages, responses, and pass-through routes.
- Recorded model access, user, budget, pricing, and provider resolution failures with resolved model and provider values.
- Excluded authentication failures, unknown users, and rate limits from logging.
- Preserved reservation refunds and ensured rejected requests have no cost and do not increase spend.
- Centralized rejection and usage-row creation.
- Updated dashboard documentation and assets.
- Added integration and unit test coverage for rejection logging and refund behavior.

## Technical notes

- Added `log_gateway_rejection` and `unresolvable_model_detail`.
- Updated `resolve_dispatch_provider` to accept a format adapter and refund reservations.


src/gateway/api/routes/_pipeline.py
```diff
@@ -640,16 +677,40 @@ async def resolve_request_context(
...
+                    log_writer=log_writer,
+                    api_key_id=api_key_id,
+                    user_id=api_key.user_id,
+                    model=model,
+                    provider=None,
+                    endpoint=adapter.endpoint,
+                    detail=user_forbidden_detail,
+                    started_at=started_at,
+                )
+            raise
```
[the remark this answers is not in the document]
njbrake: _Note: this reply was drafted by Claude via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

Valid, fixed in 594da822.

Confirmed the gap is real rather than theoretical: all ten call sites log and then re-raise, `SingleLogWriter.put` only absorbs `SQLAlchemyError` (session setup and teardown are outside its `try`), and `log_usage` builds the row before the writer is ever reached. So a writer problem would have surfaced to the client as a 500 in place of a clean 403 or 400, which is the wrong trade: observability should not change the response contract.

Guarded inside `log_gateway_rejection` rather than at each caller, since one helper covers every site. `except Exception` is deliberately broad here and is annotated as such; `asyncio.CancelledError` derives from `BaseException`, so a cancelled request still unwinds instead of being swallowed.

Worth noting for anyone reading later: nothing leaks when a row is dropped. Every rejection site refunds its reservation before calling this, never after, so a failed log costs only the log row. That ordering is now stated in the docstring so a future site does not invert it.

Pinned by `tests/unit/test_gateway_rejection_logging_best_effort.py`, with a control asserting a healthy writer still records the row, so "swallow failures" cannot quietly become "swallow everything". Narrowing the catch to `ValueError` fails that test.

tests/integration/test_gateway_rejection_logging.py
```diff
@@ -0,0 +1,332 @@
...
+    assert _chat(client, master_key_header, model="ghostprovider:some-model", user="stranded-user") == 400
+
+    session = db_session_factory()
+    try:
+        reserved = session.execute(
+            text("SELECT reserved FROM users WHERE user_id = 'stranded-user'")
+        ).scalar_one()
+    finally:
+        session.close()
+    assert float(reserved) == 0.0
```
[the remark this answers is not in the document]
njbrake: _Note: this reply was drafted by Claude via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

Good catch, fixed in 594da822. You are right that the assertion was one-sided: `reserved == 0.0` also holds when the estimate is 0, which is precisely the case where the refund guards nothing, so the test would have gone green while the regression walked past it.

Added the A/B control in the shape you suggested. The same priced selector is first sent under a budget smaller than the estimate (`max_budget=0.001`), which is refused at the budget gate with a 403 rather than reaching dispatch. That outcome is only reachable if the raw selector really did match the stored pricing row and really did produce a nonzero estimate, so it pins the precondition directly instead of inferring it. The main leg then keeps the existing `reserved == 0.0` check, scoped by user since the control now leaves a rejection row of its own.

Both legs are mutation-checked rather than assumed:

- removing `release_reservation` from `resolve_dispatch_provider` fails the main leg with `assert 0.010261875 == 0.0`
- forcing `estimate_cost` to return 0 fails the control with `assert 400 == 403`

That second one is the gap you identified, now closed.

src/gateway/api/routes/_pipeline.py
```diff
@@ -640,16 +677,40 @@ async def resolve_request_context(
...
+            # the key's own user, so that rejection has a user to attribute the
+            # drop to. resolve_user_id's other refusals (a master key with no
+            # `user` field, a key with no user) name no existing user, and
+            # usage_logs.user_id is a foreign key, so they stay unlogged.
+            # This row carries the raw selector and no provider, unlike the gates
+            # below it: nothing has been resolved this early, and resolving a
+            # selector purely to shape a log row is not worth the work on a path
+            # that is refusing the request anyway.
+            if exc.status_code == status.HTTP_403_FORBIDDEN and api_key is not None:
+                await log_gateway_rejection(
```
khaledosman, line 717: This log write happens before `check_rate_limit(...)` a few lines below, unlike every other gate this PR adds (allow-list, budget, unresolvable selector all sit after it). So a valid key can loop requests with a mismatched `user` field and append usage rows without ever being throttled — DB write amplification plus an inflated error count on its own key. Moving the write after the rate-limit check, or skipping it when the request would be throttled anyway, closes it.

src/gateway/api/routes/_passthrough.py
```diff
@@ -173,43 +186,102 @@ async def run_passthrough(
...
+    try:
+        user_id = resolve_passthrough_user_id(auth_result, user, reject_mismatch=config.reject_user_mismatch)
+    except HTTPException as exc:
+        # Only the user/key mismatch (403) has a user to attribute the drop to;
+        # see log_gateway_rejection for the rejections that deliberately do not
+        # log. Like its counterpart in the pipeline, this row carries the raw
+        # selector and no provider: nothing is resolved this early, and resolving
+        # purely to shape a log row is not worth it on a refusal path.
+        if exc.status_code == status.HTTP_403_FORBIDDEN and api_key is not None:
+            await log_gateway_rejection(
```
khaledosman, line 206: Same ordering as the pipeline counterpart: this runs before `check_rate_limit(raw_request, user_id)`, so the mismatch gate is unthrottled on both scaffolds.

src/gateway/api/routes/_pipeline.py
```diff
@@ -1199,6 +1289,81 @@ async def release_reservation(ctx: RequestContext) -> None:
...
+            db=db,
+            log_writer=log_writer,
+            api_key_id=api_key_id,
+            model=model,
+            provider=provider,
+            endpoint=endpoint,
+            user_id=user_id,
+            error=detail,
+            latency_ms=_elapsed_ms(started_at),
+            counts_toward_budget=True,
```
khaledosman, line 1402: Two notes on this writer:

1. `unpriced_requests` in the usage summary is `sum(case(cost IS NULL))`, rendered on the Usage page as "N unpriced" — its pricing-gap signal. Every row written here carries `cost=null`, so a budget incident now reads as a pricing misconfiguration. #449 introduced this for the 402 rows; this PR widens it to every gate. Scoping that count to `status == "success"` would fix both.
2. #470 adds `usage_logs.status_code` and stamps the 402 rows this function now owns. This signature needs a `status_code` parameter on whichever rebase lands second, or every gate here logs NULL and classifies as `unknown`.

src/gateway/api/routes/_pipeline.py
```diff
@@ -500,6 +515,28 @@ def resolve_dispatch_provider(ctx: RequestContext, config: GatewayConfig, model_
...
+        # already held: refund it, then record the drop. The hold is not always
+        # zero, so this refund is load-bearing rather than a formality. The
+        # preamble carries an unresolvable selector into the pricing lookup as
+        # the bare model with no provider, and find_model_pricing then keys on
+        # the model alone, which is exactly the `provider:model` form stored
+        # pricing rows use. An instance removed from config while its pricing row
+        # survives therefore prices, reserves a real estimate, and only fails
+        # here; before this refund existed the hold stayed on users.reserved
+        # until the next budget reset (forever, for a budget with no period).
+        await release_reservation(ctx)
```
khaledosman, line 536: `refund_reservation` is not idempotent — `_release_reserved` clamps at 0, but a second call still decrements again. This release-then-raise is safe only because nothing above catches the 400 and releases too; I checked `chat.py`, `messages.py` and `responses.py` and that holds today. Worth a one-line comment here, since the invariant is invisible from this call site and the next person adding an outer handler will not know to look.

khaledosman (approved):
Two things worth resolving before this merges, both inline: the user/key-mismatch gate writes its row **before** `check_rate_limit`, making it the one logged rejection a caller can spam unthrottled; and `cost=null` rejection rows feed `unpriced_requests`, which the Usage page renders as a pricing-gap signal.

Also flagged: the `status_code` coordination with #470 (that PR stamps the very rows this one now routes through `log_gateway_rejection`), and the non-idempotent-refund invariant the new `release_reservation` call depends on. The A/B structure in `test_unresolvable_selector_releases_the_reservation` is the right way to write that test — without the tiny-budget control it would pass while guarding nothing.

---

PR #473, opened 2026-08-02
feat(search): add a billed POST /v1/search pass-through

### LINKED ISSUE #400: Search passthrough endpoint (POST /v1/search)

## Context

Otari exposes web search only as an in-loop tool (`otari_web_search`), run inside the agentic/tool loop. There is no standalone, billed, auth-gated search endpoint for callers that want a direct search request (for example proxying to Exa) with the same key management and usage tracking as chat completions.

## Proposal (backend)

- Add `POST /v1/search` (and/or `POST /v1/search/{tool}`) registered in standalone mode alongside chat/messages/responses.
- Run it through the same auth, budget, and usage-logging path as completions; record a `UsageLog` row.
- Declare available search providers/tools in `config.yml`, mirroring model/provider declarations; store any secrets via the existing `secret_box` / provider-credentials machinery.

## Web dashboard

Lightest dashboard touch of the batch, and partly optional:

- If search providers use API-key credentials, surface them so operators can configure a search provider without editing `config.yml`. The natural homes are the existing `ProvidersPage.tsx` (provider credentials CRUD via `/v1/provider-credentials`) or `ToolsGuardrailsPage.tsx` (tool configuration). Pick one; avoid a new nav entry unless the config is substantial.
- Search usage should appear in `ActivityPage.tsx` / `UsagePage.tsx` automatically if the endpoint writes `UsageLog` rows, but confirm the Activity endpoint filter includes the new route and that a search row (no completion tokens) renders sensibly.
- If any UI changes land, update the relevant `*.test.tsx`, rebuild, and commit the bundle.

## Acceptance criteria

- An authenticated key can call `POST /v1/search`; the call is auth-gated, budget-checked, and usage-logged.
- At least one provider (Exa) is wired end to end with credentials from config or the dashboard.
- Search requests are visible in the Activity/Usage views.
- OpenAPI regenerated.

## Open questions

- Match LiteLLM's `/v1/search` request/response shape for drop-in compatibility, or an otari-native shape?
- Does search spend reconcile against the same budget as completions, or a separate line item?

## Background

Surfaced while scoping a migration of an auth/policy sidecar (Mozilla's MLPA) from the LiteLLM proxy onto otari. The sidecar proxies an Exa search endpoint through the gateway today, so an equivalent otari surface is needed to keep search on the single-gateway billing path.

njbrake:
## Description

Otari exposed web search only as an in-loop tool (`otari_web_search`), so a caller that wants a direct search request had to hold its own provider key and fell off the gateway's billing path entirely. That is the gap blocking MLPA's move from the LiteLLM proxy onto otari: it proxies an Exa search endpoint today and needs an equivalent surface here to keep search on one gateway.

This adds `POST /v1/search` and `POST /v1/search/{search_tool_name}`, registered standalone-only alongside the other pass-throughs. Both run the same scaffold as a completion (auth, rate limit, budget reservation, usage log, reconcile or refund) and both log `endpoint="/v1/search"`, so one Activity filter covers every search. Exa is wired end to end.

**Configuration.** Search tools are declared under `search_tools:` in `config.yml`, keyed by the name callers pass as `search_tool_name` (or in the path), and validated at startup so an unsupported provider or a missing `api_key` fails before the first request:

```yaml
search_tools:
  exa-search:
    provider: exa
    api_key: ${EXA_API_KEY}
    options:
      type: fast
```

`options` is the escape hatch for provider-native knobs the request body has no field for. Request-derived fields win over it, and `query` always comes from the caller.

**Wire shape** (issue open question 1): LiteLLM's `/v1/search`, itself shaped after Perplexity's Search API, so a client migrating off that proxy needs no request changes. Translation to and from Exa's native shape lives in `search_backend.py`; the route never sees a provider payload.

**Billing** (issue open question 2): the same budget as completions, one `UsageLog` row, no separate line item. Search bills per request rather than per token, so the row carries zero tokens and a cost taken from Exa's own reported `costDollars.total` when present, falling back to a flat per-request rate configured for `<provider>:<tool>` under the same convention moderations uses. Since the provider meters itself, search is exempt from `require_pricing`, like moderations and audio.

`find_model_pricing` grows a `use_defaults` flag that the route sets to `False`. The genai-prices fallback also matches on a bare name, so a search tool an operator happened to name after a real model would otherwise inherit that model's per-million-token rate and be billed under a per-request convention.

**Dashboard.** No change was needed for search to show up: the Activity endpoint filter is a free-form pass-through (`web/src/api/hooks.ts:653`) rather than an allowlist, and a zero-token row renders as it should (`formatTokens(0)` gives `"0"`, `formatCost(0.007)` gives `"$0.0070"`). Credentials come from `config.yml` only; dashboard CRUD for search-tool credentials is left for a follow-up, which the issue allows ("credentials from config or the dashboard").

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [x] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes #400

## Checklist

Local results: 1059 unit passed / 1 skipped, 839 integration passed / 9 skipped, `ruff` and `mypy --strict` clean, `make openapi-check` and `make postman-check` clean. Note for anyone reproducing locally: `test_error_detail_leakage` and `test_streaming_error_event` fail if `OPENAI_API_KEY` is set in the environment, because they then reach the live API instead of failing locally. Both pass with it unset, which is how CI runs.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

Implemented by Claude through back-and-forth with @njbrake. The design decisions (LiteLLM-compatible wire shape, shared budget, config-only credentials for now) are his; the code and prose are Claude's.

### DIFF
--- a/src/gateway/api/routes/search.py
+++ b/src/gateway/api/routes/search.py
@@ -8,9 +8,13 @@ so a client that proxies a search API can keep one gateway on the billing path
 instead of holding a second provider key of its own.
 
 The request and response follow LiteLLM's ``/v1/search`` (itself shaped after
-Perplexity's Search API), so a caller moving off the LiteLLM proxy needs no
-changes; :mod:`gateway.services.search_backend` translates to and from the
-provider's native shape and owns the per-provider adapters.
+Perplexity's Search API), so a caller moving off the LiteLLM proxy keeps its
+request shape; :mod:`gateway.services.search_backend` translates to and from the
+provider's native shape and owns the per-provider adapters. Two Perplexity
+features are deliberately not modeled: ``query`` is a single string rather than
+the multi-query array, and the filters with no field below (recency, context
+size, published-date) are ignored rather than rejected. Both are called out in
+``docs/api-reference.md`` so a migrating caller can check for them.
 
 Both the body-selected (``POST /v1/search``) and path-selected
 (``POST /v1/search/{search_tool_name}``) forms log ``endpoint="/v1/search"``,
@@ -48,6 +52,7 @@ from gateway.models.entities import APIKey, UsageLog
 from gateway.rate_limit import check_rate_limit
 from gateway.services.budget_service import reconcile_reservation, refund_reservation, reserve_budget
 from gateway.services.log_writer import LogWriter
+from gateway.services.model_access import is_model_allowed, model_not_allowed_detail, resolve_request_allowlist
 from gateway.services.pricing_service import find_model_pricing, flat_request_cost
 from gateway.services.search_backend import (
     MAX_RESULTS_CAP,
@@ -91,7 +96,12 @@ class SearchRequest(BaseModel):
         max_length=_MAX_DOMAIN_FILTERS,
         description="Restrict results to these domains; prefix a domain with '-' to exclude it instead",
     )
-    country: str | None = Field(default=None, description="Two-letter ISO country code to localize results to")
+    country: str | None = Field(
+        default=None,
+        min_length=2,
+        max_length=2,
+        description="Two-letter ISO country code to localize results to",
+    )
     max_tokens_per_page: int | None = Field(
         default=None, ge=1, description="Approximate cap on the page content returned per result"
     )
@@ -217,6 +227,18 @@ async def _dispatch_search(
         raise HTTPException(status_code=status.HTTP_400_BAD_REQUEST, detail=str(exc)) from exc
 
     pricing_key = f"{tool.provider}:{tool.name}"
+
+    # Model access control (per-key), keyed on the same <provider>:<tool> string
+    # pricing uses. A key with no list of its own inherits its user's default, so
+    # an unrestricted key is unaffected; a restricted one must name the search
+    # tool, which is the fail-closed side to be on for a brand-new spend surface.
+    key_allowlist = await resolve_request_allowlist(db, api_key)
+    if key_allowlist is not None and not is_model_allowed(key_allowlist, pricing_key):
+        raise HTTPException(
+            status_code=status.HTTP_403_FORBIDDEN,
+            detail=model_not_allowed_detail(pricing_key),
+        )
+
     # A search tool is not a model, so the community default-pricing dataset can
     # only produce a false match on the tool's name.
     pricing = await find_model_pricing(db, tool.provider, tool.name, use_defaults=False)
@@ -224,7 +246,13 @@ async def _dispatch_search(
         db,
         user_id,
         flat_request_cost(pricing),
-        model=pricing_key,
+        # Deliberately not the pricing key: ``model`` exists only to drive
+        # reserve_budget's free-model shortcut, which splits the string through
+        # any-llm. A search tool is not an any-llm model, so passing it logs a
+        # warning on every request, and for a tool whose name did happen to be an
+        # any-llm provider the shortcut would run an unguarded default-pricing
+        # lookup on the tool name and could skip the reservation outright.
+        model=None,
         strategy=config.budget_strategy,
         counts_toward_budget=not budget_exempt,
     )
--- a/src/gateway/api/routes/search.py
+++ b/src/gateway/api/routes/search.py
@@ -21,6 +21,11 @@ Both the body-selected (``POST /v1/search``) and path-selected
 so one Activity filter covers every search regardless of how the tool was
 named.
 
+A request the gateway itself turns away (an unknown or ambiguous tool name, a
+tool the caller's key may not use) writes a usage row too, with a null cost:
+refused traffic is dropped traffic, and it should be visible in Activity and
+countable as a failure rather than invisible outside the caller's own logs.
+
 Billing: search is priced per request, not per token, so a usage row carries
 zero tokens and a cost taken from the provider's own reported charge when it
 reports one. Otherwise the cost is the flat per-request rate configured for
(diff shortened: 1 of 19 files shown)


njbrake:
@coderabbitai review

njbrake:
@coderabbitai full review

src/gateway/api/routes/search.py
```diff
@@ -0,0 +1,327 @@
...
+
+    # Model access control (per-key), keyed on the same <provider>:<tool> string
+    # pricing uses. A key with no list of its own inherits its user's default, so
+    # an unrestricted key is unaffected; a restricted one must name the search
+    # tool, which is the fail-closed side to be on for a brand-new spend surface.
+    key_allowlist = await resolve_request_allowlist(db, api_key)
+    if key_allowlist is not None and not is_model_allowed(key_allowlist, pricing_key):
+        raise HTTPException(
+            status_code=status.HTTP_403_FORBIDDEN,
+            detail=model_not_allowed_detail(pricing_key),
```
khaledosman: This 403 (and the unknown-tool 400 above) raises without writing a usage row, which puts search on the wrong side of the invariant #472 is establishing for every other billed surface: a gateway-side refusal should be visible in Activity and countable by the failure tile. The success/error rows also carry no `status_code` (#470). Both are the kind of thing that never gets backfilled once an endpoint ships — worth doing here, or filing before merge.

src/gateway/services/search_backend.py
```diff
@@ -0,0 +1,301 @@
...
+
+    if query.country:
+        payload["userLocation"] = query.country
+
+    # Exa returns page text only when asked for it, and the response's
+    # ``snippet`` is that text, so contents are always requested. A tool that
+    # pins its own ``contents.text`` keeps it unless the caller asked for a
+    # specific per-page size.
+    contents = dict(payload.get("contents") or {})
+    if query.max_tokens_per_page is not None or "text" not in contents:
```
khaledosman: There is no way to turn page content off. When the caller sends no `max_tokens_per_page`, this still forces `contents.text` whenever the tool has not pinned it, and an operator pinning `contents: {highlights: true}` gets text injected anyway — so every search pays Exa's per-page content charge even when the caller only wants ranked URLs. Since cost comes back from the provider the billing stays truthful, it is just needlessly higher. Letting an explicit `contents.text: null`/`false` in `options` suppress it would cover the snippet-only case.

src/gateway/api/routes/search.py
```diff
@@ -0,0 +1,327 @@
...
+        )
+
+    # A search tool is not a model, so the community default-pricing dataset can
+    # only produce a false match on the tool's name.
+    pricing = await find_model_pricing(db, tool.provider, tool.name, use_defaults=False)
+    reservation = await reserve_budget(
+        db,
+        user_id,
+        flat_request_cost(pricing),
```
khaledosman, line 287: With no `<provider>:<tool>` rate configured this is 0.0, so nothing is reserved and the real cost only lands at `reconcile_reservation`. A user already over the cap is still refused (`reserve_budget`'s `spend + reserved < max_budget` catches zero-estimate requests), but one near the cap can overshoot by a search's cost, and concurrent searches cannot see each other's holds. Worth saying in `docs/configuration.md` that a flat rate is recommended rather than optional, or reserving a small default.

src/gateway/core/config.py
```diff
@@ -732,6 +748,52 @@ def validate_provider_instances(self) -> None:
...
+                    f"search_tools.{name}.provider '{provider}' is not a supported search provider "
+                    f"(one of: {', '.join(SEARCH_PROVIDERS)})."
+                )
+                raise ValueError(msg)
+            if not entry.get("api_key"):
+                msg = f"search_tools.{name}.api_key is required."
+                raise ValueError(msg)
+            timeout = entry.get("timeout")
+            if timeout is not None and (isinstance(timeout, bool) or not isinstance(timeout, (int, float))):
+                msg = f"search_tools.{name}.timeout must be a number of seconds."
```
khaledosman: A negative `timeout` passes this check and only blows up inside httpx at request time, and `timeout: 0` is silently replaced by the 30s default (the `or` in `resolve_search_tool`). Since the point of this function is failing before the first request, reject non-positive values here too.

src/gateway/services/search_backend.py
```diff
@@ -0,0 +1,301 @@
...
+    payload["contents"] = contents
+
+    payload["query"] = query.query
+    return payload
+
+
+async def _search_exa(tool: SearchTool, query: SearchQuery) -> SearchOutcome:
+    payload = build_exa_payload(tool, query)
+    try:
+        async with httpx.AsyncClient(timeout=tool.timeout_s) as client:
```
khaledosman: Fresh client (so fresh TCP + TLS) per search. `guardrails.py` does the same, but `web_search_backend.py` keeps a pooled client on an `AsyncExitStack` — for the sustained search traffic this endpoint exists to serve, the pooled shape is the one to copy.

khaledosman (approved):
The wire-shape and pricing decisions look right, and the two traps you avoided are the non-obvious ones: `use_defaults=False` on `find_model_pricing` and `model=None` into `reserve_budget` (with a test pinning the warning, which is how that stays fixed).

Inline: search rejections write no usage row (which contradicts #472 landing alongside), Exa page content cannot be opted out of, the zero-reservation case when no flat rate is configured, and a `timeout` validation gap.

---

PR #474, opened 2026-08-02
feat(dashboard): browsable activity log and rebuilt usage analytics

### LINKED ISSUE #427: Activity page works as a monitor but is not browsable

Related: #304 (built the log viewer), #312 (UX polish), #299 (dashboard epic).

## Summary

Activity works well as a *monitor* — it answers "are requests flowing, and is anything failing." It does not work as a *browser*: there is no good way to explore or navigate the request log to find a specific request or an interesting one. Today it is `Time | User | Model | API key | Tokens | Cost | Total time | Status`, filterable by time preset / status / model / user / API key, click-to-expand for detail, 50 rows per offset page.

The reasons are specific and mostly cheap to fix.

## Why it isn't browsable

**1. It's a flat firehose, and rows are near-identical.** A measured Claude Code workload of 3,551 requests across 50 sessions is ~71 pages where consecutive rows share the same user, model, and API key. The fields that actually vary are the ones not shown as columns.

**2. `Tokens` shows `total_tokens`, which for cached workloads is ~98% cache read.** Every row displays a large, similar-looking number that carries almost no discriminating information — the widest column and the least useful one.

**3. Session is a dead end.** `source_label` is not a filter on `/v1/usage` (see `list_usage`'s signature), not a column, and appears only inside the expand panel. The UI shows you a session id and then offers no way to act on it — you cannot say "show me the rest of this session."

**4. Fixed sort, `timestamp desc` only.** No sort by cost, tokens, or latency. The most common browsing intent — find the expensive or slow outlier — is unsupported, leaving "filter a window and read 71 pages" as the only path.

**5. Refresh throws away your position.** `refresh()` re-anchors `startDate` to `isoAgo(rangeSeconds)`, which changes `filters`, which trips `useEffect(() => setPage(0), [filters])`. Press refresh on page 12 and you land on page 1. Combined with offset pagination over a table that inserts at the head, deep browsing is unreliable as well as tedious.

**6. No `source` filter, though the API supports one.** Five filter controls exist and `source` isn't among them, so imported agent traffic can't be separated from gateway traffic while browsing — even though the page already defines friendly source labels (`SOURCE_LABELS`).

**7. No free-text search, and the id shown is the wrong one.** `RequestDetail` shows "Request ID" = `entry.id`, otari's internal uuid — not `source_event_id`. An operator holding a provider or agent request id has no way to look it up. Cross-system traceability breaks at the last inch.

**8. Error browsing terminates in "read the server logs."** The status filter finds error rows, but each must be expanded individually and the panel renders a fixed string ("The provider returned an error. Inspect gateway logs for details."). Browsing *why* things failed is impossible by construction.

## Proposed fixes, roughly in value order

| # | change | scope |
|---|---|---|
| A | **Sessions as the entry point** — a sessions list (cost / tokens / requests / duration / model mix) drilling into that session's requests as a timeline. Make the per-request log the leaf, not the front door. | server + UI |
| B | **`source_label` as filter + column**, with the session id in the detail panel linking to a filtered view. Needs `source_label` in `_usage_filters`. | small server + UI |
| C | **Sortable columns** (cost, tokens, latency) via an `order_by` param. Converts browsing from paging into querying. | small server + UI |
| D | **Don't reset the page on refresh**, and move to keyset pagination on `(timestamp, id)` so rows stop shifting. | UI + small server |
| E | **Add the `source` filter control.** API already supports it — pure UI. | UI only |
| F | **Show `source_event_id` and make it searchable.** | small server + UI |
| G | **Token composition per row** — a thin stacked bar (fresh / output / cache read / cache write) instead of one total, so rows become visually distinguishable. | UI only |
| H | **Error grouping** — group by normalized error with counts and first/last seen, so "23 rate-limit errors in this session" is one row instead of 23 expansions. (Depends on a status-code column — see the related error-taxonomy issue.) | server + UI |
| I | **Brush-to-filter volume sparkline** above the table, plus keyboard row navigation. | UI only |

E, G, and the first half of D are UI-only and would improve the page immediately. A is the structural fix.

njbrake:

## Description

Two related dashboard overhauls, plus the backend they need. The design follows what the mainstream usage dashboards (OpenAI usage page, Anthropic console, Grafana-family tooling) converged on: one main time-series chart with a metric selector and a group-by dimension, token counts split into their billed composition, and drag-on-the-chart time selection.

### Activity log: browsable, not just watchable

- Source filter (shown only when the window holds more than one source), a token composition bar per row (fresh input / cache read / cache write / output), and the billed token total in the request detail panel.
- Refresh keeps your place: it no longer re-anchors the rolling window, so deep pagination survives; bookmarked `?page=` URLs open on the right page; re-picking the active preset is the explicit re-anchor gesture.
- Histogram rebuilt on the standard interaction: drag across the plot to zoom (recharts event coordinates, replacing the edge-handle slider and pan strip and their pixel math), a minimap pan rail when zoomed (pointer and keyboard), and failed requests as a red segment right on the strip.

### Usage page: metric by dimension analytics

- One main chart with metric tabs (Cost, Tokens, Requests) and a Group by select (model, user, API key, source): stacked bars over the top 8 groups plus a reconciling Other fold, on a fixed categorical palette validated for CVD separation and surface contrast.
- Ungrouped Tokens stacks the billed composition, the same encoding as the Activity token bar; ungrouped Requests splits succeeded/failed.
- Dragging across the chart zooms the window, matching the Activity strip.
- Tiles: a cache hit rate (reads over billed input) replaces the three raw cache counters; the token tile is the billed total.
- Two breakdown cards: spend by model / user beside spend by session / endpoint / provider / source (session first: it usually names the work behind a bill). There is deliberately no API-key spend table; the chart's group-by can still split by key. The duplicate per-request table is gone; the Activity page owns that view and its bulk actions.

### Backend

- `/v1/usage/summary` series points now carry per-bucket error counts and the billed token composition, normalized through each row's `billing_meters` (with the same raw-column fallback the dashboard uses); totals carry `billed_input_tokens`.
- New `GET /v1/usage/series?group_by=model|user_id|api_key_id|source`: a per-group time series for the stacked charts, folding past-top-8 groups in SQL so the payload stays bounded.
- Breakdown and CSV token counts are billed (input including cache, plus output) so every analytics surface agrees on what a token count means.

**Compatibility note:** the `tokens` field in `/v1/usage/summary` breakdowns and in the `/v1/usage/summary.csv` export changes meaning from the raw provider-reported total to the billed total (fresh input, both cache buckets, and output). For additive-convention providers (Anthropic) the numbers step up accordingly. Both endpoints are recent, master-key-gated analytics surfaces, but downstream consumers of the CSV should be told.

Tested on SQLite and PostgreSQL; OpenAPI spec and Postman collection regenerated; dashboard bundle rebuilt and committed.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

N/A

## Checklist

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

Claude Code (design discussion and drafting), kimi (PR template alignment)

The implementation was driven by Claude through several rounds of discussion with @njbrake; the design decisions and final review are his.

**NOTE:**

🤖 Generated by kimi

## Summary

- Rebuilt the Activity page with source filters, billed-token details, failed-request charts, drag-to-zoom selection, navigation controls, and refresh-safe pagination.
- Rebuilt the Usage page with cost, token, and request metrics; grouped charts; cache hit rates; token composition; and breakdown tables.
- Removed the per-request table from Usage and moved request-level investigation to Activity.
- Added `/v1/usage/series` with grouping, filtering, bucket limits, error counts, and billed-token metrics.
- Standardized billed-token totals across summaries, breakdowns, and CSV exports.
- Updated API types, tests, documentation, OpenAPI-related assets, and the generated dashboard bundle.

## Benefits

Users can analyze usage trends more clearly, compare sources and groups, and investigate failed or costly requests from one activity view. Billed usage now remains consistent across dashboard views and exports.


src/gateway/api/routes/usage.py
```diff
@@ -911,6 +1025,108 @@ async def usage_summary(
...
+    "user_id": UsageLog.user_id,
+    "api_key_id": UsageLog.api_key_id,
+    "source": UsageLog.source,
+}
+
+
+@router.get("/series", dependencies=[Depends(verify_master_key)])
+async def usage_series(
+    db: Annotated[AsyncSession, Depends(get_db)],
+    group_by: SeriesGroupBy = Query(description="Dimension to split the series by"),
```
khaledosman, line 1046: **`/series` silently drops three filters `/summary` accepts.** The signature omits `provider`, `source_label`, and `status_code`, but the docstring below claims "Same filters and window bounds as `/summary`". FastAPI ignores unknown query params, so nothing errors.

The frontend already makes this reachable: `useUsageGroupedSeries` serializes the whole `UsageFilters` object (`web/src/api/hooks.ts:816`), and `UsageFilters` carries `provider` and `source_label` (`web/src/api/types.ts:435,441`). The moment UsagePage gains a provider or session filter — Activity already has both, and the drill-down flows point that way — the stacked chart aggregates over a wider set than the tiles beside it, with no error to signal it.

Either add the three params here, or narrow the docstring and have `useUsageGroupedSeries` take an explicit filter subset rather than the full `UsageFilters`.

web/src/components/charts.tsx
```diff
@@ -1,77 +1,277 @@
...
+          {showDimming && dimmed && dimmed.endIndex < last ? (
+            <ReferenceArea
+              x1={data[dimmed.endIndex + 1].x}
+              x2={data[data.length - 1].x}
+              fill="var(--otari-bg)"
+              fillOpacity={0.75}
+              stroke="none"
+            />
+          ) : null}
+          {drag && drag.start !== drag.end ? (
```
khaledosman: **Unguarded index into `data` during a drag.** `data[Math.min(drag.start, drag.end)].x` dereferences indices captured from a previous render's `activeTooltipIndex`. A background refetch landing mid-drag with a shorter series throws.

The `window` prop is clamped ~90 lines up for exactly this reason ("an out-of-range index would blank the chart instead of degrading it"). The drag path should reuse the same `clampIndex`.

src/gateway/api/routes/usage.py
```diff
@@ -625,17 +719,21 @@ async def _breakdown(
...
+    ``limit=None`` returns every group (used by the CSV export, which must not
+    truncate).
     """
     cost_sum = func.coalesce(func.sum(UsageLog.cost), 0.0)
     stmt = (
         select(
             column,
             cost_sum,
-            func.coalesce(func.sum(UsageLog.total_tokens), 0),
+            _billed_input_sum() + _billed_output_sum(),
```
khaledosman, line 743: **`tokens` changes meaning on two existing surfaces.** Breakdown rows (`by_model`, `by_user`, …) and the CSV export (see the `usage_summary_csv` docstring) switch from `sum(total_tokens)` to billed. For an additive-convention provider that is a silent step up in every existing consumer's numbers, including the finance CSV.

The PR body frames this as an internal consistency fix, but it is a contract change on `/v1/usage/summary` and `/v1/usage/summary.csv`. Since git-cliff parses the squash title, `feat(dashboard):` files it under features with no compat note — worth a `!` / `BREAKING CHANGE` footer, or a new field name alongside the old one.

web/src/pages/UsagePage.tsx
```diff
@@ -639,46 +447,238 @@ export function UsagePage() {
...
+      : billedInput !== undefined
+        ? billedInput + (totals.billed_output_tokens ?? totals.completion_tokens)
+        : totals.total_tokens;
+  const prevBilledTotal =
+    prevTotals === undefined
+      ? null
+      : prevTotals.billed_input_tokens !== undefined
+        ? prevTotals.billed_input_tokens + (prevTotals.billed_output_tokens ?? prevTotals.completion_tokens)
+        : prevTotals.total_tokens;
+  const cacheHitRate = billedInput !== undefined && billedInput > 0 && totals ? totals.cache_read_tokens / billedInput : null;
```
khaledosman: **Cache hit rate mixes a raw numerator with a normalized denominator.** `totals.cache_read_tokens` is the raw column sum (`_totals` row[4], unchanged by this PR), while `billedInput` is meter-normalized. The tile's own sparkline further down uses the normalized `p.cache_read_tokens` from the series, so the headline number and its trendline are computed differently.

The divergence is small in practice (`billable_usage` only clamps `cache_read` under the subset convention), but `UsageTotals.billed_input_tokens`' own docstring justifies the field precisely by convention-independence. Adding a `_billed_expr`-normalized cache_read/write to `_totals` and reading those here would make the tile internally consistent.

src/gateway/api/routes/usage.py
```diff
@@ -911,6 +1025,108 @@ async def usage_summary(
...
+    # value never matches ``IN``, so it lands in the CASE's ``else`` arm; the
+    # fold flag then separates a NULL group that ranked in the top N (a real
+    # ``key=None`` series, e.g. a deleted user) from the past-top-N remainder.
+    named = {g.key for g in groups if g.key is not None}
+    keeps_null = any(g.key is None and not g.is_other for g in groups)
+    key_expr = case((column.in_(named), column), else_=null())
+    if keeps_null:
+        fold_expr = case((column.is_(None), 0), (column.in_(named), 0), else_=1)
+    else:
+        fold_expr = case((column.in_(named), 0), else_=1)
```
khaledosman, line 1117: **No point cap on `/series`.** `/summary` bounds its series with `_MAX_SERIES_POINTS = 1000`; this has no analogue. `bucket=hour` over the 366-day max window yields up to 8784 × 10 ≈ 88k points (~10MB JSON) from a single master-key call.

The comment above says the row count "stays bounded by buckets × (top N + 2)" — that bounds cardinality, not time. Either cap buckets the way `/summary` does, or reject `bucket=hour` past some span.

src/gateway/api/routes/usage.py
```diff
@@ -578,6 +642,32 @@ def _dialect_name(db: AsyncSession) -> str:
...
+    column sums cannot be split into a billed composition. The pricing writers
+    resolve that into ``billing_meters`` when a row is priced; prefer that, and
+    fall back to the raw column under the subset convention for meterless rows,
+    the same fallback the dashboard's per-row token bar applies. JSON extraction
+    of a missing key and a NULL column both yield NULL, so the fallback chain
+    covers both, ending at 0. ``as_integer`` compiles to the dialect's JSON
+    number cast on SQLite and PostgreSQL alike.
+    """
+    return func.coalesce(UsageLog.billing_meters[meter].as_integer(), fallback, 0)
+
```
khaledosman, line 666: **Per-row JSON extraction on a Postgres `json` (not `jsonb`) column.** `billing_meters` is `mapped_column(JSON)` (`models/entities.py:404`), which compiles to `json` on Postgres, where every `->>` reparses the row's text.

A single Usage-page render now issues: totals (2 extracts/row), the series (4/row), and 6 breakdown passes (2/row each) — each a separate full scan of the window, now with JSON parsing on top of what used to be plain column sums. On a 366-day window that is a real multiplier.

Worth measuring before ship. If it bites, `jsonb` plus a migration, or persisting the two billed totals as columns at write time, are the escapes.

web/src/pages/UsagePage.tsx
```diff
@@ -52,50 +53,118 @@ function formatLatency(ms: number | null): string {
...
-  onDrill: (key: string) => void;
-  loading: boolean;
   // How a real group whose column was NULL reads. The default suits an id that
   // has gone missing (a deleted user); a dimension where NULL is a normal state
   // (gateway rows carry no session label) passes its own wording.
   unknownLabel?: string;
-  // Optional control rendered beside the heading (the dimension picker).
-  action?: ReactNode;
+  // Maps a row key to a human label (e.g. API key id -> key name).
+  labelFor?: (key: string) => string;
```
khaledosman: `labelFor` is declared here and on `BreakdownDimensionDef`, but no dimension ever sets it — `activePrimary.labelFor` / `activeSecondary.labelFor` are always `undefined`.

Related: the PR description claims breakdown tables are "now including API key and source, with key-name labels". There is no API-key table (the comment above `PAGE_BREAKDOWNS` explicitly rejects one) and no key-name labelling outside the chart legend. Either wire `labelFor` up or drop it, and fix the description either way.

web/src/pages/UsagePage.tsx
```diff
@@ -169,284 +235,34 @@ function BreakdownTable({
...
-];
+// ---------- which breakdowns the page asks for ----------
 
-// The breakdowns this page renders, and only those: the two fixed tables plus
-// every dimension the picker can switch to. All four picker dimensions ship in one
-// response so flipping tabs is instant instead of another round trip, while
-// `api_key` stays out because no table here shows it. Derived from the tab list, so
-// a new tab is requested automatically.
-const PAGE_BREAKDOWNS: SummaryDimension[] = ["model", "user", ...DIMENSION_TABS.map((tab) => tab.key)];
+// Every dimension a tab below can show. Requested explicitly so a new
```
khaledosman: Two stacked comment paragraphs describing the same constant; this first one reads as a leftover from an earlier revision (it talks about "every dimension a tab below can show", which the paragraph below then contradicts with "and nothing more").

web/src/components/charts.tsx
```diff
@@ -1,77 +1,277 @@
...
+  const dimmed =
+    windowRange && data.length > 0
+      ? { startIndex: clampIndex(windowRange.startIndex), endIndex: clampIndex(windowRange.endIndex) }
+      : null;
+  const showDimming = dimmed !== null && (dimmed.startIndex > 0 || dimmed.endIndex < last);
+
   return (
-    <div role="img" aria-label={ariaLabel} className="w-full">
+    <div
+      role="img"
```
khaledosman: `role="img"` on a container that now owns press-and-drag selection tells AT the whole subtree is a single image. It was defensible on a static chart.

The Activity strip keeps a keyboard path via the pan rail plus −/+, though those only reach power-of-two spans; the Usage chart has no keyboard equivalent for zoom at all.

web/src/components/charts.tsx
```diff
@@ -1,77 +1,277 @@
...
+          onMouseLeave={commit}
+        >
           <XAxis
-            dataKey="label"
+            dataKey="x"
             tickLine={false}
             axisLine={false}
             interval="preserveStartEnd"
-            minTickGap={24}
+            minTickGap={40}
```
khaledosman, line 242: `touch-pan-y select-none` is set, but only `onMouseDown`/`onMouseMove`/`onMouseUp` are wired, so drag-to-zoom is pointer-only. On the PWA the Usage chart loses its only zoom affordance (the Activity strip still has the −/+ buttons).

web/src/pages/ActivityPage.tsx
```diff
@@ -576,21 +738,41 @@ export function ActivityPage() {
     );
   };
 
-  // Refetch every window-scoped query (see the UsagePage refresh note): the
-  // model typeahead summary would otherwise stay on cached data whenever the
-  // re-anchored window resolves to the same query key.
+  // Refresh means "same view, newer rows", so it deliberately does *not* re-anchor
```
khaledosman, line 744: Not re-anchoring on refresh is the right call for pagination, but `start` stays an `isoAgo()` snapshot taken at mount. Over a long session a "Last 24 hours" view quietly becomes 24h-plus-session-length, and the caption keeps saying 24 hours.

The re-pick gesture is documented in code, but nothing in the UI hints the window has drifted.

web/src/pages/ActivityPage.tsx
```diff
@@ -369,13 +500,41 @@ export function ActivityPage() {
...
 
+  // Source options: the sources with usage in the window. Like the model
+  // suggestions, this must ignore the source filter itself, or picking Claude Code
+  // would hide every other source and switching would mean clearing first.
+  //
+  // While no source is picked the model-suggestion summary is already computed
+  // without one, so its provenance breakdown is that full list and no second query
+  // is needed. Only a picked source needs its own, so it is fetched then and only
+  // then (each summary runs four grouped aggregations plus the series).
+  const sourceSuggestFilters: UsageFilters = useMemo(
```
khaledosman, line 521: `sourceSuggestFilters` omits `sessionFilter` / `endpointFilter` / `providerFilter`, which `modelSuggestFilters` includes. So the source list is scoped differently depending on which of the two summaries feeds it (drill-down chips active or not).

Probably intentional (a broader list is friendlier), but worth either commenting as deliberately asymmetric or making the two match.

khaledosman (approved):
Backend logic holds up: the `other` fold reconciles by construction, the `key`/`is_other` NULL encoding avoids sentinel collisions, and the `keeps_null` branch correctly separates a real NULL group from the past-top-N remainder. Test coverage on both sides is unusually good for a change this size.

Nothing blocking. Inline findings, roughly in priority order:

- `/series` silently drops `provider`, `source_label`, and `status_code` while its docstring claims filter parity with `/summary` — latent divergence between the stacked chart and the tiles beside it.
- `TrendChart`'s drag path dereferences `data[index]` unguarded, while the `window` prop is explicitly clamped against exactly that race.
- The breakdown and CSV `tokens` columns change meaning (raw -> billed); worth a compat note since git-cliff will file this under `feat`.
- `/series` has no `_MAX_SERIES_POINTS` analogue.
- Per-row JSON extraction on a Postgres `json` (not `jsonb`) column, multiplied across totals + series + 6 breakdown passes.

Plus a few small ones: a dead `labelFor`, a stale duplicated comment, `role="img"` on a now-interactive chart, and mouse-only drag on a PWA surface.

---
🤖 Review by [Claude Code](https://claude.com/claude-code)

PR #475, opened 2026-08-03
feat(lint): enforce gateway layer rules via check_architecture.py

peteski22:
## Description

Adds an AST-based architecture check (`scripts/check_architecture.py`) over `src/gateway/` and wires it into `make lint` via a new `check-architecture` target, so the existing lint workflow gates every PR with no workflow changes.

Enforced layer rules (all hold on the current tree):

- `services` must not import the API layer.
- `repositories` must not import services or the API layer.
- API routes must not import `sqlalchemy.orm` (async sessions come from `get_db`).
- Repository modules end in `_repository.py`.

Deliberately not enforced:

- Routes may import repository helpers (e.g. `get_active_user`), per the repo conventions in AGENTS.md.
- No `*_service.py` naming rule: service modules are purpose-named (`guardrails.py`, `url_safety.py`, ...).

Also includes empty, no-op `gateway/ports` and `gateway/adapters` rule entries as scaffolding for future open-core boundary rules (see ARCHITECTURE.md).

Verification: `tests/unit/test_check_architecture.py` adds 11 unit tests covering flagged and allowed imports (including `from gateway import api`, relative imports, and dot-boundary prefixes), the no-op scaffolding, naming, and a real-tree-is-clean gate. A planted `gateway.api` import in a services file fails the check with file:line and a non-zero exit. Docs were updated so `make lint` is no longer described as ruff-only (AGENTS.md, CONTRIBUTING.md, lint workflow step name).

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [x] Infrastructure / CI

## Relevant issues

Tracked internally; no public issue.

## Checklist

- [x] I understand the code I am submitting.
- [x] I have added or updated tests that cover my change (`tests/unit`, `tests/integration`).
- [x] I ran the Definition of Done checks locally (`make lint`, `make typecheck`, `make test`).
- [x] Documentation was updated where necessary.
- [x] If the API contract changed, I regenerated the OpenAPI spec (`uv run python scripts/generate_openapi.py`).

## AI Usage

- [ ] No AI was used.
- [x] AI was used for drafting/refactoring.
- [ ] This is fully AI-generated.

**AI Model/Tool used:** Claude Code (Opus 4.8)

**Any additional AI details you'd like to share:**

Drafted under maintainer direction and review; lint, typecheck, and the full test suite were run locally and code-review feedback was applied before push.

- [x] I am an AI Agent filling out this form (check box if true)

PR #476, opened 2026-08-03
fix(lint): apply every enclosing layer's rules in the architecture check

njbrake:
## Description

Follow-up to #475, from reviewing that PR after it merged.

`check_file()` picked the first `RULES` entry whose fragment prefixed the file path, so with nested fragments declaration order decided which rule applied and the other was silently dropped: no violation reported, exit 0, enforcement gone with no signal at all.

The first attempt here picked the most specific entry instead, which only relocated the problem: a broader `"gateway/api"` rule would then stop applying to everything under `gateway/api/routes`, the bulk of that layer. Reproduced against both versions, with a hypothetical `gateway/api` rule forbidding `gateway.db` on a routes file that imports both `gateway.db` and `sqlalchemy.orm`:

```
first-match wins   -> only gateway.db flagged      (the routes rule is shadowed)
most-specific wins -> only sqlalchemy.orm flagged  (the api rule is shadowed)
accumulate         -> both flagged
```

So restrictions now accumulate. A file answers to its own layer's rules and to every enclosing layer's, so no declaration order can shadow anything in either direction, and a nested layer can add restrictions but cannot opt out of an enclosing one. Violations are attributed to the closest layer that forbids the import, so existing messages are unchanged. `allowed` is documentation only, so no relaxation mechanism is given up.

Nothing regresses on the current tree. No two fragments nest today, and a differential run of the merged checker against this one over 18 synthetic files (relative imports, relative imports escaping the source root, function-scoped imports, `from gateway import api, models`, `from sqlalchemy import orm`, dot-boundary near-misses, a deliberate syntax error, `api/deps.py`, `ports`/`adapters`, `core/`) agrees on every case. The reason to fix it now is that #475 deliberately added the empty `gateway/ports` and `gateway/adapters` entries as scaffolding for future boundary rules, and adding a layer rule is precisely the change that would trip this.

Changes:

- `check_file()` collects the forbidden lists of every matching layer, most specific first.
- The `RULES` comment states the accumulate semantics, since rule authors read that block rather than `check_file`.
- `test_nested_and_enclosing_layer_rules_both_apply` and `test_violation_is_attributed_to_the_closest_layer` cover both directions plus attribution. Each fails on one of the two earlier behaviours.
- `test_relative_import_above_src_root_is_ignored` covers `from ... import x` walking above the source root, which `_resolve_relative()` already handled but nothing pinned. This was a CodeRabbit suggestion on #475.

Deliberately out of scope: the sync-ORM rule still covers only `gateway/api/routes/`, so `gateway/api/deps.py` and `gateway/api/main.py` remain unchecked. That is a policy question rather than a bug. It is now a safe change to make on top of this, which was not true before.

Verification: `make lint`, `uv run mypy` (282 files), and `uv run pytest tests/unit/test_check_architecture.py` (14 passed) all green locally.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [x] Infrastructure / CI

## Relevant issues

Follow-up to #475; no separate issue.

## Checklist

## AI Usage

- [ ] No AI was used.

Found while reviewing #475 with Claude Code after it merged, and the incomplete first attempt was caught by reviewing this PR the same way. Both the original bug and the half-fix were reproduced by executing the respective checker versions before the fix was written; lint, typecheck, and the unit tests were run locally.

### DIFF
--- a/scripts/check_architecture.py
+++ b/scripts/check_architecture.py
@@ -36,7 +36,9 @@ class LayerRule(TypedDict):
 # Rules are keyed by the layer's path below src/. Only "forbidden" is enforced;
 # "allowed" documents the layer contract for reviewers. Cross-cutting top-level
 # modules (gateway.log_config, gateway.metrics, ...) are always importable and
-# are not listed.
+# are not listed. Restrictions accumulate down the tree, so a file under
+# gateway/api/routes answers to a gateway/api entry as well as its own; a nested
+# layer can add restrictions but cannot opt out of an enclosing layer's.
 RULES: dict[str, LayerRule] = {
     "gateway/services": {
         "allowed": ["gateway.repositories", "gateway.models", "gateway.core", "gateway.auth"],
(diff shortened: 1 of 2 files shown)


peteski22 (approved):
🙏🏼

PR #477, opened 2026-08-03
docs: split AGENTS.md into scoped memory files and trim derivable content

njbrake:
> _Note: this PR was drafted by Claude via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

## Description

`AGENTS.md` (and therefore `CLAUDE.md`, which imports it) loads into every session in full, so anything a session could reconstruct on its own was paying rent on every turn. This PR does two things.

**1. Removed content derivable from the repo itself.** Things a session gets from `ls`, `pyproject.toml`, the Makefile, or `--help`:

* Project Snapshot: Python version, `uv`, app type, source root, test directories, DB drivers (all in `pyproject.toml`).
* Setup, Run, and Build command blocks: standard `uv venv` / `uv sync` / `uv build` invocations, `make dev`, and the `otari serve` / `init-db` / `migrate` subcommands.
* Ruff rule list, CI workflow filenames, dev dependency names, `make typecheck`. The ruff list was also stale: it said `E`, `F`, `I` while the real config is `E`, `F`, `I`, `PERF`, which is a good illustration of why copying config into a memory file rots.
* The `make test` variants, the CI pytest invocations, and the whole "Running a Single Test" section (textbook pytest node ids).
* The OpenAPI check command and output path. The generator invocation was kept, since `make openapi-check` only validates and there is no `make openapi` target.
* Style rules already enforced mechanically: import grouping and order (ruff rule `I`), type hints (mypy `strict`), plus generic naming and formatting conventions.
* The 12 line "Key Paths" list.

**2. Moved subdirectory-scoped guidance into nested memory files** that load only when the relevant tree is being worked on, instead of on every session:

* `web/AGENTS.md` (with `web/CLAUDE.md` as a one-line `@AGENTS.md` import): the dashboard auth and session model, runtime provider management, build plus the bundled user guide, PWA, and serving. This was the single largest block in the root file.
* `src/gateway/AGENTS.md` (with `src/gateway/CLAUDE.md` as a one-line `@AGENTS.md` import): request lifecycle, budget enforcement, built-in tools vs pass-through, data/sessions/migrations, config layering, plus the database and logging patterns.

**Deliberately kept in the root file**, because none of it is derivable and some of it is safety critical: the `gateway` vs `otari` package naming rationale, the two runtime modes including the two rejected `OTARI_MODE`/token combinations and the managed-vs-BYO trust boundary, the `make lint` gotcha (ruff alone is not equivalent), the architecture layer rules, the no global `reruns` rationale, the Postman goes stale on a docstring change trap, the "do not hand edit `CHANGELOG.md`" rule, never log secrets, the CI 3.14 vs package 3.13+ discrepancy, the writing style rules, and the Change Validation Checklist.

**3. Used `@AGENTS.md` imports rather than symlinks.** Git for Windows defaults to `core.symlinks=false`, so a symlinked `CLAUDE.md` checks out as a plain text file whose entire content is the string `AGENTS.md`. Claude Code then loads a memory file that says nothing, with no error, so a Windows contributor silently gets none of this guidance. All three `CLAUDE.md` files are therefore regular one-line files containing `@AGENTS.md`. Verified with headless runs that the content loads exactly once, at the repo root and in a subdirectory, and that no external includes prompt fires. Note `.claude/skills` is still a symlink to `.github/skills` and has the same Windows problem; it cannot be fixed this way since it is a directory, so it is left for a separate change.

The cross directory trap stays in the root as well: editing `docs/dashboard.md` makes the committed dashboard bundle stale even though that file lives outside `web/`, so that rule cannot live only in `web/AGENTS.md`.

Net effect: `AGENTS.md` goes from 21,051 to 9,022 characters, roughly 5.3k to 2.3k estimated always-resident tokens, with about 2k of that relocated rather than dropped. No behavior change; documentation only.

## PR Type

- [ ] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [x] Documentation
- [ ] Infrastructure / CI

## Relevant issues

None.

## Checklist

- [ ] I have added or updated tests that cover my change (`tests/unit`, `tests/integration`). Not applicable: this PR touches only Markdown memory files and changes no behavior.
- [x] I ran the Definition of Done checks locally (`make lint`, `make typecheck`, `make test`). Results: `make lint` passed (no architecture violations, ruff clean), `make typecheck` passed (282 source files, no issues), `make test-unit` passed (1131 passed, 1 skipped). `make test-integration` could not run in this environment because Docker is unavailable for Testcontainers and `TEST_DATABASE_URL` is unset; CI will cover it, and a Markdown only change cannot affect it.
- [x] Documentation was updated where necessary. This PR is the documentation update.
- [ ] If the API contract changed, I regenerated the OpenAPI spec (`uv run python scripts/generate_openapi.py`). Not applicable: no API contract change, so `docs/public/openapi.json` and the Postman collection are untouched and remain current.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

Claude Opus 5 via Claude Code, invoked through the `/doctor` health check.

This came out of a `/doctor` run that audited context cost across the setup. The audit measured what each always-loaded memory file contributes, classified each section as derivable from the repo or not, and proposed the cuts and relocations; @njbrake reviewed the proposed set and approved it before anything was edited. Where a call was borderline, the content was kept rather than cut. Every removed block was quoted back in the audit report first, so anything here is easy to restore if a reviewer disagrees with a specific cut.

**NOTE:**


peteski22 (approved):
I think we should consider just having the `CLAUDE.md` files have the line:

```text
@AGENTS.md
```

Because symlinks don't work on Windows so that hinders Windows based contributors who are using agentic coding tools.

Food for thought/discussion/future PR? 😅

njbrake:
_Note: this reply was drafted by Claude via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

Good catch, done in 8dd4a4ba. All three `CLAUDE.md` files are now regular one-line files containing `@AGENTS.md`.

Worth recording what the failure mode actually was, because it is quieter than "symlinks don't work": Git for Windows defaults to `core.symlinks=false`, so the symlink checks out as a plain text file whose entire content is the string `AGENTS.md`. Claude Code then loads a memory file that says nothing, with no error and no warning, so a Windows contributor silently gets none of the repo guidance and has no signal anything is wrong.

I checked the replacement rather than assuming:

* The content loads exactly once, so no double loading. This matters more than it sounds: Claude Code discovers `CLAUDE.md`, `.claude/CLAUDE.md`, `CLAUDE.local.md` and `.claude/rules/`, but not `AGENTS.md`, so the import is load bearing rather than decorative.
* No external includes prompt fires. An import only counts as external when it resolves outside the project, and `@AGENTS.md` is a sibling.
* Both hold at the repo root and in a subdirectory, which is the case that matters for `web/` and `src/gateway/`.

One thing your point applies to that this PR does not fix: `.claude/skills` is a symlink to `.github/skills`, so Windows contributors lose the scoped skills entirely. That cannot be solved with an import since it is a directory, so it needs its own decision. Happy to open an issue.

PR #481, opened 2026-08-03
test(messages): cover prompt caching contract

### LINKED ISSUE #480: [FEATURE] Pin the block-level cache_control round-trip with a contract test
### Problem or use case

Block-level `cache_control` currently survives `/v1/messages` untouched, and cache usage reaches the client correctly. Individual helper behavior is unit-tested, but no route-level contract test covers the complete request and response round trip. This is the mechanism downstream clients depend on for Anthropic prompt caching, so it is worth pinning against silent regression.

Current behavior, verified by reading `main` at `0ea2336`:

- `MessagesRequest` derives its wire fields from any-llm's `MessagesParams` (`src/gateway/api/routes/messages.py:63`), so content blocks pass through structurally.
- `_strip_gateway_fields` (`src/gateway/api/routes/_tools.py:78-86`) removes only `mcp_servers`, `mcp_server_ids`, `guardrails`, `tools_header`, `max_tool_iterations`, `session_label`, `user`. Message and system content are untouched.
- `_MessagesAdapter.extract_usage` and `_messages_stream_usage` retain cache-token information for gateway metering. Client-facing passthrough occurs when non-streaming responses and streaming SSE events are serialized from the original any-llm response models (`src/gateway/api/routes/messages.py:263-264`, `:513`, `:529`).
- `inject_purpose_hints_anthropic` (`src/gateway/services/tool_format.py:77`) preserves a caller-supplied `cache_control` block unchanged when prepending tool purpose hints.

All of that is correct today. Existing unit tests pin the hint helper and cache-token extraction, but not the endpoint boundary. A refactor of the strip list, schema derivation, hint injection, or response serialization could therefore break the contract without failing a route-level test.

### Proposed solution

Exercise `POST /v1/messages` with the provider mocked, capture the arguments received by `amessages`, and inspect the returned JSON or SSE events. Add contract coverage confirming that:

1. `cache_control` on a `system` content block reaches the provider call structurally unchanged.
2. `cache_control` on a message content block does the same.
3. A caller-supplied `cache_control` block remains structurally unchanged after tool purpose hints are injected.
4. `cache_creation_input_tokens` and `cache_read_input_tokens` reach the client on both the streaming and non-streaming paths.

### Alternatives considered

Relying on integration tests against a real provider. That would catch it, but only for whoever runs them with credentials, and a cache miss is not an error, so a broken pass-through would show up as a cost change rather than a failure.

HareeshBahuleyan:
## Description

Adds route-level contract coverage for Anthropic prompt caching so future refactors cannot silently drop cache markers or cache-usage data.

The tests verify system and message `cache_control` passthrough, preservation during purpose-hint injection, and cache usage in JSON and SSE responses. This is a test-only change with no runtime behavior changes.

### Validation

- `make lint`
- `make typecheck`
- `uv run pytest tests/unit/test_mcp_loop_messages.py tests/integration/test_messages_route_dispatch.py -q` (45 passed)

## PR Type

- [ ] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [x] Infrastructure / CI

## Relevant issues

Closes #480

## Checklist

- [ ] I ran the Definition of Done checks locally (`make lint`, `make typecheck`, `make test`). Targeted tests, lint, and typecheck were run; the full test suite passed in CI.
- [x] Documentation was updated where necessary. No documentation changes are needed for this test-only change.
- [x] If the API contract changed, I regenerated the OpenAPI spec (`uv run python scripts/generate_openapi.py`). The API contract is unchanged.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:** Codex (GPT-5)

**Any additional AI details you'd like to share:** The agent investigated the existing coverage, implemented the route-level tests, and ran the listed validation commands.

## Summary

- Added route-level contract tests for Anthropic prompt caching on `POST /v1/messages`.
- Verified cache controls remain unchanged during request handling, including tool purpose-hint injection.
- Verified cache creation and cache read usage data in JSON and SSE responses.
- Added response helper support for cache usage values and SSE parsing.

These tests help prevent regressions for downstream prompt-caching clients.

### DIFF
--- a/tests/integration/test_messages_route_dispatch.py
+++ b/tests/integration/test_messages_route_dispatch.py
@@ -3,7 +3,9 @@
 These complement :mod:`tests.unit.test_mcp_loop_messages` (which tests the
 Anthropic tool loop in isolation) by exercising the FastAPI route handler:
 tool extraction, mutual-exclusivity validation, error-body mapping to the
-Anthropic shape, and the per-tool dispatch into the right backend.
+Anthropic shape, the per-tool dispatch into the right backend, and the
+prompt-cache contract (``cache_control`` markers reaching the provider
+unchanged, cache usage reaching the client).
 """
 
 from __future__ import annotations
--- a/tests/integration/test_messages_route_dispatch.py
+++ b/tests/integration/test_messages_route_dispatch.py
@@ -8,6 +8,7 @@ Anthropic shape, and the per-tool dispatch into the right backend.
 
 from __future__ import annotations
 
+import json
 from collections.abc import AsyncIterator
 from typing import Any, cast
 from unittest.mock import AsyncMock, patch
@@ -29,7 +30,12 @@ from any_llm.types.messages import (
 from fastapi.testclient import TestClient
 
 
-def _text_response(text: str = "ok") -> MessageResponse:
+def _text_response(
+    text: str = "ok",
+    *,
+    cache_creation_input_tokens: int | None = None,
+    cache_read_input_tokens: int | None = None,
+) -> MessageResponse:
     return MessageResponse(
         id="msg_test",
         type="message",
@@ -41,8 +47,8 @@ def _text_response(text: str = "ok") -> MessageResponse:
         usage=MessageUsage(
             input_tokens=5,
             output_tokens=2,
-            cache_creation_input_tokens=None,
-            cache_read_input_tokens=None,
+            cache_creation_input_tokens=cache_creation_input_tokens,
+            cache_read_input_tokens=cache_read_input_tokens,
             cache_creation=None,
             server_tool_use=None,
             service_tier=None,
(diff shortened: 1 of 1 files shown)


njbrake (approved):

Approving. Prompt-cache pass-through is a silent-failure class of bug (a broken marker shows up as a cost increase, never as an error), so pinning it at the route level is well worth the ~1.2s of added runtime.

The thing I wanted to confirm for a test-only PR is that these tests actually fail when the behavior they name breaks, so I mutation-tested them. Six independent regressions in the production code, each caught by the intended test on the intended assertion line:

| Mutation | Target | Caught by |
| --- | --- | --- |
| add `"system"` to `_GATEWAY_INTERNAL_FIELDS` | `_tools.py` | non-stream round-trip |
| strip `cache_control` from system blocks | `messages.py` | non-stream round-trip |
| strip `cache_control` from message content blocks | `messages.py` | non-stream round-trip |
| non-stream response excludes cache usage (line 529) | `messages.py` | non-stream round-trip |
| hint injection merges into the caller's cached block instead of prepending | `tool_format.py` | purpose-hint test |
| `format_chunk` excludes cache usage from SSE | `messages.py` | stream test |

A few things I liked:

* The purpose-hint test patches `gateway.services.mcp_loop_messages.amessages` rather than faking the loop, so the real `anthropic_tool_loop` runs against `FakePool`. Sibling tests in this file stub `anthropic_tool_loop_stream` outright, so this one is strictly stronger and pins the route to `inject_hints` to loop wiring.
* Equality assertions (`captured["system"] == [system_block]`) rather than probing a single field, which is what makes the two strip-mutations fail.
* `ttl: "1h"` is pinned here; the existing unit test only covers `{"type": "ephemeral"}`.
* The factory widening is keyword-only with `None` defaults matching the previously hard-coded values, so the 20 pre-existing tests in the file are behaviorally untouched.

Locally: lint, mypy strict, the changed file (23/23), and the full unit suite (1134 passed, 1 skipped) all pass, matching green CI.

Two non-blocking gaps, both follow-up material rather than reasons to hold this:

1. The hybrid non-streaming client body is still unpinned. I verified this is a real hole rather than a guess: mutating `messages.py:513` to drop cache usage from the response body leaves 41 tests passing. The existing hybrid tests assert cache tokens in the upstream usage report, not in what the caller receives, and a self-hosted hybrid gateway does return that body to its callers.
2. `cache_control` on `tools` is untested, and one path silently drops it. An Anthropic-shaped tool keeps the field, but an OpenAI-shaped function tool loses it because `openai_to_anthropic_tools` copies only `name`, `description`, and `input_schema`. Pre-existing behavior on an unlikely input, so out of scope here.

One note on the CodeRabbit walkthrough: its complaint that the title should be `test:` rather than `test(messages):` is a false positive. `otari-pr-title.yml` lists `test` in `types` and sets no `scopes` or `requireScope`, and `cliff.toml`'s `^test` skip rule matches the scoped form too.

PR #482, opened 2026-08-04
feat(dashboard): show the failure diagnostic and status code in the activity detail

njbrake:

## Description

The Activity detail panel rendered a fixed `This request failed. Inspect gateway logs for details.` for every error row, so an operator could not tell a budget rejection from a provider outage without leaving the dashboard. The backend already serialized the stored `error_message` and `status_code` on every usage row; the panel just declined to render them.

This renders `error_message` verbatim and shows the classifying HTTP status beside the Error heading (for example `Error (502)`), omitted when no status was recorded (such as a stream that finished without usage data).

Why this is safe: the dashboard is master-key admin-only, and both fields were already on the wire in the `/v1/usage` response, so this surfaces existing data to the operator rather than exposing anything new. The text is source-neutral by nature, carrying either a fixed gateway rejection string (missing pricing, forbidden user, unresolvable model) or the raw upstream provider error. Raw provider text can occasionally include an upstream request URL or provider-side detail; it is behind admin auth and was already transmitted, so this does not newly expose it to non-admins, it just renders in the panel instead of requiring devtools.

Changes:
- `web/src/api/types.ts`: add `status_code` to `UsageEntry` (serialized by the backend but previously untyped on the frontend).
- `web/src/pages/ActivityPage.tsx`: render `error_message` verbatim, show the status code next to the heading, update the stale "provider diagnostics stay server-side" comment.
- `web/src/pages/ActivityPage.test.tsx`: flip the old masking test to assert the diagnostic and status code show; add a null-status-code case; add `status_code` to the fixture.
- `docs/dashboard.md`: replace the line that promised diagnostics stay in the logs.
- Rebuild the committed dashboard bundle under `src/gateway/static/dashboard/`.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

None.

## Checklist

- [x] I have added or updated tests that cover my change (frontend Vitest suite in `web/src/pages/ActivityPage.test.tsx`; the change is frontend-only, so no `tests/unit` or `tests/integration` changes apply).
- [ ] I ran the Definition of Done checks locally (`make lint`, `make typecheck`, `make test`). Ran the frontend equivalents instead (`npm --prefix web run typecheck`, `npm --prefix web test`: 379 passed); no backend Python code changed.
- [x] Documentation was updated where necessary (`docs/dashboard.md`).
- [x] If the API contract changed, I regenerated the OpenAPI spec. Not applicable: no route or schema changed, so the OpenAPI and Postman artifacts are unaffected.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:** Claude Opus 4.8 (1M context) via Claude Code.

**Any additional AI details you'd like to share:** The code and this description were written by Claude; the direction, the security tradeoff (surfacing raw provider text to the admin panel), and the decisions are @njbrake's.

**NOTE:**


web/src/api/types.ts
```diff
@@ -409,6 +409,10 @@ export interface UsageEntry {
   cost: number | null;
   status: string;
   error_message: string | null;
+  // HTTP status that classified the failure (402 no-pricing/over-budget, 403
```
khaledosman, line 412: "402 no-pricing/over-budget" doesn't match the backend taxonomy: `error_class_for` (`src/gateway/api/routes/usage.py:775-790`) records **403** for an over-budget or blocked user and 402 for missing pricing only. Move "over-budget" to the 403 clause, or the next reader goes hunting for 402 budget rows that don't exist.

web/src/pages/ActivityPage.tsx
```diff
@@ -307,19 +307,22 @@ function DetailField({ label, children }: { label: string; children: ReactNode }
...
+// under `require_pricing`) or the raw upstream provider error, so the heading
+// stays "Error" rather than blaming the provider for every failure.
 function RequestDetail({ entry }: { entry: UsageEntry }) {
   return (
     <div className="flex flex-col gap-4 px-4 py-4">
       {entry.error_message ? (
         <div className="flex flex-col gap-1.5">
-          <span className="text-[11px] font-medium uppercase tracking-wide text-[var(--otari-muted)]">Error</span>
+          <span className="text-[11px] font-medium uppercase tracking-wide text-[var(--otari-muted)]">
+            Error{entry.status_code !== null ? ` (${entry.status_code})` : ""}
```
khaledosman, line 322: `Error (400)` reads as "the status the client received", which it deliberately isn't: `failure_status_code` stores the upstream code while the caller sees something else (401/403 → 502 today; 400 → 502 once #483 lands, where the log keeps 400). `docs/dashboard.md` says "the HTTP status that classified the failure", but the panel carries no such hint. A label ("recorded status") or the existing coarse bucket from `error_class_for` would remove the ambiguity.

khaledosman (approved):
Exposure claim checks out: the usage list endpoint is `Depends(verify_master_key)` (`src/gateway/api/routes/usage.py:263`) and `status_code` was already serialized on every row (`usage.py:129-131`, always emitted, so the `!== null` check is safe). Bundle is consistent — `index.html` updated, stale `ActivityPage-*` chunk deleted, and the churn across unrelated page chunks is the expected shared-chunk cascade. Two non-blocking inline notes.

PR #483, opened 2026-08-04
fix(errors): classify provider billing exhaustion instead of calling it a bad request

njbrake:
## Description

Providers disagree on the HTTP status for "this account is out of money". Anthropic returns a **400** `invalid_request_error` carrying `Your credit balance is too low to access the Anthropic API`; OpenAI has a 400 `billing hard limit reached`; DeepSeek uses 402 `Insufficient Balance`.

Two things went wrong with that shape:

1. `classify_provider_error` mapped it to `PROVIDER_BAD_REQUEST_DETAIL`, so an operator whose provider account ran dry was told to "check the model name and parameters".
2. `_FALLBACK_NON_RETRYABLE_STATUS_CODES = {400, 422}` treats a 400 as "malformed, every provider would reject this", so a multi-attempt route never fell through to the other providers on the same alias. Recovering from a single provider going bad is one of the main reasons an operator configures fallback.

This came out of a real incident on a self-hosted otari: an Anthropic account ran out of credit, and the downstream app failed every agent turn for about 10 hours with the model-name detail while two other providers sat unused on the same alias. The raw cause was only visible in the gateway's own `_pipeline.py:2353` log line.

### What changed

- New `is_provider_billing_error` in `_platform.py`, gated on 400/402/422 and matched against a narrow tuple of fixed provider phrases. An unrecognized phrasing degrades to the existing generic handling rather than being misclassified, since a false positive would burn an attempt against every provider in the route.
- Billing failures surface as a **502** with a new fixed `PROVIDER_BILLING_DETAIL`, following the reasoning already applied to a rejected credential: a provider rejecting the gateway's account is a gateway-side fault, not the caller's, so it must not come back as a client 4xx. Still a fixed string, so the no-leak guarantee holds (covered by a test).
- `_classify_upstream_error` now returns `(True, "http_<code>_billing")` for these, so the route fails over and an empty wallet stays separable from a malformed request in logs and in what the gateway reports to the platform.
- `failure_status_code` is unchanged: `usage_logs.status_code` keeps the upstream 400, so "how much of my error rate is an empty wallet" stays answerable.
- 429 is deliberately left alone. OpenAI's `insufficient_quota` arrives as a 429, which is already failover-eligible and already tells the caller to back off, so reclassifying would only change wording.
- `_upstream_error_message` moves to `_platform.py` as `upstream_error_message` so the billing probe and the existing reasoning-effort probe share one implementation without `_platform` importing the API layer above it. `make check-architecture` passes.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

None filed; found while debugging a downstream outage.

## Checklist

Verification:

- `make lint`: architecture check clean, ruff clean
- `make typecheck`: no issues in 282 source files
- `uv run pytest tests/unit`: 1149 passed, 1 skipped
- The 10 new behaviour tests were confirmed to fail with the predicate stubbed to `return False`, while the two negative-control tests (unrecognized 400 stays generic; probe does not reinterpret 429/500/503) keep passing.
- No OpenAPI change: the detail strings are response bodies, not schema.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

Written while investigating the downstream incident described above, so the fixed provider phrases come from an observed Anthropic body plus documented phrasing for the other providers, not from guesswork alone. The probe list is best-effort by design and degrades safely.

## Summary

- Detects provider billing exhaustion for recognized 400, 402, and 422 responses.
- Reports these errors as safe 502 billing errors instead of malformed-request errors.
- Allows route failover to another provider.
- Preserves upstream status codes in usage logs.
- Keeps 429 handling unchanged.
- Moves shared upstream error handling to `_platform.py`.
- Documents billing error classifications and troubleshooting guidance.
- Adds coverage for recognized, unrecognized, and wrapped billing errors.

### DIFF
--- a/src/gateway/api/routes/_pipeline.py
+++ b/src/gateway/api/routes/_pipeline.py
@@ -260,8 +260,9 @@ def classify_provider_error(exc: BaseException) -> ProviderErrorMapping | None:
         return ProviderErrorMapping(status.HTTP_504_GATEWAY_TIMEOUT, PROVIDER_TIMEOUT_DETAIL)
     if status_code is None:
         return None
-    # Account billing exhaustion, which several providers report as a 400/402/422
-    # rather than a 402. Like a rejected credential this is a gateway-side account
+    # Account billing exhaustion, which several providers report as a 400/422
+    # rather than the 402 the condition deserves (and DeepSeek does report as a
+    # 402). Like a rejected credential this is a gateway-side account
     # fault rather than anything wrong with the caller's request, so it surfaces as
     # a 502 and never as a client-facing 400: telling the caller to "check the model
     # name and parameters" for an empty wallet sends operators debugging the wrong
--- a/src/gateway/api/routes/_pipeline.py
+++ b/src/gateway/api/routes/_pipeline.py
@@ -71,7 +71,9 @@ from gateway.api.routes._platform import (
     _resolve_platform_credentials,
     _resolve_platform_mcp_servers,
     _resolve_platform_web_search,
+    is_provider_billing_error,
     run_platform_attempts,
+    upstream_error_message,
     upstream_exception_shape,
 )
 from gateway.api.routes._platform import (
@@ -151,6 +153,10 @@ PROVIDER_REASONING_TOOLS_UNSUPPORTED_DETAIL = (
 )
 PROVIDER_MODEL_NOT_FOUND_DETAIL = "The requested model was not found on the provider"
 PROVIDER_CREDENTIALS_DETAIL = "The provider rejected the gateway's credentials"
+PROVIDER_BILLING_DETAIL = (
+    "The upstream provider account is out of credit or over its billing limit. "
+    "Top up the provider account, or route this model to a provider that has credit."
+)
 PROVIDER_RATE_LIMITED_DETAIL = "The provider rate-limited this request"
 ALL_PROVIDERS_FAILED_DETAIL = "All upstream providers failed"
 ALL_PROVIDERS_TIMED_OUT_DETAIL = "All upstream providers timed out"
@@ -211,26 +217,6 @@ def _upstream_error_param(exc: BaseException) -> str | None:
     return None
 
 
-def _upstream_error_message(exc: BaseException) -> str:
-    """Best-effort human-readable text from an upstream exception (and its wrapper).
-
-    Concatenates the exception's ``message`` attribute and ``str()`` for both the
-    exception and any ``original_exception``, so a substring probe still matches
-    whether otari receives the raw SDK error today or the wrapped
-    ``AnyLLMError`` after any-llm flips to unified exceptions. Used only for
-    fixed-string classification, never echoed back to the caller.
-    """
-    parts: list[str] = []
-    for candidate in (exc, getattr(exc, "original_exception", None)):
-        if candidate is None:
-            continue
-        message = getattr(candidate, "message", None)
-        if isinstance(message, str):
-            parts.append(message)
-        parts.append(str(candidate))
-    return " ".join(parts)
-
-
 def _is_reasoning_effort_tools_conflict(exc: BaseException) -> bool:
     """True when a 400/422 is the 'function tools + reasoning_effort' rejection.
 
@@ -250,7 +236,7 @@ def _is_reasoning_effort_tools_conflict(exc: BaseException) -> bool:
     """
     if _upstream_error_param(exc) != "reasoning_effort":
         return False
-    return "function tools" in _upstream_error_message(exc).lower()
+    return "function tools" in upstream_error_message(exc).lower()
 
 
 def classify_provider_error(exc: BaseException) -> ProviderErrorMapping | None:
@@ -274,6 +260,15 @@ def classify_provider_error(exc: BaseException) -> ProviderErrorMapping | None:
         return ProviderErrorMapping(status.HTTP_504_GATEWAY_TIMEOUT, PROVIDER_TIMEOUT_DETAIL)
     if status_code is None:
         return None
+    # Account billing exhaustion, which several providers report as a 400/402/422
+    # rather than a 402. Like a rejected credential this is a gateway-side account
+    # fault rather than anything wrong with the caller's request, so it surfaces as
+    # a 502 and never as a client-facing 400: telling the caller to "check the model
+    # name and parameters" for an empty wallet sends operators debugging the wrong
+    # thing. Checked ahead of the status branches so it wins over both the generic
+    # bad-request detail and the 402 fall-through to a bare 502.
+    if is_provider_billing_error(exc):
+        return ProviderErrorMapping(status.HTTP_502_BAD_GATEWAY, PROVIDER_BILLING_DETAIL)
     if status_code in (400, 422):
         # Preserve the actionable remedy for the one 400/422 a caller can fix by
         # retrying: function tools combined with a non-'none' reasoning_effort.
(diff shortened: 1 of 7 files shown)


src/gateway/api/routes/_platform.py
```diff
@@ -59,9 +59,41 @@
...
+_BILLING_MESSAGE_PROBES = (
+    "credit balance is too low",  # anthropic
+    "purchase credits",  # anthropic
+    "plans & billing",  # anthropic
+    "billing hard limit",  # openai
+    "insufficient balance",  # deepseek, moonshot
+    "insufficient_quota",  # openai (error code, echoed in some messages)
+    "exceeded your current quota",  # openai
+    "insufficient credits",  # openrouter, together
+    "payment required",  # generic 402 bodies
```
HareeshBahuleyan: **[P2] Make this probe status-specific to avoid classifying echoed client input as billing**

The billing check scans the entire provider message, but providers commonly echo invalid client values in 400 responses. For example, the gateway accepts an arbitrary string for `tool_choice`, and a request using `"payment required"` can produce an upstream message such as `Invalid value: 'payment required'. Supported values are: 'none', 'auto', and 'required'.` This substring then makes `_classify_upstream_error` return `(True, "http_400_billing")` and maps the response to the billing 502, causing a malformed request to run against every fallback provider instead of returning the expected 400.

Could generic phrases such as `payment required` be restricted to status 402, while 400/422 use structured provider error codes or exact known billing messages? A regression test with a billing phrase echoed as an invalid parameter value would cover this case.

src/gateway/api/routes/_platform.py
```diff
@@ -611,6 +688,13 @@ def _classify_upstream_error(exc: BaseException) -> tuple[bool, str]:
         return True, kind
 
     if isinstance(status_code, int):
+        # Checked before the non-retryable set: a billing 400/402/422 is an
+        # account condition on THIS provider, so the route's next attempt is
+        # worth making. The distinct error_class keeps "we ran out of credit"
+        # separable from "the request was malformed" in logs and in what the
+        # gateway reports back to the platform.
+        if is_provider_billing_error(exc):
+            return True, f"http_{status_code}_billing"
```
khaledosman, line 710: Undocumented wire-contract change: `docs/hybrid-mode-protocol.md:313` still enumerates `http_<code>` as the only status-derived tag, and this value is POSTed to the platform's `/gateway/usage`. That same doc (~line 103) warns a handler that rejects unexpected input 422s the report, which is non-retryable, so the report is silently dropped. Add the `http_<code>_billing` row to the table, and confirm otari-ai doesn't match `http_\d+$`. The checklist says docs were updated, but the diff has none.

src/gateway/api/routes/_pipeline.py
```diff
@@ -151,6 +153,10 @@
 )
 PROVIDER_MODEL_NOT_FOUND_DETAIL = "The requested model was not found on the provider"
 PROVIDER_CREDENTIALS_DETAIL = "The provider rejected the gateway's credentials"
+PROVIDER_BILLING_DETAIL = (
```
khaledosman, line 156: This detail rarely reaches the caller on the routes this PR targets: `_platform.py:305` and `_pipeline.py:2172` both collapse an exhausted multi-attempt route to `ALL_PROVIDERS_FAILED_DETAIL`, so in the incident's own shape (three providers on one alias, all dry) the operator still sees "All upstream providers failed". `PROVIDER_BILLING_DETAIL` only surfaces on a single-attempt route or after tool-loop lock-in. If every recorded failure's `error_class` ends in `_billing`, the aggregate could return this detail instead.

Separately, style: this is the only multi-sentence imperative detail in the set, and it instructs an action only the operator can take. The first sentence alone matches `PROVIDER_CREDENTIALS_DETAIL`.

src/gateway/api/routes/_platform.py
```diff
@@ -599,6 +631,51 @@ def upstream_exception_shape(exc: BaseException) -> tuple[UpstreamErrorKind | No
...
+
+    Gated on :data:`_BILLING_CANDIDATE_STATUS_CODES` so the probe can only ever
+    reinterpret a status that is otherwise a dead end, and matched against the
+    narrow :data:`_BILLING_MESSAGE_PROBES` so an unrecognized phrasing falls
+    through to the existing generic handling instead of being misclassified.
+    """
+    _kind, status_code = upstream_exception_shape(exc)
+    if status_code not in _BILLING_CANDIDATE_STATUS_CODES:
+        return False
+    message = upstream_error_message(exc).lower()
```
khaledosman, line 688: `upstream_error_message` concatenates `str(exc)`, which can include request content the provider echoed back. So a malformed request whose body contains e.g. "insufficient balance" now burns an attempt against every provider in the route and returns 502 instead of 400 — caller-influenced, if unlikely. Anchoring the probe on the provider's structured `message` / error-type field rather than the full stringification closes it.

src/gateway/api/routes/_platform.py
khaledosman: `"payment required"` matches any httpx-stringified 402 (`Client error '402 Payment Required' for url ...`), so the 402 arm is effectively status-based, not phrase-based. Fine by me, but the comment above claims "a phrase only earns a place here if it cannot plausibly describe a malformed request", which doesn't describe this entry.

src/gateway/api/routes/_pipeline.py
```diff
@@ -274,6 +260,15 @@ def classify_provider_error(exc: BaseException) -> ProviderErrorMapping | None:
...
     if status_code is None:
         return None
+    # Account billing exhaustion, which several providers report as a 400/402/422
+    # rather than a 402. Like a rejected credential this is a gateway-side account
+    # fault rather than anything wrong with the caller's request, so it surfaces as
+    # a 502 and never as a client-facing 400: telling the caller to "check the model
+    # name and parameters" for an empty wallet sends operators debugging the wrong
+    # thing. Checked ahead of the status branches so it wins over both the generic
+    # bad-request detail and the 402 fall-through to a bare 502.
+    if is_provider_billing_error(exc):
```
khaledosman, line 271: `status_code` is already in hand from line 258, but `is_provider_billing_error` re-walks the exception chain through `upstream_exception_shape`. Pass the status in, or move the probe inside the `(400, 422)` / 402 arms so the 401/403/404/429 paths skip it entirely.

tests/unit/test_classify_upstream_error.py
```diff
@@ -181,3 +181,64 @@ def test_classifies_provider_sdk_status_errors(sdk_error: type, status: int, ret
...
+
+@pytest.mark.parametrize(
+    "status, message",
+    [
+        (400, _ANTHROPIC_BILLING_MSG),
+        (402, "Insufficient Balance"),
+        (422, "insufficient credits for this request"),
+    ],
+)
+def test_billing_exhaustion_falls_through_to_the_next_attempt(status: int, message: str) -> None:
```
khaledosman, line 211: Both new test files stop at the classifier, so nothing exercises the behavior the fix exists for: a two-attempt route where attempt 1 returns a billing 400 and attempt 2 succeeds. That's where a future refactor of the retryable ladder would regress silently — an integration test driving `run_platform_attempts` would pin it.

khaledosman (approved):
Classification is right, and the negative controls (unrecognized 400 stays terminal, 429/503 untouched, no-leak, `failure_status_code` preserved) are the ones that matter. Nothing blocking; the two inline notes I'd act on before merge are the undocumented `error_class` shape in `docs/hybrid-mode-protocol.md` and the new detail being swallowed by the multi-attempt aggregate.

One scope check while reading: `run_platform_attempts` is only reached under `ctx.hybrid_mode` (`chat.py:368-375`, `responses.py:374`) and standalone aliases are single-target (`docs/models.md:216+`), so the failover half of this only takes effect in hybrid mode. Worth stating in the PR body, since it currently reads as a general fallback fix.

PR #485, opened 2026-08-04
feat(dashboard): make ids selectable and copyable in the tables

### LINKED ISSUE #478: Web Dashboard: add little copy icons for easy copying of model ids to clipboard
For example, in the alias page I can't highlight to select a model id. I need an easy way to copy pertinent info into my clipboard in the web ui.

njbrake:
## Description

Ids in the dashboard tables could not be copied. The issue asked for copy icons, and investigating why highlighting failed turned up the underlying cause, so this PR fixes both.

**Why highlighting was impossible.** Three things stacked up, each confirmed in a real browser:

1. The row carries `user-select: none` (HeroUI's styling for a selectable row, and react-aria's `usePress` sets the same inline while a press is in flight, since `useSelectableItem` never passes `allowTextSelectionOnPress`). Because `user-select: auto` on a cell resolves to its parent's behavior, the value inside inherited "not selectable", so a drag across a model id selected nothing at all.
2. With that overridden, the press toggles the row's selection on pointer *down*, and the re-render that causes lands mid-drag and discards the selection the browser had just started. A drag then selected the row instead of the text.
3. With the press suppressed, the *first* drag in a freshly loaded table still failed while later ones worked: the initial focus move into the react-aria cell re-renders the row.

**The fix.** A CSS override for (1), plus a new `CopyableValue` in `components/ui.tsx` that renders an id as focusable, selectable text with a copy control beside it: keeping the press off the value handles (2), and making the value itself focusable (`tabIndex={-1}`) handles (3). Where rows have a drill-in action, `DataTable` now owns pointer down and click on data cells, which keeps the action firing exactly once and makes any cell's text selectable; it skips the action for the click that ends a drag, so highlighting an id no longer opens the detail panel.

**Where copy controls went.** The values an operator pastes elsewhere, favoring the ones where the displayed text is not the whole value:

- Models: the `provider:model` id (the table shows a name with the provider prefix stripped) and the detail panel's selector.
- Budgets: the budget id (the table renders only a short prefix).
- Activity detail: request id, user, and api key, the UUIDs that go into a log search or a support thread.
- Aliases: the alias name, its target, and the user id in "Applies to".
- Users: the user id, which every management endpoint and usage filter takes.

Deliberately skipped: the Keys table's truncated `key_prefix` (a partial secret, so copying it is a trap), the Providers instance (short, and the cell is already a link), and the Usage/Settings aggregates.

**Clipboard on plain HTTP.** A self-hosted dashboard is usually reached over plain HTTP, and a LAN address is not a secure context, so `navigator.clipboard` is undefined there and the first pass reported "copy failed" on the most common deployment. `lib/clipboard.ts` now falls back to `document.execCommand("copy")` from an offscreen textarea, restoring the operator's own selection afterwards, and reports failure only when neither path works. The helper moved out of `ActivityPage`, where it was exported but never rendered.

A press confirms with a short "Copied!" over the icon that was pressed. It is a controlled tooltip (never hover-opened, since it reports an event rather than hinting) rendered in an overlay, so it neither reflows the row it reports on nor gets clipped by the table's scroll container.

**How I verified.** Beyond the unit tests, I drove a real Chromium against a running gateway: dragging across an id selects it (first drag included), double-click selects a word, the click ending a drag does not drill in, a plain click still drills in exactly once, checkboxes still select rows on both table kinds, and a copy on an insecure origin pastes correctly into a form field. Two e2e locators needed tightening: `getByLabel("User ID")` and `getByLabel("Alias name")` also matched the new "Copy user id" and "Copy alias name" controls.

## PR Type

- [x] New Feature
- [x] Bug Fix
- [ ] Refactor
- [x] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes #478

## Checklist

Notes on the checklist, so nothing is overstated:

- Tests are Vitest, since this is a dashboard-only change and no Python was touched: 419 pass, including new coverage for `CopyableValue`'s press suppression, the clipboard fallback paths, `DataTable` skipping the row action for a selection-ending click, and a copy assertion on each page that gained a control. The 7 Playwright e2e flows pass too.
- `make lint`, `make typecheck` (mypy, 282 files), and `make test-unit` (1134 tests) all pass. The integration suite does not run in my sandbox: without Docker it errors on Testcontainers, and against a hand-run PostgreSQL it hits async fixture teardown errors under Python 3.13 rather than CI's 3.14. No Python changed here, so CI is the check that matters for those.
- No API contract change, so the OpenAPI spec and Postman collection are untouched. `docs/dashboard.md` gained a short "Copying ids" section, and the committed dashboard bundle is rebuilt.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

Written by Claude through back and forth with @njbrake, who directed the work and made the design calls: asking for the confirmation to be a small "Copied!" anchored above the click rather than a toast, and pushing back on the first pass to ask whether the copied values were the right ones and whether the broken drag-to-select could be fixed instead of worked around. That question is what turned this from copy icons into a root-cause fix. The prose here is Claude's.

**NOTE:**

## Summary

- Added selectable, copyable values for model, alias, user, budget, and activity identifiers.
- Added copy controls with success and failure feedback.
- Added clipboard support for secure and plain HTTP origins.
- Updated data tables to preserve text selection and prevent accidental row actions.
- Added documentation for copying identifiers.
- Regenerated dashboard assets and updated end-to-end selectors.

## Technical notes

- Added `CopyableValue`, `CopyButton`, and `copyToClipboard`.
- Added CSS overrides for table-row text selection.
- Added unit, component, page, and end-to-end test coverage.
- Test results: 419 Vitest tests and 7 Playwright tests passed.


web/src/pages/ModelsPage.tsx
```diff
@@ -904,11 +915,13 @@ function ModelTable({
       header: "Model",
       isRowHeader: true,
       allowsSorting: true,
+      // The copy yields the provider-qualified key, which is what a caller sends
+      // as `model`; the visible text drops the provider prefix.
       cell: (row) => (
-        <span className="font-medium break-all">
+        <CopyableValue value={row.key} label="model id" className="font-medium break-all">
           {row.model}
           <span className="sr-only">{row.key}</span>
```
khaledosman: The `sr-only` key is now *inside* the selectable value, so a manual drag across this cell copies the id twice.

Tailwind v4's `sr-only` is `position:absolute;width:1px;clip:rect(0,0,0,0)` — rendered, so its text is part of a range selection and of `Selection.toString()`. Dragging across the Model cell and pressing Ctrl-C yields `gpt-4oopenai:gpt-4o`. Invisible before this PR because selection was impossible; now it's reachable, in the exact cell #478 was filed about.

```suggestion
          <span className="sr-only select-none">{row.key}</span>
```

An own declaration outranks the inherited `user-select: text`, the same reasoning the PR already uses for `select-text`.

web/src/components/DataTable.tsx
```diff
@@ -225,6 +236,11 @@ export function DataTable<Row extends object>({
   const fireRowAction = useCallback(
     (key: string) => {
       if (!onRowAction) return;
+      // A click that ended a text drag inside the table is a selection, not an
+      // activation. Cells are selectable by design (see globals.css), and
+      // drilling in mid-highlight both loses the selection and moves the page
+      // under the operator, so the action is skipped for that click only.
+      if (hasTextSelectionIn(rootRef.current)) return;
```
khaledosman: This guard also gates **keyboard** activation. `fireRowAction` is shared with `Table.Content`'s `onRowAction`, so Enter on a focused row is silently dropped while a text selection is anchored anywhere in the table — e.g. immediately after drag-selecting an id in that row, where focus is inside the row and Enter is a deliberate activation.

Moving the check to the click path keeps the drag behavior and leaves the keyboard alone:

```ts
onClickCapture={(e: ReactMouseEvent) => {
  const key = dataCellRowKey(e);
  if (key == null) return;
  e.stopPropagation();
  if (!hasTextSelectionIn(rootRef.current)) fireRowAction(key);
}}
```

Note the propagation stop stays unconditional — you still want the click swallowed on the drag-ending case, just not the action fired.

web/src/components/ui.tsx
```diff
@@ -210,6 +211,96 @@ export function RefreshButton({
...
+  return (
+    <Tooltip.Root isOpen={state !== "idle"}>
+      <Button size="sm" variant="ghost" isIconOnly aria-label={`Copy ${label}`} onPress={copy}>
+        <svg viewBox="0 0 24 24" fill="none" stroke="currentColor" strokeWidth="2" className="h-3.5 w-3.5" aria-hidden="true">
+          <rect x="9" y="9" width="11" height="11" rx="2" />
+          <path d="M5 15V5a2 2 0 0 1 2-2h8" strokeLinecap="round" strokeLinejoin="round" />
+        </svg>
+      </Button>
+      <Tooltip.Content placement="top" showArrow>
+        {state === "failed" ? "Copy blocked, press Ctrl/Cmd-C" : "Copied!"}
```
khaledosman: Nothing is selected at this point, so the instruction can't be followed as given. `copy` only calls `copyToClipboard` and sets state.

`KeysPage`'s `CopyField` shows the identical message and it works there because it does `ref.current?.select()` *before* attempting the copy (`src/pages/KeysPage.tsx:87-88`). The comment above this component claims "the same rule as the Keys page's CopyField", but the behavior differs.

On failure, select the sibling value's contents (`range.selectNodeContents` on the value span, via a ref from `CopyableValue`) so Ctrl/Cmd-C actually copies the right thing.

web/src/lib/clipboard.ts
```diff
@@ -0,0 +1,53 @@
+// Copy text to the clipboard, reporting whether it actually landed there. A
+// caller must never claim a copy it did not make.
+//
+// Two paths, because the async Clipboard API is gated on a secure context and
+// this dashboard is routinely served from a plain-HTTP LAN address (a gateway on
+// 10.x/100.x is not localhost, so `navigator.clipboard` is undefined there).
+// document.execCommand("copy") is deprecated but is the only clipboard write
+// such an origin has, so it is the fallback rather than an error message.
```
khaledosman, line 8: Two call sites still hand-roll this and don't get the fallback: `KeysPage.tsx:89` and `SettingsPage.tsx:280` both do a bare `navigator.clipboard?.writeText`.

So on the plain-HTTP LAN deployment this helper's whole rationale is built around, the one-time-visible **secret key** and the **master key** still require a manual Ctrl-C, while a model id copies programmatically. That's backwards from a value-importance standpoint.

Routing both through `copyToClipboard` is a small change; they can keep their select-and-hint path as the last resort when this returns `false`.

web/src/lib/clipboard.ts
```diff
@@ -0,0 +1,53 @@
...
+  source.readOnly = true;
+  // Fixed and transparent rather than display:none, which is not selectable.
+  source.style.position = "fixed";
+  source.style.top = "-1000px";
+  source.style.opacity = "0";
+  document.body.appendChild(source);
+
+  const selection = document.getSelection();
+  const previous = selection && selection.rangeCount > 0 ? selection.getRangeAt(0) : null;
+  source.select();
```
khaledosman, line 41: `select()` focuses the textarea in Chrome, and `source.remove()` below then drops focus to `<body>` — a keyboard user who activates the copy button loses their tab position. The *selection* is restored (nice touch), focus isn't:

```ts
const previousFocus = document.activeElement as HTMLElement | null;
// … after source.remove() and the selection restore:
previousFocus?.focus?.();
```

Also worth knowing for the fallback path: iOS Safari won't select a `readOnly` textarea, so `execCommand` is a no-op there. Not a regression, just a limit on how far the fallback reaches.

web/src/components/ui.tsx
```diff
@@ -210,6 +211,96 @@ export function RefreshButton({
...
+  const keepPressFromRow = (event: { stopPropagation: () => void }) => event.stopPropagation();
+  return (
+    <span className="inline-flex items-center gap-1">
+      <span
+        // Focusable, but not tabbable: pressing here focuses the value itself
+        // instead of the react-aria table cell, whose focus bookkeeping re-renders
+        // the row and (again) discards a drag that has only just begun. Without
+        // this, the first drag in a freshly loaded table selected nothing and only
+        // subsequent ones worked.
+        tabIndex={-1}
```
khaledosman, line 251: Clicking a value now focuses this span rather than the `<tr>`, so react-aria's collection focus (`focusedKey`) doesn't move — a subsequent ArrowDown navigates from whatever row was focused before. Deliberate trade for the drag fix as I read it, but was it checked? If it matters, `manager.setFocusedKey` on pointer-up (after the drag can no longer be disturbed) would recover it.

web/src/pages/AliasesPage.tsx
```diff
@@ -249,15 +249,25 @@ export function AliasesPage() {
...
-      cell: (a) => <span className="font-medium break-all text-[var(--otari-ink)]">{a.name}</span>,
+      cell: (a) => (
+        <CopyableValue value={a.name} label="alias name" className="font-medium break-all text-[var(--otari-ink)]" />
+      ),
+    },
+    {
+      id: "target",
+      header: "Target",
+      cell: (a) => (
+        <CopyableValue value={a.target} label="target model id" className="break-all text-[var(--otari-muted)]" />
```
khaledosman, line 260: Three `Tooltip.Root` + HeroUI `Button` pairs per row, and this table has no `TablePagination` — same for Users and Budgets, so the full list renders.

`DataTable.tsx` already warns that per-row render cost mattered enough to add the row cache ("checkbox clicks lag by whole seconds at large page sizes"). Worth a quick check at a few thousand aliases/users rather than assuming it's free. If it does bite, mounting `Tooltip.Root` lazily isn't viable (it would remount the button and drop focus), so a single lifted overlay anchored to the pressed button would be the way.

web/src/styles/globals.css
```diff
@@ -93,6 +93,26 @@
...
 
+/* Let the text in a row be highlighted with the mouse (issue #478). Two things
+   block it out of the box: the row carries `user-select: none` (HeroUI's styling
+   for a selectable row), and react-aria's usePress sets the same inline on the row
+   for the duration of a press, since useSelectableItem never passes
+   `allowTextSelectionOnPress`. Because `user-select: auto` on a cell resolves to
+   its parent's behavior, the value inside inherits "not selectable" and a drag
+   across an id selects nothing at all. `!important` is what outranks the inline
+   style, so it is deliberate here. Scoped to the body cells, leaving the header's
+   sort buttons and the selection checkboxes untouched.
```
khaledosman: Comment-only: `.otari-table .table__row .table__cell` matches the selection cell too, so "the selection checkboxes untouched" isn't quite right. Harmless in effect (a checkbox has no text to select), but the comment is load-bearing documentation for a `!important` rule, so it's worth being accurate.

khaledosman (approved):
Good root-cause work: the three stacked causes are correctly identified and the comments explain the *why*. Approving, with one thing I'd fix before merge (inline on `ModelsPage.tsx`): the `sr-only` full model key now sits inside the newly selectable value, so dragging across the Model cell copies the id twice. Two-word fix.

Other inline notes are non-blocking: the selection guard also swallowing keyboard `Enter`, the "press Ctrl/Cmd-C" message with nothing selected, the two remaining hand-rolled clipboard call sites, and focus not being restored after the `execCommand` path.

Two risks not anchorable to a line:

- **The fix hangs on HeroUI internals with nothing guarding them.** `.table__row` / `.table__cell` are HeroUI 3.2.1 class names, and the pointer interception depends on react-aria's press implementation. The new unit tests assert the *mechanism* (`className` contains `select-text`, `stopPropagation` was called), not the outcome, because jsdom can't. A Playwright test doing a real `mouse.down/move/up` across a model id and asserting `getSelection().toString()` would catch a HeroUI bump silently reverting #478; the e2e harness is already there.
- **Chromium-only verification.** The drag-vs-click heuristic depends on *when* the engine collapses a selection relative to the `click` event, which differs in Firefox and Safari. Worth a manual pass on both, since a wrong answer means either "highlighting drills in" or "clicking never drills in".

Test coverage is otherwise solid: all six clipboard paths including "no mechanism at all", both the drag-ending click and the selection-outside-the-table case, and a copy assertion per page. The `pressRaw` helper and its comment on why `userEvent.click` can't model this are the right call.

<sub>Reviewed with [Claude Code](https://claude.com/claude-code)</sub>

Issue #401, opened 2026-07-23
Routing and cost observability on responses


njbrake:
> _Note: this issue was drafted by Claude via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

## Context

Otari already records provider, cost, and latency per request and surfaces them in `ActivityPage.tsx` (Provider, Cost, Total time fields) from `UsageLog`. What it does not expose is:

1. That metadata on the API response itself (headers), for programmatic consumers and sidecars.
2. Fallback/retry/attempt counts, even though otari performs fallback in hybrid mode.

## Proposal (backend)

- Emit otari-namespaced response headers on chat/messages/responses: resolved backend, response cost (USD), upstream duration, and attempt/fallback/retry counts where applicable (for example `x-otari-response-cost`, `x-otari-backend`, `x-otari-attempted-fallbacks`, `x-otari-attempted-retries`).
- Ensure the same fields exist on the streaming path (trailing usage/cost is already parsed there).
- Persist fallback/retry/attempt counts on `UsageLog` so they are queryable, not only header-transient.

## Web dashboard

Cost, provider, and latency are already shown, so the dashboard delta is the routing signals:

- Add fallbacks / retries (and, if useful, resolved backend) to the request detail drawer in `ActivityPage.tsx`, next to the existing Provider / Cost / Total time fields. Show them only when non-zero/known so historical rows without the data stay clean.
- Optionally add a small aggregate (fallback rate / retry rate) to `UsagePage.tsx` or `OverviewPage.tsx` if the persisted counts make it cheap.
- Extend the activity/usage API client types for the new `UsageLog` fields; update tests, rebuild, and commit the bundle.

## Acceptance criteria

- Non-streaming and streaming responses carry cost, backend, and attempt metadata via headers.
- Fallback/retry counts are persisted on `UsageLog` and populated in the modes where otari performs them (absent/zero otherwise, documented).
- The Activity detail view shows routing signals when present.
- Header names and semantics documented; OpenAPI regenerated.

## Open questions

- Headers only, or also a structured field in the non-streaming JSON envelope?
- Expose backend as an opaque label rather than a raw `api_base` URL to avoid leaking provider topology?

## Background

[...]

Issue #187, opened 2026-06-18
Smart routing: pluggable LLM router with kNN routing memory + preference collection


njbrake:
_Note: this issue was drafted by Claude via back and forth with @njbrake. The reasoning and technical decisions are his; the prose is Claude's._

## Summary

Add a router to otari that sends each request to the cheapest or fastest model
that is still good enough for the task, and gets better at that decision over
time by learning from the user's own preferences. Ships as a pluggable interface
with a default reference implementation, so the routing algorithm is swappable.

Tracks: mozilla-ai/otari-ai#1039

## Motivation

The gateway is the only component that sees every request, model, dollar, and
latency number, so it is the right place to decide which model should handle a
given prompt. Routing simple prompts to a smaller model can cut cost
dramatically (RouteLLM reports about 95% of GPT-4 quality at about 14% of the
GPT-4 calls) without the caller changing anything.

## Prior art and how this differs

- **LiteLLM Router** (OSS) load-balances across deployments of the same model
  group via `latency-based`, `least-busy`, and `cost-based` strategies.
  `cost-based-routing` picks the cheapest deployment and is not quality-aware: it
  does not decide whether a weaker model would suffice for a prompt.
  https://docs.litellm.ai/docs/routing
- **LiteLLM Auto-Router** (built on `semantic-router`) is content-aware, but it
  matches prompts against hand-authored utterance routes with a similarity
  threshold. The routes are configured by the operator, not learned, and there is
  no cost-quality objective. https://docs.litellm.ai/docs/proxy/auto_routing
  (An open request for preference-aligned routing exists:
  https://github.com/BerriAI/litellm/discussions/25703)
- **Closed products** (Martian, NotDiamond) train per-user routers from eval data
  but are not open and not embeddable in a self-hosted gateway.

otari's proposal: a per-tenant router that learns from the user's own ranked
preferences and live traffic, with no hand-authored routes, an explicit
cost-versus-quality dial, a cascade fallback, and a pluggable backend so other
routers (including a trained one) can be dropped in.

## Approach

1. **`RouterBackend` protocol** (same pattern as `SandboxBackend` /
   `WebSearchBackend`): `route(ctx) -> RoutingDecision` returns an ordered
   candidate model list plus a confidence; `record_outcome(...)` for online
   learning. The routing algorithm lives behind this seam.
2. **Default impl: kNN "routing memory."** No training pipeline. Embed the task
[...]

PR #492, opened 2026-08-04
feat(routing): named routing policies with failover, budget tier-down, and enforced guardrails

### LINKED ISSUE #463: Feature: advanced alias features (routing, fallbacks, budget based fallbacks, etc)
The 'alias' feature is the perfect launching point for our advanced routing features, and even for things like budget based fallback logic, failure fallback etc.

Here's the idea:

1. you make an alias, and by default you make an alias point to a specific model that otari manages
2. However, you can instead configure that alias to perform any number of advanced logic:
    a. You can set it so that once a certain budget is exceeded, it will fall back to a different model
    b. you can set it so that if a first model fails the request will be resent to a different model
   c. you can point it to a router (see the feature about knn based model routing) that will dynamically pick the right model for the job.

Alias is the easy launching point to sort of create 'custom' models which are more dynamic and advanced model selection algorithms.

njbrake:
## Description

Closes #463. Also delivers the observability half of #401.

An alias maps one name to one model. This turns that into a **routing policy**: a
model name your callers send, which decides which real model serves the request,
what is tried after a retryable failure, and which guardrails always run. An alias
is the one-target case, so `aliases:` keeps working as its shorthand and nothing
existing changes behavior.

The gap this closes: a standalone gateway had **no failover at all**. The
multi-attempt walker existed but was hybrid-only, driven by the otari.ai resolve
payload, so any provider blip was a 502.

```yaml
routing:
  policies:
    fast:
      select:
        - when: {budget_used_pct: {gte: 80}}
          target: openai:gpt-5-nano      # tier down as the budget fills up
        - default: openai:gpt-5-mini
      on_failure:
        - anthropic:claude-haiku-4-5     # tried if the selected model fails
      guardrails:
        - {profile: prompt-injection, mode: block, on_unavailable: block}
```

Policies are also fully manageable from the dashboard (new **Routing** page) and
over `/v1/routing/policies`, so nothing here requires editing a file and
restarting.

## Design decisions worth reviewing

**Two axes, deliberately separate.** `select` decides where the plan starts;
`on_failure` is what runs after a failure. One combined list would make "did this
entry not apply, or did it fail?" ambiguous in every log line. The fallthrough is
an explicit `default` rather than a positional last entry, so a misordered policy
is refused at load instead of silently carrying dead rules.

**`instance` and `provider` are carried separately** through the walker. Pricing,
budgets, and the usage log key on the instance, while any-llm dispatches against
the implementation. Collapsing them would silently re-key `usage_logs.provider`
and break pricing for every named-instance deployment, and it would ship green
because the existing test does not cover the fallback path.

**`absorbed` is a third outcome, not an error.** A recovered attempt writes its own
usage row correlated by `request_group_id`, with `status="absorbed"`. Every error
metric in the product counts `status == "error"` exactly, and `request_count`
excludes absorbed rows, so a working fallback chain cannot read as an outage. This
is why `lib/overview.ts`'s 2% amber threshold and the activity timeline stay honest.

**The reservation tops up before each later candidate.** A chain that falls over to
a pricier model would otherwise run against the cheaper model's reservation and
take spend past a cap the gate already approved. A refused top-up stops the chain
rather than overshooting.

**401/403 are terminal locally**, unlike hybrid mode, which retries them because a
workspace key can be rotated upstream. A standalone operator owns both keys, so
failing over would move traffic and spend to another provider and hide the
misconfiguration.

**Every schema model is `extra="forbid"`.** `GatewayConfig` is `extra="ignore"`, so
a typo'd key inside a policy would otherwise vanish and the policy would quietly
not do what it says.

**A budget threshold at or above 100 is refused at load.** The budget gate rejects
the request before selection, so such a rule could never fire; an operator writing
one believes they have configured "keep serving on a cheaper model after the budget
runs out" when they have configured nothing.

## Verified against a live gateway, not only tests

Real OpenAI traffic through a real gateway, because several of these claims are
only checkable end to end:

| Check | Result |
| --- | --- |
| Failover against a real provider | 404 head, fallback served, caller saw 200 |
| Two correlated rows | `absorbed` attempt 1/2 + `success` attempt 2/2, same group id |
| Every candidate fails | both rows written plus a clean 502 (this used to write nothing) |
| Absorbed excluded from counts | 12 rows, 2 absorbed, `request_count: 10`, `error_count: 5` |
| Conditional selection | per-user targets with `reason=condition:user_id` |
| Catalog | policies listed, dynamic ones as `pricing_source: "dynamic"` with no price |
| Guardrail, no service configured | served, check recorded as skipped |

That exercise caught three real bugs that the test suite had missed, including
stored policies being absent from `/v1/models` (the existing test covered only
config policies) and an exhausted chain writing no usage row at all.

## Not included

- The kNN router (#187). `select: [{router: knn}]` validates, and currently warns
  and falls through to the default rather than pretending.
- Policy-shaped allow-list grants, so "this key may only use policy `fast`" is not
  yet expressible.
- Output-direction guardrails, which the policy schema deliberately refuses rather
  than accepting as a silent no-op.
- Hybrid mode refuses a policy name with a 400: the platform resolves the model
  there, so a local policy name would reach it as an unknown model.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Closes #463. Delivers the observability half of #401.

## Checklist

## Checks

`make lint`, `make typecheck`, `make openapi-check`, `make postman-check`, 1222
unit, 964 integration, 442 dashboard tests, and the committed bundle rebuilt.

Migrations: `e8b1d3f5a7c9` (routing_policies) and `f4c6a8b0d2e5` (usage
attribution), both verified up and down against PostgreSQL.

## Review round

An independent review found six real defects, all now fixed in `84d62d3c` with
the tests that would have caught them. The most serious: policy guardrails were
never enforced, so the third item in this PR's title was a no-op with a full
supporting cast of schema, API, CLI output, dashboard editor, and documentation
asserting otherwise. Also fixed: failover reached only `/v1/chat/completions`;
streamed requests recorded no attribution; a terminal failure was attributed to a
candidate that was never called; `require_pricing` was bypassable through a
fallback; and `limits` was a knob that could not take effect, so it was removed
rather than shipped.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

Written by Claude through back-and-forth with @njbrake, who made the design
calls: value-first sequencing, the `routing policies` name, no `RoutingPort` for
now, one usage row per attempt with a distinct `absorbed` outcome, refusing policy
names in hybrid mode, and the `on_failure` / explicit `default` key naming.

The implementation was reviewed by a separate agent with no knowledge of how it
was built, which is what caught the six defects above; the verification was also
run against a live gateway with real provider traffic, not only against tests.

## Summary

- Added named routing policies with conditional model selection, budget tier-down, retry failover, guardrails, and usage attribution.
- Preserved aliases as a single-target routing shorthand and moved stored aliases into routing policies.
- Added policy management through configuration, CLI, API, dashboard, and documentation.
- Added policy caching, migrations, validation, model catalog support, and comprehensive tests.
- Updated activity and usage views to show recovered attempts as `absorbed` instead of successful requests.
- Replaced the dashboard Aliases page with the Routing page.

## Technical notes

- Routing policies are available in standalone mode only. Hybrid mode rejects policy names.
- Guardrails fail closed by default when unavailable. `on_unavailable: monitor` enables monitoring behavior.
- The implementation includes an attempt walker for ordered provider attempts and accurate attribution.


src/gateway/api/routes/_pipeline.py
```diff
@@ -2297,9 +2819,64 @@ async def run_standalone_non_stream(
...
+                    base_request_fields=base_request_fields,
+                    run_attempt=_run_candidate,
+                    max_tool_iterations=tool_ctx.max_tool_iterations,
+                    policy_name=ctx.plan.policy_name,
+                    build_kwargs=adapter.local_attempt_kwargs,
+                    on_absorbed=_absorbed,
+                    on_terminal=stopped_on.append,
+                )
+            except HTTPException as exhausted:
+                await log_exhausted_plan(ctx, adapter, exhausted, stopped_on[0] if stopped_on else None)
```
khaledosman, line 2870: A refused reservation top-up is attributed to a candidate that was never called.

`top_up_reservation_for_attempt` raises `HTTPException` from inside `run_attempt`, and `walk_attempts` re-raises it at `_attempts.py:184` *without* calling `on_terminal`. So `stopped_on` is empty here and `log_exhausted_plan` falls back to `ctx.plan.attempts[-1]` (`:2742`).

With plan `[A, B, C]`, A failing retryably and B's top-up refused, the 402 row records `model=C`, `provider=C.instance`, `attempt_position=3`: a provider that was never dispatched, at a position the request never reached. That row feeds the by-provider breakdown and the error taxonomy, so it is exactly the misattribution class the PR body says was fixed for the walker path.

Same at `:2288` for the streaming path.

Fix options: record the attempt in `_run_candidate`/`_open_candidate` before letting the top-up exception escape, or have the walker treat an `HTTPException` out of `run_attempt` as terminal-on-this-attempt and call `on_terminal` for it.

src/gateway/services/routing/compiler.py
```diff
@@ -0,0 +1,263 @@
...
+                provider=resolved.provider,
+                model=resolved.model,
+                kwargs=resolved.kwargs,
+                display_model=policy_name,
+                selection_reason=selection_reason,
+            )
+        )
+
+    if not attempts:
+        raise NoEligibleCandidatesError(policy_name, dropped, status_code=403)
```
khaledosman: Every empty plan becomes a 403 that tells the caller model access filtered it, but `dropped` can be entirely `unresolvable`.

That is the case the comment at `:196` anticipates ("the provider set changed under a running gateway"). An operator who deletes a provider instance then gets a 403 whose `caller_detail` (`:65`) states "Every candidate was filtered out before dispatch by model-access rules", which is false and points them at the wrong config.

`status_code` is already a constructor parameter but only ever receives 403, so the flexibility is there unused. Suggest deriving it: all-`unresolvable` should be a 5xx with wording about provider configuration, and `caller_detail` should not name a filter that didn't run.

tests/integration/test_routing_policies.py
```diff
@@ -0,0 +1,1094 @@
...
+    assert "gpt-5-mini" not in detail
+    assert "claude-haiku" not in detail
+
+
+# ---------------------------------------------------------------------------
+# Budget-conditional selection
+# ---------------------------------------------------------------------------
+
+
+def test_tier_down_fires_once_the_budget_threshold_is_crossed(
```
khaledosman, line 426: Nothing in `tests/` exercises `top_up_reservation_for_attempt`. Grepping for it, for `budget_exhausted_mid_failover`, or for `increase_reservation` in this file returns nothing.

That function is the only thing preventing a fallover to a pricier candidate from spending past a cap the gate already approved, which is one of the four headline design decisions in the PR body. It also carries the `require_pricing` gate that the same body lists as one of the six defects found and fixed. A defect found once and fixed without a regression test comes back.

Three cases worth covering, all reachable from this fixture:
- cheaper fallback candidate: top-up is a no-op, chain proceeds
- pricier fallback candidate: reservation grows before dispatch
- refused top-up: chain stops with 402 and `budget_exhausted_mid_failover_detail`, original hold refunded, plus the attribution question raised on `_pipeline.py:2870`

alembic/versions/b5d7f9a1c3e6_move_stored_aliases_into_routing_policies.py
```diff
@@ -0,0 +1,135 @@
...
+                "id": str(uuid.uuid4()),
+                "name": row["name"],
+                "spec": _spec_for(row["target"]),
+                "user_id": row["user_id"],
+                "created_at": row["created_at"],
+                "updated_at": row["updated_at"],
+            },
+        )
+
+    conn.execute(sa.text("DELETE FROM model_aliases"))
```
khaledosman, line 96: Two things about this line.

**It is an undocumented breaking change to `/v1/aliases`.** After the move, for an alias created through the API: `GET /v1/aliases` no longer lists it, `DELETE /v1/aliases/{name}` 404s, and re-creating it with `POST /v1/aliases` now 400s ("already a routing policy", `aliases.py:99`). `docs/routing.md:16` says "`/v1/aliases` remains as the one-target API", which is true of the endpoint but not of existing rows, and the PR body doesn't mention the migration at all. The dashboard is fine because `RoutingPage` reads both stores, but any script driving `/v1/aliases` breaks on upgrade. This belongs in the PR description and a release note.

**It is the only destructive migration here and it has no test.** The other two are pure DDL and need none; this one moves rows, deletes the source table's contents, and has a deliberately lossy downgrade. Worth an integration test that seeds a global alias and a user-scoped one, runs upgrade, asserts both resolve as policies and both rows are gone, then downgrades and asserts they come back with scope and timestamps intact.

Minor: the downgrade docstring's "left in place rather than silently flattened" is moot in a full downgrade, since `e8b1d3f5a7c9` drops the table on the next step, so a multi-target policy is lost either way.

khaledosman (commented):
Reviewed the full diff (backend, migrations, dashboard, docs). The routing design holds up and the usage-accounting side is unusually careful: `_row_status`, `_request_count_expr` and `_breakdown`'s `residual_requests` all agree, so the `other` row still reconciles with the tiles once a request can write more than one row. That is the part of an "extra rows per request" change that normally breaks silently, and it didn't.

One confirmed correctness bug (attribution on a refused reservation top-up), one wrong status/message pairing, and a test gap on the only code path that stops a fallover from overshooting a budget cap. Details inline, most substantive first.

Also worth adding to the PR description and the release notes: migration `b5d7f9a1c3e6` is a breaking change to `/v1/aliases` for existing rows. It is documented in `docs/routing.md:16`, but the PR body doesn't mention it at all.

src/gateway/api/routes/usage.py
```diff
@@ -642,6 +659,24 @@ def _dialect_name(db: AsyncSession) -> str:
     return bind.dialect.name
 
 
+def _request_count_expr() -> Any:
```
khaledosman: `_request_count_expr()` returns 0 when the caller filters to the status it excludes. `/v1/usage/summary?status=absorbed` reports `request_count: 0` alongside non-zero cost and tokens, and the same applies to `usage_series` and every `_breakdown`.

The Activity page now offers an "Absorbed" filter option (`ActivityPage.tsx:104`), so an operator who spots an absorbed row and filters to the rest of them will hit this on the Usage page and read it as broken tiles. Either fall back to `count()` when the conditions already pin `status`, or have the summary surfaces state that absorbed rows are counted as attempts rather than requests.

src/gateway/api/routes/models.py
```diff
@@ -249,6 +251,64 @@ def _alias_target_keys(config: GatewayConfig, aliases: dict[str, str]) -> set[st
...
+    dynamic: dict[str, PolicySpec] = {}
+    for name, spec in effective_policies(config, caller_user_id).items():
+        if spec.is_dynamic:
+            dynamic[name] = spec
+        else:
+            static[name] = spec.default_target
+    return static, dynamic
+
+
+def _policy_target_keys(config: GatewayConfig, caller_user_id: str | None) -> set[str]:
```
khaledosman, line 298: This withholds every policy candidate from `/v1/models` for every caller, not just the policy's default target. Same rule aliases already had, but the blast radius is larger now that one policy can name five selectors: a real model a caller may still call directly disappears from the catalog because some unrelated policy lists it as a fallback.

The docstring justifies it and it is defensible, but the visible consequence (a directly-callable model vanishing from the catalog) isn't stated anywhere a user would look. Worth a line in `docs/models.md` or `docs/routing.md`.

src/gateway/cli.py
```diff
@@ -226,6 +226,122 @@ def gen_secret_key() -> None:
...
+    Reads config only. No database, no provider call, nothing billed. The budget
+    options let a tier-down rule be exercised without waiting for real spend to
+    cross the threshold.
+    """
+    from gateway.models.routing import PolicySpec
+    from gateway.services.routing import BudgetState, NoEligibleCandidatesError, compile_policy
+
+    cfg = load_config(config)
+    if not cfg.routing.policies:
+        click.echo("No routing policies are configured. Add a `routing.policies` block to config.yml.")
```
khaledosman: `otari routing explain` reads config only, so it cannot see a stored policy. `config.example.yml` advertises this command as the way to check what a policy compiles to, and the dashboard is the primary way policies get created, so an operator whose policies are all stored gets "No routing policies are configured" and exit 1 on a gateway that has several.

The docstring documents the limitation; this message should too, e.g. "No routing policies are configured in config.yml. Policies created through the dashboard or API are stored in the database; use POST /v1/routing/policies/explain or the dashboard's Dry run for those."

src/gateway/api/routes/_attempts.py
```diff
@@ -0,0 +1,245 @@
...
+                reason = "timeout" if error_class == "timeout" else "upstream_error"
+                record_abandoned_attempt(attempt.instance, attempt.model, reason, attempt.position)
+            if locked_in or not retryable:
+                if on_terminal is not None:
+                    on_terminal(attempt)
+                raise _provider_failure_http_exc(exc, fallback_detail="LLM provider error") from exc
+            failures.append(AttemptFailure(attempt.position, attempt.instance, attempt.model, error_class))
+            # Only a failure with somewhere left to go is "absorbed"; the last one is
+            # the request's own outcome and is logged by the caller as an error.
+            if on_absorbed is not None and attempt is not attempts[-1]:
```
khaledosman: Nit: `attempt is not attempts[-1]` is correct (the compiler dedupes, so no object repeats) but reads as an identity trick that a future edit could break. `attempt.position < len(attempts)` says "there is somewhere left to go", which is the actual condition the docstring describes.

src/gateway/api/routes/models.py
```diff
@@ -249,6 +251,64 @@ def _alias_target_keys(config: GatewayConfig, aliases: dict[str, str]) -> set[st
...
+    rather than inventing a second field; ``pricing`` stays null, and a client that
+    does not know the new value sees "unpriced" instead of a fabricated rate.
+    """
+    return ModelObject(
+        id=policy_name,
+        created=0,
+        owned_by=ALIAS_OWNED_BY,
+        pricing=None,
+        pricing_source="dynamic",
+    )
```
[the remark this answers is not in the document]
njbrake: _Note: this reply was drafted by Claude via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

Disagreeing with this one. A dynamic policy model never reaches that mapping.

`_dynamic_policy_model` builds the object with `owned_by=ALIAS_OWNED_BY` (`models.py:267`), and `ModelsPage.tsx:1139` skips those before computing `priceStatus`:

```tsx
if (model.owned_by === ALIAS_OWNED_BY) {
  continue;
}
```

So there is no row for a dynamic policy on the models page, which means no `source: "none"` assignment and no `PriceSource` rendering to break. Policies are managed on the Routing page, not the catalogue.

`pricing_source: "dynamic"` is still meaningful over the API, which is where it is for: a client reads "unpriced, and deliberately so" instead of a fabricated rate.

src/gateway/static/dashboard/assets/OverviewPage-CTPTo-KN.js
```diff
@@ -0,0 +1 @@
...i-muted)]",title:new Date(r.timestamp).toLocaleString(),children:ht(r.timestamp)})},{id:"model",header:"Model",isRowHeader:!0,cell:r=>t.jsx("span",{className:"text-[var(--otari-ink)]",children:r.model})},{id:"cost",header:"Cost",align:"end",cell:r=>r.cost===null?"—":D(r.cost)},{id:"status",header:"Status",cell:r=>t.jsx("span",{className:`inline-flex items-center rounded-full border px-2 py-0.5 text-xs font-medium ${r.status==="error"?"border-red-200 bg-red-50 text-red-700":"border-[var(--otari-line)] bg-[var(--otari-brand-tint)] text-[var(--otari-brand-dark)]"}`,children:Ct(r.status)})}];return t.jsxs("div",{className:"flex flex-col gap-3",children:[t.jsxs("div",{className:"flex items-center justify-between",children:[t.jsx("h2",{className:"text-sm font-semibold text-[var(--otari-ink)]",children:"Recent activity"}),t.jsx(Y,{to:"/activity",className:"text-sm text-[var(--otari-brand-dark)] hover:underline",children:"View all →"})]}),t.jsx(V,{error:a}),t.jsx(xt,{ariaLabel:"Recent activity",columns:l,rows:e,getRowKey:r=>r.id,isLoading:n,emptyContent:"No requests yet. Once the gateway serves traffic, it appears here."})]})}export{Bt as OverviewIndex,Rt as OverviewPage,T as localDayKey};
```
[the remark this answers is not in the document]
njbrake: _Note: this reply was drafted by Claude via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

Agreed, and confirmed in the source rather than the bundle: `OverviewIndex` derives `needsSetup` from the provider query (`OverviewPage.tsx:83`), but `loadError` at line 137 aggregates only `today`, `period`, `health`, `budgets`, `keys`, and `users`. A failed provider call leaves `isSuccess` false, so the setup block is skipped and nothing is reported.

Out of scope for this PR though. `git diff origin/main...HEAD -- web/src/pages/OverviewPage.tsx` touches only the `absorbed` status word and its pill styling; this code is pre-existing and surfaced here because the bundle asset hash changed.

Filed as #499.

src/gateway/static/dashboard/assets/ProvidersPage-dKyb-Stc.js
```diff
@@ -0,0 +1 @@
...iting config.yml. Keys are encrypted at rest.",action:b||A?null:e.jsx(j,{variant:"primary",isDisabled:!P,onPress:()=>{u(null),x(!0)},children:"Add provider"})}),e.jsx(I,{error:t.error??s.error??a.error??n.error??d.error??i.error}),P?null:e.jsxs(ne,{tone:"warning",children:[e.jsx("code",{children:"OTARI_SECRET_KEY"})," is not set, so provider keys can't be encrypted at rest and adding providers from the dashboard is disabled. Set it on the server and restart to add providers here. Providers defined in"," ",e.jsx("code",{children:"config.yml"})," keep working without it."]}),A?e.jsx(Ce,{onAddProvider:()=>{u(null),x(!0)},needsPricing:N,onEnablePricing:()=>d.mutate({default_pricing:!0}),enabling:d.isPending,secretKeyConfigured:P}):null,b&&P?e.jsx(ye,{onClose:()=>x(!1)}):null,y?e.jsx(ke,{provider:y,onClose:()=>u(null),onSaved:K}):null,!o&&l.length>0&&n.data&&n.data.total>0?e.jsx(Se,{healthy:n.data.healthy,degraded:n.data.degraded,total:n.data.total,checkedAt:n.data.checked_at}):null,A?null:e.jsx(me,{ariaLabel:"Providers",columns:G,rows:l,getRowKey:r=>r.instance,isLoading:o,emptyContent:"No providers yet. Add your first provider to start serving models."})]})}export{Ke as ProvidersPage};
```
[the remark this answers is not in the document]
njbrake: _Note: this reply was drafted by Claude via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

Agreed. The source is `ProvidersPage.tsx:750`, a raw `<a href="/welcome">` in a HashRouter app, so it navigates the server path and leaves the router hash empty.

Out of scope for this PR: that file is untouched by the diff, and it is flagged here only because the bundle asset hash changed.

Filed as #500.

src/gateway/static/dashboard/assets/SettingsPage-ByleVDJK.js
```diff
@@ -0,0 +1 @@
...[var(--otari-muted)]",children:[e.jsx("input",{type:"checkbox",checked:o,onChange:l=>c(l.target.checked),className:"h-4 w-4 accent-[var(--otari-brand)]"}),"Settable only"]})]}),s?e.jsxs("p",{className:"text-xs text-[var(--otari-muted)]",children:["Showing ",g.length," of ",x.length," settings"]}):null,s&&g.length===0?e.jsx("p",{className:"text-sm text-[var(--otari-muted)]",children:"No settings match your search."}):null,t.isLoading?e.jsx(E,{}):null,k.map(l=>e.jsxs("section",{className:"flex flex-col gap-2",children:[e.jsxs("h2",{className:"text-sm font-semibold text-[var(--otari-ink)]",children:[l.name," ",e.jsxs("span",{className:"font-normal text-[var(--otari-muted)]",children:["(",l.fields.length,")"]})]}),e.jsx(v,{children:e.jsx(v.Content,{className:"flex flex-col divide-y divide-[var(--otari-line)] px-5 py-1",children:l.fields.map(f=>e.jsx(U,{field:f,patch:m,disabled:!s||a},f.key))})})]},l.name)),s?e.jsx(Z,{}):null,s?e.jsx(Q,{masterKeySource:s.master_key_source}):null,s?e.jsxs("p",{className:"text-xs text-[var(--otari-muted)]",children:["Mode: ",s.mode," · Version ",s.version,s.require_pricing?" · require_pricing on":""]}):null]})}export{ne as SettingsPage,M as fieldMatches};
```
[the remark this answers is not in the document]
njbrake: _Note: this reply was drafted by Claude via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

Disagreeing with this one. Read from the source (`SettingsPage.tsx:419`) rather than the bundle, the close is blocked, not completed with stale state:

```tsx
const onOpenChange = (isOpen: boolean) => {
  if (isOpen) {
    rotateMasterKey.reset();
    setDialogOpen(true);
  } else if (newKey === undefined) {
    closeDialog();
  }
};
```

`isOpen` is a controlled prop (`<AlertDialog isOpen={dialogOpen} ...>`). Once a key has been generated, `newKey !== undefined`, so the `else if` does not fire and `dialogOpen` is never set false. The dialog stays open. Backdrop and Escape cannot dismiss it, which is the deliberate guard: the key is revealed once, so it must not be dismissable before it has been copied. The only way out is the explicit Close button, which calls `closeDialog()` and clears `newKey`.

So there is no close path that leaves `newKey` set, and no reopen that could pass the old key back. The analysis assumed `m(false)` closes the dialog, which is true for an uncontrolled dialog but not for this one.

Also pre-existing: `SettingsPage.tsx` is untouched by this PR's diff.

alembic/versions/b5d7f9a1c3e6_move_stored_aliases_into_routing_policies.py
```diff
@@ -0,0 +1,135 @@
...
+                {
+                    "id": str(uuid.uuid4()),
+                    "name": row["name"],
+                    "target": select[0]["default"],
+                    "user_id": row["user_id"],
+                    "created_at": row["created_at"],
+                    "updated_at": row["updated_at"],
+                },
+            )
+        conn.execute(sa.text("DELETE FROM routing_policies WHERE id = :id"), {"id": row["id"]})
```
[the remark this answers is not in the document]
njbrake: _Note: this reply was drafted by Claude via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

Partly disagreeing, and the concrete example does not apply.

`limits` cannot appear in a stored spec. `PolicySpec` is `extra="forbid"`, so a document carrying it is rejected at write time and at startup. There is no lossy shape for the check to miss.

Converting a one-target policy into an alias on downgrade is deliberate rather than an oversight. A rolled-back binary does not read `routing_policies`, so leaving such a policy in place is what loses it. Writing the alias is what preserves the behavior, which is why the downgrade converts instead of skipping. Provenance tracking would make a policy created after the upgrade vanish on rollback, which is worse than what happens today.

One real gap you and @khaledosman both landed on: the clash branch deleted the policy without writing the alias, dropping the target entirely. Both directions now abort naming the conflict instead. Fixed in b9592804, with `tests/unit/test_alias_to_policy_migration.py` covering the move, scope and timestamp preservation, both clash refusals, and the chained-policy downgrade skip.

web/e2e/dashboard.spec.ts
```diff
@@ -109,20 +109,40 @@ test.describe("dashboard core flows", () => {
...
     // close the popover so it does not aria-hide the submit button.
-    await page.getByRole("combobox", { name: /Target/ }).fill("openai:gpt-4o");
+    await page.getByRole("combobox", { name: /Serves/ }).fill("openai:gpt-4o");
     await page.keyboard.press("Escape");
-    await page.getByRole("button", { name: "Create alias" }).click();
+    await page.getByRole("button", { name: "Create policy" }).click();
 
-    // The alias name is the table's row-header cell (react-aria rowheader).
+    // The policy name is the table's row-header cell (react-aria rowheader).
     await expect(page.getByRole("rowheader", { name: "fast" })).toBeVisible();
```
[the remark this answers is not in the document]
njbrake: _Note: this reply was drafted by Claude via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

Agreed on the assertion scoping, fixed in b9592804:

```ts
const chained = page.getByRole("row").filter({ has: page.getByRole("rowheader", { name: "chained" }) });
await expect(chained).toBeVisible();
await expect(chained).toContainText(/\+1 on failure/);
```

A `chained` saved without its fallback now fails, where the page-wide text match would have passed.

Not adding per-test isolation: `web/e2e/serve.sh` already deletes `e2e.db` and its WAL sidecars before starting the gateway, and `playwright.config.ts` disables retries and always starts a fresh server precisely so a stray instance cannot skip that reset. Each run starts empty, so the only policies present are the ones these tests create.

src/gateway/api/routes/_pipeline.py
njbrake: _Note: this reply was drafted by Claude via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

This one was already fixed in cf13d185, which landed after the commit this review was against (47300a0c), so the code you read is not current. Confirming rather than asserting it:

`walk_attempts` now reports on the `HTTPException` path (`_attempts.py`):

```python
except HTTPException:
    if on_terminal is not None:
        on_terminal(attempt)
    raise
```

and both call sites wire it up, so `stopped_on` is populated for a refused top-up:

```python
on_terminal=stopped_on.append,
...
except HTTPException as exhausted:
    await log_exhausted_plan(ctx, adapter, exhausted, stopped_on[0] if stopped_on else None)
```

Streaming (`:2288`) and non-streaming (`:2870`) both do this. The `MaxToolIterationsExceeded` path reports too, since it is terminal for the same reason. Regression tests: `test_a_gateway_side_refusal_reports_the_candidate_it_happened_on` and `test_the_tool_iteration_cap_reports_the_candidate_it_happened_on`.

The analysis was right about the consequence, and it is the reason the fix exists. Thanks for independently landing on it.

src/gateway/services/routing/compiler.py
njbrake: _Note: this reply was drafted by Claude via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

Agreed, and fixed in 566cf78b. Copilot flagged the same thing independently, which was a good signal it was not a corner case.

The status is now derived rather than hardcoded, because the two causes are different faults with different fixes:

```python
unresolvable_only = bool(dropped) and all(item.reason == "unresolvable" for item in dropped)
self.status_code = 502 if unresolvable_only else 403
```

502 with "None of its candidates resolve to a configured provider" for the all-unresolvable case, 403 with "Its candidates were all filtered out before dispatch" otherwise. Targets still stay out of `caller_detail`; the enumeration stays in `operator_detail`.

A mixed drop stays 403 on purpose: if one candidate is a real model the caller may not use, asking for access is the fix, and a 502 would tell them to wait for an operator instead.

New `tests/unit/test_routing_compiler.py` covers all three, including that the wrong config is not named in either message. Worth noting `compile_policy` had no direct unit tests before this; that file is now the place for them.

tests/integration/test_routing_policies.py
njbrake: _Note: this reply was drafted by Claude via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

Correct, and the grep result was worse than the comment implies: nothing anywhere in `tests/` referenced `top_up_reservation_for_attempt`. The only nearby coverage was of `increase_reservation` itself, which is the primitive rather than the failover decision on top of it.

Added `tests/unit/test_pipeline_failover_topup.py` in 566cf78b, covering the three cases you named plus the gate's edges:

- pricier candidate grows the hold, and by the delta rather than the full price
- cheaper candidate is a no-op, and so is an equally priced one
- refused top-up raises 402 carrying `budget_exhausted_mid_failover_detail()`
- unpriced fallback refused under `require_pricing`, allowed with it off
- budget-exempt request skips the pricing gate, matching the admission-time rule
- no reservation and no estimate inputs are both no-ops rather than crashes

The 402 attribution question is answered separately on the `_pipeline.py:2870` thread: already fixed in cf13d185, which post-dates the commit this review was against.

alembic/versions/b5d7f9a1c3e6_move_stored_aliases_into_routing_policies.py
njbrake: _Note: this reply was drafted by Claude via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

All three points taken, fixed in b9592804 and 566cf78b.

**The `/v1/aliases` break is now documented.** `docs/routing.md` carries it as a callout listing the exact symptoms: `GET` stops listing a pre-upgrade alias, `DELETE` 404s, re-`POST` 400s because the name is a policy now, the dashboard is unaffected because it reads both stores, and a code-only rollback leaves the rows unreadable. Also filed #503 for the larger question Copilot raised from the other end: `POST /v1/aliases` still writes to `model_aliases`, so the second store is drained rather than retired. That is a product decision (deprecate the endpoint, reroute it at policies, or document both stores as permanent), not something to settle inside this PR.

**The migration now has tests.** `tests/unit/test_alias_to_policy_migration.py` drives `upgrade()` and `downgrade()` against SQLite with `op.get_bind` pointed at the connection, so the data behavior is pinned without a Postgres fixture: the move, scope and timestamp preservation for a user-scoped alias, the round trip, the chained-policy downgrade skip, and both clash refusals.

On clashes, you and CodeRabbit found the same thing and the downgrade was the sharper half: it deleted the policy without writing the alias, so the target was dropped entirely. Both directions now abort naming the conflict, since which of the two the operator meant is not knowable from inside a migration.

**The docstring claim was wrong and is gone.** You are right that "left in place rather than silently flattened" is moot in a full downgrade, because `e8b1d3f5a7c9` drops the table on the next step. It now says so, and says that stopping at `f4c6a8b0d2e5` is what makes the alias round trip lossless.

src/gateway/api/routes/usage.py
njbrake: _Note: this reply was drafted by Claude via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

Confirmed and fixed in 566cf78b. `_request_count_expr()` was `sum(case((status != "absorbed", 1), else_=0))`, so a query pinned to `status=absorbed` scored every row zero.

Took the first of your two options: the expression now takes the caller's status filter and counts rows when that filter is `absorbed`, on the reasoning that filtering *to* the attempts makes them the unit being asked about. Threaded through all four sites, so `usage_series`, every `_breakdown`, and the CSV export agree with the tiles instead of contradicting them.

Regression test in `tests/integration/test_routing_policies.py`: `test_filtering_to_absorbed_counts_the_attempts_rather_than_reporting_zero` asserts the filtered total is 1, the unfiltered total is still 1 (the served row), and the by-provider breakdown sums to the tile.

src/gateway/api/routes/models.py
njbrake: _Note: this reply was drafted by Claude via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

Agreed that the consequence was undocumented. Added to `docs/models.md` in 566cf78b, in the section that already explains alias withholding, and it states the surprising part plainly: a policy can name up to five selectors and every one is withheld, so a model you can still call directly can disappear from `GET /v1/models` because an unrelated policy lists it as a fallback. It also says the catalogue is therefore not a complete inventory of what is callable, and points at `otari routing explain` and the Routing page for the operator view.

Keeping the behavior. The listing is the wrong place to publish where a policy sends traffic, and narrowing the rule to the default target would leak the failure chain, which is the part most worth not advertising.

src/gateway/cli.py
njbrake: _Note: this reply was drafted by Claude via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

Agreed, fixed in 566cf78b. The message now reads:

> No routing policies are configured in config.yml. Add a `routing.policies` block there, or, if your policies were created through the dashboard or the API, note that this command reads config only: it has no database. Use `POST /v1/routing/policies/explain` against a running gateway to compile a stored policy.

One change from your suggested wording: it does not mention the dashboard's Dry run, because that panel was cut from the Routing page before merge. The endpoint that backed it is still there, which is what the message points at.

src/gateway/api/routes/_attempts.py
njbrake: _Note: this reply was drafted by Claude via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

Fair, and taken in 566cf78b: `attempt.position < len(attempts)`. It states the condition the docstring describes rather than relying on the compiler's dedup keeping object identity meaningful.

src/gateway/services/routing/compiler.py
```diff
@@ -0,0 +1,263 @@
...
+
+    def __init__(self, policy_name: str, dropped: list[DroppedCandidate], status_code: int) -> None:
+        self.policy_name = policy_name
+        self.dropped = dropped
+        self.status_code = status_code
+        self.caller_detail = (
+            f"Routing policy '{policy_name}' has no usable candidate for this request, so it cannot be "
+            "served. Every candidate was filtered out before dispatch by model-access rules. Ask your "
+            "operator to check the policy."
+        )
```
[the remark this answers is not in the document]
njbrake: _Note: this reply was drafted by Claude via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

Agreed, fixed in 566cf78b. @khaledosman flagged the same thing independently, which was a good signal it was not a corner case.

`caller_detail` is now reason-derived rather than asserting model access, and the status comes with it: 502 plus "None of its candidates resolve to a configured provider" when every drop is `unresolvable`, 403 plus "Its candidates were all filtered out before dispatch" otherwise. Detailed per-candidate reasons stay in `operator_detail`, and targets stay out of the caller string.

`duplicate` and `over_cap` cannot empty a plan on their own (a dedup or a cap only fires where a survivor already exists), so they fall in the 403 wording alongside `not_allowed`. Covered by `tests/unit/test_routing_compiler.py`.

src/gateway/services/routing/compiler.py
```diff
@@ -0,0 +1,263 @@
...
+        if entry.default is not None:
+            return entry.default, "default"
+        if entry.router is not None:
+            # Router backends are not wired yet. Falling through to the default
+            # is the safe reading: a router is an optimization, and it must never
+            # be the reason a request cannot be served.
+            logger.warning(
+                "Routing policy names router '%s', which is not available yet; using the default target",
+                entry.router,
+            )
```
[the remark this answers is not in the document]
njbrake: _Note: this reply was drafted by Claude via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

Agreed, fixed in 566cf78b with the warn-once approach. Throttling per `(policy, router)` rather than per router alone, so suppression cannot hide a second misconfigured policy behind the first:

```python
if _warn_missing_router(policy_name, entry.router):
    logger.warning(
        "Routing policy '%s' names router '%s', which is not available yet; using the default "
        "target. This is logged once per router per process.",
        policy_name, entry.router,
    )
```

The set is keyed off policy documents, so it is bounded by config size rather than traffic. The message now also names the policy, which it did not before, and says the logging is throttled so nobody reads a single line as a single occurrence.

Two tests in `tests/unit/test_routing_compiler.py`: three compiles of one policy produce one warning, and two policies naming the same router produce two.

src/gateway/api/routes/aliases.py
```diff
@@ -94,6 +95,18 @@ def _validate(config: GatewayConfig, name: str, target: str, user_id: str | None
...
     # so this is for consistency rather than to close a hole.
+    # Symmetry with the policy write path, which already refuses a policy named
+    # after an alias. Without this, creating an alias named after an existing
+    # policy would silently shadow it: alias resolution runs first, so the policy
+    # would stop taking effect and nothing would say so.
+    if name in all_policy_names(config):
+        raise HTTPException(
+            status_code=status.HTTP_400_BAD_REQUEST,
+            detail=(
+                f"'{name}' is already a routing policy. An alias resolves before a policy, so this would "
```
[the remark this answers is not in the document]
njbrake: _Note: this reply was drafted by Claude via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

Right on both halves, and this is the sharpest framing of it: the migration drained the second store rather than retiring it, and `POST /v1/aliases` refills it.

Doing the second of your two options here, and the first as a follow-up decision:

- `docs/routing.md` now carries the break as an explicit callout: for a pre-upgrade alias, `GET` stops listing it, `DELETE` 404s, and re-`POST` 400s. It also notes the dashboard is unaffected because it reads both stores, and that a code-only rollback leaves the moved rows unreadable.
- #503 tracks whether `/v1/aliases` should be rerouted at `routing_policies`, deprecated, or documented as a permanent second store, with the tradeoffs written out. Recommendation there is deprecate then reroute, so an existing endpoint does not silently change what it writes mid-release.

Worth noting for anyone reading this thread: the two write paths already refuse each other's names (`aliases.py` rejects a name that is a policy, and the policy route rejects a name that is an alias), so the stores cannot shadow each other today. The cost is conceptual plus the upgrade break, not a live resolution bug.

web/src/pages/ActivityPage.tsx
```diff
@@ -866,6 +880,30 @@ export function ActivityPage() {
...
+        cell: (e) =>
+          e.policy_name == null ? null : (
+            <span className="flex flex-col leading-tight">
+              <span className="text-[var(--otari-ink)]">{e.policy_name}</span>
+              <span className="text-xs text-[var(--otari-muted)]">
+                {/* Non-color signal: the position is spelled out, so the amber row
+                    tint is never the only thing carrying the meaning. */}
+                {e.attempt_position != null && e.attempt_count != null && e.attempt_count > 1
+                  ? `attempt ${e.attempt_position}/${e.attempt_count} · ${e.selection_reason ?? ""}`
+                  : (e.selection_reason ?? "")}
```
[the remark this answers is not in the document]
njbrake: _Note: this reply was drafted by Claude via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

Confirmed and fixed in 566cf78b. Built by joining the present parts instead of interpolating both:

```tsx
{[
  e.attempt_position != null && e.attempt_count != null && e.attempt_count > 1
    ? `attempt ${e.attempt_position}/${e.attempt_count}`
    : null,
  e.selection_reason,
]
  .filter(Boolean)
  .join(" · ")}
```

Three tests added to `ActivityPage.test.tsx`: position with no reason renders `attempt 1/2` and asserts no trailing separator, both present renders `attempt 2/2 · on_failure`, and a single-candidate policy renders the reason alone.

PR #496, opened 2026-08-05
feat(hybrid): forward provider-specific extra_params through resolved attempts

tbille:
## Description

Hybrid-mode requests routed to AWS Bedrock always failed with botocore's `NoRegionError` ("You must specify a region.") because the hybrid wire contract had no way to carry a provider-specific credential field like Bedrock's mandatory `region_name` from the platform's resolve response through to the `acompletion()` call.

Root cause: `ResolvedAttempt` / `default_attempt_kwargs()` (`src/gateway/api/routes/_platform.py`) only ever forwarded `api_key` and an optional `api_base`. Standalone mode already supports arbitrary passthrough kwargs (`providers.bedrock.region_name` in `config.yml` via `get_provider_kwargs()`), but hybrid mode's per-request credential resolution had no equivalent. Any-llm's `BedrockProvider._init_client` builds its boto3 client with no default region, so a Bedrock attempt was unusable through hybrid mode on both a self-hosted gateway and mozilla.ai's own default gateway, regardless of how the operator's own AWS credentials were configured.

**Update:** the first commit's fix (merging `extra_params` flat into the completion kwargs) turned out not to actually work against the real any-llm SDK — see "Follow-up fix" below. The second commit corrects it and also adds support for AWS Bedrock's bearer-token ("Bedrock API key") credential shape, which the first commit's docs incorrectly described as unrepresentable over the wire.

### What this PR does

- `ResolvedAttempt` gains `extra_params: dict[str, str] | None`, parsed from both the multi-attempt and legacy single-attempt resolve payload shapes in `_parse_resolve_payload`.
- `default_attempt_kwargs()` merges `extra_params` (nested under `client_args`, see below) in after the caller's own request fields and before the forced `model` key, so it can never be shadowed by a same-named field in the request body: the same non-overridable treatment `api_key`/`model` already get.
- `responses.py`'s own `attempt_kwargs` override (it doesn't delegate to `default_attempt_kwargs`) gets the identical treatment.
- `docs/hybrid-mode-protocol.md`: documents the new field and its non-overridable precedence.

### Follow-up fix (second commit): `client_args`, not flat kwargs

any-llm's `acompletion()` only forwards a separate `client_args` mapping to the provider's client constructor (`AnyLLM.create(provider, api_key=..., api_base=..., **client_args)`); everything else in `**kwargs` goes to the completion *call* instead. The first commit merged `extra_params` flat, so `region_name`/`aws_access_key_id` landed in the completion call and never reached `boto3.client()` — feeding that commit's own `default_attempt_kwargs()` output straight into the real SDK still raised `NoRegionError`. All the original tests passed only because they monkeypatched `acompletion()`/`aresponses()` directly, which never exercises any-llm's own kwarg-splitting.

The fix:
- `default_attempt_kwargs()`/`responses.py` now nest `extra_params` under `client_args`.
- New `src/gateway/services/bedrock_gateway_auth.py` adds Bedrock-specific handling on top of that, since any-llm's `BedrockProvider` never reads a plain `api_key` when building its boto3 client, and AWS has two credential shapes:
  - Classic IAM access-key/secret-key pair: aliases the secret into `client_args["aws_secret_access_key"]`.
  - Bearer token ("Bedrock API key", no `aws_access_key_id`): builds an unsigned boto3 `bedrock-runtime` client with the `Authorization: Bearer` header injected via a `before-sign` hook (mirroring otari.ai's own in-process `bedrock_bearer_auth.py`), passed as `client_args["client"]`. The pinned boto3 version has no native `AWS_BEARER_TOKEN_BEDROCK`/`aws_bearer_token` support, so this hook is the only way to authenticate this shape today. This corrects the first commit's overly pessimistic doc claim (also flagged by CodeRabbit) that this shape "can't cross the wire": the token itself is a plain string, and the in-process client can be built by whichever process actually calls boto3, which is the gateway for gateway-routed traffic, not the platform.
- New `tests/unit/test_hybrid_client_args_routing.py` exercises the real any-llm SDK (patching only `boto3.client`, not `acompletion`) so this class of bug can't silently regress again; it fails against the first commit's implementation and passes against the second.

This is one half of a two-repo fix. The platform side (populating `extra_params` in the `/gateway/provider-keys/resolve` response as a plain passthrough for both credential shapes, no secret aliasing) is mozilla-ai/otari-ai#1534.

## PR Type

- [x] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Companion PR: mozilla-ai/otari-ai#1534

## Checklist

- [x] If the API contract changed, I regenerated the OpenAPI spec (`uv run python scripts/generate_openapi.py`). N/A: `ResolvedAttempt`/`ResolvedRoute` are internal types for the hybrid-mode platform client, not exposed on any gateway-facing route; `make openapi-check` passes unchanged.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:** OpenCode (Claude Sonnet 5), driven from a shared session that also implemented the companion mozilla-ai/otari-ai#1534 change.

Root-caused from a pasted gateway error log ("All streaming attempts failed ... You must specify a region."), by tracing through the hybrid-mode streaming fallback path, any-llm's Bedrock provider, and a sibling in-progress worktree of the otari-ai platform repo that was adding Bedrock provider-key support but hadn't yet wired it through the gateway-resolve response. The first commit's fix was validated only against mocked `acompletion` calls; testing it against the real any-llm SDK during a live debugging session (asked to reproduce a user's curl request against a locally running gateway) surfaced that it didn't actually work, which the second commit corrects, along with adding the bearer-token credential shape the user explicitly asked for.

## Summary

- Added provider-specific `extra_params` support for hybrid-mode attempts.
- Forwarded these parameters through `client_args` for chat and Responses requests.
- Added Bedrock support for IAM credentials and bearer tokens.
- Protected platform-supplied client settings from request-field overrides.
- Forwarded HTTP 400 resolve errors to clients.
- Documented the protocol and Bedrock credential changes.
- Added unit and integration tests, including tests with the real any-llm SDK.

### DIFF
--- a/docs/hybrid-mode-protocol.md
+++ b/docs/hybrid-mode-protocol.md
@@ -122,10 +122,15 @@ the caller's own request fields so they can never be shadowed by a
 same-named field in the request body: the same non-overridable treatment
 `api_key` and `model` already get. This field is sourced only from the
 trusted platform peer; Otari never accepts it from the caller. AWS Bedrock's
-bearer-token ("Bedrock API key") credential shape is not representable here,
-since it requires an in-process boto3 client with a custom signing hook, so
-the platform only sends `extra_params` for the classic IAM access-key/secret-key
-shape.
+bearer-token ("Bedrock API key") credential shape is a current gap, not a
+fundamental one: AWS's own recommended path for it is the
+`AWS_BEARER_TOKEN_BEDROCK` environment variable (or, in newer boto3 releases,
+an `aws_bearer_token` constructor kwarg), either of which is in principle
+JSON-serializable and forwardable like any other `extra_params` value. The
+reference platform (otari.ai) currently builds an in-process boto3 client
+with a custom signing hook for that shape instead, which has no
+representable form once serialized over HTTP, so it only sends
+`extra_params` for the classic IAM access-key/secret-key shape today.
 
 `request_id` groups every `attempt_id` from the same resolve call so the
 platform can attribute spend, render trace timelines, and emit fallback events.
--- a/docs/hybrid-mode-protocol.md
+++ b/docs/hybrid-mode-protocol.md
@@ -89,6 +89,19 @@ Content-Type: application/json
       "api_key": "sk-...",
       "api_base": "https://api.openai.com/v1",
       "managed": false
+    },
+    {
+      "attempt_id": "01HX3...",
+      "position": 2,
+      "provider": "bedrock",
+      "model": "anthropic.claude-3-5-sonnet-20241022-v2:0",
+      "api_key": "wJalrXUtnFEMI...",
+      "api_base": null,
+      "managed": false,
+      "extra_params": {
+        "region_name": "us-east-1",
+        "aws_access_key_id": "AKIAIOSFODNN7EXAMPLE"
+      }
     }
   ]
 }
@@ -98,6 +111,21 @@ Otari iterates `attempts` in order. On a retryable failure it moves to the
 next entry; on success it stops. The `attempt_id` of the entry that ultimately
 succeeded (or the last one tried, on total failure) is what Otari echoes
 back via `X-Correlation-ID` and reports through `/gateway/usage`.
+
+`extra_params` (optional, omitted for most providers) carries provider-specific
+credential/client fields beyond `api_key`/`api_base`: for example AWS
+Bedrock's mandatory `region_name` and, for the classic IAM-key-pair shape,
+`aws_access_key_id` (the paired secret access key travels in `api_key`, since
+that's the only field the wire contract treats as sensitive). Otari merges
+these verbatim into the `acompletion()` call for that attempt, applied after
+the caller's own request fields so they can never be shadowed by a
+same-named field in the request body: the same non-overridable treatment
+`api_key` and `model` already get. This field is sourced only from the
+trusted platform peer; Otari never accepts it from the caller. AWS Bedrock's
+bearer-token ("Bedrock API key") credential shape is not representable here,
+since it requires an in-process boto3 client with a custom signing hook, so
+the platform only sends `extra_params` for the classic IAM access-key/secret-key
+shape.
 
 `request_id` groups every `attempt_id` from the same resolve call so the
 platform can attribute spend, render trace timelines, and emit fallback events.
--- a/docs/hybrid-mode-protocol.md
+++ b/docs/hybrid-mode-protocol.md
@@ -116,21 +116,38 @@ back via `X-Correlation-ID` and reports through `/gateway/usage`.
 credential/client fields beyond `api_key`/`api_base`: for example AWS
 Bedrock's mandatory `region_name` and, for the classic IAM-key-pair shape,
 `aws_access_key_id` (the paired secret access key travels in `api_key`, since
-that's the only field the wire contract treats as sensitive). Otari merges
-these verbatim into the `acompletion()` call for that attempt, applied after
-the caller's own request fields so they can never be shadowed by a
-same-named field in the request body: the same non-overridable treatment
-`api_key` and `model` already get. This field is sourced only from the
-trusted platform peer; Otari never accepts it from the caller. AWS Bedrock's
-bearer-token ("Bedrock API key") credential shape is a current gap, not a
-fundamental one: AWS's own recommended path for it is the
-`AWS_BEARER_TOKEN_BEDROCK` environment variable (or, in newer boto3 releases,
-an `aws_bearer_token` constructor kwarg), either of which is in principle
-JSON-serializable and forwardable like any other `extra_params` value. The
-reference platform (otari.ai) currently builds an in-process boto3 client
-with a custom signing hook for that shape instead, which has no
-representable form once serialized over HTTP, so it only sends
-`extra_params` for the classic IAM access-key/secret-key shape today.
+that's the only field the wire contract treats as sensitive). This field is
+sourced only from the trusted platform peer; Otari never accepts it from the
+caller, and it can never be shadowed by a same-named field in the caller's
+own request body: the same non-overridable treatment `api_key` and `model`
+already get.
+
+Otari does not merge `extra_params` into the completion call's kwargs
+directly. any-llm's `acompletion()` only forwards a separate `client_args`
+mapping to the provider's client constructor (everything else in its own
+`**kwargs` goes to the completion call instead), so `extra_params` is nested
+under `client_args` before the call. A flat merge would silently forward
+`region_name` to the completion call rather than boto3's client constructor,
+which is how an earlier version of this forwarding path still hit boto3's
+`NoRegionError` despite carrying the right value end to end.
+
+AWS Bedrock gets additional handling on top of that generic nesting, since
+any-llm's Bedrock provider never reads a plain `api_key` when building its
+boto3 client and AWS has two distinct credential shapes:
+
+- Classic IAM access-key/secret-key pair (`aws_access_key_id` present in
+  `extra_params`): the paired secret (`api_key`) is aliased into
+  `client_args["aws_secret_access_key"]`, boto3's own constructor kwarg for it.
+- Bearer token ("Bedrock API key", no `aws_access_key_id`): Otari builds a
+  boto3 client itself, with signing disabled and the `Authorization: Bearer
+  <token>` header injected via a `before-sign` event hook, and passes it as
+  `client_args["client"]` (an any-llm-sdk `BedrockProvider` constructor
+  parameter it already supports overriding its client with). The pinned
+  boto3 version has no native support for `AWS_BEARER_TOKEN_BEDROCK` or an
+  `aws_bearer_token` constructor kwarg, so this hook-based approach is the
+  only way to authenticate this shape today; it is a per-request, in-process
+  client, safe under concurrent load. See
+  `src/gateway/services/bedrock_gateway_auth.py`.
 
 `request_id` groups every `attempt_id` from the same resolve call so the
 platform can attribute spend, render trace timelines, and emit fallback events.
@@ -168,7 +185,7 @@ multi-attempt shape.
 
 | Status | Behaviour |
 |---|---|
-| `401`, `402`, `403`, `404`, `429` | Status code is forwarded to the client; `429`'s `Retry-After` header is preserved. The `detail` is the platform's JSON `detail` string when present, otherwise the fallback `"Authorization request rejected"`. |
+| `400`, `401`, `402`, `403`, `404`, `429` | Status code is forwarded to the client; `429`'s `Retry-After` header is preserved. The `detail` is the platform's JSON `detail` string when present, otherwise the fallback `"Authorization request rejected"`. |
 | `422`, `5xx`                      | Mapped to `502 Bad Gateway` with `detail = "Authorization service unavailable"`. |
 | Network/timeout                    | Mapped to `502 Bad Gateway`. |
 
@@ -216,7 +233,7 @@ configs are resolved (SSRF guard, no bearer token over cleartext `http://`).
 
 | Status | Behaviour |
 |---|---|
-| `401`, `402`, `403`, `404`, `429` | Status code is forwarded to the client; `429`'s `Retry-After` header is preserved. The `detail` is the platform's JSON `detail` string when present, otherwise the fallback `"MCP server resolution failed"`. |
+| `400`, `401`, `402`, `403`, `404`, `429` | Status code is forwarded to the client; `429`'s `Retry-After` header is preserved. The `detail` is the platform's JSON `detail` string when present, otherwise the fallback `"MCP server resolution failed"`. |
 | `422`, `5xx`                      | Mapped to `502 Bad Gateway` with `detail = "Authorization service unavailable"`. |
 | Network/timeout                    | Mapped to `502 Bad Gateway`. |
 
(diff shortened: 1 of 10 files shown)


docs/hybrid-mode-protocol.md
```diff
@@ -99,6 +112,21 @@ next entry; on success it stops. The `attempt_id` of the entry that ultimately
...
+that's the only field the wire contract treats as sensitive). Otari merges
+these verbatim into the `acompletion()` call for that attempt, applied after
+the caller's own request fields so they can never be shadowed by a
+same-named field in the request body: the same non-overridable treatment
+`api_key` and `model` already get. This field is sourced only from the
+trusted platform peer; Otari never accepts it from the caller. AWS Bedrock's
+bearer-token ("Bedrock API key") credential shape is not representable here,
+since it requires an in-process boto3 client with a custom signing hook, so
+the platform only sends `extra_params` for the classic IAM access-key/secret-key
+shape.
```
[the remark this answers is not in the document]
tbille: Fixed: rephrased to attribute the bearer-token exclusion to the reference platform's current in-process boto3 client implementation, not to bearer-token auth being inherently unrepresentable. AWS's own `AWS_BEARER_TOKEN_BEDROCK` env var and newer boto3's `aws_bearer_token` kwarg are both plain, JSON-serializable strings that could be forwarded like any other `extra_params` value; the current gap is an implementation choice on the platform side, not a wire-format limitation.

src/gateway/api/routes/_platform.py
```diff
@@ -136,6 +136,11 @@ class ResolvedAttempt(BaseModel):
     api_base: str | None = None
     api_key: str
     managed: bool
+    extra_params: dict[str, str] | None = None
+    """Provider-specific extra client kwargs beyond api_key/api_base (e.g. AWS
+    Bedrock's ``region_name``/``aws_access_key_id``). Sourced only from the
+    trusted platform peer, never from the caller's request body — see
+    ``default_attempt_kwargs``, which merges these in non-overridably."""
```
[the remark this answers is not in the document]
tbille: Fixed: replaced the em-dash separators in the newly added docstring/comments/test docstrings (_platform.py, responses.py, and the three test files) with periods/colons per the repo's style rule.

src/gateway/api/routes/_platform.py
```diff
@@ -157,14 +162,22 @@ def default_attempt_kwargs(
...
+    """
     attempt_provider = LLMProvider(attempt.provider)
     kwargs: dict[str, Any] = {"api_key": attempt.api_key}
     if attempt.api_base:
         kwargs["api_base"] = attempt.api_base
     return {
         **kwargs,
         **base_request_fields,
+        **(attempt.extra_params or {}),
         "model": f"{attempt_provider.value}:{attempt.model}",
```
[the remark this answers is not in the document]
tbille: Good catch on the merge-order pattern, though api_key/api_base specifically are already safe here: they're in `SENSITIVE_PARAM_FIELDS` (`_schema_derive.py`) and stripped from `request_fields` by `_strip_gateway_fields` (called once in `chat.py` before either the hybrid or standalone dispatch branches) before `base_request_fields` is ever built, so they can't survive into `request_fields` regardless of what the request schema allows.

Looking closer at the merge order you flagged did surface a real, adjacent gap this PR introduced though: the new `client_args` key wasn't in that deny-list, so a caller-supplied `client_args` smuggled in through a schema that allows extra fields (Responses's extra="allow") would have reached the provider call unfiltered whenever an attempt has no `extra_params` of its own (every provider except Bedrock). Fixed in dd67d53: added `client_args` to `SENSITIVE_PARAM_FIELDS`, plus a regression test.

src/gateway/api/routes/responses.py
```diff
@@ -286,6 +286,10 @@ def attempt_kwargs(
         return {
             **kwargs,
             **{k: v for k, v in request_fields.items() if k != "provider"},
+            # extra_params (e.g. Bedrock's region_name) is platform-trusted, so
+            # it's merged in after the caller's own request fields and before
+            # the forced model/provider keys, mirroring default_attempt_kwargs.
+            **(attempt.extra_params or {}),
             "model": attempt.model,
             "provider": LLMProvider(attempt.provider),
         }
```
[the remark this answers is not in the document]
tbille: Same as the sibling comment on _platform.py: request_fields (not just base_request_fields here) already has api_key/api_base stripped before this function ever sees it (_strip_gateway_fields runs once in responses.py before either dispatch branch, and both are in SENSITIVE_PARAM_FIELDS). But the merge order you're pointing at was right to be suspicious of: the new client_args key wasn't deny-listed, so a caller-supplied client_args via extra="allow" would have survived into this function's output whenever build_attempt_client_args returns None (no extra_params on the attempt). Fixed in dd67d53: client_args is now in SENSITIVE_PARAM_FIELDS, with a regression test that fails without the fix.

src/gateway/services/bedrock_gateway_auth.py
```diff
@@ -0,0 +1,107 @@
...
+
+from typing import Any
+
+import boto3
+import botocore
+from botocore.config import Config
+
+# any-llm-sdk's BedrockProvider never forwards ``api_key`` into the boto3
+# client it builds (it's only used for a non-empty presence check) — the
+# secret must instead be forwarded under boto3's real constructor kwarg.
```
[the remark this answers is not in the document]
tbille: Fixed: replaced both em dashes (bedrock_gateway_auth.py and the test docstring) with colons/periods.

src/gateway/api/routes/_platform.py
```diff
@@ -136,6 +137,11 @@ class ResolvedAttempt(BaseModel):
     api_base: str | None = None
     api_key: str
     managed: bool
+    extra_params: dict[str, str] | None = None
```
khaledosman, line 140: `dict[str, str]` is too narrow now that these values land in `boto3.client(**client_args)`. That constructor legitimately takes non-strings: `use_ssl` and `verify` are bools, `config` is a `Config` object, and any-llm passes the whole dict through to it. The classic shape spreads `extra_params` wholesale (`bedrock_gateway_auth.py:100`), so the first non-string field the platform adds fails at parse time rather than at the call.

Suggest `dict[str, Any]`, or `dict[str, str | int | float | bool]` if you want to keep nested structures out. See the coercion point on `:576`.

src/gateway/api/routes/_platform.py
```diff
@@ -523,6 +573,7 @@ def _parse_resolve_payload(payload: dict[str, Any]) -> ResolvedRoute:
                 api_base=att.get("api_base"),
                 api_key=str(att["api_key"]),
                 managed=bool(att.get("managed", False)),
+                extra_params=att.get("extra_params"),
```
khaledosman, line 580: This is the only field in `_parse_resolve_payload` that isn't defensively coerced, and it's the one whose value types this repo controls least.

Every sibling is wrapped precisely because the platform's JSON types can't be assumed: `str(att["api_key"])`, `int(att["position"])`, `bool(att.get("managed", False))`. `extra_params` goes in raw against the `dict[str, str]` field, so any non-string value raises `pydantic.ValidationError` out of the `ResolvedAttempt(...)` construction. Neither this function nor `_resolve_platform_credentials` catches it, so it surfaces as a bare 500 on every hybrid request routed to that provider, with nothing in the response to point at the cause.

Combined with widening the annotation on `:140`, coerce the keys to `str` here so an unexpected-but-well-formed payload degrades into a forwarded kwarg instead of an outage. Same on `:599` for the legacy single-attempt shape.

src/gateway/services/bedrock_gateway_auth.py
```diff
@@ -0,0 +1,107 @@
...
+    ``boto3.client()`` via any-llm's constructor-kwargs channel.
+
+    Bearer-token shape: returns ``region_name`` plus a pre-built, per-request
+    unsigned client (under the ``client`` key any-llm-sdk's ``BedrockProvider``
+    already supports overriding its client with).
+    """
+    if extra_params.get("aws_access_key_id"):
+        return {**extra_params, _SECRET_ACCESS_KEY_KWARG: api_key}
+
+    region_name = extra_params.get("region_name", "")
```
khaledosman, line 103: Defaulting a missing region to `""` re-creates a worse version of the bug this PR exists to fix.

`boto3.client("bedrock-runtime", region_name="")` does not raise `NoRegionError` (botocore only raises that for `None`), so an empty region gets through client construction and fails later as a malformed endpoint or a DNS error, with nothing naming `region_name` as the problem. The docstring says `extra_params` always carries it and the companion platform PR enforces that, which is exactly why an explicit failure here is cheap: if the invariant ever breaks, you want the error to say which field is missing.

Suggest raising a `ValueError` (or a small module-local exception) when `region_name` is absent or blank, rather than substituting `""`.

src/gateway/services/bedrock_gateway_auth.py
```diff
@@ -0,0 +1,107 @@
...
+    listing and batch ops use a second, separate ``bedrock`` control-plane
+    client any-llm-sdk builds itself, but those aren't reachable through the
+    gateway's hybrid-mode chat/messages/responses routes.
+
+    The service name is inlined as a literal at both call sites (rather than
+    a shared module constant) because ``boto3.client``'s type stubs overload
+    on a ``Literal`` service-name argument; a ``str``-typed constant, even
+    with this exact value, would fail every overload.
+    """
+    client = boto3.client(
```
khaledosman, line 69: This builds a boto3 client per request, synchronously, on the event loop.

`build_bedrock_client_args` is called from `default_attempt_kwargs` / `_ResponsesAdapter.attempt_kwargs`, which run inline in the async request handler, and `boto3.client()` is blocking work: service-model load and parse, endpoint resolution, event-system setup. Measured in this repo's venv: 119 ms for the first construction in a process, 0.6 ms warm. So the steady-state cost is small but non-zero, it is paid per *attempt* (the walker rebuilds kwargs for each candidate), and the first Bedrock bearer request in each worker stalls the loop for over a tenth of a second for every other in-flight request too.

Worth noting that any-llm already treats boto3 as blocking and offloads the actual call (`bedrock.py:154`, `loop.run_in_executor`); this construction sits outside that.

The docstring addresses thread-safety, which is the right thing to have checked, but not cost. Two cheap options: build it inside `run_in_executor` alongside the call, or cache on `(region_name, token)` with a small bounded TTL cache. If per-request construction is a deliberate trade (a token rotation takes effect immediately, no secret held in a process-wide cache), say so in the docstring so the next reader doesn't "optimise" it away.

src/gateway/api/routes/_platform.py
```diff
@@ -466,7 +516,7 @@ async def _post_resolve(
     if response.status_code == 200:
         return response.json()
 
-    if response.status_code in {401, 402, 403, 404, 429}:
+    if response.status_code in {400, 401, 402, 403, 404, 429}:
```
khaledosman, line 523: This widening reaches further than the docstring's justification, and it is unrelated to the PR's stated scope.

`_post_resolve` is shared by all three resolve helpers, so this also forwards the platform's 400 detail on MCP-server and web-search resolution. The docs tables were updated for all three (`docs/hybrid-mode-protocol.md:188,236,287`), so the reach is clearly known, but the code rationale ("this backend's own 400s are deliberately hand-written, caller-safe rejections, e.g. a Bedrock BYO key using an auth shape that cannot be forwarded") only covers the provider-keys path. The MCP and web-search endpoints' 400s are not established as hand-written, and the invariant that keeps this safe ("the platform never returns a raw framework validation error as a 400") lives in another repo and cannot be enforced here.

Either narrow the widening to the provider-keys call (pass the accepted status set in per caller), or state the cross-repo invariant explicitly in the docstring so a platform-side change that breaks it is recognisable as breaking this. Splitting it into its own PR would also make it reviewable on its own terms; as it stands, a reviewer looking at a Bedrock region fix has no reason to look at the error-forwarding ladder.

tests/integration/test_hybrid_mode_chat.py
```diff
@@ -153,6 +153,147 @@ async def fake_acompletion(**kwargs: Any) -> ChatCompletion:
...
+    response = platform_client.post(
+        "/v1/chat/completions",
+        json=_bedrock_chat_request(),
+        headers={"Authorization": "Bearer user_test_token"},
+    )
+
+    assert response.status_code == 200
+
+
+def test_hybrid_mode_forwards_bedrock_bearer_token_via_client_args(
```
khaledosman, line 252: Minor, and much less pressing than in the earlier revision now that `build_attempt_client_args` is shared: all four new integration tests are non-streaming, but the bug report quoted in the PR description is "All streaming attempts failed ... You must specify a region."

The streaming path does go through `adapter.attempt_kwargs` (`_pipeline.py:2037`), so it is genuinely covered by construction. It just has no assertion, which leaves the one path the reporter actually hit as the one path a future refactor could regress silently. A streaming variant of the bearer test asserting `client_args` reaches `acompletion` would close it cheaply.

khaledosman (commented):
Reviewed at `d32ea8e` (the reworked `client_args` version, not the earlier flat merge). The core diagnosis is right and I verified it against the installed any-llm source: `BedrockProvider.__init__` does `kwargs.pop("client", None)` and `_init_client` calls `boto3.client("bedrock-runtime", endpoint_url=api_base, **kwargs)` with no reference to `api_key`, so both the `client_args` nesting and the `aws_secret_access_key` aliasing are necessary rather than defensive. `_verify_and_set_api_key` also short-circuits on a custom client, so the bearer path won't trip credential verification. `tests/unit/test_hybrid_client_args_routing.py` constructing a real boto3 client is the right shape of test for a bug whose whole nature was "the value was carried correctly and still went to the wrong place".

Four things inline: a latent 500 on a non-string `extra_params` value, an empty-string region fallback, per-request boto3 client construction on the event loop, and an unrelated behaviour change (the 400 widening) that reaches further than its stated justification.

PR #498, opened 2026-08-05
fix(hybrid): prevent pending cost calculations when fallback attempts stop

HareeshBahuleyan:
## Description

In hybrid mode, the platform evaluates the routing policy and returns an ordered list of provider attempts that the gateway can try in sequence. The gateway sends a usage report after each attempt it actually makes.

Previously, if a request succeeded or stopped before reaching the end of that list, the platform could not tell that the remaining attempts would not run. It could continue waiting for their usage reports, leaving the request’s cost calculation pending indefinitely.

This PR adds `is_final_attempt` to usage reports. It is `false` when the gateway will try another fallback after a retryable failure. It is `true` when the request has succeeded or no further fallback attempts will be made.

The marker is included in both background and inline error reporting. Tests cover streaming and non-streaming requests across success, fallback, and failure paths.

The hybrid-mode protocol documentation has also been updated. This is the gateway-side companion to the platform changes in mozilla-ai/otari-ai#1543.

Verification: `make lint`, `make typecheck`, `make test` (2099 passed, 9 skipped), `make openapi-check`, and `make postman-check`.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [x] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Related to mozilla-ai/otari-ai#1508 and mozilla-ai/otari-ai#1543.

## Checklist

- [x] If the API contract changed, I regenerated the OpenAPI spec (`uv run python scripts/generate_openapi.py`). No inbound OpenAPI schema changed; `make openapi-check` and `make postman-check` pass.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

Pi coding agent (GPT-5.6)

Implemented with test-driven development from the platform review finding. Tests were written and observed failing before each production path was updated.

## Summary

- Added an `is_final_attempt` boolean to hybrid-mode usage reports.
- Set `is_final_attempt` to `false` only for retryable failures that still trigger another fallback attempt.
- Set `is_final_attempt` to `true` for successful requests and for all terminal outcomes where no further fallback will run.
- Applied the marker to streaming and non-streaming reporting, including background and inline error reporting.
- Updated the hybrid-mode protocol documentation and clarified how usage reports behave when token usage data is unavailable.
- Added and expanded unit and integration tests for success, fallback, retryable failure, and non-retryable/terminal failure paths.

## Technical notes

- Updated reporting so successful attempts can still send a final usage report with `usage=None` when token usage data is missing.

### DIFF
--- a/src/gateway/api/routes/_pipeline.py
+++ b/src/gateway/api/routes/_pipeline.py
@@ -2038,6 +2038,7 @@ def build_streaming_response(
                     outcome="success",
                     usage=usage_data,
                     session_label=session_label,
+                    is_final_attempt=True,
                 ),
                 platform_correlation_id,
             )
@@ -2111,6 +2112,7 @@ def build_streaming_response(
                     outcome="error",
                     usage=None,
                     session_label=session_label,
+                    is_final_attempt=True,
                 ),
                 platform_correlation_id,
             )
@@ -2339,7 +2341,7 @@ async def run_single_attempt_stream(
 
 async def _flush_pending_usage_reports(
     config: GatewayConfig,
-    pending_error_reports: list[tuple[str, str, Any, str | None]],
+    pending_error_reports: list[tuple[str, str, Any, str | None, bool]],
     request_id: str,
     session_label: str | None = None,
 ) -> None:
@@ -2364,8 +2366,16 @@ async def _flush_pending_usage_reports(
         results = await asyncio.wait_for(
             asyncio.gather(
                 *(
-                    _report_platform_usage(config, attempt_id, outcome, usage, error_class, session_label)
-                    for attempt_id, outcome, usage, error_class in pending_error_reports
+                    _report_platform_usage(
+                        config,
+                        attempt_id,
+                        outcome,
+                        usage,
+                        error_class,
+                        session_label,
+                        is_final_attempt,
+                    )
+                    for attempt_id, outcome, usage, error_class, is_final_attempt in pending_error_reports
                 ),
                 return_exceptions=True,
             ),
@@ -2468,7 +2478,7 @@ async def run_streaming_with_fallback(
     # for the success path (it flushes once the SSE response completes), but
     # also stash the error reports so they can be flushed inline on the
     # all-failed path below.
-    pending_error_reports: list[tuple[str, str, Any, str | None]] = []
+    pending_error_reports: list[tuple[str, str, Any, str | None, bool]] = []
 
     async def _on_attempt_failed(attempt: ResolvedAttempt, failure: StreamingAttemptFailure) -> None:
         background_tasks.add_task(
@@ -2479,8 +2489,11 @@ async def run_streaming_with_fallback(
             None,
             failure.error_class,
             session_label,
+            failure.is_final_attempt,
         )
-        pending_error_reports.append((attempt.attempt_id, "error", None, failure.error_class))
+        pending_error_reports.append(
+            (attempt.attempt_id, "error", None, failure.error_class, failure.is_final_attempt)
+        )
         record_abandoned_attempt(attempt.provider, attempt.model, failure.reason, attempt.position)
         logger.warning(
             "Streaming attempt failed request_id=%s position=%d provider=%s model=%s error=%s",
@@ -2630,13 +2643,14 @@ async def run_platform_non_stream(
     # can't fire its fallback-exhausted accounting. Keep the background task for
     # the success-response path (non-blocking), but also stash the error reports
     # so they can be flushed inline if the request ends in an exception.
-    pending_error_reports: list[tuple[str, str, Any, str | None]] = []
+    pending_error_reports: list[tuple[str, str, Any, str | None, bool]] = []
 
     def _report_attempt_outcome(
         attempt: ResolvedAttempt,
         outcome: str,
         usage: Any,
         error_class: str | None,
+        is_final_attempt: bool,
     ) -> None:
         background_tasks.add_task(
             _report_platform_usage,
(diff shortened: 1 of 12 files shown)


src/gateway/api/routes/_platform.py
```diff
@@ -792,11 +794,13 @@ async def _report_platform_usage(
     usage: CompletionUsage | None,
     error_class: str | None = None,
     session_label: str | None = None,
+    is_final_attempt: bool = False,
```
khaledosman: The default fails in the unsafe direction: any call site that omits the argument reports "another attempt is coming", which is exactly the indefinite-pending bug this PR fixes. All five current call sites pass it, so this is about the next one. Suggest making it keyword-only and required (`*, is_final_attempt: bool`); `tests/unit/test_run_platform_attempts.py:219` and `:257` rely on the default and would need the argument added.

src/gateway/api/routes/_platform.py
```diff
@@ -294,7 +295,8 @@ def _mark_locked_in(_pos: int = attempt.position) -> None:
             raise
         except BaseException as exc:
             retryable, error_class = classify_error(exc)
-            report_attempt_outcome(attempt, "error", None, error_class)
+            is_final_attempt = locked_in or not retryable or is_last_planned_attempt
```
khaledosman, line 300: Terminal paths just above bypass this entirely: `MaxToolIterationsExceeded` (:278), `SandboxNotReachableError` / `WebSearchNotReachableError` (:289), and the up-front Responses guard (`responses.py:376`) all raise without calling `report_attempt_outcome`. Those requests end with zero reports carrying `is_final_attempt=true`, so the platform is still left polling. Either emit a final error report on those branches, or state in `docs/hybrid-mode-protocol.md` that a resolved request can legitimately produce no report at all, so the platform keeps a timeout as backstop.

src/gateway/api/routes/_platform.py
```diff
@@ -809,7 +813,11 @@ async def _report_platform_usage(
     usage_url = _platform_url(platform_base_url, "/gateway/usage")
     headers = {"X-Gateway-Token": config.platform_token or ""}
 
-    payload: dict[str, Any] = {"correlation_id": correlation_id, "status": outcome}
+    payload: dict[str, Any] = {
+        "correlation_id": correlation_id,
+        "status": outcome,
+        "is_final_attempt": is_final_attempt,
```
khaledosman, line 822: Rollout ordering: a 422 from `/gateway/usage` is in the non-retryable short-circuit set below, so if this ships before mozilla-ai/otari-ai#1543 and the platform handler rejects unknown fields, every usage report is dropped silently. That is lost spend, not just a lost marker. Worth confirming the platform side is deployed (or parses leniently) first.

src/gateway/api/routes/_pipeline.py
```diff
@@ -1751,6 +1752,7 @@ async def _on_error(exc: BaseException) -> None:
                     outcome="error",
                     usage=None,
                     session_label=session_label,
+                    is_final_attempt=True,
```
khaledosman, line 2138: `_on_complete` and `_on_error` both mark final now, but the third terminal outcome does not: `_on_no_usage` (:1706) returns early on `db is None`, which is always true in hybrid mode. A stream that completes without a usage chunk (or the zero-chunk `StopAsyncIteration` commit in `streaming.py:264`) therefore sends no report at all and the chosen attempt stays pending forever. Pre-existing, but it is the same failure mode this PR targets and cheap to close with a `platform_active` branch reporting `is_final_attempt=True`.

tests/integration/test_hybrid_mode_messages.py
```diff
@@ -714,6 +717,8 @@ async def fake_loop_amessages(**kwargs: Any) -> MessageResponse:
     # Both calls were to attempt 1 (rounds 1 and 2 of the tool loop). The
     # second attempt is never tried because lock-in fired on round 1.
     assert calls == ["sk-ant-broken", "sk-ant-broken"]
+    assert len(usage_reports) == 1
+    assert usage_reports[0]["is_final_attempt"] is True
```
khaledosman, line 721: This passes for the wrong reason, so the `locked_in` term at `_platform.py:298` is not actually covered. The fake raises a bare `RuntimeError`, which `_classify_upstream_error` maps to `(False, "unknown")` (`upstream_exception_shape` finds neither a kind nor a status code), so `not retryable` already forces `True`; deleting `locked_in` from line 298 leaves this test green. Worth a unit test next to `test_locked_in_failure_not_counted_as_abandoned`: two attempts, `classify_error=lambda _e: (True, "http_500")`, `on_first_response()` called before raising, asserting a single report of `(attempts[0], "error", None, "http_500", True)`.

src/gateway/api/routes/_pipeline.py
```diff
@@ -2051,7 +2061,7 @@ async def _build_for_attempt(attempt: ResolvedAttempt) -> AsyncIterator[ChunkT]:
     # for the success path (it flushes once the SSE response completes), but
     # also stash the error reports so they can be flushed inline on the
     # all-failed path below.
-    pending_error_reports: list[tuple[str, str, Any, str | None]] = []
+    pending_error_reports: list[tuple[str, str, Any, str | None, bool]] = []
```
khaledosman: Nit: this 5-tuple now appears in three places plus a matching 5-argument `Callable`, all positional with a trailing bool. A `NamedTuple` would make the unpack in `_flush_pending_usage_reports` and the call sites order-safe for the next field added.

khaledosman (approved):
The marker itself is computed correctly on both paths, and the streaming asymmetry (the timeout branch omits `not retryable` because a timeout always continues) is right. Findings inline are non-blocking: an unsafe default direction, two terminal outcomes that still send no report at all, one test that passes for the wrong reason, and a deploy-ordering risk against the platform companion PR.

PR #508, opened 2026-08-05
fix(dashboard): open the onboarding quickstart link in a new tab

### LINKED ISSUE #500: Providers page tutorial link escapes the HashRouter

Found by CodeRabbit while reviewing #492, on code that PR does not touch.

`web/src/pages/ProvidersPage.tsx:750` links to the get-started tutorial with a raw anchor:

```tsx
<a href="/welcome" className="font-medium text-[var(--otari-brand-dark)]">
```

The dashboard routes through react-router's `HashRouter`, so every route lives under a hash path (`/#/providers`, `/#/models`). A bare `href="/welcome"` is a server-path navigation: it leaves the router hash empty, so the tutorial route is never selected. Standalone mode does serve `/welcome`, which is why this is a wrong-looking navigation rather than a hard 404, but it drops out of the SPA and back to a full page load.

Fix is the router's `Link` with `to="/welcome"`, or emitting `/#/welcome` if it needs to stay an anchor. Rebuild the committed bundle after (`npm --prefix web run build`).

Worth a grep for other raw anchors to internal routes while in there.

njbrake:
## Description

The issue's premise is wrong, so this does not implement its suggested fix. `/welcome` is not a HashRouter route: it is served by the gateway itself (`src/gateway/main.py:274`, `ROOT_TUTORIAL_HTML`). The router has no `welcome` route and its `*` route redirects to `/`, so `<Link to="/welcome">` would resolve to `/#/welcome` and drop the operator on the overview page. The existing `href="/welcome"` navigates correctly and matches `Login.tsx`.

What is actually wrong: following the link replaces the signed-in SPA with a full page load, mid onboarding, with only the back button to get back. `DocsPage.tsx` and `AppShell.tsx` already handle links that leave the app with `target="_blank" rel="noreferrer"` for that reason, so this applies the same treatment and records in a comment why the anchor must stay a path anchor rather than a router `Link`.

Also grepped for other raw anchors to internal routes. The only other one is `Login.tsx:92`, which renders outside the router (`App.tsx:58`), so a router `Link` is impossible there and the path anchor is correct. Left alone.

Committed dashboard bundle rebuilt. All app chunk hashes rolled because the entry chunk changed; vendor chunks are unchanged.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes #500

## Checklist

- [x] I have added or updated tests that cover my change (`web/src/pages/ProvidersPage.test.tsx`; verified it fails against baseline HEAD and passes with the fix).
- [x] I ran the Definition of Done checks locally: `npm --prefix web run typecheck`, `npm --prefix web test` (446 passing), `npm --prefix web run build`. No Python changed, so the backend checks and the OpenAPI artifacts are unaffected.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

Claude Code (Opus 5)

Implemented by Claude via back-and-forth with @njbrake. The decision to reject the issue's suggested fix and treat this as a new-tab problem instead is his; the code and prose are Claude's.


web/src/pages/ProvidersPage.tsx
```diff
@@ -747,7 +747,16 @@ function OnboardingPanel({
           <Step n={3} title="Send your first request">
             Point your app at <code>/v1</code> on this gateway with the API key printed in the server logs
             (<code>gw-…</code>). See the{" "}
-            <a href="/welcome" className="font-medium text-[var(--otari-brand-dark)]">
+            {/* /welcome is served by the gateway itself, not by a HashRouter route, so this
+                stays a plain path anchor: a router Link would resolve to /#/welcome, which
+                the catch-all route sends back to the overview. It leaves the SPA, so open a
+                new tab and the operator keeps the dashboard (as the guide's links do). */}
+            <a
+              href="/welcome"
```
khaledosman, line 755: The link now leaves the SPA with no affordance: nothing in the text or markup signals a new tab, so screen-reader users get no warning and sighted operators lose the back button they'd expect from step 3 of onboarding. `AppShell.tsx:567` pairs its `target="_blank"` with a `title` plus an icon. Add at least `title="Opens in a new tab"` here (an `sr-only` suffix or `aria-label` would change the accessible name and break the new test's `getByRole("link", { name: "quickstart" })`).

khaledosman (approved):
Automated review by Claude Code (Opus 5).

PR #510, opened 2026-08-05
feat(keys): add per-key ignore_user_mismatch opt-out

### LINKED ISSUE #493: [FEATURE] Scope reject_user_mismatch per key, or treat telemetry-shaped metadata.user_id as opaque on /v1/messages
### Problem or use case

`docs/use-with-claude-code.md` tells standalone users to set `reject_user_mismatch: false` so Claude Code can route through the gateway. That instruction is correct, but the blast radius is larger than the page implies: the field is gateway-wide, so accommodating one client turns the check off for **every** non-master key on the deployment.

The reason it has to be turned off is that Claude Code's `metadata.user_id` is not an identity assertion at all. It is a JSON blob of client telemetry. Captured by pointing `ANTHROPIC_BASE_URL` at a local listener (Claude Code v2.1.220):

```
POST /v1/messages?beta=true
metadata.user_id = "{\"device_id\":\"<sha256>\",\"account_uuid\":\"\",\"session_id\":\"<uuid4>\"}"
```

Two consequences:

1. The value can never equal an Otari user id, so a non-master key can never satisfy the check — this is unconditional, not intermittent.
2. It embeds a **fresh `session_id` per session**, so provisioning an Otari user whose id matches is not a workaround either. There is no value an operator can configure that would make it pass.

The check also runs ahead of the per-key model allowlist, so it is the first thing every Claude Code request hits.

Net effect: the only supported way to use a documented, first-class client is a gateway-wide relaxation of an access check. Spend attribution is unaffected either way (with the flag off, spend is still bound to the key's own user), so what is being given up is the guarantee that a non-master key cannot name an arbitrary `user` in the payload forwarded upstream — for all keys, not just the one that needs it.

### Proposed solution

Either of these would let the strict default stay on gateway-wide and scope the exception to the key that needs it:

1. **Per-key opt-out** — an `ignore_user_mismatch` flag on the API key, alongside `exclude_from_budget`. The Claude Code key gets it; every other key keeps the strict check.
2. **Shape-aware handling on `/v1/messages`** — when `metadata.user_id` is a JSON object of client telemetry (`device_id` / `session_id` / `account_uuid`) rather than a bare user id, treat it as opaque telemetry and skip the comparison. Spend binding does not change; it is already bound to the key's own user whenever the comparison is skipped today.

Option 1 is the more conservative of the two, since it keeps the decision with the operator rather than inferring intent from payload shape.

If neither is wanted, then at minimum `docs/use-with-claude-code.md` should say that the setting is gateway-wide and cannot be scoped to a single key or client, so operators can weigh it. Today it reads as a routine one-line config change.

### Alternatives considered

- **Provision an Otari user whose id matches the sent value** — impossible; the value contains a per-session UUID that changes every launch.
- **Use the master key for Claude Code** — this does satisfy the check, since the master key may bill an arbitrary user, but it hands gateway management (keys, users, budgets) to a coding agent. Non-starter.
- **Strip `metadata` in a proxy in front of Otari** — works, but adds a hop and undercuts the gateway being the single point of control.
- **Leave `reject_user_mismatch: true`** — Claude Code cannot use the gateway at all.

njbrake:
## Description

`reject_user_mismatch` is gateway-wide, so the documented way to run Claude Code turned the user-mismatch check off for every non-master key on the deployment. It has to be turned off because Claude Code's `metadata.user_id` is client telemetry (a JSON blob with a fresh `session_id` per launch), not an identity: it can never equal an Otari user id, and no user an operator provisions makes the check pass.

This adds a nullable `reject_user_mismatch` on the API key: `null` (the default, and where every existing key lands) inherits the deployment setting, `false` always accepts a mismatched `user`, `true` always rejects one. It reuses the NULL-means-inherit precedent from `allowed_models`, so one name covers both directions and a deployment that already relaxed the check globally can still pin an individual key strict.

Settable at create time and via `PATCH /v1/keys/{id}` (tri-state: an omitted field leaves the override alone, an explicit `null` clears it), honored inside `resolve_user_id` so every route that resolves a billed user picks it up, and threaded through the usage-import binding helper as well (`POST /v1/usage/external-events` plus the OTLP receiver), which has its own copy of the rule because a service cannot import the API layer. The dashboard gets a three-way picker in the key create and edit forms, and the keys table chips either pinned state.

Spend behavior is unchanged in every case: a lenient key still binds spend to its own user, and the client value is forwarded to the provider as an end-user tag only. The override only decides the 403. Docs, the settings-field description the dashboard renders, the search route docstrings that ship into the spec, and the security-review instructions all now describe both levels.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [x] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes #493

## Checklist

Tests: real Claude Code telemetry `user_id` is rejected (403) on a normal key and completes on a flagged key while the gateway-wide setting stays strict, with a second unflagged key still rejected; the flag defaults off and round-trips through create/PATCH/GET; a unit case asserts a flagged key still binds to its own user; dashboard tests assert the POST and PATCH bodies carry the flag. Two integration tests (`test_error_detail_leakage`, `test_streaming_error_event`) fail in my sandbox on clean `main` too, an asyncpg teardown artifact of running against a local Postgres instead of Testcontainers.

## AI Usage

- [ ] No AI was used.

Implemented via the fix-github-issue workflow: issue and referenced files read first, then the per-key flag, tests, docs, and regenerated artifacts.


src/gateway/models/entities.py
```diff
@@ -46,6 +46,13 @@ class APIKey(Base):
     # User.spend and never gates enforcement. Default false keeps every existing key
     # (and all keys minted before this column) on the normal enforced path.
     exclude_from_budget: Mapped[bool] = mapped_column(default=False)
+    # When true, a request on this key that names a different ``user`` is accepted
+    # instead of rejected, even while the gateway-wide ``reject_user_mismatch`` is
+    # on. Spend still binds to this key's own user, so the value is only a
+    # provider-side tag. Exists for clients whose ``user`` field is telemetry
+    # rather than an identity (Claude Code sends a per-session JSON blob), so the
+    # exception is scoped to that key instead of the whole deployment.
+    ignore_user_mismatch: Mapped[bool] = mapped_column(default=False)
```
khaledosman: The flag only loosens. A deployment already running `reject_user_mismatch: false` (exactly the setup these docs now steer people off) has no way to re-tighten individual keys, and the effective rule becomes `reject_user_mismatch and not key.ignore_user_mismatch`: two names with inverted polarity.

A tri-state `Mapped[bool | None]` named `reject_user_mismatch` (NULL = inherit the gateway default) covers both directions in the same column, reuses the NULL-means-inherit precedent that `allowed_models` uses three lines below, and keeps one name. Worth settling before it lands in `openapi.json` and the generated Go/Python/Rust/TS SDKs.

docs/configuration.md
```diff
@@ -76,7 +76,7 @@ pricing:
...
 | `default_pricing` | bool | `false` | When a model has no pricing in the database, fall back to community-maintained defaults from the bundled genai-prices dataset. Off by default (opt-in). Database pricing always wins. See [Default pricing](#default-pricing). |
-| `reject_user_mismatch` | bool | `true` | When `true`, a non-master key whose request names a `user` other than its own is rejected (HTTP 403). When `false`, the client `user` is still forwarded to the provider but spend is always bound to the key's own user. The master key may always bill an arbitrary user. |
+| `reject_user_mismatch` | bool | `true` | When `true`, a non-master key whose request names a `user` other than its own is rejected (HTTP 403). When `false`, the client `user` is still forwarded to the provider but spend is always bound to the key's own user. This applies deployment-wide; to relax the check for one client only, set `ignore_user_mismatch` on its key instead. The master key may always bill an arbitrary user. |
```
khaledosman: `core/config.py:355`'s `Field(description=...)` for `reject_user_mismatch` still presents the deployment-wide switch as the only option. That string is what the dashboard renders under Settings > Metering & budgets (`api/routes/settings.py:80`), so the one place an operator actually flips this never learns the per-key flag exists. Mirror this row's wording there, then regenerate the spec and collection.

SECURITY.md
```diff
@@ -46,7 +46,8 @@ otari is a self-hosted LLM gateway; several controls are configuration-dependent
 - **The client `user` field is not trusted for non-master keys.** Spend is bound
   to the API key's own user; only the master key may bill on behalf of an
   arbitrary user. By default a non-master key naming a different user is rejected
-  (403); set `reject_user_mismatch=false` to instead bind to the key's own user
-  while still forwarding `user` to the provider as an end-user tag.
+  (403); set `ignore_user_mismatch` on a single key, or
```
khaledosman: `.github/instructions/security-review.instructions.md:41` still lists `reject_user_mismatch=false` as the only sanctioned lenient path under the IDOR check. It auto-applies to every `src/gateway/` review, so leaving it stale means future reviews (Copilot's included) read the new `api_key.ignore_user_mismatch` branch in `resolve_user_id` as a regression. Add the per-key flag there.

src/gateway/api/routes/keys.py
```diff
@@ -41,6 +41,12 @@ class CreateKeyRequest(BaseModel):
         description="When true, requests on this key are logged with cost but never reserved, "
         "reconciled into the user's spend, or gated by budget.",
     )
+    ignore_user_mismatch: bool = Field(
```
khaledosman: `api/routes/search.py:164` and `:199` still document `reject_user_mismatch` as the only way a mismatched `user` gets ignored rather than rejected. Those docstrings ship into `docs/public/openapi.json` and the Postman collection, so they are published API docs now describing an incomplete rule.

khaledosman (approved):
Automated review by Claude Code (Opus 5).

PR #513, opened 2026-08-06
feat(responses): support compaction in managed tool loops

### LINKED ISSUE #495: [FEATURE] Support OpenAI Responses context compaction on /v1/responses
OpenAI Responses server-side context compaction is being added to any-llm in mozilla-ai/any-llm#1227. Otari needs a small follow-up before `/v1/responses` can claim end-to-end support, especially when gateway-managed tools cause an internal Responses iteration.

This is separate from #486, which covers Anthropic context management on `/v1/messages`: the provider, endpoint, request shape, upstream dependency, and tool-loop replay semantics differ.

## Most direct passthrough plumbing already works

Verified against `main` at `4795754813eaa86140a4d89187e40a6cfb0fe3f0`:

- `ResponsesRequest` derives from any-llm's `ResponsesParams` (`src/gateway/api/routes/responses.py:61`), so `context_management` becomes a typed wire field once Otari upgrades to the any-llm release containing #1227. The schema also currently uses `extra="allow"`.
- `call_provider` and `open_provider_stream` forward request kwargs wholesale through `aresponses` (`responses.py:225-228`).
- Non-streaming responses use `model_dump(exclude_none=True)`, and streaming events use `model_dump_json(exclude_none=True)`, so provider-returned compaction items/events should pass through on the direct path.

## Managed tool-loop gap

OpenAI's stateless compaction flow requires the returned `compaction` output item to be replayed in subsequent input. any-llm#1227 preserves the typed item for that purpose.

Otari's Responses tool loop currently retains only gateway-owned function calls when it starts another provider iteration:

- Non-streaming `advance_transcript` appends `_items_to_dicts(owned)` plus the corresponding function-call outputs (`src/gateway/services/mcp_loop_responses.py:243-252`). Non-function output items, including `compaction`, are omitted.
- Streaming state tracks function calls only, and `advance_stream_transcript` reconstructs only those calls plus their outputs (`mcp_loop_responses.py:343-374`).

If compaction and a gateway-owned function call occur in the same provider response, the gateway's hidden follow-up iteration can therefore omit the compaction item. The final client response may also lack the compacted replay state because the intermediate response was consumed inside the gateway.

## Remaining work

1. Bump any-llm to the first release containing mozilla-ai/any-llm#1227.
2. Add direct-path contract coverage proving `context_management` reaches `aresponses` and compaction output survives streaming and non-streaming serialization.
3. Preserve compaction items across gateway-managed Responses tool-loop iterations, on both streaming and non-streaming paths.
4. Ensure a compaction item produced during a hidden tool-loop iteration remains available to the caller so its next stateless request can replay the compacted context.

## Acceptance criteria

- Requests without `context_management` remain unchanged.
- `context_management` reaches the OpenAI Responses provider unchanged.
- Direct streaming and non-streaming calls preserve provider-returned compaction output.
- When a response contains both a compaction item and gateway-owned function calls, the next internal provider request includes the compaction item in the correct output-item order before the matching function-call output.
- After such an internal iteration, the downstream client receives sufficient compacted replay state for its next stateless request.
- Both streaming and non-streaming managed-tool paths have regression tests.

Blocked on mozilla-ai/any-llm#1227 landing and being released.

HareeshBahuleyan:
## Description

Current `main` supports context compaction for direct Responses calls through any-llm 1.24.0. Gateway-managed tool loops still omitted intermediate compaction items when running a tool, so `/v1/responses` did not yet have end-to-end compaction support.

Replay compaction items in provider output order before function-call outputs, and include compaction state from internal iterations in the final streaming and non-streaming response. Adds direct passthrough coverage and managed-tool regression tests.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes #495

## Checklist

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:** OpenAI GPT-5.6 Sol via Pi

**Any additional AI details you'd like to share:** AI inspected the related Otari and any-llm changes, implemented the tool-loop replay behavior, and added contract and regression coverage. `make lint`, `make typecheck`, 81 targeted tests, and generated OpenAPI and Postman checks passed. The full `make test` suite was not run.

**NOTE:**

## Summary

- Added OpenAI Responses compaction support for direct requests and gateway-managed tool loops.
- Preserved compaction output in streaming and non-streaming responses.
- Replayed compaction items before matching function-call outputs during tool-loop continuations.
- Exposed compaction state from internal iterations for stateless follow-up requests.
- Added integration and unit tests for request forwarding, event streaming, and tool-loop behavior.
- Updated behavior for any-llm 1.24.0.

## Benefits

Clients now receive complete compaction state across direct requests and managed tool loops without changing requests that omit `context_management`.

### DIFF
--- a/src/gateway/services/mcp_loop_responses.py
+++ b/src/gateway/services/mcp_loop_responses.py
@@ -214,6 +214,25 @@ def _web_search_items_for(owned: list[Any]) -> list[ResponseFunctionWebSearch]:
     return items
 
 
+def _compaction_items(output: list[Any]) -> list[Any]:
+    """Return provider compaction items in output order."""
+    return [item for item in output if getattr(item, "type", None) == "compaction"]
+
+
+def _replay_items(output: list[Any], owned: list[Any]) -> list[Any]:
+    """Keep compaction and owned function calls in provider output order."""
+    owned_call_ids = {getattr(item, "call_id", None) for item in owned}
+    return [
+        item
+        for item in output
+        if getattr(item, "type", None) == "compaction"
+        or (
+            getattr(item, "type", None) == "function_call"
+            and getattr(item, "call_id", None) in owned_call_ids
+        )
+    ]
+
+
 async def _execute_stream_owned(state: "_ResponsesStreamState", pool: ToolBackend) -> list[dict[str, Any]]:
     """Run the stream's gateway-owned function calls, returning their output items.
 
@@ -264,12 +283,28 @@ def _without_output_items(event: Any, call_ids: set[str]) -> Any:
         return event
 
 
+def _prepend_output_items(event: Any, items: list[Any]) -> Any:
+    """Prepend hidden-iteration output items to a terminal response."""
+    if not items:
+        return event
+    response_obj = getattr(event, "response", None)
+    if response_obj is None:
+        return event
+    try:
+        output = list(items) + list(getattr(response_obj, "output", None) or [])
+        return event.model_copy(update={"response": response_obj.model_copy(update={"output": output})})
+    except (AttributeError, TypeError):
+        logger.warning("Could not add hidden Responses output items to response.completed")
+        return event
+
+
 class _ResponsesStreamState:
     """Per-iteration bookkeeping for the Responses streaming loop."""
 
     def __init__(self) -> None:
         # output_index -> {"call_id", "name", "arguments"}
         self.function_calls: dict[int, dict[str, Any]] = {}
+        self.compaction_items: dict[int, Any] = {}
         self.deferred_completed: ResponseStreamEvent | None = None
         self.owned_specs: list[dict[str, Any]] = []
         # Output items the gateway runs itself. Their events are swallowed: the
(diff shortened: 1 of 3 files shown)


src/gateway/services/mcp_loop_responses.py
```diff
@@ -366,12 +403,15 @@ async def advance_transcript(
...
         acc: dict[str, Any] | None = None,
     ) -> None:
-        # All-owned: continue. Append the assistant's function_call items AND
-        # the matching function_call_output items so the next call's input has
-        # the full transcript.
-        transcript.extend(_items_to_dicts(owned))
+        # All-owned: continue. Replay compaction items and the assistant's
+        # function calls in their original output order, then append matching
+        # function_call_output items. Stateless compaction requires the opaque
+        # compaction item on every continuation.
```
khaledosman, line 413: `advance_transcript` appends the provider's `compaction` item to the existing transcript instead of letting it replace the history it summarizes, so each continuation carries the full pre-compaction items plus one more opaque blob. With `context_management: [{"type": "compaction", ...}]` and an owned tool, iteration 2 sees a larger input than iteration 1, re-crosses the threshold, emits another compaction item, and that one is appended too: input tokens grow monotonically, the caller is billed for the uncompacted history every round, and a longer loop ends in an upstream context-length 400, the exact outcome compaction was requested to prevent. Same shape at lines 411 and 589.

This is the one finding the verifier left as PLAUSIBLE, since it depends on whether the provider intends the compaction item to substitute for the items it covers. If it does, truncating the transcript to the compaction boundary before appending is the fix, and a test pinning transcript length across two compacting iterations would document the contract either way.

khaledosman (approved):
Approving. One inline finding on the transcript handling, flagged PLAUSIBLE rather than confirmed because it turns on what the provider expects the compaction item to stand in for. Worth a look before merge.

---

src/gateway/services/mcp_loop_responses.py
HareeshBahuleyan: Thanks for flagging this. OpenAI’s [compaction guide](https://developers.openai.com/api/docs/guides/compaction) says stateless chaining should append output items as usual, while dropping items before the latest compaction is an optional latency optimization. The current loop follows that documented pattern, so I’d leave this unchanged and consider pruning as a follow-up optimization.

PR #534, opened 2026-08-07
fix(messages): classify bridged context_management/betas as a 400, not a 500

### LINKED ISSUE #530: [BUG] context_management/betas against a bridged (non-native-Anthropic) provider surfaces as a misleading 500, not a 400
### Problem or use case

A `/v1/messages` request that sets `context_management` or `betas` against a model whose provider is **not** native Anthropic Messages API (i.e. any bridged/translated provider — self-hosted OSS models, OpenAI-compatible backends, anything reached via any-llm's generic `_amessages`) fails with a generic 500 that tells the caller to retry. It never will succeed, no matter how many times it's retried.

Root cause is in any-llm's own default `_amessages`, which is correct to reject the request — a bridged backend genuinely cannot honor Anthropic's server-side compaction:

```python
# any_llm/any_llm.py
if params.context_management is not None or params.betas:
    msg = "context_management and betas require a provider with a native Anthropic Messages API"
    raise NotImplementedError(msg)
```

The gap is on Otari's side: `classify_provider_error` (`src/gateway/api/routes/_pipeline.py`) only recognizes exceptions shaped like an HTTP error (something `upstream_exception_shape` can pull a status code from). A bare `NotImplementedError` has none, so classification falls through (`status_code is None: return None`) and the caller gets the generic, unclassified path — surfaced to a client like Claude Code as:

```
API Error: 500 {"detail":{"type":"error","error":{"type":"api_error",
  "message":"The request could not be completed by the provider"}}}
This is a server-side issue, usually temporary — try again in a moment.
```

That message is actively wrong for this case. It is not server-side, not temporary, and retrying is guaranteed to fail identically every time.

### How this gets triggered without the caller asking for it directly

Neither `context_management` nor `betas` needs to be set explicitly by hand for this to fire. Claude Code sets `betas` itself whenever MCP tool search defers loading (`ENABLE_TOOL_SEARCH=true`, or `auto` once the tool block crosses its 10% threshold) — deferred loading sends the `anthropic-beta` tool-search header, which becomes a non-empty `betas` list on the request. Anthropic's own docs note the base-URL heuristic exists specifically because gateways vary in what they forward:

> When [`ANTHROPIC_BASE_URL`] is set to a non-first-party host, MCP tool search is disabled by default.

So a default-configuration Claude Code session pointed at Otari and routed to a bridged model is already safe. The failure mode here is any session, client, or config that explicitly re-enables tool search (or otherwise sets `betas`/`context_management`) against a bridged model — which is easy to do without realizing the two are connected, since the resulting error gives no hint that the beta header is the culprit.

### Proposed solution

In `classify_provider_error` (or immediately before/around it), special-case this specific `NotImplementedError` — or, better, have any-llm raise a purpose-built, catchable exception (e.g. `UnsupportedRequestFeatureError`) instead of a bare `NotImplementedError` so Otari doesn't have to string-match a message to classify it — and map it to:

- **HTTP 400**, not 500. This is a request-shape problem, not a provider or gateway fault.
- A detail that names the actual cause and the fix, e.g. *"context_management and betas require a model backed by Anthropic's native Messages API. Route to an Anthropic model, or stop sending these fields (for Claude Code: unset ANTHROPIC_BASE_URL from a bridged host, or don't force ENABLE_TOOL_SEARCH=true/auto against one)."*
- No "try again" language anywhere in the message — this is the one thing that must never be implied here.

This closes the loop that #486 opened: #486 wired `context_management`/`betas` through to `amessages()` for the native-Anthropic path. This issue is about the other branch of that same call — the bridged-provider fallback any-llm already guards — getting a response that tells the caller what actually happened instead of a generic retry prompt.

### Alternatives considered

- **Strip `context_management`/`betas` before dispatch when the resolved provider isn't native Anthropic** — would avoid the error entirely, but silently drops a feature the caller asked for rather than telling them it's unsupported for this model. Worse than a clear 400: a caller relying on compaction would see clean-looking success while the compaction they asked for silently never happened.
- **Leave it as-is and rely on client-side docs** — already tried; Claude Code's own docs already describe the safe default and I still hit this by explicitly overriding `ENABLE_TOOL_SEARCH`. A clear gateway-side error is the more durable fix, since it catches every path that can produce this shape, not just the one documented today.

njbrake:
## Description

A `/v1/messages` request that sets `context_management` or `betas` against a model whose provider is not a native Anthropic Messages API (a bridged/translated provider) is rejected by any-llm's default `_amessages` with a bare `NotImplementedError` that carries no HTTP status. `classify_provider_error` only recognizes exceptions it can extract a status code from, so this fell through to the generic 500, telling the caller to retry a request that is guaranteed to fail identically every time.

This special-cases that rejection and maps it to HTTP 400 with an actionable detail naming the real cause, so the caller sees a request-shape error instead of a misleading transient failure.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes #530

## Checklist

- [x] I ran the lint and unit test checks locally (`make lint`, `make typecheck`, `make test-unit`).
- [ ] Documentation was updated where necessary (no docs change needed: this is a runtime error-path fix, not part of the OpenAPI spec, so no spec/Postman regen).

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**
Claude Code

Fix drafted and verified by Claude in collaboration with njbrake. The gate is message-based and forward-compatible with `ANY_LLM_UNIFIED_EXCEPTIONS=1` once any-llm makes it the default: `convert_exception` wraps the raw `NotImplementedError` in a generic `ProviderError`, but the message still surfaces via `original_exception`, so the same check matches in both configurations.

The full integration suite (`make test`) needs Postgres/Testcontainers that this environment lacks; the changed code is a leaf classification function covered by focused unit tests, and the existing integration tests for native-Anthropic pass-through are unaffected.

**NOTE:**

## Summary

- Return HTTP 400 when bridged or translated providers reject `context_management` or `betas`.
- Provide actionable guidance that these fields require a native Anthropic Messages API provider.
- Add unit tests for direct and wrapped `NotImplementedError` cases.
- Keep unrelated provider errors unclassified.

### DIFF
--- a/src/gateway/api/routes/_pipeline.py
+++ b/src/gateway/api/routes/_pipeline.py
@@ -186,6 +186,10 @@ PROVIDER_BILLING_DETAIL = (
     "Top up the provider account, or route this model to a provider that has credit."
 )
 PROVIDER_RATE_LIMITED_DETAIL = "The provider rate-limited this request"
+PROVIDER_CONTEXT_MANAGEMENT_UNSUPPORTED_DETAIL = (
+    "context_management and betas require a model backed by Anthropic's native Messages API. "
+    "Route to an Anthropic model, or leave the betas/context_management fields unset for this model."
+)
 ALL_PROVIDERS_FAILED_DETAIL = "All upstream providers failed"
 ALL_PROVIDERS_TIMED_OUT_DETAIL = "All upstream providers timed out"
 SANDBOX_NOT_CONFIGURED_DETAIL = (
@@ -287,6 +291,16 @@ def _is_reasoning_effort_tools_conflict(exc: BaseException) -> bool:
     return "function tools" in upstream_error_message(exc).lower()
 
 
+# any-llm's default Messages path raises a bare NotImplementedError carrying this
+# exact message when a request sets context_management/betas against a provider
+# without a native Anthropic Messages API (see AnyLLM._amessages). classify_provider_error
+# gates on `isinstance(exc, NotImplementedError)` plus this message so an unrelated
+# NotImplementedError is not mistaken for this specific, actionable rejection.
+_CONTEXT_MANAGEMENT_UNSUPPORTED_MSG = (
+    "context_management and betas require a provider with a native Anthropic Messages API"
+)
+
+
 def classify_provider_error(exc: BaseException) -> ProviderErrorMapping | None:
     """Map an upstream provider exception to a safe, specific (status, detail).
 
@@ -306,6 +320,20 @@ def classify_provider_error(exc: BaseException) -> ProviderErrorMapping | None:
     kind, status_code = upstream_exception_shape(exc)
     if kind == "timeout":
         return ProviderErrorMapping(status.HTTP_504_GATEWAY_TIMEOUT, PROVIDER_TIMEOUT_DETAIL)
+    # any-llm's default Messages path rejects a request that sets
+    # context_management/betas against a provider without a native Anthropic
+    # Messages API (see AnyLLM._amessages). It currently raises a bare
+    # NotImplementedError (no HTTP status); once ANY_LLM_UNIFIED_EXCEPTIONS=1
+    # becomes the default, that NotImplementedError is wrapped in a generic
+    # ProviderError, so the message is the stable signal and upstream_error_message
+    # surfaces it both directly and through original_exception. Either way this is
+    # a request-shape problem, not a transient provider fault: it will never
+    # succeed by retrying, so it must surface as a 400 with an actionable detail
+    # rather than falling through to the generic 500. Gated on the fixed message
+    # so an unrelated NotImplementedError -- already an unclassifiable 500 path --
+    # is not repurposed as this specific error.
+    if _CONTEXT_MANAGEMENT_UNSUPPORTED_MSG in upstream_error_message(exc):
+        return ProviderErrorMapping(status.HTTP_400_BAD_REQUEST, PROVIDER_CONTEXT_MANAGEMENT_UNSUPPORTED_DETAIL)
     if status_code is None:
         return None
     # Account billing exhaustion, which several providers report as a 400/422
--- a/tests/unit/test_provider_error_classification.py
+++ b/tests/unit/test_provider_error_classification.py
@@ -15,6 +15,7 @@ from openai import APITimeoutError as OpenAIAPITimeoutError
 from gateway.api.routes._pipeline import (
     PROVIDER_BAD_REQUEST_DETAIL,
     PROVIDER_BILLING_DETAIL,
+    PROVIDER_CONTEXT_MANAGEMENT_UNSUPPORTED_DETAIL,
     PROVIDER_CREDENTIALS_DETAIL,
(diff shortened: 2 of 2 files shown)


src/gateway/api/routes/_pipeline.py
```diff
@@ -306,6 +320,20 @@ def classify_provider_error(exc: BaseException) -> ProviderErrorMapping | None:
...
+    # NotImplementedError (no HTTP status); once ANY_LLM_UNIFIED_EXCEPTIONS=1
+    # becomes the default, that NotImplementedError is wrapped in a generic
+    # ProviderError, so the message is the stable signal and upstream_error_message
+    # surfaces it both directly and through original_exception. Either way this is
+    # a request-shape problem, not a transient provider fault: it will never
+    # succeed by retrying, so it must surface as a 400 with an actionable detail
+    # rather than falling through to the generic 500. Gated on the fixed message
+    # so an unrelated NotImplementedError -- already an unclassifiable 500 path --
+    # is not repurposed as this specific error.
+    if _CONTEXT_MANAGEMENT_UNSUPPORTED_MSG in upstream_error_message(exc):
```
khaledosman, line 335: The substring test on `upstream_error_message(exc)` runs before the `status_code is None` branch and carries no `isinstance(exc, NotImplementedError)` guard, although both new comments (lines 291-300 and 323-334) describe the check as gated on that type. A provider 429 or 503 whose body echoes the offending request fields (any-llm surfaces the upstream body, directly or through `original_exception`) matches the sentinel text and gets reclassified as 400 `PROVIDER_CONTEXT_MANAGEMENT_UNSUPPORTED_DETAIL`: the caller is told to reroute a well-formed request, and `failure_status_code` writes 400 to the usage log, so the real rate limit or outage disappears from the dashboard's `errors_by_status_code` taxonomy. `test_unrelated_not_implemented_error_stays_unclassifiable` passes because the message differs, not because of any type check, so it cannot catch this.

Fix: add the `isinstance(exc, NotImplementedError)` guard the comments already describe, and run the check after the `status_code` branch so a real upstream status always wins. A test with a 429 carrying the sentinel text in its body would pin it.

khaledosman (approved):
Approving. One inline finding: the classification is broader than the comments describe and can swallow a genuine upstream 429/503.

---

njbrake:

Closing in favor of #537, which fixes #530 and makes this redundant.

Folding #530 into the error-detail change from #536 showed the exception type is the whole signal: any-llm raises `NotImplementedError` for any request feature a backend cannot express, so a type check does the job with no dependency on any-llm's wording. That removes the failure mode this PR carried, where an upstream reword would silently revert to the misleading 500 with tests still green, since the test here asserted our own copy of the message against our own constant.

#537 also passes the provider's message through on caller-fault statuses, so `PROVIDER_CONTEXT_MANAGEMENT_UNSUPPORTED_DETAIL` has nothing left to do: any-llm's own message already names the unsupported feature.

The diagnosis in #530 is what led to the general version.

PR #535, opened 2026-08-07
perf(dashboard): take provider dials and whole-table reads off page loads

njbrake:
## Description

Four dashboard GETs dialed upstream providers on the request path, bounded only by `model_discovery_timeout_seconds` (10s per unreachable provider) and the models.dev fetch timeout (15s). Each held a browser connection slot and a pooled database session for the whole dial.

**Discovery and the models.dev catalog now refresh in the background** and reads answer from cache, using the same refresher shape the alias, policy, provider and price caches already use in the lifespan. A provider that has never been dialed still dials, so a cold worker never claims it has no models. `?refresh=true` forces a live re-dial on `/v1/models/discoverable` and `/v1/providers/health`. `/v1/models` deliberately has no such flag: it takes any API key, and a fanout across every provider is an operator action. No new config; `model_cache_ttl_seconds = 0` still means "dial on read" and now also sets the refresh interval (floored at 30s).

**Usage and Activity no longer read the users and api_keys tables.** They paged both in full on every visit to name filter options and label rows. `/v1/usage` now returns `user_alias` and `api_key_name` per row (outer-joined, so a row whose owner was deleted survives, unlabelled), and `by_user` / `by_api_key` carry a `label` resolved in the same GROUP BY. Both pages build their pickers from a summary they already request.

Filter options are now the in-window entities ranked by spend, capped at 100, rather than every entity that ever existed. The User and API-key pickers accept a typed or pasted id on Enter, so an entity below that rank is still reachable.

Also: `retry: false` on the four discovery-backed queries, and a 30s bound in `apiFetch` (5 minutes for the bulk usage delete and reprice, whose duration scales with the data).

### Follow-up fixes in this PR

Review found a shutdown hang that CI hit as two 120s unit-test timeouts. Cancelling a task is a request, not a guarantee: anyio's `CancelScope.__exit__` calls `host_task.uncancel()` whenever its own scope was cancelling, and httpx runs its per-operation timeouts as those scopes, so a shutdown cancel racing one of them is absorbed. The refresher loop then resumes and sleeps out a full interval, and the lifespan's unbounded `await task` never returns. `_stop_refresher` now waits a bounded 5s and abandons the task with a warning; it applies to all six refreshers, and it also stops a refresher's unexpected error from aborting the rest of shutdown. Unit tests were starting the real refreshers too, so the suppression fixture moved to the root conftest.

Plus: a failed models.dev fetch is no longer served at any age (a transient outage disabled enrichment for 24h); both refreshers re-check their runtime-settable knob per tick; `apiFetch` converts a timeout on the body read, not just on `fetch()`.

## PR Type

- [x] Bug Fix
- [ ] New Feature
- [ ] Refactor
- [x] Documentation
- [ ] Infrastructure / CI

## Relevant issues

None filed; found while profiling dashboard page loads.

## Measured

Same browser repro, gateway with six unreachable providers:

| Request | Before | After |
|---|---|---|
| `/v1/providers/health` | 10009ms | 14ms |
| `/v1/models` | 8577ms | 11ms |
| `/v1/models/discoverable` | 8577ms | 3ms |
| `/v1/models/metadata` | 236ms | 3ms |

Activity and Usage now issue 4 and 3 requests respectively, all under 25ms, and neither touches `/v1/users` or `/v1/keys`.

## Checklist

Regression tests verified in both directions, failing before the fix and passing after: the three `*_serves_an_expired_cache_without_dialing` tests, `never reads the whole users or api_keys table`, the `_stop_refresher` shutdown tests, the models.dev negative-TTL test, and the `apiFetch` body-stall test. 1414 unit, 1020 integration, 488 dashboard. `make lint`, `make openapi-check`, `make postman-check` clean. Dashboard bundle rebuilt.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

Claude Opus 5 (1M context) via Claude Code.

Implemented by Claude Code, then reviewed by a separate Claude Code agent with no access to the implementing session's reasoning. That review found the shutdown hang, the models.dev negative-TTL regression, and the filter-picker gap, each reproduced in isolation before being fixed.

**NOTE:**

## Summary

- Moved model discovery and catalog updates to background refresh tasks.
- Served model and provider data from cache to reduce dashboard request time.
- Added forced live refresh support for discovery and provider health endpoints.
- Reduced Usage and Activity data loading by using labels from usage results instead of full user and API-key lists.
- Added request time limits and longer limits for bulk operations.
- Improved refresh recovery, shutdown handling, and runtime cache configuration updates.
- Updated dashboard bundles, documentation, Postman examples, and regression tests.

## Technical notes

- Added cache refresh and stale-read controls to model discovery, provider health, and catalog services.
- Preserved dial-on-read behavior when `model_cache_ttl_seconds` is `0`.
- Added user and API-key labels to usage rows and grouped summaries.
- Disabled retries for discovery-backed queries.
- Added tests for caching, refresh failures, shutdown timeouts, usage labels, and frontend request timeouts.


src/gateway/services/model_discovery_service.py
```diff
@@ -167,7 +180,17 @@ async def get_or_discover(
         provider awaits that one task, so a slow provider is dialed once, not
         once per request. ``discover`` is expected to report failure by
         returning a ``ProviderDiscovery`` with ``error`` set rather than raising.
+
+        ``serve_stale`` accepts a cached result of any age, so a read never waits
+        on a provider that has been dialed at least once. A provider with no
+        entry at all still dials (single-flighted), which keeps a cold worker
+        correct rather than briefly claiming the provider has no models.
         """
+        if serve_stale:
```
khaledosman, line 197: `stale()` returns the cached entry regardless of whether the attempt succeeded, so a cached failure is served at any age and `model_discovery_negative_ttl_seconds` (30s) no longer applies on any read path. One timed-out dial during a refresh tick makes every model from that provider disappear from `GET /v1/models` (errored providers get dropped) and marks it unreachable on `/v1/providers/health` for the full refresh interval, 300s by default. Before this change the next request re-dialed after 30s.

Fix: gate the `serve_stale` short-circuit on a successful entry, and fall through to a live dial when the cached entry carries an `error` and is older than the negative TTL. Same shape at lines 657-658.

src/gateway/services/model_catalog_service.py
```diff
@@ -87,10 +87,20 @@ def clear_catalog_cache() -> None:
...
+    ``serve_stale`` accepts a cached blob at any age, so a read never pays the
+    15s fetch timeout; the background refresher is what keeps it current.
+    ``force`` is the refresher's own read: it treats every entry as expired so
+    the fetch actually happens.
+    """
     if _cache.at == 0.0:
         return _MISS
+    if force:
+        return _MISS
+    if serve_stale:
```
khaledosman: `_read_cache(serve_stale=True)` returns `_cache.data` before the `ok` check, so a cached failure (data `None`) is handed back as a valid answer and `_NEGATIVE_TTL_SECONDS` (60s) is dead on the read path while the refresher floor is 300s. If the first refresher tick fails (models.dev briefly unreachable, or the response exceeds `_MAX_CATALOG_BYTES`), `GET /v1/models/metadata` answers `available: false` with no entries and keeps doing so for at least 300s, so the Models page shows no capabilities, context window, or cutoff for anything.

Fix: short-circuit only when `_cache.ok`; when the entry is a failure, let the negative TTL apply as it did before. Same at line 168.

web/src/api/client.ts
```diff
@@ -73,17 +73,32 @@ export async function deleteSession(): Promise<void> {
...
+// Callers pass their own `signal` to override.
+const REQUEST_TIMEOUT_MS = 30_000;
+
 export async function apiFetch<T>(path: string, init: RequestInit = {}): Promise<T> {
   const headers = new Headers(init.headers);
   headers.set("Accept", "application/json");
   if (init.body != null && !headers.has("Content-Type")) {
     headers.set("Content-Type", "application/json");
   }
+  const signal = init.signal ?? AbortSignal.timeout(REQUEST_TIMEOUT_MS);
```
khaledosman, line 116: The unconditional 30s `AbortSignal.timeout` also covers the unbounded destructive mutations, none of which pass their own signal: `DELETE /v1/usage` with `by_filter` (the "select all N matching" path), `POST /v1/usage/set-price`, `POST /v1/pricing/refresh`, `POST /v1/provider-credentials/reencrypt`. Over a large window the statement outlives 30s, the client aborts with `ApiError(0, ...)` and the dashboard reports failure while the transaction commits server-side, so the operator retries an already-applied delete or reprice with no way to tell from the UI which rows are gone.

Fix: make the timeout overridable per call, or give the bulk mutation hooks a longer cap, instead of one blanket default.

web/src/pages/ActivityPage.tsx
```diff
@@ -1058,9 +1062,9 @@ export function ActivityPage() {
   // it is not a chip). Values show the human label where one exists.
   const labelFrom = (options: { value: string; label: string }[], value: string) =>
     options.find((o) => o.value === value)?.label ?? value;
-  const userOptionsList = (users.data ?? []).map((u) => ({
-    value: u.user_id,
-    label: u.alias ? `${u.alias} (${u.user_id})` : u.user_id,
+  const userOptionsList = realGroups(modelSummary.data?.by_user).map((r) => ({
```
khaledosman: Deriving the user and key options from the usage breakdown means the picker only offers entities that both appear in the selected window and land in the top `_BREAKDOWN_TOP_N` (100) by spend. A key whose last request falls outside the window is absent from the combo box, so there is no way to select it and widen the window; past 100 users, most of them are unreachable through the UI. The previous `fetchAllUsers`/`fetchAllKeys` pickers always listed everything. Same at line 176 and `UsagePage.tsx:431`.

Fix: keep the full list as the option source (or merge the breakdown rows into it) and use the breakdown only for ordering and the spend annotations.

web/src/pages/ActivityPage.tsx
```diff
@@ -884,8 +886,6 @@ export function ActivityPage() {
       start_date: win.start,
       end_date: win.end,
       status: statusFilter || undefined,
-      user_id: userFilters.length > 0 ? userFilters : undefined,
-      api_key_id: apiKeyFilters.length > 0 ? apiKeyFilters : undefined,
       source: sourceFilter || undefined,
```
khaledosman, line 891: Dropping `user_id`/`api_key_id` from `modelSuggestFilters` also changes the model typeahead and the source picker, which share this summary; the comment just above still says "other filters applied". With Activity filtered to one user, the typeahead now suggests models only other users called, and selecting one returns an empty table with no hint that the combination could never match. The `by_source` options below (line 913 onward) degrade the same way, and `UsagePage.tsx:417` has the same drop.

Fix: keep the entity filters on the suggestion query, or split the suggestion summary from the unfiltered one so only the intended consumer loses them.

src/gateway/main.py
```diff
@@ -169,6 +181,17 @@ async def lifespan(app: FastAPI) -> AsyncGenerator[None, None]:
             # served the confirm; reload it on a TTL so sibling workers and replicas
             # converge, the same way aliases and provider credentials do.
             price_refresher = asyncio.create_task(run_price_snapshot_refresher())
+            # Discovery is the one cache that used to be filled on the request
+            # path, which put model_discovery_timeout_seconds (10s per
+            # unreachable provider) on a dashboard page load and held that
+            # request's database session open for the whole dial. The refresher
+            # owns the dialing now and reads answer from the cache. Not awaited:
+            # priming runs on its first tick so a slow provider cannot delay boot.
+            if background_discovery_enabled(config):
```
khaledosman: The refresher is created once here from `background_discovery_enabled(config)`, but `models.py:449`, `models.py:575` and `providers.py:274` re-evaluate that same predicate per request, and `model_cache_ttl_seconds` is mutable at runtime via `PATCH /v1/settings` (`settings.py:90`). Boot with the TTL at 0 (documented as disabling caching), then set it to 300 on the Settings page: the read paths now serve stale from a cache no task refreshes, so each provider is dialed once by the first cold read and that snapshot is served until restart. A model newly added at the provider never shows up, and a provider that recovers from an outage stays red on the health page.

Fix: have the settings mutation start or stop the refresher, or make the read paths treat "no refresher running" as not eligible for stale.

src/gateway/api/routes/models.py
```diff
@@ -432,7 +441,16 @@ async def list_models(
...
-            discovered = await discover_all_models(config, provider_filter=provider)
+            # Cache-only when a refresher owns the dialing. This endpoint is
+            # reachable with any API key, so it deliberately has no ``refresh``
+            # escape hatch: forcing a fanout across every configured provider is
+            # an operator action, and lives on the master-key-gated
+            # /v1/models/discoverable and /v1/providers/health instead.
+            discovered = await discover_all_models(
+                config,
+                provider_filter=provider,
+                serve_stale=background_discovery_enabled(config),
```
khaledosman, line 452: With `list_models` no longer re-dialing, nothing on the request path renews `cached_at`, but `get_model_detail` (line 671) still peeks with the TTL-bounded `cache.get(provider_prefix, ttl)`. `run_discovery_refresher` sleeps `max(30, model_cache_ttl_seconds)` *after* each dial completes, so an entry stored at T expires exactly when the next dial starts and stays expired until that dial finishes (up to 10s per provider). In that window `GET /v1/models/{provider:model}` can 404 a model that `GET /v1/models` is listing at the same moment, for any provider model with no `model_pricing` row and no genai-prices fallback.

Fix: give `get_model_detail` the same stale-tolerant read the list path now uses.

src/gateway/services/model_discovery_service.py
```diff
@@ -585,3 +628,69 @@ async def discover_all_models(
...
+# --------------------------------------------------------------------------- #
+
+# Floor on the refresh cadence. ``model_cache_ttl_seconds`` is an operator-facing
+# freshness knob, and a very small value would otherwise turn into a re-dial storm
+# against every configured provider. Reads still serve the cache below this floor;
+# they are simply refreshed no faster than this.
+_MIN_REFRESH_INTERVAL_SECONDS = 30.0
+
+
+def background_discovery_enabled(config: GatewayConfig) -> bool:
```
khaledosman, line 652: `background_discovery_enabled` looks only at `model_cache_ttl_seconds` and ignores `config.model_discovery`, and `main.py:190` starts the refresher on that basis. An operator who set `model_discovery: false` specifically to stop the gateway dialing providers still gets a `list_models` fan-out across every configured provider every `max(30, ttl)` seconds (5 minutes by default) for the life of the process, even with nobody touching the dashboard. On a provider that meters or rate-limits `list_models` that is new, permanent, unrequested traffic.

Fix: add `and config.model_discovery` to the predicate.

khaledosman (approved):
Approving so this is not held up on review. Eight findings inline. The two `serve_stale` ones are the same bug in two caches (a cached *failure* is served as an answer at any age, so both negative TTLs are dead on every read path and one transient error is pinned for the refresh interval); those are the two I would fix before this ships.

---

src/gateway/services/model_discovery_service.py
njbrake: _Note: this reply was drafted by Claude via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

Real, and fixed in 279e9025, though not the way you suggested. Falling through to a live dial once a cached failure ages past the negative TTL puts the 10s timeout back on a read, which is the thing this PR exists to remove.

Instead the refresher comes back on the negative TTL after any round that reported a failure, floored at 30s like the success interval. Reads stay off the dial path entirely, and a recovered provider still reappears within 30s by default, so the old bound is preserved rather than traded away. See `_refresh_interval(config, had_failure=True)` and `TestRefresherCadence`.

An unexpected error in the round counts as a failure too, so a raised refresher does not sleep out the full success interval before retrying.

njbrake:

All review threads addressed and resolved. Map of comment to fix, since two commits are involved:

| Comment | Fix |
|---|---|
| `stale()` serves a cached failure at any age | 279e9025, though not as suggested. See the [thread reply](https://github.com/mozilla-ai/otari/pull/535#discussion_r3736262019). |
| `_read_cache(serve_stale=True)` returns before the `ok` check | 4d9debe4. Only a successful blob is served stale; a failure falls back to the 60s negative TTL. |
| 30s deadline on unbounded destructive mutations | 4d9debe4. The bulk usage delete, the bulk reprice, and the pricing-snapshot refresh get 5 minutes. |
| User and key pickers capped at the top 100 by spend | 4d9debe4. `allowsCustom` on both, so Enter commits a pasted id as Model already did. |
| Dropping the entity filters changed the model typeahead and source picker | 279e9025. Those two keep their entity filters; the user and key pickers get their own summary. Same split on the Usage page. |
| Refresher created once from a runtime-settable predicate | 4d9debe4. Both refreshers re-check their knob per tick. |
| `get_model_detail` still peeks with a TTL-bounded read | 279e9025. Stale-tolerant like the listing, so it cannot 404 a model the listing is serving. A negatively cached provider still reports no models. |
| `background_discovery_enabled` ignores `model_discovery` | 279e9025. Predicate now requires it, so a gateway with discovery off makes no unattended `list_models` calls. |
| Stray fourth quote opening a docstring | 279e9025. |

Also from CodeRabbit: the `model_cache_ttl_seconds` and `model_discovery` rows now say what the detail endpoint does, what a zero TTL means for the refresher, and that a health re-check coalesces with a recent dial.

The one finding I disagreed with is CodeRabbit's suggestion to start the refreshers in hybrid mode. `register_routers` never mounts the models or providers routers there, so no read path could serve a stale result, and `_validate_platform_config` rejects a non-empty `providers` block in hybrid anyway.

New tests: `TestRefresherCadence`, `test_model_detail_agrees_with_the_listing_on_a_stale_entry`, `test_model_detail_does_not_serve_a_cached_failure_as_a_model`, and an ActivityPage test pinning that the model typeahead keeps the user filter while the user picker drops it. The detail-endpoint one was verified failing against the TTL-bounded read.

Issue #533, opened 2026-08-07
[BUG] Non-streaming Anthropic requests with large max_tokens fail with opaque 502 (anthropic SDK timeout guard)


peteski22:
> **Updated 2026-09-09: the suggested fix below has changed. Do not implement the original one.**
> When this was filed, otari resolved `any-llm-sdk==1.24.0`, which had no first-class `timeout`,
> so `client_args` was the only lever. mozilla-ai/any-llm#1255 added one on 2026-08-11, and otari
> picked it up when #777 moved the floor to 1.25.0 on 2026-08-25. #799 implemented the original
> prescription and was closed for that reason.

**Otari version:** current `main` (v0.5.1 and later).
**Python version:** N/A — the failure is a client-side pre-flight check, independent of Python version (`requires-python >=3.13`).
**Run mode:** Both. Standalone resolves the client through `get_provider_kwargs` (`src/gateway/services/provider_kwargs.py`); hybrid through `build_attempt_client_args` (`src/gateway/api/routes/_platform.py`).
**How you're running Otari:** From source (`main` checkout).
**Provider and model:** `anthropic:claude-opus-5` (any Anthropic model).

### What happened

A **non-streaming** chat completion for an Anthropic model with `max_tokens` above **~21,333** fails with a generic `HTTP 502 {"detail": "LLM provider error"}` and **no tokens billed** — the request is blocked *before* anything is sent to Anthropic.

Root cause: the Anthropic provider builds its `AsyncAnthropic` client with **no explicit `timeout`**, so the anthropic SDK's `_calculate_nonstreaming_timeout` guard runs and raises a **client-side `ValueError`** for any non-streaming request whose `max_tokens` implies more than 10 minutes at the SDK's worst-case rate:

```
expected_time = 3600 * max_tokens / 128_000   # assumes a full 128k generation takes 1 hour
raise if expected_time > 600                   # i.e. max_tokens > 128000/6 ≈ 21,333
```

The guard condition is `not stream and not is_given(timeout) and self._client.timeout == DEFAULT_TIMEOUT` (`anthropic/resources/messages/messages.py`), so it fires on the `max_tokens` ceiling even when the actual response is a single sentence, and only while no timeout was set.

[...]

PR #537, opened 2026-08-07
fix(errors): return the provider's own message on a caller-fault rejection

### LINKED ISSUE #530: [BUG] context_management/betas against a bridged (non-native-Anthropic) provider surfaces as a misleading 500, not a 400

### LINKED ISSUE #536: [FEATURE] Pass upstream provider errors through on caller-fault statuses instead of curating fixed detail strings

### Problem or use case

Otari replaces every upstream provider error message with a fixed string before
returning it. Each time that string turns out to hide something the caller
needed, we add a hand-written constant and a new probe inside
`classify_provider_error`. We have now done this five times, and the rate is
climbing:

- #252 (closed): an unknown provider in the model name returned a bare 500.
- #331 / #385: OpenAI's "use /v1/responses or set reasoning_effort to none" was
  flattened to "check the model name and parameters".
- #483: an Anthropic account out of credit was reported as a bad request. A
  downstream app failed every agent turn for roughly 10 hours, and the real
  cause appeared only in the gateway's own logs.
- #530 / #534: `context_management`/`betas` against a bridged provider surfaces
  as a 500 that implies retrying will help.
- #533 (open): the Anthropic SDK's non-streaming `max_tokens` guard raises a
  bare `ValueError`, so it collapses to `502 "LLM provider error"`.
- #183 (open): "the gateway masks the underlying provider exception, so the real
  cause is not visible to the caller."

Three of those probes landed within sixteen days. Every one of them string
matches or type sniffs another package's internals, so every one is a silent
regression waiting for an upstream reword. #381 was already that failure mode
for timeout types.

The posture traces to mozilla-ai/any-llm#839 (merged 2026-02-23), when the
gateway still lived in any-llm as any-llm-gateway. It reached this repo with the
extraction, as `5ca8c7a9`, the fourth commit here.

That PR is worth reading before we decide anything. It was a proactive hardening
sweep rather than a response to a reported leak, and it bundled three unrelated
fixes: database connection strings in the health endpoint, filesystem paths in
Vertex auth errors, and provider exceptions. The first two are clearly right.
The third was three lines in `chat.py`, swapping `detail=f"Error calling
provider: {e!s}"` for a fixed string and adding a server-side log, on the
grounds that provider exceptions "may contain API keys/URLs". That is a
plausible hypothesis, not an observed leak, and it covered one route with no
classifier behind it.

What has changed since is scope, not evidence. That three-line default is now a
repo-wide invariant with a conformance test and a security-review checklist item
(`security-review.instructions.md:137`), and the reasoning behind it has not
been revisited.

Meanwhile #197, which introduced the classifier, was motivated by diagnosability
rather than security: a provider failure had been misfiled as a pricing bug
because the generic message hid it. Worth noting that #331 already recorded that
Portkey passes provider errors through with no special casing.

### Proposed solution

Split the two jobs `classify_provider_error` currently fuses.

Keep curated fixed strings wherever we deliberately reassign blame, because that
is judgment the provider cannot supply: upstream 401/403 to 502, billing to 502
along with its failover eligibility, 429, and all 5xx. Credentials and gateway
internals concentrate in those cases.

Pass the upstream message through on the caller-fault branch (400/422/404),
where it is most useful and least sensitive. Gate it behind a config flag,
default off in hybrid mode since the platform's error text is not ours to
forward, and redact key-shaped tokens (`sk-`, `Bearer `, `api_base` URLs).

That would have pre-solved #252, #183 and #530 with no new constants. Note it
does not fully cover #530 or #533: both are bare exceptions carrying no HTTP
status, so they still need a classification rule. The durable fix there is a
typed exception from any-llm rather than a message match.

Worth doing regardless of the above: mint a gateway request ID on every
response. `X-Otari-Request-ID` is set at two sites, both hybrid only, so a
standalone operator hitting the #483 scenario has no way to correlate a
sanitized 502 with the log line that holds the answer.

### Alternatives considered

- **Keep adding cases.** Works, but each costs an issue, a PR and a brittle
  probe, and only helps the cases someone bothers to report.
- **Full pass-through, including 401/403 and 5xx.** Rejected: it would blame the
  caller for the operator's expired key and would expose the most sensitive
  messages.
- **Fix each case upstream in any-llm.** Correct for typed exceptions, but
  any-llm#1189 was closed as not fixable there, so it does not generalize.

njbrake:
## Description

Implements #536. otari replaced every upstream provider message with a fixed string, so a caller who sent a bad request was told to "check the model name and parameters" while the provider's actual explanation reached only the gateway log. Each time that hid something, we added a hand-written constant and a probe into another package's wording: five bug reports (#252, #331, #483, #530, #533) and three probes in sixteen days.

Detail text now splits on whose fault the failure is.

**Caller's fault (400/422/404):** the provider's own message, redacted and capped at 400 characters. The provider is the only party that diagnosed the request, and no fixed string we write substitutes for that.

**Gateway's fault (rejected credentials, exhausted provider account, 5xx, unclassifiable):** the fixed string stays. Those name the operator's account rather than anything the caller can act on, and are where a raw message is most likely to carry a key or an `api_base`. Redaction is the second line, not the only one.

Status classification is unchanged, including the deliberate reassignments (401/403 and billing exhaustion surface as 502) and the failover eligibility #483 attached to billing errors.

**Also fixes #530, and supersedes #534.** any-llm raises `NotImplementedError` when a request asks a provider for something its backend cannot express. It carries no HTTP status, so it fell through to a generic 500 telling the caller to retry something guaranteed to fail identically. The exception type is the whole signal, so a type check replaces #534's message match and needs no knowledge of any-llm's wording. It unwraps `original_exception`, so it survives `ANY_LLM_UNIFIED_EXCEPTIONS=1`. I suggest closing #534 in favor of this; its `PROVIDER_CONTEXT_MANAGEMENT_UNSUPPORTED_DETAIL` constant would be dead on arrival here.

Removes `PROVIDER_REASONING_TOOLS_UNSUPPORTED_DETAIL` and its probe. OpenAI's message already names the remedy, so passing it through covers #331 and every future rewording of it.

### Checked against otari-ai

The platform has no coupling to any of these detail strings. Its completion proxy (`gateway_completion_proxy_service._error_detail`) reads `detail` generically and handles both envelopes, so nothing breaks. It does surface the detail in the Playground, which is the intended outcome: a workspace user debugging their own request sees what the provider objected to.

One residual worth naming: mozilla.ai's own gateway serves managed models on platform-owned credentials, so an exposed message there describes a request made on the platform's account. Credential shapes, URLs and `org-`/`proj-` style account identifiers are redacted, and the statuses where account identity actually shows up (401/403, billing) never pass text through. If we want managed traffic to be stricter than that, the clean seam is a per-mode gate, which I left out deliberately rather than shipping a flag nobody asked for.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes #536
Fixes #530
Supersedes #534

## Checklist

- [x] I ran the lint and unit test checks locally (`make lint`, `make typecheck`, `make test-unit`).
- [ ] If the API contract changed, I regenerated the OpenAPI spec. (No schema change: details are response bodies. `make openapi-check` and `make postman-check` both pass.)

Verification: `make lint` and `make typecheck` clean (310 files). Unit 1406 passed. Integration 1015 passed against a local PostgreSQL 18. The two tests that make real outbound calls (`test_error_detail_leakage::test_provider_error_does_not_leak_details`, `test_streaming_error_event::test_streaming_creation_error_returns_http_error`) were deselected; I confirmed the first fails identically on clean baseline HEAD in this sandbox, so it is the environment note in `AGENTS.md` rather than a regression.

Two changes came out of test failures rather than design: `upstream_error_message` concatenated both `exc.message` and `str(exc)`, which stuttered once the text became caller-visible, and a message with no letters (some SDKs stringify a failure as just `"404"`) now falls back rather than passing through a non-explanation.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

Drafted by Claude via back-and-forth with @njbrake; the reasoning and decisions are his. The history behind the posture this reverses is in #536.

## Summary

- Caller errors now include safe provider messages for `400`, `422`, and `404` responses.
- Gateway errors retain fixed details for credentials, billing, server errors, and unknown failures.
- Unsupported provider features now return `400` responses, including wrapped `NotImplementedError` cases.
- Removed reasoning-tools-specific error handling.
- Added documentation and expanded unit and integration tests.

## Technical notes

- Provider messages are limited to 400 characters.
- Credentials, account identifiers, URLs, payloads, and token-like data are redacted.
- Status classification and failover behavior remain unchanged.
- Lint, type checks, unit tests, and integration tests pass.
- No API schema changes were made.


src/gateway/api/routes/_pipeline.py
```diff
@@ -250,53 +250,55 @@ class _PendingUsageReport(NamedTuple):
...
+    check against ``exc`` alone would stop matching.
     """
-    for candidate in (exc, getattr(exc, "original_exception", None)):
-        param = getattr(candidate, "param", None)
-        if isinstance(param, str):
-            return param
-    return None
+    return any(
+        isinstance(candidate, NotImplementedError)
+        for candidate in (exc, getattr(exc, "original_exception", None))
```
khaledosman: This unwraps exactly one level, but the docstring above claims the check survives `ANY_LLM_UNIFIED_EXCEPTIONS=1`. `upstream_exception_shape` (`_platform.py:670`) deliberately *loops* with an `id()` seen-set for the same reason. A `ProviderError(original_exception=ProviderError(original_exception=NotImplementedError(...)))` falls through `status_code is None` to `None` and #530 regresses to a generic 502, with no test catching it.

`upstream_error_message` (`_platform.py:723`) has the same one-level limit, so a doubly-wrapped caller-fault error also loses its message and silently takes the fallback string.

Reuse the chain walk instead of re-deriving a shallower one next to it.

src/gateway/api/routes/_pipeline.py
```diff
@@ -250,53 +250,55 @@ class _PendingUsageReport(NamedTuple):
...
+    stringify a failure as bare punctuation or the status code itself, and
+    "404" is not an explanation the caller did not already have from the
+    status. Requiring one letter is a low bar deliberately: it rejects the
+    empty cases without second-guessing a provider that wrote a real sentence.
     """
-    if _upstream_error_param(exc) != "reasoning_effort":
-        return False
-    return "function tools" in upstream_error_message(exc).lower()
+    redacted = redact_upstream_message(upstream_error_message(exc))
+    if not any(char.isalpha() for char in redacted):
```
khaledosman: The guard runs on the *redacted* text, and `_REDACTION_PLACEHOLDER` contains letters, so a message that is entirely a secret shape satisfies it. Verified against `_SECRET_SHAPES`: an upstream 400 whose message is just the `api_base` (`"https://api.openai.com/v1"`) produces a detail of literally `[redacted]` rather than `PROVIDER_BAD_REQUEST_DETAIL`.

Check for alpha outside the placeholders, or fall back when stripping every placeholder leaves nothing.

src/gateway/api/routes/_platform.py
```diff
@@ -711,20 +712,88 @@ def upstream_error_message(exc: BaseException) -> str:
...
+# Applied in order, so a URL is masked before the trailing catch-all can pick
+# apart its path. Each targets a shape that carries secrets rather than meaning:
+# nothing here removes text a caller needs to understand what it got wrong.
+_SECRET_SHAPES: tuple[re.Pattern[str], ...] = (
+    # Authorization-header credentials, whatever scheme.
+    re.compile(r"(?:Bearer|Basic|Token)\s+[A-Za-z0-9._~+/=-]{8,}", re.IGNORECASE),
+    # Provider key formats: a known prefix plus the key body. Covers OpenAI
+    # (sk-, sk-proj-), Anthropic (sk-ant-), Groq (gsk_), xAI (xai-), and
+    # otari's own tk_ / gw_ tokens.
+    re.compile(r"\b(?:sk|pk|rk|gsk|xai|tk|gw|api|key)[-_][A-Za-z0-9._-]{8,}", re.IGNORECASE),
```
khaledosman: The `api` and `key` alternatives are not key prefixes: no provider issues a credential named `api-…` or `key-…`, and `api_key=…` is already covered by the named-credential pattern on line 759. They only add false positives on hyphenated identifiers (`api-inference…`, `api_response_format`), which deletes exactly the diagnostic text this PR exists to surface. Drop them from the alternation.

src/gateway/api/routes/_platform.py
```diff
@@ -711,20 +712,88 @@ def upstream_error_message(exc: BaseException) -> str:
...
+    # platform's own provider account, so an error naming that account would
+    # tell a workspace user whose credentials served them.
+    re.compile(r"\b(?:org|proj|acct|account)[-_][A-Za-z0-9]{6,}", re.IGNORECASE),
+    # Any absolute URL. A self-hosted or proxied ``api_base`` is gateway
+    # topology the caller has no business learning, and a credential is
+    # sometimes embedded in one.
+    re.compile(r"\bhttps?://\S+", re.IGNORECASE),
+    # Catch-all for key material with no recognizable prefix. Long unbroken
+    # alphanumeric runs are tokens, not prose.
+    re.compile(r"[A-Za-z0-9_-]{40,}"),
```
khaledosman, line 775: Prefixless 32-char keys are uncovered and the catch-all floor sits just above them: Azure OpenAI and Mistral keys are 32 chars with no prefix. Confirmed by running the full pattern set — `"Incorrect api-key provided: 0123456789abcdef0123456789abcdef"` passes through with the key intact, because line 759 needs `api-key` immediately followed by `=`/`:` and line 754 needs 8+ chars after the separator.

Worth closing now given this PR makes "add a shape when a provider emits an uncovered format" a review gate in `security-review.instructions.md`.

docs/api-reference.md
```diff
@@ -4,6 +4,14 @@ All endpoints are under `http://localhost:8000` by default.
 
 For full request/response schemas, see the [OpenAPI spec](public/openapi.json) or the interactive docs at `/docs` when Otari is running.
 
+## Provider error details
+
+When an upstream provider fails, what the `detail` field contains depends on whose problem it is.
+
+If the provider rejected **your request** (HTTP 400, 422 or 404), the detail is the provider's own message, so it names the parameter, model or limit at fault. Credential-shaped tokens, URLs and account identifiers are stripped from it, and it is capped at 400 characters.
```
khaledosman: These read as response statuses, but `classify_provider_error` remaps an upstream 422 to a 400, so a caller reading the API reference never sees a 422. Either say "the provider returned 400, 422 or 404", or list the response statuses as 400/404.

khaledosman (approved):
Four findings inline: one correctness, two on the redaction set, one doc. None block.

One more that has no diff line to anchor to — **`src/gateway/api/routes/_platform.py:226`**, the `_provider_failure_http_exc` docstring:

> Falls back to a 502 carrying `fallback_detail` when the failure has no signal we can safely surface. **The classified detail is always a fixed string, so it cannot leak the raw upstream message.**

That is no longer true. `_provider_failure_http_exc` is the hybrid-mode terminal path, so it returns the redacted upstream message for 400/422/404 like everything else. This PR deleted the assertion that backed the claim (`test_platform_terminal_exc_uses_classified_status` lost `assert exc.detail == PROVIDER_MODEL_NOT_FOUND_DETAIL`) but left the claim standing, in exactly the function where the managed-credential residual the description names actually lands. Rewrite it to point at the fault split in `classify_provider_error`.

Gates checked with no finding: no route/schema/docstring change so the OpenAPI + Postman artifacts stay valid; `docs/api-reference.md` is not in the bundled guide (only `docs/dashboard.md` is), so the dashboard bundle is not stale; `_classify_upstream_error` is independent of `classify_provider_error` and a bare `NotImplementedError` already returned `(False, "unknown")`, so failover behavior and the #483 billing carve-out are genuinely unchanged; the mid-stream SSE error payload is a fixed constant (`streaming.py:67-82`), so no upstream text reaches an SSE frame; and the bare `except Exception -> adapter.provider_error` funnels at `_pipeline.py:2665`/`:3344` are wider than "provider error", but sandbox/web-search/guardrails all wrap `httpx.HTTPStatusError` into status-free typed errors first, so operator-infrastructure text does not reach the classifier.

---

PR #542, opened 2026-08-07
fix(billing): record charge lines for embeddings, images, moderations, and rerank

### LINKED ISSUE #432: Passthrough usage (images, audio, embeddings, rerank) has no charge lines and no non-token unit

Related: #376 (metered-pricing epic).

## Summary

Image, audio, embedding, rerank, and moderation usage is recorded and priced, but it has no auditable charge lines and no representation in a UI that reports exclusively in tokens. Operators with a mixed workload have spend that is real and nearly invisible.

## 1. The passthrough path never writes charge lines

`run_passthrough` (`src/gateway/api/routes/_passthrough.py:244`) builds its `UsageLog` with tokens, cost, and latency — but never sets `billing_meters` or `pricing_breakdown`. Those two fields are only ever populated in:

- `src/gateway/api/routes/_pipeline.py:1153-1154` (chat/messages/responses)
- `src/gateway/services/external_usage_service.py:323-324` (imported usage)

So every chat row and every imported row carries auditable per-meter charge lines, and every image/audio/embedding/rerank/moderation row carries a bare `cost` with no explanation of how it was derived. That is an inconsistency in the audit trail rather than a cosmetic gap: for these endpoints the cost comes from an endpoint-specific `compute_cost` callback, which is exactly the case where a breakdown is most useful.

## 2. There is no non-token unit

The billable meter set is entirely token-based (`metered_pricing.py:147`: `total_input_tokens`, `fresh_input_tokens`, `cache_read_tokens`, `cache_write_tokens`, `cache_write_1h_tokens`, `completion_tokens`). There is no unit for generated images, audio seconds, or characters.

The consequence in the UI: an image-generation row renders `Tokens: —` next to a cost, contributes nothing to any token chart, and is invisible in the token totals and series. The same applies to speech and transcription. An operator running a mixed workload sees a cost number they cannot decompose or trend.

## Proposal

- Have `run_passthrough` accept and persist `billing_meters` / `pricing_breakdown` from its `compute_cost` callback, so all served endpoints produce the same audit trail.
- Introduce a unit concept alongside tokens (e.g. `units` + `unit_type`, or modality-specific meters such as `images`, `audio_seconds`, `characters`) so non-token usage is countable and chartable.
- Make the dashboard's headline measures modality-aware: cost is already universal, so lead with cost for mixed workloads and treat tokens as one unit type among several rather than the only one.

## Scope note

This sits inside #376's problem space — it is the same underlying issue (`UsageLog` cannot preserve the provider meters that explain a bill) surfacing on the passthrough endpoints rather than the chat ones. Filing separately because the missing-charge-lines half is a small, self-contained inconsistency that can be fixed independently of the wider pricing engine.

AmirF194:
## Description

Passthrough usage rows for embeddings, images, moderations, and rerank set `cost` but never `billing_meters` or `pricing_breakdown`, unlike the chat/messages/responses path (`calculate_metered_cost`) and the imported-usage repricing path, both of which set all three together. Every one of those rows carried a bare cost with no auditable per-meter breakdown.

Root cause: `run_passthrough` (`_passthrough.py`) only ever wrote `usage_log.cost` from `compute_cost`; nothing in the shared scaffold or in the four routes wrote the two audit fields.

Fix: added a `compute_meters` callback to `run_passthrough`, mirroring `compute_cost`'s signature and only invoked once a cost has been priced, and implemented it per route: input tokens for embeddings/rerank, image count for images, a flat per-request charge for moderations.

Not covered: audio (transcriptions/speech) is one of the four surfaces this issue names, but it never calls `compute_cost` at all today (no per-second/per-minute pricing model exists yet for it), so there is no cost for a meters breakdown to attach to. Audio billing needs its own pricing unit first, hence `Refs #432` rather than `Fixes #432`.

## Relevant issues

Refs #432

## Verification

- New regression tests (`test_embeddings_billing_meters_tracked_with_pricing` and the `images`/`moderations`/`rerank` equivalents) fail on `main` (`billing_meters` is `None`) and pass on this branch.
- Full suite passes: `make test` (2596 unit + integration tests), `make lint`, `make typecheck`.
- Did not check: audio billing, out of scope as noted above.

## PR Type

- [x] Bug Fix

## Checklist

## AI Usage

**AI Model/Tool used:** Claude (Anthropic), in an agentic coding session directed and verified by the account owner.

**Any additional AI details you'd like to share:** None.

## Summary

- Added billing details for passthrough embeddings, images, moderations, and rerank requests.
- Added usage tracking for input tokens, image count, and moderation requests.
- Avoided billing lines for unpriced moderation requests.
- Added integration tests for all four routes.

## Benefits

Usage logs now show how passthrough charges are calculated. This improves billing visibility and auditability.

### DIFF
--- a/src/gateway/api/routes/_passthrough.py
+++ b/src/gateway/api/routes/_passthrough.py
@@ -68,6 +68,11 @@ ResultT = TypeVar("ResultT")
 
 PASSTHROUGH_PROVIDER_ERROR_DETAIL = "The request could not be completed by the provider"
 
+# A route's non-token charge lines: the meters dict and the auditable breakdown,
+# matching the shape ``calculate_metered_cost`` and ``price_tool_calls`` already
+# write for the chat and tool-charge paths.
+BillingMeters = tuple[dict[str, Any], list[dict[str, Any]]]
+
 
 def resolve_passthrough_user_id(
     auth_result: tuple[APIKey | None, bool],
@@ -130,6 +135,7 @@ async def run_passthrough(
     enforce_require_pricing: bool = False,
     usage_tokens: Callable[[ResultT], tuple[int | None, int | None, int | None]] | None = None,
     compute_cost: Callable[[ResultT, ModelPricing | None], float | None] | None = None,
+    compute_meters: Callable[[ResultT, ModelPricing | None, float], BillingMeters | None] | None = None,
     map_provider_error: Callable[[Exception], HTTPException | None] | None = None,
     reserve_before_resolve: bool = False,
     relabel: bool = True,
@@ -167,6 +173,11 @@ async def run_passthrough(
             total)`` token counts for the usage log. Defaults to ``(0, 0, 0)``.
         compute_cost: Maps the result and pricing to the final USD cost, or
             ``None`` to leave the log's cost unset and reconcile at 0.0.
+        compute_meters: Maps the result, pricing, and the cost ``compute_cost``
+            just returned to this request's billing meters and charge lines, or
+            ``None`` to leave both unset. Only called when ``compute_cost``
+            returned a cost, so a route with no priced unit never needs to guard
+            against a missing cost itself.
         map_provider_error: Route-specific provider-exception mapping checked
             before the generic 502 (the error log and refund happen either way).
         reserve_before_resolve: Preserve the audio routes' historical ordering,
@@ -389,6 +400,9 @@ async def run_passthrough(
         cost = compute_cost(result, pricing) if compute_cost else None
         if cost is not None:
             usage_log.cost = cost
+            billing = compute_meters(result, pricing, cost) if compute_meters else None
+            if billing is not None:
+                usage_log.billing_meters, usage_log.pricing_breakdown = billing
 
         await log_writer.put(usage_log)
         await reconcile_reservation(db, reservation, cost if cost is not None else 0.0)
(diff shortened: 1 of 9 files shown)


njbrake (approved):

Approving. The mechanism is right, the rate math checks out on all four routes (`per_image_cost` stores raw USD per image, `flat_request_cost` divides by 1e6, both are reflected correctly in the lines), and the two line shapes match what the dashboard's breakdown renderer discriminates on. Full suite, lint, typecheck, and `openapi-check` all pass locally.

I pushed one commit to your branch rather than leaving it as review comments:

**Unpriced moderations recorded a zero charge line.** `compute_cost` there returns `flat_request_cost(pricing)`, which is `0.0` and not `None` for an unpriced model, so `compute_meters` ran on the free path and wrote `{"requests": 1}` plus a `$0.00` line. The dashboard renders that as a "Billed meters" block explaining a charge that never happened, which contradicts the route's own comment about unpriced moderations being intentionally silent, and diverges from `calculate_metered_cost`, which suppresses empty lines. Now returns None when nothing was charged, with a regression test.

**Renamed the embeddings and rerank input meter to `total_input_tokens`.** `input_tokens` had no readers: the billed-token SQL (`usage.py:_billed_expr`) and the dashboard's token bar (`tokenComposition`) both read `total_input_tokens` by name, so those rows were still falling through to the raw columns. Both routes log the same count as their prompt tokens, so no reported value changes; the meter just stops being decorative.

One thing I left alone deliberately. `images` and `requests` are genuinely new non-token units, which is the part of #432 this PR does not claim to solve, and they now land in the `/v1/usage` response before any reader exists for them, so a later rename is a wire change. I am fine shipping them as-is, but flagging that we are naming them now.

Minor, non-blocking: `n_images` is derived twice in `images.py`, once in each callback. Worth extracting so they cannot drift.

PR #545, opened 2026-08-10
feat(routing): weighted policy for load balancing across providers

pss-julien:
## Description

A `select` entry can now name the `weighted` router, which draws one candidate per request in proportion to declared weights, so traffic splits across providers behind a single policy name.

```yaml
select:
  - router: weighted
    candidates: [openai:gpt-5, anthropic:claude-sonnet-4-5]
    weights: {openai:gpt-5: 70, anthropic:claude-sonnet-4-5: 30}
  - default: openai:gpt-5
```

Design choices worth recording, since each is a claim a plausible implementation would break:

* **Weights are normalized, not percentages**, so 7:3 and 70:30 are one split. A candidate left out of `weights` gets zero, which drains a provider without deleting it: it takes no traffic and stays at the tail of the plan as a failover target. An all-zero split is refused, since it could never select anything.
* **Each request is an independent draw and nothing is remembered**, so the split is as correct behind many replicas as behind one, and none of the stickiness machinery in `docs/routing-scaling.md` is needed. The cost is that the ratio converges statistically rather than exactly.
* **The draw continues without replacement, so the whole ordering is the plan.** A retryable failure lands on another weighted provider before `on_failure` is reached, and a failing provider sheds its share with no health tracking.
* **No pricing is required**, unlike the kNN router which scores cost. The pricing gate is now scoped by `backend_requires_pricing`, so a weighted policy is not refused for missing rows it never reads.
* **A weighted decision is not logged per request** (that would be a log line at load-balancer volume). The usage row carries `router:weighted`, and the draw is available at debug level.
* **`explain` and the CLI render the split itself**, normalized over the candidates the caller may actually use, because a weighted decision needs no request state. The declared pool is still handed to the compiler, so a filtered candidate keeps appearing in `dropped` with its reason.
* **`Otari-Router: off` keeps its meaning** and serves the default target, which pins a caller to one provider during an incident.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [x] Documentation
- [ ] Infrastructure / CI

## Relevant issues

None filed; happy to link one if you would like this tracked.

## Checklist

Coverage added: `tests/unit/test_weighted_router.py` (schema rules for a split, the draw itself, the boundary case where a zero-weight candidate must never win the head, and the synchronous view `explain` renders), `tests/integration/test_routing_weighted.py` (route behavior end to end), and `web/src/pages/RoutingPage.test.tsx` for the split control. Both generated artifacts were confirmed in sync (`make openapi-check`, `make postman-check`; the spec change did not alter the Postman collection), and the dashboard bundle in `src/gateway/static/dashboard/` was rebuilt for the `web/src` change.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

The design decisions above were the substance of the work, so they are recorded in the commit message as well as here. The two places most worth a human eye are the boundary handling in the draw (a `random()` result at the cumulative total, or drifting past it, must not hand the head to a zero-weight candidate) and the narrowing of the pricing gate to `backend_requires_pricing`, which changes when a policy is refused for missing pricing rows.

## Summary

- Added weighted routing across providers with normalized traffic shares.
- Added validation, retries, failover, zero-weight draining, and opt-out behavior.
- Added weighted routing support to the API, CLI, dashboard, documentation, and usage records.
- Updated dashboard validation and labels for weighted and unknown router configurations.
- Added unit and integration tests for routing behavior, validation, filtering, explanations, and pricing independence.
- Updated dashboard bundles and related asset references.

### DIFF
--- a/docs/api-reference.md
+++ b/docs/api-reference.md
@@ -188,12 +188,27 @@ and counted as failures rather than only in the caller's own logs.
 |--------|------|-------------|------|
 | `POST` | `/v1/images/generations` | Generate images from text prompts. | API key or master key |
 
+Image generation bills per generated image, not per token, so a usage row carries
+zero tokens and an `images` meter. Unlike audio and moderations it is subject to
+`require_pricing`, so an unpriced image model is rejected with 402 under the
+default configuration. See
+[per-image pricing](configuration.md#per-image-pricing-image-generation) for how
+to set the rate.
+
 ### Audio
 
 | Method | Path | Description | Auth |
 |--------|------|-------------|------|
 | `POST` | `/v1/audio/transcriptions` | Transcribe audio to text (multipart upload). | API key or master key |
 | `POST` | `/v1/audio/speech` | Generate speech from text (TTS). | API key or master key |
+
+Audio bills per request rather than per token, under the same convention as
+[moderations](#moderations) and search, so a usage row carries zero tokens and a
+cost taken from the flat rate configured for the model. Like moderations, audio
+is exempt from `require_pricing`: with no rate configured the request is served
+and logged at $0 with no charge line. See
+[per-request pricing](configuration.md#per-request-pricing-audio-and-moderations)
+for how to set the rate.
 
 ### Files
 
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -397,6 +397,47 @@ pricing:
 
 At `min_input_tokens` and above, an entry replaces only the rates it lists for the entire request. Omitted rates inherit the base price. The dashboard's model detail editor exposes the same controls, and each usage row records the resulting billable meters and rate breakdown for auditability.
 
+#### Per-request pricing (audio and moderations)
+
+Audio (`/v1/audio/transcriptions`, `/v1/audio/speech`) and moderations report no token usage, so they bill a flat rate per request instead. `ModelPricing` only has per-million-token columns, so these routes reuse `input_price_per_million` with a different unit: **USD per million requests**. Divide by a million to read it as a per-request charge, so `6000.0` charges $0.006 per request.
+
+```yaml
+pricing:
+  openai:whisper-1:
+    input_price_per_million: 6000.0   # $0.006 per transcription
+  openai:tts-1:
+    input_price_per_million: 15000.0  # $0.015 per speech request
+  openai:omni-moderation-latest:
+    input_price_per_million: 0.0      # free (same effect as leaving it unset)
+```
+
+The key is the resolved `provider:model`, the same key every other route uses. [Search tools](#search-tools) follow the same convention under `<provider>:<tool>`.
+
+Two behaviors are worth knowing before you rely on this:
+
+- **Leaving the rate unset is free by design.** These endpoints are exempt from `require_pricing`, so an unpriced model is served, the usage row records a $0 cost, and no charge line is written. That is deliberate: a $0 charge line would render in Activity as a billed meter explaining a charge that never happened. Set a nonzero rate to get the charge line; a rate of `0.0` is treated the same as no rate at all.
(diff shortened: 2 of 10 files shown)


njbrake:
Hi @pss-julien ! Thank you for the contribution: would you be willing to share a bit about your rationale for the PR: what made you interested in this load balancer design?

njbrake (commented):

The implementation holds up well. I rebased it onto current main locally and lint, typecheck, both generated-artifact checks, and the unit, integration, and web suites all pass. I also reproduced the draw in isolation, including the zero-weight boundary with `random()` forced to 1.0, rather than taking the description on trust. Narrowing the pricing gate to `backend_requires_pricing` is right for a reason beyond the one you gave: it also drops the requirement for `noop` and for unknown backend names, neither of which reads a price.

Four things needed from you. I tried to push the rebase myself, but GitHub does not honor "allow edits by maintainers" for forks owned by an organization, so it has to come from your side.

1. **Rebase and rebuild the bundle.** Every conflict is in the generated dashboard bundle; no source conflicts. Rebase, then `npm --prefix web run build` and commit. The bundle also went stale on its own when main took the react-router bump in #543.

2. **`declared_shares` reports a split the runtime does not serve.** With an all-zero pool it returns an even split, but `weighted_ordering` takes its own zero-total branch and keeps declared order, so the head is fixed. With two or more drained survivors, `explain`, `router_weights`, the CLI, and `confidence` all print 50/50 for a plan that sends every request to one provider. Reachable when a caller's allow-list filters out every positively weighted candidate. Suggested fix:

```python
    if total <= 0:
        return {selector: 100.0 if index == 0 else 0.0 for index, selector in enumerate(pool)}
```

3. **"Retryable failure" is stale.** #546 landed after you forked and replaced that phrasing deliberately: fallback now happens on every pre-response failure, including 400/401/403. It is back at `docs/routing.md:168`, `cli.py:353` (user-visible output), and `weighted.py:22,86`.

4. **The allow-list CLI example is missing a line.** Under `--allowed-model 'anthropic:*'`, `gemini:gemini-2.0-flash` is dropped too, since it sits in `on_failure` and fails the same filter.

CodeRabbit's finding about the debug-logging claim is a false positive. `log_decision=False` sends the line to DEBUG rather than suppressing it, and I confirmed the record fires, so your docs are correct as written.

Still interested in the rationale question above. "Weighted, with no health awareness" is a reasonable place to stop, but it is a resting point I want to be deliberate about before it becomes public contract, since backing off a failing provider needs cross-replica state that this design deliberately avoids.

pss-julien:
> Hi @pss-julien ! Thank you for the contribution: would you be willing to share a bit about your rationale for the PR: what made you interested in this load balancer design?

Hi @njbrake, happy to explain!

Our main driver is request volume: we send enough traffic that a single provider isn't a comfortable place to put all of it. One of our providers also enforces a daily request cap, so today we'd have to track remaining quota ourselves and switch over manually before hitting it.

A load balancer solves both at once: traffic is spread across providers by default, so we stay under per-provider limits without any bookkeeping, and we get some resilience if one provider degrades.

njbrake:

Thanks for the rationale, that settles it for me: spreading load and staying under a per-provider request cap is a good reason to stop at weighted with no health awareness, at least for now.

I picked the branch up in #553 rather than wait, since your fork's `main` was reset out from under the PR and GitHub does not let a maintainer push to a fork owned by an organization. Your three commits are there unchanged, with your authorship. On top of them I rebased onto main, rebuilt the dashboard bundle, and added one commit of review fixes:

- `declared_shares` returned an even split for an all-zero pool while the draw keeps declared order, so `explain`, `router_weights` and the CLI printed a split that does not run.
- `/v1/routing/status` and `preferences/rank` treated a weighted policy as a learned one, which stopped a workspace from seeding examples for a learned policy it was preparing.
- "Retryable failure" is stale phrasing since #546, and the allow-list `explain` transcript was missing the `on_failure` drop.

The first reverses a decision you deliberately pinned in a test. If you still prefer the even split, say so on #553 and I will put it back.

PR #550, opened 2026-08-10
fix(billing): audio charge lines, no token defaults on non-token routes

### LINKED ISSUE #432: Passthrough usage (images, audio, embeddings, rerank) has no charge lines and no non-token unit

njbrake:
## Description

Audio transcription and speech recorded a bare cost: neither endpoint passed `compute_cost` to `run_passthrough`, so cost stayed unset and no `billing_meters` or `pricing_breakdown` were written. That left audio the one pass-through surface with no charge lines after #542.

Both endpoints now follow the per-request convention moderations uses: `input_price_per_million` holds the per-request rate scaled by 1e6 (`flat_request_cost`), and an unpriced model is free by design, so no $0 line is written. Audio bills per request because it has no measurable unit yet (tokens, seconds, characters); a duration meter lands with the rate-card work in #376. Pricing was never resolved on the `reserve_before_resolve` branch that only audio takes, so `run_passthrough` resolves it there and refunds the held reservation if the lookup fails.

Enabling that lookup exposed a unit bug on every route that reads `input_price_per_million` as something other than a token rate, because the genai-prices fallback quotes USD per million tokens. All three are fixed here with `pricing_use_defaults=False`, matching the search tools, the tool loop, and `pricing_init`:

- **Audio** (per request, rate scaled 1e6): `gpt-4o-transcribe` is in the dataset, so with `default_pricing` on a transcription recorded cost 2.5e-06 for a rate nobody configured. `gpt-4o-mini-tts` gave speech the same exposure.
- **Moderations** (per request): narrower, since the dataset carries no moderation models, but the provider-agnostic name match makes `openai:gpt-4o` reachable, logging cost 2.5e-06.
- **Images** (raw USD per image, unscaled): `gpt-image-1` is in the dataset at 5.0, so one generation billed $5.00 against a real price near $0.01 to $0.19. The worst of the three: this route reserves its estimate, so the $5.00 was held against the budget before the provider call, and `enforce_require_pricing=True` meant the bogus rate also satisfied the fail-closed gate that exists to catch this. No issue tracked it; it surfaced while reviewing this PR.

**Release note:** an image model priced only from the dataset is now rejected with 402 under `require_pricing` rather than billed at roughly a million times its rate.

Docs cover the per-request and per-image units (the per-image one was undocumented anywhere), the default-pricing exclusion, audio's soft budget cap under concurrency, and the fact that a $0 audio row reads as priced rather than unpriced.

## PR Type

- [x] Bug Fix

## Relevant issues

Fixes #432

## Checklist

- [ ] If the API contract changed, I regenerated the OpenAPI spec (`uv run python scripts/generate_openapi.py`). Not applicable: no route signature or docstring changed, and `make openapi-check` / `make postman-check` both pass.

## AI Usage

**AI Model/Tool used:** Claude Opus 5 (1M context), via Claude Code, in agentic sessions directed and verified by the account owner.

**Any additional AI details you'd like to share:** The images fix came out of an independent review pass over the audio work, which reproduced each rate against the real genai-prices dataset rather than trusting the code comments.

## Summary

Audio transcription and speech requests now create auditable billing records. Priced models receive one charge per request, while unpriced models remain free without zero-value charge lines.

The passthrough flow resolves pricing before calculating costs, including reserved requests. Moderation uses only explicitly configured rates. Documentation and integration tests cover these billing rules and prevent token-based defaults from charging non-token endpoints.

### DIFF
--- a/src/gateway/api/routes/_passthrough.py
+++ b/src/gateway/api/routes/_passthrough.py
@@ -165,11 +165,12 @@ async def run_passthrough(
             Audio resolves it for per-request charge lines but the reservation
             estimate stays 0 (no measurable cost unit yet, so no pre-call spend).
         pricing_use_defaults: Whether the pricing lookup may fall back to the
-            genai-prices dataset. A route that bills per request passes False for
-            the reason :func:`find_model_pricing` documents: those rates are USD
-            per million *tokens*, so under a per-request convention they would be
-            charged as USD per million *requests* and write a charge line at the
-            wrong unit for a rate nobody configured.
+            genai-prices dataset. A route whose billable unit is not a token
+            passes False for the reason :func:`find_model_pricing` documents:
+            those rates are USD per million *tokens*, so a per-request route
+            would charge them as USD per million *requests* and a per-image route
+            as USD per image, writing a charge line at the wrong unit for a rate
+            nobody configured.
         estimate: Maps the pricing row to the reservation estimate in USD.
             Defaults to 0.0, which still enforces per-user state (user exists,
             not blocked, not already over budget).
--- a/src/gateway/api/routes/_passthrough.py
+++ b/src/gateway/api/routes/_passthrough.py
@@ -161,7 +161,8 @@ async def run_passthrough(
             ``HTTPException`` raised here (e.g. an upload size check) refunds
             the reservation and propagates unchanged.
         lookup_pricing: Whether to resolve :class:`ModelPricing` for the model.
-            Audio has no measurable cost unit yet and skips the lookup.
+            Audio resolves it for per-request charge lines but the reservation
+            estimate stays 0 (no measurable cost unit yet, so no pre-call spend).
         estimate: Maps the pricing row to the reservation estimate in USD.
             Defaults to 0.0, which still enforces per-user state (user exists,
             not blocked, not already over budget).
@@ -292,6 +293,8 @@ async def run_passthrough(
         # unresolvable selector to 400; otherwise the estimate leaks.
         try:
             resolved = resolve_provider_selector(config, model, user_id)
+            if lookup_pricing:
+                pricing = await find_model_pricing(db, resolved.instance, resolved.model)
         except (ValueError, AnyLLMError) as exc:
             await refund_reservation(db, reservation)
             await _log_rejection(
--- a/src/gateway/api/routes/_passthrough.py
+++ b/src/gateway/api/routes/_passthrough.py
@@ -131,6 +131,7 @@ async def run_passthrough(
     user: str | None,
     call_provider: Callable[[ResolvedProvider], Awaitable[ResultT]],
     lookup_pricing: bool = True,
+    pricing_use_defaults: bool = True,
     estimate: Callable[[ModelPricing | None], float] | None = None,
     enforce_require_pricing: bool = False,
     usage_tokens: Callable[[ResultT], tuple[int | None, int | None, int | None]] | None = None,
@@ -163,6 +164,12 @@ async def run_passthrough(
         lookup_pricing: Whether to resolve :class:`ModelPricing` for the model.
             Audio resolves it for per-request charge lines but the reservation
             estimate stays 0 (no measurable cost unit yet, so no pre-call spend).
+        pricing_use_defaults: Whether the pricing lookup may fall back to the
+            genai-prices dataset. A route that bills per request passes False for
+            the reason :func:`find_model_pricing` documents: those rates are USD
+            per million *tokens*, so under a per-request convention they would be
+            charged as USD per million *requests* and write a charge line at the
+            wrong unit for a rate nobody configured.
         estimate: Maps the pricing row to the reservation estimate in USD.
             Defaults to 0.0, which still enforces per-user state (user exists,
             not blocked, not already over budget).
@@ -293,8 +300,6 @@ async def run_passthrough(
         # unresolvable selector to 400; otherwise the estimate leaks.
         try:
             resolved = resolve_provider_selector(config, model, user_id)
-            if lookup_pricing:
-                pricing = await find_model_pricing(db, resolved.instance, resolved.model)
         except (ValueError, AnyLLMError) as exc:
             await refund_reservation(db, reservation)
             await _log_rejection(
@@ -304,6 +309,16 @@ async def run_passthrough(
                 status_code=status.HTTP_400_BAD_REQUEST,
             )
             _raise_for_unresolvable_model(model, exc)
+        if lookup_pricing:
+            # Unlike the branch below, the reservation is already held here, so a
+            # failed lookup must refund before propagating or the estimate leaks.
+            try:
+                pricing = await find_model_pricing(
+                    db, resolved.instance, resolved.model, use_defaults=pricing_use_defaults
+                )
+            except Exception:
+                await refund_reservation(db, reservation)
+                raise
     else:
         try:
             resolved = resolve_provider_selector(config, model, user_id)
@@ -317,7 +332,9 @@ async def run_passthrough(
             )
             _raise_for_unresolvable_model(model, exc)
         if lookup_pricing:
-            pricing = await find_model_pricing(db, resolved.instance, resolved.model)
+            pricing = await find_model_pricing(
+                db, resolved.instance, resolved.model, use_defaults=pricing_use_defaults
+            )
         # Reserve first so user/blocked/budget rejections (404/403) precede the
         # missing-pricing rejection (402); refund if we then reject for no pricing.
         reservation = await _reserve(
--- a/src/gateway/api/routes/_passthrough.py
+++ b/src/gateway/api/routes/_passthrough.py
@@ -177,7 +177,10 @@ async def run_passthrough(
         enforce_require_pricing: When True and ``config.require_pricing`` is
             set, reject unpriced models with 402. The check runs after the
             reservation (so its 404/403 rejections take precedence) and the
-            reservation is refunded before raising.
+            reservation is refunded before raising. Honored only on the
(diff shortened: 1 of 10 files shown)


khaledosman (commented):
Five findings inline, most severe first. The main one is that `images.py` is the fourth consumer of the overloaded `input_price_per_million` column and is left without the new `pricing_use_defaults=False`; since it can't be commented on directly (the file isn't in the diff), it's attached to the new parameter in `_passthrough.py`.

Gates checked and clean: no route signature or docstring changes, so the OpenAPI spec and Postman collection stay valid; no `web/src` or `docs/dashboard.md` edits, so the dashboard bundle is unaffected; no layering violations; no em dashes or `--` separators in the added prose; the PR title is a conventional commit.

---

🤖 Reviewed with [Claude Code](https://claude.com/claude-code) (Opus 5), directed and verified by the account owner.

src/gateway/api/routes/_passthrough.py
```diff
@@ -131,6 +131,7 @@ async def run_passthrough(
     user: str | None,
     call_provider: Callable[[ResolvedProvider], Awaitable[ResultT]],
     lookup_pricing: bool = True,
+    pricing_use_defaults: bool = True,
```
khaledosman, line 134: **High: `images.py` needs this too, and it's the route where the unit mismatch overcharges.**

This PR sweeps three of the four consumers of the overloaded `input_price_per_million` column (audio, moderations, search). `images.py:88` still passes `pricing_use_defaults` at its default `True`.

`per_image_cost` (`pricing_service.py:417`) treats the column as **raw USD per image, no division**, while genai-prices supplies USD per million tokens. Against the installed dataset:

```
openai:gpt-image-1 -> input_price_per_million = 5.0
```

So with `default_pricing: true` and no DB row for `openai:gpt-image-1`, `per_image_cost(1, pricing)` returns `1 * 5.0` = **$5.00 per image** (real rate ~$0.01-0.19). `estimate` uses the same helper, so $5.00 is also reserved up front: budgets drain 25-500x too fast and callers hit 403 far too early. `enforce_require_pricing=True` on that route makes the fallback path load-bearing rather than incidental.

Fix is the same one line: add `pricing_use_defaults=False` to the `run_passthrough` call in `images.py`, and add images to the "Per-request routes are excluded" bullet in `docs/configuration.md` (the unit there is per-image, not per-request, but the defect is identical).

src/gateway/api/routes/_passthrough.py
```diff
@@ -301,6 +309,16 @@ async def _reserve(estimated_cost: float, *, row_model: str, row_provider: str |
...
             _raise_for_unresolvable_model(model, exc)
+        if lookup_pricing:
+            # Unlike the branch below, the reservation is already held here, so a
+            # failed lookup must refund before propagating or the estimate leaks.
+            try:
+                pricing = await find_model_pricing(
+                    db, resolved.instance, resolved.model, use_defaults=pricing_use_defaults
+                )
+            except Exception:
+                await refund_reservation(db, reservation)
```
khaledosman, line 330: **Medium: this guard can't fire for its only realistic trigger.**

`find_model_pricing` is a plain `SELECT`; it fails on DB errors essentially only. At that point the session is in a failed-transaction state, so `refund_reservation`'s `UPDATE` + `commit` raises `PendingRollbackError`, masking the original exception *and* leaking the hold, which is exactly what the comment above says it prevents.

```python
except Exception:
    await db.rollback()
    await refund_reservation(db, reservation)
    raise
```

src/gateway/api/routes/_passthrough.py
```diff
@@ -301,6 +309,16 @@ async def _reserve(estimated_cost: float, *, row_model: str, row_provider: str |
                 status_code=status.HTTP_400_BAD_REQUEST,
             )
             _raise_for_unresolvable_model(model, exc)
+        if lookup_pricing:
+            # Unlike the branch below, the reservation is already held here, so a
```
khaledosman, line 317: **Low: `enforce_require_pricing` is silently a no-op on this branch.**

Harmless today (audio doesn't set it), but before this change the branch had no pricing at all, so the gap was self-evident. Now that pricing resolves here, the `enforce_require_pricing` docstring a few lines up reads as if the flag applies to both branches, and a future per-request route setting both would skip the 402 with no signal. Either apply the gate here as well, or note the restriction in the docstring.

src/gateway/api/routes/audio.py
```diff
@@ -83,10 +84,29 @@ async def call_provider(resolved: ResolvedProvider) -> Transcription:
...
+        # an unpriced model is free by design.
+        return flat_request_cost(pricing)
+
+    def compute_meters(result: Transcription, pricing: ModelPricing | None, cost: float) -> BillingMeters | None:
+        # One request is one billed meter. Skip the $0 charge line for an unpriced
+        # model, same as moderations: a reader sees no charge line because there
+        # is no charge, not because of a bug.
+        if not cost:
+            return None
+        breakdown = [{"meter": "request", "units": 1, "unit_rate": cost, "cost": cost}]
```
khaledosman: **Low: third copy of this breakdown literal.**

`{"meter": "request", "units": 1, "unit_rate": cost, "cost": cost}` now appears here, at `audio.py:207`, and at `moderations.py:81`, each behind a near-identical `if not cost: return None`. The comment block in `pricing_service.py` argues these conventions get named helpers precisely so they can't be "miscopied into a new route and misbill by a factor of a million". A `per_request_meters(cost) -> BillingMeters | None` beside `flat_request_cost` collapses all three.

tests/integration/test_audio_endpoint.py
```diff
@@ -362,3 +364,177 @@ def test_speech_optional_fields(
...
+    api_key_header: dict[str, str],
+) -> None:
+    """Default (genai-prices) rates never price audio, even with default_pricing on.
+
+    Those rates are USD per million tokens, but audio bills per request, so
+    honoring one would write a charge line at the wrong unit for a rate the
+    operator never configured. gpt-4o-transcribe is in the dataset, so it is the
+    case that would regress.
+    """
+    configure_default_pricing(True)
```
khaledosman, line 519: **Low: this guard can go vacuous.**

The assertion is `cost == 0.0` after enabling default pricing. If genai-prices ever drops `openai:gpt-4o-transcribe`, the test passes for the wrong reason and stops guarding anything. Pin the premise so it fails loudly instead:

```python
assert default_model_pricing("openai", "gpt-4o-transcribe", datetime.now(UTC)) is not None
```

Also, the `try/finally` restoring `configure_default_pricing(False)` duplicates the autouse `_reset_default_pricing` fixture in `tests/conftest.py:47`. Harmless, just redundant.

tests/integration/test_moderations_endpoint.py
```diff
@@ -440,3 +441,44 @@ def test_moderations_no_warning_when_pricing_missing(
...
+    honoring one would charge a token rate as a request rate. The dataset has no
+    moderation models, so the case that regresses is an operator pointing this
+    endpoint at a model name the dataset does carry.
+    """
+    resp_obj = ModerationResponse(
+        id="modr-test",
+        model="gpt-4o",
+        results=[ModerationResult(flagged=False, categories={}, category_scores={})],
+    )
+    configure_default_pricing(True)
```
khaledosman, line 468: Same vacuous-guard risk as the audio test: assert that `openai:gpt-4o` actually resolves under defaults before asserting the route ignores it, otherwise a dataset change turns this green for the wrong reason.

PR #553, opened 2026-08-11
feat(routing): weighted policy for load balancing across providers

njbrake:
## Description

Continues #545 by @pss-julien, whose three commits are preserved here with their authorship. That PR's head branch was the fork's `main`, which has since been reset, and GitHub does not honor "allow edits by maintainers" for a fork owned by an organization, so the branch moved here to carry the review fixes.

What it adds: a `select` entry can name the `weighted` router, which draws one candidate per request in proportion to declared weights, so traffic splits across providers behind one policy name. The design rationale is recorded in #545 and in `docs/routing.md`.

On top of @pss-julien's work, one review commit:

- `declared_shares` reported an even split for an all-zero pool while the draw kept declared order, so `explain`, `router_weights`, the CLI, and `confidence` all printed a split the runtime does not serve.
- `/v1/routing/status` and `preferences/rank` treated a weighted policy as a learned one. The second is the sharper edge: a workspace whose only router policy was weighted could no longer seed examples for a learned policy it was preparing.
- "Retryable failure" is stale phrasing since #546. Restored in the new prose, the CLI output, and the test comments.
- The allow-list `explain` transcript omitted the `on_failure` drop, which the same filter removes.
- Rebased onto main, dashboard bundle rebuilt, OpenAPI regenerated.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [x] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Continues #545. No issue filed.

## Checklist

- [x] I ran the Definition of Done checks locally (`make lint`, `make typecheck`, `make test`). Unit and web suites plus `openapi-check` and `postman-check` ran locally; the integration suite needs PostgreSQL and ran on CI.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

Claude Opus 5 via Claude Code, for the review fixes and this description. The three underlying commits were contributed by @pss-julien, who reported them as fully AI generated with the same tool in #545.

Two places worth a human eye. The `declared_shares` change reverses a behavior @pss-julien deliberately pinned in a test, so it is a decision to confirm rather than a bug fix everyone agrees on. The routing memory change narrows what `POST /v1/routing/preferences/rank` accepts as a teachable score key, which is a behavior change for any workspace running a weighted policy.

## Summary

- Added weighted routing across providers with configurable traffic shares.
- Added zero-weight candidates as failover targets.
- Updated routing explanations, CLI output, status endpoints, confidence reporting, and OpenAPI types.
- Added dashboard support for creating, editing, validating, and explaining weighted policies.
- Updated routing documentation and fallback terminology.
- Kept weighted routing independent of pricing and routing-memory data.
- Added comprehensive unit and integration test coverage.
- Updated the dashboard bundle.


src/gateway/services/routing/backends.py
```diff
@@ -133,6 +153,43 @@ def clear_router_backend_cache() -> None:
...
+    """Whether this name selects the weighted load balancer.
+
+    A named check because the weighted backend is the one whose decision needs no
+    request state, so several synchronous surfaces (``explain``, the CLI) special
+    case it, and each of them would otherwise repeat the same normalization.
+    """
+    return name is not None and name.strip().lower() == WEIGHTED_BACKEND
+
+
+def backend_learns(name: str | None) -> bool:
```
khaledosman: Excluding `weighted` here is right, but this also drops `noop`, and this module documents `noop` (lines 16-17) as the backend that "hold[s] a policy's shape while its pool is still being taught" — its pool is precisely the one an operator is teaching. Both new call sites exclude it: `routing_memory.py:259` (`POST /v1/routing/preferences/rank`) and `routing_memory.py:315` (`GET /v1/routing/status`).

Verified against both revisions, with a config holding a `noop` placeholder plus one `knn` policy:

| | `main` | `49d5306` |
|---|---|---|
| `_learned_policies` | `['smart', 'warming']` | `['smart']` |
| scores keyed on the placeholder's pool | accepted | **400** "These score keys do not name a model any learned policy can route to" |

So a workspace preparing a policy with the placeholder backend can no longer seed its examples, nor read its warmth before flipping it to `knn`. That is the same failure the PR description gives as the reason for this change, now landing on the backend that exists for that workflow.

Gate on `not backend_is_weighted(spec.router_backend)` at both call sites, or include `NOOP_BACKEND` here.

web/src/pages/RoutingPage.tsx
```diff
@@ -519,19 +618,43 @@ function PolicyForm({
...
                   </div>
+                  {weighted ? (
+                    <div className="flex items-end gap-2">
+                      <Field
+                        label="Share"
+                        value={String(weights[index] ?? 0)}
+                        onChange={(value) =>
+                          setWeights((prev) =>
+                            prev.map((weight, i) =>
+                              i === index ? Math.max(0, Number(value) || 0) : weight,
```
khaledosman: Minor: `Number(value) || 0` on an input whose value re-renders from `String(weights[index])` swallows the decimal point (`"7"` → type `.` → `Number("7.")` is `7` → renders `"7"`), so a fractional share cannot be typed, and clearing the field snaps it to `0`. The API and `docs/routing.md` accept "any non-negative numbers".

Relative weights make integers sufficient in practice and line 564 already uses this pattern for thresholds, so this is cosmetic; holding the raw string in state and parsing on submit would close the gap.

khaledosman (approved):
Gates pass on `49d5306`: `make lint` (architecture + Ruff), `make typecheck`, `openapi-check`, `postman-check`, `tests/unit/test_weighted_router.py`, and `web` `RoutingPage.test.tsx`. The committed dashboard bundle carries the new `RoutingPage`/`DocsPage` output and is referenced from `index-D0MBX8Ui.js`.

Two inline: one confirmed regression for the `noop` placeholder backend, one cosmetic note on the share input.

---

njbrake:

All eight threads addressed in 6879b14a: six fixes, two disagreements. Resolving them all.

**@khaledosman on `noop`.** Fixed, and thank you for the before/after table, it made the regression unarguable. `backend_learns` is now `backend_pool_is_teachable`, false only for the weighted backend, so `noop` and a name from a newer build keep exactly the behavior they had on main. Test added covering both the status listing and the `rank` typo guard.

**@khaledosman on the share field, and Copilot on non-finite input.** Fixed together. The field now holds the typed text and parses once, so `7.5` survives the keystroke after the decimal point, a cleared field stays cleared instead of snapping to 0, and `Infinity` or a negative is refused in the form where the API refuses it rather than sailing through to a 400.

**Copilot on backend-name normalization.** Fixed. `normalizedBackend` trims and lowercases before comparing, so a policy spelled `" Weighted "`, which the gateway resolves happily, is editable and tagged Weighted instead of read-only and "Routed". Test added.

**CodeRabbit on the zero-share wording.** Taken, in both guides: a zero-weight candidate takes no *weighted* traffic, and still serves opt-out callers when it is also the default target.

**CodeRabbit on `on_failure` in `config.example.yml`.** Taken, reworded to say failover stops once part of a response has reached the caller.

**CodeRabbit on the dashboard chunks (Critical and Major).** Not reproducible; resolving as false positives. `RoutingPage-Dqh6Cebs.js` imports `H as $e`, and index export `H` is the tool-settings query (`queryFn:()=>i("/v1/tool-settings")`), which is what `Ve()` wants. Export `$` is indeed the pricing-delete mutation, but nothing binds `$e` to it. `O` and `y` are both present in their own chunks' import lists. The bundle is the output of a single vite build, and `npm --prefix web run build` reproduces the committed tree byte for byte, so its cross-chunk bindings cannot be inconsistent by construction.

pss-julien:
@njbrake apologies for closing the other PR, I ran into rebase/merge conflicts because other commits were touching the static asset files. I decided to start over and first remove the static assets from the Git repository in favor of compiling them at build time.

I'd suggest merging #556 before this one. I've already prepared a branch that will merge cleanly once #556 lands: https://github.com/pocketsizesun/otari/tree/feat/weighted-load-balancing

njbrake:
@pss-julien Great, thank you! Yes no worries about this. I will take the PRs from here and get them over the finish line. Thanks for the great ideas and solid contributions. All the PRs have your cmmits in them so you'll get contribution credit for them 🙏

njbrake:

Two more threads from the latest CodeRabbit pass, both now closed out in 31a161cd.

**Edge-case save paths (`RoutingPage.test.tsx`).** Fair, taken. Both cases stopped at what the form displayed. The fractional test now saves and asserts the POST body carries `7.5`, and the loosely spelled backend test now saves and asserts the emitted entry, which additionally pins that the spelling is normalized on the way out. I checked both against the pre-fix behavior: with `normalizedBackend` returning the raw name and the save path truncating, exactly these two fail, so they are regression tests rather than restatements.

**Session request serialization.** Out of scope here, so resolving without a change. The logout race is real (`AuthContext.logout` fires `void deleteSession()` and does not await it, so a fast re-login can post before the delete lands), but `web/src/api/client.ts` is byte-identical to main and untouched by this PR, which adds a routing backend. The anchor is also a dashboard bundle chunk, and #556 stopped tracking the built bundle, so that file no longer exists in the repo. Worth its own issue against the auth path.

PR #555, opened 2026-08-11
feat(dashboard): share the usage view as an image

### LINKED ISSUE #525: Dashboard: share button to give screenshot that can be posted on socials
People like to share their token usage. I would like to support a world where we can encourage a diversity of model usage and even help people share their strategies for weaning themselves off proprietary models and over to open models. Or even just to encourage choice. 

I'm thinking maybe in the usage and analytics page we can over a "share button" which compiles an image they can share on X, LinkedIn etc that shares stats about token and requests or however they configure it. 

The idea is that they can hit share and then select the filters to apply to curate the data and then we produce an image designed for sharing a high level overview of their usage of models, providers, etc.

njbrake:
## Description

Adds a share icon to the bottom-right of the Usage page histogram. It opens a dialog with a live preview of a fixed-size card and hands back a PNG to copy or download.

**Data comes from the page.** The card reads the page's existing summary instead of issuing its own query, so the window, the drag-to-zoom range and the entity filters all carry over, and the card cannot disagree with the numbers above it. It names its own scope, so a filtered figure is not read as the whole gateway. The dialog controls presentation only (lead stat, title, shape, theme, row count, whether dollars appear). Those choices persist; the data scope never does, since restoring a stale window would silently change what the card claims.

**Rasterization adds no dependency.** Serialize the card with `XMLSerializer`, wrap it in an SVG `foreignObject`, load it from a `data:` URI, draw to a canvas. A `blob:` URL taints the canvas in Chromium so `toBlob()` refuses outright, and the card uses literal colors rather than `var(--otari-*)` because custom properties do not resolve inside an `img`-loaded SVG.

**Honesty details.** Spend is asterisked when the window holds unpriced requests, with the count printed on the card itself since a PNG travels without the docs. An absent `unpriced_requests` counts as unknown rather than zero. A stat with no value is omitted, not published as a dash. No user or key names reach the card. The footer URL is hardcoded, so a self-hosted gateway never publishes its own hostname.

**Frame is fixed, so rows divide a height budget.** Hard-coded row heights failed both ways: three rows left a third of a square card blank, and nine overflowed so flex-shrink collapsed the title to zero height.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Closes #525

## Checklist

Notes on the boxes above, since a tick is easy and the detail is not:

- Tests are dashboard-side (`web/`), not `tests/unit` or `tests/integration`: no Python changed in this PR. Unit tests cover the derivations and the height budget; the e2e seeds usage and asserts the downloaded file's PNG signature and IHDR dimensions.
- Ran locally: `make lint`, `make typecheck` (mypy, 324 files), `uv run pytest tests/unit` (1561 passed), `npm --prefix web run typecheck`, `npm --prefix web test` (563 passed), and the Playwright suite (9 passed). `make test` also runs `tests/integration`, which needs PostgreSQL that this environment has no Docker for, so that half did not run here.
- OpenAPI unticked because the API contract did not change. An earlier revision added a `provider_model` summary dimension; that was reverted when the open-weights split was cut, so `src/gateway/` is untouched. `make openapi-check` and `make postman-check` both pass.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

Design and scope decisions were @njbrake's, iterated in conversation; the implementation and this description are the model's.

Worth knowing for review: two bugs in this feature were invisible to a fully green unit suite and only surfaced when driving a real browser. Drawing an SVG from a `blob:` URL taints the canvas, so `toBlob()` refused and the card never rendered at all; and a long model list overflowed the fixed frame so flex-shrink collapsed the title. jsdom has no canvas, no `toBlob` and no object URLs, which is why the e2e assertions on the actual PNG bytes are in this PR rather than left as a follow-up.


web/src/components/ShareDialog.tsx
```diff
@@ -0,0 +1,351 @@
...
+          if (!cancelled) {
+            setError(cause instanceof Error ? cause.message : "The share card could not be rendered.");
+          }
+        });
+    }, 300);
+    return () => {
+      cancelled = true;
+      clearTimeout(timer);
+    };
+  }, [presentation, hero, secondary, shown, scope]);
```
khaledosman, line 161: The preview effect never settles: it re-rasterizes the card every 300ms for as long as the dialog is open.

`secondary` (L95) and `shown` (L96) are fresh arrays on every render, so this dep array always differs. The effect's own `setPreview` (a new `blob:` URL each time) re-renders, which re-arms the 300ms timer, which rasterizes again. Confirmed against this head with a throwaway test: 10 `rasterize` calls in ~4s of idle with no user input. Each cycle is a 2160x2160 canvas draw plus a PNG encode, and the `` swaps every time.

Fix: `useMemo` `secondary` and `shown`, or key the effect on `presentation` plus primitives derived from the data.

web/src/components/ShareDialog.tsx
```diff
@@ -0,0 +1,351 @@
...
+                isDisabled={busy || blocked}
+                onPress={() => withBlob((blob) => downloadBlob(blob, shareFilename(startIso, endIso)), "Image saved")}
+              >
+                Download PNG
+              </Button>
+              {copyable ? (
+                <Button
+                  variant="primary"
+                  isDisabled={busy || blocked}
+                  onPress={() => withBlob(async (blob) => { await copyBlobAsImage(blob); }, "Image copied")}
```
khaledosman: "Image copied" is shown even when the copy failed. `copyBlobAsImage` swallows its own error and returns `false`, and `withBlob` ignores the return value before setting the notice.

```ts
onPress={() => withBlob(async (blob) => {
  if (!(await copyBlobAsImage(blob))) {
    throw new Error("The image could not be copied to the clipboard.");
  }
}, "Image copied")}
```

web/src/components/ShareDialog.tsx
```diff
@@ -0,0 +1,351 @@
...
+    if (raw === null) {
+      return DEFAULTS;
+    }
+    const parsed: unknown = JSON.parse(raw);
+    if (typeof parsed !== "object" || parsed === null) {
+      return DEFAULTS;
+    }
+    // Merged over the defaults rather than trusted: a stored shape from an older
+    // build (or a hand-edited one) must not be able to crash the panel.
+    return { ...DEFAULTS, ...(parsed as Partial<Presentation>) };
```
khaledosman: This merge does not deliver what the comment above it claims: the spread copies unvalidated values straight through. A stored `ratio` outside `CARD_SIZES` (older build, hand-edited storage) crashes the panel, reproduced here with `{"ratio":"portrait"}` in localStorage:

```
TypeError: Cannot destructure property 'width' of 'CARD_SIZES[ratio]' as it is undefined.
```

The existing "survives a corrupt stored presentation" test only covers unparseable JSON, so this path is uncovered. Validate each field against its known set (`ratio`, `theme`, `hero`, `rows` in {1,3,5,9}) and fall back per field.

web/src/components/ShareCard.tsx
```diff
@@ -0,0 +1,245 @@
...
+          gap: isLandscape ? 48 : 32,
+          alignItems: isLandscape ? "center" : "flex-start",
+        }}
+      >
+        {hero !== null ? (
+          <div style={{ flex: isLandscape ? "1 1 0" : "0 0 auto" }}>
+            <div style={{ fontSize: heroSize, fontWeight: 700, lineHeight: 0.82 }}>{hero.value}</div>
+            <div style={{ fontSize: 44, fontWeight: 500, lineHeight: 1.1, color: palette.muted, marginTop: 8 }}>
+              {hero.label}
+              {hero.caveated ? "*" : ""}
```
khaledosman, line 193: The caveat asterisk has no legend on the image. `Spend*` (and the same mark at L215) publishes a symbol a viewer of a standalone PNG cannot resolve; the explanation lives only in `docs/dashboard.md`, which does not travel with the file.

Pass the unpriced count in and print `* N requests unpriced` alongside `scope` (L235) whenever a shown stat is caveated.

khaledosman (approved):
Verified at `95fbf95` in a worktree at PR head: `npm run typecheck` clean, 558 unit tests pass, and the committed bundle carries both the new chunk and the `docs/dashboard.md` text. Four inline findings, none blocking.

One question outside the diff lines: a root `TODOS.md` duplicates issue tracking and pins line numbers (`usage.py:1320`) that will drift.

PR #560, opened 2026-08-12
feat(routing): rename policies in place, and stop them hiding candidate prices

pss-julien:
## Description

Two changes to the routing surface that were in flight together.

Rename: POST /v1/routing/policies takes `rename_from`, so an edit that both
renames a policy and re-targets it cannot land half applied, leaving the old
name serving the new spec. The new name is validated exactly as a fresh one is,
because a rename can walk a policy into every collision a create can. The
dashboard's Edit form exposes it, and the row moves rather than being copied.

Pricing: a policy no longer withholds its candidates from GET /v1/models. Every
selector of every policy in force was hidden, `on_failure` chains included, so
one failover policy could empty most of a catalogue, and a candidate priced by
the genai-prices fallback then disappeared from the dashboard together with its
rate. GET /v1/models/{key} never withheld them, so nothing was really being kept
off the wire. Alias targets are still withheld: an alias exists to stand in for
a target, a policy does not.

POST /v1/pricing now refuses a policy name the way it already refuses an alias
name, naming the candidates to price instead. Pricing, budgets, and usage key on
the model a request resolves to, so a row stored under the policy name was
written and never read.

The dashboard rebuilds rows the catalogue withheld from the discovery endpoint.
Those now report "rate unknown" instead of "not priced" when default pricing is
on, since the gateway does meter such a model; with the fallback off, unpriced
remains the truth.

## PR Type

- [x] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

None.

## Checklist

`make lint`, `make typecheck` and the full `pytest tests/unit tests/integration` run are green (2718 passed, 10 skipped). The API contract changed, so `docs/public/openapi.json` and the Postman collection were both regenerated; `make openapi-check` and `make postman-check` pass. Dashboard side: `tsc -b --noEmit` clean and the full Vitest suite green (585 passed), plus an e2e spec for the rename flow.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

The change was written and validated with Claude Code. It was previously part of a
longer branch of mine; that branch has been rebased onto current `main` and split so
each PR carries one self-contained change, reviewable on its own.

## Summary

- Added in-place renaming for routing policies through the API and dashboard.
- Preserved policy scope and usage history during renames.
- Kept aliases read-only and protected against rename conflicts.
- Made routing-policy candidates visible in model discovery, including failure chains.
- Prevented pricing policies directly and identified their candidate models instead.
- Improved dashboard pricing labels for unknown and unpriced models.
- Updated documentation, API examples, integration tests, and dashboard coverage.


njbrake (approved):
_Note: this review was drafted by Claude Opus 5 via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is the model's._

Approving. Heads up that I pushed a second commit to your branch (`7204cdd9`) instead of sending it back for round trips on small things. Take a look and say so if you disagree with any of it.

Two of them were real, both on the rename path:

- `rename_from` equal to `name` with no stored row created the policy instead of 404ing, because the 404 sat behind `renaming`. An edit form whose row was deleted underneath it would resurrect the policy rather than report it gone. The 404 now keys on the field being sent at all, and the unchanged-name upsert your test pins still works.
- A rename that lost the race between `_name_is_taken` and the commit surfaced as `Database error`. It now catches `IntegrityError` and re-reads the name to answer with the same 409. Re-reading rather than mapping the exception straight to 409, because the `user_id` foreign key raises the same class and calling a deleted user a name clash sends the operator after the wrong thing.

Both carry regression tests that fail without the fix. The rest is minor: a coverage test for the one pricing branch that cannot name candidates, the unpriced label waiting for the settings fetch instead of claiming "not priced" while the flag is still undefined, a comment, and freeing the e2e rename target so a re-run against a warm database does not trip on its own leftovers.

On the half that stops policies withholding candidates: I checked that `GET /v1/models/{key}` only ever consulted `effective_aliases`, so those candidates were served with their prices all along and the listing was withholding from the dashboard without protecting anything. The key allowlist still filters every listed row. Good change, and the docs carry the reasoning rather than just the new behavior.

PR #561, opened 2026-08-12
fix(chat): forward service_tier to the provider instead of dropping it

pss-julien:
## Description

`service_tier` is part of the OpenAI chat wire contract but absent from
any-llm's `CompletionParams`, which `ChatCompletionRequest` derives its
field set from. The derived base uses pydantic's default `extra="ignore"`,
so a caller's tier validated fine and then vanished before the provider
call; meanwhile the provider's own `service_tier` still came back on the
response, so the request looked honored when it was not.

Declare the field by hand so it survives `model_dump(exclude_unset=True)`
and rides any-llm's `**kwargs` passthrough into the provider request
(`acompletion` forwards unknown kwargs to the provider API call, and both
the OpenAI and Anthropic SDKs accept `service_tier`). `/v1/responses` never
had the gap: `ResponsesParams` declares the param already.

Typed as a free-form string rather than an enum because the tier vocabulary
is provider-specific and grows independently of this gateway, so the
provider is the right place to reject an unknown value.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

None.

## Checklist

`make lint`, `make typecheck` and the full `pytest tests/unit tests/integration` run are green (2708 passed, 10 skipped). The request schema gained a field, so `docs/public/openapi.json` was regenerated; `make openapi-check` and `make postman-check` pass.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

The change was written and validated with Claude Code. It was previously part of a
longer branch of mine; that branch has been rebased onto current `main` and split so
each PR carries one self-contained change, reviewable on its own.

## Summary

Preserve `service_tier` in chat requests and forward it to providers when set. Providers can validate their supported tier values.

## Testing

Added coverage for streaming and non-streaming requests, including requests without `service_tier`. Regenerated OpenAPI documentation and confirmed all required checks pass.

### DIFF
--- a/docs/public/openapi.json
+++ b/docs/public/openapi.json
@@ -445,7 +445,7 @@
         "type": "object"
       },
       "ChatCompletionRequest": {
-        "description": "OpenAI-compatible chat completion request.\n\nThe completion-param fields are derived from any-llm's ``CompletionParams``\n(see ``_schema_derive``) so the schema cannot silently drop a param any-llm\nforwards. Fields below either tighten a derived field (``messages``,\n``response_format``) or add gateway-internal behavior (``mcp_servers``,\n``mcp_server_ids``, ``guardrails``, ``tools_header``, ``max_tool_iterations``)\nthat is stripped before the request is forwarded upstream.",
+        "description": "OpenAI-compatible chat completion request.\n\nThe completion-param fields are derived from any-llm's ``CompletionParams``\n(see ``_schema_derive``) so the schema cannot silently drop a param any-llm\nforwards. Fields below either tighten a derived field (``messages``,\n``response_format``), declare an OpenAI wire param ``CompletionParams`` does\nnot model (``service_tier``, forwarded as an any-llm ``**kwargs`` param), or\nadd gateway-internal behavior (``mcp_servers``, ``mcp_server_ids``,\n``guardrails``, ``tools_header``, ``max_tool_iterations``) that is stripped\nbefore the request is forwarded upstream.",
         "properties": {
           "frequency_penalty": {
             "anyOf": [
@@ -652,6 +652,17 @@
               }
             ],
             "title": "Seed"
+          },
+          "service_tier": {
+            "anyOf": [
+              {
+                "type": "string"
+              },
+              {
+                "type": "null"
+              }
+            ],
+            "title": "Service Tier"
           },
           "session_label": {
             "anyOf": [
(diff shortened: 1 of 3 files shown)


njbrake (changes requested):

The diagnosis and the fix are right, and I confirmed the tests are load-bearing: removing the field declaration fails both `test_service_tier_reaches_provider` cases. Lint, typecheck, unit, and the full integration suite are green locally, and `openapi-check` and `postman-check` both pass. Requesting changes only because the two edits below are comment text and we cannot push to your fork, so they need a commit from you.

**1. Frame the declaration as a stopgap.** The underlying gap is that any-llm's `CompletionParams` does not model `service_tier` even though `ResponsesParams` does; I filed mozilla-ai/any-llm#1269 for that, tracked here as #565. Worth saying so in the comment, because a subclass field fully overrides the derived one: as long as this declaration stays, our `str | None` shadows whatever annotation any-llm eventually picks, and if upstream types it as a literal the looser type silently persists in the spec and in every generated SDK with nothing in CI to flag it. Hand-declaring is also the pattern that would invite the other eleven unmodeled OpenAI chat params in one by one, so the note is what keeps this a one-off.

**2. Anthropic's request-side values are `auto` and `standard_only`.** `priority` is usage-side, on the response. OpenAI's list also includes `auto` and `default`.

Both in one patch, on top of your current `chat.py`:

```diff
-    # provider-specific ("flex"/"priority"/"scale" on OpenAI,
-    # "standard_only"/"priority" on Anthropic) and grows independently of this
+    # provider-specific ("auto"/"default"/"flex"/"scale"/"priority" on OpenAI,
+    # "auto"/"standard_only" on Anthropic) and grows independently of this
     # gateway, so the provider is the right place to reject an unknown value.
     # ``ResponsesParams`` already declares it, so /v1/responses never had the gap.
+    #
+    # Stopgap: remove this declaration once ``CompletionParams`` models the param
+    # and the SDK pin is bumped (mozilla-ai/any-llm#1269, tracked in #565). Until
+    # then it also shadows whatever annotation any-llm picks for it.
     service_tier: str | None = None
```

This touches only an inline comment, not the class docstring, so no artifact regeneration is needed. Happy to merge once it is pushed.

Two follow-ups I will file separately, not for this PR: `/v1/messages` still drops `service_tier` (`MessagesParams` does not model it either), and `_is_unsupported_feature_error` matches only `NotImplementedError`, so an upstream `UnsupportedParameterError` would surface as a 502 rather than a 400.

njbrake (approved):

njbrake (approved):

CI is green on 704dd0ff, and the fix is unchanged from what I verified. The review edits are in. Approving.

PR #562, opened 2026-08-12
fix(models): tolerate a provider that reports no model `created`

pss-julien:
## Description

`Model.created` is typed `int`, but the OpenAI SDK builds response models with
`construct_type` and no validation, so an OpenAI-compatible provider answering
`"created": null` reaches the route as `None`. Passing that into `ModelObject`
raised a pydantic `ValidationError`, so one such model turned the whole
`GET /v1/models` listing into a 500 (same for `GET /v1/models/{id}`). Local
OpenAI-compatible servers are the common source: several omit `created`
entirely or send `null`.

Coerce through `_created_timestamp()`, falling back to 0 as the other catalog
phases already do for models with no known creation time. Raising here would
fail an entire listing over one provider's payload, and `created` is
informational in this response, so degrading that single field is the
proportionate response. The helper catches `TypeError`/`ValueError`, which
also covers a non-numeric value arriving by the same unvalidated path.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

None.

## Checklist

`make lint` and `make typecheck` are green, as is the full `pytest tests/unit tests/integration` run (2707 passed, 10 skipped). New integration coverage in `tests/integration/test_model_discovery.py` pins both routes: a provider reporting `created: null` now lists alongside a well-formed model and reports `created: 0` instead of 500ing. No schema or route signature changed, so the OpenAPI spec is untouched.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

The change was written and validated with Claude Code, on a branch off current `main`
carrying this one self-contained fix.

### DIFF
--- a/src/gateway/api/routes/models.py
+++ b/src/gateway/api/routes/models.py
@@ -2,7 +2,7 @@
 
 import calendar
 from collections.abc import Sequence
-from typing import Annotated
+from typing import TYPE_CHECKING, Annotated
 
 from fastapi import APIRouter, Depends, HTTPException, Query, status
 from pydantic import BaseModel, Field
@@ -37,6 +37,9 @@ from gateway.services.pricing_service import (
     normalize_effective_at,
 )
 from gateway.services.provider_kwargs import normalize_pricing_key
+
+if TYPE_CHECKING:
+    from any_llm.types.model import Model
 
 router = APIRouter(prefix="/v1", tags=["models"])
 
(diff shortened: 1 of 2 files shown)


njbrake (approved):

Approving. I verified the premise rather than taking it on trust: reverting `models.py` to the base version makes both new tests fail with the pydantic `int_type` error at the `ModelObject(...)` call, which sits outside the `try` wrapping `discover_all_models`, so it does 500 the whole listing. Driving a real `AsyncOpenAI` client against a mock transport confirms that `"created": null` and an omitted `created` both arrive as `None`, and that any-llm's `BaseOpenAIProvider._convert_list_models_response` passes SDK-built objects through without revalidating. The 0 fallback is consistent with the other catalog phases (`models.py:234`, `276`, `699`) and with `model_discovery_service.py:289`.

`make lint`, `make typecheck`, `make openapi-check`, and `make postman-check` all pass, as does the full suite (2701 passed, 16 skipped).

One follow-up worth folding in: `OverflowError` escapes the `except`. JSON places no range limit on a number, so a literal such as `1e400` parses to `float("inf")`, and `int()` rejects that with an `ArithmeticError` rather than a `ValueError` or `TypeError`, which 500s the listing again through the same path this PR closes. Widening to `except (TypeError, ValueError, OverflowError)` covers it. Related: both new tests reach only the `None` branch, so the non-numeric coercion the docstring advertises currently has no test (line coverage cannot tell the two branches apart).

I am also filing the same class of fix upstream in any-llm's `providers/openai/base.py`, where the `isinstance(model, Model)` guard makes the `model_validate` branch dead, while sibling providers (openrouter, requesty) already fill `created = 0`. That is not a blocker here: the gateway needs its own guard at the trust boundary regardless, and a stricter upstream would fail a whole provider listing instead of degrading one field.

PR #567, opened 2026-08-13
feat(telemetry): receive OTLP outcome metrics and read them per unit of spend

### LINKED ISSUE #429: Receive what coding agents already emit: OTLP metrics signal and non-usage behavioral events

## Summary

Coding agents already emit substantially more telemetry than otari can receive or keep. Two distinct gaps: one whole OTLP signal has no receiver, and four of Claude Code's five log events are dropped at mapping time. Together they are the difference between measuring what was *spent* and measuring what was *achieved*.

## 1. There is no OTLP metrics receiver

`src/gateway/api/routes/otlp.py` implements `POST /v1/traces` and `POST /v1/logs`. There is no `/v1/metrics`, and no reference to `ExportMetricsServiceRequest` anywhere in `src/`.

Claude Code emits eight metrics on that signal (names confirmed in the 2.1.220 binary):

```
claude_code.token.usage             claude_code.lines_of_code.count
claude_code.cost.usage              claude_code.commit.count
claude_code.session.count           claude_code.pull_request.count
claude_code.active_time.total       claude_code.code_edit_tool.decision
```

The bottom four are **outcomes**, and they are what turns a consumption dashboard into an optimization tool: cost per commit, cost per PR, cost per line changed, edit accept/reject rate, and spend per unit of active time. Otari can currently measure what an agent cost but nothing about what it produced.

Two notes on scope:

- These are counters, not usage rows. They probably do not belong in `usage_logs` — they need somewhere to live that supports "outcome per unit spend" joins against usage.
- `OTEL_METRICS_EXPORTER` is a separate env var from `OTEL_LOGS_EXPORTER`. Any setup that follows a logs-only guide ships none of these, so the docs need a corresponding change or the receiver will look broken.

## 2. Non-usage agent events are silently dropped

`_build_event` special-cases `api_request` (Claude Code) and the Codex events, then falls through to the GenAI semantic conventions and returns `None` when there is no model/provider/token data. Claude Code's other four log events — `tool_result`, `tool_decision`, `user_prompt`, `api_error` — have none of those attributes, so all four are discarded.

They carry no token counts, so they genuinely don't belong in `usage_logs`. But counted per session they yield:

- **tool call volume and mix** — what the agent actually spends its turns doing
- **tool acceptance rate** (`tool_decision`) — how often a human rejects the agent's proposed edit, which is a direct quality signal
- **turns per session** (`user_prompt`) — human interaction cost alongside token cost
- **real error and retry rates** (`api_error`) — currently only visible as a `status` on rows that produced usage

This is the raw material for most behavioral analytics. Even storing counts without pricing them would be a large step up.

## Suggested sequencing

The metrics receiver is the bigger lift but has the clearer payoff (cost-per-outcome). The behavioral events are cheaper — they arrive on a signal that already has a receiver — and mostly need a decision about where non-usage events should be stored.

arthuursantos:
## Description

Closes the remaining half of #429. PR #548 added the behavioral events on the logs signal; this adds the metrics signal Otari had no receiver for, plus the read API that gives recorded spend a denominator.

**`POST /v1/metrics`** records the four outcome counters an agent reports and Otari has no other source for: lines of code changed, commits, pull requests, and active time. Rows are content-free and non-billable. Two classes of metric are recognized and deliberately skipped rather than treated as unknown:

- `claude_code.token.usage` and `claude_code.cost.usage`, already billed from the `api_request` usage event, so recording them here would double count spend.
- `claude_code.code_edit_tool.decision`, already captured as the `tool_decision` behavioral event, so the edit acceptance rate has exactly one authoritative source.

An export carrying metric names the gateway does not know is still accepted, with the recognized points recorded, so a newer agent version never breaks reception.

**Storage.** `agent_telemetry` gains five nullable columns (`kind`, `value`, `temporality`, `series_start`, `series_key`) plus a `(series_key, timestamp)` index, the ones the behavioral-events PR reserved for this receiver rather than adding then. Behavioral rows read back as NULL, which is exactly "not a metric row", so nothing is backfilled. Series identity is the OTLP data model's own (metric name plus its full attribute set), so a dimensioned counter such as lines added against lines removed stays two series and is only summed when a combined measure is asked for.

**Cumulative counters are not diffed at ingest.** Each point is stored exactly as OTLP reported it and deduped by point identity (series identity, series start, point time, importing user), so a retried export or a re-reported running total adds nothing. The delta arithmetic happens at read time, split per series generation: a counter reset arrives as a changed `series_start` and starts a fresh generation, so the diff restarts there and never yields a negative increment.

**Read API.** `GET /v1/agent-telemetry/summary`, `/count`, and `/series`, master key only, mirroring the shape of the `/v1/usage` family. Only `summary` joins usage, reporting cost per commit, cost per pull request, cost per line changed, spend per active hour, tool call volume and mix, tool acceptance rate, turns per session, and error rate, each null rather than an error when its denominator is zero. It filters by session as well as by user and API key, matching the session on both sides of the join (`agent_telemetry.session_label` against `usage_logs.source_label`). `count` and `series` keep the purge endpoint's filter set, so `count` still sizes exactly what a "delete all N matching" would remove.

**Nothing new was built where something already existed.** Ingestion reuses the existing OTLP request plumbing (size cap, gzip-bomb guard, master-key refusal), the `capture_agent_telemetry` toggle, and the batched `ingest()` pipeline; a 1,000 point export costs one INSERT, asserted in the tests rather than assumed. The metrics endpoint adds only the one safeguard this signal specifically needs: its own per-export data-point cap, because metric attribute cardinality is caller controlled unlike the fixed behavioral-event shape. The existing purge and user-deletion cleanup already cover metric rows, which the tests confirm instead of adding a second deletion path.

**Docs.** `OTEL_METRICS_EXPORTER` is a separate setting from `OTEL_LOGS_EXPORTER`, so anyone following the previous logs-only guidance would have shipped none of this and the new receiver would look broken. The setup guide now enables both and states exactly which metric fields are captured, and which are deliberately not.

Out of scope, unchanged from the issue: dashboard charts over the new measures, any pricing or budget gating of outcome metrics, `claude_code.session.count`, and backfilling data emitted before this existed.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [x] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Closes #429

## Checklist

On the two unchecked boxes, so the record is exact rather than optimistic:

- `make lint` and `make typecheck` are green as of the final commit. `make test` was last run green in full (2800 passed, 15 skipped) before the last round of changes; after them the telemetry, OTLP, and usage suites were run green (189 tests), not the whole suite. To be ticked once the full suite is re-run.
- The first box is the author's own attestation, left for the author rather than checked by the agent that drafted this.

Tests added: `tests/unit/test_agent_telemetry_delta.py` (read-time increment math, generations, resets), `tests/integration/test_otlp_metrics.py` (receiver happy and edge paths, dedup, caps, capture toggle, bulk-insert bound), `tests/integration/test_agent_telemetry_read.py` (the three read endpoints, auth and mode gating), plus cases appended to `tests/unit/test_agent_telemetry_mapping.py` and `tests/integration/test_agent_telemetry_admin.py`. The Postman collection was regenerated alongside the OpenAPI spec; `make openapi-check` and `make postman-check` both pass.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

Produced with the repository's spec-driven workflow (specify, plan, tasks, implement, converge). The convergence pass audited the implementation against the spec and appended four remaining items, which are included here: an assertion that the bulk ingest costs a bounded number of INSERT statements rather than only checking the resulting row count, the session filter on `/summary`, a test for a metric point carrying no session id, and a line-ending normalization so the diff shows the real change size.

**NOTE:**

## Summary

- Added OTLP metrics ingestion through `POST /v1/metrics` for agent outcome data.
- Added master-key-protected telemetry summary, count, and series APIs.
- Added metric storage, deduplication, counter-delta handling, filtering, and cleanup.
- Updated documentation, configuration guidance, and Postman examples.
- Added unit and integration tests for ingestion, aggregation, limits, access control, and deletion.

These changes enable outcome reporting such as lines changed, commits, pull requests, and active time without affecting billing usage.


njbrake (approved):
Approving. Nice work: the receiver extends the existing OTLP plumbing (size cap, gzip guard, master-key refusal, batched `ingest()`) rather than building a second path beside it, and `map_metric_point` mirrors `map_behavioral_event` closely enough to read as one mechanism.

I pushed 716a5711 to the branch with the review fixes rather than sending you round again. The one that mattered: cumulative series were diffed pairwise, so the first reading of every generation was dropped. An OTel counter exports no data point until its first measurement, so that reading already carries work. I confirmed against the SDK that a fresh counter emits nothing, then value 1, 2, 3, with a constant start time. Three commits read as two, and a session whose only export carried one commit read as zero, which inflates cost per commit and cost per line. A generation whose `series_start` falls inside the window is now diffed from its own zero; one that began earlier keeps the old conservative behavior.

That bug survived a careful suite because the fixture seeded a synthetic leading `0.0` point that a real export never sends. Worth remembering for the next telemetry change: the payload builders encode our model of the agent, so they cannot catch a wrong model.

Also fixed: an unbounded metric scan in `/summary` (the delta math runs in Python, so that read grew with export count rather than window), `_point_value` storing proto3's default `0` for a valueless point, naive date bounds diverging between `/count` and the purge it promises parity with, and `_SKIPPED_METRICS` being defined but never read. Docs: `configuration.md` still described the toggle as logs-only.

I amended `test_summary_counter_reset_never_yields_a_negative_increment`. Its first generation had `series_start` equal to its own first point's timestamp, which says the counter started and reached 40 at one instant; it reached 9.0 only because of the bug. Same intent and same expectation now, on data that means what it claims.

Full suite green locally against PostgreSQL 18 (2808 passed, 16 skipped), which closes the last checklist item.

One thing this cannot verify: no test here has seen a real Claude Code export. Worth pointing one session at a staging gateway, mainly to confirm the metric attributes carry `session.id`, since without it the session filter returns zero outcomes against real spend.

PR #573, opened 2026-08-14
fix(dashboard): bound and await sign-out so it can't clobber a fresh sign-in

### LINKED ISSUE #557: [BUG] Dashboard sign-out is not awaited or bounded, so a stalled request can sign out the next session

**Run mode:** standalone (dashboard only)
**How you're running Otari:** from source and Docker, unaffected by either

### What happened

Signing out of the dashboard fires `DELETE /v1/auth/session` without awaiting it and without a timeout, so a slow sign-out can silently sign the operator out of the session they open afterwards.

`AuthContext.logout` (`web/src/auth/AuthContext.tsx:35`) calls `void deleteSession()` and immediately clears local state. `deleteSession` (`web/src/api/client.ts:69`) uses raw `fetch` with no `signal`, so unlike `apiFetch` it is not bounded by `REQUEST_TIMEOUT_MS` (30s). `createSession` has the same gap.

The damage is done by the response, not by the request. `delete_session` (`src/gateway/api/routes/auth_session.py:120`) revokes only the token in the caller's own cookie, so it cannot revoke a newer session server-side. But it always ends with `clear_session_cookie(response)`, an expiring `Set-Cookie`. If that response lands after a later `POST /v1/auth/session` has minted a new cookie, the browser drops the new cookie. The server-side session stays alive and unreferenced; the operator is bounced to the sign-in screen for no visible reason, and any in-flight admin request 401s.

Narrow but real: it needs the sign-out request to stall long enough to outlive a fresh sign-in, which is the case the missing timeout allows. An unreachable or stalled gateway at sign-out time is the trigger.

### Steps to reproduce

1. Sign in to the dashboard.
2. Make `DELETE /v1/auth/session` hang (a proxy that holds the request, or a breakpoint in the handler before the response is written).
3. Sign out. The UI returns to the sign-in screen at once, since the call is not awaited.
4. Sign in again with the master key. The dashboard renders as signed in.
5. Release the stalled request. The browser applies its expiring `Set-Cookie` and the next admin call 401s back to sign-in.

### What you expected

Sign-out completes before the sign-in screen accepts a new credential, and a hung sign-out gives up on the same deadline as every other management call.

Suggested fix: `await deleteSession()` in `logout` with an in-flight guard so a second sign-in cannot start while one is pending, and pass `AbortSignal.timeout(REQUEST_TIMEOUT_MS)` to both session calls. The 401-bounce path already tolerates a rejected sign-out, and `deleteSession` swallows failures, so awaiting it does not add a new failure mode.

### Notes

Found while reviewing #553, where CodeRabbit flagged it against a generated dashboard bundle chunk. It is unrelated to that PR: `web/src/api/client.ts` is untouched there. The bot read it as the delete revoking the new session server-side, which the handler cannot do; the cookie-clearing response is the actual mechanism.

champ18ion:
## Description
`DELETE /v1/auth/session` (`deleteSession`) had no `AbortSignal`, and its promise was never awaited in `AuthContext.logout` (`void deleteSession()`). A stalled sign-out could stay in flight past a subsequent sign-in.

The damage comes from the response, not the request: `delete_session` always ends with `clear_session_cookie(response)`, an expiring `Set-Cookie`. If that response lands after a later `POST /v1/auth/session` has already minted a new session cookie, the browser drops the fresh cookie in favor of the expiring one. The new session stays alive and unreferenced server-side, and the operator is bounced to the sign-in screen for no visible reason.

Fix, matching the approach suggested in the issue:

- `deleteSession`/`createSession` (`web/src/api/client.ts`) now pass `AbortSignal.timeout(REQUEST_TIMEOUT_MS)`, the same bound `apiFetch` already uses for every other management call.
- `AuthContext` (`web/src/auth/AuthContext.tsx`) exposes a new `isSigningOut` flag: `true` from the moment `logout()` fires the `DELETE`, `false` once it resolves (now bounded by the timeout above). Local sign-out itself stays synchronous and unconditional, exactly as before — the UI still returns to the sign-in screen immediately.
- `Login`'s submit (`web/src/components/Login.tsx`) is gated on `isSigningOut`, so a new sign-in cannot start while a prior sign-out's revocation might still be in flight and clobber the new session's cookie when it eventually lands.

## PR Type

- [x] Bug Fix

## Relevant issues
Fixes #557

## Checklist

- [x] I ran the Definition of Done checks locally (`npm run typecheck`, `npm test` — this PR only touches `web/`, so the Python `make lint`/`make typecheck`/`make test` targets don't apply; the dashboard CI workflow runs `npm run typecheck`, `npm test`, and `npm run build` instead, and I ran the first two directly).
- [ ] If the API contract changed, I regenerated the OpenAPI spec (`uv run python scripts/generate_openapi.py`). — N/A, no API contract change.

## AI Usage

**Any additional AI details you'd like to share:** I picked this one up off the good-first-issue-style backlog, read through the repro and the suggested fix in the issue, and worked through the implementation with Claude Code: deciding the `isSigningOut` gate belongs on the sign-in form rather than blocking sign-out itself, choosing to keep local sign-out synchronous/unconditional so the UX doesn't regress, and writing out both the isolated state-transition test and the full blocked-then-unblocked flow test. I reviewed and tested the diff before opening this PR.

## Summary

- Prevents stale sign-out requests from affecting a newly created session.
- Shows sign-out progress while local sign-out completes immediately.
- Blocks login until all sign-out requests finish.
- Adds request timeouts and regression tests for session handling and login flow.

## Technical notes

- Adds `isSigningOut` to `AuthContext`.
- Applies timeout-bounded `AbortSignal`s to `createSession` and `deleteSession`.
- Tracks concurrent sign-out requests until all requests settle.

### DIFF
--- a/web/src/api/client.test.ts
+++ b/web/src/api/client.test.ts
@@ -1,6 +1,6 @@
 import { afterEach, describe, expect, it, vi } from "vitest";
 
-import { ApiError, apiFetch } from "./client";
+import { ApiError, apiFetch, createSession, deleteSession } from "./client";
 
 afterEach(() => {
   vi.restoreAllMocks();
@@ -83,5 +83,51 @@ describe("apiFetch", () => {
     await expect(apiFetch("/v1/models")).rejects.toMatchObject({
       message: expect.stringContaining("could not reach the gateway"),
     });
+  });
+});
+
+describe("createSession", () => {
+  it("bounds the request the same way apiFetch does", async () => {
+    const seen: (AbortSignal | null | undefined)[] = [];
+    vi.spyOn(globalThis, "fetch").mockImplementation((_input, init) => {
+      seen.push(init?.signal);
+      return Promise.resolve(new Response("{}", { status: 200 }));
+    });
+
+    await createSession("test-key");
+
+    expect(seen[0]).toBeInstanceOf(AbortSignal);
+  });
+
+  it("maps a timeout to the same bounded-request message apiFetch uses", async () => {
+    vi.spyOn(globalThis, "fetch").mockRejectedValue(new DOMException("timed out", "TimeoutError"));
+
+    await expect(createSession("test-key")).rejects.toMatchObject({
+      status: 0,
+      message: expect.stringContaining("did not respond within 30s"),
+    });
+  });
+});
+
+describe("deleteSession", () => {
+  it("bounds the request the same way apiFetch does", async () => {
+    const seen: (AbortSignal | null | undefined)[] = [];
+    vi.spyOn(globalThis, "fetch").mockImplementation((_input, init) => {
+      seen.push(init?.signal);
+      return Promise.resolve(new Response(null, { status: 204 }));
+    });
+
+    await deleteSession();
+
+    expect(seen[0]).toBeInstanceOf(AbortSignal);
+  });
+
+  it("still resolves, rather than throwing, when the request times out", async () => {
+    // deleteSession is best-effort and swallows failures: AuthContext.logout
+    // relies on this promise always settling, timeout or not, to clear
+    // isSigningOut and unblock a subsequent sign-in (see #557).
+    vi.spyOn(globalThis, "fetch").mockRejectedValue(new DOMException("timed out", "TimeoutError"));
+
+    await expect(deleteSession()).resolves.toBeUndefined();
   });
 });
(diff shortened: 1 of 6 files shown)


tbille (approved):
super cool, thanks for the contribution @champ18ion

PR #592, opened 2026-08-14
test(dashboard): pin dashboard behaviour with an e2e parity baseline

### LINKED ISSUE #579: Behavioral-parity e2e baseline for the existing dashboard
Pin the current dashboard behavior in e2e tests before any convergence work touches it, so the UI rehome cannot silently regress what self-hosters rely on today.

## What to produce

- Playwright coverage in `web/e2e` for: standalone provider setup, API keys, budgets, fallback routing, Overview, Activity, and Usage.
- Behavioral assertions (filters, drill-downs, request detail), not page reachability.
- Screenshot/visual-regression coverage for each page, as the retirement gate: no existing page is rebuilt or retired without it.

## Definition of Done

- The suite runs as a required CI check and passes on current `main`.
- Every page listed above has behavioral and visual coverage.

Feeds the parity gate tracked as mozilla-ai/otari-ai#1600.

njbrake:
## Description

Records what the existing dashboard does, in Playwright, before the UI rehome starts moving pages around.

The e2e suite already booted a real gateway against the built bundle, but past the onboarding flows it asserted little more than "the page loaded": Activity had no coverage, Usage only the share card, Overview only its first-run screen. This adds 19 behavioural flows across Overview, Activity, Usage, provider setup, API keys, budgets and fallback routing: filters narrowing the log and their chips clearing it, priced against unpriced reconciling, a shared URL restoring its filters and page, a request's detail panel, breakdown rows drilling into the requests behind a number, and the delete guards on keys and budgets.

The fixture is imported through `POST /v1/usage/external-events` rather than served by provider calls, so it writes real priced usage rows with no upstream to reach and no network egress. Everything it writes is namespaced and every assertion is scoped to that namespace: these pages are gateway-wide and the suite shares one database, so an unscoped count would assert on whatever ran first. That sharing is also why ordering is now three declared Playwright projects (`onboarding` → `seed` → `parity`) instead of the alphabetical order of filenames, since the onboarding flows need the empty database `serve.sh` leaves behind.

**No screenshot or visual-regression layer**, which issue #579 also asked for. The UI is expected to change substantially over the coming months, so @njbrake's call was that pixel baselines would cost more than they caught. Worth noting for anyone revisiting this: baselines could be produced despite the architecture mismatch between a dev machine and the runner, by generating them in CI or in a fixed container. The reason for skipping is the churn ahead, not a technical block.

**Not done here, worth a follow-up:** the `e2e` job is scoped to `web/**` paths, so a `src/gateway/` change that breaks a usage filter or the summary endpoint will not run this suite. Widening the trigger puts a browser job on every backend PR, so it seemed worth deciding separately.

## PR Type

- [ ] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [x] Infrastructure / CI

## Relevant issues

Fixes #579

## Checklist

- [ ] I ran the Definition of Done checks locally (`make lint`, `make typecheck`, `make test`). No Python changed, so the relevant checks are the dashboard ones: `npm --prefix web run typecheck` (clean, and now covers `web/e2e` too via a new `tsconfig.e2e.json`), `npm --prefix web test` (603 passed), `npm --prefix web run e2e` (31 passed, run repeatedly for stability).
- [x] Documentation was updated where necessary. `web/AGENTS.md` describes the project layout and the react-aria popover trap the helpers work around.
- [ ] If this PR adds or changes a persistent table, a control-plane contract the otari.ai platform consumes or serves, a capability the platform also has, or a mode-specific surface, I appended an entry to the M4 reconciliation ledger. Test-only, so none of those apply.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:** Claude Opus 5, via the Claude Code CLI.

**Any additional AI details you'd like to share:** Written by Claude Opus 5 through back-and-forth with @njbrake, who made the call to drop the screenshot layer. Every flow was run against a real gateway and iterated until green.

## Summary

- Added a Playwright behavioral baseline for the dashboard.
- Added 19 flows for Overview, Activity, Usage, provider setup, API keys, budgets, and fallback routing.
- Added shared test helpers and namespaced usage fixtures.
- Ordered Playwright projects as `onboarding` → `seed` → `parity`.
- Added E2E type checking and failure trace uploads.
- Documented the test suite and fixture behavior.

## Benefits

This baseline records current dashboard behavior before UI rehoming. It checks filters, URL state, usage breakdowns, request details, and deletion safeguards.

## Technical notes

Visual-regression tests are not included. CI coverage for backend changes remains a follow-up.


web/e2e/helpers.ts
```diff
@@ -0,0 +1,126 @@
...
+
+// ---------- stat tiles ----------
+
+// A StatCard renders its label and its value as sibling spans with no
+// programmatic association between them, so the value is reached through the
+// label. Scoped to `within` where a bare label would also match a table column
+// header of the same name (Usage has both a "Requests" tile and a "Requests"
+// column).
+export function tileValue(page: Page, label: string, within?: Locator): Locator {
+  return (within ?? page).getByText(label, { exact: true }).locator("xpath=following-sibling::span[1]");
```
khaledosman, line 121: `following-sibling::span[1]` is `StatCard`'s value **wrapper** (`web/src/components/ui.tsx:50`), which holds the value span plus the optional status pill, not the value itself.

None of the four tiles asserted in `parity.usage.spec.ts:37-39` carry a `statusLabel` today, so `not.toHaveText("—")` is a real check right now. But "Error rate" and "Budget health" already do carry one: the moment one of these tiles gains a pill, its text becomes `"—Elevated"`, `not.toHaveText("—")` passes on a tile literally showing unknown, and the baseline stops catching the regression it was written for.

```suggestion
  return (within ?? page).getByText(label, { exact: true }).locator("xpath=following-sibling::span[1]/span[1]");
```

`.click()` on the inner span still triggers the `to` link (`parity.overview.spec.ts:69`), and `parity.overview.spec.ts:43` could then anchor its regex with `$` instead of relying on the status word trailing it.

web/playwright.config.ts
```diff
@@ -14,12 +14,43 @@ export default defineConfig({
...
   retries: 0,
-  reporter: process.env.CI ? "list" : "line",
+  // "list" everywhere, not "line" locally: the gateway's own request log is piped
+  // into this stream (see `webServer.stdout`), and the line reporter rewrites its
+  // single status line with carriage returns, so interleaving the two shreds the
+  // failure detail exactly when it is needed.
+  reporter: "list",
   use: {
     baseURL: "http://127.0.0.1:8000",
     trace: "on-first-retry",
```
khaledosman: `on-first-retry` with `retries: 0` (line 16) means a trace is never captured. 19 browser flows over a shared database, and a CI failure yields nothing but the list reporter's text; the `e2e` job in `otari-dashboard.yml` has no artifact-upload step either.

`trace: "retain-on-failure"` (optionally plus `screenshot: "only-on-failure"`) costs nothing on green runs, and is the difference between debugging a CI-only failure and re-running it locally hoping it reproduces.

web/e2e/parity.overview.spec.ts
```diff
@@ -0,0 +1,72 @@
...
+import { login, table, tableRows, tileValue } from "./helpers";
+import { COUNTS, PARITY } from "./parity-data";
+
+// The seeded fixture is shared, and nothing here mutates it, but the suite runs
+// one worker against one gateway either way (see web/playwright.config.ts).
+test.describe.configure({ mode: "serial" });
+
+// The scratch rows are consumed by the Activity spec's bulk-delete flow, which
+// runs before this file, so they are deliberately not counted here.
+const SEEDED = COUNTS.priced + COUNTS.unpriced + COUNTS.errors;
```
khaledosman, line 12: The comment above claims an ordering the config does not declare. `dependencies` order the three projects, but inside `parity` the four spec files still run in filename order, so "which runs before this file" is an alphabetical accident rather than a guarantee.

The assertion is a floor and holds either way, which is the point worth writing down: say the scratch rows are excluded so the floor is correct whether or not the bulk-delete has already run, rather than documenting an ordering a future rename silently breaks.

web/e2e/parity.usage.spec.ts
```diff
@@ -0,0 +1,118 @@
...
+  });
+
+  test("breaks spend down by model and by user", async ({ page }) => {
+    await openUsage(page);
+
+    // Model is the default primary dimension.
+    await expect(page.getByRole("heading", { name: "Spend by model" })).toBeVisible();
+    await expect(breakdownRow(page, "model", PARITY.models.priced.model)).toBeVisible();
+    await expect(breakdownRow(page, "model", PARITY.models.priced.model)).toContainText(
+      String(COUNTS.priced),
```
khaledosman, line 76: `toContainText(String(COUNTS.priced))` is a bare `"30"` matched against the whole row, which also carries cost, tokens and a share percentage. It passes on any row containing `30` anywhere, so a wrong request count can survive it. Assert the Requests gridcell instead of the row.

web/e2e/parity.management.spec.ts
```diff
@@ -0,0 +1,213 @@
...
+    await expect(policy).toContainText(/\+2 on failure/);
+
+    // Shrinking has to be possible too: a chain that could only grow would strand
+    // an operator who added a candidate by mistake.
+    await policy.getByRole("button", { name: "Edit" }).click();
+    await expect(page.getByRole("combobox", { name: /Fallback 2/ })).toBeVisible();
+    // One Remove per chain entry, in order. The form's other Remove controls
+    // belong to the conditions, candidate-pool and guardrail sections, none of
+    // which this policy has, so they are not rendered.
+    await page.getByRole("button", { name: "Remove" }).nth(1).click();
```
khaledosman: Positional index over a substring name match. Correct today (the chain removes are the only `Remove` buttons rendered for this policy, `web/src/pages/RoutingPage.tsx:802`), but the comment above has to enumerate which sections are *not* present to stay true, and it stops being true the first time this form grows a section.

Anchor to the entry instead: filter the container that `has` the `Fallback 2` combobox, then click its `Remove`.

khaledosman (approved):
Typecheck run at this head (`tsc -b --noEmit`, clean), and `tsconfig.e2e.json` verified to genuinely cover `web/e2e` (a planted type error in `e2e/` fails the build). Every asserted UI string and aria-label cross-checked against `web/src` at head. No bundle rebuild or OpenAPI/Postman regen owed: no `web/src` or `docs/dashboard.md` change, no route or schema change.

Five non-blocking findings inline, the first two worth taking before merge.

---

Issue #589, opened 2026-08-14
Epic: Converge the control-plane UI in otari/web (M5 frontend rehome)


njbrake:
Otari is becoming a full self-hostable open-core control plane; one shared dashboard source will serve the standalone, hosted, and hybrid deployments. This epic tracks the `otari`-side frontend work: preparing `web/` to receive the platform UI, then converging the two dashboards on one design foundation while preserving current Otari dashboard behavior.

The product contract: the current Activity, Usage, Overview, and Routing behavior is the baseline and must survive; the platform's organization, workspace, membership, fixed-role, and budget-policy workflows port in; a hybrid gateway stays data-plane only with a landing page, never a second management UI. Tracked privately as mozilla-ai/otari-ai#1452 (rehome) and mozilla-ai/otari-ai#1591 (deployment and page-ownership contract).

## Deployment contexts

Mirrored from mozilla-ai/otari-ai#1591 (private), which is authoritative; change it there first.

| Context | Full dashboard location | Authoritative control plane |
| --- | --- | --- |
| Standalone | Local Otari instance | Local database, one bootstrapped organization with multiple workspaces |
| Hosted otari.ai | Hosted control-plane application | Hosted database |
| Hybrid gateway | Hosted otari.ai, not the local gateway | Hosted database |

A hybrid gateway is data-plane only. Its local root offers health, connection state, and a link to the hosted control plane, never a competing management UI for hosted organizations, credentials, routing, budgets, or usage. #583 consumes this as a bootstrap contract; #587 builds that landing page.

## Sub-issues, in rough dependency order

- [x] #579 Behavioral-parity e2e baseline for the existing dashboard
- [x] #580 Migrate web/ to TanStack Router file-based routing
- [x] #581 Generate the web API client from the OpenAPI spec
- [x] #582 Adopt the target frontend architecture and boundary lint in web/
- [x] #583 Deployment bootstrap: server contract + shell consumption
- [x] #584 Nav registry and entitlement gate seam
- [x] #585 Rehome the shared design foundation from the platform UI
- [x] #586 Rebuild existing dashboard pages on the shared design foundation
- [x] #587 Hybrid gateway landing page
- [x] #588 Port platform tenancy workflows (orgs, workspaces, members, roles, budget policies). Delivered by #626 for organizations, workspaces, members and roles; the remainder split into the three below
- [ ] #641 Invitations for organization and workspace membership
- [ ] #642 Workspace per-member budget defaults
[...]

PR #600, opened 2026-08-14
refactor(dashboard): migrate web/ to TanStack Router file-based routing

### LINKED ISSUE #580: Migrate web/ to TanStack Router file-based routing
Replace react-router-dom v7 with TanStack Router file-based routes. The incoming platform pages are built for that convention (route files export only `Route`, page components live in feature modules and reach the route via `getRouteApi`), and swapping now, while the page count is small, is one contained PR instead of a mid-rehome rewrite.

## What to produce

- File-based route tree under `web/src/routes/`; existing pages keep their URLs.
- Route files export only `Route`, so the router code-splits every page; a route-exports test enforces it.

## Definition of Done

- All current pages served through TanStack Router with unchanged URLs; react-router-dom removed.
- The parity e2e baseline stays green.

Depends on #579.

njbrake:
## Description

Replaces react-router-dom with TanStack Router file-based routing, ahead of the platform pages arriving in `web/`. They are written for this convention, so the swap happens while the page count is small.

Routes move from the `<Routes>` table in `App.tsx` to one file per URL under `web/src/routes/`, assembled into a committed `src/routeTree.gen.ts` (typecheck runs `tsc` alone and never invokes Vite). Each route file exports `Route` and nothing else, which is what lets `autoCodeSplitting` give every page its own chunk; `src/routes.test.ts` enforces that, since a stray export silently pulls a page's whole import graph into the entry bundle.

Two things needed care to keep URLs unchanged. Hash history stays, so `/#/...` still needs no server catch-all and cannot shadow `/v1` or `/assets`. And search params keep their existing encoding: these URLs are shared and bookmarked, and a repeated key (`?model=a&model=b`) is how a multi-value filter travels from Usage into Activity, which TanStack's default codec would JSON-encode. `src/lib/search.ts` supplies a codec that does not.

Worth a reviewer's attention: `useUrlParam` is deleted rather than ported, because nothing imported it and it had no test. The index nav link's `end`/`exact` handling is also gone, because TanStack matches by route rather than path prefix, so it was a no-op; verified in a browser across five routes before removing it.

## PR Type

- [x] Refactor
- [x] Infrastructure / CI

## Relevant issues

Fixes #580. Second item in epic #589 (M5 frontend rehome); depends on the parity baseline from #579.

## Checklist

No Python changed, so the two unchecked boxes do not apply: no API contract moved, and per the ledger instructions a refactor that touches no table, cross-repo contract, or duplicated capability needs no entry. `make lint` and `make typecheck` pass; the web checks are the relevant ones here (621 vitest, 32 Playwright, typecheck, build).

**Verification.** The #579 parity baseline is green unchanged, which is what makes this checkable. New tests: route exports, the retired `/aliases` and unknown-path redirects (neither reachable from the UI), the active nav link, and the Overview error-rate link's query. Each was confirmed to fail against a deliberate break rather than only observed passing.

I also drove the built app in Chromium: 74 checks over all 12 routes as cold deep links and via sidebar clicks, both redirects, back/forward, reload of a deep-filtered URL, the repeated-param round trip, a Usage drill-down, the mobile drawer, sign-out/in, rapid navigation, and keyboard activation. 74/74 with zero console errors, page errors, or failed requests. One apparent behavior change (page scroll resetting on navigation) turned out to be identical on `main`, measured against a baseline build rather than assumed.

Not verified end-to-end: the PricingWarning failure link only renders when gateway requests failed in the last hour, which this sandbox cannot produce (no upstream egress). Its href is asserted by a unit test and the banner itself renders correctly in a browser.

## AI Usage

**AI Model/Tool used:**

Implementation and tests written by the model through back-and-forth with @njbrake; the decisions are his. The browser drive above was run at his request before opening this PR, and it is what caught the missing coverage on the Overview link.

## Summary

- Replaced `react-router-dom` with TanStack Router file-based routing.
- Added route files for all existing pages.
- Preserved URLs, hash history, query parameters, legacy redirects, and unknown-path handling.
- Updated components and tests to use the new router.
- Removed the unused `useUrlParam` hook.
- Added route validation, router test utilities, and legacy-route coverage.
- Added CI validation for the committed route tree.

## Technical notes

- Added automatic route code splitting.
- Preserved repeated query keys for multi-value filters.
- Updated frontend and testing guidance.
- Lint, typecheck, unit tests, build, parity checks, and browser verification pass.


web/src/router.tsx
```diff
@@ -0,0 +1,28 @@
...
+  parseSearch,
+  stringifySearch,
+  // Each page is its own chunk, so the first visit to one waits on a fetch.
+  // Show the loader as soon as that wait starts and drop it the moment the chunk
+  // lands: the defaults hold it back a second and then keep it up for half of
+  // one, which turns an instant navigation into a visible stall.
+  defaultPendingComponent: () => <div role="status">Loading page…</div>,
+  defaultPendingMs: 0,
+  defaultPendingMinMs: 0,
+});
```
khaledosman, line 22: No `defaultErrorComponent`. Before this PR a page that threw took the SPA down; now TanStack catches it per route and renders its built-in `ErrorComponent` — an inline-`style` "Something went wrong!" block with a Show/Hide Error toggle, which is nothing like this dashboard. Point it at `ErrorBanner` so a page crash lands in house chrome.

web/src/router.tsx
```diff
@@ -0,0 +1,28 @@
...
+  // routes live under `/#/...` and need no server catch-all that could shadow
+  // `/v1` or `/assets`.
+  history: createHashHistory(),
+  parseSearch,
+  stringifySearch,
+  // Each page is its own chunk, so the first visit to one waits on a fetch.
+  // Show the loader as soon as that wait starts and drop it the moment the chunk
+  // lands: the defaults hold it back a second and then keep it up for half of
+  // one, which turns an instant navigation into a visible stall.
+  defaultPendingComponent: () => <div role="status">Loading page…</div>,
```
khaledosman, line 19: The pending tuning treats the chunk fetch as unavoidable, but it isn't for a click that follows a hover/focus/touch. Adding `defaultPreload: "intent"` fetches the route chunk on intent, so the loader stops appearing on ordinary navigation instead of just appearing faster. One line, and it works on touch too (TanStack preloads on `touchstart`).

web/vite.config.ts
```diff
@@ -73,8 +83,14 @@ export default defineConfig({
...
           heroui: ["@heroui/react"],
-          react: ["react", "react-dom", "react-dom/client", "react/jsx-runtime", "react-router-dom"],
+          react: ["react", "react-dom", "react-dom/client", "react/jsx-runtime"],
           "tanstack-query": ["@tanstack/react-query"],
+          // Its own chunk rather than folded in with React. The router pulls in
+          // @tanstack/router-core, history and store, and putting that graph in
+          // the react chunk reorders the entry's imports enough that HeroUI
+          // evaluates before React's exports exist ("Cannot read properties of
+          // undefined (reading 'createContext')" at first paint).
+          "tanstack-router": ["@tanstack/react-router"],
```
khaledosman, line 93: Both `react` and this chunk are eager entry imports (`modulepreload` in `index.html`). Measured against a `main` build: react 74.63 → 56.87 kB gzip, plus this new 31.92 kB gzip chunk, so first load is ~14.6 kB gzip heavier. Fine trade for the routing, but the description claims no behavior change and doesn't mention it — worth stating, since the gateway serves this bundle to every operator.

web/AGENTS.md
```diff
@@ -11,6 +11,10 @@ tokens, TanStack Query, Vitest), read
...
 - Home-screen install (PWA): `web/public/pwa/` holds the web app manifest and the PNG app icons (`icon-192`, `icon-512`, a `maskable` 512, and an `apple-touch-icon` for iOS), all generated from the Otari mark in `web/public/favicon.svg` (white mark on `--otari-brand`). Vite copies the directory into the bundle and `src/gateway/main.py` mounts it at `/pwa` alongside the dashboard, so it is standalone-only; `web/index.html` links the manifest plus the `apple-*`/`theme-color` meta tags. There is deliberately no service worker: Chrome dropped the fetch-handler requirement for menu installs, and the dashboard is live API data with nothing worth caching offline.
+- Routing: TanStack Router with file-based routes in `web/src/routes/`, one file per URL, assembled into `src/routeTree.gen.ts` by the Vite plugin (`@tanstack/router-plugin`) and wired up in `src/router.tsx`. The generated tree is committed, because `npm run typecheck` runs `tsc` alone and never invokes Vite; it is regenerated by any `npm test`/`npm run build`, so a stale one shows up as a dirty file. Three rules to keep:
```
khaledosman, line 14: `.github/skills/frontend-standards/SKILL.md:16` still says the dashboard has "no router … Do not introduce those to match some other repo". `CLAUDE.md` and `.github/instructions/frontend-standards.instructions.md` both send contributors to that skill first, so it now contradicts this bullet. Update it here.

web/src/routes.test.ts
```diff
@@ -0,0 +1,38 @@
...
+// That is what lets the Vite plugin's autoCodeSplitting lift each page's
+// component into its own chunk (see vite.config.ts): it rewrites a route file
+// into a definition half and a component half, and a second export would have to
+// be resolvable from both. In practice a stray helper exported from a route file
+// pulls that route's whole import graph back into the entry bundle, silently,
+// and the only symptom is a fatter first load. It is also the convention the
+// platform pages arriving in this repo are written to, where page components live
+// in feature modules and reach their route through `getRouteApi`.
+describe("route files", () => {
+  const files = readdirSync(ROUTES_DIR).filter((name) => name.endsWith(".tsx"));
```
khaledosman: The guard only reads `.tsx`. `$.tsx` and `aliases.tsx` contain no JSX and would naturally be written `.ts`; renamed or added that way they drop out of both the export assertion and the `files.length > 10` guard, silently. Use `/\.tsx?$/`.

web/src/pages/RoutingPage.tsx
```diff
@@ -1,6 +1,6 @@
+import { Link } from "@tanstack/react-router";
```
khaledosman, line 1: Router import sits above `@heroui/react` here and in `UsagePage.tsx:1`, while `AppShell.tsx`, `ui.tsx`, `OverviewPage.tsx` and `ProvidersPage.tsx` all keep heroui + react first with the router after. Move it down to match.

khaledosman (approved):
Reviewed the branch locally: typecheck, 621 vitest tests, and a production build all pass, and the committed `routeTree.gen.ts` regenerates identically. Verified the two claims in the description that are easy to get wrong: `Link to="/"` really does stop needing `end` (TanStack's `removeTrailingSlash("/")` returns `"/"`, so `/providers` fails the segment-boundary check), and a plain `<Link to="/usage">` drops the current query string rather than carrying it, so the root-level `validateSearch` doesn't leak one page's filters into another.

Approving; the notes below are non-blocking.

🤖 Review generated by Claude Code (Claude Opus 5), acting as an agent.

Issue #581, opened 2026-08-14
Generate the web API client from the OpenAPI spec


njbrake:
Replace the hand-written `web/src/api` client (`client.ts`, `types.ts`, `hooks.ts`) with a client generated from the gateway's OpenAPI spec, matching the SDK precedent already in the repo. This establishes the generated-client boundary (`web/src/client`, mocked in tests) that the ported platform pages assume, and it is the seam that later absorbs the platform routes without frontend churn.

## What to produce

- A client-generation script wired into the dev workflow and CI drift check.
- Existing consumers migrated from the hand-written client; tests mock at the client boundary.

## Definition of Done

- No hand-maintained request/response types remain for gateway endpoints; regeneration is reproducible and drift-checked in CI.

Prose by Claude Fable 5 via back-and-forth with @njbrake; the decisions are his.

PR #606, opened 2026-08-14
refactor(dashboard): generate the web API client from the OpenAPI spec

njbrake:
## Description

The dashboard hand-maintained 1038 lines of request and response types beside the gateway that serves them, so the two could drift with nothing to say so. They now come from `docs/public/openapi.json`.

`web/src/client/schema.ts` is generated by `openapi-typescript` and committed with a CI drift check, like the route tree. `web/src/client/index.ts` is the only hand-written part and declares no structure: it aliases the generated schemas to the names the dashboard already used, derives the query-parameter enums from their operations, and holds the guards that narrow a generated union. `web/src/api/client.ts` stays hand-written, because the session exchange, the 401 bounce and the two-tier timeouts are transport policy no generator emits.

**Generating first meant fixing what the spec said.** Three fields were declared as open JSON while the frontend modelled their real shapes, so generating against the spec as it stood would have *lost* type safety exactly where the code says the modelling was deliberate. `UsageEntry.pricing_breakdown` is now a union of `TokenChargeLine` and `UnitChargeLine`, the distinction that stops a token line rendering with an undefined per-call rate; `billing_meters` names its reserved `tools` map; `ModelPricingInfo.pricing_tiers` reuses `PricingTier`. Each keeps a permissive arm, because these models validate rows read back out of a database a self-hosted gateway has been accumulating across upgrades, and a strict union would turn one unrecognised row into a 500 for the whole page. Every SDK core gets the same precision, since they generate from this spec.

The wire format is unchanged, which took two attempts. Declaring `tools` put `"tools": null` on every row that never ran one, which six integration tests caught; suppressing that with a serializer then stripped `required` from the charge lines, weakening the very types this exists to add. Both are pinned by tests now.

**One documented exception.** Routing-policy bodies stay hand-written in `web/src/client/local.ts`. The gateway has a real `PolicySpec`, but the request path hand-rolls a per-field 400 the policy form binds to inputs, and `PolicySpec` is also what the two write paths dump before storing, so typing it changes what lands in the database rather than only what is read back. That is a different blast radius than this change should carry, and the file says so. The remaining DoD item is therefore open, not silently claimed.

## PR Type

- [x] Refactor
- [x] Infrastructure / CI

## Relevant issues

Refs #581. Third item in epic #589.

## Checklist

On the ledger box: this refines two response schemas without adding a table, a capability, or a mode-specific surface, and the wire format is verified unchanged, so by the instructions' own wording it needs no entry. Happy to add one if you read it differently.

**Verification.** Backend 2937 tests, web 629, e2e 32, plus `make lint`, `typecheck`, `openapi-check`, `postman-check`. Integration tests were run against a real PostgreSQL rather than skipped, which is what caught the `"tools": null` regression that the unit tests missed.

This branch merges `main` rather than rebasing: #604 landed the search-tools API and added its six types to the file this branch deletes. They now come from the spec as aliases, which is the change working as intended. All three generated artifacts were regenerated from the merged source rather than left on the textual auto-merge.

## AI Usage

**AI Model/Tool used:**

Implementation and tests written by the model through back-and-forth with @njbrake; the decisions are his, including the call to fix the spec at its source rather than work around it in the client.

## Summary

- Replaced hand-written dashboard API types with types generated from the OpenAPI specification.
- Added CI checks to detect drift in the committed generated client.
- Improved billing and pricing models while preserving compatibility with legacy data.
- Updated dashboard pages and tests to use shared client types.
- Added reusable test fixtures and expanded coverage for billing, charge-line rendering, and hit-rate calculations.
- Preserved hand-written routing-policy types for custom validation and persistence requirements.

## Technical notes

- Added typed support for token, unit, tool, and provider-defined billing data.
- Preserved the existing wire format.
- Added type guards for charge lines and pricing tiers.
- Restored `DashboardBuild.version` and support for arbitrary MCP tool names.
- Updated client documentation and generation commands.
- Verified backend, web, end-to-end, lint, type, OpenAPI, and Postman checks.


tbille (approved):
Pre approve once comments from coderabbit are fixed

PR #610, opened 2026-08-17
refactor(dashboard): adopt the layered web/src architecture and boundary lint

### LINKED ISSUE #582: Adopt the target frontend architecture and boundary lint in web/
Land the destination layout and its enforcement before the bulk of code arrives, so every rehome PR lands in final shape and the boundary is checked by CI from day one, not by review.

## What to produce

- `web/src` layout: `features/<domain>/`, `shared/`, `app/` (composition root), `routes/`, `client/`.
- Biome with layer-boundary `noRestrictedImports` rules: `features/` never imports `app/`; `shared/` never imports `features/` or `app/`; nothing imports the (future) overlay tree.
- An architecture-boundaries test guarding the lint config, and the route-exports test from the router migration.
- Existing pages relocated into the layout.

## Definition of Done

- A deliberately planted boundary violation fails CI; existing pages pass lint in the new layout.

Depends on #580.

njbrake:
## Description

`web/src` was flat (`components/`, `pages/`, `api/`, `lib/`, `auth/`), which is fine while one person holds the whole thing in their head. The platform UI arrives next, so this lands the destination layout and its enforcement first, and every rehome PR after it lands in final shape.

The layout is `otari-ai/frontend/src`'s, checked against that repo rather than inferred: `features/<domain>/` owns a domain end to end (its page, the components only it renders, its derivations, their tests); `shared/` holds what no domain owns (`components/`, `helpers/`, and an `api/` that has no counterpart there, because they generate transport with hey-api and ours is hand-written); `app/` is the composition root and shell chrome; `src/tests/` holds the harnesses and sits outside the boundary, as it does there, so a harness may mount what the app mounts. `routes/` and `client/` do not move. Git recorded nearly all of it as renames; the only content changes in a moved file are its import specifiers.

**The boundary is checked, not reviewed.** Biome (`web/biome.jsonc`, `npm run lint`, a step in `otari-dashboard.yml`) enforces three rules: a feature may not import `app/`, `shared/` may not import `features/` or `app/`, and nothing may import an overlay's tree. Each group lists the alias and the relative form, since a boundary a `../..` walks around is decoration. `src/architecture.test.ts` plants a violation of each rule, runs Biome on it, and asserts the rejection, paired with the import that must keep working. Deleting a rule from the config fails that test, which is the point.

Biome is on for that and nothing else: formatter off, every other rule off. Reformatting the tree in the commit that moved it is a different change with a different risk profile.

**One thing fell out of the move rather than being chosen.** `shareCard.ts` landed beside `ShareCard.tsx` and those collide on a case-insensitive filesystem, so the derivations are now `shareCardData.ts`.

**Not adopted from over there: their Biome preset and formatter.** That is a separate PR, not a judgment that ours is better. It rewrites about 32,000 lines across 120 files, and `preset: recommended` turns up 44 real errors (19 `noSvgWithoutTitle`, 11 `noLabelWithoutControl`, 5 `useExhaustiveDependencies`, and some a11y) that deserve fixing rather than being switched off to land a refactor.

Feature-to-feature imports stay allowed (keys pick models, budgets pick users). Restricting those to per-feature entrypoints is a real question, but it is not in #582 and it would mean barrel files, which fight `autoCodeSplitting`.

## PR Type

- [x] Refactor
- [x] Infrastructure / CI

## Relevant issues

Fixes #582. Fourth item in epic #589.

## Checklist

Two boxes left unchecked deliberately. No Python and no route changed, so there is no spec to regenerate. And the ledger instructions say a refactor that moves code without touching a table, a cross-repo contract, or a duplicated capability needs no entry; happy to add one if you read it differently.

**Verification.** `npm run lint`, `typecheck`, 648 web tests (43 files), a build that still splits per route, and the 32-test Playwright suite in a real browser. `make lint` and `make typecheck` pass; the pytest suite was not run, because no Python file changed. Both directions of the guard were checked: planting `@/app/router` in a real feature file makes `npm run lint` exit 1, and removing the rule from `biome.jsonc` fails `architecture.test.ts`.

The frontend-standards skill previously read "there is no `features/`/`shared/` architecture, do not introduce those", so it is updated here along with `web/AGENTS.md`, `web/README.md`, and the review instructions. `web/README.md` also loses a line claiming the app has no router, which #580 made wrong.

## AI Usage

**AI Model/Tool used:**

Implementation, tests and the layer mapping written by the model through back-and-forth with @njbrake; the decisions are his, including the call to relocate every page in this PR rather than migrate them one at a time.


web/biome.jsonc
```diff
@@ -0,0 +1,117 @@
...
+        //
+        // Each group lists the alias form and the relative form. Imports in this
+        // codebase all use `@/`, but a boundary rule that a `../../` dodges is
+        // not a boundary rule.
+        "noRestrictedImports": {
+          "level": "error",
+          "options": {
+            "patterns": [
+              {
+                "group": ["@/overlay/**", "**/overlay/**"],
```
khaledosman: The groups miss the bare directory specifier. Every group lists `@/app/**` / `**/app/**` but not `@/app`, and I confirmed a file under `src/features/` with `import { Provider } from "@/app";` draws no diagnostic (same for `@/overlay` and, in the shared override, `@/features`).

It does not resolve today because there is no barrel, so this is latent rather than a hole in the current tree, but adding any `src/app/index.ts` later opens the boundary silently, and the comment above claims each group covers the alias form and the relative form.

Add the bare forms to each group, e.g. here `"group": ["@/overlay", "@/overlay/**", "**/overlay/**"]`, and the same shape at lines 74 and 102 (plus `"../../app"` if you want the relative directory import covered too). `architecture.test.ts` asserts one diagnostic per import statement, so this does not disturb those counts.

web/biome.jsonc
```diff
@@ -0,0 +1,117 @@
...
+      "includes": ["src/shared/**"],
+      "linter": {
+        "rules": {
+          "style": {
+            "noRestrictedImports": {
+              "level": "error",
+              "options": {
+                "patterns": [
+                  {
+                    "group": ["@/app/**", "**/app/**", "@/features/**", "**/features/**"],
```
khaledosman: `src/shared` may still import `@/routes/**`, which is the one path around this rule: a route file names a feature's page, so `shared -> routes -> features` re-enters the layer this group forbids. Nothing does it today (`shared/test/router.tsx` builds its own route tree rather than importing the generated one), which is what makes it worth closing now.

Add `"@/routes/**"`, `"**/routes/**"` to this group.

.github/skills/frontend-standards/data-fetching.md
```diff
@@ -1,7 +1,7 @@
 # Data fetching: TanStack Query
 
-All *authenticated* server state flows through TanStack Query hooks in `web/src/api/hooks.ts`,
-which call `apiFetch` from `web/src/api/client.ts`. Don't call `fetch()` directly for
+All *authenticated* server state flows through TanStack Query hooks in `web/src/shared/api/hooks.ts`,
+which call `apiFetch` from `web/src/shared/api/client.ts`. Don't call `fetch()` directly for
```
khaledosman, line 4: Line 18 of this file still reads "The query client (`web/src/provider.tsx`)", and this PR deletes that file: the retry policy is now `src/shared/api/queryClient.ts` and the provider is `src/app/provider.tsx`. Worth fixing in the same pass as these two lines, since this is agent-facing guidance that will otherwise point at a path that no longer exists.

.github/skills/frontend-standards/components.md
```diff
@@ -41,7 +41,7 @@ If a component exposes a prop for what you want (`variant`, `size`, `isDisabled`
 (`flex`, `gap-*`, `min-w-[…]`, responsive prefixes), not for re-skinning something HeroUI
 already styles.
 
-## Check `components/ui.tsx` before hand-rolling
+## Check `shared/ui/ui.tsx` before hand-rolling
```
khaledosman: Three sibling references in this file were not moved with the tree: lines 57-59 still name `components/FilterChips.tsx`, `components/Field.tsx`, and `components/DataTable.tsx`, which are now under `shared/ui/`. This heading and line 19 were updated, so the file is half-migrated.

Same class, one file over: `.github/skills/frontend-standards/design-tokens.md:36` still says `components/ui.tsx` and is not touched by this PR at all.

khaledosman (approved):
Verified in a clean worktree at `8278ee0`: `npm run lint` (0), `npm run typecheck` (0), 645 tests / 43 files, `npm run build` still splits per route, working tree clean after the probe tests (the `__boundary_probe__` ignore does cover `src/<layer>/`), lockfile additive (Biome plus its platform binaries only). Every moved file's diff is symmetric add/remove, so the import-specifier-only claim holds. I also probed the guard beyond what `architecture.test.ts` covers: `export * from "@/app/router"`, `await import("@/app/router")`, and `"./../../app/router"` are all rejected from `src/features/`.

Nothing blocking. Four findings inline, plus one the diff cannot anchor:

`.dockerignore:13` still says `web/src/pages/DocsPage.tsx`. The byte-identical comment in `otari-docker.yml` and `otari-dashboard.yml` was updated in this PR; this third copy was missed.

---

PR #611, opened 2026-08-17
style(dashboard): adopt otari-ai/frontend's Biome formatter and rule set

njbrake:
## Description

`web/` had no formatter and one lint rule (the layer boundary from #582). `otari-ai/frontend` has both, and its control-plane UI moves into this repo at M5, so one of the two trees had to adopt the other's style. Doing it now costs 119 files. Doing it after the ports costs 500+ mixed ones, inside the PRs where a whole-file reformat would bury the changes that can actually break something.

Two commits, deliberately separable.

**The reformat** is `biome format --write` plus the organizeImports assist, no hand edits: two-space, double quotes, semicolons as needed. Read `biome.jsonc` and trust the tool; nothing else in that commit rewards line-by-line reading. `.git-blame-ignore-revs` is added with that commit in it, so `git blame` still points at whoever last changed a line rather than at the reformat (`git config blame.ignoreRevsFile .git-blame-ignore-revs` locally; GitHub honors it automatically).

**The fixes** are the 53 findings `preset: recommended` reported. About seven were real: a pure function redefined every render, a ref-only factory that belonged in a `useCallback`, three `forEach` callbacks returning values, a button with no `type`, and a badge whose `aria-label` a generic span would not expose. The rest are the rule set meeting patterns it cannot see through, and each suppression says which. The largest group, 19 `noSvgWithoutTitle`, are icons that were already `aria-hidden`: Biome does not recognize the bare JSX shorthand when the element has children, and AppShell's own test asserts the attribute is present.

**I was wrong about this PR's value before writing it.** I quoted "44 real errors" from category names without opening them. The accurate case is not that the rule set excavates defects from old code, it is that it catches them as they are written: while writing these suppressions I put three `//` comments inside JSX children, which React renders as text, and `noCommentText` caught it before it could ship an Activity table reading `// A generic span does not reliably expose aria-label` next to every model name.

`npm run lint` is now `biome check`, which reports; `npm run lint:fix` writes. CI runs the reporting one, so formatting drift fails there instead of arriving in the next person's diff.

## PR Type

- [x] Refactor
- [x] Infrastructure / CI

## Relevant issues

Follows #582 / #610. No issue of its own; raise one if you would rather this waited.

## Checklist

Both unchecked deliberately: no Python and no route changed, and formatting is not a table, a contract, or a capability. No test was added, because the change is formatting plus fixes to existing code; the existing 649 tests are what proves it inert.

**Verification.** `npm run lint`, `typecheck`, 649 tests before and after the reformat, and a build. `make lint` and `make typecheck` pass; the pytest suite was not run, because no Python changed.

**One e2e failure, pre-existing and unrelated.** `parity.management.spec.ts` fails intermittently on main: it asserts `getByRole("button", {name: "Save changes"})` is hidden to mean "the dialog closed", but that button reads "Saving…" while the request is in flight, so the assertion passes immediately, the test reloads, and the reload aborts the PATCH (`status: -1` in the trace). The app is behaving correctly. Worth a one-line fix in its own PR; happy to open it.

## AI Usage

**AI Model/Tool used:**

The reformat is a tool's output. The fixes and suppressions were written by the model through back-and-forth with @njbrake; the decision to adopt otari-ai's config now rather than after the M5 ports is his.


web/src/architecture.test.ts
```diff
@@ -154,58 +201,78 @@ describe("layer boundaries", () => {
...
 describe("the boundary check is wired up", () => {
   it("is a package script", () => {
     const pkg = JSON.parse(readFileSync(join(WEB, "package.json"), "utf8")) as {
-      scripts: Record<string, string>;
-    };
-    expect(pkg.scripts.lint).toContain("biome lint");
-  });
+      scripts: Record<string, string>
+    }
+    expect(pkg.scripts.lint).toContain("biome lint")
```
khaledosman: This assertion still expects the old script name, but `06e0c6cb` changed `package.json` to `"lint": "biome check"`, so `npm test` fails and the dashboard CI job with it:

```
Test Files  1 failed | 42 passed (43)
      Tests  1 failed | 648 passed (649)
Expected: "biome lint"  Received: "biome check"
```

```suggestion
    expect(pkg.scripts.lint).toContain("biome check")
```

The PR body's "649 tests before and after the reformat" is accurate for `d3d0272a`, where the script was still `biome lint`; the regression is in the second commit, which the checklist's `make test` does not cover.

web/biome.jsonc
```diff
@@ -27,8 +31,12 @@
   "linter": {
     "enabled": true,
     "rules": {
-      "preset": "none",
+      "preset": "recommended",
+      // Off over there, so off here: the configs are meant to stay comparable, and
+      // this tree has 172 non-null assertions, which is its own cleanup.
+      "suspicious": { "noExplicitAny": "off", "noArrayIndexKey": "off" },
```
khaledosman, line 47: `noArrayIndexKey: "off"` is the one item in this group that gives up a React correctness rule rather than a style preference, and it is load-bearing here. Five findings, four of them the same real pattern in `RoutingPage.tsx` (743, 809, 963, 1033): removable row lists keyed by index.

```jsx
{candidates.map((entry, index) => (
  <div key={index} ...>
    <ModelComboBox label={`Model ${index + 1}`} value={entry} ... />
```

with the Remove button doing `prev.filter((_, i) => i !== index)`. Delete the middle row of three and React reuses instance 1 for what was row 2, so `ModelComboBox`'s internal state (open, highlighted option, typed filter text) and focus follow the slot instead of the model. The controlled `value` is fine; everything the child owns is not. The fifth (`charts.tsx:97`) is a derived tooltip list with no child state and is benign.

The otari-ai parity argument holds for `noExplicitAny` and `noNonNullAssertion` (both style-adjacent, both with a stated cleanup cost). This one is cheap to keep on: key the four lists by a stable id, or leave the rule enabled with four per-line suppressions so the next list added does not inherit the pattern silently.

.git-blame-ignore-revs
```diff
@@ -0,0 +1,13 @@
...
+#
+# Enable locally once:
+#   git config blame.ignoreRevsFile .git-blame-ignore-revs
+# GitHub honors this file automatically.
+#
+# Only add a commit that is provably formatting-only: produced by a formatter with
+# no hand edits, and green on the same tests before and after.
+
+# style(dashboard): adopt otari-ai/frontend's Biome formatter and rule set
+d3d0272a6824b2a4f1ada8187d2d4105e05106a7
```
khaledosman: This SHA will not exist on `main`. Otari squash-merges (AGENTS.md, and the comment at the top of `otari-pr-title.yml`), so `d3d0272a` never lands and the entry is inert the moment it merges, which defeats the file's whole purpose on its first use.

The squashed commit would also contain the fixes commit, so it would not qualify under this file's own rule ("provably formatting-only").

Two ways out: merge this PR with a merge commit instead of a squash so both commits survive with these SHAs, or land the file with the header only and add the real post-merge SHA in a one-line follow-up.

web/src/features/overview/OverviewPage.tsx
```diff
@@ -38,48 +55,54 @@ export function localDayKey(): string {
...
-    };
-  }, []);
+      document.removeEventListener("visibilitychange", check)
+      window.removeEventListener("focus", check)
+    }
+  }, [])
+  // dayKey drives the refresh: same day -> same bounds; new day -> re-derive.
+  // Nothing inside reads it, so the rule sees it as surplus; it is the whole
+  // point of the memo.
+  // biome-ignore lint/correctness/useExhaustiveDependencies: dayKey is the cache key, not an input
```
khaledosman, line 76: Biome is wrong about the dependency but right that something is off: this `useMemo` body is impure (it reads `Date.now()`), and the code depends on it being called exactly once per `dayKey`. React treats `useMemo` as a cache it may discard, so a dropped entry re-derives new bounds mid-day, which is precisely the query-key churn and shifted vs-prev baseline the comment above says it exists to prevent. React Compiler tightens this further, since it assumes memo bodies are pure.

The fix is to make the state the source of truth instead of the memo, computing the bounds in the same update that moves `dayKey`:

```ts
const [win, setWin] = useState(computeWindows)
// in `check`, when the day actually changed:
setWin(computeWindows())
```

Then no suppression is needed and the bounds cannot move without the day moving. Pre-existing, not introduced here, so a follow-up is fine; flagging it because this PR is the commit that re-states the suppression.

.github/workflows/otari-dashboard.yml
```diff
@@ -60,11 +60,12 @@ jobs:
...
-    # Biome, enabled for the layer boundary and nothing else (web/biome.jsonc): a
-    # feature may not import the composition root, shared/ may not import either, and
-    # nothing may import an overlay's tree. Before this step the layout was a
-    # convention a reviewer had to notice; now a violation cannot merge.
-    - name: Lint layer boundaries
+    # Biome (web/biome.jsonc): formatting, the recommended rule set, and the layer
+    # boundary (a feature may not import the composition root; shared/ may not import
+    # either; nothing may import an overlay's tree). `npm run lint` is `biome check`,
+    # which reports rather than rewrites, so a formatting drift fails here instead of
+    # landing as noise in the next person's diff. `npm run lint:fix` is the writing one.
```
khaledosman, line 67: This claim, and "Biome also formats `web/`" in AGENTS.md, are both wider than the config. `biome.jsonc`'s `files.includes` is `["src/**", ...]`, so the check covers 120 files, all under `src/`:

```
$ npx biome check e2e vite.config.ts playwright.config.ts
× No files were processed ... These paths were provided but ignored
```

So `web/e2e/**`, `vite.config.ts` and `playwright.config.ts` get neither formatting nor lint, and `vite.config.ts` still uses semicolons, which leaves two styles in the tree. This PR even edits that file (dropping the dead `eslint-disable no-console`) without formatting it. Either widen `includes` to cover them, or scope the sentence here and in `web/AGENTS.md` to `src/`.

web/src/features/models/ModelsPage.tsx
```diff
@@ -1,7 +1,22 @@
-import { Button, Card, Chip } from "@heroui/react";
-import { type ReactNode, useCallback, useEffect, useId, useMemo, useState } from "react";
-import type { Selection, SortDescriptor } from "react-aria-components";
-
+// biome-ignore-all lint/a11y/noLabelWithoutControl: every label here wraps MoneyInput, which renders the input and takes its own ariaLabel. The rule cannot see through a component, and it still works on this file's native inputs.
```
khaledosman, line 1: `biome-ignore-all` is broader than the problem it solves. The 11 labels in this file are all fine today (10 wrap `MoneyInput`, one wraps a native input, which is why the rule leaves it alone), but a file-level suppression over ~1200 lines means the twelfth label, with a genuinely missing control, lands silently.

Ten per-line `biome-ignore`s keep the rule live for new code. Better, if `MoneyInput` took an `id` and forwarded it to its input, the association would be explicit and the rule would pass on its own.

web/AGENTS.md
```diff
@@ -11,7 +11,8 @@ tokens, TanStack Query, Vitest), read
...
+  - Biome also formats and runs the recommended rule set, both taken from `otari-ai/frontend`: two-space, double quotes, semicolons as needed, organized imports. `npm run lint` is `biome check`, which reports; `npm run lint:fix` is the one that writes. Do not reformat by hand or with another tool.
+  - Where a recommended rule is suppressed, the suppression carries its reason, and most of them exist because the rule cannot see through a custom component or a conditional role: `noLabelWithoutControl` is off for the whole of `ModelsPage.tsx` (every label there wraps `MoneyInput`, which renders the input), and `useExhaustiveDependencies` is suppressed where the real dependency is the rendered DOM (the share card) or a deliberate re-run trigger (the detail-row host in `DataTable`). Before adding one, check the rule is actually wrong; three of the findings this rule set first reported were real.
```
khaledosman: Three descriptions of `npm run lint` were updated in this PR; line 30 of this file was missed and still reads:

> Checks: `npm --prefix web run lint` (layer boundaries), ...

which is now the narrower half of what the script does.

khaledosman (approved):
Approving so this is not held up, but the first inline finding fails `npm test` on the head commit and the dashboard job will block the merge until it lands. Verified in a clean worktree at `06e0c6cb`: `npm ci`, `npm run lint` (clean, 120 files), `npm run typecheck` (clean), `npm test` (**1 failed | 648 passed**), `npm run build` (clean, and the tree stays clean afterward, so `routeTree.gen.ts` and `client/schema.ts` are in sync).

The reformat itself I read as advertised and did not review line by line. One claim I checked rather than took on faith: reverting a single `aria-hidden="true"` back to the shorthand does re-trigger `noSvgWithoutTitle`, so the 19-file rewrite is not churn.

On the React side, since this PR is choosing the rule set: the `react` domain is auto-detected from the `react` dependency, so `useHookAtTopLevel` (rules-of-hooks) and `useExhaustiveDependencies` are live. I probed that directly. Two react-domain rules are not in `recommended` and are worth a follow-up: `noNestedComponentDefinitions` reports **0** findings on `src/`, so it is free to turn on, and `useUniqueElementIds` reports 9, most of which look like deliberate singleton ids in the shell (`app-sidebar` paired with `aria-controls`) rather than bugs.

React Compiler is not enabled: `vite.config.ts` calls `react()` with no Babel plugins, on React 19.2.7. Worth its own PR rather than this one, with one ordering note: the compiler expects hook rules to hold, so the two `useExhaustiveDependencies` suppressions here (`ShareDialog`, `DataTable`) and the `useMemo` flagged below are the spots to settle first. Biome has no react-compiler equivalent, so the compiler's own lint would be additive rather than redundant.

PR #616, opened 2026-08-17
feat(reprise): ship gateway observations to the platform on a bounded queue

dpoulopoulos:
## Description

Adds the transport for Reprise v0 observation records (one per tool-loop round, one per request). A hybrid-mode gateway has no local database, so there is nowhere on the box to keep them, and they are produced mid-request: an inline POST would compete for the event loop and connection pool with the request it is measuring.

So: a bounded in-memory queue plus a periodic background flush, batching to `POST {platform}/gateway/loop-observations` (served by mozilla-ai/otari-ai#1487). Neither existing mechanism fits. `BatchLogWriter` has the right shape but an unbounded queue; `_report_platform_usage` retries per request, which is right for billing data and wrong for disposable measurement data.

Nothing is wired to it yet. The producers land in mozilla-ai/otari-ai#1484 and mozilla-ai/otari-ai#1485, so this is transport plus `/metrics` only.

- **`emit` is a plain function, not a coroutine**, so a producer physically cannot reach the network through it. Its body is a `put_nowait`; a full queue drops the arriving record and counts it rather than applying backpressure.
- **Every failure loses records, never a request.** Timeout, 5xx, 404 from a peer without the endpoint, 401: each drops the batch, counts it, and leaves the flusher alive. Nothing is retried, since a retry would compete with the traffic being measured.
- **The drain empties the queue in `max_batch` chunks per tick**, so a backlog clears at the platform's pace, not one batch per interval.
- **Shutdown drains under a 5s total budget.** Unbounded it is worst-case `(queue / batch)` POSTs long, outlasting any stop grace and blocking uvicorn's lifespan behind an observability stream. The remainder is counted `dropped_shutdown`.
- **Loss is counted per cause**, since an unknown loss rate cannot support a measurement conclusion: `gateway_observation_queue_depth`, plus `gateway_observation_records_total{result=...}` over `queued` / `shipped` / `dropped_queue_full` / `dropped_flush_failed` / `dropped_shutdown`, and batch-size and flush-duration histograms.
- **`NullObservationShipper` in standalone mode**, so producers emit unconditionally; the lifespan owns one shipper per worker on `app.state`.
- The shipper builds its own platform URL and pooled client instead of reusing `_platform_url` / `_post_platform`, which live in the API layer that `check_architecture.py` forbids a service to import. That duplicates a two-token expression; the tidy-up would move the helper into `core/`.

Four knobs size the queue (10000 records / 500 per batch / 5s interval / 5s timeout), validated positive at load: every zero-value leaves the stream silently empty, which reads as "nothing was observable" rather than as a misconfiguration.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [x] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Closes mozilla-ai/otari-ai#1483. Issues are fully qualified throughout: a bare `#1483` here would resolve to an unrelated issue in this repo.

Follow-ups: mozilla-ai/otari-ai#1711 (the M4 ledger row, see below) and mozilla-ai/otari-ai#1712 (an operator off switch; the `PLATFORM_OBSERVATION_*` knobs only size the queue, and #1483 was scoped to transport).

## Checklist

- [x] If the API contract changed, I regenerated the OpenAPI spec. No route changed; `make openapi-check` and `make postman-check` confirm no drift.

**Unchecked deliberately:** this does add a control-plane contract, so the row is owed. The append is a comment on an existing issue, which the session that wrote this branch could not make, so it is tracked as mozilla-ai/otari-ai#1711 rather than silently skipped.

## Verification

On this exact tree: unit suite 1817 passed / 7 skipped; `make lint`, `make typecheck` (`mypy --strict`), `make openapi-check`, `make postman-check`, and the OSS-edition smoke gate all pass. Pre-rebase: integration suite 1199 passed against real PostgreSQL, and a wheel build. The rebase pulled in only dashboard and CI commits with no file overlap, so CI is the confirmation for those two.

Beyond tests, the app ran in hybrid mode against a stub platform on a real socket: emits queue with nothing posted and `/metrics` showing the depth; the flush posts one batch carrying `X-Gateway-Token`; the bound drops exactly the records past it and the backlog drains in batches of at most `max_batch`; shutdown flushes the tail; against a dead platform records count as `dropped_flush_failed` while `/health` still answers 200.

No browser verification applies: gateway-internal transport plus `/metrics`, no dashboard code touched.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

Verified against a real gateway process and a stub platform, not on test output alone. The five commits are split so each is independently green on lint, typecheck, and tests.

**NOTE:**

## Summary

Adds hybrid-mode transport for gateway observation records.

- Queues records in memory and sends them in bounded batches.
- Handles queue overflow and transport failures without affecting requests.
- Drains queued records during shutdown within a five-second limit.
- Adds observation metrics for queue depth, outcomes, batch size, and flush duration.
- Adds configuration validation and environment-variable support.
- Uses `NullObservationShipper` in standalone mode.
- Documents the ingest endpoint, payloads, failure behavior, and configuration.
- Adds comprehensive unit and integration tests.

Observation producers are not included in this change.

### DIFF
--- a/docs/hybrid-mode-protocol.md
+++ b/docs/hybrid-mode-protocol.md
@@ -18,6 +18,7 @@ Otari calls these endpoints, all rooted at the configured platform base URL:
 | `POST {base}/gateway/usage`                 | Report the outcome of an attempt back to the platform |
 | `POST {base}/gateway/mcp-servers/resolve`   | Swap workspace-scoped MCP server ids for inline server configs (called only when a request references MCP server ids) |
 | `POST {base}/gateway/web-search/resolve`    | Resolve the workspace's web-search policy (called only when a request uses the `otari_web_search` tool) |
+| `POST {base}/gateway/loop-observations`     | Report a batch of observation records (optional for a peer; see [Observation ingest](#observation-ingest)) |
 
 `{base}` means Otari platform `base_url` setting. Otari concatenates literally. The peer service is responsible for including any API-version prefix it exposes its own routes under. For the reference otari deployment that prefix is `/api/v1`, so the base URL is `http://backend:8000/api/v1` and Otari ends up POSTing to `http://backend:8000/api/v1/gateway/provider-keys/resolve`.
 
@@ -27,8 +28,11 @@ Every endpoint requires `X-Gateway-Token: <gw_...>` in the request headers. This
 proves the caller is an Otari instance configured against this platform
 deployment. The three resolve endpoints additionally require `X-User-Token:
 <tk_...>`, which is the workspace API token forwarded opaquely from the end
-user's `Authorization: Bearer ...` header. The usage endpoint sends only the
-gateway token.
+user's `Authorization: Bearer ...` header. The usage and observation endpoints
+send only the gateway token. Usage reports carry a `correlation_id` the platform
+issued at resolve time, which already identifies the caller, and an observation
+batch is not sent on behalf of any one caller: it mixes records from every
+request that was in flight when the flush timer fired.
 
 ## Extension policy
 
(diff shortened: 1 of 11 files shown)


njbrake:
Hi @dpoulopoulos ! Is this change a part of the oss refactor work, or is it ok if we pause the progress of this one while we finish up the migration?

tests/unit/test_platform_observation_config.py
```diff
@@ -0,0 +1,91 @@
...
+        GatewayConfig(platform={key: bad_value})
+
+
+@pytest.mark.parametrize("key", ["observation_max_queue", "observation_max_batch"])
+@pytest.mark.parametrize("bad_value", [0.5, 0.9, 250.5])
+def test_a_fractional_record_count_is_rejected(key: str, bad_value: float) -> None:
+    """The shipper sizes a queue and a batch with these, via ``int()``.
+
+    A fraction below 1 truncates to 0, and ``asyncio.Queue(maxsize=0)`` is
+    unbounded, so a bound of 0.5 would accept records without limit: the one
```
aittalam, line 61: would a bound of 0.5 be accepted at all, given we define max_queue as int and break if the validation does not succeed?

src/gateway/main.py
```diff
@@ -301,13 +302,23 @@ async def lifespan(app: FastAPI) -> AsyncGenerator[None, None]:
             # Same shape for the models.dev catalog, whose fetch is bounded at 15s.
             catalog_refresher = asyncio.create_task(run_catalog_refresher(config))
 
-        # Start the writer inside the try so a failure here still runs the cleanup
-        # below; the refresher tasks are already created and would otherwise leak.
+        # Hybrid mode reports observation records to the platform; standalone has
+        # nowhere to ship them, and gets a null shipper so producers can emit
+        # unconditionally. Building one contacts nothing.
+        observation_shipper = create_observation_shipper(config)
```
aittalam, line 308: Would there be any value in having a non-null local observation shipper, e.g. something that allows us to keep records of observations without the need to run a full platform? Not a request for this PR (or even this epic!) but I wanted to get a sense whether it's something we might want to have in the future

docs/hybrid-mode-protocol.md
```diff
@@ -414,6 +418,73 @@ retries on transient failures (timeout, network error, 5xx) up to
...
+with records lost and the request completing exactly as it would have:
+
+| What happens | What Otari does |
+|---|---|
+| Timeout, connection error, unreachable peer | Drops the batch, counts it |
+| `5xx` | Drops the batch, counts it |
+| `404` (peer does not implement this endpoint) | Drops the batch, counts it |
+| `401` (peer rejects the gateway token) | Drops the batch, counts it |
+
+Nothing is retried. Observation records are disposable, unlike usage reports,
```
aittalam, line 465: A clarification here: by disposable we basically mean that we prefer to lose some observations rather than impacting the gateway/platform performance, right? But what happens when records are dropped? My assumption is that the whole set of observations from that execution has to be thrown away, as we only have a partial observation. Is that correct? I think it's ok as now we are just measuring this kind of impact, just wanted to get a sense of how much we lose when observations are dropped (e.g. we might want to make our data collection pipeline more forgiving, at the cost of worse performance, to make sure we do not lose observations)

aittalam (approved):
TY @dpoulopoulos ! The code itself looks ok, I just left a few non-blocking questions for clarification.

dpoulopoulos:
@njbrake no, it's part of reprise. We can pause, but please ping me when we're ready to start working on it again.

docs/hybrid-mode-protocol.md
dpoulopoulos: We drop the observation and we count it as dropped in our metrics. But we don't have to drop the whole set. Every round can be an independent one, since it takes as input how deep is in the round calls, which tools have been called before, etc.

src/gateway/main.py
dpoulopoulos: Yes, I'll create an issue for that to track it separately.

Edit: https://github.com/mozilla-ai/otari/issues/635

PR #624, opened 2026-08-17
feat(tenancy): land the reconciled tenancy schema on an async repository base

njbrake:
## Description

Lands the persistence shape the rest of the M5 backend rehome inherits, per [otari-ai#1715](https://github.com/mozilla-ai/otari-ai/issues/1715) option A: SQLModel retained, `BaseRepository` takes an `AsyncSession`, converted on arrival. First of two PRs; the services and routes follow on top of this branch.

**What's here**

- `repositories/base_repository.py`: the async CRUD base, with the platform's flush-not-commit contract preserved (repositories stage, services own the commit boundary).
- One new dependency, `sqlmodel`, which brings no new transitive packages but does cap SQLAlchemy at `<2.1.0` where this repo's own constraint is unbounded. That cap is a consequence of the option A decision and is worth recording with it.
- `models/tenancy.py`: `user`, `organization`, `organization_member`, `workspace`, `workspace_member`, with their `Create`/`Update`/`Public` schemas. Those schemas are the endpoint contracts the dashboard client is generated from, which is why SQLModel is kept rather than converted to `entities.py`'s style.
- One hand-written Alembic revision appended to the linear chain.
- `entities.py`'s `Base` now shares `SQLModel.metadata`, and `models/__init__.py` imports every model module, so Alembic, `create_all` and `drop_all` each see one whole schema instead of whichever half the caller imported.
- Tenancy repositories in `repositories/tenancy/`, plus 30 tests.

**Adapted rather than copied**

- Timestamps are timezone-aware. The platform's mixins annotate a bare `datetime` while their `default_factory` writes an aware `datetime.now(UTC)` into it, so the offset was being dropped on the way in. That is a latent bug, not a style difference.
- `email` is a plain nullable string, not `EmailStr`: a standalone operator identity is a label, not a sign-in address (M4's "local identities have no email").
- `organization.created_by_user_id` carries `use_alter`, because it closes a foreign-key cycle with `user.active_organization_id`. Without it `drop_all` raises `CircularDependencyError`, which the integration fixtures' teardown runs.
- Hosted-only columns are left behind, apart from `workspace.activation_classification`, which stays because the reconciled schema is edition-invariant (the overlay contributes adapters and routers, never tables).

**Migration verification.** CI exercises the SQLite path only through this PR's own `tests/unit/test_tenancy_schema_chain.py`; the integration suite migrates PostgreSQL and the OSS-edition gate runs against PostgreSQL too, so the dual-engine requirement in [otari-ai#1644](https://github.com/mozilla-ai/otari-ai/issues/1644) has no other coverage. The revision handles the same cycle: `organization` is created without its reference to `user`, `user` is created with its reference to `organization`, and the remaining constraint is added through `batch_alter_table` with an explicit `copy_from`. PostgreSQL takes that as a plain `ALTER`; SQLite rebuilds the table. Both engines were verified by hand:

- SQLite: upgrade to head, downgrade, re-upgrade, downgrade to base, with nothing left behind. The named unique constraint and both halves of the cycle survive the rebuild, now pinned by `tests/unit/test_tenancy_schema_chain.py`.
- PostgreSQL 18: the same round trip, and `compare_metadata` reports no tenancy drift between the migration and the models.

Unit 1785 passed, integration 1222 passed on this branch, OSS-edition smoke passed, `make lint` and `make typecheck` clean.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Part of [mozilla-ai/otari-ai#1452](https://github.com/mozilla-ai/otari-ai/issues/1452). Assumes [otari-ai#1715](https://github.com/mozilla-ai/otari-ai/issues/1715) option A, which is still open and whose charter Decisions entry is owed.

## Checklist

- [x] If the API contract changed, I regenerated the OpenAPI spec (`uv run python scripts/generate_openapi.py`). No API change in this PR; the spec and Postman checks confirm no drift.
- [ ] If this PR adds or changes a persistent table, a control-plane contract the otari.ai platform consumes or serves, a capability the platform also has, or a mode-specific surface, I appended an entry to the M4 reconciliation ledger ([otari-ai#1587](https://github.com/mozilla-ai/otari-ai/issues/1587)). See [.github/instructions/reconciliation-ledger.instructions.md](instructions/reconciliation-ledger.instructions.md).

The ledger entry applies (this adds five persistent tables) and is drafted, covering the tables, the three deltas from the platform's schema, and the persistence-stack decision. It is not appended yet; it lands once both PRs of this pair have numbers to reference.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

The scope, the two assumed decisions, and the cut lines are njbrake's; the code and prose are the model's. The migration was verified against both engines locally, including a PostgreSQL instance installed for the purpose, rather than assumed from the SQLite run.

## Summary

- Added tenancy persistence for users, organizations, workspaces, and memberships.
- Added asynchronous repositories for tenancy data access.
- Added shared metadata and migrations for SQLite and PostgreSQL.
- Added schema, migration, timestamp, and repository tests.
- Updated backend guidance and added the SQLModel dependency.

## Benefits

- Provides a stable foundation for future tenancy services and routes.
- Preserves service-level transaction control.
- Supports reliable schema management and consistent pagination.
- Normalizes timestamps to UTC across supported databases.

## Technical notes

- `BaseRepository` uses `AsyncSession` and flushes changes without committing.
- Tenancy models define validation, relationships, indexes, and uniqueness rules.
- Shared metadata supports Alembic, `create_all`, and `drop_all`.
- Foreign-key cycle handling supports both SQLite and PostgreSQL.
- Local identity creation sets `default_organization_id`.


src/gateway/models/tenancy.py
```diff
@@ -0,0 +1,600 @@
...
+    name: str = Field(min_length=1, max_length=255)
+
+
+class OrganizationMemberBase(SQLModel):
+    organization_id: uuid.UUID = Field(foreign_key="organization.id", ondelete="CASCADE", index=True)
+    user_id: uuid.UUID = Field(foreign_key="user.id", ondelete="CASCADE", index=True)
+    role: str = Field(default="member", max_length=32)
+    status: str = Field(default="active", max_length=32)
+
+    @field_validator("role")
```
khaledosman, line 349: These `role`/`status` validators never run for a stored row. `OrganizationMember` is `table=True`, and SQLModel skips construction validation on table classes, so they fire only when a caller happens to route through the non-table `OrganizationMemberCreate`/`OrganizationMemberUpdate`. `base_repository.py:74` builds the table instance directly and `:80` `setattr`s from a raw dict, so both generic paths bypass them.

The only thing keeping `role="superadmin"` out of the column is which request model the caller picked, which #625 then had to discover twice (unknown workspace role, null workspace name). `workspace.activation_classification` gets a real `CheckConstraint` two hundred lines down; give the four role/status columns the same, so the invariant lives where it cannot be routed around.

alembic/versions/c4b6d8e0f2a3_add_tenancy_tables.py
```diff
@@ -0,0 +1,236 @@
...
+        "workspace",
+        sa.Column("id", sa.Uuid(), nullable=False),
+        sa.Column("organization_id", sa.Uuid(), nullable=False),
+        sa.Column("name", sa.String(length=255), nullable=False),
+        sa.Column("description", sa.String(length=1024), nullable=True),
+        sa.Column("created_by_user_id", sa.Uuid(), nullable=True),
+        # Edition-invariant schema: the OSS control plane never writes this, but
+        # the overlay contributes no tables of its own, so the column the hosted
+        # activation surface reads has to exist in the one schema both boot.
+        sa.Column("activation_classification", sa.String(length=32), nullable=False),
```
khaledosman, line 175: NOT NULL with no `server_default`, so only a Python-side insert can fill it. #633's migration already hard-codes `'eligible'` in its raw `INSERT INTO workspace` for exactly this reason.

Add `server_default="eligible"` here and on `Workspace.activation_classification`, so a raw-SQL insert or a later backfill does not have to know the default.

khaledosman (commented):
Review of the tenancy schema and async repository base. Three points, one of them a real gap in what keeps a bad role out of the table.

src/gateway/models/tenancy.py
```diff
@@ -0,0 +1,600 @@
...
+from sqlalchemy.engine.interfaces import Dialect
+from sqlalchemy.types import TypeDecorator
+from sqlmodel import Field, SQLModel
+
+ORGANIZATION_MEMBER_ROLES = {"owner", "admin", "member", "viewer"}
+ORGANIZATION_MEMBER_STATUSES = {"active", "invited", "suspended"}
+WORKSPACE_MEMBER_ROLES = {"owner", "admin", "member", "viewer"}
+WORKSPACE_MEMBER_STATUSES = {"active", "invited", "suspended"}
+
+WorkspaceActivationClassification = Literal["eligible", "internal", "automated", "migrated", "enterprise_assisted"]
```
khaledosman, line 70: `WorkspaceActivationClassification` is defined and exported but unused: the column at `:504` is a plain `str` and the same five values are restated as a SQL string at `:487`. Build the check constraint from `get_args(WorkspaceActivationClassification)` so the vocabulary has one source, or drop the alias.

njbrake:
Merging on my own review plus the green checks. tbille advised over Slack that this is good to go.

Reviewed locally against PostgreSQL 18: no new `compare_metadata` drift, a clean upgrade/downgrade/upgrade round trip on both engines, and the full unit and integration suites passing.

Services and routes follow in the stack. The M4 reconciliation ledger entry lands with them.

Issue #588, opened 2026-08-14
Port platform tenancy workflows (orgs, workspaces, members, roles, budget policies)


njbrake:
Port the platform's organization, workspace, membership, invitation, fixed-role, and budget-policy pages into `web/`, registered through the nav registry and gated by the capability bootstrap. Standalone boots one default organization holding multiple workspaces; the same pages serve it.

## What to produce

- The feature modules, built against the generated client and mocked at that boundary for tests.
- Pages merge dark (capability-gated) until the corresponding backend endpoints land in this repo, then light up by flipping the capability; no separate integration branch.

## Definition of Done

- With the tenancy backend present, the workflows pass e2e in standalone; without it, the OSS build boots and smokes with the pages hidden and no dead links.

Depends on #581, #584, #585, and the backend tenancy slice of mozilla-ai/otari-ai#1452.

Prose by Claude Fable 5 via back-and-forth with @njbrake; the decisions are his.

njbrake:
The "backend tenancy slice of mozilla-ai/otari-ai#1452" this depends on now resolves to two decide-issues first: mozilla-ai/otari-ai#1715 (persistence stack for the rehomed control plane) and mozilla-ai/otari-ai#1716 (standalone identity and first boot).

The implementation sub-issues get written once those land. #1715 matters to this issue directly: it decides whether the platform's Pydantic schema layer ports verbatim or gets rewritten, and that layer is what `docs/public/openapi.json` and therefore this dashboard's generated client are built from.

_Note: this comment was drafted by Claude Opus 5 via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

njbrake:
Closing. The Definition of Done here is met: #626 landed `OrganizationGeneralPage`, `OrganizationMembersPage`, `WorkspacesPage`, `WorkspaceMembersPage`, `SettingsPage` and `UsersPage`, each registered with a `surface` and gated through `useNavVisibility`, against the organization and workspace member CRUD from #625.

Three items named in the title were never in the Definition of Done and are still real work. Splitting them lets each run on its own dependency instead of queueing behind the others:

- #641, invitations. Blocked on the auth track.
- #642, workspace per-member budget defaults. Blocked on nothing, so it can start now.
- #643, organization-scoped provider keys. Blocked on a keying decision.

_Note: this comment was drafted by Claude Opus 5 via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

PR #625, opened 2026-08-17
feat(tenancy): serve organizations, workspaces, and memberships

njbrake:
## Description

**Stacked on #624; review that first.** The base branch is `feat/tenancy-async-foundation`, so this diff shows only the routes and services.

The 15 endpoints under `/v1/organizations` and `/v1/workspaces`, rehomed from the platform's organization, membership and workspace services and converted to async. Response shapes stay the platform's, because they are what [otari#588](https://github.com/mozilla-ai/otari/issues/588)'s tenancy pages will be generated from, so `docs/public/openapi.json`, the Postman collection and `web/src/client/schema.ts` are regenerated here and that work can start against the committed contract.

**First boot** ([otari-ai#1716](https://github.com/mozilla-ai/otari-ai/issues/1716) option A). Standalone Otari authenticates an operator with the master key, which names no user, while every tenancy surface needs an identity with an active organization and a role. So the master key becomes the bootstrap credential: the first authenticated request provisions the default organization, its default workspace and one owner identity, and later requests resolve that identity through a `runtime_settings` marker, which survives renaming, leaving or deleting the default organization in a way a slug lookup would not. Per-session identities land with the sign-in flow that retires the master key as a login.

**One organization per deployment.** A self-hosted gateway is one tenant with several people in it, so creating, switching between and deleting organizations is not mounted: the organization is provisioned at first boot, renamable, and fixed, and workspaces are the unit teams are separated by. The charter's settled sentence ("OSS ships multi-tenant with a default org/workspace pre-provisioned, so org scoping is an OSS concept") settles that the model is tenancy-shaped and roles are org-scoped, which the code keeps; it does not settle that a self-hoster can mint a second organization. Adding those surfaces back is additive and belongs behind the entitlement gate `ARCHITECTURE.md` describes, so the overlay contributes them without this code changing. This also restores the premise of [otari-ai#1539](https://github.com/mozilla-ai/otari-ai/issues/1539).

**Deferred, each to its own slice**, rather than ported now:

- emailed invitations and pending-membership accept/decline, which need an invitation table and mail delivery. Adding a member is ported (below); being *invited* to one is not;
- organization model pricing, reconciled platform-side under [otari-ai#1639](https://github.com/mozilla-ai/otari-ai/issues/1639) and [#1663](https://github.com/mozilla-ai/otari-ai/issues/1663);
- email-domain auto-join, which has not been decided as OSS;
- the organization dashboard aggregates, which overlap the settlement reconciliation in [otari-ai#1641](https://github.com/mozilla-ai/otari-ai/issues/1641);
- teams, which port eventually but are their own vertical.

Three of those citations are closed M4 issues, which says the reconciliation is done rather than that the port into `otari` is tracked. Each still needs an issue here before this milestone closes.

Mixpanel tracking, budget-policy materialization, the Playground token anchor and guardrail provisioning are gone from the ported paths for the same reason, which is why workspace creation is a plain two-row insert here.

**One fix, and one deliberate narrowing.** An unknown workspace role is now refused with a 400, which the platform does not do: the role arrives as a query parameter and lands on a table model, SQLModel skips validation when constructing one, and the column carries no check constraint, so `?role=superadmin` is stored verbatim upstream.

Organization deletion is **not** a fix, and an earlier version of this description wrongly called it one. The platform already repoints every identity whose `active_organization_id` names the organization before deleting it, for the same NOT NULL reason. What this port changes is narrower and worth calling out: it drops the platform's preference for moving an identity to another organization it *owns* (so an identity can land somewhere it is only a viewer), and it replaces the platform's fallback of minting a personal organization with a 400 refusal, because that fallback builds the organization's name from the user's email and an OSS local identity has none.

Hybrid mode mounts none of this and answers the same "manage this via the platform UI" stub as the other management prefixes: a hybrid gateway holds no tenancy state at all.

**Adding a member, in the third commit.** Leaving `POST /me/members` out left five of these endpoints reachable but inert: nothing else could create an identity, so the roster stayed at the operator alone, both member-mutation endpoints only ever saw the last owner, and workspace add-member had nobody to name.

The platform's two branches both end at a membership that has to be accepted, one by email invitation and one as an `invited` membership row. This edition has neither an invitation to send nor a way to accept one, so a membership in that state would have no exit. Both branches therefore land `active`, which is the other arm of the platform's own result union, so the wire contract does not move; an unknown address creates a local identity carrying it, which is M4's claimable identity, and the invitation fields stay null until that flow rehomes. Re-adding someone who was removed revives their suspended membership rather than inserting beside it. Workspace assignments are applied in the same transaction rather than parked, and one naming a workspace outside the organization fails the whole request.

Two consequences worth agreeing on before this leaves draft. The member-update request no longer accepts `invited`: it stays a valid stored status because the invitation flow produces it and will rehome, but nothing here can create or accept one, so offering it advertised a state with no producer and no exit. (The M4 re-parenting backfill does not produce it either; it maps a blocked gateway user to `suspended` and every other one to `active`.) Widening it back later is additive; narrowing later would not be. And an identity added this way cannot authenticate until Otari grows a sign-in flow, so until then it is a roster and attribution entry, which is what the gateway's own `users` already are. The address is the handle its owner will claim it by, and proving ownership of it belongs to that claim flow, which is [otari-ai#1716](https://github.com/mozilla-ai/otari-ai/issues/1716)'s remaining sub-question. That flow has to verify the address rather than trust that an admin typed it, because the platform gets that verification free from the invitation email and this edition does not.

> **Reading note.** The commit-by-commit paragraphs below were written during the review rounds that produced them, and two describe surfaces the final commit then removed: organization deletion and `POST /me/switch` are no longer mounted. They are kept as the account of how the change arrived; what ships is the paragraph above ("One organization per deployment").

**Fourth commit, from a second review round.** Adding a member made the organization permanently undeletable, since the new identity's only membership was that organization and removing them suspends rather than clears it. Such an identity now goes with the organization; the caller still blocks the delete rather than deleting themselves. The narrowing above was also invisible to the client, because a validator on a plain string publishes nothing, so every request-facing role and status is a `Literal` now and the allowed values reach the spec, the generated client and the Postman collection. Email lookup became case-insensitive, since the unique index is not and a backfilled mixed-case row would otherwise mint a second identity for the same address. The last workspace can no longer be deleted, an assignment revives a suspended workspace membership rather than skipping it, the assignment list is bounded and deduplicated, and the three races that had fixes but no tests now have them.

**Review fixes, in the second commit.** A workspace name reached the column unchecked, so a `{"name": null}` update (which the generated client types as valid) answered 500 and an empty string stored a nameless workspace. Two of the platform's `IntegrityError` guards had been dropped along with its hosted depth, so the loser of a create or add-member race answered 500 where the pre-check would have answered 409. `POST /me/switch` answered 404 for an unknown organization and 403 for one the caller is not a member of, which makes the pair an existence oracle. A bootstrap marker whose value no longer resolved was permanently fatal, because the re-provision path inserted a second row on the same primary key. And `STANDALONE_SURFACES` in `api/routes/bootstrap.py`, the hand-kept roster the dashboard's nav registry gates every page on, had gained no entry, so the tenancy pages this work exists to unblock would have had no capability to flip.

**Tests.** 71 new: the endpoints end to end, plus a service-level suite for the authorization matrix. The route tests can only act as the one operator identity a standalone deployment has, and that identity is an owner and a superuser, so what a non-owner may do is only reachable at the service layer until per-identity sign-in lands.

Unit 1785 passed, integration 1293 passed, dashboard 835 passed, OSS-edition smoke passed, `make lint`, `make typecheck`, `make openapi-check`, `make postman-check`, and `tsc` plus biome on `web/` all clean.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Part of [mozilla-ai/otari-ai#1452](https://github.com/mozilla-ai/otari-ai/issues/1452). Unblocks [otari#588](https://github.com/mozilla-ai/otari/issues/588). Assumes [otari-ai#1715](https://github.com/mozilla-ai/otari-ai/issues/1715) and [otari-ai#1716](https://github.com/mozilla-ai/otari-ai/issues/1716) option A, both still open and both owed a charter Decisions entry.

## Checklist

- [x] Documentation was updated where necessary. `docs/access-control.md` gains the tenancy section and `docs/api-reference.md` the two endpoint tables.
- [x] If the API contract changed, I regenerated the OpenAPI spec (`uv run python scripts/generate_openapi.py`). Also the Postman collection, `web/src/client/schema.ts`, and the SDK endpoint manifest, where all 15 endpoints are classified as dashboard-only.
- [ ] If this PR adds or changes a persistent table, a control-plane contract the otari.ai platform consumes or serves, a capability the platform also has, or a mode-specific surface, I appended an entry to the M4 reconciliation ledger ([otari-ai#1587](https://github.com/mozilla-ai/otari-ai/issues/1587)). See [.github/instructions/reconciliation-ledger.instructions.md](instructions/reconciliation-ledger.instructions.md).

The ledger entry applies (this adds a control-plane contract the platform also serves, and a standalone-only surface) and is drafted, covering the endpoints, the schemas, and the two assumed decisions. It is not appended yet.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

The scope, the two assumed decisions, and the deferral lines above are njbrake's; the code and prose are the model's.

## Summary

- Added 15 standalone endpoints for organization, workspace, and membership management.
- Added first-use provisioning for a default organization, workspace, and owner identity.
- Added authorization, validation, membership safeguards, case-insensitive email lookup, and race-condition handling.
- Disabled these management routes in hybrid mode.
- Updated tenancy documentation, Postman requests, OpenAPI schemas, and web client types.
- Added integration tests for API behavior, authorization, provisioning, and concurrency.


src/gateway/services/tenancy/organization_service.py
```diff
@@ -0,0 +1,498 @@
...
+        # it. Unreachable while one bootstrap operator is the only identity that
+        # can authenticate, and reachable the day per-identity sign-in lands
+        # (otari-ai#1716), which makes this the cheap moment to close it. The
+        # narrowing is widenable later; the escalation would not be.
+        if actor_membership.role != "owner" and new_role == "owner":
+            raise MembershipUpdateError("Only organization owners can grant the owner role")
+
+        target_is_active_owner = target_membership.role == "owner" and target_membership.status == "active"
+        if target_is_active_owner and (new_role != "owner" or new_status != "active"):
+            if await self.members.count_active_owners(organization_id) <= 1:
```
khaledosman, line 502: Check-then-act with nothing behind it. Two concurrent `PATCH`es demoting two *different* owners both read `count_active_owners == 2`, both pass this guard, and the organization ends with zero active owners and nobody able to manage or delete it.

Unlike the three races this PR guards, there is no unique index to lose to, so the `IntegrityError` handlers never fire. `workspace_service.py:250` has the identical shape for the last workspace.

Take `SELECT ... FOR UPDATE` on the organization row before the count, or make the demote a conditional `UPDATE` whose `WHERE` re-asserts the owner count.

khaledosman (commented):
Review of the organization/workspace surface. The authorization matrix and the 404-not-403 boundary check out; the substantive finding is that the two "cannot go below one" invariants have no constraint behind them.

src/gateway/repositories/tenancy/user_repository.py
```diff
@@ -34,24 +34,36 @@ def __init__(self, db: AsyncSession):
...
+        """Return the identity with this email address, or None.
+
+        Matched case-insensitively. The unique index is not, and rows written
+        outside this service (the M4 re-parenting backfill, an operator's own
+        SQL) can carry any casing, so an exact match would answer "no identity
+        holds this address" for one that does and mint a second identity for it.
+        An address is the handle a claim flow matches on, so it has to resolve
+        to one identity however it was typed.
+        """
+        result = await self.db.execute(select(User).where(func.lower(col(User.email)) == email.strip().lower()))
```
khaledosman: The read is case-insensitive but `ix_user_email` is not, so the two disagree in both directions. `A@b.com` and `a@b.com` can both exist (the backfill and hand-written SQL are named in the docstring as writers that do not normalize), and `.scalars().first()` on a query with no `ORDER BY` then picks an arbitrary one of them, which for a claim handle is the wrong kind of nondeterminism. `func.lower()` also makes the index unusable.

Add a functional unique index on `lower(email)` so the constraint matches the lookup; failing that, at least give the query a deterministic order.

src/gateway/services/tenancy/organization_service.py
```diff
@@ -0,0 +1,498 @@
...
+        user: User,
+        organization_name: str,
+    ) -> OrganizationMembershipContextPublic:
+        """Rename the caller's organization."""
+        organization = await self.get_active_organization_for_user(user)
+        membership = await self.require_active_organization_management_access(user=user, organization=organization)
+
+        updated = await self.organizations.update_organization(
+            organization,
+            {"name": organization_name.strip() or "Organization"},
```
khaledosman: `.strip() or "Organization"` silently renames the organization to the literal `"Organization"`: `ActiveOrganizationUpdateRequest`'s `min_length=1` admits `" "`, and the fallback then applies. `workspace_service.py:115` refuses exactly this input with `WorkspaceNameRequiredError`. Raise a 400 here for the same reason, rather than storing a name the caller did not ask for.

src/gateway/services/tenancy/provisioning_service.py
```diff
@@ -0,0 +1,213 @@
...
+    organization served. An imported identity belongs to several organizations
+    and its pointer holds whichever it was last active in, so moving the marker
+    alone adopts whatever that happens to be. See "Adopting an existing tenancy"
+    in ``docs/access-control.md``.
+
+    Not a startup error either, despite reading like one: the only caller is
+    ``ensure_bootstrap_identity``, which runs from the request dependency, so
+    this surfaces as a failed tenancy request rather than a refusal to boot.
+    """
+    organizations = (await db.execute(select(Organization))).scalars().all()
```
khaledosman: `select(Organization)` with no `WHERE` and no `LIMIT`, and this runs on every request until the marker resolves. Only the non-default slugs matter, and only a handful need naming in the message: `.where(col(Organization.slug) != DEFAULT_ORGANIZATION_SLUG).limit(N)`.

src/gateway/services/tenancy/workspace_service.py
```diff
@@ -0,0 +1,346 @@
...
+        ``name`` also lands on a table model, and the column is NOT NULL with no
+        minimum length, so an explicit ``null`` reached the database as an
+        integrity error and an empty string stored a nameless workspace.
+        """
+        trimmed = (name or "").strip()
+        if not trimmed:
+            raise WorkspaceNameRequiredError
+        return trimmed
+
+    async def _require_workspace_management_access(self, *, user: User, workspace: Workspace) -> None:
```
khaledosman, line 143: This ignores `user.is_superuser`, while `_workspace_in_active_organization:81` grants a superuser read on every workspace. A superuser with no organization membership can therefore see a workspace and not manage it. Pick one rule and apply it in both places.

tbille (approved):

PR #626, opened 2026-08-18
feat(dashboard): port the tenancy pages onto the rehomed control plane

njbrake:
## Description

Ports the tenancy pages from `otari-ai/frontend`, then rebuilds the shell around them to match the ratified [navigation prototype](https://otari-nav-prototype-705104.netlify.app/navigation-prototype/home?tenancy=single&role=admin).

**Tenancy pages.** Organization (rename), Members and roles (add by address, change role, remove), and Workspaces (create, rename, delete, manage each roster). `features/organization/roles.ts` is the client half of the server's `_validate_membership_update`, with its two deliberate narrowings recorded where they are made.

**Two nav contexts, not one.** The default is a workspace, chosen from a switcher; the organization is a separate rail reached from the sidebar footer. `navContextForPath` derives the rail from the registry rather than from a URL prefix, since the split does not follow one. Members names two different pages, as the prototype has it: the workspace roster and the organization roster.

**The switcher scopes what the gateway scopes.** Members, API keys, usage, and the request log narrow to the selection; the workspace is part of each query key so switching refetches. Routing policies and aliases deliberately do not, because the gateway stores them all in the default workspace and resolves them process-wide, so filtering would hide live policies. The popover says which is which.

Also: breadcrumbs, an account menu carrying appearance and sign-out, a working light/dark theme against tokens `globals.css` already shipped, and Tools split into per-service routes.

Users, Budgets and Settings move to the organization rail. No route changed, so bookmarks still work, but the entry to that rail is emphasized and the API keys page links to both, since that is the workflow the move splits.

## PR Type

- [x] New Feature

## Relevant issues

Part of #588. Stacked on #633, which is stacked on #625.

## Checklist

The ledger entry applies and is drafted, folding into the one owed for #625 and #633 over the same slice. Not appended yet.

## Known gaps, stated rather than hidden

- **Roles are settable but not exercisable.** Standalone issues one master-key session and everyone holding it is the operator, so nothing can authenticate as a non-owner (otari-ai#1716). The role gate is written and always open; the prototype's `View only` mode is not built, because nothing could reach it.
- **Multi-organization cannot be run locally.** No gateway here reports `hosted`, and the endpoints that mint a second organization are not mounted. That trail is covered by a component test instead.
- **Provider credentials are not workspace-scoped**, by the base's decision: process-wide config rather than a workspace row.
- **MCP servers** is in the prototype's Tools group and left out, because the gateway has no MCP server registry to manage.
- **Verified by hand, since CI cannot.** Every substantive workflow is `pull_request: branches: [main]`, so a stacked PR runs only title lint, template check, and SDK codegen. Run locally: 936 Vitest across 59 files, 38 Playwright, lint, typecheck, build, plus an upgrade of a populated pre-workspace database to confirm the backfill and that first boot adopts the seeded default rather than creating a second.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

The scope, the prototype as the design authority, and the calls on what to leave out are njbrake's; the code and prose are the model's. Three findings from an independent review round are fixed here: the switcher filtered routing and aliases the gateway does not scope, its copy described the opposite of what it did, and workspace mutations did not refresh the switcher's own source.


web/src/features/workspaces/WorkspacesPage.tsx
```diff
@@ -0,0 +1,564 @@
...
+          </div>
+        ),
+      },
+    ],
+    [expanded, manages],
+  )
+
+  const renderDetail = useCallback(
+    (workspace: Workspace) => (
+      <WorkspaceMembersPanel
```
khaledosman, line 245: `WorkspaceMembersPanel` needs `key={workspace.id}`.

Clicking "Members" on row B while A is expanded only flips `expanded`, so this element stays at the same position in the tree and React reconciles instead of remounting. `AddWorkspaceMember`'s `userId` and `role` state carries over from A into B. When the carried-over id is not in B's candidate list, the `<select>` falls back to displaying "Select a member…" while state still holds the old id and `isDisabled={userId === ""}` keeps the button enabled, so pressing "Add member" posts A's selection into workspace B.

`EditWorkspaceForm` is keyed for exactly this reason (see the comment at :505); the detail panel needs the same treatment:

```tsx
<WorkspaceMembersPanel
  key={workspace.id}
  workspace={workspace}
  …
/>
```

web/src/features/organization/OrganizationMembersPage.tsx
```diff
@@ -0,0 +1,347 @@
...
+              exists without the access they were added for. Workspace roles are
+              changed afterwards on the Workspaces page.
+            </span>
+            {workspaces.data.map((workspace) => (
+              <label
+                key={workspace.id}
+                className="flex w-fit items-center gap-2 text-sm text-foreground"
+              >
+                <input
+                  type="checkbox"
```
khaledosman: Raw `<input type="checkbox">` instead of a HeroUI component. `.github/instructions/frontend-standards.instructions.md` non-negotiable #1 is "HeroUI v3 only", and `components.md`'s "Check `components/ui.tsx` before hand-rolling" section lists the sanctioned primitives; the native `<select>` is sanctioned there by name (`FilterSelect`, "token-styled native `<select>`") and a bare checkbox is not.

It also carries no styling at all, so unlike every other control on this page it renders as an unstyled browser checkbox that ignores the design foundation entirely, and its hit target is ~13px.

Use HeroUI's `Checkbox`, or promote the token-styled `SelectionCheckbox` that `shared/components/DataTable.tsx:92` already built (react-aria `Checkbox` matched to the tokens, with a comment explaining why it exists) into `components/ui.tsx` and use it here. The same substitution is owed by `KeysPage.tsx:392`, `SettingsPage.tsx:834` and `UsersPage.tsx:620`, but those predate this PR and are a separate cleanup, not a licence for a fourth.

web/src/features/organization/OrganizationMembersPage.tsx
```diff
@@ -0,0 +1,347 @@
...
+        id: "role",
+        header: "Role",
+        cell: (member) => {
+          const blocked = membershipChangeBlockedReason({
+            member,
+            context: activeContext,
+            members: rows,
+          })
+          return (
+            <span title={blocked}>
```
khaledosman, line 247: The refusal reason reaches the user only through `title` on a wrapping `<span>`, which means mouse hover and nothing else. On touch there is no way to see it at all, and a `disabled` control carries no accessible description, so keyboard and screen-reader users get a dead control with no reason.

That defeats the design this page states for itself: "a control that would be refused is disabled with the reason rather than erroring on click". The reason is only reachable by the subset of users on a pointer device.

Render `blocked` as visible text in the cell, or put it in an `id`'d `sr-only` span and point `aria-describedby` at it from the control. Same at :264 for the Remove button.

web/src/features/workspaces/WorkspacesPage.tsx
```diff
@@ -0,0 +1,564 @@
...
+            >
+              Create workspace
+            </Button>
+          )
+        }
+      />
+
+      <ErrorBanner
+        error={
+          context.error ?? workspaces.error ?? orgMembers.error ?? remove.error
```
khaledosman: `remove.error` is rendered here and again at :549 as the `ConfirmDialog`'s own `error`, so a failed delete shows the same message twice, once behind the open dialog. Drop `remove.error` from this banner; the dialog owns that mutation's error.

Same duplication on `OrganizationMembersPage.tsx:296` vs `:336`.

web/src/features/workspaces/WorkspacesPage.tsx
```diff
@@ -0,0 +1,564 @@
...
+// spellings, published on both requests. `asMembershipRole` narrows a picker's
+// string back to it.
+const ROLE_OPTIONS = MEMBERSHIP_ROLES.map((role) => ({
+  value: role,
+  label: membershipLabel(role),
+}))
+
+const getWorkspaceRowKey = (workspace: Workspace): string => workspace.id
+
+function formatDate(value: string): string {
```
khaledosman: This duplicates `features/keys/KeysPage.tsx:50` minus its `Number.isNaN(d.getTime())` guard, so a malformed `created_at` renders the string "Invalid Date" into the table cell.

`shared/helpers/format.ts` already holds `formatDateTime`; put the date-only variant there with the guard and use it from both pages.

khaledosman (commented):
Reviewed against `AGENTS.md`, `web/AGENTS.md`, `.github/skills/frontend-standards/` and `.github/instructions/frontend-standards.instructions.md`. Five inline findings; the `WorkspaceMembersPanel` key is a state leak between workspaces.

🤖 Review generated by Claude Code (Claude Opus 5), acting as an agent on @khaledosman's behalf.

tbille (approved):

tbille (approved):

PR #627, opened 2026-08-18
feat(hybrid): initialize a local database and a shadow caller identity

tbille:
## Description

Hybrid mode now calls `init_db()` (migrations included) the same as standalone, but request-plane-only: it skips every startup step that belongs to a local management API hybrid mode does not have (master key, dashboard sessions, provider/pricing overlays, discovery, alias/policy loading, which joins in a later phase). This is what #1643's gateway survivals (aliases, routing memory, router preferences, files, batches) need: a shadow `platform_identity.ensure_local_user` row keyed on the caller identity the platform's resolve response now carries (mozilla-ai/otari#615 / mozilla-ai/otari-ai#1710), satisfying the `NOT NULL` FK four of those six tables have onto `users.user_id`.

**Deliberately narrow.** `get_db_if_needed` still yields `None` in hybrid mode, unchanged. A first pass wired the shadow-identity write through that shared dependency and it silently broke tool-pricing enforcement (a couple dozen call sites in `_pipeline.py` use `ctx.db is None` as their own implicit "are we in hybrid mode" gate: budget reservation, usage logging, tool-pricing enforcement). Widening what that dependency yields reactivated every one of them for hybrid traffic, turning expected `502`s into `402`s on the sandbox/web-search unreachable-backend tests. The shadow row is written through its own `create_session()` instead, fully isolated from that shared dependency, so nothing else in the request pipeline changes behavior.

**Reverses a previously tested invariant.** `test_hybrid_mode_surface.py` pinned that hybrid mode boots even with an unreachable database. It now requires a real, working one, the same as standalone; the complementary test (`test_hybrid_mode_fails_to_start_with_an_unreachable_database`) is added to pin that reversal explicitly rather than let it happen as a side effect.

**Test hygiene fallout.** Every other test file that boots a hybrid `GatewayConfig` without an explicit `database_url` picked up an isolated `tmp_path` one instead of silently sharing a default `./otari.db` in the working directory across test runs (confirmed one was actually being created and left behind before this fix).

**Nothing observable changes for any caller yet.** Aliases stay inert in hybrid (that's the next phase, #1643) and no new endpoint is reachable.

Rides on the Phase 0 decision request in otari-ai#1717 (proposing this narrowed exception to the M4 spec's "hybrid gateway has no local management state" line). Per direction, proceeding with implementation in parallel with that review.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Refs mozilla-ai/otari-ai#1643

## Checklist

- [x] Documentation was updated where necessary (`docs/modes.md` is deliberately **not** touched here: nothing user-observable changes until aliases actually start resolving in hybrid mode next phase; updating it now would describe a future state as present fact).
- [x] If the API contract changed, I regenerated the OpenAPI spec (N/A, no route/schema change).
- [x] If this PR adds or changes a persistent table, a control-plane contract the otari.ai platform consumes or serves, a capability the platform also has, or a mode-specific surface, I appended an entry to the M4 reconciliation ledger ([otari-ai#1587](https://github.com/mozilla-ai/otari-ai/issues/1587), [comment](https://github.com/mozilla-ai/otari-ai/issues/1587#issuecomment-5324920794)). See [.github/instructions/reconciliation-ledger.instructions.md](.github/instructions/reconciliation-ledger.instructions.md).

Local results: ruff, mypy, `check_architecture.py`, `openapi --check`, `postman --check` all clean. Full `tests/unit` (1787 passed) and `tests/integration` (1200 passed) suites pass, plus the OSS-edition smoke script.

## AI Usage

**AI Model/Tool used:** Claude Sonnet 5 via Claude Code.

## Summary

- Added hybrid-mode database initialization and migrations.
- Created a local shadow user from the platform caller identity.
- Kept hybrid mode limited to request-plane features.
- Preserved existing downstream behavior by keeping `get_db_if_needed` session-free in hybrid mode.
- Added startup validation for unreachable databases.
- Updated tests to use isolated temporary databases and verify shadow-user behavior.

## Benefits

Hybrid mode now has a reliable local identity and database foundation without enabling management features or changing caller-visible endpoints.

### DIFF
--- a/src/gateway/api/routes/_pipeline.py
+++ b/src/gateway/api/routes/_pipeline.py
@@ -93,6 +93,7 @@ from gateway.api.routes._tools import (
     declares_native_web_search,
 )
 from gateway.core.config import GatewayConfig
+from gateway.core.database import create_session
 from gateway.core.env import otari_env
 from gateway.core.usage import cache_read_tokens_of, cache_write_1h_tokens_of, cache_write_tokens_of
 from gateway.inflight import track_request
@@ -122,6 +123,7 @@ from gateway.services.mcp_loop import (
 )
 from gateway.services.metered_pricing import calculate_metered_cost
 from gateway.services.model_access import is_model_allowed, model_not_allowed_detail, resolve_request_allowlist
+from gateway.services.platform_identity import ensure_local_user
 from gateway.services.policy_store import resolve_effective_policy
 from gateway.services.pricing_service import (
     GATEWAY_TOOL_PRICING_PROVIDER,
@@ -975,7 +977,9 @@ async def resolve_request_context(
     """Run the shared handler preamble up to (and including) budget pre-debit.
 
     Hybrid mode: extract the caller's bearer token and resolve the routing
-    plan against the platform; no local DB state is touched.
+    plan against the platform. The only local DB write is best-effort:
+    ensuring a shadow ``User`` row exists for the platform's caller identity,
+    when the resolve response carries one (see ``ensure_local_user``).
 
     Standalone mode: validate the API key, resolve the billed user, check the
     rate limit, then reserve the estimated cost. The reservation is taken
@@ -1042,6 +1046,29 @@ async def resolve_request_context(
             route.fallback_enabled,
             resolve_latency_ms,
         )
+        # The resolve response is the only place hybrid mode learns a caller
+        # identity (docs/hybrid-mode-protocol.md's optional user_id). A local
+        # shadow row makes that identity usable by this gateway's own
+        # survival tables (aliases, routing memory, files, batches), which
+        # all FK to users.user_id.
+        #
+        # Deliberately opens its own session via create_session() rather than
+        # using this function's `db` parameter (which stays None in hybrid
+        # mode, unchanged): a couple dozen call sites below key off `ctx.db
+        # is None` as their own implicit "are we in hybrid mode" gate (budget
+        # reservation, usage logging, tool-pricing enforcement), and handing
+        # them a real session here would silently reactivate every one of
+        # them for hybrid traffic. This write is isolated on purpose.
+        #
+        # A peer that omits the field, or a failure creating the shadow row,
+        # both degrade to user_id=None (today's behavior) rather than failing
+        # an otherwise-servable request.
+        if route.user_id is not None:
+            try:
+                async with create_session() as identity_db:
+                    user_id = await ensure_local_user(identity_db, route.user_id)
+            except SQLAlchemyError:
+                logger.warning("Failed to ensure local shadow user for platform identity", exc_info=True)
     else:
         if db is None:
             raise adapter.error(500, DB_UNAVAILABLE_DETAIL, ErrorKind.API)
--- a/src/gateway/main.py
+++ b/src/gateway/main.py
@@ -233,10 +233,20 @@ def _create_lifespan(config: GatewayConfig) -> Callable[[FastAPI], Any]:
         price_refresher: asyncio.Task[None] | None = None
         discovery_refresher: asyncio.Task[None] | None = None
         catalog_refresher: asyncio.Task[None] | None = None
+        # Every mode initializes a database as of #1643's gateway survivals:
+        # aliases, routing memory, router preferences, files, and batches all
+        # need one in hybrid mode too, not just standalone. Hybrid mode's
+        # init is deliberately request-plane-only: it runs migrations against
+        # the same linear Alembic chain (unused management tables stay
+        # empty), but skips every startup step below that belongs to a local
+        # management API hybrid mode does not have (master key, dashboard
+        # sessions, provider/pricing overlays, discovery, alias/policy
+        # loading). Alias loading joins this branch in a later phase, once
+        # hybrid mode actually resolves aliases (see _pipeline.py).
+        init_db(config)
         if config.is_hybrid_mode:
             log_writer = NoopLogWriter()
         else:
-            init_db(config)
             async with create_session() as session:
                 # Persisted dashboard overrides win over config/env; apply them
                 # before pricing init so default-pricing behavior is consistent.
--- a/src/gateway/services/platform_identity.py
+++ b/src/gateway/services/platform_identity.py
@@ -1,0 +1,68 @@
+"""Bridge a platform-supplied caller identity onto a local shadow ``User`` row.
+
+Hybrid mode has no local user directory; the platform is the identity
+authority there. But this gateway's own survival tables (``model_aliases``,
+``routing_policies``, ``routing_memory``, ``router_preferences``,
+``file_objects``, ``batches``) all carry a foreign key to ``users.user_id``,
+four of them ``NOT NULL``, so keying anything on a hybrid caller's identity
+needs a local row for that key to point at. This module creates that row on
+first sight of a platform ``user_id`` and never touches it again: a shadow row
+is deliberately not synchronized with the platform user's alias, blocked
+status, or anything else the platform owns.
+"""
+
+from __future__ import annotations
+
+from sqlalchemy.dialects.postgresql import insert as pg_insert
+from sqlalchemy.dialects.sqlite import insert as sqlite_insert
+from sqlalchemy.ext.asyncio import AsyncSession
+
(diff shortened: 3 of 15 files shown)


tbille:
Closing: direction changed. Hybrid mode should query the control plane for what it needs (aliases, routing memory/preferences, etc.) rather than gaining a local database and a shadow identity row. This local-DB approach is no longer the right shape; a fresh design querying the platform's API follows. Refs mozilla-ai/otari-ai#1643.

tbille:
Reopening: reverting the direction change. The API-querying approach would have reversed a settled (2026-08-07), cross-referenced otari-ai architecture decision that aliases/routing-memory/router-preferences stay gateway-owned, with the platform only supplying an identity to key local storage on (there is also no existing platform-side hook to build on: no RoutingPort, no alias-resolution step, no KNN/embedding concept). Back to this PR's original approach: a local database + shadow caller identity in hybrid mode. Refs mozilla-ai/otari-ai#1643.

src/gateway/main.py
```diff
@@ -233,10 +233,20 @@ async def lifespan(app: FastAPI) -> AsyncGenerator[None, None]:
...
+        # aliases, routing memory, router preferences, files, and batches all
+        # need one in hybrid mode too, not just standalone. Hybrid mode's
+        # init is deliberately request-plane-only: it runs migrations against
+        # the same linear Alembic chain (unused management tables stay
+        # empty), but skips every startup step below that belongs to a local
+        # management API hybrid mode does not have (master key, dashboard
+        # sessions, provider/pricing overlays, discovery, alias/policy
+        # loading). Alias loading joins this branch in a later phase, once
+        # hybrid mode actually resolves aliases (see _pipeline.py).
+        init_db(config)
```
khaledosman, line 246: The "nothing user-observable changes" justification for skipping docs does not hold for this line: hybrid mode now hard-fails at boot without a reachable database, which `test_hybrid_mode_fails_to_start_with_an_unreachable_database` pins deliberately. Three places are now wrong:

- `docs/deployment.md:60` — "No local database or provider credentials are needed", then "No `providers` block, no `database_url`, no `master_key`", then "No postgres container is needed; otari.ai handles storage."
- `docs/modes.md:54` — "No local database is used for keys, users, budgets, or usage logs."
- `docs/modes.md:109` — the comparison table's `Database │ Local (SQLite/PostgreSQL) │ None` row.

With `database_url` defaulting to `sqlite:///./otari.db`, the `docker run --rm` that `deployment.md` documents (no volume) now silently creates `/app/otari.db` inside an ephemeral container, and several hybrid replicas behind a load balancer each get their own private one. That is precisely the storage the #1643 survivals are going to live in, so the requirement needs stating in the phase that introduces it, not the phase that starts reading it.

src/gateway/api/routes/_pipeline.py
```diff
@@ -1042,6 +1046,29 @@ async def resolve_request_context(
...
+        # mode, unchanged): a couple dozen call sites below key off `ctx.db
+        # is None` as their own implicit "are we in hybrid mode" gate (budget
+        # reservation, usage logging, tool-pricing enforcement), and handing
+        # them a real session here would silently reactivate every one of
+        # them for hybrid traffic. This write is isolated on purpose.
+        #
+        # A peer that omits the field, or a failure creating the shadow row,
+        # both degrade to user_id=None (today's behavior) rather than failing
+        # an otherwise-servable request.
+        if route.user_id is not None:
```
khaledosman, line 1066: `is not None` should be a truthiness check. `_platform.py:579` coerces the field with `str(raw_user_id)`, so a platform sending `"user_id": ""` yields `""`, which passes this gate and inserts a `users` row with an empty-string primary key that nothing afterwards can distinguish or clean up. `if route.user_id:` costs nothing and closes it.

src/gateway/api/routes/_pipeline.py
```diff
@@ -1042,6 +1046,29 @@ async def resolve_request_context(
...
+            try:
+                async with create_session() as identity_db:
+                    user_id = await ensure_local_user(identity_db, route.user_id)
```
khaledosman, line 1069: No integration test covers this branch. The behavior this PR adds is "a hybrid request whose resolve response carries `user_id` writes a shadow row", and only `ensure_local_user` in isolation is tested.

`tests/integration/test_hybrid_mode_chat.py` already stubs the platform resolve, so adding `user_id` to that stub buys three assertions cheaply: the row exists after a request, a resolve response omitting the field writes nothing, and a failing shadow write still serves the request rather than propagating. AGENTS.md's Change Validation Checklist asks for integration coverage on a route + database change, and the `except SQLAlchemyError` degradation path in particular is the kind of thing that only stays true if something pins it.

tests/unit/test_platform_identity.py
```diff
@@ -0,0 +1,151 @@
...
+from gateway.models.entities import Base, User
+from gateway.services.platform_identity import (
+    SHADOW_USER_ORIGIN,
+    ensure_local_user,
+    reset_known_user_cache,
+)
+
+
+@pytest.fixture(autouse=True)
+def _clean_cache() -> None:
```
khaledosman, line 28: `reset_known_user_cache` is autouse only inside this module, but `_known_user_ids` is process-global and every hybrid integration test now gets its own `tmp_path` SQLite in the same process. Once anything actually reads the shadow row, a test that reuses a platform user id another test already cached will find no row and fail for a reason that has nothing to do with it.

`tests/conftest.py` already carries autouse fixtures for exactly this class of process-wide state (`_no_background_refresh:16`, `_reset_default_pricing:47`); the reset belongs there rather than here.

src/gateway/services/platform_identity.py
```diff
@@ -0,0 +1,68 @@
...
+# Provenance marker distinguishing a platform-originated shadow row from one
+# an operator created directly (standalone /v1/users), so a future
+# re-parenting pass (see otari-ai#1643's fan-out) can tell them apart.
+SHADOW_USER_ORIGIN = "platform"
+
+# In-process cache of platform user ids already known to have a local row.
+# A write-through set, not a TTL cache like alias_service's: a shadow row is
+# created once and never edited or deleted, so there is nothing to
+# invalidate and no refresher to converge sibling workers on.
+_known_user_ids: set[str] = set()
```
khaledosman, line 31: `_known_user_ids` grows one entry per distinct platform identity for the lifetime of the process, with no eviction. The comment argues only that there is nothing to *invalidate*, which is true and a separate question from unbounded growth.

A self-hosted gateway sees a handful of identities, but the gateway mozilla.ai operates runs in hybrid mode and sees every otari.ai user, so this is a set that only ever grows in the deployment where it matters most. Either bound it (an LRU keyed on the id is enough: a false miss just re-runs a no-op upsert) or say in the comment why unbounded is acceptable here.

khaledosman (commented):
Reviewed against `AGENTS.md`, `src/gateway/AGENTS.md` and `.github/instructions/`. Five inline findings; the stale hybrid-mode docs and the missing integration coverage are the two I would want addressed before merge.

🤖 Review generated by Claude Code (Claude Opus 5), acting as an agent on @khaledosman's behalf.

njbrake:
Closing. mozilla-ai/otari-ai#1717, the spec amendment this rides on, is closed as obsolete: hybrid mode is not getting a local database, because the control plane resolves aliases and routing for it.

This PR's premise was accurate. Four of the six survival tables do carry a NOT NULL foreign key onto `users.user_id`, and a hybrid gateway has no row to satisfy it. But the fix belongs on the deciding side. `_pipeline.py` already refuses local policy names in hybrid and defers the routing decision to the platform, and after the M5 rehome the control plane is otari, so it holds these tables and their services already.

Two things worth carrying forward if a local shadow identity is ever needed again. `get_or_create_attribution_user` in `users_repository.py` landed in #633 and is the same primitive as this PR's `ensure_local_user`, with revive-on-soft-delete semantics that `on_conflict_do_nothing` plus an in-process cache does not have. And #673's single-head migration test becomes load-bearing for anything that makes hybrid run Alembic at startup, since a forked head would then be a data-plane outage rather than an upgrade failure.

PR #628, opened 2026-08-18
feat(reprise): retain the resolved workspace and organization ids

dpoulopoulos:
## Description

The platform's resolve response has always carried `workspace_id` and `organization_id`, and the gateway read past both: `_parse_resolve_payload` built a `ResolvedRoute` holding only `request_id`, `fallback_enabled`, `attempts`, and (as of #615) `user_id`. This keeps the two it already receives.

- `workspace_id: str | None` and `organization_id: str | None` on `ResolvedRoute`.
- Read off the top level of both payload shapes, the attempts-list one and the legacy flat one. Unlike `user_id`, the tenancy pair does get a legacy mirror: a peer that still answers flat may nonetheless know its own tenant, and a record that ships untenanted is a record the miner discards.
- Absence is tolerated. Both stay `None`, no error, no request impact. That is the fail-open rule the observation stream runs under.

**Why it matters.** Nothing downstream can recover these after the fact. `X-Gateway-Token` authenticates the calling gateway and carries no tenant of its own, and the platform's default gateway token serves every org. Observations carry tool arguments, so pooling every org's together would be a tenancy problem rather than only a measurement one.

**Not a protocol change.** The values already cross the wire. `docs/hybrid-mode-protocol.md` only moves them out of its "fields Otari does not read" list and documents the fail-open semantics; `provider_key_id` and `allowed_models` stay in that list.

They are recording-only, and structurally so: the pair lives on the route, never on a `ResolvedAttempt`, so it cannot reach a provider call. Nothing consumes it yet; the fingerprint that does is mozilla-ai/otari-ai#1648 and the recorder that stamps it is mozilla-ai/otari-ai#1484.

**Tests** (`tests/unit/test_platform_resolve_payload.py`):

- both ids read from the attempts-list shape
- both ids read from the legacy single-attempt shape
- both shapes resolve with the pair omitted: both `None`, no error
- neither id reaches the provider call, pinned against a future move onto `ResolvedAttempt`, whose `extra_params` is forwarded to the provider client verbatim

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Refs mozilla-ai/otari-ai#1646. That issue lives in the platform repo by design, and its code lands here.

Follow-ups that consume these fields, neither of them in this PR: mozilla-ai/otari-ai#1648 (fingerprint) and mozilla-ai/otari-ai#1484 (recorder).

## Checklist

- [x] I have added or updated tests that cover my change (`tests/unit`, `tests/integration`). Four unit tests in `tests/unit/test_platform_resolve_payload.py`, one per acceptance criterion on the issue. No integration test: the change is confined to parsing a payload the existing hybrid-mode integration tests already cover end to end.
- [x] I ran the Definition of Done checks locally (`make lint`, `make typecheck`, `make test`). `make lint` (architecture check plus Ruff), `make typecheck` (`mypy --strict`, 357 files) and `make openapi-check` are clean. On tests: the full `tests/unit` suite (1786 passed, 9 skipped) and the three hybrid-mode integration suites (66 passed) are green locally; the combined `make test` run over `tests/unit` plus all of `tests/integration` was still finishing when this description was written, so treat the CI `test` job on this head as the authority for it.
- [x] Documentation was updated where necessary. `docs/hybrid-mode-protocol.md`: both fields move out of the "fields Otari does not read" list into the documented resolve shapes, with the fail-open semantics and the legacy-shape mirror spelled out.
- [x] If the API contract changed, I regenerated the OpenAPI spec (`uv run python scripts/generate_openapi.py`). Not applicable, no public route or schema changed; confirmed with `make openapi-check`.
- [x] If this PR adds or changes a persistent table, a control-plane contract the otari.ai platform consumes or serves, a capability the platform also has, or a mode-specific surface, I appended an entry to the M4 reconciliation ledger ([otari-ai#1587](https://github.com/mozilla-ai/otari-ai/issues/1587)). See [.github/instructions/reconciliation-ledger.instructions.md](https://github.com/mozilla-ai/otari/blob/main/.github/instructions/reconciliation-ledger.instructions.md). Entry: https://github.com/mozilla-ai/otari-ai/issues/1587#issuecomment-5326037967, under the "Tenancy and access" row, confirming the existing target rather than changing it.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

Written from the issue's proposed change and acceptance criteria, test-first: the three parsing tests were written and observed failing before the fields existed. The agent also ran `make lint`, `make typecheck`, `make test`, and `make openapi-check` locally, and posted the ledger entry.

Two judgment calls a reviewer may want to weigh in on:

1. The legacy single-attempt shape reads the tenancy pair, which differs from how the immediately preceding `user_id` change (#615) deliberately left that branch alone. The issue asks for both shapes, so that is what this does, with a comment at the branch explaining why the pair is the exception. Dropping the legacy read is a two-line change if you prefer consistency with `user_id`.
2. An earlier draft passed the pair as `**_tenant_ids(payload)` at both call sites. That type-checks, but pydantic ignores an unexpected keyword, so a typo'd key would silently leave the field `None`. It now unpacks a tuple into named keyword arguments so mypy checks each field.

**NOTE:**

### DIFF
--- a/docs/hybrid-mode-protocol.md
+++ b/docs/hybrid-mode-protocol.md
@@ -38,11 +38,12 @@ Otari ignores any it does not recognize. Consumers of this contract MUST do the
 same (ignore unknown fields) so the platform can add fields without breaking
 older gateways.
 
-For example, the otari.ai resolve response also carries `workspace_id`,
-`organization_id`, `provider_key_id`, and `allowed_models`. Otari does not read
-these today, so they are intentionally absent from the shapes documented here.
-`user_id` (below) is the one exception: Otari reads it when present, but a
-peer omitting it is exactly as valid as any other unrecognized field.
+For example, the otari.ai resolve response also carries `provider_key_id` and
+`allowed_models`. Otari does not read these today, so they are intentionally
+absent from the shapes documented here. `user_id`, `workspace_id`, and
+`organization_id` (all below) are the exceptions: Otari reads each when
+present, but a peer omitting any of them is exactly as valid as a peer sending
+an unrecognized field.
 
 When the operator points Otari's web-search backend at the platform
 (`OTARI_WEB_SEARCH_URL` under `base_url`), Otari also sends `X-Gateway-Token`
@@ -74,6 +75,8 @@ Content-Type: application/json
   "request_id": "01HXY...",
   "fallback_enabled": true,
   "user_id": "3fa85f64-5717-4562-b3fc-2c963f66afa6",
+  "workspace_id": "9b1deb4d-3b7d-4bad-9bdd-2b0d7b3dcb6d",
+  "organization_id": "7c9e6679-7425-40de-944b-e07fc1f90ae7",
   "attempts": [
     {
       "attempt_id": "01HX1...",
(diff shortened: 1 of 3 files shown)


aittalam (approved):
LGTM, thank you @dpoulopoulos ! 🙏

PR #629, opened 2026-08-18
feat(otel): Add otel context propagation

### LINKED ISSUE #617: [FEATURE] OTel context header handling
### Problem or use case

When requests are passed on to Otari, OTel trace/span information may be present. Since Otari already generates OTel traces, it would be natural to include parent trace information.

### Proposed solution

The traceparent header will be examined to enable context propagation across Otari and collaborating services.

### Alternatives considered

None considered.

javiermtorres:
## Description

This PR enables OTel context propagation through the incoming endpoints, so context is preserved in a distributed setting.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Closes #617 

## Checklist

- [x] If this PR adds or changes a persistent table, a control-plane contract the otari.ai platform consumes or serves, a capability the platform also has, or a mode-specific surface, I appended an entry to the M4 reconciliation ledger ([otari-ai#1587](https://github.com/mozilla-ai/otari-ai/issues/1587)). See [.github/instructions/reconciliation-ledger.instructions.md](instructions/reconciliation-ledger.instructions.md).

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

Copilot/auto

**NOTE:**

## Summary

- Added support for incoming trace context in gateway requests.
- Preserved trace information between Otari and collaborating services.
- Added OpenTelemetry instrumentation as a runtime dependency.
- Added unit and integration tests for trace propagation and invalid or missing context.

## Technical notes

- The middleware reads `traceparent` and `tracestate` headers.
- It applies valid context during request processing and removes it afterward.
- Invalid or missing context does not interrupt request handling.


src/gateway/main.py
```diff
@@ -505,7 +506,9 @@ async def dashboard_build() -> dict[str, str]:
         async def root_index() -> str:
             return ROOT_TUTORIAL_HTML
 
+    # Middleware stack is registered in reverse order (last-added runs first)
     app.add_middleware(SecurityHeadersMiddleware)
+    app.add_middleware(TraceContextPropagationMiddleware)
```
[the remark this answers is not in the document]
javiermtorres: Context propagation is mainly meant for inter-service communication, not end-user tracing, hence CORS support for otel traces will not be enabled. This will be appropriately documented.

tests/conftest.py
```diff
@@ -66,3 +72,49 @@ def _reset_default_pricing() -> Generator[None, None, None]:
...
+        carrier: A mapping of request headers.
+
+    Returns:
+        An OpenTelemetry Context with the extracted trace context, or None if
+        extraction failed or no valid traceparent is present.
+    """
+    try:
+        context = TraceContextTextMapPropagator().extract(carrier=carrier)
+    except Exception:  # noqa: BLE001
+        return None
```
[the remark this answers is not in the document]
javiermtorres: Addressed.

src/gateway/context_propagation.py
```diff
@@ -0,0 +1,91 @@
...
+            try:
+                response = await call_next(request)
+            finally:
+                # Detach the context when the request is done
+                otel_context.detach(token)
+            return response
+
+        # If extraction fails or no context is present, proceed normally
+        response = await call_next(request)
+        return response
```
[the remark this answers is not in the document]
javiermtorres: @copilot check if this is solved

pyproject.toml
```diff
@@ -23,6 +23,7 @@ dependencies = [
     "markitdown[docx,pptx,xlsx,pdf]>=0.1.0",
     "mcp>=1.28.1",
     "opentelemetry-api>=1.39.1",
+    "opentelemetry-instrumentation>=0.50b0",
```
dni138: **major:** `opentelemetry-instrumentation` is never imported.

`grep -rn "opentelemetry.instrumentation" src tests scripts alembic` returns nothing. The new module uses only `opentelemetry-api` surfaces (`opentelemetry.context`, `opentelemetry.trace`, `opentelemetry.trace.propagation.tracecontext`), and `opentelemetry-api>=1.39.1` is already on line 25.

What the line costs, all verified locally by switching between `main` and this branch:

* It ships `opentelemetry-instrumentation`, `opentelemetry-semantic-conventions`, `packaging`, and `wrapt` into the **production** image for nothing.
* Its `wrapt>=1.0,<2` cap **downgrades locked `wrapt` from 2.1.2 to 1.17.3** project-wide. On `main` the only consumer is `testcontainers` (a dev dep); this makes `wrapt` a runtime dependency and pins the whole project off 2.x. `uv sync` going back to `main` reports exactly that: `- wrapt==1.17.3 / + wrapt==2.1.2`. cp314 wheels do exist for 1.17.3, so the 3.14 image still builds, but it is an unintended lockfile move.
* `>=0.50b0` is off the series the other otel pins track. It resolves to `0.60b1`, which is the `0.60b1`/`1.39` pairing, so the floor understates what is actually required.

**Suggested fix:** drop the line and re-lock. If it is deliberate, so operators can run `opentelemetry-instrument otari serve`, it belongs in an optional extra next to `ocr` and `s3` in `[project.optional-dependencies]` plus a line in the docs, rather than a hard runtime dependency. That also settles the question in the summary.

src/gateway/context_propagation.py
```diff
@@ -0,0 +1,93 @@
...
+
+    Args:
+        carrier: A mapping of request headers (e.g. `request.headers`).
+
+    Returns:
+        An OpenTelemetry Context with the extracted trace context, or None if
+        extraction failed or no valid traceparent is present.
+    """
+    try:
+        context = TraceContextTextMapPropagator().extract(carrier=carrier)
```
dni138: **major:** this replaces the ambient OTel context instead of extending it.

`extract()` is called with no `context=`, and upstream that means `context = Context()`, a brand new empty one. So the returned context holds **only** the remote-span key, and `otel_context.attach()` on line 61 then makes that the entire current context for the request. Everything already in the ambient context is dropped for the request's lifetime: baggage, the suppress-instrumentation key, and any enclosing span.

I ran this against the branch to be sure of the mechanism:

```
--- (a) baggage set before the request survives the middleware? ---
ambient baggage before middleware: {'tenant': 'acme'}
ambient baggage inside request   : {}

--- (b) an enclosing SERVER span (what -instrumentation-asgi opens) ---
server span id               : efc90c4c26f532d9
tool span's actual parent    : 00f067aa0ba902b7   <- the header
expected (child of SERVER)   : efc90c4c26f532d9
=> re-parented to the header, SERVER span orphaned: True
```

This is latent in today's default deployment, since nothing else populates the context at request time, and I want to be clear about that. It stops being latent the moment auto-instrumentation or baggage is in play, which the dependency added in `pyproject.toml:26` suggests is the intended direction. Under `opentelemetry-instrument`, the ASGI instrumentor's SERVER span is the thing that gets discarded, so tool and route spans re-parent to the caller's span and the gateway's own server span falls out of its local hierarchy.

**Suggested fix,** one argument plus a guard:

```python
context = propagator.extract(carrier=carrier, context=otel_context.get_current())
```

Verified that this keeps the intended behavior rather than trading it away: baggage survives (`{'tenant': 'acme'}`) and the current span is still the remote parent (`00f067aa0ba902b7`). On top of that, consider skipping the attach entirely when `otel_context.get_current()` already carries a valid span, which means something upstream already extracted and this middleware has nothing to add.

src/gateway/context_propagation.py
```diff
@@ -0,0 +1,93 @@
...
+"""
+
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, Any
+
+from opentelemetry import context as otel_context
+from opentelemetry import trace
+from opentelemetry.context import Context
+from opentelemetry.trace.propagation.tracecontext import TraceContextTextMapPropagator
```
dni138: **minor:** hard-codes one propagator instead of using the configured global one.

`opentelemetry.propagate.extract(carrier)` honors `OTEL_PROPAGATORS`, which is the standard knob operators reach for, and its default value (`tracecontext,baggage`) also carries W3C **baggage**. As written, a deployment that sets `OTEL_PROPAGATORS=b3` because its other services speak B3, or that wants baggage to flow, is silently ignored, and the docs sentence "extracts it with the standard OpenTelemetry propagator" reads as though the standard configuration applies when it does not.

One nuance if you take this, which I checked so the suggestion is actually usable: a composite propagator can return a context carrying baggage and **no** span, so the `INVALID_TRACE_ID` check on line 90 would throw that away.

```
baggage extracted   : {'tenant': 'acme', 'region': 'eu'}
trace_id is INVALID : True
=> the current guard shape returns None here and drops that baggage
```

So the guard has to change from "no valid span means return `None`" to "attach whenever anything was extracted". Also worth instantiating the propagator once at module scope rather than per request.

src/gateway/context_propagation.py
```diff
@@ -0,0 +1,93 @@
...
+    Args:
+        carrier: A mapping of request headers (e.g. `request.headers`).
+
+    Returns:
+        An OpenTelemetry Context with the extracted trace context, or None if
+        extraction failed or no valid traceparent is present.
+    """
+    try:
+        context = TraceContextTextMapPropagator().extract(carrier=carrier)
+    except Exception as exc:  # noqa: BLE001
```
dni138: **minor:** this guard cannot fire for the carrier the middleware actually passes, and the real per-request log risk is elsewhere.

`TraceContextTextMapPropagator.extract` returns the context unmodified for every malformed `traceparent` rather than raising. I checked each shape against the production function on this branch:

| carrier | result |
| --- | --- |
| `{}` | `None`, no exception |
| `traceparent: invalid` | `None`, no exception |
| all-zero trace id | `None`, no exception |
| version `ff` | `None`, no exception |
| extra fields on version `00` | `None`, no exception |
| malformed `tracestate` | `Context`, no exception |

The one input that *does* raise is a non-string header value (`TypeError: expected string or bytes-like object, got 'int'`), and that cannot come from `Headers.get`, which always returns `str`. It is reachable only because `carrier: Any` invites arbitrary mappings. So typing the carrier `Mapping[str, str]` (nit 3) makes the broad `except` and its `# noqa: BLE001` provably unnecessary for the one real caller, which is also what the backend-standards skill asks for ("prefer specific exceptions over broad `except Exception`"). The `INVALID_TRACE_ID` check on line 90 is already the graceful-degradation path the tests are aiming at.

Separately, and more useful: a client-controlled malformed `tracestate` makes **OpenTelemetry's own** logger emit a WARNING per request, which no amount of handling here suppresses.

```
WARNING:opentelemetry.trace.span:Member doesn't match the w3c identifiers format =====not
WARNING:opentelemetry.trace.span:There can't be more than 32 key/value pairs.
```

That is unauthenticated log amplification driven by a header. Worth a sentence in the docs or the threat-model comment suggested on line 61, so an operator knows to set the level on `opentelemetry.trace.span`.

src/gateway/context_propagation.py
```diff
@@ -0,0 +1,93 @@
...
+            await self.app(scope, receive, send)
+            return
+
+        extracted_context = _extract_context_from_carrier(Headers(scope=scope))
+
+        if extracted_context is None:
+            await self.app(scope, receive, send)
+            return
+
+        token = otel_context.attach(extracted_context)
```
dni138, line 70: **minor:** this is a trust decision, and right now it is made implicitly.

Middleware runs before any route auth dependency, so any caller, with no credential at all, sets the trace id, the parent span id, the sampled flag, and up to 32 `tracestate` entries for the spans the operator's collector then ingests. That is enough to stitch chosen ids into an operator's traces or to force sampling.

To be clear about severity: this is what `opentelemetry-instrumentation-fastapi` does by default, so it is **not a vulnerability**, and I am not asking for it to be turned off. The point is a convention one. The repo's own guidance is that new security-relevant behavior is either gated by a flag that fails closed (backend-standards, "Config & env"; security-review, "Security misconfiguration") or carries a threat-model comment, the way `api/routes/chat.py` documents the sandbox-URL decision. `MetricsMiddleware` is already gated on `enable_metrics` for a weaker reason than this.

**Suggested fix:** either a `trust_incoming_trace_context` config flag, or, if unconditional is the deliberate answer, a short threat-model comment in the module docstring plus a line in the docs saying this trusts the caller and belongs behind a proxy that strips the two headers at the edge. Either one makes the decision reviewable later; the docs already say the feature is "intended for backend and service-to-service callers", and nothing currently enforces that reading.

tests/conftest.py
```diff
@@ -66,3 +72,46 @@ def _reset_default_pricing() -> Generator[None, None, None]:
     configure_default_pricing(False)
     configure_provider_types(None)
     reset_price_refresh_state()
+
+
+def extract_trace_context_from_carrier(carrier: Mapping[str, str]) -> Span | None:
```
dni138: **minor:** these helpers duplicate the production function, and the unit tests assert against the copy.

`extract_trace_context_from_carrier` and `_extract_otel_context_from_carrier` reimplement `gateway.context_propagation._extract_context_from_carrier` almost line for line, including the `INVALID_TRACE_ID` check. The five `test_extract_trace_context_from_carrier_*` cases then exercise **the copy**, so the production function's contract is covered nowhere.

Demonstrated on this branch: I replaced the production guard with `if False:` (deleting the entire "return `None` on absent, malformed, or all-zero ids" behavior) and the suite still reported `23 passed`. Two copies of one rule will drift, and the copy is the one the tests defend.

Two smaller things ride along. A non-fixture helper plus four OpenTelemetry imports land in the **root** conftest, which every test in the repo loads, for the benefit of one module. And `from conftest import ...` inside the test bodies works only because pytest happens to put `tests/` on `sys.path` in rootdir mode, which is a fragile way to reach a helper.

**Suggested fix:** import the real function in the unit test (reaching for a private is fine in a test), or promote it to a public name and import that, then delete both helpers from here. That single change is what makes those five tests start defending the shipped code.

tests/unit/test_context_propagation.py
```diff
@@ -0,0 +1,334 @@
...
+
+@pytest.mark.parametrize("invalid_traceparent", [
+    "invalid",  # completely wrong format
+    "00",  # too short
+    "00-short-parts",  # not enough parts
+    "00-short-id-00f067aa0ba902b7-01",  # trace ID too short
+    "00-4bf92f3577b34da6a3ce929d0e0e4736-short-01",  # span ID too short
+    "00-4bf92f3577b34da6a3ce929d0e0e4736-00f067aa0ba902b7-x",  # invalid trace flags
+])
+def test_invalid_traceparent_format_is_handled_gracefully(client: TestClient, invalid_traceparent: str) -> None:
```
dni138: **minor:** these cases pass whether or not the feature exists.

Each asserts only `response.status_code == 200` against `/health`, which returns 200 with or without the middleware. Measured on this branch, with `app.add_middleware(TraceContextPropagationMiddleware)` commented out:

```
FAILED test_http_request_span_inherits_traceparent
FAILED test_streaming_response_span_inherits_traceparent
2 failed, 21 passed
```

So 21 of 23 cases are not testing trace propagation. The same applies to `test_traceparent_header_is_extracted_from_request` (the name promises extraction, the body asserts a status code), `test_traceparent_header_with_correct_format`, `test_traceparent_with_trace_state_extension`, and both zero-id tests.

**Suggested fix:** this file already contains the right tool. `test_http_request_span_inherits_traceparent` registers a probe route and returns `format(span.get_span_context().trace_id, "032x")`. Point these cases at the same probe and assert what each one is named for: an invalid header should yield a trace id *different* from the injected one, which is what "handled gracefully" actually means and what would catch a future propagator swap that started accepting garbage. Six parametrized invalid inputs are genuinely worth having once they assert that.

Also drop the `expected_status` parameter on line 56: all three cases pass `200`, so it reads as a variable while being a constant.

tests/unit/test_context_propagation.py
```diff
@@ -0,0 +1,334 @@
...
+    response = client.get(
+        "/health",
+        headers={"traceparent": invalid_traceparent}
+    )
+    
+    # Should still return 200 (graceful degradation)
+    assert response.status_code == 200
+
+
+def test_multiple_requests_maintain_independent_contexts(client: TestClient) -> None:
```
dni138: **minor:** nothing covers the `detach` that the pure-ASGI design exists for.

This test is named for exactly the invariant that matters, request isolation, and then checks two 200s, which would hold even if `otel_context.detach(token)` were deleted. The `finally` in the middleware is the whole reason the module is raw ASGI rather than `BaseHTTPMiddleware`, so it is the one line most deserving a regression test.

**Suggested fix:** with the probe route from the previous comment, the real assertion is three lines. Send a request carrying `traceparent`, then one carrying none, and assert the second reports a *different* trace id. That fails loudly if the context ever leaks past the response, which is the bug that would otherwise show up in production as unrelated requests sharing a trace.

tests/unit/test_context_propagation.py
```diff
@@ -0,0 +1,334 @@
...
+    assert format(span_context.trace_id, "032x") == (
+        "4bf92f3577b34da6a3ce929d0e0e4736"
+    )
+
+
+def test_streaming_response_span_inherits_traceparent() -> None:
+    """Verify that spans created while streaming inherit incoming trace context."""
+    from opentelemetry import trace
+
+    app = create_app(GatewayConfig())
```
dni138: **minor:** this test boots the lifespan against the default database, so a unit run migrates `./otari.db`.

This is the only new test using `with TestClient(app)`, so the lifespan runs, and `create_app(GatewayConfig())` leaves `database_url` at its default `sqlite:///./otari.db` (`core/config.py:296`) with `auto_migrate=True`. Confirmed by running it on a clean tree:

```
$ rm -f otari.db && uv run pytest tests/unit/test_context_propagation.py -k streaming
1 passed
$ ls -l otari.db
-rw-r--r--  1  352256  otari.db
```

Three consequences. It runs `alembic upgrade head` against a file in the CWD (`core/database.py:136`), which on a developer machine is their real gitignored local dev DB. It calls `ensure_master_key` on it (`main.py:267`), minting and printing a key into that database. And it leaves the process-wide `_engine` and `_SessionLocal` pointed at that file for the rest of the pytest worker, since nothing calls `reset_db()`. The migration chain running here is also what puts the new `SAWarning: Table '_alembic_tmp_model_pricing' ...` into the unit suite's output.

**Suggested fix:** follow the pattern already in `tests/unit/test_dashboard_session.py:29`, which is there for this reason:

```python
config = GatewayConfig(
    database_url=f"sqlite:///{tmp_path / 'trace-test.db'}",
    master_key="sk-test-master",
    require_pricing=False,
)
```

Related, and worth fixing in the same pass: both new app builds call a bare `GatewayConfig()`, which reads ambient env, so a developer with `OTARI_AI_TOKEN` exported gets a hybrid-mode app with a different router set than CI builds.

docs/external-usage.md
```diff
@@ -170,6 +170,31 @@ not accepted). Authenticate with a budget-exempt API key via
 here: OTLP events carry no user attribution, so usage always binds to the key's
 user (to import a mixed feed for many users, use `/v1/usage/external-events`).
 
+### HTTP trace context propagation
```
dni138: **minor:** this section splits the one it was inserted into, and one of its claims is not true yet.

**Placement.** It lands between the OTLP authentication paragraph and the "Otari reads only the content-free usage attributes" table that belongs with it, so that table and its attribute list now read as a continuation of trace-context propagation rather than of the OTLP receiver. Beyond the local seam, the host page is "Importing external usage", which states at line 15 that "This is a standalone-mode feature; hybrid mode has no local usage database", while `TraceContextPropagationMiddleware` is registered unconditionally in **both** modes. A reader in hybrid mode has been told this page does not apply to them.

`docs/configuration.md` is probably the better home, since that is where `enable_metrics` and the Prometheus endpoint already live and this is the same kind of operator-facing observability knob. `docs/deployment.md` would work too. Failing a move, putting it after "Recognized coding agents" would at least keep the OTLP section intact.

**Accuracy,** line 185: "The `tracestate` value is preserved for downstream instrumentation." There is no `inject(` anywhere in `src/`, so nothing forwards either header to a provider or to the platform. What is preserved is the `trace_state` on the span context for spans created *in this process*, which is real and worth saying, but "downstream" will read to an operator as "the header reaches my provider call", and it does not unless they add httpx instrumentation themselves. Suggest naming that explicitly, since it is also the natural follow-up issue.

The CORS paragraph, for contrast, is accurate: `traceparent` and `tracestate` are genuinely absent from `allow_headers` in `main.py`.

dni138 (changes requested):
## Review summary

Reviewed against `48b7655`, with the branch checked out locally and every claim below reproduced by running it. `make lint`, `make typecheck`, `tests/unit/test_context_propagation.py` (23 passed) and `tests/integration/test_context_propagation.py` (1 passed) are all green, so nothing here is something CI could have caught for you.

The middleware design is right, and the part that is easy to get wrong is correct: pure ASGI so the `finally` outlives a streamed body, matching `InFlightMiddleware` and `MetricsMiddleware`, with `test_streaming_response_span_inherits_traceparent` genuinely proving it rather than asserting around it. Logging follows the `gateway.log_config` + `%s` convention, and the CORS scoping you documented is accurate.

Two items want a change before merge, and one question decides both of them.

### The question

**Which deployment model is this for: `opentelemetry-instrument` auto-instrumentation, or an SDK the operator wires up by hand?** The gateway configures no `TracerProvider` of its own today, so the answer changes the review:

* **Auto-instrumentation.** `opentelemetry-instrumentation-asgi` installs the outermost middleware, extracts the same header, and opens a SERVER span. This middleware then becomes redundant *and* actively harmful, because it discards that span from context (see the note on `context_propagation.py:85`).
* **Hand-configured SDK.** The middleware is exactly what is needed, and the newly added dependency is dead weight (see `pyproject.toml:26`).

### 2 major, 8 minor, 10 nits

Majors are on `pyproject.toml:26` and `context_propagation.py:85`. Minors cover the propagator choice, the exception guard, the trust boundary, four test issues, and the docs placement.

The largest of the test findings, quantified: **21 of the 23 new unit cases still pass with `app.add_middleware(TraceContextPropagationMiddleware)` commented out.** Only `test_http_request_span_inherits_traceparent` and `test_streaming_response_span_inherits_traceparent` fail. And separately, I deleted the production `INVALID_TRACE_ID` guard outright, the single behavior those five `test_extract_trace_context_from_carrier_*` cases exist to check, and all 23 still passed, because they exercise the conftest copy instead of the real function.

### Nits

1. **21 trailing-whitespace lines** in `tests/unit/test_context_propagation.py`, and `tests/integration/test_context_propagation.py` has **no final newline**. `tests/` on `main` has zero trailing-whitespace lines, so this is a real deviation; ruff's `W` rules are not in `select` and pre-commit runs only `ruff check`, so CI cannot see it. Strip them by hand rather than running `ruff format`: the repo is not format-clean (106 files would change), so the formatter would bury the diff.
2. **Duplicate import.** `from starlette.types import ASGIApp, Receive, Scope, Send` appears at runtime (line 21) and again under `TYPE_CHECKING` (line 26). The second is dead. `inflight.py` shows the house pattern.
3. **`carrier: Any`.** The only caller passes `starlette.datastructures.Headers`, and the conftest twin already types it `Mapping[str, str]`. Annotating it lets the `Any` import go, and it is what makes the broad `except` provably unnecessary (see the inline note).
4. **`config_with_tracing`** returns a plain `GatewayConfig()` with nothing tracing-related, and the `app` fixture docstring says "with tracing enabled" though nothing enables tracing. Rename or inline.
5. **`@pytest.mark.asyncio` on `test_context_propagation_with_span_generation`**, which awaits nothing. It also asserts OpenTelemetry's own `set_span_in_context` behavior rather than the gateway's, so it passes with this PR fully reverted.
6. **Inconsistent span-context access** between the two new files: `spans[0].get_span_context()  # type: ignore[no-untyped-call]` versus `provider_span.context.trace_id`. The latter needs no ignore.
7. **The new comment** `# Middleware stack is registered in reverse order (last-added runs first)` states a Starlette fact rather than the intent. What matters is that this middleware sits inside `CORSMiddleware` (so a preflight is answered without attaching) and outside anything that emits spans. `InFlightMiddleware` and `MetricsMiddleware` are registered after it and so run outside the attached context, which is harmless now since neither emits spans, but worth a word for whoever adds an instrumented middleware next.
8. **Duplicate `traceparent` headers are accepted.** Starlette's `Headers.get` returns the first value, so `len(header) == 1` upstream and the header is honored, where W3C says a request carrying two should be treated as carrying none. Very low impact.
9. **Placement.** These are app-level HTTP tests in `tests/unit/`, which `AGENTS.md` scopes to `tests/integration`. Soft, since the root conftest explicitly documents that the unit suite builds apps too, but the genuinely unit-level part is the carrier-extraction function, and the `./otari.db` note below is the concrete cost.
10. **Two follow-ups worth naming in the PR body** (outside #617's scope, not blockers): the gateway opens no SERVER span for the request itself, and there is no outbound `inject(` anywhere in `src/`, so the trace stops at Otari unless the operator also instruments httpx. That last point makes one docs sentence inaccurate; see the inline note.

### Checked and clean

No budget, reservation, or metering path touched, so there is nothing to reconcile or refund. No tenant-isolation or `resolve_user_id` surface. No SQL, no N+1, no new list endpoint, no migration, no entity. No OpenAPI-affecting change. No secrets or payloads logged. `scripts/check_architecture.py` passes and the module sits at `src/gateway/` alongside `inflight.py` and `metrics.py`, which is the right home. No M4 ledger entry looks required: no table, no control-plane contract, no capability the platform also has, and not mode-specific.

tests/conftest.py
```diff
@@ -66,3 +72,46 @@ def _reset_default_pricing() -> Generator[None, None, None]:
...
+        context is present.
+    """
+    context = _extract_otel_context_from_carrier(carrier)
+    if context is None:
+        return None
+
+    return trace.get_current_span(context)
+
+
+def _extract_otel_context_from_carrier(carrier: Mapping[str, str]) -> Context | None:
```
dpoulopoulos: **[major]** `_extract_otel_context_from_carrier` here is a near-verbatim fork of `_extract_context_from_carrier` in `src/gateway/context_propagation.py` (same propagator call, same `INVALID_TRACE_ID` guard), and commit `91720ff` repointed the five `test_extract_trace_context_from_carrier_*` tests from the production function to this copy. Those tests now assert on the fork, so the shipped function's invalid-format, all-zero-trace-id and all-zero-span-id handling has no direct coverage: deleting the guard in `context_propagation.py` leaves the whole suite green. Prefer making the production helper public (e.g. `extract_trace_context`) and importing it in the tests, rather than duplicating the logic. A helper in the root `conftest.py` imported as `from conftest import ...` also relies on pytest's rootdir sys.path insertion, which is fragile compared to a `tests/helpers.py` module.

pyproject.toml
dpoulopoulos: **[minor]** `opentelemetry-instrumentation` is never imported anywhere in `src/`, `tests/` or `scripts/`. The propagator this PR uses (`opentelemetry.trace.propagation.tracecontext`) and `opentelemetry.context` both ship in `opentelemetry-api`, which was already a dependency; I confirmed at runtime that the module resolves to the `opentelemetry-api` 1.39.1 install. This line adds a beta-versioned package (0.60b1 resolves today) to every runtime install of a published distribution for no benefit. Drop it, or, if the intent was to run the gateway under `opentelemetry-instrument` auto-instrumentation, note that `OpenTelemetryMiddleware` already performs this extraction and this middleware would then be redundant.

tests/unit/test_context_propagation.py
dpoulopoulos: **[minor]** Roughly 140 of the 334 lines here (`test_traceparent_header_is_extracted_from_request`, the valid/invalid `traceparent` parametrizations, `test_missing_traceparent_header_behaves_normally`, `test_traceparent_with_trace_state_extension`, the zero-ID cases, `test_multiple_requests_maintain_independent_contexts`) assert only that `GET /health` returns 200. Every one of them passes with the middleware deleted, so they document intent but do not guard it. Only `test_http_request_span_inherits_traceparent` and `test_streaming_response_span_inherits_traceparent` actually exercise the feature. Consider collapsing the status-code tests into a couple of cases that assert the resulting span's trace ID (a handler that reports `trace.get_current_span()` makes that a one-line change per case), and asserting a *new* root trace ID for the invalid/missing cases rather than just a 200.

src/gateway/context_propagation.py
dpoulopoulos, line 70: **[minor]** The middleware trusts client-supplied trace context unconditionally, on every HTTP request, with no way to turn it off. Otari is an internet-facing, API-key-authenticated gateway, so any key holder can now pick the trace ID their requests are filed under (joining or polluting an unrelated trace) and can force `sampled=1` to drive telemetry volume and cost, which I confirmed propagates through to handler spans. The docs say this is "intended for backend and service-to-service callers," but nothing enforces that. Accepting inbound context at the edge is normal OTel behavior, so this is a design call rather than a defect, but for a multi-tenant gateway it is worth a `GatewayConfig` toggle (default on or off, your call) so an operator exposing the gateway publicly can decline to trust caller-supplied context.

src/gateway/context_propagation.py
```diff
@@ -0,0 +1,93 @@
...
+from opentelemetry import context as otel_context
+from opentelemetry import trace
+from opentelemetry.context import Context
+from opentelemetry.trace.propagation.tracecontext import TraceContextTextMapPropagator
+from starlette.datastructures import Headers
+from starlette.types import ASGIApp, Receive, Scope, Send
+
+from gateway.log_config import logger
+
+if TYPE_CHECKING:
```
dpoulopoulos: **[nit]** `ASGIApp`, `Receive`, `Scope` and `Send` are imported at runtime on line 21 and then re-imported inside the `if TYPE_CHECKING:` block on lines 25-26. The second import is dead; drop either it or the runtime one (with `from __future__ import annotations` already in place, the `TYPE_CHECKING`-only form is the one the repo tends to prefer).

src/gateway/context_propagation.py
```diff
@@ -0,0 +1,93 @@
...
+        token = otel_context.attach(extracted_context)
+        try:
+            # Awaiting here spans the full response, including a streamed body,
+            # so the context stays attached until streaming completes.
+            await self.app(scope, receive, send)
+        finally:
+            otel_context.detach(token)
+
+
+def _extract_context_from_carrier(carrier: Any) -> Context | None:
```
dpoulopoulos: **[nit]** `carrier: Any` gives up the type information the caller actually has. The only call site passes `starlette.datastructures.Headers`, and the propagator needs a `Mapping[str, str]`; typing it as such (the way the `tests/conftest.py` copy already does) keeps mypy strict useful here.

src/gateway/context_propagation.py
dpoulopoulos: **[nit]** This `except Exception` is effectively unreachable: `TraceContextTextMapPropagator.extract` already swallows malformed headers internally and returns the passed-in context, which the `INVALID_TRACE_ID` check on the next lines catches. Harmless as belt-and-braces, but the `# noqa: BLE001` suppresses a rule this repo does not enable (`[tool.ruff.lint] select = ["E", "F", "I", "PERF"]`), so it reads as copied from another project.

tests/integration/test_context_propagation.py
```diff
@@ -0,0 +1,67 @@
...
+                "model": "ollama:test-model",
+                "messages": [{"role": "user", "content": "Hello"}],
+                "user": test_user["user_id"],
+            },
+        )
+
+    assert response.status_code == 200, response.text
+    spans = exporter.get_finished_spans()
+    provider_span = next(span for span in spans if span.name == "provider_request")
+    assert format(provider_span.context.trace_id, "032x") == "4bf92f3577b34da6a3ce929d0e0e4736"
```
dpoulopoulos, line 66: **[nit]** File ends without a trailing newline, and `tests/unit/test_context_propagation.py` has 21 lines with trailing whitespace. Ruff's selected rule set (`E`, `F`, `I`, `PERF`) does not include `W291`/`W292`, so CI is green, but both files are off the formatting the rest of the suite uses.

dpoulopoulos (commented):
Adds a pure-ASGI `TraceContextPropagationMiddleware` that extracts W3C `traceparent`/`tracestate` from incoming requests and attaches the OpenTelemetry context for the duration of the request (including streamed response bodies), plus docs and unit/integration tests. I verified the behavior at runtime against this checkout: valid context propagates trace ID, parent span ID and `tracestate` into handler-created spans; missing/garbage headers degrade gracefully; the context is detached after the response; and streaming chunks still see the incoming trace ID.

### General notes
- **[minor]** `src/gateway/main.py:520` — `traceparent`/`tracestate` are missing from the CORS `allow_headers` list. I verified the effect with `cors_allow_origins` configured: an `OPTIONS` preflight carrying `Access-Control-Request-Headers: traceparent` is rejected with **400 `Disallowed CORS headers`**, while the same preflight with `authorization` returns 200. So a cross-origin browser client whose OpenTelemetry fetch instrumentation adds the header has its request blocked outright, not merely untraced, which is stronger than what `docs/external-usage.md:192-197` says ("cannot propagate these headers"). Not a regression from this PR, but adding the two names here is a two-line fix that makes the newly documented feature work for browser callers too; otherwise, tighten the doc wording to say the preflight fails.

pyproject.toml
javiermtorres: @copilot remove the intrumentation package. The server will be manually configured, since we generate explicit spans for AI related events instead of generic HTTP traffic events.

src/gateway/context_propagation.py
javiermtorres: @copilot the only important baggage would be the one coming in the HTTP request. we assume this is the starting point for traces. add a comment or docstring to the tracing middleware class to clarify this if needed.

src/gateway/context_propagation.py
javiermtorres: add a logger warning in the case of otel trace errors first. then, use the configured global propagator. also, make the distinction between no existing trace information (no traceparent header, for example) and invalid trace information (traceparent header with an invalid value). the latter should be logged as an error, but the request should be allowed to proceed.

src/gateway/context_propagation.py
javiermtorres: @copilot won't the OTEL environment variables control the tracing details? the user deploying otari can make the decision there.

src/gateway/context_propagation.py
javiermtorres: Already addressed.

tests/unit/test_context_propagation.py
javiermtorres: Addressed.

javiermtorres:
@copilot check that all previous comments have been addressed

dni138 (approved):
## Approving

Re-reviewed against `e3e8577`, branch checked out locally, tests run, and the probes from the last round re-run. **All 2 majors and all 8 minors are addressed**, several of them more thoroughly than asked. `make lint`, `make typecheck` clean (370 files), `tests/unit/test_context_propagation.py` 19 passed, `tests/integration/test_context_propagation.py` + `test_cors_configuration.py` 5 passed, and all 8 CI checks green.

| Was | Now |
| --- | --- |
| **major** unused `opentelemetry-instrumentation` | Dependency dropped from `pyproject.toml` and out of the lock. One residual, see below. |
| **major** ambient context replaced, not extended | `propagate.extract(carrier=carrier, context=otel_context.get_current())`, with a comment saying why. Re-ran the probe: ambient baggage now survives the request (`{'tenant': 'acme'}` in and out), and the caller is still the parent. Exactly the fix. |
| **minor** hard-coded propagator | Now the configured global propagator, plus `test_extract_trace_context_uses_configured_global_propagator` to hold it there. One nuance residual, see below. |
| **minor** unreachable `except Exception` | Gone entirely, and the `opentelemetry.trace.span` log-noise point is now documented with mitigation advice. |
| **minor** implicit trust decision | `accept_incoming_trace_context`, **default off**, so it fails closed the way the repo asks new security-relevant flags to. Threat-model paragraph in the module docstring, a docs section, a `config.example.yml` entry, the dashboard settings group, and tests for both default-off and explicit-off. |
| **minor** tests asserted a copy of the helper | Production function promoted to `extract_trace_context` and imported by the tests; both conftest forks deleted. Confirmed the direction is right: the tests now fail if the shipped guard changes. |
| **minor** 21 of 23 cases passed without the feature | Rewritten around probe routes that assert real trace ids, with invalid inputs asserting a *new* root trace id rather than a 200. Shorter file, more coverage. |
| **minor** `detach` uncovered | `test_context_is_detached_after_request_completes` does precisely the traceparent-then-no-traceparent check. |
| **minor** unit test migrated `./otari.db` | `_test_config(tmp_path)` with a tmp database, `_env_file=None`, and an autouse fixture stripping `OTARI_*`. Verified: a clean run leaves no `otari.db`. The autouse `_reset_db_state` in the root conftest also closes the leaked-engine half. |
| **minor** docs in the wrong page | Moved to `docs/configuration.md`, `docs/external-usage.md` reverted, and the inaccurate "preserved for downstream instrumentation" line replaced with an explicit statement that Otari does not inject propagation headers outbound. |

Nits 1 through 6, 9 and 10 are all cleared too: no trailing whitespace, final newlines present, `TYPE_CHECKING`-only starlette import, `carrier: Mapping[str, str]`, the misnamed fixture and the `asyncio`-marked SDK test both gone, and the `# type: ignore[no-untyped-call]` no longer needed. The CORS regression test is a nice addition that turns a docs claim into something enforced.

### Three nit-level residuals, none blocking

1. **`uv.lock` still carries the `wrapt` downgrade the removed dependency caused.** Against the true merge base (`3ef8ac8`), the *entire* package-level lockfile diff on this branch is `wrapt 2.1.2 -> 1.17.3`, and nothing left in the tree requires `<2.0` (only `testcontainers` consumes it, and it resolves 2.1.2 on `main`). Since the PR no longer changes any dependency, its lockfile diff should be zero. Worth knowing: plain `uv lock` is a no-op here, it preserves the existing pin. `git checkout origin/main -- uv.lock` is the precise fix; I confirmed `uv lock --check` then passes against this branch's `pyproject.toml` and the diff versus the merge base becomes empty. (`uv lock --upgrade-package wrapt` also clears it, landing 2.3.0.) No functional impact, purely lockfile hygiene, but it is one command and it removes the last unexplained change from the diff.

2. **The baggage nuance from the propagator thread is still open.** The `INVALID_TRACE_ID` guard returns `None` for a context that carries baggage but no span, so baggage arriving without a `traceparent` is discarded. Verified in a clean process (the earlier attempt was contaminated by a prior `attach`):

   ```
   {'baggage': 'tenant=acme,region=eu'}                  -> None   (baggage dropped)
   {'traceparent': '00-4bf9...-01', 'baggage': 'tenant=acme'} -> Context, baggage carried
   ```

   The docs now list `baggage` as part of the default propagator set and describe it as carrying metadata across service boundaries, so there is a small gap between what they promise and what a baggage-only request gets. Narrow in practice, since an instrumented caller sends both headers, which is why this is a nit rather than a re-request. Either widen the guard to "attach whenever the extracted context differs from the current one", or say in the docs that baggage rides along with a valid trace context.

3. **`accept_incoming_trace_context` is missing from the config table** in `docs/configuration.md`. `enable_metrics` and `enable_docs` each have a row (lines 79 and 80); the new flag appears only in the prose section below, so someone scanning the table for available settings will not find it.

Two nits from the previous round were left as-is and I am not asking for them: the middleware-order comment still states the Starlette rule rather than the intent, and duplicate `traceparent` headers are still accepted where W3C says to treat them as absent. The second matters much less now that the feature is opt-in and scoped to trusted callers.

Nice work on the turnaround, particularly choosing to fail closed on the config flag rather than defending the unconditional behavior.

src/gateway/context_propagation.py
```diff
@@ -0,0 +1,108 @@
...
+        token = otel_context.attach(extracted_context)
+        try:
+            # Awaiting here spans the full response, including a streamed body,
+            # so the context stays attached until streaming completes.
+            await self.app(scope, receive, send)
+        finally:
+            otel_context.detach(token)
+
+
+def extract_trace_context(carrier: Mapping[str, str]) -> Context | None:
```
dpoulopoulos: **[major]** The review instruction on the propagator thread was: "add a logger warning in the case of otel trace errors ... make the distinction between no existing trace information (no traceparent header, for example) and invalid trace information (traceparent header with an invalid value). The latter should be logged as an error, but the request should be allowed to proceed." The module still has no logger and no such distinction: both cases fall into the same `return None`. Verified against a running gateway from this branch: `curl -H 'traceparent: garbage'` returns a fresh root trace and the gateway logs nothing.

Note the distinction has to be drawn from the carrier itself, not from the guard: now that extraction is seeded with `otel_context.get_current()`, a request under an ambient span yields a valid trace id even when the header is garbage, so the `INVALID_TRACE_ID` check can never tell "invalid header" from "no header".

src/gateway/context_propagation.py
```diff
@@ -0,0 +1,108 @@
...
+        extraction failed or no valid context is present.
+    """
+    # Extract into the current context so we preserve any existing baggage,
+    # suppress-instrumentation flags, or enclosing spans that may already
+    # be set (e.g., by an outer instrumentation layer).
+    context = propagate.extract(
+        carrier=carrier, context=otel_context.get_current()
+    )
+
+    if trace.get_current_span(context).get_span_context().trace_id == trace.INVALID_TRACE_ID:
```
dpoulopoulos: **[major]** Baggage-only requests silently lose their baggage. The default propagator set is `tracecontext,baggage` and `docs/configuration.md:141` documents `baggage` as extracted, but this guard returns `None` whenever no valid *span* context was extracted, so `attach()` never runs and extracted baggage is discarded.

Verified live against this branch: `baggage: tenant=acme,region=eu` with no `traceparent` reaches the handler as `{}`; the same header alongside a valid `traceparent` reaches it as `{'tenant': 'acme'}`.

This is the nuance raised last round ("the guard has to change from 'no valid span means return None' to 'attach whenever anything was extracted'"), and the reply that "the only important baggage would be the one coming in the HTTP request" reads as agreement that it should flow. Consider gating on "the extracted context differs from the current one" instead of span validity.

dpoulopoulos (changes requested):
Adds a pure-ASGI `TraceContextPropagationMiddleware` that extracts incoming OpenTelemetry context (W3C `traceparent`/`tracestate` by default, via the globally configured propagator) and attaches it for the life of the request, including a streamed body, behind a new default-off `accept_incoming_trace_context` config flag surfaced read-only in the dashboard's config viewer, with unit + integration tests and an operator-facing `docs/configuration.md` section.

Most of the previous round's feedback is genuinely addressed: the middleware is now raw ASGI with detach-after-streaming, extraction seeds from `otel_context.get_current()` so ambient context survives, the global propagator is used instead of a hard-coded `TraceContextTextMapPropagator`, the duplicated conftest extract helper is gone and the unit tests import the shipped `extract_trace_context`, the status-code-only tests were replaced with real span-probe assertions (plus detach-isolation and streaming cases), the tests use a `tmp_path` database and scrub `OTARI_*` env, `opentelemetry-instrumentation` was dropped from `pyproject.toml`, the docs moved to `configuration.md` with the incorrect "downstream" claim corrected, CORS was explicitly declined and documented with a regression test, and the trust decision became an explicit default-off flag. Four items are still open: the author's own request for logging that distinguishes missing from invalid incoming context, the baggage-drop nuance raised in the last round, a leftover `uv.lock` downgrade from the removed dependency, and the suggestion to assert the incoming parent span id.

src/gateway/context_propagation.py
javiermtorres: @copilot the otel sdk itself will issue a warning in the case the trace is not valid, right?

javiermtorres: @copilot true, but no exception will be thrown in the case of invalid traces, so no logging is possible here from our side

src/gateway/context_propagation.py
javiermtorres: @copilot is there a way to tell an invalid traceparent from a missing traceparent with the existing otel propagators?

javiermtorres: @copilot We will remove this line and the following and rely on the propagator, then. We can check afterwards if we really need to cover this case.

tbille (approved):

dpoulopoulos (approved):

PR #630, opened 2026-08-18
feat(reprise): per-format normalization hooks for the v0 fingerprint

dpoulopoulos:
## Description

The v0 fingerprint has to group the same automation together whichever wire format it arrived through. Several of its inputs are currently unobtainable in a format-neutral way, and each one silently halves or destroys support if the implementer guesses. This adds the normalization hooks the fingerprint reads from, and nothing else: **nothing calls any of them from a request path yet.** Computing the fingerprint is mozilla-ai/otari-ai#1648; emitting the record is mozilla-ai/otari-ai#1484.

Five commits, each closing one gap.

**Purpose hints have three authors and only one is the customer.** Hints come out in the order the request body listed the servers, the preamble header is a deployment knob (`OTARI_TOOLS_HEADER`), and a hint the caller never supplied resolves to whatever the dashboard, env, or YAML holds. Hashing the block as assembled means an operator editing a text field, or a release rewording a built-in, resets every pile in the deployment mid-measurement with nobody having touched the automation. `normalize_purpose_hints` sorts the pairs, excludes the header, and replaces an undeclared hint with the literal sentinel `"<default>"`. The injected block keeps the caller's order, because that is what the model sees.

**The caller's prompt lives in three places and is rewritten before dispatch.** `client_system_prompt` and `first_user_message` land on `FormatAdapter` because `inject_hints` runs at the route layer and all three adapters copy before injecting, so that is the last point at which `call_kwargs` still holds the caller's original text. Hashing the *injected* result is not available: the gateway prepends the block for two formats and appends it for the third, so the same automation through two SDKs would never group. Both accessors normalize rather than merely read, so a string `system` and an equivalent single text block agree, a `developer`-led chat request agrees with a `system`-led one, and a `cache_control` marker (which moves money, not meaning) changes nothing.

**Tool definitions are format-shaped.** `normalize_tools` unwraps nested `function.parameters`, `input_schema`, and flat `parameters` into one triple, feeding two values that move for unrelated reasons: `tool_set_hash` (sorted names, a fingerprint input) and `tool_definitions_hash` (names, descriptions, canonicalized schemas; payload only). They are split because both description and schema arrive in the same live `list_tools()` response and `MCPClientPool` is rebuilt per request, so an upstream server adding one optional property, backward compatibly and breaking nothing, would otherwise split every pile in the workspace from that moment on.

**Usage field names and semantics differ, and the cache term is captured nowhere.** OpenAI's `prompt_tokens` includes cached tokens; Anthropic's `input_tokens` excludes them; Responses has no cache field at all. Since the saving estimate v0 exists to produce is mostly the cache re-read, recording "whatever this provider calls input" understates the number in the direction that cancels the project. `usage_snapshot` / `stream_usage_snapshot` return a `GatewayUsage` rather than a bespoke triple: the carrier already holds cache reads, cache writes, the 1-hour write subset, and `cache_tokens_in_prompt`, which is the per-format mapping expressed as data rather than as a table. A hand-written triple would drop cache writes, the priciest input class (1.25x base input at 5m, 2x at 1h, against 0.1x for a read), and would need editing for every new any-llm provider.

Streaming returns the same populated snapshot, not a partial one. Anthropic splits a round across `message_start` (input and cache) and `message_delta` (cumulative output), and both reach `observe` even on a continuing round, where the strategy defers rather than drops them. The one genuine gap is a chat caller who explicitly sets `include_usage: False`: there the provider sends nothing and the snapshot is `None` rather than a zero-filled carrier, because zero and unknown are different answers and mozilla-ai/otari-ai#1488 has to tell them apart.

The snapshots read the provider's own fields and do **not** touch the loop's usage accumulators, which exist to fold totals into the client response and keep doing exactly that. A test asserts that.

### Two things worth a reviewer's attention

**The hooks are on the Protocols, not duck-typed**, so `mypy --strict` names any format that forgot one. `normalize_tools` and `usage_snapshot` sit on `ToolLoopStrategy` even though nothing in the loop calls either; the Protocol docstring says so out loud, because the merged tool list is built there and the raw provider result is only in scope there.

**Chat completions still assembles the hint block inline** in `inject_purpose_hints`, duplicating `build_hint_block`. Unifying them means moving code between `mcp_loop` and `tool_format` (the import already runs one way, `tool_format` → `mcp_loop`), and this PR is meant to change no behavior. Instead a test pins all three formats to one string, so a future divergence fails there rather than silently splitting `injected_block_hash` from what chat actually sent.

### One fix, from reviewing the diff

An adversarial review pass over the diff after opening found three defects, each fixed in `f685f47` with a test written failing first:

- **The Anthropic streaming snapshot understated input on a compaction round.** It took the whole input side from `message_start`, which is emitted before the compaction pass, so only the `message_delta` carries the `iterations` list that sums to the billed total. It reported output summed over every iteration and input from only the first: 42 in against the 142 the gateway bills for the exact stream `tests/integration/test_messages_streaming_usage.py` already pins. Now reconciled per field, last non-zero winning, the convention `gateway.streaming._merge_usage` already applies to this stream.
- **Every Responses provider-native tool normalized to an empty name.** Anthropic's native tools carry `name`; the Responses ones carry none, since identity lives in `type`. So `web_search`, `file_search`, and `mcp` shared one `tool_set_hash`. Name now falls back to `type`.
- **A Responses round ending in `response.incomplete` reported unknown usage.** Truncation by `max_output_tokens` or a content filter is billed, but the snapshot read only `response.completed`, so a paid round returned the value reserved to mean "the provider reported nothing".

Two of the three were untested cases rather than mistyped code: the acceptance criterion "streamed round == non-streamed round" passed while being false, because the fixture carried no `iterations`.

### One refactor, separated

`c96fb87` is a pure move: the Anthropic usage readers (`_billable_messages_usage` including the compaction `iterations` sum, `_cache_write_1h_tokens`) and the Responses one (`_usage_to_completion_usage`) go from their route modules into `core/usage.py`. The tool-loop strategies need the same readings and a service may not import the API layer, so `core` is the only place both sides can reach, and it already owns `GatewayUsage`. Same code, same declared types, no behavior change.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [x] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Closes mozilla-ai/otari-ai#1647

## Checklist

- [x] I ran the Definition of Done checks locally (`make lint`, `make typecheck`, `make test`). Also `make openapi-check`.
- [x] Documentation was updated where necessary. No user-facing surface changed, so no doc change; the reasoning lives in the module and Protocol docstrings, which is where the next reader of these hooks will be.
- [x] If the API contract changed, I regenerated the OpenAPI spec (`uv run python scripts/generate_openapi.py`). No contract change; `make openapi-check` clean.
- [x] If this PR adds or changes a persistent table, a control-plane contract the otari.ai platform consumes or serves, a capability the platform also has, or a mode-specific surface, I appended an entry to the M4 reconciliation ledger ([otari-ai#1587](https://github.com/mozilla-ai/otari-ai/issues/1587)). Entry to follow on this PR, against the **Usage and observability** row; target disposition unchanged, since these hooks add no second accounting implementation.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

Implementation and tests written by Claude from the issue body, test-first. A multi-agent adversarial review over the diff ran after opening; it raised 12 findings, 6 survived independent refutation, and those were the 3 distinct defects fixed in `f685f47` (the streaming-compaction one was found by four separate lenses). The agent also ran `make lint`, `make typecheck`, the full unit suite, and the integration suite locally.

**NOTE:**

### DIFF
--- a/src/gateway/core/observation.py
+++ b/src/gateway/core/observation.py
@@ -58,13 +58,22 @@ def normalized_tool(definition: dict[str, Any], schema_key: str) -> NormalizedTo
 
     ``schema_key`` names where the format keeps the JSON Schema (``parameters``
     for chat completions and Responses, ``input_schema`` for Anthropic Messages).
-    An entry carrying neither a description nor a schema, such as a provider-native
-    server tool the gateway never converts, still contributes its name, which is
-    what :func:`tool_set_hash` counts.
+
+    The name falls back to ``type`` when the entry has none. A provider-native
+    server tool the gateway never converts still has to contribute an identity,
+    which is what :func:`tool_set_hash` counts, and only some of them spell it
+    ``name``: Anthropic's carry one (``{"type": "web_search_20250305", "name":
+    "web_search"}``) while the Responses ones carry none at all (``{"type":
+    "web_search"}``, ``{"type": "file_search", ...}``). Reading ``name`` alone
+    collapses every Responses server tool onto the empty string, so a workspace
+    swapping one for another would fingerprint as an unchanged tool set. What such
+    an entry configures (an ``mcp`` tool's ``server_label``, a ``file_search``
+    tool's vector stores) stays out of both hashes, consistent with a key made of
+    names.
     """
     schema = definition.get(schema_key)
     return NormalizedTool(
-        name=str(definition.get("name") or ""),
+        name=str(definition.get("name") or definition.get("type") or ""),
         description=str(definition.get("description") or ""),
         input_schema=schema if isinstance(schema, dict) else {},
     )
--- a/src/gateway/core/observation.py
+++ b/src/gateway/core/observation.py
@@ -1,0 +1,133 @@
+"""Format-neutral vocabulary for the Reprise loop-round observation.
+
+Reprise groups repeated automation by a fingerprint computed over a handful of
+request properties (otari-ai#1482). The same logical request reaches the gateway
+through three wire formats, and a property read "however this format spells it"
+hashes differently in each, so one nightly automation driven through two SDKs
+would never group and the report would understate its own support. Every value
+feeding the fingerprint is therefore normalized to a format-neutral form before
+it is hashed; this module owns those forms and the hashes over them.
+
+The accessors that produce the values live where the format is known:
+``FormatAdapter`` (``gateway.api.routes._pipeline``) for the request fields only
+the route layer still holds unmodified, ``ToolLoopStrategy`` /
+``StreamToolLoopStrategy`` (``gateway.services._tool_loop``) for what the loop
+sees, and :func:`gateway.services.tool_format.normalize_purpose_hints` for the
+purpose-hint block, which is format-neutral already.
+
+Assembling the fingerprint out of these is otari-ai#1648 and emitting the record
+is otari-ai#1484, so nothing here is called from a request path yet.
+"""
+
+from __future__ import annotations
+
+import hashlib
+import json
+from collections.abc import Sequence
+from typing import Any, NamedTuple
+
+UNDECLARED_HINT = "<default>"
+"""Stands in for a purpose hint the caller did not declare in the request body.
+
+An undeclared hint resolves to whatever the deployment supplies (dashboard
+override, env, YAML, or a built-in default), so hashing the resolved text would
+let an operator editing a text field, or a gateway release rewording a built-in,
+reset every pile in the deployment with nobody having touched the automation.
+The sentinel records "the caller said nothing here", which is the property that
+stays true across such an edit. What the model actually saw is still recorded,
+unhashed, as ``injected_block_hash`` over the assembled block.
+"""
+
+
+class NormalizedTool(NamedTuple):
+    """One tool in the shape all three wire formats agree on.
+
+    The formats wrap the same JSON Schema object differently (nested
+    ``function.parameters`` for chat completions, ``input_schema`` for Anthropic
+    Messages, flat ``parameters`` for Responses), so each loop strategy unwraps
+    its own shape into this triple.
+    """
+
+    name: str
+    description: str
+    input_schema: dict[str, Any]
+
+
+def message_text(content: Any) -> str:
+    """Flatten a message-content value to the text the model was shown.
+
+    Covers the two legal shapes every format has for a prompt: a plain string,
+    and a list of content blocks (dicts or SDK objects) whose text is joined with
+    a blank line, so ``"foo"`` and ``[{"type": "text", "text": "foo"}]`` agree.
+    A block carrying no text (an image, an unknown block type) contributes
+    nothing, and ``None`` flattens to ``""``, so absent and empty normalize alike.
+
+    ``cache_control`` markers are ignored for free, since only ``text`` is read:
+    they move money rather than meaning, and a caller adding one mid-run must not
+    split its own pile.
+    """
+    if isinstance(content, str):
+        return content
+    if isinstance(content, list | tuple):
+        return "\n\n".join(text for block in content if (text := _block_text(block)))
+    return ""
+
+
+def _block_text(block: Any) -> str:
+    """The text carried by one content block, or ``""`` when it carries none."""
+    if isinstance(block, str):
+        return block
+    text = block.get("text") if isinstance(block, dict) else getattr(block, "text", None)
+    return text if isinstance(text, str) else ""
+
+
+def text_hash(text: str) -> str:
+    """sha256 of one normalized text value.
+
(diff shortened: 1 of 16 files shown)


src/gateway/core/observation.py
```diff
@@ -0,0 +1,133 @@
...
+    nothing, and ``None`` flattens to ``""``, so absent and empty normalize alike.
+
+    ``cache_control`` markers are ignored for free, since only ``text`` is read:
+    they move money rather than meaning, and a caller adding one mid-run must not
+    split its own pile.
+    """
+    if isinstance(content, str):
+        return content
+    if isinstance(content, list | tuple):
+        return "\n\n".join(text for block in content if (text := _block_text(block)))
```
aittalam, line 72: Is this a recognised standard? Do you have a link documenting it?

src/gateway/services/mcp_loop_messages.py
```diff
@@ -399,6 +405,11 @@ def fold_usage(self, result: MessageResponse, acc: _MessagesUsageAccumulator) ->
             except (AttributeError, TypeError):
                 logger.warning("Could not add native web-search blocks to the response content")
 
+    def usage_snapshot(self, result: MessageResponse) -> GatewayUsage | None:
+        if result.usage is None:
+            return None
+        return anthropic_billable_usage(result.usage)
```
aittalam, line 411: Seeing how this is implemented for chat completions (`return GatewayUsage.from_completion_usage(state.usage)`) I wanted to understand why we do not have a `from_responses_usage` or something like that and checked the code in core/usage.py. I [found that](https://github.com/mozilla-ai/otari/blob/c96fb87c190bbc2d9b3fade51a44371e5b3c17b5/src/gateway/core/usage.py#L22) `GatewayUsage` is a subclass of any_llm's `CompletionUsage`, and that there are separate classes for e.g. responses and openresponses. 
I think it's ok to decide "let's translate everything into a completions-like format", but I wonder whether the current implementation could be improved to be more readable / maintainable.

src/gateway/core/observation.py
```diff
@@ -58,13 +58,22 @@ def normalized_tool(definition: dict[str, Any], schema_key: str) -> NormalizedTo
...
-    An entry carrying neither a description nor a schema, such as a provider-native
-    server tool the gateway never converts, still contributes its name, which is
-    what :func:`tool_set_hash` counts.
+
+    The name falls back to ``type`` when the entry has none. A provider-native
+    server tool the gateway never converts still has to contribute an identity,
+    which is what :func:`tool_set_hash` counts, and only some of them spell it
+    ``name``: Anthropic's carry one (``{"type": "web_search_20250305", "name":
+    "web_search"}``) while the Responses ones carry none at all (``{"type":
+    "web_search"}``, ``{"type": "file_search", ...}``). Reading ``name`` alone
```
aittalam, line 67: Is there any documentation (for anthropic/responses) we can point to so we can ground these into actual implementation details of the different protocols? It's also ok to keep it simple and empirical now, but in the long term I think it'd be better if we grounded this in the standards (e.g. checking whether there's a schema that says that at least `type` has to be always not null)

src/gateway/services/mcp_loop_messages.py
```diff
@@ -636,11 +636,22 @@ def accumulate_stream_usage(self, acc: _MessagesStreamAccumulator, state: _Messa
...
+        input fields on the delta too. Picking ``message_start`` for the whole input
+        side would then report a compaction round's output summed over every
+        iteration and its input from only the first, understating the cache re-read
+        the estimate is mostly made of, and disagreeing with what the same stream is
+        billed.
+
+        So the two are reconciled per field, last non-zero winning, which is the
+        convention ``gateway.streaming._merge_usage`` already applies to this stream
+        for the client-facing totals. Both events reach ``observe`` even on a
+        continuing round, where they are deferred rather than dropped.
```
aittalam, line 654: I have no idea what this comment is saying 😅

aittalam (commented):
TY @dpoulopoulos ! I left a few comments around.
Overall I think the code is ok, the only two main concerns I had are:
1. how usage class(es) are organized between completions, responses, messages
2. some generated comments which I had a very very hard time understanding.

For the former, it looks like this comes from previous code organization, so feel free to check it out and postpone it to a follow-up PR to keep this one light. For the latter, I believe the main issue is that the model knows things from other parts of the code or from previous iterations which we don't, so it should either add proper references if they are available in the current code or drop them when they are just considerations from previous failed attempts.

PR #633, opened 2026-08-18
feat(tenancy): give the tenancy model a request plane

njbrake:
## Description

Three things that give the tenancy model a request plane.

**A member can hold a key.** Keys, budgets and usage attach to the gateway's string-keyed `users` row; a member is a UUID identity in `user`. Nothing joined them, so a member could be created and then given nothing. Creating a tenancy identity now mints the matching `users` row, keyed on the identity's UUID, and `attribution_user_id` carries it to the dashboard. Forward-compatible with M4: otari-ai's `GatewayIdentityBridge` keys idempotency off the same legacy id.

**The request plane carries a workspace.** `api_keys`, `usage_logs`, `model_aliases` and `routing_policies` gain a NOT NULL `workspace_id`, backfilled onto a default the migration seeds under the slug provisioning already looks up, so first boot adopts those rows rather than minting a second default. The workspace is read off the authenticating key, never a header. It is an optional filter on the four list endpoints, and on `UsageSelection`, so a filtered "delete all N matching" cannot reach another workspace's rows.

**Tenancy scopes get their own budgets.** A fifth persistent table, `scoped_budgets`, and five endpoints under `/v1/scoped-budgets`, cap an organization, a workspace, either kind of membership, or a single API key, optionally narrowed to one provider. It sits beside `budgets` rather than replacing it, and both are enforced: `budgets` is many-to-one from `users` and is checked against `users.spend + users.reserved`, so folding counters onto it would silently turn N users each getting a limit into a pooled cap. Reservation stays lock-free, one conditional UPDATE per ceiling, taken in one total order and compensated when a later ceiling refuses.

The membership context also carries the caller's own workspaces, so the dashboard shell can pick a context without a second round trip.

Known limits of the scoped ceilings, written down in `docs/access-control.md`: a provider-narrowed ceiling is resolved from the provider a routing policy names first, so a fallover to a second provider neither checks nor charges that provider's ceiling; and batch, external and imported spend settle against the per-user counters alone, because those paths have no request scope to resolve a workspace from.

Deliberately not done, with reasons in the commit bodies: alias and policy name uniqueness stays deployment-wide, and provider and search-tool credentials stay process-wide config.

## PR Type

- [x] New Feature

## Relevant issues

Part of #588. Unblocks the workspace switcher on #626.

## Checklist

- [x] Documentation was updated where necessary (`docs/access-control.md`, `docs/api-reference.md`, `src/gateway/AGENTS.md`).

The ledger entry applies (four persistent tables change shape, `scoped_budgets` is a new one, and this is a capability the platform also has) and is owed. It is not appended yet, and folds into the entry drafted for #625 and #626 covering the same slice.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

The scope and the sequencing are njbrake's; the code and prose are the model's. This description was written by a different session from the one that wrote the code, so treat it as a summary of the diff rather than the author's own account.

## Summary

- Added workspace ownership to request-plane resources and usage records.
- Linked tenancy members to gateway users for key ownership and usage attribution.
- Added workspace context to authenticated membership data and relevant list, usage, and export endpoints.
- Added scoped budgets for organizations, workspaces, memberships, API keys, and providers.
- Enforced scoped budgets alongside existing per-user budgets.
- Added workspace-aware migrations, provisioning, seed data, deletion safeguards, documentation, API clients, and tests.
- Preserved standalone-only behavior for scoped-budget management routes.

## Technical notes

- Workspace deletion returns `409` when dependent request-plane records remain.
- Provider fallback and request paths without workspace scope remain known limitations.
- The M4 reconciliation ledger entry remains outstanding.


khaledosman (commented):
Review of the workspace-scoped request plane and the scoped-budget mechanism. The reserve/settle/refund lifecycle is threaded correctly through both mechanisms, including the compensation on a per-user refusal. The findings are in the process-global workspace cache; the first one makes the `create_all` fallback path fail closed in a way that persists for the process lifetime.

src/gateway/services/workspace_scope.py
```diff
@@ -0,0 +1,145 @@
...
+    ).scalar_one_or_none()
+    if resolved is None:
+        # Fall back to any workspace before creating one: an operator may have
+        # renamed the default, and minting a second beside it would be worse than
+        # billing to the one that exists.
+        resolved = (await db.execute(select(col(Workspace.id)).order_by(col(Workspace.created_at)))).scalars().first()
+    if resolved is None:
+        resolved = await _create_default_workspace(db)
+
+    _default_workspace_id = resolved
```
khaledosman: The cache is populated from a row that is only flushed, and `_create_default_workspace` never commits, so on a `create_all`-built schema this poisons itself on first use.

Path: the first master-key completion reaches `_pipeline.log_usage` → `workspace_for_key_id(db, None)` → `default_workspace_id` → `_create_default_workspace`, which flushes and returns the id. `log_usage` then hands the row to `LogWriter`, which opens its own session (`services/log_writer.py:34`), and the request session closes with no commit, rolling the organization and workspace back. `_default_workspace_id` keeps their id for the rest of the process, so every later key create, alias, policy and usage row fails its `workspace.id` foreign key, and the usage failures are swallowed by `SingleLogWriter`, so it is silent.

Commit inside `_create_default_workspace`, or only cache after the caller's commit. As written, the fallback the docstring says exists for non-Alembic bootstraps does not work on the path that reaches it.

src/gateway/services/workspace_scope.py
```diff
@@ -0,0 +1,145 @@
...
+)
+
+_default_workspace_id: uuid.UUID | None = None
+# api_key id -> its workspace. A key belongs to exactly one workspace and never
+# moves, so this cannot go stale; it exists because a usage row is written on the
+# hot path and must not pay a lookup for something immutable.
+_key_workspace: dict[str, uuid.UUID] = {}
+
+
+def reset_default_workspace_cache() -> None:
```
khaledosman: This clears one process. `tenancy/workspace_service.py:283` calls it after deleting an empty default workspace, but with more than one uvicorn worker or replica every other process keeps the stale id and 500s on the foreign key until restart, which is the exact failure the reset exists to prevent.

Either drop the memoization of the default, or treat a foreign-key failure on `workspace_id` as a cache miss and re-resolve once.

src/gateway/services/workspace_scope.py
```diff
@@ -0,0 +1,145 @@
...
+            select(col(Workspace.id))
+            .join(Organization, col(Organization.id) == col(Workspace.organization_id))
+            .where(col(Organization.slug) == DEFAULT_ORGANIZATION_SLUG, col(Workspace.name) == DEFAULT_WORKSPACE_NAME)
+        )
+    ).scalar_one_or_none()
+    if resolved is None:
+        # Fall back to any workspace before creating one: an operator may have
+        # renamed the default, and minting a second beside it would be worse than
+        # billing to the one that exists.
+        resolved = (await db.execute(select(col(Workspace.id)).order_by(col(Workspace.created_at)))).scalars().first()
```
khaledosman: No `LIMIT`, and no tiebreaker. Every workspace id is fetched in order to take `.first()` in Python, and two workspaces sharing a `created_at` (which one transaction produces) make "where deployment-wide writes land" differ between processes. `.order_by(col(Workspace.created_at), col(Workspace.id)).limit(1)`.

src/gateway/services/scoped_budget_service.py
```diff
@@ -0,0 +1,399 @@
...
+                    and_(
+                        ScopedBudget.current_spend + ScopedBudget.reserved_spend < ScopedBudget.max_budget,
+                        ScopedBudget.current_spend + ScopedBudget.reserved_spend + amount <= ScopedBudget.max_budget,
+                    ),
+                ),
+            )
+            .values(reserved_spend=ScopedBudget.reserved_spend + amount)
+            .execution_options(synchronize_session=False)
+        )
+        await db.commit()
```
khaledosman, line 350: One commit per ceiling, in the loop, before the provider call. A request billing to five identities with an aggregate and a narrowed ceiling each pays up to ten commits on the hot path.

The separate commits buy nothing here: no lock spans the provider call either way, and the loop's compensating `release` already covers a mid-loop refusal. Issue the conditional UPDATEs in one transaction and commit once after the loop.

src/gateway/services/tenancy/errors.py
```diff
@@ -145,6 +145,25 @@ class ForeignTenancyError(TenancyError):
     status_code = status.HTTP_500_INTERNAL_SERVER_ERROR
 
 
+class WorkspaceInUseError(TenancyConflictError):
```
khaledosman, line 160: The RESTRICT guard covers the four request-plane tables but not `scoped_budgets`, whose `scope_id` is deliberately not a foreign key. A workspace holding only a ceiling therefore deletes cleanly and leaves an orphan row that lists fine and enforces nothing, which is the state `api/routes/scoped_budgets.py:125` `_require_scope_exists` says it exists to prevent, just at the other end of the lifecycle.

Delete the workspace's own and its memberships' ceilings in `delete_workspace`, or refuse while any exist.

alembic/versions/d5e7f1a2b3c4_workspace_scope_request_plane.py
```diff
@@ -0,0 +1,147 @@
...
+    for table in SCOPED_TABLES:
+        # Added NOT NULL with the default in one statement, which both engines
+        # accept on a populated table, so there is no add-then-backfill-then-
+        # tighten dance. It does not avoid SQLite's table rebuild, as an earlier
+        # draft of this comment claimed: the foreign key below needs one anyway
+        # (see the docstring). The partial indexes on the alias and policy tables
+        # survive that rebuild, which `test_tenancy_schema_chain` pins.
+        op.add_column(
+            table,
+            sa.Column("workspace_id", sa.Uuid(), nullable=False, server_default=sa.text(f"'{literal}'")),
```
khaledosman, line 142: The `server_default` stays on the column but `entities.py` declares none, so a migrated database has it and a `create_all` one does not. The docstring's reason for keeping it ("a write path that has not yet learned to carry a workspace lands its row in the default one rather than failing in production") holds on only half the deployments. Declare `server_default` on the four model columns too, or drop it here.

src/gateway/api/routes/scoped_budgets.py
```diff
@@ -0,0 +1,292 @@
...
+from gateway.models.entities import APIKey, ScopedBudget
+from gateway.models.tenancy import Organization, OrganizationMember, Workspace, WorkspaceMember
+
+# ``ScopeType`` comes from the service that resolves a scope, not from a copy
+# here: the ``Literal`` is what puts the allowed values in the OpenAPI schema,
+# and a second roster would eventually let a client create a scope enforcement
+# does not know.
+from gateway.services.scoped_budget_service import ScopeType
+
+router = APIRouter(prefix="/v1/scoped-budgets", tags=["scoped-budgets"])
```
khaledosman: Auth is repeated on all five decorators instead of declared here, which is the pattern `api/routes/organizations.py:45` argues against in a comment ("a future handler that did not would be unauthenticated with nothing to notice"). Move `dependencies=[Depends(verify_master_key)]` onto this `APIRouter(...)` call.

tbille (approved):

src/gateway/api/routes/_pipeline.py
```diff
@@ -1200,6 +1202,12 @@ async def resolve_request_context(
                 pricing_provider=gate_instance,
                 strategy=config.budget_strategy,
                 counts_toward_budget=not budget_exempt,
+                # The tenancy-scoped ceilings resolve from the key's workspace and
+                # the identity behind it, and from the provider this attempt is
+                # about to call. A fallover to a different provider keeps the
+                # ceilings resolved here: repricing changes the amount held, not
+                # which caps the request was admitted against.
+                scope=BudgetScopeRequest(api_key=api_key, provider_instance=gate_instance),
```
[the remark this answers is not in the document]
njbrake: Confirmed, and it is a wider hole than repricing. With an OpenAI head and an Anthropic fallover, the Anthropic ceiling is neither checked nor charged, which defeats the split cap `docs/access-control.md` uses as its own example. `resolve_request_context` resolves the scope from `plan.head`, and `increase_reservation` reuses `handle.scoped_budgets` by design, so nothing downstream can recover it. A selector the gate cannot resolve leaves `gate_instance` null, which drops every narrowed ceiling for that request too.

Not fixed here. Both candidate designs change the settle contract: hold across every plan candidate and release the unused narrowings at settle, or transfer the narrowed hold at dispatch. Picking one is @njbrake's call rather than a review fix. The limit is now written down in `docs/access-control.md` so nobody narrows a ceiling expecting it to survive a fallover.

_Drafted by Claude Opus 5 via back-and-forth with @njbrake._

src/gateway/api/routes/batches.py
```diff
@@ -306,6 +310,7 @@ async def create_batch(
         model=request.model,
         strategy=config.budget_strategy,
         counts_toward_budget=not budget_exempt,
+        scope=BudgetScopeRequest(api_key=api_key, provider_instance=resolved.instance),
```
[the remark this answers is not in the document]
njbrake: Agreed. `create_batch` reserves and then reconciles at $0, and the real cost arrives later through `record_external_spend`, which skips scoped ceilings, so batch spend never reaches a workspace cap. Persisting the scope on the batch record is the right fix and it is a schema change, so it is not in this PR. It is now named as a known limit in `docs/access-control.md` beside external and imported usage.

_Drafted by Claude Opus 5 via back-and-forth with @njbrake._

src/gateway/services/budget_service.py
```diff
@@ -365,6 +442,15 @@ async def reconcile_reservation(db: AsyncSession, handle: ReservationHandle, act
         values["spend"] = User.spend + actual_cost
     if handle.reserved:
         values["reserved"] = _release_reserved(handle.estimate)
+    # Every scoped ceiling the reservation held against has to be unwound too, or
+    # the hold outlives the request and permanently shrinks that ceiling.
+    await settle_scoped(
+        db,
+        handle.scoped_budget_ids,
+        actual_cost=actual_cost,
+        held=handle.scoped_estimate,
```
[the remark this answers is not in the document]
njbrake: Not changing this. The split is the existing design rather than something the scoped path introduced: `reserve_budget` commits the per-user reservation on its own so that no row lock spans the provider call, and `reconcile_reservation` writes `users.spend` in a later transaction for the same reason. Staging both mechanisms into one transaction means holding that lock across the work the current design exists to avoid.

A crash between the two calls leaves scoped spend recorded and the per-user counter short, which is the same exposure the per-user path already carries against `usage_logs`. Making both atomic is a change to how the gateway reserves in general, not to scoped budgets.

_Drafted by Claude Opus 5 via back-and-forth with @njbrake._

alembic/versions/d5e7f1a2b3c4_workspace_scope_request_plane.py
```diff
@@ -0,0 +1,147 @@
...
+    else:
+        organization_id = uuid.UUID(str(organization_id))
+
+    workspace_id = bind.execute(
+        sa.text("SELECT id FROM workspace WHERE organization_id = :org AND name = :name"),
+        {"org": _uuid_literal(bind, organization_id), "name": DEFAULT_WORKSPACE_NAME},
+    ).scalar()
+    if workspace_id is not None:
+        return uuid.UUID(str(workspace_id))
+
```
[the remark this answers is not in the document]
njbrake: Good catch, fixed. `_ensure_default_workspace` now adopts the organization's oldest workspace before minting one, ordered on `created_at, id` so the migration and `workspace_scope.default_workspace_id` resolve the same row instead of disagreeing after a rename.

_Drafted by Claude Opus 5 via back-and-forth with @njbrake._

src/gateway/models/entities.py
```diff
@@ -43,6 +43,17 @@ class APIKey(Base):
...
+    # tables below that carry the same column. NOT NULL: a workspace is the unit
+    # the dashboard scopes by, so "no workspace" is never a real state, only an
+    # unmigrated one. Existing rows were backfilled onto the deployment's default
+    # workspace, which the same migration seeds when tenancy was never touched.
+    # RESTRICT rather than cascade: deleting a workspace must not silently take
+    # its keys, usage, aliases and policies with it. Which workspace a write
+    # lands in is resolved in `services/workspace_scope.py`.
+    workspace_id: Mapped[uuid.UUID] = mapped_column(
+        Uuid, ForeignKey("workspace.id", ondelete="RESTRICT"), nullable=False, index=True
+    )
```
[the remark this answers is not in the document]
njbrake: Confirmed by running them. Both fail on a database migrated to head, which is what their own usage lines document:

```
sqlalchemy.exc.IntegrityError: NOT NULL constraint failed: api_keys.workspace_id
[SQL: INSERT INTO api_keys (id, key_hash, workspace_id, ...) VALUES (?, ?, ?, ...)]
[parameters: ('key-prod', 'hash-key-prod', None, ...)]
```

The retained `server_default` does not save them: SQLAlchemy sends an explicit NULL for a mapped column it was given no value for rather than omitting it from the INSERT, so the default never applies. Both scripts now resolve the default workspace once and stamp it on every row they build, and the migration docstring that claimed otherwise is corrected. Verified by seeding both databases clean.

_Drafted by Claude Opus 5 via back-and-forth with @njbrake._

alembic/versions/a3c7e1b9d5f2_add_scoped_budgets.py
```diff
@@ -0,0 +1,81 @@
+"""Add the tenancy-scoped budget table.
+
+A second, fully enforced budget mechanism beside ``budgets``, which is left
```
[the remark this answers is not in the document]
njbrake: Already acknowledged at the bottom of the PR description, and still owed. Worth noting it is not just a shape change to covered rows: `scoped_budgets` is a new table with no ledger row at all, which is the "surface has no row" trigger in `.github/instructions/reconciliation-ledger.instructions.md`, so it needs a proposed row with its own Current scope, Target and Tracking cells rather than folding into the #625 and #626 entry.

Leaving this open for @njbrake to append before merge.

_Drafted by Claude Opus 5 via back-and-forth with @njbrake._

src/gateway/api/routes/scoped_budgets.py
```diff
@@ -0,0 +1,292 @@
+"""Manage the tenancy-scoped USD ceilings in ``scoped_budgets``.
+
+Standalone-mode only and master-key authed, following ``routes/budgets.py``.
+Deliberately minimal: a scope's ceiling is created, listed, retimed and removed
+here, and everything about how one is enforced lives in
+``services/scoped_budget_service.py``.
```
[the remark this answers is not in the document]
njbrake: Fair, and the description now names the third piece: the `scoped_budgets` table, the five endpoints, why it sits beside `budgets` instead of replacing it, and the two known limits.

The docs were already covering it (`docs/api-reference.md` has the endpoint table, `docs/access-control.md` the semantics), so the gap was the description alone. That still matters here, because merges are squashed and git-cliff parses the PR title, so the release notes would have carried "give the tenancy model a request plane" for a change that adds a budget mechanism.

_Drafted by Claude Opus 5 via back-and-forth with @njbrake._

src/gateway/services/workspace_scope.py
njbrake: Confirmed, including the path you traced. `get_db` (`core/database.py:140`) never commits, so the flushed organization and workspace roll back when the request session closes while the id lives on for the process, and `SingleLogWriter` swallows the resulting foreign key failures.

Fixed, and then made moot: the default is not memoized at all any more. See the reply on your comment about the reset.

_Drafted by Claude Opus 5 via back-and-forth with @njbrake._

src/gateway/services/workspace_scope.py
njbrake: Agreed, and the memoization is gone rather than patched. `RESTRICT` never covered a default workspace holding nothing, and no per-process reset can fix what the other workers and replicas hold, so the cache could not be made correct by clearing it in more places.

`default_workspace_id` now resolves per call: two indexed reads, and only on the master-key path, which is operator traffic. Keyed requests still read the workspace straight off the key, and `_key_workspace` stays because a key never moves between workspaces. This also removed the local import `workspace_service` needed to break the cycle, so that whole comment is gone.

_Drafted by Claude Opus 5 via back-and-forth with @njbrake._

src/gateway/services/workspace_scope.py
njbrake: Fixed: `.order_by(col(Workspace.created_at), col(Workspace.id)).limit(1)` with `scalar_one_or_none`. The migration's own fallback, added for a related finding, orders the same way, so the two resolve the same row rather than disagreeing about where deployment-wide writes land.

_Drafted by Claude Opus 5 via back-and-forth with @njbrake._

src/gateway/services/scoped_budget_service.py
njbrake: The measurement is right and I think the change is worth making. One transaction with a single commit after the loop would also delete the compensation path, since a refusal could roll back instead of releasing, and the lock would span a few UPDATEs rather than a network call either way.

Not doing it in this pass: the lock-free property is argued at length in the module docstring, and reversing a documented design decision is @njbrake's call rather than a review fix. Leaving this open.

_Drafted by Claude Opus 5 via back-and-forth with @njbrake._

src/gateway/services/tenancy/errors.py
njbrake: Right, and fixed. `delete_workspace` now removes the ceilings naming the workspace and its memberships in the same transaction, reading the membership ids before the cascade takes those rows away. A refused delete rolls the ceiling deletes back with it, so a workspace that still holds keys is left exactly as it was. Two integration tests cover both scopes.

Ceilings on an `api_token` or an `org_member` whose subject later goes away are the same shape and are not covered here.

_Drafted by Claude Opus 5 via back-and-forth with @njbrake._

alembic/versions/d5e7f1a2b3c4_workspace_scope_request_plane.py
njbrake: Addressed a third way, because neither option works as stated. `server_default` cannot be declared on the model columns: the value is a per-deployment UUID that is not known at model definition time. Dropping it from the migration breaks the backfill it exists for.

What is wrong is the claim, not the default. It never protected a write path that omits a workspace, because SQLAlchemy sends an explicit NULL for a mapped column with no value rather than omitting it, so the column default never applies to an ORM writer. Both seed scripts were failing on migrated databases for that reason. They now carry a workspace, and the docstring says what the default actually does.

_Drafted by Claude Opus 5 via back-and-forth with @njbrake._

src/gateway/api/routes/scoped_budgets.py
njbrake: Done, moved onto the `APIRouter(...)` call with the `routes/organizations.py` reasoning carried over. The module docstring said it followed `routes/budgets.py`, which is the per-handler form; the organizations argument is the better one and the docstring now says so.

_Drafted by Claude Opus 5 via back-and-forth with @njbrake._

PR #638, opened 2026-08-19
feat(dashboard): bring the navigation shell onto the design

jigjigjig:
## Description

Brings the dashboard's navigation shell onto the navigation design: the rail's
geometry and interaction states, the scope switcher and its menu, both rails'
grouping, and the top bar and account menu around them.

`feat/tenancy-dashboard-pages` (#626) has since merged, so this now targets
`main` and the diff below is only this work.

### The rail
- The design's geometry: 44px rows, 16px glyphs, uppercase group headings, and
  264px expanded / 72px collapsed.
- The glyphs are `react-icons` Feather components now, replacing 359 lines of
  hand-rolled inline SVG. `otari-ai/frontend` names its marks as `IconType`,
  and matching that is the point: a destination on both rails should wear the
  same mark at M5, which it cannot while one rail draws its own paths. This is
  the one field where the two registries differed, and it stops differing.
- The selected row lifts off the rail instead of wearing a teal tint, and hover
  moves the opposite way, so the two are never confusable. Pressed and
  selected-while-hovered complete the two directions, and a keyboard ring sits
  on every row whatever else it wears. All of it is one module,
  `nav/rowStyles.ts`, mirroring the reference's own.
- A collapsed group opens a flyout of its children rather than linking past
  them, which is what makes Guardrails, Web search, and Code execution
  reachable from a collapsed rail. A group gated down to one child becomes it.
- Each rail resumes its last destination instead of resetting, which is the
  design's `Interaction · A ⇄ B` note.

### Scope and both rails
- The switcher is sized to the design and sits on the rail's ground, not on a
  white card that read as the first item in the list. Its menu gained labeled
  groups, the current scope checked at both levels, and create actions.
- The organization rail is the design's four groups, and the row that crosses
  into it is an ordinary nav row rather than a bordered box under the rail.
- **Model pricing** is one of those groups' rows, under Cost & billing, with
  `features/pricing/` behind it. A rate applies to every workspace and every key
  in the deployment, so it is tenant-scoped in fact as well as in the design;
  before this it had no page at all, and the default-catalog refresh flow sat in
  the gateway's runtime Settings next to the master key. One model's rate still
  lives on Models, beside the model it prices.

### Chrome
- The top bar gained the design's right-hand cluster, the account menu was
  rebuilt to its order with appearance folded into one row, and the mobile
  drawer is a full panel below the bar, so its backdrop, focus trap, and
  `role="dialog"` are gone along with the modal it no longer is.

### Deliberately absent
Four organization rows the design draws (Billing, Gateways, the organization's
own provider credentials, the guardrail ceiling) are declared and gated on
surfaces the standalone bootstrap does not report. They are absent here and
present on a deployment that serves them. Create organization is gated the same
way, on an entitlement this build does not grant; the row's handler belongs to
the build that grants it, which `WorkspaceSwitcher` says at the site.

### Two known limitations, both deliberate
Neither is a gap to patch without a design answer, so each is recorded where
someone would look for it rather than left to be discovered:

- **The bundled guide has no entry point below `md`.** It is **Documentation**
  in the top bar, and that cluster is `hidden md:flex`; the design's account
  menu has no row for the guide, so `/#/docs` is the way in on a phone. Noted in
  `web/AGENTS.md`.
- **`/tools`, the three tool services on one page, is reachable by URL only.**
  A group with more than one visible child is a disclosure when expanded and a
  flyout when collapsed, and neither navigates to the parent. Each service has
  its own destination, and Guardrails belongs to Routing, so a row for the
  combined page would duplicate all three. Noted in `nav/registry.ts`.

## Review round

Every thread on this PR is answered and resolved. Six changed code, three were
answered with a decision or a pointer at the existing behavior.

- The mobile drawer was offset from the viewport while the header it sits below
  is in flow under the banners, so with the pricing alarm up it covered the
  header, and with it the only control that closes a drawer that no longer has a
  backdrop. It is absolute within the header's own row now.
- The rail memory re-validated registration but not visibility, so a stored
  destination whose surface a restart stopped reporting resumed onto "not
  available here", and that visit was recorded as somewhere to return to. Both
  halves now read the shell's own predicate, and after a follow-up thread there
  is exactly one implementation of it (`isPathVisible`) rather than two rules
  that disagreed about nesting.
- **Create workspace** was offered to everyone in the scope menu while
  `POST /v1/workspaces` is owners and admins only, so a member met the refusal
  as a 403 after typing a name. Gated on `canManage`, the predicate the
  Workspaces page uses.
- `UserComboBox` resolved a typed owner by id *or* label in one scan. Now that a
  member is labeled by the roster, a member whose name equals another user's
  `user_id` took that id and the key was billed to the wrong owner. An id match
  wins outright.
- The account and scope dialogs named their trigger but not themselves, so a
  screen reader announced an unnamed dialog after activating a named control.
- Model pricing rendered a per-million rate at two fraction digits, so a
  cache-read price of `0.0025` read as `$0.00`, which is what that column's em
  dash means. It uses `formatCost` now, as Models does for the same values.
- `UsageFilters` stays hand-written (it is a filter set assembled across five
  usage endpoints, several values repeatable), but its keys are pinned to
  `GET /v1/usage`, so a key the spec drops is a build error rather than a param
  the server ignores in silence.
- The e2e suite went red on this branch and this PR is why: a group is a
  disclosure now, and HeroUI puts its trigger inside an `h3`, so
  `getByRole("heading", { name: "Routing" })` matched the rail row and the page
  title both and failed strict mode. One failure took 35 tests with it; a
  `pageHeading()` helper scopes past the rail.
- Docs followed the controls that moved: `docs/dashboard.md` described an
  account control that still held the guide, a light/dark radio group, and
  "sign-out", and listed Providers under Gateway where the rail files it under
  Access.

## One commit that is not this PR's work

`fix(migrations): serialize the two revisions that claimed one parent`.

`main` has **two alembic heads**: #667 (`f2a4c6d8b0e3`) and #640
(`b6e8c2a4d7f1`) were both authored against `a3c7e1b9d5f2` and both merged, so
`upgrade head` has nothing to choose between them. Two revisions naming one
parent is not a merge conflict, which is why nothing stopped it.

The gateway migrates to `head` on boot, so it does not start: every branch off
current `main` fails the same three ways this one did (`e2e` could not boot the
webServer, `serving` failed seven tests, and the Docker liveness check failed on
application startup), and `test_the_revision_chain_has_one_head` is red on `main`
itself.

The fix serializes the two, since they are independent (a `scoped_budgets`
rebuild against three `user` columns), and keeps the chain linear as all 41
revisions before it are. The later-merged revision is the one that moves, because
nothing can be sitting on it: reaching `b6e8c2a4d7f1` required resolving the very
ambiguity this fixes.

It is carried here only because this PR cannot go green without it, and it is one
commit, so it lifts out into its own change cleanly if you would rather unblock
`main` ahead of this.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Was stacked on #626, which has merged; this targets `main` directly.

## Checklist

- [x] I ran the Definition of Done checks locally (`make lint`, `make typecheck`, `make test`). *(Web-only diff, so the dashboard counterparts: `pnpm --dir web run lint`, `typecheck`, the full Vitest suite (1029 tests), and `pnpm --dir web run e2e` (38 tests).)*
- [ ] If the API contract changed, I regenerated the OpenAPI spec (`uv run python scripts/generate_openapi.py`). *(No API contract change: this PR touches `web/` and `docs/` only. The committed client was regenerated after merging `main` and matches the spec.)*
- [ ] **M4 ledger:** this PR either needs no entry (an existing row already covers it and its target is unchanged), or the entry is appended to [otari-ai#1587](https://github.com/mozilla-ai/otari-ai/issues/1587). One entry per *change*, not per PR: a stack delivering one surface change gets one. See [.github/instructions/reconciliation-ledger.instructions.md](.github/instructions/reconciliation-ledger.instructions.md).

## Notes for the reviewer

- **Adds a dependency, `react-icons`** (`^5.7.0`), which is the point of the
  glyph change rather than a side effect of it: it is what makes a mark on this
  rail the same object as a mark on otari.ai's. Worth a deliberate ack.
- **The Vitest suite needs Node 22.** Node 26 ships a built-in `localStorage`
  global that is `undefined` without `--localstorage-file` and shadows jsdom's,
  which fails 133 tests locally for reasons that have nothing to do with this
  branch. CI pins 22.
- **`pull_request` workflows have not been firing for pushes to this branch.**
  Only `Otari PR Title Check` (a `pull_request_target` workflow) has run since
  `653da03`; the path filters match and the PR is not a draft, so this looks like
  a GitHub-side rather than a config issue. The dashboard workflow was dispatched
  manually to get a real signal, and `dashboard`, `e2e` and `screenshots` all
  passed. Worth confirming the checks fire before relying on them to merge.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:** Claude Code (Claude Opus)

The navigation shell was implemented with Claude Code against the navigation
design, reviewed and adjusted by hand.

**NOTE:**

## Summary

- Updated dashboard navigation to match the new design across desktop and mobile.
- Added expanded and collapsed rails, grouped navigation, icons, flyouts, scope switching, and destination restoration.
- Reorganized workspace and organization destinations with entitlement and deployment checks.
- Moved pricing management to Cost & billing and added the model pricing page.
- Improved account controls, top-bar actions, mobile documentation access, and shared navigation styling.
- Added navigation, pricing, tenancy, usage scoping, and user-selection test coverage.
- Added `react-icons` and merged Alembic migration branches into one head.
- Updated dashboard documentation and usage filter typing.

## Technical notes

- Gated destinations are hidden and excluded from navigation restoration when unavailable.
- Standalone deployments now advertise the `pricing` surface.
- Vitest requires Node 22.


web/src/app/AppShell.tsx
```diff
@@ -411,271 +443,184 @@ export function AppShell() {
...
+                  "fixed inset-x-0 top-14 bottom-0 z-40 w-full transition-transform duration-250 ease-out-fluid motion-reduce:transition-none",
                   mobileNavOpen ? "translate-x-0" : "-translate-x-full",
                 )
               : clsx(
-                  "relative shrink-0",
-                  !resizing && "transition-[width] duration-150",
+                  // 150ms, which is what `otari-ai/frontend`'s shell collapses
+                  // its own rail in.
+                  "relative shrink-0 transition-[width] duration-150 motion-reduce:transition-none",
+                  collapsed ? COLLAPSED_SIDEBAR_WIDTH : SIDEBAR_WIDTH,
```
jigjigjig, line 502: **The mobile drawer's `top-14` assumes the header is flush with the viewport top; `PricingWarning` breaks that, and there is no longer any other way to dismiss it.**

`PricingWarning` renders as an in-flow block (`<div className="shrink-0 px-6 pt-3">`, `features/models/PricingWarning.tsx:47`) *above* the `flex min-h-0 flex-1` row that holds the aside and the header. When it is up (`require_pricing` on and `default_pricing` off, the documented onboarding state), the header no longer starts at y=0 — it starts ~60px down — while this drawer stays pinned at a viewport-relative `top-14`. The drawer then covers the header outright, including the "Close navigation" button in it.

That button is now the *only* pointer affordance that closes the drawer: this PR removed the backdrop and the focus trap, and the aside is no longer `role="dialog"`. So on a phone in the pricing-alarm state, opening the nav leaves no way out except navigating to some page or pressing Escape (which a touch user does not have). Before this PR the backdrop covered that case.

Drive the offset from the header instead of a literal `3.5rem` (render the drawer inside the content pane below the header, or measure it), or bring the dismiss affordance back.

Related: the comment at `AppShell.tsx:652` still says the mobile drawer "is dismissed from the control to its left or from the backdrop" — the backdrop is gone.

web/src/app/nav/navigationHistory.ts
```diff
@@ -0,0 +1,101 @@
...
+  try {
+    window.localStorage.setItem(STORAGE_KEYS[navContextForPath(pathname)], to)
+  } catch {
+    // Storage can throw when it is disabled (blocked cookies, private mode).
+    // Losing the memory costs a landing page, so it is not worth an error path.
+  }
+}
+
+/** Where you last were on this rail, if it is still somewhere you can go. */
+export function lastLocation(
```
jigjigjig, line 109: **`lastLocation` re-validates registration but not visibility, so the resume link can dead-end on the "not available here" panel.**

The docstring above this function says the re-validation exists because "a destination that has been removed or moved to the other rail since it was stored would otherwise send someone to the shell's 'not available here' panel". `destinationAt` + `navContextForPath` only answer "is this still a registered destination on this rail" — neither consults `isVisible`/the surface set.

Concrete case: a session stores `otari.dashboard.lastOrganizationLocation = "/settings"`, then the gateway is restarted with a config (or mode) that no longer reports the `settings` surface. `/settings` is still registered and still an organization destination, so `lastLocation("organization")` returns it, and the footer's **Organization** row — the only way onto that rail — now lands straight on `EmptyState "Settings is not available here"`. `rememberLocation` has the mirror-image gap: a gated-off route reached by URL still shows the panel *and* gets recorded as somewhere to return to.

`AppShell` already has `isVisible` from `useNavVisibility()` (line 297); passing it in (or filtering the two `lastLocation` results through `navItemForPath` + `isVisible`) closes both halves.

web/src/app/nav/WorkspaceSwitcher.tsx
```diff
@@ -30,115 +68,177 @@ export function WorkspaceSwitcher({ collapsed }: { collapsed: boolean }) {
...
-          </div>
-          <p className="mt-2 border-t border-border px-2 pt-2 text-xs text-muted">
-            Scopes members, API keys, usage, and the request log. Routing and
-            provider credentials stay deployment-wide.
-          </p>
-        </Popover.Dialog>
-      </Popover.Content>
-    </Popover>
+            <div className={`${MENU_DIVIDER} my-1`} />
+            <button
```
jigjigjig: **"Create workspace" is offered to everyone here, but the endpoint is owners/admins only and the Workspaces page gates the same action.**

`WorkspacesPage` renders its own create control only behind `manages = canManage(context.data)` (`WorkspacesPage.tsx:161`, used at :253), and the server documents `POST /v1/workspaces` as "Create a workspace in the caller's organization. Owners and admins only." (`src/gateway/api/routes/workspaces.py:53`).

This row has no such check, so an organization `member` or `viewer` gets the full create form from the scope menu and only discovers the refusal as a 403 in `ErrorBanner` after typing a name. `context` (`useOrganizationContext()`) is already in scope at line 57, so this is `canManage(context.data) ? <button…> : null` — the same predicate the page uses, and the same shape as the `EntitlementGate` on the "Create organization" row right below it.

web/src/app/nav/TopBarActions.tsx
```diff
@@ -0,0 +1,72 @@
...
+ * in this build supplies one; it is the seam an overlay build fills.
+ */
+export function TopBarActions({ balance }: { balance?: number }) {
+  const { management_url } = useDeployment()
+  const platform = management_url?.replace(/\/$/, "")
+
+  return (
+    // Hidden below the md breakpoint, where the mobile header has room for the
+    // dismiss control and the trail and nothing else.
+    <div className="hidden shrink-0 items-center gap-5 md:flex">
```
jigjigjig, line 36: **Moving the guide here makes it unreachable on mobile.** 

This container is `hidden … md:flex`, and the PR also removed the `/docs` row from `AccountMenu` (which lived inside the drawer and was therefore reachable at every width). `grep -rn 'to="/docs"' web/src` now returns exactly one hit: this link. So below the `md` breakpoint the bundled user guide has no entry point at all.

That is the width the guide is most likely to be read at: `docs/dashboard.md` has an "Install it on your phone" section and the repo ships PWA icons for it, so the installed app is expected on a phone — where the only way to the guide it bundles is now typing `/#/docs` by hand.

Either keep a `/docs` row in the account menu (it is in the drawer, which is full-width on mobile) or let this one link through below `md`.

web/src/features/pricing/ModelPricingPage.tsx
```diff
@@ -0,0 +1,325 @@
...
+interface PriceRow {
+  modelKey: string
+  input: number
+  output: number
+  cacheRead: number | null
+  tiers: number
+  updatedAt: string
+}
+
+const rate = (value: number) =>
```
jigjigjig: **`rate()` rounds sub-cent per-million rates to `$0.00`, and disagrees with the Models page about the same numbers.**

`maximumFractionDigits: 2` on a per-million rate is not enough precision: a cache-read price of `0.0025` renders as `$0.00` — indistinguishable from the `—` this very table uses to mean *no cache-read rate at all* (the comment on that column makes exactly that distinction) — and `0.075` renders as `$0.08`.

`shared/helpers/format.ts#formatCost` already solves this ("Show more precision for tiny per-request costs so they don't read as $0.00": 4 digits below 0.01) and is what `ModelsPage` uses for these same `*_price_per_million` values (`ModelsPage.tsx:733-755`). As written, the two pages print different numbers for one stored price.

Secondary: `toLocaleString(undefined, …)` takes the runtime locale while the `$` is hard-coded, so a comma-decimal locale gives `$2,50`; `formatCost` pins `en-US` for that reason.

```suggestion
const rate = (value: number) => formatCost(value)
```
(with `import { formatCost } from "@/shared/helpers/format"`)

web/src/app/nav/registry.ts
```diff
@@ -102,16 +111,20 @@ const BASE_NAV_SECTIONS = [
         to: "/tools",
```
jigjigjig, line 112: **`/tools` is no longer reachable from the sidebar in either rail state.**

With both children visible (the normal case — only `Guardrails` carries its own `surface`), this entry renders as a `Disclosure.Trigger` when expanded, which does not navigate, and as a flyout of just `Web search` + `Code execution` when collapsed. Neither offers `/tools`.

`routes/tools.index.tsx` serves the combined `ToolsGuardrailsPage()` (no `only` prop, i.e. all three services on one page), so that page is now URL-only. Before this PR the collapsed rail rendered `<Link to={item.to}>` for a group, so `/tools` was reachable there — the new flyout closes the Web-search/Code-execution gap the docstring describes and opens this one.

If the combined view is meant to stay reachable, the flyout and the disclosure need a row for the parent (otari-ai's rail uses an "Overview"-style first child for this); if it is not, `tools.index.tsx` is better off redirecting to `/tools/web-search` than being a page nothing links to.

web/src/shared/components/ui.tsx
```diff
@@ -451,6 +451,24 @@ export function EmptyState({
...
+/**
+ * A destination this build declares but does not serve.
+ *
+ * The shell already answers a *gated-off* registered path with its own panel, so
+ * in a standalone gateway this never renders: the surface is absent, the sidebar
+ * drops the link, and the shell intercepts the route. It is what a deployment
+ * that reports the surface but has not composed the overlay serving it would
+ * see, which is the one case that would otherwise paint a blank page.
+ */
+export function UnavailableHere({ title }: { title: string }) {
```
jigjigjig, line 585: **This function was inserted between `PageLoading`'s comment and `PageLoading`.**

The block comment directly above (`// A full-width loading placeholder for a page (or section) whose content is gated on a first fetch … `role="status"` announces the wait`) documents `PageLoading`, which now sits below `UnavailableHere`. As it stands the file reads as if that paragraph describes this empty-state component.

Move `UnavailableHere` above that comment, or below `PageLoading`.

web/AGENTS.md
```diff
@@ -33,7 +33,7 @@ tokens, TanStack Query, Vitest), read
...
 - Build: `make dashboard` from the repo root (`npm --prefix web ci && npm --prefix web run build`). Output goes to `src/gateway/static/dashboard/` (set in `web/vite.config.ts`), which is gitignored, not committed: Vite content-hashes every asset filename, so a committed bundle turned any two branches touching `web/src` into a merge conflict. Consequences: the Docker image builds the bundle in a `node:22-slim` stage (`Dockerfile`), a source checkout without a build degrades to the tutorial page (`gateway/dashboard.py` returns `None`), and a wheel ships the dashboard only if `make dashboard` ran first. There is nothing to commit after a `web/src` change.
 - Bundled user guide: the dashboard renders `docs/dashboard.md` at `/#/docs` (linked from the sidebar footer) by importing it with Vite's `?raw` (`web/src/features/docs/DocsPage.tsx`), so the running dashboard ships the guide it matches. The guide lives outside `web/`, so `web/vite.config.ts` grants `server.fs` access to it. Because it is a build input from outside `web/`, it is also a Docker image input: `.dockerignore` excludes `docs/` but re-includes this one file, and `otari-docker.yml` triggers on it.
```
jigjigjig: **Stale after this PR:** the bundled guide is no longer "linked from the sidebar footer" — this PR moved it to the top bar (`src/app/nav/TopBarActions.tsx`, labeled "Documentation") and removed the row from `AccountMenu`. Line 13 of this same file was updated for the `icon` change in this diff, so this line is stale within one edit.

docs/dashboard.md
```diff
@@ -500,9 +509,9 @@ put it behind HTTPS, as the security notes below describe.
   server, expires the cookie, and clears any cached admin data. Rotating the
   master key revokes every session and re-mints the one you are using, so other
   signed-in browsers are logged out.
-- **Sign out on a machine you share.** A session runs for its full
+- **Log out on a machine you share.** A session runs for its full
```
jigjigjig, line 514: **One place this rename did not reach: line 165.**

> At the very bottom sits the account control: the bundled user guide, the light/dark preference, and sign-out.

All three drifted in this PR:

- the bundled user guide is now **Documentation** in the top bar, not in the account control (`src/app/nav/TopBarActions.tsx`) — and `AccountMenu` no longer has a row for it;
- "sign-out" is now **Log out**, the rename applied on this very line;
- the light/dark preference is now a cycling **Appearance** row rather than a radio group.

`features/docs/DocsPage.tsx` renders this file as the in-app guide, so an operator reading it is pointed at a control that is no longer there.

web/src/app/AppShell.tsx
jigjigjig: Fixed in a437e90, by driving the offset from the header rather than restoring a dismiss affordance.

The row that holds the aside and the content pane now carries `relative`, and the drawer is `absolute inset-x-0 top-14 bottom-0` inside it instead of `fixed`. That row's top edge *is* the header's top edge, so anything in flow above it (the update prompt, the connection status, `PricingWarning`) moves the drawer down together with the header. `top-14` still pairs with the header's `min-h-14`, which is what it was always meant to measure.

That keeps the header, and the "Close navigation" button in it, visible and hittable under an open drawer in every banner state, which is what makes the missing backdrop survivable.

The stale comment at the collapse control is corrected in the same commit: it now says the drawer is dismissed from the control to its left, and why that control stays visible underneath it.

`AppShell.test.tsx` pins both halves (the aside is `absolute`, not `fixed`; its parent is `relative`), following how that suite already pins `fixed` and `w-[16.5rem]`.

web/src/app/nav/navigationHistory.ts
jigjigjig: Fixed in a437e90. Both halves take the predicate.

`rememberLocation(pathname, isVisible)` and `lastLocation(context, isVisible)` now receive the same `useNavVisibility()` result the shell already holds, so a gated-off destination is neither written nor returned. `destinationAt` gives back the entry (and, for a nested destination, the child) alongside the path, and `isReachable` composes them the way `NavGroup` does: the entry's own three gates, plus the parent's gates with the child's `surface` substituted. That last part matters for exactly the case the docstring names, Guardrails being grouped under Routing but served by `tools`.

The docstring now lists three things that keep a stored value honest rather than two, and says why the read-side gate is not redundant with the write-side one: a gateway restarted against a config reporting fewer surfaces changes the answer under an entry that was legitimately written.

Covered by four new tests: `navigationHistory.test.ts` for both directions plus the nested-surface case, and `AppShell.test.tsx` for the end-to-end shape (a stored `/settings` with the surface withdrawn falls back to `/organization/members`, and a URL visit that lands on the panel records nothing).

web/src/app/nav/WorkspaceSwitcher.tsx
jigjigjig: Fixed in a437e90, with the predicate you named: `canManage(context.data)`, failing closed, the same call and the same outcome as `WorkspacesPage`.

One thing came with it. Hiding the row would have left the divider above it drawing a line across the foot of the menu, since nothing in this build grants `organizations.create`. So both create rows resolve as predicates now (`useEntitlement("organizations.create")` in place of `EntitlementGate`) and the divider renders only when at least one of them does. That is the same reason the sidebar takes the gates as a predicate rather than a wrapper: a rule has to disappear along with the last row under it.

Two tests in `AppShell.test.tsx` cover the owner and the member, scoped to the popover so they do not match the `Modal.Trigger` of the same name.

web/src/app/nav/TopBarActions.tsx
jigjigjig: Correct as described, and deliberately left as it is: the guide stays reachable from `md` up only, and below that the URL is the way in.

The two alternatives both cost more than the gap does. A `/docs` row in the account menu puts back the row the design's menu does not have, and this PR's whole claim is that the menu now matches that artboard. Letting this one link through below `md` means an icon-only affordance in a bar the design gives three worded links, invented here rather than drawn there.

So this is a known limitation rather than an oversight, and `web/AGENTS.md` now records it where someone would look for it (0bc05b6): the cluster is `hidden md:flex`, the account menu carries no row for the guide, so below `md` `/#/docs` is the only entry point. If the phone case turns out to matter in practice, it wants a design answer for the mobile bar rather than a row added back here.

web/src/features/pricing/ModelPricingPage.tsx
jigjigjig: Fixed in 331a215, taking the suggestion: the local `rate` helper is gone and the three cells call `formatCost` directly, which is what `ModelsPage` formats the same `*_price_per_million` values with, so the two pages cannot print different numbers for one stored price.

That settles both halves. Sub-cent rates keep four digits, so `0.0025` reads as `$0.0025` and stays distinguishable from the em dash this table uses for "no cache-read rate at all"; and `formatCost` pins `en-US`, where the old helper took the runtime locale while hard-coding the `$`.

A test in `ModelPricingPage.test.tsx` pins `$0.0025` and `$0.08` and asserts no `$0.00` appears.

The em dash in this file's header comment went with it, since prose in the repo does not use them.

web/src/app/nav/registry.ts
jigjigjig: Correct, and the answer is the second half of your comment: the combined view is not meant to stay reachable from the rail. `/tools` is a URL-only page, and a437e90 records that in the registry rather than changing behavior.

An "Overview"-style first child was the obvious fix and is the wrong one here. Each of the three services already has its own destination, and Guardrails is one of them but belongs to Routing, so a row for the combined page would duplicate all three and reach across the grouping the design draws. (Routing leading with Policies is not the same pattern: `/routing` *is* the policies page, so that child is a destination in its own right rather than a handle for its parent.)

The redirect you offer as the alternative is also out, and for a rule this repo writes down: `web/AGENTS.md` says a nav entry points at a page and never at a redirect, and `/tools` is an entry's `to`. It is what gates and names the group.

So the comment in `registry.ts` now says `to` here is a gate and a name rather than somewhere the rail navigates, and why the combined page has no row. The page keeps its route and its tests; if it ever earns a row, that is a design decision about the rail rather than a gap to patch.

web/src/shared/components/ui.tsx
jigjigjig: Fixed in a437e90: `UnavailableHere` moved below `PageLoading`, so the block comment sits against the function it describes again.

web/AGENTS.md
jigjigjig: Fixed in 0bc05b6. The line now names **Documentation** in the top bar (`src/app/nav/TopBarActions.tsx`) and, while it is being rewritten anyway, records what that costs: the cluster is `hidden md:flex` and the account menu carries no row for the guide, so below `md` the URL is the only way in. See the reply on the `TopBarActions.tsx` thread for why that is where it stays.

docs/dashboard.md
jigjigjig: Fixed in 0bc05b6. All three drifts on that line are corrected: the account control is now described as account settings, a cycling **Appearance** row, the hosted legal pages where a deployment has them, and **Log out**, and the sentence says outright that the bundled guide is not in it, it is **Documentation** in the top bar.

Grepped the rest of the file for the same drift ("sign out", "user guide", "light/dark", "account control") and line 165 was the only place left; the `Log out` mentions further down were already right.

web/src/app/nav/types.ts
```diff
@@ -55,7 +56,39 @@ type NavItemGating =
...
+  icon: IconType
+  /**
+   * The surface this destination needs, when it is not the parent's.
+   *
+   * Grouping is an editorial choice and gating is a fact about the deployment,
+   * so the two can disagree: Guardrails is grouped under Routing, where the
+   * navigation prototype puts it, but the page is served by the tools surface.
+   * Omitted, the child inherits the parent's, which is the ordinary case.
+   */
+  surface?: string
```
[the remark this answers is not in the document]
jigjigjig: Already the case, so no change here.

`NAV_CHILD_PARENTS` (`registry.ts`) does exactly the substitution this asks for. It maps each child path to *the parent carrying the child's surface when the child declares one*:

```ts
child.surface ? { ...item, surface: child.surface } : item
```

So `navItemForPath("/tools/guardrails")` answers with `{ ...Routing, surface: "tools" }`, and the `isVisible(currentItem)` call in `AppShell` gates on `tools` while keeping Routing's label (for the panel title and the sidebar highlight) and Routing's capability and flag axes. Three gates, still three.

What was missing was a test: the sidebar half was covered ("drops a nested destination whose own surface the deployment lacks") but not the route half, which is what a bookmark hits. 132724a adds one that visits `/tools/guardrails` with only the `routing` surface and asserts the panel rather than the page.

The same substitution is now also what `navigationHistory` uses to decide whether a remembered destination is still reachable (a437e90), so the rule has one shape in both places.

web/src/app/nav/WorkspaceSwitcher.tsx
```diff
@@ -0,0 +1,244 @@
...
+              <button
+                type="button"
+                className={`${MENU_ROW} text-muted hover:bg-surface-alt hover:text-foreground`}
+                onClick={() => setOpen(false)}
+              >
+                <PlusMark />
+                <span className="min-w-0 flex-1 truncate">
+                  Create organization
+                </span>
+              </button>
```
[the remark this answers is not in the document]
jigjigjig: Right that the row does nothing, and it stays, with the reason written at the site (6318171).

There is nothing here to wire it to. A self-hosted gateway is one tenant: it mounts no create-organization endpoint and serves no page for one, so `setOpen(false)` is the whole of what this build can honestly do. The row is gated on `organizations.create`, which `BASE_CAPABILITIES` never grants, so it does not render in this build at all.

Removing it would lose the one thing it does earn: the row's place in the menu, in the order the design draws, proven against real markup rather than described in a comment. The comment now says the handler belongs to the build that grants the capability, that the grant and the handler belong in the same change, and that a deployment granting one without the other ships a row that does nothing which nothing in this repository can fail on. That last part is the honest cost, and naming it is better than leaving it to be discovered.

web/src/client/local.ts
```diff
@@ -102,6 +102,11 @@ export interface DashboardBuild {
  * spec for this to be generated from.
  */
 export interface UsageFilters {
+  // The workspace a row was recorded in. Set from the sidebar's switcher rather
+  // than by the operator, so the request log, its counts, and the summary and
+  // series the charts are drawn from all show the workspace the shell is looking
+  // at. Every usage endpoint this object is spread onto takes it.
+  workspace_id?: string
```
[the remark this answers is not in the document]
jigjigjig: Half taken, in b7e4412.

The drift risk is real and is now a build error. Every `UsageFilters` key is pinned to the operation that carries all of them, `GET /v1/usage`:

```ts
export type UsageFilterKeysArePinned = Assert<
  keyof UsageFilters extends keyof RequestsQuery ? true : false
>
```

A key the spec drops, or a typo, fails `tsc` instead of being serialized and ignored by the server. Verified by planting a bogus key and watching it fail; without it, that key compiled fine.

The type itself stays hand-written, which is the documented call rather than an oversight (`web/AGENTS.md` names `UsageFilters` as one of three sanctioned exceptions in `local.ts`). Deriving it is not available: it is one filter set the hooks spread onto five endpoints whose query sets differ (`series` adds `group_by`, `summary` adds `bucket` and `dimensions`), and three of its fields are `string | string[]` because they go on the wire as repeated params, which the generated type expresses as a single value. A generated alias would either be one endpoint's shape (wrong for the other four) or an intersection that rejects the array form the hooks assemble.

Checked all fourteen keys against the spec while doing this, incidentally: all are live query parameters on `/v1/usage`, `workspace_id` included, so nothing was drifting today.

web/src/features/keys/KeysPage.tsx
```diff
@@ -974,6 +1017,28 @@ export function KeysPage() {
...
+        </Link>
+        , and their limits under{" "}
+        <Link
+          to="/budgets"
+          className="font-medium text-link hover:text-link-hover"
+        >
+          Spend &amp; budgets
+        </Link>
+        .
+      </p>
```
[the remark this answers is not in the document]
jigjigjig: Not taking this one, because the guideline is narrower than the reading here and the fix it asks for is not available.

The rule is about *declaring a destination*: "Declare a new destination in the nav registry, never as a hand-written link, with whichever of the three gates it needs." `/users` and `/budgets` are both declared there, on the organization rail. These two are in-page cross-links to destinations that already exist, which is what the guide sends someone to when a key needs an owner or a spending limit, and the codebase does that throughout (`RoutingPage` links to Tools, `ModelPricingPage` to Models, the Workspaces and Members pages to each other).

The suggested indirection cannot be built as described either. `src/app` is exactly what a feature may not import, Biome-enforced and proven by `src/architecture.test.ts`, so a "registry-backed helper" reachable from `features/keys` would mean either moving the registry into `src/shared` (it depends on `navContextForPath` and the generated route tree, and its whole point is to be the app shell's data) or adding a second table of paths in `shared`, which is the duplication this is meant to prevent, one indirection further from the routes.

The safety net the rule really wants is already there and stronger than a helper: `to` on a TanStack `Link` is checked against the generated route tree, so a path that is not a real route fails `tsc`. A renamed route breaks the build here rather than 404ing at runtime.

web/src/features/users/UserComboBox.tsx
```diff
@@ -29,25 +30,43 @@ export function UserComboBox({
...
+      return {
+        id: u.user_id,
+        name: u.alias ? `${u.user_id} (${u.alias})` : u.user_id,
+        isMember: false,
+      }
+    })
+    .sort((a, b) => {
+      if (a.isMember !== b.isMember) return a.isMember ? -1 : 1
+      return a.name.localeCompare(b.name)
+    })
```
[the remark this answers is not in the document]
jigjigjig: Good catch, and fixed as suggested in bbccbdb.

The collision is real: a member's label is now a free-form roster name, members sort to the front, and one scan matching `o.id === trimmed || o.name === trimmed` therefore let a roster name outrank another user's actual id. `resolveId` now looks for an id match first and falls back to a name match, so typing an id always submits that id.

Took the diff as written, with a comment saying why the order matters, and added a test that plants exactly the collision described (a member whose roster name is `ci-bot` alongside a real user with `user_id: "ci-bot"`, the member sorted first) and asserts the typed `ci-bot` submits `ci-bot`.

web/src/features/workspaces/WorkspaceMembersPage.test.tsx
```diff
@@ -0,0 +1,74 @@
...
+  const client = new QueryClient({
+    defaultOptions: { queries: { retry: false } },
+  })
+  return render(
+    <QueryClientProvider client={client}>
+      <SelectedWorkspaceProvider>
+        <WorkspaceMembersPage />
+      </SelectedWorkspaceProvider>
+    </QueryClientProvider>,
+  )
```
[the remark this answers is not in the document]
jigjigjig: Leaving this one as it is.

The guideline is conditional on the page keeping state in the URL: "A component test needs a real router, since the pages keep filters and pagination in the URL." `WorkspaceMembersPage` keeps none. It reads the selected workspace from `SelectedWorkspace` (a context over localStorage, which the test does provide), the organization context and the member list from TanStack Query, and it renders no `Link`: not in the page, and not in `WorkspaceMembersPanel` under it. There are no search params to validate and no navigation to await, so the harness would add a router the component never consults, plus the `await` that comes with it.

The harness is worth reaching for the moment that changes, and it will announce itself: rendering a `Link` or calling a router hook without a router throws, so the test fails rather than quietly passing on a wrong assumption.

docs/dashboard.md
```diff
@@ -301,7 +301,7 @@ hand.
 
 ### Gateway
 
-- **Provider credentials**: add, edit, test, and delete provider credentials at runtime
+- **Providers**: add, edit, test, and delete provider credentials at runtime
```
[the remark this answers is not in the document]
jigjigjig: Fixed in cd9a210, in the direction suggested.

The full description of Providers moves to **Access**, where the registry puts it, and the sentence explaining the filing ("a provider credential is what a key spends upstream") goes with it as the entry's opening clause. Gateway is now Models, Routing and Tools, which is the group as `registry.ts` declares it, so the line promising the groups match the dashboard rail by rail is true again.

web/src/app/nav/AccountMenu.tsx
```diff
@@ -120,149 +201,74 @@ export function AccountMenu({ collapsed }: { collapsed: boolean }) {
...
+              aria-hidden="true"
+              className={`text-muted ${navIndicatorClass({ open })}`}
+            />
+          </>
         )}
       </Button>
       {/* Opens upward: the control is pinned to the bottom of the rail. */}
       <Popover.Content placement="top start">
-        <Popover.Dialog className="flex w-56 flex-col gap-0.5">
+        <Popover.Dialog className="flex w-[17rem] flex-col gap-1.5">
```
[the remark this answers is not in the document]
jigjigjig: Fixed in 6318171: `aria-label="Account"` on the account menu's `Popover.Dialog` and `aria-label="Switch workspace"` on the switcher's, so all four surfaces this PR opens name themselves.

Worth stating why the trigger's own name does not cover it, since both triggers are carefully named: the name belongs to the control, not to the dialog it opens, so a screen reader read the label on activation and then announced an unnamed dialog. Both sites now carry a comment pointing at the flyout and the modal that already did this, so the next popover added to the rail has the pattern next to it.

web/src/app/nav/registry.ts
```diff
@@ -151,45 +167,116 @@ const BASE_NAV_SECTIONS = [
...
+        to: "/organization/billing",
+        label: "Billing",
+        surface: "billing",
+        icon: FiCreditCard,
+      },
+      // Tenant-scoped in fact as well as in the design: a rate applies to every
+      // workspace and every key in the deployment. The catalog had no home
+      // before — its refresh flow sat in the gateway's runtime Settings next to
+      // the master key — so this is where it lives, while one model's rate stays
+      // on Models, beside the model it prices.
```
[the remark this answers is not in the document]
jigjigjig: Fixed in cd9a210: the em dashes become parentheses, which keeps the aside without the separator.

Swept the rest of this PR's prose for the same thing while in there; `ModelPricingPage.tsx` had one in its header comment (331a215) and there were no others.

web/src/app/nav/navigationHistory.ts
```diff
@@ -0,0 +1,149 @@
...
+ * same as visible, and the two resume links are where that difference would land
+ * on a panel instead of a page.
+ */
+function isReachable(
+  found: Destination,
+  isVisible: (item: NavItem) => boolean,
+): boolean {
+  if (!isVisible(found.item)) return false
+  const surface = found.child?.surface
+  return surface === undefined || isVisible({ ...found.item, surface })
```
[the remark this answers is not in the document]
jigjigjig: Right, and fixed in cabd441 by deleting the second rule rather than adjusting it.

The disagreement was real: `isReachable` required the parent's surface *and* the child's, while `navItemForPath` substitutes the child's for the parent's, so a deployment serving `tools` without `routing` would have refused to resume onto `/tools/guardrails` even though the shell renders that page.

Substitution is the rule to keep, for two reasons beyond consistency. It is what `NavChild.surface` documents ("the surface this destination needs, when it is not the parent's"), and it is the one that follows the panel's purpose: the page is served, so refusing it would hide something that works, which is the opposite of the failure the panel exists to prevent.

Rather than copy that rule into a second place, the shell's question is now a function, `isPathVisible(pathname, isVisible)` in `registry.ts`, and both callers ask it: `AppShell` for `routeIsGatedOff`, `navigationHistory` for both halves of the memory. The local `isReachable` and the `Destination` shape it needed are gone, so the answer here is by construction the answer the shell will give when the link is followed.

Coverage, since the ask was specifically for it: `isPathVisible` now has its own block in `registry.test.ts`, including both directions of the nested case (`/tools/guardrails` refused when `tools` is withheld, served when `routing` is), the deeper-path inheritance, and an unregistered path staying ungated. The two `navigationHistory` tests for the nested child and the AppShell test that visits the route both still hold.

One thing the fix deliberately does not touch, because it is a design question rather than a gate: on a deployment serving `tools` without `routing`, the *rail* still has no Guardrails row, since a group whose own surface is absent takes its children's rows with it. Where an orphaned child should sit is not something to settle in a predicate, so the route serves the page and the rail stays as the grouping describes it. The commit message says so too, for whoever meets that combination first.

alembic/versions/b6e8c2a4d7f1_add_scoped_budget_reset_alignment.py
```diff
@@ -29,7 +37,7 @@
 from alembic import op
 
 revision: str = "b6e8c2a4d7f1"
-down_revision: str | Sequence[str] | None = "a3c7e1b9d5f2"
+down_revision: str | Sequence[str] | None = "f2a4c6d8b0e3"
```
khaledosman: Re-parenting rather than adding a merge revision strands any DB already stamped `b6e8c2a4d7f1`. The docstring's argument ("nothing can be sitting on it: reaching it required resolving the very ambiguity this fixes") holds only for someone who reached it through main's two-head state. Anyone who ran the app on #640's branch **before** it merged got there from the single head `a3c7e1b9d5f2` with no ambiguity to resolve, as did anyone who ran `alembic upgrade heads`. After this change `b6e8c2a4d7f1` is the head, so `upgrade head` is a no-op for them and #667's three nullable `user` credential columns are never created, silently, until tenancy code touches them.

`alembic merge -m "serialize the two revisions that claimed one parent" f2a4c6d8b0e3 b6e8c2a4d7f1` converges from either side and still leaves exactly one head, so `test_the_revision_chain_has_one_head` stays green. If you keep the re-parent instead, the docstring should say what such a DB has to do to recover, because `alembic stamp` then re-runs a revision whose column already exists.

Blast radius is local dev DBs only: `git tag --contains` shows neither revision in a release.

web/src/app/nav/AccountMenu.tsx
```diff
@@ -1,58 +1,100 @@
...
 const THEME_LABELS: Record<ThemePreference, string> = {
   system: "System",
   light: "Light",
   dark: "Dark",
 }
 
+// 36px rows at 13.5px, which is the menu's own scale: a step down from the
+// rail's 44px/14px, because a menu row is read once on the way to a decision
+// rather than scanned as a standing list.
+const MENU_ROW = `flex min-h-9 w-full items-center gap-2.5 rounded-md px-2.5 text-left text-[0.84375rem] leading-[1.125rem] font-medium ${NAV_TRANSITION}`
```
khaledosman: `min-h-9` is a 36px row, and this menu renders inside the mobile drawer, where Log out is a control people tap. The phone floor is 44px (`responsiveness.md`: "Touch targets are at least 44px", sized up below `md`). `min-h-11 md:min-h-9` keeps the 36px menu scale on the desktop rail and meets the floor on a phone.

web/src/app/nav/TopBarActions.tsx
khaledosman, line 36: This cluster is the guide's only entry point now, and it is `hidden md:flex`, so on a phone no control reaches `/#/docs` at all, on an app that installs to the home screen. The account menu does render inside the drawer, so a Documentation row there closes the gap for the price of one row the design does not draw; that reads better than recording the limitation in `web/AGENTS.md` and leaving the phone with a URL as its only way in.

web/src/routes/organization.pricing.tsx
```diff
@@ -0,0 +1,7 @@
+import { createFileRoute } from "@tanstack/react-router"
+
+import { ModelPricingPage } from "@/features/pricing/ModelPricingPage"
+
+export const Route = createFileRoute("/organization/pricing")({
```
khaledosman, line 5: New page, no screenshot entry. `web/e2e/screenshots/authenticated.spec.ts`'s `organization rail` block needs a case for `/organization/pricing` (heading `/model pricing/i`), which is what will cover it at three viewports in both themes once that suite becomes a gate. Owed even though nothing fails without it today.

web/src/app/nav/registry.ts
```diff
@@ -152,45 +178,116 @@ const BASE_NAV_SECTIONS = [
...
+      },
+      // Tenant-scoped in fact as well as in the design: a rate applies to every
+      // workspace and every key in the deployment. The catalog had no home
+      // before (its refresh flow sat in the gateway's runtime Settings next to
+      // the master key), so this is where it lives, while one model's rate stays
+      // on Models, beside the model it prices.
+      {
+        to: "/organization/pricing",
+        label: "Model pricing",
+        surface: "settings",
```
khaledosman: The gate names the wrong surface for this page's own data: the table and the refresh flow are `/v1/pricing`, which is its own router (`app.include_router(pricing.router)`), while `settings` covers `/v1/settings`, which this page reads only for the policy banner. In this gateway the set is all-or-nothing so it makes no difference, but the surface axis exists for the deployment where it comes apart, and there `settings` shows a page whose data is not served.

`pricing` is a mountable route name, so `STANDALONE_SURFACES` takes it without failing `test_deployment_bootstrap`, and the bootstrap response is `list[str]` so no schema regeneration. One line each side.

web/src/app/nav/AccountMenu.tsx
```diff
@@ -120,149 +201,80 @@ export function AccountMenu({ collapsed }: { collapsed: boolean }) {
       <Button
         variant="ghost"
         aria-label="Account"
-        className={`h-auto w-auto! justify-start gap-2.5 rounded-lg px-2 py-2 text-left ${
-          collapsed ? "mx-2 justify-center" : "mx-3"
-        }`}
+        className={`${navRowClass({ collapsed })} w-auto! justify-start`}
       >
-        <span className="flex h-6 w-6 shrink-0 items-center justify-center rounded-full bg-surface-alt text-[10px] font-semibold text-foreground">
+        <span className="flex h-[1.625rem] w-[1.625rem] shrink-0 items-center justify-center rounded-full border border-border bg-surface-alt text-[0.5625rem] font-semibold text-muted">
```
khaledosman: Off-scale type, and it is a small family now: `text-[0.5625rem]` here, `text-[0.84375rem]` at 51 and 266, `text-[0.71875rem]` at 269, plus `WorkspaceSwitcher.tsx:127` (`text-[0.6875rem]`) and `TopBarActions.tsx:23` (`text-[0.8125rem]`). `design-tokens.md`: the scale is the whole set of sizes, and a one-off combination that repeats is a role that should be added to `globals.css` instead. Worth resolving here rather than later, since this PR replaced a `text-[11px]` with `text-overline` in `AppShell` in the same breath.

web/src/app/nav/TopBarActions.tsx
```diff
@@ -0,0 +1,72 @@
...
+      </span>
+    </span>
+  )
+}
+
+/**
+ * @param balance Spendable credit, when the deployment holds a wallet. Nothing
+ * in this build supplies one; it is the seam an overlay build fills.
+ */
+export function TopBarActions({ balance }: { balance?: number }) {
```
khaledosman: `balance` is unreachable in any build this repository can produce: `AppShell` renders `<TopBarActions />`, and an overlay replaces `overlaySections.ts` / `overlayLabelOverrides.ts` only, so nothing can pass it without editing a base source file. The `Balance` component and the `EntitlementGate capability="billing"` behind it are dead with it. Dropping both until there is a seam that feeds them costs nothing to add back, and leaves one less prop that reads as wired.

web/src/app/nav/navigationHistory.ts
```diff
@@ -0,0 +1,130 @@
...
+ * gated-off route is still registered and still reachable by URL, so without
+ * that check the resume link lands on the shell's "not available here" panel,
+ * and the visit that saw the panel is recorded as somewhere worth returning to.
+ * That second check is `isPathVisible` rather than a rule of its own, so the
+ * answer here is by construction the one the shell will give when the link is
+ * followed. And because the first check resolves against the registry, what
+ * comes back is a `NavPath`: the caller gets something a `Link` will accept, not
+ * a string it has to cast.
+ *
+ * It lives here rather than in `shared/helpers` — where `otari-ai` keeps its
```
khaledosman: Em dashes in a doc comment, which the repo's writing style rules out (`AGENTS.md`: docs, doc comments, commit messages). A comma pair or parentheses reads the same. Same in `AppShell.test.tsx:586`. The `"—"` table cells in `ModelPricingPage` are UI glyphs and fine.

khaledosman (approved):
Approving; nothing here blocks the dashboard work. Verified at `0b3622b8`: `pnpm --dir web run lint`, `typecheck` and the full 1029-test Vitest suite pass, and a test run leaves no generated-file drift. `main` is confirmed red (`get_heads()` returns `['b6e8c2a4d7f1', 'f2a4c6d8b0e3']`, `test_the_revision_chain_has_one_head` fails there), and the head of this branch passes `tests/unit/test_tenancy_schema_chain.py` in full, so the carried migration commit is doing real work.

The one finding to settle before merge is the shape of that migration fix (inline on `down_revision`): serializing rather than merging strands a local DB that already ran #640's revision on its branch. It is the only finding that can damage something outside this PR.

One thing checked and cleared rather than flagged: `lastLocation()` reads `localStorage` during render, which the React Compiler is free to cache. A live navigation probe under the real vite/compiler config (workspace → `/usage` → Organization → back link) tracked the current session correctly, so the resume link is not stale today.

🤖 Review generated with [Claude Code](https://claude.com/claude-code) (Claude Opus)

alembic/versions/b6e8c2a4d7f1_add_scoped_budget_reset_alignment.py
jigjigjig: Agreed, and taken as written: \`d0f2b4a6c8e1\` now merges the two strands and \`b6e8c2a4d7f1\` keeps \`a3c7e1b9d5f2\` as its parent (626561c, first commit).

You were right about who the re-parent stranded. I checked the case rather than reasoning about it: upgrading a database stopped on each strand to the merged head, and asserting on the resulting schema. Both arrive with #667's credential columns *and* \`reset_alignment\` present; before the fix, the #640 side reached \`upgrade head\` as a no-op and silently lacked the three \`user\` columns, exactly as you described.

\`test_the_revision_chain_has_one_head\` stays green (\`get_heads()\` is \`['d0f2b4a6c8e1']\`), and the revision is empty by design, since the two strands are independent. The docstring on \`b6e8c2a4d7f1\` now records why it keeps its parent instead of arguing that nothing could be sitting on it.

web/src/app/nav/registry.ts
jigjigjig: Fixed in 327b9bf: the row is \`surface: "pricing"\`, and \`STANDALONE_SURFACES\` grew the name.

Your reasoning about the axis is the part I kept in the comment: this gateway mounts its management API all or nothing, so the two are the same answer here, and the surface axis exists for the deployment where they come apart. There the old gate offered a page whose data is not served.

Two things your note did not have to mention but that the change needed: \`test_every_surface_names_a_route_the_gateway_mounts\` passes because \`/v1/pricing\` is mounted, and \`surfaces\` is a \`list[str]\` on the wire so nothing regenerates. The frontend mirrors that tuple in two places (\`web/src/tests/fixtures.ts\` and \`registry.test.ts\`), so both took \`pricing\` too, or the row would have been gated off in every test instead.

web/src/app/nav/TopBarActions.tsx
jigjigjig: Done in 0a70fbe, the way you suggested: the account menu carries a Documentation row, under \`md:hidden\`.

The reason I went with the row rather than recording the limitation is the one you gave, that this is the width the guide is most likely to be read at. \`md:hidden\` rather than showing it everywhere keeps the design's menu unchanged at the desk: the row is present exactly where the top-bar cluster is not.

It is a router \`Link\`, not the \`MenuExternalLink\` next to it, since \`/docs\` is a route in this app and an \`<a href>\` would reload the shell. Covered by a test that opens the menu and asserts the href and the breakpoint class.

\`docs/dashboard.md\` and \`web/AGENTS.md\` both stated the gap as a limitation, so they describe the row now instead.

web/src/app/nav/AccountMenu.tsx
jigjigjig: Fixed in 0a70fbe: \`min-h-11 md:min-h-9\`, exactly as you wrote it.

Your framing is what the comment now says: 36px is the menu's own scale, but it is a desk figure, and this menu also renders inside the mobile drawer where the rows are tapped. The 44px floor applies below \`md\` and the menu scale resumes at the breakpoint.

web/src/app/nav/TopBarActions.tsx
jigjigjig: Removed in 0a70fbe: the \`balance\` prop, the \`Balance\` component, and the \`EntitlementGate capability="billing"\` behind it are gone.

Your unreachability argument checks out end to end. \`AppShell\` renders \`<TopBarActions />\` with no prop, and an overlay replaces \`overlaySections.ts\` / \`overlayLabelOverrides.ts\` and not this file, so nothing in any build this repository produces could supply one.

I kept the fact rather than the code: the header comment now says the design draws a balance and why it is not built (this gateway meters spend but holds no wallet, and there is no seam that could feed one), so the next person reads it as a deliberate absence rather than an omission.

web/src/app/nav/AccountMenu.tsx
jigjigjig: Taken, and taken as a family rather than one site, since that was your point: \`globals.css\` now carries three chrome roles and the six inline sizes are gone (0a70fbe).

\`text-chrome-row\` (13.5px), \`text-chrome-meta\` (11.5px), \`text-chrome-initials\` (9px), grouped in the scale's comment block with a note that the chrome reads a step under the content scale because it is furniture around a page rather than part of one.

Two decisions worth flagging, since neither is forced by your comment:

- **The roles carry metrics only, no color and no weight.** One menu row is \`text-foreground\` resting and \`text-muted\` disabled, and the identity block is medium over semibold, so the call site keeps saying which. It also avoids a specificity fight with \`text-muted\`.
- **Two sizes moved half a pixel to land on a role**: the top bar's links 13px to 13.5px, and the switcher's organization line 11px to 11.5px. Consolidating is the whole point, so I would rather have three roles than five, but say the word if the artboard means 13px exactly and I will give it its own role.

\`foundation.test.ts\` asserts both halves: the roles are declared, and no file in \`app/nav\` carries an arbitrary size again. That second one is what keeps a seventh from appearing.

I left the pre-existing \`text-[11px]\` uses in \`features/activity\` and friends alone as out of scope for this PR.

web/src/routes/organization.pricing.tsx
jigjigjig: Added in 03667fe: an \`organization model pricing\` case in the \`organization rail\` block, reaching \`/organization/pricing\` and asserting the \`/model pricing/i\` heading, captured as \`organization-model-pricing\`.

Reached by URL like the rest of that file, for the reason its comment already gives: on the mobile viewport the footer link lives inside the closed drawer, so clicking it times out and the page is never captured at the size most worth looking at.

Agreed it was owed even though nothing fails without it today.

web/src/app/nav/navigationHistory.ts
jigjigjig: Fixed in 626561c. \`navigationHistory.ts:25\` takes parentheses, and \`AppShell.test.tsx\` the comma pair (that one rode along in 0a70fbe, which rewrote the same block).

I also caught one you did not flag, \`ui.tsx:137\`, which is in this diff and had the same problem; it takes a semicolon, since the two halves are independent clauses.

And confirmed your carve-out rather than assuming it: the \`"—"\` cells in \`ModelPricingPage\` are UI glyphs and stay, as does the \`Reset history — \` assertion in \`parity.management.spec.ts\`, which matches copy in \`BudgetsPage.tsx\` that this PR does not touch.

PR #640, opened 2026-08-19
feat(budgets): let a scoped budget reset on a UTC calendar boundary

### LINKED ISSUE #639: [FEATURE] Calendar-aligned resets for scoped_budgets
### Problem or use case

`scoped_budgets` expresses a period only as `budget_duration_sec`, an integer added to the last reset. Two things follow.

A calendar month is not a fixed number of seconds, so a monthly budget cannot be expressed at all. 2592000 is a different product: a year holds 12.17 thirty-day periods against 12 months, so every monthly cap rises about 1.5 percent.

Separately, and affecting every period: `_roll_expired_periods` sets `period_start = now` and `period_end = now + budget_duration_sec`, where `now` is the first request *after* expiry rather than the moment of expiry. A window is therefore "at least N seconds, restarted on next use" and not "every N seconds". On a quiet deployment a daily budget rolls at whenever traffic resumed and stays there, and each delay is permanent.

The second is the more interesting one: when the reset happened leaks into the data and never washes out. A calendar-aligned window is derived from the boundary instead, so a budget rolled late still lands on the correct window and the reset becomes independent of when anything looked at it.

### Proposed solution

A nullable `reset_alignment` column holding `calendar_day` / `calendar_week` / `calendar_month`. A plain string rather than a database enum, for the same reason `scope_type` is one.

Three states, with a CHECK so a fourth cannot be stored:

| `budget_duration_sec` | `reset_alignment` | meaning |
| --- | --- | --- |
| NULL | NULL | never resets (unchanged) |
| set | NULL | rolling (unchanged) |
| NULL | set | calendar aligned (new) |

`CHECK (NOT (budget_duration_sec IS NOT NULL AND reset_alignment IS NOT NULL))`. Without it the two columns encode one concept with an implicit "ignored when" rule, and `(86400, calendar_month)` becomes storable and meaningless.

When alignment is set, `_roll_expired_periods` computes the window containing `now` (midnight UTC, Monday 00:00, or the 1st) instead of `now + duration`. Same lock-free compare-and-swap on `period_end`, and the same no-backfill semantics: a budget untouched for two months lands in the current month with fresh counters.

This is not a period enum. Periods stay in seconds; the column only says whether a reset snaps to a boundary.

**UTC only, deliberately.** Boundaries are UTC, so a deployment in Tokyo sees a daily budget reset at 09:00 local. That is weighed and left out of scope rather than missed: a local calendar day is 23 or 25 hours across a DST transition, which is real complexity for a benefit nobody has asked for yet. A `reset_timezone` column stays additive if that changes.

### Alternatives considered

**Infer alignment from the duration** (86400 means daily, so snap it). No schema change at all, but it silently changes every existing daily budget on upgrade, which is the one thing this must not do.

**Rolling 30 days for monthly.** Decided against in otari-ai#1725 for the 1.5 percent above.

**Refusing to migrate monthly budgets.** After cutover the gateway is what enforces, so those caps stop existing rather than being enforced somewhere else.

### Definition of Done

- Existing rows behave identically with `reset_alignment` NULL.
- An aligned budget resets on the boundary and not on next use, tested by rolling one across a boundary late.
- The CHECK rejects a row with both columns set.
- One Alembic revision with a real downgrade, roundtrip tested.
- `docs/access-control.md` updated alongside the existing scoped-budget section.

Consumer: otari-ai#1725 blocks on this. Its migration script already has the mode wired and refusing until the column exists.

njbrake:
## Description

`scoped_budgets` expressed a period only as `budget_duration_sec`, seconds added to the last reset, so a calendar month could not be written at all (2592000 makes a year of 12.17 periods against 12 months, about 1.5 percent more generous than a monthly cap reads) and a rolling window restarted from the first request after expiry rather than from the moment of expiry, which lets the reset time walk forward on a quiet deployment.

A nullable `reset_alignment` column (`calendar_day`, `calendar_week`, `calendar_month`) is the other place a period can come from, and a CHECK keeps every row to one of three states: neither field never resets, a duration is rolling, an alignment is calendar aligned. Both set is not storable, so the pair never needs an "ignored when" rule to be read.

When alignment is set, the roll at the gate writes the window containing `now` (midnight, Monday 00:00, or the 1st) instead of `now + duration`. Same lock-free compare-and-swap on `period_end`, same no-backfill: a ceiling nothing touched for two months lands in the current month with fresh counters, and when the reset happened stops leaking into the data. Existing rows keep the column NULL and behave exactly as before.

The window derivation is one function, shared by creation, retiming, and the roll, so the three cannot drift apart. Boundaries are UTC only, deliberately: a local calendar day is 23 or 25 hours across a DST transition, and a reset timezone stays additive if that is ever wanted. An alignment value the API cannot create (only reachable by a write that went around it) logs a warning and leaves the exhausted window in place, which refuses requests rather than guessing a cadence.

The migration adds the column and the CHECK through `batch_alter_table` with `copy_from`, because SQLite has no `ALTER TABLE ... ADD CONSTRAINT` and `copy_from` is what carries the table's two partial unique indexes through SQLite's rebuild.

`otari-ai#1725` blocks on this column.

One unrelated line rides along: a repo-root `.pnpm-store/` is now ignored. pnpm falls back to one when it cannot use a store on the same filesystem as the install target, which is what happens in a container, and the untracked result is large enough to push a diff past the file window GitHub applies path filters within, which silently skips every path-filtered workflow. The gap dates from the pnpm migration in #637.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [x] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes #639

## Checklist

- [x] **M4 ledger:** this PR either needs no entry (an existing row already covers it and its target is unchanged), or the entry is appended to [otari-ai#1587](https://github.com/mozilla-ai/otari-ai/issues/1587). One entry per *change*, not per PR: a stack delivering one surface change gets one. See [.github/instructions/reconciliation-ledger.instructions.md](instructions/reconciliation-ledger.instructions.md).

No ledger entry: the `scoped_budgets` row already covers this table and its target disposition is unchanged.

## Test plan

- Unit: the window arithmetic, including December, both Februaries, the set of month lengths, an aware timestamp in another offset, and an unrecognized alignment.
- Unit: the revision's chain on SQLite, since nothing else migrates it. Column, CHECK, both partial unique indexes surviving the rebuild, the three legal states storable, the fourth refused, downgrade, and upgrade/downgrade/upgrade.
- Integration: a daily ceiling rolled three days late lands on today's midnight; a monthly one asleep for two months lands on the 1st with zeroed counters; an unrecognized alignment stays exhausted and 403s; the database refuses a row carrying both periods; and over the API, an aligned ceiling opens on its boundary, a cadence switch is one request, both fields is a 400, an unknown alignment is a 422.
- `make lint`, `make typecheck`, `make openapi-check`, `make postman-check`, the OSS-edition smoke gate, and the full unit (1830) and integration (1355) suites, against a local PostgreSQL 18.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

Implemented from the issue, which @njbrake wrote up with the design decisions already made (the mutual-exclusion CHECK, UTC only, no inference from the duration). The reasoning and decisions are his; the code and prose are Claude's.

**NOTE:**

## Summary

Adds calendar-aligned reset support for `scoped_budgets`.

- Supports UTC day, week, and month boundaries.
- Preserves rolling duration resets and no-reset behavior.
- Rejects conflicting or invalid reset settings.
- Updates the API, client schema, documentation, and Postman example.
- Adds migration, rollover, validation, API, and period-calculation tests.
- Ignores repository-local `.pnpm-store/` directories.


tbille (approved):

alembic/versions/b6e8c2a4d7f1_add_scoped_budget_reset_alignment.py
```diff
@@ -0,0 +1,101 @@
...
+Create Date: 2026-08-19
+"""
+
+from collections.abc import Sequence
+
+import sqlalchemy as sa
+from alembic import op
+
+revision: str = "b6e8c2a4d7f1"
+down_revision: str | Sequence[str] | None = "a3c7e1b9d5f2"
```
khaledosman, line 32: Chain collision, not a defect in this revision: five open PRs all set `down_revision = "a3c7e1b9d5f2"`, which is main's current head (#668 `7ff4e082eb0c`, #671 `f7a2c4e6b8d1`, #670 `e1c3a5b7d9f2`, #667 `f2a4c6d8b0e3`, this one `b6e8c2a4d7f1`). Whichever lands first is fine; the second forks the chain, and `alembic upgrade head` then refuses to choose — startup `auto_migrate` and `tests/integration/conftest.py` both run it. Needs a merge order with the rest rebased onto the new head.

Second-order hazard from this revision specifically: `_scoped_budgets()` hard-codes the column list for `copy_from`. No other open PR rebuilds `scoped_budgets`, so nothing is wrong today, but any future `batch_alter_table` on this table that omits `reset_alignment` will silently drop it on SQLite.

src/gateway/services/scoped_budget_service.py
```diff
@@ -190,9 +203,49 @@ async def _resolve_identities(
...
+        start = day - timedelta(days=day.weekday())
+        return start, start + timedelta(days=7)
+    if alignment == ALIGN_MONTH:
+        start = day.replace(day=1)
+        end = start.replace(year=start.year + 1, month=1) if start.month == 12 else start.replace(month=start.month + 1)
+        return start, end
+    raise ValueError(f"Unknown reset alignment: {alignment!r}")
+
+
+def period_window(
```
khaledosman, line 227: Coordination note with #671: `WorkspaceBudgetDefault` has no `reset_alignment` column, and `WorkspaceBudgetDefaultService._create_member_budget` derives its window with an inline `now + timedelta(...)` rather than through `period_window`. Once both land, a workspace budget default can only ever materialise a rolling ceiling, and there are two derivations of a period again — which is the drift this function exists to prevent. Worth agreeing which PR closes it.

khaledosman (approved):
No findings in the enforcement path.

Checked the two things a reader would most want proven: every aligned window returned by `_aligned_window` has `end > now` strictly (day, ISO week, and month, including the December rollover), so a rolled aligned ceiling can never re-enter `expired` and spin; and the CAS guard `period_end <= now` is still sound for an aligned roll for the same reason, so a loser at the boundary sees zero rows exactly as it did for a rolling window. The `window is None` arm in `_roll_expired_periods` reproduces the old `period_start=now, period_end=None` behavior, so a cadence-less row that somehow carries a `period_end` is unchanged.

`ScopedBudgetResponse.reset_alignment: str | None` rather than the `Literal` reads as a mismatch against the request until you notice `scope_type` is `str` there too, and that typing it as the union would turn a row written around the API into a 500 on `GET`. Consistent, and right.

One process note: the `.gitignore` entry is unrelated to the feature and the title is `feat(budgets)`. The reasoning for riding along is good and I would not split it, but it is the kind of thing `CONTRIBUTING.md`'s "keep diffs focused" is aimed at.

---

PR #667, opened 2026-08-19
feat(tenancy): add the nullable credential columns to `user`

### LINKED ISSUE #645: Add the nullable credential columns to `user`
- Part of mozilla-ai/otari-ai#1452
- **Blocks the entire auth track.**

## Why this exists

mozilla-ai/otari-ai#1716 settled that the master key bootstraps first boot and is retained as the API credential, while sessions become the dashboard login. Every ported tenancy page reads off a session that resolves a user, an organization and a role, and `user` (`src/gateway/models/tenancy.py`) carries no credential of any kind today: `email`, `is_active`, `is_superuser`, `full_name`, `active_organization_id`, and nothing else.

That decision put these columns ahead of the rehome deliberately. They exist in production on `otari-ai` and in no table here, the overlay contributes no tables, and landing them now gives mozilla-ai/otari-ai#1644 one target schema instead of two.

## What to produce

One additive migration adding `hashed_password`, `oauth_provider`, `email_verification_token`, `email_verified_at` and `terms_accepted_at` to `user`, all nullable. Nothing reads them yet. `email` stays nullable and unique, because a standalone operator needs no address.

## Out of scope

`webauthn_credential` is a table rather than a column, and there is no WebAuthn anywhere in this tree, so it rides the passkey issue instead. No authentication behavior changes here.

## Definition of Done

- The migration appends to the linear Alembic chain and roundtrips up and down on SQLite and PostgreSQL.
- `alembic heads` stays single.
- An existing v0.x database upgrades with every row untouched and every new column null.

## Carried from mozilla-ai/otari-ai#1716

`docs/access-control.md` and `docs/dashboard.md` still document the standalone model as master-key authenticated end to end. That correction has no owner and belongs with whoever lands the session work. Named here so it is not lost.

njbrake:
## Description

`user` carried no credential of any kind, so the session flow settled in mozilla-ai/otari-ai#1716 had nowhere to store one and the re-parenting migration (mozilla-ai/otari-ai#1644) faced one target schema per edition. One additive revision (`f2a4c6d8b0e3`) adds `hashed_password`, `terms_accepted_at`, `oauth_provider`, `email_verification_token` and `email_verified_at`, all nullable, none read anywhere in this tree. `email` is untouched: still nullable, still uniquely indexed, because a standalone operator needs no address. No authentication behavior changes, and `webauthn_credential` stays with the passkey issue.

**No table rebuild.** Five `ADD COLUMN` statements and one `CREATE UNIQUE INDEX`, which both engines take as plain DDL. Rebuilding `user` on SQLite would mean recreating the table four others hold foreign keys into, plus the half of the `organization` cycle that points at it, so `email_verification_token` is unique through an index rather than a constraint (as it is in production), and `downgrade` drops the index before the column because SQLite refuses `DROP COLUMN` while one covers it.

**Two departures from the platform's production shape**, both stated next to the columns and both now settled: they stay, and mozilla-ai/otari-ai#1644 carries the cast rather than this tree carrying production's spelling.

- `oauth_provider` is `VARCHAR(50)`, not production's native `oauthprovider` enum. `op.add_column` does not create a PostgreSQL enum type, the same type renders as VARCHAR plus a CHECK on SQLite, and the enum stores uppercase member *names* rather than the flow's values. The vocabulary belongs to the OAuth flow that has not rehomed; the tenancy tables already keep their own vocabularies as strings (`role`, `status`).
- `terms_accepted_at` and `email_verified_at` are timezone-aware (`UtcDateTime`), unlike production's naive `DateTime`, matching the departure the tenancy revision already applied to every other timestamp in this schema.

Per column, against production's `3c08d06718ce`: `hashed_password` and `email_verification_token` are DDL-identical, unbounded VARCHAR and the same `ix_user_email_verification_token` unique index name. The other three (`oauth_provider`, `terms_accepted_at`, `email_verified_at`) are the two departures above, so three of the five columns diverge from production DDL by design. Both departures are recorded on mozilla-ai/otari-ai#1644 as planned casts, each with the cast it owes: native enum to text needs an explicit `::text`, and naive to aware needs localizing to UTC rather than reinterpreting in the server timezone. `alembic heads` stays single, and a new unit test pins that so a parallel M5 revision cannot quietly fork it.

**Adjacent, not fixed here** (flagged rather than touched, since several migrations are landing in parallel): `docs/access-control.md` and `docs/dashboard.md` still describe standalone as master-key authenticated end to end. Issue #645 carries that correction forward to whoever lands the session work, and it is still unowned.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes #645
Part of mozilla-ai/otari-ai#1452, ahead of mozilla-ai/otari-ai#1644 and mozilla-ai/otari-ai#1716.

## Checklist

- [x] Documentation was updated where necessary (the module docstring that said these columns were deliberately absent; the two docs corrections issue #645 carries are named below and still unowned).
- [ ] **M4 ledger:** this PR either needs no entry (an existing row already covers it and its target is unchanged), or the entry is appended to [otari-ai#1587](https://github.com/mozilla-ai/otari-ai/issues/1587). One entry per *change*, not per PR: a stack delivering one surface change gets one. See [.github/instructions/reconciliation-ledger.instructions.md](https://github.com/mozilla-ai/otari/blob/main/.github/instructions/reconciliation-ledger.instructions.md).

Deliberately unchecked: an entry **is** owed and has not been posted yet. The "Tenancy and access" row's Current scope cell reads "otari's `user` carries no credential columns", which this PR makes false, so the third trigger fires. The target disposition does not move (otari's schema still survives the re-parent). @njbrake to post it.

### Verified locally, and what CI re-verified

- `make lint`, `make typecheck` (mypy over 384 files), `make openapi-check`, `make postman-check`: clean. The spec is unchanged, which is the check that the five columns stayed off the wire (they are on the table class, not on `UserBase`), and `openapi.json`, the Postman collection and `web/src/client/schema.ts` contain none of the five names.
- `uv run pytest tests/unit`: 1816 passed locally, 1817 on CI. The eight new tests are named in the CI log, so they ran there rather than only here.
- Up, down, up on **both** engines, with a pre-existing row inserted at `a3c7e1b9d5f2` first: the row reads back byte-identical, all five columns null, and repeated null tokens coexist under the unique index. `compare_metadata` reports no diff for `user` on either engine, and PostgreSQL `information_schema` confirms the DDL (`character varying` unbounded, `character varying(50)`, `timestamp with time zone`, all nullable).
- The tests were checked for revert-sensitivity rather than assumed to have it. Removing the migration and the model fields fails 7 of the 8; reverting only the model fields fails the parity test and nothing else; making the token index non-unique fails both uniqueness tests; downgrading the two timestamps to naive `DateTime` fails the parity test. The one test that stays green under a full revert is the single-head guard, which is a chain invariant rather than a claim about these columns.
- OSS-edition smoke gate on SQLite and on PostgreSQL: passed.

**CI covers the PostgreSQL upgrade.** `tests/integration/conftest.py` runs `alembic upgrade head` against a Testcontainers `postgres:17`, so the earlier caveat about having used a locally installed PostgreSQL 18 instead of Testcontainers is resolved: CI's `test` job exercised that path on this branch (1347 passed, 9 skipped).

### Not covered by CI

- **The PostgreSQL downgrade.** Nothing in the repo runs `alembic downgrade` against PostgreSQL; the integration fixtures tear down with `DROP TABLE ... CASCADE` instead, and the only downgrade tests are the SQLite ones in `tests/unit/test_tenancy_schema_chain.py`. I ran the PostgreSQL down and up myself and it round-trips, but that is a local result, not a standing guarantee. Adding a permanent PostgreSQL downgrade test would introduce a pattern the repo does not have yet, which felt like the wrong thing to smuggle into a migration PR.
- The dashboard suites were not run: no `web/` file changes, and no generated dashboard artifact moves.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

Claude Opus 5 (1M context) via the Claude Code CLI.

Column shapes were read off the platform's `backend/app/models/user.py` and its `3c08d06718ce` init revision rather than guessed, which is where the two documented departures come from. Every check above was run, not inferred.

**NOTE:**

## Summary

- Added five nullable credential and identity fields to the `User` model.
- Added a migration to create and remove these fields without rebuilding the table.
- Added documentation that explains why the fields are present but unused.
- Expanded schema tests for nullability, uniqueness, data preservation, migration round trips, model alignment, and Alembic head integrity.
- Preserved existing authentication behavior and the nullable unique `email` field.

## Technical notes

- `oauth_provider` uses `VARCHAR(50)`.
- Timestamp fields use timezone-aware types.
- The downgrade removes the token index before dropping the token column for SQLite compatibility.

### DIFF
--- a/alembic/versions/f2a4c6d8b0e3_add_user_credential_columns.py
+++ b/alembic/versions/f2a4c6d8b0e3_add_user_credential_columns.py
@@ -1,0 +1,81 @@
+"""Add the nullable credential columns to ``user``.
+
+``hashed_password``, ``terms_accepted_at``, ``oauth_provider``,
+``email_verification_token`` and ``email_verified_at``, all nullable, all
+unread. They exist in the platform's production ``user`` table and in no table
+here, and the overlay contributes no tables, so landing them now gives the
+re-parenting migration (otari-ai#1644) one target schema instead of one per
+edition. otari-ai#1716 settled the flow they belong to: the master key stays the
+API credential and sessions become the dashboard login, so a standalone row with
+all five null is the normal state rather than an unmigrated one. No
+authentication behavior changes here.
+
+``email`` is deliberately untouched: it stays nullable and uniquely indexed,
+because a standalone operator identity is a label rather than a sign-in address.
+
+Three things this revision settles, all of them stated in ``models/tenancy.py``
+next to the columns:
+
+- **No table rebuild.** Five ``ADD COLUMN`` statements and one
+  ``CREATE UNIQUE INDEX``, which both engines take as plain DDL.
+  ``batch_alter_table`` is what a constraint would have cost, and rebuilding
+  ``user`` on SQLite would mean recreating the table four other tables hold
+  foreign keys into, plus the half of the ``organization`` cycle that points at
+  it. ``email_verification_token`` is therefore unique through an index rather
+  than a constraint, which is what the platform does too, and the index is
+  dropped before the column in ``downgrade`` because SQLite refuses
+  ``DROP COLUMN`` on an indexed column.
+- **``oauth_provider`` is a VARCHAR, not the platform's native ``oauthprovider``
+  enum.** ``op.add_column`` does not create a PostgreSQL enum type for you, and
+  the same type renders as VARCHAR plus a CHECK on SQLite, which the OSS edition
+  ships by default. The vocabulary belongs to the OAuth flow that has not
+  rehomed; the tenancy tables already store their own vocabularies as strings.
+- **The two timestamps are timezone-aware**, unlike the platform's naive
+  ``DateTime``, matching the departure the tenancy revision already applied to
+  every ``created_at`` and ``updated_at`` in this schema.
+
+Revision ID: f2a4c6d8b0e3
+Revises: a3c7e1b9d5f2
+Create Date: 2026-08-19
+"""
+
+from collections.abc import Sequence
+
+import sqlalchemy as sa
+from alembic import op
+
+revision: str = "f2a4c6d8b0e3"
+down_revision: str | Sequence[str] | None = "a3c7e1b9d5f2"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+_VERIFICATION_TOKEN_INDEX = "ix_user_email_verification_token"
+
+
+def upgrade() -> None:
+    # Unbounded, as the platform has it: a hash carries its own algorithm and
+    # cost parameters, so a ceiling here would be a bet on which one the session
+    # flow picks.
+    op.add_column("user", sa.Column("hashed_password", sa.String(), nullable=True))
+    op.add_column("user", sa.Column("terms_accepted_at", sa.DateTime(timezone=True), nullable=True))
+    op.add_column("user", sa.Column("oauth_provider", sa.String(length=50), nullable=True))
+    op.add_column("user", sa.Column("email_verification_token", sa.String(), nullable=True))
+    op.add_column("user", sa.Column("email_verified_at", sa.DateTime(timezone=True), nullable=True))
+    # Unique so two identities cannot hold one token, and repeated NULLs stay
+    # legal on both engines, which is what every existing row will have.
+    op.create_index(
+        op.f(_VERIFICATION_TOKEN_INDEX),
+        "user",
+        ["email_verification_token"],
+        unique=True,
+    )
+
+
+def downgrade() -> None:
+    # Before the column: SQLite refuses ``DROP COLUMN`` while an index covers it.
+    op.drop_index(op.f(_VERIFICATION_TOKEN_INDEX), table_name="user")
+    op.drop_column("user", "email_verified_at")
+    op.drop_column("user", "email_verification_token")
+    op.drop_column("user", "oauth_provider")
+    op.drop_column("user", "terms_accepted_at")
+    op.drop_column("user", "hashed_password")
--- a/alembic/versions/f2a4c6d8b0e3_add_user_credential_columns.py
+++ b/alembic/versions/f2a4c6d8b0e3_add_user_credential_columns.py
@@ -13,26 +13,21 @@ authentication behavior changes here.
 ``email`` is deliberately untouched: it stays nullable and uniquely indexed,
 because a standalone operator identity is a label rather than a sign-in address.
 
-Three things this revision settles, all of them stated in ``models/tenancy.py``
-next to the columns:
+**No table rebuild**, which is the one thing here that belongs to the revision
+rather than to the columns. Five ``ADD COLUMN`` statements and one
+``CREATE UNIQUE INDEX``, which both engines take as plain DDL;
+``batch_alter_table`` is what a constraint would have cost, and rebuilding
+``user`` on SQLite would mean recreating the table four other tables hold
+foreign keys into, plus the half of the ``organization`` cycle that points at it.
+``email_verification_token`` is therefore unique through an index rather than a
+constraint, as it is on the platform, and ``downgrade`` drops that index before
+the column because SQLite refuses ``DROP COLUMN`` while one covers it.
 
-- **No table rebuild.** Five ``ADD COLUMN`` statements and one
-  ``CREATE UNIQUE INDEX``, which both engines take as plain DDL.
-  ``batch_alter_table`` is what a constraint would have cost, and rebuilding
-  ``user`` on SQLite would mean recreating the table four other tables hold
-  foreign keys into, plus the half of the ``organization`` cycle that points at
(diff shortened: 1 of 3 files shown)


njbrake:
Self-review pass on this PR, run against the issue's Definition of Done bullet by bullet. Two test changes pushed, one PR-body correction, nothing changed in the migration or the model.

**Changed.** `test_identities_without_a_token_coexist_under_that_index` counted every null-token row and asserted 2, which quietly also asserted that no revision in the chain ever seeds an identity. It now counts the delta across its own two inserts. The insert helper's docstring said its literals are SQLite-shaped for UUIDs; it also writes booleans as 1/0, so it now says that too.

**Corrected in the description.** The old "Not verified locally" heading said the integration suite had run against a locally installed PostgreSQL rather than Testcontainers. CI's `test` job resolves that: `tests/integration/conftest.py` runs `alembic upgrade head` against a Testcontainers `postgres:17`, and it passed on this branch. What is genuinely not covered by CI is the PostgreSQL **downgrade**, since nothing in the repo runs `alembic downgrade` on PostgreSQL. I ran it locally and it round-trips; the description now says which half is a standing guarantee and which half is my word.

**Tests were checked for revert-sensitivity rather than assumed to have it.** Removing the migration and the model fields fails 7 of the 8 new tests; reverting only the model fields fails the parity test and nothing else; making the token index non-unique fails both uniqueness tests; downgrading the two timestamps to naive `DateTime` fails the parity test. The single-head guard stays green under a full revert, correctly, because it is a chain invariant rather than a claim about these columns.

**Deliberately left alone.**

1. `email_verification_token` is an unbounded VARCHAR under a unique btree index. PostgreSQL errors past roughly 2704 bytes in an index entry. There is no writer today, and the column is DDL-identical to production on purpose, so bounding or hashing it belongs with the flow that starts writing it.
2. Production carries `terms_accepted_at` and `email_verified_at` on `UserBase`, so they reach its wire contract. Here all five sit on the table class only. The DDL is identical either way, so otari-ai#1644 is unaffected, and no route in this tree serializes an identity, so there is no contract to diverge from yet.
3. `oauth_provider` stays a VARCHAR rather than production's native `oauthprovider` enum, for the reasons in the model comment. This is the one call worth a second opinion.
4. `test_the_revision_chain_has_one_head` is a repo-wide invariant living in a tenancy-named file. A general migrations module would be a better home, but creating one now adds a file for the parallel M5 branches to collide on.
5. The M4 ledger entry is still owed and still unposted. The "Tenancy and access" row's Current scope cell reads "otari's `user` carries no credential columns", which this PR makes false, so trigger 3 fires while the target disposition does not move.

_Note: this comment was drafted by Claude Opus 5 (1M context) via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

tbille (approved):

alembic/versions/f2a4c6d8b0e3_add_user_credential_columns.py
```diff
@@ -0,0 +1,76 @@
...
+Create Date: 2026-08-19
+"""
+
+from collections.abc import Sequence
+
+import sqlalchemy as sa
+from alembic import op
+
+revision: str = "f2a4c6d8b0e3"
+down_revision: str | Sequence[str] | None = "a3c7e1b9d5f2"
```
khaledosman, line 43: Chain collision, not a defect in this revision: five open PRs all set `down_revision = "a3c7e1b9d5f2"`, which is main's current head (#668 `7ff4e082eb0c`, #671 `f7a2c4e6b8d1`, #670 `e1c3a5b7d9f2`, this one `f2a4c6d8b0e3`, #640 `b6e8c2a4d7f1`). Whichever lands first is fine; the second forks the chain, and `alembic upgrade head` then refuses to choose — startup `auto_migrate` and `tests/integration/conftest.py` both run it. The single-head test this PR adds is what will catch it, after the fact. Needs a merge order with the rest rebased onto the new head.

khaledosman (approved):
No findings in the revision or the model.

Verified: `UtcDateTime.impl` is `DateTime(timezone=True)`, so the migration's DDL matches the model on both engines and `compare_metadata` has nothing to report; SQLModel's `unique=True, index=True` renders the same `ix_user_email_verification_token` unique index the revision creates by name; and all five columns sit on `User` rather than `UserBase`, so no `*Public` schema and no route can serialize them — no tenancy route returns a bare `User`.

Two non-blocking notes:

- `test_two_identities_cannot_share_a_verification_token` asserts only that *some* `IntegrityError` fires. The `organization.slug` and `user.email` uniques are also in play in that helper, so the test would pass for the wrong reason if the token index were dropped. Matching on the constraint name would pin it to the thing under test.
- The ledger entry you flagged as owed and unposted is the one open item; nothing here is blocked on it.

---

PR #669, opened 2026-08-19
docs: describe the screenshot suite as it now works

khaledosman:
## Description

Follow-up to #637. Its last revision replaced the screenshot harness with otari-ai/frontend's, after the guides describing that harness had already been written, so `testing.md` and `web/AGENTS.md` merged describing the setup it replaced: the old snapshot layout, and masking as the whole answer to nondeterminism.

They now describe what actually runs, and say the settings are otari-ai's so the next person extends one suite instead of inventing a second answer to the same problem:

- real phone emulation for the mobile projects (`isMobile`, `hasTouch`, `deviceScaleFactor`, iPhone UA), and the note that Playwright captures at CSS scale so the PNG stays 390 wide
- the frozen clock, and why it is `setFixedTime` rather than `clock.install`, which would deadlock TanStack Query's refetching and React's scheduler
- the animation, scrollbar, SVG-timeline and font handling that `waitForStable` does
- pinned `locale`/`timezoneId`/`reducedMotion`, on the screenshot projects only
- the two comparison budgets, and why `maxDiffPixelRatio` alone lets a tall full-page capture absorb a whole changed paragraph
- what is still masked and why that goes away once captures run against mocked responses

Plus two things a reader would otherwise rediscover the hard way: the suite depends on `seed` rather than `parity`, because a single flaky behavioral test otherwise takes all 108 captures with it (it did), and `pnpm run <script> -- --flag` silently drops the flag, which is why the update script carries no separator.

One workflow fix comes with it, because the section above cannot be true without it. `otari-dashboard.yml` still uploaded the `screenshot-baselines` artifact from `web/e2e/screenshots/__snapshots__/`, the layout from before the suite adopted otari-ai/frontend's `snapshotPathTemplate`. Captures land beside the spec that took them now, so that directory is never created, and `if-no-files-found: ignore` turned the empty match into a green run whose artifact carried no PNGs. That artifact is the only way to obtain a Linux baseline set, so it also blocked the gate migration this section documents.

No runtime behavior change: two markdown files and one artifact path.

## PR Type

- [ ] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [x] Documentation
- [x] Infrastructure / CI

## Relevant issues

None. Follows #637.

## Checklist

- [x] **M4 ledger:** this PR either needs no entry (an existing row already covers it and its target is unchanged), or the entry is appended to [otari-ai#1587](https://github.com/mozilla-ai/otari-ai/issues/1587). One entry per *change*, not per PR: a stack delivering one surface change gets one. See [.github/instructions/reconciliation-ledger.instructions.md](.github/instructions/reconciliation-ledger.instructions.md).

No tests: two markdown files and a CI artifact path, which only a dispatched run exercises. No ledger entry: no table, no contract, no surface.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

## Summary

- Updated screenshot testing documentation to match the current harness.
- Added guidance for reliable captures, comparisons, masking, snapshot paths, and commands.
- Updated screenshot artifact collection for the current Playwright snapshot path.
- Removed outdated guidance to reduce confusion and flaky results.

## Technical notes

- Documented mobile emulation, CSS scaling, fixed time, locale, stability waits, animation handling, comparison budgets, masking, and `seed` usage.
- Changed the ignored snapshot pattern from `__snapshots__` to `*-snapshots/`.
- Refined frozen-clock and pixel-comparison guidance.


njbrake (approved):
Approving. I reproduced the non-obvious claims rather than reading them: the `snapshotPathTemplate` resolution (a capture lands at `authenticated.spec.ts-snapshots/`), the CSS-scale capture width (390 wide at `deviceScaleFactor: 3`, which relies on the viewport meta in `index.html`), pnpm 11.4.0 forwarding the literal separator into argv, Playwright dropping everything after it (the same 29 tests list with and without, so it is neither a filter nor an error), and `Math.min` on the two budgets in `playwright-core@1.62.1`. The 108 figure checks out too: 18 tests times six projects.

The sweep for stale `__snapshots__` references stopped one short, at `otari-dashboard.yml`'s artifact upload path, now fixed on this branch. It was the last instance in the repo and the one that broke something: the artifact these docs tell you to download carried no PNGs, silently, because `if-no-files-found: ignore` swallowed the empty match. That artifact is also the only route to a Linux baseline set, so it blocked the gate migration the section ends on.

CodeRabbit's two wording findings went in as well. Both were fair, and the frozen-clock bullet contradicted the masks bullet three lines below it.

One thing worth a follow-up rather than this PR: the guide says running the suite locally is worth doing, but not that the run is red by construction, since with no baselines committed Playwright writes each one and fails the test. A first-timer sees 108 failures and assumes they broke something. The CI job comment explains this; the guide does not.

PR #670, opened 2026-08-20
feat(provider-keys): add organization-scoped BYO provider keys

### LINKED ISSUE #643: Organization-scoped provider keys
- Labels: `otari-migration`, `open-core`, `area/backend`
- Part of #589, replaces part of #588

## Why this exists

The two products hold provider credentials in shapes that do not map onto each other, and neither one replaces the other.

`provider_credentials` here is keyed on instance name, is deployment-global, and is merged over `config.yml` into an in-memory overlay, because the streaming dispatch path is synchronous and holds no session. Its own docstring scopes it to standalone mode.

`otari-ai`'s `ProviderKey` is organization-scoped, with archival, organization defaults, per-workspace overrides, workspace access lists, and model restrictions.

So this is not a duplicate to dedupe. `provider_key_service` ports, and `provider_store_service` survives beneath it as the config-merge cache.

## Decision

Resolved by mozilla-ai/otari-ai#1748 (recorded in mozilla-ai/otari-ai#1760): an org-scoped source of truth, kept disjoint from the instance overlay rather than merged into it. `provider_key`'s shape ports into `otari` as new, additive tables that become the source of truth for organization-scoped BYO provider credentials. `provider_credentials`/`provider_store_service` are unchanged, staying the mechanism for config.yml-defined and legacy deployment-global stored credentials, addressed by instance name exactly as before. The two mechanisms never merge: an `instance:model` selector and a bare `provider:model` selector are different requests that never contend for the same overlay entry, so neither can "win over" the other, and there is nothing for the two stores to disagree about (this is not literally #1748's Option A, which would derive the instance overlay from `provider_key`; see the docs for why that mechanism was not carried over).

New otari-side tables: `org_provider_keys`, `workspace_provider_key_overrides`, `workspace_provider_model_restrictions` (the platform's managed/phantom-key bucket has no otari-side equivalent and is dropped — every otari-side key is BYO).

## Definition of Done

- [x] Migration + models (`org_provider_keys`, `workspace_provider_key_overrides`, `workspace_provider_model_restrictions`)
- [x] Service (`services/tenancy/org_provider_key_service.py`): CRUD, `set_org_default`, workspace override tri-state auto-resolution, per-workspace model allow-lists, dispatch-path overlay cache
- [x] Dispatch integration: `get_provider_kwargs`/`resolve_provider_selector` gain an optional `workspace_id`, threaded through chat/messages/responses, batches (including retrieve/cancel/list/results), and the embeddings/images/audio/rerank/moderations pass-through scaffold
- [x] Workspace model restrictions enforced at dispatch (403), not only stored
- [x] SSRF gate (`validate_provider_api_base`) applied to organization provider-key writes, matching `/v1/provider-credentials`
- [x] Routes under `/v1/organizations/me/provider-keys` and `/v1/workspaces/{workspace_id}/provider-keys`, reusing existing tenancy authorization
- [x] Unit + integration test coverage, including regressions for a null-name update and a re-pin-after-clear race
- [x] `provider_credentials`/`provider_store_service` behavior verified unchanged (full regression suite green)
- [x] `search_tool_credentials` explicitly left out of scope (separate future decision)

Implemented in mozilla-ai/otari#670.

## References

`src/gateway/services/provider_store_service.py`, `src/gateway/models/entities.py` (`ProviderCredential`); `otari-ai` `backend/app/services/provider_key_service.py`. Disposition recorded in `otari-ai` `docs/architecture/rehome-inventory.md` and `docs/architecture/m4-reconciliation.md`.

tbille:
## Description

Implements organization-scoped provider keys, unblocked by the decision recorded at [otari-ai#1748](https://github.com/mozilla-ai/otari-ai/issues/1748): an org-scoped source of truth, kept disjoint from the instance overlay rather than merged into it (not literally #1748's Option A; see the linked docs for why that mechanism was not carried over).

`provider_credentials` (the existing `config.yml`-merged, instance-keyed, deployment-global store) is **unchanged**. This PR adds new, additive tables that become the source of truth for **organization-scoped** BYO provider credentials: archival, one default per organization+provider, per-workspace pin/disable overrides, and per-workspace model allow-lists. The two mechanisms are disjoint by construction — an `instance:model` selector always resolves through `config.providers` exactly as before and never touches the new tables; only a bare `provider:model` selector with no matching instance falls back to an organization-scoped key when a `workspace_id` is available.

### What's here

- **Schema**: `org_provider_keys`, `workspace_provider_key_overrides`, `workspace_provider_model_restrictions` (migration `e1c3a5b7d9f2`, one forward migration, no legacy data to backfill). The platform's "managed/phantom key" bucket has no otari-side equivalent and is dropped entirely — every otari-side key is BYO.
- **Service** (`services/tenancy/org_provider_key_service.py`): CRUD, `set_org_default` (partial unique index + `IntegrityError`→domain-error mapping for the race), workspace override tri-state auto-resolution (pinning re-enables a disabled key, disabling un-pins and cascades deleting that workspace's model restrictions), and a dispatch-path overlay cache mirroring `provider_store_service.py`'s shape (30s TTL, refreshed on write) but keyed by `(workspace_id, provider)`.
- **Dispatch integration**: `get_provider_kwargs`/`resolve_provider_selector` gain an optional `workspace_id`, consulted only when no `config.providers` entry answers the selector, threaded through the chat/messages/responses preamble and `batches.py` via the workspace already resolvable off the request's `APIKey` (zero extra I/O for a keyed request).
- **Routes**: `/v1/organizations/me/provider-keys` and `/v1/workspaces/{workspace_id}/provider-keys`, reusing `OrganizationService`/`WorkspaceService`'s existing tenancy authorization (org owner/admin for org-scoped mutations; org owner/admin *or* that workspace's own owner/admin for workspace overrides and restrictions — `WorkspaceService.workspace_in_active_organization`/`require_workspace_management_access` promoted to public for this reuse).

### Addressed in review

CodeRabbit and Copilot flagged several real issues, all fixed on top of the original implementation:

- **SSRF gate missing.** `create_key_for_user`/`update_key_for_user` now call `validate_provider_api_base`, matching `/v1/provider-credentials`.
- **Misleading 409 on an explicit `null` name.** An explicit `null`/blank name on update now raises a 400 before it can reach the database as a NOT NULL violation reported as a duplicate-name conflict.
- **`clear_workspace_pinned_default` correctness bug.** The bulk `UPDATE` (`synchronize_session=False`) didn't exclude the key being (re-)pinned, so re-pinning an already-pinned key silently un-pinned it once the ORM's change tracking saw no net change on the loaded object. Fixed by excluding the target key, plus a `WorkspaceRepository.lock()` row lock serializing the whole read-decide-write sequence (no unique index spans the variable set of override rows a pin can land in). Regression test added.
- **Workspace model restrictions were stored but never enforced.** A workspace restricted to one model could still dispatch any model with the same org key. The dispatch-path cache now carries the active key's model allow-list per `(workspace, provider)`, enforced in `_pipeline.py` alongside the existing per-key allow-list check.
- **`workspace_id` was missing from several dispatch paths.** Threaded through the embeddings/images/audio/rerank/moderations pass-through scaffold and every batch lifecycle endpoint (retrieve/cancel/list/results, not only creation), so organization-scoped credentials are visible on every path that resolves a bare selector, not only chat/messages/responses.
- Two low-risk nitpicks: a bulk `DELETE` instead of a per-row loop, and a tightened cache-reset test.

Test coverage grew accordingly: **12 unit + 26 integration** (was 12 + 22), including regressions for the null-name and re-pin-after-clear bugs and three new tests pinning the model-restriction cache behavior.

### Verification

- `make lint` (architecture check + ruff): clean
- `uv run mypy` (390 files, `src`+`tests`+`scripts`): clean
- Full unit suite: **1819 passed**, 0 regressions
- Full integration suite (real PostgreSQL): **1369 passed**, 0 regressions
- New tests: **12 unit + 22 integration**, all passing
- Migration roundtrip (`upgrade` → `downgrade` → `upgrade`) on SQLite: clean
- `scripts/oss_edition_smoke.py` (standalone boot, BYO credential, fallback completion, usage): passes
- OpenAPI spec + Postman collection regenerated and checked

### Decision record

The upstream keying decision and its three required sub-question answers (instance-name addressing stays disjoint and unchanged; one global cache refresh keyed by `(workspace_id, provider)`, not per-organization; `search_tool_credentials` explicitly deferred to a separate decision) are recorded in `otari-ai`'s `docs/architecture/m4-reconciliation.md` and `docs/architecture/rehome-inventory.md` (companion PR pending in that repo).

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes #643

## Checklist

- [x] Documentation was updated where necessary. (Decision recorded in `otari-ai`'s architecture docs; companion PR pending there.)
- [x] **M4 ledger:** the Target disposition for the "API keys and BYO providers" row changes (it read "Dedupe under org/workspace"; the actual decision keeps `provider_credentials`/`provider_key` disjoint rather than deduped). Entry appended: mozilla-ai/otari-ai#1587 (comment).

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:** Claude Code (Claude Sonnet 5), agentic implementation from a reviewed plan.

Implemented end-to-end by Claude Code from a plan reviewed and approved by the repo owner: schema, migration, repositories, service, dispatch-path integration, routes, and tests. All verification steps above were run and their output checked before opening this PR.

## Summary

- Added organization-scoped provider keys with secure storage, lifecycle management, defaults, and workspace controls.
- Added workspace overrides, inheritance, disablement, and model restrictions.
- Added authorized organization and workspace APIs and dashboard client support.
- Updated routing, pricing, usage tracking, streaming, and batch processing to use workspace context.
- Added caching, startup refresh, validation, tenancy protection, database migrations, and regression tests.

These changes let organizations manage BYO provider keys while preserving existing deployment-level behavior. Workspace controls improve provider selection and cost attribution.

## Technical notes

- Bare `provider:model` selectors can use organization keys when a workspace is available.
- Explicit provider instances continue to use configured credentials.
- Batch lifecycle operations retain the originating workspace for credential resolution.

### DIFF
--- a/alembic/versions/e1c3a5b7d9f2_add_org_provider_keys.py
+++ b/alembic/versions/e1c3a5b7d9f2_add_org_provider_keys.py
@@ -1,0 +1,179 @@
+"""Add organization-scoped provider keys.
+
+Decided at otari-ai#1748 (mozilla-ai/otari#643): the platform's
+``ProviderKey`` shape ports into otari as new, additive tables that become the
+source of truth for organization-scoped BYO provider credentials.
+``provider_credentials`` and its config.yml-merge overlay are untouched by
+this revision; they stay the mechanism for config.yml-defined and legacy
+deployment-global stored credentials.
+
+Three tables, no legacy data to migrate (they are brand new here), so this is
+a single forward migration with no expand/backfill/contract dance:
+
+- ``org_provider_keys``: one BYO credential per organization+provider+name,
+  with a partial unique index enforcing at most one ``is_org_default`` row per
+  ``(organization_id, provider)`` among non-archived rows. The "managed
+  bucket" dimension of the platform's equivalent index does not carry over:
+  every otari-side key is BYO, so there is one default per provider, not one
+  per bucket.
+- ``workspace_provider_key_overrides``: a workspace's departure from its
+  organization's default for one key (pin it as the workspace default, or
+  disable it for that workspace).
+- ``workspace_provider_model_restrictions``: a per-workspace, per-key model
+  allow-list.
+
+Rebased onto `f7a2c4e6b8d1` (workspace per-member budget defaults), the actual
+head `main` had gained by the time this revision landed, rather than the
+original `a3c7e1b9d5f2` this was first written against: `main` grew a new head
+several times over the life of this PR, and each rebase re-parents here rather
+than leaving a redundant merge revision beside `main`'s own.
+
+Revision ID: e1c3a5b7d9f2
+Revises: f7a2c4e6b8d1
+Create Date: 2026-08-19 00:00:00.000000
+"""
+
+from collections.abc import Sequence
+
+import sqlalchemy as sa
+from alembic import op
+
+revision: str = "e1c3a5b7d9f2"
+down_revision: str | Sequence[str] | None = "f7a2c4e6b8d1"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+_ORG_DEFAULT_INDEX = "uq_org_provider_keys_org_default"
+
+
+def upgrade() -> None:
+    op.create_table(
+        "org_provider_keys",
+        sa.Column("id", sa.Uuid(), nullable=False),
+        sa.Column("organization_id", sa.Uuid(), nullable=False),
+        sa.Column("provider", sa.String(length=255), nullable=False),
+        sa.Column("name", sa.String(length=255), nullable=False),
+        sa.Column("api_base", sa.String(length=1024), nullable=True),
+        sa.Column("client_args", sa.JSON(), nullable=True),
+        sa.Column("encrypted_api_key", sa.String(), nullable=True),
+        sa.Column("last4", sa.String(length=8), nullable=True),
+        sa.Column("archived_at", sa.DateTime(timezone=True), nullable=True),
+        sa.Column("is_org_default", sa.Boolean(), nullable=False, server_default=sa.false()),
+        sa.Column("created_at", sa.DateTime(timezone=True), server_default=sa.func.now(), nullable=False),
+        sa.Column("updated_at", sa.DateTime(timezone=True), nullable=True),
+        sa.ForeignKeyConstraint(["organization_id"], ["organization.id"], ondelete="CASCADE"),
+        sa.PrimaryKeyConstraint("id"),
+        sa.UniqueConstraint("organization_id", "provider", "name", name="uq_org_provider_keys_org_provider_name"),
+        # Covers (organization_id, id) so the two link tables below can carry a
+        # composite FK to it: Postgres requires the referenced columns of a
+        # composite foreign key to be backed by a unique constraint, not only
+        # ``id`` being the primary key on its own.
+        sa.UniqueConstraint("organization_id", "id", name="uq_org_provider_keys_org_id"),
+    )
+    op.create_index(op.f("ix_org_provider_keys_organization_id"), "org_provider_keys", ["organization_id"])
+    op.create_index(
+        _ORG_DEFAULT_INDEX,
+        "org_provider_keys",
+        ["organization_id", "provider"],
+        unique=True,
+        postgresql_where=sa.text("is_org_default AND archived_at IS NULL"),
+        sqlite_where=sa.text("is_org_default AND archived_at IS NULL"),
+    )
+
+    op.create_table(
+        "workspace_provider_key_overrides",
+        sa.Column("id", sa.Uuid(), nullable=False),
+        sa.Column("workspace_id", sa.Uuid(), nullable=False),
+        # Denormalized from the workspace's own organization (the service layer
+        # only ever creates an override when the key and the workspace already
+        # agree on organization), so the composite FK below can enforce that
+        # invariant at the database rather than trusting every write path to
+        # keep re-deriving it correctly.
+        sa.Column("organization_id", sa.Uuid(), nullable=False),
+        sa.Column("org_provider_key_id", sa.Uuid(), nullable=False),
+        sa.Column("is_default", sa.Boolean(), nullable=False, server_default=sa.false()),
+        sa.Column("disabled", sa.Boolean(), nullable=False, server_default=sa.false()),
+        sa.Column("created_at", sa.DateTime(timezone=True), server_default=sa.func.now(), nullable=False),
+        sa.Column("updated_at", sa.DateTime(timezone=True), nullable=True),
+        # Composite, not a plain FK on org_provider_key_id alone: pins the
+        # referenced key to *this row's own* organization_id, so a cross-organization
+        # override (a workspace in org A pointing at org B's key) is a foreign-key
+        # violation rather than a silently-persisted row.
+        sa.ForeignKeyConstraint(
+            ["organization_id", "org_provider_key_id"],
+            ["org_provider_keys.organization_id", "org_provider_keys.id"],
+            ondelete="CASCADE",
+        ),
+        sa.ForeignKeyConstraint(["workspace_id"], ["workspace.id"], ondelete="CASCADE"),
+        sa.PrimaryKeyConstraint("id"),
+        sa.UniqueConstraint("workspace_id", "org_provider_key_id", name="uq_workspace_provider_key_overrides_ws_key"),
+    )
+    op.create_index(
+        op.f("ix_workspace_provider_key_overrides_workspace_id"),
+        "workspace_provider_key_overrides",
+        ["workspace_id"],
+    )
+    op.create_index(
+        op.f("ix_workspace_provider_key_overrides_org_provider_key_id"),
+        "workspace_provider_key_overrides",
+        ["org_provider_key_id"],
+    )
+
+    op.create_table(
+        "workspace_provider_model_restrictions",
+        sa.Column("id", sa.Uuid(), nullable=False),
+        sa.Column("workspace_id", sa.Uuid(), nullable=False),
+        # Same denormalization and reasoning as the override table above.
+        sa.Column("organization_id", sa.Uuid(), nullable=False),
+        sa.Column("org_provider_key_id", sa.Uuid(), nullable=False),
+        sa.Column("model", sa.String(length=255), nullable=False),
+        sa.Column("created_at", sa.DateTime(timezone=True), server_default=sa.func.now(), nullable=False),
+        sa.ForeignKeyConstraint(
+            ["organization_id", "org_provider_key_id"],
+            ["org_provider_keys.organization_id", "org_provider_keys.id"],
+            ondelete="CASCADE",
+        ),
+        sa.ForeignKeyConstraint(["workspace_id"], ["workspace.id"], ondelete="CASCADE"),
+        sa.PrimaryKeyConstraint("id"),
+        sa.UniqueConstraint(
+            "workspace_id",
+            "org_provider_key_id",
+            "model",
+            name="uq_workspace_provider_model_restrictions_ws_key_model",
+        ),
+    )
+    op.create_index(
+        op.f("ix_workspace_provider_model_restrictions_workspace_id"),
+        "workspace_provider_model_restrictions",
+        ["workspace_id"],
+    )
+    op.create_index(
+        op.f("ix_workspace_provider_model_restrictions_org_provider_key_id"),
+        "workspace_provider_model_restrictions",
+        ["org_provider_key_id"],
+    )
+
+
+def downgrade() -> None:
+    op.drop_index(
+        op.f("ix_workspace_provider_model_restrictions_org_provider_key_id"),
+        table_name="workspace_provider_model_restrictions",
+    )
+    op.drop_index(
+        op.f("ix_workspace_provider_model_restrictions_workspace_id"),
+        table_name="workspace_provider_model_restrictions",
+    )
+    op.drop_table("workspace_provider_model_restrictions")
+
+    op.drop_index(
+        op.f("ix_workspace_provider_key_overrides_org_provider_key_id"),
+        table_name="workspace_provider_key_overrides",
+    )
+    op.drop_index(
+        op.f("ix_workspace_provider_key_overrides_workspace_id"), table_name="workspace_provider_key_overrides"
+    )
+    op.drop_table("workspace_provider_key_overrides")
+
+    op.drop_index(_ORG_DEFAULT_INDEX, table_name="org_provider_keys")
+    op.drop_index(op.f("ix_org_provider_keys_organization_id"), table_name="org_provider_keys")
+    op.drop_table("org_provider_keys")
--- a/src/gateway/api/routes/_passthrough.py
+++ b/src/gateway/api/routes/_passthrough.py
@@ -65,7 +65,8 @@ from gateway.services.pricing_service import (
 )
 from gateway.services.provider_kwargs import ResolvedProvider, resolve_provider_selector
 from gateway.services.scoped_budget_service import BudgetScopeRequest
-from gateway.services.workspace_scope import organization_for_key_id, workspace_for_key_id
+from gateway.services.tenancy.org_provider_key_service import cached_org_model_restriction
+from gateway.services.workspace_scope import organization_for_workspace_id, resolve_workspace_id
 
 ResultT = TypeVar("ResultT")
 
@@ -209,10 +210,17 @@ async def run_passthrough(
     started_at = time.monotonic()
     api_key, is_master_key = auth_result
     api_key_id = api_key.id if api_key else None
-    # Resolved once for this request: it decides both what the model costs and
-    # whether the free-model shortcut applies, and a master-key caller's
-    # workspace lookup is deliberately never memoized.
-    organization_id = await organization_for_key_id(db, api_key_id)
+    # Zero I/O for a keyed request (see `_pipeline.resolve_request_context`'s
+    # equivalent comment); only a master-key request pays a lookup. Fixed for
+    # the whole request, so this also becomes the usage row's workspace below.
+    workspace_id = await resolve_workspace_id(db, api_key)
+    # Derived from the workspace already resolved above, not via
+    # `organization_for_key_id` (which would re-derive the same workspace from
+    # `api_key_id` internally): a master-key request's workspace lookup is
+    # deliberately never memoized, so re-deriving it here would pay that cost
+    # twice for one request. Decides both what the model costs and whether the
+    # free-model shortcut applies.
+    organization_id = await organization_for_workspace_id(db, workspace_id)
     # A key flagged exclude_from_budget logs cost but is never reserved or folded
     # into users.spend. Threaded through the reservation handle (so reconcile skips
     # the spend write) and stamped on the usage row.
@@ -312,7 +320,7 @@ async def run_passthrough(
         # The reservation is already held, so refund it before mapping an
         # unresolvable selector to 400; otherwise the estimate leaks.
         try:
-            resolved = resolve_provider_selector(config, model, user_id)
+            resolved = resolve_provider_selector(config, model, user_id, workspace_id=workspace_id)
         except (ValueError, AnyLLMError) as exc:
             await refund_reservation(db, reservation)
             await _log_rejection(
@@ -344,7 +352,7 @@ async def run_passthrough(
                 raise
     else:
         try:
-            resolved = resolve_provider_selector(config, model, user_id)
+            resolved = resolve_provider_selector(config, model, user_id, workspace_id=workspace_id)
         except (ValueError, AnyLLMError) as exc:
             # Nothing is reserved yet on this branch, so there is no refund to do.
             await _log_rejection(
@@ -412,9 +420,30 @@ async def run_passthrough(
             detail=not_allowed_detail,
         )
 
-    # Resolved here rather than in the closure below: the builder is
-    # synchronous, and the workspace is fixed for the whole request anyway.
-    usage_workspace_id = await workspace_for_key_id(db, api_key_id)
+    # Organization-scoped model restriction (otari#643): mirrors `_pipeline.py`'s
+    # equivalent gate. Applies only when the selector did not name a
+    # configured instance, the same condition `provider_kwargs.get_provider_kwargs`
+    # uses to decide whether to consult the organization overlay at all.
+    if workspace_id is not None and resolved.instance not in config.providers:
+        org_allowlist = cached_org_model_restriction(workspace_id, resolved.provider.value)
+        if org_allowlist is not None and resolved.model not in org_allowlist:
+            await refund_reservation(db, reservation)
+            not_allowed_detail = model_not_allowed_detail(model)
+            await _log_rejection(
+                not_allowed_detail,
+                row_model=resolved.model,
+                row_provider=resolved.instance,
+                status_code=status.HTTP_403_FORBIDDEN,
+            )
+            raise HTTPException(
+                status_code=status.HTTP_403_FORBIDDEN,
+                detail=not_allowed_detail,
+            )
+
+    # Already resolved above (needed there for organization-scoped provider
+    # credentials); reused here rather than re-derived, since it is fixed for
+    # the whole request either way.
+    usage_workspace_id = workspace_id
 
     def _usage_row(row_status: str, **outcome: Any) -> UsageLog:
         """Build this request's usage row, varying only the outcome columns.
--- a/src/gateway/api/routes/_pipeline.py
+++ b/src/gateway/api/routes/_pipeline.py
@@ -148,6 +148,7 @@ from gateway.services.sandbox_backend import (
     SandboxNotReachableError,
 )
 from gateway.services.scoped_budget_service import BudgetScopeRequest
+from gateway.services.tenancy.org_provider_key_service import cached_org_model_restriction
 from gateway.services.tool_usage import (
     MAX_TOOL_NAMES,
     OVERFLOW_TOOL_NAME,
@@ -157,8 +158,8 @@ from gateway.services.tool_usage import (
 from gateway.services.url_safety import UnsafeURLError, validate_mcp_url
 from gateway.services.web_search_backend import WEB_SEARCH_TOOL_NAME, WebSearchNotReachableError
 from gateway.services.workspace_scope import (
-    organization_for_key_id,
     organization_for_workspace_id,
+    resolve_workspace_id,
     workspace_for_key_id,
 )
 from gateway.streaming import (
@@ -567,6 +568,7 @@ class RequestContext:
         rate_limit_info: RateLimitInfo | None,
         reservation: ReservationHandle | None,
         started_at: float,
+        workspace_id: uuid.UUID | None = None,
         resolved_provider: ResolvedProvider | None = None,
         plan: CompiledPlan | None = None,
         estimate_inputs: "EstimateInputs | None" = None,
@@ -589,6 +591,12 @@ class RequestContext:
         # pay that per candidate.
         self.organization_id = organization_id
         self.user_id = user_id
+        # Standalone-only, `None` in hybrid mode: the workspace this request
+        # bills to, resolved once in the preamble (`resolve_workspace_id`) and
+        # reused here for the rare fallback resolve in `resolve_dispatch_provider`,
+        # so a request whose gate-check selector was unparseable still gets
+        # organization-scoped provider keys on its real dispatch attempt.
+        self.workspace_id = workspace_id
         self.rate_limit_info = rate_limit_info
         self.reservation = reservation
         # USD already written onto a failure row for gateway-run tool calls. A
@@ -672,7 +680,7 @@ async def resolve_dispatch_provider(
     if ctx.resolved_provider is not None:
         return ctx.resolved_provider
     try:
-        return resolve_provider_selector(config, model_selector, ctx.user_id)
+        return resolve_provider_selector(config, model_selector, ctx.user_id, workspace_id=ctx.workspace_id)
     except (ValueError, AnyLLMError) as exc:
         # The preamble deliberately tolerated this selector, so a reservation is
         # already held: refund it, then record the drop. The hold is not always
@@ -915,6 +923,7 @@ async def _compile_request_plan(
     endpoint: str,
     started_at: float,
     routing_signal: Callable[[], RoutingSignal] | None = None,
+    workspace_id: uuid.UUID | None = None,
 ) -> CompiledPlan | None:
     """Compile ``model`` into a plan when it names a routing policy, else ``None``.
 
@@ -967,6 +976,7 @@ async def _compile_request_plan(
             allowlist=allowlist,
             budget=budget,
             router_ordering=router_ordering,
+            workspace_id=workspace_id,
         )
     except NoEligibleCandidatesError as exc:
         logger.warning("%s", exc.operator_detail)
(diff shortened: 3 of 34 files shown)


src/gateway/services/tenancy/org_provider_key_service.py
```diff
@@ -0,0 +1,626 @@
...
+            await db.execute(
+                select(OrgProviderKey)
+                .where(col(OrgProviderKey.archived_at).is_(None))
+                .order_by(col(OrgProviderKey.created_at), col(OrgProviderKey.id))
+            )
+        )
+        .scalars()
+        .all()
+    )
+    overrides = (await db.execute(select(WorkspaceProviderKeyOverride))).scalars().all()
```
khaledosman, line 204: **`workspace_provider_model_restrictions` enforces nothing.** This refresh reads `Workspace`, `OrgProviderKey` and `WorkspaceProviderKeyOverride`; the restriction table is never queried here, and grepping the head branch, the only reads of `self.restrictions` are the four CRUD call sites (lines 514, 568, 591, 611). `provider_kwargs.py` never sees it, and neither does any dispatch path.

So a workspace allow-listed to `gpt-4o-mini` on a key still dispatches `openai:gpt-4o` through that same key. Concretely: `POST /v1/workspaces/{ws}/provider-keys/{key}/models` with `{"model": "gpt-4o-mini"}`, then a chat request for `openai:gpt-4o` from an API key in that workspace — the request succeeds and bills the org credential.

What makes this a finding rather than a stated deferral is that everything around it reads as enforcement: the route docstrings say "Narrow this workspace's allow-list", the model docstring explains allow-list-not-deny-list semantics, and two tests (`test_disabling_a_key_cascades_deleting_its_model_restrictions`, `test_restricting_models_on_a_disabled_key_is_refused`) treat the rows as load-bearing. Nothing in the PR says the table is inert. An operator who sets an allow-list will believe it holds.

Either enforce it — the natural spot is alongside the existing `api_keys.allowed_models` check, which would need the restriction set in the overlay, so the cache entry grows an `allowed_models` field populated from a fourth query here — or say plainly in the route descriptions and the module docstring that this is storage only, pending a follow-up issue. The first is better; the second is at least honest. Shipping it as-is is the one option I would rule out, because an access control that silently does not apply is worse than one that visibly does not exist yet.

src/gateway/services/provider_kwargs.py
```diff
@@ -85,6 +102,8 @@ def get_provider_kwargs(
     lookup = instance if instance is not None else provider.value
     kwargs: dict[str, Any] = {}
     raw_config = config.providers.get(lookup)
+    if raw_config is None and workspace_id is not None:
+        raw_config = cached_org_provider_kwargs(workspace_id, provider.value)
```
khaledosman, line 112: `workspace_id` reaches this fallback from the chat/messages/responses preamble and from `batches.py`, but not from the pass-through scaffold: `_passthrough.py:310` and `:338` still call `resolve_provider_selector(config, model, user_id)` with no workspace, and neither does `_dispatch_search`. Neither file is in this diff.

The result is an endpoint-dependent credential: a workspace whose only credential for a provider is an org-scoped key gets a working `/v1/chat/completions` and a credential-less failure on `/v1/embeddings`, `/v1/images/generations`, `/v1/audio/*`, `/v1/rerank`, `/v1/moderations` and `/v1/search`. `test_dispatch_without_workspace_id_ignores_organization_scoped_keys` pins that behavior without naming which endpoints land in it.

`run_passthrough` and `_dispatch_search` already resolve the `APIKey`, so threading it is the same one-line change as `batches.py`. If some of those are meant to be out of scope for v1, the module docstring is the place to say which endpoints org-scoped keys serve — right now it reads as "consulted when `workspace_id` is passed", which is true and tells a reader nothing about coverage.

src/gateway/api/routes/_pipeline.py
```diff
@@ -1049,6 +1059,13 @@ async def resolve_request_context(
             raise adapter.error(500, DB_UNAVAILABLE_DETAIL, ErrorKind.API)
         api_key, is_master_key = await verify_api_key_or_master_key(raw_request, db, config)
         api_key_id = api_key.id if api_key else None
+        # Zero I/O for a keyed request: `api_key.workspace_id` is already an
+        # in-memory attribute on the row just loaded. Only a master-key request
+        # pays a lookup, which `resolve_workspace_id` itself accepts as
+        # operator traffic (see its docstring). Used below to resolve
+        # organization-scoped provider keys for a bare provider selector; an
+        # instance-addressed one never consults it (`provider_kwargs.py`).
+        workspace_id = await resolve_workspace_id(db, api_key)
```
khaledosman, line 1111: For a keyed request this is genuinely free, as the comment says. For a **master-key** request it is not: `resolve_workspace_id` falls through to `default_workspace_id`, which is two indexed reads and is deliberately not memoized (`workspace_scope.py`'s module docstring explains why, and justifies the cost on the grounds that it is "operator traffic").

That justification assumed one lookup per request. This adds a second: the usage row still resolves its own workspace later through `workspace_for_key_id(None)`, which for a master-key request is the same un-memoized `default_workspace_id` call. Passing `ctx.workspace_id` down to the usage-row path instead of re-resolving would keep it at one.

Worth noting deliberately, since it is a behavior change and not just cost: a master-key inference request now resolves org-scoped provider keys belonging to the *default workspace's* organization for a bare `provider:model` selector. That is defensible for the deployment operator, but it is not stated anywhere and it is the kind of thing that surprises someone later.

khaledosman (commented):
Structurally this is sound where it matters most. I specifically checked for cross-tenant leakage and could not construct one: `refresh_org_provider_cache` builds each workspace's candidate list from `keys_by_org_provider[(organization_id, provider)]` where the organization comes off the workspace's own row, so a stray override pointing at a foreign key is never looked up; `get_in_organization` scopes the read rather than checking afterwards, so a cross-org id is indistinguishable from a missing one; and the disjointness claim holds — `lookup` is the instance name, so an `instance:model` selector that matches `config.providers` short-circuits before the overlay is consulted. The migration is clean: every FK indexed with an explicit `ondelete`, the partial unique index for one default per `(organization, provider)` with the `IntegrityError` mapping behind it, a real `downgrade`, and `archived_at` correctly on `_timestamp_field` for the asyncpg reason the comment gives. `resolve_active_key`'s three tiers each have a unit test.

Three inline findings. The first is the one that matters; I would not merge without either enforcing that table or saying in the API description that it does not enforce.

Two without a line to hang on:

- **No documentation.** `docs/api-reference.md`'s endpoint table gains none of the thirteen new routes, and `src/gateway/AGENTS.md`'s "Runtime credential stores" section still describes two overlays where this makes three — same shape, same TTL-refresher-in-the-lifespan pattern, different cache key, which is exactly what that section exists to tell the next reader. #640 and #668 both updated `api-reference.md`. The checklist points at otari-ai's architecture docs, but those are the decision record, not this repo's operator-facing surface.
- **Alembic chain:** five open PRs all set `down_revision = "a3c7e1b9d5f2"` (this one `e1c3a5b7d9f2`, plus #668, #671, #667, #640). The second to merge forks the chain and `alembic upgrade head` refuses to choose, which startup `auto_migrate` and `tests/integration/conftest.py` both run. Needs a merge order with the rest rebased.

One semantic sharp edge I looked at and would leave as ported, but is worth a comment at `resolve_active_key`: because tier 3 falls back to the earliest-created enabled key, `disabled` denies a workspace *that key*, not that provider. An operator who disables the credential they think of as the one that matters still reaches the provider through a sibling key. Correct per the tier ladder, surprising in the dashboard copy that will eventually sit on top of it.

---

src/gateway/api/routes/_pipeline.py
```diff
@@ -1168,6 +1187,36 @@ async def resolve_request_context(
...
+                    user_id=user_id,
+                    model=gate_model,
+                    provider=gate_instance,
+                    endpoint=adapter.endpoint,
+                    detail=not_allowed_detail,
+                    status_code=status.HTTP_403_FORBIDDEN,
+                    started_at=started_at,
+                )
+                raise adapter.error(403, not_allowed_detail, ErrorKind.PERMISSION)
+
```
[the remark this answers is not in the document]
tbille: This is already enforced end to end, not only at `plan.head`.

`compile_policy` (`services/routing/compiler.py`) checks `cached_org_model_restriction` inside its per-candidate loop over `ordered` (selected head + every `on_failure` entry), dropping any candidate that violates the workspace's organization-key restriction before it can become an attempt — including a candidate that would otherwise have been promoted to `plan.head` after an earlier one was dropped. `test_a_fallover_candidate_violating_the_org_restriction_is_dropped` (`tests/unit/test_routing_compiler.py`) pins exactly this: a restricted first candidate is dropped and does not silently let the next one through unchecked.

The check at `_pipeline.py:1200-1221` you're pointing at runs earlier, before `_compile_request_plan` is even called, against the plain-model/static-policy resolution (`gate_instance`/`gate_model`) -- it's the single-candidate-equivalent gate, not a substitute for the compiler's per-candidate one. For a routing policy with a router or `on_failure` chain, `resolve_provider_selector` deliberately fails to resolve (a dynamic policy "is left unresolved" by design), so `gate_instance` is `None` there and this block is skipped entirely for that path; enforcement for that path is exactly the compiler's per-candidate check.

So the two blocks are not overlapping/duplicated views of the same thing, they're the head-of-plan check (this file) and the every-attempt check (`compiler.py`), covering the plain-model and routing-policy paths respectively. Let me know if you're seeing a concrete request shape that gets through both.

alembic/versions/c7e9a1f3b5d0_merge_user_credentials_and_budget_.py
```diff
@@ -0,0 +1,29 @@
...
+
+Revision ID: c7e9a1f3b5d0
+Revises: f2a4c6d8b0e3, b6e8c2a4d7f1
+Create Date: 2026-08-20 00:00:00.000000
+"""
+
+from collections.abc import Sequence
+
+revision: str = "c7e9a1f3b5d0"
+down_revision: str | Sequence[str] | None = ("f2a4c6d8b0e3", "b6e8c2a4d7f1")
```
njbrake: Duplicate merge revision. #673 landed `c8f2a6b4e9d3` on main naming the same two parents (`b6e8c2a4d7f1`, `f2a4c6d8b0e3`), so once this branch takes main the graph forks again. I reconstructed it and `ScriptDirectory.get_heads()` returns `['7c5ba82a601b', 'c8f2a6b4e9d3']`: `alembic upgrade head` exits 1 and `tests/unit/test_migration_single_head.py` fails.

Fix: drop this file, point `e1c3a5b7d9f2.down_revision` at `c8f2a6b4e9d3`, and update that revision's docstring, which currently narrates the rebase onto `c7e9a1f3b5d0`.

src/gateway/api/routes/batches.py
```diff
@@ -414,10 +451,13 @@ async def retrieve_batch(
 ) -> dict[str, Any]:
     """Retrieve the status of a batch."""
     api_key, is_master_key = auth_result
-    provider_enum, provider_kwargs = _resolve_batch_provider(config, provider)
+    record = await get_batch_record(db, batch_id)
+    provider_enum, provider_kwargs = _resolve_batch_provider(
+        config, provider, workspace_id=await _lifecycle_workspace_id(db, record, api_key)
```
njbrake, line 509: Cross-tenant credential use ahead of authorization. `get_batch_record` is keyed on `batch_id` alone with no tenancy scoping, so `_lifecycle_workspace_id` returns the batch owner's workspace even when the caller belongs to a different organization. The upstream call on line 459 then runs with that organization's BYO key, and `_authorize_batch` only refuses afterward. `cancel_batch` and `retrieve_batch_results` have the same ordering.

Before this PR the credential here was deployment-global, so the pre-auth call carried nothing tenant-specific; now a 404'd caller can make the gateway spend another organization's provider credential.

The strict path does not need the provider batch in order to authorize: only the legacy record-less fallback does, and that fallback already resolves the caller's own workspace. Authorizing on `record` first when it is present, then resolving credentials, closes this.

src/gateway/api/routes/_passthrough.py
```diff
@@ -307,7 +311,7 @@ async def _reserve(estimated_cost: float, *, row_model: str, row_provider: str |
         # The reservation is already held, so refund it before mapping an
         # unresolvable selector to 400; otherwise the estimate leaks.
         try:
-            resolved = resolve_provider_selector(config, model, user_id)
+            resolved = resolve_provider_selector(config, model, user_id, workspace_id=workspace_id)
```
njbrake, line 323: The organization model allow-list is not enforced on this surface. This resolves the org credential for a bare selector, but `cached_org_model_restriction` appears nowhere in this file, so a workspace restricted to one model can dispatch any model on the same org key through `/v1/embeddings`, `/v1/images/*`, `/v1/audio/*`, `/v1/rerank` and `/v1/moderations`. `_pipeline.py:1198` has the gate; this scaffold needs the equivalent, and a test for it.

src/gateway/api/routes/batches.py
```diff
@@ -268,8 +301,11 @@ async def create_batch(
 
     rate_limit_info = check_rate_limit(raw_request, user_id)
 
+    # Zero I/O for a keyed request (see `_pipeline.resolve_request_context`'s
+    # equivalent comment); only a master-key request pays a lookup.
+    workspace_id = await resolve_workspace_id(db, api_key)
     try:
-        resolved = resolve_provider_selector(config, request.model, user_id)
+        resolved = resolve_provider_selector(config, request.model, user_id, workspace_id=workspace_id)
```
njbrake, line 341: Same gap on batch creation. The org credential resolves here, and the only allow-list check below is the per-key `is_model_allowed`, so a workspace restricted to `gpt-4o` can still submit a batch for any model on the org key.

njbrake (commented):
Five findings; three of them look blocking to me.

Verified locally against `e2361de`: `make lint`, `make openapi-check`, `make postman-check`, a SQLite `upgrade`/`downgrade`/`upgrade` roundtrip, `compare_metadata` against the model metadata (the three new tables and `batches.workspace_id` match, on SQLite and in the rendered Postgres DDL), and the unit tests this PR touches (88 passed). No Docker in my environment, so the Postgres integration suite and the `web/src/client/schema.ts` drift check are unverified.

**Alembic heads (inline).** `c7e9a1f3b5d0` duplicates #673's `c8f2a6b4e9d3`: same two parents. I reconstructed the post-merge graph and `ScriptDirectory.get_heads()` returns `['7c5ba82a601b', 'c8f2a6b4e9d3']`, so `alembic upgrade head` exits 1 again and `tests/unit/test_migration_single_head.py` (new on main) goes red.

**Cross-tenant credential on the batch lifecycle (inline).** `retrieve`/`cancel`/`results` resolve the batch's origin workspace and call the provider with that organization's key before `_authorize_batch` runs.

**Model allow-list enforced on two dispatch surfaces of four (inline, two places).** `_pipeline.py` and `compiler.py` check it; `_passthrough.py` and `create_batch` resolve the org credential without it.

**Provider name unvalidated (inline).** A typo'd or aliased `provider` is accepted with a 201 and then silently never resolves at dispatch.

**Docs.** Thirteen endpoints and a new access-control layer land with no user-facing documentation. `docs/configuration.md` ("Runtime provider management") and `docs/access-control.md` both describe the surfaces this sits beside, and neither mentions organization-scoped keys or the workspace model allow-list.

The design fit is good otherwise. The claimed disjointness from `provider_credentials` holds where it claims to (I checked `instance:model` against a bare selector both ways), hybrid mode is genuinely untouched because every new startup hook sits inside `main.py`'s standalone branch, and the authorization reuses `WorkspaceService`'s existing two-level check rather than inventing a parallel one.

src/gateway/services/tenancy/org_provider_key_service.py
```diff
@@ -0,0 +1,727 @@
...
+        self,
+        *,
+        user: User,
+        request: OrgProviderKeyCreateRequest,
+    ) -> OrgProviderKeyPublic:
+        """Create a key in the caller's organization. Organization owners and admins only."""
+        organization = await self.organizations.get_active_organization_for_user(user)
+        await self.organizations.require_active_organization_management_access(user=user, organization=organization)
+
+        name = _validated_name(request.name)
```
njbrake, line 405: `request.provider` is stored verbatim and is what the dispatch cache is keyed by, matched against `LLMProvider.value`. Nothing validates it, so `provider="OpenAI"`, `"azure-openai"`, or a value with stray whitespace is accepted with a 201 and then never resolves at dispatch, with no error at either point.

`/v1/provider-credentials` refuses these through `_validate_instance` (blank name, `:` or `/`, unknown `provider_type`) and honors `PROVIDER_TYPE_ALIASES`. Worth the same guard here, given the PR positions this surface as following that route's conventions.

src/gateway/services/tenancy/org_provider_key_service.py
```diff
@@ -0,0 +1,727 @@
...
+    return entry
+
+
+def cached_org_provider_kwargs(workspace_id: uuid.UUID, provider: str) -> dict[str, Any] | None:
+    """The decrypted overlay entry this worker last loaded for this workspace+provider, if any."""
+    entry = _org_cache.get((workspace_id, provider))
+    return dict(entry) if entry is not None else None
+
+
+def cached_org_model_restriction(workspace_id: uuid.UUID, provider: str) -> list[str] | None:
```
njbrake, line 141: Worth stating in the docstring: this is now an access-control input, not only a credential cache, so tightening an allow-list takes up to `ORG_PROVIDER_CACHE_TTL_SECONDS` to reach sibling workers and replicas. The credential and the allow-list are also two separate reads on the request path, so a refresh landing between them can pair one key's credential with another key's list.

njbrake (approved):
Approving to let you get this over the finish line and sync this up alongside 671 🚀

alembic/versions/c7e9a1f3b5d0_merge_user_credentials_and_budget_.py
tbille: Fixed. Dropped the redundant `c7e9a1f3b5d0` merge revision and re-chained `e1c3a5b7d9f2` onto `main`'s own merge (`c8f2a6b4e9d3`) instead — confirmed via `alembic heads` back down to a single head, and an upgrade/downgrade/upgrade roundtrip. The branch has since rebased through several more `main` heads (org pricing, invitations, dashboard sessions, workspace budget defaults) as they landed; each rebase re-parents the same way rather than leaving a second merge revision beside `main`'s.

src/gateway/api/routes/batches.py
tbille: Fixed. Split `_authorize_batch` into `_authorize_record` (record-only, no upstream call needed — checked *before* `_resolve_batch_provider`/`_retrieve_batch_or_502` now run) and `_authorize_legacy_batch` (the metadata-anchored fallback, which still runs after the provider call since a record-less batch has no stored workspace to leak — credentials there already resolve to the caller's own). All three lifecycle routes (`retrieve_batch`, `cancel_batch`, `retrieve_batch_results`) now call `_authorize_record` first when a record exists, and only fetch from the provider afterward.

Added regression tests proving the ordering, not just the 404: `test_recorded_batch_foreign_caller_never_reaches_the_provider` and its cancel/results counterparts assert `aretrieve_batch` was never awaited for a foreign caller, which is the only way to see this specific bug (asserting the 404 alone would have passed either way, before or after the fix).

src/gateway/api/routes/_passthrough.py
tbille: Fixed. Added the same gate right after the existing per-key allow-list check in `run_passthrough`: when `workspace_id is not None and resolved.instance not in config.providers`, look up `cached_org_model_restriction(workspace_id, resolved.provider.value)` and refund + reject (403) if the model isn't in it. Covers embeddings, images, audio, rerank, and moderations in one place, since they all funnel through this shared scaffold.

src/gateway/api/routes/batches.py
tbille: Fixed. Same gate added right after the per-key allow-list check in `create_batch`, using `provider`/`model`/`resolved.instance` from the selector already resolved just above it.

src/gateway/services/tenancy/org_provider_key_service.py
tbille: Fixed. Added `_validated_provider`: strips, applies `PROVIDER_TYPE_ALIASES` (so `"azure-openai"` still resolves the same as it does for a configured instance), then requires `LLMProvider(canonical)` to succeed — 400 via a new `OrgProviderKeyUnknownProviderError` otherwise. The canonical value is what gets stored, not the alias, so the cache key (`refresh_org_provider_cache`, keyed on the raw `provider` string) always matches what `resolved.provider.value` produces at dispatch. Wired into `create_key_for_user` only, since `provider` isn't in `OrgProviderKeyUpdateRequest` — it's immutable after creation.

src/gateway/services/tenancy/org_provider_key_service.py
tbille: Added to `cached_org_model_restriction`'s docstring: it's an access-control input, not only a credential cache, so a narrowed allow-list takes up to `ORG_PROVIDER_CACHE_TTL_SECONDS` to reach a sibling worker/replica, and the credential + allow-list are two separate reads on the request path with no shared snapshot, so a refresh landing between them can pair one key's credential with a different key's list for one request. Called out as a real but bounded trade-off (the TTL caps the staleness window; the pairing mismatch is one request, not standing state), not a guarantee this cache doesn't make.

src/gateway/services/tenancy/errors.py
```diff
@@ -180,9 +180,106 @@ class LastWorkspaceError(TenancyValidationError):
...
+class SecretBoxUnavailableTenancyError(TenancyValidationError):
+    """`OTARI_SECRET_KEY` is not configured, so a provider key cannot be stored.
+
+    Wraps `services.secret_box.SecretBoxUnavailableError` as a tenancy error so
+    the route stays thin (see the module docstring): the underlying error
+    carries no key material, and neither does this one.
+    """
+
+    def __init__(self) -> None:
+        super().__init__("OTARI_SECRET_KEY is not set; it is required to store provider credentials")
```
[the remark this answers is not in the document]
tbille: Fixed. Changed the base class from `TenancyValidationError` to `TenancyError` directly (which defaults to 500), matching `ForeignTenancyError`'s precedent for the same reasoning: a deployment configuration gap the caller cannot fix and shouldn't be blamed for, and needs to surface in 5xx alerting.

src/gateway/services/tenancy/org_provider_key_service.py
```diff
@@ -0,0 +1,727 @@
...
+        ``client_args`` is redacted for a plain member: this read is open to
+        every active member, not only owners/admins, and `OrgProviderKey.to_public`
+        explains why that field cannot be assumed safe for that wider audience.
+        """
+        organization = await self.organizations.get_active_organization_for_user(user)
+        rows = await self.keys.list_for_organization(organization.id, include_archived=include_archived)
+        membership = await self.organizations.members.get_active_by_organization_and_user(organization.id, user.id)
+        is_admin = user.is_superuser or (membership is not None and membership.role in MANAGEMENT_ROLES)
+        data = [row.to_public(include_client_args=is_admin) for row in rows]
+        return OrgProviderKeysPublic(data=data, count=len(rows))
```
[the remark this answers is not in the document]
tbille: Partially addressed. Added `skip`/`limit` (default 100, `le=1000`) to the organization-level list (`list_org_provider_keys` / `list_keys_for_user` / `OrgProviderKeyRepository.list_for_organization`), which returns the total in `count` now (mirrors `WorkspaceService.list_workspaces`'s shape) rather than the page size.

Deliberately left `list_effective_keys_for_workspace` (and the model-restriction list) unpaginated for now: that endpoint groups every candidate by provider and resolves precedence (`resolve_active_key`) per group to compute `is_effective_default`. Paginating the underlying rows at the SQL level risks splitting a provider's candidate group across pages, which would make the wrong key look "effective" on a partial page — a correctness bug, not just a completeness one. Getting that right (page over resolved provider groups, not raw rows) is real additional work I didn't want to rush alongside the other fixes in this round; happy to open a follow-up issue for it rather than let it block this PR further, if that's preferable to doing it now.

alembic/versions/e1c3a5b7d9f2_add_org_provider_keys.py
```diff
@@ -0,0 +1,157 @@
...
+        sa.ForeignKeyConstraint(["org_provider_key_id"], ["org_provider_keys.id"], ondelete="CASCADE"),
+        sa.ForeignKeyConstraint(["workspace_id"], ["workspace.id"], ondelete="CASCADE"),
+        sa.PrimaryKeyConstraint("id"),
+        sa.UniqueConstraint(
+            "workspace_id",
+            "org_provider_key_id",
+            "model",
+            name="uq_workspace_provider_model_restrictions_ws_key_model",
+        ),
+    )
```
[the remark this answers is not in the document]
tbille: Fixed. Added a unique constraint on `org_provider_keys(organization_id, id)` and switched both link tables' FK on `org_provider_key_id` to a composite `(organization_id, org_provider_key_id) -> org_provider_keys(organization_id, id)`, with `organization_id` denormalized onto both link tables (set from the workspace the service layer already resolved, which by construction always agrees with the key's own organization). A cross-organization override or restriction is now a foreign-key violation, not a silently-persisted row. Folded into the still-unmerged `e1c3a5b7d9f2` migration rather than a new one, since these tables have no data yet. Verified with a fresh upgrade/downgrade/upgrade roundtrip and the full test suite.

web/src/client/schema.ts
```diff
@@ -8932,7 +9268,7 @@ export interface operations {
                     [name: string]: unknown;
                 };
                 content: {
-                    "application/json": components["schemas"]["PricingResponse"][];
+                    "application/json": components["schemas"]["OrgProviderKeysPublic"];
                 };
```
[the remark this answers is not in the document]
tbille: Addressed for the organization-level list endpoint (`GET /v1/organizations/me/provider-keys` now takes `skip`/`limit`, default 100/`le=1000`, regenerated into this file). See my reply on the sibling org_provider_key_service.py comment for why the workspace-level effective-view endpoint is deliberately left unpaginated for now.

src/gateway/api/routes/org_provider_keys.py
```diff
@@ -0,0 +1,232 @@
...
+
+
+@workspace_router.get("")
+async def list_workspace_provider_keys(
+    service: OrgProviderKeyServiceDep,
+    current_identity: CurrentIdentity,
+    workspace_id: uuid.UUID,
+) -> WorkspaceProviderKeyOverridesPublic:
+    """The effective view of every key visible to this workspace. Any member of the workspace may read it."""
+    return await service.list_effective_keys_for_workspace(user=current_identity, workspace_id=workspace_id)
```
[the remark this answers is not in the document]
tbille: `list_workspace_provider_key_model_restrictions` deliberately kept unpaginated too, for a different reason than the effective-keys view: it's a complete allow-list, not a browsable collection. A client evaluating access needs the whole set to reason correctly, so returning a page of it would be actively misleading — a caller seeing 100 of 150 restricted models could conclude a model is unrestricted when it's simply on a later page. The realistic growth path here is also narrow: there's no bulk-add endpoint, so the list only grows one `POST .../models` call at a time, which self-limits how large it practically gets for one workspace+key.

If this needs bounding in the future, I'd want to change its shape deliberately (e.g. keep an unpaginated 'give me the whole allow-list' read for correctness, and add a separate paginated 'browse' endpoint if one is ever needed for a UI), not thread `skip`/`limit` onto the existing one — that would silently turn a complete-allow-list contract into a partial one for every existing caller. Given the practical risk here is low, I'd rather not rush that shape change in this PR; happy to track it separately.

PR #671, opened 2026-08-20
feat(tenancy): workspace per-member budget defaults

### LINKED ISSUE #642: Workspace per-member budget defaults
- Part of #589, replaces part of #588

## Why this exists

`scoped_budgets` (#633) holds concrete budget rows keyed on scope and provider key. It has no notion of a template.

`otari-ai`'s `budget_policy_service` manages `BudgetDefault`: a workspace-level per-member spend template, eagerly materialised into a real budget row when a member joins, or when a default is created on a workspace that already has members, so the budget is visible in the UI immediately rather than on first spend.

Nothing here does that today, and nothing else in the tenancy port supplies it.

## What to produce

- The default table, and the materialisation on member join and on default creation.
- CRUD routes and the workspace-settings UI.
- Keep the `WorkspaceMemberBudgetPolicy*` DTO names, so the generated client stays recognisable across both trees.

## Definition of Done

- Creating a default on a populated workspace materialises one budget per member; a member joining afterwards gets one.
- Editing a default is specified either way, retroactive or not, and tested for whichever is chosen.
- Deleting a default states what happens to rows it already materialised.

## Dependencies

None. This is the piece of the tenancy port that can start immediately.

tbille:
## Description

Adds a workspace-level template for the per-member `scoped_budgets` ceiling (`scoped_budgets`, #633, has no notion of a template).

- `WorkspaceBudgetDefault` model + migration: one row per `(workspace_id, provider_key_id)`, no counters of its own.
- Materialization: creating a default fans it out into a `ScopedBudget(scope_type="workspace_member")` row for every existing active member; a member joining afterwards (`WorkspaceService.add_member`, workspace creation, or organization-member creation with `workspace_assignments`) gets one from every default already on the workspace. A member-specific override always wins over the template.
- Editing a default is **not retroactive**: members already materialized keep the value in effect at the time; only a member materialized afterwards sees the new one.
- Deleting a default **preserves** the `ScopedBudget` rows it already produced; a member joining afterwards no longer gets one from it.
- `services/tenancy/authorization.py` extracts `WorkspaceService`'s workspace-visibility and workspace-management checks into a shared module, so the new service enforces the same rules rather than a second, driftable copy.
- Routes: `/v1/workspaces/{workspace_id}/member-budget-policies` (master-key, standalone only), DTOs named `WorkspaceMemberBudgetPolicy*` per the issue, so the generated client stays recognizable against `otari-ai`'s hosted equivalent. That equivalent's now-removed CRUD surface (`208a552c`, `4ecdc263` in `otari-ai`'s history) is what the edit/delete semantics above are ported from.
- Dashboard: a "Budget defaults" page/panel beside Workspace Members, gated on the `workspaces` surface.
- Hardened against review findings: a `WorkspaceRepository.lock` closes the race between creating a default and a member joining concurrently; `materialize_for_default` batches its per-page existence check and insert, with a savepoint-based fallback so a single colliding row (e.g. a concurrent direct write to `/v1/scoped-budgets`) never fails the whole default; a suspended membership revived via `workspace_assignments` is materialized too; list endpoints are paginated; `budget_duration_sec` is bounded; the dashboard preserves a reset period it cannot represent in its days-only field instead of silently clearing it on save; and `canManageWorkspace` matches the server's organization-or-workspace-role rule (also applied to the pre-existing Members page, which had the same gap).

### Why a local (not top-level) import in `organization_service.py`

`WorkspaceBudgetDefaultService` reaches back into `OrganizationService` for its own authorization checks. A top-level import of it from `organization_service.py` would close that into a real cycle (A imports B, B imports A), so `_apply_workspace_assignments` imports it locally instead, the same kind of avoidance `workspace_service.py` already documents for `ScopedBudget` deletion. `tests/unit/test_service_module_imports.py` is extended to pin the new modules importing cleanly first, in a fresh interpreter.

### Testing

- `tests/integration/test_workspace_member_budget_policies.py`: materialization on create/join through all three membership-creation paths, member-override precedence, non-retroactive edit, delete-preserves-rows, uniqueness conflict, authorization (management vs. plain member, foreign workspace), a two-session race between creating a default and a member joining (confirmed to fail reliably without the lock and pass reliably with it).
- Regression: `test_scoped_budgets.py`, `test_tenancy_authorization.py`, `test_tenancy_races.py`, `test_tenancy_api.py`, `test_workspace_scope.py`, `test_service_module_imports.py`, full unit suite, all pass.
- Migration verified both directions against real PostgreSQL (upgrade, downgrade, re-upgrade).
- `make lint`, `make typecheck`, `make openapi-check`, `make postman-check`, `scripts/oss_edition_smoke.py`, all pass.
- Dashboard: `pnpm run lint`, `pnpm run typecheck`, `pnpm test`, `pnpm run build`, all pass.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Closes #642

## Checklist

- [x] **M4 ledger:** this PR either needs no entry (an existing row already covers it and its target is unchanged), or the entry is appended to [otari-ai#1587](https://github.com/mozilla-ai/otari-ai/issues/1587). One entry per *change*, not per PR: a stack delivering one surface change gets one. See [.github/instructions/reconciliation-ledger.instructions.md](https://github.com/mozilla-ai/otari/blob/main/.github/instructions/reconciliation-ledger.instructions.md).

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:** Claude Code (Claude Sonnet 5), via a background agent session.

**Any additional AI details you'd like to share:** Implementation, review-comment triage, and fixes were all done by the agent; a human (@tbille) is the operator directing the session.

## Summary

Adds workspace-level per-member budget defaults for `scoped_budgets`.

- Adds controls to create, edit, and delete workspace defaults.
- Applies defaults to active members when defaults are created, members join, or memberships are reactivated.
- Preserves member-specific overrides and existing budgets.
- Makes default updates non-retroactive.
- Adds authorization, validation, pagination, concurrency handling, SDK support, and database migration updates.
- Adds dashboard navigation, UI controls, generated client types, and comprehensive tests.

### DIFF
--- a/src/gateway/services/tenancy/organization_service.py
+++ b/src/gateway/services/tenancy/organization_service.py
@@ -535,9 +535,28 @@ class OrganizationService:
         An existing membership is updated rather than skipped, as the platform's
         own assignment path does: a suspended row that is left alone would leave
         the member listed in a workspace they were just granted while still
-        being refused everything in it.
+        being refused everything in it. Reviving one is materialized exactly
+        as creating one is: a suspended member could have missed a default
+        created while they were out, and the revive is the only signal that
+        they are back to being covered by the workspace's defaults again.
+
+        Each target workspace is locked (`WorkspaceRepository.lock`) before
+        its create-or-revive-and-materialize step, same as
+        `WorkspaceService.add_member`, and in a stable order (`wanted` is
+        walked sorted by id, not in insertion order) so two requests naming
+        the same workspaces in different orders cannot deadlock each other.
+        Imported locally, not at module top: `WorkspaceBudgetDefaultService`
+        reaches back to `OrganizationService` (for its own authorization
+        checks), and importing it at the top of this module would close that
+        into a real cycle. See `tests/unit/test_service_module_imports.py`.
         """
+        from gateway.services.tenancy.workspace_budget_default_service import (
+            WorkspaceBudgetDefaultService,
+        )
+
         members = WorkspaceMemberRepository(self.db)
+        workspaces = WorkspaceRepository(self.db)
+        budget_defaults = WorkspaceBudgetDefaultService(self.db)
         wanted: dict[uuid.UUID, str] = {}
         for assignment in assignments:
             wanted.setdefault(assignment.workspace_id, assignment.role)
@@ -548,12 +567,18 @@ class OrganizationService:
         existing_by_workspace = {
             member.workspace_id: member for member in await members.get_by_workspaces_and_user(wanted, user_id)
         }
-        for workspace_id, role in wanted.items():
+        for workspace_id, role in sorted(wanted.items(), key=lambda item: str(item[0])):
+            # Serialized against a concurrent `WorkspaceBudgetDefaultService.create_default`
+            # on this workspace; see `WorkspaceService.add_member`'s identical lock
+            # for why.
+            await workspaces.lock(workspace_id)
             existing = existing_by_workspace.get(workspace_id)
             if existing is not None:
-                await members.update(existing, WorkspaceMemberUpdate(role=role, status="active"))
+                revived = await members.update(existing, WorkspaceMemberUpdate(role=role, status="active"))
+                await budget_defaults.materialize_for_member(revived)
                 continue
-            await members.create(workspace_id=workspace_id, user_id=user_id, role=role)
+            member = await members.create(workspace_id=workspace_id, user_id=user_id, role=role)
+            await budget_defaults.materialize_for_member(member)
 
     async def invite_active_organization_member_for_user(
         self,
--- a/src/gateway/services/tenancy/organization_service.py
+++ b/src/gateway/services/tenancy/organization_service.py
@@ -539,6 +539,11 @@ class OrganizationService:
         as creating one is: a suspended member could have missed a default
         created while they were out, and the revive is the only signal that
         they are back to being covered by the workspace's defaults again.
+        Gated on the row actually having been inactive: re-applying the same
+        assignment to an already-active membership (a repeat invitation
+        accept, say) is not a join, and materializing it would resurrect a
+        per-member ceiling an admin deliberately deleted through
+        `/v1/scoped-budgets`.
 
         Each target workspace is locked (`WorkspaceRepository.lock`) before
         its create-or-revive-and-materialize step, same as
@@ -574,8 +579,10 @@ class OrganizationService:
             await workspaces.lock(workspace_id)
             existing = existing_by_workspace.get(workspace_id)
             if existing is not None:
+                was_inactive = existing.status != "active"
                 revived = await members.update(existing, WorkspaceMemberUpdate(role=role, status="active"))
-                await budget_defaults.materialize_for_member(revived)
+                if was_inactive:
+                    await budget_defaults.materialize_for_member(revived)
                 continue
             member = await members.create(workspace_id=workspace_id, user_id=user_id, role=role)
             await budget_defaults.materialize_for_member(member)
--- a/src/gateway/services/tenancy/workspace_budget_default_service.py
+++ b/src/gateway/services/tenancy/workspace_budget_default_service.py
@@ -75,7 +75,7 @@ def _period_window(budget_duration_sec: int | None) -> tuple[datetime | None, da
     of two; worth reconciling further, into one definition shared with
     ``scoped_budgets.py`` itself, wherever that shared home ends up landing.
     """
-    if not budget_duration_sec:
+    if budget_duration_sec is None:
         return None, None
     now = datetime.now(UTC)
     return now, now + timedelta(seconds=budget_duration_sec)
@@ -161,7 +161,14 @@ class WorkspaceBudgetDefaultService:
         """Create a per-member ``ScopedBudget`` for every default on the member's workspace.
 
         Skips any ``provider_key_id`` the member already has a ceiling for; a
-        member-specific override always wins over the template.
+        member-specific override always wins over the template. Never raises
+        ``IntegrityError``: a collision on one default (see
+        :meth:`_insert_member_budgets`) is swallowed there, precisely so this
+        method's callers (``WorkspaceService.add_member``,
+        ``OrganizationService._apply_workspace_assignments``) can keep their
+        own ``except IntegrityError`` narrow to the membership row they
+        actually insert, without a materialization-time collision escaping
+        and being misreported as a duplicate membership.
         """
         defaults = (
             (
@@ -178,7 +185,7 @@ class WorkspaceBudgetDefaultService:
         for default in defaults:
             if await self._existing_member_budget(member.id, default.provider_key_id) is not None:
                 continue
-            created.append(await self._create_member_budget(member.id, default))
+            created.extend(await self._insert_member_budgets([member.id], default))
         return created
 
     async def materialize_for_default(self, default: WorkspaceBudgetDefault) -> list[ScopedBudget]:
@@ -204,28 +211,20 @@ class WorkspaceBudgetDefaultService:
                 break
             active_ids = [member.id for member in members if member.status == "active"]
             if active_ids:
-                created.extend(await self._create_member_budgets(active_ids, default))
+                created.extend(await self._materialize_batch(active_ids, default))
             skip += len(members)
             if skip >= total:
                 break
         return created
 
-    async def _create_member_budgets(
+    async def _materialize_batch(
         self, member_ids: list[uuid.UUID], default: WorkspaceBudgetDefault
     ) -> list[ScopedBudget]:
         """Materialize ``default`` onto every id in ``member_ids`` that does not already have one.
 
-        One query for the whole batch's existing rows, one flush for the whole
-        batch's new ones, in the common case. The existence check and the
-        insert are still two round trips with nothing locking the gap between
-        them, and the one surface that can land in it (a direct
-        ``POST /v1/scoped-budgets`` for one of these members, racing this
-        fan-out) takes no lock on this workspace: `routes/scoped_budgets.py`
-        is a master-key admin surface with no notion of one. A batch that
-        collides falls back to inserting one row at a time, each in its own
-        savepoint, so a single colliding member costs one skipped row rather
-        than failing the whole page, and, one level up, the default's own
-        creation with it (see :meth:`create_default`).
+        One query for the whole batch's existing rows: the insert itself is
+        :meth:`_insert_member_budgets`, which is what actually tolerates a
+        collision on any one id.
         """
         existing_stmt = select(ScopedBudget.scope_id).where(
             ScopedBudget.scope_type == _SCOPE_WORKSPACE_MEMBER,
@@ -240,39 +239,67 @@ class WorkspaceBudgetDefaultService:
         candidates = [member_id for member_id in member_ids if str(member_id) not in already_covered]
         if not candidates:
             return []
+        return await self._insert_member_budgets(candidates, default)
 
-        def build(member_id: uuid.UUID) -> ScopedBudget:
-            period_start, period_end = _period_window(default.budget_duration_sec)
-            return ScopedBudget(
-                scope_type=_SCOPE_WORKSPACE_MEMBER,
-                scope_id=str(member_id),
-                provider_key_id=default.provider_key_id,
-                max_budget=default.max_budget,
-                budget_duration_sec=default.budget_duration_sec,
-                period_start=period_start,
-                period_end=period_end,
-            )
+    async def _insert_member_budgets(
+        self, member_ids: list[uuid.UUID], default: WorkspaceBudgetDefault
+    ) -> list[ScopedBudget]:
+        """Insert a ``ScopedBudget`` for each id in ``member_ids``, tolerating a uniqueness collision on any one.
 
-        budgets = [build(member_id) for member_id in candidates]
-        self.db.add_all(budgets)
+        The existence check the two callers run first is a separate round
+        trip with nothing locking the gap after it, and the one surface that
+        can land there (a direct ``POST /v1/scoped-budgets`` for one of these
+        members, racing this insert) takes no lock on the workspace:
+        `routes/scoped_budgets.py` is a master-key admin surface with no
+        notion of one. One flush for the whole batch in the common case;
+        falls back to one row at a time, each in its own savepoint, on
+        conflict, so a single colliding id costs one skipped row rather than
+        the whole batch.
+
+        Each attempt's ``add()`` happens *inside* its ``begin_nested()``
+        block, not before it: ``AsyncSession.begin_nested()`` unconditionally
+        flushes whatever is already pending before it opens the ``SAVEPOINT``,
+        so an add staged beforehand is flushed *outside* any savepoint, and a
+        conflict there fails the outer transaction rather than just rolling
+        back to the point the savepoint was meant to protect. The same reason
+        rules out an explicit ``expunge`` in the ``except`` branch: a
+        savepoint rollback already expunges the row it was protecting, so a
+        second, manual expunge finds nothing there and raises
+        ``InvalidRequestError`` instead.
+        """
         try:
             async with self.db.begin_nested():
+                budgets = [self._build_member_budget(member_id, default) for member_id in member_ids]
+                self.db.add_all(budgets)
                 await self.db.flush()
             return budgets
         except IntegrityError:
             pass  # fall back below; the savepoint already rolled the batch back
 
         created: list[ScopedBudget] = []
-        for member_id in candidates:
-            budget = build(member_id)
-            self.db.add(budget)
+        for member_id in member_ids:
             try:
                 async with self.db.begin_nested():
+                    budget = self._build_member_budget(member_id, default)
+                    self.db.add(budget)
                     await self.db.flush()
                 created.append(budget)
             except IntegrityError:
-                self.db.expunge(budget)
+                pass  # this id's savepoint rolled back and expunged it; move on
         return created
+
+    @staticmethod
+    def _build_member_budget(member_id: uuid.UUID, default: WorkspaceBudgetDefault) -> ScopedBudget:
+        period_start, period_end = _period_window(default.budget_duration_sec)
+        return ScopedBudget(
+            scope_type=_SCOPE_WORKSPACE_MEMBER,
+            scope_id=str(member_id),
+            provider_key_id=default.provider_key_id,
+            max_budget=default.max_budget,
+            budget_duration_sec=default.budget_duration_sec,
+            period_start=period_start,
+            period_end=period_end,
+        )
 
     async def _existing_member_budget(self, member_id: uuid.UUID, provider_key_id: str | None) -> ScopedBudget | None:
         stmt = select(ScopedBudget).where(
@@ -285,21 +312,6 @@ class WorkspaceBudgetDefaultService:
             else ScopedBudget.provider_key_id.is_(None)
         )
         return (await self.db.execute(stmt)).scalars().first()
-
-    async def _create_member_budget(self, member_id: uuid.UUID, default: WorkspaceBudgetDefault) -> ScopedBudget:
-        period_start, period_end = _period_window(default.budget_duration_sec)
-        budget = ScopedBudget(
-            scope_type=_SCOPE_WORKSPACE_MEMBER,
-            scope_id=str(member_id),
-            provider_key_id=default.provider_key_id,
-            max_budget=default.max_budget,
-            budget_duration_sec=default.budget_duration_sec,
-            period_start=period_start,
-            period_end=period_end,
-        )
-        self.db.add(budget)
-        await self.db.flush()
-        return budget
 
     # ------------------------------------------------------------------
     # CRUD (commit on success)
@@ -385,7 +397,7 @@ class WorkspaceBudgetDefaultService:
         # write `WorkspaceBudgetDefaultAlreadyExistsError`'s message describes
         # (the template's own unique index). `materialize_for_default` handles
         # its own, unrelated `IntegrityError`s internally (see
-        # `_create_member_budgets`) precisely so one never reaches here and
+        # `_insert_member_budgets`) precisely so one never reaches here and
         # gets misreported as "this default already exists", rolling back a
         # template that in fact does not conflict with anything.
         try:
--- a/src/gateway/services/tenancy/workspace_budget_default_service.py
+++ b/src/gateway/services/tenancy/workspace_budget_default_service.py
@@ -1,0 +1,465 @@
+"""Workspace-level templates for per-member ``scoped_budgets`` ceilings.
+
+A default (``WorkspaceBudgetDefault``) is a workspace-level template for a
+per-member spend limit. When a member joins the workspace (or when a default
+is created on a workspace that already has members) it is **eagerly
+materialized** into a per-member :class:`ScopedBudget` row, so the ceiling is
+visible immediately rather than on first spend.
+
+The wire DTOs keep the ``WorkspaceMemberBudgetPolicy*`` names from
+``otari-ai``'s ``budget_policy_service`` (the hosted equivalent this is ported
+from), so the generated client stays recognizable across both trees; the
+stored fields follow this repo's own ``ScopedBudget`` vocabulary
+(``max_budget``, ``budget_duration_sec``) rather than otari-ai's
+(``budget_limit``, ``spend_period``), since OSS budgets are USD/seconds-only
+with no token or request dimension.
+
+Materialization methods (``materialize_for_member``, ``materialize_for_default``)
+are flush-only: they do not commit, so they fold into the enclosing
+transaction at each call site (workspace creation, adding a member,
+organization-member creation with workspace assignments). CRUD methods commit
+on success.
+"""
+
+from __future__ import annotations
+
+import uuid
+from datetime import UTC, datetime, timedelta
+
+from pydantic import BaseModel, Field
+from sqlalchemy import func, select
+from sqlalchemy.exc import IntegrityError
+from sqlalchemy.ext.asyncio import AsyncSession
+from sqlmodel import col
+
+from gateway.models.entities import ScopedBudget, WorkspaceBudgetDefault
+from gateway.models.tenancy import User, Workspace, WorkspaceMember
+from gateway.repositories.tenancy import WorkspaceMemberRepository, WorkspaceRepository
+from gateway.services.tenancy import authorization
+from gateway.services.tenancy.errors import (
+    WorkspaceBudgetDefaultAlreadyExistsError,
+    WorkspaceBudgetDefaultNotFoundError,
+)
+from gateway.services.tenancy.organization_service import OrganizationService
+
+# The scope name is spelled out rather than imported from
+# `scoped_budget_service`: that module imports `workspace_scope`, which imports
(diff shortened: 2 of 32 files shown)


khaledosman (commented):
The materialisation semantics are the right ones and the reasoning for each is stated where it belongs: non-retroactive edit, delete-preserves-rows, member override beating the template. Extracting `authorization.py` out of `WorkspaceService` rather than growing a second copy of the visibility and management rules is the correct call, and the extraction reads as faithful. Router placement is right too — it sits after `register_routers`' hybrid `return`, and `hybrid_mode.py`'s `/v1/workspaces/{path:path}` catch-all already gives it the friendly 404.

Six inline findings; the N+1 and the over-wide `except IntegrityError` are the two I would fix before merge.

Three without a line to hang on:

- **No screenshot entry** for the new `/budget-defaults` page. It owes a line in `web/e2e/screenshots/authenticated.spec.ts` — nothing fails without it today, which is why it gets forgotten.
- **Test coverage is thin on the panel.** `WorkspaceBudgetDefaultsPage.test.tsx` has two read-only cases against a 363-line component; create, edit, delete and the `parseLimit` / `parseDays` / `daysString` helpers are untested, and those helpers are pure and cheap to test directly. The eight integration tests on the service side are good.
- **Alembic chain:** five open PRs all set `down_revision = "a3c7e1b9d5f2"` (this one `f7a2c4e6b8d1`, plus #668, #670, #667, #640). The second to merge forks the chain and `alembic upgrade head` refuses to choose, which startup `auto_migrate` and `tests/integration/conftest.py` both run. Needs a merge order with the rest rebased.

One thing I checked and would leave alone: `defaults.isLoading` as the loading guard contradicts the written standard (`isPending && !data`), but `WorkspaceMembersPanel.tsx:156` does exactly the same thing, so this matches its sibling. Worth fixing as its own sweep, not here.

---

src/gateway/services/tenancy/workspace_budget_default_service.py
```diff
@@ -0,0 +1,334 @@
...
+        while True:
+            members, total = await self.workspace_members.get_by_workspace(
+                default.workspace_id, skip=skip, limit=_MATERIALIZE_PAGE_SIZE
+            )
+            if not members:
+                break
+            for member in members:
+                if member.status != "active":
+                    continue
+                if await self._existing_member_budget(member.id, default.provider_key_id) is not None:
```
khaledosman: N+1 inside the page loop: one `_existing_member_budget` SELECT plus one INSERT-and-flush per member, so a 500-member page is ~1000 round trips in one request. The performance instructions call out `await db.execute(select(...))` in a `for` loop specifically.

Both halves batch cleanly. Fetch the existing scope ids once per page:

```python
member_ids = [str(m.id) for m in members if m.status == "active"]
stmt = select(col(ScopedBudget.scope_id)).where(
    ScopedBudget.scope_type == _SCOPE_WORKSPACE_MEMBER,
    col(ScopedBudget.scope_id).in_(member_ids),
    ScopedBudget.provider_key_id == default.provider_key_id
    if default.provider_key_id is not None
    else ScopedBudget.provider_key_id.is_(None),
)
already = set((await self.db.execute(stmt)).scalars().all())
```

then build the rows and `self.db.add_all(...)` with a single flush. `materialize_for_member` has the same shape but is bounded by the number of defaults on one workspace, so it matters much less.

src/gateway/services/tenancy/workspace_budget_default_service.py
```diff
@@ -0,0 +1,334 @@
...
+            name=request.name,
+            max_budget=request.max_budget,
+            budget_duration_sec=request.budget_duration_sec,
+        )
+        self.db.add(default)
+        try:
+            await self.db.flush()
+            await self.materialize_for_default(default)
+            await self.db.commit()
+        except IntegrityError:
```
khaledosman, line 405: The `try` spans `materialize_for_default`, so any `IntegrityError` raised while fanning out is reported as `WorkspaceBudgetDefaultAlreadyExistsError` — a 409 reading "Workspace X already has a budget default for Y" — and the default itself is rolled back.

That is reachable, not theoretical: `_existing_member_budget` is check-then-act with no lock, so a concurrent `WorkspaceService.add_member` (or a second `create_default`) materialising the same `(workspace_member, member_id, provider_key_id)` triple trips `scoped_budgets`' partial unique index. The operator is then told a default exists when none does.

Narrow the `try` to the default row's own flush, which is the only write the mapped error describes, and handle the materialisation conflict on its own terms (skipping a row another writer just created is the benign outcome).

src/gateway/services/tenancy/workspace_budget_default_service.py
```diff
@@ -0,0 +1,334 @@
...
+
+    async def _create_member_budget(self, member_id: uuid.UUID, default: WorkspaceBudgetDefault) -> ScopedBudget:
+        now = datetime.now(UTC)
+        budget = ScopedBudget(
+            scope_type=_SCOPE_WORKSPACE_MEMBER,
+            scope_id=str(member_id),
+            provider_key_id=default.provider_key_id,
+            max_budget=default.max_budget,
+            budget_duration_sec=default.budget_duration_sec,
+            period_start=now if default.budget_duration_sec else None,
```
khaledosman: This duplicates the period-window arithmetic #640 is centralising: that PR adds `period_window()` to `scoped_budget_service` precisely so "creation, retiming, and the roll cannot drift apart", and this becomes a fourth derivation. It also means `WorkspaceBudgetDefault` has no `reset_alignment`, so once #640 lands a workspace default can only ever materialise a rolling ceiling and a monthly per-member cap cannot be templated at all.

The cycle note above rules out importing `scoped_budget_service` from here, so closing this probably means `period_window` moving somewhere neutral (`core/`). Worth agreeing with #640 which PR owns it rather than both landing as-is.

src/gateway/services/tenancy/workspace_budget_default_service.py
```diff
@@ -0,0 +1,334 @@
...
+from gateway.services.tenancy.errors import (
+    WorkspaceBudgetDefaultAlreadyExistsError,
+    WorkspaceBudgetDefaultNotFoundError,
+)
+from gateway.services.tenancy.organization_service import OrganizationService
+
+# The scope name is spelled out rather than imported from
+# `scoped_budget_service`: that module imports `workspace_scope`, which imports
+# `tenancy.provisioning_service`, which imports `tenancy/__init__`, which
+# imports `workspace_service`, which imports this module — see
```
khaledosman: Em dash in a code comment; `AGENTS.md` rules those out in prose including doc comments. A comma or a semicolon does the same work here.

Same paragraph, wider point: this PR carries **~26 British spellings** (`materialis*`, `recognis*`) across this module, the route docstrings, the migration, `models/entities.py`, `organization_service.py`, and the dashboard. The repo is US English, and this is not only a comment style question: the route docstrings become endpoint descriptions, so `materialises` ships into `docs/public/openapi.json`, the Postman collection and `web/src/client/schema.ts`. `materialize`/`materialized` is already the spelling the method names use (`materialize_for_member`), so the prose is inconsistent with its own identifiers.

web/src/shared/api/hooks.ts
```diff
@@ -1626,6 +1629,88 @@ export function useUpdateWorkspaceMemberRole() {
...
 }
 
+// A workspace's budget-default templates. Nested under the workspaces key so
+// deleting a workspace drops them with it, same as `useWorkspaceMembers`.
+export function useWorkspaceBudgetDefaults(workspaceId: string | null) {
+  return useQuery({
+    queryKey: [WORKSPACES, workspaceId, "budget-defaults"],
+    queryFn: () =>
+      fetchAllPaged<WorkspaceBudgetDefault>(
+        `/v1/workspaces/${encodeURIComponent(workspaceId as string)}/member-budget-policies`,
```
khaledosman, line 1716: `fetchAllPaged` appends `?skip=&limit=`, but `list_workspace_budget_defaults` declares no pagination parameters — FastAPI ignores unknown query params, so the route returns the full list every time. It happens to work because the two partial unique indexes bound a workspace to one aggregate default plus one per provider instance, so `data.length < TENANCY_PAGE_SIZE` breaks the loop on the first pass.

At or past the page size it would refetch the same full list up to `TENANCY_MAX_PAGES` times and duplicate every row. Every other `fetchAllPaged` target genuinely paginates; either give this route `skip`/`limit` like the sibling tenancy lists, or use a plain `apiFetch<Paged<...>>` and read `data`.

Minor, same line: `workspaceId as string` inside a hook already gated by `enabled: workspaceId !== null` — the cast is avoidable by taking the id in a narrowed local.

web/src/features/workspaces/WorkspaceBudgetDefaultsPanel.tsx
```diff
@@ -0,0 +1,363 @@
...
+                </span>
+                <Chip size="sm" color="default">
+                  {providerLabel(row.provider_key_id)}
+                </Chip>
+                <span className="text-sm text-muted">
+                  {row.max_budget === null
+                    ? "No limit"
+                    : usd.format(row.max_budget)}
+                  {row.budget_duration_sec
+                    ? ` / ${row.budget_duration_sec / DAY} days`
```
khaledosman: Unguarded division: `budget_duration_sec` is any integer `> 0` on the wire, so a default created through the API with `3600` renders `/ 0.041666666666666664 days`. `daysString` above already guards this for the form (`seconds % DAY === 0`); the read view does not.

Either reuse that guard and fall back to hours/seconds, or reject non-day multiples at the API so the display assumption is actually true.

tbille:
Addressed all 8 findings from Copilot's review in 60fb4eb:

1. **Race** (create_default vs. a concurrent member join): closed with a new `WorkspaceRepository.lock`, taken by `create_default` and by every membership-creation path before its own read. Added a two-session test (`test_concurrent_default_create_and_member_add_both_land`) that I confirmed fails reliably without the lock and passes reliably with it.
2. **Suspended-membership revive doesn't materialize**: fixed in `_apply_workspace_assignments`; added `test_reviving_a_suspended_workspace_membership_is_materialized`.
3. **N+1 in `materialize_for_default`**: batched into one existence query + one insert per page instead of two round trips per member.
4. **Unbounded list endpoint**: `list_defaults`/the GET route now take `skip`/`limit`, matching every other list endpoint here.
5. **Duration round-trip bug** (a non-whole-day `budget_duration_sec` silently nulled on an unrelated save): the days field now preserves an untouched value it can't display, with a test that fails without the fix.
6. **Frontend gating narrower than the server's rule**: added `canManageWorkspace` (org role OR the caller's own role in the selected workspace, which `useSelectedWorkspace` already carries for free) and applied it to both the new page and the same pre-existing gap on the Members page.
7. **Missing e2e screenshot coverage**: added `/budget-defaults` to `WORKSPACE_ROUTES`.
8. **Unbounded `budget_duration_sec`**: capped at ~10 years on create/update, so an oversized value 422s instead of overflowing `timedelta` into a 500.

Also caught in the process (not from the review, but surfaced once the spec was regenerated after these fixes): the four new endpoints were missing from `scripts/sdk_codegen/sdk-endpoints.txt`, which `test_sdk_endpoint_coverage.py` enforces — added under `[excluded]` alongside the sibling workspace-member routes.

All backend and frontend suites, `make lint`, `openapi-check`/`postman-check`, and the OSS-edition smoke test pass.

tbille:
Addressed the human review findings (@khaledosman) and CodeRabbit's, in f96325c:

1. **N+1 in `materialize_for_default`**: already batched in the prior commit; confirmed it matches your proposed diff.
2. **`IntegrityError` misattribution**: real gap. The outer `try` in `create_default` spanned `materialize_for_default`, so a per-member insert collision (racing a direct `POST /v1/scoped-budgets` for one of the members, the one surface the new workspace lock doesn't and can't cover) rolled back the default's own row and reported a false "already exists" 409. Narrowed the `try` to the default row's own flush, and made the batched insert itself resilient: a savepoint around the batch, falling back to one-row-at-a-time (each in its own savepoint) on collision, skipping only the row that actually conflicts. I didn't add a dedicated test for this exact interleaving: it needs two genuinely concurrent connections racing a narrow window with no lock to make deterministic (unlike the earlier race, where the lock itself makes the outcome deterministic regardless of scheduling), so I verified it by reasoning + the existing conflict test (`test_duplicate_aggregate_default_conflicts`, still exercising the narrowed `try` correctly) rather than a flaky interleaving test.
3. **Period-window duplication / #640**: extracted the two in-file derivations into one `_period_window` helper. Left the cross-PR reconciliation with `scoped_budget_service`'s own `period_window()` as a follow-up since I can't see #640's design; happy to coordinate on which PR owns it.
4. **Em dash + British spellings**: swept all of it back to house style, US English and no em-dash separators, including the route docstrings that ship into the OpenAPI spec, Postman collection, and generated client.
5. **Unguarded division in the read view**: same bug class as the edit-form fix, just in `WorkspaceBudgetDefaultsPanel`'s row display. Added `formatDuration` with an hours/seconds fallback.
6. **`fetchAllPaged` cast**: left as-is, it's the same idiom `useWorkspaceMembers` already uses one line above it, not something introduced here.

From CodeRabbit: fixed the `make typecheck` failures (six missing `assert ... is not None` guards in the new integration test). The Postman example showing `budget_duration_sec: 0` is a pre-existing gap in the generator's example-value heuristic, already present for `/v1/budgets` and `/v1/scoped-budgets`, not something this PR introduced; fixing the shared generator felt out of scope for this change, so I left it.

Also updated the PR description with the template's required PR Type, Checklist, and AI Usage sections.

All suites, `make lint`, `make typecheck`, OpenAPI/Postman checks, and the OSS-edition smoke test pass.

src/gateway/services/tenancy/workspace_budget_default_service.py
```diff
@@ -0,0 +1,465 @@
...
+                scope_id=str(member_id),
+                provider_key_id=default.provider_key_id,
+                max_budget=default.max_budget,
+                budget_duration_sec=default.budget_duration_sec,
+                period_start=period_start,
+                period_end=period_end,
+            )
+
+        budgets = [build(member_id) for member_id in candidates]
+        self.db.add_all(budgets)
```
njbrake: This fallback cannot run, and the failure is worse than the case it was written for.

`add_all` happens before `begin_nested()`, so the objects are pending when the savepoint opens. When the flush fails, SQLAlchemy rolls back to the savepoint boundary and leaves the session's transaction deactivated, so the first `begin_nested()` in the loop below raises `PendingRollbackError`, which is not an `IntegrityError` and escapes both handlers. A single colliding member therefore 500s the whole `create_default` instead of costing one skipped row.

Reproduced against PostgreSQL 18 / asyncpg with this exact shape (a pre-existing `workspace_member` row that the batch's existence check missed, plus a prior successful flush in the same transaction, as `create_default` has):

```
batch flush raised IntegrityError, as designed
FAILED PendingRollbackError: This Session's transaction has been rolled back
due to a previous exception during flush.
```

Moving `add_all` and each per-row `add` inside their own `begin_nested()` block, and dropping the `expunge` below, gives the intended behavior on the same fixture:

```
batch IntegrityError -> per-row fallback
created=2 (expect 2), rows=4 (expect 4)
```

Worth a test that pre-inserts a colliding `scoped_budgets` row for one member of the page, since that is what would have caught this.

src/gateway/services/tenancy/workspace_budget_default_service.py
```diff
@@ -0,0 +1,465 @@
...
+        created: list[ScopedBudget] = []
+        for member_id in candidates:
+            budget = build(member_id)
+            self.db.add(budget)
+            try:
+                async with self.db.begin_nested():
+                    await self.db.flush()
+                created.append(budget)
+            except IntegrityError:
+                self.db.expunge(budget)
```
njbrake: Even with the savepoint ordering fixed, this line raises. A savepoint rollback already expunges the object it was about to insert, so `expunge` gets an instance that is no longer in the session:

```
InvalidRequestError: Instance <ScopedBudget at 0x...> is not present in this Session
```

Verified on PostgreSQL. Deleting the line is enough once the `add` moves inside the `begin_nested()` block.

src/gateway/services/tenancy/workspace_budget_default_service.py
```diff
@@ -0,0 +1,465 @@
...
+    """Request body for creating a default."""
+
+    provider_key_id: str | None = Field(
+        default=None,
+        max_length=255,
+        description="Narrow the default to one provider instance; null applies to every provider",
+    )
+    name: str | None = Field(default=None, max_length=200, description="Admin-facing label")
+    max_budget: float | None = Field(default=None, ge=0, description="Maximum USD spend per member in the period")
+    budget_duration_sec: int | None = Field(
```
njbrake, line 94: A template cannot express `reset_alignment`, which #640 added to `ScopedBudget` and which is now on main. So a workspace that wants calendar-month per-member ceilings still has to create every row by hand, which is the case this feature exists to remove.

Fine to defer, but worth an issue rather than silence: as it stands the template is a strict subset of what the row it materializes can hold, and the CHECK constraint on `scoped_budgets` means adding it later is an either/or field, not an extra one.

src/gateway/services/tenancy/workspace_service.py
```diff
@@ -401,10 +381,18 @@ async def add_member(
...
+        # read of the current defaults and that one's read of the current
+        # members can each run before the other's write commits, and the new
+        # member gets neither.
+        await self.workspaces.lock(workspace.id)
+
         # As in create_workspace: the pre-check races the insert, and the unique
         # constraint is what actually decides.
         try:
             member = await self.members.create(workspace_id=workspace.id, user_id=user_id, role=role)
+            await self.budget_defaults.materialize_for_member(member)
```
njbrake, line 395: The misattribution you narrowed in `create_default` is still here.

`materialize_for_member` reaches `_create_member_budget`, which does an unguarded `add` plus `flush` with no savepoint. A concurrent direct `POST /v1/scoped-budgets` for this member (the one surface the workspace lock cannot cover, by your own analysis in `_create_member_budgets`) raises `IntegrityError` inside this `try`, which rolls the membership back and answers `WorkspaceMemberAlreadyExistsError` for a member who does not exist.

`_apply_workspace_assignments` has the same shape one level up, where its caller's `except IntegrityError` turns it into `OrganizationMemberAlreadyExistsError`. Either give the per-member insert the same savepoint the batch path is meant to have, or move materialization out of the membership-creation `try` so an unrelated collision is not reported as a duplicate.

src/gateway/services/tenancy/organization_service.py
```diff
@@ -457,12 +476,18 @@ async def _apply_workspace_assignments(
...
+        for workspace_id, role in sorted(wanted.items(), key=lambda item: str(item[0])):
+            # Serialized against a concurrent `WorkspaceBudgetDefaultService.create_default`
+            # on this workspace; see `WorkspaceService.add_member`'s identical lock
+            # for why.
+            await workspaces.lock(workspace_id)
             existing = existing_by_workspace.get(workspace_id)
             if existing is not None:
-                await members.update(existing, WorkspaceMemberUpdate(role=role, status="active"))
+                revived = await members.update(existing, WorkspaceMemberUpdate(role=role, status="active"))
+                await budget_defaults.materialize_for_member(revived)
```
njbrake: This materializes for every existing membership, not only a revived one: `existing is not None` is also true for an already-active member whose role is being re-applied unchanged, and the local name `revived` reads as though it were not.

The visible consequence is that re-posting the same `workspace_assignments` re-creates a per-member ceiling an admin deliberately deleted through `/v1/scoped-budgets`. If revive is the intent, gate on `existing.status != "active"`. If re-application is meant to be a re-materialization point, the docstring should say that, because it is a different promise from the one above it.

njbrake (commented):
Reviewed the full diff and ran it locally against PostgreSQL 18: the new integration file plus the scoped-budget and tenancy regressions (129 tests, green), the migration up, down and up again, and `alembic check` clean for this table. CI is green too.

**Alembic is fine after the rebase.** `f7a2c4e6b8d1` descends from `c8f2a6b4e9d3`, `alembic heads` reports a single head, and `tests/unit/test_migration_single_head.py` passes on this branch. The DDL matches the model on PostgreSQL: `timestamptz` both sides, both partial unique indexes, FK cascade.

**One real defect:** the savepoint fallback in `_create_member_budgets` cannot run as written, and 500s instead of skipping one row. Inline, with the reproduction. CodeRabbit's comment at line 274 is right; the fix needs one more piece than it names.

**The narrowed try/except is narrowed in only one of the three call sites.** `add_member` and `_apply_workspace_assignments` still wrap `materialize_for_member` in an `except IntegrityError` that reports a duplicate membership, so the misattribution you fixed in `create_default` is still live in both. Inline.

**Test gap:** nothing reaches the fallback path, which is why it shipped broken. `test_duplicate_aggregate_default_conflicts` covers the template's own unique index, not a per-member collision.

Two things I would take as follow-ups rather than in this PR:

- `remove_member` deletes the `WorkspaceMember` row and leaves its materialized `ScopedBudget` behind. That gap predates this PR, but defaults turn an occasional orphan into one per departing member, and `_delete_scoped_budgets_for` is already the codebase's answer for a ceiling that outlives its scope.
- `ensure_bootstrap_identity` in `provisioning_service.py` is a fourth `WorkspaceMember`-creating path and is not materialized, so `create_workspace`'s new comment ("every WorkspaceMember-creating path materializes the same way") is not quite true. First boot only, so small.

**Bots:** I agree with CodeRabbit on line 274 and on its singular-label nit. On the Postman `budget_duration_sec: 0` finding, your reply is correct: main already carries three such examples, so it is a pre-existing generator gap, not a regression here.

Two smaller notes: `web/src/app/nav/registry.test.ts` now conflicts with #638 on main, and two British spellings survived the sweep (`materialises` in `web/src/app/nav/registry.ts`, `recognisable` in `web/src/client/index.ts`).

_Note: this review was drafted by Claude Opus 5 in a session directed by @njbrake. The analysis and the prose are Claude's; what to do with the PR is his call._

tbille:
Rebased onto main (invitations #668, org model pricing #674, and the nav shell redesign #638 had all landed) and addressed @njbrake's review, in `fcdbb32`.

**The real defect, reproduced and confirmed fixed:** `_insert_member_budgets`'s savepoint fallback couldn't run as written. `AsyncSession.begin_nested()` unconditionally flushes pending state *before* opening the `SAVEPOINT`, so staging `add`/`add_all` before entering the block flushed those rows outside any savepoint; a collision there failed the outer transaction, and the next `begin_nested()` in the per-row fallback raised `PendingRollbackError` (not `IntegrityError`), 500ing the request instead of skipping one row. I reproduced this exactly (same `PendingRollbackError`) by temporarily restoring the old ordering, then confirmed the fix by moving every `add`/`add_all` inside its own `begin_nested()` block and dropping the manual `expunge()` (a savepoint rollback already expunges the row it was protecting, so the second call raised `InvalidRequestError` finding it already gone). Added `test_materialize_batch_recovers_from_a_missed_collision`, which calls the insert helper directly with a pre-existing colliding row so the collision is deterministic rather than timing-dependent; verified it fails on the old ordering and passes on the new one.

**The misattribution was live in the other two call sites**, exactly as flagged: fixed at the root rather than per call site. `materialize_for_member`/`materialize_for_default` now both route through the same savepoint-safe helper, so neither can raise `IntegrityError` to `add_member` or `_apply_workspace_assignments` at all anymore.

**The revive-branch over-materialization**: fixed by gating on `existing.status != "active"`, with `test_reapplying_an_active_assignment_does_not_rematerialize_a_deleted_override` (confirmed it fails without the gate, reproducing the "admin deletes an override, a re-post resurrects it" scenario exactly).

**`_period_window`'s zero-duration truthiness check**: changed to `is None`, per CodeRabbit.

**Frontend `formatDuration`**: singularized "1 day"/"1 hour".

**Filed #680** for the three deferred items (reset_alignment on the template, `remove_member` orphaning ceilings, `ensure_bootstrap_identity` not materializing) rather than leaving them unacknowledged.

**Ran the full manual smoke test** against a live server (`make dashboard`, `uv run otari serve`) beyond the automated suites: created a workspace default and confirmed it materialized onto the existing member; added a new member and confirmed join-time materialization; edited the default and confirmed existing rows were untouched (non-retroactive); deleted it and confirmed materialized rows survived; confirmed auth (401 unauthenticated/wrong key), the duplicate-default 409, and the oversized-duration 422 all behave correctly; confirmed the dashboard bundle serves, the budget-defaults chunk is reachable, and the session sign-in flow works.

All CI checks are green.

tbille:
Addressed CodeRabbit's latest review:

1. **Dispose the `sessions` fixture's engine** (real, fixed): the fixture now disposes its engine in a `finally`, converted to an async generator (`@pytest_asyncio.fixture`, no `asyncio_mode` is set in `pytest.ini` so strict mode applies). Verified the file's tests, including the race test that uses it, still pass.

2. **`created_at`/`updated_at` should use `UtcDateTime()`**: declined. `WorkspaceBudgetDefault`'s sibling and materialization target, `ScopedBudget`, uses plain `DateTime(timezone=True)` for the identical fields (and for `period_start`/`period_end` too), so matching it is the consistent choice, not a gap. This repo's own precedent for when `UtcDateTime` is actually needed is `OrganizationModelPricing`'s docstring: it's for a timestamp round-tripped through an *editable* control, where a naive-on-SQLite value would corrupt on save. Neither `WorkspaceBudgetDefault`'s nor `ScopedBudget`'s `created_at`/`updated_at` are rendered anywhere in the dashboard today (confirmed: no reference to either field in `WorkspaceBudgetDefaultsPanel.tsx`), so there's no render-then-corrupt path this closes. Fixing it on the new table alone while leaving `ScopedBudget` as is would be inconsistent without removing the underlying exposure.

3. **The repeated `materialize_for_default`/`commit` `try`/`except SQLAlchemyError` suggestion**: still declined, same reasoning as before. To be explicit about why this isn't an oversight: `update_default` and `delete_default` two methods below it in this exact file, and every write in `workspace_service.py`/`organization_service.py`, commit bare with no `SQLAlchemyError` wrapping either. The tenancy service layer's convention is to let an unexpected database error propagate to `get_db`'s teardown rather than catch-rollback-reraise at every commit; the AGENTS.md rule this cites describes the older route-embedded pattern (`routes/budgets.py`, `routes/scoped_budgets.py`), which has no service layer between the route and the session. Applying the suggestion to only this one call site would make it the one write in the tenancy services that's inconsistent with its own neighbors.

tbille:
Addressed the remaining CodeRabbit finding, in `d6dea4f`:

**`workspace_budget_defaults.created_at`/`updated_at` should use `UtcDateTime()`**: real, fixed. Unlike the earlier general style question on this same pattern, this one is concrete: `WorkspaceMemberBudgetPolicyPublic.from_model` serializes both columns with `.isoformat()` for the budget-defaults dashboard page, and on SQLite (this repo's default `database_url`) `DateTime(timezone=True)` round-trips naive, so the wire value would carry no offset and a browser would read it as local time. `UtcDateTime.impl` is `DateTime(timezone=True)`, so the migration's DDL is unchanged.

Verified: `make lint`, `make typecheck`, the targeted integration suites (`test_workspace_member_budget_policies.py`, `test_scoped_budgets.py`) against a real PostgreSQL container, and confirmed `make openapi-check`/`make postman-check` show no drift (no wire-shape change).

All 27 review threads are now resolved and CI is green.

njbrake (approved):
Approving to unblock and let you get this over the finish line! 🚀

PR #675, opened 2026-08-20
feat(platform): return settled cost inline

### LINKED ISSUE #672: Return the settled cost inline on platform-mode responses
## Desired behavior

In platform mode, a successful response carries the authoritative cost, in every
response shape:

| Shape | Non-streaming | Streaming |
|---|---|---|
| `/v1/messages` | `Message.usage` | `message_delta.usage` |
| `/v1/chat/completions` | `usage` | final usage chunk |
| `/v1/responses` | `response.usage` | final response object |

Standalone mode is unchanged.

## Why

Cost is unknowable to the caller at response time. The usage report is
fire-and-forget after the response (`_schedule_usage_report`,
`src/gateway/api/routes/_pipeline.py:2317`), so a caller enforcing budgets is
always a poll interval behind its own spend.

## Notes

- Blocked by mozilla-ai/otari-ai#1762 — the platform must return the cost first.
  The gateway must not price locally: org-scoped and managed-model pricing live
  on the platform, so a second computation would produce a different number.
- Implement as the inverse of the existing per-shape adapter protocol
  (`_pipeline.py:455-470`, `extract_usage` / `extract_stream_usage`). Adding
  `attach_cost` there forces every current and future shape to implement it, so
  a messages-only version is not expressible.
- On the success path the report becomes awaited before the final chunk instead
  of scheduled. It runs after generation, so it delays only the terminal event.
  Every non-success path keeps `_schedule_usage_report`.
- **Field naming needs a decision.** `cost_usd` matches the platform's existing
  `/request-costs` contract (`cost_usd` + `currency: "USD"`); `cost` matches
  OpenRouter and existing ecosystem tooling. One-way door.
- **Chat streaming gap.** `chat.py:181` injects `include_usage: True` only when
  the client sent no `stream_options`, so a client explicitly sending
  `{"include_usage": false}` gets no usage object to carry cost. Proposal: emit
  the usage chunk regardless in platform mode, documented in §"Streaming" as a
  deliberate deviation — the client opted out of token counts, not out of being
  billed. Verified safe for any-llm and the Python SDK; residual risk is
  third-party app code indexing `chunk.choices[0]`.
- Streaming has a hard limit: per §"Streaming", errors after the first chunk
  flushes are out of reach, so a mid-content stream death carries no inline cost
  and still needs the `/request-costs` fallback.
- Chat, messages, responses only. Embeddings, images, rerank, moderations and
  audio have no platform-mode path (no resolve, no `correlation_id`) and the
  platform prices tokens only. Separate issue if wanted.

HareeshBahuleyan:
## Why

Hybrid clients need the platform's authoritative settled cost with the model response, without duplicating pricing logic in the gateway or requiring a second lookup.

## What changed

Await the successful usage report under a configurable inline budget and attach its exact cost and pricing source to existing usage objects across Chat Completions, Messages, and Responses. Streaming delays only the terminal usage suffix; unpriced, unavailable, legacy, and timed-out settlements keep the original response shape, while failed-attempt reporting and standalone behavior remain unchanged.

## Notes

- **Deployment dependency.** Blocked on deployment of mozilla-ai/otari-ai#1762.
- **Platform follow-up.** Add coverage for a success report arriving before an earlier attempt's failure report.
- **Platform follow-up.** Return `202` when result building fails after accounting commits, avoiding a retry that collides with settlement and returns `409`.
- **M4 ledger.** No entry is needed. The existing Pricing and request cost and Usage and observability rows cover this contract, and their targets are unchanged.

Fixes #672

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes #672

## Checklist

- [x] **M4 ledger:** this PR either needs no entry (an existing row already covers it and its target is unchanged), or the entry is appended to [otari-ai#1587](https://github.com/mozilla-ai/otari-ai/issues/1587). One entry per *change*, not per PR: a stack delivering one surface change gets one. See [.github/instructions/reconciliation-ledger.instructions.md](instructions/reconciliation-ledger.instructions.md).

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:** Pi coding agent (`openai-codex/gpt-5.6-sol`)

**Any additional AI details you'd like to share:** Used to investigate CI and restore the required PR template sections.

## Summary

- Added inline settled-cost reporting for platform-mode Chat Completions, Messages, and Responses.
- Added configurable timeout handling for successful usage settlement.
- Added cost and pricing details to response and terminal streaming usage objects.
- Preserved existing behavior for failed, unpriced, unavailable, timed-out, legacy, and standalone requests.
- Added documentation, metrics, configuration validation, and comprehensive tests.

## Technical notes

- Streaming responses delay only the terminal usage event.
- Added `PLATFORM_USAGE_INLINE_TIMEOUT_MS` with a default of 1500 ms.
- Preserved `/request-costs` fallback behavior when inline settlement is unavailable.

### DIFF
--- a/src/gateway/api/routes/_platform.py
+++ b/src/gateway/api/routes/_platform.py
@@ -40,12 +40,13 @@ from gateway.services.web_search_backend import WebSearchNotReachableError
 T = TypeVar("T")
 
 # Status codes returned by the platform's usage-report endpoint that the
-# gateway should NOT retry. Auth / payment-required / not-found / conflict /
-# unprocessable are all permanent rejection signals — retrying would just
+# gateway should NOT retry. Auth, payment-required, not-found, conflict, gone,
+# and unprocessable are all permanent rejection signals: retrying would just
 # hammer the platform (an overdrawn or missing wallet won't recover within the
-# retry window). 402 is already excluded by the >= 500 retry predicate below;
-# listing it keeps the intent explicit and robust to changes in that predicate.
-_USAGE_NON_RETRYABLE_STATUS_CODES = {401, 402, 404, 409, 422}
+# retry window). Several of these are already excluded by the >= 500 retry
+# predicate below; listing them keeps the intent explicit and robust to changes
+# in that predicate.
+_USAGE_NON_RETRYABLE_STATUS_CODES = {401, 402, 404, 409, 410, 422}
 
 # Statuses on which the billing-message probe runs. A 402 is payment required by
 # definition, so it is handled directly in ``is_provider_billing_error`` without
@@ -1063,7 +1064,7 @@ async def _report_platform_usage(
                     cost_usd=completed.cost_usd,
                     pricing_source=completed.pricing.source,
                 )
-            if response.status_code in _USAGE_NON_RETRYABLE_STATUS_CODES or response.status_code == 410:
+            if response.status_code in _USAGE_NON_RETRYABLE_STATUS_CODES:
                 return None
             should_retry = response.status_code >= 500
         except (httpx.TimeoutException, httpx.NetworkError):
--- a/src/gateway/api/routes/_platform.py
+++ b/src/gateway/api/routes/_platform.py
@@ -128,7 +128,7 @@ class _UsagePricing(BaseModel):
 
 
 class _CompletedUsageSettlement(BaseModel):
-    correlation_id: uuid.UUID
+    correlation_id: str
     status: Literal["completed"]
     outcome: Literal["success"]
     cost_usd: str = Field(pattern=r"^-?\d+\.\d{6}$")
@@ -1035,14 +1035,13 @@ async def _report_platform_usage(
                     return None
                 try:
                     completed = _CompletedUsageSettlement.model_validate(response.json())
-                    expected_correlation_id = uuid.UUID(correlation_id)
                 except (ValueError, ValidationError):
                     logger.warning(
                         "Platform usage report returned an invalid completed body correlation_id=%s",
                         correlation_id,
                     )
                     return None
-                if completed.correlation_id != expected_correlation_id:
+                if completed.correlation_id != correlation_id:
                     logger.warning(
                         "Platform usage report correlation mismatch expected=%s received=%s",
                         correlation_id,
--- a/src/gateway/api/routes/_platform.py
+++ b/src/gateway/api/routes/_platform.py
@@ -25,7 +25,7 @@ from any_llm.types.completion import CompletionUsage
 from fastapi import HTTPException, Request, status
 from openai import APIConnectionError as _OpenAIAPIConnectionError
 from openai import APITimeoutError as _OpenAIAPITimeoutError
-from pydantic import BaseModel
+from pydantic import BaseModel, Field, ValidationError
 
 from gateway.core.config import GatewayConfig
 from gateway.core.usage import cache_read_tokens_of, cache_write_tokens_of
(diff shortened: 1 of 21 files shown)


src/gateway/streaming.py
```diff
@@ -204,32 +226,110 @@ async def streaming_generator(
...
+                    continue
+
+                if buffering_terminal:
+                    if len(terminal_buffer) >= _MAX_TERMINAL_BUFFER_CHUNKS:
+                        # The designated carrier was not terminal after all. Stop
+                        # holding it so the stream keeps flowing, but stay open to
+                        # a later carrier rather than giving up on cost outright.
+                        if not overflow_logged:
+                            logger.warning(
+                                "Terminal usage carrier was not terminal for %s; streaming without holding it",
```
khaledosman, line 252: `WARNING` is the wrong level for a condition this code documents as expected. A hybrid chat tool loop forwards one `include_usage` chunk per iteration (`mcp_loop.py` `observe` returns `FORWARD` for a usage-only chunk), so any iteration that emits more than three chunks after its usage chunk trips this, which is close to every tool-loop stream. I hit it on the first probe with a two-round stream:

```
WARNING  Terminal usage carrier was not terminal for probe; streaming without holding it
```

Nothing is wrong at that point and the stream self-corrects onto the real terminal carrier, so an operator watching warnings gets noise proportional to tool-loop traffic. `logger.debug` (or `info`) fits what it now reports.

src/gateway/api/routes/_platform.py
```diff
@@ -116,6 +117,27 @@ class ResolvedAttempt(BaseModel):
...
 
+class SettledCost(BaseModel):
+    """An attachable platform settlement: a priced cost and the basis for it."""
+
+    cost_usd: str = Field(pattern=r"^-?\d+\.\d{6}$")
+    pricing_source: str
+
+
+class _UsagePricing(BaseModel):
+    source: str | None
```
khaledosman, line 128: `source` is required, so a completed body with `pricing: {}` or no `pricing` key fails `model_validate` and logs `WARNING "...returned an invalid completed body"` instead of taking the "settlement is not attachable" path a few lines down, which is where `docs/hybrid-mode-protocol.md` says an unpriced completed response should land ("Completed response without a price source... return the provider response without inline cost").

```python
source: str | None = None
```

puts a pricing-less body on the intended `debug` branch and keeps the warning for bodies that are actually malformed.

khaledosman (approved):
Verified on `cabbc87b`: `make lint` (architecture + ruff) clean, `mypy` clean on the three changed modules, `openapi-check` clean (no request/response model changed, so no generated artifact is stale), and the six affected unit files pass (149 passed, 1 skipped). Integration tests not run here: no Docker and no `TEST_DATABASE_URL`.

I reviewed the previous head first and had flagged the chat tool loop tripping the old abandon guard (round-1 `include_usage` chunk designated as carrier, four content chunks withheld, inline cost then permanently disabled). `8b96af07` resolves it: the last carrier wins, and I reproduced the new behavior against the real generator with a two-round tool-loop stream, where content resumes streaming, settlement lands on the terminal usage chunk, and `[DONE]` stays last. The six new tests in `test_pipeline_settlement.py` cover the paths that were previously untested.

Two nits inline. Also worth confirming platform-side before the dependent deploy: the success report is now awaited inline while failed-attempt reports stay in `BackgroundTasks`, so a success can now reach `/gateway/usage` before an earlier attempt's failure report, which reverses the old ordering.

---

src/gateway/api/routes/_platform.py
HareeshBahuleyan: Thanks. I checked the producer contract: `RequestCostCompletedSuccess.pricing` is required, and `RequestCostPricing.source` is required but nullable. The producer route test also asserts that `source` is present. An unpriced result is represented as `pricing.source: null`, which the current model already validates and routes to the debug branch. `pricing: {}` or an omitted `pricing` field is malformed, so the warning is intentional. Keeping the current validation.

Reviewed with Pi.

PR #699, opened 2026-08-21
fix(deps): cap mcp below 2.0.0, which drops streamablehttp_client

### LINKED ISSUE #689: mcp is floored with no ceiling, so any fresh resolution installs mcp 2.0.0 and breaks every gateway command on import
`pyproject.toml:24` floors `mcp>=1.28.1` with no upper bound. mcp 2.0.0 removed `streamablehttp_client` from `mcp.client.streamable_http`, and `src/gateway/services/mcp_client.py:22` imports it at module scope (used at line 82). Any fresh resolution therefore installs mcp 2.0.0 and every `gateway` command dies on import:

```
ImportError: cannot import name 'streamablehttp_client' from 'mcp.client.streamable_http'
```

Add a ceiling to the floor in `pyproject.toml`, or port the import to the 2.x API.

**This is invisible to CI precisely because the lock is committed.** `uv.lock` holds mcp 1.28.1, so otari never re-resolves and never sees the break. Anyone resolving fresh does. A regression test has to resolve without the lock, or CI keeps passing while the bug is live.

Blast radius is any fresh `uv sync` of otari today, not only the superset build that surfaced it. That is what makes it an otari bug rather than an integration detail of the consumer.

Found while wiring otari as a pinned submodule of `otari-ai` (mozilla-ai/otari-ai#1456): the superset resolves both dependency trees together, so otari's floor is re-resolved rather than read from this lock. That branch carries a local `mcp<2` constraint as a workaround pointing at this issue; it should come out once the ceiling lands here.

champ18ion:
## Description
`pyproject.toml` floored `mcp>=1.28.1` with no ceiling. mcp 2.0.0 removed `streamablehttp_client` from `mcp.client.streamable_http`, and `gateway/services/mcp_client.py` imports that at module scope, so any fresh resolution that picks up mcp 2.0 breaks every `gateway` command on import:

```
ImportError: cannot import name 'streamablehttp_client' from 'mcp.client.streamable_http'
```

`uv.lock` is committed and pinned to 1.28.1, so `uv sync` here never re-resolves and never sees this. It surfaces on any fresh, unlocked resolution instead: a fresh clone before the first `uv lock`, or a consuming superset resolving otari's tree alongside its own — which is how this was actually found (wiring otari as a pinned submodule of `otari-ai`).

Reproduced directly before writing the fix: a clean venv installing `"mcp>=1.28.1"` resolves to mcp 2.0.0 and hits exactly the ImportError above; the same venv with `"mcp>=1.28.1,<2.0.0"` resolves to mcp 1.29.0 and imports cleanly.

## Why the regression test is more than the one-line fix

A static check on the `pyproject.toml` constraint's *shape* (does it have an upper bound at all) would pass even if the ceiling were technically present but placed too high, or wrong in some other way that still lets the break through. `tests/unit/test_mcp_dependency_ceiling.py` does two things:

1. A fast static check that the declared constraint has an upper bound at all (clear failure message pointing at this issue if that regresses).
2. The real regression: resolves `mcp` fresh into a throwaway venv via `uv pip install --python <venv>`, **ignoring `uv.lock` entirely**, and asserts the actual import that broke (`streamablehttp_client`) still works against whatever the constraint currently allows. This is the only way to catch a fresh-resolution break, since the committed lock can't.

Also added `packaging` as an explicit dev dependency (was only present transitively before) since the test now relies on it directly to parse the `mcp` requirement out of `pyproject.toml`.

## PR Type

- [x] Bug Fix

## Relevant issues
Fixes #689

## Checklist

- [x] I ran the Definition of Done checks locally (`make lint`, `make typecheck`, `make test` — ran `ruff check` and `mypy` directly against the changed files; couldn't run the full suite through `uv sync` on this machine since `uv.lock` only supports `sys_platform == 'linux'`/`'darwin'`, but ran `uv lock` directly, which isn't platform-gated the same way, and it reproduced the exact same lockfile diff deterministically. Verified the actual fix and regression-test mechanism manually against real `uv pip install` runs before writing the test, described above).
- [ ] If the API contract changed, I regenerated the OpenAPI spec (`uv run python scripts/generate_openapi.py`). — N/A, no API contract change.

## AI Usage

**Any additional AI details you'd like to share:** I picked this up off your own issue since it was clean, uncontested, and high-value (breaks every gateway command on a fresh resolve). Had Claude Code implement it, and pushed back when the first pass toward the regression test leaned static-only — asked for the actual fresh-resolution proof the issue calls for instead, since a manifest-string check alone wouldn't have caught the real failure mode. Verified the bug reproduction and the fix independently (real `ImportError` without the ceiling, clean import with it) before accepting either.

## Summary

- Limited `mcp` to versions below 2.0.0.
- Added a static dependency check and an integration test for fresh dependency resolution.
- Verified that `streamablehttp_client` remains importable.
- Added `packaging` as a development dependency.
- Added timeouts for dependency installation and import checks.
- Documented the integration test’s network requirement and retry behavior.

## Benefit

Fresh installations avoid an incompatible `mcp` release. CI now detects dependency regressions without relying on `uv.lock`.


champ18ion:
Addressed the flaky-reruns suggestion (pushed), skipping the exact-specifier tightening on `_has_upper_bound()` intentionally.

That function is the fast, static half of the regression: a cheap early check for the specific shape of the original bug (a floor with no ceiling at all). The actual correctness proof is the behavioral test below it, which does a real, unlocked resolve and import check - that one already catches a ceiling that's present but placed wrong (e.g. `<3.0.0`), which the docstring calls out explicitly. Hardcoding the exact `<2.0.0` value in the static check would make it brittle for no real safety gain: a legitimate future ceiling bump (say to `<2.5.0`, if a compatible 2.x line ever ships) would fail this test for no reason, when the behavioral test is already the one actually proving safety.

champ18ion:
Good call, done. Split the file: the pure static check (_mcp_requirement, _has_upper_bound) stayed in tests/unit, the real fresh-install-and-import behavioral test moved to tests/integration/test_mcp_dependency_ceiling.py, keeping its assertions and flaky retry as-is. Both suites already run in the same CI job here so this doesn't change when it runs, just where it lives.

njbrake (approved):
Approving. The fix is right, and I checked the load-bearing parts rather than taking them on trust: I pulled the mcp wheels for 1.28.1, 1.29.0 and 2.0.0 and grepped them, and `streamablehttp_client` is genuinely gone in 2.0.0 (renamed `streamable_http_client`, with new `httpx2` and `mcp_types` deps), so the ceiling is the right call over a port. I also ran the new tests against a deliberately loosened `mcp>=1.28.1,<3.0.0`: the integration test fails with the real #689 ImportError while the static check passes, which is the split your docstrings describe, and it holds.

Two things I would like fixed before merge.

**1. `tests/integration/test_mcp_dependency_ceiling.py:60`, use `with_pip=False`.** `uv pip install` with the `--python` flag does the installing and pip inside that venv is never used, but bootstrapping it needs `ensurepip`, which distro interpreters ship separately (`python3.14-venv` on Debian). On one of those the test fails with an ensurepip error, after burning both reruns and both 5s delays, on something unrelated to the constraint it is meant to check. CI is fine today because setup-uv's CPython has ensurepip; a contributor's machine may not. With `with_pip=False` it passes there in 0.6s and still fails correctly on a loosened ceiling.

**2. Update the `AGENTS.md` Test Notes bullet.** It currently says "Two tests assert the provider-error sanitization by making a real outbound call", and that bullet is where the repo records which tests need egress. This adds a third, and unlike those two it fails hard after its retries rather than looking like a status mismatch. Worth noting there that it also skips when `uv` is not on PATH.

Non-blocking: `_mcp_requirement()` is duplicated verbatim across the two files, and the unit docstring re-tells the whole #689 story the integration one already tells; I would trim it to a few lines plus a pointer. On CodeRabbit's open comment asking for a literal `<2.0.0` assertion, I disagree. Pinning the string would fail the day otari ports to 2.x, and the behavioral test already covers a ceiling placed too high.

champ18ion:
Thanks for the thorough check, especially pulling the actual wheels to confirm the rename rather than trusting the issue's description. Working on both blocking items now: with_pip=False, and the AGENTS.md Test Notes bullet. Will also dedupe _mcp_requirement() and trim the unit docstring per the non-blocking notes.

champ18ion:
Both blocking items pushed:

1. with_pip=False, verified end to end (real venv, real uv pip install --python, real import check) - confirms what you said, uv doesn't touch the venv's own pip at all.
2. AGENTS.md Test Notes updated with the egress requirement and the fail-hard-after-retries distinction from the other two.

Also did the two non-blocking ones: trimmed the unit docstring to point at the integration test instead of repeating the story, and looked into deduping _mcp_requirement() but left it duplicated - this repo doesn't have a precedent for sharing helpers between tests/unit and tests/integration (tests/unit has no __init__.py, tests/integration does), and setting one up felt like more machinery than an 8-line pure function warrants. Happy to do it differently if you'd rather.

Agreed on the CodeRabbit literal-<2.0.0> point, for the same reason you gave.

champ18ion:
Added timeouts, with a different value than suggested: pytest.ini already caps every test at 120s globally, so 300s would never fire, the global bound kills the test first. Used 60s for the install and 30s for the import check instead, both well under that ceiling.

Still worth adding independent of the global timeout though: pytest-timeout's thread-based mechanism isn't guaranteed to terminate a subprocess blocked on I/O, so a genuinely stuck `uv` could outlive a test that already reported timing out. This bounds the subprocess itself and fails with a clearer, more specific message first. Verified both values don't cause spurious failures on a normal run.

PR #704, opened 2026-08-21
fix(dashboard): let the table's container own its corners, and give the nav rows a pointer

jigjigjig:
## Description

Two things HeroUI's defaults draw that the dashboard's own chrome does not want.

**Table corners.** HeroUI rounds the body's corner cells to `min(32px, var(--radius-2xl))`, a 16px arc off its `--radius: 0.5rem`, because there the body is a card inside the root. `.otari-table` makes the root the only card, at 12px, so that wider arc landed just inside the root's clip: at the bottom the cell's border curved away from the container's and the two left a crescent, and at the top the first row notched away from the header. Fixed at the token, per review: `--radius-2xl: 0` on the root makes HeroUI compute the corners flat itself, on every rule that reads it, so no selector mirrors its DOM. Scoped by what reads the token, which in the built stylesheet is Table rules and the Drawer dialog only; a chip in a cell reads `--radius` and keeps its pill. Separately, the row separator is painted by the cells while HeroUI clears only the *row's* border on the last row, so the last row drew a line on top of the container's, doubled along the flat edge and clipped to a stub inside the corner. That one has no token or prop route, so it stays a rule.

**Nav cursor.** The arrow was on the bare `<button>`s, which carry no HeroUI class: the account menu's rows, the workspace switcher's, and the header's mobile-navigation and collapse toggles. The rail's own rows are `Link`s, HeroUI `Button`s and a `Disclosure.Trigger`, all of which already resolve to `pointer`, so `ROW_BASE` is deliberately left alone and says why.

Base is `feat/nav-shell-to-design`, not `main`: `web/src/app/nav/rowStyles.ts` is introduced there.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

None filed; both were spotted in the running dashboard.

## Checklist

- [x] I have added or updated tests that cover my change (`tests/unit`, `tests/integration`). No test changes: both diffs are presentational, and the screenshot matrix is what covers them once it becomes a gate. Verified in a browser against the real stylesheet instead, and every claim above is a reading rather than an inference: the corner cell measured 16px against the root's 12px before the fix and 0 after, the last row's `border-bottom` went 1px to 0 while mid-table rows and a mid-table detail panel kept theirs, a `Chip` in a cell stayed at 16px under the token override, and each nav element was measured with the cursor utility off (`<a href>` and HeroUI's `.button` / `.disclosure__trigger` are `pointer`, a disabled Button is `not-allowed`, only a bare `<button>` is `default`).
- [x] I ran the Definition of Done checks locally (`make lint`, `make typecheck`, `make test`). The dashboard's counterparts, since the diff is `web/` only: `pnpm --dir web run lint` and `typecheck` pass, and `pnpm --dir web test` is unchanged against this branch's baseline (143 failing, 891 passing, all pre-existing).
- [x] Documentation was updated where necessary. The reasoning is in the stylesheet and at the row strings, where the next reader of those lines will be. One correction there: the claim that these selectors outrank HeroUI's is wrong and is now gone, including where it predated this PR. `@heroui/styles` imports its component CSS with `layer(components)` while this section is unlayered, so specificity never entered into it.
- [x] If the API contract changed, I regenerated the OpenAPI spec (`uv run python scripts/generate_openapi.py`). No contract change.
- [x] **M4 ledger:** this PR either needs no entry (an existing row already covers it and its target is unchanged), or the entry is appended to [otari-ai#1587](https://github.com/mozilla-ai/otari-ai/issues/1587). No entry: no table and no cross-repo surface, only styling on surfaces existing rows already cover.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

Review turned up two things the first pass got wrong, both now measured rather than argued: the selectors never needed their specificity (layers, not specificity, were deciding), and the cursor utility on the rail's shared row was a no-op that would also have overridden a disabled control's `not-allowed`. `DataTable.tsx` ended up back to identical with `main`.

### DIFF
--- a/web/src/app/nav/AccountMenu.tsx
+++ b/web/src/app/nav/AccountMenu.tsx
@@ -55,9 +55,10 @@ const THEME_LABELS: Record<ThemePreference, string> = {
 // where these rows are tapped, so below `md` they take the 44px touch floor and
 // the menu's own scale resumes at the breakpoint.
 const MENU_ROW = `flex min-h-11 w-full items-center gap-2.5 rounded-md px-2.5 text-left text-chrome-row font-medium md:min-h-9 ${NAV_TRANSITION}`
-// `cursor-pointer` because these rows are `<button>`s and Tailwind's reset
-// leaves a button on the default arrow, which reads as inert next to the
-// anchors in the same menu that get the hand for free.
+// `cursor-pointer` because these rows are bare `<button>`s, which the user agent
+// leaves on the default arrow, unlike an `<a href>` or a HeroUI `.button`. On the
+// row states rather than on `MENU_ROW`, so `MENU_ROW_DISABLED` keeps
+// `cursor-not-allowed`.
 const MENU_ROW_RESTING = "cursor-pointer text-foreground hover:bg-surface-alt"
 const MENU_ROW_DISABLED = "cursor-not-allowed text-muted opacity-60"
 // No vertical margin: the dialog's own 6px gap is the menu's rhythm, and a
--- a/web/src/app/nav/AccountMenu.tsx
+++ b/web/src/app/nav/AccountMenu.tsx
@@ -55,7 +55,10 @@ const THEME_LABELS: Record<ThemePreference, string> = {
 // where these rows are tapped, so below `md` they take the 44px touch floor and
 // the menu's own scale resumes at the breakpoint.
 const MENU_ROW = `flex min-h-11 w-full items-center gap-2.5 rounded-md px-2.5 text-left text-chrome-row font-medium md:min-h-9 ${NAV_TRANSITION}`
-const MENU_ROW_RESTING = "text-foreground hover:bg-surface-alt"
+// `cursor-pointer` because these rows are `<button>`s and Tailwind's reset
+// leaves a button on the default arrow, which reads as inert next to the
+// anchors in the same menu that get the hand for free.
+const MENU_ROW_RESTING = "cursor-pointer text-foreground hover:bg-surface-alt"
 const MENU_ROW_DISABLED = "cursor-not-allowed text-muted opacity-60"
 // No vertical margin: the dialog's own 6px gap is the menu's rhythm, and a
 // divider that adds to it sits in a wider gutter than the design draws.
--- a/web/src/app/nav/rowStyles.ts
+++ b/web/src/app/nav/rowStyles.ts
@@ -105,12 +105,15 @@ const ROW_FOCUS =
  * wherever a row is rendered, rather than patching the one heading that has a
  * control inside it.
  *
- * `cursor-pointer` is here for the same reason. The `Link`s get the hand from
- * the user agent, while the expand triggers and the account control are
- * `<button>`s, which Tailwind's reset leaves on the default arrow, so the rail
- * answered the pointer two different ways depending on the row.
+ * Deliberately no `cursor-pointer`: every element this dresses already resolves
+ * to `pointer` on its own. The leaves and both footer rows are `Link`s, so the
+ * `<a href>` takes it from the user agent, and the expand triggers and the
+ * account control carry HeroUI's `.disclosure__trigger` or `.button`, which set
+ * `cursor: var(--cursor-interactive)`. Naming it here would be a no-op on all
+ * five call sites and would also outrank `status-disabled`'s `--cursor-disabled`
+ * on the account control, so a disabled row would promise a click.
  */
-const ROW_BASE = `flex min-h-11 w-full cursor-pointer items-center gap-3 rounded-lg px-3 font-sans text-sm font-medium leading-[1.375rem] ${NAV_TRANSITION} ${ROW_FOCUS}`
+const ROW_BASE = `flex min-h-11 w-full items-center gap-3 rounded-lg px-3 font-sans text-sm font-medium leading-[1.375rem] ${NAV_TRANSITION} ${ROW_FOCUS}`
 
 /**
  * `data-pressed` alongside `active` because a rail row is three different
--- a/web/src/app/nav/rowStyles.ts
+++ b/web/src/app/nav/rowStyles.ts
@@ -104,8 +104,13 @@ const ROW_FOCUS =
  * Naming the family on the shared row is what keeps the rail one typeface
  * wherever a row is rendered, rather than patching the one heading that has a
  * control inside it.
+ *
+ * `cursor-pointer` is here for the same reason. The `Link`s get the hand from
+ * the user agent, while the expand triggers and the account control are
+ * `<button>`s, which Tailwind's reset leaves on the default arrow, so the rail
+ * answered the pointer two different ways depending on the row.
  */
-const ROW_BASE = `flex min-h-11 w-full items-center gap-3 rounded-lg px-3 font-sans text-sm font-medium leading-[1.375rem] ${NAV_TRANSITION} ${ROW_FOCUS}`
+const ROW_BASE = `flex min-h-11 w-full cursor-pointer items-center gap-3 rounded-lg px-3 font-sans text-sm font-medium leading-[1.375rem] ${NAV_TRANSITION} ${ROW_FOCUS}`
 
 /**
  * `data-pressed` alongside `active` because a rail row is three different
--- a/web/src/shared/components/DataTable.tsx
+++ b/web/src/shared/components/DataTable.tsx
@@ -184,6 +184,10 @@ export function DataTable<Row extends object>({
       // Its content stays in the accessibility tree as ordinary elements.
       row.setAttribute("role", "presentation")
       const cell = document.createElement("td")
+      // Named so globals.css can style it at a specificity that outranks
+      // HeroUI's rounding of the body's corner cells: a panel expanded under the
+      // bottom row makes this cell one of those corners.
+      cell.className = "otari-detail-cell"
       row.appendChild(cell)
       hostRef.current = { row, cell }
     }
--- a/web/src/shared/components/DataTable.tsx
+++ b/web/src/shared/components/DataTable.tsx
@@ -184,10 +184,6 @@ export function DataTable<Row extends object>({
       // Its content stays in the accessibility tree as ordinary elements.
       row.setAttribute("role", "presentation")
       const cell = document.createElement("td")
-      // Named so globals.css can style it at a specificity that outranks
-      // HeroUI's rounding of the body's corner cells: a panel expanded under the
-      // bottom row makes this cell one of those corners.
-      cell.className = "otari-detail-cell"
       row.appendChild(cell)
       hostRef.current = { row, cell }
     }
--- a/web/src/styles/globals.css
+++ b/web/src/styles/globals.css
@@ -843,15 +843,36 @@ h6 {
    the edges and the root's `overflow: hidden` clips it (and the last row) to one
    12px radius, so header and card share the same corner.
 
-   Column selectors are replicated with the `.otari-table` prefix so they outrank
-   HeroUI's own first/last-column radius rules (which are more specific than a
-   bare `.table__column`). */
+   Nothing in this section wins on specificity, and none of it needs to:
+   `@heroui/styles` imports its component CSS with `layer(components)` while this
+   file's rules are unlayered, and an unlayered declaration beats a layered one
+   whatever its specificity. So the replicated column selectors are here to
+   *match* HeroUI's virtualized and non-virtualized shapes, not to outrank
+   anything. */
 .otari-table.table-root {
   background-color: var(--color-surface);
   border: 1px solid var(--color-border);
   border-radius: 0.75rem; /* rounded-xl */
   padding: 0;
   overflow: hidden;
+
+  /* HeroUI rounds the body's four corner cells (and a virtualized body) to
+     `min(32px, var(--radius-2xl))`, a 16px arc off its `--radius: 0.5rem`. That
+     is right for the primary variant, where the body is a card inside the root,
+     and wrong here, where the root is the only card: the wider arc was drawn
+     just inside the root's 12px clip, so at the bottom the cell's border curved
+     away from the container's and the two crossing arcs left a crescent, while
+     at the top the first row notched away from the header.
+
+     Setting the variable rather than the radius lets HeroUI compute the corners
+     flat itself, on every rule that reads it, with nothing to keep in step with
+     its selectors. Aliasing a HeroUI variable to our own value is the pattern
+     the theme blocks already use; the radius ramp is simply not aliased yet.
+     Scoped by what actually reads the token: in the built stylesheet the only
+     `var(--radius-2xl)` consumers are Table rules and the Drawer dialog, and no
+     drawer opens inside a table. A chip or an avatar in a cell reads `--radius`,
+     so both keep their own shape. */
+  --radius-2xl: 0;
 }
 .otari-table .table__header {
   background-color: var(--color-primary-subtle);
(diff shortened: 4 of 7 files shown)


web/src/app/nav/AccountMenu.tsx
khaledosman, line 69: we could make these tailwind utilities instead of javascript variables so they can be reused elsewhere outside of the variable's scope

web/src/styles/globals.css
```diff
@@ -877,6 +877,36 @@ h6 {
...
+   styled. */
+.otari-table.table-root
+  .table__body
+  tr:is(:first-child, :last-child)
+  .table__cell {
+  border-radius: 0;
+}
+.otari-table.table-root .table__body > tr:last-child > .table__cell {
+  border-bottom: 0;
+}
```
khaledosman: is there anyway we can do this in a more generic way as part of the heroui table customisation itself instead of being dependent on the html structure which can be brittle?

web/src/app/nav/AccountMenu.tsx
```diff
@@ -55,7 +55,10 @@ const THEME_LABELS: Record<ThemePreference, string> = {
 // where these rows are tapped, so below `md` they take the 44px touch floor and
 // the menu's own scale resumes at the breakpoint.
 const MENU_ROW = `flex min-h-11 w-full items-center gap-2.5 rounded-md px-2.5 text-left text-chrome-row font-medium md:min-h-9 ${NAV_TRANSITION}`
-const MENU_ROW_RESTING = "text-foreground hover:bg-surface-alt"
+// `cursor-pointer` because these rows are `<button>`s and Tailwind's reset
+// leaves a button on the default arrow, which reads as inert next to the
+// anchors in the same menu that get the hand for free.
+const MENU_ROW_RESTING = "cursor-pointer text-foreground hover:bg-surface-alt"
```
khaledosman, line 62: `WorkspaceSwitcher.tsx` duplicates this exact pair (`MENU_ROW` / `MENU_ROW_RESTING`, lines 15-16) and gets none of the fix, so the popover the rail's scope row opens still answers with the arrow while the account menu beside it answers with the hand: the split this PR sets out to close.

`cursor-pointer` is owed there on `MENU_ROW_RESTING` (16) and `MENU_ROW_CURRENT` (21), plus the two inline row strings at 212 and 239 that bypass `MENU_ROW_RESTING`. The non-control organization `<div>` at 153 correctly keeps the default, which is also why it cannot go on that file's base `MENU_ROW`.

web/src/styles/globals.css
```diff
@@ -877,6 +877,36 @@ h6 {
...
+   away from the container's 12px one and the two crossing arcs left a crescent,
+   and at the top the first row's corners notched away from the header. The
+   root's radius plus its `overflow: hidden` is the whole shape, so flatten these.
+
+   And the separator between rows is painted by the cells, while HeroUI clears
+   only the *row's* border on the last row, so the last row still drew a line
+   directly on top of the container's bottom border: doubled along the flat edge,
+   clipped to a stub inside the corner. The container's border closes the table.
+
+   Both carry `.table-root` alongside `.otari-table`, for the specificity to
```
khaledosman: The stated mechanism is not the one doing the work. `@heroui/styles/dist/index.css` imports its component CSS with `layer(components)`, while this whole section of `globals.css` is unlayered, and an unlayered declaration beats any layered one regardless of specificity.

Confirmed in the built bundle: HeroUI's `tr:first-child td:first-child` lands inside `@layer components`, `.otari-table.table-root{...}` lands after the utilities layer closes. So `.table-root` buys nothing in any of the four selectors it was added to, and the comment (plus the same claim already at 846, about the column rules) teaches the next reader a rule that does not hold. Either drop the compound or restate the reason as the section being unlayered.

web/src/shared/components/DataTable.tsx
```diff
@@ -184,6 +184,10 @@ export function DataTable<Row extends object>({
       // Its content stays in the accessibility tree as ordinary elements.
       row.setAttribute("role", "presentation")
       const cell = document.createElement("td")
+      // Named so globals.css can style it at a specificity that outranks
+      // HeroUI's rounding of the body's corner cells: a panel expanded under the
+      // bottom row makes this cell one of those corners.
+      cell.className = "otari-detail-cell"
```
khaledosman: `otari-detail-cell` is now a load-bearing contract between this file and `globals.css`, with nothing pinning it. `DataTable.test.tsx` already reaches for `.otari-detail-row`, so one assertion there (the host row's `firstElementChild` carries `otari-detail-cell`) keeps a rename from silently restoring the 32px corner under an expanded bottom row.

khaledosman (approved):
Three findings inline. Checked and fine: the empty/loading state renders a class-less `<td>`, so HeroUI's corner rounding still matches it but paints nothing (no background, no border); the detail row under the last data row is covered by both new rules; nothing sits inside `Table.Root` below the tbody, so dropping the last row's cell border is safe. Biome passes at head, so the removed `noDescendingSpecificity` suppression was correctly retired.

_Review by Claude Opus 5 via the Claude Code CLI._

web/src/styles/globals.css
```diff
@@ -877,6 +877,36 @@ h6 {
...
+   outrank HeroUI's own corner-cell rules, which are three classes deep. The
+   detail row's cell needs the same pair and gets them below, where it is
+   styled. */
+.otari-table.table-root
```
khaledosman: Adding to my earlier comment on this block: the corner half has a token hook that removes the need for the selector.

Every radius being fought here is `min(32px, var(--radius-2xl))`, and in the built stylesheet that resolves to `calc(var(--radius) * 2)` off HeroUI's `--radius: 0.5rem`, which is the 16px arc measured in the description. `--radius-2xl` is one of HeroUI's own variables, and aliasing those to our tokens is the pattern the theme blocks already use (`--focus`, `--surface`, `--surface-secondary`, ...); the radius ramp is just not aliased yet. Setting `--radius-2xl: 0` on `.otari-table.table-root` makes HeroUI compute the corners flat itself, on every rule that reads it, with nothing to keep in sync with `.table__body tr:first-child td:first-child` and no priority contest, so the specificity question in my other comment disappears along with the selector. It inherits into cell content, so a HeroUI surface rendered inside a cell or the detail panel flattens too: the one `rounded-2xl` in `web/src` (`UsagePage.tsx:1388`) is outside a table, and re-setting the variable on `.otari-detail-cell` covers the panel.

Separately, and pre-existing rather than this PR's: `Table.Root` takes `variant`, and `secondary` is "No background, padding, or rounding on the root", which is what 849-855 neutralizes by hand. `components.md`'s "props over `className`" points at the prop. Not a clean swap (secondary brings its own standalone-rounded-header treatment, which this block already overrides), so a follow-up at most.

The last row's `border-bottom` has no prop or token route (HeroUI clears the *row's* border while the cells paint the separator), so that rule stands as written.

_Follow-up by Claude Opus 5 via the Claude Code CLI._

web/src/app/nav/rowStyles.ts
```diff
@@ -104,8 +104,13 @@ const ROW_FOCUS =
...
  * wherever a row is rendered, rather than patching the one heading that has a
  * control inside it.
+ *
+ * `cursor-pointer` is here for the same reason. The `Link`s get the hand from
+ * the user agent, while the expand triggers and the account control are
+ * `<button>`s, which Tailwind's reset leaves on the default arrow, so the rail
+ * answered the pointer two different ways depending on the row.
  */
-const ROW_BASE = `flex min-h-11 w-full items-center gap-3 rounded-lg px-3 font-sans text-sm font-medium leading-[1.375rem] ${NAV_TRANSITION} ${ROW_FOCUS}`
+const ROW_BASE = `flex min-h-11 w-full cursor-pointer items-center gap-3 rounded-lg px-3 font-sans text-sm font-medium leading-[1.375rem] ${NAV_TRANSITION} ${ROW_FOCUS}`
```
njbrake: Measured no-op, and the reason above it does not hold. With this utility reverted, every element `navRowClass` dresses already resolves to `pointer`: the leaves and both footer rows are `Link`s, so an `<a href>` takes it from the user agent, while a group's expand trigger carries HeroUI's `.disclosure__trigger` and the collapsed triggers and the account control carry `.button`, both of which set `cursor: var(--cursor-interactive)`, and that variable is `pointer`. Only a bare `<button>` would need it, and none of the five `navRowClass` call sites is one.

The `AccountMenu` half is the real fix, since `MenuRow` is a bare button. Suggest dropping this and the paragraph above it, and spending it on `WorkspaceSwitcher` and the two `AppShell` toggles instead.

web/src/styles/globals.css
```diff
@@ -877,6 +877,36 @@ h6 {
...
   border-color: var(--color-border);
 }
+
+/* Two things HeroUI's Table draws for a body that is its own card inside the
+   root, both of which fight the single card `.otari-table` makes of the root,
+   and both of which showed up as artifacts in the corners.
+
+   The body's four corner cells are rounded to the primary variant's radius
+   (`min(32px, --radius-2xl)`), a second and much wider arc drawn just inside the
+   root's 12px clip: at the bottom the cell's own border followed the 32px curve
```
njbrake: The arc is 16px, not 32px: `min(32px, var(--radius-2xl))` resolves off HeroUI's `--radius: 0.5rem`, which is also the figure the PR description gives. The expression on the line above is right; it is the prose reading of it that overstates.

web/src/shared/components/DataTable.tsx
```diff
@@ -184,6 +184,10 @@ export function DataTable<Row extends object>({
       // Its content stays in the accessibility tree as ordinary elements.
       row.setAttribute("role", "presentation")
       const cell = document.createElement("td")
+      // Named so globals.css can style it at a specificity that outranks
```
njbrake: The reason given here is the specificity one, which does not hold (see the thread on `globals.css`). The reason that does hold is in the commit message and not in the tree: matching this cell on `td` inside `.otari-table` puts the rule in the descending-specificity group Biome tracks against the `.otari-markdown td` rules further down the stylesheet. I confirmed both directions, so the class genuinely earns its keep: rewriting the two selectors to use `td` produces three `noDescendingSpecificity` warnings, and keeping the retired suppression produces an unused-suppression warning. Worth moving that sentence up here, since it is what justifies adding a class hook at all.

njbrake (approved):
Approving the change: the corner fix is right and I reproduced it. Four things to settle before this merges.

**1. The base is a merged branch, and it costs this PR its CI.** `feat/nav-shell-to-design` landed on `main` as #638 on 2026-08-20, and all four files here are byte-identical between `main` and that branch, so the stated reason for the non-main base no longer holds. The consequence is concrete: `otari-dashboard.yml` triggers on `pull_request: branches: [ main ]`, so Biome, tsc, vitest, the build and the two drift checks never ran, and CodeRabbit skipped for the same reason. Only the title lint and the template check are green. Please do not retarget in the UI: the merge base is `e798d3eb9`, so a merge would replay the stale branch over `main`'s newer work. Cherry-pick the two commits onto `main` instead, which is conflict free since the files match.

**2. The suite figure in the checklist does not reproduce.** After `pnpm install --frozen-lockfile`, `pnpm --dir web test` at this head is 1034 passed, 0 failed, not 143 failing. Worth correcting, since "the failures are pre-existing" is the kind of note that hides a real one later.

**3. Two more pointer surfaces.** Alongside `WorkspaceSwitcher.tsx`, the mobile drawer toggle (`AppShell.tsx:650`) and the sidebar collapse toggle (`AppShell.tsx:673`) are bare `<button>`s and still compute `default`.

**4. Three inline**, on the two comment claims and on the rail row string.

I agree with the findings already open (layer over specificity, the switcher gap, the missing `otari-detail-cell` assertion), and I verified the `--radius-2xl: 0` route works: it flattens all four corner cells, the detail cell in both positions, and the empty-state cell the selector cannot reach.

Everything above was measured in Chromium against the real `DataTable` and stylesheet, on the base branch and on this head.

web/src/styles/globals.css
jigjigjig: Confirmed and fixed, both the compound and the claim. Checked in the built bundle rather than by reading: HeroUI's `.table__body tr:first-child td:first-child` lands inside `@layer components`, `.otari-table.table-root{...}` lands unlayered, so specificity never entered into it. The `.table-root` compound is gone from the rules that had it, and the comment now says what actually holds, including at 846 where the same claim was made about the column selectors: they are there to match HeroUI's virtualized and non-virtualized shapes, not to outrank anything.

web/src/styles/globals.css
jigjigjig: Adopted, and it is better. `--radius-2xl: 0` on `.otari-table.table-root` replaces the selector entirely: HeroUI computes the corners flat itself, on the body corner cells, the virtualized body, and the secondary-variant column caps alike, with nothing to keep in step with `.table__body tr:first-child td:first-child`.

I went looking for the inheritance leak you flagged and it does not exist, which is worth recording. In the built bundle `.chip` compiles to `border-radius: calc(var(--radius) * 2)`, so it reads `--radius`, not `--radius-2xl`: Tailwind inlines `rounded-2xl` to the theme expression rather than the named variable. Measured a `Chip` inside a cell with the override live and it stayed at 16px. Every `var(--radius-2xl)` consumer that ships is a Table rule or `.drawer__dialog`, and no drawer opens inside a table, so the override needs no counter-set on the panel and `.otari-detail-cell` is retired with it. `DataTable.tsx` is back to what it was.

The last row's `border-bottom` stands as written, as you say. On the `variant="secondary"` note: agreed it is not a clean swap and out of scope here.

web/src/app/nav/AccountMenu.tsx
jigjigjig: Fixed. `cursor-pointer` on `MENU_ROW_RESTING` and `MENU_ROW_CURRENT`, plus the two inline strings at the create-workspace and create-organization rows that bypass them. The organization `<div>` keeps the arrow for exactly the reason you give, and that is why it is on the row states rather than on that file's base `MENU_ROW`; there is now a comment saying so, since the same string in `AccountMenu.tsx` can safely carry it and the difference would otherwise look accidental.

web/src/styles/globals.css
jigjigjig: Yes, and that is what it does now. Every radius involved is `min(32px, var(--radius-2xl))`, so setting `--radius-2xl: 0` on the table root makes HeroUI compute them flat itself; the structural selector is gone. What is left depends on the DOM only where HeroUI leaves no other route: the last row's separator is painted by the cells while HeroUI clears the row's border, so there is no prop and no token for it.

web/src/app/nav/AccountMenu.tsx
jigjigjig: The concrete half of this, the same pair duplicated in `WorkspaceSwitcher.tsx` and missing the fix, is done in your other comment.

On the form: a `@utility` would work, but the two files' rows are not the same row. `AccountMenu`'s is `text-chrome-row md:min-h-9` and the switcher's is a flat `text-sm`, so one utility means one of them takes a shape it was not drawn with, and a parameterized pair is what `rowStyles.ts` already does for the rail (`navRowClass`). Extending that module to own the menu rows too is the version I would want, and it is a refactor across both files rather than part of a cursor fix, so I left it out. Happy to do it as a follow-up if you want it before the shell lands.

web/src/app/nav/rowStyles.ts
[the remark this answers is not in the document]
jigjigjig: Added, in `rowStyles.test.ts` alongside the other shared-shape assertions: `cursor-pointer` is checked on the resting, selected, nested and collapsed variants.

web/src/shared/components/DataTable.tsx
jigjigjig: Resolved by removing the contract rather than pinning it. The corner half is now a token on the root (`--radius-2xl: 0`), which the detail cell inherits like any other, so the stylesheet has nothing to target it by and `otari-detail-cell` is gone; `DataTable.tsx` is byte-identical to what it was. What is left for the panel is the border-bottom pair, expressed on `> td` in the rules next to the detail row itself.

web/src/app/nav/rowStyles.ts
jigjigjig: You are right, and I reproduced it before reverting. Rendered each element with the utility off and read the computed cursor: `<a href>` is `pointer` from the user agent, `.button` and `.disclosure__trigger` are `pointer` off `--cursor-interactive`, and only a bare `<button>` is `default`. All five `navRowClass` call sites are one of the first three, so the rail never wore the arrow.

One more reason to drop it that the measurement turned up: a disabled HeroUI Button resolves to `not-allowed` via `status-disabled`, and a utility outranks that, so the account control would have promised a click while disabled. Reverted, the assertion that pinned the no-op is gone, and the measurement is now the comment above `ROW_BASE` so the next reader does not add it back.

Spent on the bare buttons instead, as you suggest: the header's mobile-navigation and collapse toggles are in the new commit, and `WorkspaceSwitcher` is in the one before it. The skip link is bare too and left alone deliberately, since it is `sr-only` until focused and reached by keyboard.

web/src/styles/globals.css
jigjigjig: Right, and gone. That prose block was replaced wholesale by the token comment, which now reads `min(32px, var(--radius-2xl))`, a 16px arc off its `--radius: 0.5rem`, matching the figure measured off the corner cell.

web/src/shared/components/DataTable.tsx
jigjigjig: Both directions confirmed, and the class is gone anyway: the change that retires it is the token, not a different selector. With `--radius-2xl: 0` on the root there is no radius rule left to write, so the only cell rule remaining is the last row's `border-bottom`, one `> td` selector at (0, 2, 2) placed directly after `.otari-detail-row > td` at (0, 2, 1). That is ascending, the retired suppression is restored where it was, and Biome is clean at head, which is what your three-warnings result was measuring against the version that still carried the radius selector. `DataTable.tsx` is byte-identical to `main` now.

PR #712, opened 2026-08-21
fix(dashboard): open the organization rail inside the mobile drawer

jigjigjig:
## Description

On a phone the sidebar's **Organization** row navigated, and the drawer closed over the result. So the rail that row swaps to was never a thing you got to read: you landed on the organization's first page and had to reopen the menu to reach any of the others. On the desk the same row is fine, because the rail it swaps to stays on screen beside the page.

Below the breakpoint that row is now a submenu trigger:

- It opens the organization rail one level down **inside** the drawer, with nothing routed yet. `showOrganizationRail = inOrganization || (isMobile && mobileOrgNavOpen)` drives the section list, the head of the rail, and the footer row, so the drawer swaps to the organization's own sections.
- The tap that picks a destination in it is the tap that navigates and dismisses the drawer.
- The head of that level is a "Back to `<workspace>`" **button**, not the link the organization rail's own head is: the page under the drawer never left the workspace, so there is nothing to go back to yet.
- Escape unwinds one level at a time (the submenu, then the drawer). Closing or reopening the drawer returns to the workspace menu, and crossing to a desktop viewport drops both, so a submenu cannot outlive the drawer that framed it.
- Focus follows the level in each direction. A breakpoint change that unmounts the submenu keeps focus on the navigation landmark.
- The trailing chevron is back on the row, now that it promises a submenu that opens. `rowStyles.ts` explains why those two bare buttons name `cursor-pointer` themselves while the shared row still does not.

**Desktop behavior is unchanged**: the row is still a `Link` to the remembered organization location, and the submenu trigger never renders beside it.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

None filed; reported directly.

## Checklist

- [x] I have added or updated tests that cover my change, including opening without navigating, destination dismissal, back navigation, Escape unwinding, drawer and breakpoint resets, focus restoration, and unchanged desktop linking.
- [x] I ran the dashboard Definition of Done checks on Node 22: Biome checked 217 files, TypeScript passed, all 1,145 Vitest tests passed, and the production build succeeded.
- [x] Documentation was updated where necessary (the reasoning comments in `AppShell.tsx` and `rowStyles.ts`; no user-facing guide text changes, since the row keeps its name and its destinations).
- [x] If the API contract changed, I regenerated the OpenAPI spec. Not applicable: no backend change.

Not verified in a browser: signing in needs the deployment's master key. The behavior is covered at the `matchMedia` level in Vitest, and the mobile screenshot projects capture pages rather than menu states.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:** Claude Opus 5 in Claude Code and OpenAI Codex

The bug report, the "treat Organization as a submenu" framing, and the scope (mobile only, no desktop change) came from the maintainer. AI agents implemented the change, added the tests, addressed review feedback, and prepared this description.

## Summary

- Added mobile organization submenu navigation inside the drawer.
- Added back navigation, Escape handling, focus restoration, breakpoint reset, and drawer reset behavior.
- Preserved direct organization links on desktop.
- Updated `ConnectionStatus` to subscribe to query-cache state safely.
- Expanded `AppShell` tests for mobile and desktop navigation behavior.
- Documented cursor behavior for shared navigation row styles.

These changes make mobile organization navigation more predictable while preserving existing desktop behavior.

### DIFF
--- a/web/src/app/AppShell.tsx
+++ b/web/src/app/AppShell.tsx
@@ -349,6 +349,7 @@ export function AppShell() {
   const toggleRef = useRef<HTMLButtonElement>(null)
   const orgNavTriggerRef = useRef<HTMLButtonElement>(null)
   const orgNavBackRef = useRef<HTMLButtonElement>(null)
+  const restoreSidebarFocusRef = useRef(false)
   const [collapsed, setCollapsed] = useState<boolean>(readStoredCollapsed)
   const [isMobile, setIsMobile] = useState<boolean>(readIsMobile)
   const [mobileNavOpen, setMobileNavOpen] = useState(false)
@@ -393,8 +394,12 @@ export function AppShell() {
       return
     const query = window.matchMedia(MOBILE_QUERY)
     const onChange = (event: MediaQueryListEvent) => {
+      if (!event.matches) {
+        restoreSidebarFocusRef.current =
+          asideRef.current?.contains(document.activeElement) ?? false
+        closeMobileNav()
+      }
       setIsMobile(event.matches)
-      if (!event.matches) closeMobileNav()
     }
     // Safari < 14 (and some older engines) only expose the deprecated
     // addListener/removeListener; fall back to it so the shell doesn't throw.
@@ -424,11 +429,17 @@ export function AppShell() {
   // Focus management for the mobile drawer, which is a modal overlay: move focus
   // into it when it opens and restore focus to the toggle when it closes, so
   // keyboard and screen-reader users are neither stranded inside a hidden panel
-  // nor dropped back to the top of the document. The isMobile guard means a
-  // breakpoint change to desktop (which also closes the drawer) never yanks focus
-  // to the now-hidden toggle.
+  // nor dropped back to the top of the document. A breakpoint change to desktop
+  // keeps focus on the sidebar when it came from the drawer, rather than moving
+  // it to the now-hidden toggle or losing it when the submenu unmounts.
   useEffect(() => {
-    if (!isMobile) return
+    if (!isMobile) {
+      if (restoreSidebarFocusRef.current) {
+        restoreSidebarFocusRef.current = false
+        asideRef.current?.focus()
+      }
+      return
+    }
     if (mobileNavOpen) {
       asideRef.current?.focus()
     } else if (asideRef.current?.contains(document.activeElement)) {
@@ -524,7 +535,10 @@ export function AppShell() {
           // lives in that bar. While closed it is off-canvas, so inert takes its
           // links out of the tab order and the accessibility tree until opened.
           aria-label={isMobile ? "Navigation" : undefined}
-          tabIndex={isMobile ? -1 : undefined}
+          // Programmatically focusable on both layouts so a breakpoint change
+          // can preserve focus on the navigation landmark without adding it to
+          // the natural tab order.
+          tabIndex={-1}
           inert={isMobile && !mobileNavOpen ? true : undefined}
           className={clsx(
             "flex flex-col gap-4 border-r border-border bg-background-alt p-3 focus:outline-none",
--- a/web/src/app/AppShell.tsx
+++ b/web/src/app/AppShell.tsx
@@ -7,6 +7,7 @@ import type { IconType } from "react-icons"
 import {
   FiArrowLeft,
   FiChevronDown,
+  FiChevronRight,
   FiMenu,
   FiSettings,
   FiSidebar,
@@ -312,12 +313,6 @@ export function AppShell() {
   // workspace one, so the two never render together.
   const navContext = navContextForPath(pathname)
   const inOrganization = navContext === "organization"
-  // Filtered before it is indexed, so the divider and top margin below key off
-  // the first *rendered* section rather than the first registered one.
-  const visibleSections = visibleNavSections(
-    inOrganization ? ORG_NAV_SECTIONS : NAV_SECTIONS,
-    isVisible,
-  )
   const organization = useOrganizationContext()
   const { selected: selectedWorkspace } = useSelectedWorkspace()
   // Always true in a standalone deployment, where the one session is the local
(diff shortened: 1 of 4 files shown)


web/src/app/ConnectionStatus.tsx
```diff
@@ -16,23 +16,27 @@ function isUnreachable(error: unknown): boolean {
...
-  const [unreachable, setUnreachable] = useState(false)
+  const cache = useQueryClient().getQueryCache()
 
-  useEffect(() => {
-    const cache = queryClient.getQueryCache()
-    const compute = () =>
+  // Read through `useSyncExternalStore` rather than by writing a subscription
+  // into state, and that is load-bearing rather than tidying: mounting a
+  // component that holds a query builds that query into this cache *during
+  // render*, and the cache notifies its subscribers synchronously. A `setState`
```
khaledosman: The premise here does not hold. `QueryCache.notify()` wraps every event in `notifyManager.batch()`, and that flush goes through `systemSetTimeoutZero` (`@tanstack/query-core@5.101.4/build/modern/notifyManager.js`); nothing in `@tanstack/react-query` calls `setScheduler`. A cache subscriber therefore never runs during a render, so there is no cross-component `setState` for React to warn about.

Verified rather than reasoned: a probe that mounts a fresh query key during a sibling's render emits no warning under either the previous `useEffect` + `setState` version or this one, and `AppShell.test.tsx` prints none either way.

Keep `useSyncExternalStore` if you prefer the shape (it is fewer renders and no effect), but drop or correct the "notifies its subscribers synchronously ... which React warns about" claim, since the comment is presented as load-bearing. Related: nothing in the suite distinguishes the two implementations, so a revert of this file would be silent.

web/src/app/AppShell.tsx
```diff
@@ -506,20 +556,42 @@ export function AppShell() {
...
+                // *to* yet. Same row and same name as the link above, because it
+                // means the same thing to whoever reads it and differs only in
+                // whether the route has moved yet. `cursor-pointer` because a
+                // bare button resolves to the default arrow, which is the one
+                // thing `navRowClass` leaves to its call sites.
+                <button
+                  type="button"
+                  ref={orgNavBackRef}
+                  onClick={() => setMobileOrgNavOpen(false)}
+                  className={`${navRowClass()} cursor-pointer`}
```
khaledosman, line 601: `cursor-pointer` hardcodes the value the rest of the rail resolves through a variable. `--cursor-interactive` is defined as `pointer` in `@heroui/styles/dist/themes/default/variables.css:30`, and it is what `.button` (`button.css:18`) and `.disclosure__trigger` (`disclosure.css:11`) read, which is what the neighboring rows pick it up from.

Use `cursor-(--cursor-interactive)` here and on line 701, and name the variable rather than the literal in the `rowStyles.ts` note (lines 108-118), so a theme changing that variable moves these two rows with the others.

web/src/app/AppShell.tsx
```diff
@@ -506,20 +556,42 @@ export function AppShell() {
...
+              ) : (
+                // The submenu's way back, which closes a level rather than
+                // navigating: the page under the drawer is still the workspace
+                // page you opened the menu from, so there is nothing to go back
+                // *to* yet. Same row and same name as the link above, because it
+                // means the same thing to whoever reads it and differs only in
+                // whether the route has moved yet. `cursor-pointer` because a
+                // bare button resolves to the default arrow, which is the one
+                // thing `navRowClass` leaves to its call sites.
+                <button
```
khaledosman, line 597: The row body (icon, then a `min-w-0 flex-1 truncate` span) is now hand-written two more times, here and at line 697. `NavRowLink` exists in this file precisely so that cannot happen: its docstring says the shared row is what keeps the leaves, a group's children, and a collapsed single child from drifting.

Extract a `NavRowButton` beside it, taking label, icon, an optional trailing indicator, a ref, and the handler, and route both new controls through it. That also puts the `cursor` fix and the `text-left` in one place instead of two.

khaledosman (approved):

One note that is not a finding: the branch is behind `main` by #710 (Vite 8, plugin-react 6), so the local run in the description was on the previous toolchain. I rebased onto `main` and re-ran there: `tsc -b --noEmit` clean, Biome clean, 143 `src/app` tests passing. Mergeable, just stale.

<sub>Reviewed by Claude Opus 5 in Claude Code.</sub>

web/src/app/AppShell.tsx
khaledosman: Withdrawing this one. `--cursor-interactive` exists so a design system can opt out of pointer cursors, and this dashboard never sets it to anything but `pointer`, so routing through it buys no adaptability here and `cursor-pointer` is the more legible of the two at the call site. `cursor-pointer` as written is fine, and so is the `rowStyles.ts` note naming it.

<sub>Claude Opus 5 in Claude Code.</sub>

web/src/app/AppShell.tsx
khaledosman: Withdrawing this one too; the standards prescribe the opposite of what I asked for.

`components.md` puts a repeating look on a shared class string and names this exact one: "a value-shaped decision, rather than a markup-shaped one, is where a utility or a shared class string comes in ... `navRowClass` and `NAV_TRANSITION` in `app/nav/rowStyles.ts` carry a chrome row." Both new buttons already go through `navRowClass()`, so the part that can drift is already owned.

And `component-architecture.md` scopes the no-copied-markup rule to a wrapper appearing in two files, then says "a `variant` prop beats a second near-identical component. Two components that differ by 10% of their markup will not stay in step" — a `NavRowButton` beside `NavRowLink` is that second component. Two short JSX bodies in one file, sharing the class string, is the shape the standards ask for. No change wanted here.

<sub>Claude Opus 5 in Claude Code.</sub>

web/src/app/AppShell.tsx
```diff
@@ -369,7 +394,7 @@ export function AppShell() {
     const query = window.matchMedia(MOBILE_QUERY)
     const onChange = (event: MediaQueryListEvent) => {
       setIsMobile(event.matches)
-      if (!event.matches) setMobileNavOpen(false)
+      if (!event.matches) closeMobileNav()
```
jigjigjig: [P2] Preserve focus when this breakpoint change closes the organization submenu. If its Back button or a destination owns focus, `closeMobileNav()` unmounts that node; after `isMobile` becomes false, both focus effects return early, so focus falls to `<body>`. Restore focus to a surviving desktop target after the layout switch, and extend the breakpoint test to assert focus.

jigjigjig (commented):
Codex review: one accessibility blocker around focus restoration during a mobile-to-desktop breakpoint change.

web/src/app/AppShell.tsx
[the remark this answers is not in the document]
jigjigjig: Addressed in adf4e63: the test now enters the submenu, crosses to desktop and back, verifies preserved focus, reopens the drawer, and confirms the workspace level is reset.

web/src/app/ConnectionStatus.tsx
```diff
@@ -16,23 +16,27 @@ function isUnreachable(error: unknown): boolean {
...
+  // in that callback is therefore an update to this component while a different
+  // one is rendering, which React warns about. Swapping the sidebar's head
+  // (leaving the mobile drawer's organization submenu remounts the workspace
+  // switcher, which holds two queries) is one way to reach it. This is the API
+  // for an external store that can change mid-render.
+  return useSyncExternalStore(
+    (onStoreChange) => cache.subscribe(onStoreChange),
```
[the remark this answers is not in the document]
jigjigjig: Addressed in adf4e63 by removing the unrelated ConnectionStatus refactor. QueryCache notifications are batched, so the proposed render-warning regression test is not applicable.

web/src/app/ConnectionStatus.tsx
jigjigjig: Addressed in adf4e63: reverted the unnecessary ConnectionStatus refactor and removed its incorrect load-bearing rationale from the PR description.

web/src/app/AppShell.tsx
jigjigjig: Addressed in adf4e63: a breakpoint change now restores focus to the surviving sidebar landmark, with regression coverage for the desktop and mobile round trip.

Issue #661, opened 2026-08-19
Adopt the reconciled Decimal cost-math core


njbrake:
- Labels: `otari-migration`, `open-core`
- Milestone: `otari-migration-m5`
- Part of mozilla-ai/otari-ai#1452, pricing and usage track

## Why this exists

The pricing lookup chain here is the better of the two and survives: it prices tool calls and images, carries the one-hour cache-write meter and tiered rates, and has a refresh and init pair the platform has no equivalent of.

The arithmetic underneath it does not. `services/metered_pricing.py` uses `float` throughout, represents a single cached-token convention, and reads rates from `model_pricing` columns that are themselves `float`. The platform's cost-math core is `Decimal` throughout and represents **both** cached-token conventions, because the two sides of the merge disagreed on which one they speak and it reconciled them rather than picking one.

So the reconciliation work is already done, on the side that would otherwise be deleted, and the implementation that survives by default is the one carrying float money across settlement rows.

## What to produce

Replace the arithmetic with the reconciled `Decimal` core, keeping the lookup chain as it is. The rate columns become exact rather than binary floating point, which is a migration on `model_pricing` and not only a code change.

## Definition of Done

- Costs are computed in `Decimal` end to end, from the stored rate to the settled row.
- Both cached-token conventions are representable, and which one a caller speaks is explicit rather than assumed.
- Existing priced rows reprice to the same values within the rounding the change defines, and that rounding is stated.

_Note: this issue was drafted by Claude Opus 5 via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

PR #722, opened 2026-08-21
feat(budgets): make the spend ledger exact, like the settled rows feeding it

### LINKED ISSUE #691: Make the budget spend ledger exact, like the settled rows feeding it
- Labels: `area/backend`, `otari-migration`
- Follows mozilla-ai/otari#661

## What is true today

Settled costs are exact after #661: `usage_logs.cost` is `NUMERIC(18, 6)` and the arithmetic behind it is `Decimal` end to end. The counters that enforce budgets are not. `users.spend`, `users.reserved`, `budgets.max_budget`, `scoped_budgets.current_spend` / `reserved_spend` / `max_budget`, and `workspace_budget_defaults.max_budget` are all `Float`, so an exact settled amount is narrowed to reach them.

This is visible rather than theoretical. Driving four completions through a live gateway on PostgreSQL left `users.spend` at `0.6619999999999999` while every row that summed to it is exact to the micro-dollar. The drift is tiny per request and unbounded over a period: it accumulates across every reconcile until the budget resets, and it lands in the value a 403 is decided against.

## Why #661 did not address it

That PR's scope was the stored rate and the settled row, which mozilla-ai/otari-ai#1751 named the accounting truth. The budget engine is a different concern with its own conditional-UPDATE arithmetic, its own clamp expression, and its own API and dashboard surface, and mozilla-ai/otari-ai#1725 owns the hosted spend counters. Converting it in the same change would have put two independent risky migrations in one PR, one of which already rewrites `usage_logs`. The narrowing is instead confined to two named boundaries in `budget_service.py`, each commented at the site.

## What it would take

- Retype the six counters to `NUMERIC(18, 6)` using the `UsdCost` type `usage_logs.cost` already uses. One migration, the same shape as #661's, with the same PostgreSQL-rewrites-the-table caveat.
- Make the reservation arithmetic Decimal-safe: the conditional `UPDATE` that pre-debits, the `CASE` clamp in `_release_reserved`, and the `spend + reserved < max_budget` comparison.
- Drop the `float()` narrowing in `reconcile_reservation`, so a settled `Decimal` reaches the counter unchanged.
- Decide deliberately whether the reserve-time estimate stays float. It is an upper bound reconciled against an exact settlement, so it can, but the decision should be written down rather than left as an accident of where the conversion happens to sit.
- The API and dashboard boundaries stay float; the wire contract does not change.

## Definition of Done

- A user's spend equals the exact sum of the rows that produced it, over an arbitrary number of requests.
- Budget enforcement compares exact values, and the reservation lifecycle still holds under the existing race tests.
- `src/gateway/AGENTS.md`'s cost-math section is updated: it currently documents where a `Decimal` narrows to reach a float counter, and those boundaries would no longer exist.

njbrake:
## Description

Settled costs became exact in #661, but the counters the budget gate enforces against stayed `Float`, so an exact amount was narrowed to reach them. Four completions through a live gateway left `users.spend` at `0.6619999999999999` while every row that summed to it was exact to the micro-dollar, and the error accumulated across every reconcile until the budget reset, in the value a 403 is decided against.

Eight columns become `NUMERIC(18, 6)` via the `UsdCost` type `usage_logs.cost` already uses (`users.spend` / `reserved`, `budgets.max_budget`, `scoped_budgets.max_budget` / `current_spend` / `reserved_spend`, `workspace_budget_defaults.max_budget`, `budget_reset_logs.previous_spend`), and the reservation arithmetic goes with them: the conditional UPDATE, both CASE clamps, and the cap comparison.

**The reserve-time estimate stays exact rather than narrowing, and that is the decision the issue asked to be made deliberately.** It could have stayed float, but on the stream-without-usage path it is what settles the usage row (`cost_override=reservation.estimate`), which makes it accounting; and a hold released at the amount it was taken needs both sides written the same way. The reasoning is in `src/gateway/AGENTS.md` rather than left implicit.

Two things beyond the issue's list, both flagged for review. `budget_reset_logs.previous_spend` converts too, or the narrowing the DoD removes would come straight back at that site. And `MAX_USD_LIMIT` bounds the three cap request models, because the exact column has a ~$1T ceiling the float one did not: a larger cap used to return 500 from a numeric overflow, and an already-stored one aborted the migration with a bare `NumericValueOutOfRange` at startup.

The wire contract keeps its float types; amounts narrow on the way out only.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes #691. Follows #661.

## Checklist

Run locally against PostgreSQL 18 (no Docker in the sandbox, so Testcontainers was replaced by a local server; CI uses `postgres:17`): `make lint`, `mypy` (452 files), `make openapi-check`, `make postman-check`, 2204 unit tests, 1511 integration tests, and the OSS-edition smoke gate on SQLite. The migration was upgraded and downgraded by hand on both engines with data in place, and the SQLite result diffed against `create_all`. The only OpenAPI change is the new `maximum` on six cap schemas; `web/src/client/schema.ts` regenerates with no diff.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

Ported from the equivalent implementation in `mozilla-ai/otari-ai` (`BudgetService`, `BudgetRepository.try_reserve` / `release_reserved` / `reset_period`), with two deliberate differences: the platform clamps with `func.greatest`, which SQLite has no equivalent for, so this repo's `CASE` stays and both arms became `Decimal` so PostgreSQL resolves it as `numeric`; and the platform's budget rows are `NUMERIC(12, 4)` where its wallet counter is `NUMERIC(18, 6)`. This uses 18/6 throughout, since 12/4 would round every settled row on the way into the counter that sums it.

After implementation a separate review agent read the diff cold; four of its five findings were acted on (the $1T ceiling above, a `coalesce(sum(spend), 0.0)` that PostgreSQL resolved as `double precision`, and two overstated claims in prose).

_Note: this PR description was drafted by Claude Opus 5 via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

## Summary

- Converted budget and spend counters to exact `NUMERIC(18, 6)` values.
- Updated reservation, settlement, and reconciliation calculations to preserve exact amounts.
- Centralized budget settings and added calendar-based reset alignment.
- Kept float conversion at API and dashboard boundaries.
- Added request limits to prevent numeric overflow.
- Added migration, ledger, precision, and validation tests.
- Documented the updated cost-handling rules.

These changes prevent rounding drift and keep budget enforcement consistent with settled usage.


alembic/versions/b3f8d1c6a4e7_exact_budget_counters.py
```diff
@@ -0,0 +1,226 @@
...
+# keeps a default written against the column's old type and rewrites it as
+# ``'0'::double precision`` when the column is retyped: still correct, but a
+# migrated database would then describe the column differently from one built by
+# ``create_all``, which is the same drift the SQLite rebuild exists to avoid.
+_COUNTERS: tuple[tuple[str, str, bool, str | None], ...] = (
+    ("budgets", "max_budget", True, None),
+    ("users", "spend", False, None),
+    ("users", "reserved", False, "0"),
+    ("budget_reset_logs", "previous_spend", False, None),
+    ("scoped_budgets", "max_budget", True, None),
```
khaledosman: `scoped_budgets.max_budget` no longer exists, and neither does `workspace_budget_defaults.max_budget` on line 93: `f3a5c7e9d1b4` dropped both when it made a scoped ceiling name a budget instead of carrying its own figure. On PostgreSQL these two `op.alter_column` calls fail with `column "max_budget" does not exist`.

After the rebase the set is five columns, not eight: `users.spend` / `reserved`, `budgets.max_budget`, `budget_reset_logs.previous_spend`, `scoped_budgets.current_spend` / `reserved_spend`.

alembic/versions/b3f8d1c6a4e7_exact_budget_counters.py
```diff
@@ -0,0 +1,226 @@
...
+    ("scoped_budgets", "current_spend", False, "0"),
+    ("scoped_budgets", "reserved_spend", False, "0"),
+    ("workspace_budget_defaults", "max_budget", True, None),
+)
+
+
+def _scoped_budgets_table(numeric: bool) -> sa.Table:
```
khaledosman: This hand-written table is now wrong in both directions against `f3a5c7e9d1b4`. It declares `max_budget`, `budget_duration_sec`, `reset_alignment` and `ck_scoped_budgets_single_period_source`, all of which that revision removed, and it omits `budget_id` (NOT NULL, FK to `budgets.budget_id`) and `ix_scoped_budgets_budget_id`, which it added.

This is exactly what SQLite's rebuild copies from, so as written the rebuild would silently drop `budget_id`, its foreign key and its index.

alembic/versions/b3f8d1c6a4e7_exact_budget_counters.py
```diff
@@ -0,0 +1,226 @@
...
+    old_type: sa.types.TypeEngine[object] = _FLOAT_TYPE if to_numeric else _COST_TYPE
+    using = "numeric" if to_numeric else "double precision"
+
+    for table in ("budgets", "users", "budget_reset_logs", "scoped_budgets", "workspace_budget_defaults"):
+        columns = [entry for entry in _COUNTERS if entry[0] == table]
+        if is_sqlite:
+            # The rebuild starts from the table as it is now, so ``copy_from``
+            # describes the *old* column types.
+            copy_from = _scoped_budgets_table(not to_numeric) if table == "scoped_budgets" else None
+            with op.batch_alter_table(table, copy_from=copy_from) as batch_op:
```
khaledosman, line 143: `budgets` now carries `ck_budgets_single_period_source` (`f3a5c7e9d1b4` moved the period onto it and brought the CHECK along), and it goes through this loop with `copy_from=None`. A reflected SQLite `Table` carries no CHECK constraints, which is the precise reason `_scoped_budgets_table` is hand-written above, so the rebuild drops it.

`budgets` needs its own hand-written `sa.Table` here, with the CHECK and `reset_alignment` declared.

alembic/versions/b3f8d1c6a4e7_exact_budget_counters.py
```diff
@@ -0,0 +1,226 @@
...
+"""
+
+from collections.abc import Sequence
+
+import sqlalchemy as sa
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision: str = "b3f8d1c6a4e7"
+down_revision: str | Sequence[str] | None = "db8fbf901ee0"
```
khaledosman: `down_revision` has to move to `f3a5c7e9d1b4`. `db8fbf901ee0` is no longer the head, so a rebase that leaves this alone gives the chain two heads and `tests/unit/test_tenancy_schema_chain.py::test_the_revision_chain_has_one_head` fails.

src/gateway/services/tenancy/workspace_budget_default_service.py
```diff
@@ -90,7 +91,9 @@ class WorkspaceMemberBudgetPolicyCreate(BaseModel):
         description="Narrow the default to one provider instance; null applies to every provider",
     )
     name: str | None = Field(default=None, max_length=200, description="Admin-facing label")
-    max_budget: float | None = Field(default=None, ge=0, description="Maximum USD spend per member in the period")
+    max_budget: float | None = Field(
+        default=None, ge=0, le=MAX_USD_LIMIT, description="Maximum USD spend per member in the period"
```
khaledosman: This field has no column behind it after `f3a5c7e9d1b4`: `workspace_budget_defaults` now carries `budget_id` and no cap, name or duration of its own. Same for `WorkspaceMemberBudgetPolicyUpdate.max_budget` on line 114, the `as_float(default.max_budget)` in `from_model`, and both `to_usd_or_none(request.max_budget)` writes.

The `MAX_USD_LIMIT` bound is still worth having; it belongs on whatever writes `budgets.max_budget` now.

khaledosman (approved):
The arithmetic is careful: every narrowing point is accounted for, `ZERO`/`to_usd` is what keeps PostgreSQL from resolving a CASE arm or a `coalesce` as `double precision`, and the migration's pre-flight that names the offending row beats the bare `numeric field overflow` an `ALTER` would raise at startup.

What the findings below are all one thing: `main` moved to `f3a5c7e9d1b4` (#711, "Move the period onto `budgets`") after this branch was cut, and that revision deleted two of the eight columns this converts, added a NOT NULL `scoped_budgets.budget_id`, and moved the period CHECK onto `budgets`. The rebase is not mechanical.

alembic/versions/b3f8d1c6a4e7_exact_budget_counters.py
```diff
@@ -0,0 +1,226 @@
...
+    if bind.dialect.name != "postgresql":
+        # Nothing to guard: SQLite has no numeric storage class, so the value
+        # stays a REAL and the rebuild cannot overflow. Refusing here would be a
+        # failure this engine would not otherwise have had.
+        return
+    for table, column, _nullable, _default in _COUNTERS:
+        offending = bind.execute(
+            sa.text(f"SELECT count(*) FROM {table} WHERE {column} >= :ceiling"),  # noqa: S608 (names are literals above)
+            {"ceiling": _OVERFLOWS_AT},
+        ).scalar_one()
```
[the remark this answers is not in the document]
njbrake: Real, and fixed. `NUMERIC(18, 6)` is symmetric, so a counter at or below -1e12 overflows the same way, and a one-sided check would have let through the very case the guard exists to catch.

Guarded on magnitude rather than on two bounds: `WHERE ABS({column}) >= :ceiling`. Same coverage as the two-parameter form you proposed, and it states the symmetry instead of implying it. The refusal now reads "an amount of $1000000000000 or more in magnitude".

The negative boundary is covered by `tests/integration/test_exact_budget_columns.py::test_the_migration_names_the_row_it_cannot_convert`, which seeds a too-large cap, asserts the upgrade refuses, flips that same row to the negative bound, and asserts it refuses again. Confirmed against PostgreSQL 18: the guard raises before any DDL runs, so the schema is left untouched and the operator can correct the row and migrate again.

tests/integration/test_atomic_spend_update.py
```diff
@@ -20,15 +22,17 @@ async def test_spend_update_uses_sql_expression(async_db: AsyncSession) -> None:
...
     )
 
     await async_db.refresh(user)
     assert user is not None
 
     expected_new_spend = 5.0 + 3.5
-    assert abs(user.spend - expected_new_spend) < 0.001, f"Expected spend {expected_new_spend}, got {user.spend}"
+    assert abs(user.spend - Decimal(str(expected_new_spend))) < Decimal("0.001"), (
+        f"Expected spend {expected_new_spend}, got {user.spend}"
+    )
```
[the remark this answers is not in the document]
njbrake: Real, and fixed in both files.

`test_atomic_spend_update.py` now settles `Decimal("3.500001")` and asserts `user.spend == Decimal("8.500001")`; the accumulation test settles `Decimal("20.000001")` three times and asserts `Decimal("60.000003")`. Both tolerance comparisons are gone. A float caller still reaches `reconcile_reservation`, which widens through `to_usd`, and that path stays covered in `test_budget_race_condition.py`.

`test_budget_cost_estimate.py` compares the estimate against a `_micro_dollars(...)` helper doing the same arithmetic in `Decimal`, so the expected value stays the arithmetic under test rather than a constant somebody would have to re-derive. No `pytest.approx` is left in either file.

The same reasoning drove converting `test_budget_race_condition.py` earlier in this PR: those counters are exact now, so an approximate assertion would keep passing on one that had regressed to binary floating point.

alembic/versions/b3f8d1c6a4e7_exact_budget_counters.py
njbrake: Correct, and fixed by the rebase now pushed (`db6a049`). The roster was written against `db8fbf901ee0`, before `f3a5c7e9d1b4` moved the limit onto the budget.

`_COUNTERS` is now the six that exist: `budgets.max_budget`, `users.spend`, `users.reserved`, `budget_reset_logs.previous_spend`, `scoped_budgets.current_spend`, `scoped_budgets.reserved_spend`. That is the same list you gave; I read your "five" as a slip, since the names below it number six. Say the word if you meant to drop one.

alembic/versions/b3f8d1c6a4e7_exact_budget_counters.py
njbrake: Correct, and gone. `_scoped_budgets_table` is deleted rather than corrected: after `f3a5c7e9d1b4` that table has no CHECK, so there is nothing reflection cannot carry, and `scoped_budgets` is now rebuilt straight from reflection.

Verified on a SQLite database migrated to `f3a5c7e9d1b4` and then to head: `budget_id` keeps its column and its `REFERENCES budgets` clause, and the index set is identical across the rebuild, `ix_scoped_budgets_budget_id`, `ix_scoped_budgets_scope`, and both partial unique indexes with their `sqlite_where` intact.

alembic/versions/b3f8d1c6a4e7_exact_budget_counters.py
njbrake: Correct, and this is where the hand-written table went. `_budgets_table` now declares `reset_alignment` and `ck_budgets_single_period_source`, and `budgets` is the only table in the loop handed a `copy_from`.

`tests/unit/test_exact_budget_schema_chain.py::test_the_budget_period_check_survives_the_rebuild` asserts the CHECK is present after the upgrade and that an insert naming both a duration and an alignment is refused, and `test_the_conversion_is_reversible` asserts it survives the downgrade too.

alembic/versions/b3f8d1c6a4e7_exact_budget_counters.py
njbrake: Agreed, and done: `down_revision` is `f3a5c7e9d1b4`, which leaves `b3f8d1c6a4e7` as the single head. `tests/unit/test_tenancy_schema_chain.py` passes, 28 tests including `test_the_revision_chain_has_one_head`.

Worth recording that this contradicted the brief I was given, which said to keep `db8fbf901ee0`. That was written before #711 landed. You are right and the brief was stale.

src/gateway/services/tenancy/workspace_budget_default_service.py
njbrake: Correct, and removed. Those fields and both `to_usd_or_none` writes are gone; the only change left in this file against `main` is one line, `max_budget=as_float(budget.max_budget)` in `from_model`, which narrows the budget's now-exact cap for a wire contract that stays float.

Agreed on the bound, and that is where it went: `MAX_USD_LIMIT` is on `CreateBudgetRequest` and `UpdateBudgetRequest` in `api/routes/budgets.py`, the two bodies that write `budgets.max_budget`. It appears nowhere else.

PR #723, opened 2026-08-21
feat(usage): record the cached-token convention on usage_logs

### LINKED ISSUE #690: Persist the cached-token convention on usage_logs instead of recovering it
- Labels: `area/backend`, `otari-migration`
- Follows mozilla-ai/otari#661

## What is true today

`usage_logs` records the token counts a provider reported but not which cached-token convention they were reported under, and the two are indistinguishable from the numbers alone. Repricing therefore recovers the convention rather than reading it: `usage_admin_service._row_cache_tokens_included` compares the stored `billing_meters["total_input_tokens"]` against `prompt_tokens`, which are equal under the inclusive (OpenAI) shape and differ by the cache buckets under the additive (Anthropic) one.

That works for every row the gateway priced, because settlement writes those meters. It does not work for a row that was ingested without pricing, which has no meters to read, and there the fallback is the additive shape. An OpenAI-shaped import that arrived while its model had no rate row, and is repriced later through `POST /v1/usage/set-price`, will therefore have its cached tokens added to a prompt that already contained them. On a 1,000-token prompt with 500 cached at $3/$15/$0.3 per million that is `0.00615` where the row should settle at `0.00465`, a 32% overcharge.

## Why #661 did not address it

#661 made the recovery explicit where it had previously been a silent assumption, which closes the case that actually occurs: every gateway-priced row carries meters. Storing the convention is a column on `usage_logs`, and that PR already rewrites `usage_logs` once to retype `cost`. Adding a second change to the same table in the same migration would have widened the risky half of an already risky change for a case that only reaches unpriced imported rows.

## What it would take

- A nullable `cache_tokens_in_prompt` boolean on `usage_logs`, written by the external-usage ingest from the value the submitter already sends, and by settlement from `GatewayUsage.cache_tokens_in_prompt`.
- `_row_cache_tokens_included` reads the column when it is set and keeps the meter recovery as the fallback, so historical rows keep working unchanged.
- No backfill: the meters already answer for every priced row, and a row with neither meters nor a flag is one nothing can answer for.
- Nullable rather than defaulted, because "not recorded" and "inclusive" are different answers and the fallback needs to tell them apart.

## Definition of Done

- An imported row reprices under the convention it was submitted with, whether or not it was priced at ingest.
- The meter-based recovery still covers rows written before the column existed.
- A test covers the case this issue exists for: an OpenAI-shaped import, unpriced at ingest, repriced afterwards.

njbrake:
## Description

`usage_logs` recorded the token counts a provider reported but not which cached-token convention they arrived under, and the two shapes are indistinguishable from the numbers alone. Repricing recovered it from `billing_meters`, which only answers for a row something already priced: an OpenAI-shaped import that landed while its model had no rate row, then repriced through `POST /v1/usage/set-price`, had its cached tokens added to a prompt that already contained them (`0.00615` where the row should settle at `0.00465`).

A nullable `cache_tokens_in_prompt` on `usage_logs` now records it, written by the external-usage ingest from the value the submitter sends and by settlement from `GatewayUsage.cache_tokens_in_prompt`. Repricing reads the column when set and keeps the meter recovery for rows that predate it. Nullable with no default and no backfill on purpose: "not recorded" and "inclusive" are different answers, and a default would make every historical row claim a convention nothing checked.

The hosted platform has the same ingest contract but no such column, because it prices at ingest and has no repricing surface. The persistence is new here rather than lifted.

Two money-column chain tests pin a database to an older revision and loaded the whole `UsageLog` entity, so they broke on the new column; each now selects the one column it asserts on. The PostgreSQL one also migrates back to head when it is done, since the integration fixtures share one database.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes #690

## Checklist

Full unit suite (2141 passed) and full integration suite (1488 passed) run against a local PostgreSQL, plus the OSS-edition smoke gate since migrations changed. `make openapi-check` and `make postman-check` both report no drift: the column is billing-internal and is not on any response schema.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

Implemented from the issue by Claude Opus 5 in discussion with @njbrake, then reviewed by a second agent reading the diff cold. That review caught the PostgreSQL half of the chain-test breakage and the missing migrate-back-to-head, both fixed before this PR was opened.


alembic/versions/b3d5f7a9c1e6_add_usage_logs_cache_tokens_in_prompt.py
```diff
@@ -0,0 +1,43 @@
...
+"""
+
+from collections.abc import Sequence
+
+import sqlalchemy as sa
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision: str = "b3d5f7a9c1e6"
+down_revision: str | Sequence[str] | None = "b3f8d1c6a4e7"
```
khaledosman, line 31: `b3f8d1c6a4e7` is #722's revision and is on neither this branch nor `main`, so the chain has nothing to walk from. `tests/unit/test_tenancy_schema_chain.py::test_the_revision_chain_has_one_head` is failing on this PR, and `test`, `oss-edition` and `serving` are all red for the same reason (every test that boots the app or migrates a database errors).

Either chain onto `main`'s head (`f3a5c7e9d1b4`), or state the merge order in the description (#722 → #723 → #724 → #725) and rebase each branch onto the one before it. #724 chains onto this revision and #725 onto #724, so all three are in the same state.

tests/integration/test_exact_money_columns.py
```diff
@@ -277,10 +277,21 @@ def test_the_migration_round_trips_on_postgresql_with_rows_in_the_table(
...
-    assert settled.cost == Decimal("0.123457")
+    assert settled_cost == Decimal("0.123457")
+
+    # Put the database back at head. The integration fixtures migrate once per
+    # session and share the database, so leaving it a revision behind would hand
+    # every test after this one a schema its models no longer match. Released
+    # first for the same reason the downgrade above is: the reads reopened a
+    # transaction, and an idle one blocks the ALTER.
+    test_db.close()
+    command.upgrade(config, "head")
```
khaledosman: The restore to head is not failure-safe. If any assertion between the `command.downgrade` on line 276 and here raises, the session-shared PostgreSQL is left a revision behind and every integration test after this one fails on a missing column, which turns one real failure into a suite-wide cascade that hides it.

Put the downgrade/upgrade pair in `try`/`finally`, or move the revision pinning into a fixture that restores on teardown.

khaledosman (approved):
The nullable-with-no-backfill reasoning holds: "not recorded" and "inclusive" really are different answers, and reading the column before the meters cannot disagree with itself, because ingest prices with the same `event.cache_tokens_in_prompt` it now stores. Test coverage pins both conventions from one call site, which is the right shape for a fact that is invisible in the numbers.

Two things below: the migration chains onto an unmerged sibling, and one test leaves a shared database behind head on any failure.

alembic/versions/b3d5f7a9c1e6_add_usage_logs_cache_tokens_in_prompt.py
njbrake: Correct when you wrote it, and now resolved by the merge rather than by an edit: #722 landed as d2ba90e, so `b3f8d1c6a4e7` is on main and is main's single head. `f3a5c7e9d1b4` is no longer the head; it is that revision's parent, because #722 rechained under #711's two budget migrations. Chaining onto it now would recreate the second head you are describing, so `down_revision` stays as it is.

The branch is rebased onto d2ba90e. `alembic heads` reports one head, `b3d5f7a9c1e6`, and a fresh PostgreSQL database round-trips upgrade, downgrade -1, upgrade with the column appearing, disappearing and returning. `test_migration_single_head.py`, `test_tenancy_schema_chain.py`, `test_exact_money_schema_chain.py` and `test_exact_money_columns.py` pass, along with the full unit and integration suites.

The merge order is as you set out, #722 then #723 then #724 then #725, each rebased onto the one before.

tests/integration/test_exact_money_columns.py
njbrake: Agreed, and fixed. The downgrade, the re-upgrade and the assertions are now inside a `try`, with the close and `command.upgrade(config, "head")` in a `finally`, so the restore runs on the failure path too. I kept it in the test rather than a fixture because this is the only test that pins a revision on the shared database, and a fixture would put the restore a file away from the downgrade that makes it necessary.

Verified by injecting a deliberate failure between the downgrade and the restore: before the change the injected failure was followed by a cascade of missing-column errors, after it the injected failure is the only one and the 33 tests that run next, including every `test_usage_admin.py` case that reads the new column, still pass. The injection is not in the diff.

Issue #657, opened 2026-08-19
Per-workspace code execution policy


njbrake:
- Labels: `otari-migration`, `open-core`
- Milestone: `otari-migration-m5`
- Part of mozilla-ai/otari-ai#1452, config planes track
- Depends on #655

## Why this exists

Code execution here is a sandbox backend the tool loop dispatches to, configured deployment-wide through the tool settings surface. There is no per-workspace policy: every workspace on a deployment gets whatever the operator configured.

The hosted platform carries a per-workspace policy (which tools are permitted, and the limits around them), read and upserted behind an organization role gate and a capability gate.

## What to produce

Per-workspace code execution policy above the deployment-wide sandbox configuration, following #655. The sandbox wire contract is unchanged: this is about who may ask for what, not about how execution happens.

## Definition of Done

- A workspace's policy is readable and editable by an appropriately-roled member, and the request path enforces it.
- The sandbox contract and its conformance check are untouched.
- A deployment that sets no per-workspace policy behaves as it does today.

## Not in scope

The hosted resolve endpoint that hands a gateway its code execution configuration is hosted protocol and stays where it is.

_Note: this issue was drafted by Claude Opus 5 via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

Issue #656, opened 2026-08-19
Per-workspace web search configuration


njbrake:
- Labels: `otari-migration`, `open-core`
- Milestone: `otari-migration-m5`
- Part of mozilla-ai/otari-ai#1452, config planes track
- Depends on #655

## Why this exists

Search tools are configured deployment-wide here: `api/routes/search_tools.py` is gated on the master key throughout, and `search_tool_store_service` merges stored rows over `config.yml` into a process-wide overlay. One operator, one deployment, one set of tools.

The hosted platform has a per-workspace layer above that: a self-serve toggle gated on the organization owner or admin role for the workspace's organization. It stores **no secret**, because the credential is held by the gateway-side adapter, which is the shape that survives the rehome.

## What to produce

Per-workspace web search configuration above the existing deployment-wide tools, following the resolution and composition answers from #655. The credential stays where it is; what is new is which workspaces may use search and with what settings.

## Definition of Done

- A workspace with the right role can turn search on or off for itself, and the request path honours it.
- No credential moves into the per-workspace layer.
- A deployment that configures nothing per-workspace behaves as it does today.

## Not in scope

The platform-hosted search backend that this repository's client calls over HTTP is a wire contract on the hosted side and does not move.

_Note: this issue was drafted by Claude Opus 5 via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

Issue #654, opened 2026-08-19
Organization and workspace guardrail configuration


njbrake:
- Labels: `otari-migration`, `open-core`
- Milestone: `otari-migration-m5`
- Part of mozilla-ai/otari-ai#1452, config planes track

## Why this exists

Otari has guardrail execution and a deployment-wide operator setting. It has no per-tenant guardrail configuration at all.

Execution is `run_input_guardrails` in `services/guardrails.py`, a flat pre-provider interceptor that all three route handlers call right after auth, talking to `any-guardrail` over HTTP with an SSRF check on caller-supplied URLs. Configuration is `api/routes/tool_settings.py` over the `runtime_settings` key/value table, gated on the master key, holding `guardrails_url` and its neighbours and surfaced by `ToolsGuardrailsPage.tsx`.

That is one operator configuring one deployment, which is the right shape for a single-tenant gateway and the wrong shape once an organization has workspaces. What is missing is the layer above: which guardrails an organization has configured, with which credentials, and which of its workspaces may use them.

## What to produce

A per-organization guardrail configuration plane, sitting **above** the deployment setting rather than replacing it. `guardrails_url` stays a deployment concern in `runtime_settings`; a deployment that configures no organization guardrails keeps behaving as it does today.

The plane covers guardrail entries per organization, each with its own credential, an enabled state, and the set of workspaces that may use it, including what a newly created workspace inherits.

## A design question, now shared rather than owned here

Per-workspace opt-in on top of organization defaults has never been built on either side. The hosted platform's plane is organization-level only, and the workspace-level table it once carried was dead code that has since been removed.

This turned out to be the same question all four config planes hit, so it moved to #655 and is answered once for guardrails, web search, code execution and MCP together. What follows is the guardrail-specific form of it, as input to that decision rather than a separate answer.

So "override" has to be specified before it is implemented. Can a workspace enable a guardrail its organization has not configured? Disable one the organization requires? Or only narrow the parameters of one it has been granted? Answer this in the issue before writing the schema, because the answer decides whether the workspace row is a grant, a veto, or a refinement.

## What does not come with it

- **Enforcement.** Otari already enforces, at the interceptor above. The hosted platform's enforcement layer is reachable only through its own completion route and retires with it.
- **Execution.** Otari already calls `any-guardrail`.
[...]

PR #724, opened 2026-08-21
feat(mcp): let a workspace configure its own MCP servers

### LINKED ISSUE #658: Workspace MCP server management
- Depends on #655, and least of the four

## Why this exists

This repository's MCP support is entirely client-side: `mcp_client` and the tool loop connect to servers a request names. There is no management surface, no route, and no stored server configuration.

The hosted platform has workspace-scoped CRUD over configured MCP servers, encrypting auth tokens on save, decrypting on resolve, and never returning them in public responses.

## What to produce

Workspace-scoped MCP server management: create, list, update and delete, with tokens encrypted at rest and absent from responses, plus URL validation and the role gate.

## Why #655 matters less here

MCP is the one of the four with **no deployment-wide equivalent to compose over**, so the composition half of #655 may not apply at all. The resolution half still does: the tool loop reaches these servers on the request path.

## Definition of Done

- A workspace configures an MCP server with an auth token; the token is encrypted at rest and never returned.
- The tool loop reaches a workspace's configured servers.
- Duplicate server names within a workspace are rejected rather than silently collapsing.

njbrake:
## Description

Standalone deployments can now configure MCP servers per workspace. `mcp_server_ids` was hybrid-only and returned a 400 outside it; it now resolves in both modes, through the platform in hybrid and against a new `workspace_mcp_servers` table in standalone. Tokens are Fernet-encrypted at rest and absent from every response (`has_token` instead), names are unique per workspace at the database, and URLs face the same SSRF and TLS checks a request-body server does.

Ported from otari-ai's `MCPServerService` rather than designed fresh. Three deliberate divergences: this repository's own workspace authorization (a workspace owner/admin qualifies, not only an organization owner/admin), one async URL check in the service instead of the platform's split between a sync ingress validator and a service-layer SSRF pass, and no analytics or capability-status field. Its stricter read gate is kept: every operation including the list needs management access, since these rows name the endpoints the gateway connects to and authenticates against.

Resolution follows #678 exactly: at admission in `prepare_gateway_tools`, off `RequestContext.workspace_id`, against the request's live session, landing on `ToolContext`. No overlay, no cache. MCP is the exception that decision names, having no deployment-wide list to narrow.

**Two contract changes**, both unreachable while the field was hybrid-only. An unknown id is now 404 in both modes rather than a standalone 400, and `mcp_server_ids` gains a ceiling of 50, the most servers a workspace can hold on either side.

**Not included:** a dashboard page. The issue carries `area/dashboard`, but its Definition of Done is entirely backend and the generated client already has the types. Happy to follow up if you want the UI in this milestone.

Worth noting before merge: your own comment on #658 asks whether any otari.ai users actually use the MCP server feature, and floats deferring it out of m5 if not. That question is still open.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes #658.

## Checklist

- [x] I have added or updated tests that cover my change (`tests/unit`, `tests/integration`). 29 new integration tests covering CRUD, encryption at rest, the token's three update states, URL validation, the role gate, resolution, and the zero-rows case #678 asks each of the four surfaces for.
- [x] I ran the Definition of Done checks locally (`make lint`, `make typecheck`, `make test`). All pass. One caveat on the integration suite: the sandbox has no Docker, so I ran it against a local PostgreSQL 18 via `TEST_DATABASE_URL`. The first full run was clean at 1509 passed. A second run after review fixes took 44 minutes instead of 12 under load and reported one failure in `test_agent_telemetry_read.py` plus two socket-timeout errors in `test_otlp_metrics.py`; all three pass in isolation and none touch this change. Also ran the OSS-edition smoke gate, and the migration up, down and up again on PostgreSQL with `alembic check` showing no drift.
- [x] Documentation was updated where necessary. `docs/mcp.md` and `src/gateway/AGENTS.md`.
- [x] If the API contract changed, I regenerated the OpenAPI spec. Postman collection, `sdk-endpoints.txt` and `web/src/client/schema.ts` regenerated too.

## AI Usage

- [ ] No AI was used.

Grounded in the otari-ai implementation this issue describes, read directly rather than reconstructed from the issue text, and in #678's recorded decision. A fresh reviewer agent read the diff cold and found a real bug: an explicit `null` on a required field reached a NOT NULL column and came back as a 409 naming a name conflict that did not exist. Fixed, along with two unbounded-input findings. Prose is the model's; the decisions are Nathan's.

https://claude.ai/code/session_01UYFMEkk1ppKiP6CXuDeBPZ

_Note: this PR description was drafted by Claude Opus 5 via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

## Summary

- Added workspace-scoped MCP server management with create, list, update, and delete operations.
- Added encrypted token storage. API responses expose only `has_token`.
- Added workspace authorization, unique server names, URL safety checks, and server limits.
- Enabled MCP server resolution in standalone and hybrid deployments.
- Added disabled-server handling, request limits, documentation, API artifacts, migration, and integration tests.
- No dashboard UI is included.

## Technical notes

- `mcp_server_ids` supports up to 50 IDs.
- Unknown server IDs return `404`.
- Tokens use Fernet encryption at rest.
- Workspace deletion cascades to its MCP server configuration.

### DIFF
--- a/alembic/versions/c5e9a1d3f7b2_add_workspace_mcp_servers.py
+++ b/alembic/versions/c5e9a1d3f7b2_add_workspace_mcp_servers.py
@@ -1,0 +1,61 @@
+"""Add the workspace-scoped MCP server table.
+
+One row is one MCP server a workspace has configured: a URL, an optional
+bearer token encrypted with ``OTARI_SECRET_KEY``, and the purpose hint and
+tool allow-list the tool loop already understands from an inline
+``mcp_servers`` entry. A request references them by id with
+``mcp_server_ids``, which until now was hybrid-only.
+
+``(workspace_id, name)`` is unique: the name is how an operator and the tool
+loop both identify a server, so two rows sharing one within a workspace would
+silently hide a server rather than fail.
+
+``workspace_id`` cascades, matching ``workspace_budget_defaults`` and the
+provider-key tables: this is a workspace-owned configuration row, not durable
+request-plane history like ``usage_logs`` or ``api_keys``.
+
+Revision ID: c5e9a1d3f7b2
+Revises: b3d5f7a9c1e6
+Create Date: 2026-08-21
+"""
+
+from collections.abc import Sequence
+
+import sqlalchemy as sa
+from alembic import op
+
+revision: str = "c5e9a1d3f7b2"
+down_revision: str | Sequence[str] | None = "b3d5f7a9c1e6"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+_UNIQUE_NAME = "uq_workspace_mcp_servers_workspace_name"
+
+
+def upgrade() -> None:
+    op.create_table(
+        "workspace_mcp_servers",
+        sa.Column("id", sa.Uuid(), nullable=False),
+        sa.Column("workspace_id", sa.Uuid(), nullable=False),
+        sa.Column("name", sa.String(), nullable=False),
+        sa.Column("url", sa.String(), nullable=False),
+        sa.Column("encrypted_token", sa.Text(), nullable=True),
+        sa.Column("purpose_hint", sa.Text(), nullable=True),
+        sa.Column("allowed_tools", sa.JSON(), nullable=True),
+        sa.Column("enabled", sa.Boolean(), nullable=False),
+        sa.Column("created_at", sa.DateTime(timezone=True), nullable=False),
+        sa.Column("updated_at", sa.DateTime(timezone=True), nullable=False),
+        sa.PrimaryKeyConstraint("id"),
+        sa.ForeignKeyConstraint(["workspace_id"], ["workspace.id"], ondelete="CASCADE"),
+        sa.UniqueConstraint("workspace_id", "name", name=_UNIQUE_NAME),
+    )
+    op.create_index(
+        op.f("ix_workspace_mcp_servers_workspace_id"),
+        "workspace_mcp_servers",
+        ["workspace_id"],
+    )
+
+
+def downgrade() -> None:
+    op.drop_index(op.f("ix_workspace_mcp_servers_workspace_id"), table_name="workspace_mcp_servers")
+    op.drop_table("workspace_mcp_servers")
--- a/src/gateway/AGENTS.md
+++ b/src/gateway/AGENTS.md
@@ -97,8 +97,17 @@ workspace row is a veto and a refinement and never a grant, and no row means no 
 (MCP, which has no deployment-level server list to narrow, is the stated exception). The
 question is [#655](https://github.com/mozilla-ai/otari/issues/655) and the reasoning is in
 the PR that answered it,
-[#678](https://github.com/mozilla-ai/otari/pull/678). Read both before building any of
-#654, #656, #657 or #658, each of which settles its own surface under the rule above.
+[#678](https://github.com/mozilla-ai/otari/pull/678). Read both before building
+any of #654, #656 or #657, each of which settles its own surface under the rule above.
+
+MCP's own surface is the one that has landed (#658). `workspace_mcp_servers`
+(`models/entities.py`) holds a workspace's configured servers, with the bearer token
+encrypted by the same `secret_box` the two stores above use; it is managed over
+`/v1/workspaces/{workspace_id}/mcp-servers` through
+`services/tenancy/workspace_mcp_server_service.py`, and read on the request path by
+`_pipeline._resolve_mcp_server_ids`, the standalone half of an `mcp_server_ids` field the
+platform answers in hybrid mode. There is no overlay and no cache: it resolves per
+request against `ctx.db`, which is what the decision above says the seam should do.
 
 ## The first-request setup guide
 `services/tenancy/workspace_activation_service.py` is the state behind the dashboard's setup guide (`routes/workspace_activation.py`, `/v1/workspaces/{id}/activation`): whether to offer it, the API key it hands out, what the workspace's traffic says about the attempt, and the dismissal that retires it. Ported from the platform's `WorkspaceActivationService`, and the one departure to know is that **activation is derived, not recorded**: the first successful `source="gateway"` row in `usage_logs` for the workspace *is* the evidence, read through `ix_usage_logs_workspace_source_status_timestamp`. The platform stores that telemetry in columns because its usage pipeline is asynchronous and crosses services; here the row is written by this process into this database, so a second copy could only drift. `workspace_activation_state` therefore holds one row per workspace carrying what cannot be observed elsewhere (the dismissal, when a key was last issued, and which key it was).
--- a/src/gateway/api/routes/_pipeline.py
+++ b/src/gateway/api/routes/_pipeline.py
@@ -33,6 +33,7 @@ Settlement invariants owned here:
 from __future__ import annotations
 
 import asyncio
+import contextlib
 import time
 import uuid
 from collections.abc import AsyncIterator, Awaitable, Callable, Coroutine
@@ -156,7 +157,10 @@ from gateway.services.sandbox_backend import (
     SandboxNotReachableError,
 )
 from gateway.services.scoped_budget_service import BudgetScopeRequest
+from gateway.services.secret_box import SecretBoxUnavailableError, SecretDecryptionError
+from gateway.services.tenancy.errors import WorkspaceMcpServerNotFoundError
 from gateway.services.tenancy.org_provider_key_service import cached_org_model_restriction
+from gateway.services.tenancy.workspace_mcp_server_service import resolve_workspace_mcp_servers
 from gateway.services.tool_usage import (
     MAX_TOOL_NAMES,
     OVERFLOW_TOOL_NAME,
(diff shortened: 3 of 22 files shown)


alembic/versions/c5e9a1d3f7b2_add_workspace_mcp_servers.py
```diff
@@ -0,0 +1,61 @@
...
+Create Date: 2026-08-21
+"""
+
+from collections.abc import Sequence
+
+import sqlalchemy as sa
+from alembic import op
+
+revision: str = "c5e9a1d3f7b2"
+down_revision: str | Sequence[str] | None = "b3d5f7a9c1e6"
```
khaledosman, line 28: `b3d5f7a9c1e6` is #723's revision and is on neither this branch nor `main`, so the chain is unresolvable as pushed: `test`, `oss-edition`, `serving`, `build` and `e2e` are all red, and `test_the_revision_chain_has_one_head` is the test that names it.

Same fix as on #723: either chain onto `main`'s head, or declare the merge order (#722 → #723 → #724 → #725) and rebase each onto the one before.

src/gateway/models/mcp.py
```diff
@@ -4,6 +4,18 @@
...
+# configured servers, in standalone (``MAX_MCP_SERVERS_PER_WORKSPACE`` in
+# ``services/tenancy/workspace_mcp_server_service.py``) and on the platform,
+# which uses the same value, so no reachable request loses anything. It is
+# declared separately from that constant rather than imported: a wire bound has
+# to be enforceable at parse time, with no database and no mode to consult, and
+# a models module reaching into a service to learn its own limits would be the
+# wrong direction. Without it a single request could hand the resolver an
+# unbounded ``IN`` list, which is why the usage filters carry the same kind of
+# ceiling (``core/sql.MAX_FILTER_VALUES``).
+MAX_MCP_SERVER_IDS = 50
```
khaledosman, line 17: This bound justifies itself by being equal to `MAX_MCP_SERVERS_PER_WORKSPACE`, and nothing pins that. `tests/integration/test_workspace_mcp_servers.py` imports both constants but never asserts they agree, so raising the service cap to 100 would leave a workspace able to store 100 servers and a request able to name only 50, surfacing as a 422 with nothing pointing at the cause.

Add `assert MAX_MCP_SERVER_IDS >= MAX_MCP_SERVERS_PER_WORKSPACE` to a unit test, so the comment's claim is checked rather than stated.

src/gateway/api/routes/_pipeline.py
```diff
@@ -1617,6 +1621,47 @@ def merge_policy_guardrails(
...
+        return await _resolve_platform_mcp_servers(
+            config=ctx.config,
+            user_token=ctx.user_token,
+            mcp_server_ids=mcp_server_ids,
+        )
+
+    if ctx.db is None or ctx.workspace_id is None:
+        raise adapter.error(400, MCP_SERVER_IDS_UNAVAILABLE_DETAIL, ErrorKind.INVALID_REQUEST)
+    try:
+        return await resolve_workspace_mcp_servers(ctx.db, workspace_id=ctx.workspace_id, server_ids=mcp_server_ids)
```
khaledosman, line 1663: `resolve_workspace_mcp_servers` can raise `SQLAlchemyError`, which is not an `HTTPException`, so it escapes `prepare_gateway_tools`' closing `except HTTPException: await release_reservation(ctx)` and leaks the hold on `users.reserved`.

The gap pre-exists this PR (`_require_tool_pricing` reads the database inside the same block, and #725's description names it as a property of the whole block), but this is the change that puts a second database read in there. Either widen that handler to `except (HTTPException, SQLAlchemyError)` here, or file the follow-up from this PR rather than the next one.

khaledosman (approved):
Faithful port, and the parts that are easy to get wrong are right: the token's three PATCH states, the unique index as arbiter rather than a preceding existence check, reading `server.name` before the flush so the conflict handler does not lazy-load into `MissingGreenlet`, standalone-only router placement, the resolve taking `workspace_id` off the key rather than a header, and `asyncio.gather` over the per-server URL checks so 50 servers is not 50 serial DNS lookups.

alembic/versions/c5e9a1d3f7b2_add_workspace_mcp_servers.py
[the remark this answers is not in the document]
njbrake: Stale as of the rebase, and correct when you wrote it. This was reviewed against 053ec95a, when `b3d5f7a9c1e6` lived only on #723's branch. #723 has since merged (d80beabc) and that revision is now main's single head, so the chain resolves: `alembic heads` reports one head, `c5e9a1d3f7b2`, and upgrade, `downgrade -1`, upgrade round-trips against PostgreSQL 18. Chain is `db8fbf901ee0 -> e2f4a6c8b0d3 -> f3a5c7e9d1b4 -> b3f8d1c6a4e7 -> b3d5f7a9c1e6 -> c5e9a1d3f7b2`.

Pointing at main's head instead would have been wrong here: four PRs are landing as a serial train, and chaining each onto main would give four heads rather than one.

njbrake: Right at the time, and fixed by the rebase rather than by changing the target. You named the two options; the second is what is happening. The merge order is #722 -> #723 -> #724 -> #725, each rebased onto the one before, so chaining onto main's head would produce four heads instead of one.

#723 merged as d80beabc, so `b3d5f7a9c1e6` is now main's single head and this resolves for real. `alembic heads` gives one head, `c5e9a1d3f7b2`; `test_the_revision_chain_has_one_head` passes, as do the other five schema-chain files (54 tests); and upgrade, `downgrade -1`, upgrade round-trips on PostgreSQL. The five red jobs you listed were all the same `KeyError: b3d5f7a9c1e6`.

src/gateway/services/tenancy/workspace_mcp_server_service.py
```diff
@@ -0,0 +1,472 @@
...
+                select(func.count())
+                .select_from(WorkspaceMcpServer)
+                .where(WorkspaceMcpServer.workspace_id == resolved_workspace_id)
+            )
+        ).scalar_one()
+        # Check-then-act, deliberately, unlike the name conflict below: "at most
+        # N rows for this workspace" is not a constraint either engine can hold,
+        # so two creates racing at N-1 can both land and leave N+1. This is a
+        # management-gated surface off the request path, and the cap is a
+        # resource bound rather than a correctness invariant, so one row over is
```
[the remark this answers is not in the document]
njbrake: Fixed in f5f5e72. `create_server` now takes the workspace row lock before counting.

I had left this racy on purpose, with a comment saying one row over a resource bound was not worth serializing every create for. That reasoning was weaker than I thought: `WorkspaceRepository.lock` already exists for exactly this read-decide-write shape and is used by `workspace_budget_default_service` and `org_provider_key_service`, so the fix is one line of established house pattern rather than new machinery. The comment now explains the lock instead of excusing its absence.

Tested by asserting the lock is taken, not by racing two sessions, since a working lock is precisely what makes the race unobservable.

src/gateway/services/tenancy/workspace_mcp_server_service.py
```diff
@@ -0,0 +1,472 @@
...
+    write-only field needs: omit it to leave the stored token alone, send
+    ``""`` to clear it, send a value to rotate it. An explicit ``null`` also
+    leaves it alone, matching the platform: a client that serializes its whole
+    form back, with an empty token box it never filled in, must not destroy a
+    token it was never shown.
+    """
+
+    name: str | None = Field(default=None, min_length=1, max_length=128)
+    url: str | None = Field(default=None, min_length=1, max_length=2048)
+    authorization_token: str | None = Field(default=None, max_length=8192)
```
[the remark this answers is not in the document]
njbrake: Fixed in f5f5e72, and thank you: the contract really did advertise a payload the server refuses.

`name`, `url` and `enabled` back NOT NULL columns and are omittable but not nullable, so they now carry `SkipJsonSchema[None]`. The spec and the generated client say `name?: string`, `url?: string`, `enabled?: boolean`; `purpose_hint`, `allowed_tools` and `authorization_token` stay nullable because those genuinely can be cleared. Runtime behavior is unchanged: `None` is still the "not sent" marker and an explicit null is still refused.

Artifacts regenerated, and a test now asserts the schema shape so the two cannot drift apart again.

src/gateway/models/mcp.py
njbrake: Fixed in f5f5e72. You were right that nothing checked the claim: the integration file imported both constants and asserted neither against the other.

`test_the_request_bound_admits_every_server_a_workspace_can_hold` in `tests/unit/test_mcp_server_id_resolve.py` now asserts `MAX_MCP_SERVER_IDS >= MAX_MCP_SERVERS_PER_WORKSPACE`. Put in a unit test rather than beside the others, since it needs no database, and the file it landed in is where the rest of the id-resolution behavior is pinned.

src/gateway/api/routes/_pipeline.py
njbrake: Fixed here rather than deferred, in f5f5e72. I checked #725 before deciding: its description lists this among "two things left undone, each worth its own follow-up", so nobody was going to pick it up and fixing it here conflicts with nothing.

The handler gains an `except SQLAlchemyError` arm. It rolls the session back before releasing, because a failed statement leaves the session unusable and `release_reservation` writes, so releasing first raises `PendingRollbackError` and the hold survives anyway. Both calls are best-effort so a database still refusing work re-raises the original failure rather than a confusing second one. That covers `_require_tool_pricing` as well as the new read, which is the right scope given the gap is a property of the block.

Two unit tests cover it: the hold is released and the session rolled back, and a release that also fails still surfaces the first error.

src/gateway/AGENTS.md
```diff
@@ -98,7 +98,16 @@ workspace row is a veto and a refinement and never a grant, and no row means no
 question is [#655](https://github.com/mozilla-ai/otari/issues/655) and the reasoning is in
 the PR that answered it,
 [#678](https://github.com/mozilla-ai/otari/pull/678). Read both before building any of
-#654, #656, #657 or #658, each of which settles its own surface under the rule above.
+#654, #656 or #657, each of which settles its own surface under the rule above.
```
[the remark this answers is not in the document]
njbrake: Reflowed in cff4a9a8, prose byte-identical, no backticks added so the reference style stays consistent with the rest of the paragraph.

Two notes on the finding itself. The `#` at the start of that line predates this PR: main already had `#654, #656, #657 or #658, ...` opening the line, and my edit only shortened the list. And there is no markdownlint job in this repo, so nothing was failing on it. Fixed anyway, since the line is one I am already touching and the reflow costs nothing.

Issue #655, opened 2026-08-19
Decide: how per-workspace tool configuration composes over deployment defaults


njbrake:
- Labels: `otari-migration`, `open-core`, `needs-decision`
- Milestone: `otari-migration-m5`
- Part of mozilla-ai/otari-ai#1452, config planes track
- **Gates #654, and the web search, code execution and MCP configuration issues.**

## Why this exists

Four tool configuration surfaces are heading into this repository (guardrails, web search, code execution, MCP servers) and they all hit the same wall. Filing the shared question once beats four contributors inventing four answers.

Every tool here is configured the same way today: deployment-wide, gated on the master key, and merged over `config.yml` into a **process-wide in-memory overlay** so the synchronous dispatch path can read it with no database session. `search_tool_store_service` says this out loud, and it is "deliberately the same shape" as `provider_store_service`. `tool_settings.py` over `runtime_settings` is the same idea for scalar settings.

That design cannot serve per-workspace configuration. A process-wide overlay has exactly one answer, and the request path needs the answer for *this* request's workspace.

## The question

**How does per-workspace tool configuration reach the synchronous dispatch path, and what does it mean when it disagrees with the deployment default?**

Two parts, and both need an answer before any of the four are built:

**Resolution.** Does the overlay become workspace-keyed and stay in memory, does the dispatch path gain a session, or does resolved configuration ride on the request context established earlier at auth time? The third looks cheapest from here, because auth already resolves the caller, but it is not obviously right and it is not free.

**Composition.** Is a workspace row a grant (it may use what the deployment offers), a veto (it may disable what the deployment offers), or a refinement (it may narrow parameters only)? The answer is not required to be identical for all four, but where they differ that has to be deliberate. MCP servers are pure workspace CRUD with no deployment-level equivalent, so it may not even apply there.

## Definition of Done

- Both questions answered and written down where the four implementation issues can cite them.
- A deployment that configures nothing per-workspace behaves as it does today, and that is stated as a requirement rather than assumed.
- #654 updated to follow the answer rather than carrying its own.

_Note: this issue was drafted by Claude Opus 5 via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

PR #725, opened 2026-08-21
feat(gateway): give a workspace a code-execution policy over the deployment sandbox

### LINKED ISSUE #657: Per-workspace code execution policy

njbrake:
## Description

Code execution was configured deployment-wide and nowhere else, so every workspace on a deployment got whatever the operator had configured. This adds a per-workspace policy above that, following #655's answer as recorded in #678.

A workspace has a policy row or none. The row can refuse `otari_code_execution` for that workspace (403 at admission), lower the tool-loop and per-execution ceilings, and supply the purpose hint a request did not give. It carries no URL and no credential, so the sandbox stays deployment-scoped. No row means no narrowing, which is what keeps a deployment that configures nothing behaving as it did; that is asserted at both layers rather than assumed.

Reading and writing the policy both take an organization owner/admin, or an owner/admin of the workspace. Enforcement sits in `prepare_gateway_tools`, beside the hybrid resolve it mirrors, reading the workspace off the key that authenticated the request rather than a header. The dashboard gets a card under the existing sandbox settings on `/tools/code-execution`.

Ported from otari-ai's `CodeExecutionConfig`, with four deliberate divergences: `enabled` is required rather than defaulted (absence of a row means unnarrowed here and disabled there, so either default would surprise somebody); `tools` and `image` are dropped (this backend runs one tool and picks no image, so both would be inert); reads are role-gated, as the hosted service does; and a limit above the deployment's own ceiling is refused at the write instead of clamped at resolve time.

Two things left undone, each worth its own follow-up. `GET /v1/tools` still reports availability deployment-wide, so a blocked workspace is told the tool is available and then gets a 403; hybrid mode has the identical gap, so that fix should cover both modes at once. And a database error inside `prepare_gateway_tools` leaks the budget reservation, which is a property of the whole block (`_require_tool_pricing` reads the database there too) rather than of this new read.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes #657. Follows the decision in #655, recorded in #678.

## Checklist

- [x] I have added or updated tests that cover my change (`tests/unit`, `tests/integration`). Two new integration files: the management surface (CRUD, the role gates, the zero-rows case, the insert race) and the request path (the veto, both ceilings, the hint precedence, streaming, a master-key request). Plus a dashboard test file for the card.
- [x] I ran the Definition of Done checks locally (`make lint`, `make typecheck`, `make test`). Full backend suite: 3643 passed, 17 skipped. Dashboard: lint, typecheck and 1149 tests pass. Also `make openapi-check`, `make postman-check` and the OSS-edition smoke gate. Integration tests ran against a local PostgreSQL 18 rather than Testcontainers, since the sandbox had no Docker. Not run: the Playwright e2e and screenshot suites.
- [x] Documentation was updated where necessary. `docs/tools.md`, plus the seam note in `src/gateway/AGENTS.md` and the page inventory in `web/AGENTS.md`.
- [x] If the API contract changed, I regenerated the OpenAPI spec. Both generated artifacts are committed (`openapi.json` and the Postman collection), along with `web/src/client/schema.ts` and the SDK endpoint manifest.

## AI Usage

- [ ] No AI was used.

The hosted implementation in otari-ai was read before anything was designed here, and this is a port of it rather than a second design; the divergences above are the places where the standalone shape genuinely differs, and each is commented at its site. The finished branch was then reviewed by a separate agent with no memory of writing it, which found the upsert race and the stale-read bug behind it.


alembic/versions/c4d7f2a9b6e1_add_workspace_code_execution_policies.py
```diff
@@ -0,0 +1,59 @@
...
+Create Date: 2026-08-21
+"""
+
+from collections.abc import Sequence
+
+import sqlalchemy as sa
+from alembic import op
+
+revision: str = "c4d7f2a9b6e1"
+down_revision: str | Sequence[str] | None = "c5e9a1d3f7b2"
```
khaledosman, line 26: `c5e9a1d3f7b2` is #724's revision and is on neither this branch nor `main`, so the chain is unresolvable as pushed and CI is red for it. Same fix as on #723 and #724: chain onto `main`'s head, or declare the merge order (#722 → #723 → #724 → #725) and rebase each onto the one before.

src/gateway/api/routes/_pipeline.py
```diff
@@ -1714,6 +1724,31 @@ async def prepare_gateway_tools(
             # `bool` is an `int` subclass — exclude it so a JSON `true` isn't read as 1.
             if isinstance(resolved_iters, int) and not isinstance(resolved_iters, bool) and resolved_iters > 0:
                 sandbox_max_iterations = resolved_iters
+        elif use_sandbox and ctx.db is not None and ctx.workspace_id is not None:
```
khaledosman: This branch fails open. With no workspace the policy is never read, so a workspace whose row says `enabled=False` gets code execution anyway: the veto is skipped rather than enforced.

#724's `_resolve_mcp_server_ids` treats the identical condition (`ctx.db is None or ctx.workspace_id is None`) as unresolvable and refuses with a 400. It is unreachable today, since `resolve_workspace_id` always returns a UUID (falling back to the default workspace) and `db is None` already 500s upstream, which is exactly why a veto's defensive arm should refuse or assert rather than default to allow: the day it becomes reachable, the failure is silent.

web/src/features/tools/WorkspaceCodeExecutionPolicyCard.tsx
```diff
@@ -0,0 +1,239 @@
...
+// Three states, not two, which is why the first control is a select rather than
+// a toggle: a workspace can be allowed, blocked, or carry no policy at all.
+// "Deployment default" is the last of those and is a delete, not a saved
+// `enabled: true`.
+
+type Stance = "default" | "allowed" | "blocked"
+
+// The server's own ceilings (`workspace_code_execution_policy_service`): a value
+// above either could never take effect, so it is refused rather than stored.
+export const MAX_ITERATIONS = 25
```
khaledosman, line 35: These are a third copy of `MAX_TOOL_ITERATIONS_CAP` (25) and `DEFAULT_EXEC_TIMEOUT_S` (60), and `parseCeiling` rejects client-side, so raising either backend constant makes this form refuse a value the API would accept, with nothing failing anywhere to say so.

The generated client carries the types but not the `maximum`, so the enforceable version is a backend test pinning both numbers with a comment naming this file, the way the MCP cap pair needs one on #724.

khaledosman (approved):
Narrow-never-grant is implemented consistently and asserted at both layers, the required rather than defaulted `enabled` is the right call given that absence of a row means unnarrowed here, and the PUT race resolves to the idempotent outcome instead of a conflict nobody could act on.

src/gateway/services/tenancy/workspace_code_execution_policy_service.py
```diff
@@ -0,0 +1,323 @@
...
+            # required rather than tidy.
+            await self.db.rollback()
+            policy = await self.db.get(WorkspaceCodeExecutionPolicy, resolved_id)
+            if policy is None:
+                raise  # not the race: nothing is there to have collided with
+            self._apply(policy, request)
+            await self.db.commit()
+        except SQLAlchemyError:
+            await self.db.rollback()
+            raise
```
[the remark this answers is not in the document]
njbrake: Fixed in ac6f87db2. Both commits now go through a `_commit()` helper that rolls back before any failure escapes, so the retry's commit is guarded exactly like the first one, and `clear_policy` uses it too so there is one commit shape in the file.

The rollback matters for the reason you give: SQLAlchemy leaves a session with a failed flush unusable, so without it the next statement raises `PendingRollbackError` and masks the error that actually happened.

src/gateway/api/routes/_pipeline.py
```diff
@@ -1794,6 +1829,7 @@ async def prepare_gateway_tools(
         sandbox_tool_entry=sandbox_tool_entry,
         sandbox_url=sandbox_url,
         sandbox_auth_token=sandbox_auth_token,
+        sandbox_exec_timeout_s=sandbox_exec_timeout_s,
         use_web_search=use_web_search,
         web_search_tool_entry=web_search_tool_entry,
         web_search_url=web_search_url,
```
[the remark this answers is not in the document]
njbrake: Correct, and it is fixed on main rather than here, so this branch does not re-fix it.

#724 added an `except SQLAlchemyError` arm to this same block. It rolls back before releasing, because a poisoned session makes `release_reservation` itself raise `PendingRollbackError` and the hold survives anyway, and both calls are best-effort so the original failure is what surfaces. After rebasing onto it, the code-execution policy read sits inside that arm, so the case you describe is covered.

Two things I did add: a regression test for the code-execution path specifically (`test_a_policy_read_that_fails_releases_the_budget_reservation`, which fails without the arm), and a one-word correction to the arm's own comment, which counted two database reads in this block where there are now three.

alembic/versions/c4d7f2a9b6e1_add_workspace_code_execution_policies.py
njbrake: Right when you wrote it, and resolved the way your second option describes.

The four PRs landed as a train (#722, #723, #724, #725), so `c5e9a1d3f7b2` is #724's revision on main and is main's single head. This branch is rebased onto it: `alembic heads` reports one head, `c4d7f2a9b6e1`, and upgrade, downgrade, upgrade round-trips against PostgreSQL. The schema-chain tests pass, `test_migration_single_head` among them.

Nothing about the migration changed except the parent pointer.

src/gateway/api/routes/_pipeline.py
njbrake: Good catch, fixed in ac6f87db2.

The arm is now `elif use_sandbox:` and refuses with a 500 (`CODE_EXEC_POLICY_UNRESOLVABLE_DETAIL`) when `ctx.db is None or ctx.workspace_id is None`, rather than falling through to no policy read. That is the shape `_resolve_mcp_server_ids` uses at the identical condition.

Your reasoning is also why it refuses rather than asserts. It is unreachable today, for the two reasons you name, and a veto that silently skips itself on the day one of those invariants stops holding is worth a branch that says so out loud. Pinned by `test_an_unresolvable_workspace_refuses_rather_than_running_the_sandbox`.

web/src/features/tools/WorkspaceCodeExecutionPolicyCard.tsx
njbrake: Half of this was already covered and half was not, so I added the half that was missing.

Already covered: the card's own test compares both literals against `maximum` in the committed `docs/public/openapi.json` (`keeps its ceilings equal to the ones the server enforces`). You are right that `openapi-typescript` drops `maximum` from `schema.ts`, but the spec itself carries it, so that drift does fail somewhere.

What it missed is locality, which is your actual point: someone raising `MAX_TOOL_ITERATIONS_CAP` in `mcp_loop.py` runs pytest, not vitest. Added `tests/unit/test_code_execution_policy_limits.py`, which reads both literals out of `WorkspaceCodeExecutionPolicyCard.tsx` and compares them to the server's own derived ceilings, so the failure lands in the suite the person editing the constant is already running.

Issue #653, opened 2026-08-19
The frontend auth and session surface


njbrake:
- Labels: `otari-migration`, `open-core`
- Part of mozilla-ai/otari-ai#1452, identity and auth track and #589
- Depends on #647, then each flow on its own backend issue

## Why this exists

`web/src/features/auth/` is `Login.tsx` and `AuthContext.tsx`: one text box that posts a master key. The platform carries ten route files covering login, signup, check-email, verify-email, resend-verification, recover-password, reset-password, accept-invitation, and the Google and GitHub callbacks, plus the feature behind them.

## What to produce

Those routes and their feature, with the master-key box replaced by the session login from #649.

## Lands incrementally

Each flow merges dark and lights up when its backend arrives, the same pattern #588 used for the tenancy pages. This does not wait for #649 through #652 to all land, and it should not be one pull request.

Mapping, so the order is not guesswork:

| Flow | Backend issue |
| --- | --- |
| login | #649 |
| signup, check-email, verify-email, resend-verification, recover-password, reset-password | #650 |
| Google and GitHub callbacks | #651 |
| passkey registration and sign-in affordances | #652 |
| accept-invitation | #641 |

## Does not port: the data-region selector

The hosted product is adding a United States / Europe data-region dropdown to its own login and signup pages, resolving the API base URL from a pair of per-region environment variables and persisting the choice across the OAuth redirect. It lands in the same files this issue ports: the login and signup routes, the auth hook, the OAuth callback handler, and two new helper modules.

**None of it comes here.** A self-hosted Otari is one deployment with one backend and no region to choose, so the dropdown would be an affordance that cannot do anything. Regional residency is hosted-only surface.

Called out because the exclusion is invisible from the file: whoever ports the login route will find the selector sitting in it and has no reason to know it should be left behind.

## Definition of Done

- Every route has a page and there are no dead links.
- A flow whose backend has not landed is hidden rather than broken.
- The master-key text box is gone from the steady-state login, per mozilla-ai/otari-ai#1716.

_Note: this issue was drafted by Claude Opus 5 via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

njbrake:
Closing. Both rows still open on the table are now owned by the backend issue that lights them up, so there is nothing left here to sequence.

[...]

Issue #652, opened 2026-08-19
WebAuthn and passkeys


njbrake:
- Labels: `otari-migration`, `open-core`
- Part of mozilla-ai/otari-ai#1452, identity and auth track
- Depends on #647

## Why this exists

WebAuthn appears nowhere in this repository. The table, the ceremonies and the management surface are all net-new here, and the hosted product depends on them.

## What to produce

- A `webauthn_credential` table. It is a table rather than a column, which is why it was left out of the additive migration in #645.
- Registration and login ceremonies, and challenge handling.
- Credential list, rename and delete in settings.

## Standing constraint from mozilla-ai/otari-ai#1716

Migrating otari.ai users **import their credentials rather than claiming new accounts**. `webauthn_credential` binds a passkey to a relying-party ID derived from the origin, so that import holds only while the origin stays `otari.ai`. This is recorded as a standing constraint on mozilla-ai/otari-ai#1592 and must not be lost between here and the cutover.

## Definition of Done

- A passkey registers and authenticates.
- Credentials list, rename and delete.
- The relying-party ID derivation is explicit and documented, rather than implied by whatever URL the deployment happens to serve.

_Note: this issue was drafted by Claude Opus 5 via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

Issue #651, opened 2026-08-19
OAuth sign-in, and make IdentityProviderPort real


njbrake:
- Labels: `otari-migration`, `open-core`
- Part of mozilla-ai/otari-ai#1452, identity and auth track
- Depends on #647

## Why this exists

Two things, and they belong together because the second exists to serve the first.

`IdentityProviderPort` is in ARCHITECTURE.md's port table and in no Python file in this repository. It is not alone: `services/routing/__init__.py` says the same of its own, "Plain core code, with no port. ARCHITECTURE.md names a RoutingPort". Identity is where that has to stop being true, because the enterprise overlay binds SSO through this port and cannot bind to a table row in a document.

## The OAuth work is lighter than it looks

mozilla-ai/otari-ai#1740 moved the mechanics onto apron-auth, which owns provider endpoints, code exchange, and the userinfo fetch that normalizes a provider response into an identity profile. What ports is the platform's own part: which providers are configured, which scopes are requested, and how a fetched identity maps onto a user.

## Two carry-overs that must survive the port

- **PKCE is deliberately off.** Authorize and callback are independent requests with no store between them, so a verifier minted at authorize time has nowhere to live. For the same reason the authorization URLs are built by hand rather than by `OAuthClient.get_authorization_url`. Turning PKCE on is a real change with a real prerequisite, not a default to restore.
- **`email_verified` is tri-state in apron-auth and resolved on a bool here**, so an unasserted value collapses to unverified rather than being laundered into a verified identity. mozilla-ai/otari-ai#1551 moves resolution onto the tri-state model and onto `(provider, subject)` keying. It should land in one place, not be ported and then fixed twice.

## What to produce

- `IdentityProviderPort` as a real interface with a working core adapter, per ARCHITECTURE.md's cardinal property that every port ships with one.
- Google and GitHub sign-in behind it, on apron-auth.

## Definition of Done

- Google and GitHub sign-in work against a real provider, with the live exchange gated behind an opt-in test flag rather than assumed from a green unit suite.
- An overlay binds a different identity adapter without editing Otari source.
- The core adapter is a real implementation, not a stub.

## The dashboard half

[...]

Issue #650, opened 2026-08-19
Signup, email verification, and password reset


njbrake:
- Labels: `otari-migration`, `open-core`
- Part of mozilla-ai/otari-ai#1452, identity and auth track
- Depends on #647 and #648

## Why this exists

`user` gains `hashed_password`, `email_verification_token`, `email_verified_at` and `terms_accepted_at` in #645, and nothing reads them. This is the issue that reads them.

## What to produce

Port `create_user_for_signup`, `verify_email` and `resend_verification_email` from `otari-ai`'s `user_service`, plus the recover and reset password pair.

## Definition of Done

- A new account signs up, receives a verification message, verifies, and signs in.
- The policy for an unverified account is recorded here and enforced: what it can do before verifying, and for how long.
- A password reset round-trips.
- Verification and reset tokens expire and are single use.
- With no mail transport configured (#648), these surfaces are unavailable rather than broken.

_Note: this issue was drafted by Claude Opus 5 via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

PR #726, opened 2026-08-21
feat(dashboard): give signup, verification, and password reset their pages

njbrake:
## Description

The #650 tranche of #653, which the issue says lands incrementally: `#/signup`, `#/check-email`, `#/verify-email`, `#/resend-verification`, `#/recover-password`, `#/reset-password`. Login (#649) already landed; the OAuth callbacks (#651), passkeys (#652) and accept-invitation (#641) are not here.

They are hash paths rendered by `DeploymentRoot` ahead of the auth gate, not files in `src/routes/`, for the reason `AcceptInvitationPage` is not one either: every route there sits behind that gate, and someone opening an emailed link holds neither a session nor the master key. So there is no route tree to commit. Two of the six spellings are the gateway's, not ours: `user_service.py` mails `/#/verify-email?token=...` and `/#/reset-password?token=...`.

Four of the six begin by sending a message, so they gate on the bootstrap's `mail_ready` the way the invitation form already does (#648): the sign-in screen hides the link, and a URL a bookmark still reaches answers with a panel saying so rather than a form whose only outcome is the route's 503. The two that a token in hand reaches never gate, since the message carrying that token was sent while mail did work.

### Deliberately not ported

- **The data region dropdown**, which #653 calls out: a self-hosted gateway is one deployment with one backend, so it would be an affordance with nothing to choose. Its API base URL resolver and helper modules stay behind with it.
- **The Google and GitHub buttons** (#651).
- **The newsletter opt-in and the terms checkbox**, which have no counterpart on a deployment that publishes no terms and runs no marketing list. The request omits `terms_accepted` rather than asserting acceptance of a document that does not exist.
- **`PreloginLayout`'s chrome** (Mozilla byline, marketing footer, forced light theme), in favor of the card shape `Login` and `AcceptInvitationPage` already use.

Signup's copy says "claim", not "create", because the backend only ever completes an identity already on the roster. The password policy moves to `shared/helpers/password.ts` so the three doors that mint or replace a password state it once.

The second commit closes a cold review of the first. The one worth naming: verifying an email fired a mutation from an effect, and StrictMode's development-only remount spent the single-use token twice, landing the second call's 400 over the first call's success. `useVerifyEmail` is a query keyed on the token now, the shape `useValidateInvitation` already uses, and the page has no effect at all.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Part of #653 (the #650 row of its table). Deliberately does not close it: #651, #652 and #641 are still to come.

## Checklist

### Verification detail

- `pnpm --dir web run lint`, `run typecheck`, and `test` (1216 tests) pass, as do `make lint`, `make typecheck`, `make openapi-check` and `make postman-check`. No backend change, so nothing generated was owed, and no `schema.ts` drift after `client:generate`.
- **Walked in a browser against a live gateway** on the console mail transport: admin adds a member, claim from the sign-in screen, check email, sign-in refused for an unverified address with the gateway's own wording, follow the mailed verify link, sign in, request a reset, follow the mailed reset link, sign in with the new password, replayed token refused. Resend verification walked separately with a second address.
- A link crawl from the sign-in screen: every hash link resolves to a real page.
- Screenshot entries added for all six pages, run on `screenshots-desktop-large-light` and `screenshots-mobile-dark` against a real gateway. Not run: the behavioral e2e suite, and the other four screenshot projects.
- The e2e fixture deliberately reports `mail_ready: false` (`parity.bootstrap.spec.ts` asserts it), so the four mail-gated captures stub that one boolean through `page.route` instead of changing a fixture another spec depends on.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:** Claude Code (Claude Opus 5), driven by @njbrake.

**Any additional AI details you'd like to share:** Scoped to one tranche and to a stated exclusion by @njbrake up front. The platform's route files were read from a throwaway clone of `mozilla-ai/otari-ai` outside this worktree and ported rather than pasted; no cross-repo commit was made. The first commit was reviewed cold by a separate agent that was given the branch and nothing else, and the second commit is what that review produced.

**NOTE:**

https://claude.ai/code/session_019bkkgqV9JAZfJtEZZKzUfn

## Summary

- Added standalone hash-based pages for signup, email verification, verification resend, password recovery, and password reset.
- Added shared authentication layouts, navigation, token handling, API hooks, and password validation.
- Added mail-readiness checks for email-based flows.
- Updated login links, documentation, tests, and screenshots.
- Improved verification safety, error handling, route matching, and accessibility.

These changes provide clear self-service account and password recovery flows while respecting gateway configuration and token security rules.

## Technical notes

- Token-based verification and reset pages remain available after mail delivery is disabled.
- Email verification uses a token-keyed query with automatic refetching disabled.
- Password validation uses shared Unicode character and UTF-8 byte limits.
- Login, OAuth callbacks, passkeys, and invitation acceptance remain outside this tranche.


njbrake:
All seven addressed in 2c0bf1007. Six were real.

**Empty token.** `?token=` parses as `""`, not null, so callers skipped the missing-token branch and then declined to fetch on an empty string: the page sat on "Confirming your address…" forever having asked the gateway nothing. An empty token is a malformed link now. `AcceptInvitationPage` shares the helper and the fix.

**`useVerifyEmail` was not quite a one-shot.** Correct that a failed query holds no data for `staleTime` to keep fresh. `retryOnMount`, `refetchOnMount`, `refetchOnReconnect` and `refetchOnWindowFocus` are off by name now rather than by the provider's defaults.

**`docs/dashboard.md`.** Right: `PUT /v1/auth/password` always acts on the caller's own identity, so master-key recovery is how the operator gets back in and sets nobody else's password. With no mail a member has no fallback at all, and the guide says that now.

**`web/AGENTS.md`.** Three pages send a message, not four. `#/check-email` is gated with them because it is where signup and a resend land.

**`cancelling`.** Fixed in the four new pages and in the one pre-existing instance in `PasswordCard`, which this branch already touches.

**Reset page stale-error test.** Added.

The `QueryClient` one did not reproduce. Measured, the call count stays at 1 after the rerender, both synchronously and after settling. Taken anyway, since swapping the client makes what the test proves ambiguous.

web/src/features/auth/PublicAuthLayout.tsx
```diff
@@ -0,0 +1,92 @@
...
+}: {
+  title: string
+  /** Optional subhead under the title. */
+  description?: ReactNode
+  children: ReactNode
+  /** Links below the divider: where to go next when this page is a dead end. */
+  footer?: ReactNode
+}) {
+  return (
+    <div className="flex min-h-full items-center justify-center p-6">
```
khaledosman, line 30: This is the third copy of the same shell. `Login.tsx:133-140` and `AcceptInvitationPage.tsx:37-44` both repeat `flex min-h-full items-center justify-center p-6`, `Card` with `w-full max-w-md`, `Card.Content` with `flex flex-col gap-5 p-7`, the favicon mark at `h-12 w-12`, and an `h1` with `text-lg font-semibold text-foreground`, verbatim.

Both files are already in this diff, so migrating them onto `PublicAuthLayout` is in scope, and it is what stops the next tweak to the card from landing in one of three places.

web/src/features/auth/CheckEmailPage.tsx
```diff
@@ -0,0 +1,45 @@
...
+ * identity to act on, so this page has to be written in the same conditional
+ * or it would leak, on the client, the fact the server declined to state.
+ *
+ * `?type=resend` only changes which of the two sentences reads naturally; a
+ * missing or unrecognized value falls back to the signup wording rather than
+ * being validated, since the page is correct either way and a visitor who
+ * hand-edited the URL should still get a page.
+ */
+export function CheckEmailPage({ hash }: { hash: string }) {
+  const type = new URLSearchParams(hash.split("?")[1] ?? "").get("type")
```
khaledosman, line 17: This hand-rolls the parsing that `shared/helpers/hashParams.ts` was added for in this same PR, so there are now two places that know parameters live after the `?` inside the fragment.

Add a `hashParam(hash, name)` there and have `tokenFromHash` call it, keeping the empty-string-is-not-a-token rule where it is.

web/src/features/auth/PublicAuthPage.tsx
```diff
@@ -0,0 +1,70 @@
...
+  path: PublicAuthPath
+  hash: string
+}) {
+  const { mail_ready } = useDeployment()
+
+  if (!isPublicAuthPageAvailable(path, mail_ready)) {
+    return <MailUnavailable />
+  }
+
+  switch (path) {
```
khaledosman, line 40: The switch is exhaustive today, but nothing enforces that it stays so: adding a key to `PUBLIC_AUTH_PAGES` without a case here infers `Element | undefined` and renders a blank screen, which is the exact failure `publicAuthPaths.ts`'s own comment about `Object.hasOwn` names.

Annotating the return type (`: ReactElement`) turns a missing case into a typecheck error instead of a blank page.

khaledosman (approved):
Coherent tranche. The mail gating matches how #648 settled it for the invitation form, the enumeration-safe wording is consistent at every door including `#/check-email` (which is where a client-side leak of what the server declined to state would actually have happened), and turning verification into a token-keyed query with all three automatic refetches named individually, rather than leaning on the provider's defaults, is the right fix for a single-use token under StrictMode.

Worth naming because it is easy to miss: sharing `tokenFromHash` also fixes `#/accept-invitation?token=` with an empty value, which used to sit on "Checking your invitation" forever and now says the link carries nothing.

Three things below, none blocking.

PR #730, opened 2026-08-23
fix(gateway): stop purpose-hint injection from corrupting list-shaped system content

### LINKED ISSUE #590: Purpose-hint injection corrupts list-shaped system content and can insert a second system message
Verified against `main` at commit `9458131` (`9458131764a90968c9369d83a909bc576cfb9bdf`).

`inject_purpose_hints` merges the MCP purpose-hint block into the leading system message with a bare f-string and no type branch. Two distinct defects fall out of that: list-shaped system content is corrupted into a Python `repr`, and a system message that is not at index 0 causes a second system message to be inserted.

## 1. List-shaped system content becomes a Python repr in the prompt

[`src/gateway/services/mcp_loop.py:88-94`](https://github.com/mozilla-ai/otari/blob/9458131764a90968c9369d83a909bc576cfb9bdf/src/gateway/services/mcp_loop.py#L88-L94):

```python
out = list(messages)
if out and out[0].get("role") == "system":
    existing = out[0].get("content") or ""
    out[0] = {**out[0], "content": f"{existing}\n\n{block}" if existing else block}
else:
    out.insert(0, {"role": "system", "content": block})
return out
```

OpenAI chat legally allows system content to be either a string or a list of text content parts. From the vendored SDK's own type, `openai.types.chat.ChatCompletionSystemMessageParam`:

```python
content: Required[Union[str, Iterable[ChatCompletionContentPartTextParam]]]
```

When a caller sends the list form, `f"{existing}\n\n{block}"` stringifies the list, so the model receives the Python `repr` of the content-part dicts as literal prompt text.

Observed on this commit:

```python
from gateway.services.mcp_loop import inject_purpose_hints

msgs = [
    {"role": "system", "content": [{"type": "text", "text": "be helpful"}]},
    {"role": "user", "content": "hi"},
]
out = inject_purpose_hints(msgs, [("cal", "for scheduling")])
print(repr(out[0]["content"]))
```

```
"[{'type': 'text', 'text': 'be helpful'}]\n\nYou have access to the following tools:\n- cal: for scheduling"
```

The system prompt now literally begins with `[{'type': 'text', 'text': 'be helpful'}]`. The original instruction survives only as quoted debris inside a repr, and the message silently changes shape from list to string.

The Anthropic-side helper already handles exactly this, at [`src/gateway/services/tool_format.py:98-105`](https://github.com/mozilla-ai/otari/blob/9458131764a90968c9369d83a909bc576cfb9bdf/src/gateway/services/tool_format.py#L98-L105):

```python
if existing is None:
    call_kwargs["system"] = block
elif isinstance(existing, str):
    call_kwargs["system"] = f"{block}\n\n{existing}" if existing else block
elif isinstance(existing, list):
    call_kwargs["system"] = [{"type": "text", "text": block}, *existing]
else:
    call_kwargs["system"] = block
```

The chat path is missing the `isinstance` ladder its own sibling has.

## 2. A system message at index >= 1 yields two system messages

The merge check tests `out[0]` only. Nothing in the OpenAI chat contract requires the system message to be first, and the `developer` role (o1 and newer) commonly precedes it. When the system message sits at any index >= 1, the `else` branch runs and inserts a *second* system message at the front:

```python
msgs = [
    {"role": "developer", "content": "dev preamble"},
    {"role": "system", "content": "be helpful"},
    {"role": "user", "content": "hi"},
]
out = inject_purpose_hints(msgs, [("cal", "for scheduling")])
print([m["role"] for m in out])
```

```
['system', 'developer', 'system', 'user']
```

Two system messages, and the hints are attached to the one the caller did not write.

## Reachability

Both shapes are reachable from the wire. `ChatCompletionRequest.messages` is typed as `list[dict[str, Any]]` ([`src/gateway/api/routes/chat.py:80`](https://github.com/mozilla-ai/otari/blob/9458131764a90968c9369d83a909bc576cfb9bdf/src/gateway/api/routes/chat.py#L80)) and the only structural validation is a `role` key presence check ([`chat.py:103-110`](https://github.com/mozilla-ai/otari/blob/9458131764a90968c9369d83a909bc576cfb9bdf/src/gateway/api/routes/chat.py#L103-L110)):

```python
@field_validator("messages")
@classmethod
def validate_message_structure(cls, v: list[dict[str, Any]]) -> list[dict[str, Any]]:
    for i, message in enumerate(v):
        if "role" not in message:
            msg = f"messages[{i}]: 'role' is required"
            raise ValueError(msg)
    return v
```

Content shape and message ordering are never constrained, so both inputs pass validation and reach the injector unchanged.

The injector only runs when there are hints, so this needs at least one MCP server carrying a `purpose_hint`. Given that, any caller using list-shaped system content (a normal choice, and the one required for Anthropic-style `cache_control` on system blocks) silently gets a corrupted prompt.

## Test gap

The existing coverage at [`tests/unit/test_mcp_loop.py:143-161`](https://github.com/mozilla-ai/otari/blob/9458131764a90968c9369d83a909bc576cfb9bdf/tests/unit/test_mcp_loop.py#L143-L161) exercises string content and a system-at-index-0 request only:

- `test_inject_purpose_hints_no_hints_returns_unchanged`
- `test_inject_purpose_hints_prepends_when_no_system`
- `test_inject_purpose_hints_extends_existing_system`

No case covers list-shaped content or a system message at index >= 1.

## Suggested fix

Give `inject_purpose_hints` the same `isinstance` ladder as `tool_format.py`, preserving the list shape by prepending or appending a text part rather than stringifying. Separately, locate the system message by scanning for the first `role == "system"` rather than testing index 0 only, so an existing system message anywhere in the list is extended instead of shadowed. Worth deciding at the same time whether a `developer` message should be treated as the merge target for o1-class models.

Add unit coverage for: list-shaped system content, system at index >= 1, and a `developer` + `system` combination.

Note: this is a pre-existing gateway defect, found while verifying the v0 observation schema against this repo. Not Reprise work.

shoemoney:
## Description

`inject_purpose_hints` (`src/gateway/services/mcp_loop.py:66-94`) merges the MCP purpose-hint block into the leading system message with a bare f-string and no type check on the content shape. Two bugs come out of that:

1. **List-shaped system content gets stringified into a Python repr.** OpenAI's chat contract allows `content` on a system message to be either a string or a list of content-part dicts (the shape needed for `cache_control`-style caching). At `mcp_loop.py:90-91` (before the fix), `f"{existing}\n\n{block}"` calls `str()`/`repr()` on the list when `existing` is a list, so the model receives the literal Python repr of the content parts as prompt text, e.g. `"[{'type': 'text', 'text': 'be helpful'}]\n\nYou have access to the following tools:\n..."`. The original system prompt survives only as quoted debris inside that repr.

2. **A system message not at index 0 causes a second system message to be inserted.** The merge check only tested `out[0].get("role") == "system"` (`mcp_loop.py:89`). Nothing in the OpenAI chat contract requires `system` to be the first message, and a `developer` message (o1-class models) commonly precedes it. When `system` sits at index >= 1, the `else` branch ran and inserted a *second* system message at the front, so the hints land on a message the caller never wrote while the caller's actual system message is untouched.

Both are reachable from the wire: `ChatCompletionRequest.messages` (`src/gateway/api/routes/chat.py:80`) is typed `list[dict[str, Any]]`, and its validator (`chat.py:103`) only checks for a `role` key — content shape and message order are unconstrained. `chat.py:263` calls `inject_purpose_hints(kwargs["messages"], hints, header=header)` on wire-supplied messages whenever any configured MCP server has a `purpose_hint`. Any caller sending list-shaped system content, or a `developer` message ahead of `system`, gets a silently corrupted request on every call.

`src/gateway/services/tool_format.py:98-105` (`inject_purpose_hints_anthropic`, used by the `/v1/messages` and Responses routes) already has the correct fix for the content-shape half of this: an isinstance ladder over `None` / `str` / `list` / `else` that keeps a list a list. The chat-completions path is the one place that never got that ladder. This PR brings it in line: scans for the first message with `role == "system"` instead of testing index 0 only, and branches on the existing content's type the same way `tool_format.py` does, rather than assuming `str`.

## PR Type

- [x] Bug Fix

## Relevant issues

Fixes #590

## What I ran

```
uv sync --frozen
uv run pytest tests/unit/test_mcp_loop.py -v
```

RED (source reverted to the pre-fix version, new tests left in place):
```
FAILED tests/unit/test_mcp_loop.py::test_inject_purpose_hints_preserves_list_shaped_system_content
FAILED tests/unit/test_mcp_loop.py::test_inject_purpose_hints_developer_then_system_extends_the_system_message
2 failed, 26 passed in 79.55s
```

GREEN (fix restored):
```
28 passed in 74.89s
```

`uv run pytest tests/unit` (the full unit suite) did not finish in this environment after 10+ minutes at ~16% progress, so I scoped the RED/GREEN runs to `tests/unit/test_mcp_loop.py`, which is the file covering this change, rather than block on that.

## Checklist

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

Claude Opus 5 (Claude Code)

Written in conjunction with my pair programmer Claude. The finding was verified against upstream before any code was written, and the fix is covered by tests that fail against the unpatched source.

**NOTE:**

## Summary

Updated `inject_purpose_hints` to handle system messages consistently across message layouts.

- Preserves list-shaped system content.
- Extends the first system message wherever it appears.
- Prevents duplicate system messages when a developer message appears first.
- Preserves existing string-content behavior.
- Adds regression tests for these cases and for conversations without a system message.

Targeted tests pass: 28 tests.

### DIFF
--- a/src/gateway/services/mcp_loop.py
+++ b/src/gateway/services/mcp_loop.py
@@ -71,6 +71,12 @@ def inject_purpose_hints(
 ) -> list[dict[str, Any]]:
     """Prepend or extend the system message with per-tool usage hints.
 
+    The merge target is the first message whose ``role`` is ``"system"``, wherever
+    it sits (a ``developer`` message may legally precede it); one is inserted at
+    the front when the conversation has none. ``content`` may be a string or a
+    list of content parts, so a list keeps its shape and gains a leading text
+    part rather than being stringified.
+
     Header resolution priority:
       1. ``header`` arg (per-request override, set from the request body)
       2. ``OTARI_TOOLS_HEADER`` env (per-deployment override)
--- a/src/gateway/services/mcp_loop.py
+++ b/src/gateway/services/mcp_loop.py
@@ -86,11 +86,21 @@ def inject_purpose_hints(
     block = "\n".join(lines)
 
     out = list(messages)
-    if out and out[0].get("role") == "system":
-        existing = out[0].get("content") or ""
-        out[0] = {**out[0], "content": f"{existing}\n\n{block}" if existing else block}
-    else:
+    system_idx = next((i for i, m in enumerate(out) if m.get("role") == "system"), None)
+    if system_idx is None:
         out.insert(0, {"role": "system", "content": block})
+        return out
+
+    existing = out[system_idx].get("content")
+    if existing is None:
+        new_content: Any = block
+    elif isinstance(existing, str):
+        new_content = f"{existing}\n\n{block}" if existing else block
+    elif isinstance(existing, list):
+        new_content = [{"type": "text", "text": block}, *existing]
+    else:
+        new_content = block
+    out[system_idx] = {**out[system_idx], "content": new_content}
     return out
 
 
--- a/tests/unit/test_mcp_loop.py
+++ b/tests/unit/test_mcp_loop.py
@@ -159,6 +159,51 @@ def test_inject_purpose_hints_extends_existing_system() -> None:
     assert out[0]["role"] == "system"
     assert "be helpful" in out[0]["content"]
     assert "cal" in out[0]["content"]
+    # regression: string content behaves exactly as today (existing text kept ahead of the hint block)
+    assert out[0]["content"] == "be helpful\n\nYou have access to the following tools:\n- cal: use it"
+
+
+def test_inject_purpose_hints_no_system_inserts_at_front() -> None:
+    """Regression: with no system message present, one is still inserted at index 0."""
+    msgs = [{"role": "user", "content": "hi"}]
+    out = inject_purpose_hints(msgs, [("cal", "for scheduling")])
+    assert len(out) == 2
+    assert out[0] == {"role": "system", "content": "You have access to the following tools:\n- cal: for scheduling"}
+    assert out[1] == {"role": "user", "content": "hi"}
+
+
+def test_inject_purpose_hints_preserves_list_shaped_system_content() -> None:
+    """List-shaped system content (Anthropic-style cache_control parts) must not be stringified into a repr."""
+    msgs = [
+        {"role": "system", "content": [{"type": "text", "text": "be helpful"}]},
+        {"role": "user", "content": "hi"},
+    ]
+    out = inject_purpose_hints(msgs, [("cal", "for scheduling")])
+    assert out[0]["role"] == "system"
+    content = out[0]["content"]
+    assert isinstance(content, list)
+    # the hint block is prepended as its own text part; the original part survives untouched.
+    assert content[0] == {"type": "text", "text": "You have access to the following tools:\n- cal: for scheduling"}
+    assert content[1] == {"type": "text", "text": "be helpful"}
+    # no repr() debris anywhere in the output.
+    assert "{'type'" not in str(content[0]["text"])
+    assert "{'type'" not in str(content[1]["text"])
+    assert len([m for m in out if m.get("role") == "system"]) == 1
+
+
+def test_inject_purpose_hints_developer_then_system_extends_the_system_message() -> None:
+    """A developer message ahead of system must not cause a second system message to be inserted."""
+    msgs = [
+        {"role": "developer", "content": "dev preamble"},
+        {"role": "system", "content": "be helpful"},
+        {"role": "user", "content": "hi"},
+    ]
+    out = inject_purpose_hints(msgs, [("cal", "for scheduling")])
+    system_messages = [m for m in out if m.get("role") == "system"]
+    assert len(system_messages) == 1
+    assert [m["role"] for m in out] == ["developer", "system", "user"]
+    assert "be helpful" in out[1]["content"]
+    assert "cal" in out[1]["content"]
 
 
 def test_finalize_tool_calls_orders_by_index() -> None:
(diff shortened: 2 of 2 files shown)


njbrake (approved):
Approving. Both defects reproduce on main and the fix clears them: list-shaped system content stayed a Python repr in the prompt, and a `system` message behind a `developer` message produced a second system message carrying the hints.

The approach matches `tool_format.inject_purpose_hints_anthropic`, which already had the None/str/list/else ladder for the same problem on the Anthropic side, so this brings the chat-completions path in line rather than adding a second mechanism beside it.

I pushed one commit to the branch. The new list-shaped-content test built a mixed literal, so mypy inferred `list[object]` and the Typecheck job failed; annotating the local clears it. I also documented the merge-target scan and the content-shape handling on the function, matching how the Anthropic helper documents its own. All checks are green on 4335146.

Two things I left as follow-ups rather than blocking on:

1. The hint block lands in a different position depending on content type. The string arm appends it after existing text, the list arm prepends it. The Anthropic helper prepends in both, so this is worth unifying one way or the other.
2. A `system` message sitting mid-conversation now has the hints merged into it instead of receiving a fresh system message at index 0, so the tool description arrives at the end of the prompt. Rare and low impact, but nothing tests it and nothing would surface it in production.

Thanks for the clear writeup and for including the RED/GREEN runs.

PR #731, opened 2026-08-24
ci(dashboard): gate the parity suite on gateway changes, and cover a hybrid deployment

khaledosman:
## Description

Two gaps in the dashboard's browser suite, both named in mozilla-ai/otari-ai#1600.

**It did not run on backend changes.** The suite was scoped to `web/**`, so a change under `src/gateway` could break a flow the dashboard serves and land green; the workflow said as much in its own comment. The Playwright job moves into `otari-dashboard-parity.yml`, triggered by `src/gateway/**` as well as `web/**`. It is a separate workflow rather than a widened path list because a path filter applies to a whole workflow, so widening `otari-dashboard.yml` would run vitest, tsc and a bundle build on every backend PR. `otari-dashboard-serving.yml` was split off the same file for the same reason and owns the pytest half.

**It only asserted the standalone deployment shape.** A hybrid gateway is data-plane only: its organizations, credentials, routing, budgets and usage live on otari.ai, so what it may show an operator is its own health, whether it can reach that control plane, the base URL to point a client at, and a link out. Nothing else. That is decided by the server, so `parity.hybrid.spec.ts` runs in a `hybrid` project against a second gateway booted in hybrid mode beside the standalone one. The standalone gateway's `/health` stands in for the control plane, so the connection row is covered in its connected state and not only its failure one.

The component half (`HybridLanding` against mocks) and the per-route 404s (`tests/integration/test_hybrid_mode_surface.py`) already existed and are not repeated here. What only this suite can do is put the server's real answer and the real bundle in one browser.

## PR Type

- [ ] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [x] Infrastructure / CI

## Relevant issues

Refs mozilla-ai/otari-ai#1600

## Checklist

Local run: `pnpm --dir web run e2e` (43 passed, including the 3 new hybrid specs), `pnpm --dir web run lint`, `pnpm --dir web run typecheck`, `pnpm --dir web test` (1209 passed), `make lint`. No Python changed. The connection assertion was checked against a dead control plane to confirm it is not vacuous.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

Implementation, tests and prose drafted by the agent; the scoping decisions come from the issue.

## Summary

- Reorganized dashboard CI to separate gateway parity tests from focused dashboard checks.
- Added browser coverage for a real hybrid gateway deployment.
- Added hybrid checks for health, control-plane connectivity, client URLs, management links, and route behavior.
- Added hybrid gateway configuration and startup support.
- Updated test guidance and scripts to include the hybrid project.
- Prevented inherited tokens from switching standalone tests to hybrid mode.

These changes improve hybrid deployment confidence and keep dashboard checks focused.

## Technical notes

- The parity workflow runs for `web/**` and `src/gateway/**` changes.
- Serving checks run in a separate workflow.
- Playwright runs `onboarding`, `seed`, `parity`, and `hybrid` projects.
- Existing `HybridLanding` component tests and hybrid route integration tests remain unchanged.


njbrake (approved):
Approving. I pushed one commit to this branch (c1aeb8b): `serve-hybrid.sh` defaulted `OTARI_AI_TOKEN` rather than setting it, so a real otari.ai token in the operator's shell was loaded into the throwaway gateway instead of the placeholder. That is the hazard the standalone half closes by unsetting it, pointing the other way. Hardening rather than a live leak, since `platform.base_url` is localhost and nothing in the suite spends the token. Re-ran the hybrid project with a token exported; all 3 pass.

I booted both gateways from the scripts here and checked the two claims the change rests on, rather than reading them:

- **The connection row is not vacuous.** With the standalone gateway up, `:8010/health` reports `platform_reachable: "yes"`; kill :8000 and it flips to `"no"` (readiness to 503). The `health_path` override does what its comment says.
- **`unset OTARI_AI_TOKEN` in `serve.sh` is load-bearing, not tidying.** Booting `web/e2e/otari.yml` with an ambient token gives `deployment_type: "hybrid"`, which fails every parity spec.

Also confirmed `:8010/v1/bootstrap` matches the spec's whole-object assertion field for field, `/v1/providers` 404s, the hybrid config creates no files, and the full suite is 43 passed locally.

The split follows `otari-dashboard-serving.yml` closely enough that there is nothing to argue with, and `main` has no required status checks, so removing the `e2e` job leaves nothing dangling.

One non-blocking note: the rationale for the split is now written in four places (the new workflow header, `otari-dashboard.yml`, `web/AGENTS.md`, `testing.md`), and the two mechanical rules about `testMatch` and the project list in four more (`playwright.config.ts`, `e2eProjects.test.ts`, and both docs). `AGENTS.md` asks for the narrowest layer and a link. The cut I would make is `web/AGENTS.md`'s closing paragraph on line 65, which restates `testing.md`'s new paragraph nearly claim for claim. Your call; the existing bullet was already this long.

CodeRabbit's title-length warning carries nothing here: `cliff.toml` skips `^ci` commits, so this title never reaches the changelog.

khaledosman:
Took the cut. `web/AGENTS.md`'s closing paragraph is gone in 8031893; the two mechanical rules now live in `testing.md` alone, with a link from the bullet that used to restate them. The hybrid spec's exclusion from the `parity` project stayed put, next to the `testIgnore` that performs it, which is the narrowest layer it could sit in.

Your `serve-hybrid.sh` fix is right and my reasoning for the default was wrong. I picked `${VAR:-...}` for consistency with `serve.sh`'s `OTARI_SECRET_KEY`, but the two are not analogous: that one is a local encryption key with no meaning outside the process, and this one is a credential for a real external service. It also pointed the opposite way from the unset I had added one commit earlier, which should have been the tell.

Re-ran the full suite after merging your commit, with a token exported: 43 passed.

(via Claude Code)

PR #733, opened 2026-08-24
fix(dashboard): login design improvements

jigjigjig:
## Description

Before:
image

After:
image

Redesigns `web/src/features/auth/Login.tsx` for craft and accessibility. Both
credential branches (`master_key` and `password`) get the work, and the
`sign_in_methods: []` state gets the frame and the mark.

Design reference: https://app.paper.design/file/01M0SD8XYZMTZ7GTB9G884NY7H/1-0

## Accessibility floor

| | Before | After |
| --- | --- | --- |
| Credential inputs | 14px, so iOS Safari zooms on focus | 16px (`text-base`) |
| Sign in button | 36px | 44px |
| Disclosure summary | 16px | 44px row, 14px medium label |
| Footer links | 44px already | unchanged |
| Reveal toggle | did not exist | 36px visible, 44px hit area |
| Credential inputs on mount | `autoFocus` | not focused |
| The mark | `alt="Otari"` beside an `<h1>` naming the product | `alt=""` |

The submit button is no longer disabled on an empty box. Disabled, it renders
white on `#a5bec7` (1.95:1), and an empty form is this screen's resting state,
so that unreadable pairing was the first thing an operator saw on every visit.
Emptiness is checked on submit and names the box to fill. **The `isSigningOut`
disable stays** ; that is the #557 guard, and its test still covers it.

Making that work needed `validationBehavior="aria"` on the fields. react-aria's
`native` default puts a `required` attribute on the input, so the browser
cancels the submit event itself and shows an unstyleable "Please fill out this
field" bubble that is gone on the next scroll. `aria` keeps `aria-required` and
lets this form render the refusal where it decided to.

## Layout stability

The card was vertically centered, so anything that grew it moved its top edge
up by half the growth. Opening the first-run disclosure walked the submit
button out from under a pointer already resting on it.

Centering and stability look like a straight trade, and are not. The offset is
now half the viewport minus a **constant** half-height rather than minus the
card's own: `pt-[max(2rem,calc(50vh-17.5rem))]`. Because the figure is a
constant, growth moves nothing (measured: 0px, against a card going 537px to
612px), and the card still reads as centered. Measured off the running page:

| state | card | off true center |
| --- | --- | --- |
| master key, 1 footer link | 537px | 12px high |
| master key, 2 links | 581px | 12px low |
| password, 1 link | 589px | 14px low |
| password, 4 links | 721px | 80px low, but it nearly fills a 900px window |
| sign-in unavailable | 351px | 9px high |

The unavailable card has its own constant: at 351px it is far enough from the
form that sharing the figure left it 110px high.

`vh`, not `dvh` ; the dynamic unit shrinks when a phone's soft keyboard opens,
which would re-center the card at the moment someone is typing into it. `max()`
floors the offset on a window shorter than the card, where the page scrolls.

`<ErrorBanner>` is gone from this form. It inserted about 46px between the last
field and the button, in the same spot. A refusal now renders on the field's own
label row: label and required marker grouped left, the message right, sharing
the existing 20px line, so a refusal costs no height at all. `role="alert"` is
kept, and a message too long for the row (the gateway's master-key retirement
notice) wraps rather than being clipped. **`ErrorBanner` itself is untouched**;
other screens use it.

## Visual and spacing

- Heading to 24px Zilla Slab 600, `tracking-tight` ; the `text-display` role.
- The guide's tinted code block becomes inline `<code>` chips on a 13px line,
  with shorter copy: "No `OTARI_MASTER_KEY` set? Otari printed one to the
  server logs on startup. Find it with `docker logs <container>`."
- The `border-t` above the footer links is gone. At 1.38:1 it was separating
  two things nobody was confusing.
- The security note is one line of copy, and the same line for both branches.
- Every section boundary reads **24px on screen**. A 44px hit-area row carries
  12px of invisible padding above and below its text, so the column beside one
  runs at 12px to land on the same 24px, and the card's padding is optically
  symmetric: 49px top against 37px + a 12px tail on the desk, 41px against
  29px + 12px on a phone.

One substitution from the brief: the chips take **`bg-surface-alt`**, not
`bg-surface-muted`. `--color-surface-muted` is declared in both theme blocks but
registered in `@theme` only as `--color-surface-alt`, so `bg-surface-muted`
compiles to nothing at all ; the dead-class trap `design-tokens.md` warns about.
Same token, live utility.

## Checks

- `pnpm --dir web run lint`, `pnpm --dir web run typecheck` clean
- `pnpm --dir web test` ; 1211 passed (85 files)
- Verified in a real browser against a local gateway at 1280px and 390px, in
  both themes, across the master-key and password branches, with the
  disclosure open and closed and with a refusal on the row. Geometry above is
  measured from the running page, not from the design.
- No OpenAPI, Postman, client-schema or route-tree artifacts affected. The
  dashboard bundle is gitignored.

## Test changes

The test at `Login.test.tsx` asserting the button stays disabled until both
halves are typed no longer describes the intent, so it is replaced by one
asserting that submitting an empty credential does not POST, then names the
missing box, then succeeds once both halves are there. Three more cover the
empty master key (whitespace included), the reveal toggle round-tripping the
input's `type`, and nothing being focused on mount.

## Follow-ups, deliberately not in this PR

- **The rest of the dashboard's inputs are still 14px**, so iOS Safari zooms on
  focus everywhere else too. Same one-line fix per field, but it touches every
  form in `web/src` and belongs in its own PR.
- `PublicAuthLayout` still draws its card the old way: 18px heading, `p-7`,
  vertically centered, `border-t` footer. Signup, verification and password
  reset render into it, so the auth surface now reads as two designs. Bringing
  that layout onto this treatment is the obvious next step.

## Summary

- Redesigned the login screen for consistent spacing, styling, and layout across all sign-in states.
- Improved accessibility with clearer validation, stable inline errors, decorative logo treatment, and accessible controls.
- Added master-key visibility toggling and removed automatic input focus.
- Kept the initial sign-in action enabled while validating incomplete submissions locally.
- Updated screenshot and unit tests for validation, visibility toggling, whitespace-only master keys, and initial focus behavior.
- Verified lint, typecheck, and all 1,211 tests across 85 files.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

None. #557's sign-out guard is deliberately preserved rather than changed, and
its test still covers it.

## Checklist

- [x] I have added or updated tests that cover my change. This is a `web/`-only
      change, so they are Vitest rather than `tests/unit` / `tests/integration`:
      `web/src/features/auth/Login.test.tsx` replaces the test that asserted the
      old disabled-button behavior and adds four covering the empty-credential
      guard, whitespace-only keys, the reveal toggle, and nothing being focused
      on mount.
- [x] I ran the Definition of Done checks locally. `make lint` and
      `make typecheck` pass. `make test` reports 19 failures, all in Python
      modules this PR does not touch (master-key service, dashboard sessions,
      config env loading, model aliases and discovery, batches); a change to a
      React component cannot reach them. The checks that do cover this change
      are green: `pnpm --dir web run lint`, `pnpm --dir web run typecheck`, and
      `pnpm --dir web test` (1211 passed, 85 files).
- [x] Documentation was updated where necessary: nothing here changes behavior
      that the docs describe.
- [x] If the API contract changed, I regenerated the OpenAPI spec. Not
      applicable, no route or schema changed.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

The brief specified `bg-surface-muted` for the code chips; that utility is not
registered anywhere in `globals.css` and would have compiled to nothing, so the
chips use `bg-surface-alt`, the live utility for the same
`--color-surface-muted` token. The brief also specified a `⚠` glyph, which
renders as a color emoji on Apple platforms and would have ignored
`text-danger`, so the mark is an inline SVG on `currentColor`. Both substitutions
are noted in the diff's comments. Geometry quoted above is measured from the
page running against a local gateway, not read off a design or inferred.

### DIFF
--- a/web/src/features/auth/Login.tsx
+++ b/web/src/features/auth/Login.tsx
@@ -21,6 +21,14 @@ type CredentialField = "email" | "password" | "masterKey"
  */
 const VALIDATION = "aria" as const
 
+const EMAIL_PATTERN = /^[^@\s]+@[^@\s]+\.[^@\s]+$/
+
+const ERROR_IDS: Record<CredentialField, string> = {
+  email: "login-email-error",
+  password: "login-password-error",
+  masterKey: "login-master-key-error",
+}
+
 /**
  * The page frame: optically centered, and stable while the card grows.
  *
@@ -149,15 +157,24 @@ function DisclosureCaret() {
  * The row above a credential box: label and required marker grouped left, the
  * field's refusal right, both on the label's existing 20px line. This is where
  * an `<ErrorBanner>` used to go, between the last field and the button, and it
- * inserted about 46px right where the pointer already was. On the label's line
- * a refusal costs no height at all for the messages that fit; one long enough
- * not to fit wraps to its own line rather than being clipped, which the
- * gateway's master-key retirement notice does.
+ * inserted about 46px right where the pointer already was. The row stays one
+ * line high so a refusal cannot move the submit button. Sighted users get a
+ * truncated message, while assistive technology retains the complete detail.
  */
-function LabelRow({ label, error }: { label: string; error: unknown }) {
+function LabelRow({
+  label,
+  error,
+  errorId,
+}: {
+  label: string
+  error: unknown
+  errorId: string
+}) {
+  const message = error ? errorMessage(error) : null
+
   return (
-    <div className="flex min-h-5 flex-wrap items-center justify-between gap-x-3">
-      <span className="flex items-center">
+    <div className="flex h-5 items-center justify-between gap-x-3">
+      <span className="flex shrink-0 items-center">
         {/* `text-foreground` holds the label at its resting ink while the field
             is invalid: HeroUI turns any `.label` under `[data-invalid]` red, and
             a red label beside a red message says the same thing twice. */}
@@ -173,13 +190,19 @@ function LabelRow({ label, error }: { label: string; error: unknown }) {
           *
         </span>
       </span>
-      {error ? (
+      {message ? (
         <span
+          id={errorId}
           role="alert"
-          className="flex items-center gap-1 text-sm text-danger"
+          className="flex min-w-0 flex-1 items-center justify-end gap-1 text-sm text-danger"
         >
           <AlertIcon />
-          <span>{errorMessage(error)}</span>
+          <span
+            aria-hidden="true"
+            data-message={message}
+            className="min-w-0 truncate after:content-[attr(data-message)]"
+          />
+          <span className="sr-only">{message}</span>
         </span>
       ) : null}
     </div>
@@ -269,6 +292,10 @@ export function Login() {
         fail("email", "Enter your email.")
         return null
       }
+      if (!EMAIL_PATTERN.test(email.trim())) {
+        fail("email", "Enter a valid email address.")
+        return null
+      }
       if (!password) {
         fail("password", "Enter your password.")
         return null
@@ -329,12 +356,10 @@ export function Login() {
           <Card.Content className={CARD_FLAT}>
             {/* Grouped so the mark sits 16px from the heading it belongs to,
                 the way it does on the form. alt="" because the <h1> under it
-                already names the product: a screen reader announcing "Otari,
-                heading Sign-in is unavailable" reads the mark as a word of the
-                sentence. */}
+                already names the product. */}
             <div className="flex flex-col items-center gap-4">
               <img src="/favicon.svg" alt="" className="h-10 w-11" />
-              <h1 className={HEADING}>Sign-in is unavailable</h1>
+              <h1 className={HEADING}>Otari sign-in is unavailable</h1>
             </div>
             <p className="text-sm text-muted">
               This gateway cannot start a session at the moment, which usually
@@ -375,6 +400,7 @@ export function Login() {
               password branch has no such row and spaces at the full 24px. */}
           <form
             className={`flex flex-col ${usesPassword ? "gap-6" : "gap-3"}`}
+            noValidate
             onSubmit={(event) => {
               event.preventDefault()
               void submit()
@@ -397,6 +423,7 @@ export function Login() {
                   <LabelRow
                     label="Email"
                     error={errorField === "email" ? error : null}
+                    errorId={ERROR_IDS.email}
                   />
                   {/* 16px, not the 14px HeroUI drops to from `sm:` up: under
                       16px iOS Safari zooms the page on focus. No autoFocus,
@@ -405,6 +432,9 @@ export function Login() {
                   <Input
                     placeholder="you@example.com"
                     autoComplete="username"
+                    aria-describedby={
+                      errorField === "email" ? ERROR_IDS.email : undefined
+                    }
                     className="h-11 text-base"
                   />
                 </TextField>
@@ -423,9 +453,13 @@ export function Login() {
                   <LabelRow
                     label="Password"
                     error={errorField === "password" ? error : null}
+                    errorId={ERROR_IDS.password}
                   />
                   <Input
                     autoComplete="current-password"
+                    aria-describedby={
+                      errorField === "password" ? ERROR_IDS.password : undefined
+                    }
                     className="h-11 text-base"
                   />
                 </TextField>
@@ -447,11 +481,17 @@ export function Login() {
                   <LabelRow
                     label="Master key"
                     error={errorField === "masterKey" ? error : null}
+                    errorId={ERROR_IDS.masterKey}
                   />
                   <div className="relative">
                     <Input
                       placeholder="otari-mk-…"
                       autoComplete="off"
+                      aria-describedby={
+                        errorField === "masterKey"
+                          ? ERROR_IDS.masterKey
+                          : undefined
+                      }
                       fullWidth
                       className="h-11 pr-11 font-mono text-base"
                     />
--- a/web/src/features/auth/Login.tsx
+++ b/web/src/features/auth/Login.tsx
@@ -3,10 +3,161 @@ import { useState } from "react"
 import { useAuth } from "@/features/auth/AuthContext"
 import type { SignInCredential } from "@/shared/api/client"
 import { createSession } from "@/shared/api/client"
-import { ErrorBanner } from "@/shared/components/ui"
+import { errorMessage } from "@/shared/components/ui"
 import { useDeployment } from "@/shared/hooks/useDeployment"
 
 import { PublicAuthLink } from "./PublicAuthLayout"
+
+/** Which box an error belongs beside. */
+type CredentialField = "email" | "password" | "masterKey"
+
+/**
+ * `aria` rather than react-aria's default `native`, which puts a `required`
+ * attribute on the input and lets the browser cancel the submit event outright.
+ * That is a correct refusal announced in the wrong place: an unstyleable bubble
+ * saying "Please fill out this field", gone on the next scroll, instead of the
+ * message this form renders on the field's own label row. The field still
+ * carries `aria-required`, which is the part a screen reader reads.
+ */
+const VALIDATION = "aria" as const
+
+/**
+ * The page frame both states share: anchored to a fixed offset from the top
+ * rather than vertically centered. Centering means anything that grows the card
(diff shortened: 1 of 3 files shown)


web/src/features/auth/Login.tsx
```diff
@@ -130,25 +352,29 @@ export function Login() {
...
               </p>
             </div>
           </div>
 
+          {/* Section boundaries read 24px on screen throughout. A 44px row
+              carries 12px of invisible padding above and below its own text, so
+              the column next to one runs at 12px to land on the same 24px: the
+              master-key branch has the disclosure's summary row in it, and the
+              password branch has no such row and spaces at the full 24px. */}
           <form
```
jigjigjig, line 399: `validationBehavior="aria"` does not disable the browser’s `type="email"` constraint. Entering a nonempty malformed address causes native validation to cancel the submit event before this handler runs, so the browser bubble returns instead of the new `LabelRow` error. Add `noValidate` here, validate the email in `readCredential`, and add a test that it renders the inline error without issuing a request.

jigjigjig (commented):
Reviewed by Codex.

The PR Template Check is failing because the description is missing the required `## PR Type`, `## Checklist`, and `## AI Usage` sections.

web/src/features/auth/Login.tsx
jigjigjig: Addressed in 1f29374: the form now disables native validation, validates malformed email addresses locally, and covers the no-request path.

jigjigjig:
Addressed CodeRabbit feedback in 1f29374:

- Added local malformed-email validation and disabled native form validation.
- Linked each field error to its input with stable IDs and `aria-describedby`.
- Kept error rows one line high, with visually truncated text and full assistive-technology detail.
- Named Otari in the unavailable-state heading.

The PR description now includes the required `## PR Type`, `## Checklist`, and `## AI Usage` sections.

web/src/features/auth/Login.tsx
```diff
@@ -241,37 +513,44 @@ export function Login() {
...
+                  Need a new verification link?
+                </PublicAuthLink>
+              ) : null}
+              {/* Not a `PublicAuthLink`: `/welcome` is a page the gateway
+                  serves, so this one really is a navigation and not a hash
+                  change. Sized to match the links above it. */}
+              <Link
+                href="/welcome"
+                className="inline-flex min-h-11 items-center text-sm hover:text-link-hover"
+              >
```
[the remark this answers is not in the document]
jigjigjig: Addressed in c769176: the `/welcome` link now uses the same base color and weight as the public-auth links.

web/src/features/auth/Login.tsx
```diff
@@ -3,11 +3,212 @@ import { useState } from "react"
...
+          className="flex min-w-0 flex-1 items-center justify-end gap-1 text-sm text-danger"
+        >
+          <AlertIcon />
+          <span
+            aria-hidden="true"
+            data-message={message}
+            className="min-w-0 truncate after:content-[attr(data-message)]"
+          />
+          <span className="sr-only">{message}</span>
+        </span>
```
khaledosman, line 204: The fixed `h-5` row (line 176) plus `truncate` clips the gateway's own refusals with no way to read them. `MASTER_KEY_SIGN_IN_RETIRED` (`src/gateway/api/routes/auth_session.py:65`) is 190 characters; on a 20px line beside the label an operator sees about `Master-key sign-in is ret…` and never sees "Sign in with your email and password", which is the entire instruction on the one path where they are stuck. Commit 1 let it wrap; commit 3 traded that away for a stability guarantee the constant top offset already gives the card's top edge (only the button moves).

Let the message wrap (`min-h-5 flex-wrap`, as before), or render an over-long one under the field, and add `title={message}` either way.

Same lines: the three-element construction is not needed. A screen reader reads the full text of a `truncate`d node, so

```tsx
<span className="min-w-0 truncate" title={message}>{message}</span>
```

replaces the `aria-hidden` + `after:content-[attr(data-message)]` span and the `sr-only` duplicate, and stops the visible text living in a CSS `content` value.

web/src/features/auth/Login.tsx
```diff
@@ -3,11 +3,212 @@ import { useState } from "react"
...
+
+/**
+ * The same card for the unavailable state, which ends in a paragraph rather
+ * than a 44px row, so it has no invisible 12px to borrow and pays the padding
+ * itself.
+ */
+const CARD_FLAT = "flex flex-col gap-6 px-6 py-6 text-center sm:px-8 sm:py-8"
+
+/** 24px Zilla Slab. The screen's one page-defining line, so `text-display`. */
+const HEADING = "text-display font-semibold tracking-tight"
```
khaledosman: This re-skins the role it just picked: `@utility text-display` is `font-weight: 700` with `letter-spacing: 0`, and `globals.css:602` says the zero is deliberate ("neutral for the Zilla Slab slab serif"). Non-negotiable 3 puts type in the seven roles, so if 600 with tight tracking is the intent for a display line, change the `text-display` block once rather than overriding it at this call site. Same shape at line 527, where `text-caption leading-relaxed` overrides that role's 1.45.

web/src/features/auth/Login.tsx
```diff
@@ -3,11 +3,212 @@ import { useState } from "react"
...
+ * `aria` rather than react-aria's default `native`, which puts a `required`
+ * attribute on the input and lets the browser cancel the submit event outright.
+ * That is a correct refusal announced in the wrong place: an unstyleable bubble
+ * saying "Please fill out this field", gone on the next scroll, instead of the
+ * message this form renders on the field's own label row. The field still
+ * carries `aria-required`, which is the part a screen reader reads.
+ */
+const VALIDATION = "aria" as const
+
+const EMAIL_PATTERN = /^[^@\s]+@[^@\s]+\.[^@\s]+$/
```
khaledosman, line 27: Character for character the gateway's `_ADDRESS_PATTERN` (`src/gateway/core/addresses.py:17`), which is the shape check every stored address already passed through `validated_email`. Nothing connects the two copies, so drift on either side either locks a stored address out of the dashboard or accepts one the gateway will refuse. Add the pointer in a comment, the way `core/addresses.py` documents being the single answer.

khaledosman (approved):
Approving; finding 1 is worth fixing before merge but does not need to block a design PR.

Two notes that have no line in this diff to sit on:

- `web/e2e/screenshots/public.spec.ts:21-25` still documents the layout this PR removes ("it adds a banner above the form"), and its `getByText(/invalid|unauthor/i)` now resolves to the `sr-only` copy of the message rather than to anything an operator sees. Worth repointing at the label row's `role="alert"` while the behavior is being changed.
- The PR body uses em dashes. Squash-merge puts that body in the commit message, where `AGENTS.md`'s prose rule applies.

Checked and clear: no OpenAPI, Postman, client-schema or route-tree artifact is owed, no new page so no screenshot entry, the #557 sign-out guard and its test are intact, and the file's `afterEach` restores the `fetch` spy. The `h-11` / `text-base` overrides are the right rung after all: `Input` has no `size` variant, `.button--lg` drops to 40px at `md`, and HeroUI's component CSS sits in `@layer components`, so the call-site utility does win.

🤖 Reviewed with [Claude Code](https://claude.com/claude-code) (Claude Opus 5)

web/src/features/auth/Login.tsx
jigjigjig: Addressed in 04b7757: gateway errors wrap with their full visible instruction, and the message has a title.

web/src/features/auth/Login.tsx
jigjigjig: Addressed in 04b7757: headings and the guide now use their semantic type roles without call-site overrides.

web/src/features/auth/Login.tsx
jigjigjig: Addressed in 04b7757: the mirrored email shape check now points to the gateway source of truth.

jigjigjig:
Addressed the remaining pre-merge notes in 04b7757:

- Screenshot coverage now waits for the inline alert, not hidden implementation text.
- The PR description now follows the repository prose punctuation rule.

PR #735, opened 2026-08-24
fix(dashboard): stop the bulk action bar sliding in from the left

### LINKED ISSUE #734: fix(dashboard): bulk action bar slides in from the left instead of appearing centered
## What happens

Selecting one or more rows in any dashboard table makes the bulk action bar appear off to the left of where it belongs, then slide right into its centered position. It reads as a glitch rather than as an intentional entrance animation.

Affected pages (every `BulkActionBar` consumer): Keys, Models, Budgets, Activity.

## Steps to reproduce

1. Open the dashboard and go to any table page with row selection (for example API keys).
2. Tick the checkbox on a row.
3. Watch the bar at the bottom of the viewport: it renders roughly half its own width to the left of center and then moves to center over ~160ms.

## Cause

`BulkActionBar` (`web/src/shared/components/BulkActionBar.tsx`) centers itself with `left-1/2 -translate-x-1/2`, and its entrance animation lives in `web/src/styles/globals.css`:

```css
.otari-bulk-bar {
  animation: otari-bulk-bar-in 160ms ease-out;
}
@keyframes otari-bulk-bar-in {
  from { opacity: 0; transform: translate(-50%, 0.5rem); }
  to   { opacity: 1; transform: translate(-50%, 0); }
}
```

Under Tailwind v4, `-translate-x-1/2` no longer compiles to `transform`; it compiles to the separate `translate` property:

```css
.-translate-x-1\/2 {
  --tw-translate-x: calc(calc(1 / 2 * 100%) * -1);
  translate: var(--tw-translate-x) var(--tw-translate-y);
}
```

(verified by compiling the utility against the pinned `tailwindcss@4.3.3`)

`translate` and `transform` both apply and compose, so while the animation runs the bar is shifted by `-50%` twice, i.e. `-100%`. The keyframes have no `forwards` fill, so at the end of the 160ms the `transform` drops and the bar snaps back to the correct `-50%`. Hence: appears left, then moves centered.

## Suggested fix

Make the centering and the animation use the same property, so they don't stack. Either:

- keep the Tailwind `-translate-x-1/2` and animate only the properties it doesn't own (`opacity`, plus vertical offset via `translate` or `--tw-translate-y`), or
- drop `-translate-x-1/2` and do the whole centering inside the keyframes' `transform`.

Worth adding a regression check while we're in there: the existing `BulkActionBar.test.tsx` covers behavior, not the computed offset, so this class of stacking bug can come back silently on the next Tailwind bump.

## Notes

`prefers-reduced-motion: reduce` disables the animation, so the bar is correctly centered from the first frame there. The bug is only visible with motion enabled.

jigjigjig:
## Description

Selecting rows in any dashboard table made the bulk action bar appear about half its own width to the left of center and then slide into place, reading as a glitch rather than an entrance animation. Affected every `BulkActionBar` consumer: Keys, Models, Budgets, Activity.

`BulkActionBar` centers itself with Tailwind's `-translate-x-1/2`, which under Tailwind v4 compiles to the `translate` property rather than `transform`:

```css
.-translate-x-1\/2 {
  --tw-translate-x: calc(calc(1 / 2 * 100%) * -1);
  translate: var(--tw-translate-x) var(--tw-translate-y);
}
```

The `otari-bulk-bar-in` keyframes set `transform: translate(-50%, …)`, so both properties applied and composed into a -100% offset for the 160ms the animation ran; with no `forwards` fill the bar then snapped back to the correct -50%. The fix animates `translate` instead, so the centering and the entrance share one property.

CSS-only: two changed declarations in `web/src/styles/globals.css`, plus a comment recording why `transform` is the wrong property here, since reaching for it again would reintroduce the bug.

`prefers-reduced-motion: reduce` already disabled the animation, so that path was never affected.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Closes #734

## Checklist

- [ ] I have added or updated tests that cover my change (`tests/unit`, `tests/integration`). **No tests added, at the author's explicit request to keep the PR minimal.** For the record, this bug class is not covered: `BulkActionBar.test.tsx` asserts rendering and behavior, not computed offsets, so a jsdom test would not have caught it and would not catch a recurrence. Catching it needs a browser-rendered check (the screenshot suite), which is on-demand and gitignored while the dashboard is mid-migration.
- [ ] I ran the Definition of Done checks locally (`make lint`, `make typecheck`, `make test`). Ran Biome on the changed file (clean). Did not run the Python-side `make lint`/`typecheck`/`test`, which cover `src/gateway/` and are untouched by a `web/src/styles/globals.css` change; leaving this unchecked rather than implying the full suite was run.
- [x] Documentation was updated where necessary. (Not needed; the reasoning lives in a comment next to the keyframes.)
- [x] If the API contract changed, I regenerated the OpenAPI spec. (No API change.)

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

Claude Code, Claude Opus 5 (1M context)

The bug was reported from observed behavior; the Tailwind v4 `translate`-vs-`transform` cause was found by compiling the `-translate-x-1/2` utility against the repo's pinned `tailwindcss@4.3.3` and reading the emitted property, rather than inferred. Branch was built in a worktree off `origin/main` so the author's unrelated in-progress work stayed out of the diff.

### DIFF
--- a/web/src/styles/globals.css
+++ b/web/src/styles/globals.css
@@ -960,14 +960,18 @@ h6 {
 .otari-bulk-bar {
   animation: otari-bulk-bar-in 160ms ease-out;
 }
+/* Animates `translate`, not `transform`: the bar centers itself with Tailwind's
+ * `-translate-x-1/2`, which under Tailwind v4 sets the `translate` property, and
+ * a `transform` here would compose with it into a -100% offset for the length of
+ * the animation, so the bar slid in from half its width to the left. */
 @keyframes otari-bulk-bar-in {
   from {
     opacity: 0;
-    transform: translate(-50%, 0.5rem);
+    translate: -50% 0.5rem;
   }
   to {
     opacity: 1;
-    transform: translate(-50%, 0);
+    translate: -50% 0;
   }
 }
 


jigjigjig:
Reviewed by Codex: no actionable comments.

khaledosman (approved):
Verified the root cause against the pinned `tailwindcss@4.3.3` rather than taking it on faith: `-translate-x-1/2` compiles to `--tw-translate-x: calc(calc(1 / 2 * 100%) * -1); translate: var(--tw-translate-x) var(--tw-translate-y)`, and `--tw-translate-y` is registered with `initial-value: 0`. The static state is therefore exactly `translate: -50% 0`, which the new `to` frame matches, so the absence of a fill mode is correct here and there is no snap when the animation ends.

Swept the rest of the file for the same composition hazard: `otari-detail-open` is the only other keyframes block and animates grid tracks plus opacity, no transform property. The two `transition-transform` call sites (`AppShell.tsx:559`, `SettingsPage.tsx:80`) are unaffected, since v4 expands that utility to `transform, translate, scale, rotate`.

Biome on the head revision of the file: no formatting drift and no new diagnostics versus `main`. No generated artifacts owed (no route or schema touched, no `schema.ts` or `routeTree.gen.ts` change, bundle uncommitted, no new page so no screenshot entry).

No findings. The test gap you flagged is real but not actionable at the rung available: jsdom does not compute keyframes, and the screenshot suite is on-demand with gitignored baselines.

---
🤖 Review by [Claude Code](https://claude.com/claude-code) (Claude Opus 5, 1M context)

PR #744, opened 2026-08-24
feat(gateway): let a workspace configure its own web search

### LINKED ISSUE #656: Per-workspace web search configuration

njbrake:
## Description

Per-workspace web search configuration, the third of the four config planes under #655/#678. A workspace gets a `workspace_web_search_configs` row, or none, managed over `/v1/workspaces/{id}/web-search` and resolved at admission in `prepare_gateway_tools`. The row holds no URL and no credential: the backend stays on `/v1/tool-settings` and the `/v1/search` tools stay on `/v1/search-tools`. No row means no narrowing, asserted rather than assumed.

Three things need a reviewer's eye.

**Standalone narrows where the hybrid arm defaults.** Hybrid applies the policy it resolves from otari.ai as defaults a request overrides; that is the platform's contract and is untouched. It fails open under this repo's rule, though: a request sheds a workspace's block-list by sending a block-list of its own. So the standalone arm unions `blocked_domains`, intersects `allowed_domains`, and floors `max_results`. Only `purpose_hint` and the opaque `provider_options` bag fill in rather than narrow, because neither permits anything. An empty allow-list intersection is a 403, not `[]`, since `_build_web_search_backend` reads an empty list as *no* allow-list.

**`POST /v1/search` honors the veto too**, and only the veto. It is the second door into the same capability, so leaving it open would make the switch bypassable by any key in the workspace. The row's other fields shape the in-loop backend's request and have no counterpart in that endpoint's provider adapters.

**One fix reaches past this surface.** Writing a workspace `max_results` into the tool entry *raised* the operator's own cap, because the tool entry beats `web_search_max_results`. It now floors against `web_search_max_results_baseline(config)`, extracted from `_build_web_search_backend`; `MAX_RESULTS_CAP` and `DEFAULT_MAX_RESULTS` become public for it.

Ported from otari-ai's `WorkspaceWebSearchConfig` with three deliberate divergences: `enabled` is required on the write (absence of a row means unnarrowed here and disabled there, so either default surprises somebody), the `provider` column is dropped (the operator picks the backend by URL, so a per-workspace provider is inert or a widening), and reads take the management role because the row is the workspace's posture rather than one member's allowance.

## PR Type

- [x] New Feature

## Relevant issues

Fixes #656

## Checklist

Full pytest run was 3851 passed / 17 skipped against a local PostgreSQL, plus the OSS-edition smoke under `--no-dev`, `pnpm lint` / `tsc --noEmit` / vitest (1222 passed), and `make openapi-check` / `make postman-check`. `alembic heads` is a single head, `d8b3e5c1f7a2` onto `c4d7f2a9b6e1`.

## AI Usage

**AI Model/Tool used:**

Implemented from #656 against the two worked examples already in main (#724, #725), with otari-ai's own `WorkspaceWebSearchConfig` read for the port. The composition rule and the two divergences above were the decisions; the rest follows the sibling planes.

## Summary

- Added per-workspace web search settings for standalone deployments.
- Added workspace web search management endpoints and dashboard controls.
- Applied workspace enablement, result limits, purpose hints, and domain rules to search requests.
- Preserved deployment-wide backend and credential settings.
- Kept existing behavior when a workspace has no custom settings.
- Added backend, frontend, integration, unit, documentation, and OpenAPI coverage.

## Technical notes

- Workspace settings store no credentials.
- Workspace limits can only narrow deployment limits.
- Disabled workspaces and unresolved workspace settings fail closed.
- `POST /v1/search` uses the same workspace enforcement rules.


src/gateway/services/tenancy/workspace_web_search_service.py
```diff
@@ -0,0 +1,516 @@
...
+
+def _union(requested: list[str] | None, workspace: tuple[str, ...]) -> list[str]:
+    """Every domain either side named, in request-then-workspace order, de-duplicated."""
+    merged: dict[str, None] = {}
+    for host in (*(requested or ()), *workspace):
+        merged.setdefault(host, None)
+    return list(merged)
+
+
+def _intersect(requested: list[str], workspace: tuple[str, ...]) -> list[str]:
```
khaledosman, line 404: `_intersect` compares domains as exact strings, but the backend matches by suffix: `web_search_backend.py:386` keeps a host when `host == d or host.endswith("." + d)`, so a stored `python.org` covers `docs.python.org`.

That makes a request which is *strictly narrower* than its workspace's allow-list get refused:

- workspace `allowed_domains: ["python.org"]`, request `allowed_domains: ["docs.python.org"]` → empty intersection → 403 `WorkspaceWebSearchDomainsExcludedError` (line 321-323), even though every result the request could get was already permitted by the workspace.
- The reverse is wrong too: workspace `["docs.python.org"]`, request `["python.org"]` → 403, where the correct narrowing is `["docs.python.org"]`.

Fix: treat a pair as intersecting when either side covers the other under the same `==` / `endswith("." + d)` rule, and keep the more specific of the two.

`test_web_search_narrowing.py` only exercises disjoint and exactly-equal labels, so neither case is caught today.

src/gateway/services/tenancy/workspace_web_search_service.py
```diff
@@ -0,0 +1,516 @@
...
+    ``None`` rather than ``[]``, because an empty list here would read as "an
+    allow-list permitting nothing" to a reader and as "no allow-list" to
+    :func:`narrow_web_search_tool_entry`, and only one of those is what a
+    cleared form means.
+    """
+    if value is None:
+        return None
+    seen: dict[str, None] = {}
+    for raw in value:
+        host = raw.strip().lower()
```
khaledosman: `_normalize_domains` only strips, lower-cases, de-dupes and length-checks, so `https://evil.example`, `evil.example/path` and `.evil.example` are all stored verbatim. None of them can equal a `urlparse(...).hostname` or suffix-match one at `web_search_backend.py:384-386`.

A `blocked_domains` entry entered that way blocks nothing, silently: the row reads as configured on the dashboard and in the `GET` response while the guardrail is inert. Allow-lists fail the other way, where one such entry filters every result out.

Reject anything containing `/`, `:`, whitespace, or a leading `.` at the write, or strip to the host before storing. `parseDomains` in `web/src/features/tools/WorkspaceWebSearchCard.tsx:59` should do the same so the error lands on the field rather than arriving as a 422 banner.

khaledosman (approved):
Two findings inline, both in `workspace_web_search_service.py`'s domain handling. Neither blocks.

Non-blocking, on the point you flagged for review: the `/v1/search` veto (`search.py:305`) gates *every* tool configured under `/v1/search-tools`, not just the SearXNG-backed `otari_web_search` path. An operator who sets `enabled: false` to stop the in-loop tool also silently kills that workspace's Tavily/Exa calls. It is the right call for bypassability and `docs/tools.md` says so, but `WorkspaceWebSearchConfigUpdate.enabled`'s description ("False refuses web search for this workspace") and the card's "Blocked" label do not, and that is where an admin will read it.

Generated artifacts check out: `openapi.json` and the Postman collection both carry `/v1/workspaces/{workspace_id}/web-search`, `sdk-endpoints.txt` is updated, the router is standalone-only with the hybrid-surface test extended, and no `routeTree.gen.ts` change is owed. Migration `d8b3e5c1f7a2` is a single head onto `c4d7f2a9b6e1`.

---
🤖 Review by [Claude Code](https://claude.com/claude-code) (Claude Opus 5), posted by @njbrake's reviewer.

src/gateway/services/tenancy/workspace_web_search_service.py
njbrake: Real, and fixed. `_intersect` now treats a pair as overlapping when either side covers the other under the backend's own `==` / `endswith("." + d)` rule, and keeps the more specific of the two; genuinely disjoint lists still intersect to nothing and still raise.

Your two cases, run against the current head:

- workspace `["python.org"]`, request `["docs.python.org"]` gives `["docs.python.org"]`
- workspace `["docs.python.org"]`, request `["python.org"]` gives `["docs.python.org"]`

You were right that the tests only covered disjoint and exactly-equal labels. There are four new unit cases (both directions, the silent case where an overlapping entry vanished because a sibling matched exactly, and `notexample.com` which is not under `example.com`), plus an integration case so the subdomain path is exercised end to end through admission rather than only at the helper.

src/gateway/services/tenancy/workspace_web_search_service.py
njbrake: Real, and fixed. This was the one that worried me most of the three, for the reason you gave: on a block-list it read as configured while blocking nothing.

`_normalize_domains` now refuses any entry carrying `/`, `:`, `@`, `?`, `#`, `*`, a backslash or whitespace, so `https://evil.example`, `evil.example/path` and `evil.example:8080` are all rejected at the write with a message naming the entry. `parseDomains` in the card applies the same rule and puts the error on the field, so it does not arrive as a 422 banner.

One deliberate difference from what you suggested: a leading dot is stripped rather than refused. An entry here already covers its subdomains, so `.example.com` has exactly one reading and it is the rule this surface already applies, written in cookie syntax. Refusing it would be turning a correct intent into an error.

alembic/versions/d8b3e5c1f7a2_add_workspace_web_search_configs.py
```diff
@@ -0,0 +1,58 @@
...
+        _TABLE,
+        sa.Column("workspace_id", sa.Uuid(), nullable=False),
+        sa.Column("enabled", sa.Boolean(), nullable=False),
+        sa.Column("max_results", sa.Integer(), nullable=True),
+        sa.Column("purpose_hint", sa.Text(), nullable=True),
+        sa.Column("allowed_domains", sa.JSON(), nullable=True),
+        sa.Column("blocked_domains", sa.JSON(), nullable=True),
+        sa.Column("provider_options", sa.JSON(), nullable=True),
+        sa.Column("created_at", sa.DateTime(timezone=True), nullable=False),
+        sa.Column("updated_at", sa.DateTime(timezone=True), nullable=False),
```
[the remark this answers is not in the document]
njbrake: Right, and now fixed. I declined this first time round and the reason I gave was wrong, so it is worth correcting on the record: I claimed no table-creating migration in the chain carries a `server_default`. Five of twenty-three do, and one of them is `c4b6d8e0f2a3`, which creates `workspace` itself and gives every `created_at` a `server_default=sa.func.now()`. That is the table this one hangs its foreign key off.

`enabled`, `created_at` and `updated_at` now carry defaults, mirrored on the ORM columns so autogenerate sees no drift. `sa.func.now()` rather than a dialect literal, since the chain runs on SQLite too.

Verified on a fresh database rather than assumed: `INSERT INTO workspace_web_search_configs(workspace_id) VALUES (...)` now succeeds and lands `true` / `now()` / `now()`, where before it failed on three not-null violations. Downgrade still drops the table cleanly.

The guideline you cited is written about existing rows, which a `create_table` has none of. The reason to do it anyway is the one in your comment: inserts that are not this service's ORM mapping.

Issue #676, opened 2026-08-20
[FEATURE] One period derivation, and calendar alignment for workspace budget defaults


njbrake:
### Problem or use case

Two things arrive together when #671 lands, both about where a budget period comes from.

`WorkspaceBudgetDefault` has `budget_duration_sec` and no `reset_alignment`, so a default can only ever materialize a rolling ceiling. otari.ai's own defaults are the opposite: per mozilla-ai/otari-ai#1741, the platform has no rolling mode at all and `calendar_window` is its only period model. A default ported from there cannot express what it meant, and #640 added the column that would let it.

Separately, there are two derivations of a period again. #640 made `scoped_budget_service.period_window()` the single one, shared by creation, retiming, and the roll at the gate, precisely so a window cannot be computed two ways. #671 adds `workspace_budget_default_service._period_window()`, which is the same arithmetic minus alignment. That is not an oversight: importing `scoped_budget_service` from `services/tenancy/` closes an import cycle (`scoped_budget_service` imports `workspace_scope`, which imports `tenancy.provisioning_service`), which `workspace_service.py` already documents at its own `_delete_scoped_budgets_for`, and `tests/unit/test_service_module_imports.py` pins. So the fix is a shared home, not a call.

### Proposed solution

Move the derivation into a module both can import and that imports neither, the way `core/sql.py` already holds conditions a route and a service both need, and have both call it. Then add a nullable `reset_alignment` to the defaults table, with the same CHECK that keeps a row to one period source, and carry it into the `ScopedBudget` each default materializes.

### Alternatives considered

**Leave the local duplicate.** It drifts in one direction only: the next alignment added to one is silently absent from the other, and the symptom is a ceiling that resets on a cadence nobody configured.

**Break the cycle instead, so the defaults service can import the service directly.** Heavier than moving one function, and it relitigates a layering decision that is already documented and tested.

### Definition of Done

- One period derivation, reached from both call sites, with `tests/unit/test_service_module_imports.py` still passing.
- A default carrying an alignment materializes a calendar-aligned ceiling, tested across a boundary.
- The CHECK on the defaults table refuses a row with both period sources, matching `scoped_budgets`.
- One Alembic revision with a real downgrade, and the SQLite chain covered as in `tests/unit/test_scoped_budget_alignment_chain.py`.

Follows #640 (merged), raised in review there. Part of #642, and #671 is where the second derivation appears. The platform-side counterpart is mozilla-ai/otari-ai#1741.

[...]

PR #745, opened 2026-08-24
feat(budgets): close the last two gaps in workspace budget defaults

### LINKED ISSUE #680: Workspace budget defaults: reset_alignment, orphaned ceilings on member removal, bootstrap materialization
Three follow-ups from @njbrake's review of #671 (workspace per-member budget defaults), noted there as worth tracking rather than fixing in that PR:

## 1. `WorkspaceBudgetDefault` cannot express `reset_alignment`

#640 added `reset_alignment` to `ScopedBudget` (calendar-aligned resets: daily/weekly/monthly, as an alternative to a rolling `budget_duration_sec` window). `WorkspaceBudgetDefault` only carries `budget_duration_sec`, so a workspace that wants a calendar-month per-member ceiling still has to create every member's row by hand, which is exactly the case this feature exists to remove.

Worth doing deliberately rather than bolting on: the `scoped_budgets` CHECK constraint makes `budget_duration_sec` and `reset_alignment` an either/or pair, not two independent fields, so `WorkspaceBudgetDefault` and its `WorkspaceMemberBudgetPolicyCreate`/`Update` schemas need the same either/or shape, and `_period_window` in `workspace_budget_default_service.py` (a local duplicate of `scoped_budget_service.period_window`, kept local for import-cycle reasons — see that module's docstring) would need the alignment branch too, ideally by finally reconciling the two into one shared definition rather than growing a third copy.

## 2. `remove_member` leaves a materialized `ScopedBudget` behind

`WorkspaceService.remove_member` deletes the `WorkspaceMember` row but the `ScopedBudget(scope_type="workspace_member", scope_id=<member id>)` rows it materialized are not cleaned up. This gap predates #671, but workspace defaults turn an occasional hand-created orphan into one per departing member. `WorkspaceService._delete_scoped_budgets_for` (used on workspace deletion) is already this codebase's answer for a ceiling that has outlived its scope; `remove_member` likely wants the same treatment for the one member's rows.

## 3. `ensure_bootstrap_identity` doesn't materialize

`provisioning_service.ensure_bootstrap_identity` is a fourth path that creates a `WorkspaceMember` row (the first-boot operator identity) and isn't wired into `WorkspaceBudgetDefaultService.materialize_for_member`. First-boot only, and a workspace has no defaults yet at that point in practice, so the practical impact is small, but `WorkspaceService.create_workspace`'s comment ("every WorkspaceMember-creating path materializes the same way") is not quite true while this one is missing.

_Filed by an agent while addressing review feedback on #671; the analysis is the reviewer's (@njbrake, via a Claude session), not new investigation._

njbrake:
## Description

Most of #680 and all of #676 turned out to be closed already by #711, which moved a limit and its period off the ceiling onto the `budgets` row a ceiling names. That gave `WorkspaceBudgetDefault` its calendar alignment (it names a budget carrying one), put the either/or CHECK on `budgets`, and reconciled the duplicated derivation into `services/budget_periods.py`, a leaf both call sites import with the documented cycle intact. #676 proposed `core/sql.py` as the shared home; what landed is that shape under a name that says what it holds. `remove_member` already sweeps the ceilings keyed on the membership it deletes, which is also what otari-ai's `remove_member` does.

Two ends were left uncovered.

**A default naming a calendar-aligned budget read back with every period field null**, which is how a budget that never resets looks. The two period fields are exclusive, and this read shape only carried `budget_duration_sec`. `ScopedBudgetResponse` already carries both off the budget for exactly this reason, and otari-ai's own policy read shape carries its `spend_period`, so this one was the outlier.

**First boot created a `WorkspaceMember` without materializing the workspace's defaults onto it**, the one path of four that did not. A genuine first boot has nothing to materialize, since creating a default needs an identity and there is none until provisioning returns. What makes the call bind is the marker being unresolved on a database that has already run: the identity it names deleted, which `_load_marked_identity` reports as no marker, or the row cleared by hand. The workspace is adopted then rather than created (migration `d5e7f1a2b3c4` seeds the default organization and its workspace, and `_provision` looks both up first), and the operator joined it as the one member with no ceiling.

That same state is why this path takes `WorkspaceRepository.lock` like `add_member` does, rather than arguing it cannot race: `get_current_identity` returns a dashboard session's identity without consulting the marker, so a signed-in operator can be creating a default while a session-less request provisions here. `WorkspaceRepository.lock`'s docstring already says every membership-creation path takes it, with `create_workspace` the one stated exception, and this path is not that exception.

No migration. Nothing schema-shaped is left; #676's "one Alembic revision" is unbuildable now, since the column it asked for would be dead. Head is unchanged at `c4d7f2a9b6e1`.

`tests/unit/test_service_module_imports.py` still passes.

## PR Type

- [x] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes #680
Refs #676

#676 is #680's item 1 promoted into its own issue. Its substance shipped in #711 and its Definition of Done is written against a schema that no longer exists, so it is worth closing as delivered rather than reinterpreting; #680 is the one whose remaining scope matched what was actually left.

## Checklist

Also run green: `make openapi-check`, `make postman-check` (collection came out unchanged), the OSS-edition smoke gate, 2218 unit and 1582 integration tests, 1212 dashboard tests, `tsc -b --noEmit`, and biome. `web/src/client/schema.ts` regenerated and committed.

The bootstrap test was verified to fail without its fix.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

Implemented via back-and-forth with @njbrake, who directed the scope and the call on which of the two issues should own the work. The reasoning and decisions are his; the prose and the code are Claude's.


src/gateway/services/tenancy/provisioning_service.py
```diff
@@ -206,11 +206,37 @@ async def _provision(db: AsyncSession) -> User:
...
+    # No workspace row lock, unlike the other three call sites: they serialize
+    # against a concurrent ``create_default``, which cannot run here, since it
+    # resolves a tenancy identity first and that resolution is this function.
+    # Flush-only, so it lands in the commit below and a lost race rolls it back
+    # with everything else.
+    from gateway.services.tenancy.workspace_budget_default_service import (
+        WorkspaceBudgetDefaultService,
+    )
+
+    await WorkspaceBudgetDefaultService(db).materialize_for_member(member)
```
khaledosman: **Blocker: a broken workspace default now makes the deployment permanently unbootstrappable.**

`materialize_for_member` reaches `_budget_for`, which raises `WorkspaceBudgetDefaultBudgetNotFoundError`. `ensure_bootstrap_identity` catches only `IntegrityError` (line 83), so that escapes, the marker upsert at line 247 never runs, and every later tenancy request re-enters `_provision` and fails identically. The dashboard and every management route answer 404 `Budget ... not found` forever, with no way out over the API: deleting the offending default needs an authorized identity, which needs the marker.

Reproduced against a Testcontainers Postgres, seeding a `workspace_budget_defaults` row that names a missing budget (foreign key dropped, i.e. the restore case), then calling `ensure_bootstrap_identity` twice:

```
PR head:  attempt 1 FAILED WorkspaceBudgetDefaultBudgetNotFoundError / marker = None
          attempt 2 FAILED WorkspaceBudgetDefaultBudgetNotFoundError / marker = None
main:     attempt 1 SUCCEEDED / marker = 51bd4558-...
          attempt 2 SUCCEEDED / marker = 51bd4558-...
```

`_budget_for`'s own docstring names "a database restored without foreign keys enforced" as the only thing that reaches it, which is the same adopted or imported database the comment above cites as this change's reason for existing. So the one scenario the change is for is the one that bricks it.

Fix:

```python
try:
    await WorkspaceBudgetDefaultService(db).materialize_for_member(member)
except WorkspaceBudgetDefaultBudgetNotFoundError:
    logger.warning("Skipping budget-default materialization on first boot: %s", ...)
```

An operator who joins uncapped is the pre-PR state and is recoverable from the dashboard. A gateway that cannot bootstrap is not.

src/gateway/services/tenancy/provisioning_service.py
```diff
@@ -206,11 +206,37 @@ async def _provision(db: AsyncSession) -> User:
...
+    # carry defaults the operator would otherwise join underneath uncapped.
+    #
+    # Imported inside the function, not at module top: this module is reached
+    # through ``tenancy/__init__``, which imports ``workspace_service``, which
+    # imports the default service, so naming it up there would make the cycle
+    # depend on which module was imported first. Same reason
+    # ``OrganizationService._apply_workspace_assignments`` imports it locally;
+    # ``tests/unit/test_service_module_imports.py`` pins the graph.
+    #
+    # No workspace row lock, unlike the other three call sites: they serialize
```
khaledosman: The reason given for skipping the lock is false, and it invalidates a docstring elsewhere.

"`create_default` ... cannot run here, since it resolves a tenancy identity first and that resolution is this function" holds only for master-key requests. `api/deps.py:417` returns `session_identity` and never calls `ensure_bootstrap_identity`, and `auth_session._sign_in_with_password` is what mints such a session. In both scenarios the comment just above invokes (a marker cleared by hand, a tenancy imported under the `default` slug) password identities exist and can sign in, so a concurrent `create_default` on the adopted workspace is reachable, producing exactly the interleaving the lock exists to prevent.

Separately, `repositories/tenancy/workspace_repository.py:45-49` states that every membership-creation path except `create_workspace` takes this lock. This adds a fourth that does not, and the `create_workspace` exemption ("a just-created workspace has no default that could race it") does not cover it, since the case that matters here is the *adopted* workspace.

`workspaces` is already in scope at line 201, so `await workspaces.lock(workspace.id)` is a one-liner. Otherwise amend both docstrings.

tests/integration/test_workspace_member_budget_policies.py
```diff
@@ -709,3 +714,90 @@ async def test_removing_a_member_takes_their_workspace_ceiling_with_them(async_d
...
+    await async_db.commit()
+
+    operator = await ensure_bootstrap_identity(async_db)
+
+    member = await WorkspaceMemberRepository(async_db).get_by_workspace_and_user(workspace.id, operator.id)
+    assert member is not None
+    ceiling = await _member_budget(async_db, member.id)
+    assert ceiling is not None, "the operator identity must be capped by the workspace it joined"
+    assert ceiling.budget_id == budget_id
+    # Materialization is flush-only, so it has to have landed in the commit
```
khaledosman: The comment claims more than the assertions test. "Materialization is flush-only, so it has to have landed in the commit `ensure_bootstrap_identity` makes rather than in a transaction of its own" is not what `period_start is not None` and `.day == 1` distinguish: a separate, committed transaction would read back identically. What they do prove is that the calendar window got stamped.

Restate to that, or assert the coupling for real by failing `_provision` after materialization and checking no ceiling survives.

src/gateway/services/tenancy/workspace_budget_default_service.py
```diff
@@ -106,6 +106,12 @@ class WorkspaceMemberBudgetPolicyPublic(BaseModel):
     name: str | None
     max_budget: float | None
     budget_duration_sec: int | None
+    # The other way the budget can carry a period, alongside
+    # ``budget_duration_sec`` and never with it (a CHECK on ``budgets`` refuses
+    # both). Carried for the same reason ``ScopedBudgetResponse`` carries it: a
+    # default naming a calendar-aligned budget would otherwise read back with
+    # every period field null, which is how a row that never resets looks.
+    reset_alignment: str | None
```
khaledosman, line 114: Nothing consumes this field yet, so the operator-visible half of the gap is still open.

The description frames it as the dashboard rendering the period as none, but `WorkspacesPage.tsx` shows only `budgetLabel(budget)` (the name) for a default and never a period, and `BudgetsPage.tsx:609` renders `formatDuration(b.budget_duration_sec)` with no alignment arm. The wire shape is right now; what an operator sees is unchanged.

Worth a follow-up issue, or drop the dashboard framing from the description.

khaledosman (dismissed):
One blocker: the new materialization call can make a deployment permanently unbootstrappable. Reproduced against Postgres and confirmed `main` does not do it. Three smaller items inline.

Gates verified at this head: `make openapi-check`, `make postman-check` (collection genuinely unchanged), `make lint` (architecture + ruff), `make typecheck` (mypy strict, 464 files), `tsc -b --noEmit`, biome, and `test_workspace_member_budget_policies.py` 18/18. `test_tenancy_races.py::test_concurrent_demotions_cannot_strip_the_last_owner` fails, but it fails on `main` too, so it is not this PR's.

<sub>Reviewed by Claude Opus 5 via the Claude Code CLI, on @khaled-mozilla's request.</sub>

khaledosman (approved):
forgot to pre-approve

PR #746, opened 2026-08-24
fix(auth): gate master-key sign-in on the operator's password, not on any

### LINKED ISSUE #702: has_password_identity gates the operator's login on any identity's password
Split out of the review of #684 (fix PR: #701). Not a bug today; it is a predicate that stops meaning what it says at two known future points.

## What it does

`POST /v1/auth/session` decides whether the master key is still accepted as a dashboard login by calling `has_password_identity` (`src/gateway/services/tenancy/user_service.py`):

```python
found = await db.execute(select(col(User.id)).where(col(User.hashed_password).is_not(None)).limit(1))
return found.first() is not None
```

Its docstring calls this "the deployment has been claimed", and `GET /v1/bootstrap` publishes the answer as `sign_in_methods`. But the question it actually answers is "does *any* row hold a password", while the thing it gates is *the operator's* login.

The two are the same thing only because `PUT /v1/auth/password` is currently the sole writer of `hashed_password`, and that route always acts on the caller's own identity, which on a standalone deployment is the bootstrap operator.

## Where they come apart

**The re-parenting backfill (#1644 over in otari-ai).** `models/tenancy.py` carries the credential columns specifically so that migration has one target schema, and platform identities arrive *with* `hashed_password` set. On a re-parented deployment `has_password_identity` is true on first boot, so `/v1/bootstrap` reports `["password"]` and the sign-in screen asks for an address and a password before anyone has claimed anything. The operator holding the master key cannot use it at the login screen and has to claim by curl first. Arguably defensible, but it is not what "the deployment has been claimed" describes, and nobody chose it.

**Signup and reset (#650).** Once a member can set their own password, the first one to do so retires master-key sign-in for an operator who has none. The operator is not locked out of the API, and `PUT /v1/auth/password` with the master key in a header is still the way back, but they are locked out of the dashboard by somebody else's action, with no UI path back in. `user_service`'s module docstring already flags the current state ("only the operator identity can hold a password") as the reason this is safe for now.

## Options

1. Scope the probe to the bootstrap operator: read `tenancy_bootstrap_user_id` and ask whether *that* identity holds a password. Says exactly what the docstring says, and both cases above behave.
2. Record the claim explicitly in `runtime_settings` rather than deriving it. Rejected once already in the docstring ("Read off the identities rather than a settings row, so it cannot disagree with whether a password sign-in could actually succeed"), which is a good reason, though option 1 keeps that property.
3. Leave it and make `/v1/bootstrap` report both methods while the operator has no password, so the master-key box survives alongside the password form.

Option 1 looks right to me, but this is a decision about #649's contract rather than a fix to it, which is why it is an issue and not part of #701.

## Not urgent

Nothing here is reachable on a deployment that has not been re-parented, so this wants settling before #650 lands or before the first re-parenting run, whichever comes first.

_Filed by Claude Opus 5 via Claude Code, from a post-merge review of #684._

njbrake:
## Description

`has_password_identity` answered "does any row hold a `hashed_password`" while gating *the operator's* dashboard login. Those were the same answer only while `PUT /v1/auth/password` was the sole writer of that column, and it no longer is.

**This is reachable today**, which the issue predates: #650 has landed, so the first roster member to complete signup retires the dashboard login of an operator who never claimed, by an act of their own, with no UI path back in. The re-parenting backfill (mozilla-ai/otari-ai#1644) breaks it the other way, since platform identities arrive with `hashed_password` set: a re-parented deployment reads as claimed on first boot, `/v1/bootstrap` publishes `["password"]`, and the sign-in screen asks the operator holding the master key for credentials that are not theirs.

Issue option 1. The predicate becomes `operator_has_password`, which resolves the `tenancy_bootstrap_user_id` marker and asks what that one identity holds. That is the identity a header master key resolves to (`deps.get_current_identity`), so the credential and the password that retires it now name the same person. `provisioning_service._load_marked_identity` becomes public as `load_bootstrap_identity` to support the read.

`PUT /v1/auth/password` returned `master_key_sign_in_retired: true` unconditionally, which was the same conflation in the response body. It now reads the predicate back, so a member changing their own password on an unclaimed deployment gets `false` and the dashboard leaves the master-key affordances alone (`useDeployment` turns them off on that field alone).

**This changes behavior on existing installs, not just new ones.** A deployment where a member signed up post-#650 and the operator never claimed currently publishes `["password"]` and lets that member sign in through the dashboard. After this it publishes `["master_key"]`, and that member loses the UI login (and gets the claim form on Account settings) until the operator claims. The window closes on one operator action and the master key was never the member's way in, but it is a live change rather than a new-deployment one, and nothing surfaces it to the operator.

**One consequence is deliberate, and documented rather than fixed here.** While a deployment is unclaimed, `sign_in_methods` stays `["master_key"]`, so a member who has signed up signs in by calling `POST /v1/auth/session` directly until the operator claims it. Reporting both methods at once is truthful (the two predicates are independent) but needs a login screen that renders two credentials, which is its own change. The same predicate also gives that member the claim form on **Account settings**, which the endpoint refuses for the missing `current_password`. Both follow from `sign_in_methods` describing the deployment while the dashboard has no route for asking what the signed-in identity itself holds. `docs/access-control.md` and `docs/dashboard.md` now say so, and the guide advises claiming before adding anyone.

Verification, all local:

- Three new tests in `tests/unit/test_password_sign_in.py`, one per direction plus the response field. All three fail on `main` (checked in a worktree at `origin/main`) and pass here.
- `make lint`, `make typecheck` (464 files), `uv run pytest tests/unit` (2221 passed, 8 skipped), `uv run pytest tests/integration` against PostgreSQL 18 (1580 passed, 9 skipped).
- `make openapi-check` and `make postman-check` pass; both artifacts and `web/src/client/schema.ts` regenerated for the changed field description.
- `pnpm --dir web run lint` and vitest (1212 passed), plus the OSS-edition smoke gate.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes #702

## Checklist

## AI Usage

- [ ] No AI was used.

**Any additional AI details you'd like to share:** The scoping decision (option 1 backend-only, accepting the member-sign-in gap rather than growing this into a dual-credential login screen) is @njbrake's, taken after the agent surfaced that #650 had landed and made the trade-off live. Implementation, tests and the verification above are the agent's, run in a container against this repo's own gates.


src/gateway/api/routes/bootstrap.py
```diff
@@ -27,18 +27,19 @@
...
 # credential and ``password`` the steady-state one, and a standalone gateway
-# offers exactly one of them: the master key until an identity has a password,
-# and the password from then on (mozilla-ai/otari-ai#1716). A list rather than a
-# single value because #651 and #652 add methods that coexist with the password
-# rather than replacing it, and because a hybrid gateway offers none.
+# offers exactly one of them: the master key until the operator claims the
+# deployment with a password, and the password from then on
+# (mozilla-ai/otari-ai#1716). A list rather than a single value because #651 and
+# #652 add methods that coexist with the password rather than replacing it, and
+# because a hybrid gateway offers none.
```
khaledosman, line 42: `sign_in_methods`'s own `Field` description below (line 116) still says `master_key` "is offered until **some identity** on this deployment has a password". That is the predicate this PR removes. It is published at `docs/public/openapi.json:2951`, in the Postman collection, and at `web/src/client/schema.ts:4540`, so a client reading the spec gets the old rule. Reword it to the operator identity and regenerate the three artifacts.

src/gateway/api/routes/bootstrap.py
```diff
@@ -180,7 +181,7 @@ async def _sign_in_methods(db: AsyncSession) -> list[SignInMethod]:
     way, since minting one writes a row.
     """
     try:
-        claimed = await has_password_identity(db)
+        claimed = await operator_has_password(db)
```
khaledosman, line 185: `get_bootstrap`'s docstring (line 148) is now wrong twice: "The one database read is a ``LIMIT 1`` probe for **any identity** holding a password". This is two primary-key lookups scoped to the bootstrap operator. `web/src/main.tsx` was updated to say exactly that; this docstring was not, and it is the one that gets published (`openapi.json:13361`, Postman `:866`, `schema.ts:597`).

src/gateway/api/routes/auth_session.py
```diff
@@ -204,7 +211,7 @@ async def _sign_in_with_master_key(
         record_auth_failure("invalid_key")
         _check_login_rate_limit(request)
         raise HTTPException(status_code=status.HTTP_401_UNAUTHORIZED, detail="Invalid master key")
-    if await has_password_identity(db):
+    if await operator_has_password(db):
```
khaledosman, line 215: The `master_key` request field's description at line 110 carries the same stale predicate: "Accepted only while **no identity** on this deployment has a password". Published at `openapi.json:2761` and `schema.ts:4436`. This plus the two in `bootstrap.py` land in one regeneration pass, so `make openapi-check` and `make postman-check` need rerunning after.

web/src/app/nav/AccountMenu.tsx
```diff
@@ -73,9 +73,9 @@ const MENU_DIVIDER = "h-px shrink-0 bg-border"
 // instead of inventing an address.
 //
 // Which credential that is comes from the bootstrap, not from an assumption:
-// a deployment publishes `master_key` in `sign_in_methods` exactly while no
-// identity holds a password, so a claimed one (otari#649) signs in with an
-// address and would be mislabeled by "Master-key session".
+// a deployment publishes `master_key` in `sign_in_methods` exactly while its
+// operator identity holds no password, so a claimed one (otari#649) signs in
+// with an address and would be mislabeled by "Master-key session".
```
khaledosman: `signInMethods.includes("master_key")` at line 91 is now a fact about the deployment, not about the caller, so this comment's reasoning no longer holds in one direction. A member who signed in with their password on an unclaimed deployment (which `docs/access-control.md` now explicitly tells them to do via `POST /v1/auth/session`) gets `detail: "Master-key session"`. Pre-PR that state was unreachable, because their password made the deployment read as claimed. Either take a per-caller signal, or fall back to a neutral string rather than naming a credential this can no longer infer.

web/src/features/account/PasswordCard.tsx
```diff
@@ -74,8 +74,12 @@ function PasswordField({
  *   always cookie-authenticated, so that field is always required here.
  *
  * Which of the two applies is read from the bootstrap's `sign_in_methods`
- * rather than probed: `master_key` is published exactly while no identity on
- * this deployment holds a password. That context cannot be refetched, so a
+ * rather than probed: `master_key` is published exactly while this deployment's
+ * operator identity holds no password (otari#702). It is a fact about that
+ * identity and not about the reader, so a member signed in on an unclaimed
+ * deployment is shown the claim form and the server refuses it for the missing
```
khaledosman: Same overload as `AccountMenu`, with a worse outcome than "the server refuses it": that member is shown the **Claim this deployment** form, and no input to it can succeed. Retyping their own address hits `CurrentPasswordRequiredError`; any other address hits `EmailChangeNotSupportedError`. This PR converts a working change-password form into a dead end for a non-operator on an unclaimed deployment. At minimum render the explanation instead of a form that always 400s.

tests/unit/test_password_sign_in.py
```diff
@@ -733,3 +734,133 @@ def test_a_malformed_stored_address_is_not_reported_as_the_caller_s_mistake(tmp_
...
+        assert client.get("/v1/bootstrap").json()["sign_in_methods"] == ["master_key"]
+
+        # And the master key is not refused as a retired login. It cannot mint a
+        # session on this database either, but for the reason that actually
+        # applies: ``ForeignTenancyError``, whose 500 names the marker to point
+        # in the log rather than in the body. A 403 here would be the bug, since
+        # it would send an operator holding the only credential there is off to
+        # find a password.
+        refused = client.post("/v1/auth/session", json={"master_key": MASTER_KEY})
+        assert refused.status_code == 500
```
khaledosman, line 881: This cannot distinguish the `ForeignTenancyError` the comment above says is the whole point from any other 500: `_tenancy_error_handler` renders every 5xx tenancy error as `{"detail": "Internal server error"}`, so the body is no help either, and a broken fixture or an unrelated 500 passes as the documented behavior. Assert the handler's `logger.error("Tenancy request failed: %s", ...)` line through `caplog` so the test pins the reason it describes.

khaledosman (approved):
Predicate change and the service/route wiring look right, and both directions are pinned by tests. Four things: three published API descriptions still state the predicate this PR just made false (and they ship in the generated artifacts), and two dashboard call sites read `sign_in_methods` as "how did *you* sign in" when it now only answers "is the deployment claimed".

_Review by Claude Opus 5 via Claude Code._

src/gateway/api/routes/bootstrap.py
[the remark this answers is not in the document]
njbrake: Fixed in 5788344, which landed just after this review. The `sign_in_methods` field description now names the operator identity, and the handler docstring describes the two primary-key lookups (the marker, then the identity it names) instead of the `LIMIT 1` probe. OpenAPI, Postman and `web/src/client/schema.ts` were all regenerated; `make openapi-check` and `make postman-check` pass.

_Drafted by Claude Opus 5 via @njbrake._

src/gateway/api/routes/auth_session.py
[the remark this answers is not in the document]
njbrake: Fixed in 5788344. It now reads "Accepted only while the operator identity has no password, which is to say while nobody has claimed this deployment", and the three generated artifacts were regenerated with it.

_Drafted by Claude Opus 5 via @njbrake._

src/gateway/api/routes/auth_password.py
```diff
@@ -203,7 +207,12 @@ async def set_dashboard_password(
             keep_session_token_hash=keep_session,
         )
     assert identity.email is not None  # guaranteed by set_password
-    return PasswordResponse(email=identity.email, master_key_sign_in_retired=True)
+    # Read back rather than hardcoded True: this route acts on the caller's own
+    # identity, and only the operator's password retires master-key sign-in
+    # (#702). A member changing theirs on an unclaimed deployment answers False,
+    # which is what stops the dashboard telling them it retired a login that is
+    # still the only one their operator has.
+    return PasswordResponse(email=identity.email, master_key_sign_in_retired=await operator_has_password(db))
```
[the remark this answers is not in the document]
njbrake: Agreed, and fixed in e9a1184. This was a real window I opened: the read sat after `set_password`/`update_password` had already committed, so a failure there reported 500 for a change that had landed, and a session retry would then send a `current_password` that was no longer current.

Both facts the answer needs are knowable before the write, so they are now resolved there: whether the operator already holds a password, and whether this caller is the operator. The operator was either claimed already or is being claimed by this very call, and no other identity's password moves the switch. No I/O after the commit.

_Drafted by Claude Opus 5 via @njbrake._

docs/dashboard.md
```diff
@@ -114,11 +114,13 @@ the logs.
...
-on, which is what you want as soon as more than one person needs the dashboard,
-or as soon as you would rather not paste a deployment-wide credential into a
-browser.
+The master key signs you in because the operator identity has no password yet.
+Giving it an address and a password makes that pair the sign-in from then on,
+which is what you want as soon as more than one person needs the dashboard, or
+as soon as you would rather not paste a deployment-wide credential into a
+browser. Do it before you add anyone: until you claim, the sign-in screen offers
+the master key rather than the password form, so a member who has signed up
+cannot reach the dashboard from it.
```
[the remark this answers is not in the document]
njbrake: Agreed, fixed in e9a1184. That paragraph predated #650 and was already false before this PR; my new guidance three sections above it made the contradiction visible. It now says signup and password reset have landed, that OAuth and passkeys have not, and that a member setting a password claims their account rather than the deployment.

_Drafted by Claude Opus 5 via @njbrake._

src/gateway/api/routes/bootstrap.py
njbrake: Fixed in 5788344, pushed just after you reviewed. Reworded to the operator identity and all three artifacts regenerated, with `make openapi-check` and `make postman-check` green.

Worth naming the shape of the miss: I rewrote the module comment sitting directly above `SignInMethod` and left the `Field` description ten lines below it saying the opposite.

_Drafted by Claude Opus 5 via @njbrake._

src/gateway/api/routes/bootstrap.py
njbrake: Fixed in 5788344. The docstring now reads "two primary-key lookups: the `tenancy_bootstrap_user_id` marker, and the identity it names". You were right that this is the copy that gets published while the `web/src/main.tsx` mirror I did update is not.

_Drafted by Claude Opus 5 via @njbrake._

src/gateway/api/routes/auth_session.py
njbrake: Fixed in 5788344, in the same regeneration pass as the two in `bootstrap.py`. Both checks rerun and green.

_Drafted by Claude Opus 5 via @njbrake._

web/src/app/nav/AccountMenu.tsx
njbrake: Agreed, and taken the second option in e9a1184: there is no per-caller signal to take. `POST /v1/auth/session` and `DELETE` are the only routes on it, so after a reload the tab has nothing that says which credential minted the session.

The unclaimed case now reports "Unclaimed deployment", which is the part this genuinely knows, and the claimed case keeps "Password sign-in" (the master key is refused outright by then, so every session is a password one and there is nothing to infer). It reads to the operator as the standing invitation to claim, which the docs now also advise doing before adding anyone.

_Drafted by Claude Opus 5 via @njbrake._

web/src/features/account/PasswordCard.tsx
njbrake: Right on the diagnosis, and it is worse than my comment said: neither input can succeed, for the two different reasons you name.

I have taken the "at minimum" and stopped short of the rest, deliberately. The claim copy now says up front that claiming is the operator's to do and that an account already holding a password will be refused, so a member reads it before submitting rather than after. What I have not done is switch the form, because the only client-side signal for "you are not the operator" is the 400 itself, and branching a form on a guess at a refusal string is a fragile fix. The non-fragile one is a route for asking what the signed-in identity holds, which the management API does not have; that is the same follow-up `Login.tsx` needs and is out of this PR's agreed scope.

Two things bound it meanwhile: reaching this form at all means having gone around the sign-in screen, which offers such a member no password form either, and `docs/access-control.md` now documents both halves.

_Drafted by Claude Opus 5 via @njbrake._

tests/unit/test_password_sign_in.py
njbrake: Good catch, fixed in e9a1184. The test now routes the `gateway` logger into `caplog` (it does not propagate) and asserts three things: the handler's "Tenancy request failed" line, the `tenancy_bootstrap_user_id` marker the message tells the operator to point, and the `acme-a1b2` slug that put it there. A bare fixture break or an unrelated 500 no longer passes. Kept the `{"detail": "Internal server error"}` assertion alongside it, since the sanitized body is itself part of the contract.

_Drafted by Claude Opus 5 via @njbrake._

Issue #738, opened 2026-08-24
Telemetry seam: the shell has no way for an overlay to observe navigation or the auth funnel


njbrake:
Add the cross-cutting-module grain to `web/`: a no-op telemetry module in the
base that a build-time override swaps, so an overlay can record events fired
from shell and auth code it does not own.

## Why

`web/` carries no telemetry, and that is correct rather than an oversight: there
is no mixpanel, matomo, intercom or sentry dependency in `web/package.json`, and
an OSS gateway that reported to a vendor by default would be a defect.

But it leaves a deployment that does have consent no way to observe anything,
because the events that matter are fired from files an overlay may not edit under
cardinal rule 6. otari-ai fires twelve events from nine files that all become
`web/`'s once its shell is the base: the acquisition funnel (`SIGNUP_STARTED`,
`SIGNUP_SUCCESS`, `SIGNUP_FAILED`, `LOGIN_SUCCESS`, `LOGIN_FAILED`, `LOGOUT`,
`EMAIL_VERIFICATION_SUCCESS`, `EMAIL_VERIFICATION_FAILED`,
`RESEND_VERIFICATION_CLICKED`, `FORM_VALIDATION_FAILED`), plus `TAB_CHANGED` from
the sidebar and `CHECKOUT_CANCELLED`. Every one comes from sign-in, sign-up,
verification, the shell, or the nav registry. Re-implementing those pages in an
overlay to keep the funnel is the duplicate shell this track exists to retire.

## What to produce

A base module exporting a no-op tracker (an event call and an identity call),
reached by its `@/…` specifier so `OVERLAY_MODULE_OVERRIDES` can replace it. The
base default records nothing and adds no dependency, so an OSS build stays free
of telemetry code as well as telemetry traffic.

Two constraints from the working implementation in otari-ai: the identity half
needs the active organization as well as the user, and every call is gated on
stored consent, so the seam has to carry consent through rather than assume it.

## Definition of done

The base build declares no telemetry dependency and emits nothing. A superset
build records the events above from unmodified `web/` source.

Follows #584 and #644.

One of three frontend seams the shell convergence needs, with #736 and #737.

Prose by Claude Opus 5 via back-and-forth with @njbrake; the decisions are his.

Issue #644, opened 2026-08-19
Port the nav label-override seam


njbrake:
- Labels: `otari-migration`, `open-core`
- Part of #589
- Replaces mozilla-ai/otari-ai#1539, whose premise was overtaken

## Why this exists

mozilla-ai/otari-ai#1539 assumed the OSS registry would need "Organization" renamed to something tenancy-neutral, with the enterprise overlay renaming it back. That is overtaken. The registry rehomed in #619, and the two-rail information architecture that came with it has no "Account & Org" group at all. The surviving "Organization" item sits under "General" and points at `/organization`, the organization's own settings page, which is the right label for a single-tenant deployment.

What is missing is the seam itself. `otari-ai/frontend/src/app/nav/` carries `NavLabelOverride`, `applyNavLabelOverrides`, and an empty `OVERLAY_NAV_LABEL_OVERRIDES`. `web/src/app/nav/` has `overlaySections.ts` and no label equivalent, so an overlay cannot relabel a base nav entry without editing an Otari source file, which ARCHITECTURE.md's cardinal rule 6 forbids.

## What to produce

The label-override module and the type and helper behind it, empty in this build, applied by `registry.ts` to section headers and disclosure labels. Same shape and same name as the `otari-ai` module, so the two registries compose when the control-plane UI converges.

## There is a reference implementation

The platform grew this exact seam in mozilla-ai/otari-ai#1693 (@tbille), and it is merged there. Port that rather than deriving a second design: the two registries are meant to compose, so the type, the helper and the module name should match on both sides.

## Definition of Done

- An overlay renames a base section header or disclosure label by replacing one module at build time.
- The base build's labels are unchanged and the seam ships empty, matching `overlaySections.ts`.
- A test asserts the base default stays empty, as the overlay-sections seam already does.

_Note: this issue was drafted by Claude Opus 5 via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

Issue #584, opened 2026-08-14
Nav registry and entitlement gate seam


njbrake:
Navigation becomes data: pages register into a nav registry, and an entitlement gate decides what renders (entitlement AND feature flag AND capability). The OSS base ships the full mechanism plus an empty overlay seam; a private overlay adds entries at build time without editing OSS source.

## What to produce

- The nav registry primitive and `EntitlementGate`, ported from the platform's M2 machinery; all existing pages registered.
- The empty overlay seam module the OSS build compiles (the overlay tree itself never lands here).
- Composition with the deployment bootstrap's capability list.

## Definition of Done

- Sidebar and routes render from the registry; a registry entry gated off by capability or flag disappears without dead links.

Depends on #582 and #583.

Prose by Claude Fable 5 via back-and-forth with @njbrake; the decisions are his.

PR #747, opened 2026-08-24
chore(dashboard): drop the overlay-owned Billing and Gateways rows

njbrake:
## Description

`overlaySections.ts` already names Billing as ARCHITECTURE.md's canonical overlay-only capability, and names Gateways in the same breath, while `registry.ts` declared both inline twelve files away with an `UnavailableHere` route file behind each. This drops the two entries and both route files, so the base declares neither the row nor the route.

**The failure mode, confirmed.** The composed build mounts the base tree and the overlay's at one root (two `physical("", ...)` entries in the superset `routes.ts`), so a base route file at a path the overlay also serves is not shadowed by it, it collides with it. Reproduced against this repo's own `src/routes` with the pinned generator (1.167.32), the base file restored and an overlay `organization.billing.tsx` mounted alongside:

```
Conflicting configuration paths were found for the following routes: "/organization/billing", "/organization/billing".
Please ensure each Route has a unique full path.
```

The generator throws and writes no tree. Under the Vite plugin that throw is caught and logged (`router-generator-plugin.js`), so the composed build carries on against whatever `routeTree.gen.ts` is already on disk, where the overlay's page has no route at all; a step that generates the tree standalone and checks it fails outright. Either way the overlay page cannot mount while the base declares the path.

**One seam added, because removal alone leaves a hole.** `OVERLAY_ORG_NAV_SECTIONS` appends whole sections, and neither row wants one: Billing belongs inside "Cost & billing" beside `/budgets` and `/organization/pricing`, Gateways inside "Gateway" beside `/organization/guardrails`. Both are base sections with base neighbours, so an overlay would have to edit `registry.ts` to place either, which cardinal rule 6 rules out. `overlayNavItems.ts` plus `composeNavItems` is that seam, one grain finer than the section one, empty here like the other two: a contribution names a base section and its rows land after the base's own. Note the issue points at `NavItemContribution`/`composeNavItems` in otari-ai as the reference; I could not find either there (current `main`, `frontend/src/app/nav`), so the shape follows this repo's own `composeNavSections`/`applyNavLabelOverrides` instead.

`overlayNavItems.test.ts` mocks the module the way an overlay replaces it, because `registry.test.ts` cannot see an unwired `composeNavItems` while the contribution list is empty: unwiring the organization rail leaves that suite and `AppShell.test.tsx` green (95 tests) and turns this file red. The same reason `overlayLabelOverrides.test.ts` exists beside it.

[...]

PR #748, opened 2026-08-24
feat(dashboard): add the intra-page slot seam for the top bar's balance

### LINKED ISSUE #736: Intra-page slot seam: the top bar has no way for an overlay to contribute the balance
Add the intra-page slot grain to `web/`: a named module, empty in the base, that
an overlay's build-time module override swaps. `src/app/nav/TopBarActions.tsx` is
the first case and already names the gap:

> The design also draws a balance here. It is not built: this gateway meters
> spend but holds no wallet, and there is no seam that could feed one, since an
> overlay replaces `overlaySections.ts` / `overlayLabelOverrides.ts` and not this
> file.

## Why

Cardinal rule 6 in [ARCHITECTURE.md](https://github.com/mozilla-ai/otari/blob/main/ARCHITECTURE.md): "If extending
an overlay *requires* editing an Otari file, that is a missing seam, and the seam
belongs in Otari." The nav grain landed in #584 and #644. This is the grain below
it, for what an overlay puts *inside* a page or a piece of chrome the base owns.

It is also one of three frontend seams the shell convergence needs. otari-ai's
frontend is now marketing plus overlay: every logged-in page it still ships is
either an overlay route or something Otari already serves. Its shell stays alive
on those three gaps, and this is the one with no workaround, because a chip in
the base top bar cannot be contributed from anywhere else. Its auth pages are a
separate matter and wait on #651 and #652.

## What to produce

A base module exporting a no-op slot, reached by its `@/…` specifier so a
build-time alias can replace it. `otari-ai/frontend/src/app/overlayWalletSlot.tsx`
(private) is the working reference: an empty `WalletNavSlot` plus a no-op
`useRefreshWallet`, with the overlay's version owning the modal state so the base
chrome holds none.

The specifier is load-bearing. A relative import resolves to the base default in
every edition, and the contribution then vanishes with no error.

## Definition of done

`TopBarActions.tsx` renders the slot and drops the comment above; the base build
renders nothing there; a superset build renders the overlay's chip.

Follows #584 and #644.

The other two frontend seams the shell convergence needs are #737 and #738.

Prose by Claude Opus 5 via back-and-forth with @njbrake; the decisions are his.

njbrake:
## Description

The nav seams let an overlay append a destination to a rail (`overlaySections.ts`) or rename one the base declares (`overlayLabelOverrides.ts`). Neither covers a contribution that is not a destination: the design draws a balance at the end of the top-bar cluster, and a build with a figure to report could only put it there by editing `TopBarActions.tsx`. That is the missing seam cardinal rule 6 describes, and that file's own comment said so.

`web/src/app/nav/overlayWalletSlot.tsx` is the grain below the nav seams: a slot inside a piece of chrome the base owns. It renders nothing here, and a superset build replaces the module so its version renders the chip and owns everything behind it (query, modal, gate), leaving the base chrome no state for a capability it does not have.

Three things worth a reviewer's attention:

**One export, not the two the issue's reference has.** otari-ai's base module pairs the slot with a no-op `useRefreshWallet` because its post-checkout return is handled in `src/routes/_layout.tsx`, a base file. There is no such call site here, and the slot renders on every page of the shell, so the overlay's own component can own that effect. A no-op hook with no base caller reads as wired while reaching nothing, which is what the deleted comment warned against.

**No `web/vite.config.ts` change.** The alias table (`OVERLAY_MODULE_OVERRIDES`) belongs to the superset build: its values resolve against an `OVERLAY_ROOT` this repository has none of and per `web/AGENTS.md` is not meant to grow one of. What this repository owes the mechanism is the inert default plus the `@/…` specifier the alias keys on.

**`web/src/overlaySeams.test.ts` makes that specifier enforceable.** A relative import of a seam resolves to a path the alias never sees, so an overlay's contribution vanishes with no error, in the one kind of build whose only symptom is the empty default it wanted anyway. The test fails on a relative import of a seam and on a seam nothing imports. It caught that #584 and #644 left both nav seams reachable only relatively, so `registry.ts` now imports them by specifier; that is the one change here outside the issue's scope, kept in this PR because it is the same hunk as the test.

Coordinated with #738, which needs the same swap mechanism for a no-op telemetry module: one mechanism, no duplicated plumbing, and no shared files beyond one bullet each in `web/AGENTS.md`.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [x] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes #736

## Checklist

Frontend change, so the checks that apply are the dashboard's: `pnpm --dir web test` (1223 passed), `run lint`, `run typecheck` and `run build` all clean, plus `make lint` at the root. No API contract change, so no spec to regenerate and no generated artifact drift.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

Written from the issue by back-and-forth with @njbrake; the design decisions are his. The design was checked against mozilla-ai/otari-ai's own `OVERLAY_MODULE_OVERRIDES` and `overlayAdminTabs.ts` rather than assumed, which is where the "no vite.config change" conclusion came from.


web/src/overlaySeams.test.ts
```diff
@@ -0,0 +1,110 @@
...
+ * renders nothing) and is swapped at build time by an alias keyed on its `@/…`
+ * specifier, the technique GitLab documents as `ee_else_ce`. The alias table
+ * itself belongs to the superset build's config in `otari-ai`, not here: its
+ * replacements resolve against an overlay tree that this repository has none of
+ * and is not meant to grow one of (see AGENTS.md). What this repository owes the
+ * mechanism is the other half, and it is the half this file checks.
+ *
+ * Add a module here when you add a seam.
+ */
+const SEAM_MODULES = [
```
khaledosman, line 23: `SEAM_MODULES` is hand-maintained, so a seam added to neither this list nor an alias import is invisible to all three assertions: "are all present" only iterates the list, the `@/…` check is `it.each` over the list, and "are never reached relatively" filters on `seamPaths.has(...)`, which is built from the list. A new seam therefore fails nothing and is silently unaliasable.

That is not hypothetical: #747 adds `app/nav/overlayNavItems.ts` with a relative import, and this file stays green.

Derive the candidate set from the filesystem so a new seam is enrolled by existing, and keep the constant only for seams that live outside that directory:
```ts
const NAV_SEAMS = readdirSync(join(SRC, "app/nav"))
  .filter((name) => /^overlay.*\.tsx?$/.test(name))
  .map((name) => `app/nav/${name}`)
```
Failing that, one assertion that every `app/nav/overlay*` module appears in `SEAM_MODULES` gets the same coverage. As written the test protects the seams that existed when it was written and nothing added after.

khaledosman (approved):
The seam and its two-sided test are well aimed: `overlayWalletSlot.test.tsx` mocking the module by the same `@/…` specifier the alias replaces is what makes deleting `<WalletNavSlot />` from the cluster fail something, which `TopBarActions.test.tsx` alone could not.

**Coordinate with #747.** The two branches conflict in four files (`ARCHITECTURE.md`, `web/AGENTS.md`, `web/src/app/nav/registry.ts`, `web/src/app/nav/TopBarActions.tsx`). #747 adds a third nav seam (`overlayNavItems.ts`) imported *relatively*, which this PR's guard will not catch for the reason in the inline comment. It also renumbers `overlayLabelOverrides` to "the third seam" in `web/AGENTS.md`, colliding with this PR's "third seam"; renumber to four on merge. The `TopBarActions.tsx` comment #747 edits is the one this PR deletes, so resolve toward this branch.

Non-blocking.

---

web/src/overlaySeams.test.ts
njbrake: Stale in its premise, and the concrete case does not hold as described. Both worth showing, since the concern underneath was the right one.

**The list is no longer hand-maintained.** `2d8bc4f48` added a third assertion, "are the whole of what the list claims to cover": it discovers every `overlay<Capital>` module under `src/`, excluding tests, and asserts set equality with `SEAM_MODULES`. Your derivation is the same idea scoped to `app/nav`; I went broader deliberately, because #738's seam is `shared/telemetry/overlayTelemetry.ts` and an `app/nav` derivation would not see it.

**On #747, I checked out its head and ran this file against it.** Two corrections. Its production import is already the alias form: `registry.ts:19` is `import { OVERLAY_NAV_ITEMS } from "@/app/nav/overlayNavItems"`. Its only relative import of that module is `registry.test.ts:5`, and test files are exempt by design, the same way `overlayLabelOverrides.test.ts` imports relatively on purpose to pin the base default.

And this file does not stay green. Dropping #747's `overlayNavItems.ts` into the tree fails the completeness assertion, naming the path:

```
+   "app/nav/overlayNavItems.ts",
```

**What is real is the merge-order cost**, so thank you for pointing at #747. Whichever of #747 and #748 lands second owes one line in `SEAM_MODULES`. The two also conflict textually in `TopBarActions.tsx`: #747 edits the "no seam that could feed one" comment to name `overlayNavItems.ts`, while this PR deletes that comment and mounts the slot in its place. Both are one-line resolutions at merge rather than design problems, but the second author should not have to discover them.

**One limitation worth stating**, since what landed is not the filesystem derivation you proposed: the discovery rule is the naming convention, so it makes the convention enforceable rather than the seam set provable. A seam named for what it does, rather than `overlay<Capital>`, is still invisible to it. Nothing on this side distinguishes a seam from an ordinary module, so that is the trade for having an automated check at all. It is written down at the assertion so the next person does not assume more coverage than exists.

Resolving with no code change, since the gap was closed before this thread landed.

njbrake: Following up on the coordination note in your approval, since it named a collision worth removing rather than deferring.

You are right that #747 and this branch both claim "the third seam" in `web/AGENTS.md`. I took a different resolution than renumbering to four: `763a9c065` names this one by kind ("the intra-page seam, one grain below the nav seams above") and drops the ordinal. Renumbering would be correct once and wrong again the next time a nav seam lands above it, and this seam is not a fourth of the same thing in any case, which is the distinction the rest of the bullet is about.

The other three conflicts you listed stand as you described them, and I agree with your resolutions:

- `TopBarActions.tsx`: resolve toward this branch. #747 edits the "no seam that could feed one" comment; this branch deletes it, because the comment's claim stops being true here.
- `registry.ts` and `ARCHITECTURE.md`: ordinary text conflicts, no design question in either.

One correction to the note itself, covered in the inline thread above: #747's `overlayNavItems.ts` is imported by the alias form at its `registry.ts:19`, not relatively, and the guard on this branch does flag it. It fails the completeness assertion naming the path, so whichever of the two lands second owes one line in `SEAM_MODULES` and will be told so by a red test rather than by a reviewer.

Thanks for catching the #747 overlap. It was the one thing about this change I could not see from inside it.

PR #749, opened 2026-08-24
feat(gateway): freeze dashboard sign-ins with a maintenance-mode switch

### LINKED ISSUE #717: Note for migration: 'feature flags' page can go, only need to keep the 'maintainence mode toggle
We need to keep the 'maintanance mode' toggle so that a platform admin has a way to freeze new logins so we can redeploy. all the other feature flags can go, which means we can probably migrate that one toggle to a different page and delete the feature flags page for rn

njbrake:
## Description

Builds the one thing #718 left behind. Nothing was portable: otari.ai has a flag of the same name, but `_layout.tsx` swaps in a placard for a non-superuser *after* login, `login.tsx` checks nothing, and the evaluation endpoint takes `current_user`, so it cannot be reached before a session exists. The login still succeeds and the API stays open. #717 asks for the opposite, so the refusal is server-side here.

**`POST /v1/auth/session` refuses every credential with 503 while the freeze is on**, before either one is verified: a frozen deployment should not spend a bcrypt verification per attempt, and the refusal is not about the credential. No identity is exempt because none needs to be. The master key still authenticates the management API through the header, which never passes through the door the freeze closes, so `PATCH /v1/settings/maintenance-mode` turns it back off even from a browser holding no cookie.

**The flag shares the `runtime_settings` table but deliberately not `runtime_settings_service`.** No migration; that table is a schema-light key/value store. But every key in that service mirrors a `GatewayConfig` field, is applied by mutating this worker's in-memory config, and reaches other replicas only on their next startup. There is no config field here, and a freeze holding on one replica is not a freeze, so it is read from the stored row on every sign-in attempt.

**Scope is what the issue asks for and no more.** Sessions already minted keep working, so the operator running the redeploy is never booted by the switch they just flipped. Signup and invitation acceptance mint no session and are untouched. The data plane is untouched. `GET /v1/bootstrap` publishes the flag unauthenticated so the sign-in screen says what is happening instead of presenting a form that can only be refused, and a tab already open when the freeze started renders the gateway's own 503 wording the way it renders every other refusal.

`test_sdk_endpoint_coverage` required classifying the two new routes in `scripts/sdk_codegen/sdk-endpoints.txt`; both are filed as not yet wrapped.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes #717. Follows #718, which deleted the rest of the axis and recorded that this toggle is a setting rather than a flag.

## Checklist

- [x] I have added or updated tests that cover my change (`tests/unit`, `tests/integration`). `tests/unit/test_maintenance_mode.py` covers the refusal for both credentials, the surviving session, the header master key staying reachable, the freeze surviving a restart (a second app over the same database standing in for a second replica), and the bootstrap. Dashboard: the card, the sign-in screen's maintenance branch, and 503 as a refusal in `createSession`.
- [x] I ran the Definition of Done checks locally (`make lint`, `make typecheck`, `make test`). Lint and typecheck green. Unit 2229 passed, integration 1580 passed (no Docker here, so against a local PostgreSQL 18 rather than left unrun). Dashboard lint, typecheck and test green (1224). OSS edition smoke passes.
- [x] Documentation was updated where necessary. `docs/access-control.md` gains a "Freezing sign-ins for a redeploy" section; `docs/dashboard.md` describes the card.
- [x] If the API contract changed, I regenerated the OpenAPI spec. Both artifacts regenerated and committed; `make openapi-check` and `make postman-check` pass.

## AI Usage

- [ ] No AI was used.

@njbrake stopped the work before any code was written and asked for the facts first, which is what changed the shape of it. Verified against both repos rather than taking #717's phrasing as settled: the whole flag axis is gone from this repo on main, and otari.ai's `feature_flags.yaml`, `feature_flag_service`, `_layout.tsx`, `login.tsx` and the `require_feature_flag` family in `deps.py` (which has zero callers) show its maintenance mode is cosmetic. So the reported job, "migrate one toggle to another page", was really "build a login freeze". He then chose the scope: sign-in only, dashboard only, its own endpoint. Prose is the model's; the decisions are his.


src/gateway/api/routes/auth_session.py
```diff
@@ -263,7 +274,23 @@ async def create_session(
...
+    refuses both. Before, because a frozen deployment should not spend a bcrypt
+    verification per attempt and the refusal is not about the credential
+    anyway; both, because the way back out is the master key against
+    ``PATCH /v1/settings/maintenance-mode`` through the header, which never
+    passes through this door. That is what keeps the switch from locking out
+    its own operator, and it is why no identity needs an exemption here. It
+    leaks nothing either: ``GET /v1/bootstrap`` already publishes the same flag
+    unauthenticated, so the sign-in screen can render the right page.
     """
+    if await is_maintenance_mode(db):
```
khaledosman, line 298: "Refuses every credential" does not survive the passkeys work. `POST /v1/auth/webauthn/authenticate` mints a session at `src/gateway/api/routes/auth_webauthn.py:232` (commit `25ed0b51`, not yet on `main`) with no freeze check, so once that lands the freeze has a second open door and the docs' guarantee in `docs/access-control.md` becomes false.

Whichever branch merges second needs the same check on that route. Worth pinning it with a test that enumerates the session-minting routes rather than naming the two that exist now, since this is the second time a new sign-in path has had to remember a cross-cutting rule.

src/gateway/api/routes/auth_session.py
```diff
@@ -263,7 +274,23 @@ async def create_session(
...
+        record_auth_failure("maintenance_mode")
```
khaledosman: This files a non-credential refusal into the `auth_failures` counter, and unlike the `invalid_password` / `email_not_verified` branches it deliberately does not throttle. So any alert on auth-failure rate fires during a *planned* freeze, which is the one time an operator least wants to be paged.

Give it its own counter, or drop the call: a 503 that never looked at a credential is not an authentication failure.

web/src/features/auth/Login.tsx
```diff
@@ -347,6 +347,42 @@ export function Login() {
...
 
+  // An operator has frozen sign-ins to redeploy this gateway. Rendering the
+  // form instead would offer a credential whose only outcome is a 503, and one
+  // whose refusal reads as "wrong key" to anyone who does not already know a
+  // freeze is on. Read from the bootstrap, so this is what the page shows on a
+  // fresh load; a tab that was already open when the freeze started still has
+  // the form, and submitting it renders the gateway's own 503 wording on the
+  // label row, the same way every other refusal arrives.
+  //
+  // Checked after `signInUnavailable` because the two cannot both be true: the
```
khaledosman, line 388: The comment and the code disagree. It says "Checked after `signInUnavailable`" and argues "cannot reach its database" is the more actionable of the two if they collide, but the `maintenance_mode` branch returns at line 362 and `signInUnavailable` only at 386, so maintenance wins the collision the comment says it should lose.

Move the block below `if (signInUnavailable)`, which is what the stated reasoning asks for, or drop the claim.

web/src/shared/api/client.ts
```diff
@@ -86,7 +87,17 @@ export async function createSession(
...
+  // 503 is maintenance mode, and it belongs with the other two rather than on
+  // the throw path: it is the gateway deliberately refusing this sign-in, with
+  // wording written for the person reading it, not a fault to be surfaced as
+  // one. The sign-in screen normally renders a notice instead of the form on a
+  // frozen deployment (it reads the same flag from the bootstrap); this is the
+  // tab that was already open when the freeze started.
+  if (
+    response.status === 401 ||
+    response.status === 403 ||
+    response.status === 503
```
khaledosman: Keying the pass-through on the status alone conflates maintenance mode with every other 503. A proxy with no healthy upstream, or the gateway mid-restart, is exactly the window this feature exists for, and there `extractErrorMessage` gets a non-JSON body, falls through to `response.statusText`, and the sign-in label row renders "Service Unavailable" as a credential refusal rather than as an unreachable-gateway fault.

Gate the 503 arm on a parsed JSON body carrying `detail`, and let anything else keep taking the `ApiError` path.

web/src/features/settings/SettingsPage.tsx
```diff
@@ -52,38 +54,6 @@ export function fieldMatches(field: ConfigField, query: string): boolean {
 
 // A small on/off switch. role="switch" so it reads correctly to assistive tech
```
khaledosman: This comment stayed behind when `Toggle` moved to `Toggle.tsx`, so two comment blocks now stack above `NumberSetting` and the first describes nothing in this file. Delete it, or fold its point about `role="switch"` and the accessible name into `Toggle.tsx`'s own docstring, which does not currently carry it.

khaledosman (approved):
Design is sound where it matters. The router is inside the standalone-only block of `register_routers`; `verify_master_key` accepts the session cookie *or* the header, so the switch genuinely cannot lock out its own operator; and `create_dashboard_session` has exactly two call sites on `main` (`auth_session.py` and master-key rotation in `settings.py`, which re-mints the already-authenticated caller's own session), so the claim that the freeze covers the only sign-in door holds today. Reading the stored row per attempt rather than mirroring it into `runtime_settings_service` is the right call for a multi-replica freeze, and both generated artifacts are regenerated and in sync.

Five non-blocking notes inline, the first of which is a gap that opens the moment the passkeys branch lands.

---

src/gateway/api/routes/auth_session.py
njbrake: Already dropped, in 3770a1f7f, before this landed. The freeze branch now carries a comment saying why rather than a counter call: nothing failed to authenticate, because the check runs before either credential is verified and the gateway declined to try.

Your second reason is the one I put in the comment, since it is the one someone would otherwise re-add the call without knowing.

web/src/features/auth/Login.tsx
njbrake: Already fixed, in 3770a1f7f. The block moved below `if (signInUnavailable)`, which is what the comment argued for; `signInUnavailable` now returns at 353 and `maintenance_mode` at 392.

I kept the claim rather than dropping it because it is still the right tie-break, even though the two cannot actually collide: a database failure is what empties `sign_in_methods`, and it is also what makes the bootstrap report no freeze.

web/src/features/settings/SettingsPage.tsx
njbrake: Already fixed, in 3770a1f7f. Its point about `role="switch"` and the accessible name folded into `Toggle.tsx`, which is the option you named that keeps it: the docstring there now says the control is `role="switch"` so it reads correctly to assistive tech and can be targeted by its accessible name, whether that is a config key or the maintenance switch label.

web/src/shared/api/client.ts
njbrake: Fixed in ff85d822a, your way. The 503 arm now gates on a JSON body carrying `detail` and everything else keeps the ApiError path.

You are right that the redeploy is exactly the window a proxy answers its own 503, which makes this the likely case rather than the exotic one. I had argued in a comment that both readings mean "not signing anyone in right now", and that misses what you point at: the label row is scoped to a credential, so rendering `statusText` there says the credential was rejected by a gateway that never saw it.

`extractErrorMessage` became `readRefusal`, returning `{ detail, message }` off one read of the body, since the body can only be read once and the caller has to tell who wrote it. Covered by a test with an HTML 503 that expects the throw.

src/gateway/api/routes/auth_session.py
njbrake: Good catch, and the enumerating test is the part I took. `auth_webauthn.py` is not on `main`, so I could not add the check to it; instead `tests/unit/test_maintenance_mode.py` now scans `api/routes/` for `create_dashboard_session` call sites and fails on one that neither references `is_maintenance_mode` nor appears in `SESSION_MINTING_EXEMPT` with a reason. Master-key rotation is the single exemption today: it re-mints the caller's own session and hands nothing to a caller who has none, so it opens no door.

That gives your "whichever branch merges second" a mechanism rather than a memory. If 25ed0b51 rebases onto this, its new call site fails the test until the route checks the freeze; if this rebases onto that, the same. I mutation-checked it by removing the one exemption and confirming it fails.

The service docstring names WebAuthn and OAuth (#651, #652) as the two that will answer to it. I left `docs/access-control.md` saying "every credential" because it is true of what is on `main`; the test is what keeps it true.

web/src/shared/api/hooks.ts
```diff
@@ -620,6 +622,43 @@ export function useUpdateSettings() {
...
+ * Not read from the bootstrap, which carries the same flag: that one is fetched
+ * once per page load and cached for the life of the tab, which is right for the
+ * sign-in screen (it renders before there is anything to poll with) and wrong
+ * for the switch that changes it. This is the live value the card renders.
+ */
+export function useMaintenanceMode() {
+  return useQuery({
+    queryKey: [MAINTENANCE_MODE],
+    queryFn: () => apiFetch<MaintenanceMode>("/v1/settings/maintenance-mode"),
+    staleTime: 60_000,
```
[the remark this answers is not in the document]
njbrake: Fixed in ff85d822a. You are right, and more so than the finding says: this app sets `refetchOnWindowFocus: false` globally in `app/provider.tsx`, so the value was not refreshing on focus either and a mounted card really could hold its mount-time answer indefinitely. My own docstring called it "the live value", which it was not.

It now polls every 30s while mounted, refetches on focus, and sets `staleTime: 0`, the same treatment `useDashboardBuild` gets and for the same reason. One indexed single-row read, and only while the settings page is open.

web/src/features/settings/Toggle.tsx
```diff
@@ -0,0 +1,38 @@
...
+    <button
+      type="button"
+      role="switch"
+      aria-checked={checked}
+      aria-label={label}
+      disabled={disabled}
+      onClick={() => onChange(!checked)}
+      className={`relative inline-flex h-6 w-11 shrink-0 items-center rounded-full transition-colors disabled:opacity-50 ${
+        checked ? "bg-accent" : "bg-surface-subtle"
+      }`}
```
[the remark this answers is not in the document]
njbrake: Fixed in ff85d822a. The track is `h-6 w-11`, so 24px tall against the 44px floor `responsiveness.md` sets for the phone viewport.

I kept the 24px visual track, which is what the settings rows are drawn around, and grew the hit area with `before:absolute before:-inset-y-2.5 before:inset-x-0` on the already-`relative` button: 24 + 20 = 44px in both dimensions, absolutely positioned so no row moves. That is the same device the master-key reveal toggle uses on the sign-in screen, which is why I did not reach for the HeroUI primitive. It also gains `focus-visible:ring-2 focus-visible:ring-accent`, since you are right that it had no visible focus state.

Worth noting this control is pre-existing: this PR moved it out of `SettingsPage.tsx` so a second caller could use it, and the fix now covers every settable boolean on that page rather than only the new switch.

src/gateway/api/routes/auth_session.py
```diff
@@ -263,7 +274,23 @@ async def create_session(
...
+        raise HTTPException(
+            status_code=status.HTTP_503_SERVICE_UNAVAILABLE,
+            detail=MAINTENANCE_MODE_REFUSAL,
+        )
```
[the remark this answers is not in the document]
njbrake: Not taking this one, and the reason is what the feature guarantees rather than whether the race is real. The mechanism is described correctly: a request can read `false`, spend a bcrypt verification, and commit its session after a `PATCH` has returned.

What that costs is a session minted a few hundred milliseconds after the toggle instead of a few hundred before. This freeze deliberately does not revoke sessions already issued (#717 asks to freeze new logins, and booting the operator who just flipped the switch mid-redeploy is the failure mode that ruled the alternative out). So a session from one second *before* the toggle is equally live, equally allowed, and lasts the whole redeploy. The one the race lets through is indistinguishable from it in every consequence, which means closing the window does not change what an operator can rely on: sessions exist during the redeploy either way.

The lock is not free either. A shared lock held from the check through the session commit puts a database round trip and a held lock on every sign-in, on the path that already pays ~200ms of bcrypt, so that a planned operation is 200ms tidier. That is contention on the login path in exchange for no change in guarantee.

If the goal is the stronger property, that there are no live sessions after the toggle returns, the lock is not what gets you there: revoking existing sessions on toggle-on is, and `revoke_all_dashboard_sessions` already exists (master-key rotation uses it). That was considered and declined as this PR's scope, because it signs out the operator running the redeploy. I would rather leave the door visible than half-close it with a lock that reads like the guarantee is stronger than it is.

Resolving on that basis. Happy to reopen if you want the revoke-on-freeze variant; it is a scope call for @njbrake rather than a correctness one, and it belongs in its own change with the lockout question that comes with it.

PR #752, opened 2026-08-24
feat(guardrails): let an organization mandate guardrails over its workspaces

### LINKED ISSUE #654: Organization and workspace guardrail configuration

njbrake:
## Description

The fourth and last config plane (#654). Otari had guardrail execution and one deployment-wide operator setting; an organization had no way to say "every request from these workspaces is checked".

Adds `organization_guardrails` plus `organization_guardrail_workspaces`: one row per profile an organization mandates, with an optional endpoint and credential of its own, the two failure-handling modes a request-body entry already has, an enabled flag, and either the workspaces it runs in or the switch that runs it in all of them. Managed over `/v1/organizations/me/guardrails` (master key on the router, organization owner or admin in the service, on the read as well as the writes) and surfaced by a card under the deployment guardrail settings.

**This plane sits above the deployment rather than below it**, which is a departure from #655/#678 worth stating rather than glossing: a guardrail is a restriction the tenant accepts, not a capability it acquires, so an entry can only ever make fewer requests succeed. Three things follow, and none generalize to the other three planes.

- An entry may name its own endpoint and credential, where the workspace code-execution policy may not. A caller can already point a request-body guardrail at an endpoint of their own, SSRF-checked on the request path, so storing one reaches nothing new. The credential is encrypted with the same `secret_box` and rides as `Authorization: Bearer`, not in the body: the guardrails service forbids unknown body fields, and a credential field on `GuardrailConfig` would be one a caller could set.
- The scope is the organization's, with no workspace opt-out, since a veto is the one move that would widen what succeeds.
- The resolve is unconditional, where the MCP and code-execution ones fire only on a request that opts in: a mandate that ran only when the caller asked for it would not be a mandate. One indexed read per completion request in standalone; hybrid is untouched, since the platform has no guardrail resolve endpoint.

`merge_guardrail_layers` folds caller, then organization, then routing policy, stricter setting winning on every axis, with the outermost layer owning the endpoint a shared profile is sent to. A credential is dropped for any profile the policy layer takes over, since it was stored for an endpoint that is no longer the one being called.

Ported from otari-ai's `OrganizationGuardrailKey` with four deliberate divergences: `profile` is unique per organization rather than a nickname (this request path has always merged by profile, so a second row of one profile could never run); no `key_source`, since platform-managed sources are commercial depth (otari-ai#1699); an `enabled` flag the hosted row has no equivalent of; and no `extra_kwargs_for_creation`, because the guardrails container builds its guardrails at boot from the operator's YAML and `POST /validate` accepts only `validate_kwargs`.

`guardrails_url` stays a `runtime_settings` concern and nothing about it moved. An organization with no entries changes no request, asserted on both halves, and a new workspace inherits the entries marked `applies_to_all_workspaces` and nothing else.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes #654.

## Checklist

- [x] I have added or updated tests that cover my change (`tests/unit`, `tests/integration`). Unit: the three-layer merge and the bearer routing. Integration: CRUD, roles, scope, encryption at rest, and the request path, including the zero-rows case #655 requires of each plane. Dashboard: `OrganizationGuardrailsCard.test.tsx`.
- [x] I ran the Definition of Done checks locally (`make lint`, `make typecheck`, `make test`), plus `pnpm --dir web run lint`, `typecheck`, `test`, and the OSS-edition smoke gate.
- [x] Documentation was updated where necessary. `docs/guardrails.md` describes both layers and how they compose; `docs/dashboard.md`, `src/gateway/AGENTS.md` and `web/AGENTS.md` follow.
- [x] If the API contract changed, I regenerated the OpenAPI spec. Postman collection, the SDK endpoint manifest, and the dashboard client are regenerated with it.

## AI Usage

- [ ] No AI was used.

The design was checked against mozilla-ai/otari-ai's `organization_guardrail_key` plane and against the `otari-anyguardrails-container` and `any-guardrail` sources rather than assumed. The container's `/validate` forbids unknown body fields and builds its guardrails at boot, which is what settled where a stored credential can go. Prose is the model's; the decisions are Nathan's.

## Summary

Added organization-level guardrail management above workspace configuration.

- Organizations can create, update, pause, and delete guardrail profiles.
- Guardrails support workspace-specific scope or inheritance across current and future workspaces.
- Credentials are encrypted, write-only in API responses, and routed as bearer tokens.
- Requests now merge caller, organization, and routing-policy guardrails. Stricter settings take precedence.
- Organization guardrails apply unconditionally when configured.
- Preserved deployment-wide `guardrails_url` fallback behavior.
- Added authorized CRUD APIs, dashboard controls, generated client types, documentation, and comprehensive tests.
- Sanitized public errors to prevent endpoint details from leaking.


src/gateway/services/tenancy/organization_guardrail_service.py
```diff
@@ -0,0 +1,670 @@
...
+                .select_from(OrganizationGuardrail)
+                .where(OrganizationGuardrail.organization_id == organization_id)
+            )
+        ).scalar_one()
+        if existing >= MAX_GUARDRAILS_PER_ORGANIZATION:
+            raise OrganizationGuardrailLimitReachedError(MAX_GUARDRAILS_PER_ORGANIZATION)
+
+        url = _blank_to_none(request.url)
+        credential = _blank_to_none(request.credential)
+        if url is not None:
```
khaledosman, line 599: A credential stored without a `url` is never scheme-checked, and still ships.

`_validate_url` runs only when `url is not None`, but `run_input_guardrails` keys credentials by profile and sends `Authorization: Bearer` regardless of whether the entry carries a URL (`services/guardrails.py::_validate_one`), while the SSRF/TLS pre-check loop skips entries with `g.url is None`. So `POST {"profile": "x", "credential": "s3cret"}` stores a secret that then travels to `default_url`, which is commonly plain http; this PR's own `_DEPLOYMENT_URL` is `http://anyguardrails:8000`.

The field description on line 132 says the credential "requires an https URL". With a null url it does not. Same gap at line 580 on the update path, which is the likelier way in: `GuardrailRow` offers a credential box and no URL field, so an operator can add a credential to a stored entry that has no endpoint of its own.

Refuse a credential when the effective URL is None, or check the deployment default's scheme when a credential is in play. No test covers the combination today.

src/gateway/api/routes/_pipeline.py
```diff
@@ -1599,44 +1607,118 @@ async def _validate_mcp_server_urls(adapter: FormatAdapter[Any, Any], mcp_server
...
+    this one is unconditional: those run only when a request opts into the
+    feature, and a mandate that only ran when the caller asked for it would not
+    be a mandate. The cost is one indexed query on a table an organization edits
+    by hand.
+
+    A workspace always has an organization (``workspace.organization_id`` is not
+    nullable), so a `None` here means the preamble resolved no workspace row at
+    all, and there is no organization plane above such a request to consult.
+    """
+    if ctx.hybrid_mode or ctx.organization_id is None:
```
khaledosman: The fail-open arm swallows the case the fail-closed arm below was written for.

`ctx.organization_id is None` returns `[]`, while the next line raises 500 on `ctx.workspace_id is None` because "what this guards is an *enforcement* decision". But `organization_for_workspace_id` (`services/workspace_scope.py:189`) returns `None` exactly when the workspace row is missing, and such a request still has a non-`None` `ctx.workspace_id`, so it reaches this line and skips every mandate silently. The docstring's justification ("a `None` here means the preamble resolved no workspace row at all") describes the state that never reaches the raise.

```python
if ctx.hybrid_mode:
    return []
if ctx.db is None or ctx.workspace_id is None or ctx.organization_id is None:
    raise adapter.error(500, ORGANIZATION_GUARDRAILS_UNRESOLVABLE_DETAIL, ErrorKind.API)
```

web/src/features/tools/OrganizationGuardrailsCard.tsx
```diff
@@ -0,0 +1,490 @@
...
+  )
+  const [scope, setScope] = useState<string[]>([...guardrail.workspace_ids])
+  // Blank means "keep the stored credential". The field is write-only, so it
+  // never shows what is stored, only whether something is.
+  const [credential, setCredential] = useState("")
+  const [error, setError] = useState("")
+
+  // Rehydrate from whatever the server last said, so the row never drifts from
+  // the stored entry after a save.
+  useEffect(() => {
```
khaledosman, line 160: Saving one row discards unsaved edits in every other row.

The dependency is the whole `guardrail` object, and every row mutation invalidates `[ORGANIZATION_GUARDRAILS]`. Saving row A refetches the list, hands row B a fresh object, and resets B's mode, if-unreachable, status, and scope with no indication that it happened.

`SearchToolsCard.tsx:79` is the pattern: depend on the server-owned values, not the object.

```ts
}, [
  guardrail.mode,
  guardrail.on_unavailable,
  guardrail.enabled,
  guardrail.applies_to_all_workspaces,
  guardrail.workspace_ids.join(","),
])
```

web/src/features/tools/OrganizationGuardrailsCard.tsx
```diff
@@ -0,0 +1,490 @@
...
+// the caller asked for them or not, and in which workspaces.
+//
+// This is the one tool plane that adds rather than narrows, and the reason is
+// worth keeping in view while editing: a guardrail is a restriction the
+// organization accepts, not a capability it acquires, so an entry here can only
+// ever make fewer requests succeed. That is why an entry may name an endpoint
+// and a credential of its own where the workspace code-execution policy below
+// may not. See `src/gateway/AGENTS.md`.
+
+const INPUT_CLASS =
```
khaledosman: `INPUT_CLASS` here and `Badge` on line 48 are byte-identical to `SearchToolsCard.tsx:28` and `:31`, in the same feature folder. Second copy; hoist both into `shared/components/ui` before a third one lands.

src/gateway/models/entities.py
```diff
@@ -1490,3 +1490,96 @@ class WorkspaceCodeExecutionPolicy(Base):
         default=lambda: datetime.now(UTC),
         onupdate=lambda: datetime.now(UTC),
     )
+
+class OrganizationGuardrail(Base):
```
khaledosman, line 1590: One blank line before the class where every other class in this file has two. `ruff format` rewrites it, but `make lint` runs `ruff check` only, so nothing in CI catches it.

khaledosman (approved):
Five inline findings, none blocking. The merge-order reasoning holds up: `_overlay_mandate` always replaces with the mandating entry's whole config, so a caller can never redirect a mandated check, and the policy-layer credential drop is correct and tested.

One thing outside the diff, so it could not be anchored inline: `_pipeline.py:1972` still says "Three reads in this block touch the database". With `_resolve_organization_guardrails` it is four.

*Reviewed with [Claude Code](https://claude.com/claude-code) (Claude Opus 5).*

src/gateway/services/tenancy/organization_guardrail_service.py
njbrake: Fixed in 04a0a1e62. You are right, and two other reviewers landed on the same place, so I dropped the position an earlier commit took (which corrected the field description and left the behavior).

A credential now requires the entry to name its own `url`, which is what routes it through `validate_mcp_url` and its https rule. `_require_url_for_credential` is checked against the effective pair on update as well, so both ways in are closed: adding a credential to a url-less entry, and clearing the url from one that keeps its credential. Three integration tests cover those.

I did not take the other option you offered, checking the deployment default's scheme, because that setting can change after the row is written and the row cannot see it, so the promise would still be conditional on something this plane does not control.

One part of your note was true when you wrote it and is not now: `GuardrailRow` gained a URL field in 0f08bf2b9, which landed after your review.

src/gateway/api/routes/_pipeline.py
njbrake: Fixed in 04a0a1e62. Agreed on both counts: the justification described a state that never reaches the raise, and the two arms disagreed for no reason. All three standalone preconditions now fail closed together, as you wrote it.

Worth flagging one consequence, since it moved other tests. The organization resolve is the first gate in `prepare_gateway_tools`, so a context with no workspace is now refused there rather than by the `mcp_server_ids` guard or the sandbox one. Rather than let those two lose their coverage, `test_mcp_server_ids_without_a_workspace_releases_reservation` is rewritten to assert what the input can actually reach, #725's sandbox arm neutralizes the earlier gate so it still exercises its own, and the new gate gets its own test in `test_organization_guardrail_enforcement.py`. `_resolve_mcp_server_ids` keeps its guard; it is simply no longer first.

web/src/features/tools/OrganizationGuardrailsCard.tsx
njbrake: Changed in 04a0a1e62, but I could not reproduce the clobber, so the reasoning here differs from the one in the finding.

TanStack Query's structural sharing (default in v5, not overridden here) hands an untouched row back its previous object, so the refetch a sibling save triggers does not re-run that row's effect. I checked this rather than assuming it: I wrote the two-row test you would expect, typed into row B, saved row A, and ran it against the old dependency array. It passes either way.

I took the change anyway, because depending on the stored values rather than on object identity means the behavior does not rest on that library detail, and because `SearchToolsCard` is the pattern. The test is kept and describes itself as the property worth holding rather than as a regression test for the dependency array, since it cannot fail for that reason.

web/src/features/tools/OrganizationGuardrailsCard.tsx
njbrake: Fixed in 04a0a1e62. Both moved to `shared/components/ui`, with `SearchToolsCard` importing them from there, so there is one copy rather than two and no third to land.

src/gateway/models/entities.py
njbrake: Already fixed, in 0f08bf2b9, which landed after your review. Same diagnosis you gave: `ruff format` rewrites it and `make lint` runs `ruff check` only, so nothing in CI catches it.

src/gateway/services/tenancy/organization_guardrail_service.py
```diff
@@ -0,0 +1,670 @@
...
+            await _validate_url(url, has_credential=credential is not None)
```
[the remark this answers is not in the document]
njbrake: Real, and fixed in 04a0a1e62. Two other reviewers reached the same finding.

A credential now requires the entry to name its own `url`, so `validate_mcp_url` sees it and refuses `http` outright. The update path checks the effective pair rather than either half, which closes the case you name at the end: clearing an explicit URL while retaining a credential is refused, as is adding a credential to an entry that never had one.

I did not take the alternative of validating the effective deployment URL's transport, because `guardrails_url` can change after the row is written and the row cannot see it, so the guarantee would still be conditional. Three integration tests cover the create path, the add-a-credential-later path, and the clear-the-url path.

src/gateway/services/tenancy/organization_guardrail_service.py
```diff
@@ -0,0 +1,670 @@
...
+            count=total,
+        )
+
+    async def create_guardrail(
+        self, *, user: User, request: OrganizationGuardrailCreate
+    ) -> OrganizationGuardrailPublic:
+        """Mandate a guardrail, encrypting its credential before it is stored."""
+        organization_id = await self._manageable_organization_id(user)
+
+        existing = (
```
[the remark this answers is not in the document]
njbrake: Not taking this one, and the premise does not hold.

The closest precedent is `workspace_mcp_server_service.create_server`, which enforces `MAX_MCP_SERVERS_PER_WORKSPACE` with the same unlocked count-then-insert and leaves the unique index as the arbiter, with a comment saying so. I read it rather than assuming; the tenancy services do not take the organization row lock for this class of bound.

The cap is a fan-out guard rail, not an invariant anything downstream depends on, so two concurrent creates landing 11 rows where 10 was advertised has no consequence beyond the count. The same-profile case you mention is not a race that surfaces either: the unique index refuses the second insert and the service maps it to a 409, which `test_one_profile_per_organization` covers, alongside a second test for the rename collision found later in review.

Serializing the whole plane on the organization row to make a soft cap firm would cost more than the anomaly does.

src/gateway/services/tenancy/organization_guardrail_service.py
```diff
@@ -0,0 +1,670 @@
...
+        await self._commit()
+        await self.db.refresh(guardrail)
+        return await self._public(guardrail)
+
+    async def update_guardrail(
+        self, *, user: User, guardrail_id: uuid.UUID, request: OrganizationGuardrailUpdate
+    ) -> OrganizationGuardrailPublic:
+        """Apply the fields this request set, leaving the rest as they were."""
+        organization_id = await self._manageable_organization_id(user)
+        guardrail = await self._get_or_404(organization_id, guardrail_id)
```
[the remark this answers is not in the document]
njbrake: Not taking this one either, but the mechanism is real so it is worth saying why rather than deferring silently.

Two concurrent PATCHes replacing the scope can leave the union rather than either submission. What that is not is a safety problem on this surface: a union means the guardrail runs in more workspaces than either admin asked for, so the failure narrows what succeeds rather than widening it, which is the direction #655/#678 cares about. A guardrail running somewhere it was not meant to is visible and correctable; one silently not running is the failure this plane exists to prevent.

The neighbouring code takes no lock for the same shape either (`organization_service._apply_workspace_assignments` replaces a member's workspace grants the same way). Worth a lock if two admins editing one guardrail's scope in the same instant turns out to be a real workflow, which is not something I can justify serializing the plane for today.

src/gateway/services/tenancy/organization_guardrail_service.py
```diff
@@ -0,0 +1,670 @@
...
+    WorkspaceNotFoundError,
+)
+from gateway.services.tenancy.organization_service import OrganizationService
+from gateway.services.url_safety import UnsafeURLError, validate_mcp_url
+
+# What one organization may mandate. Every guardrail in scope for a workspace is
+# one more call the request waits on before the provider is reached, and
+# `run_input_guardrails` runs them one after another, so this bounds latency an
+# organization spends on every request rather than only rows it stores.
+MAX_GUARDRAILS_PER_ORGANIZATION = 25
```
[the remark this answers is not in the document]
njbrake: Half right, and the half that is right was my comment rather than the code. Corrected in 04a0a1e62.

The cap never bounded latency and the comment claimed it did. `run_input_guardrails` awaits entries one at a time with a 30s timeout each, so an unavailable set costs the sum. The comment now says that, and the cap is down from 25 to 10, which puts what this plane adds near the 8 a single request may already carry in its own `guardrails` field. One thing bounds the bad case in practice and is now written down: a fail-closed entry raises on the first failure, so reaching the worst case needs a scope full of observing entries all pointed at something unreachable.

The real fix is the one you name, one admission deadline across the whole set, or concurrent evaluation. Both change the shared execution path rather than this plane, and concurrency is not free here (the guardrails container runs a threadpool of 8, so fanning 10 requests at it trades one queue for another). I have left it out of #654 deliberately and raised it rather than deciding it inside a PR about configuration.

src/gateway/services/tenancy/organization_guardrail_service.py
```diff
@@ -0,0 +1,670 @@
...
+    the only direction the request path enforces
+    (`services.guardrails.run_input_guardrails`); an organization that could
+    store an output-direction mandate would be storing one nothing runs.
+    """
+
+    profile: str = Field(
+        min_length=1,
+        max_length=128,
+        description="Profile name configured on the guardrails service, unique within the organization",
+    )
```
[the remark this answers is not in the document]
njbrake: Real, fixed in 04a0a1e62, and the consequence you name checks out.

I confirmed the reservation half rather than taking it on trust: `prepare_gateway_tools` releases on `HTTPException` and `SQLAlchemyError`, and the `ValidationError` that `GuardrailConfig` raises on an empty profile is neither, so the hold would have been stranded on `users.reserved` until the budget's next reset.

Both schemas now use `StringConstraints(strip_whitespace=True, min_length=1, max_length=128)`, so a whitespace-only profile is refused at ingress and a padded one is stored trimmed. The service no longer strips, since the schema already did. Covered by a test that asserts both halves, including the line 191 case you point at.

web/src/client/index.ts
```diff
@@ -436,4 +436,12 @@ export type WorkspaceCodeExecutionPolicy =
 export type UpdateWorkspaceCodeExecutionPolicyRequest =
   Schemas["WorkspaceCodeExecutionPolicyUpdate"]
 
+// The organization-level guardrail plane above the deployment-wide guardrail
+// settings; see `src/gateway/services/tenancy/organization_guardrail_service.py`.
+export type OrganizationGuardrail = Schemas["OrganizationGuardrailPublic"]
+export type CreateOrganizationGuardrailRequest =
+  Schemas["OrganizationGuardrailCreate"]
```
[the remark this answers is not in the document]
njbrake: Already fixed, in 8f8838e91, which landed before this thread was posted; GitHub has it marked outdated for the same reason. `CreateOrganizationGuardrailRequest` is wrapped in `Defaulted` over `enabled`, `on_unavailable`, `mode` and `applies_to_all_workspaces`.

Worth recording how it got in: I typechecked locally with `tsc --noEmit -p tsconfig.json`, and the root config is a solution file with project references, so that command checked nothing. CI runs `pnpm run typecheck`, which is `tsc -b --noEmit`. That is the command I have used since.

src/gateway/AGENTS.md
```diff
@@ -89,16 +89,18 @@ wired into the lifespan, standalone-only.
...
+`ToolContext` for the tool loop to read (guardrails, which never enter that loop, go
+straight into the pre-provider check instead). A lower layer may only narrow what the
+layer above it permits, so a workspace row is a veto and a refinement and never a grant,
+and no row means no narrowing (MCP, which has no deployment-level server list to narrow,
+is the stated exception). The
 question is [#655](https://github.com/mozilla-ai/otari/issues/655) and the reasoning is in
 the PR that answered it,
 [#678](https://github.com/mozilla-ai/otari/pull/678). Read both before building
-either of #654 or #656, each of which settles its own surface under the rule above.
+#656, which settles its own surface under the rule above.
```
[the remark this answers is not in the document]
njbrake: Fixed in 04a0a1e62. Real: the line began with `#656`, so MD018 is right about it. Written as a linked issue reference mid-sentence now, which also makes it navigable rather than bare.

src/gateway/services/guardrails.py
```diff
@@ -188,9 +233,14 @@ async def run_input_guardrails(
...
     # maps it to a 400.
+    credentials = credentials or {}
     if guardrails:
         await asyncio.gather(
-            *(validate_mcp_url(g.url, has_authorization_token=False) for g in guardrails if g.url is not None)
+            *(
+                validate_mcp_url(g.url, has_authorization_token=bool(credentials.get(g.profile)))
+                for g in guardrails
+                if g.url is not None
+            )
```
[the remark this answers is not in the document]
njbrake: Real, fixed in 04a0a1e62, and independently found by two other reviewers.

A credential now requires the entry to name its own `url`, so the https requirement applies unconditionally rather than depending on how the operator happened to set `guardrails_url`. Both halves you ask for are covered: at write time by `_require_url_for_credential` plus `validate_mcp_url`, checked against the effective pair so clearing the url from a credentialed entry is refused too; and on the request path by the existing SSRF pass, which now sees `has_authorization_token=True` for the entries that carry one and only those.

I did not take the alternative of validating the effective default URL, because that setting can change after the row is written and the row cannot see it. Three regression tests: the create path, adding a credential to a url-less entry, and clearing the url from a credentialed one.

docs/public/otari.postman_collection.json
```diff
@@ -2353,6 +2353,149 @@
...
+          "name": "Update Organization Guardrail",
+          "request": {
+            "body": {
+              "mode": "raw",
+              "options": {
+                "raw": {
+                  "language": "json"
+                }
+              },
+              "raw": "{\n  \"applies_to_all_workspaces\": false,\n  \"credential\": \"string\",\n  \"enabled\": false,\n  \"mode\": \"block\",\n  \"on_unavailable\": \"block\",\n  \"profile\": \"string\",\n  \"url\": \"string\",\n  \"validate_kwargs\": {},\n  \"workspace_ids\": [\n    \"00000000-0000-0000-0000-000000000000\"\n  ]\n}"
```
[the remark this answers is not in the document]
njbrake: Fixed in de748d538, and the finding is right that the sample would be refused.

Worth saying where it came from, since the file is generated and hand-editing it would fail `make postman-check`. `scripts/generate_postman.py` derives every string field as `"string"`, so `"url": "string"` appears at four call sites, three of which predate this PR (`/tool-settings` and both `/mcp-servers` bodies, which have the same property: a URL that `validate_mcp_url` would refuse).

Rather than change the generator's placeholder for the whole collection, which is well outside what #654 asks, I gave both guardrail request models a schema-level example. `example_for_schema` prefers an explicit `example`, and `CreateSearchToolRequest` already carries one for the same reason, so this is the idiomatic fix here. The regenerated sample is now a body that would actually be accepted:

```json
{
  "credential": "sk-guardrails-...",
  "mode": "monitor",
  "url": "https://guardrails.internal.example/validate"
}
```

Both examples are asserted to validate against their own schemas rather than eyeballed. The other three sites keep the generic placeholder; they are a pre-existing property of the generator and not this PR's to change.

PR #754, opened 2026-08-24
feat(dashboard): add a no-op telemetry seam an overlay build can replace

### LINKED ISSUE #738: Telemetry seam: the shell has no way for an overlay to observe navigation or the auth funnel

njbrake:
## Description

Adds the no-op telemetry module `web/` needs so an overlay build can observe the auth funnel and sidebar navigation without editing an Otari source file (ARCHITECTURE.md, cardinal rule 6).

`web/` carrying no telemetry is correct, not an oversight, and stays that way: the base tracker records nothing, `web/package.json` declares no analytics vendor, and `overlayTelemetry.test.ts` fails if either stops being true. This is a seam, not telemetry.

`src/shared/telemetry/` holds three modules and only one is swapped. `overlayTelemetry.ts` is the seam, reached by its `@/…` specifier so a build-time alias can replace it; `events.ts` (the event names) and `types.ts` (the contract) are base-owned, so a replacement cannot drift from the vocabulary its call sites use. The twelve names are `otari-ai/frontend/src/shared/helpers/mixpanel.ts`'s `EVENTS` verbatim, checked against a clone, because a superset build reports into the analytics workspace the platform already reports into.

Consent travels through the seam rather than being assumed inside it. `Telemetry.consent` is `"unknown"` here, meaning nothing has been asked, which is not a synonym for granted, and `src/app/TelemetryIdentity.tsx` withholds the identity until a decision is stored. That is otari-ai's `AnalyticsIdentity.tsx` gate kept on the base side. The actor and the active organization go together because `GET /v1/organizations/me` reports both in one response and is the only management route that reports the caller at all.

Eleven of the twelve fire from unmodified `web/` source. `CHECKOUT_CANCELLED` is named and fired by nobody: this gateway meters spend but holds no wallet, so no base page sees a checkout return. It is in the catalog so the overlay that owns that return uses the platform's existing string instead of a second spelling, and a test asserts it is the only unfired name.

Three judgement calls worth a reviewer's attention:

1. `FORM_VALIDATION_FAILED` fires only from the sign-in screen, which deliberately keeps its submit live and validates on submit. Signup disables its submit until the form is complete, so there is no moment to record and none was manufactured.
2. `VerifyEmailPage` gained an effect, which its docstring previously said it deliberately had none of. Verification runs on arrival, so there is no handler, and TanStack Query v5 dropped the query `onSuccess`. It is ref guarded against StrictMode's double run, and that paragraph is rewritten.
3. The seam imports its contract as `@/shared/telemetry/types` rather than `./types`. A replacement is a file in the overlay's tree, so a relative sibling path is the one thing in the module that could not survive being copied.

Coordinated with #736 (PR #748): neither of us touches `web/vite.config.ts`, since the override table belongs to the superset build's config and `web/` has no overlay root to resolve against. #748 owns the seam guard test and the `registry.ts` relative-import fix. Once it merges I owe one line adding `shared/telemetry/overlayTelemetry.ts` to its `SEAM_MODULES`.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes #738

## Checklist

This change is dashboard only, so the checks that apply are the `web/` ones: `pnpm --dir web run lint`, `pnpm --dir web run typecheck`, and `pnpm --dir web test`, all green at 87 files and 1263 tests. `make lint` was run too and passes. The gateway test suite was not run because no Python changed, and the API contract did not change, so there is no spec to regenerate.

A note on running the suite in parallel: a `pnpm test` on a loaded machine produced spurious `Test timed out in 5000ms` failures across files this branch never touches. That is contention, not this change. It was confirmed by running the full suite on a throwaway checkout of `main` (green) and this branch with `--no-file-parallelism` (green, 1263 tests).

Tests added: the base default is inert and stable and imports nothing but its contract; the manifest declares no vendor; the catalog matches the platform's strings; every name except `CHECKOUT_CANCELLED` is fired from a base source file, and that one from none; and the identity half plus each of the six call sites, each mocking the seam by the `@/…` specifier an alias replaces.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

The code, tests and prose here were written by Claude Opus 5 through back and forth with @njbrake; the reasoning and the decisions are his. The design was checked against a clone of `mozilla-ai/otari-ai` rather than recalled: the event strings, the consent gate, and the tab-name derivation are each read off that tree and named where they are used. The seam's placement and the `@/…` specifier rule were settled in conversation with the session working #736 so the two seams share one mechanism.


web/src/app/AppShell.tsx
```diff
@@ -102,13 +117,28 @@ function NavRowLink({
...
       activeOptions={{ exact: true }}
-      onClick={onNavigate}
+      onClick={() => {
+        // Only a move: clicking the row you are already on is not a navigation,
+        // and counting it would inflate whichever page people sit on longest.
+        // The context is read from the destination rather than from the rail
+        // that was showing, so a row that crosses between the two is recorded
+        // where it landed.
+        if (!isActive) {
+          recordEvent(TELEMETRY_EVENTS.TAB_CHANGED, {
```
khaledosman: `TAB_CHANGED` misses the two rows that actually cross rails. Both context-switch links go around `NavRowLink`: the back-arrow at `AppShell.tsx:611` (`to={workspaceLanding?.to ?? "/"}`) and the Organization row at `:762` (`to={organizationLanding?.to ?? "/organization/members"}`). Both navigate to a registry destination (`lastLocation("workspace"|"organization")`), so entering and leaving the organization rail records nothing while every row inside it records.

The new test asserting "names the rail the destination belongs to, not the one on screen" covers a registry row that crosses; these two dedicated switch rows are the uncovered case. Give them the same `recordEvent` call, or lift the recording into a helper both paths use.

web/src/shared/telemetry/types.ts
```diff
@@ -0,0 +1,81 @@
...
+  organizationName: string
+  /** The actor's role in that organization ("owner", "admin", "member"). */
+  role: string
+}
+
+/** The tracker a call site reaches, whichever build supplied it. */
+export interface Telemetry {
+  /** The stored decision, for a caller that must not assume one. */
+  readonly consent: TelemetryConsent
+  readonly recordEvent: (
```
khaledosman, line 86: Consent gates `identify` but not `recordEvent`, and the contract does not say so. `TelemetryIdentity` refuses to identify unless `consent === "granted"`, while all six `recordEvent` call sites fire unconditionally.

A replacement author reads this interface, sees one gated call in base code, and has nothing telling them the other gate is theirs. State it here: the base fires `recordEvent` whatever consent says, so a replacement owns that check.

web/src/features/auth/Login.tsx
```diff
@@ -242,8 +244,14 @@ function LabelRow({
...
 export function Login() {
   const { login, isSigningOut } = useAuth()
+  const { recordEvent } = useTelemetry()
   const { sign_in_methods, mail_ready } = useDeployment()
   const usesPassword = sign_in_methods.includes("password")
+  // Which credential this attempt presented, recorded on every outcome so the
+  // two sign-ins a deployment can offer stay separable in the funnel. The
+  // platform's own vocabulary for this property, minus the OAuth values whose
+  // buttons are still otari#651's and #652's.
+  const authenticationMethod = usesPassword ? "password" : "master_key"
```
khaledosman: `authentication_method` is derived from what the deployment offers, not from what the attempt presented. `usesPassword ? "password" : "master_key"` is exact only while `sign_in_methods` has exactly two values.

#652 adds `passkey`, and on a deployment offering `["password", "passkey"]` a passkey sign-in would report `authentication_method: "password"` with nothing failing. Derive it from the credential `readCredential` returned, so a third method cannot inherit an existing label.

web/src/tests/telemetry.ts
```diff
@@ -0,0 +1,66 @@
...
+ * then agree with themselves about it.
+ *
+ * A test reaches it through the async factory form, because `vi.mock` is hoisted
+ * above the imports and a factory referencing one directly would run before it
+ * is initialized:
+ *
+ * ```ts
+ * vi.mock("@/shared/telemetry/overlayTelemetry", async () => {
+ *   const { telemetrySpy } = await import("@/tests/telemetry")
+ *   return { useTelemetry: () => telemetrySpy }
```
khaledosman: The mock factory should wrap the hook in `vi.fn()`. `imports-and-modules.md` ("Test mocking"): "Wrap a hook-like export in `vi.fn()` when you replace one, so it behaves as a function identity that lint and the React tooling can follow" (`useNavigate: vi.fn(() => navigate)`).

This snippet is copied verbatim into all six test files, so fixing the canonical one here fixes them:

```ts
return { useTelemetry: vi.fn(() => telemetrySpy) }
```

web/src/features/auth/Login.tsx
```diff
@@ -290,21 +318,29 @@ export function Login() {
...
+        failValidation(
+          "email",
+          "Enter a valid email address.",
+          "email_invalid_format",
+        )
         return null
       }
       if (!password) {
-        fail("password", "Enter your password.")
+        failValidation("password", "Enter your password.", "password_required")
```
khaledosman, line 345: `password_required` is the one validation reason with no test. `email_required`, `email_invalid_format` and `master_key_required` each have one in `Login.test.tsx`.

khaledosman (approved):
Five findings, none blocking. Gates that apply are clean: no `schema.ts` / `routeTree.gen.ts` drift, no new dependency, no new page (so no screenshot entry owed), and `src/shared/telemetry/` sits under an existing layer, so `architecture.test.ts`'s closed-set assertion is unaffected.

*Review written by Claude Code (Opus 5).*

web/src/app/TelemetryIdentity.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+  const actorId = data?.organization_member_id
+  const organization = data?.organization
+  const organizationId = organization?.id
+  const organizationName = organization?.name
+  const role = data?.role
+
+  useEffect(() => {
+    if (consent !== "granted") {
+      return
```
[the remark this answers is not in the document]
njbrake: Real, and fixed in c4fb619.

You are right that the gate was written as an early return, so it withheld a *new* identity but never took back one already sent. `TelemetryIdentity` now keeps an `identified` ref and calls `identify(null)` on the transition out of `"granted"`, which is the same "forget the last actor" signal a sign-out uses.

Two tests: grant then revoke expects `identify(null)`, and a browser that never consented expects no `identify` at all, so the fix cannot degrade into a spurious forget on every mount.

web/src/features/auth/AuthContext.tsx
```diff
@@ -56,6 +59,13 @@ export function AuthProvider({ children }: { children: ReactNode }) {
   const pendingSignOutsRef = useRef(0)
 
   const logout = useCallback(() => {
+    // Recorded before anything is torn down, and from here rather than from the
+    // account menu, because this is also the path a 401 takes: a session that
+    // expired or was revoked ends the same funnel a deliberate sign-out does.
+    // `identify(null)` is the other half of it, and it is what stops the next
+    // session in this tab from being attributed to the identity that just left.
+    recordEvent(TELEMETRY_EVENTS.LOGOUT)
+    identify(null)
```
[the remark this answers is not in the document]
njbrake: Real, and fixed in c4fb619. A second review pass reproduced it independently: three `logout()` calls produced three `LOGOUT` events.

`AuthContext` now holds a `sessionOpenRef`, seeded from the same stored marker the rendered state is, cleared on the first `logout` and set again by `login`. That covers both halves of what you describe: the concurrent path records once, and a 401 arriving with no session open records nothing rather than inventing an ending.

Two tests added, one for three concurrent calls and one for the no-session-open case.

web/src/features/auth/ResendVerificationPage.tsx
```diff
@@ -43,7 +46,13 @@ export function ResendVerificationPage() {
       return
     }
     resend.mutate(email.trim(), {
-      onSuccess: () => goToPublicAuthPage("#/check-email?type=resend"),
+      onSuccess: () => {
+        // On the send rather than on the click, so a request the gateway
+        // refused is not counted as a link somebody was sent. Enumeration-safe
+        // like signup, so there is nothing in the response to read.
+        recordEvent(TELEMETRY_EVENTS.RESEND_VERIFICATION_CLICKED)
```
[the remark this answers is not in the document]
njbrake: Keeping this as it is, and here is the reason rather than a silent resolve.

The platform fires this event in `onSuccess` too: `frontend/src/routes/resend-verification.tsx` in `mozilla-ai/otari-ai` calls `track(events.RESEND_VERIFICATION_CLICKED)` inside the mutation's success handler. I checked a clone rather than recalling it.

That matters more than the name reading oddly, because the reason this PR pins the twelve event strings character for character is that a superset build reports into the analytics workspace the platform already reports into. Moving *when* an event fires while keeping its name is the same hazard as renaming it, just in the time dimension: one series would silently change meaning mid-history and nothing would fail.

Your two observations are both correct on the merits, and the name is inaccurate. I think that makes it a question about an existing platform series rather than one this PR should settle on its own, so I have flagged it to @njbrake. If he wants the semantics changed, the right shape is to change it on both sides at once.

The behavior is documented at the call site and pinned by a test asserting a refused send records nothing.

web/src/app/AppShell.tsx
```diff
@@ -102,13 +117,28 @@ function NavRowLink({
...
+            tab_name: tabNameForPath(to),
+            context: navContextForPath(to),
+          })
+        }
```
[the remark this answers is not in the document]
njbrake: Real, and fixed in c4fb619. @khaledosman raised the same gap independently, which is a fair sign it was the significant miss in this PR.

Centralized rather than duplicated, as you suggested: a `useRecordNavigation` hook now owns the `!isActive` guard, the tab name and the context, and `NavRowLink` plus both footer rows call it. Tests cover entering the organization rail by the `Organization` row and leaving it by the back arrow.

web/src/app/AppShell.tsx
njbrake: Real, and fixed in c4fb619. Thank you, this was the significant miss: Copilot flagged the same two rows independently.

Lifted into a helper rather than copied, which is the option you offered second and the better one here. `useRecordNavigation` owns the `!isActive` guard, the tab name and the context; `NavRowLink` and both footer rows call it. Tests cover entering the rail by the `Organization` row and leaving it by the back arrow.

Worth noting the two rows navigate to `lastLocation(...)`, so the destination varies; reading the context from the destination rather than from the rail on screen is what makes that come out right.

web/src/shared/telemetry/types.ts
njbrake: Real, and fixed in c4fb619.

You are right that the split was undocumented, and it is deliberate rather than an oversight: an identity is the one call that hands a tracker something durable about a person, so the base withholds it, while the platform's own consent gate for events lives inside `trackEvent` in `shared/helpers/mixpanel.ts` and belongs to a replacement.

`Telemetry` now says that on the interface, and `recordEvent` carries its own line: not consent-gated by the base, a replacement must check `consent` itself. That is also the answer to why `consent` is on the interface at all, since an implementation cannot gate what it is never told.

web/src/features/auth/Login.tsx
njbrake: Real, and fixed in c4fb619. Good catch: this would have failed silently, which is the worst shape for a bug in a funnel.

`authenticationMethod` is now a function of the `SignInCredential` that `readCredential` returned (`"masterKey" in credential`), so it describes what the attempt presented rather than what the deployment offers, and #652's passkey cannot inherit an existing label.

While in there I also stopped collapsing the two refusals. `SignInResult` now carries the status, so `LOGIN_FAILED` distinguishes 401 (wrong credential) from 403 (master-key sign-in retired), which this screen already treats as load-bearing. The error code is the platform's `http_<status>` form rather than a bucket of our own, for the same reason the event names are verbatim.

web/src/tests/telemetry.ts
njbrake: Real, and fixed in c4fb619. Correct citation, and I had read that guidance and still written the bare arrow.

Fixed in the canonical snippet in `src/tests/telemetry.ts` and in all seven call sites, with a line in the harness docstring saying why so the next copy does not reintroduce it.

web/src/features/auth/Login.tsx
njbrake: Real, and fixed in c4fb619. `password_required` now has a test alongside the other three.

It also turned out to be the reason with the most to say, since it is the only one reached with a valid email already typed.

PR #759, opened 2026-08-24
feat(dashboard): add a seam an overlay resolves entitlements through

### LINKED ISSUE #758: Give the dashboard a seam to resolve entitlements from the server
The display half of what #659 left. That issue closed with the axis narrowed by #718: the flag axis is gone, entitlement answers at deployment grain, and enforcement is `require_capability` on routes, which #756 landed. All settled. What remains is that the dashboard still resolves the axis from `BASE_CAPABILITIES`, a build-time constant that grants no overlay capability, so a nav row or page tagged with one hides in every build, including a build whose routes are entitled and serving.

There is no way for an overlay to supply the client-side answer: the five seam modules in `overlaySeams.test.ts` all contribute inside the shell, and `useEntitlements`' docstring ("supplies its own answer by rendering `EntitlementProvider` above the shell") describes a mount point that does not exist. The superset already serves the server's answer at `GET /v1/entitlements`, resolved through `EntitlementPort` per request, and has nowhere to feed it in.

The ask: a sixth seam, for example `app/overlayEntitlementResolver.tsx`, whose base default renders its children unchanged and which the shell mounts above nav and routes. An overlay's version fetches `/v1/entitlements` and renders `EntitlementProvider` with the answer; `EntitlementGate`'s existing `loading` prop already covers the resolving state. Added to `SEAM_MODULES` so the completeness check guards it.

An alternative shape, if a seam feels wrong here: the base itself probes `/v1/entitlements` and falls back to the constant on 404. Fewer moving parts, but it puts an overlay-only wire path in the base, which is why the seam is the recommendation. Vocabulary questions stay with mozilla-ai/otari-ai#1806; nothing here depends on them, since the capability names on the wire are the ones `require_capability` already enforces.

njbrake:
## Description

The dashboard decides whether to show a nav row or a page by asking whether this deployment is licensed for the feature behind it. Until now it answered that question from a hard-coded list baked into the build, and the list is empty, so anything tagged as a paid feature was hidden everywhere, even on a deployment that had paid for it and whose server was already serving it. The enterprise build knows the real answer and had nowhere to hand it over.

This adds one small file the enterprise build can swap out to supply that answer, and mounts it at the top of the app so both the sidebar and the pages read from it. In this open-source build the file does nothing at all: it passes its contents straight through, so the dashboard behaves exactly as it did before.

## PR Type

- [x] New Feature

## Relevant issues

Fixes #758

## Checklist

<sub>Dashboard-only change, so the checks that apply are `web/`'s: `pnpm run lint`, `pnpm run typecheck`, `pnpm test` (1372 passed / 96 files) and `pnpm run build`, all clean with no drift in either committed generated artifact. No API contract change, so nothing to regenerate.</sub>

### What changed

- **`web/src/app/overlayEntitlementResolver.tsx`** (new): the sixth seam. Its base default renders its children unchanged, deliberately rather than rendering `EntitlementProvider` with `BASE_CAPABILITIES`, which looks equivalent: a provider shadows whatever is above it, so that version would overwrite a superset build's own answer and would silently hand the empty base answer to every test that wraps the shell in a provider of its own.
- **`web/src/app/AppShell.tsx`**: `AppShell` becomes a wrapper that mounts the seam above a new `AppShellChrome` holding the previous body. Two components rather than one because a provider is invisible to the component that renders it, and `useNavVisibility` is called in that body.
- **`web/src/overlaySeams.test.ts`**: sixth entry in `SEAM_MODULES`, so the completeness check guards it.
- **`web/src/app/overlayEntitlementResolver.test.tsx`** (new): three tests on the base default and two on the mount point. The mount-point tests build the whole superset shape, mocking the resolver and `overlayNavItems` together, then assert a capability-gated row reaches the rail, a capability-gated page reaches the `Outlet`, and a still-resolving answer is not reported as a page this deployment does not serve. All three fail with the wrapper reverted.
- Docs: a `web/AGENTS.md` bullet pair, one sentence in `useEntitlements.tsx` (whose docstring described a mount point that did not exist), and the `ARCHITECTURE.md` line that said an overlay answers "by rendering `EntitlementProvider`" now names where.

### One shell fix came with it

A cold review caught that the seam alone was not enough. The shell decides whether the current path is gated off from the same predicate the rail uses, and that predicate reads capabilities while ignoring `isLoading`. Right for the rail, wrong for the route: an entitled deep link into an overlay page rendered the "is not available here" panel, asserting the deployment does not serve the page, and only then the page. The second commit makes the shell render the router's pending copy while the axis is resolving, which is the branch `EntitlementGate` has had a `loading` prop for since it was written. Base behavior is unchanged, `isLoading` being always false here, so it is a branch only a superset build's asynchronous resolver reaches. Without it every replacement would have had to hold its children back until its query settled, which blanks the whole dashboard on every load.

### Why the seam and not the base probe

The issue names an alternative: the base requests `/v1/entitlements` itself and falls back to the constant on 404. It is rejected here for a reason stronger than the issue's own ("an overlay-only wire path in the base"). This gateway mounts no such route at all, `require_capability` being `EntitlementPort`'s only caller, so the request would 404 on every load of every OSS deployment. It would also have to be hand-written, since `web/src/client/schema.ts` is generated from `docs/public/openapi.json` and has no operation to type it with. The constant is not a placeholder standing in for a fetch; the build that resolves the axis from a server is the build that owns the request for it.

## AI Usage

**AI Model/Tool used:**

Implemented from the issue by an agent working with @njbrake, who reviewed the design choice (seam vs. base probe) before the code was written. The base default's passthrough shape and the two-component split of `AppShell` were both derived from reading the existing seam modules rather than assumed.


web/src/app/AppShell.tsx
```diff
@@ -894,7 +934,13 @@ export function AppShell() {
             className="flex-1 overflow-y-auto focus:outline-none"
           >
             <div className="mx-auto flex max-w-[1800px] flex-col gap-6 px-4 py-5 md:px-6 md:py-6">
-              {routeIsGatedOff ? (
+              {routeIsGatedOff && entitlementsResolving ? (
+                // The router's own pending copy, so a route waiting on the
+                // entitlement answer reads as the same kind of wait as a route
+                // waiting on its chunk (`defaultPendingComponent` in
+                // `app/router.tsx`).
+                <div role="status">Loading page…</div>
```
khaledosman: This duplicates `app/router.tsx:18`'s `defaultPendingComponent` character for character, and the comment above it says being that same copy is the point — which is exactly the case for not writing it twice. Export it once (both files are in `src/app/`) so the two cannot drift into two different waits.

Minor, and unrelated: the branch is `routeIsGatedOff && entitlementsResolving`, so in a superset build a route gated off by `surface` — an answer already known — also renders the pending copy for as long as the entitlement query is in flight. Narrowing the branch to the capability axis would avoid that flash. Base behavior is unaffected either way, `isLoading` being always false here.

khaledosman (approved):
Checks run against the head revision: `pnpm lint` + `typecheck`, and `client:generate` plus a `build` confirming no drift in either committed artifact.

The passthrough base default is the right shape, and the reason given for it is the real one — an `EntitlementProvider` carrying `BASE_CAPABILITIES` would shadow a superset build's own answer, and the third test in the file is what keeps that from being re-"simplified" later. Splitting `AppShell` so the resolver sits above `useNavVisibility` is also correct: mounting it inside the chrome would have answered the pages and not the rail that decides which pages are reachable.

One finding inline.

web/src/app/AppShell.tsx
njbrake: Both taken, in 96ac771.

On the duplication: agreed, and a comment naming the source is not a mechanism. `app/PendingPage.tsx` is now the one declaration; `router.tsx` passes it as `defaultPendingComponent` and the shell renders it. It is a third module rather than an export from `router.tsx` because `AppShell` is what `routes/__root.tsx` names and `router.tsx` reaches that file through `routeTree.gen.ts`, so importing the router from the shell would close that circle.

On the second point: also right, and it was not only a flash. `isPathVisible` composes the two axes and reports only that something hid the route, so the branch could not tell them apart. The surfaces are settled by the bootstrap before the page renders, so waiting on the entitlement query held back a panel that was already correct and that no query could change. `useNavVisibility.ts` now exports the surface half beside its composition (`useSurfaceVisibility`), and the shell asks it separately rather than rebuilding the predicate at the call site, so there is still one implementation of the axis. Covered by a test that fails without the narrowing: a deep link to `/organization/provider-keys`, gated on a surface `STANDALONE_SURFACES` does not report, answers with the panel while `isLoading` is true.

Base behavior is unchanged either way, as you say, `isLoading` being a literal false against the constant.

PR #760, opened 2026-08-24
feat(dashboard): manage a workspace's MCP servers from Tools

### LINKED ISSUE #739: Dashboard surface for per-workspace MCP servers
> **Re-scoped 2026-08-24.** As filed this asked for the whole MCP configuration plane. Almost all of it is already on `main`, and only the dashboard half is missing. Original text at the bottom.

- Lands in: `otari`

## What is already there

- **Table** `WorkspaceMcpServer` (`src/gateway/models/entities.py:1369`) with `UniqueConstraint("workspace_id", "name")`, which is also the answer to #591 inside a workspace
- **Service** `src/gateway/services/tenancy/workspace_mcp_server_service.py` (480 lines), with its own not-found, already-exists and unsafe-URL errors
- **Routes** `src/gateway/api/routes/workspace_mcp_servers.py`, registered in `api/main.py`
- **Honored on the request path** `_pipeline.py:1673` calls `resolve_workspace_mcp_servers` for the requesting key's workspace

## The gap

- **No dashboard surface.** Nothing under `web/src` manages MCP servers. It belongs alongside `SearchToolsCard` and `WorkspaceCodeExecutionPolicyCard` on the `/tools` page.
- **The nav registry's comment is stale.** `web/src/app/nav/registry.ts:117` still says "the gateway has no MCP server registry", which stopped being true when the service landed. Worth fixing in the same PR so the next reader is not misled the way this issue was.

## Definition of Done

- A workspace's MCP servers are manageable from `/tools`, with duplicate names surfaced as the 409 the constraint already produces.
- The stale nav comment is gone.

<details>
<summary>Original text as filed</summary>

Asked for the tables, service, routes, org-role gating, dashboard surface and `mcp_client` wiring as if none existed. It was written from `docs/architecture/rehome-inventory.md` in the private repo, which predates the tenancy work that landed on 8/20 and 8/21.
</details>

njbrake:
## Description

Otari already let a workspace register MCP servers, but only over the API with the master key. There was no way to do it from the dashboard, so anyone setting one up had to reach for curl. This adds that screen. You pick a workspace, see the MCP endpoints it has registered, and add, edit, enable, disable or delete them. Requests can then name a server by id instead of carrying its URL and its access token every time.

Two details the screen is careful about. Access tokens are stored encrypted and never sent back to the browser, so the form cannot show you one; leaving the token box empty on an edit keeps the token that is already stored rather than wiping it, and there is a separate checkbox for when you really do want it gone. And the limit on how many servers a workspace may hold is not repeated in the browser: going over it shows the gateway's own refusal, which names the real number, so the two cannot drift apart.

MCP gets its own entry under Tools rather than joining the Web search and Code execution views. Those two narrow a backend the deployment configures; there is no deployment-wide MCP server list for these rows to narrow, so a row here is the configuration rather than a narrowing of one. The same card also closes the combined `/tools` page.

Checks run: `pnpm --dir web run lint`, `pnpm --dir web run typecheck`, `pnpm --dir web test` (1378 passing), and `pnpm --dir web run build`. No Python changed, so no OpenAPI or Postman regeneration is owed.

Also drops the nav registry comment claiming the gateway has no MCP server registry, which stopped being true when #724 landed.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [x] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes #739

## Checklist

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

The scope, the placement decision and the call not to copy the server-side limit into the client were settled in discussion with @njbrake before any code was written. The backend was verified as already present on `main` first, since this issue was re-scoped after being filed against a stale inventory.

**NOTE:**

## Summary

- Added a workspace MCP server management dashboard under **Tools → MCP servers**.
- Users can select a workspace, view servers, and add, edit, enable, disable, or delete entries.
- Protected bearer tokens remain encrypted and hidden from the browser. Edits can preserve, replace, or remove tokens.
- Server limits remain enforced by the gateway. Gateway errors are shown to users.
- Added navigation, API hooks, shared types, documentation, fixtures, and tests.
- Separated MCP servers from deployment-wide tool views and removed the stale navigation comment.
- Reused the shared `SecretField` component for credential inputs.


web/src/features/tools/WorkspaceMcpServersCard.tsx
```diff
@@ -0,0 +1,272 @@
...
+          Add MCP server
+        </Button>
+      </div>
+
+      <p className="text-sm text-muted">
+        Endpoints requests billed to {selected.name} can use by naming their ids
+        in <code className="font-mono">mcp_server_ids</code>. A request may
+        still pass its own servers inline; these are the ones it does not have
+        to carry a URL or a token for. Tokens are stored encrypted and are never
+        shown again.
```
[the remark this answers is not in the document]
njbrake: Applied in 255cf62a. Agreed: without the pronoun the sentence garden-paths as a compound noun ("Endpoints requests") before it resolves as a reduced relative clause, so the extra word earns its place.

This thread shows as outdated because the branch was rebased onto main afterwards to clear a `web/AGENTS.md` conflict with #752, so the line numbers above no longer line up.

_Note: this reply was written by Claude Opus 5, working on @njbrake's PR at his direction. The prose is Claude's._

web/src/features/tools/WorkspaceMcpServersCard.tsx
```diff
@@ -0,0 +1,269 @@
...
+          // hundreds), so the count is the cell and the list is in the dialog.
+          <span>{row.allowed_tools.length} allowed</span>
+        ),
+    },
+    { id: "token", header: "Token", cell: tokenChip },
+    {
+      id: "enabled",
+      header: "Status",
+      cell: (row) => (
+        <Chip size="sm" variant="secondary">
```
khaledosman: A `text-success` span inside `<Chip variant="secondary">` re-skins what the component already styles, which the standards put last after a variable and the component's own prop.

`color` is a separate axis from `variant` on HeroUI's Chip — `@heroui/styles/dist/components/chip.css` defines `.chip--success` setting `--chip-fg: var(--success-soft-foreground)` — and the two cards next door already use it (`WorkspaceCodeExecutionPolicyCard.tsx:197`, `WorkspaceWebSearchCard.tsx:251`).

`color={row.enabled ? "success" : "default"}` also gets the pairing right: `text-success` here sits on the chip's `--default` fill rather than a subtle success fill, which is the combination the design-tokens guide calls out as the one people get wrong.

web/src/features/tools/McpServerDialog.tsx
```diff
@@ -0,0 +1,263 @@
...
+                  <Checkbox isSelected={enabled} onChange={setEnabled}>
+                    Enabled
+                  </Checkbox>
+                  <span className="text-xs text-muted">
+                    A disabled server keeps its row and its token. A request
+                    that names it skips it rather than failing.
+                  </span>
+                </div>
+
+                {urlReason ? (
```
khaledosman: `urlProblem`'s comment says the message lands "under the field rather than arriving as a banner after a round trip", but it renders here, at the foot of the dialog body — five controls below the URL field at `:168`, past the token box, the hint, the allow-list and the Enabled checkbox. It is also not associated with the input, so a screen reader on the URL field is told nothing.

Either move the `role="alert"` paragraph directly under the URL `Field`, or give `Field` an `isInvalid`/`errorMessage` pass-through; HeroUI's `TextField` has both, and `isInvalid` is the component's own prop the standards point at.

khaledosman (approved):
Checks run against the head revision: `pnpm lint` + `typecheck`, and `client:generate` plus a full `build` leaving both committed artifacts (`schema.ts`, `routeTree.gen.ts`) unchanged — so the route addition is accounted for. The screenshot entry for the new page is there, and the route file exports `Route` and nothing else.

Gating the read on `canManageWorkspace` rather than only the writes matches what the service does, and leaving the server's cap uncopied is the right call.

Two findings inline, both presentation.

web/src/features/tools/WorkspaceMcpServersCard.tsx
```diff
@@ -0,0 +1,269 @@
...
+      id: "allowed_tools",
+      header: "Tools",
+      cell: (row) =>
+        row.allowed_tools === null ? (
+          <span className="text-muted">All</span>
+        ) : (
+          // The names themselves would be an unbounded column (the list runs to
+          // hundreds), so the count is the cell and the list is in the dialog.
+          <span>{row.allowed_tools.length} allowed</span>
+        ),
```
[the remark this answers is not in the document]
njbrake: Correct, and fixed in a3dde5fe. Verified against `src/gateway/services/mcp_client.py:88`, which is `allowed = set(cfg.allowed_tools) if cfg.allowed_tools else None`: `[]` is falsy, so it collapses to `None` and every tool is exposed. The cell was saying no tools were exposed when all of them were.

The cell now reads `null` and `[]` identically as "All". This form never sends `[]` (`parseAllowedTools` returns null for an empty input), but the create schema accepts one, so a row can arrive here holding it, which is the case you named.

Covered by a regression test, "reads an empty allow-list as every tool, the way the gateway does", which fails against the old cell.

_Note: this reply was written by Claude Opus 5, working on @njbrake's PR at his direction. The prose is Claude's._

web/src/features/tools/WorkspaceMcpServersCard.tsx
njbrake: Agreed, fixed in a3dde5fe. The chip is now `color={row.enabled ? "success" : "default"}` with no span inside it.

I checked the two things this turns on before changing it: `Chip` does expose `color` as its own axis (`color?: ChipVariants["color"]` in `@heroui/react/dist/components/chip/chip.d.ts`), and `.chip--success` does set `--chip-fg: var(--success-soft-foreground)`, so the foreground comes from the fill rather than from a class fighting it. Both sibling cards use the same prop.

I left `tokenChip`'s "Stored" chip on `variant="secondary"`: it carries no nested color class, so it is not re-skinning anything, and changing it would be a different edit from the one you flagged. Say the word if you want the two chips on one idiom.

_Note: this reply was written by Claude Opus 5, working on @njbrake's PR at his direction. The prose is Claude's._

web/src/features/tools/McpServerDialog.tsx
njbrake: You are right that the comment was describing something the code did not do, and it is fixed in a3dde5fe.

I took the second option you offered, the pass-through, since it is the one the standards point at. `Field` now accepts `isInvalid` and `errorMessage` and renders the message through HeroUI's `FieldError`, which is exported from `@heroui/react` and renders into the field's own error slot. The URL field passes `isInvalid={urlReason !== undefined}`, and the loose paragraph at the foot of the dialog body is gone. Both props are optional, so no other caller of `Field` changes.

The test now asserts the association rather than the text: `expect(getByLabelText("URL")).toHaveAccessibleDescription(/needs an https URL/)`. It fails if the `FieldError` is removed, which is the part that was previously untested.

`urlProblem`'s comment about the message landing under the field is now accurate rather than aspirational.

_Note: this reply was written by Claude Opus 5, working on @njbrake's PR at his direction. The prose is Claude's._

PR #762, opened 2026-08-24
feat(gateway): consult ModelProviderPort when no stored credential serves a candidate

### LINKED ISSUE #757: Consult ModelProviderPort on the request path when no stored credential serves a candidate
#756 landed `ModelProviderPort` with its core adapter, and nothing calls it: the credential ladder in `services/provider_kwargs.py` walks org-scoped keys, then stored instances, then config.yml, and refuses when all three miss. An overlay adapter bound behind the port (hosted inference from a deployment-owned fleet) is therefore unreachable, however correctly it is bound.

The ask: when the existing ladder yields no credential for a candidate, resolve `ModelProviderPort` and ask `resolve_hosted_credential` before treating the candidate as unserved. The core adapter answers `None` for everything, so a deployment with no overlay behaves as it does today, rung for rung.

The port contract already carries what the call site needs: `HostedCredential.response_provider` names the upstream that usage and telemetry key on, and `HostedAccessDeniedError` is the refusal a caller handles without naming the adapter. What this issue adds is only the call, in whichever of `provider_kwargs.py` or `_pipeline.py` owns that decision, plus a test proving the plain build's behavior is unchanged and an overlay-bound adapter is reached.

Ordering note: BYO stays upstream and untouched. The port is asked last, never before an organization's own key, which preserves the precedence ladder as documented.

njbrake:
## Description

Otari can be extended by an outside build that plugs its own pieces into named slots. One of those slots is "where do the credentials come from when a request brings none of its own", and #756 built the slot but nothing ever looked in it, so a build that filled it correctly would still never be asked. This PR adds the question.

When a request names a model and the gateway cannot find any key for it, whether from the organization's own stored key, a provider configured in the dashboard, the config file, or the provider's own environment variable, it now asks the slot whether this particular build can serve the request from inference capacity the deployment itself owns. Otari's own answer is always "no", so a plain install behaves exactly as it did before, down to the same error for the same request. A build that supplies its own answer gets its credentials used, and the request is billed and logged against whichever upstream actually served it. If that build says the organization is not allowed to use that upstream, the caller gets a plain refusal that says nothing about how the decision was made, and the budget hold taken earlier in the request is given back.

**Where the call went, and why.** The issue left this open between `services/provider_kwargs.py` and `api/routes/_pipeline.py`. It went in `_pipeline.py`:

1. `provider_kwargs.py` cannot host it. `provider_store_service.py`'s docstring already states the constraint: "Resolution has to stay synchronous. `get_provider_kwargs` and `resolve_provider_selector` read `config.providers` directly and are called from the streaming hot path, which has no database session." That is the reason stored providers are an in-memory overlay rather than a lookup. Every port method is `async` and a port factory takes a session, so putting the call there would mean making the whole ladder async and re-entering it from `routing/knn.py`, `routing/decide.py`, `routing/compiler.py`, and the CLI.
2. The port keys its access decision on `organization_id`. `provider_kwargs.py` never sees one; `_pipeline.py` resolved it in the preamble already.
3. `resolve_dispatch_provider` is the single funnel all three standalone completion routes use for dispatch, so one call site covers chat, messages, and responses.

**Precedence is unchanged.** A new predicate, `credential_ladder_exhausted`, gates the call: the port is asked only when `get_provider_kwargs` produced nothing at all *and* the provider's own SDK environment variable is unset. It reuses the same env check the keyless placeholder makes, so both agree on what counts as a credential already in hand. BYO is resolved upstream and never displaced.

**Deliberately out of scope, and stated rather than silent:** a routed multi-candidate plan does not reach the port. Those candidates are credentialed by `routing/compiler.py`, which is synchronous and documented as doing no I/O, so what `resolve_dispatch_provider` returns is not what they are called with. Asking anyway would meter a resolve whose answer is discarded, so the code skips it explicitly and says so. `_passthrough.py` and `batches.py` are likewise untouched. Reaching the port from the compiler is separate work.

Three refusals the call site owns, each non-leaky: `HostedAccessDeniedError` becomes a 403 carrying the same detail an organization-scoped model restriction already uses, with the reservation refunded before it surfaces; `None` leaves the candidate exactly as it was; and an adapter naming an upstream any-llm does not implement is a defect in that build, logged in full and surfaced as a generic 502.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [x] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes #757

## Checklist

Tests: `tests/unit/test_hosted_credential_port.py` covers the plain build through the real `build_container()` binding, all four ways a BYO credential keeps the port unasked (config.yml, an organization-scoped key, the provider SDK env var, and a request with no organization), both refusals, alias preservation, and the multi-candidate skip. `tests/integration/test_hosted_credential_request_path.py` boots two real apps from one config, one with an `OTARI_BOOTSTRAP` overlay binding a fleet adapter and one without, and compares what reaches any-llm: `together:gpt-4o` with the fleet key and base versus `openai:gpt-4o` uncredentialed.

Also ran the full unit suite (2332 passed, 8 skipped), the full integration suite (1662 passed, 9 skipped), the OSS-edition smoke script, `make lint`, `make typecheck`, `make openapi-check` and `make postman-check`. The API contract did not change: the new route dependency injects no parameters, so the regenerated spec is byte-identical and neither generated artifact moved.

Docs: `ARCHITECTURE.md`'s "Where the ports are today" callout said nothing on the request path calls a port, which is no longer true; it now names both ports that have callers and the routed-chain gap. `src/gateway/AGENTS.md` gains the last rung of the credential ladder next to the runtime credential stores it sits below.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

Coordinated with the session on #758 (the dashboard entitlement seam, also following #756) to keep the vocabulary disjoint: "hosted credential" and "hosted inference" stay on the `ModelProviderPort` side, using the port's own wording from #756, and "entitlement" and "capability" stay on the `EntitlementPort` side. No files overlap.

One judgement worth flagging for review: a resolved credential re-keys `ResolvedProvider.instance` onto `response_provider`, because the port documents that field as the upstream usage and telemetry name. The pricing and budget gates ran earlier in the preamble against the name the caller asked for, so an overlay whose `response_provider` differs from the public name owes a pricing row for the upstream. That is the same property a routing fallover to a differently priced candidate already has, but it is a real consequence rather than an oversight.

_Note: this pull request description was drafted by Claude Opus 5 via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

### DIFF
--- a/src/gateway/api/routes/_pipeline.py
+++ b/src/gateway/api/routes/_pipeline.py
@@ -808,9 +808,12 @@ async def _serve_from_hosted_credential(
         credential = await port.resolve_hosted_credential(
             organization_id=ctx.organization_id,
             workspace_id=ctx.workspace_id,
-            # The public name the caller addressed, which is the instance rather
-            # than the implementation; ``response_provider`` comes back naming
-            # whichever upstream actually serves it.
+            # The instance rather than the implementation: that is the name
+            # pricing, budgets and usage key on, and for a bare selector it is
+            # the one the caller wrote. (An alias resolves to its target first,
+            # so an aliased request names the target's instance here, never the
+            # alias.) ``response_provider`` comes back naming whichever upstream
+            # actually serves it.
             provider=resolved.instance,
             model=resolved.model,
         )
@@ -863,8 +866,34 @@ async def _serve_from_hosted_credential(
             resolved.model,
         )
         await release_reservation(ctx)
+        # Logged like the 400 and 403 refusals beside it: an operator watching
+        # the activity log during an overlay rollout would otherwise see dropped
+        # traffic as nothing at all.
+        await log_gateway_rejection(
+            db=ctx.db,
+            log_writer=ctx.log_writer,
+            api_key_id=ctx.api_key_id,
+            user_id=ctx.user_id,
+            model=resolved.model,
+            provider=resolved.instance,
+            endpoint=adapter.endpoint,
+            detail=HOSTED_CREDENTIAL_UNUSABLE_DETAIL,
+            status_code=status.HTTP_502_BAD_GATEWAY,
+            started_at=ctx.started_at,
+        )
         raise adapter.error(502, HOSTED_CREDENTIAL_UNUSABLE_DETAIL, ErrorKind.API) from exc
 
+    # Re-keying moves what settlement prices on. The pricing and budget gates
+    # ran in the preamble against the name the caller asked for, and settlement
+    # reprices on ``instance``, so an overlay owes a pricing row for every
+    # ``response_provider`` it returns: without one the row's cost is NULL and
+    # the request settles free, which is `require_pricing` bypassed rather than
+    # enforced. A routed fallover to a differently priced candidate has an
+    # explicit guard for the same hazard (`top_up_reservation_for_attempt`
+    # refuses an unpriced candidate with a 402); this path has none, because the
+    # substitution is the adapter's decision rather than a plan's and refusing it
+    # here would make a build's own fleet unusable until it were priced. An
+    # overlay binding this port owns that.
     kwargs: dict[str, Any] = {"api_key": credential.api_key}
     if credential.api_base is not None:
         kwargs["api_base"] = credential.api_base
--- a/src/gateway/api/routes/_pipeline.py
+++ b/src/gateway/api/routes/_pipeline.py
@@ -113,6 +113,7 @@ from gateway.models.entities import ModelPricing, UsageLog
 from gateway.models.guardrails import GuardrailConfig
 from gateway.models.mcp import McpServerConfig
 from gateway.models.money import to_usd
+from gateway.ports.model_provider_port import HostedAccessDeniedError, ModelProviderPort
 from gateway.rate_limit import RateLimitInfo, check_rate_limit
 from gateway.services.budget_service import (
     ZERO,
@@ -142,7 +143,7 @@ from gateway.services.pricing_service import (
     price_tool_calls,
     pricing_required_but_missing,
 )
-from gateway.services.provider_kwargs import ResolvedProvider, resolve_provider_selector
+from gateway.services.provider_kwargs import ResolvedProvider, credential_ladder_exhausted, resolve_provider_selector
 from gateway.services.routing import (
     BudgetState,
     CompiledPlan,
@@ -254,6 +255,11 @@ ORGANIZATION_GUARDRAILS_UNRESOLVABLE_DETAIL = "Organization guardrails could not
 ORGANIZATION_GUARDRAIL_CREDENTIAL_UNREADABLE_DETAIL = (
     "A configured organization guardrail's credential could not be read"
 )
+# The bound ``ModelProviderPort`` adapter answered with an upstream any-llm has
+# no implementation for, so there is nothing to dispatch against. Deliberately
+# says nothing about hosted inference or about which adapter answered: that is a
+# defect in this build, not something a caller can act on.
+HOSTED_CREDENTIAL_UNUSABLE_DETAIL = "No upstream provider is available to serve this model"
 SANDBOX_UNREACHABLE_DETAIL = (
     "code_execution sandbox unreachable. Check the sandbox URL in the dashboard's "
     "Tools settings, or OTARI_SANDBOX_URL, and that the container is running."
@@ -701,6 +707,7 @@ async def resolve_dispatch_provider(
     model_selector: str,
     *,
     adapter: FormatAdapter[Any, Any],
+    model_provider: ModelProviderPort,
 ) -> ResolvedProvider:
     """Get the ``ResolvedProvider`` for dispatch, reusing the one computed for
     the pricing/budget gate (``ctx.resolved_provider``) instead of resolving
@@ -712,11 +719,16 @@ async def resolve_dispatch_provider(
     check couldn't parse the selector (an unparseable selector has no
     pricing, but dispatch still needs its own resolution attempt so any-llm's
     own error surfaces instead of a stale gate-check failure).
+
+    Whichever of the two produced it, the result then passes through
+    :func:`_serve_from_hosted_credential`, which is where ``model_provider`` is
+    asked to serve a candidate no stored credential could. That is the last rung
+    and never displaces an earlier one; see that function.
     """
     if ctx.resolved_provider is not None:
-        return ctx.resolved_provider
+        return await _serve_from_hosted_credential(ctx, ctx.resolved_provider, adapter=adapter, port=model_provider)
     try:
-        return resolve_provider_selector(config, model_selector, ctx.user_id, workspace_id=ctx.workspace_id)
+        resolved = resolve_provider_selector(config, model_selector, ctx.user_id, workspace_id=ctx.workspace_id)
     except (ValueError, AnyLLMError) as exc:
         # The preamble deliberately tolerated this selector, so a reservation is
         # already held: refund it, then record the drop. The hold is not always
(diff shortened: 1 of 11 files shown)


src/gateway/api/routes/_pipeline.py
```diff
@@ -749,6 +761,153 @@ async def resolve_dispatch_provider(
...
+        # what those candidates are called with. Asking anyway would meter a
+        # resolve that never serves, so a routed chain keeps the ladder's own
+        # answer for now. Reaching the port from the compiler is separate work:
+        # the compiler is synchronous and documented as doing no I/O.
+        return resolved
+    if not credential_ladder_exhausted(resolved.provider, resolved.kwargs):
+        return resolved
+
+    try:
+        credential = await port.resolve_hosted_credential(
```
khaledosman, line 856: Only `HostedAccessDeniedError` is caught. An overlay adapter that fetches credentials over the network is the expected implementation, so an `httpx` timeout, a DB error, or any other exception is ordinary rather than exotic — and it propagates straight out of `resolve_dispatch_provider`, which all three routes call outside any `try` (`chat.py:415` and `:455`, `messages.py:617` and `:663`, `responses.py:524`).

There is no global refund: every site in this module releases explicitly, and `gateway.main` registers handlers only for `TenancyError` and `RequestValidationError`. So the reservation taken in the preamble leaks, which is the §0.3 invariant.

Wrap it the way the `LLMProvider(...)` `ValueError` branch below already does — log in full, `release_reservation`, then the non-leaky 502.

src/gateway/api/routes/_pipeline.py
```diff
@@ -749,6 +761,153 @@ async def resolve_dispatch_provider(
...
+            model=resolved.model,
+            provider=resolved.instance,
+            endpoint=adapter.endpoint,
+            detail=HOSTED_CREDENTIAL_UNUSABLE_DETAIL,
+            status_code=status.HTTP_502_BAD_GATEWAY,
+            started_at=ctx.started_at,
+        )
+        raise adapter.error(502, HOSTED_CREDENTIAL_UNUSABLE_DETAIL, ErrorKind.API) from exc
+
+    # Re-keying moves what settlement prices on. The pricing and budget gates
```
khaledosman, line 978: Taking this up since the description flags it for review.

The consequence is stated correctly, so this is only about the observability half: an overlay whose `response_provider` has no pricing row settles the request at `cost=NULL`, which is `require_pricing` bypassed rather than enforced, and nothing anywhere records that it happened. The comparison to a routing fallover is not quite even — `top_up_reservation_for_attempt` refuses an unpriced candidate with a 402, so that hazard has a guard and this one has neither a guard nor a signal.

I take the point that refusing here would make a build's own fleet unusable until it is priced. But a warning log when the re-keyed instance resolves no pricing costs nothing and turns a silent free-settlement into something an operator can find. Worth adding even if the refusal stays out.

khaledosman (changes requested):
Checks run against the head revision: `make lint`, `generate_openapi.py --check` (byte-identical, as claimed), and `test_hosted_credential_port.py` / `test_bad_model_name_returns_400.py` / `test_provider_instances.py` (88 passed).

The precedence work is right — `credential_ladder_exhausted` reusing the same env check as the keyless placeholder, and the multi-candidate skip matching the `len(attempts) > 1` condition dispatch actually branches on, so the port is never metered for a resolve that gets discarded.

Holding on the first finding: the port call has no guard for anything but `HostedAccessDeniedError`, and there is nothing above it that refunds.

src/gateway/api/routes/_pipeline.py
njbrake: Fixed in eb80b20c0. You are right on every step of it, including that nothing above would catch one: I checked the three call sites and `messages.py:526` is the only enclosing `try`, and it takes `HTTPException` alone.

Wrapped it the way you describe, matching the `LLMProvider(...)` branch: log with `exc_info`, `release_reservation`, rejection row, then the same non-leaky 502. `except Exception` rather than `BaseException`, so `asyncio.CancelledError` still propagates; a disconnected client is not a failed lookup and swallowing it would mark a cancelled task as handled.

I wrote the reproduction first: `test_an_adapter_failure_refunds_and_becomes_a_502` is parametrized over a `RuntimeError`, a `TimeoutError` and a `ValueError`, and all three failed before the guard. `test_cancellation_is_not_swallowed` pins the other half.

src/gateway/api/routes/_pipeline.py
njbrake: Added the warning in eb80b20c0. Your framing of the asymmetry is the part that convinced me: `top_up_reservation_for_attempt` gives the routing hazard a 402, so this one having neither a guard nor a signal was the real gap, not the missing refusal.

A re-key now looks up pricing for the upstream and, when there is none, warns naming both the candidate the caller asked for and the upstream that replaced it, so the line is greppable and distinguishable from the generic unpriced-model warning at settlement. Best-effort like the rejection log, since observability must not be able to fail a request that is otherwise about to succeed, and skipped entirely when the port answers `None`, so a plain build never pays for it. `test_an_unpriced_re_keyed_upstream_is_logged` covers it end to end.

Refusal stays out, as you said.

src/gateway/api/routes/_pipeline.py
```diff
@@ -749,6 +761,153 @@ async def resolve_dispatch_provider(
...
+        # the compiler is synchronous and documented as doing no I/O.
+        return resolved
+    if not credential_ladder_exhausted(resolved.provider, resolved.kwargs):
+        return resolved
+
+    try:
+        credential = await port.resolve_hosted_credential(
+            organization_id=ctx.organization_id,
+            workspace_id=ctx.workspace_id,
+            # The instance rather than the implementation: that is the name
```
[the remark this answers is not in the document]
njbrake: Agreed, and fixed in eb80b20c0. @khaledosman raised the same thing independently and it was the blocker on this PR.

Every failure but the port's own refusal now logs in full, releases the reservation, writes a rejection row, and returns a non-leaky 502. Cancellation semantics are preserved as you asked: the guard is `except Exception`, so `asyncio.CancelledError`, which derives from `BaseException`, still propagates. `test_cancellation_is_not_swallowed` pins that, and `test_an_adapter_failure_refunds_and_becomes_a_502` is parametrized over three failure shapes that all leaked before.

On the suppressed in-flight comment in your summary: `track_request` runs in the preamble, before this function, so a hosted refusal does briefly appear in `/v1/usage/in-flight`. That is true of the unresolvable-selector 400 that has always lived here too, so it is pre-existing rather than something this rung introduces, and moving registration is out of scope for this PR.

src/gateway/api/routes/_pipeline.py
```diff
@@ -749,6 +761,153 @@ async def resolve_dispatch_provider(
...
+    # here would make a build's own fleet unusable until it were priced. An
+    # overlay binding this port owns that.
+    kwargs: dict[str, Any] = {"api_key": credential.api_key}
+    if credential.api_base is not None:
+        kwargs["api_base"] = credential.api_base
+    return ResolvedProvider(
+        instance=credential.response_provider,
+        provider=upstream,
+        model=resolved.model,
+        kwargs=kwargs,
```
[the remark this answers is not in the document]
njbrake: Half taken, half pushed back on.

Taken: an unpriced upstream settling at `cost=NULL` is real, and eb80b20c0 adds a warning that names the candidate asked for and the upstream that replaced it. @khaledosman adjudicated the same thread and asked for the signal rather than the refusal, on the grounds that refusing would make a build's own fleet unusable until every upstream is priced. That is the call I followed. An overlay owes a pricing row for each `response_provider` it returns, and this makes the omission findable instead of silent.

Pushed back on the rest. Repricing and topping up here would mean a second budget gate keyed on a name the caller never wrote, and the substitution is the adapter's decision rather than a plan's, which is why the routed-chain guard does not transfer.

On overspend specifically: `reconcile_reservation` records actual spend and does not re-check the cap, by design. The budget is enforced fail-closed at reserve time and settlement records truth, so any request whose real token usage exceeds its estimate settles above what was held. That is the existing estimate-then-reconcile contract, not a hazard re-keying introduces, so I have not changed it here.

src/gateway/services/provider_kwargs.py
```diff
@@ -79,6 +106,46 @@ def keyless_placeholder_api_key(provider: LLMProvider, api_base: Any, api_key: A
...
+    backends ollama, llamacpp and llamafile, and Vertex AI, which authenticates
+    through the cloud SDK), and those declaring one any-llm does not insist on
+    (``_KEYLESS_SELF_HOSTED_PROVIDERS`` and ``_AMBIENT_CREDENTIAL_PROVIDERS``).
+
+    ``provider_credential_env_names`` returns ``None`` rather than ``()`` for a
+    provider it cannot inspect at all, and that stays exhausted: nothing is known
+    to serve the candidate, which is the same position an uncredentialed provider
+    with a declared variable is in.
+    """
+    if kwargs:
```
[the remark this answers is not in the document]
njbrake: Correct, and fixed in eb80b20c0. I verified the `client_args` case before changing anything: `providers.openai: {client_args: {timeout: 60}}` does resolve a non-empty dict with no credential in it, so the predicate reported the ladder as having answered when it had found a provider entry and no way to call it.

`credential_ladder_exhausted` now asks whether anything in kwargs could authenticate a call rather than whether the dict is non-empty, ignoring transport-only keys and falsy values. Deliberately narrow: `api_base` is not treated as non-credential, because an instance with a base URL and no key already takes the keyless placeholder and so never arrives alone. Ambient credentials are unaffected, since those providers are excluded by name a few lines above and never reach this check.

Covered by a parametrized test over `client_args` only, `api_key:` with no value, an empty key, and both together, plus the converse case pinning that transport tuning beside a real key still counts as credentialed.

src/gateway/api/routes/_pipeline.py
```diff
@@ -750,6 +762,242 @@ async def resolve_dispatch_provider(
...
+    one from a usage report. The settlement warning alone does not distinguish
+    it: an unpriced re-key looks exactly like an ordinary unpriced model there.
+
+    Best-effort, like the rejection log: observability must not be able to fail a
+    request that is otherwise about to succeed.
+    """
+    if ctx.db is None:
+        return
+    try:
+        pricing = await find_model_pricing(
```
khaledosman, line 791: This runs on every request the port re-keys, which for a build whose fleet serves its traffic is every request, and `find_model_pricing` is uncached: an organization-override query, the canonical `provider:model` key, then the legacy key on a miss, so 2 to 3 round-trips per request to decide whether to emit a log line. Settlement repeats the same lookup a moment later (`:2407`).

In the healthy case (fleet priced) that buys nothing at all, since the warning never fires. Memoize the `(upstream, model, organization_id)` combinations already warned about in a bounded set so it costs once per combination, or emit the warning from settlement where the lookup has already happened.

src/gateway/api/routes/_pipeline.py
```diff
@@ -750,6 +762,242 @@ async def resolve_dispatch_provider(
...
+            ctx.db,
+            upstream,
+            resolved.model,
+            organization_id=ctx.organization_id,
+        )
+    except SQLAlchemyError:
```
khaledosman, line 797: Swallowing this does not make the helper best-effort. A failed SELECT leaves the session's transaction aborted, so the settlement and log writes that follow on `ctx.db` raise `PendingRollbackError`, which is the docstring's promise inverted: observability failing a request that was otherwise about to succeed, and taking the hold with it.

`await ctx.db.rollback()` before returning. (`_held_handle` in `budget_service.py` already does exactly this for the same reason.)

khaledosman (approved):
Re-reviewed the delta since my last pass (`91f1652e` and `3bfb310d`; the rest of the range is the rebase). Both findings are addressed: the bare-exception refund is there with `Exception` rather than `BaseException`, parametrized over three failure shapes and backed by an integration test that asserts `users.reserved` itself, and the unpriced re-key now warns naming both the asked-for candidate and the upstream.

Also checked the Copilot-driven `credential_ladder_exhausted` change: `_kwargs_carry_a_credential` treating `api_base` as credentialing holds up, because an instance with a base URL and no key takes the keyless placeholder unless the provider's own env var is set, and that case is credentialed anyway.

Two non-blocking findings on the new pricing warning, inline.

PR #766, opened 2026-08-24
feat(tools): let a workspace pin its sandbox image and exposed tool set

### LINKED ISSUE #740: Workspace code-execution policy: sandbox image and exposed tool set
> **Re-scoped 2026-08-24** after @njbrake's note. #725 built the configuration plane; the gap is two columns narrower than this issue claimed. Original text at the bottom.

- Lands in: `otari`

## What is already there (#725)

Table `WorkspaceCodeExecutionPolicy`, `workspace_code_execution_policy_service.py`, org-role-gated routes, `WorkspaceCodeExecutionPolicyCard.tsx`, and enforcement tests. Its columns: `enabled`, `default_purpose_hint`, `max_iterations`, `exec_timeout_s`.

## The gap, stated as a column diff

The platform's `CodeExecutionConfig` (`backend/app/models/code_execution_config.py`) carries those four and two more that have no counterpart here:

- **`image`** the sandbox image, defaulted to `DEFAULT_SANDBOX_IMAGE`. This is the backend-selection half: a policy row today decides who may execute and for how long, never against what.
- **`tools`** which tool names the policy exposes, defaulted to `["code_execution"]`.

`SandboxBackend` is constructed at `_pipeline.py:2550` from deployment configuration alone, so neither has anywhere to be read from yet.

## What to produce

- Add `image` and `tools` to the policy, its upsert schema, its routes and its dashboard card.
- `SandboxBackend` selection reads the workspace's policy, falling back to deployment configuration.
- Guard what an operator may pin as an image; a workspace-settable image is a supply-chain surface, not just a string.

## Definition of Done

- A workspace can pin its sandbox image and its exposed tool set, honored on the next request.
- An unset policy behaves exactly as today.

<details>
<summary>Original text as filed</summary>

Asked for the table, service, routes and dashboard as if none existed, from `docs/architecture/rehome-inventory.md`, which predates #725.
</details>

njbrake:
## Description

A workspace could already say *whether* its people may run code and for how long, but not *what* that code runs in or which tools it gets. This adds those two. An operator now approves a list of sandbox images, and an owner of a workspace can pick one of them for their workspace, plus tick which code-execution tools the workspace is allowed to use. Picking an image the operator never approved is refused, and if the operator later withdraws an image, a workspace still pinned to it is told so rather than quietly being switched to a different one. Nothing changes for anyone who does not set these: a deployment that approves no images and a workspace with no policy behave exactly as before.

#725 built the configuration plane; the gap was two columns. The platform's `CodeExecutionConfig` carries `enabled`, `default_purpose_hint`, `max_iterations` and `exec_timeout_s`, which #725 has, plus `image` and `tools`, which it does not. `SandboxBackend` was built from deployment config alone, so neither had anywhere to be read from.

**`image`** is a nullable column, plus two deployment settings: `sandbox_image` (the deployment default, editable on the Tools page beside `sandbox_url`) and `sandbox_allowed_images` (the operator's comma-separated allow-list, deliberately config/env-only rather than dashboard-editable, matching how the SSRF gates are treated). Resolution is workspace pin, then deployment, then nothing. The guard runs twice: at the write, and again at admission, because an operator can shrink the list after a workspace pinned from it, and a stale pin refuses rather than falling back. The resolved image reaches the backend as an optional additive field on the session it leases, specified in the code-execution contract as a hint a fixed-pool backend may ignore. With no image resolved the create-session body stays `{}`, byte for byte as before.

**`tools`** is a nullable JSON column over the same vocabulary the hosted config uses. A stored list intersects with what the backend serves, so it only ever removes one; a list leaving nothing runnable refuses with a 403 rather than opening a backend advertising no tools, because a request that succeeds while silently running no code is the worse outcome. `SandboxBackend` advertises and dispatches the surviving set from one predicate, so the two cannot disagree.

On the fail-open note from #725's review: both reads sit inside the existing policy arm, behind the guard that already refuses when the session or workspace is missing, so the absent-context path still refuses rather than falling through.

Two things worth a reviewer's disagreement:

1. **`image` is weaker here than the issue implies.** On the platform it is explicitly a control-plane field (`sandbox.py`: "NOT enforced here: the warm pool serves a fixed pre-baked image") locked to one value. This gateway does not run containers either, so a session hint is the only honest wiring, and a backend that ignores it means the pin does nothing. The allow-list guard is the half that is real regardless.
2. **`tools` narrows to `code_execution` or to nothing today**, which is the objection #725's model docstring raised against the column ("a no-op or a second spelling of `enabled`"). I overrode it and rewrote that paragraph rather than leave it contradicting the column beside it, using the hosted three-name vocabulary so a policy written now keeps its meaning when a backend serves the other two, instead of being rejected as unknown today and silently ungated later. If you would rather honor #725's reasoning, dropping `tools` and shipping only `image` is a clean subtraction.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [x] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes #740

## Checklist

Run locally: `make lint`, mypy over 205 files, unit (2329 passed), integration (1674 passed, 9 skipped, against a local PostgreSQL 18 since this box has no Docker), the dashboard suite (1370 passed), `make openapi-check` and `make postman-check`, and the OSS-edition smoke gate under `--no-dev`.

Three hybrid tests failed on the first full integration run and are fixed here: `_FakeSandboxBackend` in `test_hybrid_mode_chat.py` declared a narrower `__init__` than the real class, so the new keyword arguments raised `TypeError` and surfaced as a 502 naming nothing. The double now mirrors the real signature, and a new test pins hybrid's arrangement, which was previously only a comment: `image` still comes from this gateway's config, and the `tools` the platform's resolve returns stay enforced by the `/v1/sandbox` proxy rather than a second time here.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

Implemented from the re-scoped issue by @njbrake working with Claude Opus 5. The design decisions above (the operator allow-list as the guard, refusing rather than falling back on a withdrawn image, and overriding #725's argument against a `tools` column) are his; the prose and the code are the model's.

## Summary

- Added workspace controls for sandbox images and exposed code-execution tools.
- Added operator-managed image allow-lists and deployment defaults.
- Validated image and tool selections when saved and when a session starts.
- Passed the resolved image to sandbox sessions.
- Updated dashboard controls, APIs, configuration, documentation, migrations, and tests.
- Preserved existing behavior when no workspace policy is configured.
- Added fail-closed handling for withdrawn images and unrunnable tool selections.


src/gateway/api/routes/_pipeline.py
```diff
@@ -1909,6 +1951,25 @@ async def prepare_gateway_tools(
                     sandbox_tool_entry["purpose_hint"] = workspace_policy.default_purpose_hint
                 sandbox_max_iterations = workspace_policy.max_iterations
                 sandbox_exec_timeout_s = workspace_policy.exec_timeout_s
+                if workspace_policy.tools is not None:
+                    # An intersection, so it only ever removes. Nothing left to run
+                    # is refused here rather than handed to a backend advertising an
+                    # empty tool list, which the model would answer by not calling
+                    # the tool at all: an unusable-but-successful request is the
+                    # failure mode a policy exists to make loud.
+                    if CODE_EXECUTION_TOOL_NAME not in workspace_policy.tools:
```
khaledosman: This tests membership of `CODE_EXECUTION_TOOL_NAME`, where the write-time guard tests `set(tools) & set(SERVED_TOOL_NAMES)` (`_require_runnable_tools`). The two agree only while `SERVED_TOOL_NAMES` has one entry.

The day the backend serves `bash_code_execution` and it joins that tuple, a policy of `["bash_code_execution"]` becomes storable — the write guard passes it — and every request then 403s here. That is precisely the storable-but-unusable state `_require_runnable_tools` was added to prevent, and it is the forward-compatibility case `CODE_EXECUTION_TOOL_NAMES` exists for.

Intersect against `SERVED_TOOL_NAMES` here as well, so the storable rule and the admission rule are one rule.

web/src/features/tools/WorkspaceCodeExecutionPolicyCard.tsx
```diff
@@ -222,6 +263,69 @@ export function WorkspaceCodeExecutionPolicyCard({
             description="Lowers how long one execution may run. It never raises it."
           />
 
+          {allowedImages.length > 0 ? (
```
khaledosman: The options are `allowedImages` plus the sentinel, so once an operator withdraws an image there is no `<option>` matching the stored pin. `FilterSelect` renders a controlled native `<select>` (`ui.tsx:810`), so a `value` with no matching option shows as the first option — "Deployment default" — while React state still holds the dead image. Save then sends it and earns the 400 from `_require_allowed_image`, with nothing on screen naming the value that caused it.

That is the withdrawn-image scenario this PR leads with, and the card currently does the opposite of telling the workspace its pin is dead. Include `policy.image` as an option when it is not in `allowedImages`, labeled as no longer approved.

alembic/versions/a7f3c9e2b481_add_code_execution_policy_image_and_tools.py
```diff
@@ -0,0 +1,42 @@
+"""Add the sandbox image and exposed tool set to the workspace code-execution policy.
+
+Two columns, both nullable, on ``workspace_code_execution_policies``:
+
+* ``image`` names the sandbox image the workspace's code runs in, chosen from
+  the allow-list an operator curated (``sandbox_allowed_images``). A workspace
```
khaledosman: `sandbox_allowed_images` is not the setting — it is `sandbox_allowed_session_images` everywhere else (config, `AGENTS.md`, docs, the service, the card). This is the only occurrence of the shorter spelling in the tree, and the `sandbox_image` / `sandbox_session_image` near-miss the config comment warns about is exactly why it is worth fixing rather than leaving.

khaledosman (approved):
Checks run against the head revision: `make lint`, `generate_openapi.py --check`, Postman regenerated from this PR's own spec (both artifacts current), `pnpm lint` + `typecheck`, `client:generate` and `build` leaving `schema.ts` and `routeTree.gen.ts` unchanged, and `test_code_execution_contract.py` / `test_sandbox_image_allowlist.py` / `test_sandbox_backend.py` / `test_tool_settings_endpoint.py` / `test_tool_settings_read_sites.py` (80 passed).

On the two things flagged for disagreement: the allow-list-as-the-real-guard framing holds up, and I checked that `sandbox_allowed_session_images` is genuinely not writable — it is in `ENV_BRIDGED_FIELDS` and the `/v1/settings` read grouping, but not in `UpdateToolSettingsRequest`, so config/env-only is enforced and not just documented.

Three findings inline.

src/gateway/api/routes/_pipeline.py
njbrake: Fixed in ac2f26c3c. You are right, and the failure mode is the one both guards exist to prevent: the write guard passes `["bash_code_execution"]` the moment that name joins `SERVED_TOOL_NAMES`, and admission then refuses every request against it.

Admission now intersects against `SERVED_TOOL_NAMES` too, so it is one rule rather than two that coincide while the tuple has one entry.

One thing your comment led me to that I want to flag, because it makes the constant load-bearing in a second way: `SERVED_TOOL_NAMES` is only a correct thing to intersect against while it names what a backend really offers. Growing it without teaching `SandboxBackend` the new kind would admit a policy naming only that kind, then hand the model a backend advertising nothing, which is the silently-successful request rather than the loud refusal. So `test_served_tool_names_is_what_the_backend_actually_advertises` now pins the tuple to what a `SandboxBackend` actually puts in `openai_tools`, and growing one without the other fails there.

web/src/features/tools/WorkspaceCodeExecutionPolicyCard.tsx
njbrake: Fixed in ac2f26c3c. Confirmed the mechanism you describe: `FilterSelect` is a controlled native `<select>` (`ui.tsx:810`), so with no `<option>` matching the stored value the browser shows the first one while React state still holds the withdrawn image, and the save sends it and earns the 400 from `_require_allowed_image` naming a value that was never on screen.

The stored image is now carried as an option of its own when it is not in `allowed_images`, labeled `<image> (no longer approved)`, above a warning saying the workspace's requests are being refused and that picking another or asking an operator to restore it are the ways out. The picker also renders now when the withdrawn pin is the *only* thing to show, which the previous `allowedImages.length > 0` gate would have skipped, leaving the operator no control at all on the deployment that had dropped its whole list.

Three tests: the withdrawn pin is displayed and named, it renders with an empty allow-list, and moving it back to Deployment default sends `image: null`.

alembic/versions/a7f3c9e2b481_add_code_execution_policy_image_and_tools.py
njbrake: Fixed in ac2f26c3c. You are right that it was the only occurrence left: the rename swept `src`, `tests`, `docs` and `web/src`, and `alembic` was simply not in the search path, so the one file nobody greps again is the one that kept the old spelling.

Re-swept the whole tree afterwards. The single remaining hit for the short form is the comment in `core/config.py` explaining why the field is deliberately *not* called `sandbox_image`, which is the near-miss you mention and wants to keep saying it.

alembic/versions/a7f3c9e2b481_add_code_execution_policy_image_and_tools.py
```diff
@@ -0,0 +1,42 @@
...
+  image is a supply-chain surface rather than a free string, which is why the
+  service refuses one that is not on that list.
```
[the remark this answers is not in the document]
njbrake: Correct, and fixed in ac2f26c3c. @khaledosman raised the same thing independently.

The migration docstring was the only place the pre-rename `sandbox_allowed_images` survived: the rename pass covered `src`, `tests`, `docs` and `web/src` and never looked in `alembic`. It now reads `sandbox_allowed_session_images`, matching the config field, the env var and the rest of the tree.

web/src/features/tools/WorkspaceCodeExecutionPolicyCard.tsx
```diff
@@ -149,6 +194,11 @@ export function WorkspaceCodeExecutionPolicyCard({
           default_purpose_hint: hint.trim() === "" ? null : hint.trim(),
           max_iterations: iterations.value,
           exec_timeout_s: timeout.value,
+          image: image === DEPLOYMENT_IMAGE ? null : image,
+          tools:
+            tools && tools.length > 0 && tools.length < availableTools.length
+              ? tools
+              : null,
```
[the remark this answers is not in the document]
njbrake: Valid, and fixed in 0b59cf2ae. I reproduced it before changing anything: with stored `["bash_code_execution"]` and served `["code_execution"]`, `tools.length < availableTools.length` is `1 < 1`, so the save sent `tools: null`. Admission refuses that policy (`_pipeline` intersects the stored list against `SERVED_TOOL_NAMES`), so dropping it turned a refusal into permission, from a save that meant to change the timeout. Agreed on the severity.

The root defect was asking a set question with a length comparison. A list narrows nothing when it covers everything served *and* names nothing else, which is now what `toolsForSave` tests, so a stale entry survives the save.

For the surfacing half I followed the shape @khaledosman's withdrawn-image finding one field over already set, since it is the same class of bug: stale state the server is actively refusing, silently normalized away by the card. Stale kinds are listed as their own checkboxes labeled `(no longer served)`, above a warning naming them and saying the workspace's requests are being refused. One thing that needed more than your comment asked for: the control previously rendered only when more than one kind was served, which on a single-tool deployment would have shown the warning with no checkbox to untick and no way out, so it now renders whenever there are stale entries.

Three tests. The regression one asserts an unrelated save (changing max_iterations) preserves the stale list; I checked it fails against the old length comparison and passes against the set-cover, so it pins the behavior rather than restating it.

src/gateway/api/routes/tool_settings.py
```diff
@@ -84,6 +84,7 @@ class UpdateToolSettingsRequest(BaseModel):
     web_search_intercept: bool | None = None
     sandbox_url: str | None = None
     sandbox_purpose_hint: str | None = None
+    sandbox_session_image: str | None = None
```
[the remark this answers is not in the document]
njbrake: Verified the mechanism and it is right: `UpdateToolSettingsRequest` takes Pydantic v2's default, so a PATCH naming `sandbox_allowed_session_images` parses to a model with every field `None`, builds an empty `updates`, and answers 200 having changed nothing.

Not taking `extra="forbid"` in this PR, for scope rather than disagreement. Six config fields are bridged into the environment but absent from `UpdateToolSettingsRequest`, so all six already answer a PATCH this way: `tools_header`, `mcp_allow_loopback`, `mcp_allow_private_hosts`, `provider_allow_private_hosts`, `web_search_allow_private_hosts`, and the one this PR adds. Forbidding extras is a behavior change to a published endpoint that affects all of them and any client sending a key it does not declare, which is a different concern from #740 and would ride in on a PR nobody would think to look in for it.

The narrower reading, that this PR made the hazard more likely by documenting a field as config-only that an operator might then try to PATCH, is fair. What it buys against is thin though: the field is refused today, quietly rather than loudly, and the failure is an operator wondering why their allow-list did not take rather than anything unsafe. @khaledosman checked this design explicitly in his review and endorsed it, noting the field is genuinely not writable rather than only documented as such.

Worth doing as its own change across all six fields, with the OpenAPI regeneration (`additionalProperties: false`) and a 4xx test that comes with it. Resolving here rather than leaving it open, since the decision is made rather than pending.

Issue #742, opened 2026-08-24
Give budget reservations a ledger, and add budget policies


khaledosman:
> **Re-scoped 2026-08-24** after @njbrake's note. The premise as filed was wrong: `otari` has a live reserve path. Original text at the bottom.

- Milestone: `otari-migration-m5`
- Labels: `otari-migration`, `open-core`
- Lands in: `otari`
- Gated by mozilla-ai/otari-ai#1782

## Correcting the premise

The filed version said a budget engine with no reserve path enforces ceilings only after the money is spent. Not true here. On `main`: `scoped_budget_service.reserve()/.settle()/.release()`, `budget_service.reserve_budget()`, `release_reservation()` in `_pipeline.py`, and a `reserved` column whose ceiling is `spend + reserved`, reconciled into `spend`. #724 fixed a leak in that path last week, so it is live code.

## The actual gap

**Holds are a counter, not a ledger.** There are no reservation rows, so:

- a leaked hold cannot be reclaimed *by identity*, only detected in aggregate;
- settle has nothing to be idempotent *against*, which matters most on the gateway path where a report can arrive twice.

**Budget policies** are the other genuinely absent half: `otari` has budgets, not a policy layer above them.

## What to produce

- A reservation ledger: one row per hold, with the identity that makes settle idempotent and leaked holds reclaimable. `reserved` stays the fast path; the rows are what make it auditable.
- Budget policies, with their dashboard surface.

Sequencing: mozilla-ai/otari-ai#1782 decides whether the platform's reserve-then-settle orchestration ports onto this or re-parents into the overlay. Take that decision against the state described above, not against the filed premise.

## Definition of Done

- Every hold is a row; settle is idempotent by reservation identity; a leaked hold is reclaimable individually.
- Two concurrent requests against a budget with room for one cannot both settle.

<details>
<summary>Original text as filed</summary>

Claimed `otari` had no reserve path at all, from `docs/architecture/rehome-inventory.md`, which predates the current state.
</details>

njbrake:
Flagging a factual problem with the premise before anyone picks this up.

"A budget engine with no reserve path enforces ceilings only after the money is spent" is not true of `otari` today. On `main`:

- `scoped_budget_service.reserve()`, `.settle()` and `.release()` (`src/gateway/services/scoped_budget_service.py`)
- `budget_service.reserve_budget()` (`src/gateway/services/budget_service.py`)
[...]

PR #767, opened 2026-08-24
feat(budgets): give reservations a ledger

njbrake:
## Description

Refs #742, the ledger half. Not "Closes": the issue's budget-policy half is already shipped, see below.

`users.reserved` and `scoped_budgets.reserved_spend` held in-flight budget as counters with no identity behind them, which is why two of the issue's own DoD lines were unsatisfiable. `budget_reservations` gives each hold a row, a status and a TTL; `budget_reservation_scopes` records what it placed on each scoped ceiling. The counters stay the fast path the gate reads.

Reconcile and refund claim the row before touching a counter, so the second settlement site to fire for one request is a no-op rather than a second refund. That mattered more than it looks: `release_reservation` is reachable from roughly seven sites in `_pipeline.py`, nothing but a `raise` after each kept two of them from firing, and because the release expression clamps at zero a double release passed silently as an under-count of live holds. Leaked holds are now reclaimable individually, by a per-user reclaim on the reserve path and by a lifespan sweeper.

Two invariants worth keeping: the row is written after the holds it records, never before, since a row with no hold would have the sweep return money nobody is holding; and a top-up grows the existing row rather than opening a second one with its own TTL.

**Budget policies are already here** and I did not rebuild them: `workspace_budget_default_service.py` (whose docstring names itself a port of the platform's `budget_policy_service`), `/v1/workspaces/{id}/member-budget-policies`, and the "Default member budget" control on `WorkspacesPage`. That half of #742 needs a re-scope, not code.

On sequencing: #742 is marked gated by mozilla-ai/otari-ai#1782. That decision is now proposed in mozilla-ai/otari-ai#1824 (delete `ReservationService.settle()`), and the ledger here is needed under either outcome, so it does not wait on it.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Refs #742 (the ledger half; the budget-policy half needs a re-scope rather than code).

## Checklist

- [x] If the API contract changed, I regenerated the OpenAPI spec (`uv run python scripts/generate_openapi.py`). No contract change here; `make openapi-check` and `make postman-check` are clean.

Also verified: migration round-trip (upgrade, downgrade, upgrade) on PostgreSQL 18 and SQLite, and `scripts/oss_edition_smoke.py`, which drives a real completion through the reserve path on SQLite.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

Written by Claude Opus 5 in back-and-forth with @njbrake; the reasoning and decisions are his. The change was then reviewed by a separate agent with no stake in it, which reproduced the load-bearing claims against a live PostgreSQL and SQLite rather than taking the description's word for them.

## Summary

Adds a budget reservation ledger for active holds, scoped allocations, expiration, and terminal states.

- Makes settlement and refund operations idempotent.
- Reclaims expired reservations per user and through a background sweeper.
- Extends existing reservations for top-ups.
- Preserves late settlement after reclamation.
- Retains terminal records and removes them after the configured retention period.
- Keeps existing counters as the fast path for budget checks.
- Adds configuration, migration, documentation, schema checks, and integration tests.


src/gateway/models/entities.py
```diff
@@ -844,9 +845,7 @@ class BatchRecord(Base):
     # this record is the strict ownership anchor, so it must always name an owner.
     # CASCADE: deleting the user drops the ownership record (the user's keys are
     # gone too, and usage_logs remain the billing history).
-    user_id: Mapped[str] = mapped_column(
-        ForeignKey("users.user_id", ondelete="CASCADE"), nullable=False, index=True
-    )
+    user_id: Mapped[str] = mapped_column(ForeignKey("users.user_id", ondelete="CASCADE"), nullable=False)
```
khaledosman: `index=True` was dropped from `BatchRecord.user_id` in this reformat. `ix_batches_user_id` is still created by `f3b6d8a1c4e7`, so this is autogenerate drift in one direction, and `BatchRecord` has no `__table_args__`, so nothing else indexes this CASCADE FK — user deletion and every `batches` read by owner lose their index.

Unrelated to the ledger; looks like the flag went missing while the identically-shaped `BudgetReservation.user_id` line was being written. Restore it.

src/gateway/models/entities.py
```diff
@@ -1100,6 +1099,101 @@ class ScopedBudget(Base):
...
+        # whole deployment's backlog of expired rows. Leading on ``user_id`` also
+        # serves the FK cascade, so this replaces the plain index on that column
+        # rather than joining it.
+        Index("ix_budget_reservations_user_status_expires", "user_id", "status", "expires_at"),
+    )
+
+    id: Mapped[str] = mapped_column(primary_key=True, default=lambda: str(uuid.uuid4()))
+    # Indexed for the opportunistic per-user reclaim on the reserve path, which
+    # asks for one user's expired holds on every request that reserves.
+    user_id: Mapped[str] = mapped_column(
```
khaledosman: This declares `index=True`, which renders an implicit `ix_budget_reservations_user_id`, but `c7a1e4d8f3b6` creates only the two composite indexes. So a migrated PostgreSQL and a `create_all` schema (which several unit tests build) disagree, and `alembic revision --autogenerate` would propose the missing index.

The comment at `:1139` already says the composite "replaces the plain index on that column rather than joining it", so the flag looks like the leftover rather than the intent — drop it.

src/gateway/services/budget_reservation_ledger.py
```diff
@@ -0,0 +1,461 @@
...
+    if older_than_sec <= 0:
+        return 0
+    cutoff = datetime.now(UTC) - timedelta(seconds=older_than_sec)
+    ids = (
+        (
+            await db.execute(
+                select(BudgetReservation.id)
+                .where(
+                    BudgetReservation.status != RESERVATION_ACTIVE,
+                    BudgetReservation.updated_at < cutoff,
```
khaledosman: `prune_terminal` filters `status != 'active' AND updated_at < cutoff`, which neither index supports: `ix_budget_reservations_status_expires_at` leads on an inequality here, and `updated_at` is not indexed at all.

The `.limit(batch_size)` bounds the result, not the scan, so the steady state is the worst case: with nothing yet old enough to prune, every tick (300s by default) scans the whole table to return zero rows, on a table that gains a row per billable request.

Either add `(status, updated_at)`, or key retention off `expires_at`, which is already the leading-adjacent column in both indexes.

src/gateway/services/budget_service.py
```diff
@@ -429,38 +519,44 @@ async def reserve_budget(
...
 
 async def reconcile_reservation(db: AsyncSession, handle: ReservationHandle, actual_cost: Decimal | float) -> None:
     """Settle a reservation: record actual spend and release the held estimate.
 
     Note: if this UPDATE/commit fails (e.g. a transient DB error after the
     provider call succeeded), the held estimate is not released and stays in
-    ``users.reserved``. That shrinks the user's effective budget until the next
-    budget reset zeroes it; a future enhancement could add a stale-reservation
-    sweep. This is the cost of fail-closed pre-debit and is rare in practice.
+    ``users.reserved``. The ledger row stays active with it, so the TTL sweep
```
khaledosman: This now says the wrong thing: `try_terminate` a few lines below commits the `SETTLED` transition *before* the spend/release UPDATE, so if that UPDATE fails the row is already terminal and no sweep will look at it again. The hold is then leaked permanently and the spend is lost, which is strictly worse than the pre-ledger behavior this paragraph is describing.

The module docstring does own the window ("a process that dies between the two"), but a transient DB error is a much likelier trigger than process death and the sweeper's `except Exception` swallows it. At minimum this docstring should stop promising the sweep will recover it.

khaledosman (changes requested):
Checks run against the head revision: `make lint` (architecture + ruff), `generate_openapi.py --check` (clean, so the no-contract-change claim holds), `make postman-check` equivalent, and `tests/unit/test_pipeline_vision_billing.py`.

Holding on the first two findings: `BatchRecord.user_id` loses its index in a reformat unrelated to this PR, and `BudgetReservation.user_id` declares an index the migration never creates. Nothing in the suite compares metadata against the migrated schema, so neither fails CI.

The ledger design itself reads well — writing the row after the holds, the CAS on the status transition, and the late-settlement path for a request that outlives its own TTL all hang together.

src/gateway/models/entities.py
njbrake: Fixed in 7eee7a066. You read the cause exactly right: an edit meant for `BudgetReservation.user_id` matched the identically-shaped line above it instead, so `batches` lost the flag and the ledger kept one it should not have had. Both are corrected.

Verified rather than eyeballed: `create_all` against a fresh PostgreSQL and the migrated schema now report the same index set for `batches` (`ix_batches_api_key_id`, `ix_batches_user_id`, `ix_batches_workspace_id`), so the autogenerate drift in both directions is gone.

src/gateway/models/entities.py
njbrake: Dropped in 7eee7a066. You are right that it was the leftover rather than the intent, and the comment at `:1139` was the giveaway.

The line now carries a short note saying why there is deliberately no `index=True` here, so the next person to add a composite does not put the flag back. Confirmed that `create_all` and a migrated schema now agree on `budget_reservations` (`ix_budget_reservations_status_expires_at`, `ix_budget_reservations_user_status_expires`, and nothing else).

src/gateway/services/budget_reservation_ledger.py
njbrake: Fixed in 7eee7a066, taking your second option. The query now asks for the terminal statuses by equality (`status IN (settled, released, expired)`) and ages rows by `expires_at`, so the whole predicate rides `ix_budget_reservations_status_expires_at` and no third index is needed.

Measured on PostgreSQL 18 with 200k rows and nothing eligible, which is the steady state you described: `Index Scan`, 6 buffer hits, 0.02ms. It was a sequential scan of the whole table before.

`updated_at` is the more literal answer to "when did this go terminal", and I dropped it for the reason you gave: it would need its own index to ask for. `expires_at` is creation plus the TTL, so against a retention window measured in days the two differ by minutes. That tradeoff is written down at the query.

src/gateway/services/budget_service.py
njbrake: You are right, and the docstring was the smaller half of it. Fixed properly in 7eee7a066 rather than reworded.

The claim now defers its commit and lands in the same transaction as the spend and release it authorizes, so a failure there rolls the row back to active and the sweep still finds it. `scoped_budget_service.release`/`settle` and the ledger claims grew a `commit=False` to make that possible, and the reclaim path is atomic for the same reason, which also closes the window the module docstring previously documented as open. The cost is a row lock on the reservation for the length of that transaction, which is two or three UPDATEs and never spans the provider call.

New test `test_a_failed_settlement_leaves_the_row_reclaimable` injects a failure into the user-row UPDATE that lands after the claim, then asserts the row is still `active` and that a subsequent sweep returns the hold on both counters. It fails against the previous code.

On your point that a transient error is likelier than process death and the sweeper swallows it: agreed, and that is what moved this from a docstring fix to a real one.

alembic/versions/c7a1e4d8f3b6_add_budget_reservation_ledger.py
```diff
@@ -0,0 +1,109 @@
...
+    # one and filters user_id, so one user's reclaim pays for the whole
+    # deployment's backlog of expired rows (measured at 200k rows: 1415 buffer
+    # hits per reserving request against a 100k-row backlog belonging to someone
+    # else). Leading on user_id also serves the FK cascade, so this replaces a
+    # plain index on that column rather than joining one.
+    op.create_index(
+        "ix_budget_reservations_user_status_expires",
+        "budget_reservations",
+        ["user_id", "status", "expires_at"],
+    )
```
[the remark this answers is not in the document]
njbrake: Both halves fixed in 7eee7a066, though not quite the way suggested.

`BatchRecord.user_id` gets its `index=True` back, as you say. For `BudgetReservation.user_id` I dropped the model flag rather than adding `ix_budget_reservations_user_id` to the migration: the composite `ix_budget_reservations_user_status_expires` already leads on `user_id`, so a plain index would be a second one covering a prefix of the first, paying write cost on every insert for nothing. Creating it would have made the two schemas agree on the wrong answer.

Verified both directions: `create_all` against a fresh PostgreSQL and a migrated schema now report identical index sets for `batches`, `budget_reservations` and `budget_reservation_scopes`.

src/gateway/services/budget_reservation_ledger.py
```diff
@@ -0,0 +1,461 @@
...
+                )
+                .limit(batch_size)
+            )
+        )
+        .scalars()
+        .all()
+    )
```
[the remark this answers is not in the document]
njbrake: Fixed in 7eee7a066, without the new index.

You are right that the retention query was unserved, and the diagnosis matches the one @khaledosman gave. I took the other remedy: the predicate now asks for the terminal statuses by equality and ages rows by `expires_at`, which makes it ride the existing `ix_budget_reservations_status_expires_at`. Measured with `EXPLAIN (ANALYZE, BUFFERS)` on PostgreSQL 18 at 200k rows with nothing eligible: index scan, 6 buffer hits, versus a full scan before.

A `(status, updated_at)` index would also have worked, and it is the more literal fit. It loses on cost: a third index on a table that takes an insert per billable request, to serve one query on a 300s timer, when an existing index can serve it for the price of measuring in `expires_at` instead. Over a retention window in days those differ by the TTL, which is minutes.

src/gateway/models/entities.py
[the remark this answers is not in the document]
njbrake: Correct, and fixed in 7eee7a066. It was collateral from an edit aimed at the identically-shaped `BudgetReservation.user_id` line; `ix_batches_user_id` is still created by `f3b6d8a1c4e7`, so the model and the migration had drifted apart. Restored, and verified that `create_all` and a migrated schema now agree on the `batches` indexes.

src/gateway/models/entities.py
```diff
@@ -1100,6 +1099,101 @@ class ScopedBudget(Base):
...
+        ForeignKey("users.user_id", ondelete="CASCADE"), nullable=False, index=True
+    )
```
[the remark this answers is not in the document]
njbrake: Correct, and fixed in 7eee7a066: the flag is dropped and the composite is the only index on that column. It also had the second problem you did not mention, which is that the migration never created the implicit index, so a `create_all` schema and a migrated one disagreed.

src/gateway/services/budget_reservation_ledger.py
```diff
@@ -0,0 +1,461 @@
...
+        return True
+    result = await db.execute(
+        update(BudgetReservation)
+        .where(
+            BudgetReservation.id == reservation_id,
+            BudgetReservation.status == RESERVATION_ACTIVE,
+        )
+        .values(estimate=BudgetReservation.estimate + user_delta)
+        .execution_options(synchronize_session=False)
+    )
```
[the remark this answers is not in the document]
njbrake: This one does not hold, and I checked it rather than reasoning about it, because it would have been a real bug if true.

SQLAlchemy applies a column-level `onupdate` to Core `update()` statements, not only to ORM flushes, so `grow()` and both claim functions do bump `updated_at` even though it is absent from their `values()`. Probe against the real model and engine, using the same statement shape:

```
before=2026-08-25 09:54:57.028802+00:00
after =2026-08-25 09:54:57.081755
onupdate fired on Core UPDATE: True
```

Separately, and for an unrelated reason, `prune_terminal` no longer keys off `updated_at` at all as of 7eee7a066: it ages rows by `expires_at` so the query can use an existing index. So the concern is moot on both counts. No code change from this thread.

src/gateway/services/budget_service.py
```diff
@@ -257,6 +271,51 @@ def estimate_cost(
...
+    return ReservationHandle(
+        user_id=user_id,
+        estimate=estimate,
+        reserved=user_reserved,
+        strategy=strategy,
+        counts_toward_budget=counts_toward_budget,
+        scoped_budgets=scoped,
+        scoped_estimate=scoped_estimate,
+        reservation_id=reservation_id,
+    )
```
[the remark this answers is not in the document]
njbrake: Valid, and fixed in 7eee7a066.

The ordering itself stays: the row is written after the holds on purpose, because a row with no hold would have the sweep return an amount nobody is holding, which is worse than the reverse. What was wrong is what happened when the write failed. Raising left the caller with no handle to reconcile or refund, which guaranteed the leak the ledger exists to prevent.

`_held_handle` now catches `SQLAlchemyError` around the ledger write, rolls back so the session stays usable for the caller own settlement writes, logs a warning, and returns the handle with `reservation_id=None`. The hold then settles through the handle exactly as every reservation did before this table existed. It is not reclaimable by the sweep in that case, which is the honest cost and is what the log line says.

src/gateway/services/budget_service.py
```diff
@@ -429,38 +519,44 @@ async def reserve_budget(
...
+    and leaves ``reserved`` untouched, so such a hold was never given back.
+
+    Idempotent by reservation identity: the first caller to claim the ledger row
+    does the work and any later one is a no-op, so two settlement sites firing for
+    one request cannot release the hold twice. A request that outlived its own TTL
+    is the one exception: the sweep has already returned its hold, so this records
+    the spend it still owes and releases nothing.
 
     This is the single authority for writing ``users.spend`` on the billable
     path — the usage-log writer no longer touches spend, so reconciliation must
```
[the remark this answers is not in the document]
njbrake: Fixed in 7eee7a066 for the line you pointed at, which is inside a docstring this PR rewrites.

Worth noting for accuracy: that em dash is not introduced here. It is on `origin/main` at `budget_service.py:466`, along with two more in `increase_reservation` that this PR does not touch. I left those two alone to keep the diff scoped to the change under review rather than sweeping the file.

src/gateway/models/entities.py
njbrake: Follow-up: the gap you named in the review body ("nothing in the suite compares metadata against the migrated schema, so neither fails CI") now has a guard, in 5d793bab1.

`tests/integration/test_schema_metadata_parity.py` builds the schema twice against the same server, once through the migration chain and once through `metadata.create_all`, and compares tables, columns with types and nullability, indexes, and unique constraints.

I checked it fails on both of the defects you found rather than just passing today:

```
budget_reservations: indexes disagree. Migrated only: []; models only: [ix_budget_reservations_user_id]
batches: indexes disagree. Migrated only: [ix_batches_user_id]; models only: []
```

It surfaced one pre-existing difference I did not fix: `APIKey.key_hash` is `unique=True, index=True`, which renders as one unique index, while `28d153c22616` emitted a `UniqueConstraint` alongside it, so a migrated database enforces that rule twice. Real drift, but dropping the redundant constraint wants its own migration against an auth-adjacent table rather than riding along here. It is named and excluded explicitly in the test, with a note to remove the exclusion with that migration.

src/gateway/services/budget_reservation_ledger.py
```diff
@@ -0,0 +1,495 @@
...
+            .where(BudgetReservationScope.reservation_id == reservation_id)
+            .values(amount=BudgetReservationScope.amount + scoped_delta)
+            .execution_options(synchronize_session=False)
+        )
+    await db.commit()
+    return True
+
+
+async def try_terminate(
+    db: AsyncSession, reservation_id: str | None, status: str, *, commit: bool = True
```
khaledosman: After this commit no caller passes `commit=True`: `budget_service.py:597`, `:605` and `:681` all pass `False`, `_reclaim` at `:310` does too, and no test calls either function. So the default is dead, and it is precisely the ordering this commit removed: the claim committing ahead of the release it authorizes, leaving a terminal row with a live hold that no later sweep revisits.

Default it to `False`, or drop the parameter, so a new call site cannot reintroduce that window by taking the default. Same for `try_settle_reclaimed` at `:236`.

src/gateway/services/budget_service.py
```diff
@@ -473,6 +588,34 @@ async def reconcile_reservation(db: AsyncSession, handle: ReservationHandle, act
...
+    if not await ledger.try_terminate(db, handle.reservation_id, ledger.RESERVATION_SETTLED, commit=False):
+        # Losing that claim has two causes and they settle differently. Another
+        # settlement site for this request already ran, and there is nothing left
+        # to do; or the TTL sweep reclaimed the hold while the request was still
+        # alive, in which case the hold is gone but the spend it went on to incur
+        # is still owed. Dropping it would leave ``users.spend`` permanently short
+        # of the sum of that user's rows, which is the counter a 403 is decided
+        # against.
+        if not await ledger.try_settle_reclaimed(db, handle.reservation_id, commit=False):
+        # Commit rather than roll back the empty transaction: the guarded UPDATE
```
khaledosman: This comment block is indented at the `if` level rather than inside the branch it describes. `ruff format` rewrites it (confirmed against this branch), but `make lint` runs `ruff check` only and `.pre-commit-config.yaml` has no `ruff-format` hook, so nothing in CI will catch it.

src/gateway/services/budget_service.py
```diff
@@ -523,16 +671,29 @@ async def record_external_spend(db: AsyncSession, user_id: str, cost: Decimal |
...
-    if not handle.reserved:
+    """Release a reservation without recording spend (e.g. provider failure).
+
+    Idempotent by reservation identity, for the same reason as
+    :func:`reconcile_reservation`: ``release_reservation`` in ``_pipeline`` is
+    reachable from roughly seven sites, and only control flow (a ``raise`` after
+    each) has kept two of them from firing for one request.
+    """
+    if not await ledger.try_terminate(db, handle.reservation_id, ledger.RESERVATION_RELEASED, commit=False):
+    # Commit rather than roll back the empty transaction: the guarded UPDATE
```
khaledosman: Same indentation slip as `:606`; these two are the only places `ruff format` would touch in this file.

khaledosman (approved):
Re-reviewed the delta since my last pass (`7eee7a06`). Everything I raised is addressed: `BatchRecord.user_id`'s index is back and `BudgetReservation.user_id`'s is gone, so `create_all` and `c7a1e4d8f3b6` agree; the claim and the release it authorizes are one transaction; retention keys off `(status IN (...), expires_at)` and rides `ix_budget_reservations_status_expires_at`.

Also checked, since the transactions got longer: lock order is reservations, then `scoped_budgets`, then `users` at every settlement site, and the reserve path commits per step, so no deadlock cycle is introduced. Two non-blocking findings inline.

src/gateway/services/budget_reservation_ledger.py
njbrake: Good catch, and I took the second option: the parameter is gone from both `try_terminate` and `try_settle_reclaimed` rather than defaulted to `False`. Fixed in b27869dd7.

A default nobody uses is still a way to reintroduce the window by accident, and there was no call site that wanted the old behavior, so removing it means a new caller cannot opt back into the bad ordering at all. Both docstrings now say they never commit and why.

`scoped_budget_service.release` keeps its switch, since two compensation paths (`budget_service.py:533` and `:783`) still own their own commit and legitimately take the default.

src/gateway/services/budget_service.py
njbrake: Fixed in b27869dd7, both this one and `:682`. The comment blocks now sit inside the branch they describe.

You are right about the gap that let it through: `make lint` runs `ruff check` only and there is no `ruff-format` hook, so `ruff format --check` on this file is now clean but nothing in CI asserts that.

Worth knowing before anyone wires one up: `ruff format` currently wants to rewrite four files this PR touches (`_pipeline.py`, `core/config.py`, `models/entities.py`, `scoped_budget_service.py`), and all four drift identically on `origin/main`, so it is pre-existing rather than anything this branch introduced. I left them alone, since AGENTS.md asks for formatter changes in their own PR. I did format the one file this PR adds.

src/gateway/services/budget_service.py
njbrake: Fixed in b27869dd7 together with `:606`; see that thread for the detail on why CI did not catch either.

PR #769, opened 2026-08-25
fix(gateway): Address Gateway 502s on max_completion_tokens (OpenAI's current field name)

aittalam:
## Description

Addresses https://github.com/mozilla-ai/otari-ai/issues/1062

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues
Closes https://github.com/mozilla-ai/otari-ai/issues/1062

## Checklist

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**
Claude Code Opus 5

## Summary

- Support both OpenAI output-limit fields.
- Prefer `max_completion_tokens` when both fields are provided.
- Document this precedence in the API and client schemas.
- Return clear `400` errors for provider-rejected parameters.
- Scope prompt-cache keys to the authenticated user and workspace.
- Pass workspace context to routing and message-normalization logic.
- Apply configured TTLs to budget reservations.
- Add unit and integration tests for these behaviors.

These changes improve OpenAI compatibility, protect cache isolation, and make provider errors easier to understand.

## Technical notes

- Dispatch the effective output cap as `max_tokens` for any-llm compatibility.
- Log a warning when both fields have different values.
- Apply the same behavior to streaming requests.
- Limit rejected-parameter detection to fields supported by the gateway.
- Preserve safe, actionable provider error details without leaking internal implementation names.


src/gateway/api/routes/_pipeline.py
```diff
@@ -308,6 +309,37 @@ def _is_unsupported_feature_error(exc: BaseException) -> bool:
...
 
+_UNEXPECTED_KWARG = re.compile(r"unexpected keyword argument '([^']+)'")
+
+
+def provider_rejected_param_detail(param: str) -> str:
+    """Detail for a request param the provider serving this model cannot express."""
+    return f"The provider serving this model does not accept the '{param}' parameter"
+
+
+def _rejected_param(exc: BaseException) -> str | None:
```
khaledosman, line 393: `_rejected_param` matches *any* `TypeError: … unexpected keyword argument 'x'` anywhere in the chain, and `classify_provider_error` is reached from broad `except Exception` handlers (`_pipeline.py:3249`, `:3690`) that wrap gateway-internal code, not only the `acompletion` call.

So a signature drift after an any-llm bump, or a test double with a narrower `__init__` — exactly the `_FakeSandboxBackend` bug #766's description reports — now surfaces as `400 "The provider serving this model does not accept the 'image' parameter"`, and `failure_status_code` records 400. A gateway defect gets attributed to the caller and stops reading as an outage on the error-rate panel.

Gate the match on the name being something the gateway actually forwarded (`CompletionParams.model_fields`, say) and fall through to the generic 502 otherwise.

src/gateway/api/routes/chat.py
```diff
@@ -283,6 +293,38 @@ def prepare_platform_call_kwargs(self, kwargs: dict[str, Any]) -> dict[str, Any]
...
+
+def _effective_output_cap(max_tokens: int | None, max_completion_tokens: int | None) -> int | None:
+    """The single output cap to dispatch, given OpenAI's current and legacy fields.
+
+    OpenAI renamed the cap to ``max_completion_tokens`` and deprecated
+    ``max_tokens``, so an OpenAI-compatible client sends the former
+    (this is what any-llm's own SDK does).
+
+    any-llm's OpenAI layer remaps ``max_tokens`` to ``max_completion_tokens`` on
+    the way out, so nothing is lost by folding everything into ``max_tokens`` and a
```
khaledosman: Trailing whitespace at the end of this line. Ruff's `W291` is not enabled here, so `make lint` will not catch it.

`:315` in the same docstring also reads "while just ``max_completion_tokens`` value is used, everything is logged", which does not parse.

khaledosman (approved):
Checks run against the head revision: `make lint` (architecture + ruff), `generate_openapi.py --check`, and the Postman collection regenerated from this PR's own spec — no drift, so the `openapi-spec` job is satisfied by the one committed spec change.

Two findings inline. The first is the one worth acting on: the `TypeError` probe is wider than the failure it was written for.

dpoulopoulos:
@aittalam, I have a question. Please let me know if I'm understanding this correctly:

The issue states:

> OpenAI deprecated `max_tokens` in favor of `max_completion_tokens`, so OpenAI-compatible clients (including the any-llm SDK) send `max_completion_tokens` by default, which breaks all of them against otari.

The solution folds the two limit fields into a single `max_tokens` parameter before dispatch, with `max_completion_tokens` taking precedence and a warning logged if the two disagree.

So, why are we folding the two parameters into the one that has been deprecated? Is it just to make it work with Otari? And if so, shouldn't we update Otari instead?

src/gateway/api/routes/_pipeline.py
```diff
@@ -365,6 +397,13 @@ def classify_provider_error(exc: BaseException) -> ProviderErrorMapping | None:
     # names the unsupported feature.
     if _is_unsupported_feature_error(exc):
         return ProviderErrorMapping(status.HTTP_400_BAD_REQUEST, _caller_fault_detail(exc, PROVIDER_BAD_REQUEST_DETAIL))
+    # The same shape one layer down: a param any-llm forwards that the resolved
+    # provider's SDK has no parameter for. Also permanent, also the caller's to
+    # fix (by dropping the param or routing to a provider that has it), and the
+    # response names it, since a caller cannot otherwise tell which param the
+    # provider objected to.
+    if (param := _rejected_param(exc)) is not None:
```
dpoulopoulos, line 490: **[major]** `_rejected_param` matches *any* `TypeError` carrying Python's "unexpected keyword argument" wording anywhere in the `original_exception` chain, without checking that the name is something the caller could have sent. `classify_provider_error` is reached from broad `except Exception` handlers that wrap far more than the `acompletion` call (`_pipeline.py:3249` around `open_stream`, and the standalone non-stream handler around `dispatch_non_stream`/the tool loop), and `acompletion` itself forwards operator-owned `client_args` into `AnyLLM.create(**client_args)` (`any_llm/api.py:105`; the gateway populates it from `config.yml` / `ProviderCredential.client_args` via `services/provider_kwargs.py:128` and `bedrock_gateway_auth.py`).

Running the classifier against the pinned SDK confirms both cases:
- `TypeError("AsyncOpenAI.__init__() got an unexpected keyword argument 'timeoutt'")` → `400 "The provider serving this model does not accept the 'timeoutt' parameter"`, and `failure_status_code` records `400`.
- `TypeError("_FakeSandboxBackend.__init__() got an unexpected keyword argument 'image'")` → the same 400.

So an operator `client_args` typo, or gateway-internal signature drift after an any-llm bump, is returned to the caller as their fault, names an internal parameter they cannot remove (the internal detail the repo's error-detail boundary exists to keep out of public responses), and drops off the error-rate panel as a 4xx instead of an upstream failure.

Gate the match on the name being a param the gateway actually forwards (e.g. `param in CompletionParams.model_fields`) and fall through to the generic 502 otherwise.

src/gateway/api/routes/_pipeline.py
```diff
@@ -365,6 +397,13 @@ def classify_provider_error(exc: BaseException) -> ProviderErrorMapping | None:
...
+        return ProviderErrorMapping(status.HTTP_400_BAD_REQUEST, provider_rejected_param_detail(param))
```
dpoulopoulos: **[minor]** This branch also changes `/v1/messages` and `/v1/responses`, which share `classify_provider_error` (`messages.py:359`, `responses.py:263`), but the only endpoint-level coverage is the chat one (`tests/integration/test_output_cap_fold.py:151`). Worth one assertion on a non-chat route so the mapping is pinned where it is not obvious from the chat tests.

src/gateway/api/routes/chat.py
```diff
@@ -283,6 +293,38 @@ def prepare_platform_call_kwargs(self, kwargs: dict[str, Any]) -> dict[str, Any]
...
+    reasoning model still receives the field it requires. On the other hand,
+    folding the other way would break every provider that never learned the new
+    name.
+
+    The current name wins when a request carries both, matching OpenAI's own
+    deprecation and the precedence any-llm's OpenAI layer already applies. The
+    result is what the budget estimate reserves against as well, so the cost
+    reserved and the cap dispatched can never disagree about which field won.
+    Two *different* caps in one request is the caller contradicting itself, and
+    while just ``max_completion_tokens`` value is used, everything is logged.
```
dpoulopoulos: **[minor]** "Two *different* caps in one request is the caller contradicting itself, and while just ``max_completion_tokens`` value is used, everything is logged." does not parse. This is the docstring a maintainer reads to understand the precedence rule; Copilot flagged the same sentence.

src/gateway/api/routes/chat.py
dpoulopoulos: **[nit]** Trailing whitespace at end of line. Ruff's `W` rules are not in `select = ["E", "F", "I", "PERF"]` (`pyproject.toml:142`), so `make lint` will not catch it.

tests/unit/test_chat_output_cap.py
```diff
@@ -0,0 +1,100 @@
...
+def test_neither_leaves_the_cap_unset() -> None:
+    """No cap stays no cap: the provider applies its own default, and the budget
+    estimate falls back to its configured default output tokens."""
+    assert _effective_output_cap(None, None) is None
+
+
+def test_explicit_null_current_name_does_not_erase_the_legacy_one() -> None:
+    """A client that sends ``"max_completion_tokens": null`` alongside a real
+    ``max_tokens`` asked for the latter, so the null must not win the fold."""
+    assert _effective_output_cap(20, None) == 20
```
dpoulopoulos: **[nit]** `test_explicit_null_current_name_does_not_erase_the_legacy_one` asserts `_effective_output_cap(20, None) == 20`, byte-for-byte the same call as `test_legacy_name_alone_is_the_cap` at `:35`. At this layer an explicit JSON `null` and an unset field are indistinguishable, so the test cannot cover what its name promises. The real distinction is covered by `tests/integration/test_output_cap_fold.py:106`; either delete this one or point its docstring at where the behavior actually lives.

src/gateway/api/routes/_pipeline.py
```diff
@@ -308,6 +309,37 @@ def _is_unsupported_feature_error(exc: BaseException) -> bool:
     return any(isinstance(candidate, NotImplementedError) for candidate in upstream_exception_chain(exc))
 
 
+_UNEXPECTED_KWARG = re.compile(r"unexpected keyword argument '([^']+)'")
+
+
+def provider_rejected_param_detail(param: str) -> str:
```
dpoulopoulos: **[nit]** `provider_rejected_param_detail` is public but has no consumer outside this module (its siblings `_caller_fault_detail` and `_is_unsupported_feature_error` are private, and no test imports it). Making it private keeps the module's exported surface honest.

dpoulopoulos (changes requested):
Fixes gateway 502s when an OpenAI-compatible client sends `max_completion_tokens` (OpenAI's current name for the output cap): `/v1/chat/completions` now folds the two cap fields into a single `max_tokens` before dispatch, with `max_completion_tokens` winning and a warning logged when the two disagree, and the same folded value feeds the budget reservation. It also adds a `classify_provider_error` branch that maps a provider SDK's `TypeError: ... unexpected keyword argument 'x'` to a 400 naming `x` instead of a generic 502, plus unit/integration coverage and the regenerated OpenAPI/dashboard schema.

The fold is correct and well covered. The new `TypeError` branch is the one thing to change before merge: it matches on wording alone, so operator- and gateway-side faults now come back as caller 400s naming an internal parameter.

Not verified in a browser: no user-facing UI change (the only `web/` edit is a regenerated JSDoc `@description` in `web/src/client/schema.ts`, referenced nowhere in `web/src`). Verified instead by reading the changed files and their callers, checking the fold against the pinned `any_llm` package in this checkout, and running `classify_provider_error` directly against the misclassification cases. CI is green on all 15 checks, including `openapi-spec` (so the Postman collection is not stale).

aittalam:
> @aittalam, I have a question. Please let me know if I'm understanding this correctly:
> 
> 
> 
> The issue states:
> 
> 
> 
> > OpenAI deprecated `max_tokens` in favor of `max_completion_tokens`, so OpenAI-compatible clients (including the any-llm SDK) send `max_completion_tokens` by default, which breaks all of them against otari.
> 
> 
> 
> The solution folds the two limit fields into a single `max_tokens` parameter before dispatch, with `max_completion_tokens` taking precedence and a warning logged if the two disagree.
> 
> 
> 
> So, why are we folding the two parameters into the one that has been deprecated? Is it just to make it work with Otari? And if so, shouldn't we update Otari instead?

I had the same question! any-llm takes max_tokens and translates it properly to any other model, while it does not do the same with max_completion_tokens (as the error shows). I asked Claude to double check and it confirmed that what gets to the model is the actual correct parameter.

I added a test in 3e0c15dc15395adbe444b742613211201c0b3872 so this is explicitly verified by the code itself.

src/gateway/api/routes/chat.py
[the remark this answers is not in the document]
aittalam: Addressed in c4bfec3c41c9f20f7e60c45fdbdfe3c8bb18b78c

src/gateway/api/routes/chat.py
aittalam: Addressed in f8ec530b0f131dfdc300243ec3172ae5344f741e

src/gateway/api/routes/_pipeline.py
aittalam: Addressed in 57cff3622d377af4da8b18c8223d3428f805c4ec

src/gateway/api/routes/_pipeline.py
aittalam: Addressed in 57cff3622d377af4da8b18c8223d3428f805c4ec

src/gateway/api/routes/_pipeline.py
```diff
@@ -291,6 +292,37 @@ def _is_unsupported_feature_error(exc: BaseException) -> bool:
...
+    502 and report a permanent, caller-fixable mismatch between a model and a
+    param as an upstream outage. Only the param name is returned: the SDK's
+    method name is an internal detail and never reaches the client (see
+    ``test_error_detail_leakage``).
+    """
+    for candidate in upstream_exception_chain(exc):
+        if isinstance(candidate, TypeError):
+            match = _UNEXPECTED_KWARG.search(str(candidate))
+            if match:
+                return match.group(1)
```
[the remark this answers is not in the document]
aittalam: Addressed in 57cff3622d377af4da8b18c8223d3428f805c4ec

src/gateway/api/routes/_pipeline.py
aittalam: Good point! Addressed in d088912d59b0c1363e0e8a5752c85709f26c24a4

src/gateway/api/routes/chat.py
aittalam: Agree - addressed in c4bfec3c41c9f20f7e60c45fdbdfe3c8bb18b78c

src/gateway/api/routes/chat.py
aittalam: Addressed in f8ec530b0f131dfdc300243ec3172ae5344f741e

tests/unit/test_chat_output_cap.py
aittalam: Addressed in d28642cc8f65ca593cb65d38c7dd69298a4f8fec

src/gateway/api/routes/_pipeline.py
aittalam: Addressed in e8d6d7b14cc7161e5d40cfee7530e4cd7b3dd736

aittalam:
@copilot resolve the merge conflicts in this pull request

aittalam:
@copilot resolve the merge conflicts in this pull request

dpoulopoulos (approved):

PR #772, opened 2026-08-25
fix(usage): reserve the otari-ai: source prefix in external-usage ingest

khaledosman:
## Description

The external-usage `source` slug pattern allows a colon, so an importer could post rows tagged `otari-ai:gateway` or `otari-ai:claude_code`, the same tags otari.ai's backfill writes. Those lookalike rows inflate the migration's reconciliation totals and show up as a false MISMATCH during a cutover.

Any source starting with `otari-ai:` (case-insensitive) is now rejected with the same reserved-source 422 the bare `gateway` slug already gets. The exact-match `gateway` check is unchanged, and this is a prefix check rather than a colon ban, so `foo:bar` is still accepted. Validation on new imports only: no migration, no change to stored rows.

The OTLP route is the second door into the same rows: it derives a `source` from the caller-supplied `otari.client_name` and hands it to the batch schema, so the new reservation would have turned such an export into a 500. The reservation now lives in one `reserved_source_reason` helper that both doors read: a 422 for a direct import, and a fallback to the default `otel` source for an exporter, matching how `gateway` is already handled there.

Follow-up (not here): otari-ai#1815 reserves the same prefix in `usage_admin_service.py` and `workspace_activation_service.py`. Once both land, the three copies of the prefix could converge on one shared definition.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes mozilla-ai/otari-ai#1801

## Checklist

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

Implementation, tests, and the local check runs (`make lint`, `make typecheck`, the external-usage integration file, `tests/unit`, `make openapi-check`, `make postman-check`) were done by the agent.

Authored by Claude Opus 5 (1M context) via Claude Code.

### DIFF
--- a/docs/external-usage.md
+++ b/docs/external-usage.md
@@ -103,7 +103,7 @@ The body is a batch that shares a `source` and a default `user_id`:
 
 | Field | Required | Notes |
 | --- | --- | --- |
-| `source` | yes | Provenance slug, e.g. `claude_code`. Generic: add your own sources. |
+| `source` | yes | Provenance slug, e.g. `claude_code`. Generic: add your own sources. `gateway` and any slug starting with `otari-ai:` are reserved (otari.ai writes those tags itself) and rejected with a 422. |
 | `user_id` | with master key | Default attribution. Optional with an API key (binds to the key's user); required with the master key. Must be an existing user. |
 | `events` | yes | 1 to 1000 events. |
 
--- a/docs/external-usage.md
+++ b/docs/external-usage.md
@@ -182,7 +182,7 @@ Otari reads only the content-free usage attributes, preferring the
 | `gen_ai.usage.output_tokens` | output |
 | `gen_ai.usage.cache_read_tokens` (or `cached_tokens`) | cache read |
 | `gen_ai.usage.cache_write_tokens` | cache write |
-| `otari.client_name` | `source` (provenance) |
+| `otari.client_name` | `source` (provenance); a reserved slug (`gateway`, or anything starting with `otari-ai:`) falls back to `otel` |
 | `otari.user_session_label` / `otari.session_label` | session label |
 
 A record with no model/provider/tokens (a non-LLM span, a prompt log, a metric) is
--- a/src/gateway/api/routes/otlp.py
+++ b/src/gateway/api/routes/otlp.py
@@ -75,10 +75,10 @@ from gateway.services.agent_telemetry_service import (
 from gateway.services.agent_telemetry_service import ingest as ingest_telemetry
 from gateway.services.external_usage_service import (
     MAX_EVENTS_PER_BATCH,
-    RESERVED_SOURCES,
     ExternalEventsRequest,
     ExternalUsageEvent,
     ingest_external_events,
+    reserved_source_reason,
 )
 
 router = APIRouter(tags=["otel"])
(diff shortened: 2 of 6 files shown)


njbrake (approved):
Approving. Reserving `otari-ai:` alongside `gateway` by extending the existing validator, rather than adding a second check beside it, is the right shape, and the OTLP door is handled the same way `gateway` already is.

What I verified rather than took on faith:

- The backfill is not locked out of its own tag. otari-ai#1798 writes directly to the target database via `--target-url`, not through `POST /v1/usage/external-events`, so this validator is not in its path.
- The boundary matches the consumer. `settlement_target.read_target_totals` scopes with `source LIKE 'otari-ai:%'`, colon included, so reserving the prefix with the colon and leaving bare `otari-ai` free is the correct cut.
- The 500 the second commit prevents is real: `ExternalEventsRequest(...)` in `otlp.py` is constructed with no `ValidationError` handling.
- Both new tests fail when `RESERVED_SOURCE_PREFIXES` is emptied and pass when it is restored. The two touched integration files are green locally against PostgreSQL, along with lint, typecheck, `openapi-check`, and `postman-check`.

Non-blocking nits:

1. `external_usage_service.py:73` hardcodes `'gateway'` in the exact-match message, so a second entry in `RESERVED_SOURCES` would get the wrong text. An f-string on the matched value produces today's string and stays correct later.
2. Prefix matching depends on the tuple entries being lowercase. `prefix.lower()` or a one-line comment would say so.
3. `reserved_source_reason` is pure logic with no unit test; a table-driven case would pin the case handling and the bare `otari-ai` boundary cheaply.

The red `e2e` job is unrelated: it is the budget-assignment combobox timing out in `dashboard.spec.ts`, this PR touches no `web/` code, and the e2e seeds import under `parity-e2e` and `e2e-seed`. Worth a re-run.

Ordering note for afterwards: otari-ai#1815 is still open, so `usage_admin_service._selection_conditions` still reads a backfilled row as ordinary imported usage.

khaledosman:
All three nits applied in bb4edae.

1. The exact-match message is now an f-string on the matched value, so a second `RESERVED_SOURCES` entry gets its own text; the string is byte-identical today, pinned by an assertion in the new unit test.
2. The prefix comparison lowercases the tuple entry (`lowered.startswith(prefix.lower())`), so the match no longer rests on how the constant is written. Comment not added, since the code now says it.
3. `tests/unit/test_reserved_source_slugs.py`: table-driven over reserved spellings (both cases of `gateway` and `otari-ai:`) and free ones, pinning the colon boundary you verified against `read_target_totals`, so bare `otari-ai` and `otari-ai-mirror` stay legal.

Green locally: `make lint`, `make typecheck`, the new unit file (17 passed), and both touched integration files (43 passed). The `e2e` job stays red until #779 merges, which fixes that flake at its root: the spec seeded alice after `login(page)`, racing the app's own `GET /v1/users`, and `useUsers` caches with `staleTime: 60_000`, so a list fetched before the seed stayed fresh through the page mount. Reproduced on `main` with none of this PR applied, so no re-run here.

(via Claude Opus 5 (1M context) in Claude Code)

PR #773, opened 2026-08-25
fix(usage): read a backfilled hosted row as usage this deployment served

khaledosman:
## Description

The hosted-history backfill tags a migrated row with its origin behind a legacy prefix, so traffic the hosted deployment served itself arrives as `otari-ai:gateway`. Two places compared `source` against the bare literal `"gateway"` and read such a row as the wrong kind of row, in opposite directions:

- `usage_admin_service._selection_conditions` let delete and reprice reach it. Budget exemption is not an imported-only flag, which is why the source guard is the load-bearing one.
- `workspace_activation_service` read a workspace whose whole history was migrated as never having called this gateway, so the guide asked a long-standing customer for their first request.

Both now ask one question, in the new `gateway.core.usage_source`: the bare `gateway` slug, or that same slug behind the legacy prefix. Deliberately **not** a blanket `otari-ai:%` match: `otari-ai:claude_code` is imported usage that happens to have been migrated, and repricing imported usage is exactly what the admin surface is for, so a wildcard would quietly shield every migrated import. Tests cover both directions.

Ingest-side reservation of the prefix (`external_usage_service.py`, `otlp.py`) is mozilla-ai/otari-ai#1801, deliberately untouched here; converging the two definitions is a follow-up.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes mozilla-ai/otari-ai#1815

## Checklist

## AI Usage

- [ ] No AI was used.

**Any additional AI details you'd like to share:** Implemented, tested, and self-reviewed by the agent; `make lint`, `make typecheck`, and the unit plus integration suites were run locally.

🤖 Authored by Claude Opus 5 (1M context) via [Claude Code](https://claude.com/claude-code)

## Summary

- Recognize both `gateway` and `otari-ai:gateway` as usage served by the current deployment.
- Preserve migrated imported usage, such as `otari-ai:claude_code`, as imported usage.
- Prevent hosted usage from being deleted or repriced.
- Ensure migrated gateway usage activates workspaces correctly.
- Add shared source handling and integration test coverage.
- Update related documentation.


njbrake (approved):
Approving. The narrow match is the right call: reading the slug behind the legacy prefix, rather than the prefix itself, is what keeps `otari-ai:claude_code` repriceable while a migrated hosted row is not, and the tests prove that direction as well as the obvious one. `core/usage_source.py` follows `core/sql.py`, so the shared predicate lands where the codebase already puts conditions two layers have to agree on.

I confirmed the new tests are real regression tests by narrowing `SERVED_HERE_SOURCES` back to `("gateway",)`; both fail. Also checked that these were the only two `source`-vs-`"gateway"` comparisons in the backend, and that the `source: "gateway"` activity link in `PricingWarning.tsx` is scoped to the last hour, where a backfilled row cannot appear.

One thing I pushed onto the branch (5ca51a0): the ingest was still reserving only the bare `gateway` literal, and `_SLUG_PATTERN` allows a colon, so `POST /v1/usage/external-events` accepted `source="otari-ai:gateway"`. Before this PR that bought a caller nothing. After it, those rows activate a workspace that never called the gateway and drop out of reach of delete and set-price for good. Rather than leave that window open until otari-ai#1801, both the events endpoint and the OTLP receiver now read the slug through `slug_behind_legacy_prefix` before checking `RESERVED_SOURCES`, which reserves the served-here spellings and nothing more. Two tests, one per direction. If #1801 intends a blanket prefix reservation, or the backfill writes through this endpoint, drop that commit.

Two notes, neither blocking:

- The admin half of the fix is load-bearing only if backfilled hosted rows land with `counts_toward_budget = False`. I could not verify that from this repo.
- `GET /v1/usage/count` applies no source guard, and it backs the dashboard's "select all N matching", which picks rows by `!counts_toward_budget`. An operator can see a count, hit delete, and get a smaller number back with no explanation. Pre-existing, widened slightly here, worth an issue.

Verified locally: `make lint`, `make typecheck`, `make openapi-check`, `make postman-check`, the full unit suite (2330 passed), and the four affected integration modules against PostgreSQL 18.

PR #775, opened 2026-08-25
feat(usage): add pricing-provenance columns to usage_logs

khaledosman:
## Description

`usage_logs` records the settled cost but nothing about where that amount came from. Five nullable columns add it, mirroring the platform's `gateway_usage_settlement`: `pricing_source`, `pricing_reference`, `pricing_effective_at`, `pricing_version`, and `calculated_at` (when the amount was priced, which is not `timestamp`, when the request ran).

Additive, with no backfill: this edition's settlement path records no provenance, so null means "not recorded". The platform's hosted usage backfill is what writes them, and because it is insert-only, the columns have to exist before it runs rather than after.

Verified on both engines. `tests/unit/test_pricing_provenance_schema_chain.py` covers SQLite (types, nullability, the migration matching the model, and a downgrade/upgrade round trip that leaves the rows intact); `tests/integration/test_pricing_provenance_columns.py` covers PostgreSQL (the enforced varchar ceilings, an aware `timestamptz` round trip, and the same round trip through `ALTER TABLE`). `alembic heads` stays single, at `a9c4e2b6d8f1`.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes mozilla-ai/otari-ai#1800

## Checklist

- [x] I ran the Definition of Done checks locally: `make lint`, `make typecheck`, all of `tests/unit`, the usage, pricing and money integration tests, and the OSS-edition smoke.
- [x] Documentation was updated where necessary (no docs affected; the columns are internal schema).
- [x] If the API contract changed, I regenerated the OpenAPI spec (no contract change; `make openapi-check` and `make postman-check` are clean).

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

Written from mozilla-ai/otari-ai#1800, with the column types read off the platform's `gateway_usage_settlement` model rather than the issue's table.

Authored by Claude Opus 5 (1M context) via Claude Code.

### DIFF
--- a/alembic/versions/a9c4e2b6d8f1_add_usage_logs_pricing_provenance.py
+++ b/alembic/versions/a9c4e2b6d8f1_add_usage_logs_pricing_provenance.py
@@ -1,0 +1,56 @@
+"""Add pricing provenance to usage_logs.
+
+Records why a row's ``cost`` is the amount it is: which price list settled it,
+which entry in that list was applied, from when that rate was in force, which
+revision of the list it came from, and when the pricing happened.
+
+The platform's ``gateway_usage_settlement`` carries this and ``usage_logs``
+cannot re-derive it, so without these columns the hosted-usage backfill
+(mozilla-ai/otari-ai#1798) drops the provenance of every row it writes and the
+truth table cannot answer why an amount is what it is. That backfill is
+insert-only, so a row it has already written is never revisited: these columns
+have to exist before it runs, not after.
+
+``calculated_at`` is deliberately separate from ``timestamp``. One is when the
+amount was priced, the other when the request ran, and usage settled or repriced
+later moves them apart.
+
+All five are nullable with no backfill, and the lengths mirror the platform's
+settlement columns so a value copied across always fits. The gateway's own
+settlement path records no provenance, so null reads correctly as "not
+recorded"; inventing a source for a row nobody priced that way would be a lie.
+
+Revision ID: a9c4e2b6d8f1
+Revises: d8b3f1c6a4e9
+Create Date: 2026-08-25 00:00:00.000000
+
+"""
+
+from collections.abc import Sequence
+
+import sqlalchemy as sa
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision: str = "a9c4e2b6d8f1"
+down_revision: str | Sequence[str] | None = "d8b3f1c6a4e9"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+
+def upgrade() -> None:
+    """Upgrade schema."""
+    op.add_column("usage_logs", sa.Column("pricing_source", sa.String(length=32), nullable=True))
+    op.add_column("usage_logs", sa.Column("pricing_reference", sa.String(length=511), nullable=True))
+    op.add_column("usage_logs", sa.Column("pricing_effective_at", sa.DateTime(timezone=True), nullable=True))
+    op.add_column("usage_logs", sa.Column("pricing_version", sa.String(length=255), nullable=True))
+    op.add_column("usage_logs", sa.Column("calculated_at", sa.DateTime(timezone=True), nullable=True))
+
+
+def downgrade() -> None:
+    """Downgrade schema."""
+    op.drop_column("usage_logs", "calculated_at")
+    op.drop_column("usage_logs", "pricing_version")
+    op.drop_column("usage_logs", "pricing_effective_at")
+    op.drop_column("usage_logs", "pricing_reference")
+    op.drop_column("usage_logs", "pricing_source")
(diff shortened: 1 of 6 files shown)


njbrake (approved):
Approving. The change is a careful copy of its own precedent (`f4c6a8b0d2e5_add_usage_logs_routing_attribution.py`: five nullable additive columns on `usage_logs`, no backfill, a docstring arguing why null is the honest value), and the ordering argument for landing schema ahead of an insert-only backfill holds, so dormant columns are the right call rather than a smell. Verified locally: single `alembic heads` at `a9c4e2b6d8f1`, both new test files green (SQLite, and PostgreSQL 18 interleaved with the usage suites to check for schema contamination), `make lint` and `mypy` clean.

One thing would have bitten, and I pushed the fix to this branch as 2a8dc85. `set_usage_price` selects on `source != "gateway"` and `counts_toward_budget = False`, which is the same set an external backfill writes, and it rewrites `cost` / `billing_meters` / `pricing_breakdown` while leaving the five new columns naming the price list that produced the amount the row no longer has. That is the failure these columns exist to prevent, one endpoint away. The fix clears them on any row the pass updates, with a regression test in `test_usage_admin.py`.

Three follow-ups, none blocking:

**How does the backfill write these?** There is no write path here and none anywhere in the repo. If it is `/v1/usage/external-events`, `ExternalUsageEvent` needs the five fields. If the platform writes directly into the gateway's database, that is a new coupling the ingest API was built to avoid, and worth saying out loud.

**The platform settlement body already carries this and we drop it.** `_UsagePricing` models only `source`; the payload also has `pricing.reference` and a top-level `calculated_at` (`tests/unit/test_run_platform_attempts.py:431`). Standalone settlement knows its source too, since `find_model_pricing` resolved it. So "this edition's settlement path records no provenance" is true of the code but not of the information available to it; these could stop being dormant with modest follow-up.

**`pricing_source` now means three things in this codebase**: this column, `_platform.SettledCost.pricing_source` (the hybrid wire field callers see as `usage.pricing_source`), and `models.py`'s "configured"/"default"/"dynamic"/"none". One line in the model comment naming which vocabulary the column speaks would save a reader the collision.

Minor: consider dropping `test_the_revision_round_trips_on_postgresql_with_rows_in_the_table`. It downgrades the shared per-worker schema, and `conftest._SCHEMA_READY` means that if its `finally` upgrade ever fails, nothing repairs it for the rest of the worker's session. `ADD COLUMN ... NULL` is about the safest DDL there is and the SQLite round trip already proves rows survive.

khaledosman:
Thanks. `2a8dc85` kept and built on, not reordered or squashed: the repricing hole is real and one endpoint away, exactly as you describe. `tests/integration/test_usage_admin.py` is green on the new head (26 passed, including `test_set_price_clears_the_provenance_of_the_amount_it_replaced`).

**Vocabulary named** (`ae387ef`). The comment now says the column speaks the platform's settlement vocabulary, the values `_platform.SettledCost.pricing_source` already carries on the hybrid wire, and that it is not `api/routes/models.py`'s `"configured"` / `"default"` / `"dynamic"` / `"none"`, which answers where a price list entry came from rather than what settled one row's amount.

**PostgreSQL downgrade pass dropped** (`41c4b69`). Agreed on `_SCHEMA_READY`: a failed restore poisons the rest of that worker's session, and `ADD COLUMN ... NULL` does not need a second engine to be believed. What only PostgreSQL can show stays (enforced varchar ceilings, an aware `timestamptz` round trip, the oversized value refused rather than truncated), and the SQLite chain test still covers the downgrade and re-upgrade with rows in the table.

**How the backfill writes these: directly to the target database.** Not through `/v1/usage/external-events`, so `ExternalUsageEvent` does not need the five fields. On the platform side, `backend/app/migration/settlement_target.py:630` does `sa.insert(USAGE_LOGS)` narrowed by a `columns` set so it names only what the target actually has, which is the same mechanism that detects these columns and why `PRICING_PROVENANCE_COLUMNS` needs no change. You are right that this is direct coupling and worth saying out loud, and you reached the same conclusion yourself while reviewing #772. The mitigation is scope rather than design: it is one-shot cutover machinery pointed at an explicit `--target-url`, so no serving code acquires the coupling and it is not an ongoing path. Whether that gets recorded as an architectural note is your call.

**The other two follow-ups are real and out of scope here.** `_UsagePricing` dropping `pricing.reference` and the top-level `calculated_at`, and standalone settlement already knowing its source through `find_model_pricing`, are both otari-side product calls about who writes these columns rather than schema work, so they want their own issue rather than folding into this PR. I have not filed one.

_Reply written by Claude Opus 5 (1M context) via Claude Code._

khaledosman:
Correction to the note above: the two follow-ups are filed after all, not left unfiled.

- #782 covers settlement recording provenance from what it already resolved, including `_UsagePricing` dropping `pricing.reference` and the top-level `calculated_at`.
- #781 covers the usage count and the usage mutations resolving different row sets, from your review of #773.
- #783 covers the two definitions of the `otari-ai:` prefix, now that #772 has landed.

Filed by Claude Opus 5 (1M context) at @khaledosman's request.

PR #776, opened 2026-08-25
chore(dashboard): ui cleanup, drop the account menu identity block

jigjigjig:
## Description

Removes the identity block at the foot of the account menu: the avatar, the "Operator" name, and the session detail line under it.

It repeated the name and initials the rail's own trigger already shows one row above it, and spent its second line on a session detail that is neither actionable nor, on an unclaimed deployment, reliably about this caller. Nothing else in the menu depended on it.

Consequences:

- `sessionIdentity` keeps the name and initials the trigger uses and loses its `detail` half, so `AccountMenu` no longer reads `sign_in_methods`.
- The divider above Log out stays, so that row is still set apart.
- The two `AccountMenu` tests that asserted only the removed line's wording go with it; `openMenu` loses its now-unused `signInMethods` parameter.

One behavior change worth naming, since it goes past "delete a redundant block": `identity` is now consumed only by the menu's trigger, and the trigger renders the name only while the rail is expanded. On a collapsed rail the open menu therefore shows no name at all, and a `hosted_user` session is reduced to the trigger's `··` over "Signed in". That is intended: the rail expands, and `/account` is where an identity is actually read.

Nothing else outside this menu changes. `PasswordCard` still reads `sign_in_methods` and still owns the claim flow, with its own tests.

## PR Type

- [ ] New Feature
- [ ] Bug Fix
- [x] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

None. Reported directly: the block reads as dead weight in the menu.

## Checklist

Notes on the boxes above:

- Tests: this deletes UI, so the change is covered by removing the two assertions that named the deleted text. No new test is owed; the surviving `AccountMenu` test still covers the menu's one real destination.
- Checks: this touches `web/` only, so the counterpart commands are the ones that apply there, all green on this branch: `pnpm --dir web run lint`, `pnpm --dir web run typecheck`, `pnpm --dir web vitest run src/app/nav/AccountMenu.test.tsx`. `src/app/nav/navigationHistory.test.ts` fails on this branch and on `main` alike and is unrelated.
- Documentation: nothing to update. `docs/dashboard.md` does not describe this block.
- OpenAPI: no API contract change, so the spec and the Postman collection are untouched.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

The removal, the commit, and this description were written by the agent. The decision to remove the block was the author's; the agent was asked to agree or push back before making the change.

**NOTE:**

## Summary

- Removed the redundant identity block from the account menu.
- Kept the user name and initials in the menu trigger.
- Kept the divider above “Log out”.
- Updated tests to match the simpler menu.
- Removed unused test setup and parameters.

## Technical notes

- Typed `sessionIdentity` with `DeploymentBootstrap["session_type"]`.
- Removed the `sign_in_methods` dependency from `AccountMenu`.
- Validation passed for linting, type checking, and account menu tests.


web/src/app/nav/AccountMenu.tsx
```diff
@@ -65,42 +65,18 @@ const MENU_ROW_DISABLED = "cursor-not-allowed text-muted opacity-60"
...
-// but a member who signed up and signed in by calling `POST /v1/auth/session`
-// directly holds one too, and naming a credential here would name theirs
-// wrongly. So the unclaimed case reports the deployment's state, which is the
-// part this actually knows, and the operator reads it as the standing
-// invitation to claim that it is.
-function sessionIdentity(
-  sessionType: string,
-  signInMethods: readonly string[],
-): {
+// Whose session this is, as the trigger at the foot of the rail names it. A
```
jigjigjig, line 72: While this comment is being rewritten: the file header above it (line 31) is now stale. It still lists what the menu opens onto as "account settings, appearance, the legal pages, who you are signed in as, and the way out", and the fourth of those is what this PR deletes. Drop that clause, or the next reader goes looking for the row it promises.

web/src/app/nav/AccountMenu.tsx
```diff
@@ -247,9 +223,9 @@ function AppearanceControl() {
 
 export function AccountMenu({ collapsed }: { collapsed: boolean }) {
   const { logout } = useAuth()
-  const { session_type, sign_in_methods, management_url } = useDeployment()
+  const { session_type, management_url } = useDeployment()
```
jigjigjig: Worth stating explicitly: with the footer block gone, `identity` is consumed only by the trigger, and the trigger renders the name only when the rail is expanded (`collapsed ? null : ...`). So on a collapsed rail the open menu now shows no name anywhere, and a `hosted_user` session is reduced to `··` over "Signed in". Fine if intended (the rail expands, and /account names the identity), but it is a behavior change past "remove a useless component" and belongs in the PR description rather than being found later.

web/src/app/nav/AccountMenu.test.tsx
```diff
@@ -3,20 +3,17 @@ import userEvent from "@testing-library/user-event"
 import { afterEach, describe, expect, it, vi } from "vitest"
```
jigjigjig: `vi` is dead now, and so is the `afterEach(() => vi.restoreAllMocks())` below: the only test that mocked anything (`vi.spyOn(globalThis, "fetch")`) left with the claim describe. Biome will not catch it, since the `afterEach` still references `vi`. Drop `vi` from this import and delete the `afterEach`.

jigjigjig (commented):
Reviewed by Claude Opus 5 (Claude Code) against AGENTS.md, web/AGENTS.md, and the frontend-standards instructions.

The removal is clean: no orphaned references, `sign_in_methods` is still read by `PasswordCard` and `Login` where it means something, `text-chrome-meta` and `text-chrome-initials` both still have callers so no token goes dead, and no e2e or screenshot fixture touched the deleted block. Three inline notes, none blocking. (COMMENT rather than APPROVE only because GitHub will not let an author approve their own PR.)

web/src/app/nav/AccountMenu.tsx
```diff
@@ -65,42 +65,18 @@ const MENU_ROW_DISABLED = "cursor-not-allowed text-muted opacity-60"
...
-// invitation to claim that it is.
-function sessionIdentity(
-  sessionType: string,
-  signInMethods: readonly string[],
-): {
+// Whose session this is, as the trigger at the foot of the rail names it. A
+// standalone gateway issues one, for the operator identity it provisioned
+// itself, and no management route reports the caller's own name or address, so
+// this is the most it can honestly say.
+function sessionIdentity(sessionType: string): {
```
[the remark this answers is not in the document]
jigjigjig: Fixed in d0def9a. Agreed, and it was a pre-existing weakness this PR made easier to fix: the parameter is now typed `DeploymentBootstrap["session_type"]`, so the helper is checked against the generated union (`local_operator | hosted_user | none`) and a change there reaches this branch instead of silently falling through to the default.

web/src/app/nav/AccountMenu.tsx
jigjigjig: Fixed in d0def9a. The header now says the menu opens onto account settings, appearance, the legal pages, and the way out, and states where the identity actually lives (the trigger's own line, and /account) so the removal reads as deliberate rather than as an omission.

web/src/app/nav/AccountMenu.tsx
jigjigjig: Intended, and now written down: the PR description carries a paragraph naming it. The rail expands, and /account is the page that actually reports the identity, so nothing is lost that a click does not recover. Restoring a name inside the menu would put back the duplication this PR removes, so no code change here.

web/src/app/nav/AccountMenu.test.tsx
jigjigjig: Fixed in d0def9a. `vi` is out of the vitest import and the `afterEach` is deleted; the file now imports only `describe`, `expect`, `it`. Lint, typecheck, and the suite stay green.

khaledosman (approved):
No findings.

The deleted claim-propagation test is not a coverage loss: `useDeployment.test.tsx:64` covers the retirement, and no remaining `sign_in_methods` consumer lost a test. `identity.detail` and `sign_in_methods` are gone from the file with no leftovers.

PR #777, opened 2026-08-25
fix(api): expose prompt cache keys

### LINKED ISSUE #514: Expose prompt_cache_key in gateway schemas
## Desired behavior

Upgrade to the any-llm release containing `prompt_cache_key` on `CompletionParams` and `MessagesParams`. Confirm the derived `/v1/chat/completions` and `/v1/messages` schemas expose it, then regenerate the OpenAPI document and published client SDKs.

Add gateway tests proving an OpenAI target receives the key and an unsupported final provider returns a clear client error.

## Why

Otari derives both request schemas from any-llm, but the current dependency and generated clients omit `prompt_cache_key`. Existing generated `MessagesRequest.from_dict()` therefore drops the field before HTTP, even though the any-llm Otari provider can pass it to the gateway.

## Notes

Blocked by mozilla-ai/any-llm#1247. This issue covers the OSS gateway and generated Otari SDKs; the hand-written platform endpoint in `mozilla-ai/otari-ai` is a separate surface.

HareeshBahuleyan:
## Description

`prompt_cache_key` was dropped from chat completion and Messages requests because Otari was locked to an older any-llm release. Upgrade to any-llm-sdk 1.25.0, expose and forward the field on both schemas, and return an actionable 400 when the selected provider cannot support it.

The OpenAPI document and dashboard client schema are regenerated. Tests cover schema exposure, OpenAI forwarding, Messages forwarding, unsupported-provider errors, and the Bedrock client construction change in any-llm 1.25.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes #514

## Checklist

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:** GPT-5.6 Sol with Pi

**Any additional AI details you'd like to share:** Implemented and verified directly against the backend, OpenAPI/Postman, and dashboard client checks.

**NOTE:**

## Summary

- Upgraded `any-llm-sdk` to 1.25.0.
- Added `prompt_cache_key` to chat completion, Messages, and Responses requests.
- Scoped cache keys to authenticated users and removed them when no authenticated scope exists.
- Added clear HTTP 400 errors for unsupported providers.
- Updated generated API and dashboard client schemas.
- Added tests for schema exposure, key scoping, forwarding, error handling, and Bedrock routing.

These changes provide safer, consistent prompt caching across supported gateway endpoints.

### DIFF
--- a/src/gateway/api/routes/_pipeline.py
+++ b/src/gateway/api/routes/_pipeline.py
@@ -332,8 +332,17 @@ def _unsupported_feature_detail(exc: BaseException) -> str:
         if isinstance(candidate, UnsupportedParameterError):
             # AnyLLMError.__str__ prefixes the provider to ``message``. Feeding
             # both through upstream_error_message would repeat the same reason.
-            return redact_upstream_message(candidate.message)
+            return _redacted_caller_fault_detail(candidate.message, PROVIDER_BAD_REQUEST_DETAIL)
     return _caller_fault_detail(exc, PROVIDER_BAD_REQUEST_DETAIL)
+
+
+def _redacted_caller_fault_detail(message: str, fallback: str) -> str:
+    """Return redacted explanatory text, or a fixed detail when none remains."""
+    redacted = redact_upstream_message(message)
+    explanatory = redacted.replace("[redacted]", "")
+    if not any(char.isalpha() for char in explanatory):
+        return fallback
+    return redacted
 
 
 def _caller_fault_detail(exc: BaseException, fallback: str) -> str:
@@ -350,11 +359,7 @@ def _caller_fault_detail(exc: BaseException, fallback: str) -> str:
     A message made entirely of redaction placeholders is empty for this
     purpose, too.
     """
-    redacted = redact_upstream_message(upstream_error_message(exc))
-    explanatory = redacted.replace("[redacted]", "")
-    if not any(char.isalpha() for char in explanatory):
-        return fallback
-    return redacted
+    return _redacted_caller_fault_detail(upstream_error_message(exc), fallback)
 
 
 def classify_provider_error(exc: BaseException) -> ProviderErrorMapping | None:
--- a/src/gateway/api/routes/_pipeline.py
+++ b/src/gateway/api/routes/_pipeline.py
@@ -46,7 +46,7 @@ from typing import Any, Generic, Literal, NamedTuple, NoReturn, Protocol, TypeVa
 from urllib.parse import ParseResult, urlparse
 
 from any_llm import LLMProvider
-from any_llm.exceptions import AnyLLMError
+from any_llm.exceptions import AnyLLMError, UnsupportedParameterError
 from any_llm.types.completion import (
     ChatCompletion,
     ChatCompletionChunk,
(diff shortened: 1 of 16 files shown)


src/gateway/api/routes/_pipeline.py
```diff
@@ -300,12 +300,23 @@ class _PendingUsageReport(NamedTuple):
...
+    return any(isinstance(candidate, unsupported_types) for candidate in upstream_exception_chain(exc))
+
+
+def _unsupported_feature_detail(exc: BaseException) -> str:
+    """Return one actionable reason for a locally rejected request feature."""
+    for candidate in upstream_exception_chain(exc):
+        if isinstance(candidate, UnsupportedParameterError):
+            # AnyLLMError.__str__ prefixes the provider to ``message``. Feeding
+            # both through upstream_error_message would repeat the same reason.
+            return redact_upstream_message(candidate.message)
```
khaledosman: This drops the emptiness guard `_caller_fault_detail` applies two lines below. `redact_upstream_message` returns `""` outright on a `_PAYLOAD_ECHO` hit, and that regex's own comment says such a message should "make the caller-fault classifier use its fallback". An `UnsupportedParameterError` whose `additional_message` echoes the request (bedrock's variant already appends free text, and the URL in it is redacted today) would ship a 400 with `detail: ""`, or with a bare `"[redacted]"` if the secret shapes eat the rest.

Not reachable with any-llm 1.25's current message set, but that is the only thing holding it, and the guard exists because provider text is unpredictable. Route the result through the same check:

```python
detail = redact_upstream_message(candidate.message)
explanatory = detail.replace("[redacted]", "")
return detail if any(char.isalpha() for char in explanatory) else PROVIDER_BAD_REQUEST_DETAIL
```

tests/integration/test_messages_endpoint.py
```diff
@@ -36,18 +36,22 @@ def test_messages_endpoint_basic_completion(
...
+    with patch("gateway.api.routes.messages.amessages", new=mock_amessages):
         response = client.post(
             "/v1/messages",
             json=messages_request_body,
             headers=master_key_header,
         )
 
     assert response.status_code == 200
+    assert mock_amessages.await_args is not None
+    assert mock_amessages.await_args.kwargs["prompt_cache_key"] == "tenant-session-123"
```
khaledosman: Non-blocking, but worth a deliberate call: `prompt_cache_key` is caller-controlled and forwarded verbatim. On a deployment that multiplexes tenants onto one upstream provider key it becomes a shared cache-routing namespace, so two tenants can pin themselves to the same machine and read each other's `cached_tokens` as a prefix oracle. `resolve_user_id` is the precedent for binding a client-supplied tag to the authenticated principal; namespacing the value per resolved user before forwarding would close it. Fine to accept as-is, but the derivation comment in `_schema_derive.py` calls out that automatic pickup is "the point for benign params", so this one deserves the explicit answer.

khaledosman (approved):
Title: `_schema_derive.py` picks up a new caller-facing request field automatically, which is a feature rather than a fix. `cliff.toml:89` routes `^fix` to Bug Fixes, and squash-merge makes the title the changelog line, so `feat(api):` is the one that lands it in the right section.

Checked `python3 scripts/generate_postman.py --check` against this head: clean, so no Postman regeneration is owed (both touched schemas carry `EXAMPLE_BODY_OVERRIDES`).

src/gateway/api/routes/_pipeline.py
HareeshBahuleyan: Fixed in 8ad4f499. Typed unsupported-parameter details now use the same explanatory-text fallback guard as other caller-fault messages. I also added regression coverage for a payload echo that redacts to an empty string.

tests/integration/test_messages_endpoint.py
HareeshBahuleyan: Fixed in 47523daa. Chat, Messages, and Responses now hash each caller key with the authenticated user scope before provider dispatch. Hybrid mode uses the platform-resolved user, falls back to the resolved workspace for older peers, and omits the routing hint if neither identity is available. The tests cover standalone isolation, hybrid isolation, the legacy fallback, and all three request formats.

PR #780, opened 2026-08-25
fix(dashboard): end the create-workspace flow inside the new workspace

jigjigjig:
https://github.com/user-attachments/assets/d51bee78-59ec-487d-8b5e-4e5b3f100a69

## Description

Creating a workspace from the scope switcher closed the modal and left the operator where they started, so the workspace they had just asked for existed only as a row in a list they had to go back to and pick. The switcher now follows it: the shell's scope moves to the new workspace and the app lands on the overview, which is where that scope reads. The Workspaces list page keeps its own behavior, because it is already showing the row that appeared.

The submit that navigates says so ("Create and open") and holds the press for 800ms with a spinner in the label's place, so the page changes after an acknowledged press rather than under an unacknowledged one. The label is faded rather than hidden, so the button keeps its accessible name for the length of the wait, and the spinner takes the label's place rather than sitting beside it, so nothing moves.

Two things the error path was getting wrong come with it, both found while exercising this flow:

- A refused create is now carried by the name that caused it, as the input's own error state in the description's place. Every way `POST /v1/workspaces` refuses is about the name (taken, empty, too long), and the banner above the form resized the modal that framed it. The default-budget call keeps the banner, since it fails after the workspace already exists and is about no field.
- The modal's width is pinned to the `max-w-md` cap it was already growing to. HeroUI's `.modal__container` is `sm:w-fit` and `size="md"` only caps the dialog, so the modal was as wide as whatever was in it and jumped the moment a message longer than the fields appeared.

`Field` gained two optional props (`isInvalid`, `errorMessage`) so a form can put a refusal on the input that caused it. Additive: no existing caller behaves differently.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

None; reported directly.

## Checklist

Notes on the three above that are not a plain yes:

- Tests are the dashboard's, not `tests/unit` / `tests/integration`: this change is entirely under `web/src`. A new `WorkspaceSwitcher.test.tsx` covers create then enter, the spinner and label swap, and the absence of either at rest; `WorkspacesPage.test.tsx` gains the refused-create case (input invalid, message in the description's place, cleared on edit).
- The Definition of Done targets are the Python ones and this PR touches no Python. Their dashboard counterparts were run and are clean: `pnpm --dir web run lint` (Biome), `pnpm --dir web exec tsc -b`, and vitest (176 passing over the touched suites, and the full `dashboard` job green in CI).
- No documentation and no API contract changed: dashboard-only, no route, schema, or docstring touched, so neither generated artifact moves.

## Review follow-ups

- **Dismissal during the hold** (CodeRabbit, and Copilot's suppressed comment: same finding). The 800ms outlives the render that started it, so dismissing the form during it still ran the completion and took the operator into the new workspace after they had said not to. Fixed in 3fc6d038: the completion drops when the form is no longer the one on screen, and Cancel is disabled while the create is in flight the way `ConfirmDialog` already does with its own. Cancel alone would not have covered it, since Escape and a click outside dismiss the modal without going through a button, which is the path the new test uses.
- **Test isolation** (found by CI, not by a reviewer). The selected workspace is remembered per browser and the switcher tests never cleared it, so the first test's choice seeded the second. It passed locally only because that Node build leaves `window.localStorage` undefined. Fixed in b4427c73.
- **A refusal the name cannot fix** (Copilot, second pass). Any create error reddened the Name field, so a 403, a 5xx or a dropped connection told the operator and assistive tech that the name was the problem. Fixed in ee234626: only 400, 409 and 422 reach the field; everything else keeps the banner, and a test covers the 500 case.
- **Declined, with reasons in the thread**: guarding `setHolding(false)` against unmount (React 18 made unmounted `setState` a no-op and dropped the warning; the `active` ref guards the navigation, which is an observable effect, not state writes), and converting the switcher test to fake timers (`userEvent` needs `advanceTimers` wired through react-aria's pointer events, and the explicit 3s override is the honest expression of a hold longer than Testing Library's 1s default).
- **A doc comment on the wrong declaration** (found reviewing my own diff). The hold constant had landed between `CreateWorkspaceForm`'s JSDoc and the function, so the comment read as describing the constant. Fixed in 28817d31.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

Written and iterated against a locally running gateway and dashboard, with the reviewer driving the flow in the browser between rounds. The spinner color and the modal width were both diagnosed from HeroUI's own CSS (`.spinner--accent` resolving to the button's fill, `.modal__container` being `sm:w-fit`) rather than guessed at.

**NOTE:**

## Summary

- The scope switcher now supports organization switching and organization creation.
- The create-workspace flow now uses “Create and open” and opens the new workspace after creation.
- The flow shows loading feedback, disables Cancel during creation, and prevents navigation after dismissal.
- Name-related API errors now appear beside the name field. Other errors remain in the banner.
- Workspace and switcher tests now cover navigation, loading, dismissal, error handling, organization actions, and deterministic completion handling.


web/src/features/workspaces/WorkspacesPage.tsx
```diff
@@ -273,23 +326,61 @@ export function CreateWorkspaceForm({ onClose }: { onClose: () => void }) {
                   // here leaves the workspace created and undefaulted, which the
                   // banner reports and the edit form can finish.
                   onSuccess: (workspace) => {
-                    if (budgetId === NO_DEFAULT) {
+                    const finish = async () => {
+                      await held
+                      // No setHolding here: the form unmounts on close, and the
+                      // failure paths below are what release the button.
                       onClose()
+                      onCreated?.(workspace)
```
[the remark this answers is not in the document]
jigjigjig: Confirmed and fixed in 3fc6d038: the hold outlives the render that started it, so dismissal during it ran the completion and took the operator into a workspace they had just said not to enter.

The completion now drops when the form is no longer the one on screen (a ref cleared on unmount, checked after `await held`). Cancel is disabled while the create is in flight as well, the way `ConfirmDialog` already does with its own, but that alone would not have covered this: Escape and a click outside dismiss the modal without going through a button, which is the path the new test uses.

`WorkspaceSwitcher.test.tsx` gains "does not enter a workspace the operator dismissed the form over", asserting past the hold that the route did not change and the switcher still names the old workspace. It fails without the guard, so it pins the behavior rather than restating it.

The workspace itself is still created, since the request had already been accepted; it just shows up in the list and the switcher instead of being navigated into.

web/src/features/workspaces/WorkspacesPage.tsx
```diff
@@ -273,25 +338,71 @@ export function CreateWorkspaceForm({ onClose }: { onClose: () => void }) {
...
                         workspaceId: workspace.id,
                         body: { budget_id: budgetId },
                       },
-                      { onSuccess: onClose },
+                      {
+                        onSuccess: () => {
+                          void finish()
+                        },
+                        onError: () => setHolding(false),
+                      },
```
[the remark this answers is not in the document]
jigjigjig: Skipping this one, and it is a factual disagreement rather than a preference: since React 18, `setState` on an unmounted component is a no-op and the "can't perform a React state update on an unmounted component" warning was removed precisely because guarding it was cargo-culted noise. `setHolding(false)` after unmount neither warns nor leaks here.

The `active` ref is not there to police state writes. It guards `onClose()` and `onCreated()`, which are observable side effects: closing a modal that is already gone is harmless, but `onCreated` navigates and changes the shell's scope, which is exactly the bug you and CodeRabbit both caught. Extending the ref to the `onError` paths would imply those writes are dangerous too, and a reader would then reasonably wonder which of the two reasons applied where.

Your other two comments from this pass:

- **`aria-invalid` for non-name failures** (line 305): agreed and fixed in ee234626. Only the refusals this endpoint gives when the input is at fault (400, 409, 422) reach the field; a 403, a 5xx or a dropped connection keeps the banner, since none of them is something the operator can retype their way out of. A test covers the 500 case.
- **Real timers in the switcher test** (line 162): keeping them. `userEvent` needs `advanceTimers` wired to `vi.advanceTimersByTime` to work under fake timers, and driving react-aria's pointer events through that is its own source of flake. The explicit `timeout: 3000` is there because the hold is deliberately longer than Testing Library's 1s default, so the override is the honest expression of the wait rather than a workaround. Cost is roughly 2.5s of wall time in a suite that runs 1384 tests.

jigjigjig:
Review round closed out. State: all checks green, both bot reviewers' findings addressed, no unresolved threads.

**Applied**

| Finding | From | Fix |
| --- | --- | --- |
| Dismissing the form during the 800ms hold still entered the new workspace | CodeRabbit, and Copilot's suppressed comment (same bug) | 3fc6d038 — completion drops when the form is no longer on screen; Cancel disabled while in flight |
| Any create error reddened the Name field, including ones the name cannot fix | Copilot, second pass | ee234626 — only 400/409/422 reach the field, everything else keeps the banner |
| Hold constant sat between `CreateWorkspaceForm`'s JSDoc and the function | my own re-read of the diff | 28817d31 |
| Switcher tests leaked a remembered workspace between cases | CI (passed locally, where this Node build has no `localStorage`) | b4427c73 |

**Declined**, both argued in their threads rather than waved off: guarding `setHolding(false)` against unmount (React 18 made unmounted `setState` a no-op and removed the warning; the `active` ref guards the navigation, which is an observable effect), and converting the switcher test to fake timers (`userEvent` needs `advanceTimers` wired through react-aria's pointer events; the explicit 3s override is the honest expression of a hold longer than Testing Library's 1s default).

Two of the four applied fixes were bugs in the review-round work itself rather than in the original change, which is worth knowing if you are reading the commits in order.

Ready for a human pass. @njbrake @khaledosman — the behavior is easiest to judge by running it: switcher, Create workspace, then try both a duplicate name and dismissing during the spinner.

web/src/features/workspaces/WorkspacesPage.tsx
```diff
@@ -225,28 +233,92 @@ function NarrowedDefaults({
...
+  // The hold is running, on a form whose submit navigates. Kept separately from
+  // the mutations' own pending flags, because it outlives them: the request can
+  // answer in 90ms and the button still has a beat left to serve.
+  const [holding, setHolding] = useState(false)
+  // Whether this form is still the one on screen. The hold outlives the render
+  // that started it, so a dismissal during it would otherwise still hand the
+  // page over: the operator cancels and is taken somewhere anyway. Disabling
+  // Cancel is not enough on its own, because Escape and a click outside dismiss
+  // the modal too, and neither goes through a button.
+  const active = useRef(true)
```
khaledosman, line 282: **Blocker.** A cleanup-only effect over a ref initialized at declaration breaks under StrictMode, which `web/src/main.tsx:41` wraps the whole app in. The dev remount runs mount, cleanup, mount, so line 266 sets `active.current` to `false` and nothing ever sets it back. Every create then returns early at `if (!active.current) return` inside `finish()`, so `onClose()` and `onCreated()` never fire: the modal stays open and the button spins forever.

It hits both call sites, not just the switcher. `WorkspacesPage`'s own create reaches the same guard through `finish()` even with `entersWorkspace` false.

The suite misses it because `renderWithRouter` does not wrap in StrictMode. `VerifyEmailPage.tsx:30` is the precedent for this hazard, and `VerifyEmailPage.test.tsx:76` is the precedent for testing it.

```tsx
useEffect(() => {
  active.current = true
  return () => {
    active.current = false
  }
}, [])
```

web/src/app/nav/WorkspaceSwitcher.test.tsx
```diff
@@ -1,206 +1,195 @@
...
-      "Organization not found",
-    )
-    // Still open, which is what makes the message readable at all.
-    expect(screen.getByRole("dialog")).toBeInTheDocument()
+    // Both halves of "entering" it: the shell's scope moved to the new
+    // workspace, and the flow ended on the page that scope reads, rather than
+    // back on the page the modal was opened from. The timeout clears the hold
+    // with room to spare.
+    expect(
+      await screen.findByText("OVERVIEW PAGE", undefined, { timeout: 3000 }),
```
khaledosman: Per-assertion timeout override. `testing.md`: "A per-assertion `{ timeout: 5000 }` is a review blocker. It hides which wait was slow." This one exists only to outlast the 800ms `ENTER_HOLD_MS`, so the fix is to control the clock rather than to widen the ceiling: `vi.useFakeTimers()` around the press, or make the hold injectable so the test can set it to 0. Either way the assertion goes back to a plain `findByText`.

web/src/app/nav/WorkspaceSwitcher.test.tsx
```diff
@@ -1,206 +1,195 @@
...
+    // Escape rather than Cancel: Cancel is disabled while the create is in
+    // flight, so dismissal is what is left, and it is the path that bypasses
+    // every button.
+    await user.keyboard("{Escape}")
+
+    // Past the hold, so this is the completion having been dropped rather than
+    // not having run yet. The workspace was created and the switcher will list
+    // it; what must not happen is being taken there after saying not to.
+    await new Promise((resolve) => {
+      setTimeout(resolve, ENTER_HOLD_MS * 2)
```
khaledosman: A literal 1.6s sleep. `testing.md`: "Wait on the event, not the clock. `findBy*` and `waitFor` resolve the instant the DOM changes; their timeout is a ceiling, not a sleep." Same root cause as the `{ timeout: 3000 }` above: with fake timers this becomes `await vi.advanceTimersByTimeAsync(ENTER_HOLD_MS)` and the two cases stop costing real seconds on every run. It also removes the need to mirror the private `ENTER_HOLD_MS` at line 19, which the comment there already flags as a value that would keep passing if the constant were deleted.

web/src/app/nav/WorkspaceSwitcher.test.tsx
```diff
@@ -1,206 +1,195 @@
...
+    expect(screen.queryByText("OVERVIEW PAGE")).toBeNull()
+    // Still found by its name while pending: the label is faded rather than
+    // hidden, so the button does not go unnamed for the length of the wait.
+    const submit = within(form).getByRole("button", { name: /Create and open/ })
+    expect(submit).toHaveAttribute("data-pending", "true")
+    // The spinner takes the label's place, in the label's color rather than the
+    // default accent, which on this variant is the button's own fill.
+    const spinner = submit.querySelector("[data-slot='spinner']")
+    expect(spinner).not.toBeNull()
+    expect(spinner?.className).toContain("spinner--current")
```
khaledosman: These three assert HeroUI internals: a `data-slot` selector, the library's own `spinner--current` class, and a Tailwind class on the label span. `testing.md`: "never a class selector: `.bg-surface` is a token that will be renamed, and a test that breaks on a restyle teaches everyone to stop trusting the suite." `data-pending` at line 148 plus the accessible name already carry the behavior the comment describes (the button stays named while pending); the color and the opacity are styling that a HeroUI bump can move without breaking anything real.

web/src/app/nav/WorkspaceOrganizationSwitcher.test.tsx
```diff
@@ -0,0 +1,214 @@
+import { screen, within } from "@testing-library/react"
```
khaledosman: There is no `WorkspaceOrganizationSwitcher.tsx`; this file covers `WorkspaceSwitcher`. `naming-conventions.md`: "Tests: `<Name>.test.ts(x)`, beside the file they cover." Fold these back into `WorkspaceSwitcher.test.tsx` as a second `describe`, or name the file after what it actually covers. As it stands, someone editing `WorkspaceSwitcher.tsx` and looking for its tests finds only half of them.

khaledosman (changes requested):
One blocker, plus four test-hygiene findings against `testing.md` and `naming-conventions.md`. The blocker is dev-only, which is exactly why the suite is green on it: `renderWithRouter` does not wrap in StrictMode, and `main.tsx:41` does.

web/src/features/workspaces/WorkspacesPage.tsx
jigjigjig: Confirmed and fixed. You were right that the suite could not see it: with the effect changed back to cleanup-only, the new StrictMode case fails, so this was a real hang and not a theoretical one.

```tsx
useEffect(() => {
  active.current = true
  return () => {
    active.current = false
  }
}, [])
```

`still completes after StrictMode's development remount` renders the switcher inside `<StrictMode>` and asserts the flow still lands on the overview, following `VerifyEmailPage.test.tsx`'s precedent. Both call sites are covered by it, since the guard is inside `finish()` where the list page's create reaches it too.

web/src/app/nav/WorkspaceSwitcher.test.tsx
jigjigjig: Fixed: the override is gone and the assertion is a plain `findByText`.

The hold is 800ms, which is inside `findBy*`'s own 1s ceiling, so nothing needs widening; the override was defensive and unjustified. I tried the fake-timer route you suggested first and backed it out: `userEvent` drives react-aria's pointer events through real timers, and with `vi.useFakeTimers()` (even wired through `advanceTimers`) all three cases deadlocked in the interactions rather than in the wait, while main's six organization cases kept passing. The comment in the file records that, so the next person does not spend the same hour on it.

web/src/app/nav/WorkspaceSwitcher.test.tsx
jigjigjig: Mostly fixed. The 1.6s sleep is gone, and the mirrored constant is gone with it: `ENTER_HOLD_MS` is now exported from `WorkspacesPage` and the one remaining wait is `ENTER_HOLD_MS + 100`, so it tracks the hold instead of a copy of it.

One clock wait survives, in the dismissal case only, and it is deliberate: that assertion is an *absence* (nothing navigated), and there is no event to wait on when the completion is correctly dropped. `findBy*` cannot express "still not there after the hold". It is commented as the exception rather than left to look like the rule. If you would rather that case did not wait at all, the alternative is making the hold injectable so it can be zero, which trades a real 0.9s in one test for a prop that exists only for tests; happy to switch if you prefer that trade.

web/src/app/nav/WorkspaceSwitcher.test.tsx
jigjigjig: Fixed: all three styling assertions are gone. What is left is `data-pending` plus the accessible name, which is exactly the pair you pointed at, and the `spinner--current` class, the `data-slot` selector and the `opacity-0` check are no longer asserted anywhere. A HeroUI bump can now move any of that without touching this suite.

For the record on the behavior they were standing in for: the spinner replacing the label is styling, but "the button keeps its accessible name while pending" is not, and that is still asserted, since the label is faded rather than hidden precisely so the name survives.

web/src/app/nav/WorkspaceOrganizationSwitcher.test.tsx
jigjigjig: Fixed by deleting the file. Both halves are back in `WorkspaceSwitcher.test.tsx` as two describes: `the organization half of the scope switcher` (main's cases, unchanged) and `the workspace half of the scope switcher` (this branch's three), on main's harness, which spies `apiFetch` rather than global fetch and clears the remembered workspace around every case.

Worth flagging how it got there, since it was not in the diff you first reviewed: the branch's merge of main resolved this file by keeping this branch's version and moving main's organization cases out into that new name. So the split was a merge artifact, and the four findings above were all pointing at pre-review code that the same merge had preserved. This branch now carries a second merge that keeps one file with both halves in it.

jigjigjig:
@khaledosman all five addressed, and the blocker was real: reverting the effect to cleanup-only makes the new StrictMode case fail, so this was a hang in development rather than a theoretical one.

| Finding | Outcome |
| --- | --- |
| **Blocker**: cleanup-only effect over a ref, cleared by StrictMode's remount | Fixed. Effect sets `active.current = true` on the way in. New case renders the switcher inside `<StrictMode>` and asserts the flow still lands, following `VerifyEmailPage.test.tsx` |
| Per-assertion `{ timeout: 3000 }` | Gone. The hold is inside `findBy*`'s own 1s ceiling, so a plain `findByText` does it |
| 1.6s sleep, and the mirrored `ENTER_HOLD_MS` | Sleep gone from two of three cases; the constant is now exported, so nothing mirrors it |
| Three HeroUI-internals assertions | Gone. `data-pending` plus the accessible name is all that is left, as you suggested |
| `WorkspaceOrganizationSwitcher.test.tsx` | File deleted. Both halves are two describes in `WorkspaceSwitcher.test.tsx` on main's harness |

Two things worth your attention beyond the list.

**The file split was a merge artifact, not an authoring choice.** The merge of main already on this branch resolved `WorkspaceSwitcher.test.tsx` by keeping this branch's version and moving main's six organization cases out into that new name. That is also why the other four findings pointed at pre-review code: the same merge preserved it. `Field` was the mirror image, kept as this branch's version over main's; main had reached the same two props first, so `Field` is now main's, untouched, and the create form drops its own description while a refusal is up rather than teaching a shared component to swap rows.

**One clock wait survives, in the dismissal case only**, and I left that thread open rather than resolving it. What it asserts is an absence, so there is no event for `findBy*` to wait on. The alternative is making the hold injectable so a test can set it to zero, which trades 0.9s in one test for a prop that exists only for tests. Your call; I did not want to pick it silently.

**Fake timers do not work here.** I tried your suggestion first: with `vi.useFakeTimers()`, even wired through `advanceTimers`, all three cases deadlock in `userEvent`'s interactions (react-aria drives pointer events through real timers) while main's six organization cases keep passing. That is recorded in a comment in the file.

Separately, and not from your review: this flow now diverges from `CreateOrganizationForm` beside it, which does the same create-and-move-into-it with a plain label, no hold, and a banner. The label, the 800ms spinner and the field-level error were all explicit product decisions from @jigjigjig rather than my invention, but if the two should match, the sibling is the one to align and that is a separate PR.

All checks green. No rush from my side.

web/src/features/workspaces/WorkspacesPage.tsx
```diff
@@ -217,6 +219,14 @@ function NarrowedDefaults({
   )
 }
 
+// How long the submit holds before a create that navigates hands the page over.
+// Long enough for the spinner to register as this press having done something,
+// short enough not to be a wait: the operator should read it as the button
+// acknowledging them, not as the gateway being slow. Exported so a test can wait
+// past it by name rather than by a copy of the number, which would keep passing
+// if this were changed or removed.
+export const ENTER_HOLD_MS = 800
```
khaledosman: This constant is the only thing making the workspace-half cases slow, and the coverage does not depend on its value. Measured on this head, `WorkspaceSwitcher.test.tsx` plus `WorkspacesPage.test.tsx`, 18 tests, warm:

| case | at 800 | at 50 |
|---|---|---|
| enters the workspace it just created | 891ms | 141ms |
| does not enter a workspace the operator dismissed | 990ms | 237ms |
| still completes after StrictMode's remount | 892ms | 143ms |

That is 2773ms of the 3740ms of test time across both files, about 74%, spent waiting on a value chosen for how a button feels. At 50 all 9 cases in the file still pass, and the dismissal case still fails when `if (!active.current) return` is deleted, so nothing is being traded away.

Make the hold injectable so the tests set it near zero and the default stays 800 for the app. A single optional prop threaded through is enough:

```tsx
export function CreateWorkspaceForm({ onClose, onCreated, holdMs = ENTER_HOLD_MS }: { … })
// WorkspaceSwitcher forwards an optional `createHoldMs` it takes as a prop
```

If threading a knob through `WorkspaceSwitcher` reads worse to you than the second it costs, say so and I will not push it again. But the fix is cheap and the number is real.

web/src/app/nav/WorkspaceSwitcher.test.tsx
```diff
@@ -204,3 +236,144 @@ describe("the organization half of the scope switcher", () => {
...
+
+    // The press is acknowledged before the page moves, rather than the create
+    // landing them somewhere else with nothing in between. The button keeps its
+    // name through the wait, so it is still the control it was.
+    expect(submit).toHaveAttribute("data-pending", "true")
+    expect(screen.queryByText("OVERVIEW PAGE")).toBeNull()
+
+    // Both halves of "entering" it: the shell's scope moved to the new
+    // workspace, and the flow ended on the page that scope reads.
+    expect(await screen.findByText("OVERVIEW PAGE")).toBeInTheDocument()
```
khaledosman, line 352: Related to the hold, and the reason widening is not the answer: this wait has less slack than it looks. Nothing configures `asyncUtilTimeout` in `src/tests/setup.ts`, so `findBy*` gets Testing Library's 1000ms default, and the hold eats 800 of it before the create, the context refetch and the router transition get their turn.

I found the boundary by moving `ENTER_HOLD_MS` alone: it passes at 1000 and fails at 1150 with "Unable to find an element with the text: OVERVIEW PAGE". So roughly 200 to 350ms of headroom over the current value. That is inside the range a contended CI runner adds, and `testing.md` rules out the obvious patch (a per-assertion timeout), which is what makes shrinking the hold the fix here rather than a nicety. Same applies to the StrictMode case at line 377.

web/src/app/nav/WorkspaceSwitcher.test.tsx
```diff
@@ -204,3 +236,144 @@ describe("the organization half of the scope switcher", () => {
...
+  })
+
+  afterEach(() => {
+    vi.restoreAllMocks()
+    window.localStorage.clear()
+  })
+
+  // The submit holds for `ENTER_HOLD_MS` before handing the page over, which is
+  // inside `findBy*`'s own ceiling, so the cases below wait on the DOM rather
+  // than on the clock. Fake timers are not an option: `userEvent` drives
```
khaledosman: This comment is right that fake timers are not usable here, but the reason it gives is not the dangerous one, and someone reading it will try the version it does not rule out.

The react-aria deadlock only applies while `userEvent` is driving. In the dismissal case nothing is driving after `{Escape}`, so this looks safe and is not:

```tsx
vi.useFakeTimers()
await vi.advanceTimersByTimeAsync(ENTER_HOLD_MS + 1)
vi.useRealTimers()
```

It passes, in 202ms instead of 990ms, which is exactly why it is tempting. It also passes with `if (!active.current) return` deleted from `WorkspacesPage.tsx`, because the router transition never lands inside the fake-timer window. Adding a real-timer flush afterwards did not recover it either. The current real sleep does fail when that guard is deleted, so what is here now is the one of the two that actually tests something.

Worth saying in the comment: fake timers make this case vacuous, not just deadlocked.

web/src/app/nav/WorkspaceSwitcher.test.tsx
```diff
@@ -204,3 +236,144 @@ describe("the organization half of the scope switcher", () => {
...
+    )
+    // Escape rather than Cancel: Cancel is disabled while the create is in
+    // flight, so dismissal is what is left, and it is the path that bypasses
+    // every button.
+    await user.keyboard("{Escape}")
+    // The only case here that waits on the clock, because what it asserts is an
+    // absence: there is no event to wait on when the completion is correctly
+    // dropped. Bounded by the hold itself rather than by a copy of it.
+    await new Promise((resolve) => {
+      setTimeout(resolve, ENTER_HOLD_MS + 100)
```
khaledosman: Minor, and it comes with the raw sleep: this case logs "An update to ForwardRef(Button) inside a test was not wrapped in act(...)" on every run, because a raw promise is not Testing Library's `asyncWrapper`, so the org-context refetch settles outside `act`. It survives at any hold value, so shrinking the hold does not clear it; wrapping the settle (`await act(async () => { … })`, or a `waitFor` on something the create does produce) does.

khaledosman (changes requested):
The blocker is fixed and covered: I reverted the effect to cleanup-only against this head and `still completes after StrictMode's development remount` fails with "Unable to find an element with the text: OVERVIEW PAGE", so that case earns its keep. The three test-hygiene points are resolved too: no `data-slot` selector, `spinner--current` or `opacity-0` assertion survives, the per-assertion timeout is gone, and both describes are back in `WorkspaceSwitcher.test.tsx`.

One item left, on the part you kept: the hold is still real wall clock in the suite. I measured it rather than asserting it, and there is also a trap worth writing down before someone else steps in it. Details inline.

web/src/features/workspaces/WorkspacesPage.test.tsx
```diff
@@ -145,6 +153,63 @@ describe("WorkspacesPage", () => {
...
+
+  it("keeps a failure the name cannot fix off the name", async () => {
+    // A 500 or a dropped connection is not something the operator can retype
+    // their way out of, so reddening the field would tell them (and assistive
+    // tech) that the name was the problem.
+    mockApi({
+      createRefusal: { status: 500, detail: "Internal server error" },
+    })
+    const user = userEvent.setup()
+    renderPage(<WorkspacesPage />)
```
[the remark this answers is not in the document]
jigjigjig: Skipping, as out of scope rather than as a disagreement with the guideline.

`renderPage` is this file's own helper on main, and all nine cases in it use it, including the six that predate this branch. My two additions follow the file's existing pattern; they do not introduce the gap. Swapping the helper for `renderWithRouter` changes every case in the file, and `WorkspacesPage` renders fine without a router today because nothing in the create or edit form reads the URL, so the change would be a test-only refactor with no behavior behind it. That belongs in its own PR, where a failure in it is attributable.

Worth noting the guideline is already satisfied where it bites: `WorkspaceSwitcher.test.tsx` does use `renderWithRouter`, because the create flow there navigates, and a test without a router could not observe where it landed.

web/src/features/workspaces/WorkspacesPage.tsx
jigjigjig: Done, threaded exactly as you proposed: `holdMs` on `CreateWorkspaceForm` defaulting to `ENTER_HOLD_MS`, forwarded from an optional `createHoldMs` on `WorkspaceSwitcher` that only the tests pass. The app never sets it, so the shipped hold is still 800.

You asked whether the knob reads worse to me than the second it costs. It did before your next comment; it does not now. On its own this was "74% of a 3.7s run", which I would have left alone rather than put a test-only prop in a component's signature. Paired with the 200ms of headroom under `findBy*`'s ceiling, it stops being about speed: the value was one contended runner away from a red build that would have looked like a real failure. That is worth the prop.

The three cases are now 141/185/143ms instead of 891/990/892ms, and I checked the coverage survived rather than assuming it: at 50 the dismissal case still fails with `if (!active.current) return` deleted, and the StrictMode case still fails against a cleanup-only effect.

web/src/app/nav/WorkspaceSwitcher.test.tsx
jigjigjig: This is the comment that changed my mind, so thank you for measuring the boundary rather than asserting it. I had argued the opposite on the other thread: that a single commented sleep for an absence assertion was defensible and the alternatives were worse. That reasoning was about *legibility* and it ignored the ceiling entirely. Passing at 1000 and failing at 1150 is not slack, it is a red build waiting for a slow runner, and `testing.md` ruling out a per-assertion timeout is what leaves shrinking the hold as the only fix.

Fixed for both cases: the hold is 50 in tests, so `findBy*` now has ~950ms of its ceiling for the create, the context refetch and the router transition instead of ~200ms.

web/src/app/nav/WorkspaceSwitcher.test.tsx
jigjigjig: Corrected, and this is the best catch of the review: the version I ruled out was the one I had tried, and the version I left unruled-out is the one that quietly proves nothing. "It deadlocks" reads as "try it somewhere nothing is driving", which is precisely the dismissal case.

The comment now says both, with the vacuity as the reason that matters: advancing the clock past the hold makes that case pass with the guard deleted, because the router transition never lands inside the fake-timer window. I reproduced that before rewriting the comment rather than taking it on trust.

web/src/app/nav/WorkspaceSwitcher.test.tsx
jigjigjig: Fixed. The wait is inside `await act(async () => { … })`, so the context refetch settles under React's watch, and the run is clean: grepping the output for "not wrapped in act" returns 0 where it previously fired on every run of that case.

Kept as a wait inside `act` rather than a `waitFor` on something the create produces, because what the case asserts is still an absence; `act` is what makes the settle legal, not what makes it observable.

web/src/app/nav/WorkspaceSwitcher.test.tsx
jigjigjig: Closing this one out, since the trade-off I left open here has been settled by your later comment about the ceiling.

I was wrong to keep the sleep on legibility grounds. With `findBy*` on its 1000ms default and 800 of that eaten by the hold, this was a latent red build rather than a slow test, which makes the injectable hold the fix and not a nicety. The hold is now 50 in tests, the remaining wait is `TEST_HOLD_MS + 50` inside `act`, and I verified the case still fails when the guard is deleted so the speed cost no coverage.

jigjigjig:
@khaledosman all four applied, replies on each thread, everything resolved and green.

The measurement is what moved me, so it is worth saying plainly: I had argued on the earlier thread that keeping the sleep was defensible and that shrinking the hold was a nicety. The ceiling evidence made that wrong. Passing at 1000 and failing at 1150 is not slack, it is a red build waiting for a slow runner, and with `testing.md` ruling out a per-assertion timeout there was no third option. `holdMs` is threaded exactly as you proposed, defaulting to 800, and only the tests pass it.

| | before | after |
|---|---|---|
| enters the workspace it just created | 891ms | 141ms |
| does not enter a workspace the operator dismissed | 990ms | 185ms |
| still completes after StrictMode's remount | 892ms | 143ms |
| `findBy*` headroom | ~200ms | ~950ms |
| act warnings per run | 1 | 0 |

The fake-timer catch was the sharpest of the four: the version my comment left unruled-out is the one that passes with the guard deleted. I reproduced that before rewriting the comment, and it now leads with the vacuity rather than the deadlock.

Coverage checked rather than assumed at the new value: the dismissal case still fails with `if (!active.current) return` deleted, and the StrictMode case still fails against a cleanup-only effect.

Shipped behavior is untouched by this round: 800ms hold, "Create and open", spinner in the label's place, refusal on the name, fixed modal width.

web/src/app/nav/WorkspaceSwitcher.test.tsx
```diff
@@ -204,3 +242,154 @@ describe("the organization half of the scope switcher", () => {
...
+    // every button.
+    await user.keyboard("{Escape}")
+    // The only case here that waits on the clock, because what it asserts is an
+    // absence: there is no event to wait on when the completion is correctly
+    // dropped. Inside `act` so the context refetch the create triggers settles
+    // under React's watch; a raw promise is not Testing Library's `asyncWrapper`,
+    // and the update lands outside it with a warning on every run.
+    await act(async () => {
+      await new Promise((resolve) => {
+        setTimeout(resolve, TEST_HOLD_MS + 50)
```
khaledosman: The last duration the suite depends on. `+ 50` is a margin, and a margin is a guess about the slowest runner that will ever execute this, which is the same failure mode as the per-assertion timeout we removed.

The patch in the review body replaces it with a gate the test opens: press, `{Escape}`, `beat.release()`, then reopen the switcher and wait for the new workspace to be visible in it. Getting to that visible element is what puts the absence assertions after the completion's turn, so there is no duration left anywhere. 183ms to 106ms, and it still fails when `if (!active.current) return` is deleted.

web/src/features/workspaces/WorkspacesPage.tsx
```diff
@@ -225,29 +237,110 @@ function NarrowedDefaults({
...
  */
-export function CreateWorkspaceForm({ onClose }: { onClose: () => void }) {
+export function CreateWorkspaceForm({
+  onClose,
+  onCreated,
+  holdMs = ENTER_HOLD_MS,
+}: {
+  onClose: () => void
+  /** The acknowledged-press hold, in ms. Shortened by tests; see ENTER_HOLD_MS. */
+  holdMs?: number
```
khaledosman: `holdMs?: number` is what forces the test to sleep: a duration can be shortened but never awaited, so the test can only guess when the window closed. `hold?: () => Promise<void>` with a `defaultHold` of the same 800ms `setTimeout` lets the test own the window instead, and the app is unchanged since it never passes the prop.

Worth knowing why the cheaper alternative does not work: deferring the create response so the test controls that instead moves the dismissal before `onSuccess`, and TanStack Query drops `mutate`-level callbacks once the component has unmounted, so `finish()` never runs and the case passes with the guard deleted. I checked.

khaledosman (changes requested):
Thanks for the measurements and for reproducing the fake-timer vacuity rather than taking my word for it. Four of five are settled. The fifth is the sleep, and I owe you more than an objection to it, so I built and verified the replacement instead: patch at the bottom, applies clean on 3de882e.

**Why the sleep is worth one more round.** `TEST_HOLD_MS + 50` is a margin, and a margin is a guess about the slowest machine that will ever run this. It is the same shape as the `{ timeout: 3000 }` we already took out, just smaller, and it is the last thing in this file whose correctness depends on how fast the runner is.

**What I tried and threw away, so nobody re-tries them.** Two obvious routes both go vacuous, verified by deleting `if (!active.current) return` and watching them still pass:

1. Fake timers, the case you already documented.
2. Deferring the create response so the test owns when it lands. This one is subtler and worth knowing: deferring the create moves the dismissal *before* `onSuccess`, and TanStack Query does not run `mutate`-level callbacks once the component has unmounted. `finish()` never runs at all, guard or no guard.

That is what makes the third route the right one. The guard's window is "`onSuccess` has fired and the beat has not elapsed", so the beat is the thing the test has to own. Passing it as a duration cannot express that; passing it as a **gate** can:

```tsx
hold?: () => Promise<void>   // instead of holdMs?: number
```

The app never passes it and `defaultHold` is the same 800ms `setTimeout` as today. The tests hand over a promise they resolve themselves, so the window is entered and left on purpose, and every assertion is `findBy*` / `getBy*` on visible state.

**Results, same machine, both files:**

| | at head (3de882e) | with this patch |
|---|---|---|
| enters the workspace it just created | 127ms | 98ms |
| does not enter a workspace the operator dismissed | 183ms | 106ms |
| still completes after StrictMode's remount | 143ms | 88ms |
| `setTimeout` / sleeps in the suite | 1 | 0 |
| duration constants the tests depend on | 2 | 0 |

Mutation-checked, not assumed: deleting `if (!active.current) return` fails the dismissal case, and reverting to a cleanup-only effect fails the StrictMode case. 18/18 green across `WorkspaceSwitcher.test.tsx` and `WorkspacesPage.test.tsx`, `tsc --noEmit` clean, `biome check` clean, no `act` warnings.

One assertion changed shape and is worth a look: the last one used to read the trigger's accessible name, which the reopened menu hides from the accessibility tree, so it now reads the `Selected` marker inside the menu and additionally asserts the new workspace is *not* marked. That is a slightly stronger claim than before.

<details><summary>The patch</summary>

```diff
diff --git a/web/src/app/nav/WorkspaceSwitcher.test.tsx b/web/src/app/nav/WorkspaceSwitcher.test.tsx
index 0502a96c..7069e43b 100644
--- a/web/src/app/nav/WorkspaceSwitcher.test.tsx
+++ b/web/src/app/nav/WorkspaceSwitcher.test.tsx
@@ -1,5 +1,5 @@
 import { Outlet } from "@tanstack/react-router"
-import { act, screen, within } from "@testing-library/react"
+import { screen, within } from "@testing-library/react"
 import userEvent from "@testing-library/user-event"
 import { StrictMode } from "react"
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest"
@@ -19,13 +19,6 @@ import { renderWithRouter } from "@/tests/router"
 
 const SECOND_ORGANIZATION_ID = "99999999-9999-9999-9999-999999999999"
 const CREATED_WORKSPACE_ID = "77777777-7777-7777-7777-777777777777"
-// The create form's hold, shortened for the cases below. Nothing here depends on
-// its real value: what is under test is that the press is acknowledged before the
-// page moves and that a dismissal drops the completion, neither of which is a
-// function of how long the beat lasts. At its shipped 800ms these three cases
-// spend most of their runtime waiting, and they clear `findBy*`'s 1000ms ceiling
-// by only ~200ms, which a loaded runner can close.
-const TEST_HOLD_MS = 50
 
 /** One membership per organization, the second one being the one to switch to. */
 function twoOrganizations(): CallerOrganizationMembership[] {
@@ -247,11 +240,22 @@ describe("the organization half of the scope switcher", () => {
 // Rendered in the shell slot rather than as the page, and started away from "/",
 // so the navigation has somewhere to land and something to leave: a switcher
 // rendered as the page's own content would unmount the moment it navigated.
-function renderSwitcherOnAPage(options: { strict?: boolean } = {}) {
+/** A hold the test opens, so the beat is a gate rather than a duration. */
+function pendingHold() {
+  let release = () => {}
+  const gate = new Promise<void>((resolve) => {
+    release = resolve
+  })
+  return { hold: () => gate, release: () => release() }
+}
+
+function renderSwitcherOnAPage(
+  options: { strict?: boolean; hold?: () => Promise<void> } = {},
+) {
   const shell = (
     <Provider>
       <SelectedWorkspaceProvider>
-        <WorkspaceSwitcher collapsed={false} createHoldMs={TEST_HOLD_MS} />
+        <WorkspaceSwitcher collapsed={false} createHold={options.hold} />
         <Outlet />
       </SelectedWorkspaceProvider>
     </Provider>
@@ -318,8 +322,9 @@ describe("the workspace half of the scope switcher", () => {
 
   it("enters the workspace it just created", async () => {
     mockApi({ context: startedInAWorkspace })
+    const beat = pendingHold()
     const user = userEvent.setup()
-    await renderSwitcherOnAPage()
+    await renderSwitcherOnAPage({ hold: beat.hold })
 
     const form = await fillCreateForm(user)
     // The label names the navigation, which is the only warning the operator
@@ -329,10 +334,12 @@ describe("the workspace half of the scope switcher", () => {
 
     // The press is acknowledged before the page moves, rather than the create
     // landing them somewhere else with nothing in between. The button keeps its
-    // name through the wait, so it is still the control it was.
+    // name through the beat, so it is still the control it was.
     expect(submit).toHaveAttribute("data-pending", "true")
     expect(screen.queryByText("OVERVIEW PAGE")).toBeNull()
 
+    beat.release()
+
     // Both halves of "entering" it: the shell's scope moved to the new
     // workspace, and the flow ended on the page that scope reads.
     expect(await screen.findByText("OVERVIEW PAGE")).toBeInTheDocument()
@@ -344,9 +351,13 @@ describe("the workspace half of the scope switcher", () => {
   })
 
   it("does not enter a workspace the operator dismissed the form over", async () => {
+    // The beat is a gate this test opens, so the window the guard exists for is
+    // entered and left on purpose rather than by sleeping long enough to have
+    // been inside it. Nothing here waits on a duration.
     mockApi({ context: startedInAWorkspace })
+    const beat = pendingHold()
     const user = userEvent.setup()
-    await renderSwitcherOnAPage()
+    await renderSwitcherOnAPage({ hold: beat.hold })
 
     const form = await fillCreateForm(user)
     await user.click(
@@ -356,24 +367,28 @@ describe("the workspace half of the scope switcher", () => {
     // flight, so dismissal is what is left, and it is the path that bypasses
     // every button.
     await user.keyboard("{Escape}")
-    // The only case here that waits on the clock, because what it asserts is an
-    // absence: there is no event to wait on when the completion is correctly
-    // dropped. Inside `act` so the context refetch the create triggers settles
-    // under React's watch; a raw promise is not Testing Library's `asyncWrapper`,
-    // and the update lands outside it with a warning on every run.
-    await act(async () => {
-      await new Promise((resolve) => {
-        setTimeout(resolve, TEST_HOLD_MS + 50)
-      })
+    beat.release()
+
+    // Reopening the switcher is a real interaction, and the workspace listed in
+    // it is the visible result of the create. Getting there is what puts the
+    // assertions below after the completion has had its turn.
+    await user.click(screen.getByLabelText(/^Switch workspace/))
+    const menu = await screen.findByRole("dialog", {
+      name: "Switch workspace or organization",
     })
+    const staging = await within(menu).findByRole("button", { name: /Staging/ })
+    expect(staging).toBeVisible()
 
-    // The workspace was created and the switcher will list it; what must not
+    // The workspace was created and the switcher offers it; what must not
     // happen is being taken there after saying not to.
     expect(screen.queryByText("OVERVIEW PAGE")).toBeNull()
     expect(screen.getByText("USAGE PAGE")).toBeInTheDocument()
+    // Read from the menu rather than the trigger, which the open menu hides
+    // from the accessibility tree: the scope never moved.
     expect(
-      screen.getByRole("button", { name: /^Switch workspace/ }),
-    ).toHaveAccessibleName(/currently Default Workspace/)
+      within(menu).getByRole("button", { name: /Default Workspace/ }),
+    ).toHaveTextContent("Selected")
+    expect(staging).not.toHaveTextContent("Selected")
   })
 
   it("still completes after StrictMode's development remount", async () => {
@@ -382,13 +397,15 @@ describe("the workspace half of the scope switcher", () => {
     // would then hang with the modal open and the button spinning, in
     // development only, where `main.tsx` wraps the app in StrictMode.
     mockApi({ context: startedInAWorkspace })
+    const beat = pendingHold()
     const user = userEvent.setup()
-    await renderSwitcherOnAPage({ strict: true })
+    await renderSwitcherOnAPage({ strict: true, hold: beat.hold })
 
     const form = await fillCreateForm(user)
     await user.click(
       within(form).getByRole("button", { name: /Create and open/ }),
     )
+    beat.release()
 
     expect(await screen.findByText("OVERVIEW PAGE")).toBeInTheDocument()
   })
diff --git a/web/src/app/nav/WorkspaceSwitcher.tsx b/web/src/app/nav/WorkspaceSwitcher.tsx
index 57aaf92a..35bbc664 100644
--- a/web/src/app/nav/WorkspaceSwitcher.tsx
+++ b/web/src/app/nav/WorkspaceSwitcher.tsx
@@ -61,14 +61,14 @@ function PlusMark() {
 // workspace row.
 export function WorkspaceSwitcher({
   collapsed,
-  createHoldMs,
+  createHold,
 }: {
   collapsed: boolean
   // Forwarded to the create form's own hold, and only ever set by tests: at its
   // real value the three cases over this flow spend most of their run waiting,
   // and they have less headroom under `findBy*`'s ceiling than is comfortable on
   // a loaded runner. The app never passes it, so the default stands.
-  createHoldMs?: number
+  createHold?: () => Promise<void>
 }) {
   const { memberships, selected, select, isLoading } = useSelectedWorkspace()
   const context = useOrganizationContext()
@@ -348,7 +348,7 @@ export function WorkspaceSwitcher({
                   select(workspace.id)
                   void navigate({ to: "/" })
                 }}
-                holdMs={createHoldMs}
+                hold={createHold}
               />
             </Modal.Dialog>
           </Modal.Container>
diff --git a/web/src/features/workspaces/WorkspacesPage.tsx b/web/src/features/workspaces/WorkspacesPage.tsx
index 7a1fc4d5..78c1c909 100644
--- a/web/src/features/workspaces/WorkspacesPage.tsx
+++ b/web/src/features/workspaces/WorkspacesPage.tsx
@@ -229,6 +229,14 @@ function NarrowedDefaults({
 // eat. `holdMs` is how a test shortens it.
 const ENTER_HOLD_MS = 800
 
+// The shipped beat, as a gate. A caller that wants to own when it opens (the
+// tests do, so the dismissal window is entered and left on purpose rather than
+// slept through) passes its own.
+const defaultHold = () =>
+  new Promise<void>((resolve) => {
+    setTimeout(resolve, ENTER_HOLD_MS)
+  })
+
 /**
  * The one form that creates a workspace, wherever it is offered from.
  *
@@ -240,11 +248,16 @@ const ENTER_HOLD_MS = 800
 export function CreateWorkspaceForm({
   onClose,
   onCreated,
-  holdMs = ENTER_HOLD_MS,
+  hold = defaultHold,
 }: {
   onClose: () => void
-  /** The acknowledged-press hold, in ms. Shortened by tests; see ENTER_HOLD_MS. */
-  holdMs?: number
+  /**
+   * The acknowledged-press beat, as a gate rather than a duration. Defaults to
+   * `ENTER_HOLD_MS` of wall clock; a test supplies one it opens itself, so the
+   * dismissal window is entered and left deterministically instead of by a
+   * sleep long enough to have outlasted it.
+   */
+  hold?: () => Promise<void>
   // Fired once the workspace exists (and its default budget, when one was
   // picked), so the caller that opened the form can follow the new workspace.
   // The list page does not need it: it is already looking at the row that
@@ -362,11 +375,7 @@ export function CreateWorkspaceForm({
               // together: the operator waits a beat, not a beat plus a round
               // trip. A create slower than the hold keeps the spinner until it
               // answers, which is the honest reading of the same indicator.
-              const held = entersWorkspace
-                ? new Promise<void>((resolve) => {
-                    setTimeout(resolve, holdMs)
-                  })
-                : Promise.resolve()
+              const held = entersWorkspace ? hold() : Promise.resolve()
               setHolding(entersWorkspace)
               create.mutate(
                 { name: trimmed, description: description.trim() || null },
```

</details>

If you would rather keep `holdMs` and live with the margin, say so and I will drop it. But the gate is the version where nothing in this file can be made to fail by a slow machine.

jigjigjig:
@khaledosman re-requesting review when you have a moment.

Everything from both of your passes is applied and resolved, and the head is 3de882ea. The short version of what changed since you last looked:

- `holdMs` threaded as you proposed, default 800, only the tests pass it (50). The three cases are 141/185/143ms, and `findBy*` headroom went from ~200ms to ~950ms.
- The fake-timer comment now leads with the vacuity rather than the deadlock, since the variant it previously left unruled-out passes with the guard deleted.
- The remaining absence-wait runs inside `act`; the act warning is gone.
- Earlier pass: the StrictMode effect fix (with a case that fails against a cleanup-only effect), the timeout override removed, the styling assertions dropped, and `WorkspaceOrganizationSwitcher.test.tsx` deleted with both halves back in one file.

Two things I did not change, both flagged rather than assumed:

1. The `renderPage` helper in `WorkspacesPage.test.tsx` still lacks a router (CodeRabbit raised it). It is main's helper and all nine cases in the file use it; swapping it is a file-wide test refactor with no behavior behind it, so I left it for its own PR. Say the word if you would rather it came with this one.
2. This flow diverges from `CreateOrganizationForm` beside it, which does the same create-and-move-into-it with a plain label, no hold, and a banner. The label, the hold and the field-level error here were product decisions from @jigjigjig, so aligning the two is a call for you both rather than something I should quietly do in this PR.

No rush, and thank you for the two rounds; the measured boundary in particular changed the fix rather than just the wording.

khaledosman:
A third pass landed at 13:46, six minutes before your comment, and the head is still 3de882ea, so I think it arrived after you had written this: [review 5019656677](https://github.com/mozilla-ai/otari/pull/780#pullrequestreview-5019656677). It carries a verified patch that removes the last sleep by passing the beat as a gate (`hold?: () => Promise<void>`) instead of a duration, with the two dead ends written up so nobody re-tries them. "No, keeping `holdMs`" is a fine answer to it; leaving it unanswered is the only thing I would rather avoid.

Your two questions:

**1. `renderPage` and the router.** Not a deferral, a false positive for this file. `WorkspacesPage.tsx` imports no `Link`, no `useNavigate`, no `useSearch` and no `useParams`; nothing on that page reads the router, which is why all nine cases pass without one. Adding a provider that nothing consumes would be noise. Leave it, and I would not open the follow-up PR either.

**2. `CreateOrganizationForm`.** Split it, because one of the three differences is correct and the others share a single cause.

The field-level name error is not a divergence to fix. That form's own description says names do not have to be unique, so its endpoint has no 409 to put on a field. A workspace name is unique within the organization, so yours does. Different contracts, correctly different UI.

The label, the hold, and the disabled Cancel are genuine, and all three trace to the hold. Worth naming what it has cost so far: the `active` ref, the StrictMode footgun that took a review round to find, the prop now threaded through `WorkspaceSwitcher`, and every timing-sensitive assertion in the file. Delete the hold and `held` becomes `Promise.resolve()`, `finish()` runs inside `onSuccess`, and TanStack Query already drops `mutate`-level callbacks once the component has unmounted, so the guard becomes dead code and the whole apparatus goes with it, tests included.

The pending state does not depend on the hold either: react-aria sets `data-pending` on press, which is why that assertion passes before anything resolves. What the hold buys is a *minimum duration* for a spinner, on a flow whose real feedback is the page changing.

That is a product call and I am not making it. But it should be made knowing the price, and the sibling form in the same menu already ships the same create-and-move-into-it without paying it. If the beat stays, the gate patch is how the tests stop paying for it.

web/src/features/workspaces/WorkspacesPage.tsx
jigjigjig: Applied as sent, in b579b221. You are right about the shape of it: a duration can be shortened but never awaited, which is why the test was still guessing. The gate removes the guess rather than shrinking it.

Verified here rather than taken from the review, since applying a patch unchecked is how a green suite stops meaning anything: 18/18 across both files, `tsc -b` and `biome check` clean, no act warnings, and both mutation checks still bite (dismissal fails with `if (!active.current) return` deleted, StrictMode fails against a cleanup-only effect). The three cases land at 92ms and under here.

The TanStack Query detail is the part I would not have found: `mutate`-level callbacks being dropped after unmount means the deferred-response seam proves nothing precisely where it looks strongest. That is now in the file's comment rather than only in this thread, since the next person to speed these up will reach for it first.

Two comments went stale against your patch and I updated them: the note on `ENTER_HOLD_MS` still described `holdMs`, and the describe-level note still referred to `TEST_HOLD_MS`. Content unchanged otherwise.

web/src/app/nav/WorkspaceSwitcher.test.tsx
jigjigjig: Gone, along with the last duration in the file: `grep -cE "setTimeout|TEST_HOLD_MS|timeout:"` over `WorkspaceSwitcher.test.tsx` is 0 now.

You were right that the margin was the same failure mode as the timeout override, just quieter. I had defended it as "a bounded wait for an absence assertion", which was true and beside the point: the assertion did not need a duration, it needed the window to be closed on purpose, and reopening the switcher to find the new workspace is a better way of getting there than any margin.

The stronger closing assertion came with it: the menu marks Default Workspace `Selected` and does not mark Staging, rather than reading a trigger name the open menu hides from the accessibility tree.

jigjigjig:
@khaledosman patch applied in b579b221, both threads closed, everything green. Nothing outstanding on my side.

You were right on both counts, and the second one is the one I had argued against twice: a margin is not a bounded wait, it is a smaller version of the timeout override we already removed. The assertion never needed a duration, it needed the window closed on purpose, and reopening the switcher to find the new workspace gets there without one. `grep -cE "setTimeout|TEST_HOLD_MS|timeout:"` over the file is 0.

Verified rather than trusted, since applying a patch unchecked is how a green suite stops meaning anything:

| check | result |
|---|---|
| both files | 18/18 |
| `tsc -b`, `biome check` | clean |
| act warnings | 0 |
| guard deleted → dismissal case | fails, as it must |
| cleanup-only effect → StrictMode case | fails, as it must |
| durations left in the file | 0 |

Two comments had gone stale against the patch and I updated them (the `ENTER_HOLD_MS` note still described `holdMs`; the describe-level note still referred to `TEST_HOLD_MS`), and the TanStack Query finding is now in the file rather than only in the thread, since the deferred-response seam is the first thing the next person will reach for and it proves nothing.

Shipped behavior is unchanged across all three rounds: 800ms beat, "Create and open", spinner in the label's place, refusal on the Name field, fixed modal width. `defaultHold` is the same `setTimeout(ENTER_HOLD_MS)` the app has had since the first commit.

The two things I still have not changed are both on the record above rather than dropped: `renderPage` lacking a router in `WorkspacesPage.test.tsx` (main's helper, all nine cases, a file-wide refactor with no behavior behind it) and the divergence from `CreateOrganizationForm`. Happy to fold either in if you would rather they came with this.

web/src/app/nav/WorkspaceSwitcher.tsx
```diff
@@ -58,9 +59,20 @@ function PlusMark() {
...
-export function WorkspaceSwitcher({ collapsed }: { collapsed: boolean }) {
+export function WorkspaceSwitcher({
+  collapsed,
+  createHold,
+}: {
+  collapsed: boolean
+  // Forwarded to the create form's own hold, and only ever set by tests: at its
+  // real value the three cases over this flow spend most of their run waiting,
+  // and they have less headroom under `findBy*`'s ceiling than is comfortable on
+  // a loaded runner. The app never passes it, so the default stands.
```
khaledosman: Stale against the change directly below it. This paragraph is the `createHoldMs` rationale: runtime and `findBy*` headroom were why a *duration* had to be shortened. A gate is not about speed, it is about the test owning when the beat ends, which is what makes the dismissal case deterministic rather than merely faster.

Suggest something like:

```
// Forwarded to the create form's own hold, and only ever set by tests: supplying
// the beat as a gate is what lets a test enter and leave the dismissal window on
// purpose, instead of sleeping past a duration it cannot await. The app never
// passes it, so the default stands.
```

You went through the other two stale ones in this commit; this is the one the prop rename left behind.

khaledosman (approved):
Verified against b579b221 rather than read: 18/18 across both files, `tsc --noEmit` and `biome check` clean, no `act` warnings, and `grep -cE "setTimeout|TEST_HOLD_MS|timeout:"` over the test file is 0. The three cases run 98/109/93ms.

Mutation-checked three ways, including one you did not claim:

| mutation | result |
|---|---|
| `if (!active.current) return` deleted | dismissal case fails |
| effect reverted to cleanup-only | StrictMode case fails |
| `onCreated?.(workspace)` commented out | "enters the workspace" case fails |

Also confirmed `Field.tsx` is no longer in this PR because `FieldError` landed on main; the `isInvalid` / `errorMessage` call sites here resolve against it.

One stale comment, inline. Everything else is settled, and the two you left open are settled correctly: `renderPage` needs no router (nothing on that page reads one) and the `CreateOrganizationForm` divergence is a product call rather than a defect.

web/src/app/nav/WorkspaceSwitcher.tsx
jigjigjig: Fixed in a1df5b6c, with your wording. You are right that it was the wrong rationale rather than just an out-of-date one: runtime and headroom were the argument for shortening a number, and a gate is not a faster duration, it is the test owning when the beat ends. Leaving the old paragraph there would have taught the next reader that the prop exists for speed.

That is the third stale comment this rename produced, all three of them mine, which is a fair signal that a prop with a test-only reason for existing needs its reason kept in one place rather than restated at every hop.

On your two answers, both noted with thanks:

**`renderPage`** stays as it is, and no follow-up PR. Your check is stronger than my "no behavior behind it": `WorkspacesPage` imports no `Link`, `useNavigate`, `useSearch` or `useParams`, so a router provider there would be a provider nothing consumes.

**`CreateOrganizationForm`** split as you framed it. The field-level error being a contract difference rather than a divergence is the part I had not reasoned through: workspace names are unique within the organization so there is a 409 to put on the field, and that form's own copy says organization names need not be unique, so there is nothing to put. Correctly different, not inconsistent.

The label, the hold and the disabled Cancel are @jigjigjig's call, and I have put your price list in front of them rather than deciding it here. For the record it is accurate: the `active` ref, the StrictMode footgun, the threaded prop and every timing-sensitive assertion in the file all descend from the beat, and `data-pending` is set on press by react-aria rather than by the hold, so the acknowledgement is not what the beat buys.

PR #793, opened 2026-08-25
feat(dashboard): give the shell a seam an overlay mounts a post-sign-in step in

### LINKED ISSUE #789: Give the shell a seam an overlay mounts a post-sign-in step in
`AppShell` mounts `EntitlementResolver` above the chrome and the `<Outlet>`, and
there is no second place like it. An overlay that needs to put something in front
of the whole app once a session exists, and before the dashboard behind it, has
nowhere to do it. The six modules `overlaySeams.test.ts` guards are four nav
contributions, that resolver, and telemetry: the nav four contribute inside the
chrome, and the other two are composition with no surface of their own.

mozilla-ai/otari-ai#1822 is the caller. A hosted signup asks a few profile and
use-case questions once, after the first sign-in of a new account, and maps the
answers onto its CRM. The overlay owns the questions, the answers, the storage
and the vendor mapping. What it cannot do is be shown.

**Not a change to the setup guide.** `SetupGuideCard`'s docstring argues why this
repository shows a panel where the platform shows a blocking sheet, and that
argument stands for a self-hosted operator. This seam leaves it alone: the base
default renders its children unchanged, so an OSS deployment sees the Overview
panel and nothing else, and a sheet exists only in a build that has a signup to
serve.

**The ask:** a seventh seam, `app/overlayPostSignInGate.tsx`, whose base export
renders `{children}` unchanged, mounted by `AppShell` between
`EntitlementResolver` and `AppShellChrome`:

```tsx
export function AppShell() {
  return (
    <EntitlementResolver>
      <PostSignInGate>
        <AppShellChrome />
      </PostSignInGate>
    </EntitlementResolver>
  )
}
```

Added to `SEAM_MODULES`, and reached by its `@/app/overlayPostSignInGate`
specifier.

That position is the contract, and all three boundaries are load-bearing. Inside
the resolver, so a replacement may gate on a capability. Behind the auth gate
`App.tsx` puts the router behind, so there is a session to resolve a user from.
Above the chrome, so a replacement that renders a step replaces the rails with
it, rather than covering a sidebar the person cannot use.

A replacement owns the rest: its own query for whether this user has answered,
its own step, its own record of the answer. This gateway stores no onboarding
columns (`models/tenancy.py`: "purely hosted CRM and onboarding columns ... are
simply not part of the reconciled schema") and should not start.

**One thing it must not do is hold its children back while it decides.** A
replacement that renders nothing until its query settles blanks the dashboard on
every load, which is the disappointment `overlayEntitlementResolver`'s docstring
already warns about for the entitlement axis. Render children while resolving,
and the step once the answer calls for one.

**Alternative shape,** if a whole-app gate reads as too blunt: a slot inside
`SetupComplete`, so the questions ride the activation payoff instead of the
sign-in. Narrower and non-blocking, and second choice here because it can only
reach a user who got to that card in a session that saw the offer, which a signup
funnel cannot rely on.

**Not what this needs:** `GrowthSignalPort.record_onboarding_completed`. The
answers reach the overlay's own contributed route, so the port stays uncalled for
the reason its docstring already gives. Worth filing separately: nothing in this
tree resolves `GrowthSignalPort` at any lifecycle moment (`api/deps.py:518`, zero
call sites), so signup, activation, profile edits and deletion notify nothing in
any build.

_Note: this issue was drafted by Claude Opus 5 via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is the model's._

njbrake:
## Description

The dashboard is built so that a bigger, hosted version of it can slot its own
pieces in without editing files this repository owns. There were places to add a
sidebar link, or a chip in the top bar, and a couple of hooks with no visible
piece of their own, but no place to put something in front of the whole app right
after somebody signs in. The hosted product wants to ask a
new account a few short questions the first time they arrive, and it had the
questions and somewhere to store the answers, but no way to actually show them.

This adds that one place. In this repository it does nothing at all: it hands
whatever is inside it straight through, so a self-hosted install sees the same
dashboard it saw yesterday, with no extra step and nothing asking its operator
anything. In the hosted build, it is where the questions get shown, covering the
app until they are answered instead of appearing on top of a sidebar nobody can
use yet.

Two rules come with it. Whatever gets slotted in has to keep the dashboard on
screen while it works out whether it needs to ask anything: going blank while it
checks would mean everyone stares at an empty page on every load, including the
people who answered months ago. And while its questions are up, the rest of the
app is not running, so anything it wants to record has to say who is answering,
because the piece that normally does that is part of what it replaced.

Mechanically: a new seam module `web/src/app/overlayPostSignInGate.tsx` whose base
export renders `{children}` unchanged, mounted by `AppShell` between
`EntitlementResolver` and `AppShellChrome`, added to `SEAM_MODULES` and reached by
its `@/app/overlayPostSignInGate` specifier. All three boundaries of that position
are the contract: inside the resolver, so a replacement may gate on a capability;
behind the auth gate `App.tsx` puts the router behind, so there is a session to
resolve a user from; above the chrome, so a step replaces the rails rather than
covering them.

That last boundary has a consequence the seam docstring now records:
`TelemetryIdentity`, `UpdatePrompt` and `ConnectionStatus` live inside
`AppShellChrome`, so a step rendered instead of the children unmounts all three.
The identity is the one that costs something, since a signup step's events are the
acquisition funnel the telemetry seam exists to carry. Hoisting `TelemetryIdentity`
above the seam is deliberately not the fix: it reads `GET /v1/organizations/me`,
and the account answering a signup's first questions may hold no membership for it
to report, so a replacement that records during its step names the actor itself.

The setup guide is deliberately untouched. `SetupGuideCard`'s argument for a panel
where the platform shows a blocking sheet stands for a self-hosted operator, and a
sheet exists only in a build that has a signup to serve. This gateway also stores
nothing for the seam, per `models/tenancy.py` on the hosted CRM and onboarding
columns, and `GrowthSignalPort.record_onboarding_completed` is not what is being
wired: the answers reach the overlay's own contributed route.

Not addressed here, since the issue says to file it separately: nothing in this
tree resolves `GrowthSignalPort` at any lifecycle moment.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes #789

## Checklist

Dashboard-only change, so the checks that apply are `pnpm --dir web run lint`,
`pnpm --dir web run typecheck` and `pnpm --dir web test` (101 files, 1465 tests),
all passing. No API contract change, so no spec to regenerate. The new tests were
confirmed to fail without the `AppShell` mount.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

Implemented and tested by Claude Opus 5 through back-and-forth with @njbrake. The
reasoning and decisions are his.

## Summary

- Added a post-sign-in overlay seam for hosted onboarding flows.
- Mounted the seam between entitlement checks and the dashboard shell.
- Kept the default dashboard unchanged for self-hosted deployments.
- Added documentation, seam registration, and coverage for passthrough and replacement behavior.

This allows hosted builds to show onboarding steps without changing the dashboard or setup guide.


web/src/app/overlayPostSignInGate.tsx
```diff
@@ -0,0 +1,73 @@
...
+ * The base default renders its children unchanged, and that is the whole of it.
+ * An OSS deployment sees the dashboard it saw before, with no step in front of
+ * it and nothing asking its operator anything.
+ *
+ * One export, and its shape is the contract: a component named `PostSignInGate`
+ * taking `{ children }`.
+ *
+ * **The position is the rest of the contract, and all three boundaries are
+ * load-bearing.** `AppShell` mounts it inside `EntitlementResolver`, so a
+ * replacement may gate its step on a capability; behind the auth gate `App.tsx`
```
khaledosman, line 26: The contract grants a capability gate here but never says what to do while the axis is still resolving. In a superset build the resolver is a query reporting `isLoading: true` with capabilities unresolved, and `AppShellChrome` is what absorbs that for the route gate (`answerIsStillComing`, `AppShell.tsx:431-441`). Nothing absorbs it above this seam.

So a replacement reading `useEntitlement(...)` on first render gets `entitled: false` and, having also been told not to hold its children back, renders the dashboard and then swaps a blocking step in when the query settles, pulling the app out from under someone already using it. Or the reverse: step, then no step.

Add the rule the other four paragraphs have: read `isLoading` alongside `entitled`, and hold the *decision* until the axis settles rather than the children.

web/src/app/overlayPostSignInGate.tsx
```diff
@@ -0,0 +1,73 @@
...
+ * **It must not hold its children back while it decides.** A replacement that
+ * renders nothing until its query settles blanks the dashboard on every load,
+ * for every user, including the ones who answered months ago. That is the
+ * disappointment `overlayEntitlementResolver`'s docstring already warns about on
+ * the entitlement axis, and it lands harder here because this seam sits above
+ * the chrome as well as the routes. Render children while resolving, and the
+ * step only once the answer calls for one.
+ *
+ * **A replacement owns the telemetry identity while its step is up.** The three
+ * chrome-level components that render nothing (`TelemetryIdentity`,
```
khaledosman: "render nothing" is true of one of the three. `UpdatePrompt` renders a pill once the gateway serves a new bundle (`UpdatePrompt.tsx:39-50`) and `ConnectionStatus` a `role="alert"` toast on an unreachable gateway (`ConnectionStatus.tsx:47-55`); only `TelemetryIdentity` renders nothing unconditionally.

That undercuts the conclusion two sentences down. `ConnectionStatus` is the app-level "the gateway is not answering", and a step POSTing its answers is precisely when someone needs it, so the identity is not the only one that costs something.

Correct the claim, and give the replacement the connectivity rule too: its step owns its own submit-failure feedback for as long as the toast is unmounted. `web/AGENTS.md:31` carries the same sentence and needs the same fix.

web/src/app/overlayPostSignInGate.test.tsx
```diff
@@ -0,0 +1,155 @@
...
+vi.mock("@/app/overlayPostSignInGate", () => ({
+  PostSignInGate: ({ children }: { children: ReactNode }) =>
+    replacement.current ? replacement.current(children) : children,
+}))
+
+// The other half of the position under test: a resolver granting a capability
+// this build never grants. `AppShell` mounts the gate inside this, so a
+// replacement that reads the axis gets the resolved answer rather than the
+// context default, which is what lets it gate its step on a capability.
+vi.mock("@/app/overlayEntitlementResolver", async () => {
```
khaledosman, line 44: This mock is file-wide, so the test at `:98` commented "the state every OSS deployment is in" actually runs with `billing` granted. Inert today because no base nav entry gates on a capability, but once one does, the test pinning the OSS baseline is the one asserting against a nav OSS does not render.

Same shape `replacement.current` already uses: hoist a `capabilities` ref defaulting to `[]` and set it only in the capability test at `:121`.

khaledosman (approved):
Verified at head: the two touched suites pass (15 tests), `tsc --noEmit` and `biome check` clean. No generated artifacts owed (no route/schema change, no `schema.ts` or `routeTree.gen.ts` touch, no new page, no dependency).

Three notes inline, all on the docstring contract and the test's baseline.

PR #795, opened 2026-08-25
feat(api): expose managed MCP activity in Messages streams

### LINKED ISSUE #755: Expose managed MCP activity in Messages streams
## Desired behavior

When `/v1/messages` executes an MCP tool inside the gateway, expose truthful tool-start and tool-completion signals to the streaming client without transferring execution ownership. Include enough information for an existing activity UI: stable call ID, tool/server name, input, and completion result or error.

## Why

The Messages MCP loop currently hides gateway-owned `tool_use` and `tool_result` blocks. Downstream clients receive narration and the final answer, but cannot show live execution cards, record an action trail, or distinguish a slow MCP call from a slow model response. The client cannot reconstruct those details after Otari removes them.

## Notes

Keep the single logical Messages stream and internal tool loop. Do not expose credentials or add tool inputs/results to logs. A custom stream event may require matching any-llm/SDK support.

Ref: mozilla-ai/octonous#4916, mozilla-ai/octonous#4914

HareeshBahuleyan:
## Why

Gateway-owned MCP calls were hidden from Messages streaming clients, so activity UIs could not show tool execution or its result. Echoed activity also needed to be removed before budget admission so hidden result content could not cause false budget rejection or estimate-based charging.

## What changed

Messages streams now emit paired server-owned `mcp_tool_use` and `mcp_tool_result` blocks while Otari retains execution ownership. The gateway preserves explicit MCP errors, avoids exposing transport details, strips replayed activity before prompt estimation, and keeps provider-native blocks intact.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes #755

## Checklist

Targeted verification: `make lint`, `make typecheck`, 171 relevant unit and integration tests, and the OSS-edition smoke gate. The streaming response is not represented by the OpenAPI response model, so generated artifacts are unchanged.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:** GPT-5.6 Sol via pi

**Any additional AI details you'd like to share:** The agent implemented the change, added tests and documentation, investigated a budget-estimation review finding, and ran the listed checks.

**NOTE:**

## Summary

- Added gateway-managed MCP activity events to `/v1/messages` streams.
- Included call IDs, tool and server names, inputs, results, and errors.
- Kept MCP execution within the gateway and preserved one logical stream.
- Removed replayed gateway activity before prompt estimation and provider dispatch.
- Protected credentials and transport details.
- Added documentation and tests for streaming, errors, replay handling, and logging.

### DIFF
--- a/docs/tools.md
+++ b/docs/tools.md
@@ -73,6 +73,9 @@ Responses and Messages can expose native server-tool result blocks when their
 wire format expects them. Chat Completions returns only the final assistant
 message. Calls for tools the client must execute are preserved.
 
+For gateway-run MCP activity in Messages streams, see
+[Messages streaming activity](mcp.md#messages-streaming-activity).
+
 ## Code execution
 
 Start the bundled sandbox:
--- a/src/gateway/api/routes/messages.py
+++ b/src/gateway/api/routes/messages.py
@@ -167,26 +167,7 @@ def _is_gateway_minted_mcp_block(block: Any) -> bool:
     return str(activity_id or "").startswith(MCP_ACTIVITY_ID_PREFIX)
 
 
-def _has_gateway_minted_mcp_blocks(messages: Any) -> bool:
-    """Whether an inbound transcript contains Otari-provenance MCP activity."""
-    if not isinstance(messages, list):
-        return False
-    for message in messages:
-        if not isinstance(message, dict):
-            continue
-        content = message.get("content")
-        if not isinstance(content, list):
-            continue
-        if any(_is_gateway_minted_mcp_block(block) for block in content):
-            return True
-    return False
-
-
-def _strip_gateway_minted_blocks(
-    messages: Any,
-    *,
-    strip_web_search: bool = True,
-) -> Any:
+def _strip_gateway_minted_blocks(messages: Any) -> Any:
     """Drop this gateway's own server-tool blocks from inbound ``messages``.
 
     Continuing an Anthropic conversation means echoing the previous assistant turn.
@@ -214,11 +195,9 @@ def _strip_gateway_minted_blocks(
             continue
         # Two passes: identify our web-search results and our provenance-prefixed
         # MCP uses, then drop each complete pair. A provider's pair matches neither.
-        minted_web_ids = (
-            {block.get("tool_use_id") for block in content if _is_gateway_minted_result(block)}
-            if strip_web_search
-            else set()
-        )
+        minted_web_ids = {
+            block.get("tool_use_id") for block in content if _is_gateway_minted_result(block)
+        }
         minted_mcp_ids = {
             block.get("id") if block.get("type") == "mcp_tool_use" else block.get("tool_use_id")
             for block in content
@@ -593,16 +572,12 @@ async def create_message(
     """
     user_from_metadata = request.metadata.get("user_id") if request.metadata else None
 
-    # Remove replayed gateway-owned MCP activity before admission derives prompt
+    # Remove replayed gateway-owned activity before admission derives prompt
     # size. Waiting until request_fields are built below would reserve against
     # result content that never reaches the provider and can falsely reject or
-    # overcharge the request. Web-search stripping remains gated later because
-    # its provenance depends on the resolved interception configuration.
-    if _has_gateway_minted_mcp_blocks(request.messages):
-        request.messages = _strip_gateway_minted_blocks(
-            request.messages,
-            strip_web_search=False,
-        )
+    # overcharge the request. Provenance comes from each block, so this is
+    # independent of whether the current request enables the same tool again.
+    request.messages = _strip_gateway_minted_blocks(request.messages)
 
     async def _normalize(
         user_id: str,
@@ -693,9 +668,6 @@ async def create_message(
     scope_prompt_cache_key(request_fields, ctx)
     if request_fields.get("tools"):
         request_fields["tools"] = openai_to_anthropic_tools(request_fields["tools"])
-    inbound_messages = request_fields.get("messages")
-    if inbound_messages and tool_ctx.intercepts_web_search:
-        request_fields["messages"] = _strip_gateway_minted_blocks(inbound_messages)
     if tool_ctx.use_sandbox:
         # ``container`` addresses Anthropic's own code-execution container, and
         # the gateway sandbox owns execution for this request, so the provider
--- a/src/gateway/api/routes/messages.py
+++ b/src/gateway/api/routes/messages.py
@@ -153,8 +153,22 @@ def _is_gateway_minted_result(block: Any) -> bool:
     return all(isinstance(hit, dict) and not hit.get("encrypted_content") for hit in hits)
 
 
+def _is_gateway_minted_mcp_block(block: Any) -> bool:
+    """Whether ``block`` carries this gateway's reserved MCP activity prefix."""
+    if not isinstance(block, dict):
+        return False
+    block_type = block.get("type")
+    if block_type == "mcp_tool_use":
+        activity_id = block.get("id")
+    elif block_type == "mcp_tool_result":
+        activity_id = block.get("tool_use_id")
+    else:
+        return False
+    return str(activity_id or "").startswith(MCP_ACTIVITY_ID_PREFIX)
+
+
 def _has_gateway_minted_mcp_blocks(messages: Any) -> bool:
-    """Whether an inbound transcript contains an Otari-provenance MCP use block."""
+    """Whether an inbound transcript contains Otari-provenance MCP activity."""
     if not isinstance(messages, list):
         return False
     for message in messages:
@@ -163,12 +177,7 @@ def _has_gateway_minted_mcp_blocks(messages: Any) -> bool:
         content = message.get("content")
         if not isinstance(content, list):
             continue
-        if any(
-            isinstance(block, dict)
-            and block.get("type") == "mcp_tool_use"
-            and str(block.get("id") or "").startswith(MCP_ACTIVITY_ID_PREFIX)
-            for block in content
-        ):
+        if any(_is_gateway_minted_mcp_block(block) for block in content):
             return True
     return False
 
@@ -211,11 +220,9 @@ def _strip_gateway_minted_blocks(
             else set()
         )
         minted_mcp_ids = {
-            block.get("id")
+            block.get("id") if block.get("type") == "mcp_tool_use" else block.get("tool_use_id")
             for block in content
-            if isinstance(block, dict)
-            and block.get("type") == "mcp_tool_use"
-            and str(block.get("id") or "").startswith(MCP_ACTIVITY_ID_PREFIX)
+            if _is_gateway_minted_mcp_block(block)
         }
         kept_blocks = [block for block in content if not _is_minted_pair_member(block, minted_web_ids, minted_mcp_ids)]
         if len(kept_blocks) == len(content):
@@ -687,11 +694,7 @@ async def create_message(
     if request_fields.get("tools"):
         request_fields["tools"] = openai_to_anthropic_tools(request_fields["tools"])
     inbound_messages = request_fields.get("messages")
-    if inbound_messages and (
-        tool_ctx.intercepts_web_search
-        or tool_ctx.mcp_server_configs
-        or _has_gateway_minted_mcp_blocks(inbound_messages)
-    ):
+    if inbound_messages and tool_ctx.intercepts_web_search:
         request_fields["messages"] = _strip_gateway_minted_blocks(inbound_messages)
     if tool_ctx.use_sandbox:
         # ``container`` addresses Anthropic's own code-execution container, and
--- a/src/gateway/api/routes/messages.py
+++ b/src/gateway/api/routes/messages.py
@@ -72,6 +72,14 @@ from gateway.types.attempt import Attempt
 router = APIRouter(prefix="/v1", tags=["messages"])
 
 
+def _merge_anthropic_betas(body_betas: list[str] | None, raw_request: Request) -> list[str] | None:
+    """Combine legacy body betas with Anthropic's standard beta header."""
+    betas = list(body_betas or [])
+    for header_value in raw_request.headers.getlist("anthropic-beta"):
+        betas.extend(beta.strip() for beta in header_value.split(",") if beta.strip())
+    return list(dict.fromkeys(betas)) or None
+
+
 class MessagesRequest(derive_request_base(MessagesParams)):  # type: ignore[misc]
     """Anthropic Messages API-compatible request.
 
@@ -571,6 +579,9 @@ async def create_message(
     applies up to the pre-lock-in point, same as chat).
     """
     user_from_metadata = request.metadata.get("user_id") if request.metadata else None
+    merged_betas = _merge_anthropic_betas(request.betas, raw_request)
+    if merged_betas is not None:
+        request.betas = merged_betas
 
     # Remove replayed gateway-owned activity before admission derives prompt
     # size. Waiting until request_fields are built below would reserve against
--- a/src/gateway/api/routes/messages.py
+++ b/src/gateway/api/routes/messages.py
@@ -62,6 +62,7 @@ from gateway.services.mcp_loop import ToolBackend
 from gateway.services.mcp_loop_messages import (
     MAX_TOOL_ITERATIONS_CAP,
     MCP_ACTIVITY_ID_PREFIX,
+    MCP_CLIENT_BETA,
     anthropic_tool_loop,
     anthropic_tool_loop_stream,
 )
@@ -78,6 +79,21 @@ def _merge_anthropic_betas(body_betas: list[str] | None, raw_request: Request) -
     for header_value in raw_request.headers.getlist("anthropic-beta"):
         betas.extend(beta.strip() for beta in header_value.split(",") if beta.strip())
     return list(dict.fromkeys(betas)) or None
+
+
+def _split_mcp_client_beta(kwargs: dict[str, Any]) -> tuple[dict[str, Any], bool]:
+    """Consume the client capability without forwarding it to the model provider."""
+    betas = kwargs.get("betas")
+    if not isinstance(betas, list) or MCP_CLIENT_BETA not in betas:
+        return kwargs, False
+
+    provider_kwargs = {**kwargs}
+    provider_betas = [beta for beta in betas if beta != MCP_CLIENT_BETA]
+    if provider_betas:
+        provider_kwargs["betas"] = provider_betas
+    else:
+        provider_kwargs.pop("betas")
+    return provider_kwargs, True
 
 
 class MessagesRequest(derive_request_base(MessagesParams)):  # type: ignore[misc]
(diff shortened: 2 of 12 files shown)


src/gateway/services/mcp_loop_messages.py
```diff
@@ -331,16 +309,118 @@ def __init__(self) -> None:
...
+    *,
+    native_blocks: list[Any] | None = None,
+) -> AsyncGenerator[MessageStreamEvent, None]:
+    """Execute owned calls and yield live MCP start/completion representations."""
+    for spec in state.owned_specs:
+        name = str(spec["name"])
+        parsed_input = _parsed_stream_input(state, spec)
+        server_name = _mcp_server_name(pool, name)
+        activity_id: str | None = None
+        if server_name is not None:
```
khaledosman: MCP activity is emitted to every streaming caller, with no equivalent of `emit_native_web_search`.

`mcp_tool_use` / `mcp_tool_result` are beta-only blocks: they are absent from the non-beta `ContentBlock` union in `anthropic/types/raw_content_block_start_event.py`, so a stock (non-beta) Anthropic SDK client mis-parses them. Driving the installed anthropic 0.88.0's own `construct_type` + `accumulate_event` over this stream shape:

```
block: ParsedTextBlock text            'hi'
block: ParsedTextBlock mcp_tool_use    None
block: ParsedTextBlock mcp_tool_result None
```

The accumulated message ends up carrying two `TextBlock`s with `text=None`. That is the class of SDK-accumulator breakage the index renumbering in this module exists to prevent, and it contradicts the rule the file already states for web search at line 78: native vocabulary goes out only for a caller that declared it. Opting into `mcp_servers` / `mcp_server_ids` is an Otari extension, not a declaration of Anthropic's beta block vocabulary.

Gate emission on the caller having declared it, e.g. `"mcp-client-2025-04-04" in (request.betas or [])` (`betas` is already on `MessagesParams`), plumbed through the strategy the same way `emit_native_web_search` is.

src/gateway/api/routes/messages.py
```diff
@@ -197,13 +228,22 @@ def _strip_gateway_minted_blocks(messages: Any) -> Any:
...
     if _is_gateway_minted_result(block):
-        return True
-    return block.get("type") == "server_tool_use" and block.get("id") in minted_ids
+        return block.get("tool_use_id") in minted_web_ids
+    block_type = block.get("type")
+    if block_type == "server_tool_use":
+        return block.get("id") in minted_web_ids
+    if block_type == "mcp_tool_use":
+        return block.get("id") in minted_mcp_ids
+    return block_type == "mcp_tool_result" and block.get("tool_use_id") in minted_mcp_ids
```
khaledosman, line 257: An orphaned gateway `mcp_tool_result` is forwarded upstream.

`minted_mcp_ids` is built only from `mcp_tool_use` blocks, so a result whose use block is not in the same message survives the strip. `_has_gateway_minted_mcp_blocks` (line 155) does not see it either, so the pre-admission pass skips the transcript entirely. Anthropic rejects a `mcp_tool_result` with no matching use block, which surfaces as a provider 400.

The result's own `tool_use_id` carries `MCP_ACTIVITY_ID_PREFIX`, so match on that prefix directly rather than requiring the pair:

```python
if block_type == "mcp_tool_result":
    tool_use_id = str(block.get("tool_use_id") or "")
    return tool_use_id in minted_mcp_ids or tool_use_id.startswith(MCP_ACTIVITY_ID_PREFIX)
```

and widen `_has_gateway_minted_mcp_blocks` the same way. `test_mcp_result_is_removed_only_with_its_gateway_use` only exercises the complete pair, so this shape is untested.

src/gateway/api/routes/messages.py
```diff
@@ -584,8 +635,13 @@ async def _normalize(
     scope_prompt_cache_key(request_fields, ctx)
     if request_fields.get("tools"):
         request_fields["tools"] = openai_to_anthropic_tools(request_fields["tools"])
-    if tool_ctx.intercepts_web_search and request_fields.get("messages"):
-        request_fields["messages"] = _strip_gateway_minted_blocks(request_fields["messages"])
+    inbound_messages = request_fields.get("messages")
+    if inbound_messages and (
+        tool_ctx.intercepts_web_search
+        or tool_ctx.mcp_server_configs
```
khaledosman: Both added clauses are inert for MCP, and the net effect is unrelated widening.

`request.messages` was already MCP-stripped at line 553 and `request_fields` derives from it, so `_has_gateway_minted_mcp_blocks(inbound_messages)` on the next line is always false. What `tool_ctx.mcp_server_configs` actually does is run *web-search* stripping on requests where `intercepts_web_search` is off, which drops the interception gate that backed the provenance check (the docstring paragraph explaining that gate was removed in this diff).

Revert the condition to `tool_ctx.intercepts_web_search`.

src/gateway/services/mcp_loop_messages.py
```diff
@@ -331,16 +309,118 @@ def __init__(self) -> None:
...
+        if preserve_mcp_error and callable(call_outcome):
+            outcome = await call_outcome(name, arguments)
+            return str(outcome.content), bool(outcome.is_error), False
+        text = await pool.call_tool(name, arguments)
+        return text, is_tool_error(text), False
+    except Exception as exc:  # noqa: BLE001 - recoverable tool failure is model input
+        # Do not put exception details in logs: transports may include URLs or
+        # headers in them. The internal model result keeps the detail for recovery;
+        # the client-facing activity event receives a fixed error below.
+        logger.warning("Gateway tool %s execution failed", name)
```
khaledosman: The exception is dropped from the log entirely, leaving operators with no clue why any gateway tool call failed on the Messages streaming path, web search included and not just MCP.

Log the exception class: it carries no URL, header, or token, and preserves the diagnostic.

```python
logger.warning("Gateway tool %s execution failed: %s", name, type(exc).__name__)
```

src/gateway/services/mcp_loop_messages.py
```diff
@@ -331,16 +309,118 @@ def __init__(self) -> None:
...
+    try:
+        value = json.loads(state.tool_use_json_bufs.get(spec["index"], "") or "{}")
+    except json.JSONDecodeError:
+        return {}
+    return value if isinstance(value, dict) else {}
+
+
+def _mcp_server_name(pool: ToolBackend, tool_name: str) -> str | None:
+    """Return MCP ownership metadata when ``pool`` is an MCP client pool."""
+    resolver = getattr(pool, "server_name_for_tool", None)
```
khaledosman: This probe and the `call_tool_outcome` one at line 349 reach past `ToolBackend`, so mypy checks neither call. That is why `str(outcome.content)` / `bool(outcome.is_error)` need defensive coercion at line 351. A narrow `MCPToolBackend(Protocol)` declaring the two members would let `--strict` check the shape instead of `getattr` + `callable()`.

khaledosman (commented):
Verified on the PR head: `make lint`, `make typecheck` (mypy strict, 522 files) and the 65 relevant unit tests pass, and no OpenAPI-affecting surface changed, so the generated-artifact gate is clean.

Five findings inline. The first two are client-visible regressions.

<sub>Review by [Claude Code](https://claude.com/claude-code) (Opus 5), run by @HareeshBahuleyan's reviewer.</sub>

tests/integration/test_messages_route_dispatch.py
```diff
@@ -802,6 +806,75 @@ async def _stream_iter(*events: MessageStreamEvent) -> AsyncIterator[MessageStre
...
+    ]
+    captured: dict[str, Any] = {}
+
+    async def fake_normalize_messages(input_messages: Any, **kwargs: Any) -> Any:
+        captured["normalized_messages"] = input_messages
+        return input_messages, SimpleNamespace(vision_usage=lambda: None)
+
+    async def fake_resolve_request_context(**kwargs: Any) -> Any:
+        captured.update(kwargs)
+        await kwargs["normalize_messages"]("user", None, "model", None)
```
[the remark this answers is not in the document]
HareeshBahuleyan: Fixed in 21d1c224. The test now passes `workspace_id=None` to the normalization hook, and the focused regression test passes.

docs/tools.md
```diff
@@ -97,16 +97,17 @@ a tool for you":
...
-The gateway's own tool calls are deliberately withheld from streaming clients: a
-client shown a `tool_use` block can never be sent the matching `tool_result`,
-because Otari consumed it, and an SDK accumulating that stream would be left with
-an unanswered call. When a model asks for one of your tools *and* a gateway tool in
-the same message, Otari runs its own, hides it, renumbers what is left so the
-indices stay gapless for SDK accumulators, and hands you only the call you can
-dispatch.
+The gateway's caller-owned `tool_use` blocks are deliberately withheld from
+streaming clients: Otari consumes those calls, so exposing that vocabulary would
+incorrectly hand execution ownership to the client. Messages streams instead
```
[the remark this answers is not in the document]
HareeshBahuleyan: Fixed in 21d1c224. The documentation now identifies the withheld `tool_use` blocks as gateway-owned.

src/gateway/services/mcp_loop_messages.py
```diff
@@ -331,16 +309,118 @@ def __init__(self) -> None:
...
+            outcome = await call_outcome(name, arguments)
+            return str(outcome.content), bool(outcome.is_error), False
+        text = await pool.call_tool(name, arguments)
+        return text, is_tool_error(text), False
+    except Exception as exc:  # noqa: BLE001 - recoverable tool failure is model input
+        # Do not put exception details in logs: transports may include URLs or
+        # headers in them. The internal model result keeps the detail for recovery;
+        # the client-facing activity event receives a fixed error below.
+        logger.warning("Gateway tool %s execution failed", name)
+        return f"[tool error] {exc}", True, True
```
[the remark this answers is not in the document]
HareeshBahuleyan: Fixed in 21d1c224. Streaming tool exceptions now produce a fixed model-facing error instead of forwarding exception details. The regression test captures the next provider call and verifies the sensitive detail is absent.

src/gateway/services/mcp_loop_messages.py
HareeshBahuleyan: Fixed in 21d1c224. Added a runtime-checkable `MCPToolBackend` protocol for `server_name_for_tool` and `call_tool_outcome`, replacing the untyped capability probes and defensive result coercions. Focused strict mypy passes.

src/gateway/api/routes/messages.py
HareeshBahuleyan: Addressed in 3d75b13b. Gateway provenance detection now recognizes the reserved prefix on `mcp_tool_result.tool_use_id`, so orphaned results are stripped before admission. Added focused unit coverage for this shape.

src/gateway/api/routes/messages.py
HareeshBahuleyan: Addressed in 3d75b13b. Restored the late stripping condition to `tool_ctx.intercepts_web_search`; MCP replay stripping remains in the early pre-admission pass. Added an integration test proving `mcp_servers` does not strip web-search replay when interception is off.

src/gateway/services/mcp_loop_messages.py
HareeshBahuleyan: Addressed in 3d75b13b. Tool failure warnings now include only the exception class while omitting the exception message and other details. The focused test verifies `RuntimeError` is logged and sensitive text is absent.

src/gateway/services/mcp_loop_messages.py
HareeshBahuleyan: Addressed in 925087ac. MCP activity blocks are now emitted only when `betas` contains `mcp-client-2025-04-04`. MCP tools still execute for stable streaming clients, but their streams omit the beta-only blocks. Focused unit and integration coverage verifies both modes.

_AI response by `openai-codex/gpt-5.6-sol` via pi._

tests/unit/test_mcp_loop_messages.py
```diff
@@ -769,13 +791,219 @@ async def fake_amessages(**kwargs: Any) -> AsyncIterator[MessageStreamEvent]:
...
+    use = cast(Any, activity_start).content_block
+    assert use.type == "mcp_tool_use"
+    assert use.id.startswith("otari_mcptoolu_")
+    assert use.name == "fetch_url"
+    assert use.server_name == "fixture-server"
+    assert use.input == {"url": "https://example.test"}
+    assert not pool.started.is_set(), "execution must not precede the client-visible start"
+
+    assert (await anext(stream)).type == "content_block_stop"
+    pending_completion = asyncio.create_task(anext(stream))
```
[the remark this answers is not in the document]
HareeshBahuleyan: Verified with Pi: this does not reproduce on the project's supported Python versions. Here, `stream` is an async generator, so `anext(stream)` returns `async_generator_asend`, which implements `collections.abc.Coroutine`; `asyncio.iscoroutine()` returns `True`, and `asyncio.create_task()` accepts it. I verified this on CPython 3.13.11 and 3.14.5, and the focused parametrized test passes. No wrapper is needed, so I am resolving this as a non-issue.

src/gateway/services/mcp_loop_messages.py
```diff
@@ -331,31 +325,143 @@ def __init__(self) -> None:
...
+            outcome = await mcp_backend.call_tool_outcome(name, arguments)
+            return outcome.content, outcome.is_error, False
+        text = await pool.call_tool(name, arguments)
+        return text, is_tool_error(text), False
+    except Exception as exc:  # noqa: BLE001 - recoverable tool failure is model input
+        # Transport exceptions may include URLs, headers, or credentials. Neither
+        # logs, client events, nor the model-facing result may receive that detail.
+        logger.warning("Gateway tool %s execution failed: %s", name, type(exc).__name__)
+        detail = "MCP tool execution failed" if mcp_backend is not None else "Gateway tool execution failed"
+        return f"[tool error] {detail}", True, True
```
khaledosman: The sanitization landed on the streaming Messages path only. The comment two lines up asserts transport exceptions may carry URLs, headers, or credentials, and `test_stream_mcp_exception_emits_error_without_logging_detail` locks that in, but four sibling paths still do `logger.warning(..., exc)` plus `text = f"[tool error] {exc}"`:

- `mcp_loop_messages.py:204` (`_execute_tool_uses`, non-streaming `/v1/messages`)
- `mcp_loop.py:202` (`_execute_mcp_calls`, `/v1/chat/completions`)
- `mcp_loop_responses.py:103` and `:252` (`/v1/responses`)

All four reach `MCPClientPool.call_tool`, so an httpx transport failure against a stored MCP server puts the server URL into the log and into the model's context on the next turn. Either the exception text is safe to expose, in which case this branch's change is unnecessary, or it isn't, in which case a shared helper returning `(text, is_error)` should back all five call sites.

src/gateway/services/mcp_loop_messages.py
```diff
@@ -331,31 +325,143 @@ def __init__(self) -> None:
...
+            mcp_backend=mcp_backend if server_name is not None else None,
+        )
+        if not raised and native_blocks is not None:
+            native_blocks.extend(_native_blocks_for_call(pool, name, parsed_input))
+        results.append({"type": "tool_result", "tool_use_id": spec["id"], "content": text})
+
+        if activity_id is not None:
+            # A transport exception may embed URLs or headers. Expose only the
+            # truthful failure state and the same fixed message sent to the model.
+            activity_content = "[tool error] MCP tool execution failed" if raised else text
```
khaledosman: `text` here is the model-facing rendering, so the client-visible `mcp_tool_result.content` carries the gateway's internal `[tool error] ` sentinel: `call_tool_outcome` (`mcp_client.py:156`) prefixes it whenever `result.isError`, and the block already sets `is_error: true` alongside it.

Anthropic's own `mcp_tool_result` carries the server's raw content plus the flag, so an activity UI rendering this shows a marker no provider emits, and a client that string-matches the content sees a gateway-internal token. `tests/unit/test_mcp_client.py` currently pins the prefixed form.

Carry the unprefixed `flattened` on `MCPToolCallOutcome` alongside `content`, use it for the block, and keep the sentinel on the `tool_result` fed back to the model.

src/gateway/api/routes/messages.py
```diff
@@ -505,6 +552,17 @@ async def create_message(
     """
     user_from_metadata = request.metadata.get("user_id") if request.metadata else None
 
+    # Remove replayed gateway-owned MCP activity before admission derives prompt
+    # size. Waiting until request_fields are built below would reserve against
+    # result content that never reaches the provider and can falsely reject or
+    # overcharge the request. Web-search stripping remains gated later because
```
khaledosman: The stated reason does not hold, and the budget bug this block fixes survives for web search.

`_is_gateway_minted_result` (`messages.py:130`) decides provenance from the block alone, the empty `encrypted_content`, and never consults the resolved interception configuration. `intercepts_web_search` is a "could one of ours be here at all" shortcut on the late call, not a provenance requirement.

So a replayed gateway-minted `web_search_tool_result` still counts toward `estimate_prompt_chars` at line 600 and toward the reservation `resolve_request_context` places, and is only stripped afterwards at line 645, which is exactly the false-rejection / estimate-overcharge this early pass exists to prevent, and the results payload is typically larger than an MCP one.

Either strip both here, or replace this sentence with the real reason (widening the unconditional strip changes behavior for a hand-crafted block when interception is off) and file the remainder.

khaledosman (approved):
Verified locally on the head: `make lint` (architecture + alembic heads + ruff), `make typecheck` (mypy strict, 527 files), and `tests/unit/test_mcp_loop_messages.py tests/unit/test_mcp_client.py tests/unit/test_messages_minted_block_stripping.py tests/unit/test_mcp_loop.py` (92 passed). The earlier Copilot / CodeRabbit / review threads all look addressed. Three findings inline. The branch currently conflicts with `main`.

<sub>Reviewed with [Claude Code](https://claude.com/claude-code).</sub>

src/gateway/services/mcp_loop_messages.py
HareeshBahuleyan: Addressed in 88fac934. MCPClientPool.call_tool() now delegates to call_tool_outcome(), which catches transport failures at the shared MCP boundary, logs only the exception class, and returns generic model-facing text. This covers the Messages, Chat, and Responses call sites listed here.

src/gateway/services/mcp_loop_messages.py
HareeshBahuleyan: Addressed in 88fac934. MCPToolCallOutcome now carries separate model-facing content and client-facing activity_content. The model retains the [tool error] sentinel, while mcp_tool_result receives unprefixed content alongside is_error.

src/gateway/api/routes/messages.py
HareeshBahuleyan: Addressed in b827edf8. Gateway-minted web-search and MCP pairs are now stripped before prompt-size estimation and admission. The early pass uses per-block provenance and no longer depends on whether the current request enables the same interception.

PR #799, opened 2026-08-26
fix(gateway): default an Anthropic client timeout so the non-streaming pre-flight guard does not 502

### LINKED ISSUE #533: [BUG] Non-streaming Anthropic requests with large max_tokens fail with opaque 502 (anthropic SDK timeout guard)

AmirF194:
## Description

A non-streaming Anthropic request with `max_tokens` above roughly 21,333 fails
with a generic `502 {"detail": "LLM provider error"}` and bills no tokens. The
Anthropic SDK builds its client with no explicit `timeout` here, so its own
non-streaming pre-flight guard (`_calculate_nonstreaming_timeout`) fires
whenever `max_tokens` implies more than 10 minutes at its worst-case rate, and
raises a bare, status-less `ValueError` before anything is sent.
`classify_provider_error` returns `None` for a status-less error, so it falls
through to the generic 502 with no hint that a client-side limit was hit.

`build_attempt_client_args` (`src/gateway/api/routes/_platform.py`) and
`get_provider_kwargs` (`src/gateway/services/provider_kwargs.py`) now give an
Anthropic client an explicit `timeout` (the SDK's own default value, so this
only disables the guard's own-default check rather than changing the wire
timeout for any request the guard already let through) whenever the attempt
carries none, including the hybrid-mode path that sources `client_args` from
`extra_params`. An operator-configured timeout is left untouched.
`classify_provider_error` also gets a defense-in-depth branch: an
operator-supplied timeout equal to the SDK's default can still re-arm the
guard, and it maps that status-less failure to a 400 naming the cause instead
of the generic 502.

## PR Type

- [x] Bug Fix

## Relevant issues

Fixes #533

## Checklist

## AI Usage

**AI Model/Tool used:**

An autonomous coding agent (Sonnet 5).

**Any additional AI details you would like to share:**

Root cause, fix and tests were investigated and written by the agent, running
under the account owner's standing authorization for this repo's disclosure
policy.

## Verification

- `tests/unit/test_provider_error_classification.py`,
  `tests/unit/test_provider_instances.py`, and
  `tests/unit/test_hybrid_client_args_routing.py`: fail on `main`
  (`6eade8378`) with the new symbols missing or the pre-flight `ValueError`
  reaching the transport unmodified, pass on this branch; ran both ways in a
  clean `python:3.14` container, matching this repo's own CI matrix.
  `test_anthropic_default_timeout_reaches_the_real_sdk_client[_in_hybrid_mode]`
  drive the real `any-llm`/Anthropic SDK stack, patching only
  `httpx.AsyncClient.send`, so they prove the guard is actually skipped rather
  than only that our own classifier returns the value we expect.
- Full `tests/unit` suite (2540 passed, 8 skipped), `ruff check`,
  `check_architecture.py`, and `mypy --strict` all clean on the branch, same
  container. `tests/integration` was not run (needs a Postgres
  Testcontainer); none of it imports the changed symbols.
- Not verified: behavior against a live Anthropic account. This only
  exercises the SDK's own client construction and pre-flight guard; no
  network credentials were used or needed.

## Summary

Anthropic requests with large `max_tokens` values now avoid premature client-side rejection. Configured timeouts remain unchanged, including in hybrid mode.

Status-less timeout guard failures now return an actionable HTTP 400 response instead of a generic 502 error.

Added regression tests and documentation for timeout handling, hybrid routing, error classification, and SDK behavior.


src/gateway/services/provider_kwargs.py
```diff
@@ -91,6 +91,28 @@
...
+
+    A no-op for every other provider. An operator's own ``timeout`` (set via
+    ``client_args`` on the provider config or a platform attempt's
+    ``extra_params``) is left untouched; this only fills the gap when nothing
+    was configured.
+    """
+    if provider != LLMProvider.ANTHROPIC:
+        return client_args
+    merged = dict(client_args) if client_args else {}
+    merged.setdefault("timeout", ANTHROPIC_DEFAULT_TIMEOUT_SECONDS)
```
khaledosman: A flat float raises the Anthropic client's **connect** timeout from 5s to 600s on every request, including the streaming ones the guard never touched. Verified against the installed SDK:

```
anthropic._constants.DEFAULT_TIMEOUT   -> Timeout(connect=5.0, read=600, write=600, pool=600)
_calculate_nonstreaming_timeout(...)   -> Timeout(connect=5.0, read=600, write=600, pool=600)
AsyncAnthropic(timeout=600.0).timeout  -> Timeout(timeout=600.0)   # connect=600.0
```

The guard's trigger is `self._client.timeout == DEFAULT_TIMEOUT` (`anthropic/resources/messages/messages.py:1061`), so `600.0` disarms it only because it is *unequal* to the default, and the inequality is exactly the connect dimension. There is no client-level value that both disarms the guard and keeps `connect=5.0`: `httpx.Timeout(600.0, connect=5.0)` compares equal to `DEFAULT_TIMEOUT` and re-arms it. This also reaches `alist_models`, where the surrounding code already treats a 600s hang as a hazard (`model_discovery_service.py:317`, "so a black-holed endpoint cannot hang the test button for the SDK's ~600s default").

A **per-request** `timeout` disarms the guard and leaves the client at `DEFAULT_TIMEOUT`. Against the real any-llm/Anthropic stack with `httpx.AsyncClient.send` patched:

```
client_args {"timeout": 600.0}                 -> APIConnectionError   (guard skipped, connect=600)
client_args httpx.Timeout(600.0, connect=5.0)  -> InvalidRequestError  (guard re-armed)
acompletion(..., timeout=600.0)                -> APIConnectionError   (guard skipped, connect=5)
```

Fix: inject `timeout` as a completion kwarg instead of into `client_args`. That moves the injection to where request kwargs are assembled (`default_attempt_kwargs`, `responses.py:381`, and the standalone path) rather than `get_provider_kwargs`, so it is a wider change than this one. If it has to stay client-level, `httpx.Timeout(601.0, connect=5.0)` keeps the 5s connect.

src/gateway/services/provider_kwargs.py
```diff
@@ -91,6 +91,28 @@
 # and no key takes the keyless placeholder, so it never reaches a caller alone.
 _NON_CREDENTIAL_KWARGS = frozenset({"client_args"})
 
+# The anthropic SDK's non-streaming pre-flight guard fires for a large max_tokens
+# only while the client's timeout is untouched (otari#533); this is also the
+# guard's own fallback value, so setting it explicitly disables the guard alone.
```
khaledosman: Nothing ties `600.0` to `anthropic.DEFAULT_TIMEOUT`, so if the SDK bumps its default the constant diverges silently and this comment's "also the guard's own fallback value" goes stale. Add the drift assertion this repo already uses for the same class of coupling (`test_the_uncredentialed_provider_roster_has_not_drifted`):

```python
assert anthropic.DEFAULT_TIMEOUT.read == ANTHROPIC_DEFAULT_TIMEOUT_SECONDS
```

src/gateway/api/routes/_pipeline.py
```diff
@@ -403,6 +421,10 @@ def classify_provider_error(exc: BaseException) -> ProviderErrorMapping | None:
     # whole signal here, and its message names the unsupported feature.
     if _is_unsupported_feature_error(exc):
         return ProviderErrorMapping(status.HTTP_400_BAD_REQUEST, _unsupported_feature_detail(exc))
+    # Defense in depth for otari#533: an operator-supplied timeout equal to the
```
khaledosman: The scenario named here cannot occur. A YAML `client_args: {timeout: 600}` becomes `httpx.Timeout(600.0)`, which is **not** equal to `DEFAULT_TIMEOUT` (`connect=5.0`), so an operator-supplied timeout equal to the SDK's default is not expressible from config and never re-arms the guard.

Keep the branch: it still catches any-llm's `InvalidRequestError` translation and any future Anthropic client built without `with_anthropic_default_timeout`. Reword the justification, or the next reader will trust it.

tests/unit/test_provider_instances.py
```diff
@@ -405,6 +407,47 @@ def test_get_provider_kwargs_strips_provider_type_and_models() -> None:
     assert kwargs == {"api_base": "http://x/v1", "api_key": "k"}
 
 
+# ---------------------------------------------------------------------------
+# get_provider_kwargs / with_anthropic_default_timeout (otari#533)
+# ---------------------------------------------------------------------------
+
+
+def test_get_provider_kwargs_fills_a_default_anthropic_timeout() -> None:
```
khaledosman, line 439: `get_provider_kwargs(GatewayConfig(), ANTHROPIC)` now returns `{"client_args": {"timeout": 600.0}}` where it returned `{}`. It still reports exhausted, because `client_args` is already in `_NON_CREDENTIAL_KWARGS`, but that is load-bearing for the organization-key and fallback-rung paths and nothing here holds it. Worth pinning alongside these:

```python
def test_a_default_anthropic_timeout_does_not_look_like_a_credential() -> None:
    kwargs = get_provider_kwargs(GatewayConfig(), LLMProvider.ANTHROPIC)
    assert credential_ladder_exhausted(LLMProvider.ANTHROPIC, kwargs) is True
```

khaledosman (dismissed):
Notes, not a block. Finding 1 is a behavior change on paths this fix was not aimed at, so worth a look before merge.

Two docs are now out of step with the code (neither file is in the diff, so no inline anchor):

- `docs/hybrid-mode-protocol.md:140` still presents Bedrock as the only provider that gets handling beyond the generic `extra_params` -> `client_args` nesting. Anthropic is now a second one. The `_platform.py` docstring was updated for this; the doc was not, and per the root `AGENTS.md` the stale copy is the one someone believes.
- `docs/configuration.md:397` lists `client_args` as "Extra client options: `custom_headers`, `timeout` (optional)" with no note that an Anthropic instance now gets a `timeout` filled in when none is configured.

Lint, `mypy --strict`, and the three touched test files (139 passed) are clean locally. No generated-artifact obligation: no route or schema changed.

<sub>Review written by Claude Code (Opus 5), posted by @KhaledOsm.</sub>

src/gateway/services/provider_kwargs.py
AmirF194: You are right on every part of this, and I reproduced it before changing anything. In a Docker container with anthropic 0.88.0 installed, AsyncAnthropic(timeout=600.0)._client.timeout is Timeout(timeout=600.0), and its connect attribute reads 600.0, not 5.0. anthropic._constants.DEFAULT_TIMEOUT is Timeout(connect=5.0, read=600, write=600, pool=600), and the guard in messages.py checks self._client.timeout == DEFAULT_TIMEOUT exactly as you quoted. I also patched httpx.AsyncClient.send and ran your three cases through the real SDK: a flat 600.0 skips the guard and reaches the transport layer with connect=600, httpx.Timeout(600.0, connect=5.0) compares equal to DEFAULT_TIMEOUT and puts the guard back, and httpx.Timeout(601.0, connect=5.0) skips the guard and reaches the transport with connect still at 5.0.

I went with your client-level alternative rather than moving this to a completion kwarg. with_anthropic_default_timeout now injects httpx.Timeout(601.0, connect=5.0) instead of a flat float. That function already backs both call sites, get_provider_kwargs and build_attempt_client_args, so the platform-attempt path gets the fix too without touching responses.py or the completion-kwarg assembly. A new test in test_hybrid_client_args_routing.py drives an attempt through the real any-llm/anthropic stack and captures the constructed httpx.AsyncClient's own timeout to confirm connect lands at 5.0 end to end, not just in our own dict construction.

Pushed at 4bdef8c83f0d2ebc178c6593cdbd7fb102e6a6cf.

src/gateway/services/provider_kwargs.py
AmirF194: Agreed on the drift risk, though the exact assertion changed shape along with the fix. Since with_anthropic_default_timeout now injects httpx.Timeout(601.0, connect=5.0) instead of a flat 600.0 (see the reply on the finding above), the constant that has to track the SDK is the connect value, not the read value. tests/unit/test_provider_instances.py now has test_anthropic_default_timeout_constants_track_the_sdks_own_default, asserting anthropic.DEFAULT_TIMEOUT.connect == ANTHROPIC_DEFAULT_CONNECT_TIMEOUT_SECONDS and that the httpx.Timeout we build stays unequal to anthropic.DEFAULT_TIMEOUT. Either one drifting fails the test.

Pushed at 4bdef8c83f0d2ebc178c6593cdbd7fb102e6a6cf.

src/gateway/api/routes/_pipeline.py
AmirF194: You're right about the mechanism, and I rewrote the comment rather than the branch. An operator's client_args.timeout can only ever arrive as a flat number from config, and a flat number sets every httpx.Timeout dimension including connect, so it can equal DEFAULT_TIMEOUT only by also setting connect to 5.0, at which point read/write/pool are 5.0 too, not 600. Config alone can't reproduce the real default, so the old comment's stated scenario doesn't happen.

The branch itself stays, and the comment now says what it actually catches: a client built without with_anthropic_default_timeout, i.e. some future call path that skips get_provider_kwargs and build_attempt_client_args. That client is left at the SDK's real default and can still trip the guard for a large max_tokens.

Pushed at 4bdef8c83f0d2ebc178c6593cdbd7fb102e6a6cf.

tests/unit/test_provider_instances.py
AmirF194: Added, close to what you proposed. tests/unit/test_provider_instances.py now has test_a_default_anthropic_timeout_does_not_look_like_a_credential, asserting that get_provider_kwargs(GatewayConfig(), LLMProvider.ANTHROPIC) still reports credential_ladder_exhausted() as True. It pins exactly the thing you flagged: a default-filled client_args stays in _NON_CREDENTIAL_KWARGS and never gets read as a credential by itself.

Pushed at 4bdef8c83f0d2ebc178c6593cdbd7fb102e6a6cf.

AmirF194:
Both of these are right, and both are fixed now. docs/hybrid-mode-protocol.md has a new paragraph right after the Bedrock section describing Anthropic's default-timeout fill-in and why it's an httpx.Timeout rather than a flat number. docs/configuration.md's client_args row now has a note that an anthropic instance gets a default timeout filled in when none is configured, so an operator reading that table sees it without having to find the code.

Pushed at 4bdef8c83f0d2ebc178c6593cdbd7fb102e6a6cf.

src/gateway/services/provider_kwargs.py
```diff
@@ -91,6 +92,45 @@
...
+# pool=600), checked in anthropic/resources/messages/messages.py). A flat float
+# disarms the guard but is not a narrow disarm: the anthropic SDK expands a flat
+# float into httpx.Timeout(timeout=<value>), which sets *every* dimension
+# including connect, so client_args={"timeout": 600.0} silently raises the
+# client's connect timeout from 5s to 600s on every Anthropic request, streaming
+# included (otari#799 review). Building the httpx.Timeout explicitly, with
+# read/write/pool one second above the SDK's own default and connect pinned at
+# the SDK's own default, disarms the guard (unequal to DEFAULT_TIMEOUT on
+# read/write/pool) without touching connect at all.
+ANTHROPIC_DEFAULT_TIMEOUT_SECONDS = 601.0
```
khaledosman, line 107: The name reads as "the SDK's default" but this is default + 1, and the PR description states it *is* "the SDK's own default value". Derive it instead of restating it:

```python
httpx.Timeout(anthropic.DEFAULT_TIMEOUT.read + 1, connect=anthropic.DEFAULT_TIMEOUT.connect)
```

Then neither constant can go stale against an SDK bump, and `test_anthropic_default_timeout_constants_track_the_sdks_own_default` stops being load-bearing for correctness.

src/gateway/api/routes/_pipeline.py
```diff
@@ -485,6 +503,18 @@ def classify_provider_error(exc: BaseException) -> ProviderErrorMapping | None:
...
+    # operator-supplied client_args timeout can't reproduce the SDK's own
+    # DEFAULT_TIMEOUT, because config only ever supplies a flat float and a flat
+    # float sets every httpx.Timeout dimension including connect, so it can only
+    # equal DEFAULT_TIMEOUT (connect=5.0, read/write/pool=600) by also setting
+    # connect to 5.0, at which point read/write/pool are 5.0 too, not 600. What
+    # this still catches is a client built without with_anthropic_default_timeout
+    # (any future call path that does not route through get_provider_kwargs or
+    # build_attempt_client_args), which is left at the SDK's real default and can
+    # still trip the guard for a large max_tokens.
+    if _is_anthropic_nonstreaming_timeout_guard(exc):
```
khaledosman, line 516: This branch is unreachable through either client-construction path once the timeout is injected, and it couples the classifier to an English sentence in a vendored SDK. Meanwhile any-llm 1.25's `_translating_nonstreaming_guard` already converts the guard into an `InvalidRequestError` whose message names the actual `max_tokens` value, which this replaces with a fixed string.

Classifying a status-less `any_llm.exceptions.InvalidRequestError` as a 400 through `_redacted_caller_fault_detail` is reachable today, generalizes to every other status-less `InvalidRequestError`, keeps the more specific message, and sits on the same rung as the existing `_is_unsupported_feature_error` branch.

docs/configuration.md
```diff
@@ -396,6 +396,11 @@ Each provider entry in the config file supports:
 | `api_base` | Custom API base URL (optional) |
 | `client_args` | Extra client options: `custom_headers`, `timeout` (optional) |
 
+For an `anthropic` instance, a `timeout` is filled in automatically when
+`client_args` does not set one, so the anthropic SDK's non-streaming
+pre-flight guard does not turn a large `max_tokens` into an opaque 502
```
khaledosman, line 401: The guard has a second raise condition the injected timeout also disarms: `MODEL_NONSTREAMING_TOKENS` (the opus-4 family, capped at 8192). Those requests now round-trip to Anthropic for the rejection instead of failing pre-flight. Worth naming here, since this paragraph and the hybrid-protocol note both describe only the 10-minute case.

khaledosman (approved):
Checked the mechanism against the installed SDK and it holds up: the guard is `not stream and not is_given(timeout) and self._client.timeout == DEFAULT_TIMEOUT`, and its non-raising branch returns `Timeout(600, connect=5.0)`, identical to what the client would have used anyway. So disarming it by inequality costs nothing for a small `max_tokens`, and building the `httpx.Timeout` explicitly to keep `connect` at 5s rather than passing a flat float is the right call. `credential_ladder_exhausted` staying pinned is the load-bearing test here.

Three non-blocking points inline.

---

njbrake:
Hey @AmirF194 , thanks for your patience on this one. Tbh we're in the middle of a huge codebase refactor to open source almost all of our otari.ai platform, so we didn't have time to dig into this ticket as much as we needed to.

The code here is well implemented and really appreciate the iterations based on comments. However, I think at the core, the problem that is being fixed here mostly belongs in any-llm and not otari. And, since we originally created the issue this PR solves, we have since added a toggle into any-llm so that otari only needs a small change to benefit from the any-llm feature.

Sorry again about all the back and forth, and hopefully you're not discouraged from continuing to make contributions! Things will soon be back to normal in the next few weeks once we finish this big migration. 🙏 


Here's Claudes full feedback for context: 

#533 was written against any-llm-sdk==1.24.0, and its suggested fix named client_args in build_attempt_client_args. That was right at 1.24.0. mozilla-ai/any-llm#1255 then landed a first-class timeout on acompletion for this same SDK guard, reported upstream as mozilla-ai/any-llm#1251, and otari picked it up when #777 moved the floor to 1.25.0 on 2026-08-25. This PR opened a day later, so it implements an instruction whose premise had already changed. The work here follows the issue faithfully; the issue was the stale part.

Passing any-llm's timeout lifts the guard, and on current main it is documented on messages and responses as well, with TIMEOUT_SUPPORT declaring it per provider, native on anthropic and inherited by azureanthropic and vertexaianthropic. Going through client_args instead couples otari to the dimension structure of anthropic.DEFAULT_TIMEOUT and to a hand-maintained set of anthropic-SDK providers that BaseAnthropicProvider already groups. That set is short by two here: azureanthropic and vertexaianthropic reach the same messages.create and still fail on the request this fixes for anthropic.

Upstream declined to default the timeout as a deliberate choice, making it settable and the error legible instead, since a silent 600s default turns a request that needs streaming into a ten-minute hang.

Two follow-ups:

- Map a status-less any_llm.exceptions.InvalidRequestError to a 400 in classify_provider_error, keyed on the exception type rather than the SDK's message text. This is the part of this PR worth keeping, and it clears the opaque 502 for every provider that raises it. No dependency change.
- Bump any-llm-sdk to >=1.27.1 and pass timeout on the completion, messages and responses calls, gated on TIMEOUT_SUPPORT. azure, huggingface, lmstudio, ollama, sagemaker, watsonx and xai declare unsupported and raise UnsupportedParameterError once a timeout is set, which this classifier already surfaces as a client-facing 400.

The root-cause analysis, the max_tokens > 128000/6 boundary and the MODEL_NONSTREAMING_TOKENS second trigger are accurate and carry into both.

AmirF194:
Appreciate the detailed writeup, that pins down exactly why this landed a day too late: any-llm's own timeout support closes the guard at a lower layer than client_args ever could, and it covers azureanthropic/vertexaianthropic too, which my approach didn't. Good call closing in favor of that.

The status-less InvalidRequestError to 400 mapping in classify_provider_error still stands on its own regardless of which timeout path ships, so #989 makes sense as separate work.

PR #800, opened 2026-08-26
fix(dashboard): anchor the filter dropdowns under their trigger

jigjigjig:
## Description

Every filter on every page was a token-styled native `<select>`, so the browser drew the option menu wherever the platform wanted. On macOS that is *over* the control, with the selected row on top of the button that opened it, which reads as two overlapping dropdowns.

The report also asked for more of HeroUI, and the two are the same fix: `FilterSelect` (`web/src/shared/components/ui.tsx`) is the one shared primitive behind all of these, so moving it to HeroUI's `Select` (a button plus a listbox popover, placement `bottom`) corrects every filter bar in the dashboard at once. Nothing else about the control changes: same props, same tokens, now taken from HeroUI rather than from a class string of ours.

`KeysPage`'s "Mismatched `user` field" was the last hand-rolled native `<select>` on a filter-shaped control, so it takes the primitive too; leaving it would have shipped a fix for "the dropdowns cover their trigger" with one that still did. It keeps its own `<label>`, because the text carries markup (`user` as `<code>`) that the `label` prop cannot.

**What that costs**

- **`children` leaves the API.** Raw `<option>`/`<optgroup>` belongs to the native element. The four Activity-page callers that used it pass `options` instead, which is the array they were already assembling by hand.
- **Keys carry a prefix.** react-aria reads an empty key as "nothing selected", and `""` is a real filter value here ("All", "Any price"), so `OPTION_KEY_PREFIX` wraps every key and one helper strips it back off.
- **Tests drive the control, not the element.** `user.selectOptions` has no `<select>` to act on. `pickOption` / `selectTrigger` (`web/src/tests/select.ts`, 20 test files) open the popover and click the option by its visible label, and `selectTrigger(label)` with `toHaveTextContent` replaces `toHaveValue`. `e2e/helpers.ts` carries the Playwright counterpart.

Docs follow the code: `components.md` and `testing.md` both described `FilterSelect` as a native `<select>`, and three in-code comments explained themselves by what a native select does with an unmatched value (shows its first option, where HeroUI shows nothing) — the code they justify is still right, the stated reason was not.

**Verified in a browser**, because two questions only a browser answers:

- **Mobile.** The stated reason for keeping this native was that a filter bar wants the platform picker on a phone. At 390x664 the popover opens below the trigger (trigger ends y=306, popover starts y=314), sits fully inside the viewport, and scrolls to its last option, which selects. That objection does not hold, and `components.md` now says so.
- **Scroll containers.** Walked the popover subtree with computed styles: exactly one scrollable element, the `ListBox` at `max-h-72`. HeroUI's popover is `overflow-y: auto` but is not scrolling, since its child fits. No stacked scrollers, no double scrollbar.
- **Width.** The `KeysPage` select lost the `w-full` its native predecessor carried; as a stretch item in a `flex flex-col` the trigger measures 100% of its column at both 1440px and 390px, so no width prop is needed.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [x] Documentation
- [ ] Infrastructure / CI

## Relevant issues

No filed issue: reported directly with a screenshot of the Models page capability filter, showing the option menu drawn over the button that opened it.

## Checklist

Notes on the four above, since this change is entirely in `web/`:

- **Tests.** No Python test changed; the dashboard's counterpart is Vitest, where 20 colocated test files moved onto the new harness, plus three Playwright specs. There is no new test *file* beyond the harness itself: the control is exercised by the ~30 call-site assertions that now drive its popover.
- **DoD checks.** Left unchecked rather than claimed: `make lint` (architecture + alembic heads + ruff) and `make typecheck` (mypy, 527 files) both pass on this tree, but `make test` did not run to completion. Not much is riding on it — this branch changes no Python, so the backend suite has nothing of this change to exercise, and CI runs it regardless. The three that do cover this change are the dashboard's, in the table below.
- **OpenAPI.** Left unchecked because it does not apply, not because it was skipped: no route, schema, or docstring changed, so neither generated artifact moves. `web/src/client/schema.ts` and `routeTree.gen.ts` are likewise untouched.

| Check | Result |
|---|---|
| `pnpm --dir web run lint` | clean |
| `pnpm --dir web run typecheck` | clean |
| `pnpm --dir web test` | 1495 / 1496 — see caveat |
| `pnpm --dir web run e2e` | **not run** — see caveat |
| `make lint` | clean |
| `make typecheck` | clean (527 files) |
| `make test` | **not run** — see note |

**Two caveats a reviewer should not read as regressions.**

`pnpm test` is not green locally, and neither failure is this change. This machine runs Node 26, where jsdom's `window.localStorage` comes back undefined (`localStorage is not available because --localstorage-file was not provided`); on a clean `main` checkout that alone fails ~196 tests. Running with `--localstorage-file` to see past it leaves one wandering failure, because that single file is shared across eight vitest workers: three consecutive runs failed a different test each time (Login's master-key exchange, AppShell's per-rail history, OAuthCallbackPage's StrictMode double-mount, SelectedWorkspace's remount), all four read localStorage, none touches `FilterSelect`, and each passes in isolation with a storage file of its own. CI is on Node 22 and sees none of this.

`pnpm run e2e` never ran locally: port 8000 was held by another gateway. **Now retired as a caveat** — CI has run the `parity` project green on this branch, including the scoped `pickOption(page, /Role for /, "Admin", member)` form, which is the one line of the Playwright helper no run had reached until the review fixes landed.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

Claude Opus 5, in Claude Code.

Written by one agent and reviewed by a second before opening, which caught two real defects now folded into the second commit: a regex built from an unescaped label, where the shipped `label="Priced?"` compiles to "Price" plus an optional "d" and would have matched the wrong control instead of failing; and a missing null guard on the selection key, where react-aria clearing a selection would have pushed `String(null).slice(2)` — the string `"ll"` — into a filter and, on the Activity page, into the URL. The same review caught that the review gateway's throwaway config and SQLite file had been committed to the repo root, which is why the second commit is amended rather than followed by a deletion.

The human ran the dashboard by hand against these changes before this PR was opened.

**NOTE:**

## Summary

- Replaced native dashboard filter controls with HeroUI `Select` components.
- Anchored option popovers below their triggers for improved scrolling and mobile behavior.
- Updated `FilterSelect` to use required `options`, preserve empty filter values, and display unmatched active values.
- Applied `FilterSelect` to the `KeysPage` mismatched-user filter.
- Updated unit and E2E tests to use shared selection helpers and visible trigger text.
- Added documentation for the new control and testing patterns.

## Verification

- Lint and typecheck passed.
- All 1497 unit tests passed on a clean run.
- An unrelated localStorage-related failure remains reproducible with the documented Node 26/jsdom setup.
- E2E tests passed in CI twice, including the previously unexecuted Playwright helper path.


web/src/shared/components/ui.tsx
```diff
@@ -779,63 +780,67 @@ export function Badge({
...
-      className={FILTER_SELECT_CLASS}
+      isDisabled={disabled}
+      selectedKey={optionKey(value)}
+      // A null key is react-aria clearing the selection, which no filter here
+      // asks for: reporting it would push the strip of a non-string ("ll") into
+      // the filter, and on a URL-backed page into the query string with it.
+      onSelectionChange={(key) => {
+        if (key != null) onChange(optionValue(String(key)))
+      }}
+      className={label ? "flex flex-col gap-1" : undefined}
```
khaledosman: This restates HeroUI's own base. `@heroui/styles` ships:

```css
.select { gap: calc(var(--spacing) * 1); flex-direction: column; display: flex }
```

Same axis, same 0.25rem gap, and `selectVariants` puts `base: "select"` on `Select.Root` unconditionally, so the labelled case already stacked with this exact gap before the class was added. components.md is explicit that `className` is for layout and position, not for restyling a component HeroUI already styles.

Drop the prop entirely. The labelled and unlabelled cases then differ only in whether a `<Label>` renders, which is the actual difference.

Related, since the description reasons about it: `fullWidth` is a real `Select.Root` prop (`select--full-width` / `select__trigger--full-width`). Your measurement that the `KeysPage` trigger already spans its column is right, because `.select` is a flex item stretching in a `flex-col` parent and the trigger stretches inside it, so nothing is broken. But if a future call site does need it, that prop is the sanctioned knob rather than a `w-full` at the call site.

web/src/tests/select.ts
```diff
@@ -0,0 +1,32 @@
+import { screen } from "@testing-library/react"
+import type { UserEvent } from "@testing-library/user-event"
+
+// A FilterSelect is a HeroUI Select, so its control is a button that opens a
+// listbox popover rather than a native <select>: `user.selectOptions` cannot
+// drive it, and react-aria names the button with the current value *and* the
+// label ("Error Status"), which is why these match the end of the name.
+const named = (label: string) => ({
+  name: (name: string) => name === label || name.endsWith(` ${label}`),
```
khaledosman, line 15: This matcher picks the wrong control as soon as one `FilterSelect` label is a suffix of another's on the same page, and it fails by silently driving a different select rather than by not matching. `endsWith(" Status")` is satisfied by a trigger named `... Error Status` just as well as by the one named `... Status`.

No live collision today: I went through every `FilterSelect` call site and no two labels on one page are suffix-related. But twenty test files route through this helper, and `e2e/helpers.ts`'s `pickOption` already grew a `scope` parameter for exactly this hazard, with a comment explaining why. The Vitest half should take the same `scope` (a container element to search `within`), so a page that develops the collision has an escape that does not require rewriting the helper.

The regex escaping in the Playwright twin is a genuinely good catch, incidentally: `label="Priced?"` compiling to `Price` plus an optional `d` would have matched the wrong control rather than failing.

.gitignore
```diff
@@ -22,6 +22,11 @@ otari.db
 otari-dev.db
 smoke.db
 otari-smoke.db
+# A gateway booted by hand to look at a change: its config carries a throwaway
+# master key and provider keys, and its database is a binary blob, so both are
+# named here rather than trusted to `git add -p`.
+review-otari.db
```
khaledosman: `review-otari.db` / `review-otari.yml` are not names any project convention produces; they are one workflow's local scratch files. `.git/info/exclude` is where those belong, which keeps the shared ignore file describing the project rather than a tool's habits, and it is also where `web/src/**/*.stories.tsx` already lives per #801's description.

Also unrelated to this PR's subject, so it is a line a reader of `git log` for the dropdown fix will not expect to find.

khaledosman (approved):
Checked the migration against the library rather than the description: `Select.Root` / `Trigger` / `Value` / `Indicator` / `Popover` are the v3 shapes, `isDisabled` is the right prop, and `.select__popover` already carries `min-width: var(--trigger-width)` and `overflow-y: auto`, so `max-h-72` on the inner `ListBox` is the only height constraint and the single scroller you measured is the expected one.

One thing that looks like an a11y bug and is not, so nobody re-flags it: passing `id` to `Select.Trigger` does override react-aria's generated trigger id, but RAC's `Button` merges through `mergeProps`, which routes `id` through `mergeIds` and remaps the `aria-labelledby` self-reference `useSelect` builds. The `KeysPage` control keeps its name.

Three inline, none blocking.

---

web/src/shared/components/ui.tsx
jigjigjig: Confirmed and dropped in 4167b062. `@heroui/styles/dist/components/select.css` opens with

```css
.select { @apply flex flex-col gap-1; }
```

and `select.styles.js:9` has `base: "select"` in `selectVariants`, so `Select.Root` was already carrying that exact declaration before the prop was added. The prop was a verbatim restatement of the base, which is what components.md rules out. The labelled and unlabelled cases now differ only in whether a `<Label>` renders.

Noted on `fullWidth` too: agreed that `select--full-width` / `select__trigger--full-width` is the sanctioned knob if a call site ever needs it, rather than a `w-full` at the call site. Nothing needs it today.

web/src/tests/select.ts
jigjigjig: Fixed in 4167b062: both `pickOption` and `selectTrigger` now take an optional `scope` and search `within` it, mirroring the Playwright twin. The comment says why the suffix match is what makes the parameter necessary, and that the failure mode is driving the wrong control silently rather than not matching.

One asymmetry worth naming, since it is not obvious from the signature: `scope` narrows the *trigger* lookup only. The popover is portaled to the body, so the option is still found on `screen` even when the trigger came from inside `scope` — same as the e2e helper, which clicks the option on `page` rather than in its scope.

Agreed there is no live collision; I re-checked every call site as well.

.gitignore
jigjigjig: Agreed, reverted in 4167b062. `.gitignore` is back to what is on `main`, and the two names now live in `.git/info/exclude` alongside `web/src/**/*.stories.tsx`, with a comment saying what booted them. You are right that the neighbors there (`otari.db`, `otari-dev.db`, `smoke.db`, `otari-smoke.db`) are names the CLI defaults and the smoke script actually produce, which `review-otari.*` is not, and right that it was a line in `git log` for a dropdown fix that nobody would expect to find.

njbrake (changes requested):
Two things to fix, both small. The design is right: this extends the shared primitive and puts it on the same footing as `FilterMultiComboBox` sitting beside it, rather than adding a second control.

**The e2e failure is a miss in this PR's own migration, not flake.** `e2e/parity.tenancy.spec.ts:76` and `:79` still drive the role control as a native select. The trigger's accessible name is value first, label second (`"Owner Role for Operator (…)"`), so `/^Role for /` cannot match, and `toHaveValue` means nothing on a button. The three sibling assertions at 106-113 were converted; these two were not. The disabled trigger carries a real `disabled` attribute rather than `aria-disabled`, so `toBeDisabled` is fine once the locator is:

```ts
const role = operator.getByRole("button", { name: /Role for / })
await expect(role).toHaveText(/Owner/)
await expect(role).toBeDisabled()
```

Line 85's `getByLabel(/^Status for /)).toHaveCount(0)` has the same cause and is now vacuous: no `FilterSelect` can match that locator any more, so the guard would no longer catch a regression. It wants `getByRole("button", { name: /Status for / })` too.

**An unmatched value renders `"Select an item"`, not empty.** react-aria falls back to its own placeholder. Three comments say otherwise (`ModelsPage.tsx:1670`, `WorkspaceCodeExecutionPolicyCard.tsx:140`, and its test's mirror at `:315`). One live path reaches it: Activity's `status`, `priced` and `tool` read straight off the URL against static lists, so `/activity?status=bogus` puts library boilerplate in the filter bar where the native select showed its first option. `source` is safe, since line 1366 prepends an unmatched value to its options. Either thread a `placeholder` through `FilterSelect`, or give those three what `source` already has.

**One residual risk worth knowing.** `parity.tenancy.spec.ts` is `mode: "serial"`, so the failure aborted 4 later tests, including line 110's `pickOption(page, /Role for /, "Admin", member)`. That scoped RegExp form is the most novel line in the Playwright helper and CI has never run it. The activity and usage `pickOption` calls did run and passed, so the plain form is proven.

@khaledosman's three inline points all hold. The `className` is redundant: `@heroui/styles` ships `.select{gap:calc(var(--spacing) * 1);flex-direction:column;display:flex}`. On the `scope` suggestion, I enumerated every `FilterSelect` label; no live collision, but two near-misses survive only on letter case (`Mode` vs `Guardrail mode`, `Budget` vs `Default member budget`).

Not conflicting with today's four merges, incidentally. Merged onto current `main` locally: clean, typecheck and lint clean, 1522/1522 vitest green on Node 24.

jigjigjig:
Both fixed, plus the residual risk you named. @njbrake

**1. The e2e miss** — fixed in 4167b062, with your locator verbatim, and line 85 with it. Your diagnosis of the cause is exactly right: the name is value-first, so `/^Role for /` cannot match, and CI is now green on the `parity` project including the four tests the serial abort had been hiding.

That covers your residual-risk note too: line 110's `pickOption(page, /Role for /, "Admin", member)` — the scoped RegExp form — ran for the first time in that green run. It was the one line of the Playwright helper CI had never reached, and it passes.

**2. `"Select an item"`, not empty** — you are right, and I checked rather than reasoned about it: rendering `FilterSelect` with an unmatched value puts `Select an item` in the trigger's `textContent`. My own earlier review of this PR asserted the opposite, so the "reads as empty" rewording was a correction in the wrong direction.

Fixed in 8088679f, and in the primitive rather than at the three call sites: `FilterSelect` carries an unmatched value as its own option, so the bar names what is actually filtering. Reasons for choosing that over threading a `placeholder`:

- It is the fallback the page already makes for itself twice — `labelFrom` is `?? value` (`ActivityPage.tsx:1537`), and `sourceOptions` prepends a drilled-into source with no rows in the window. A `placeholder` would have replaced react-aria's boilerplate with ours while still not naming the applied filter.
- It covers the next list read off a URL without that call site having to know about the trap.
- `source` needing no change falls out of it rather than being a special case.

I checked the empty-label edge before doing it in the primitive: every option list today carries its own default (`{ value: "", label: "Any price" }`, `{ value: "all", label: "All providers" }`, and so on), so reaching it needs a call site that omits its own default, which is a bug there.

There is a test for it, on the live path you identified (`/activity?status=bogus`), and it fails with `Select an item` when the fallback is removed — I verified that rather than trusting it.

The three comments are corrected to what happens now. The code-execution card's test title went with them: "names a withdrawn pin instead of showing it as the deployment default" described the native select's first-option fallback, which is not anything HeroUI does. Both surviving reasons still hold — a bare image string says nothing about being refused, and an invented provider still ends in zero rows — so `withdrawnImage` and the `ModelsPage` reset effect both stay.

**On @khaledosman's three** — all applied in 4167b062, replies on each thread. I confirmed the `className` claim against the installed package: `select.css` opens `.select { @apply flex flex-col gap-1 }` and `select.styles.js:9` has `base: "select"`, so the prop restated the base it already had. Your two case-only near-misses (`Mode` / `Guardrail mode`, `Budget` / `Default member budget`) are the argument for the `scope` parameter landing now rather than when one of them stops differing.

**One caveat now retired.** The PR description says `pnpm run e2e` never ran. It still has not run on my machine (port 8000 is held by another gateway), but CI has now run it green twice on this branch, so the Playwright helper is no longer covered by review alone.

Unit suite is 1497/1497 on a clean run; the wandering single failure the description documents reproduced here too, landing on `Login` this time and passing in isolation, which is the shared `--localstorage-file` and not this branch.

njbrake (approved):
Both findings are fixed. Verified against current `main` rather than only trusting the checks: merged onto `4cd3f5c6d` locally, typecheck and lint clean, 1528/1528 vitest, and CI is green on `47f1015` with e2e included, so the four tenancy tests that the serial abort had never run have now run.

Carrying an unmatched value as its own option in the primitive is the better fix than patching the three call sites, and it matches the fallback `sourceOptions` already made for the Source filter.

One thing I raised and then talked myself out of, recorded so nobody re-finds it: `UsagePage.tsx:1213` filters `source` out of the group options, which looked like it could strand a selected value under the new fallback. It cannot. Line 723 sets `showSource = multiSource || groupBy === "source"`, so the option is present whenever it is the one selected, and the comment above it already says that is why.

PR #801, opened 2026-08-26
feat(dashboard): add a trend chip for period-over-period change

jigjigjig:
## Description

Adds `TrendChip`, a pill that pairs an arrow with a signed percentage, replicating the trend chip drawn on the "Trend chips" design canvas.

It is built on HeroUI v3's own `Chip` rather than a hand-rolled `<span>`. v3 already ships the shape, the four fill treatments (primary/secondary/tertiary/soft) and the three sizes the canvas draws, and `globals.css` already aliases the status bases its CSS reads (`--success`, `--danger`, `--default`) onto our tokens. Nothing in the component names a color, so light and dark both follow the foundation with no work at the call site.

What the wrapper adds is the part a generic chip cannot know:

- **Direction** from the sign of the fraction. The arrow and the sign both carry it, so it survives without hue.
- **Polarity**, the metric's own axis. Spend and error rate improve by falling, so `down-is-good` paints a fall green while the arrow keeps telling the truth about which way it went. It defaults to `neutral`: most numbers on this dashboard are cost and error rate, and a forgotten prop that painted rising spend green would be worse than no color at all.
- **Tabular figures**, so a chip on a live tile does not change width as its digits tick.
- **A flat state** below half of the smallest printable percent, which stops a chip reading `-0.0%` beside no arrow.
- **A screen-reader word for the direction**, since polarity puts good and bad in hue alone: a `-2.1%` that is good and a `-2.1%` that is bad render the same text and the same arrow, and a reader of the text cannot recover the difference.

`trendState` is exported so the mapping from a fraction to a direction and a status color is tested as a function, rather than by asserting HeroUI's class names on the rendered chip.

```tsx
<TrendChip fraction={0.033} polarity="up-is-good" />
<TrendChip fraction={0.12} polarity="down-is-good" />
<TrendChip fraction={0.184} text="+$1,234" />
<TrendChip fraction={-0.059} caption="vs last month" className="ml-auto" />
```

**No call sites yet, on purpose.** `DeltaHint` in `shared/components/ui.tsx` renders the same idea as plain text and has six call sites across `UsagePage` and `OverviewPage`. Migrating them is a behavior change to two pages and belongs in its own PR, so this one lands the primitive and its tests only.

Verified in Storybook across all four variants, three sizes and both themes against the design artboards. The story file is not in the diff because `web/src/**/*.stories.tsx` is excluded in `.git/info/exclude`: the catalog is deliberately local-only.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

None. This came from the design canvas rather than a filed issue.

## Checklist

Dashboard-only change, so the checks that apply are the dashboard's: `pnpm --dir web run lint`, `pnpm --dir web run typecheck` and `pnpm --dir web test` all pass on a clean checkout of this branch. No API contract changed, so the OpenAPI spec and the Postman collection are untouched. No page was added, so no screenshot entry is owed.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

The component, its tests and its Storybook stories were written by the agent from the Paper design canvas, then reviewed against a component-API design rubric, which is where the `text` prop name (it was `value`, which reads as the datum rather than the printed string) and the `className` passthrough came from.

## Summary

- Added the `TrendChip` dashboard component for period-over-period changes.
- Displays signed percentages with directional arrows in a pill.
- Supports polarity-aware colors, captions, custom text, sizes, variants, and accessible labels.
- Treats very small or invalid values as flat.
- Added tests for formatting, states, accessibility, styling, and empty values.

### DIFF
--- a/web/src/shared/components/TrendChip.test.tsx
+++ b/web/src/shared/components/TrendChip.test.tsx
@@ -77,10 +77,22 @@ describe("TrendChip", () => {
     expect(hasArrow(container)).toBe(false)
   })
 
-  it("says the direction in a word, since hue carries good vs bad", () => {
-    render(<TrendChip fraction={-0.021} polarity="down-is-good" />)
-    // A fall that is good and a fall that is bad render the same text and the
-    // same arrow, so the word is all a screen reader has to go on.
+  it("says the judgment, not just the direction, when there is one to make", () => {
+    // The same fall under each polarity: identical text, identical arrow, and
+    // the color is the only visible difference, so the announced phrase is all a
+    // screen reader has to tell the two apart.
+    const good = render(
+      <TrendChip fraction={-0.021} polarity="down-is-good" />,
+    ).container
+    expect(good).toHaveTextContent("down, better")
+    const bad = render(
+      <TrendChip fraction={-0.021} polarity="up-is-good" />,
+    ).container
+    expect(bad).toHaveTextContent("down, worse")
+  })
+
+  it("says the direction alone when the metric has no good direction", () => {
+    render(<TrendChip fraction={-0.021} polarity="neutral" />)
     expect(screen.getByText("down")).toBeInTheDocument()
   })
 
@@ -118,7 +130,7 @@ describe("TrendChip", () => {
     const { container } = render(
       <TrendChip fraction={0.033} className="ml-auto" />,
     )
-    expect(container.querySelector(".chip")?.className).toContain("ml-auto")
+    expect(container.firstElementChild).toHaveClass("ml-auto")
   })
 
   it("hides the arrow from the accessibility tree", () => {
--- a/web/src/shared/components/TrendChip.test.tsx
+++ b/web/src/shared/components/TrendChip.test.tsx
@@ -1,0 +1,131 @@
+import { render, screen } from "@testing-library/react"
+import { describe, expect, it } from "vitest"
+
+import { TrendChip, trendState } from "./TrendChip"
+
+// The arrow is aria-hidden decoration and which hue HeroUI paints is what the
+// stories and the screenshot matrix are for, so the rendered assertions here
+// stay on text, sign, and whether an arrow is there at all. The mapping from a
+// fraction to a direction and a status color is tested as a function instead.
+function hasArrow(container: HTMLElement): boolean {
+  return container.querySelector("svg") !== null
+}
+
+describe("trendState", () => {
+  it("reads the sign as the direction", () => {
+    expect(trendState(0.033, "neutral").direction).toBe("up")
+    expect(trendState(-0.021, "neutral").direction).toBe("down")
+  })
+
+  it("calls a change too small to print flat rather than a fall", () => {
+    // -0.0004 rounds to "0.0%", and a down arrow beside that reads as a bug.
+    expect(trendState(-0.0004, "neutral").direction).toBe("flat")
+    expect(trendState(0, "neutral").direction).toBe("flat")
+    expect(trendState(-0, "neutral").direction).toBe("flat")
+  })
+
+  it("treats NaN as flat rather than falling through to a fall", () => {
+    expect(trendState(Number.NaN, "up-is-good")).toEqual({
+      direction: "flat",
+      color: "default",
+    })
+  })
+
+  it("passes no judgment when the metric has no good direction", () => {
+    expect(trendState(0.12, "neutral").color).toBe("default")
+    expect(trendState(-0.12, "neutral").color).toBe("default")
+  })
+
+  it("colors by the metric's own axis, not the number's", () => {
+    expect(trendState(0.033, "up-is-good").color).toBe("success")
+    expect(trendState(-0.021, "up-is-good").color).toBe("danger")
+    expect(trendState(-0.059, "down-is-good").color).toBe("success")
+    expect(trendState(0.12, "down-is-good").color).toBe("danger")
+  })
+
+  it("leaves a flat change uncolored whatever the polarity", () => {
+    expect(trendState(0, "up-is-good").color).toBe("default")
+    expect(trendState(0, "down-is-good").color).toBe("default")
+  })
+})
+
+describe("TrendChip", () => {
+  it("renders nothing when there is no comparable previous value", () => {
+    expect(
+      render(<TrendChip fraction={null} />).container,
+    ).toBeEmptyDOMElement()
+    expect(
+      render(<TrendChip fraction={undefined} />).container,
+    ).toBeEmptyDOMElement()
+  })
+
+  it("signs a rise and shows an arrow", () => {
+    const { container } = render(<TrendChip fraction={0.033} />)
+    expect(screen.getByText("+3.3%")).toBeInTheDocument()
+    expect(hasArrow(container)).toBe(true)
+  })
+
+  it("keeps the minus sign on a fall", () => {
+    const { container } = render(<TrendChip fraction={-0.021} />)
+    expect(screen.getByText("-2.1%")).toBeInTheDocument()
+    expect(hasArrow(container)).toBe(true)
+  })
+
+  it("drops the arrow and prints a bare zero when nothing moved", () => {
+    const { container } = render(<TrendChip fraction={-0.0004} />)
+    expect(screen.getByText("0.0%")).toBeInTheDocument()
+    expect(hasArrow(container)).toBe(false)
+  })
+
+  it("says the direction in a word, since hue carries good vs bad", () => {
+    render(<TrendChip fraction={-0.021} polarity="down-is-good" />)
+    // A fall that is good and a fall that is bad render the same text and the
+    // same arrow, so the word is all a screen reader has to go on.
+    expect(screen.getByText("down")).toBeInTheDocument()
+  })
+
+  it("names a flat change rather than leaving a bare number", () => {
+    render(<TrendChip fraction={0} />)
+    expect(screen.getByText("no change")).toBeInTheDocument()
+  })
+
+  it("takes an absolute value as its text and still reads the sign", () => {
+    const { container } = render(<TrendChip fraction={0.184} text="+$1,234" />)
+    expect(screen.getByText("+$1,234")).toBeInTheDocument()
+    expect(hasArrow(container)).toBe(true)
+  })
+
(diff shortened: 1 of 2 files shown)


web/src/shared/components/TrendChip.tsx
```diff
@@ -0,0 +1,142 @@
...
+        <Arrow className={`${ARROW_SIZE[size]} shrink-0`} aria-hidden="true" />
+      )}
+      {/* The arrow is decoration a screen reader never sees, and `polarity`
+          puts the good/bad distinction in hue alone, which no reader of the text
+          can recover: a -2.1% that is good and a -2.1% that is bad differ only
+          in green vs red. So direction is also said in a word. It sits outside
+          `Chip.Label` so it does not join the visible label's text, and
+          `sr-only` is rendered (clipped) rather than hidden, so `select-none`
+          keeps it out of a drag selection over the chip, as in ModelsPage. */}
+      <span className="sr-only select-none">{DIRECTION_LABEL[direction]}</span>
```
jigjigjig: The word does not carry the distinction the comment above it names. `DIRECTION_LABEL` says direction, and direction is already in the sign: `-2.1%` is a fall whether or not the arrow renders. What a reader of the text cannot recover is the *judgment*, and that is exactly what the comment says this fixes. Both of these announce `down -2.1%` and differ only in green vs red:

```tsx
<TrendChip fraction={-0.021} polarity="up-is-good" />   // danger
<TrendChip fraction={-0.021} polarity="down-is-good" /> // success
```

So WCAG 1.4.1 is still open and the sr-only span is redundant with the sign. Say the judgment when `polarity` is not `neutral` (`better` / `worse`, keyed off `color` rather than `direction`), leave `neutral` announcing nothing extra, and cover both outcomes for one fraction. CodeRabbit landed on the same finding independently.

Either way the comment needs to stop claiming a fix it does not deliver.

web/src/shared/components/TrendChip.tsx
```diff
@@ -0,0 +1,142 @@
...
+  className?: string
+}) {
+  if (fraction == null) return null
+  const { direction, color } = trendState(fraction, polarity)
+  const Arrow = direction === "up" ? FiArrowUp : FiArrowDown
+  // A flat chip prints a bare zero: the sign of a change too small to render is
+  // noise ("-0.0%"), and it would contradict the missing arrow beside it.
+  const printed =
+    text ??
+    `${direction === "up" ? "+" : ""}${formatPct(direction === "flat" ? 0 : fraction)}`
```
jigjigjig, line 139: `text` and the flat state can contradict each other. `direction` comes from `fraction` while the printed string comes from `text`, so `<TrendChip fraction={0.0002} text="+$1,234" />` renders no arrow, no color, and announces `no change +$1,234`. Either drop the direction word when `text` is supplied, or say on the prop that `fraction` has to agree in magnitude with what `text` prints, not just in sign (the current comment only promises the sign).

web/src/shared/components/TrendChip.test.tsx
```diff
@@ -0,0 +1,131 @@
...
+    )
+  })
+
+  it("passes a call site's layout classes through to the chip", () => {
+    // Position, not restyling: a chip is placed by the row it sits in, and
+    // without this the call site needs a wrapper element to do it.
+    const { container } = render(
+      <TrendChip fraction={0.033} className="ml-auto" />,
+    )
+    expect(container.querySelector(".chip")?.className).toContain("ml-auto")
```
jigjigjig: Class selector, which this file's own header comment (lines 6-9) is written to avoid and which `testing.md` rules out outright: "Not `getByTestId`, and never a class selector: a test that breaks on a restyle teaches everyone to stop trusting the suite." `.chip` is HeroUI's internal slot name, so a HeroUI bump renames it and this test fails on a change that broke nothing.

The assertion does not need it, since the chip is the only root:

```tsx
expect(container.firstElementChild).toHaveClass("ml-auto")
```

jigjigjig (commented):
Review by Claude Opus 5 (1M context) via Claude Code, against `.github/skills/review/SKILL.md`, `.github/skills/frontend-standards/SKILL.md` (plus its `testing.md`), `.github/instructions/frontend-standards.instructions.md` and `web/AGENTS.md`. `COMMENT` rather than `APPROVE` only because GitHub does not accept either verdict from the PR author's own session.

Three inline findings. Two notes that are not about a line in the diff:

**Dead code at merge.** No call sites, and `DeltaHint` (`shared/components/ui.tsx:228`) renders the same idea at 7 sites across `UsagePage` and `OverviewPage`. Landing the primitive alone is a reasonable split, but until the follow-up lands the tree carries two components for one concept in the same directory, which is what `component-architecture.md` is aimed at. File the migration issue and link it here so the duplicate cannot become permanent by default.

**Storybook is the stated evidence and no reviewer can see it.** `web/package.json` tracks `storybook`, `@storybook/react`, `@storybook/react-vite` and `@storybook/addon-docs` pinned at 10.5.10, while `web/.storybook/`, `web/src/**/*.stories.tsx` and `web/storybook-static/` are excluded through `.git/info/exclude`, which is per-clone and untracked. `git ls-files 'web/src/**/*.stories.tsx'` returns zero. So every clone installs the toolchain and only one machine can run it, and "verified in Storybook across all four variants, three sizes and both themes" is unreviewable here. Not this PR's to fix, but worth resolving one way or the other: track the config and the stories, or move the exclusion into a tracked ignore file and drop the four devDependencies.

One thing that is genuinely non-obvious and worth keeping: exporting `trendState` so the fraction-to-color mapping is tested as a function rather than by asserting HeroUI's class names is the right call, and the contrast of the fills this lands on holds in both themes (worst case 5.30:1, `primary` danger in dark, computed through HeroUI's oklab derivations with the alpha fill composited over `--color-surface`).

web/src/shared/components/TrendChip.tsx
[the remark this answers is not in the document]
jigjigjig: Fixed in 11fbcebe. `announce()` now appends the judgment where there is one to make, so the same fall separates:

```
polarity="down-is-good"  ->  "down, better -2.1%"
polarity="up-is-good"    ->  "down, worse -2.1%"
polarity="neutral"       ->  "down -2.1%"   (unchanged, as you asked)
```

Keyed off the resolved `color` rather than off `polarity` directly, so a flat change stays uncolored and announces "no change" under every polarity. Went with "better" / "worse" over "improvement" / "regression": shorter to hear in front of the number, and plainer, which is the copy register the rest of this dashboard uses.

Covered both outcomes for one fraction in a single test, since the point is that the two are indistinguishable without it, plus a `neutral` case to pin that the judgment is omitted rather than defaulted.

jigjigjig: Fixed in 11fbcebe. `announce()` appends "better" / "worse" off the resolved `color`, so `neutral` still says the direction alone and a flat change still says "no change" under every polarity. Comment rewritten to describe what it now does. CodeRabbit filed the same finding, replied there too.

web/src/shared/components/TrendChip.tsx
jigjigjig: Addressed in 11fbcebe, with the doc rather than the code. Making the word conditional on `text` would drop it from exactly the case that needs it most, since `text="+$1,234"` may not carry a sign at all. So the prop now says `fraction` has to describe the same change and not merely agree in sign, and names the failure: one inside the flat threshold announces "no change" whatever string sits beside it.

web/src/shared/components/TrendChip.test.tsx
jigjigjig: Fixed in 11fbcebe: `expect(container.firstElementChild).toHaveClass("ml-auto")`. No selector, and it fails for the right reason if the passthrough ever stops reaching the root.

khaledosman (approved):
Nothing to raise.

Checked the parts a description cannot settle:

- `variant` / `color` / `size` match `chipVariants` exactly (`primary|secondary|soft|tertiary`, `success|danger|default`, `sm|md|lg`), so no prop is silently ignored.
- The theme claim holds through the derivation, not just the alias. `.chip--soft` reads `--{status}-soft` / `--{status}-soft-foreground`, and `@heroui/styles` derives both from `--success` / `--danger` / `--default` by `color-mix`, which `globals.css` aliases onto our tokens. Nothing in the component names a color and nothing reads a HeroUI variable directly.
- Measured the default `soft` pairings rather than assuming them: about 8.8:1 for light success, 7.9:1 for light danger, 7.1:1 for dark success on `--color-surface`. Comfortably AA at 12px, which matters because `#807` drops the weight floor under this.
- `FLAT_THRESHOLD = 0.0005` encodes `formatPct`'s one-decimal contract, and `format.test.ts:59` already pins that precision, so the coupling is guarded.
- `-0`, `NaN`, and the exact-threshold boundary all land where the tests say.

Exporting `trendState` so the fraction-to-status mapping is tested as a function instead of by asserting HeroUI's class names is the right call, and the `sr-only` phrase is a real accessibility gap closed rather than a decorative `aria-label`: polarity does put good and bad in hue alone, and the sign cannot recover it.

---

web/src/shared/components/TrendChip.test.tsx
```diff
@@ -0,0 +1,143 @@
...
+    // The same fall under each polarity: identical text, identical arrow, and
+    // the color is the only visible difference, so the announced phrase is all a
+    // screen reader has to tell the two apart.
+    const good = render(
+      <TrendChip fraction={-0.021} polarity="down-is-good" />,
+    ).container
+    expect(good).toHaveTextContent("down, better")
+    const bad = render(
+      <TrendChip fraction={-0.021} polarity="up-is-good" />,
+    ).container
```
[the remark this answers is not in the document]
jigjigjig: Skipping this one. The router harness exists for components that touch routing, and TrendChip has no router surface: no Link, no useNavigate, no useSearch, no useUrlState. web/AGENTS.md scopes the rule that way ("A component test needs a real router, since the pages keep filters and pagination in the URL"), and seven leaf-component tests in this same directory use plain render for the same reason (BulkActionBar, ConfirmDialog, EntitlementGate, Field, FilterChips, TablePagination, charts). Wrapping this one would add a router to a component that has nothing to do with routing.

Issue #806, opened 2026-08-26
[BUG] An older gateway's bootstrap white-screens the dashboard sign-in page


jigjigjig:
**Title:** `fix(dashboard): guard the bootstrap's oauth_providers so an older gateway does not white-screen the sign-in page`

## What happens

A dashboard built from `main` served by a gateway whose `/v1/bootstrap` omits
`oauth_providers` renders a blank page at every route. No error text, no fallback UI,
nothing in the DOM: `document.getElementById("root").innerHTML.length === 0`. The only
signal is the console:

    TypeError: Cannot read properties of undefined (reading 'filter')
      at renderableOAuthProviders (features/auth/oauthProviders.ts:34)
      at Login (features/auth/Login.tsx:455)

## Cause

`renderableOAuthProviders(configured)` does `configured.filter(isOAuthProvider)` with
`configured: readonly string[]`, and the field is absent rather than `[]`.

TypeScript raises nothing, because the generated client types it as **required**:
`src/client/schema.ts` has `oauth_providers: string[]` with no `?`. That is the generator
trap `web/AGENTS.md` already documents ("a schema field with a **default** is emitted as
required, which is right for a response and wrong for a request body") appearing on the
*response* side, where the field is required in the type and optional on the wire from any
gateway older than the commit that added it.

## Why it takes the whole app down

`grep -rn 'ErrorBoundary|componentDidCatch|errorComponent' web/src` returns **nothing**.
There is no error boundary anywhere in the dashboard, so any throw during render unmounts
to a blank document. The sign-in page is the worst possible place for that, because it is
the only page an operator can reach before anything else.

## Reproduce

Serve the `main` dashboard against a gateway built before `oauth_providers` was added (any
branch cut earlier), open `/`, observe a blank page. Found in exactly this way, running a
`main` dashboard against a gateway from a `fix/*` branch.

## Fix, two parts

1. `renderableOAuthProviders(configured = [])`, or narrow at the bootstrap boundary the way
   `Defaulted<T, K>` in `src/client/index.ts` already does for the cases that bite.
2. **An error boundary around the router**, so the next missing field is a message instead
   of a white screen.

**(2) is the more valuable half.** (1) fixes this instance; (2) fixes the class. A dashboard
with no error boundary will keep converting any render-time type mismatch into a blank page,
and the version skew that caused this one is a normal condition during development.

Issue #805, opened 2026-08-26
[BUG] The dashboard test suite cannot run on Node 26, and nothing pins the version locally


jigjigjig:
**Corrected.** The original report said the suite was red on `main`. That was wrong, and the
correction matters more than the original claim: **CI is green.** The same commit that reports
303 failures locally passes 102 of 103 files in the `dashboard` job. So this is not a broken
suite, it is a suite that cannot run on a Node version nothing stops a contributor from using.

**Scope:** dashboard (`web/`), local development. Repo-level, so the environment fields of the
bug template do not apply.

### What happens

`pnpm --dir web test` on **Node 26** fails 303 tests across 21 files. On **Node 22**, which is
what CI and the Docker web stage use, the same tree passes.

    Node 26 (local, macOS)   Test Files  21 failed | 82 passed (103)
                             Tests      303 failed | 1192 passed (1495)

    Node 22 (CI, ubuntu)     Test Files   1 failed | 102 passed (103)

Whole files fail rather than scattered assertions. Two observed symptoms, offered as leads and
not as a diagnosis:

- `src/app/nav/navigationHistory.test.ts` fails 9 of 9 with
  `TypeError: Cannot read properties of undefined (reading 'clear')` on
  `window.localStorage.clear()`. It is a pure-logic file with no DOM, no router and no
  providers, so whatever takes it down is not component-level. Note jsdom 30 instantiated
  directly under Node 26 *does* provide `localStorage`, so the gap is somewhere between that
  and vitest's `environment: "jsdom"` rather than in jsdom alone.
- `src/shared/hooks/useTheme.test.tsx` fails differently:
  `getByRole("button", { name: "system" })` reports **multiple elements found**, which reads
  like a duplicate render rather than a missing one.

### Why it is worth fixing rather than closing

`web/AGENTS.md` already states that Node 22 is the version that builds this bundle, and that
`@types/node` is deliberately pinned to it for that reason. But **nothing enforces it locally**:
there is no `.nvmrc`, no `engines.node` in `web/package.json`, and no check. A contributor on a
current Node gets a 303-failure wall on a clean checkout with no indication that the version is
the problem.

The concrete cost is not the inconvenience. It is that **a red baseline destroys the signal.**
This was found because a PR (#807) shipped a real test regression that CI caught and the author
did not: three `getByRole("columnheader")` queries broke when column header copy changed, and
those three failures were invisible inside 303 pre-existing ones. A suite that is red for
[...]

Issue #804, opened 2026-08-26
[FEATURE] Type scale, phase 2: Routing column widths, identifiers in mono, and the last arbitrary sizes


jigjigjig:
> **Retraction first.** The original version of this issue led with "five meanings share 14px"
> and named the Activity table's explanation line as the headline instance. **Both claims were
> wrong.** They came from misreading a computed-style census: a bucket at `14px/400` printed
> three samples that were all the same element, and I generalized from the bucket without
> probing the other line by name.
>
> Measured in the live DOM, on a seeded database, at 1280px:
>
>     "demo-guarded"                            14px/400  #1a1a1a
>     "the policy's default target"             12px/400  #545b62
>     "attempt 2 of 2 failed, plan exhausted"   12px/400  #545b62
>
> `features/activity/ActivityPage.tsx:749` has read `text-xs break-words text-muted` throughout,
> so that size step was always there. And classifying every 14px node on Activity by structure:
> **386 of ~427 are table cells**, which is what body size is for on a 161-row table, not
> evidence of a collapse. There is no section heading, card title, or secondary explanation in
> that population.
>
> The issue below is what survived re-measurement. It is a much smaller piece of work than the
> original framing claimed.

## What is actually wrong

Three findings, each with a measurement behind it.

**1. Section headings render at body size.** On Usage, `Spend by model` is Zilla Slab
**14px/550**, the same size as the table rows beneath it. Same for `Spend by session` and
`Gateway-run tools`. These are `<h2>`/`<h3>` elements carrying `text-sm font-semibold`, so a
section heading is distinguished from its own content by weight alone. They want `text-heading`
(18) or `text-title` (16) depending on whether they head a page section or a card.

**2. One sub-table hand-rolls its column headers.** `features/activity/ActivityPage.tsx:824-842`
renders the per-attempt plan table's headers as `<th className="px-3 py-2 text-left
font-medium">`. The main tables now wear the single overline spec through
`.otari-table .table__column`; this one does not, so two tables on one page label their columns
differently.

**3. A dialog title at body size.** `ActivityPage.tsx:227`, `Popover.Heading className="text-sm
font-medium"`. Wants `text-title`.

## What is NOT wrong, recorded so it is not "fixed" later

- **Table cells at 14px are correct.** That is `text-body`. A dense table having hundreds of
  body-size text nodes is the intended outcome.
- **Secondary explanation lines are already a step down**, at 12px muted. Moving them to
[...]

PR #807, opened 2026-08-26
fix(dashboard): repair the text-subtle contrast failure, and retune the type scale

### LINKED ISSUE #803: [BUG] --color-text-subtle fails WCAG AA on every surface in both themes
**Scope:** dashboard (`web/`), repo-level. Not a deployment problem, so the environment fields
of the bug template do not apply: reproducible on any build of `main`, currently 065fe511.

### What happened

`--color-text-subtle` fails WCAG AA for normal text on **every** background and surface token
the design foundation declares, in **both** themes. It is the token behind tertiary text:
timestamps, helper text, and `text-chrome-meta`, the second line under a nav row or an account
menu row.

Measured with WCAG 2.x relative luminance, sRGB, `(L_lighter + 0.05) / (L_darker + 0.05)`,
against the six grounds declared at `web/src/styles/globals.css:112-117` (light) and `349-354`
(dark). AA for normal text is 4.5.

**Light, `--color-text-subtle: #80868d`**

| ground | ratio | |
| --- | --- | --- |
| `background` / `surface` `#ffffff` | 3.68 | fail |
| `surface-muted` `#f1f4f6` | 3.33 | fail |
| `background-subtle` `#eef1f4` | 3.24 | fail |
| `background-muted` `#e8ecef` | 3.09 | fail |
| `surface-subtle` `#e4e9ed` | 3.01 | fail |

**Dark, `--color-text-subtle: #6e7882`**

| ground | ratio | |
| --- | --- | --- |
| `background` `#15191f` | 3.92 | fail |
| `surface` `#1b1f27` | 3.67 | fail |
| `background-muted` `#222731` | 3.33 | fail |
| `surface-muted` `#232831` | 3.29 | fail |
| `background-subtle` `#292e38` | 3.03 | fail |
| `surface-subtle` `#2b313b` | 2.91 | fail |

Every value clears 3.0, so all of it passes AA for *large* text and none of it passes for the
small text these tokens actually pair with.

The pairing worth calling out is `background-subtle`. Its own comment in `globals.css` reads
"hover/pressed on muted", so it is the hovered nav row, and `text-chrome-meta` sits on hovered
rows. The worst-contrast state of that component is one a person reaches by moving the mouse.

### Steps to reproduce

1. Run the dashboard (`pnpm --dir web run dev` against a gateway).
2. Read any timestamp, helper line, or the second line under a nav row, in either theme.
3. Or compute the ratios directly from the token values at the two `globals.css` line ranges above.

### What you expected

Text tokens intended for normal-size body text clear 4.5:1 against every surface they can land
on, not against a subset.

### Notes

`--color-text-muted` is fine in both themes and is not part of this (light worst case 5.63, dark
5.17); only the `subtle` step fails.

Worth knowing for whoever picks this up: the failure was easy to miss because weight 500 was
disguising it. The dashboard had no regular-weight text in its chrome, and the extra stem weight
reads as more contrast than the ratio gives. Dropping the weight floor is what made it visible.

jigjigjig:
## What this does, and what it does not

The dashboard's type carried its hierarchy in boldness. Nothing in the shell chrome was
regular weight: every nav row, button, chip and field label sat at 500, so weight could
not signal anything because every row had already spent it. Above that floor, five
different meanings shared a single size.

This change fixes the **weight** and **tracking** axes completely, and the **size** axis
only where a shared primitive owns it. It does not fix the size axis at the call sites,
because that lives in which utility a call site names and no token can reach it. That
work is scoped as a follow-up and is linked below. The before/after is modest and
structural rather than dramatic, and that is the honest description of it.

## The mechanism, and two cascade bugs found by compiling rather than reasoning

**1. A `@theme` override of these keys is a silent no-op.** `@heroui/styles` is not a
consumer of Tailwind's type variables, it is a *prebuilt Tailwind stylesheet* that ships
its own `@layer theme { :root, :host { … } }` carrying the full default theme:
`--text-xs`, `--text-sm`, `--font-weight-medium: 500`, `--tracking-tight` and every
`--text-*--line-height`. Tailwind hoists all `@theme` blocks into its own theme layer at
the position of `@import "tailwindcss"` (line 1), so HeroUI's literal theme layer lands
second. Same layer, same `:root` specificity: source order decides, and HeroUI wins.

An override written in `@theme` therefore compiles clean, emits, passes every check, and
changes nothing. Worse, it half-works: the *new* keys survive, because HeroUI never
declares them, so tracking would have landed while every size and weight silently
reverted. `globals.css` already carried a comment saying its `@theme` block was
"non-overlapping with HeroUI's tokens", which is the shape of a workaround for this
without the diagnosis.

So the scale is split in two, and both halves are load-bearing:

- **`@theme`** registers only the keys HeroUI never declares: `--text-caption` and the
  `--text-*--letter-spacing` set. Registration is what makes Tailwind emit the
  `letter-spacing` line inside `.text-xs` and friends at all. A tracking value declared
  only at `:root` is never read, because the utility would not reference it.
- **An unlayered `:root` block**, after the HeroUI import, carries every value HeroUI
  also declares. Unlayered beats every `@layer`, so it wins because of *what it is* and
  not because of where it sits. That distinction is commented in the file, because the
  next `@import` somebody adds would quietly outrank a merely-later block.

**2. The `h1` to `h6` font-family rule was unlayered.** `@utility` lands in
`@layer utilities`, and an unlayered rule beats every layer, so
`h1,…,h6 { font-family: var(--font-display) }` silently overrode the `font-family` of any
role applied to a heading element. `text-heading` and `text-title` are set in the body
face now and are worn by `<h2>` and `<h3>`, so this would have kept them in Zilla Slab
and made the role map a suggestion with no symptom to chase. It is now in `@layer base`:
a bare heading still gets the display face, and a role wins.

Both bugs are invisible at every stage except a rendered page, which is why they were
found by compiling the stylesheet and reading the emitted CSS rather than by reasoning
about it.

## The thing this PR is about, found inside this PR

A reviewer recompiled this branch and found that **the fix contained the bug it exists to
prevent.**

`--text-caption` was registered in the `--text-*` namespace, which makes Tailwind generate a
`.text-caption` **size** utility. That landed in `@layer utilities` beside the `@utility
text-caption` **role** of the same name and, being second, won. The role's own `font-size`,
`line-height` and `letter-spacing` were dead declarations. Editing them compiled, emitted, passed
lint, typecheck and every test, and changed nothing on screen.

I had checked that collision earlier and cleared it, because I verified that the role's `color`
survived the merge. It does. Its metrics do not. The check that settles it is setting the role to
`font-size: 99px` and reading the compiled output, where `var(--text-caption)` still wins.

The step is now `--text-caption-step`, declared in the unlayered value block where it generates no
utility, so the role owns its whole spec and no collision is possible. And the `@theme` assertion
had been whitelisting `--text-caption` and `--text-caption--line-height` to accommodate it: both
exceptions are gone, so it now reads flatly as *letter-spacing keys only, no exceptions*.

This is stated up front rather than in a fix list because it is the strongest argument the PR has.
The failure mode described below is not hypothetical and it is not rare enough to reason your way
past: it got into the change whose entire subject is that failure mode, past three new tests
written to catch it, and it took a human recompiling the branch to see it. That is what the tests
are for, and it is why the rule they now enforce has no exceptions left to erode.

## The scale

Whole pixels throughout; a fractional size reflows every dense table for no legible gain,
tested at 100%, 125% and 150% zoom. Tracking is positive on the small steps, where tight
setting goes muddy, and negative on the large ones, where it goes loose.

| step | size / line-height | tracking |
| --- | --- | --- |
| `--text-xs` | 12 / 18 | +0.01em |
| `--text-caption` | 13 / 19 | +0.0075em |
| `--text-sm` | 14 / 20 | +0.005em |
| `--text-base` | 16 / 24 | 0 |
| `--text-lg` | 18 / 26 | -0.01em |
| `--text-xl` | 22 / 28 | -0.015em |
| `--text-2xl` | 26 / 32 | -0.02em |

`--text-caption` at 13 is a new seventh step. Tailwind ships nothing between 12 and 14,
and that absence is why five meanings were sharing 14px: a secondary line had nowhere
lighter to go than the value it was explaining.

Weights: `normal` 400, **`medium` 400**, `semibold` 550, `bold` 600.

`--font-weight-medium: 400` is deliberate and **the token name is now a lie.** It is how
roughly 140 `font-medium` call sites and HeroUI's own 93 internal uses come off 500 in two
lines rather than in 140 edits, and it is the single largest visual change here. Deleting
the now-inert utilities is follow-up, not this change.

## Contrast: a pre-existing bug this found, not one it caused

`--color-text-subtle` fails WCAG AA for normal text today, on every surface, in both
themes. Dropping the weight floor to 400 did not break it; it removed what was disguising
it. Measured against all six background and surface tokens each theme declares:

**Light: `#80868d` to `#62686f`**

| ground | before | after |
| --- | --- | --- |
| `background` / `surface` `#ffffff` | 3.68 | **5.63** |
| `surface-muted` `#f1f4f6` | 3.33 | **5.10** |
| `background-subtle` `#eef1f4` | 3.24 | **4.97** |
| `background-muted` `#e8ecef` | 3.09 | **4.74** |
| `surface-subtle` `#e4e9ed` | 3.01 | **4.61** |

**Dark: `#6e7882` to `#919ba5`**

| ground | before | after |
| --- | --- | --- |
| `background` `#15191f` | 3.92 | **6.24** |
| `surface` `#1b1f27` | 3.67 | **5.85** |
| `background-muted` `#222731` | 3.33 | **5.30** |
| `surface-muted` `#232831` | 3.29 | **5.24** |
| `background-subtle` `#292e38` | 3.03 | **4.82** |
| `surface-subtle` `#2b313b` | 2.91 | **4.63** |

Every ground clears 4.5 after. The worst case is 4.61 light and 4.63 dark, so neither
theme rides the line. WCAG 2.x relative luminance, sRGB, `(L_hi + 0.05)/(L_lo + 0.05)`,
measured independently twice.

Every figure above was computed twice, independently, by two people working from the token
values in `globals.css` rather than one carrying numbers across. The first pass sampled
three of the six grounds and reported the dark repair as passing; it was not. Which is the
point of the next paragraph.

The two deepest surfaces are the ones that matter and were the ones missed on the first
pass: `background-subtle` is the **hover/pressed state of a nav row**, and
`text-chrome-meta` sits on hovered rows, so the failing pairing was one an operator hits
by moving the mouse.

## Roles

The seven roles now read the scale instead of hardcoding rem, which is what makes the
follow-up mechanical: fixing a call site becomes changing which role it names, not
re-deriving a number per site.

| role | step | weight | face |
| --- | --- | --- | --- |
| `text-display` | `--text-2xl` 26 | 500 | Zilla Slab, -0.01em |
| `text-heading` | `--text-lg` 18 | 550 | body |
| `text-title` | `--text-base` 16 | 550 | body |
| `text-body` | `--text-sm` 14 | 400 | body |
| `text-emphasis` | `--text-sm` 14 | 550 | body |
| `text-caption` | `--text-caption` 13 | 400 | body, muted |
| `text-overline` | `--text-xs` 12 | 550 | body, +0.06em, uppercase, muted |

`text-display` overrides the step's -0.02em to -0.01em: Zilla Slab's slab serifs already
close the counters, so the same negative tracking that reads as resolved on the sans reads
as condensed on the serif. Any future sans consumer of `text-2xl` should keep -0.02em.

`text-chrome-initials` stays hardcoded at 9px, off-scale, as the one documented exception:
two letters in a 26px avatar are recognized, not read, so putting them on the scale would
either overflow the avatar or drag a real text step down to suit a monogram.

`text-title` and `text-body` have never had a call site in product code and become
correct and available here without being called. That is deliberate.

## Call sites, bounded

- **Column headers** adopt the overline spec, replacing a hardcoded `font-weight: 600` in
  `.otari-table .table__column` that no token could reach. With the rest of the scale down
  at 400/550 that literal had become the heaviest text in the product, so a header row
  out-weighted every cell it labeled. Uppercase costs column width, and three Models
  headers were shortened rather than the spec weakened: `Base in / out $ / 1M` ->
  `Base in / out / 1M`, `Caching policy` -> `Caching`, `Pricing policy` -> `Pricing`. The
  `$` comes out because every cell in that column already shows it; `/ 1M` stays because
  nothing else on the page carries the basis. The tooltip on that column explains how
  unpriced models are metered and never states the unit, so cutting `/ 1M` too would have
  left the basis for `$0.90` stated nowhere. Measured
  on the page: header row 56px -> 41px, Model column 354px -> 389px, and the longest
  identifier stops wrapping.

  **Known costs of the uppercase spec, accepted deliberately and consistently.** Uppercase plus
  0.06em tracking makes every header wider, and where that forces a choice we have taken the
  wrap rather than a rename, every time.

  On Models, `Pricing policy` and `Base in / out / 1M` both wrap to two lines, the header row is
  56px rather than 41px, and the 24px this costs the Model column (389px to 365px) re-wraps the
  longest identifier, `anthropic.claude-3-5-sonnet-20241022-v2:0`. `Pricing policy` was briefly
  shortened to `Pricing` and that is reverted: alone it sits two columns from
  `Base in / out / 1M` and reads as a second price column rather than as the policy that produced
  the price. `POLICY` was considered and rejected, because `Caching policy` is already `Caching`,
  so `CACHING` beside `POLICY` breaks the parallel between two headers naming the same kind of
  thing and shows a reader two categories where there is one. Shortening that pair would have to
  be done as a pair, which is a naming exercise and not a line in a type PR.

  `APPLIES TO` also wraps to two lines on Routing.
  Shortening it was considered and rejected. The three cuts above each remove a word that
  carries nothing (every value in the caching column is a policy), where the candidate here
  was replacing a plain phrase with a shorter noun, and `RoutingPage.tsx:1335` gives that
  column the id `scope`, which is evidence the shorter word is our internal name for it
  rather than the reader's. Routing also has two wraps that predate this change,
  `demo-failover` in the Policy column and `Every caller` in this one, so a naming change
  would have fixed one third of a layout problem. All three are column-width work in the
  follow-up issue.
- **Three overline spellings collapse into one.** Seven call sites carried
  `text-[11px] font-medium uppercase tracking-wide`, `text-xs font-semibold uppercase
  tracking-wide` and `text-xs font-medium uppercase tracking-wide`; the rail's label, a
  stat-card label and a column header now measure identically.
- **The ladder is no longer inverted.** A page title was `text-xl font-semibold` (20px)
  while a stat value inside a card was `text-xl … sm:text-2xl` (24px), so a number
  outranked the name of the page. The title moves up to `text-display` (26) and the stat
  down to `text-xl` (22): both halves move, so the ladder is correct by construction
  rather than by two call sites agreeing.
- **Page descriptions get `max-w-prose`.** They ran the full container width: 968px at
  14px is ~138 characters per line, roughly twice a comfortable measure, on every page.
  Now ~89.
- **The delta slot reserves two lines.** A delta like `5.8% errors · ▲ 214.1% vs prev`
  wrapped in a narrow tile while its four neighbors stayed on one line, and the row of
  five stat cards stopped aligning. This is a layout fix, not a type fix: sizing type to
  the longest string is sizing type by accident.
- **`tabular-nums` in one place, not thirteen.** `DataTable` already applies it to every
  `align: "end"` column, and a DOM scan found no numeric *cell* without it. What lacked it
  was numbers inside sentences, where proportional figures are correct. The one addition
  is the delta slot, where a column of five percentages does align.

## Type is not a width tool

> Type is what you reach for when the layout is right and it still reads badly.

Three of the fixes above were layout problems wearing a type problem's clothes, and they
are worth naming together because the pattern is the more useful finding:

- **The delta slot.** A delta wrapped in a narrow tile while its neighbors did not, so the
  row of stat cards stopped aligning. The tempting fix is a smaller font. The fix is a
  reserved two-line height.
- **The `Over budget` chip.** It had wrapped onto its own line because the stat value beside
  it had grown. It needed no work at all: it came back inline the moment the stat took its
  correct step on the scale.
- **The column headers.** Uppercase costs width, and the first instinct was to give up the
  uppercase. What it actually needed was three header strings with a dead word each
  (`CACHING POLICY` -> `CACHING`; every value in that column is a policy).

In each case a type value had been, or was about to be, tuned to survive a layout
constraint. That is how a scale degrades quietly: each value looks defensible on its own
and only the pile-up is visible. It is also the honest explanation for how this codebase
got five meanings onto 14px. Nobody made a bad decision; they made forty locally
reasonable ones.

## Deliberately not in this change

The size axis at the call sites: a handful of headings that render at body size, and one
sub-table that hand-rolls its column headers instead of wearing the overline spec. Tracked
in #804, along with measure control on the descriptions the two shared helpers do not
reach, the Routing column widths, and identifiers in the mono face.

**Correction to an earlier version of this description.** It claimed five meanings share
14px, and that the Activity table renders its explanation line at the same size as the
value it explains. **Both were wrong and I am the author of the error.** Measured in the
live DOM: the policy name is 14px and the explanation beside it is 12px muted
(`ActivityPage.tsx:749` has read `text-xs ... text-muted` throughout), so that size step
was always there. And of the ~427 nodes at 14px on Activity, 386 are table cells, which is
what body size is for on a 161-row table rather than evidence of a collapse. The original
claim came from reading a computed-style census bucket whose three printed samples were all
the same element, and generalizing without probing the second line by name.

Nothing in this PR depends on that claim: the weight floor, the tracking, the contrast
repair, the two cascade bugs and the title/stat ladder were each measured directly and are
unaffected. #804 has been corrected in the same way, leading with the retraction.

## Checks

`pnpm --dir web run lint` and `run typecheck` clean.

Three new tests in `src/styles/foundation.test.ts` pin the split from both sides: the
unlayered `:root` block declares every contested key; neither theme block declares any
`--text-*` / `--font-weight-*` / `--tracking-*`; and `@theme` registers the letter-spacing
keys and does **not** register a size or weight HeroUI also declares. The third guards
against reintroducing the silent no-op above.

Type metrics are deliberately declared once rather than in both theme blocks. Type has no
theme axis, and declaring them twice would allow one block to be edited and the other
missed, giving the two themes different type metrics: invisible in whichever theme you are
looking at, and undetectable by the existing both-themes assertion, because the key sets
would still match.

The vitest suite is red on `main` in this environment independent of this change: 303
failures across 21 files, including `AppShell.test.tsx` (60/60), `Login.test.tsx` (42/42),
`ModelsPage.test.tsx` (39/39) and `navigationHistory.test.ts` (9/9), all reproducible with
this branch's changes stashed. This branch's delta is +3 passing tests, which are the
three it adds.

The screenshot suite is `workflow_dispatch`-only, its baselines are gitignored, and CI
runs it with `--update-snapshots=missing`, so it writes baselines rather than failing. It
is a before/after tool here, not a regression gate.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [x] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

Both, deliberately. The Bug Fix half is the WCAG AA contrast failure (#803) and the two
cascade bugs described above. The Refactor half is the scale itself, which changes no
behavior an operator can name.

## Relevant issues

Fixes #803

Related, and deliberately **not** fixed here:

- #804 is the follow-up this change was scoped against. It carries the call-site migration
  that separates the five meanings still sharing 14px, the Routing column widths, and
  identifiers in the mono face. Nothing in it is fixed by this PR, and the "Deliberately not
  in this change" section above says why.
- #805 (the dashboard suite cannot run on Node 26, and nothing pins the version locally) was
  filed from this branch on a wrong premise and has been corrected. It is not fixed here. The
  Checklist note below explains why it is worth reading before trusting any local test result
  on this repo.
- #806 (an older gateway's bootstrap white-screens the sign-in page) was found while standing
  up the environment used to verify this change. Unrelated to typography and not fixed here.

## Checklist

      Three tests in `web/src/styles/foundation.test.ts`. Not `tests/unit` or
      `tests/integration`: this change is entirely under `web/`, whose suite is Vitest.
      Ticked with a note: those three are the Python side and this PR touches no Python. Their
      dashboard counterparts are what ran, and CI is the authority on the third.
      - `pnpm --dir web run lint` clean, locally and in CI.
      - `pnpm --dir web run typecheck` clean, locally and in CI.
      - `pnpm --dir web test`: green in CI. **Correction, and it matters, so it is here rather
        than buried:** I originally reported this suite as red on `main` independent of this
        change, with a 303-failure baseline. That was wrong. Those 303 failures are a **local
        Node 26** problem; CI runs Node 22 and passes. Which means the baseline I was measuring
        against was noise, and it hid a real regression this branch introduced: three
        `getByRole("columnheader")` queries in `ModelsPage.test.tsx` still named the old header
        copy. CI caught it, I did not. Fixed in 9521fd6e. #805 has been rewritten from "the
        suite is red on main" to what it actually is, a missing local Node pin, with this as
        the argument for why that is worth fixing.
      The reasoning lives in `globals.css` itself, at the two places where the next person
      would otherwise reintroduce the bugs: a comment on the unlayered `:root` block saying
      why it is unlayered rather than merely later, and one on the `@layer base` wrap saying
      what it stops overriding.
      **Not applicable.** No route, schema, or docstring changed; no API contract is touched.
      `web/src/client/schema.ts` and `web/src/routeTree.gen.ts` are both unchanged, so there
      is no generated artifact to commit. The dashboard bundle is gitignored as usual.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

Three agent sessions: a design engineer producing the scale, a coordinator, and this one
implementing and verifying. Worth knowing for review, because it shaped what is here: the
numbers in the scale were proposed by one session and every one of them was re-measured by
another before landing. That caught the two cascade bugs, and it caught a proposed dark
contrast value that had been checked against three of the six surfaces and reported as
passing when it failed on the other two.

Every quantitative claim in the description above was measured on the running dashboard
against a seeded database, or by compiling the stylesheet and reading the emitted CSS. None
of it is inferred. Where a measurement contradicted a plausible assumption, the measurement
won, and the description says so rather than presenting the conclusion alone.

**NOTE:**

### DIFF
--- a/web/src/features/activity/ActivityTimeline.tsx
+++ b/web/src/features/activity/ActivityTimeline.tsx
@@ -281,7 +281,7 @@ export function ActivityTimeline({
             <ChartLegend series={chartSeries} />
           </span>
           <div className="flex items-center gap-1.5">
-            <span className="hidden text-[11px] text-muted sm:inline">
+            <span className="hidden text-xs text-muted sm:inline">
               drag across the chart to zoom
             </span>
             <Button
--- a/web/src/features/activity/ActivityTimeline.tsx
+++ b/web/src/features/activity/ActivityTimeline.tsx
@@ -275,7 +275,7 @@ export function ActivityTimeline({
       <div className="rounded-xl border border-border bg-surface p-2">
         <div className="flex items-center justify-between gap-2 px-1 pb-1">
           <span className="flex items-center gap-3">
-            <span className="text-[11px] font-medium uppercase tracking-wide text-muted">
+            <span className="text-overline">
               Requests / {bucket === "hour" ? "hour" : "day"}
             </span>
             <ChartLegend series={chartSeries} />
--- a/web/src/features/models/ModelsPage.tsx
+++ b/web/src/features/models/ModelsPage.tsx
@@ -593,9 +593,7 @@ function PanelSection({
 }) {
   return (
     <div className="flex flex-col gap-2">
-      <span className="text-xs font-semibold uppercase tracking-wide text-muted">
-        {title}
-      </span>
+      <span className="text-overline">{title}</span>
       {children}
     </div>
   )
@@ -1304,7 +1302,7 @@ function ModelTable({
         id: "input",
         header: (
           <span className="inline-flex items-center gap-1">
-            {`${comparisonLabel} in / out $ / 1M`}
+            {`${comparisonLabel} in / out / 1M`}
             <PricingInfo />
           </span>
         ),
@@ -1326,7 +1324,10 @@ function ModelTable({
       },
       {
         id: "caching",
-        header: `Caching ${comparisonContextTokens == null ? "policy" : comparisonLabel}`,
+        header:
+          comparisonContextTokens == null
+            ? "Caching"
+            : `Caching ${comparisonLabel}`,
         align: "end",
         cell: (row) => (
           <CachingCell
@@ -1338,7 +1339,7 @@ function ModelTable({
       },
       {
         id: "policy",
-        header: "Pricing policy",
+        header: "Pricing",
         align: "end",
         cell: (row) => (
           <PricingPolicyCell row={row} onEdit={() => onEditPricing(row.key)} />
--- a/web/src/features/models/ModelsPage.tsx
+++ b/web/src/features/models/ModelsPage.tsx
@@ -1339,7 +1339,13 @@ function ModelTable({
       },
       {
         id: "policy",
-        header: "Pricing",
+        // Deliberately not shortened, unlike `Caching` beside it. `Pricing`
+        // alone sits two columns from `Base in / out / 1M` and reads as a
+        // second price column rather than as the policy that produced the
+        // price, so it wraps instead. See the header comment in globals.css on
+        // the overline spec: uppercase costs width, and a wrapped header costs
+        // a glance where a misleading one costs a wrong conclusion.
+        header: "Pricing policy",
         align: "end",
         cell: (row) => (
           <PricingPolicyCell row={row} onEdit={() => onEditPricing(row.key)} />
--- a/web/src/shared/components/ui.tsx
+++ b/web/src/shared/components/ui.tsx
@@ -180,11 +180,16 @@ export function StatCard({
     // Card.Content's, which otherwise doubled the tile's height (most visible at
     // two-up on mobile). Content owns the padding: tighter on mobile, roomier up.
     <Card.Content className="flex flex-col gap-1 p-4 sm:p-5">
-      <span className="text-xs font-medium uppercase tracking-wide text-muted">
-        {label}
-      </span>
+      <span className="text-overline">{label}</span>
       <span className="flex flex-wrap items-center gap-2">
-        <span className="text-xl font-semibold text-foreground sm:text-2xl">
+        {/* text-xl (22px), deliberately a step *below* the page title's
+            text-display (26px). It used to be text-xl rising to text-2xl at
+            `sm`, which made a number inside a card the largest text on the
+            page, 4px bigger than the name of the page itself. Fixing that by
+            raising the title alone would have left the two agreeing by
+            coincidence; both halves move so the ladder is correct by
+            construction. `tabular-nums` so a column of these aligns. */}
+        <span className="text-xl font-semibold tabular-nums text-foreground">
           {value}
         </span>
         {status && statusLabel ? (
@@ -195,7 +200,16 @@ export function StatCard({
           </span>
         ) : null}
       </span>
-      {hint ? <span className="text-xs text-muted">{hint}</span> : null}
+      {/* Two lines reserved, always. A delta like "5.8% errors · ▲ 214.1% vs
+          prev" wraps in a narrow tile while its neighbours stay on one line,
+          and a row of five tiles then stops aligning. This is a layout fix and
+          not a type fix on purpose: sizing the type to the longest string would
+          be sizing the type by accident. min-h-9 is two 18px lines. */}
+      {hint ? (
+        <span className="block min-h-9 text-xs tabular-nums text-muted">
+          {hint}
+        </span>
+      ) : null}
       {chart ? <div className="mt-2">{chart}</div> : null}
     </Card.Content>
   )
@@ -288,9 +302,12 @@ export function PageHeader({
   return (
     <div className="flex flex-col gap-3">
       <div>
-        <h1 className="text-xl font-semibold text-foreground">{title}</h1>
+        <h1 className="text-display">{title}</h1>
         {description ? (
-          <p className="mt-1 text-sm text-muted">{description}</p>
+          // max-w-prose because this ran to the full container width: 968px at
+          // 14px is ~138 characters per line, roughly twice a comfortable
+          // measure, and it is the same paragraph on every page.
+          <p className="mt-1 max-w-prose text-sm text-muted">{description}</p>
         ) : null}
       </div>
       {/* The primary action sits on its own left-aligned row under the heading,
@@ -641,9 +658,9 @@ export function EmptyState({
     <Card>
       <Card.Content className="flex flex-col gap-4 p-6">
         <div>
-          <h2 className="text-lg font-semibold text-foreground">{title}</h2>
+          <h2 className="text-heading">{title}</h2>
           {description ? (
-            <p className="mt-1 text-sm text-muted">{description}</p>
+            <p className="mt-1 max-w-prose text-sm text-muted">{description}</p>
           ) : null}
         </div>
         {children}
--- a/web/src/shared/components/ui.tsx
+++ b/web/src/shared/components/ui.tsx
@@ -201,7 +201,7 @@ export function StatCard({
         ) : null}
       </span>
       {/* Two lines reserved, always. A delta like "5.8% errors · ▲ 214.1% vs
-          prev" wraps in a narrow tile while its neighbours stay on one line,
+          prev" wraps in a narrow tile while its neighbors stay on one line,
           and a row of five tiles then stops aligning. This is a layout fix and
           not a type fix on purpose: sizing the type to the longest string would
           be sizing the type by accident. min-h-9 is two 18px lines. */}
--- a/web/src/styles/foundation.test.ts
+++ b/web/src/styles/foundation.test.ts
@@ -75,6 +75,78 @@ const HEROUI_VARIABLES = [
 
 const TYPE_NAMESPACES = /^--(text|font-weight|tracking)-/
 
+/**
+ * WCAG 2.x relative luminance, sRGB. Kept here rather than pulled in as a
+ * dependency: it is eight lines and the alternative is a package in the
+ * dependency graph for one assertion.
+ */
+function luminance(hex: string): number {
+  const m = /^#([0-9a-f]{6})$/i.exec(hex.trim())
+  if (!m) throw new Error(`not a six-digit hex color: ${hex}`)
+  const channel = (v: number): number => {
+    const c = v / 255
+    return c <= 0.03928 ? c / 12.92 : ((c + 0.055) / 1.055) ** 2.4
+  }
+  const n = Number.parseInt(m[1], 16)
+  return (
+    0.2126 * channel((n >> 16) & 0xff) +
+    0.7152 * channel((n >> 8) & 0xff) +
+    0.0722 * channel(n & 0xff)
+  )
+}
+
+function contrast(a: string, b: string): number {
+  const [hi, lo] = [luminance(a), luminance(b)].sort((x, y) => y - x)
+  return (hi + 0.05) / (lo + 0.05)
+}
+
+describe("text on every ground it can land on", () => {
+  // The repair in this change was measured against all six grounds each theme
+  // declares, and the measurement is the assertion. Without it the WCAG AA fix
+  // regresses on the next palette edit with no symptom until somebody renders
+  // it, which is the same invisible-until-rendered class of failure the cascade
+  // tests above exist for.
+  //
+  // Both of these failed AA on every ground before this change (light worst case
+  // 3.01, dark 2.91), and the reason it went unnoticed for so long is that the
+  // dashboard had no regular-weight text: extra stem weight reads as more
+  // contrast than the ratio gives.
+  const GROUNDS = [
+    "--color-background",
+    "--color-background-muted",
+    "--color-background-subtle",
+    "--color-surface",
+    "--color-surface-muted",
+    "--color-surface-subtle",
+  ] as const
+  const INK = [
+    "--color-text",
+    "--color-text-muted",
+    "--color-text-subtle",
+  ] as const
+  // 4.5:1 is AA for normal-size text, which is what all three of these pair with.
+  const AA_NORMAL = 4.5
+
+  it.each([
+    ["light", LIGHT],
+    ["dark", DARK],
+  ] as const)("clears AA for normal text in the %s theme", (_theme, tokens) => {
+    for (const ink of INK) {
+      for (const ground of GROUNDS) {
+        const fg = tokens.get(ink)
+        const bg = tokens.get(ground)
+        expect(fg, `${ink} is not declared`).toBeDefined()
+        expect(bg, `${ground} is not declared`).toBeDefined()
+        const ratio = contrast(fg as string, bg as string)
+        expect(
+          Number(ratio.toFixed(2)),
+          `${ink} on ${ground} is ${ratio.toFixed(2)}:1, under the ${AA_NORMAL}:1 AA floor for normal text`,
+        ).toBeGreaterThanOrEqual(AA_NORMAL)
+      }
+    }
+  })
+})
+
 describe("the type scale's two halves", () => {
   // The scale is split on purpose and the split is easy to undo by accident, so
   // it is pinned from both sides. `@heroui/styles` is a prebuilt Tailwind
@@ -84,7 +156,19 @@ describe("the type scale's two halves", () => {
   // order: it compiles, emits, and changes nothing. The values therefore live in
   // an unlayered `:root` block, which beats every layer, and only the keys
   // HeroUI never declares are registered in `@theme`.
+  // `block()` is a first-match `indexOf`, and globals.css has two top-level
+  // `:root {` blocks: this one and the base reset further down. It resolves
+  // correctly only because this one comes first, so assert that rather than
+  // depend on it: a reordering would otherwise point every failure below at the
+  // wrong block.
   const TYPE_VALUES = declarations(block(":root"))
+
+  it("matches the type block rather than the base reset", () => {
+    expect(
+      TYPE_VALUES.get("--text-xs"),
+      "block(':root') matched a different :root block (the base reset?); anchor it on something distinguishing",
+    ).toBeDefined()
+  })
 
   it("declares every contested type key in the unlayered :root block", () => {
     for (const key of [
@@ -105,6 +189,9 @@ describe("the type scale's two halves", () => {
       "--font-weight-semibold",
       "--font-weight-bold",
       "--tracking-tight",
+      "--text-caption-step",
+      "--text-caption-step--line-height",
+      "--text-caption-step--letter-spacing",
     ]) {
       expect(
         TYPE_VALUES.get(key),
@@ -139,9 +226,6 @@ describe("the type scale's two halves", () => {
     // :root alone is never read, because the utility would not reference it.
     const theme = declarations(block("@theme"))
     for (const key of [
-      "--text-caption",
-      "--text-caption--line-height",
-      "--text-caption--letter-spacing",
       "--text-xs--letter-spacing",
       "--text-sm--letter-spacing",
       "--text-base--letter-spacing",
@@ -156,12 +240,14 @@ describe("the type scale's two halves", () => {
     }
     // The other direction: a size or weight registered here is the silent-no-op
     // mistake this split exists to prevent.
+    // No exceptions any more: the only type keys `@theme` may carry are the
+    // letter-spacing registrations. A *size* here would also generate a
+    // same-named utility that outranks the role of that name in the same layer,
+    // which is how the caption step silently killed `@utility text-caption`'s
+    // own metrics before it was renamed to `--text-caption-step`.
     const contested = [...theme.keys()].filter(
       (name) =>
-        TYPE_NAMESPACES.test(name) &&
-        !name.endsWith("--letter-spacing") &&
-        name !== "--text-caption" &&
-        name !== "--text-caption--line-height",
+        TYPE_NAMESPACES.test(name) && !name.endsWith("--letter-spacing"),
     )
     expect(
       contested,
--- a/web/src/styles/foundation.test.ts
+++ b/web/src/styles/foundation.test.ts
@@ -138,8 +138,11 @@ describe("text on every ground it can land on", () => {
         expect(fg, `${ink} is not declared`).toBeDefined()
         expect(bg, `${ground} is not declared`).toBeDefined()
         const ratio = contrast(fg as string, bg as string)
+        // Compare the unrounded ratio. Rounding first would let 4.496 pass as
+        // 4.50, which is a test that goes green on a value that fails AA.
+        // Rounding belongs in the message and nowhere else.
         expect(
-          Number(ratio.toFixed(2)),
+          ratio,
           `${ink} on ${ground} is ${ratio.toFixed(2)}:1, under the ${AA_NORMAL}:1 AA floor for normal text`,
         ).toBeGreaterThanOrEqual(AA_NORMAL)
       }
--- a/web/src/styles/foundation.test.ts
+++ b/web/src/styles/foundation.test.ts
@@ -73,6 +73,103 @@ const HEROUI_VARIABLES = [
   "--overlay-shadow",
 ]
 
+const TYPE_NAMESPACES = /^--(text|font-weight|tracking)-/
+
+describe("the type scale's two halves", () => {
+  // The scale is split on purpose and the split is easy to undo by accident, so
+  // it is pinned from both sides. `@heroui/styles` is a prebuilt Tailwind
+  // stylesheet carrying its own `@layer theme` with the default `--text-*`,
+  // `--font-weight-*` and `--tracking-*` in it, and it is imported after
+  // Tailwind, so an override of one of those written in `@theme` loses on source
+  // order: it compiles, emits, and changes nothing. The values therefore live in
+  // an unlayered `:root` block, which beats every layer, and only the keys
+  // HeroUI never declares are registered in `@theme`.
+  const TYPE_VALUES = declarations(block(":root"))
+
+  it("declares every contested type key in the unlayered :root block", () => {
+    for (const key of [
+      "--text-xs",
+      "--text-xs--line-height",
+      "--text-sm",
+      "--text-sm--line-height",
+      "--text-base",
+      "--text-base--line-height",
+      "--text-lg",
+      "--text-lg--line-height",
+      "--text-xl",
+      "--text-xl--line-height",
+      "--text-2xl",
+      "--text-2xl--line-height",
+      "--font-weight-normal",
+      "--font-weight-medium",
+      "--font-weight-semibold",
+      "--font-weight-bold",
+      "--tracking-tight",
+    ]) {
+      expect(
+        TYPE_VALUES.get(key),
+        `${key} is not declared in the unlayered :root block, so @heroui/styles' own value wins`,
+      ).toBeDefined()
+    }
+  })
+
+  it("keeps type metrics out of the theme blocks", () => {
+    // Not a style preference. Type has no theme axis, and declaring these twice
+    // would allow one block to be edited and the other missed, giving the two
+    // themes different type metrics: invisible in whichever theme you are
+    // looking at, and undetectable by the both-themes assertion below, because
+    // the key sets would still match.
+    for (const [theme, tokens] of [
+      ["light", LIGHT],
+      ["dark", DARK],
+    ] as const) {
+      const offenders = [...tokens.keys()].filter((name) =>
+        TYPE_NAMESPACES.test(name),
+      )
+      expect(
+        offenders,
+        `the ${theme} theme block declares type metrics (${offenders.join(", ")}); they belong in the unlayered :root block, declared once`,
+      ).toEqual([])
+    }
+  })
+
+  it("registers in @theme only the keys HeroUI does not declare", () => {
+    // Tracking is emitted into `.text-xs` and friends only if the key is
+    // registered at build time, so these have to be in `@theme`; a value at
+    // :root alone is never read, because the utility would not reference it.
+    const theme = declarations(block("@theme"))
+    for (const key of [
+      "--text-caption",
+      "--text-caption--line-height",
+      "--text-caption--letter-spacing",
+      "--text-xs--letter-spacing",
+      "--text-sm--letter-spacing",
+      "--text-base--letter-spacing",
+      "--text-lg--letter-spacing",
+      "--text-xl--letter-spacing",
+      "--text-2xl--letter-spacing",
+    ]) {
+      expect(
+        theme.get(key),
+        `${key} is not registered in @theme, so its utility never references it`,
+      ).toBeDefined()
+    }
+    // The other direction: a size or weight registered here is the silent-no-op
+    // mistake this split exists to prevent.
+    const contested = [...theme.keys()].filter(
+      (name) =>
+        TYPE_NAMESPACES.test(name) &&
+        !name.endsWith("--letter-spacing") &&
+        name !== "--text-caption" &&
+        name !== "--text-caption--line-height",
+    )
+    expect(
+      contested,
+      `@theme registers type values @heroui/styles also declares (${contested.join(", ")}); it loses on source order, so these belong in the unlayered :root block`,
+    ).toEqual([])
+  })
+})
+
 describe("design foundation tokens", () => {
   it("declares the same token set in both themes", () => {
     // Each theme block owns the complete set of variables it needs rather than
--- a/web/src/styles/globals.css
+++ b/web/src/styles/globals.css
@@ -119,7 +119,7 @@
   /* ── Text — brand black + cool neutral muted ─────────────────────────── */
   --color-text: #1a1a1a; /* primary body and headings (Otari brand black) */
   --color-text-muted: #545b62; /* secondary text, labels, captions */
-  --color-text-subtle: #80868d; /* tertiary, timestamps, helper text */
+  --color-text-subtle: #62686f; /* tertiary, timestamps, helper text */
   --color-text-on-primary: #ffffff; /* text on primary-colored bg */
   --color-text-on-inverse: #ffffff; /* text on inverse/dark surface */
 
@@ -356,7 +356,7 @@
   /* ── Text ──────────────────────────────────────────────── */
   --color-text: #f0ede5; /* warm off-white (echoes brand B26 inverted) */
   --color-text-muted: #9aa4ae;
-  --color-text-subtle: #6e7882;
+  --color-text-subtle: #919ba5;
   --color-text-on-primary: #0a1419; /* deep teal-black for contrast on light teal */
   --color-text-on-inverse: #15171c;
 
@@ -555,6 +555,97 @@
   --shadow-elevation-md: var(--shadow-md);
   --shadow-elevation-lg: var(--shadow-lg);
   --shadow-modal: var(--shadow-modal);
+  /* Type scale, registration half. Two kinds of key live here and only these
+   * two, because they are the keys `@heroui/styles` does not declare:
+   *
+   *   --text-caption          a seventh step Tailwind does not ship, between
+   *                           --text-xs (12) and --text-sm (14). Registering it
+   *                           here is what makes it a real step rather than a
+   *                           number repeated in a role.
+   *   --text-*--letter-spacing  tracking per size step. Registration is what
+   *                           makes Tailwind emit the `letter-spacing` line in
+   *                           `.text-xs` and friends at all; a value declared
+   *                           only at :root is never read, because the utility
+   *                           would not reference it. So these must be here and
+   *                           the sizes must not (see the :root block below).
+   *
+   * Every value that HeroUI also declares lives in the unlayered `:root` block
+   * further down, NOT here. The comment there says why. */
+  --text-caption: 13px;
+  --text-caption--line-height: 19px;
+  --text-caption--letter-spacing: 0.0075em;
+  --text-xs--letter-spacing: 0.01em;
+  --text-sm--letter-spacing: 0.005em;
+  --text-base--letter-spacing: 0em;
+  --text-lg--letter-spacing: -0.01em;
+  --text-xl--letter-spacing: -0.015em;
+  --text-2xl--letter-spacing: -0.02em;
+}
+
+/* =============================================================
+ * Type scale, value half. READ THIS BEFORE ADDING AN @import.
+ *
+ * These are the keys `@heroui/styles` declares too, and that is the whole
+ * reason this block exists and is shaped the way it is.
+ *
+ * `@heroui/styles` is a prebuilt Tailwind stylesheet: it ships its own
+ * `@layer theme { :root, :host { … } }` carrying the full default theme,
+ * `--text-sm`, `--font-weight-medium: 500` and `--tracking-tight` included.
+ * Tailwind hoists every `@theme` block into its own theme layer at the
+ * position of `@import "tailwindcss"`, which is line 1, so HeroUI's literal
+ * theme layer lands after ours. Same layer, same `:root` specificity: source
+ * order decides, and HeroUI is second. An override of one of these keys
+ * written in `@theme` therefore compiles clean, emits, and loses. It changes
+ * nothing and reports nothing.
+ *
+ * So this block is deliberately **unlayered**. An unlayered declaration beats
+ * every `@layer`, which means it wins because of what it is and not because of
+ * where it sits. Do not move these into `@theme`, and do not rely on this
+ * block merely coming after the import: the next `@import` somebody adds would
+ * quietly outrank it if the win were only positional.
+ *
+ * Deliberately NOT in the light and dark theme blocks. Type metrics have no
+ * theme axis, and declaring twenty-odd of them twice creates a failure the
+ * single declaration cannot have: one block edited and the other missed, giving
+ * the two themes different type metrics, invisible in whichever theme you are
+ * looking at, and undetectable by the both-blocks assertion because the key
+ * sets would still match. `foundation.test.ts` pins this decision from both
+ * sides so the next reader finds a rule rather than an accident.
+ * ============================================================= */
+:root {
+  /* Size / line-height per step. Whole pixels throughout: a fractional size
+   * reflows every dense table for no legible gain, which was tested at 100%,
+   * 125% and 150% zoom before these landed. Tracking lives in the @theme
+   * registration above, positive on the small steps where tight setting goes
+   * muddy and negative on the large ones where it goes loose. */
+  --text-xs: 12px;
+  --text-xs--line-height: 18px;
+  --text-sm: 14px;
+  --text-sm--line-height: 20px;
+  --text-base: 16px;
+  --text-base--line-height: 24px;
+  --text-lg: 18px;
+  --text-lg--line-height: 26px;
+  --text-xl: 22px;
+  --text-xl--line-height: 28px;
+  --text-2xl: 26px;
+  --text-2xl--line-height: 32px;
+
+  /* Hierarchy comes from size and tracking, so the weight axis is two values
+   * doing real work instead of four crowded together. 550 is the one emphasis
+   * weight; 400 is the floor and is where nearly everything sits.
+   *
+   * `--font-weight-medium: 400` is deliberate and the name is now a lie. It is
+   * how ~140 `font-medium` call sites and HeroUI's own 93 internal uses come
+   * off 500 in two lines rather than in 140 edits, and it is the largest
+   * single visual change here: before it, no text in the shell chrome was
+   * regular, so weight could not signal anything because every row had already
+   * spent it. Deleting the now-inert utilities is follow-up, not this change. */
+  --font-weight-normal: 400;
+  --font-weight-medium: 400;
+  --font-weight-semibold: 550;
+  --font-weight-bold: 600;
+  --tracking-tight: -0.015em;
 }
 
 /* =============================================================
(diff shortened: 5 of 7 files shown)


web/src/styles/globals.css
```diff
@@ -555,6 +555,97 @@
...
+ *
+ * Deliberately NOT in the light and dark theme blocks. Type metrics have no
+ * theme axis, and declaring twenty-odd of them twice creates a failure the
+ * single declaration cannot have: one block edited and the other missed, giving
+ * the two themes different type metrics, invisible in whichever theme you are
+ * looking at, and undetectable by the both-blocks assertion because the key
+ * sets would still match. `foundation.test.ts` pins this decision from both
+ * sides so the next reader finds a rule rather than an accident.
+ * ============================================================= */
+:root {
```
jigjigjig, line 611: Self-review, and this is the load-bearing detail of the whole PR, so it is worth stopping on.

This block is unlayered on purpose and it must stay that way. `@heroui/styles` is a *prebuilt* Tailwind stylesheet: it ships its own `@layer theme { :root, :host { ... } }` carrying `--text-sm`, `--font-weight-medium: 500`, `--tracking-tight` and the whole default theme. Tailwind hoists every `@theme` block to the position of `@import "tailwindcss"` on line 1, so HeroUI's theme layer lands second. Same layer, same specificity, so source order decides and HeroUI wins.

I verified this by compiling rather than reasoning about it. Written inside `@theme`, the block emits and then loses:

```
--text-xs: 12px      <- ours
--text-xs: .75rem    <- HeroUI, later, wins
```

It compiles clean, passes lint and typecheck, and changes nothing. The half that makes it dangerous is that the *new* keys survive, because HeroUI never declares them, so the tracking would have landed while every size and weight silently reverted and it would have looked like it half-worked.

The reason it is unlayered rather than simply placed after the import: an unlayered declaration beats every `@layer` unconditionally, so it wins because of what it is. Winning on source order alone would make the next `@import` somebody adds a silent regression. `foundation.test.ts` pins this from both directions.

web/src/styles/globals.css
```diff
@@ -555,6 +555,97 @@
...
+   * weight; 400 is the floor and is where nearly everything sits.
+   *
+   * `--font-weight-medium: 400` is deliberate and the name is now a lie. It is
+   * how ~140 `font-medium` call sites and HeroUI's own 93 internal uses come
+   * off 500 in two lines rather than in 140 edits, and it is the largest
+   * single visual change here: before it, no text in the shell chrome was
+   * regular, so weight could not signal anything because every row had already
+   * spent it. Deleting the now-inert utilities is follow-up, not this change. */
+  --font-weight-normal: 400;
+  --font-weight-medium: 400;
```
jigjigjig, line 649: `--font-weight-medium: 400` is the single largest visual change here and the token name is now false. Flagging it rather than hoping a reviewer reads the comment.

The trade: this drops roughly 140 `font-medium` call sites plus HeroUI's own 93 internal uses from 500 to 400 in one line, instead of 140 edits. Before it, **nothing in the shell chrome was regular weight**, which is why weight could not signal anything: the selected nav row had to be carried entirely by its background fill because all nine rows were already at 500.

Deleting the now-inert `font-medium` utilities is tracked in #804 rather than done here, because it is a 140-site sweep and this PR is deliberately bounded to the stylesheet plus the shared primitives.

web/src/styles/globals.css
```diff
@@ -119,7 +119,7 @@
   /* ── Text — brand black + cool neutral muted ─────────────────────────── */
   --color-text: #1a1a1a; /* primary body and headings (Otari brand black) */
   --color-text-muted: #545b62; /* secondary text, labels, captions */
-  --color-text-subtle: #80868d; /* tertiary, timestamps, helper text */
+  --color-text-subtle: #62686f; /* tertiary, timestamps, helper text */
```
jigjigjig, line 122: These two token changes (light here, dark at line 359) are the `fix` in the title, and they are a pre-existing defect rather than fallout from the retune. Both current values fail WCAG AA for normal text on **all six** grounds each theme declares: light worst case 3.01, dark 2.91.

Worth knowing how it hid for so long: the dashboard had no regular-weight text, and weight 500 reads as more contrast than the ratio gives. Dropping the floor to 400 did not break this, it removed what was disguising it.

The full six-surface table for both themes is in the PR description and in #803. The pairing I would draw a reviewer's attention to is `background-subtle`, whose own comment above calls it "hover/pressed on muted": that is the hovered nav row, `text-chrome-meta` sits on it, and at 2.91 the worst-contrast state of that component was one a person reaches by moving the mouse.

Every ratio was computed twice independently. The first pass sampled three of the six grounds and reported a dark repair of `#87919b` as passing; it fails at 4.25 and 4.08 on the two deepest surfaces. `#919ba5` clears 4.5 everywhere with its worst case at 4.63, within 0.02 of light's 4.61, so neither theme rides the line.

jigjigjig (commented):
Self-review before anyone else reads this. Nine comments on the places I would stop if this were somebody else's diff: the two cascade bugs, the token whose name is now false, the contrast repair and how it was verified, the user-visible copy change, and the one test that reads like a convention being evaded.

Nothing in this pass changed the diff. Every claim below is a measurement from the running dashboard or from the compiled stylesheet, and where a measurement contradicted a plausible assumption I have said which won.

web/src/styles/globals.css
```diff
@@ -691,13 +816,23 @@ body {
...
-  font-family: var(--font-display);
+/* In `@layer base` rather than unlayered, which is a cascade fix and not a
+   tidy-up: `@utility` lands in `@layer utilities`, and an unlayered rule beats
+   every layer, so while this was unlayered it silently overrode the
+   `font-family` of any role applied to a heading element. `text-heading` and
+   `text-title` are set in the body face now, and both are worn by `<h2>` and
+   `<h3>`, so unlayered here would have kept them in Zilla Slab and made the
+   role a suggestion. Layered, a bare heading still gets the display face and a
+   role still wins. */
+@layer base {
```
jigjigjig, line 836: Second cascade bug, same family as the `@theme` one and found the same way.

This rule was **unlayered**, and `@utility` lands in `@layer utilities`. An unlayered rule beats every layer, so while it sat outside one it silently overrode the `font-family` of any role applied to a heading element. That matters now specifically: under the new map `text-heading` and `text-title` are set in the body face, and both are worn by `<h2>` and `<h3>`, so this would have kept them in Zilla Slab and left the role map as a suggestion with no symptom to chase.

Wrapped in `@layer base`, a bare heading still gets the display face and a role now wins. Behavior for every existing heading with no role on it is unchanged.

web/src/styles/globals.css
```diff
@@ -904,8 +1039,21 @@ h6 {
...
+  /* The single overline spec, matching `@utility text-overline`, because a
+     column header is a label above a group and that is exactly the role. It
+     replaces a hardcoded `font-weight: 600` that no token could reach: with the
+     rest of the scale down at 400/550 that literal had become the heaviest text
+     in the product, so the header row out-weighted every cell it labeled. A
+     header is now a point smaller, 50 units lighter, muted, and letterspaced,
+     which is what stops it competing with its own rows. Kept as a rule here
+     rather than a class at the call site because these three selectors reach
+     into HeroUI's table DOM, which no call site can. */
+  color: var(--color-text-muted);
```
jigjigjig: This replaces a hardcoded `font-weight: 600` that no token could reach, which is why it needed a call-site-shaped fix inside a stylesheet rule. With the rest of the scale down at 400/550 that literal had become the heaviest text in the product, so a column header out-weighted every cell it labeled.

One consequence a reviewer should see rather than discover: **uppercase costs column width.** Measured on the Models table at 1280px, uppercase plus 0.06em took 26px off the Model column and wrapped both the header row and the longest identifier. Rather than weaken the spec, three header strings lost a dead word each (`Caching policy` to `Caching`; every value in that column is a policy). Result: header row 56px to 41px, Model column 354px to 389px, nothing wraps.

`APPLIES TO` on Routing does still wrap to two lines, and that is accepted deliberately. The shorter candidate was `SCOPE`, which is the column's internal id at `RoutingPage.tsx:1335`, so it is our vocabulary rather than the reader's. Routing has two wraps that predate this change as well, so a naming edit would have fixed one third of a column-width problem. All three are in #804.

web/src/shared/components/ui.tsx
```diff
@@ -180,11 +180,16 @@ export function StatCard({
...
       <span className="flex flex-wrap items-center gap-2">
-        <span className="text-xl font-semibold text-foreground sm:text-2xl">
+        {/* text-xl (22px), deliberately a step *below* the page title's
+            text-display (26px). It used to be text-xl rising to text-2xl at
+            `sm`, which made a number inside a card the largest text on the
+            page, 4px bigger than the name of the page itself. Fixing that by
+            raising the title alone would have left the two agreeing by
+            coincidence; both halves move so the ladder is correct by
+            construction. `tabular-nums` so a column of these aligns. */}
+        <span className="text-xl font-semibold tabular-nums text-foreground">
```
jigjigjig, line 192: Both halves of this move, and that is the point rather than an accident.

The page title was `text-xl font-semibold` (20px) while this stat value was `text-xl ... sm:text-2xl` (24px), so a number inside a card was the largest text on the page, 4px bigger than the name of the page. Raising the title alone would have fixed the symptom and left the two agreeing by coincidence. The title goes up to `text-display` (26) and this comes down to `text-xl` (22), so the ladder is ordered by construction.

Verified on the running app: `h1` measures Zilla Slab 26px/500, this measures 22px/550.

A side effect worth naming because it looks like luck: the `Over budget` chip on the Overview budget card had been wrapping onto its own line, making that card taller than its neighbours. It came back inline with no work once this took its correct step. That is usually the tell that the number was right rather than tuned.

web/src/shared/components/ui.tsx
```diff
@@ -195,7 +200,16 @@ export function StatCard({
...
         ) : null}
       </span>
-      {hint ? <span className="text-xs text-muted">{hint}</span> : null}
+      {/* Two lines reserved, always. A delta like "5.8% errors · ▲ 214.1% vs
+          prev" wraps in a narrow tile while its neighbours stay on one line,
+          and a row of five tiles then stops aligning. This is a layout fix and
+          not a type fix on purpose: sizing the type to the longest string would
+          be sizing the type by accident. min-h-9 is two 18px lines. */}
+      {hint ? (
+        <span className="block min-h-9 text-xs tabular-nums text-muted">
```
jigjigjig, line 209: This is a layout fix wearing a type problem's clothes, and it is the clearest of the three in this PR.

A delta like `5.8% errors / 214.1% vs prev` wraps in a narrow tile while its four neighbours stay on one line, and a row of five stat cards then stops aligning. The tempting fix is a smaller font. The fix is reserving the two lines. `min-h-9` is two 18px lines at the 12px step.

`tabular-nums` here is the one place I added it. A DOM scan for numeric text lacking tabular figures found 11 hits on Usage and 5 on Overview, and **every one was outside a table cell**: `DataTable` already applies it to each `align: "end"` column. What was left is numbers inside sentences, where proportional figures are correct. A column of five percentages, one per card, is the exception that genuinely aligns.

web/src/features/models/ModelsPage.tsx
```diff
@@ -1304,7 +1302,7 @@ function ModelTable({
         id: "input",
         header: (
           <span className="inline-flex items-center gap-1">
-            {`${comparisonLabel} in / out $ / 1M`}
+            {`${comparisonLabel} in / out / 1M`}
```
jigjigjig, line 1305: User-visible copy change, so flagging it explicitly: `Base in / out $ / 1M` becomes `Base in / out / 1M`.

The `$` comes out because every cell in this column already shows it. `/ 1M` stays, and that is deliberate against the shorter cut that was proposed: the info tooltip on this column explains how *unpriced* models are metered and never states the unit, so dropping `/ 1M` too would have left the basis for `$0.90` stated nowhere on the page, and a reader could not tell per-million from per-thousand. I read `PricingInfo` before cutting rather than assuming the tooltip carried it.

Keeping `/ 1M` costs nothing: the 389px Model column measured in the header comment is measured with it still in.

web/src/styles/foundation.test.ts
```diff
@@ -73,6 +73,103 @@ const HEROUI_VARIABLES = [
...
+      "--tracking-tight",
+    ]) {
+      expect(
+        TYPE_VALUES.get(key),
+        `${key} is not declared in the unlayered :root block, so @heroui/styles' own value wins`,
+      ).toBeDefined()
+    }
+  })
+
+  it("keeps type metrics out of the theme blocks", () => {
```
jigjigjig, line 206: This test will read like a rule being evaded unless the reasoning is visible, so: it is asserting that type metrics are declared **once**, outside both theme blocks, and that is a deliberate departure from the file's own both-blocks convention.

That convention exists for a colour reason, stated in the sibling test's docstring: a token in one block and missing from the other falls back to the light value in dark mode and reads as a contrast bug three pages away. Type has no theme axis, so none of that applies.

What declaring them twice *would* buy is a new failure that cannot happen today: twenty-odd paired declarations where someone edits `--text-sm` in one block and misses the other, giving the two themes different type metrics. That is invisible in whichever theme you are looking at, and the existing both-themes assertion cannot catch it because the key sets would still match.

So rather than route around the convention silently, the decision is pinned from both sides: this test says the metrics must not be in the theme blocks, the one above says the unlayered block must declare all of them, and the one below says `@theme` must not register a value HeroUI also declares. The third is the guard against reintroducing the silent no-op that this PR exists to fix.

web/src/shared/components/ui.tsx
```diff
@@ -195,7 +200,16 @@ export function StatCard({
           </span>
         ) : null}
       </span>
-      {hint ? <span className="text-xs text-muted">{hint}</span> : null}
+      {/* Two lines reserved, always. A delta like "5.8% errors · ▲ 214.1% vs
+          prev" wraps in a narrow tile while its neighbours stay on one line,
```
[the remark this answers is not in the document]
jigjigjig: Fixed in 0724d255. Changed to `neighbors`. You are right that it was inconsistent within the file as well as against the repo rule: `ui.tsx` already had `neighbors` at line 776 and `neighboring` at 218. I also swept the same word out of the PR description and the follow-up issue drafts.

web/src/features/activity/ActivityTimeline.tsx
```diff
@@ -275,7 +275,7 @@ export function ActivityTimeline({
       <div className="rounded-xl border border-border bg-surface p-2">
         <div className="flex items-center justify-between gap-2 px-1 pb-1">
           <span className="flex items-center gap-3">
-            <span className="text-[11px] font-medium uppercase tracking-wide text-muted">
+            <span className="text-overline">
```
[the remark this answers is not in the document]
jigjigjig: Fixed in 0724d255. `text-[11px]` on the zoom hint is now `text-xs`, so it is 12px from the scale and matches the `text-overline` three lines above it rather than sitting a pixel under it.

Good catch on the location: it is inside a hunk this branch touches, so leaving it left the change half done.

Two more arbitrary sizes remain in `ActivityPage.tsx` (`text-[10px]` at 861, `text-[11px]` at 1882). Both are outside any hunk this branch touches, and they are part of the call-site migration tracked in #804 rather than something to sweep here. Flagging them explicitly so they are not mistaken for an oversight.

jigjigjig:
Correcting a claim I made in the description, flagged here rather than left as a silent body edit.

The **Deliberately not in this change** section asserted that five meanings share 14px and that Activity renders its explanation line at the same size as the value it explains. Both are false and the error is mine.

Measured in the live DOM on a seeded database:

    "demo-guarded"                            14px/400  #1a1a1a
    "the policy's default target"             12px/400  #545b62
    "attempt 2 of 2 failed, plan exhausted"   12px/400  #545b62

`ActivityPage.tsx:749` has carried `text-xs ... text-muted` throughout, so that size step predates this work. And classifying every 14px node on that page by structure: 386 of ~427 are table cells, which is what body size is for on a 161-row table.

Where it came from: a computed-style census bucketed `14px/400/lh:17.5px` and printed three samples, all of which happened to be the policy name. I read the bucket as covering both lines of the cell and never probed the second line by name.

**Nothing else in the description rests on it.** The weight floor, the per-step tracking, the contrast repair, the two cascade bugs and the title/stat ladder were each measured directly and stand. #804 has been corrected the same way, leading with the retraction, and its scope is now the smaller set of findings that survived re-measurement.

web/src/features/models/ModelsPage.tsx
```diff
@@ -1326,7 +1324,10 @@ function ModelTable({
       },
       {
         id: "caching",
-        header: `Caching ${comparisonContextTokens == null ? "policy" : comparisonLabel}`,
+        header:
+          comparisonContextTokens == null
+            ? "Caching"
+            : `Caching ${comparisonLabel}`,
```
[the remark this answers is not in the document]
jigjigjig: Addressed in 9521fd6e. Updated to `Caching`, and the other two in the same file: `Base in / out $ / 1M` to `Base in / out / 1M` and the `at 200K` variant of the same header.

Worth recording that you caught this before I did. CI caught it too, but my local run could not: the suite fails 303 tests on Node 26 for an unrelated reason (#805), so three real failures were invisible inside the noise.

web/src/shared/components/ui.tsx
```diff
@@ -195,7 +200,16 @@ export function StatCard({
...
+          and a row of five tiles then stops aligning. This is a layout fix and
+          not a type fix on purpose: sizing the type to the longest string would
+          be sizing the type by accident. min-h-9 is two 18px lines. */}
```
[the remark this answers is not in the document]
jigjigjig: Addressed in 0724d255. `neighbours` to `neighbors`. The file already used `neighbors` and `neighboring` elsewhere, so it was inconsistent internally as well as against the repo rule. Also swept it out of the PR description and the follow-up issue drafts.

jigjigjig:
## For whoever merges this: two things to do, in this order

**1. Retarget the stacked PR first.** #810 is based on `fix/type-scale`, not `main`. This repo has `delete_branch_on_merge` enabled, and merging a parent with its branch deleted while a child still points at it **closes the child permanently**: GitHub refuses to reopen a PR whose base branch was gone (422 `state cannot be changed`), even if the branch is recreated. So before merging this one:

    gh pr edit 810 --base main

There is no rush: see (3) below.

**2. Squash-commit line, if the release note is being written by hand.** This PR and #810 are two halves of one change, and #810 is `refactor`, which `cliff.toml` skips. So the changelog will carry this PR's line and nothing about the headings, even though every card title in the dashboard changes size. Suggested line so the visible half is not split from the invisible one:

> Repaired the dashboard's tertiary text contrast, which failed WCAG AA on every surface in both themes, and retuned the type scale it sits in: hierarchy now comes from size and tracking with one emphasis weight rather than from boldness, and headings that had been rendering at body size now sit above it.

**3. State right now:** all checks green, all 13 review threads resolved, both bots (Copilot and CodeRabbit) through and addressed. `mergeStateStatus` is `BLOCKED` for one reason only: the `protect-main` ruleset requires **one approving human review**, and a bot's `COMMENTED` review does not count. Repository admins are bypass actors, so a merge is mechanically possible without one, but the rule is asking for a human to look. #811 documents this, because the skill that describes this repo's merge mechanics said `main` was unprotected.

web/src/styles/globals.css
```diff
@@ -555,6 +555,97 @@
...
+   *   --text-*--letter-spacing  tracking per size step. Registration is what
+   *                           makes Tailwind emit the `letter-spacing` line in
+   *                           `.text-xs` and friends at all; a value declared
+   *                           only at :root is never read, because the utility
+   *                           would not reference it. So these must be here and
+   *                           the sizes must not (see the :root block below).
+   *
+   * Every value that HeroUI also declares lives in the unlayered `:root` block
+   * further down, NOT here. The comment there says why. */
+  --text-caption: 13px;
```
khaledosman: Registering `--text-caption` in the `--text-*` namespace makes Tailwind generate a `.text-caption` **size** utility that collides with the `@utility text-caption` **role** below (line 748).

Compiled output of this branch has `.text-caption` twice inside `@layer utilities`, the `@utility` copy first and the theme-derived copy second:

```css
.text-caption { font-family: var(--font-sans); font-size: var(--text-caption); line-height: var(--text-caption--line-height); letter-spacing: var(--text-caption--letter-spacing); font-weight: var(--font-weight-normal); color: var(--color-text-muted); }
/* ... */
.text-caption { font-size: var(--text-caption); line-height: var(--tw-leading, var(--text-caption--line-height)); letter-spacing: var(--tw-tracking, var(--text-caption--letter-spacing)); }
```

Same layer, same specificity, so the later one wins and the `font-size` / `line-height` / `letter-spacing` lines inside `@utility text-caption` are dead. The values agree today, so nothing looks wrong. Edit the `@utility`'s size, though, and it will compile, emit, pass every check, and change nothing: exactly the failure mode this PR's three new tests exist to catch. And `registers in @theme only the keys HeroUI does not declare` explicitly whitelists `--text-caption`, so the test cannot see it either.

Two ways out: name the step outside the utility namespace (`--text-caption-step`), or drop the three metric lines from the `@utility` and let the theme copy own them, with a comment saying which half owns metrics.

web/src/styles/globals.css
```diff
@@ -596,60 +687,94 @@
...
+  letter-spacing: var(--text-caption--letter-spacing);
+  font-weight: var(--font-weight-normal);
   color: var(--color-text-muted);
 }
 @utility text-overline {
-  font-size: 0.6875rem; /* 11px */
-  font-weight: 600;
-  line-height: 1.2;
+  /* 12px rather than 11: of the three hand-rolled overline spellings this
+     replaces, the most common one was already 12, and an eighth theme step for
```
khaledosman: The measurement behind this is wrong. Of the seven hand-rolled overline sites this replaces, **five** were `text-[11px]`:

```
features/activity/ActivityPage.tsx      4  text-[11px] font-medium uppercase tracking-wide text-muted
features/activity/ActivityTimeline.tsx  1  text-[11px] font-medium uppercase tracking-wide text-muted
features/models/ModelsPage.tsx          1  text-xs font-semibold uppercase tracking-wide text-muted
shared/components/ui.tsx                1  text-xs font-medium uppercase tracking-wide text-muted
```

So the most common spelling was 11, not 12, and the three pre-existing `text-overline` consumers (`AppShell`, `nav/WorkspaceSwitcher`, `nav/rowStyles`) go **up** a point rather than staying put. 12 is still the right call, since an eighth step for one role is not worth a point of size, but the reason stated here is not the reason.

web/src/styles/globals.css
```diff
@@ -904,8 +1039,21 @@ h6 {
   > [role="presentation"]:last-of-type
   > .table__column {
   background-color: transparent;
-  color: var(--color-text);
-  font-weight: 600;
+  /* The single overline spec, matching `@utility text-overline`, because a
```
khaledosman, line 1051: This is a second copy of the overline spec, not "the single overline spec". Six declarations are duplicated from `@utility text-overline` (line 749): `font-size`, `line-height`, `letter-spacing`, `font-weight`, `text-transform`, `color`. Retune the role and the header row silently keeps the old metrics, with no test and no visible symptom pointing here.

`@apply text-overline` inside this rule reaches the values and keeps the part that genuinely cannot be a call-site class: the three selectors into HeroUI's table DOM, which is the justification the comment already gives for the rule existing at all.

web/src/styles/foundation.test.ts
```diff
@@ -73,6 +73,103 @@ const HEROUI_VARIABLES = [
...
+describe("the type scale's two halves", () => {
+  // The scale is split on purpose and the split is easy to undo by accident, so
+  // it is pinned from both sides. `@heroui/styles` is a prebuilt Tailwind
+  // stylesheet carrying its own `@layer theme` with the default `--text-*`,
+  // `--font-weight-*` and `--tracking-*` in it, and it is imported after
+  // Tailwind, so an override of one of those written in `@theme` loses on source
+  // order: it compiles, emits, and changes nothing. The values therefore live in
+  // an unlayered `:root` block, which beats every layer, and only the keys
+  // HeroUI never declares are registered in `@theme`.
+  const TYPE_VALUES = declarations(block(":root"))
```
khaledosman, line 167: Two things about this block.

**The headline fix ships unguarded.** Three new assertions pin the cascade split; nothing pins `--color-text-subtle` (globals.css:122 and :359). The WCAG AA repair regresses silently on the next palette edit, which is the same invisible-until-rendered class of failure the rest of this PR is about. The description says the token was already measured against all six background and surface tokens each theme declares, and that measurement is the assertion. A relative-luminance helper over those pairs is roughly fifteen lines, and `LIGHT` / `DARK` are already parsed right here.

**`block(":root")` is a first-match `indexOf(":root {")`, and `globals.css` now has two top-level `:root {` blocks**: the type scale at 615 and the pre-existing base reset at 785. It resolves correctly today only because the type scale happens to come first. Move the reset above it and the failure message (`is not declared in the unlayered :root block`) points at the wrong block. Worth anchoring on something distinguishing, or asserting up front that the matched block is the one carrying `--text-xs`.

web/src/features/models/ModelsPage.tsx
```diff
@@ -1338,7 +1339,7 @@ function ModelTable({
       },
       {
         id: "policy",
-        header: "Pricing policy",
+        header: "Pricing",
```
khaledosman: Three user-visible column labels change here and the description mentions none of them: `Pricing policy` to `Pricing`, `Caching policy` to `Caching` (line 1329), and `$` dropped from `Base in / out $ / 1M` (line 1305).

Dropping `$` is fine, since `formatCost` puts it in every cell. `Caching` is fine. `Pricing` is not: it now sits two columns from `Base in / out / 1M` and reads as a second price column rather than as the policy that produced the price. Keep `Pricing policy`, or if uppercase plus 0.06em tracking forced the headers shorter, say so in the description. A type PR silently renaming product copy is the part that is hard to review.

khaledosman (approved):
Verified the central cascade claim by compiling this branch's `globals.css` through Tailwind's own `compile()` API with `@heroui/styles` resolved as it is at runtime, then reading the emitted CSS: HeroUI does ship `@layer theme` with `--font-weight-medium: 500`, and the unlayered `:root` block does beat it (the `400` lands last, unlayered). `@theme` registration is what makes `.text-xs` and friends emit a `letter-spacing` line at all. The diagnosis and the split are right, and `foundation.test.ts` pinning it from both sides is the right shape.

Also confirmed `Mozilla Text` is a variable face at `font-weight: 200 700`, so `550` is a real instance rather than a value that snaps to a static cut.

Five inline: one silent no-op the split does not cover, one duplicated spec, one comment whose measurement is wrong, the contrast fix shipping without an assertion, and an undocumented copy change.

---

web/src/styles/globals.css
jigjigjig: You are right, and this is the one that matters most: it is the same silent no-op the PR exists to prevent, sitting inside the fix.

I had checked this collision earlier and concluded there was no problem, because I only verified that the role's `color` survived the merge. It does. Its **metrics** do not. Proved it properly this time by setting the role to `font-size: 99px` and compiling: the emitted rule carries both, and `var(--text-caption)` still wins.

Fixed in 1febc0e7 by your first option, renaming the step out of the utility namespace: `--text-caption-step` in the unlayered value block, where it generates no utility, so no collision can exist and the role owns its whole spec in one place. I preferred it over dropping the role's metric lines because a role should be readable as a complete definition.

It made the test stricter as a side effect, which is the part I like: `registers in @theme only the keys HeroUI does not declare` had to whitelist `--text-caption` and `--text-caption--line-height`. Both exceptions are gone, so the assertion is now flatly 'letter-spacing keys only, no exceptions', and a size registered there fails.

web/src/styles/globals.css
jigjigjig: Correct, and my comment asserted the opposite of the data. Five of the seven were `text-[11px]` and two were `text-xs`, so 11 was the more common spelling.

Rewritten in 1febc0e7 to give the real reason (an eighth theme step for a single role is not worth a point of size, and `--text-xs` is already exactly this size) and to state the consequence I had left out: the three sites that already used the role, `AppShell`, `nav/WorkspaceSwitcher` and `nav/rowStyles`, go **up** a point rather than staying put.

Thank you for checking the count rather than the claim.

web/src/styles/globals.css
jigjigjig: Agreed, and `@apply` is the right instrument. Fixed in 1febc0e7: the rule is now `@apply text-overline` plus `background-color` and `border-radius`, so the role is the single definition and retuning it moves the header row with it. What stays is the only thing that genuinely cannot be a call-site class, the three selectors into HeroUI's table DOM, which is the justification the comment already gave.

Verified in the compiled output that the role's declarations land in the rule rather than `@apply` silently resolving to nothing.

web/src/styles/foundation.test.ts
jigjigjig: Both right, and the first is the fairest hit in the review: the headline fix was the one thing shipping unguarded.

Added in 1febc0e7. Three inks (`--color-text`, `-muted`, `-subtle`) against all six grounds each theme declares, both themes, so 36 assertions, using a WCAG 2.x relative-luminance helper inline rather than a dependency. **Confirmed non-vacuous**: putting `#80868d` back fails with `--color-text-subtle on --color-background is 3.68:1, under the 4.5:1 AA floor for normal text`. It also covers `--color-text` and `-muted`, which were already fine, so the floor is now pinned for the whole ink family rather than just the token this PR moved.

On `block(":root")`: also right, and it resolved correctly only by accident of ordering. Rather than re-anchor the matcher and risk being clever, there is now an explicit assertion that the matched block is the one carrying `--text-xs`, so a reordering fails on a message that says so instead of misdirecting every assertion below it.

web/src/features/models/ModelsPage.tsx
jigjigjig: Two halves here, and I am leaving this thread open because the second one is your call rather than mine.

**The description gap is real and fixed.** All three renames are now in the description under the column-headers bullet, with the reason (uppercase plus 0.06em tracking made every header wider, which took 26px off the Model column and wrapped both the header row and the longest identifier) and the measurement.

**On `Pricing` specifically, you are right about the reading and I do not have a free fix.** Measured on the page at 1280px:

    Pricing            Model column 389px   nothing wraps
    Pricing policy     Model column 365px   `Base in / out / 1M` AND `Pricing policy` both wrap

So restoring `Pricing policy` costs 24px off the Model column and re-introduces two wraps on the densest table in the product. That is the trade, and it is a real one rather than me defending the short version.

You offered two ways out: restore it, or say in the description why the headers got shorter. I have done the second, and I do not think it settles your objection: a header two columns from `BASE IN / OUT / 1M` reading as a second price column is a readability defect I introduced, and explaining it in a PR description does not fix it for the operator.

So I would rather propose a third option than pick from the two: **`POLICY`**. One word, so it keeps the width, and it cannot read as a price. It is also arguably more accurate than `Pricing policy` was, since the column's values are `Base only`, `Tiered` and so on, which are policies rather than prices.

I have **not** made that change, for two reasons. It is a product-naming decision rather than a typography one, and this is a type PR, which is exactly the objection you raised. And I could not measure `POLICY` reliably in-page (swapping header text in the DOM re-flows the table and makes the last column report as wrapping regardless of content), so I would want to verify it properly against a real build before claiming it fits.

Say which you want and I will do it: restore `Pricing policy` and accept the two wraps, keep `Pricing` with the description note, or `POLICY` with a proper measurement first. Leaving this unresolved for you.

jigjigjig: Restored in a1d22072. `Pricing policy` is back; `Caching` stays short.

Your objection was right and my proposed alternative was wrong for a reason I had not seen. **`POLICY` breaks a parallel.** `Caching policy` is already shortened to `Caching`, so `CACHING` beside `POLICY` gives a reader two different categories where there is one kind of thing, which trades a local ambiguity for a structural one. If that pair is ever shortened it has to be shortened as a pair, and that is a naming exercise with a product owner in it rather than a line in a type PR. I would not have got there from the single header.

This also makes us consistent with the `APPLIES TO` decision on Routing, where we took a wrapped header over a rename on the same grounds. The asymmetry is what decides both: a wrapped header costs a reader a glance, a misleading header costs them a wrong conclusion about what a column holds.

**The measured cost, stated fully, because it is three things and not one:**

    Model column        389px -> 365px
    header row height    41px ->  56px
    wraps                none -> `Base in / out / 1M` and `Pricing policy`
    longest identifier  one line -> wraps (`anthropic.claude-3-5-sonnet-20241022-v2:0`)

That last one I had not priced when I offered the trade: the 24px lost from the Model column re-wraps the longest model id, so restoring the header costs a data-cell wrap as well as two header wraps. It does not change the call, but you should have the whole number rather than the part that supports it.

The description now carries all of this in one place beside the `APPLIES TO` note, so the two read as one consistent decision about uppercase costing width rather than two ad hoc ones, and it records that `POLICY` was considered and why it was rejected.

Leaving this for you to resolve or push back on.

web/src/styles/foundation.test.ts
```diff
@@ -73,6 +73,189 @@ const HEROUI_VARIABLES = [
...
+      for (const ground of GROUNDS) {
+        const fg = tokens.get(ink)
+        const bg = tokens.get(ground)
+        expect(fg, `${ink} is not declared`).toBeDefined()
+        expect(bg, `${ground} is not declared`).toBeDefined()
+        const ratio = contrast(fg as string, bg as string)
+        expect(
+          Number(ratio.toFixed(2)),
+          `${ink} on ${ground} is ${ratio.toFixed(2)}:1, under the ${AA_NORMAL}:1 AA floor for normal text`,
+        ).toBeGreaterThanOrEqual(AA_NORMAL)
```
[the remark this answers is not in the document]
jigjigjig: Correct, and it is a worse bug than minor. A guard that goes green on a failing value is worse than no guard, and this one was added in this very change to stop that class of thing.

Fixed in beb3252c. Comparing `ratio` directly; the rounding stays in the failure message only.

Demonstrated the dead zone rather than taking it on faith: `#547e7f` on `#ffffff` is **4.4989**, which `toFixed(2)` turns into `4.50`, so the old form accepted a value 0.0011 under the floor. Any ratio in [4.495, 4.5) had the same hole.

Two of these in one PR now: a size utility silently outranking the role that owns its name, and a test rounding its way past its own threshold. Both were invisible without compiling or arithmetic, which is the argument the PR is making.

web/src/features/models/ModelsPage.tsx
jigjigjig: Resolving this now rather than leaving it open, since the question in it has been answered and the change is pushed: `Pricing policy` is restored in a1d22072, `Caching` stays short, and the reasoning for rejecting `POLICY` (it breaks the parallel with `Caching`, showing two categories where there is one) is on the thread above along with the full measured cost.

It is a decision with its reasoning on the record now rather than an open question. Please reopen if you disagree with any part of it; the argument is all here to push back on.

jigjigjig:
**Superseding the state paragraph in my earlier comment**, which is now stale. Items 1 and 2 there still stand; item 3 does not.

- The retarget instruction is **moot**: #810 now targets `main` directly, so the `delete_branch_on_merge` hazard is gone. Merge order still matters, but only for the changelog: **#807 before #810**, because #810 is `refactor` and `cliff.toml` skips those, so merging it first would land the type scale on `main` under a title the release notes never show.
- The release-note line is unchanged and is the one above.
- **Current state: all checks green, all 19 review threads resolved, `CLEAN`, `MERGEABLE`, `APPROVED`.** The earlier note said BLOCKED pending a human approval; that approval is in, and what had actually been gating was thread resolution.

Issue #809, opened 2026-08-26
[BUG] A long database_url overflows its Settings row and squeezes the description column


jigjigjig:
**Scope:** dashboard (`web/`), Settings page. Repo-level, so the bug template's environment
fields do not apply.

**Check this first:** the value that triggered it in my environment is a pathologically long
SQLite path (a temp directory nested six levels deep). It may not reproduce with a normal
`database_url`, in which case this is worth closing. I am filing it rather than dropping it
because a real deployment can have a long value here too: a Postgres URL with a long host, or a
container path.

### What happened

On `#/settings`, under **Server & database**, a long `database_url` value overflows its row
horizontally: the value runs past the right edge of the card and past the page, and it squeezes
the neighbouring description column hard enough that `Database connection URL (SQLite default
for local use; PostgreSQL recommended for production)` wraps to roughly one word per line, about
twelve lines tall.

Every other row on the page is a single line. So one long value makes one row about ten times
taller than its neighbours and pushes content outside the viewport.

### Steps to reproduce

1. Run the gateway with a long `--database-url`, for example a SQLite path several directories
   deep.
2. Open `#/settings` and look at the `database_url` row under **Server & database**.

### What you expected

The value truncates, wraps inside its own column, or scrolls within the row. A settings row
should not be able to push the page wider than the viewport, and the description column should
not lose its width to it.

### Notes

The value is a path or a URL, so it has no spaces to wrap at, which is why it overflows rather
than wrapping. `min-w-0` on the flex child plus `break-all` or a `truncate` with the full value
in a `title` are the usual answers; which one depends on whether an operator needs to read the
whole value at a glance or only confirm it.

Not a type problem, and mentioned here only because it was found during typography work:
nothing about the scale causes or fixes it.

Issue #808, opened 2026-08-26
[BUG] Sidebar disclosure groups are <h3> elements, so they sit in every page's document outline


jigjigjig:
**Scope:** dashboard (`web/`), repo-level. Not a deployment problem, so the bug template's
environment fields do not apply.

### What happened

The sidebar's collapsible groups (`Routing`, `Tools`) render as real `<h3>` elements, so they
join the page's document outline. They are chrome rather than content, and they carry the nav's
own metrics rather than a content heading's, so the outline of a settings page reads:

    H1  26px/500   Settings
    H2  18px/550   Server & database (9)
    H2  16px/550   Email delivery
    H3  14px/400   Routing        <- sidebar group, not part of this page
    H3  14px/400   Tools          <- sidebar group, not part of this page

An `<h3>` at regular weight, smaller than the `<h2>` above it and the same size as body text,
sitting in the outline of every page regardless of what that page is about.

Found while putting content headings onto the type roles (#807, and the follow-up that does the
call sites). It is not caused by that work: the nav's metrics are deliberate and correct for a
rail row. The question is whether these should be headings at all.

### What you expected

Either a heading that reads as one in the outline, or not a heading. A screen-reader user
navigating by heading currently lands on two sidebar group labels between the page's own
sections.

### Suggestion, not a prescription

These are disclosure toggles: they expand and collapse a group of links. `<button
aria-expanded>` inside a `<nav>` describes that better than `<h3>` does, and it takes them out
of the outline without changing anything visual. But there is a real argument for keeping them
as headings (a nav landmark with labelled groups is navigable), in which case they want to be
consistent with the rest of the outline instead.

Whoever picks this up should decide which, rather than treating it as a styling fix. Nothing
about the type scale forces the answer.

PR #810, opened 2026-08-26
refactor(dashboard): put headings that render at body size onto the type roles

jigjigjig:
### The collapse was real. We had it in the wrong place, twice.

#807 and an earlier version of #804 both claimed the dashboard had "five meanings sharing 14px"
and named the Activity table's secondary explanation line as the headline instance. **Both
claims were wrong**, they are retracted in those two places, and this PR is what the finding
actually was.

The explanation line was never 14px. It is 12px muted and `ActivityPage.tsx:749` has read
`text-xs break-words text-muted` throughout. And classifying every 14px node on Activity by
structure: 386 of ~427 are table cells, which is what body size is *for* on a 161-row table.

The real collapse was in the **headings**, and it was uniform:

    text-sm font-semibold text-foreground     35 sites, 26 files

That is the card-and-section-heading idiom of the whole dashboard, and `text-sm` is 14px. So
every card title and every section heading rendered at the same size as the body text inside
it, distinguished from its own content by weight alone. `text-title` (16) existed for exactly
this and had **zero consumers**.

Settings is the clearest before and after. Its document outline was:

    H1  26px/500   Settings
    H2  14px/550   Server & database (9)
    H2  14px/550   Metering & budgets (6)
    ... ten more, every one at body size

A 12px jump from the page title to the first heading, then nothing at all between a heading and
the rows it labels. It now reads:

    H1  26px/500   Settings
    H2  18px/550   Server & database (9)        <- section, heads a group of cards
    H2  16px/550   Email delivery               <- a card's own title
    body 14px/400

Four levels where there were two.

One limit on that, raised in review: **roughly fifteen of the 33 `text-title` sites are `<div>`s
rather than heading elements**, so the document-outline improvement the Settings example
describes reaches the `<h2>` subset and not all of them. They look right either way; the outline
only improves where the element was already a heading. The Settings example happens to be all
`<h2>`, which is what made the claim look general. Turning those `<div>`s into headings is a
separate change, and #808 covers the outline question.

### The split, and the rule behind it

35 sites, one substitution, two destinations. The rule is what the heading *heads*, not where it
sits in the file:

- **`text-title` (16/550), 33 sites.** A card's own title: it lives inside a bordered surface
  and titles that surface's contents.
- **`text-heading` (18/550), 2 sites.** An `<h2>` in a `<section>` that *contains* a `Card`. It
  heads a group of cards rather than titling one, which is the distinction the role names. Both
  are on Settings, where every section heading was previously the same size as the settings rows
  beneath it.

One corroboration worth noting: `shared/components/ui/SettingsSection.tsx` was already at
`text-base font-semibold`, which is 16/550, the exact metrics `text-title` names. Somebody had
reached for that size by hand in the one place the role map independently says 16. It now names
the role instead of restating the numbers.

### Two more from the same pass

- **Activity's per-attempt plan sub-table hand-rolled its column headers** as
  `<th className="px-3 py-2 text-left font-medium">`, so it missed the single overline spec
  `.otari-table .table__column` gives every other table on that page. Two tables on one screen
  labelling their columns differently. Six headers, now `text-overline`.
- **A `Popover.Heading` was at body size.** It is a dialog title; it takes `text-title`.

### Activity barely moved, and that is expected

Six sub-table headers and one popover title. **Its row heights are unchanged**, and they were
never going to change here: the rows are 79px median because the Routing column's explanation
renders into a 117px column, so a sentence of that length wraps three ways at any size in the
scale. Simulated before deciding: moving those lines to `text-caption` (13px) makes rows
*taller*, 79 to 81px median, and the wrap does not resolve.

So Activity's problem is column width, tracked in #804 along with the deliberate-wrapping
comment at `ActivityPage.tsx:748` that whoever picks it up should read first. Please do not
reopen Activity as a type question.

`text-caption` is therefore still unconsumed, deliberately. The sites that would want it are
already at 12px and moving them up a step costs wrapping. A step that exists and is not yet
needed is fine; manufacturing a consumer to justify it is not.

### Retargeted to `main`, and why the diff is bigger than the change

This PR was opened against `fix/type-scale` (#807) to keep the diff to its own single commit.
**I have retargeted it to `main`, because stacking turned out to cost far more than the clean
diff was worth.** A PR based on a topic branch in this repo gets:

- **no dashboard CI.** `otari-dashboard.yml`, `otari-lint.yml`, `otari-typecheck.yml`,
  `otari-tests.yml` and the rest all declare `pull_request: branches: [main]`, so **none of them
  ran.** No lint, no typecheck, no vitest, no build. Only the three repo-wide meta checks fired,
  and a green tick on those looks like a green PR.
- **no CodeRabbit review.** `.coderabbit.yaml` sets `base_branches: ["main"]`, so it posted
  "Review skipped" rather than reviewing.
- **no `protect-main`.** The ruleset covers the default branch only, so this reported
  `mergeStateStatus: CLEAN` and could have been merged into `fix/type-scale` with no approval and
  no thread resolution, folding two reviews into one.

CI is the check that caught the only real regression in #807, so trading three checks for a
tidier diff was the wrong trade and I made it. Documented in #811 so the next person does not.

**What this costs a reviewer, and how to work around it:** until #807 merges, the diff below
shows **four commits and 33 files**, three of which belong to #807. **This PR's change is the
single commit `5574e601`**, 27 files, all of them a one-line class substitution. Reviewing that
commit alone is reviewing this PR. Once #807 merges the diff collapses to just it.

## PR Type

- [ ] New Feature
- [ ] Bug Fix
- [x] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

`refactor` rather than `feat` or `fix`: nothing became newly possible and no defect is repaired
here. One note on the consequence, since it is a real one: `cliff.toml` has `refactor` as
`skip = true`, so this will not appear in the changelog even though it changes what every card
title in the product looks like. That is the right call given #807 carries the user-visible
claim and these two will land together, but it is worth knowing rather than discovering.

## Relevant issues

Part of #804. Deliberately not `Fixes`, because #804 also carries the column widths, the
descriptions the shared helpers do not reach, and identifiers in the mono face, none of which
are here.

Filed while doing this work and **not** fixed here: #808 (sidebar disclosure groups are `<h3>`,
so they sit in every page's document outline) and #809 (a long `database_url` overflows its
Settings row). Neither is caused by this change; both were visible because of it.

## Checklist

      No new tests, and that is the honest answer rather than a skipped box: this changes which
      utility class a heading carries. It changes no element, no text, no behavior, and no
      query. The suite queries headings by role and text, so it covers this change as-is, and
      the two `toHaveClass` assertions in the whole suite are both on `md:hidden`. Adding a test
      that asserts a class name would pin the implementation rather than the behavior and would
      have to be rewritten by the next retune. `web/src/styles/foundation.test.ts` already
      guards the scale these roles read.
      Those three are the Python side and this PR touches no Python. Their dashboard
      counterparts:
      - `pnpm --dir web run lint` clean.
      - `pnpm --dir web run typecheck` clean.
      - `pnpm --dir web test`: I compared the **failing-file sets** against this branch's base
        rather than comparing counts, because a count can hide a swap. **Zero newly failing,
        zero newly passing**, sets byte-identical. The 303 failures in both are the Node 26
        problem in #805, and CI on Node 22 is the real check.
      No doc change owed: the roles and their reasoning already live in `globals.css`, and this
      only adds consumers. The role-to-step map is in #804.
      **Not applicable.** No route, schema, or docstring changed. `web/src/client/schema.ts` and
      `web/src/routeTree.gen.ts` are unchanged.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

Same three-session arrangement as #807, and the same rule: a claim does not ship until it has
been measured by someone other than whoever proposed it.

That rule is why this PR exists in this shape. The migration this started as was a twelve-page
sweep based on the "five meanings on 14px" claim. Measuring the first page before editing it
showed the claim was false, that the sweep as specified would have made Activity's rows taller,
and that the real finding was a different one with 35 instances. The scope came out larger than
what was agreed (35 sites rather than 3) and that was a judgment call: a uniform idiom
half-migrated reads as a bug where all-14 read as a choice. It is one commit and revertible
whole.

**NOTE:**

### DIFF
--- a/web/src/features/activity/ActivityPage.tsx
+++ b/web/src/features/activity/ActivityPage.tsx
@@ -224,9 +224,7 @@ function InFlightControl({
       <Popover.Content placement="bottom end">
         <Popover.Dialog>
           <div className="flex w-80 flex-col gap-2">
-            <Popover.Heading className="text-sm font-medium">
-              In flight
-            </Popover.Heading>
+            <Popover.Heading className="text-title">In flight</Popover.Heading>
             <p className="text-xs text-muted">
               Running right now, across the whole gateway; longest-running
               first. Not narrowed by the filters above.
@@ -821,25 +819,25 @@ function RoutingPlan({ entry }: { entry: UsageEntry }) {
         >
           <thead className="text-muted">
             <tr className="border-b border-border">
-              <th scope="col" className="px-3 py-2 text-left font-medium">
+              <th scope="col" className="px-3 py-2 text-left text-overline">
                 #
               </th>
-              <th scope="col" className="px-3 py-2 text-left font-medium">
+              <th scope="col" className="px-3 py-2 text-left text-overline">
                 Target
               </th>
-              <th scope="col" className="px-3 py-2 text-left font-medium">
+              <th scope="col" className="px-3 py-2 text-left text-overline">
                 Selected as
               </th>
-              <th scope="col" className="px-3 py-2 text-left font-medium">
+              <th scope="col" className="px-3 py-2 text-left text-overline">
                 Outcome
               </th>
               {/* Every attempt's `latency_ms` is measured from the start of the
                   request, not from the start of that attempt, so this is the same
                   "Total time" the row column shows, not a per-candidate duration. */}
-              <th scope="col" className="px-3 py-2 text-right font-medium">
+              <th scope="col" className="px-3 py-2 text-right text-overline">
                 Total time
               </th>
-              <th scope="col" className="px-3 py-2 text-right font-medium">
+              <th scope="col" className="px-3 py-2 text-right text-overline">
                 Cost
               </th>
             </tr>
--- a/web/src/features/activity/ActivityPage.tsx
+++ b/web/src/features/activity/ActivityPage.tsx
@@ -817,7 +817,10 @@ function RoutingPlan({ entry }: { entry: UsageEntry }) {
           className="w-full text-xs"
           aria-label={`Routing plan for policy ${entry.policy_name}`}
         >
-          <thead className="text-muted">
+          {/* No `text-muted` here: `text-overline` on each `<th>` sets the color
+              itself, so a second declaration on the parent is one more place to
+              keep in step for no effect. */}
+          <thead>
             <tr className="border-b border-border">
               <th scope="col" className="px-3 py-2 text-left text-overline">
                 #
--- a/web/src/features/settings/SettingsPage.tsx
+++ b/web/src/features/settings/SettingsPage.tsx
@@ -564,7 +564,7 @@ function SecurityKeysSection({
 }) {
   return (
     <section className="flex flex-col gap-2">
-      <h2 className="text-sm font-semibold text-foreground">
+      <h2 className="text-heading">
         Credential security <span className="font-normal text-muted">(2)</span>
       </h2>
       <Card>
@@ -683,7 +683,7 @@ export function SettingsPage() {
 
       {groups.map((group) => (
         <section key={group.name} className="flex flex-col gap-2">
-          <h2 className="text-sm font-semibold text-foreground">
+          <h2 className="text-heading">
             {group.name}{" "}
             <span className="font-normal text-muted">
               ({group.fields.length})
--- a/web/src/features/usage/UsagePage.tsx
+++ b/web/src/features/usage/UsagePage.tsx
@@ -1315,7 +1315,7 @@ export function UsagePage() {
           <div className="grid gap-6 xl:grid-cols-2">
             <div className="flex flex-col gap-3">
               <div className="flex flex-wrap items-center justify-between gap-3">
-                <h2 className="text-sm font-semibold text-foreground">
+                <h2 className="text-title">
                   Spend by {activePrimary.label.toLowerCase()}
                 </h2>
                 <div className="inline-flex gap-1.5">
@@ -1348,7 +1348,7 @@ export function UsagePage() {
             </div>
             <div className="flex flex-col gap-3">
               <div className="flex flex-wrap items-center justify-between gap-3">
-                <h2 className="text-sm font-semibold text-foreground">
+                <h2 className="text-title">
                   Spend by {activeSecondary.label.toLowerCase()}
                 </h2>
                 <div className="inline-flex gap-1.5">
@@ -1388,9 +1388,7 @@ export function UsagePage() {
           {toolRows.length ? (
             <div className="rounded-2xl border border-border bg-surface p-4">
               <div className="mb-3 flex flex-col gap-1">
-                <h2 className="text-sm font-semibold text-foreground">
-                  Gateway-run tools
-                </h2>
+                <h2 className="text-title">Gateway-run tools</h2>
                 <p className="text-xs text-muted">
                   Tools Otari ran itself, billed per call. MCP tools are not
                   listed here: their names come from your own server, so they
--- a/web/src/shared/components/ui/SettingsSection.tsx
+++ b/web/src/shared/components/ui/SettingsSection.tsx
@@ -23,7 +23,7 @@ export const SettingsSection = ({
     <section className="flex flex-col gap-3">
       <div className="flex items-start justify-between gap-4">
         <div className="flex flex-col gap-1">
-          <h2 className="text-base font-semibold">{title}</h2>
+          <h2 className="text-title">{title}</h2>
           {description && <p className="text-sm text-muted">{description}</p>}
         </div>
         {actions && <div className="shrink-0">{actions}</div>}
--- a/web/src/app/HybridLanding.tsx
+++ b/web/src/app/HybridLanding.tsx
@@ -78,9 +78,7 @@ export function HybridLanding() {
           </div>
 
           <div className="flex flex-col gap-2">
-            <h2 className="text-sm font-semibold text-foreground">
-              Point a client here
-            </h2>
+            <h2 className="text-title">Point a client here</h2>
             <CopyableValue
               value={baseUrl}
               label="gateway base URL"
--- a/web/src/features/account/PasskeysCard.tsx
+++ b/web/src/features/account/PasskeysCard.tsx
@@ -230,7 +230,7 @@ export function PasskeysCard() {
 
   return (
     <section className="flex flex-col gap-2">
-      <h2 className="text-sm font-semibold text-foreground">Passkeys</h2>
+      <h2 className="text-title">Passkeys</h2>
       <Card>
         <Card.Content className="flex flex-col gap-4 px-5 py-5">
           <p className="max-w-3xl text-sm text-muted">
--- a/web/src/features/account/PasswordCard.tsx
+++ b/web/src/features/account/PasswordCard.tsx
@@ -174,7 +174,7 @@ export function PasswordCard() {
 
   return (
     <section className="flex flex-col gap-2">
-      <h2 className="text-sm font-semibold text-foreground">
+      <h2 className="text-title">
         {isClaimed ? "Dashboard password" : "Claim this deployment"}
       </h2>
       <Card>
--- a/web/src/features/budgets/BudgetsPage.tsx
+++ b/web/src/features/budgets/BudgetsPage.tsx
@@ -256,7 +256,7 @@ function BudgetForm({
   return (
     <Card>
       <Card.Content className="flex flex-col gap-4 p-5">
-        <div className="text-sm font-semibold text-foreground">{title}</div>
+        <div className="text-title">{title}</div>
         <ErrorBanner error={error} />
         <Field
           label="Name (optional)"
--- a/web/src/features/keys/KeysPage.tsx
+++ b/web/src/features/keys/KeysPage.tsx
@@ -418,9 +418,7 @@ function CreateKeyForm({
   return (
     <Card>
       <Card.Content className="flex flex-col gap-4 p-5">
-        <div className="text-sm font-semibold text-foreground">
-          Create API key
-        </div>
+        <div className="text-title">Create API key</div>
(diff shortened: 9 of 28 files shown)


web/src/features/settings/SettingsPage.tsx
```diff
@@ -683,7 +683,7 @@ export function SettingsPage() {
 
       {groups.map((group) => (
         <section key={group.name} className="flex flex-col gap-2">
-          <h2 className="text-sm font-semibold text-foreground">
+          <h2 className="text-heading">
```
jigjigjig, line 686: Self-review. This is one of only **two** sites in the 35 that took `text-heading` (18) rather than `text-title` (16), so it is the place to check whether the rule I applied is the right one.

The rule: what does the heading *head*? This `<h2>` sits in a `<section>` whose child is a `<Card>`, so it heads a group rather than titling one. `text-heading` is the section role; `text-title` is the card role. The other 33 sites are inside a bordered surface and title that surface's contents.

The outcome on this page is the clearest before/after in the PR. It read `H1 26 / H2 14 / H2 14`, so there was a 12px jump from the page title and then nothing between a heading and the settings rows it labelled. It now reads 26 / 18 / 16 / 14.

If a reviewer disagrees with the split, this is the site to argue about, and the argument is about what a Settings group is rather than about the numbers.

web/src/shared/components/ui/SettingsSection.tsx
```diff
@@ -23,7 +23,7 @@ export const SettingsSection = ({
     <section className="flex flex-col gap-3">
       <div className="flex items-start justify-between gap-4">
         <div className="flex flex-col gap-1">
```
jigjigjig, line 25: Worth flagging because it is evidence rather than a change.

This was `text-base font-semibold`, which is 16px at weight 550, the **exact metrics `text-title` names**. Somebody had already reached for that size by hand, and they reached for it in the one place the role map independently says 16. So this line is not being restyled; it is being given the name of what it already was.

That is the strongest single argument in the PR that the role map matches how people were already thinking, rather than imposing a scheme on them. The other 34 sites were 14px, and this one, the shared settings-section primitive, was the outlier that had it right.

web/src/features/usage/UsagePage.tsx
```diff
@@ -1315,7 +1315,7 @@ export function UsagePage() {
           <div className="grid gap-6 xl:grid-cols-2">
             <div className="flex flex-col gap-3">
               <div className="flex flex-wrap items-center justify-between gap-3">
-                <h2 className="text-sm font-semibold text-foreground">
+                <h2 className="text-title">
```
jigjigjig, line 1318: These three (`Spend by model`, `Spend by session`, `Gateway-run tools`) were the only sites named when this work was scoped. I grepped for the idiom rather than the three strings and found 35 across 26 files, all identical: `text-sm font-semibold text-foreground`.

**I did all 35 rather than the 3, and that was a scope call worth challenging.** The reasoning: a uniform idiom half-migrated is worse than either end state. Some cards at 16 and some at 14 reads as a bug, where all of them at 14 read as a choice. It is one mechanical substitution, one commit, revertible whole.

The cost of being wrong is low and the cost of being half-right is a dashboard that looks broken, so I took the larger scope. Say the word and it cuts back to three.

web/src/features/activity/ActivityPage.tsx
```diff
@@ -821,25 +819,25 @@ function RoutingPlan({ entry }: { entry: UsageEntry }) {
         >
           <thead className="text-muted">
             <tr className="border-b border-border">
-              <th scope="col" className="px-3 py-2 text-left font-medium">
+              <th scope="col" className="px-3 py-2 text-left text-overline">
                 #
               </th>
```
jigjigjig, line 827: These six were the second of the three findings. The main tables on this page get the single overline spec from `.otari-table .table__column`, which reaches into HeroUI's table DOM; this sub-table is hand-rolled `<table>`/`<th>`, so no rule reached it and it kept `font-medium`.

The visible result before this was two tables on one screen labelling their columns two different ways. Now both are 12px/550/+0.06em/uppercase/muted.

`text-overline` on a `<th>` rather than a class on the parent, because the alignment (`text-left` on four, `text-right` on two) has to stay per-header.

web/src/features/activity/ActivityPage.tsx
```diff
@@ -224,9 +224,7 @@ function InFlightControl({
       <Popover.Content placement="bottom end">
         <Popover.Dialog>
           <div className="flex w-80 flex-col gap-2">
-            <Popover.Heading className="text-sm font-medium">
-              In flight
-            </Popover.Heading>
+            <Popover.Heading className="text-title">In flight</Popover.Heading>
```
jigjigjig, line 227: Third finding, one site. A `Popover.Heading` at `text-sm font-medium` is a dialog title rendered at body size and at the *body* weight, so it was the weakest heading in the product.

Note what it did **not** get: `text-heading`. It titles the contents of a popover surface, which is the card case, so it takes `text-title` under the same rule as the other 33.

jigjigjig (commented):
Self-review. Five comments on the decisions rather than the mechanics, since the mechanics are one substitution repeated 35 times.

The three worth a reviewer's attention: the 33-versus-2 split and the rule behind it, the call to do all 35 sites when three were scoped, and `SettingsSection` turning out to already be at the exact metrics the role names.

Nothing in this pass changed the diff.

jigjigjig:
@coderabbitai review

Requesting manually: `.coderabbit.yaml` sets `base_branches: ["main"]`, so auto review skipped this PR because it is stacked on `fix/type-scale` rather than on `main`. Not a reason to go unreviewed.

jigjigjig:
@coderabbitai review

jigjigjig:
**Retargeted from `fix/type-scale` to `main`, and the reason is worth recording.**

Stacking bought a one-commit diff and silently gave up three things, which I found only by checking why a job was missing:

1. **No dashboard CI ran at all.** `otari-dashboard.yml`, `otari-lint.yml`, `otari-typecheck.yml` and `otari-tests.yml` all declare `pull_request: branches: [main]`. A PR based on a topic branch triggers none of them. This PR had no lint, no typecheck, no vitest and no build in CI, and the three repo-wide meta checks passing made it look green.
2. **No CodeRabbit review.** `.coderabbit.yaml` sets `base_branches: ["main"]`, so it skipped rather than reviewed.
3. **No `protect-main`.** The ruleset covers the default branch only, so this reported `CLEAN` and could have been merged into `fix/type-scale` with no approval and no thread resolution.

CI is what caught the only real regression in #807 (a stale test assertion, invisible in my local run for an unrelated reason). Trading it for a tidier diff was the wrong call.

**For reviewers:** until #807 merges the diff shows four commits and 33 files, three commits of which are #807's. **This PR is the single commit `5574e601`** (27 files, one class substitution repeated). Once #807 merges, the diff collapses to just that.

The retarget also removes the `delete_branch_on_merge` hazard entirely, so the instruction on #807 to retarget this before merging is now moot.

khaledosman (approved):
Note first that this PR now targets `main`, so the diff GitHub shows carries #807's three commits as well as this one. The review below is only about this PR's own commit; the `globals.css` / `foundation.test.ts` / `ModelsPage` type-scale hunks are reviewed on #807.

Verified the substitution rather than the count: no `text-sm font-semibold` survives anywhere under `web/src` outside test files, so the idiom is fully migrated and not left half-done. The `text-title` / `text-heading` split reads correctly against the rule you state (what the heading heads, not where it sits), and `SettingsSection.tsx` naming the role instead of restating `text-base font-semibold` is the right direction.

One miss.

**`web/src/features/models/ModelsPage.tsx:827`** still reads:

```tsx
<h2 className="text-base font-semibold break-all text-foreground">
  {row.model}
</h2>
```

That is an `<h2>` inside `Card.Content` titling that card's contents, at 16/550: the exact metrics `text-title` names, and the exact spelling you migrated in `SettingsSection.tsx` in this diff. It is the second of the two hand-rolled instances of the size the role map independently lands on, and leaving it is the "uniform idiom half-migrated reads as a bug where all-14 read as a choice" state the description argues against. (No inline comment because the line falls outside every hunk.)

Two smaller things, neither worth changing on its own:

- The `<thead className="text-muted">` on the Activity sub-table is now redundant, since `text-overline` sets `color: var(--color-text-muted)` on each `<th>` itself.
- Roughly fifteen of the 33 `text-title` sites are `<div>`s rather than headings, so the document-outline improvement the Settings example describes does not reach them. That is a separate change from this one and #808 already covers the outline, so it is only worth knowing that the outline claim applies to the `<h2>` subset.

Reverting the "five meanings sharing 14px" claim in the two places it was made, with the measurement that disproves it, is the part of this PR most worth keeping in the record.

---

jigjigjig:
Thank you for reviewing the commit rather than the diff GitHub showed, and for checking that no `text-sm font-semibold` survives rather than taking the count on trust. All three points were right; two are fixed and the third is now stated as a limit.

**`ModelsPage.tsx:827` was a real miss**, and the most annoying kind: the other hand-rolled instance of exactly the size the role map lands on, in the same file I had already edited. Fixed in 655e5717, now `text-title break-all`. Re-grepped afterwards rather than trusting the fix: no `text-sm font-semibold` or `text-base font-semibold` survives anywhere under `web/src` outside test files.

**The redundant `<thead className="text-muted">`** is gone in the same commit, with a comment saying why, since `text-overline` sets the color on each `<th>` itself.

**The `<div>` point is the one I would have got wrong in the description**, so it is now in the body as an explicit limit: about fifteen of the 33 sites are `<div>`s, the outline improvement reaches the `<h2>` subset only, and #808 owns the outline question. The Settings example I led with happens to be all `<h2>`, which is what made the claim look general.

Rebased onto #807's review fixes, so this branch now carries the caption-step and `@apply text-overline` changes in its base.

PR #812, opened 2026-08-26
feat(mcp): replace inline MCP execution with stored-server endpoints

### LINKED ISSUE #791: Add stateless MCP tool execution endpoint
## Desired behavior

Add an authenticated, stateless `POST /v1/mcp/execute` endpoint that accepts one inline MCP server config, a consented tool name, and arguments. Validate the URL and live allowlist, execute the exact call through `MCPClientPool`, and return a typed result or error. Expose it in hybrid mode.

## Why

Applications such as Octonous need to approve, reject, or edit mutating MCP calls before execution. The managed loop currently executes gateway-owned calls internally, so the caller cannot apply its approval and audit flow. A one-shot endpoint keeps approval with the caller while Otari remains the MCP executor, without holding streams, MCP sessions, or cross-worker approval state open.

## Notes

Reuse existing API authentication, URL-safety checks, allowlist enforcement, cancellation, and error sanitization. Never log credentials, arguments, or results. This complements #755, which covers activity signals for calls that remain inside the managed loop.

Ref: mozilla-ai/octonous#4916

HareeshBahuleyan:
## Why

Otari's stateless MCP endpoint took an inline server configuration, so a caller
transmitted a URL and a credential on every request, chose its own outbound
destination, and supplied its own policy. In hybrid mode it authenticated by
resolving an empty id list, which bound the call to no stored server at all.
Nothing tied the tool a caller asked for to a server its workspace was
authorized to reach.

 MCP servers expose tools that an AI application can call. This PR adds a safer, stateless way for 
 an application to use those tools:                                                                
                                                                                                   
 1. The application asks Otari which tools a stored MCP server exposes.                            
 2. The application decides whether to allow a proposed tool call, potentially asking a human.     
 3. The application sends the approved tool name and arguments back to Otari.                      
 4. Otari opens a short-lived MCP connection, runs exactly that call, returns the native MCP       
    result, and closes the connection.                                                             
                                                                                                   
 Nothing remains open while approval is happening. 

## What changed

`POST /v1/mcp/execute` now takes a stored server id, and
`GET /v1/mcp/servers/{mcp_server_id}/tools` is new. Together they let an
application own its own MCP tool loop: discover once per run, apply its own
authorization policy to what a model proposes, then ask Otari to run one exact
authorized call. Otari holds nothing open in between, which is what lets the
application pause for a person.

Otari executes a caller-authorized call and does not verify a human approval.
It enforces authentication, stored-server access, the stored allowlist, URL
safety, and its own bounds.

`server_revision` binds an authorization decision to the configuration it was
made against. It is derived from the resolved URL, a digest of the credential,
the enabled state and the sorted allowlist, so no migration or platform field
is needed and every worker agrees by construction. A stale revision is a 409.

## Notes

- **Legacy resolver compatibility.** New platform peers should return `id` and
  `enabled`. A legacy peer may omit both; Otari binds its one returned entry to
  the requested id and treats it as enabled. An empty list remains the legacy
  disabled-server response. Contract in `docs/hybrid-mode-protocol.md`.
- **`/v1/mcp/execute` must never be retried automatically**, including by a
  proxy or service mesh. `outcome_unknown` means the tool may already have run.
  Verify ingress policy before enabling for anyone real.
- **The inline `server` field is gone**, and now rejected rather than ignored.
- **Empty MCP allowlists now deny all tools.** This changes managed completion behavior. Omit `allowed_tools` or send `null` for unrestricted access; use `[]` for deny-all.
- Execution never calls `tools/list`; the stored allowlist is the authority.
- Errors share one body (`detail`, `code`, `execution_state`, `request_id`) with
  a fixed safe message per code. `docs/mcp.md` has the full table and limits.
- `mcp` is pinned to one minor line. The endpoint answers with mcp's own
  `CallToolResult`, so ten mcp-owned schemas reach three committed artifacts;
  see the constraint comment and AGENTS.md before raising it.

Fixes #791

## PR Type

- [x] New Feature
- [x] Refactor
- [x] Documentation

## Checklist

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:** GPT-5 Codex for the original inline-server version; Claude Opus 5 for this refactor.

**Any additional AI details you'd like to share:** The refactor was driven test-first against the design doc, and the review findings on the first version were each verified against the new code rather than assumed fixed. Two were reproduced before fixing: a blocked user's call returning 200, and a dead MCP server escaping the error contract entirely.

**NOTE:**
When responding to reviewer questions, please respond yourself rather than copy/pasting reviewer comments into an AI and pasting back its answer. We want to discuss with you, not an AI :)

## Summary

- Replaced inline MCP server settings with stored-server references.
- Added stored-server tool discovery.
- Added checks for access, revisions, allowlists, URL safety, and execution limits.
- Standardized errors and added safe handling for uncertain execution outcomes.
- Updated documentation, API artifacts, dependency controls, and tests.

These changes support secure, stateless, caller-controlled MCP approvals without open sessions or shared approval state.

### DIFF
--- a/docs/mcp.md
+++ b/docs/mcp.md
@@ -168,9 +168,11 @@ an approval credential.
 It detects Otari-side and platform-side configuration changes only. A remote
 server that changes its own catalog, or a tool's behavior, behind an unchanged
 URL will not move it; catching that would need another live `tools/list` at
-execution time, which this version does not do. What the revision binds is the
-stored server id, its stored configuration, the remote tool name, and the exact
-arguments, not an immutable remote implementation.
+execution time, which this version does not do. The revision binds the request
+only to the stored server configuration returned by discovery. The calling
+application is responsible for binding the server id,
+revision, remote tool name, and final arguments to its authorization or approval
+record. Neither mechanism binds the call to an immutable remote implementation.
 
 `client_execution_id` is a canonical UUID you generate, for correlation across
 your service and Otari. It is not proof of approval and **not an idempotency
--- a/docs/mcp.md
+++ b/docs/mcp.md
@@ -270,13 +270,12 @@ Gateway-owned ceilings, not caller-controlled request fields:
 Discovery and execution hold separate concurrency ceilings, so slow or hostile
 discovery cannot starve calls an application has already authorized.
 
-The 1 MiB transport ceiling is enforced at two points: a `Content-Length`
-pre-check that refuses an oversized response before its body is read, and a
-measurement of the decoded result. It is **not** a streaming bound. A chunked or
-SSE response carries no `Content-Length`, and the MCP SDK's Streamable HTTP
-transport buffers a JSON response before returning a result, so that case is
-bounded by the call and total deadlines and by your deployment's HTTP timeouts
-until a streaming-bounded transport lands.
+The 1 MiB transport ceiling is enforced while each response stream is read, so
+chunked JSON and SSE responses cannot bypass it by omitting `Content-Length`.
+Otari requests identity encoding and refuses compressed responses before
+reading their bodies, preventing a small encoded response from expanding past
+the ceiling during decompression. The decoded MCP result is measured again
+before it is returned.
 
 Redirects are disabled. Otari returns a redirect response rather than following
 it, so no credential and no call data is ever sent to a redirect destination.
--- a/docs/mcp.md
+++ b/docs/mcp.md
@@ -24,6 +24,29 @@ When the model emits an MCP tool call, Otari:
 The loop stops when the model returns a normal assistant response or hits
 `max_tool_iterations`.
 
+## Stateless tool execution
+
+`POST /v1/mcp/execute` runs one caller-approved tool against one inline MCP
+server, without calling a model or retaining a session:
+
+```json
+{
+  "server": {
+    "name": "github",
+    "url": "https://mcp.example.com/github",
+    "allowed_tools": ["create_issue"]
+  },
+  "tool_name": "create_issue",
+  "arguments": {"repository": "mozilla-ai/otari", "title": "Approved issue"}
+}
+```
+
+Otari checks URL safety, `allowed_tools`, and live tool discovery before
+execution. It returns the native MCP result, including `isError`; transport and
+protocol failures become a sanitized `502`. The endpoint accepts the usual API
+key or master key in standalone mode, and an otari.ai bearer token in hybrid
+mode.
+
 ## Messages streaming activity
 
 For streaming `/v1/messages` requests with the standard Anthropic header
--- a/docs/mcp.md
+++ b/docs/mcp.md
@@ -114,6 +114,10 @@ one with a `$ref` pointing outside itself:
 {"tool_name": "create_issue", "code": "mcp_tool_schema_unsupported"}
 ```
 
+A tool name outside the 1–256-character execution limit is omitted with
+`mcp_tool_name_unsupported`, so every tool returned by discovery can be sent
+back to the execution endpoint.
+
 Otari does not validate an `inputSchema` against a JSON Schema dialect, and never
 truncates, rewrites, or drops a keyword. It bounds the size, depth and property
 count it will carry, and it refuses a schema that would have to be fetched over
--- a/docs/mcp.md
+++ b/docs/mcp.md
@@ -413,7 +413,7 @@ are managed there.
 - `mcp_server_ids` accepts at most 50 ids, which is also the most servers a workspace can have; a repeated id resolves once
 - Server names must be unique across everything one request uses, because Otari routes each tool call to its server by name. Two `mcp_servers` entries sharing a name are refused with a `400` naming it, the way a bad request-body URL is. An `mcp_servers` name colliding with one of your workspace's stored servers is a `400` too, but a fixed one that repeats neither name, since a stored server is not yours to read. Two stored servers sharing a name is a `500` with the names in the log, on the same grounds as a stored URL that fails its safety check
 - MCP URLs are validated to reduce SSRF risk; by default, private and reserved addresses are blocked, loopback is allowed, and `http://` is rejected when `authorization_token` is present
-- Redirects are followed only when the scheme, host, and port stay the same, or when a default-port `http://` URL upgrades to `https://` on the same host; configure the final URL when a redirect changes host or port
+- The managed MCP loop follows redirects only when the scheme, host, and port stay the same, or when a default-port `http://` URL upgrades to `https://` on the same host. Caller-orchestrated discovery and execution do not follow redirects. Configure the final URL when a managed-loop redirect changes host or port
 - `OTARI_MCP_ALLOW_LOOPBACK=false` disables loopback; `OTARI_MCP_ALLOW_PRIVATE_HOSTS=true` relaxes the private-host restriction
 
 For the hybrid platform contract behind `mcp_server_ids`, see
--- a/docs/mcp.md
+++ b/docs/mcp.md
@@ -234,6 +234,8 @@ exception text.
 |---|---:|---|---|
 | Invalid request body | 422 | `invalid_request` | `not_started` |
 | Authentication failure | 401 | `authentication_failed` | `not_started` |
+| Payment required or insufficient funds | 402 | `payment_required` | `not_started` |
+| Platform authorization refused | 403 | `forbidden` | `not_started` |
 | Authenticated request rate exceeded | 429 | `rate_limit_exceeded` | `not_started` |
 | Server inaccessible to caller, or disabled | 404 | `mcp_server_not_found` | `not_started` |
 | Stored server revision changed | 409 | `mcp_server_changed` | `not_started` |
@@ -244,6 +246,7 @@ exception text.
 | Discovery bound or pagination failure | 502 | `mcp_discovery_limit_exceeded` | `not_started` |
 | Discovery admission deadline exceeded | 503 | `mcp_discovery_capacity_unavailable` | `not_started` |
 | Execution admission deadline exceeded | 503 | `mcp_capacity_unavailable` | `not_started` |
+| Local service temporarily unavailable | 503 | `service_unavailable` | `not_started` |
 | Connection or initialization failure | 502 | `mcp_connection_failed` | `not_started` |
 | Tool-call deadline exceeded after dispatch | 504 | `mcp_outcome_unknown` | `outcome_unknown` |
 | Transport, cancellation, protocol or parse failure after dispatch | 502 | `mcp_outcome_unknown` | `outcome_unknown` |
@@ -281,8 +284,10 @@ reading their bodies, preventing a small encoded response from expanding past
 the ceiling during decompression. The decoded MCP result is measured again
 before it is returned.
 
-Redirects are disabled. Otari returns a redirect response rather than following
-it, so no credential and no call data is ever sent to a redirect destination.
+Redirects are disabled and never forwarded to the caller. A redirect during
+connection or initialization becomes `502 mcp_connection_failed`; after tool
+dispatch it becomes `502 mcp_outcome_unknown`. No credential or call data is
+sent to the redirect destination.
 
 ### Availability
 
@@ -333,7 +338,7 @@ to hide this activity because they have no equivalent server-owned MCP vocabular
 - `url`: streamable HTTP MCP endpoint, reachable from the gateway
 - `authorization_token`: optional bearer token; when set, the `url` must use `https://`
 - `purpose_hint`: optional hint Otari prepends to the system message to help the model choose the tool
-- `allowed_tools`: optional allow-list; only these tools are exposed from that server
+- `allowed_tools`: optional allow-list with three states: omit it or send `null` to expose every listed tool, send `[]` to expose none, or send a non-empty list to expose only those named tools
 
 ## Workspace-scoped servers
 
--- a/docs/mcp.md
+++ b/docs/mcp.md
@@ -24,28 +24,269 @@ When the model emits an MCP tool call, Otari:
 The loop stops when the model returns a normal assistant response or hits
 `max_tool_iterations`.
 
-## Stateless tool execution
+## Caller-orchestrated MCP
 
-`POST /v1/mcp/execute` runs one caller-approved tool against one inline MCP
-server, without calling a model or retaining a session:
+Otari supports two MCP integration modes, and they are alternatives rather than
+layers.
+
+**Managed MCP** is the loop above: you send `mcp_servers` or `mcp_server_ids`
+with a completion, and Otari owns the model/tool loop.
+
+**Caller-orchestrated MCP** hands the loop back to your application. Otari
+provides two stored-server endpoints, and your application decides whether a
+proposed call may run:
+
+1. `GET /v1/mcp/servers/{mcp_server_id}/tools` returns the tool definitions a
+   stored MCP server exposes to the authenticated workspace.
+2. Your application exposes those tools to its model or workflow, and applies
+   its own authorization policy to whatever gets proposed. That may be a user
+   approval prompt, an administrator rule, or trusted read-only
+   auto-authorization.
+3. `POST /v1/mcp/execute` runs one exact authorized call and returns the remote
+   server's native MCP result.
+
+**Otari executes a caller-authorized call. It does not verify a user approval
+and does not claim to.** The calling application is the authorization boundary
+and owns any human approval, argument editing, cancellation, and action history.
+What Otari enforces independently is authentication, that the stored server is
+one the authenticated workspace may reach, the stored tool allowlist, URL
+safety, and its own execution bounds.
+
+Between the two requests Otari holds nothing open: no model stream, no MCP
+session, no database session, no worker-local state. That is why your
+application can pause for a person for as long as it needs to.
+
+Both endpoints refer to a server by the id it was registered under. There is no
+inline server field: registering a remote MCP server through the control plane
+once, rather than transmitting a URL and a credential per request, is what keeps
+authorization, credential storage, revocation, and allowlist policy on Otari's
+side of the boundary. Local `stdio` servers stay outside this: a hosted Otari
+deployment cannot reach a machine-local process.
+
+### Discovery
+
+```http
+GET /v1/mcp/servers/2c948a61-dc96-4cd8-96bb-8e1434bf424e/tools
+Authorization: Bearer <token>
+```
 
 ```json
 {
-  "server": {
-    "name": "github",
-    "url": "https://mcp.example.com/github",
-    "allowed_tools": ["create_issue"]
-  },
-  "tool_name": "create_issue",
-  "arguments": {"repository": "mozilla-ai/otari", "title": "Approved issue"}
+  "server_id": "2c948a61-dc96-4cd8-96bb-8e1434bf424e",
+  "server_revision": "9f2c41ae7b0d4e5c8a13d6f0b27e5c91",
+  "tools": [
+    {
+      "name": "create_issue",
+      "description": "Create an issue",
+      "input_schema": {"type": "object", "properties": {"title": {"type": "string"}}},
+      "annotations": {"readOnlyHint": false}
+    }
+  ],
+  "warnings": []
 }
 ```
 
-Otari checks URL safety, `allowed_tools`, and live tool discovery before
-execution. It returns the native MCP result, including `isError`; transport and
-protocol failures become a sanitized `502`. The endpoint accepts the usual API
-key or master key in standalone mode, and an otari.ai bearer token in hybrid
-mode.
+Call this once per server when you prepare a run, and reuse the answer for every
+tool from that server for the length of the run. There is no cross-run cache, so
+a later run rediscovers.
+
+Where the stored server has an allowlist, `tools` is its intersection with the
+live catalog, so a tool the server added later is not exposed just because it
+was listed. Where it has none, the whole live catalog is returned, the same
+meaning the column already carries in the managed loop. An explicit empty
+allowlist is a deny-all: `tools` is empty and Otari opens no connection.
+
+The response never contains the server URL, its credential, or an allowlist entry
+the live catalog did not return.
+
+`annotations` is the server's own metadata, passed through as untrusted data.
+Otari never turns `readOnlyHint` into an authorization decision, and never infers
+safety from a tool's name or description. Your risk policy is yours.
+
+`tools` is either the complete authorized catalog or absent: a pagination
+failure or a discovery ceiling returns `502 mcp_discovery_limit_exceeded` with no
+tools at all, because a short catalog would be read as the complete input to
+your authorization policy. The one exception is `warnings`, which names a tool
+whose descriptor Otari could not carry, for example a schema that is too large or
+one with a `$ref` pointing outside itself:
+
+```json
+{"tool_name": "create_issue", "code": "mcp_tool_schema_unsupported"}
+```
+
+Otari does not validate an `inputSchema` against a JSON Schema dialect, and never
+truncates, rewrites, or drops a keyword. It bounds the size, depth and property
+count it will carry, and it refuses a schema that would have to be fetched over
+the network to be understood at all. An unknown dialect or vendor keyword survives
+the round trip unchanged.
+
+### Execution
+
+```http
+POST /v1/mcp/execute
+Authorization: Bearer <token>
+Content-Type: application/json
+```
+
+```json
+{
+  "mcp_server_id": "2c948a61-dc96-4cd8-96bb-8e1434bf424e",
+  "tool_name": "create_issue",
+  "arguments": {"title": "Approved title", "body": "Approved body"},
+  "server_revision": "9f2c41ae7b0d4e5c8a13d6f0b27e5c91",
+  "client_execution_id": "6e51b3bc-6f48-4c68-a61e-0786bc80cd67"
+}
+```
+
+Otari resolves the stored server, applies its allowlist to `tool_name`, compares
+`server_revision`, validates the resolved URL, and then calls that one tool
+directly. It never calls `tools/list` here: MCP permits calling a known tool, and
+what authorizes the call is the stored allowlist and the authenticated workspace,
+not a live catalog.
+
+The successful body is the native MCP result:
+
+```json
+{
+  "content": [{"type": "text", "text": "Created issue #42"}],
+  "structuredContent": {"issue_number": 42},
+  "isError": false
+}
+```
+
+`isError: true` is still an HTTP 200. The server handled the call and classified
+the outcome itself, so it is a definitive result rather than a transport failure,
+and it is not a retry signal.
+
+`server_revision` is the value discovery returned. Persist it with the proposed
+call and send it back. It is an opaque revision of the stored server's URL,
+credential, enabled state and allowlist, so a configuration change between your
+authorization decision and this request is refused with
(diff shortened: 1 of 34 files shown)


src/gateway/api/routes/mcp.py
```diff
@@ -0,0 +1,120 @@
...
+            if not pool.owns_tool(request.tool_name):
+                raise HTTPException(
+                    status_code=status.HTTP_404_NOT_FOUND,
+                    detail=MCP_TOOL_NOT_FOUND_DETAIL,
+                )
+            return await pool.call_tool_result(request.tool_name, request.arguments)
+    except HTTPException:
+        raise
+    except Exception as exc:
+        logger.warning("Stateless MCP request failed error_class=%s", type(exc).__name__)
```
khaledosman: `type(exc).__name__` will log `ExceptionGroup` for essentially every real transport failure, so the sanitized 502 leaves behind no usable diagnostic. Verified against the pinned `mcp`: `streamable_http_client` yields inside `anyio.create_task_group()`, so a dead server surfaces as `ExceptionGroup('unhandled errors in a TaskGroup', [ConnectError('All connection attempts failed')])`. Unwrap the leaves before logging, e.g. log `[type(e).__name__ for e in exc.exceptions]` when `isinstance(exc, BaseExceptionGroup)`. `test_transport_and_protocol_failures_are_sanitized` asserts `"RuntimeError"` from a mocked pool, so it never sees this shape.

Related: a `BaseExceptionGroup` with a non-`Exception` leaf (a cancellation) escapes `except Exception` entirely and returns a 500 instead of the sanitized 502.

src/gateway/api/routes/mcp.py
```diff
@@ -0,0 +1,120 @@
...
+) -> CallToolResult:
+    """Execute one caller-approved tool against one inline MCP server.
+
+    The server is connected only for this request. Otari validates its URL,
+    applies its configured tool allowlist, confirms the tool through live MCP
+    discovery, and then executes exactly the named call. A server-returned
+    ``isError`` remains a typed MCP result; transport and protocol failures are
+    returned as a sanitized gateway error.
+    """
+    await _authenticate(raw_request, db, config)
```
khaledosman: No `check_rate_limit`. Rate limiting in this repo is per-route, not middleware (`_pipeline.py:1603`, `_passthrough.py:262`), so this route has none. It performs caller-directed DNS + outbound HTTP, `pool.call_tool` is issued with no `read_timeout_seconds` and the transport's `sse_read_timeout` defaults to 300s, and there is no budget or metering on this path. Any valid API key can hold gateway workers open indefinitely. Resolve the key's user and call `check_rate_limit(raw_request, user_id)` after `_authenticate`, and pass an explicit `read_timeout_seconds` to the tool call.

src/gateway/api/routes/mcp.py
```diff
@@ -0,0 +1,120 @@
...
+        await _resolve_platform_mcp_servers(config, user_token, [])
+        return
+
+    if db is None:
+        raise HTTPException(
+            status_code=status.HTTP_503_SERVICE_UNAVAILABLE,
+            detail="Authentication temporarily unavailable, please retry",
+        )
+    session_identity = await get_session_identity(raw_request, db, config)
+    await verify_api_key_or_master_key(raw_request, db, config, session_identity)
```
khaledosman: `verify_api_key_or_master_key` checks only that the key is active and unexpired. The `users.blocked` gate lives in `reserve_budget` (`budget_service.py:400`), which this route never reaches, so a blocked user's key still drives outbound MCP calls through the gateway. Either resolve `api_key.user_id` here and refuse a blocked user, or state in the docstring why that gate is out of scope for an unbilled route.

khaledosman (commented):
Reviewed against `AGENTS.md`, `.github/skills/review`, `.github/instructions/security-review`, and `src/gateway/AGENTS.md`. Findings inline. Generated artifacts (OpenAPI, Postman, `web/src/client/schema.ts`), mode registration, and prose style all check out.

🤖 Generated with Claude Opus 5 (1M context) via Claude Code, acting as an agent.

src/gateway/api/routes/mcp.py
```diff
@@ -0,0 +1,120 @@
...
+
+
+@router.post(
+    "/execute",
+    response_model=CallToolResult,
```
khaledosman, line 589: `response_model=CallToolResult` pulls the whole `mcp` package type tree (`CallToolResult`, `TextContent`, `ImageContent`, `AudioContent`, `ResourceLink`, `EmbeddedResource`, and their resource-contents/annotations children) into `docs/public/openapi.json`, the Postman collection, and the committed `web/src/client/schema.ts`. A routine `mcp` version bump now fails `openapi-check`/`postman-check` and the dashboard client drift check on an unrelated dependabot PR. Define a gateway-owned response model that mirrors the fields you actually promise and convert at the boundary.

tests/unit/test_mcp_execute_endpoint.py
```diff
@@ -0,0 +1,288 @@
...
+    _FakePool.owns = True
+    _FakePool.enter_error = None
+    _FakePool.call_error = None
+    _FakePool.result = CallToolResult(
+        content=[TextContent(type="text", text="two open issues")],
+        structuredContent={"count": 2},
+        isError=False,
+    )
+    monkeypatch.setattr(mcp_route, "MCPClientPool", _FakePool)
+    monkeypatch.setattr(mcp_route, "validate_mcp_url", _safe_url)
```
khaledosman: The `fake_pool` fixture patches out `validate_mcp_url` for every test, and `test_unsafe_url_is_rejected_before_contacting_server` substitutes its own raiser, so nothing exercises the real SSRF gate through the route. The two properties worth pinning are that `http://169.254.169.254/mcp` is rejected, and that `http://...` plus `authorization_token` is rejected before the token goes out in cleartext (the `has_authorization_token=` argument at `mcp.py:97` is otherwise untested). Add a test that leaves `validate_mcp_url` unpatched.

src/gateway/api/routes/mcp.py
```diff
@@ -0,0 +1,120 @@
...
+    db: AsyncSession | None,
+    config: GatewayConfig,
+) -> None:
+    """Authenticate through the mode's existing data-plane path."""
+    if config.is_hybrid_mode:
+        user_token = _extract_platform_user_token(raw_request)
+        # Stateless execution has no stored ids to resolve. The empty resolve
+        # deliberately reuses the platform's MCP authorization boundary rather
+        # than inventing a second authentication call for this endpoint.
+        await _resolve_platform_mcp_servers(config, user_token, [])
```
khaledosman: This makes an empty-ids resolve load-bearing for hybrid authentication, but nothing pins that contract on the platform side. `MCPGatewayResolveService.resolve` in otari-ai does validate both tokens, org match, and the default-gateway toggle before it looks at `mcp_server_ids`, so it holds today, and `backend/overlay/tests/services/test_mcp_gateway_resolve_service.py` has no empty-list case. An "optimize the empty list to an early return" change there would silently make this endpoint unauthenticated in hybrid mode. Worth an otari-ai issue for a regression test asserting `{"mcp_server_ids": []}` still rejects a bad user token.

njbrake:
Hi @HareeshBahuleyan , what's the risk profile for this PR? We're technically in a feature freeze while we're doing the deployment of the new design so I want to be careful with adding new features while this is in progress.

HareeshBahuleyan:
Hey @njbrake its not urgent, so it can wait till the new design is ready 🙏

src/gateway/api/routes/mcp.py
```diff
@@ -0,0 +1,120 @@
...
+            status_code=status.HTTP_400_BAD_REQUEST,
+            detail=str(exc),
+        ) from exc
+
+    try:
+        async with MCPClientPool([server]) as pool:
+            if not pool.owns_tool(request.tool_name):
+                raise HTTPException(
+                    status_code=status.HTTP_404_NOT_FOUND,
+                    detail=MCP_TOOL_NOT_FOUND_DETAIL,
```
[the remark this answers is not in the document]
HareeshBahuleyan: Resolved by a design change rather than a fix: execution no longer calls `list_tools` at all. Authorization comes from the stored allowlist and the authenticated workspace, so there is no live catalog for an approved tool to be missing from.

Discovery does page, and does it as you described: it follows `nextCursor` under a 20-page ceiling, stops early once every allowlisted name has been seen, and refuses the whole response on a repeated cursor, a malformed page, or two different tools under one name. Covered by `tests/unit/test_mcp_discovery_pagination.py`, including the multi-page case.

src/gateway/api/routes/mcp.py
```diff
@@ -0,0 +1,120 @@
...
+        raise HTTPException(
+            status_code=status.HTTP_403_FORBIDDEN,
+            detail=MCP_TOOL_NOT_ALLOWED_DETAIL,
+        )
+
+    try:
+        await validate_mcp_url(
+            server.url,
+            has_authorization_token=bool(server.authorization_token),
+        )
```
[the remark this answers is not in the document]
HareeshBahuleyan: Correct, and fixed for this path. The stateless transport now passes an `httpx_client_factory` with `follow_redirects=False`, so a redirect comes back as a response instead of being followed and no credential or call data reaches its destination. `tests/unit/test_mcp_stateless_transport.py` asserts both the flag and that a 307 produces exactly one request, to the original host.

Worth flagging that the other half of your finding survives outside this diff. `services/mcp_client.py` still opens the managed pool's transport with the SDK default, so `mcp_servers` and `mcp_server_ids` on a completion still follow redirects. That path is no longer touched by this PR, so I am filing it separately rather than widening this one.

src/gateway/api/routes/mcp.py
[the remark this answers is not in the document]
HareeshBahuleyan: Added. Standalone rate-limits on the authenticated key's subject before any outbound work; hybrid defers to the platform resolver, which enforces the user and workspace limit, and Otari preserves its 429 and `Retry-After` rather than collapsing it.

There is also a per-process concurrency ceiling behind that, deliberately separate for discovery and execution so slow or hostile discovery cannot starve a call an application has already authorized. Both capacity refusals carry `Retry-After`; nothing else on this contract advertises a retry.

`tests/integration/test_mcp_stored_server_endpoints.py` covers the standalone limiter end to end, including that the refused request never reaches the MCP server.

src/gateway/api/routes/mcp.py
HareeshBahuleyan: You were right, and it turned out worse than described. Reproduced against the pinned `mcp`: a closed port surfaces from `session.initialize()` as a **bare `CancelledError`**, not an `ExceptionGroup`, and transport shutdown separately raises a group. So `except Exception` let the commonest failure there is escape the error contract completely. At the route it reached Starlette, which answered `RuntimeError: No response returned` and sent nothing at all.

And your point about the test was exactly the reason it survived the rewrite: every test raised a tidy `RuntimeError` from a substituted session, so none of them could see the real shape.

Fixed in `9af67340`. `BaseException` at the three points that classify a transport failure, with the deadline conversions kept ahead of them so a phase bound still reports as itself rather than as a connection failure. Connect stays `not_started`, since nothing was written; after dispatch stays `outcome_unknown`, grouped cancellation included, because a cancelled local transport says nothing about whether the remote server ran the tool.

Also added `failure_class`, which names a failure by its leaf types, so a dead server now logs `error_class=ConnectError` instead of `ExceptionGroup`. Types only, never a message, since an exception message can carry the URL, the credential, or the caller's arguments. Tests now use the shapes the SDK really produces, and I verified both endpoints against a real closed port.

Thanks for chasing this into the SDK's task group; it was the most valuable comment on the PR.

src/gateway/api/routes/mcp.py
HareeshBahuleyan: Both halves addressed.

Rate limiting: standalone limits on the key's subject before any outbound work, hybrid defers to the platform resolver and preserves its 429 and `Retry-After`.

The unbounded call: 5s for connect plus initialize, 30s for the tool call measured from dispatch, 45s for the whole request, and an 8-slot per-process gate with a bounded admission wait. Bounded with `asyncio.timeout` rather than `read_timeout_seconds`, which is equivalent from Otari's side; the difference is that the SDK would also send the server a cancellation notification. Happy to switch if you would rather the server be told.

Still no metering, which is deliberate for a route that dispatches to no provider and costs nothing to serve, but it is no longer possible to hold a worker open indefinitely on a valid key.

src/gateway/api/routes/mcp.py
HareeshBahuleyan: Confirmed and fixed in `13d7aa34`. Reproduced before fixing: a blocked user's execute call returned 200 with the tool result, exactly as you predicted, because `users.blocked` is read only in `reserve_budget` and this route reserves nothing.

Now checked where the route already holds the session and the key, before the database session is released, so it costs one indexed read on a session that is open anyway. Reported as `authentication_failed` rather than a code of its own: the caller's credential no longer authorizes anything, and a distinct code would tell an unauthenticated probe that a given key names a real but blocked user. The reason goes to the log instead. A key naming no user has no block to check, which is the bootstrap and convenience path.

src/gateway/api/routes/mcp.py
HareeshBahuleyan: Confirmed still true, and measured: ten `mcp`-owned schemas in `openapi.json` (`CallToolResult`, `TextContent`, `ImageContent`, `AudioContent`, `ResourceLink`, `EmbeddedResource`, `Annotations`, `BlobResourceContents`, `TextResourceContents`, `Icon`), travelling from there into the Postman collection and `web/src/client/schema.ts`.

One correction to the mechanism, in your favor and mine: nothing bumps `mcp` automatically here, since `dependabot.yml` only configures the `github-actions` ecosystem. The real trigger is a `uv lock` run for an unrelated dependency, which is arguably more surprising than a dependabot PR, because the diff that moves three API artifacts is one that added something else entirely.

I did not take the mirror model, and the reasoning is a trade rather than a disagreement about the cost you identified. A gateway-owned response model has to enumerate content-block types, so the day `mcp` adds an eleventh, Otari silently drops it and an image result reaches the caller as nothing. `ResourceLink` and `AudioContent` were both added after 1.0, so that is not hypothetical, and returning the native bounded result is an acceptance criterion of the design this endpoint implements.

What went in instead (`8b6b44ec`): `mcp` is pinned to one minor line, `>=1.28.1,<1.29.0`, so the version only moves when someone means it and a `uv lock` for something else cannot drag the artifacts with it. Patch releases still flow. `AGENTS.md` now records that raising the ceiling is API work rather than a dependency bump, naming the ten schemas, the three committed files, and why the native result is deliberate, so the next person does not "fix" this by adding the mirror model and quietly losing content types.

Closing this with the mitigations rather than the fix you proposed. If you still think the drift checks should win, reopen it and I will do the mirror model.

tests/unit/test_mcp_execute_endpoint.py
HareeshBahuleyan: Fair, and half of it was already fixed by the rewrite: the new tests never patch `validate_mcp_url`, and both endpoints exercise the real one with `http://169.254.169.254/mcp`.

Your second property genuinely was not covered, and for the reason you would expect: the metadata address fails on the private-range rule and returns before the cleartext-credential rule is ever consulted, so `has_authorization_token=` had no test behind it at all.

Added the pair that actually pins the argument, in `13d7aa34`: a public `http://` host with a stored token is refused with `unsafe_mcp_url` and opens no session, and the same URL with no stored token is allowed through, since there is nothing to leak. Verified the first fails when the argument is hardcoded to `False` while the second still passes, so the pair cannot go green on a refactor that drops it.

src/gateway/api/routes/mcp.py
HareeshBahuleyan: Moot for this endpoint now: the empty-ids resolve is gone. Hybrid resolves the id that was actually requested, and reads the answer strictly, requiring exactly one entry whose `id` matches. Zero, several, a mismatched id, or a field Otari cannot parse is a resolution failure rather than a best guess, so hybrid authentication no longer rides on an empty list being rejected.

That needed two new required fields from the resolver, `id` and `enabled`, both documented in `docs/hybrid-mode-protocol.md` along with what a peer omitting either gets.

Your otari-ai point still stands on its own for `_resolve_platform_mcp_servers`, which the tool loop calls with a real list; it is just no longer load-bearing for this endpoint.

src/gateway/api/routes/mcp.py
```diff
@@ -0,0 +1,129 @@
...
+
+    result: CallToolResult | None = None
+    try:
+        async with MCPClientPool([server]) as pool:
+            if not pool.owns_tool(request.tool_name):
+                raise HTTPException(
+                    status_code=status.HTTP_404_NOT_FOUND,
+                    detail=MCP_TOOL_NOT_FOUND_DETAIL,
+                )
+            result = await pool.call_tool_result(request.tool_name, request.arguments)
```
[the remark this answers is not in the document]
HareeshBahuleyan: Added. Both endpoints call `track_request` after validation and before the outbound MCP session opens, each under its own path so an operator watching `/v1/usage/in-flight` can tell discovery from execution. Removal stays with `InFlightMiddleware`, as you noted.

`tests/integration/test_mcp_stored_server_endpoints.py` pins it behaviorally rather than by asserting the call: the substituted MCP session reads the registry while the call is in flight, and the test checks the entry is there at that moment and gone once the response has been sent.

src/gateway/services/mcp_stateless.py
```diff
@@ -0,0 +1,741 @@
...
+async def _close_bounded(stack: AsyncExitStack) -> None:
+    """Close the transport without letting shutdown outlive or replace a result.
+
+    Shielded and bounded (R-EXEC-1). A definitive result is already in hand by
+    the time this runs, so a transport that will not close, or a request whose
+    task is being cancelled, must not turn a completed mutation into an error
+    the caller might retry. A close that overruns its deadline is cancelled and
+    awaited so stalled shutdowns cannot accumulate detached transport tasks.
+    """
+    closer = asyncio.ensure_future(stack.aclose())
```
khaledosman: `_close_bounded` closes the transport from a *different* task, which cancels the request. Both `_discover_once` (L553) and `_execute_once` (L718) hand `stack.aclose()` to `asyncio.ensure_future`. `mcp.ClientSession.__aexit__` exits an anyio task group, and `streamablehttp_client` exits an anyio cancel scope, both entered in the request task. Exiting them from the closer task (a) fires `cancel_scope.cancel()` against the request task and (b) raises `RuntimeError: Attempted to exit cancel scope in a different task than it was entered in`.

Driving the real `_close_bounded` over a real `ClientSession`:

```
WARNING  Stateless MCP cleanup failed error_class=CancelledError
handler result: RESULT
REQUEST FAILED AFTER THE CALL SUCCEEDED: CancelledError
```

So the tool has run, `CallToolResult` is in hand, and the request task dies in the `finally` before Starlette can send it. That inverts R-EXEC-1: the shield turns a completed mutation into a dropped connection.

Fix by keeping entry and exit in one task: run open -> call -> close inside a single `asyncio.Task` and shield *that*, not just the close. Every existing test monkeypatches `mcp_stateless.open_session` with a plain `asynccontextmanager`, which is why CI is green; the regression test needs a real `ClientSession`.

src/gateway/api/routes/mcp.py
```diff
@@ -0,0 +1,683 @@
...
+def _classify(exc: HTTPException) -> tuple[str, ExecutionState, int]:
+    """Map an authentication or platform refusal onto this contract's enums.
+
+    Every one of these is raised before dispatch, so all of them are
+    ``not_started``. The platform's own detail is dropped rather than forwarded:
+    it may describe a workspace, a plan, or a stored server, and R-ERR-1 lets
+    nothing platform-side through.
+    """
+    if exc.status_code in {401, 402, 403}:
+        return CODE_AUTHENTICATION_FAILED, ExecutionState.NOT_STARTED, 401
```
khaledosman, line 235: Platform 402 and 403 collapse into `401 authentication_failed`. A hybrid caller whose workspace is out of funds or forbidden is told to re-authenticate, which cannot fix either. `.github/instructions/security-review.instructions.md` requires the status contract stay stable (402 funds, 403 forbidden/over budget). Map them to 402/403 with their own codes; the platform detail can still be dropped.

src/gateway/api/routes/mcp.py
```diff
@@ -0,0 +1,683 @@
...
+    if exc.status_code == 404:
+        return CODE_SERVER_NOT_FOUND, ExecutionState.NOT_STARTED, 404
+    if exc.status_code == 429:
+        return CODE_RATE_LIMIT_EXCEEDED, ExecutionState.NOT_STARTED, 429
+    return CODE_RESOLUTION_FAILED, ExecutionState.NOT_STARTED, 502
```
khaledosman, line 246: The catch-all reports client and infrastructure errors as `502 mcp_resolution_failed`. FastAPI raises `HTTPException(400, "There was an error parsing the body")` for any body-parse failure that is not a `JSONDecodeError` (e.g. a ~50k-deep JSON body, where `json.loads` raises `RecursionError`), so a client error is labeled a gateway resolution failure; and the 503 raised at L324 can never reach a caller. Pass 400 through as `invalid_request`/422, and 503 as itself.

src/gateway/api/routes/mcp.py
```diff
@@ -0,0 +1,683 @@
...
+            track_request(raw_request, endpoint=EXECUTE_ENDPOINT, model=EXECUTE_LABEL)
+
+            result = await execute_stored_tool(
+                server,
+                request.tool_name,
+                request.arguments,
+                on_dispatch=mark_dispatched,
+                timings=timings,
+            )
+    except TimeoutError:
```
khaledosman, line 658: A timed-out execution logs nothing: this arm raises without `_log_outcome`, unlike the discovery equivalent at L539 and the `McpExecutionError` arm at L649. The one outcome an operator most needs (`outcome_unknown` from the total deadline) is the one that leaves no line. Call `_log_outcome` before raising.

src/gateway/services/mcp_client.py
```diff
@@ -148,7 +149,7 @@ async def _connect(self, cfg: McpServerConfig) -> _ConnectedServer:
         await session.initialize()
 
         listed = await session.list_tools()
-        allowed = set(cfg.allowed_tools) if cfg.allowed_tools else None
+        allowed = set(cfg.allowed_tools) if cfg.allowed_tools is not None else None
```
khaledosman, line 152: `allowed_tools: []` flips from allow-all to deny-all in the managed loop. Right default, but it changes existing `/v1/chat/completions` behavior for an inline `mcp_servers` entry or a stored server with an empty list, and neither the PR body nor `docs/mcp.md:336` ("optional allow-list") says so. Call it out so the release notes carry it, and state the three-state meaning where the managed loop documents the field.

src/gateway/services/mcp_client.py
```diff
@@ -186,6 +187,18 @@ def server_name_for_tool(self, name: str) -> str | None:
         """Return the configured server that owns ``name``, if connected."""
         return self._tool_owner.get(name)
 
+    async def call_tool_result(self, name: str, arguments: dict[str, Any]) -> CallToolResult:
```
khaledosman: `call_tool_result` is new public surface with no caller but `call_tool_outcome` and its own test. Make it `_call_tool_result` if it is only an extraction.

docs/mcp.md
```diff
@@ -24,6 +24,275 @@ When the model emits an MCP tool call, Otari:
...
+discovery cannot starve calls an application has already authorized.
+
+The 1 MiB transport ceiling is enforced while each response stream is read, so
+chunked JSON and SSE responses cannot bypass it by omitting `Content-Length`.
+Otari requests identity encoding and refuses compressed responses before
+reading their bodies, preventing a small encoded response from expanding past
+the ceiling during decompression. The decoded MCP result is measured again
+before it is returned.
+
+Redirects are disabled. Otari returns a redirect response rather than following
```
khaledosman: "Otari returns a redirect response rather than following it" is not what happens. httpx's `raise_for_status` treats a 3xx as an error, so the SDK raises and the caller gets `mcp_connection_failed` (discovery) or `mcp_outcome_unknown` (execution). No 3xx reaches the caller.

khaledosman (approved):
`make lint`, `make openapi-check`, `make postman-check` clean on `27853dd`; 294 mcp unit tests green; both generated artifacts and `web/src/client/schema.ts` committed; conventional title; no em dashes in the prose diff.

Approving as requested, but the first inline finding is a blocker in practice: on a real MCP server both endpoints compute the answer and then lose it in cleanup. Please land that fix before enabling this for anyone.

<sub>Reviewed with [Claude Code](https://claude.com/claude-code) (Opus 5)</sub>

src/gateway/services/mcp_stateless.py
HareeshBahuleyan: Fixed in 7ed5ca39. Discovery and execution now run the full session lifecycle in one owner task, so the anyio contexts are entered and exited by the same task. Added real ClientSession regression tests for both paths.

src/gateway/api/routes/mcp.py
HareeshBahuleyan: Fixed in 7ed5ca39. Platform 401 remains authentication_failed, 402 is now payment_required, and 403 is now forbidden. The responses keep fixed safe details, with tests and generated API artifacts updated.

src/gateway/api/routes/mcp.py
HareeshBahuleyan: Fixed in 7ed5ca39. The route now catches Starlette HTTPException, normalizes body-parser 400s to 422 invalid_request, and preserves local 503 as service_unavailable. Added regression tests for both cases.

src/gateway/api/routes/mcp.py
HareeshBahuleyan: Fixed in 7ed5ca39. The total-timeout arm now records the classified outcome before raising, and the test asserts that the timeout is logged.

src/gateway/services/mcp_client.py
HareeshBahuleyan: Fixed in 7ed5ca39. The managed-loop docs and PR notes now state the three values explicitly: omitted or null is unrestricted, an empty list denies all tools, and a non-empty list admits only those names.

src/gateway/services/mcp_client.py
HareeshBahuleyan: Fixed in 7ed5ca39. Renamed the extraction to _call_tool_result and updated its internal caller and test.

docs/mcp.md
HareeshBahuleyan: Fixed in 7ed5ca39. The docs now state that redirects are neither followed nor forwarded, and document the connection_failed versus outcome_unknown mapping.

HareeshBahuleyan:
Tested the caller-orchestrated MCP flow locally against Linear's hosted MCP server.

- Built and ran this branch with Docker Compose, then registered Linear through the dashboard using its Streamable HTTP endpoint and a stored Linear API key.
- Tool discovery returned HTTP 200 with 65 tools, no warnings, and a server revision.
- Executed `get_workspace` through `POST /v1/mcp/execute` and received the expected Linear workspace.
- Simulated caller-owned approval by displaying a `save_issue` payload, waiting for approval outside Otari, and only then sending the execution request once. Linear created the dummy issue successfully.
- Executed `get_issue` in a separate request and fetched the same title, description, team, and status, confirming the write persisted.

All MCP calls went through Otari rather than directly to Linear. This verifies the local happy path for stored-server resolution, credential use, discovery, revision binding, caller-orchestrated approval, and read/write tool execution.

PR #815, opened 2026-08-26
feat(dashboard): render usage deltas as trend chips

### LINKED ISSUE #814: Usage page still renders deltas as glyph text, not the trend chip added in #801
## What

`TrendChip` (added in #801) has no call sites. The Usage page still renders period-over-period
change as plain glyph text through `DeltaHint`: `▲ 212.2% vs prev`.

#801 deferred this on purpose. Its description says "No call sites yet, on purpose", on the
grounds that migrating `DeltaHint`'s call sites is a behavior change to two pages and belongs in
its own PR. That follow-up was never opened, so the component has sat unused and the tiles still
show the glyph.

## Why it matters

- The glyph carries direction in a decorative character a screen reader skips. `TrendChip`
  announces the direction, and the judgment ("worse") where the metric has a polarity.
- `▲` and `▼` are the same color whichever way the number went, so a rise in spend and a rise in
  cache hit rate read identically. `TrendChip` takes a `polarity` and paints the two differently
  while the arrow keeps telling the truth about direction.
- The chip is what the design canvas draws for this. The dashboard currently draws something else.

## Scope

The five KPI tiles at the top of the Usage page (`web/src/features/usage/UsagePage.tsx`):

| Tile | Polarity |
| --- | --- |
| Tracked cost | `down-is-good` |
| Requests | `neutral` (volume) |
| Tokens (billed) | `neutral` (volume) |
| Cache hit rate | `up-is-good` |
| Avg latency | no delta on the wire, unchanged |

`StatCard` needs somewhere to put a chip: today the delta is passed inside `hint`, which is prose.

`OverviewPage` has two more `DeltaHint` call sites. Out of scope here so the diff stays one page,
but they are the rest of the migration and `DeltaHint` cannot retire until they move.

jigjigjig:
## Description

`TrendChip` landed in #801 with no call sites. #801 deferred them on purpose ("No call sites yet,
on purpose"), on the grounds that migrating `DeltaHint` is a behavior change to two pages and
belongs in its own PR. That follow-up was never opened, so the Usage page still printed its
period-over-period change as `▲ 212.2% vs prev`: an arrow a screen reader skips, in one color
whichever way the number went, so a rise in spend and a rise in cache hit rate read identically.

This is that follow-up, for the Usage page.

### The chips

The four tiles that have a delta now render a chip that knows the metric's own axis:

| Tile | Polarity | Why |
| --- | --- | --- |
| Tracked cost | `down-is-good` | Spend falling is the improvement |
| Requests | `neutral` | Volume. More traffic is neither a win nor a regression on its own, and the error rate beside it carries the judgment |
| Tokens (billed) | `neutral` | Same |
| Cache hit rate | `up-is-good` | A higher share of reads served from cache is the win |
| Avg latency | n/a | No delta on the wire, untouched |

So a fall in spend and a rise in cache hit rate both paint as the improvement they are, while the
arrow keeps telling the truth about which way the number went. `neutral` is a deliberate choice
rather than an omission on the two volume tiles: painting rising traffic green would assert
something the gateway does not know.

### The layout

`StatCard` gains a `trend` slot. The chip and the hint share one row under the value, because
they are two halves of the same aside: the chip says how the number moved, the hint says what it
is made of ("5.8% errors", "311.2k read · 0 written"). Stacked they read as two statements and
cost a line the tile does not have.

```
CACHE HIT RATE
13.0%
[↑ +0.4% vs prev]  311.2k read · 0 written
```

The row is a flex row, which is also what makes the chip hug its text: HeroUI's `Chip` is
inline-flex, and as a stretched child of `Card.Content`'s column it ran the full width of the
tile as a pill-shaped band. `flex-wrap` is the fallback at five-up, not the layout.

Moving the delta out of `hint` also left the billed-tokens tile with no hint at all, since the
delta was all it had, and its sparkline rode above its neighbors'. The reservation #807 added for
a wrapping hint now also covers a charted tile with nothing to say, so a row of tiles keeps its
charts on one line. A tile with neither hint nor chart still reserves nothing, so a lone tile
carries no dead space.

Self-review caught that reservation being 4px short: #807 sized it as two 18px lines of plain
text, and a wrapped row is now an 18px chip line plus `gap-y-1` plus an 18px hint line, so 40px.
Raised to match. A hint long enough to wrap on its own still overruns it, so it is a floor and not
a guarantee, and the comment says so.

### Not in scope

`DeltaHint` stays. `OverviewPage` has two more call sites, and moving them is the rest of the
migration; `DeltaHint` can retire in that PR. Keeping this diff to one page is what #801 asked for.

### Verification

`pnpm --dir web run lint`, `pnpm --dir web run typecheck` and `pnpm --dir web test` all pass:
104 files, 1525 tests. Three tests are new (two on `StatCard`'s layout, one asserting the Usage
tiles go through a chip and no longer render the glyph); `TrendChip.test.tsx` already owns the
direction and polarity mapping, so the page test does not re-cover it.

Worth recording, because it cost time: the dashboard suite fails wholesale on Node 26 (554
failures, renders accumulating in one document) and passes on Node 22, which is what
`otari-dashboard.yml` uses. Run it on 22.

No API contract changed, so the OpenAPI spec and the Postman collection are untouched. No page was
added, so no screenshot entry is owed. The bundle is not committed.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes #814

## Checklist

Dashboard-only change, so the checks that apply are the dashboard's, listed under Verification
above. No Python changed, so `make lint` and `make test` have nothing to say about this diff.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

Written against a local gateway seeded with `scripts/seed_usage_smoke.py`, so the tiles were read
with real deltas in them rather than from the test fixtures alone.

## Summary

- Replaced glyph-based Usage KPI deltas with `TrendChip`.
- Applied the correct trend meaning to cost, requests, billed tokens, and cache hit rate.
- Added a dedicated trend area to `StatCard` and preserved tile alignment.
- Added regression tests for Usage rendering and `StatCard` layout.

### DIFF
--- a/web/src/features/usage/UsagePage.test.tsx
+++ b/web/src/features/usage/UsagePage.test.tsx
@@ -187,6 +187,10 @@ function jsonResponse(body: unknown, status = 200): Response {
 function mockApi(
   body: UsageSummary | null,
   extra: Record<string, unknown> = {},
+  // The previous window. Without one, that query gets the current window's body
+  // back, every delta is exactly zero, and a test cannot see a direction or a
+  // polarity color.
+  previousBody?: UsageSummary,
 ) {
   return vi.spyOn(globalThis, "fetch").mockImplementation(async (input) => {
     const url = String(input)
@@ -194,6 +198,12 @@ function mockApi(
       if (url.includes(path)) return jsonResponse(answer)
     }
     if (url.includes("/v1/usage/summary")) {
+      // The previous-period query is the only one on this page passing
+      // NO_BREAKDOWNS, which goes on the wire as the server's `none` sentinel,
+      // so that is what tells the two windows apart here.
+      if (previousBody && url.includes("dimensions=none")) {
+        return jsonResponse(previousBody)
+      }
       return jsonResponse(body ?? summary())
     }
     if (url.includes("/v1/usage/series")) {
@@ -616,11 +626,41 @@ describe("UsagePage", () => {
     // And it announces a direction, which the glyph never did: "▲" is
     // decoration a screen reader skips. TrendChip.test.tsx owns the direction
     // and polarity mapping; this only asserts the tiles go through it.
+    // Anchored, and counted. Unanchored, `up` also matches the page's own
+    // description and the "No grouping" option, so the assertion would pass with
+    // the announcement deleted. The count is the four tiles that have a delta,
+    // less cache hit rate, which this fixture gives no input-token composition
+    // to divide by.
     expect(
-      screen.getAllByText(/^(no change|up|down)(, (better|worse))?$/).length,
-    ).toBeGreaterThan(0)
+      screen.getAllByText(/^(no change|up|down)(, (better|worse))?$/),
+    ).toHaveLength(3)
     // The hand-rolled arrow glyphs are gone from the tiles.
     expect(screen.queryByText(/[▲▼]/)).not.toBeInTheDocument()
+  })
+
+  it("reads a chip against the metric's own polarity, not the direction alone", async () => {
+    // Spend rose from 827.00 to 1,240.50, a 50% rise. On `down-is-good` that is
+    // the regression, so the chip is a danger chip and says so: direction plus
+    // judgment, because polarity puts good and bad in hue alone.
+    mockApi(
+      summary(),
+      {},
+      summary({
+        totals: usageTotals({
+          cost: 827,
+          request_count: 42_000,
+          billed_input_tokens: 4_000_000,
+          billed_output_tokens: 2_200_000,
+        }),
+      }),
+    )
+    renderPage(<UsagePage />)
+
+    expect(await screen.findByText("up, worse")).toBeInTheDocument()
+    expect(screen.getByText("+50.0% vs prev")).toBeInTheDocument()
+    // Requests doubled too, but volume carries no polarity, so it announces the
+    // direction and nothing more.
+    expect(screen.getAllByText("up").length).toBeGreaterThan(0)
   })
 
   it("lists spend by model with a reconciling 'other' fold row", async () => {
--- a/web/src/features/usage/UsagePage.test.tsx
+++ b/web/src/features/usage/UsagePage.test.tsx
@@ -616,7 +616,9 @@ describe("UsagePage", () => {
     // And it announces a direction, which the glyph never did: "▲" is
     // decoration a screen reader skips. TrendChip.test.tsx owns the direction
     // and polarity mapping; this only asserts the tiles go through it.
-    expect(screen.getAllByText(/no change|up|down/).length).toBeGreaterThan(0)
+    expect(
+      screen.getAllByText(/^(no change|up|down)(, (better|worse))?$/).length,
+    ).toBeGreaterThan(0)
     // The hand-rolled arrow glyphs are gone from the tiles.
     expect(screen.queryByText(/[▲▼]/)).not.toBeInTheDocument()
   })
--- a/web/src/features/usage/UsagePage.test.tsx
+++ b/web/src/features/usage/UsagePage.test.tsx
@@ -605,6 +605,22 @@ describe("UsagePage", () => {
     expect(summaryCalls.some((u) => !u.includes("end_date="))).toBe(true)
   })
 
+  it("renders period-over-period change as a trend chip, not a glyph", async () => {
+    mockApi(summary())
+    renderPage(<UsagePage />)
+    await screen.findByText("$1,240.50")
+
+    // The chip carries the caption the old plain-text hint carried, so the
+    // comparison still says what it is being compared against.
+    expect(screen.getAllByText(/vs prev/).length).toBeGreaterThan(0)
+    // And it announces a direction, which the glyph never did: "▲" is
+    // decoration a screen reader skips. TrendChip.test.tsx owns the direction
+    // and polarity mapping; this only asserts the tiles go through it.
+    expect(screen.getAllByText(/no change|up|down/).length).toBeGreaterThan(0)
+    // The hand-rolled arrow glyphs are gone from the tiles.
+    expect(screen.queryByText(/[▲▼]/)).not.toBeInTheDocument()
+  })
+
   it("lists spend by model with a reconciling 'other' fold row", async () => {
     mockApi(summary())
     renderPage(<UsagePage />)
--- a/web/src/features/usage/UsagePage.tsx
+++ b/web/src/features/usage/UsagePage.tsx
@@ -31,8 +31,8 @@ import {
 } from "@/shared/components/charts"
 import { DataTable, type DataTableColumn } from "@/shared/components/DataTable"
 import { type FilterChip, FilterChips } from "@/shared/components/FilterChips"
+import { TrendChip } from "@/shared/components/TrendChip"
 import {
-  DeltaHint,
   EmptyState,
   ErrorBanner,
   FilterMultiComboBox,
@@ -64,7 +64,7 @@ import { useSelectedWorkspace } from "@/shared/hooks/SelectedWorkspace"
 // ---------- formatting ----------
 
 // Compact currency (formatUsd), token counts (formatTokens), percentages
-// (formatPct) and the period-over-period helpers (deltaFraction / DeltaHint) are
+// (formatPct) and the period-over-period helpers (deltaFraction / TrendChip) are
 // shared with the overview page from @/shared/helpers/format and @/shared/components/ui. Only the
 // two formatters specific to this page stay local.
 function formatBucketLabel(iso: string, bucket: UsageBucket): string {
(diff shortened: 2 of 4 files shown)


web/src/features/usage/UsagePage.tsx
```diff
@@ -1051,15 +1051,19 @@ export function UsagePage() {
...
               value={totals ? formatUsd(totals.cost) : "—"}
+              // Spend falling is the improvement, so a rise paints danger while
+              // the arrow keeps telling the truth about which way it went.
+              trend={
+                <TrendChip
+                  fraction={costDelta}
+                  polarity="down-is-good"
+                  caption="vs prev"
+                />
+              }
```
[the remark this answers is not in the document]
jigjigjig: Addressed in 95c87b30: each tile's delta is a named const now, and every chip is gated on it being non-null, so the aside row is not reserved for a chip that draws nothing. Requests and cache hit rate had the same hole, since deltaFraction also returns null when the previous value is 0.

web/src/features/usage/UsagePage.tsx
```diff
@@ -64,7 +64,7 @@ import { useSelectedWorkspace } from "@/shared/hooks/SelectedWorkspace"
 // ---------- formatting ----------
 
 // Compact currency (formatUsd), token counts (formatTokens), percentages
-// (formatPct) and the period-over-period helpers (deltaFraction / DeltaHint) are
+// (formatPct) and the period-over-period helpers (deltaFraction / TrendChip) are
 // shared with the overview page from @/shared/helpers/format and @/shared/components/ui. Only the
 // two formatters specific to this page stay local.
```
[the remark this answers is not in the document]
jigjigjig: Addressed in 95c87b30: the comment now points deltaFraction at @/shared/helpers/format and TrendChip at its own module, and stops calling the chip a helper.

web/src/features/usage/UsagePage.test.tsx
```diff
@@ -605,6 +605,22 @@ describe("UsagePage", () => {
...
+
+    // The chip carries the caption the old plain-text hint carried, so the
+    // comparison still says what it is being compared against.
+    expect(screen.getAllByText(/vs prev/).length).toBeGreaterThan(0)
+    // And it announces a direction, which the glyph never did: "▲" is
+    // decoration a screen reader skips. TrendChip.test.tsx owns the direction
+    // and polarity mapping; this only asserts the tiles go through it.
+    expect(screen.getAllByText(/no change|up|down/).length).toBeGreaterThan(0)
+    // The hand-rolled arrow glyphs are gone from the tiles.
+    expect(screen.queryByText(/[▲▼]/)).not.toBeInTheDocument()
```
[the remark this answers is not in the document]
jigjigjig: Addressed in 95c87b30: anchored to the phrases announce() actually produces, so "breakdown" no longer satisfies it. Good catch, the assertion passed with no chip on the page.

web/src/shared/components/ui.tsx
```diff
@@ -200,14 +209,37 @@ export function StatCard({
...
+          #807 reserved 36px here for two lines of plain text, which the chip's
+          own line height now overruns by 4px. A hint long enough to wrap on its
+          own still overruns this, so it is a floor and not a guarantee; the
+          type is deliberately not sized to the longest string, which would be
+          sizing it by accident. Reserved for a charted tile even with neither
+          chip nor hint, since the chart is what makes the misalignment visible;
+          a tile with no chart and nothing to say reserves nothing, so a lone
+          tile carries no dead space. */}
+      {trend || hint || chart ? (
+        <span className="flex min-h-10 flex-wrap items-center gap-x-2 gap-y-1 text-xs tabular-nums text-muted">
```
khaledosman: `min-h-10` is still 2px short of what the wrap costs, so the head commit does not close the gap it names.

The comment above sizes the chip line at 18px. It is 20px: HeroUI's `.chip--sm` does not reset the base `.chip`'s `--tw-leading`, so its `line-height` falls through to the base value rather than to the `--text-xs--line-height` fallback. From the compiled `@heroui/styles`:

```css
.chip{ …; line-height:var(--tw-leading,var(--text-xs--line-height)); --tw-leading:calc(var(--spacing) * 5); line-height:calc(var(--spacing) * 5) }
.chip--sm{ padding-inline:calc(var(--spacing) * 1); padding-block:calc(var(--spacing) * 0); font-size:var(--text-xs); line-height:var(--tw-leading,var(--text-xs--line-height)) }
```

`--tw-leading` is set on the same element, so `.chip--sm` resolves to 20px. `--spacing` is `0.25rem` and `--text-xs--line-height` is 18px, so a wrapped row costs 20 + 4 + 18 = 42px against a 40px floor. A wrapping tile still drops its sparkline below its neighbors', by 2px instead of by a line.

`min-h-10.5` compiles to exactly 42px. Whichever value you land on, the "an 18px chip line" in the comment needs correcting too.

web/src/features/usage/UsagePage.tsx
```diff
@@ -1051,15 +1051,19 @@ export function UsagePage() {
             <StatCard
               label="Tracked cost"
               value={totals ? formatUsd(totals.cost) : "—"}
+              // Spend falling is the improvement, so a rise paints danger while
+              // the arrow keeps telling the truth about which way it went.
+              trend={
+                <TrendChip
```
khaledosman: This is the one of the four tiles whose `trend` is not guarded on data, and it is the tile that pays for it.

`costDelta` is `null` until the summary lands, so `TrendChip` renders nothing, but the *element* is truthy and `trend || hint || chart` in `StatCard` tests the element rather than what it renders. Rendering the page with the summary request in flight gives 5 tiles and 1 reserved row: 40px of dead space on the tile with nothing to say, which is the case the comment on that condition says it avoids.

Guard it like its three neighbors:

```tsx
trend={
  totals ? (
    <TrendChip fraction={costDelta} polarity="down-is-good" caption="vs prev" />
  ) : null
}
```

web/src/features/usage/UsagePage.test.tsx
```diff
@@ -605,6 +605,22 @@ describe("UsagePage", () => {
...
+    renderPage(<UsagePage />)
+    await screen.findByText("$1,240.50")
+
+    // The chip carries the caption the old plain-text hint carried, so the
+    // comparison still says what it is being compared against.
+    expect(screen.getAllByText(/vs prev/).length).toBeGreaterThan(0)
+    // And it announces a direction, which the glyph never did: "▲" is
+    // decoration a screen reader skips. TrendChip.test.tsx owns the direction
+    // and polarity mapping; this only asserts the tiles go through it.
+    expect(screen.getAllByText(/no change|up|down/).length).toBeGreaterThan(0)
```
khaledosman: This assertion passes with the feature deleted.

The alternation is unanchored, so `up` matches the page's own header prose. Removing the `sr-only` `announce` span from `TrendChip` entirely and rerunning this test, it still passes, on these two matches:

```
<P>Spend, tokens, cache use, and request volume over time. Grou…
<OPTION>No grouping
```

Neither is a chip. Anchor the matcher and pin the count, so the assertion fails when the announcement stops rendering:

```tsx
expect(
  screen.getAllByText(/^(up|down|no change)(, (better|worse))?$/),
).toHaveLength(3)
```

web/src/features/usage/UsagePage.test.tsx
```diff
@@ -605,6 +605,22 @@ describe("UsagePage", () => {
     expect(summaryCalls.some((u) => !u.includes("end_date="))).toBe(true)
   })
 
+  it("renders period-over-period change as a trend chip, not a glyph", async () => {
+    mockApi(summary())
```
khaledosman, line 619: `mockApi` answers every `/v1/usage/summary` with one body, so the previous-window query gets the current window back and every delta is exactly 0. All three chips on this page render `no change`, so the test never sees a direction or a polarity color: the two things the PR is about.

The previous request is distinguishable, since it carries a different `start_date` and `NO_BREAKDOWNS`. Branch the mock on the window and assert one concrete chip, for instance the cost tile reading `up, worse` when spend rose.

khaledosman (approved):
Four findings inline, each verified by running the code rather than by reading it. Two of them undercut claims the PR makes about itself: the reservation raised in the head commit is still short, and the new page test passes with the feature deleted.

Repo gates are clean. No API contract change, so no OpenAPI or Postman regeneration is owed; `web/src/client/schema.ts` and `web/src/routeTree.gen.ts` are untouched; no new dependency, so no lockfile or `allowBuilds` entry; no new page, so no screenshot entry; `TrendChip` sits in `shared/` and the page in `features/`, which the layer rules allow; and the PR title is a conventional commit, which is what git-cliff parses on squash.

web/src/shared/components/ui.tsx
jigjigjig: Addressed in d5415901: 42px via min-h-10.5, and the comment now says 20px and why. Confirmed the fall-through in the compiled CSS, and that min-h-10.5 emits calc(var(--spacing) * 10.5) against a 0.25rem --spacing. Thanks, I had read .chip--sm's line-height as resolving to the token fallback and missed that --tw-leading is set on the same element.

web/src/features/usage/UsagePage.tsx
jigjigjig: Already guarded at HEAD: 95c87b30 (pushed 14:49Z, just before this review) made every tile's delta a named const and gated each chip on it, so the cost tile reads costDelta !== null. Worth noting the suggested totals ? guard would be looser than that: costDelta is also null once totals lands but the previous window has not, or when the previous cost is 0, and TrendChip renders nothing for a null fraction either way. Gating on the fraction covers both.

web/src/features/usage/UsagePage.test.tsx
jigjigjig: Anchored at HEAD already, in 95c87b30. Took the count pin in d5415901: toHaveLength(3), the four tiles with a delta less cache hit rate, which this fixture gives no input-token composition to divide by. That is the half that makes it fail when the announcement stops rendering.

web/src/features/usage/UsagePage.test.tsx
jigjigjig: Addressed in d5415901: mockApi takes an optional previous-window body and tells the two apart on dimensions=none, which only that query sends. A new case drops previous spend to 827.00 and asserts the cost tile reads "up, worse" and "+50.0% vs prev". This was the real gap, every chip was rendering no change.

PR #824, opened 2026-08-27
fix(api): give hosted mode a management-only posture

### LINKED ISSUE #822: Hosted control plane serves the data-plane inference endpoints unbilled
## Impact

In hosted mode (`OTARI_MODE=hosted`, the multi-tenant control plane), the data-plane inference endpoints are mounted and served **unbilled**. A workspace API key pointed at the control plane host reaches `/v1/chat/completions` (and `/v1/messages`, `/v1/responses`, `/v1/embeddings`, `/v1/images`, `/v1/audio`, `/v1/rerank`, `/v1/batches`, `/v1/moderations`, `/v1/search`), runs inference, and the organization wallet is never debited.

Verified live on a hosted control plane: added a real OpenAI key via `POST /v1/hosted-backends`, minted a key in a funded org with no BYO openai key, and ran `openai:gpt-4o-mini`. Result: `200`, a real reply, a usage event with cost `$0.000002`, and the wallet balance **unchanged** (funded org, $5.00).

## Root cause

- `src/gateway/api/main.py:_register_core_routers` mounts `chat`/`messages`/`responses` in both modes, then returns early only for `is_hybrid_mode`; standalone and hosted fall through and mount every data-plane router alongside the management API. There is no management-only posture.
- The control plane's register (`overlay.gateway_bootstrap`) binds `EntitlementPort`, `ModelProviderPort`, `TelemetryStoragePort` but **not** `BillingPort`, so otari's default `NullBillingAdapter` (`container.py:166`) is used. The wallet is debited only on the gateway usage-report path (`overlay/services/gateway_usage_service.py`).

## Intended model

Customer inference belongs on the data-plane gateway (hybrid), which resolves credentials via the control plane and reports usage back, which is what debits the wallet (a real `mzai:` debit proves it). The control plane is documented as one that "never receives inference traffic", but nothing enforces that.

## Wanted

A management-only posture so hosted mode does not serve the data-plane inference routers (a direct hit returns 404), or otherwise refuses inference on the control plane. Immediate mitigation lives in infra (a higher-priority ALB rule on the control-plane host returning 403 for the inference paths); this issue is the otari-side fix. See the sibling dashboard issue for how users are currently pointed here.

njbrake:
## Description

A hosted control plane mounted every data-plane router alongside the management API, so a workspace key pointed at the control plane host ran inference that no wallet was debited for. Verified before the change: hosted mounted all ten inference paths identically to standalone.

`_register_core_routers` mounted chat/messages/responses before any mode check and returned early only for hybrid, so hosted fell through the standalone block. The data plane is now gated on `not is_hosted_mode`.

**Not mounting is only half of it.** A bare 404 tells a customer who pointed an SDK at the wrong host nothing, and an operator who mis-set `OTARI_MODE` less. `hybrid_mode.py` already solved this for the opposite plane by doing both, so `management_only.py` mirrors it: catch-alls over the inference prefixes with a diagnostic body, mounted last so a contributed router still wins. Status stays 404 to match, since it reports the same fact: this process does not host that plane. 403 would invite a retry with a different key; 501 would misreport a posture as a missing feature.

Standalone is unchanged and still serves both planes from one process, proven by diffing route sets (paths and methods) against `main` for standalone and hybrid. `/v1/models` stays on hosted: discovery is not dispatch. `/v1/files` is gated because an upload only exists to feed a completion or a batch.

**On the issue's second cause:** `BillingPort` has no core call sites (`ARCHITECTURE.md` says so), so binding it in otari-ai would be a no-op today. Worth a separate discussion, not this PR.

The mode-keyed fix inherits the mode, so I would keep the host-keyed ALB rule rather than retire it.

**The refusal covers the reads and deletes too, deliberately.** A control plane already serving this traffic may hold files and batches, and their owners lose the API to them here rather than keeping a read-only window onto a plane the deployment no longer serves. Nothing stored is deleted.

## PR Type

- [x] Bug Fix

## Relevant issues

Closes #822

## Checklist

`make lint`, `make typecheck`, `make openapi-check` and `make postman-check` pass. 2566 unit and 1936 integration tests pass. The committed spec needs no regeneration: it is generated from a standalone app, which this change does not alter.

## AI Usage

**AI Model/Tool used:**

Written by Claude Opus 5 through back-and-forth with @njbrake, who made the design calls: refuse via stubs rather than a bare 404, keep the status at 404, and gate `/v1/files`. Reviewed by a separate Claude Opus 5 agent given only the PR number, which caught a factually wrong code comment, five stale mode descriptions elsewhere in the repo, and a missing test; those are the two follow-up commits.

## Summary

Hosted mode now serves management routes only. Inference, file, and batch requests return `404` and direct users to the data-plane gateway.

Standalone mode remains unchanged. Hosted mode retains model discovery and publishes the configured data-plane URL.

This prevents unbilled inference on the hosted control plane. Tests and documentation cover the updated behavior.

## Technical notes

- Added last-priority fallback routes for disabled hosted endpoints.
- Added hosted data-plane URL configuration and bootstrap discovery.
- Updated tests to use PostgreSQL-backed fixtures.
- Documented the management-only behavior and billing rationale.


src/gateway/api/routes/management_only.py
```diff
@@ -0,0 +1,128 @@
...
+hold files and batches, and their owners lose the API to them here rather than
+keeping a read-only window onto rows the plane no longer serves. That is
+deliberate: a plane is either on a deployment or it is not, and a half-mounted
+one is the state nobody can reason about later. Nothing stored is deleted, so
+the rows outlive the refusal and a data-plane gateway is where they belong.
+"""
+
+from fastapi import APIRouter, HTTPException, status
+
+_DISABLED_DETAIL = (
```
khaledosman: The refusal names no address, but the deployment has one. `bootstrap.py:286` already publishes `config.data_plane_url` for exactly this case, and `docs/modes.md:38-39` claims this 404 points "at the gateway that should have served it". It does not; it says "your Otari gateway", which is what the customer who mis-pointed the SDK already believed they were calling. The PR's own argument (a bare 404 tells whoever hit it nothing) applies to this wording too.

Take `config: Annotated[GatewayConfig, Depends(get_config)]` on the handlers and append `config.data_plane_url` when it is set, keeping the current sentence for the null case the bootstrap already treats as unconfigured. If the address is deliberately left out, correct the `modes.md` sentence instead.

tests/integration/test_hosted_mode_surface.py
```diff
@@ -0,0 +1,158 @@
...
+from fastapi.testclient import TestClient
+
+from gateway.api.deps import reset_config
+from gateway.core.config import GatewayConfig
+from gateway.core.database import reset_db
+from gateway.main import create_app
+
+# Every data-plane path otari#822 found served on the control plane, plus the
+# sub-paths that ride the same catch-all.
+INFERENCE_PATHS = (
```
khaledosman, line 28: Hand-maintained, so a router gated in `main.py` with no matching stub here ships a bare 404 silently, which is the outcome this PR exists to avoid. The gated-router list and the stub prefix list live in two files with nothing tying them together.

Derive it: build both apps and assert `standalone_paths - hosted_paths` is exactly the set the stubs answer. That holds today across all eleven routers, and it turns "gated a router, forgot the stub" into a failure. It cannot catch a new data-plane router added outside the guard, so the mode-diff assertion is the half that pays.

src/gateway/api/routes/management_only.py
```diff
@@ -0,0 +1,128 @@
...
+_METHODS = ["GET", "POST", "PUT", "PATCH", "DELETE", "OPTIONS"]
+
+router = APIRouter(tags=["management-only"])
+
+
+def _raise_disabled() -> None:
+    raise HTTPException(status_code=status.HTTP_404_NOT_FOUND, detail=_DISABLED_DETAIL)
+
+
+@router.api_route("/v1/chat/{path:path}", methods=_METHODS)
```
khaledosman: Twelve copies of the same four-line stub follow, which lets a prefix added later pick up a different `methods=` list unnoticed. Register from a loop over a `(prefix, why)` tuple; the per-prefix reasoning survives as data rather than as comments between near-identical blocks. Worth keeping to this file rather than also reworking `hybrid_mode.py`.

src/gateway/api/routes/management_only.py
```diff
@@ -0,0 +1,128 @@
+"""Data-plane stubs for a hosted control plane, which serves management only.
```
khaledosman: Naming: the sibling is `hybrid_mode.py` and this docstring introduces the module as its mirror image, so `hosted_mode.py` reads as the pair. `management_only` reads as a third concept alongside the two modes.

docs/modes.md
```diff
@@ -21,12 +21,36 @@ All endpoints are available. On first startup, Otari bootstraps an API key and l
...
+Two things change. The first is the data plane, described next; the second is
+the dashboard.
+
+### Hosted mode serves no inference
+
+A control plane runs no customer traffic. `/v1/chat/completions`,
+`/v1/messages`, `/v1/responses`, `/v1/embeddings`, `/v1/images`, `/v1/audio`,
+`/v1/rerank`, `/v1/moderations`, `/v1/search`, `/v1/batches` and `/v1/files` are
+not mounted, and a request to any of them returns `404` saying so and pointing
+at the gateway that should have served it.
```
khaledosman: The read-loss is explained only in the module docstring. An existing hosted deployment loses the API to files and batches it already stores, and `modes.md` is what an operator reads. One sentence here: nothing stored is deleted, but the reads and deletes go with the dispatch.

khaledosman (approved):
Verified independently: every router that dispatches to a provider (`chat`, `messages`, `responses`, `embeddings`, `images`, `audio`, `rerank`, `moderations`, `search`, `batches`, `files`) is inside the `serves_data_plane` guard and covered by a stub prefix, and nothing left mounted in hosted mode reaches `_pipeline`/`_passthrough` on a billable path. `search_tools` is correctly not shadowed by the `/v1/search` catch-all. The spec claim in the description holds: `scripts/generate_openapi.py:35` builds a default (standalone) config, so the stubs never enter it.

Five inline notes, none blocking.

---
_Review by Claude Code (Claude Opus 5)._

src/gateway/api/routes/management_only.py
njbrake: Done in 5f7ff33e6. The handlers now take `config` and the refusal names `data_plane_url` when one is set, falling back to the generic sentence otherwise, which is what bootstrap already treats as unconfigured. You were right that my own argument convicted the wording: "your Otari gateway" is what the caller who mis-pointed the SDK already believed they were calling. The `modes.md` claim is corrected to describe both cases, and there is a test covering the address-naming branch.

src/gateway/api/routes/management_only.py
njbrake: Done in 5f7ff33e6. The stubs now register from a `DATA_PLANE_PREFIXES` tuple of `(prefix, why)`, so a prefix added later cannot pick up a different `methods=` list and the per-prefix reasoning survives as data. Kept to this file; `hybrid_mode.py` is left alone as you suggested.

src/gateway/api/routes/management_only.py
njbrake: Done in 5f7ff33e6: renamed to `hosted_mode.py`, with `main.py`, the tests, `AGENTS.md` and the docs updated to match. Agreed that `management_only` read as a third concept beside the two modes.

tests/integration/test_hosted_mode_surface.py
njbrake: Done in 5f7ff33e6, as `test_every_gated_router_has_a_stub_standing_in_for_it`: it builds both apps and asserts everything standalone serves and hosted does not falls under a prefix the stubs claim. Verified by deleting the `/v1/rerank` prefix, which fails naming the uncovered path, then passes on restore. Your caveat is in the docstring: it cannot catch a new data-plane router added outside the guard, since that one mounts in both modes and never appears in the difference.

docs/modes.md
njbrake: Done in 5f7ff33e6. `modes.md` now says the refusal covers the reads and deletes, that a control plane keeps the files and batches it already stored, and that it loses the API to them without anything being deleted.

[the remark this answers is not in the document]
njbrake: Fixed in 8b58cbb20. The wording said the routers are not mounted and then that a request returns a diagnostic 404, which reads as a contradiction against the code. It now says both halves: the routers are not mounted, and a stub stands in their place so the answer explains itself. The module has since been renamed to `gateway.api.routes.hosted_mode`.

docs/modes.md
```diff
@@ -55,8 +79,8 @@ ownership, so a member of one organization can put a key in another's
 workspace, billed to that organization.
 
 That is fine for the single-tenant deployment this base build is written for,
-and it is why hosted mode changes the surface set and nothing else. Do not read
-it as isolation between tenants. Closing the gap is tracked in
+and it is why hosted mode changes the surface set and the data plane but not the
+authority model. Do not read it as isolation between tenants. Closing the gap is tracked in
 [mozilla-ai/otari-ai#1880](https://github.com/mozilla-ai/otari-ai/issues/1880),
 which scopes the fix to this repository: gating the deployment-wide routers on
```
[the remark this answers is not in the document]
njbrake: Not changing this one. The target is `mozilla-ai/otari-ai#1880`, a private repository, so it 404s for an unauthenticated fetch but resolves for the team it is written for. It is also pre-existing rather than introduced here: `git show origin/main:docs/modes.md` carries the same link at line 60, and this PR only reflowed the paragraph around it.

PR #825, opened 2026-08-27
fix(dashboard): point hosted request snippets at the data-plane gateway

### LINKED ISSUE #823: Hosted dashboard request snippets point at the control plane, not the data-plane gateway
## Impact

In hosted mode the dashboard is served by the control plane, and its "send your first request" snippets point users at the control plane for inference instead of the data-plane gateway. Users who copy the onboarding or Keys-page curl/python example send inference to the wrong host (and, per the sibling issue, an unbilled one).

## Root cause

`web/src/shared/helpers/requestSnippets.ts` builds the base URL from the caller's `origin`, and the callers pass `window.location.origin` (`web/src/features/onboarding/SetupGuideCard.tsx:225`, `web/src/features/keys/KeysPage.tsx:115`). The file's own comment states the assumption: "the base URL is the browser's own origin, because the gateway serves this dashboard." That holds in standalone, where one process is both dashboard and data plane, but not in hosted mode, where the dashboard host is the control plane and the data plane is a separate gateway.

## Wanted

In hosted mode the snippets should target the data-plane gateway base URL rather than `window.location.origin`. That URL needs surfacing to the dashboard, for example on `GET /v1/bootstrap` (today `management_url` is `null`).

## Related

Sibling of #822 (control plane serves inference unbilled): fixing the snippet stops directing users there; that issue stops the control plane serving it.

njbrake:
## Description

In hosted mode the dashboard is served by the control plane, and its "make your first call" snippets pointed users at that same host for inference, which is the one host their traffic should not reach (#822). Both callers passed `window.location.origin`: correct in standalone and hybrid, wrong here.

`GET /v1/bootstrap` now carries `data_plane_url`, the mirror of `management_url`. That one says where management lives when this deployment is not the control plane; this one says where the data plane is when this deployment is not it. Only a hosted deployment publishes it, from a new `data_plane_url` config field (`OTARI_DATA_PLANE_URL`), validated as an absolute http(s) URL with the trailing slash trimmed.

`resolveSnippetBaseUrl` is now the one place the dashboard decides a base URL: the browser's origin for standalone and hybrid, the published address for hosted, and null for a hosted deployment that published none. **Null renders no snippet**, with a line saying why. A placeholder hostname would be a value nobody reading it could replace, and the origin is the bug itself. Hosted mode without the setting warns once at startup rather than refusing to boot, so redeploying a live control plane is not an outage over a dashboard affordance.

Standalone and hybrid behavior is unchanged.

**The live fix needs a second half.** The bootstrap contract is shared with otari.ai, which serves this shape for its own hosted deployment (mozilla-ai/otari-ai#1591). Nothing changes on the real control plane until it populates `data_plane_url`.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Closes #823

## Checklist

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

Claude Opus 5, through the Claude Code CLI.

Written by the model working through the issue with @njbrake; the reasoning and the design decisions (the field name, and rendering no snippet rather than a placeholder when the address is unset) are his. Reviewed by a separate agent with no part in writing it, and the findings from that review are being worked through with him.

## Summary

- Hosted deployments now use the configured data-plane gateway in dashboard request snippets.
- Standalone and hybrid deployments continue to use the browser address.
- Hosted deployments without a gateway address hide snippets and show guidance instead.
- Added URL validation, safe snippet escaping, startup warnings, documentation, and shared dashboard messaging.
- Added coverage for configuration, bootstrap responses, routing behavior, and UI states.

## Technical notes

- Added `data_plane_url` to `/v1/bootstrap`.
- Added `OTARI_DATA_PLANE_URL` configuration support.
- Rejects non-HTTP(S) URLs, query strings, and fragments.
- Trims trailing slashes before use.


web/src/shared/helpers/requestSnippets.ts
```diff
@@ -59,14 +66,14 @@ export interface RequestSnippetInput {
...
-  origin,
+  baseUrl,
   apiKey,
   model = SNIPPET_MODEL_PLACEHOLDER,
   message = DEFAULT_MESSAGE,
 }: RequestSnippetInput): string {
   const body = `{"model": ${literal(model)}, "messages": [{"role": "user", "content": ${literal(message)}}]}`
   return [
-    `curl ${origin}/v1/chat/completions \\`,
+    `curl ${baseUrl}/v1/chat/completions \\`,
```
khaledosman: `baseUrl` is now the only interpolation in this file that is neither shell-quoted nor `literal()`-escaped, and it just became the one value that comes from the server instead of `window.location.origin`. `_validate_data_plane_url` (`src/gateway/core/config.py:1757`) checks only scheme and netloc, so all three of these load without error:

```
https://gw.example/x" ; echo pwned ; "     -> escapes the Python string literal at line 93
https://gw.example/a'; echo pwned; '       -> escapes the curl line here
https://gw.example/a b                     -> splits the curl argument in two
```

The docstrings on `literal` and `shellSingleQuoted` already state why this matters; apply them to `baseUrl` too:

```ts
`curl ${shellSingleQuoted(`${baseUrl}/v1/chat/completions`)} \\`,
// and
`client = OpenAI(base_url=${literal(`${baseUrl}/v1`)}, api_key=${literal(apiKey)})`,
```

Or tighten the validator to reject whitespace and quote characters.

web/src/shared/helpers/requestSnippets.ts
```diff
@@ -75,19 +82,49 @@ export function buildCurlSnippet({
...
+ * no snippet and says why.
+ *
+ * A configured value wins in every mode, so a standalone deployment fronted by a
+ * separate API hostname can set one; the trailing slash is trimmed again here
+ * because this is the value the gateway publishes, not the value it validated.
+ */
+export function resolveSnippetBaseUrl(
+  deployment: Pick<DeploymentBootstrap, "deployment_type" | "data_plane_url">,
+  origin: string = typeof window === "undefined" ? "" : window.location.origin,
+): string | null {
```
khaledosman: Return `string | undefined`. frontend-standards (`typescript-and-react.md`) is explicit: "`undefined`, not `null`, for absent values in your own types and props... convert at the boundary rather than letting `null` spread through the component tree." This helper *is* that boundary, and it instead hands `null` onward to `snippets === null` (`KeysPage.tsx:189`) and `snippetInput === null` (`SetupGuideCard.tsx:240`).

web/src/shared/helpers/requestSnippets.ts
```diff
@@ -75,19 +82,49 @@ export function buildCurlSnippet({
...
+ * inference belongs on the data-plane gateway rather than on it, so it has to
+ * name that address itself (`data_plane_url` on `/v1/bootstrap`).
+ *
+ * Null when a hosted deployment names none. Falling back to the origin there is
+ * the bug this replaces: it hands somebody a runnable command aimed at the one
+ * host their traffic should not reach. A placeholder host would be no better,
+ * since nobody reading it can know what to put in its place, so the caller shows
+ * no snippet and says why.
+ *
+ * A configured value wins in every mode, so a standalone deployment fronted by a
```
khaledosman: "A configured value wins in every mode, so a standalone deployment fronted by a separate API hostname can set one" is not what ships. `bootstrap.py:286` publishes the field only when `hosted`, and `docs/configuration.md:66` says standalone and hybrid ignore it. A standalone operator who sets `OTARI_DATA_PLANE_URL` gets silence, not a separate API hostname.

Either drop this sentence, or drop the `if hosted` branch server-side (and warn when the setting is ignored).

web/src/features/keys/KeysPage.tsx
```diff
@@ -175,12 +186,32 @@ function RevealSecretModal({
               Make your first call
             </div>
             <p className="text-xs text-muted">
-              Replace <code>{SNIPPET_MODEL_PLACEHOLDER}</code> with a model from
-              the Models page.
+              {snippets === null ? (
+                <>
+                  This deployment has not published the gateway address to send
```
khaledosman: Byte-identical to `SetupGuideCard.tsx:242-245`. The same argument the module docstring makes for sharing the builders ("so an operator cannot be shown two dialects of one request") applies to the copy that replaces them, and `component-architecture.md` says not to copy markup across files. Extract it next to the builders it stands in for.

docs/configuration.md
```diff
@@ -63,6 +63,7 @@ pricing:
 | `port` | int | `8000` | Server bind port |
 | `master_key` | string | none | Master key for management endpoints |
 | `public_base_url` | string | none | This deployment's own externally-reachable URL, no trailing slash. Needed to put an absolute link in outgoing email (see [Mail](#mail)), to derive the relying-party ID passkeys are bound to (see [Access control](access-control.md#passkeys)), and to derive the OAuth redirect URI (see [OAuth sign-in](access-control.md#oauth-sign-in-google-and-github)). |
+| `data_plane_url` | string | none | Where this deployment's inference traffic belongs, as an absolute `http(s)` URL with no trailing slash. Only a hosted control plane needs it; a standalone or hybrid gateway serves its own API and leaves it unset. See [The data-plane address](#the-data-plane-address). |
```
khaledosman, line 66: `OTARI_DATA_PLANE_URL` is missing from the **Common variables** table at line 202, where `OTARI_DOCS_URL` (the field the validator docstring says this one is modeled on) is listed. Minor, but the env var is the whole configuration surface for a container.

khaledosman (approved):
Verified on the PR head: `make lint`, `tsc --noEmit`, biome, the three affected Vitest files (59 pass), `tests/unit/test_deployment_bootstrap.py` (38 pass), `make postman-check`, and `client:generate` drift are all clean.

Five inline findings, none blocking.

---

web/src/shared/helpers/requestSnippets.ts
khaledosman: Fixed in 2207a728. Both builders now put the base URL through the escaping helpers that were already there for the model id: `curl ${shellSingleQuoted(`${baseUrl}/v1/chat/completions`)}` and `base_url=${literal(`${baseUrl}/v1`)}`. `apiKey` is left as it was; it is gateway-generated and this PR does not change where it comes from.

Both cases are pinned by tests, including that the whole URL stays one shell word so an embedded `;` never reaches the shell as a separator. Output for an ordinary value is unchanged apart from the quotes on the curl line.

Also tightened the validator in 32aae696, for the one bad value escaping cannot help with: a query string or fragment (see the CodeRabbit thread on `config.py`).

🤖 Applied with [Claude Code](https://claude.com/claude-code)

web/src/shared/helpers/requestSnippets.ts
khaledosman: Fixed in 2207a728. `resolveSnippetBaseUrl` returns `string | undefined`, and both call sites branch on `undefined`.

It also no longer answers `""`: a blank origin (no `window`) now returns `undefined`, which is Copilot's point on the same function, so a caller cannot build `curl /v1/chat/completions` out of an origin that is not there. New test covers it.

🤖 Applied with [Claude Code](https://claude.com/claude-code)

web/src/shared/helpers/requestSnippets.ts
khaledosman: Fixed in 2207a728. Kept the `if hosted` branch server-side, since that is what `docs/configuration.md` and `docs/modes.md` both describe, and rewrote the docstring to say what actually ships: the helper reads `data_plane_url` whichever mode published it, but only a hosted deployment does.

🤖 Applied with [Claude Code](https://claude.com/claude-code)

web/src/features/keys/KeysPage.tsx
khaledosman: Fixed in 2207a728. Extracted to `MissingGatewayAddressNotice` (`web/src/shared/components/`), with a colocated test and a row in the `components.md` inventory. Both call sites now render the one component, and the Keys page's conditional moved out of the enclosing `<p>` so the markup is identical on both sides.

🤖 Applied with [Claude Code](https://claude.com/claude-code)

docs/configuration.md
khaledosman: Fixed in 2207a728. Added beside `OTARI_DOCS_URL` in the Common variables table.

🤖 Applied with [Claude Code](https://claude.com/claude-code)

web/src/shared/helpers/requestSnippets.ts
```diff
@@ -75,19 +82,49 @@ export function buildCurlSnippet({
...
+ * separate API hostname can set one; the trailing slash is trimmed again here
+ * because this is the value the gateway publishes, not the value it validated.
```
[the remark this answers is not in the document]
khaledosman: Fixed in 2207a728. Kept the server-side behavior (only a hosted deployment publishes the field, which is what `docs/configuration.md` and `docs/modes.md` describe) and corrected the docstring to match.

🤖 Applied with [Claude Code](https://claude.com/claude-code)

web/src/shared/helpers/requestSnippets.ts
```diff
@@ -75,19 +82,49 @@ export function buildCurlSnippet({
...
+ */
+export function resolveSnippetBaseUrl(
+  deployment: Pick<DeploymentBootstrap, "deployment_type" | "data_plane_url">,
+  origin: string = typeof window === "undefined" ? "" : window.location.origin,
+): string | null {
+  const configured = deployment.data_plane_url?.trim().replace(/\/+$/, "")
+  if (configured) return configured
+  return deployment.deployment_type === "hosted" ? null : origin
+}
```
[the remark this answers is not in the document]
khaledosman: Fixed in 2207a728. The helper now returns `string | undefined` and answers `undefined` rather than `""` when there is no `window`, so the contract holds on every path. New test pins it.

🤖 Applied with [Claude Code](https://claude.com/claude-code)

docs/public/openapi.json
```diff
@@ -3154,6 +3166,7 @@
           "session_type",
           "surfaces",
           "management_url",
+          "data_plane_url",
           "docs_url",
```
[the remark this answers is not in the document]
khaledosman: No change needed here: `make postman-check` passes on this branch, verified locally against the PR head.

The collection is derived from paths, parameters and request bodies (`scripts/generate_postman.py`), not from response schemas. `GET /v1/bootstrap` has no request body, so adding a field to the `DeploymentBootstrap` *response* model leaves the collection byte-identical. The `openapi-spec` CI job, which runs both `make openapi-check` and `make postman-check`, is green.

🤖 Checked with [Claude Code](https://claude.com/claude-code)

src/gateway/core/config.py
```diff
@@ -1723,6 +1736,29 @@ def _validate_docs_url(cls, value: str | None) -> str | None:
...
+        different path on a strict router.
+        """
+        normalized = (value or "").strip().rstrip("/")
+        if not normalized:
+            return None
+        parsed = urlsplit(normalized)
+        if parsed.scheme not in {"http", "https"} or not parsed.netloc:
+            msg = f"data_plane_url must be an absolute http(s) URL, got '{value}'"
+            raise ValueError(msg)
+        return normalized
```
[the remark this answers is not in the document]
khaledosman: Fixed in 32aae696. `_validate_data_plane_url` now rejects a value carrying a query string or a fragment, with the message naming which, and `docs/configuration.md` says so beside the trailing-slash rule. Parametrized coverage added for both cases.

Good catch: this is the one bad value the http(s) check alone let through, and unlike `docs_url` (a link a person follows) nothing downstream can recover from it, since the dashboard appends `/v1/chat/completions` to this.

🤖 Applied with [Claude Code](https://claude.com/claude-code)

PR #827, opened 2026-08-27
fix(streaming): relax final forwarded-tool timeout

### LINKED ISSUE #826: Final forwarded-tool requests time out too early
## What happens

A sole/final hybrid streaming attempt with forwarded provider tools uses the plain first-chunk deadline. In the hosted deployment that is 10 seconds, so a slow tool-bearing Anthropic request returned `504 {"detail":"LLM provider timeout"}` before producing its first chunk.

Check Octonous Sentry Issue: 7695028645

## Expected

Non-final attempts keep the tight failover budget, while a final forwarded-tool attempt gets the existing 30-second tool-loop deadline.

## Repro

Send a single-attempt hybrid streaming request with provider-native tools and delay the upstream first chunk beyond the plain final deadline but below 30 seconds. The gateway currently returns 504 instead of accepting the slow first chunk.

Ref: #237

HareeshBahuleyan:
## Why

Final hybrid requests with forwarded provider tools used the plain first-chunk deadline. In the hosted configuration, that can return a 504 after 10 seconds even though tool-heavy requests already have a 30-second criterion.

## What changed

Keep the tight deadline while another fallback remains, but floor a sole or final forwarded-tool request at the existing tool-loop budget. Ordinary final requests and gateway-managed tool-loop deadlines are unchanged.

## Notes

- Targeted timeout tests: 92 passed.
- `make test-unit` reached 2557 passed and 9 skipped, then failed one docs-style check against an ignored local `docs/superpowers/` artifact that is not part of this branch.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [x] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes #826

## Checklist

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:** Pi with OpenAI Codex `gpt-5.6-sol`

The agent investigated the production symptom, implemented the timeout policy, and ran the checks listed above under human direction.

**NOTE:**

## Summary

Updated hybrid streaming timeout handling for final requests that forward provider tools. Final attempts now allow the existing tool-loop timeout, while earlier fallback attempts keep the shorter timeout. Ordinary requests remain unchanged.

Added configuration, documentation, and tests for the new behavior.

## Technical notes

- Added `STREAMING_FALLBACK_FIRST_CHUNK_TIMEOUT_MS_TOOL_LOOP`.
- Updated final-attempt timeout calculation and fallback handling.
- Added unit and configuration coverage.
- Targeted checks, lint, and type checks passed. Unit tests passed with 2,557 tests and 9 skips.

### DIFF
--- a/src/gateway/api/routes/_pipeline.py
+++ b/src/gateway/api/routes/_pipeline.py
@@ -3713,9 +3713,11 @@ def stream_final_attempt_extra_seconds(
     Added on top of the per-attempt failover budget for the terminal attempt,
     which has no next entry in the routing policy to fall over to. A request that
     forwards provider-native tools but does not run a gateway-managed tool loop
-    keeps the plain failover budget for non-final attempts, while its final
-    deadline is floored to the tool-loop budget. Tool-heavy agents therefore get
-    the existing relaxed criterion only when there is nowhere left to fail over.
+    keeps the plain failover budget on non-final attempts, and on the final one
+    is raised to the tool-loop base before the configured grace is added, so
+    tool-heavy agents get the relaxed criterion only where there is nowhere left
+    to fail over. The grace stays additive in every mode: an operator who
+    configures it always buys that much more time.
     """
     configured_extra = (
         int(
@@ -3731,7 +3733,7 @@ def stream_final_attempt_extra_seconds(
 
     plain_budget = stream_first_chunk_timeout_seconds(config, tool_mode=False)
     tool_loop_budget = stream_first_chunk_timeout_seconds(config, tool_mode=True)
-    return max(configured_extra, tool_loop_budget - plain_budget)
+    return configured_extra + max(0.0, tool_loop_budget - plain_budget)
 
 
 # ---------------------------------------------------------------------------
--- a/src/gateway/api/routes/_pipeline.py
+++ b/src/gateway/api/routes/_pipeline.py
@@ -3702,15 +3702,22 @@ def stream_first_chunk_timeout_seconds(config: GatewayConfig, *, tool_mode: bool
     )
 
 
-def stream_final_attempt_extra_seconds(config: GatewayConfig) -> float:
+def stream_final_attempt_extra_seconds(
+    config: GatewayConfig,
+    *,
+    tool_mode: bool = False,
+    has_forwarded_tools: bool = False,
+) -> float:
     """Extra first-chunk grace granted only to the sole/final streaming attempt.
 
     Added on top of the per-attempt failover budget for the terminal attempt,
-    which has no next entry in the routing policy to fall over to. Keeps that
-    attempt's wait bounded while not converting a slow-but-valid first token into
-    a timeout. Mode-agnostic (applies on top of the plain or tool-loop budget).
+    which has no next entry in the routing policy to fall over to. A request that
+    forwards provider-native tools but does not run a gateway-managed tool loop
+    keeps the plain failover budget for non-final attempts, while its final
+    deadline is floored to the tool-loop budget. Tool-heavy agents therefore get
+    the existing relaxed criterion only when there is nowhere left to fail over.
     """
-    return (
+    configured_extra = (
         int(
             config.platform.get(
                 _STREAM_FINAL_ATTEMPT_EXTRA_FIRST_CHUNK_TIMEOUT_MS_KEY,
@@ -3719,6 +3726,12 @@ def stream_final_attempt_extra_seconds(config: GatewayConfig) -> float:
         )
         / 1000
     )
+    if tool_mode or not has_forwarded_tools:
+        return configured_extra
+
+    plain_budget = stream_first_chunk_timeout_seconds(config, tool_mode=False)
+    tool_loop_budget = stream_first_chunk_timeout_seconds(config, tool_mode=True)
+    return max(configured_extra, tool_loop_budget - plain_budget)
 
 
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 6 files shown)


src/gateway/api/routes/_pipeline.py
```diff
@@ -3668,6 +3675,12 @@ def stream_final_attempt_extra_seconds(config: GatewayConfig) -> float:
         )
         / 1000
     )
+    if tool_mode or not has_forwarded_tools:
+        return configured_extra
+
+    plain_budget = stream_first_chunk_timeout_seconds(config, tool_mode=False)
+    tool_loop_budget = stream_first_chunk_timeout_seconds(config, tool_mode=True)
+    return max(configured_extra, tool_loop_budget - plain_budget)
```
khaledosman: The floor swallows the operator's configured grace instead of adding to it. With this test's own numbers (plain 3s, tool-loop 30s, extra 7s), a forwarded-tool final attempt gets 30s while a managed tool-loop final attempt gets 37s: the configured grace has no effect at all until it exceeds `tool_loop - plain` (27s).

If that is deliberate, `_platform.py:95` is now wrong. It still reads "Applied on top of whichever base budget is in effect (plain or tool-loop)", which is exactly what this branch stops doing.

Either pick the tool-loop budget as the base and keep the extra additive (`plain if not has_forwarded_tools else tool_loop`, then `+ configured_extra`), or leave the floor and update that comment to describe it.

docs/configuration.md
```diff
@@ -383,8 +383,8 @@ These are only relevant when running connected to [otari.ai](https://otari.ai).
...
 | `PLATFORM_USAGE_INLINE_TIMEOUT_MS` | `1500` | Budget for the successful usage report awaited to attach inline cost. Expiry returns the response without cost while the report continues |
 | `PLATFORM_USAGE_MAX_RETRIES` | `3` | Max retries for transient usage reporting failures |
-| `STREAMING_FALLBACK_FIRST_CHUNK_TIMEOUT_MS` | `2000` | Per-attempt timeout waiting for first streamed chunk |
-| `STREAMING_FALLBACK_FINAL_ATTEMPT_EXTRA_FIRST_CHUNK_TIMEOUT_MS` | `0` | Extra first-chunk grace for the sole/final attempt, added on top of the per-attempt budget. `0` = no grace (unchanged) |
+| `STREAMING_FALLBACK_FIRST_CHUNK_TIMEOUT_MS` | `2000` | Per-attempt timeout waiting for the first streamed chunk. Forwarded provider-tool requests keep this tight budget on non-final attempts. |
+| `STREAMING_FALLBACK_FINAL_ATTEMPT_EXTRA_FIRST_CHUNK_TIMEOUT_MS` | `0` | Extra first-chunk grace for the sole/final attempt, added on top of the per-attempt budget. A final request with forwarded provider tools gets at least the tool-loop first-chunk budget (`30000` ms by default), even when this configured grace would produce a shorter deadline. |
```
khaledosman: This row (and its twin in `hybrid-mode-protocol.md`) now makes `streaming_first_chunk_timeout_ms_tool_loop` load-bearing for a new class of request, but that key has no env var. `_apply_platform_env_overrides` (`src/gateway/core/config.py:2119`) maps only `STREAMING_FALLBACK_FIRST_CHUNK_TIMEOUT_MS` and `STREAMING_FALLBACK_FINAL_ATTEMPT_EXTRA_FIRST_CHUNK_TIMEOUT_MS`.

So on the env-configured hosted deployment #826 came from, an operator who finds the new 30s floor too long can only change it by re-supplying the whole `OTARI_CONFIG_YAML`. Add `"STREAMING_FALLBACK_FIRST_CHUNK_TIMEOUT_MS_TOOL_LOOP": ("streaming_first_chunk_timeout_ms_tool_loop", int)` to that mapping and give it a row in both tables.

khaledosman (approved):
Verified locally on the head: `make lint`, `make typecheck`, and `tests/unit/test_pipeline_settlement.py tests/unit/test_platform_timeout_config.py tests/unit/test_streaming_attempts.py` (92 passed). The floor math is correct and never regresses the existing deadline. Two findings inline. The branch currently conflicts with `main`.

<sub>Reviewed with [Claude Code](https://claude.com/claude-code).</sub>

src/gateway/api/routes/_pipeline.py
HareeshBahuleyan: Went with the first option: the tool-loop budget is now the base for a forwarded-tool final attempt and the configured grace is added on top, rather than compared against the gap.

`stream_final_attempt_extra_seconds` now returns `configured_extra + max(0.0, tool_loop_budget - plain_budget)`, so with your numbers (plain 3s, tool-loop 30s, extra 7s) a forwarded-tool final attempt gets 37s, the same as a managed tool-loop one, and the 7s always buys 7s. The `max(0.0, ...)` keeps a config with `tool_loop < plain` from subtracting time; validation only enforces that both are positive.

`_platform.py:95` is updated rather than left stale: it now names which base is in effect when (tool-loop for a managed loop or forwarded tools, plain otherwise), which is again additive as it claimed.

Test expectations moved with it: `forwarded-tools-final` 30.0 -> 37.0, and the wiring test's `final_attempt_extra_seconds` 27.0 -> 34.0.

docs/configuration.md
HareeshBahuleyan: Added, as `"STREAMING_FALLBACK_FIRST_CHUNK_TIMEOUT_MS_TOOL_LOOP": ("streaming_first_chunk_timeout_ms_tool_loop", int)` in `_apply_platform_env_overrides`. An env-configured deployment can now retune the floor without re-supplying `OTARI_CONFIG_YAML`. Env overrides run before `GatewayConfig(**config_dict)`, so the existing `_validate_platform_streaming_timeouts` validator covers the new var too: `STREAMING_FALLBACK_FIRST_CHUNK_TIMEOUT_MS_TOOL_LOOP=0` is refused at startup with `streaming_first_chunk_timeout_ms_tool_loop must be > 0, got 0`. `test_load_config_platform_env_overrides` now asserts all three streaming vars.

On the doc rows: the table this comment is on no longer exists. #860 replaced the `otari.ai variables` table in `docs/configuration.md` with a pointer to the hybrid-mode protocol doc, which is what the rebase onto main conflicted over, so the env-var rows now live only in `docs/hybrid-mode-protocol.md`. The new key has a row there, and the neighboring rows plus the step-2 prose are reworded from "floored to" to the additive behavior.

PR #828, opened 2026-08-27
fix(config): refuse a data_plane_url that already names /v1

njbrake:
## Description

Follow-up to #825, which added `data_plane_url`. Two gaps that review found after it merged.

**A value that already ends in `/v1` was accepted.** Everywhere else a client meets one, "base URL" means the `/v1` address (the OpenAI SDK's own `base_url` includes it), so writing that here is the natural mistake. #825's validator refused a bad scheme, a query string and a fragment, but this passed, and the dashboard then rendered a snippet posting to `/v1/v1/chat/completions`: it looks right and 404s on first use, with nothing in the product to surface it.

Refused rather than stripped. Stripping would be silent, and wrong for a gateway genuinely mounted under such a path, while refusing can never produce a wrong URL and costs one edit. There is also no running deployment for the refusal to take down, since the key itself shipped only in #825, which is what makes a hard failure cheap here where it would have been expensive for the unset case that only warns. Any other path prefix is left alone, because a gateway proxied at `https://api.example.com/otari` is a real deployment.

The rule is stated on the bootstrap field too, since otari.ai serves the same contract from its own code and a validator here cannot reach it.

**`data_plane_url` was missing from `_CONFIG_VIEW`.** #825 followed the `docs_url` precedent, which adds itself to that hand-maintained roster; nothing enforces its completeness, so the omission was silent. Without it an operator cannot read back the setting that explains why a hosted deployment shows no snippet, on the page that exists to answer that.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Follow-up to #825, part of #823.

## Checklist

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

Claude Opus 5, through the Claude Code CLI.

Both gaps were found by a separate review agent with no part in writing #825. The decision to refuse a `/v1` suffix rather than strip it is @njbrake's.

## Summary

- Rejects `data_plane_url` values with a `/v1` path segment or embedded credentials.
- Documents the required bare gateway address and accepted path prefixes.
- Shows `data_plane_url` in the read-only configuration view.
- Adds tests for URL validation, credential safety, configuration coverage, and displayed values.

These changes prevent malformed endpoints such as `/v1/v1/chat/completions` and prevent credentials from appearing in bootstrap or configuration responses.


src/gateway/core/config.py
```diff
@@ -1768,6 +1779,12 @@ def _validate_data_plane_url(cls, value: str | None) -> str | None:
         if parsed.query or parsed.fragment:
             msg = f"data_plane_url must carry no query string or fragment, got '{value}'"
             raise ValueError(msg)
+        if parsed.path.rsplit("/", 1)[-1].lower() == "v1":
```
khaledosman: Only the last path segment is checked, so `https://gateway.otari.ai/v1/chat/completions` still passes and renders a snippet posting to `/v1/chat/completions/v1/chat/completions`. Pasting the full endpoint is the same mistake as pasting `/v1`, and the same reasoning applies: refusal cannot produce a wrong URL and costs one edit.

```python
if any(segment.lower() == "v1" for segment in parsed.path.split("/")):
```

That still leaves `https://api.example.com/otari` alone.

docs/configuration.md
```diff
@@ -63,7 +63,7 @@ pricing:
 | `port` | int | `8000` | Server bind port |
 | `master_key` | string | none | Master key for management endpoints |
 | `public_base_url` | string | none | This deployment's own externally-reachable URL, no trailing slash. Needed to put an absolute link in outgoing email (see [Mail](#mail)), to derive the relying-party ID passkeys are bound to (see [Access control](access-control.md#passkeys)), and to derive the OAuth redirect URI (see [OAuth sign-in](access-control.md#oauth-sign-in-google-and-github)). |
-| `data_plane_url` | string | none | Where this deployment's inference traffic belongs, as an absolute `http(s)` URL with no trailing slash. Only a hosted control plane needs it; a standalone or hybrid gateway serves its own API and leaves it unset. See [The data-plane address](#the-data-plane-address). |
+| `data_plane_url` | string | none | Where this deployment's inference traffic belongs, as an absolute `http(s)` URL with no trailing slash and no `/v1` suffix, which the dashboard appends itself. Only a hosted control plane needs it; a standalone or hybrid gateway serves its own API and leaves it unset. See [The data-plane address](#the-data-plane-address). |
```
khaledosman: The env-var row at `docs/configuration.md:203` still reads "as an absolute `http(s)` URL" with no trailing-slash or `/v1` rule. `modes.md` points a hosted operator at `OTARI_DATA_PLANE_URL`, so that row is the one they read; mirror the new wording there.

src/gateway/api/routes/settings.py
```diff
@@ -152,6 +152,7 @@
             "enable_metrics",
             "enable_docs",
             "docs_url",
+            "data_plane_url",
```
khaledosman, line 155: This adds the missing key but not the guard the PR body diagnoses ("nothing enforces its completeness, so the omission was silent"), so the next field is silently omitted the same way.

35 of 101 `GatewayConfig` fields are absent from `_CONFIG_VIEW`, most deliberately (secrets, nested blocks like `providers`/`pricing`, router tuning knobs), so the enforceable shape is a test asserting `set(GatewayConfig.model_fields) == shown | _DELIBERATELY_OMITTED` with that exclusion tuple named next to the roster. Then the next omission fails a test instead of needing a reviewer.

khaledosman (approved):
Verified at `e41bff71`: `make lint`, `make typecheck`, `tests/unit/test_deployment_bootstrap.py` + `tests/unit/test_settings_endpoint.py` (69 passed), `make openapi-check` and `make postman-check` all pass. The Postman collection correctly needs no change here: it derives request bodies only, and this description sits on a GET response schema.

One finding outside the diff, so it could not be inline:

**`docs/modes.md:46`** — "Set `data_plane_url` (or `OTARI_DATA_PLANE_URL`) to that gateway's base URL" is exactly the phrase this PR identifies as the cause of the mistake, on the page a hosted operator reads while setting the value. Say "that gateway's own address, with no `/v1` suffix".

---

🤖 Review by [Claude Code](https://claude.com/claude-code) (Claude Opus 5), run by @khaledosman.

PR #830, opened 2026-08-27
feat(dashboard): finish the trend-chip migration on the overview page

jigjigjig:
## Description

`TrendChip` landed in #801 with no call sites, and #815 moved the Usage page onto it. Both left
the Overview page behind: its three tiles with a period-over-period change still printed
`▲ 225.5% vs prev` through `DeltaHint`, where the arrow is decoration a screen reader skips and
the text reads identically whichever way the number went. #815's own "Not in scope" section named
this as the rest of the migration, and said `DeltaHint` could retire in it.

This is that follow-up, and it is the last of the three.

### The chips

| Tile | Polarity | Why |
| --- | --- | --- |
| Spend, last 30 days | `down-is-good` | Spend falling is the improvement |
| Requests, last 30 days | `neutral` | Volume. More traffic is neither a win nor a regression on its own, and the error-rate tile beside it carries the judgment |
| Error rate, last 30 days | `down-is-good` | Errors falling is the improvement |
| Spend today, Budget health, Active keys, Active members | n/a | No delta on the wire, untouched |

Same polarity reasoning as #815, so the two pages agree on what a rising number means: `neutral`
on volume is a deliberate choice rather than an omission, since painting rising traffic green
would assert something the gateway does not know.

Each delta is a named const (`costDelta`, `requestDelta`, beside the `errDelta` that already
existed) and each chip is gated on it being non-null. That is the shape the review on #815
arrived at, for a reason worth restating: `TrendChip` renders nothing for a null fraction, but the
*element* is truthy, and `StatCard` tests `trend || hint || chart` to decide whether to reserve
the aside row, so an ungated chip costs a tile 42px of dead space while the previous window is in
flight. Gating on the fraction rather than on the query also covers the two cases a `periodTotals ?`
guard would miss: the current window landed and the previous has not, and a previous value of 0.

The deltas move from `hint` into the `trend` slot #815 added, so the chip leads the aside row and
`DeltaHint` retires with its last caller.

### The status pill

`StatCard`'s status word ("On track", "Elevated") moves onto HeroUI's `Chip` in the same pass.
It was a hand-rolled `<span>` carrying a `border`/`bg`/`text` class triple per status, restating
in three class strings what the library already derives from an aliased token, and it left the
status word and the trend chip on one tile drawing two different shapes at two different weights.
`globals.css` already aliases `--success`, `--warning` and `--danger` onto our tokens, so all
three statuses follow the foundation in both themes with nothing named locally, which is the same
argument #801 made for building `TrendChip` on `Chip` rather than on a `<span>`. Both are now
`soft`/`sm`, and `STAT_STATUS` carries a color name instead of a class triple. The tile's left
accent bar is untouched: that is the card, not the chip.

Not folded in: `Badge` (`ui.tsx`) is a third hand-rolled pill, with a `muted` tone that has no
status color and call sites across the Tools cards and several tables. Moving it is its own
change.

### Tests

Three new tests on `OverviewPage`, written against the findings the review on #815 raised rather
than repeating the mistakes:

- The fixture gives each tile a **distinct** delta (spend +100%, requests +25%, error rate +300%),
  so no assertion can be satisfied by a sibling tile's chip.
- The announcement matcher is **anchored and counted**
  (`/^(no change|up|down)(, (better|worse))?$/` at length 3). Unanchored, `up` and `down` also
  match this page's own prose, and the assertion would pass with the announcement deleted, which
  is exactly how it passed on #815 before review caught it.
- One test pins polarity (`up, worse` twice, a bare `up` once), so a chip that lost its judgment
  fails. `TrendChip.test.tsx` already owns the fraction-to-direction mapping, so these do not
  re-cover it.
- One test pins that a previous window of zeroes reserves no row, which is the gating above.

No test was added for the status pill. The visible behavior is the same word in the same place,
and the only assertion that could tell the two implementations apart is a class selector, which
the review on #801 rejected outright ("a test that breaks on a restyle teaches everyone to stop
trusting the suite"). The existing `"Elevated"` and `"On track"` assertions still cover that the
word renders, in both the unit suite and `parity.overview.spec.ts`.

### Verification

`pnpm --dir web run lint`, `pnpm --dir web run typecheck` and `pnpm --dir web test` all pass:
106 files, 1560 tests. `make lint` passes too, though no Python changed, so it has nothing to say
about this diff.

Worth repeating from #815, because it cost time again: the dashboard suite fails wholesale on
Node 26 and passes on Node 22, which is what `otari-dashboard.yml` and the Docker web stage use.
Run it on 22.

Read against a local gateway on this branch rather than from the fixtures alone. No API contract
changed, so the OpenAPI spec and the Postman collection are untouched. No page was added, so no
screenshot entry is owed. The bundle is not committed.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

None open. #814 scoped its follow-up to the Usage page and was closed by #815; the Overview
remainder was named in #815's description rather than filed as an issue of its own.

## Checklist

Dashboard-only change, so the checks that apply are the dashboard's, listed under Verification
above.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

The agent was pointed at #801 and #815, including their inline review threads, and asked to carry
the approach and the findings forward rather than re-derive them. The anchored-and-counted
matcher, the gate-on-the-fraction shape, and the refusal to assert on a class name all come from
those reviews.

## Summary

- Added gated trend chips to the Spend, Requests, and Error Rate tiles.
- Applied metric-specific trend meaning.
- Replaced custom status pills with shared `Chip` components.
- Removed the unused `DeltaHint` component.
- Added tests for trend values, direction announcements, polarity, and zero-valued previous periods.

These changes improve consistency, accessibility, and clarity on the Overview page.

### DIFF
--- a/web/src/features/overview/OverviewPage.test.tsx
+++ b/web/src/features/overview/OverviewPage.test.tsx
@@ -155,6 +155,70 @@ describe("OverviewPage", () => {
     expect(screen.getByText("Elevated")).toBeInTheDocument() // error-rate status word (non-hue)
   })
 
+  it("renders period-over-period change as a trend chip, not a glyph", async () => {
+    mockApi({
+      today: { cost: 5, request_count: 100, error_count: 1 },
+      period: { cost: 200, request_count: 2000, error_count: 40 },
+      prev: { cost: 100, request_count: 1600, error_count: 8 },
+    })
+    renderPage(<OverviewPage />)
+    await screen.findByText("$200.00")
+
+    // The chip carries the caption the old plain-text hint carried, so the
+    // comparison still says what it is being compared against. Distinct
+    // percentages per tile, so each assertion is about one chip: spend doubled,
+    // requests rose a quarter, and the error rate went 0.5% -> 2.0%.
+    expect(screen.getByText("+100.0% vs prev")).toBeInTheDocument()
+    expect(screen.getByText("+25.0% vs prev")).toBeInTheDocument()
+    expect(screen.getByText("+300.0% vs prev")).toBeInTheDocument()
+    // And each announces a direction, which the glyph never did: "▲" is
+    // decoration a screen reader skips. TrendChip.test.tsx owns the direction
+    // and polarity mapping; this only asserts the three tiles go through it.
+    // Anchored and counted: unanchored, `up` and `down` also match this page's
+    // own prose, so the assertion would pass with the announcement deleted.
+    expect(
+      screen.getAllByText(/^(no change|up|down)(, (better|worse))?$/),
+    ).toHaveLength(3)
+    // The hand-rolled arrow glyphs are gone from the tiles.
+    expect(screen.queryByText(/[▲▼]/)).not.toBeInTheDocument()
+  })
+
+  it("reads a chip against the metric's own polarity, not the direction alone", async () => {
+    // Spend and the error rate both rose, and both improve by falling, so both
+    // are regressions and say so: direction plus judgment, because polarity puts
+    // good and bad in hue alone. Request volume rose too, but volume carries no
+    // polarity, so it announces the direction and nothing more.
+    mockApi({
+      today: { cost: 5, request_count: 100, error_count: 1 },
+      period: { cost: 200, request_count: 2000, error_count: 40 },
+      prev: { cost: 100, request_count: 1600, error_count: 8 },
+    })
+    renderPage(<OverviewPage />)
+
+    expect(await screen.findByText("$200.00")).toBeInTheDocument()
+    expect(screen.getAllByText("up, worse")).toHaveLength(2)
+    expect(screen.getAllByText("up")).toHaveLength(1)
+  })
+
+  it("reserves no trend row for a tile with no comparable previous window", async () => {
+    // No previous window on the wire leaves every delta null, and TrendChip
+    // renders nothing for a null fraction. The chip has to be gated on the
+    // fraction rather than on the query, or StatCard reserves the aside row for
+    // an element that draws nothing.
+    mockApi({
+      today: { cost: 5 },
+      period: { cost: 200, request_count: 2000, error_count: 40 },
+      prev: { cost: 0, request_count: 0, error_count: 0 },
+    })
+    renderPage(<OverviewPage />)
+    await screen.findByText("$200.00")
+
+    expect(screen.queryByText(/vs prev/)).not.toBeInTheDocument()
+    expect(
+      screen.queryByText(/^(no change|up|down)(, (better|worse))?$/),
+    ).not.toBeInTheDocument()
+  })
+
   it("renders spend and request-volume sparklines from the 30-day series", async () => {
     const series = [
       {
--- a/web/src/features/overview/OverviewPage.test.tsx
+++ b/web/src/features/overview/OverviewPage.test.tsx
@@ -213,10 +213,13 @@ describe("OverviewPage", () => {
     renderPage(<OverviewPage />)
     await screen.findByText("$200.00")
 
-    expect(screen.queryByText(/vs prev/)).not.toBeInTheDocument()
+    // queryAllByText, not queryByText: a regression that renders all three
+    // chips makes queryByText throw on the multiple match rather than fail on
+    // the assertion, which reads as a broken test instead of a broken tile.
+    expect(screen.queryAllByText(/vs prev/)).toHaveLength(0)
     expect(
-      screen.queryByText(/^(no change|up|down)(, (better|worse))?$/),
-    ).not.toBeInTheDocument()
+      screen.queryAllByText(/^(no change|up|down)(, (better|worse))?$/),
+    ).toHaveLength(0)
   })
 
   it("renders spend and request-volume sparklines from the 30-day series", async () => {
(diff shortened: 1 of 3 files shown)


web/src/features/overview/OverviewPage.test.tsx
```diff
@@ -155,6 +155,73 @@ describe("OverviewPage", () => {
...
+    renderPage(<OverviewPage />)
+    await screen.findByText("$200.00")
+
+    // queryAllByText, not queryByText: a regression that renders all three
+    // chips makes queryByText throw on the multiple match rather than fail on
+    // the assertion, which reads as a broken test instead of a broken tile.
+    expect(screen.queryAllByText(/vs prev/)).toHaveLength(0)
+    expect(
+      screen.queryAllByText(/^(no change|up|down)(, (better|worse))?$/),
+    ).toHaveLength(0)
```
khaledosman, line 230: This test does not assert the reservation it is named for. Rewriting all three gates to `periodTotals ? (<TrendChip ... />) : null`, the exact regression the comment at `OverviewPage.tsx:248` warns about, leaves all 20 tests in this file green: `TrendChip` renders nothing for a null fraction, so no text appears while the row is still reserved.

`ui.test.tsx:71,78` already sanctions the class assertion for this (jsdom does no layout, so the reservation is only observable as the utility). Adding this passes at head and fails on that mutation:

```ts
const tile = screen.getByText("Spend, last 30 days").parentElement!
expect(tile.querySelector(".min-h-10\\.5")).toBeNull()
```

Scoped to the tile rather than the container, because the Budget health tile reserves the row off its own hint.

web/src/features/overview/OverviewPage.test.tsx
```diff
@@ -155,6 +155,73 @@ describe("OverviewPage", () => {
...
+    // And each announces a direction, which the glyph never did: "▲" is
+    // decoration a screen reader skips. TrendChip.test.tsx owns the direction
+    // and polarity mapping; this only asserts the three tiles go through it.
+    // Anchored and counted: unanchored, `up` and `down` also match this page's
+    // own prose, so the assertion would pass with the announcement deleted.
+    expect(
+      screen.getAllByText(/^(no change|up|down)(, (better|worse))?$/),
+    ).toHaveLength(3)
+    // The hand-rolled arrow glyphs are gone from the tiles.
+    expect(screen.queryByText(/[▲▼]/)).not.toBeInTheDocument()
```
khaledosman: Same flaw the second commit removed at :219: `queryByText` throws on a multiple match, so a regression that restores a glyph on all three tiles reports `TestingLibraryElementError: Found multiple elements with the text: /[▲▼]/` (confirmed by injecting one) rather than failing the assertion.

```ts
expect(screen.queryAllByText(/[▲▼]/)).toHaveLength(0)
```

web/src/features/overview/OverviewPage.test.tsx
```diff
@@ -155,6 +155,73 @@ describe("OverviewPage", () => {
...
   })
 
+  it("renders period-over-period change as a trend chip, not a glyph", async () => {
+    mockApi({
+      today: { cost: 5, request_count: 100, error_count: 1 },
+      period: { cost: 200, request_count: 2000, error_count: 40 },
+      prev: { cost: 100, request_count: 1600, error_count: 8 },
+    })
+    renderPage(<OverviewPage />)
+    await screen.findByText("$200.00")
```
khaledosman: Await the chip, not `$200.00`: every assertion below is about text derived from the *previous* window, while `$200.00` comes from the period query, so the sync `getByText` after it assumes both land in one commit. `testing.md`'s "Await the content, not its container" is the same shape. With a 200ms skew on the prev response this test and the polarity one below both fail.

`await screen.findByText("+100.0% vs prev")` here, and the same swap at :198 (await the chip instead of `$200.00`).

khaledosman (approved):
Three findings, all in the test file. The `ui.tsx` and `OverviewPage.tsx` changes check out: `warning` is a real HeroUI `Chip` color, `--warning-soft` / `--warning-soft-foreground` derive from the `--warning` alias `globals.css` sets in both theme blocks, and the resulting pill lands around 8:1 in both themes, so the darkened light-theme values the standards call out for AA on a pill are not at risk.

_Review by Claude Opus 5 (1M context), via Claude Code._

web/src/features/overview/OverviewPage.test.tsx
jigjigjig: Addressed in f44cfdfc, as written, scoped to the spend tile. Confirmed it earns its name: rewriting all three gates to periodTotals fails this test and only this test, and it passes again on revert. The comment now points at the same jsdom reason ui.test.tsx gives for asserting the utility.

web/src/features/overview/OverviewPage.test.tsx
jigjigjig: Addressed in f44cfdfc: queryAllByText with a length of 0, same as the fix on the other matcher. You are right that it was the identical flaw one test up.

web/src/features/overview/OverviewPage.test.tsx
jigjigjig: Addressed in f44cfdfc: both tests now await "+100.0% vs prev" instead of "$200.00", so nothing sync runs before the previous-window response has landed.

PR #840, opened 2026-08-27
fix(usage): scope /v1/usage/count to imported rows for the select-all affordance

### LINKED ISSUE #781: Usage count and usage mutations resolve different row sets
`GET /v1/usage/count` applies no source guard, while the delete and reprice it feeds do.

`usage_admin_service._selection_conditions` pins every mutation to imported rows: it excludes usage this deployment served itself and requires `counts_toward_budget IS FALSE`. The count endpoint behind the dashboard's "select all N matching" applies only the budget half.

So an operator sees a count, hits delete, and gets a smaller number back with no explanation for the gap. Pre-existing, and widened slightly by mozilla-ai/otari#773, which correctly moved the mutation guard to also exclude backfilled hosted rows without moving the count with it.

## What would close this

The count and the mutations resolve their target set from one place, so a number shown to an operator is the number the next click acts on. `core/usage_source.not_served_here` is the condition the mutation side now uses.

## Definition of Done

- The count and the delete/reprice selection cannot disagree.
- A test pins them to the same row set, including a row this deployment served itself.

Raised by @njbrake in review of mozilla-ai/otari#773. Filed by Claude Opus 5 (1M context) at @khaledosman's request.

AloysJehwin:
## Description

`GET /v1/usage/count` applied no provenance guard, while the delete and reprice it feeds do. `_selection_conditions` pins every mutation to imported rows with both `not_served_here(UsageLog.source)` and `counts_toward_budget IS FALSE`; the count applied only the budget half.

Those are not the same question. Gateway traffic on an `exclude_from_budget` key is also `counts_toward_budget = False`, so an operator could be shown "N matching", hit delete, and get a smaller number back with no explanation for the gap.

The count endpoint serves two callers — the paginator total for `GET /v1/usage`, which must keep matching that endpoint exactly, and the "select all N matching" affordance. Only the second one is scoped to imported rows, and it is the `counts_toward_budget=false` call, so the guard is applied on that branch rather than unconditionally. That keeps the paginator contract intact.

The existing `test_count_scopes_to_imported_rows` passes today because it only ever pairs `counts_toward_budget=False` with imported sources — it never builds the row that actually breaks this. The new test does, and asserts the count and the delete land on the same set.

## PR Type

- [x] Bug Fix

## Relevant issues

Fixes #781

## Checklist

- [ ] I ran the Definition of Done checks locally (`make lint`, `make typecheck`, `make test`). — `ruff check` and `mypy` pass; I could not run `make test` locally because the integration suite needs Docker/testcontainers and Docker is unavailable on this machine. Relying on CI for that.
- [ ] Documentation was updated where necessary. — no doc change needed; the endpoint docstring already described this scoping, the code just did not implement it.
- [ ] If the API contract changed, I regenerated the OpenAPI spec. — no contract change, same params and response shape.

## AI Usage

**Any additional AI details you like to share:** Root-cause tracing and the patch were developed with AI assistance; I reviewed every changed line and confirmed the behaviour against `_selection_conditions` and `core/usage_source`. Note the one checklist box I left unchecked above rather than claim a local test run I could not do.

## Summary

- Updated usage counts to match rows targeted by bulk delete and reprice actions.
- Added regression coverage for gateway-served usage rows.
- Kept usage listings inclusive of served-here rows.
- Clarified the behavior in the generated API and Postman documentation.
- This prevents misleading counts and keeps related actions consistent.

### DIFF
--- a/src/gateway/api/routes/usage.py
+++ b/src/gateway/api/routes/usage.py
@@ -330,8 +330,9 @@ _TOOL_DESC = (
     f"name ({', '.join(GATEWAY_TOOL_NAMES)}) matches that tool specifically."
 )
 _COUNTS_DESC = (
-    "Filter by budget participation: true = only enforced gateway rows, "
-    "false = only imported rows that never touch a budget"
+    "Filter by budget participation, which is not the same question as provenance: "
+    "true = only enforced gateway rows, false = every row that never touches a budget, "
+    "meaning imported usage and also gateway traffic on a budget-exempt key"
 )
 _WORKSPACE_DESC = "Only usage recorded in this workspace."
 # The three entity filters are repeatable on every usage endpoint, so a chart or a
--- a/src/gateway/api/routes/usage.py
+++ b/src/gateway/api/routes/usage.py
@@ -27,6 +27,7 @@ from gateway.core.sql import (
     match_any,
     utc_bound,
 )
+from gateway.core.usage_source import not_served_here
 from gateway.inflight import get_registry
 from gateway.models.entities import APIKey, UsageLog, User
 from gateway.models.money import as_float
@@ -611,6 +612,13 @@ async def count_usage(
         workspace_id=workspace_id,
         scope=None,
     )
+    if counts_toward_budget is False:
+        # This is the "select all N matching" affordance, and the delete / set-price it
+        # feeds scope to imported rows via _selection_conditions. counts_toward_budget
+        # alone does not say "imported": gateway traffic on an exclude_from_budget key
+        # is also False. Without the same provenance guard the count promises rows the
+        # mutation then refuses to touch.
+        conditions.append(not_served_here(UsageLog.source))
     stmt: Any = select(func.count()).select_from(UsageLog).where(*conditions)
     total = (await db.execute(stmt)).scalar_one()
     return UsageCount(total=total)
--- a/src/gateway/api/routes/usage.py
+++ b/src/gateway/api/routes/usage.py
@@ -383,8 +383,10 @@ def _usage_filters(
 ) -> list[ColumnElement[bool]]:
     """Build the shared WHERE conditions for the list and count endpoints.
 
-    Keeping this in one place guarantees the paginator's total (``/count``)
-    always matches the rows ``list_usage`` returns for the same filters.
+    Keeping this in one place is what makes the paginator's total (``/count``)
+    match the rows ``list_usage`` returns. One condition sits outside it:
+    ``count_usage`` also narrows ``counts_toward_budget=false`` to imported rows,
+    because that call sizes a mutation rather than a page (see its docstring).
 
     ``scope`` is the one condition that is not a filter. The deployment-wide
     routes here pass ``None`` and read every row; the organization-scoped routes
@@ -592,6 +594,11 @@ async def count_usage(
     separate request), so the ``COUNT(*)`` is not paid on every page load. With
     ``counts_toward_budget=false`` it also backs the "select all N matching this
     filter" affordance for bulk delete / set-price, which touch imported rows only.
+
+    That value is the one place this count is narrower than ``GET /v1/usage``: it
+    also excludes rows this deployment served itself, so the number an operator
+    confirms is the number the mutation can reach. The list still pages the
+    budget-exempt gateway rows it omits.
     """
     conditions = _usage_filters(
         start_date=start_date,
@@ -613,11 +620,9 @@ async def count_usage(
         scope=None,
     )
     if counts_toward_budget is False:
-        # This is the "select all N matching" affordance, and the delete / set-price it
-        # feeds scope to imported rows via _selection_conditions. counts_toward_budget
-        # alone does not say "imported": gateway traffic on an exclude_from_budget key
-        # is also False. Without the same provenance guard the count promises rows the
-        # mutation then refuses to touch.
+        # counts_toward_budget alone does not say "imported": gateway traffic on an
+        # exclude_from_budget key is also False, so without this the count would
+        # promise rows _selection_conditions then refuses to touch.
         conditions.append(not_served_here(UsageLog.source))
     stmt: Any = select(func.count()).select_from(UsageLog).where(*conditions)
     total = (await db.execute(stmt)).scalar_one()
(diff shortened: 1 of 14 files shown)


src/gateway/api/routes/usage.py
```diff
@@ -581,6 +582,13 @@ async def count_usage(
         request_group_id=request_group_id,
         workspace_id=workspace_id,
     )
+    if counts_toward_budget is False:
```
khaledosman, line 637: The behavior contract changed and the docs still state the old one. `docs/api-reference.md:418` says this endpoint takes the "Same filters as `GET /v1/usage`, so a multi-value filter counts the same rows the list returns." That is now false for `counts_toward_budget=false`: `/v1/usage?counts_toward_budget=false` still pages budget-exempt gateway rows this count no longer reports.

The scoping itself is right, and `/v1/agent-telemetry/count` is the in-house precedent for a count mirroring what its mutation can reach, but the divergence has to be written down. Update that table row, and add the caveat to the docstring above, which currently opens "Total number of usage logs matching the given filters."

A docstring edit makes both generated artifacts stale: run `uv run python scripts/generate_openapi.py`, then `make postman`, and commit both, or the single `openapi-spec` job fails. The checklist's "no contract change, same params and response shape" holds for the schema but not for the behavior.

src/gateway/api/routes/usage.py
```diff
@@ -581,6 +582,13 @@ async def count_usage(
...
+        # This is the "select all N matching" affordance, and the delete / set-price it
+        # feeds scope to imported rows via _selection_conditions. counts_toward_budget
+        # alone does not say "imported": gateway traffic on an exclude_from_budget key
+        # is also False. Without the same provenance guard the count promises rows the
+        # mutation then refuses to touch.
+        conditions.append(not_served_here(UsageLog.source))
```
khaledosman, line 641: `src/gateway/AGENTS.md`, under "Usage filters: read and mutate must agree", lists three rules the count/mutation pair has to hold. This adds a fourth: the count must apply the mutation's fixed scope, not just its filter set. That is exactly the case `test_count_scopes_to_imported_rows` missed, so it is worth a bullet there for the next reader.

tests/integration/test_usage_admin.py
```diff
@@ -758,3 +759,28 @@ def test_count_scopes_to_imported_rows(
...
+    client: TestClient, master_key_header: dict[str, str], db_session: Session
+) -> None:
+    """counts_toward_budget=False is not the same question as "imported".
+
+    Gateway traffic on an exclude_from_budget key is also False, so the count behind
+    "select all N matching" has to apply the provenance guard the mutation applies,
+    or the dialog promises a row the delete then refuses to touch.
+    """
+    _make_log(db_session, log_id="imp-1", counts_toward_budget=False, source="claude_code")
+    _make_log(db_session, log_id="gw-exempt", counts_toward_budget=False, source=SERVED_HERE_SLUG)
```
khaledosman, line 780: This fixture is not shaped like the row it names. `_make_log` derives `endpoint` and `source_event_id` from `counts_toward_budget`, not from `source`, so this row lands with `endpoint="external"` and a non-null `source_event_id`, neither of which a real gateway row on an `exclude_from_budget` key has.

The assertion holds either way, but a future source-scoped or endpoint-scoped guard would be silently untested. Derive those two fields from `source` instead.

khaledosman (approved):
Verified on the PR head: `ruff check` and `scripts/check_architecture.py` clean, `tests/integration/test_usage_admin.py` 29 passed, and the new test fails against `main`'s `usage.py` and passes with the fix, so it is a real regression test.

Two things checked that are not findings: `UsageLog.source` is non-nullable (`default="gateway"`), so `not_in` has no NULL hole; and the `priced` predicate mismatch between `_usage_filters` (`_needs_pricing_expr`) and `_selection_conditions` (`cost IS NULL`) is unreachable for imported rows, since only gateway-run tools write the tool meter.

One on the PR description rather than the code: it carries an em dash ("serves two callers — the paginator total") and "behaviour". The repo bans em dashes as separators in PR descriptions and spells prose in US English. The title is a valid conventional commit.

---

src/gateway/api/routes/usage.py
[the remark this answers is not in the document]
njbrake: Not taking the suggested fix. Guarding `list_usage` would drop budget-exempt gateway traffic off the activity table, which is the traffic the mutation guard exists to protect. `test_count_and_delete_agree_on_budget_exempt_gateway_rows` now fails if the guard moves into `_usage_filters`, asserting the list still pages that row.

The asymmetry is real and is now written down. `/v1/usage/count` publishes its own `counts_toward_budget` description rather than the shared one, and `src/gateway/AGENTS.md` records the exception to shared filter semantics.

src/gateway/api/routes/usage.py
```diff
@@ -581,6 +582,13 @@ async def count_usage(
         request_group_id=request_group_id,
         workspace_id=workspace_id,
     )
+    if counts_toward_budget is False:
+        # This is the "select all N matching" affordance, and the delete / set-price it
+        # feeds scope to imported rows via _selection_conditions. counts_toward_budget
+        # alone does not say "imported": gateway traffic on an exclude_from_budget key
+        # is also False. Without the same provenance guard the count promises rows the
```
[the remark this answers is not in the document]
njbrake: Option (b) is ruled out by test: moving the guard into `_usage_filters` fails the list assertion in `test_count_and_delete_agree_on_budget_exempt_gateway_rows`, because it takes budget-exempt gateway rows off the operator's activity table.

Option (a) stays the better shape if a second paginating caller of `counts_toward_budget=false` appears. Until then the divergence is confined to one endpoint, published in that endpoint's own parameter description, and recorded in `src/gateway/AGENTS.md`.

src/gateway/api/routes/usage.py
njbrake: That line is gone. #860 cut `docs/api-reference.md` to 96 lines and removed the endpoint tables, so there is no filter row to update. The file now names the generated OpenAPI spec as the source of truth, and the spec carries the divergence both in the count endpoint's description and in its own `counts_toward_budget` parameter.

src/gateway/api/routes/usage.py
njbrake: Added, under "Usage filters": the count narrows `counts_toward_budget=false` because it sizes a mutation, and a count that sizes a mutation applies the mutation's fixed scope and not only its filter set. `UsageEntry.bulk_editable` is named there as how a client learns which rows that scope admits.

Issue #839, opened 2026-08-27
fix(dashboard): organization provider keys reports "OTARI_SECRET_KEY is not set" to every non-operator


njbrake:
An organization owner on a hosted control plane sees a warning banner saying `OTARI_SECRET_KEY` is not set, with "Add provider key" disabled, on a deployment where the key is set and the write path works. The banner is reporting a refused request, not the key.

`web/src/features/organization/OrganizationProviderKeysPage.tsx:240` derives the flag from `useSettings()` and fails closed:

```ts
const secretKeyConfigured = settings.data
  ? settings.data.secret_key_configured !== false
  : !settings.isError   // an error reads as "not set"
```

Since #821, `GET /v1/settings` is gated on `require_deployment_operator` (`src/gateway/api/routes/settings.py:412`), which admits a header master key, a superuser, or the bootstrap identity (`src/gateway/services/tenancy/deployment_user_service.py:90`). An organization owner is none of those, so the query 403s for every tenant, and tenants are the page's whole audience in the platform edition (#816). The page landed in #820, one commit before the gate.

Nothing is wrong server-side: `org_provider_keys` is tenant-scoped, keeps `verify_master_key` plus its own role check, and encrypts normally. The UI disables the one control that would have worked, and tells the owner to go set an environment variable that is already set.

Fix: report the fact on a tenant-readable contract instead of inferring it from an operator-only endpoint. `OrganizationMembershipContextPublic` already carries `byo_provider_keys_allowed` (`src/gateway/models/tenancy.py:501`) for this reason, and the page already reads that context for `canEdit`, so a sibling field for "provider-key encryption is available" fits there. Then a 403 cannot masquerade as a missing key.

`web/src/features/providers/ProvidersPage.tsx:845` carries the same inference. That page is operator-only in the platform edition (#816) but not in standalone, so it should move to the same signal or stop inferring.

Acceptance: an organization owner on a deployment with the key set sees no banner and an enabled "Add provider key"; with the key genuinely unset, both pages still warn and disable.

Related: #834 is the same refused-settings-query pattern behind `PricingWarning`, and #832 is another user-visible consequence of the same gate.

_Note: this issue was drafted by Claude Opus 5 via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

Issue #836, opened 2026-08-27
fix(dashboard): a non-operator watches the whole sidebar render, then vanish


njbrake:
Hide `operatorOnly` rows until the operator answer is known, or better, ship that answer with a read the shell already makes.

Today `allowedByCaller` returns `operator.data !== false` (`web/src/app/nav/useNavVisibility.ts:65-71`), so every `operatorOnly: "refused"` row paints while `GET /v1/admin/access` is in flight and is stripped when it answers false. Nine rows are tagged that way: six in the workspace rail (Activity, Usage, Models, Routing, API keys, Providers) and three in the organization one (Spend & budgets, Model pricing, Settings). So on every page load, every signed-in user who is not a deployment operator watches most of the rail appear and then disappear.

`web/src/app/nav/types.ts:66-74` justifies showing first as sparing operators "a sidebar flicker on every operator's page load". That weighs a flicker for operators against a retraction for everyone else, and `AppShell.tsx` already states the principle that settles it the other way: "a row that appears late is not a row that told anyone it was missing".

Two fixes:

1. Require an explicit `true` for `"refused"` too. One line, but it collapses the field's two values into one: `"unlisted"` and `"refused"` would then differ in nothing, including which way they fail.
2. Carry `granted` on `GET /v1/organizations/me`, which `AppShellChrome` already awaits for `managesOrganization`. No in-flight window at all, and the two values keep distinct meanings when that read fails.

Option 2 is the right one, with 1 as a stopgap if 2 cannot land immediately.

Not the deployment bootstrap: it is pre-session and caller-independent by contract (`bootstrap.py:137-144`), and the shape is shared with otari.ai.

Fallout from eb900b8 (#821), like #832 and #834.

_Note: this issue was drafted by Claude Opus 5 via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

PR #842, opened 2026-08-27
feat(api): give a tenant an organization-scoped read of their own usage

### LINKED ISSUE #837: A member of an organization can see none of its usage: /v1/usage is deployment-operator only
Decide whether a non-operator gets an organization-scoped usage read, and if so, build it. Right now there is nothing on that surface a member is allowed to call.

#821 put `require_deployment_operator` on every read behind Activity and Usage (`usage.py:426,532,1336,1475,1615`) and tagged both rows `operatorOnly: "refused"` (`web/src/app/nav/registry.ts:74-85`). `has_administration_access` is `is_superuser` or the bootstrap identity (`deployment_user_service.py:77-93`), so an ordinary member of an organization loses both pages, plus Models, Routing, API keys and Providers. Unhiding the rows achieves nothing: the requests behind them 403.

That is #821 working as designed, and the gap it left is real. On a multi-tenant deployment, "show me my organization's traffic" is the ordinary case, and the operator gate is the only answer the API currently has.

Worth fixing alongside: a member who bookmarked `/usage` still gets the page rendered, with its requests failing inside it. `useRouteVisibility` drops the caller axis deliberately, and the reasoning it gives (an operator should not see a "not served here" panel while the query resolves) is sound, but the member's outcome is a page of failed requests rather than a page that explains itself.

### LINKED ISSUE #838: Promoting a member to organization admin grants nothing the sidebar gates on
Two different authorities read as one in the UI, and the one an admin can reach does not affect what the sidebar shows.

The roster's role picker (`web/src/features/organization/OrganizationMembersPage.tsx`) sets an organization role. Every gate #821 added asks `DeploymentUserService.has_administration_access` (`deployment_user_service.py:77-93`), which is `is_superuser` or being the bootstrap identity, and nothing else. So promoting somebody to `admin` or `owner` leaves the sidebar exactly as it was, with no indication of why.

The only control that changes it is Platform Admin's "Make operator" (`web/src/features/admin/DeploymentAccountsPage.tsx:163`), which sets `is_superuser` and confers authority over every tenant in the process. That is not what "make this person an org admin" is meant to mean, so it is not the workaround.

Needs a decision on whether an organization admin should confer any deployment-wide read (probably no, with an organization-scoped read as the answer instead), and then the UI has to stop implying the role picker affects it.

Separately: even a real "Make operator" is slow to land. `useDeploymentAdminAccess` caches with `staleTime: 60_000` (`web/src/shared/api/hooks.ts:2888-2903`) and the invalidation runs in the promoting operator's client, so the promoted user needs a reload before their sidebar changes.

njbrake:
## Description

An ordinary member of an organization could see none of its usage, and promoting somebody to organization admin changed nothing they could reach. Both are the same decision: what an organization role confers, and which reads exist so a tenant can see their own organization without deployment-wide authority.

**The gate #821 added stays exactly where it is.** `/v1/usage` reads every tenant's rows and its `workspace_id` is a filter the *client* supplies, so a mode that widened it for a non-operator would have rebuilt the escalation that gate closed. Instead the tenant gets a second, narrower reading of the same rows: `/v1/organizations/me/usage{,/count,/summary,/series}`, on the prefix four tenant surfaces already use.

Scope is derived, never accepted. It comes from the caller's own `active_organization_id` through `resolve_visible_workspace_scope`, which refuses a pointer with no live membership behind it; no request names an organization, and `POST /v1/organizations/me/switch` (404 outside the caller's memberships) stays the only way to move. How much of the organization follows the rule `resolve_visible_workspace` already states rather than a fifth answer: owner, admin or superuser read every workspace, member and viewer read the ones they belong to. A named `workspace_id` goes through `resolve_workspace_in_organization`, so outside the scope it is a 404 rather than a row. Belonging to no workspace is an empty page, not a refusal.

Every aggregation is the one `usage.py` already ran. `_summary_response` and `_grouped_series_response` are lifted out of the two handlers so the scoped copies cannot drift; the deployment-wide signatures are untouched, and the regenerated spec is additions only. Usage writes stay operator-only, as does `/v1/usage/in-flight`, whose registry entries carry no workspace to scope by.

**The two authorities now read as two.** The roster's role picker sets an organization role and never reached the sidebar, with nothing saying so; the page now states what a role confers and that no role there reaches the deployment's own pages. And `OrganizationMembershipContextPublic` grows the two fields three issues were each about to design separately: `deployment_operator`, resolved through the same predicate `/v1/admin/access` publishes so the two cannot disagree (#836), and `provider_key_encryption_available`, so a tenant page stops inferring a missing `OTARI_SECRET_KEY` from an operator-gated 403 (#839). **#836 and #839 are unblocked by this contract and need no new endpoint.**

Activity and Usage therefore stop declaring `operatorOnly: "refused"`. The pages behind them now serve every signed-in identity something true, which beats a prettier refusal.

### Security testing

`tests/integration/test_organization_usage_scope.py` builds two organizations with traffic in both and asserts the rows that must be *absent* rather than counting the ones returned. It carries the sibling-workspace case an organization-only scope would pass, the membership-less `active_organization_id` pointer, the cross-tenant `workspace_id` on all four routes, and the control that an organization owner is still refused by `/v1/usage`.

### Deliberately deferred

`/v1/usage/summary.csv` (no dashboard consumer), in-flight workspace attribution, org-scoped equivalents for Models/Routing/Keys/Providers/Budgets, and #838's propagation tail: `refetchOnWindowFocus` is globally off, so a promoted operator still needs a reload and nothing short of polling or a push changes that.

## PR Type

- [x] New Feature
- [x] Bug Fix

## Relevant issues

Fixes #837
Fixes #838

## Checklist

Full backend suite (2594 unit, 1971 integration), the whole dashboard suite (1566), `make lint`, `make typecheck`, `make openapi-check`, `make postman-check` and the OSS-edition smoke all pass locally. Integration tests ran against a local PostgreSQL 18 rather than Testcontainers, since this sandbox has no Docker.

## AI Usage

**AI Model/Tool used:**

The design was settled with @njbrake before any code was written: both issues were labelled `needs-decision`, so the session read #821, #836, #837, #838 and #839, verified every file and line reference against the tree, and brought a recommendation covering the authority model, the API shape, the shared organization-context contract, the UI correction and the scope split. He confirmed the model, including the call that a plain member sees the workspaces they belong to rather than the whole organization. Implementation followed from there.

## Summary

- Added organization-scoped usage endpoints for logs, counts, summaries, and series.
- Limited results by organization membership and workspace access.
- Preserved operator-only access for deployment-wide usage reads and writes.
- Updated dashboard pages to use the correct usage scope.
- Hid deployment-wide controls and data from non-operators.
- Exposed deployment-operator and encryption-availability status in organization context.
- Added tests, API documentation, OpenAPI definitions, and Postman examples.
- Updated generated Postman requests to use valid defaults for required enum parameters.


src/gateway/api/routes/usage.py
```diff
@@ -362,18 +362,29 @@ def _usage_filters(
     status_code: int | None = None,
     request_group_id: list[str] | None = None,
     workspace_id: uuid.UUID | None = None,
+    scope: ColumnElement[bool] | None = None,
```
khaledosman: `scope` defaults to `None`, but the docstring above says the parameter exists "so that a route that builds filters cannot forget the half that makes them safe." The default is what lets it be forgotten, and the forgotten case reads every tenant's rows silently.

Both this and `_summary_context` (line 1272) already take `scope` keyword-only, so dropping the default costs one explicit `scope=None` at the four deployment-wide call sites (`list_usage`, `count_usage`, `usage_summary`, `usage_series`) and turns a future tenant-scoped route that omits it into a `TypeError` rather than a cross-tenant read. That makes the signature enforce what the docstring asserts.

src/gateway/api/routes/organization_usage.py
```diff
@@ -0,0 +1,392 @@
...
+    workspace_id: uuid.UUID | None,
+) -> ColumnElement[bool]:
+    """The WHERE clause that confines a read to what this caller may see.
+
+    Resolved per request rather than cached on the session: a membership can be
+    suspended or a role changed between two requests, and the cheaper answer is
+    the one that goes stale in the unsafe direction.
+    """
+    organizations = OrganizationService(db)
+    scope = await resolve_visible_workspace_scope(db, user=user, organizations=organizations)
```
khaledosman: This runs before the `workspace_id` branch below, so a member filtering by workspace pays the deliberately-unpaged `get_workspace_ids_for_user` scan and then discards the result. The Activity page issues list + count + summary + series per filter change, so that is four wasted unbounded reads per interaction.

The `workspace_id` branch only needs `scope.organization`, which `get_active_organization_for_user` alone supplies. Either resolve just the organization before the branch and the full scope after it, or pass the workspace hint into `resolve_visible_workspace_scope` so it can skip building the id set.

web/src/shared/api/hooks.ts
```diff
@@ -1437,21 +1446,66 @@ function usageParams(filters: UsageFilters): URLSearchParams {
...
+// reaches this, since `_resolve_active_organization` 404s with no live
+// membership to fall back to. So a failure falls through to the *narrower*
+// surface, which is the safe direction (understating beats a cross-tenant read),
+// and its own refusal is what the page reports. `AppShell` resolves the same
+// error the same way, failing open rather than stranding the destinations behind
+// it.
+//
+// The base is part of every query key it feeds, so two callers on one browser
+// can never read each other's cached rows.
+export function useUsageScope(): {
```
khaledosman, line 1478: `deployment_operator`'s own contract note in `models/tenancy.py` says "a client should read one of them rather than both." After this PR the dashboard reads both: `useNavVisibility.ts:56` and `OverviewPage.tsx:118` gate on `useDeploymentAdminAccess()` (`GET /v1/admin/access`), while `useUsageScope` and `isDeploymentOperator` gate on the context field. Two independently-resolving queries for one predicate, on the same pages, so the rail and the page behind it can settle out of step.

Not a blocker and not a correctness bug (the server derives both from the same predicate). Worth converging in a follow-up, or stating in `web/AGENTS.md` which source is canonical for which layer, so that contract note stays true.

khaledosman (approved):
Verified locally against this head: `make openapi-check`, `make postman-check`, and `make lint` (architecture + alembic heads + Ruff) all pass, so the generated artifacts are in sync. Traced the isolation claim rather than taking it: `_scope_condition` lands in `conditions`, and every aggregation path (`_breakdown`, `_errors_by_status_code`, `_tool_breakdown`, the series query, `_totals`) threads that same list, so no pass escapes the scope; `UsageLog.workspace_id` is `nullable=False`, so the `IN` predicate has no NULL blind spot; and `resolve_workspace_in_organization` does 404 a non-member's own-organization workspace, as the module docstring claims.

Three non-blocking notes inline.

Worth saying: the integration suite asserts the rows that must be absent rather than counting the ones returned, and carries the sibling-workspace, suspended-membership, membership-less-pointer, and `/v1/usage`-still-403 controls. That is what makes the isolation claim checkable rather than asserted.

---

PR #853, opened 2026-08-28
feat(dashboard): adopt a better palette across both themes

jigjigjig:
## Description

Replaces the dashboard's teal and silver colour foundation with a neutral palette and indigo accent, in both themes.

The palette is four neutral levels mapped onto our six rungs, with the two families **staggered one rung apart** rather than collapsed onto each other. Two token pairs therefore hold the same value in each theme, which is the stagger working rather than a duplicate to tidy away:

```
light   #ffffff  background-muted (chrome) · surface (cards)
        #f7f8f8  background (canvas)
        #f0f1f2  background-subtle · surface-muted
        #e8e9eb  surface-subtle
dark    #08090a  background (canvas)
        #0f1011  background-muted (chrome) · surface (cards)
        #191a1b  background-subtle · surface-muted
        #1f2023  surface-subtle
```

A consequence worth expecting rather than reporting as a bug: **in light, the nav rail is pure white and the page canvas is darker than it.** That is the deliberate mirror of dark, where the rail sits one rung above the canvas, and it is what keeps chrome reading distinctly from cards in both themes.

Everything was measured on the running dashboard rather than in a specimen, across eleven routes in both themes.

### What we changed

- **`--color-primary-foreground` is `#ffffff`, not `#f7f8f8`.** The off-white `#f7f8f8` on the indigo measures 4.417:1 and misses the AA floor by 0.083 on twelve primary buttons. White measures 4.700 and clears it. A correctness repair, marked as a departure in the file.
- **Status, attention and info are ours.** There are no `-subtle` fills for these status colours and no attention or info family at all, so those inks would have landed on our old fills and produced pairings belonging to neither system. Five hues, each held near the sRGB ceiling at the lightness AA forces on it; worst pairing anywhere in either theme is 4.48:1.
- **Field borders are re-derived for WCAG 1.4.11.** The previous input spec measures 1.13:1 against its own fill where 1.4.11 asks 3.0. These clear 3.0 on the fill side and on every rung a field can land on.
- **Dark elevation is a ring, not a shadow stack.** The canvas is L\* 2.43 and the panel 4.63, so a black shadow paints nothing on either. This is an elevation-model change, not a tuning tweak.

### The vendor aliases and the table row border

This is its own item rather than part of the palette, and it is palette-independent.

Four HeroUI variables were never aliased to our tokens: `--field-foreground` and the three `--separator` tiers. HeroUI's values for them are literals, so a Select's value text, a field's shadow, and every table and accordion separator kept whatever `@heroui/styles` ships and stopped tracking the palette. Separately, HeroUI draws a bottom border on a table **row** as well as on its cells, and the existing `.otari-table` patch only covered the cells.

Measured before this change, overriding every `--color-*` token moved **98.2% of the painted properties** on eleven routes in both themes. These five are the remainder. They would be worth doing under any palette; they are noticeable under this one because the border tokens carry alpha.

The tooltip arrow gets a scoped override for the same reason: `@heroui/styles` composites its outline at 40% of `--border`, which a translucent token thins to nearly nothing.

### Nav ramp

Four distinct fills in both themes, from four levels:

| state | token | dark, L\* from rail |
| --- | --- | --- |
| resting | rail (`background-muted`) | 0 |
| pressed | `background` | -2.20 |
| selected | `surface-muted` | +4.56 |
| hover / focus | `surface-subtle` | +7.63 |

Hover lifts away from the chrome and a press sinks onto the canvas behind it, so the two are told apart by direction rather than degree. `--color-background` had never been assigned a nav state; using it is what takes this from three distinct fills to four.

Both `dark:` overrides on the selected row are gone. `dark:bg-surface` now resolves to the rail's own value and erased the selection entirely. The selected row also no longer answers a hover: a hovered resting row already gets `hover:text-foreground` as well as a fill, so a selected row taking the hover fill became indistinguishable from it, and losing "is the row under my pointer the page I am on" costs more than losing a response on a row that does nothing when clicked.

### Shadows

Thirteen sites used bare Tailwind `shadow-*` utilities, which Tailwind inlines with its own values at build time, so they never read our tokens. They now use `shadow-elevation-*`, which do. `KeysPage`'s `shadow-xl` maps to **`shadow-modal`** rather than an invented `xl` step: `shadow-xl` has no token counterpart and the element is a modal.

This matters more than it did: with dark elevation as a ring, a missed site has no elevation at all rather than a slightly wrong one. Verified two ways, because three separate counts of this family disagreed before it was settled. The source now contains no bare shadow utility, and the built stylesheet emits none, which is independent of any regex since Tailwind only generates a utility the source references.

### Accent as text

Thirteen `text-accent` call sites moved to `text-link`. `#5e6ad2` fails AA as text on every surface in both themes (3.47 to 4.70); it is a fill with white on it. Nine of the thirteen were links already wearing the accent. The other four are genuinely accent-coloured text and take `text-link` too, which passes on every surface. Naming them: a notice in `ShareDialog`, a status label in `KeysPage`, a selection count in `BulkActionBar`, and a hover state in `ModelsPage`. A dedicated ink role would be more honest than a link token for those four, and is not in this change.

### One thing worth not deriving yourself

Shipped separates the token figure from its bar on 131.7° of hue at ΔL\* 0.72; 8D separates on ΔL\* 4.45. Lightness separation survives when hue information is removed: in greyscale the shipped pair collapses to 0.72 L\*, effectively merging, while this one holds at 4.45. Under simulated colour-vision deficiency both pairs move by less than 0.4 ΔE, so CVD is not the differentiator; greyscale is.

## Interactive-state coverage

Every token below is declared in **both** theme blocks: nothing exists in one theme and not the other, which is what `foundation.test.ts` enforces and which now holds by construction. The second column is different information: how many rules in the built stylesheet actually read it.

| token | both blocks | `var()` reads in the built CSS |
| --- | --- | --- |
| `--color-primary` | yes | 4 |
| `--color-primary-hover` | yes | **0** |
| `--color-primary-active` | yes | **0** |
| `--color-primary-subtle` | yes | 7 |
| `--color-primary-foreground` | yes | 2 |
| `--color-primary-subtle-foreground` | yes | 2 |
| `--color-link` / `-hover` | yes | 10 / 3 |
| `--color-focus` | yes | 3 |
| status and attention families | yes | 1 to 3 each |
| `--field-foreground` | yes | 51 |
| `--field-border` / `-background` | yes | 26 / 20 |
| `--separator` / `-tertiary` / `-secondary` | yes | 12 / 9 / 1 |

**Declared and reaching the screen are different things, and this table is here to keep them apart.** Two tokens in the whole set are declared, symmetric and read by nothing, and both are on the accent. See the entry below.

`--color-focus` having no hover or active state, and the five status roles having none, are correct rather than gaps: a focus ring is one value by nature and a status colour is not interactive. They were considered.

The four `--field-*` and `--separator` rows are the vendor aliases described above. Their read counts are why the remainder of that 98.2% mattered.

## Known limitations, named

**A button's pressed colour cannot be expressed through the tokens, for three independent reasons, any one sufficient.** `.button:active` and `.button:hover` sit at equal specificity in `@heroui/styles` with hover later in the sheet (offsets 73414 and 73540), so a mouse press, which is always also a hover, always takes the hover background. There is no `--accent-active` variable anywhere in the dist to alias to. And `--color-primary-hover` and `--color-primary-active` are not aliased to HeroUI's accent variables, which is why they read zero above. Every button variant sets `--button-bg-pressed` to its own hover value, so pressed-equals-hover is deliberate upstream rather than an oversight; the press feedback that does exist is `transform: scale(.97)`, in both themes. Measured on a rendered button, not derived.

**Disabled is not expressed at all.** No role declares a disabled colour and `--disabled-opacity` is not aliased, so every disabled state is HeroUI's default opacity over our colours, measuring 2.20:1 in light and 2.54:1 in dark. There is no missed value here; the role was never expressed.

**Field hover and focus are HeroUI's arithmetic over our value.** `--field-border` is ours; `--field-border-hover` and `--field-border-focus` are `color-mix`es HeroUI derives from it. The chain works, measured on a hovered control: alpha 0.48 to 0.521 in light, 0.345 to 0.404 in dark. The point is that the two most-touched states of the most-touched control in the product are values we do not name, and would move silently if those ratios changed.

**Nine call sites opt out of that chain.** `border-field-border`, `border-border` or `focus:border-accent` written into a field's `className` are unlayered app utilities and outrank `@heroui/styles`' layered hover and focus rules, so those controls do not move on hover and take the accent on focus. `app/AppShell.tsx`, `features/models/ModelsPage.tsx`, `features/settings/SettingsPage.tsx`, `shared/components/TablePagination.tsx`, `shared/components/ui.tsx`. Pre-existing, a styling-layer inconsistency rather than a token one, and untouched here.

**Link has no active state, and adding one is not a small change.** The link chain runs in opposite directions per theme: hover recedes in light (6.01 to 4.70 on the card) and advances in dark (6.64 to 9.43). Both clear AA and both give a visible hover, so the inconsistency is in the convention rather than the behaviour. Continuing light's chain fails AA on all four light surfaces (3.05 to 3.71) and reversing it collides with `primary-active`. It would also argue with an existing decision at `globals.css:1263`, where `.otari-markdown a:hover` declines a hover tint on purpose. Out of scope.

**`--color-primary-subtle` has no hover state and that is not a gap:** 18 `bg-` sites and no `hover:bg-primary-subtle` anywhere in `web/src`. Checked and closed.

## What this does not do

- **Radius and disabled opacity still do not follow the tokens.** `--radius` has 183 consumers in the built stylesheet and `--disabled-opacity` 53, and neither is aliased to anything we name. That gap is real, predates this change, and is untouched here.
- **`ShareCard.tsx` will drift.** It renders to a PNG through an ``-rasterized SVG, where `var()` does not resolve, so its fifteen colors are literal by necessity rather than by neglect. They were sampled from the palette this change replaces, so the card will drift from the dashboard it represents. It is left out of scope on purpose: bringing it back into alignment is a second pass with its own constraint, which is that the card is read in someone else's timeline and not on our page.
- **The 4px accent bar for `warning` and `success` reads brown rather than amber.** These tokens are constrained by having to work as text; a non-text step would let the bar read closer to the hue the word implies. A known limitation, not scheduled here.

## Known issues we accepted

- **The warning and attention hierarchy inverts in light.** No single pair of values orders them correctly in both themes. It is a per-theme exception or it is accepted, and we accepted it.
- **The light nav rail is pure white with the canvas darker than the chrome**, as described above. Deliberate, and the mirror of the dark ramp.
- **Light `--color-warning` is a bronze rather than an amber.** Forced rather than chosen: `text-warning` is real text in 23 places so it must clear 4.5:1, and at that hue the sRGB ceiling collapses as lightness rises. It was put on screen and looked at before being accepted. If it reads badly the fix is those 23 call sites, not the token.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

No tracking issue. This change came from direct design direction rather than a reported defect. Several pre-existing contrast defects were found while measuring and are deliberately **not** fixed here; they are not yet filed.

## Checklist

- [ ] Documentation was updated where necessary. No documentation describes these values; `.github/skills/frontend-standards/design-tokens.md` describes the token families and their rules, which are unchanged.
- [ ] If the API contract changed, I regenerated the OpenAPI spec. No API contract changed; this is `web/` only.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

Values were measured on the running dashboard rather than computed and asserted. Several claims in earlier drafts of this work were withdrawn after measurement contradicted them, including one about colour-vision deficiency that appears corrected in the section above.

### DIFF
--- a/web/index.html
+++ b/web/index.html
@@ -13,7 +13,7 @@
     <!-- Browser/OS chrome tint. A meta tag cannot read a custom property, so this
          is a literal copy of --color-primary; the matching `theme_color` in
          /pwa/manifest.webmanifest is the other one. Change all three together. -->
-    <meta name="theme-color" content="#4a7d8f" />
+    <meta name="theme-color" content="#5e6ad2" />
     <meta name="mobile-web-app-capable" content="yes" />
     <meta name="apple-mobile-web-app-capable" content="yes" />
     <meta name="apple-mobile-web-app-title" content="Otari" />
@@ -62,14 +62,14 @@
         align-items: center;
         justify-content: center;
         min-height: 100vh;
-        background: var(--color-background, #ffffff);
+        background: var(--color-background, #f7f8f8);
       }
       .app-boot-spinner {
         width: 1.75rem;
         height: 1.75rem;
         border-radius: 9999px;
-        border: 2px solid var(--color-border, #d6dce1);
-        border-top-color: var(--color-primary, #4a7d8f);
+        border: 2px solid var(--color-border, rgb(0 0 0 / 0.06));
+        border-top-color: var(--color-primary, #5e6ad2);
         animation: app-boot-spin 0.7s linear infinite;
       }
       @keyframes app-boot-spin {
--- a/web/public/pwa/manifest.webmanifest
+++ b/web/public/pwa/manifest.webmanifest
@@ -7,8 +7,8 @@
   "scope": "/",
   "display": "standalone",
   "orientation": "any",
-  "background_color": "#ffffff",
-  "theme_color": "#4a7d8f",
+  "background_color": "#f7f8f8",
+  "theme_color": "#5e6ad2",
   "icons": [
     {
       "src": "/pwa/icon-192.png",
--- a/web/src/app/AppShell.tsx
+++ b/web/src/app/AppShell.tsx
@@ -652,7 +652,7 @@ function AppShellChrome() {
         type="button"
         inert={backgroundInert}
         onClick={skipToMain}
-        className="sr-only focus:not-sr-only focus:absolute focus:top-3 focus:left-3 focus:z-50 focus:rounded-lg focus:border focus:border-accent focus:bg-surface focus:px-4 focus:py-2 focus:text-sm focus:font-medium focus:text-link focus:shadow-md focus:outline-none"
+        className="sr-only focus:not-sr-only focus:absolute focus:top-3 focus:left-3 focus:z-50 focus:rounded-lg focus:border focus:border-accent focus:bg-surface focus:px-4 focus:py-2 focus:text-sm focus:font-medium focus:text-link focus:shadow-elevation-md focus:outline-none"
       >
         Skip to main content
       </button>
--- a/web/src/app/ConnectionStatus.tsx
+++ b/web/src/app/ConnectionStatus.tsx
@@ -50,7 +50,7 @@ export function ConnectionStatus() {
     <div
       role="alert"
       aria-live="assertive"
-      className="fixed right-4 bottom-4 z-50 flex max-w-sm items-start gap-2.5 rounded-lg border border-danger bg-danger-subtle px-4 py-3 text-sm text-danger shadow-lg"
+      className="fixed right-4 bottom-4 z-50 flex max-w-sm items-start gap-2.5 rounded-lg border border-danger bg-danger-subtle px-4 py-3 text-sm text-danger shadow-elevation-lg"
     >
       <svg
         viewBox="0 0 24 24"
--- a/web/src/app/UpdatePrompt.tsx
+++ b/web/src/app/UpdatePrompt.tsx
@@ -41,7 +41,7 @@ export function UpdatePrompt() {
     <div className="pointer-events-none absolute inset-x-0 top-0 z-50 flex justify-center">
       <div
         role="status"
-        className="pointer-events-auto mt-1.5 flex items-center gap-3 rounded-full border border-accent bg-primary-subtle py-1.5 pr-1.5 pl-4 text-sm text-primary-subtle-foreground shadow-md"
+        className="pointer-events-auto mt-1.5 flex items-center gap-3 rounded-full border border-accent bg-primary-subtle py-1.5 pr-1.5 pl-4 text-sm text-primary-subtle-foreground shadow-elevation-md"
       >
         <span>
           <strong className="font-semibold">An update is available.</strong>{" "}
--- a/web/src/app/nav/AccountMenu.tsx
+++ b/web/src/app/nav/AccountMenu.tsx
@@ -29,8 +29,8 @@ import {
 } from "./rowStyles"
 
 // The control that ends the sidebar, and the menu it opens: account settings,
-// appearance, the legal pages, and the way out. The design's
-// "Menu member · Linear order" artboard is the order and the geometry. Who you
+// appearance, the legal pages, and the way out. The design's account-menu
+// artboard is the order and the geometry. Who you
 // are signed in as is the trigger's own line, not a row inside the menu: the
 // menu repeated it under an avatar the trigger already draws, so that block is
 // gone and /account is where an identity is actually read.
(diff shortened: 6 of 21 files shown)


jigjigjig:
Superseded by #871, which carries the same four commits on a renamed branch.

PR #854, opened 2026-08-28
feat(api)!: remove GET /v1/usage/summary.csv

njbrake:
## Description

`GET /v1/usage/summary.csv` was added with the Usage & analytics page (#345) to serve that page's CSV download button. The button was removed on 2026-07-22 in `a304eb71`, "chore(dashboard): remove usage CSV export", which took the action and its client helpers and left the route mounted with no stated reason to keep it. Nothing has called it since: no dashboard code path, and no SDK, which lists it under `[excluded]` in the endpoint manifest as "not yet wrapped".

Removed completely rather than merely unrouted: the handler, the spreadsheet-formula guard and dimension-label map that existed only for it, the now-unused `csv` and `io` imports, the four CSV tests, the CSV stanza of the two workspace-scope tests (their summary and series assertions stay, which is what those tests are about), the API reference row, the manifest line, and the regenerated spec, Postman collection and dashboard client.

`_breakdown` keeps its `limit=None` arm, with a note saying no route asks for it today. It is what makes the top-N fold optional rather than assumed, and the next caller that must not truncate wants exactly it.

### This is a breaking change, and worth naming

It was a published route in `docs/public/openapi.json`, and the use case it was built for (an uncapped export for finance) is the kind that lives in somebody's script rather than in our UI. **It is also the only uncapped breakdown the API had**: `/v1/usage/summary` folds everything past the top N of each dimension into a reconciling `other` row, and `/v1/usage` returns raw rows rather than aggregates. So a caller who wanted every group of a breakdown must now page `/v1/usage` and aggregate it themselves.

I could not find a consumer, but "no first-party consumer" is not "no consumer", and the spec is how an external caller would have found it. If it turns out to matter, the right restoration is a tenant-scoped one: #837's organization-scoped usage surface deliberately never grew a CSV counterpart, so bringing it back deployment-wide would re-add a surface only an operator can use.

## PR Type

- [x] Refactor

## Relevant issues

None. Raised in review of #842 while auditing what the organization-scoped usage surface should and should not mirror.

## Checklist

`make lint`, `make typecheck`, `make openapi-check`, `make postman-check`, the OSS-edition smoke, the full dashboard suite (1583) and the affected integration suites (184) all pass locally. I diffed the test-function lists before and after to confirm the deletion took the four CSV tests and nothing else.

## AI Usage

**AI Model/Tool used:**

@njbrake asked why the endpoint existed and whether it could go. The session traced it to the commit that added it and the one that removed its only caller, checked the SDK manifest and the dashboard for any remaining consumer, and reported that removal was safe but breaking, with the uncapped-export caveat above. He decided to remove it; this is that change.

## Summary

- Removed the unused `GET /v1/usage/summary.csv` endpoint.
- Removed its handler, tests, documentation, API manifest entry, and generated client artifacts.
- Retained `_breakdown(limit=None)` for future uncapped callers.
- Callers that need complete breakdowns must now page through `/v1/usage` and aggregate the rows.


njbrake:
I had originally added this for a small dashboard feature that we didn't use. Deletion of this is intended in order to streamline our design. I'm open to a github issue for a feature request around usage exports, but I'd like to think through and propelry implement the idea instead of taking the one that I'm removing here.

PR #859, opened 2026-08-28
fix(hybrid): forward Anthropic 1h cache-write tokens in usage reports

### LINKED ISSUE #856: fix(hybrid): forward Anthropic one-hour cache-write tokens in usage reports
## Problem

Otari captures Anthropic cache-write TTL detail in `GatewayUsage.cache_write_1h_tokens`, but the hybrid usage report drops it.

`src/gateway/api/routes/messages.py` records the one-hour bucket, while `_report_platform_usage` in `src/gateway/api/routes/_platform.py` forwards only `cache_read_tokens` and aggregate `cache_write_tokens`. The `POST /gateway/usage` example and field description in `docs/hybrid-mode-protocol.md` omit the one-hour sub-bucket as well.

A hybrid settlement authority therefore cannot distinguish five-minute cache writes from one-hour writes and cannot apply the distinct one-hour rate implemented by #357.

## Proposed change

- Add optional `cache_write_1h_tokens` to the hybrid usage payload and protocol documentation.
- Forward it from `GatewayUsage` with the existing extraction helper.
- Keep compatibility with older receivers and reports: omitted or absent means zero.
- Extend the hybrid Messages integration test to cover a one-hour cache-write report.

## Acceptance criteria

- An Anthropic Messages response with one-hour cache creation reports both total `cache_write_tokens` and the `cache_write_1h_tokens` subset.
- Five-minute-only writes report a zero one-hour bucket.
- Chat, Responses, and reports with no cache usage remain unchanged.
- `docs/hybrid-mode-protocol.md` defines the field as an optional subset of total cache-write tokens.

## Related

- #357 implemented TTL-aware Anthropic cache-write pricing and capture.
- #376 tracks provider-metered pricing correctness.
- #298 tracks prompt-cache accounting more broadly.

AloysJehwin:
## Description

`GatewayUsage` already carries `cache_write_1h_tokens`, but `_report_platform_usage` was forwarding only `cache_read_tokens` and the aggregate `cache_write_tokens`. A settlement authority could see that a cache write happened but not that it used the one-hour TTL, so it couldn't apply the distinct 1h rate that #357 added.

This forwards it with the existing `cache_write_1h_tokens_of` helper and documents it in the protocol.

The one design call worth flagging: the acceptance criteria ask for a zero 1h bucket on five-minute-only writes *and* for Chat, Responses and no-cache reports to stay unchanged. Emitting the key unconditionally satisfies the first and breaks the second — it changed the payload for every OpenAI and Gemini report, and two existing tests failed on exactly that. So the key is emitted only when `cache_write_tokens` is non-zero. An Anthropic report always carries it (`0` meaning every write was 5m), and a provider with no cache-write concept sends the byte-identical payload it sent before. That way I didn't have to edit any existing test to accommodate the change.

Documented as an optional subset, never an addition, so the 5m portion is `cache_write_tokens - cache_write_1h_tokens` and an absent key means zero for older gateways.

## PR Type

- [x] Bug Fix
- [x] Documentation

## Relevant issues

Fixes #856

## Checklist

- [x] If the API contract changed, I regenerated the OpenAPI spec — not applicable here: `POST /gateway/usage` is an outbound call to the platform, so otari's own OpenAPI surface is unchanged.

Test results:

```
make lint       -> architecture ✅, single alembic head, ruff: All checks passed!
make typecheck  -> Success: no issues found in 535 source files
pytest tests/unit tests/integration -> 2664 passed, 9 skipped
```

The new test is `test_hybrid_mode_reports_one_hour_cache_write_subset`, which asserts the 1h bucket is reported as a subset (30 total, 10 at 1h, 20 remaining at 5m). The existing messages test now also pins `cache_write_1h_tokens: 0` for the 5m-only case, and the chat/responses payload assertions are untouched and still pass.

## AI Usage

Reproduced the gap by reading `_report_platform_usage` against `messages.py`, then ran the four hybrid suites plus the full unit and integration suites locally under Docker. The conditional-emission decision above came from the two chat/responses failures — worth a look in review, since the alternative reading is to always emit and update those two assertions instead.

## Summary

- Added Anthropic one-hour cache-write token reporting.
- Documented the optional field in the hybrid protocol.
- Preserved existing payloads for providers without cache-write data.
- Added coverage for one-hour, five-minute-only, Chat, Responses, and no-cache reports.

## Technical notes

- `cache_write_1h_tokens` is emitted only when cache-write data exists.
- The field is an optional subset of `cache_write_tokens`.
- Five-minute-only writes report zero one-hour tokens.
- Older receivers remain compatible because an omitted field means zero.

## Validation

- Lint, type checking, and unit/integration tests passed.

### DIFF
--- a/docs/hybrid-mode-protocol.md
+++ b/docs/hybrid-mode-protocol.md
@@ -340,7 +340,9 @@ Content-Type: application/json
     "completion_tokens": 7,
     "total_tokens": 20,
     "cache_read_tokens": 8,            // provider cache-read input tokens
-    "cache_write_tokens": 0           // cache-write (creation) input tokens; Anthropic only
+    "cache_write_tokens": 0,          // cache-write (creation) input tokens; Anthropic only
+    "cache_write_1h_tokens": 0        // subset of cache_write_tokens written with
+                                      // a 1-hour TTL; Anthropic only
   },
   "error_class": "http_401",           // optional on error; omitted when the
                                        // Otari can't classify the failure
@@ -406,6 +408,17 @@ that in mind:
   cache. `cache_read_tokens` and `cache_write_tokens` are reported **separately**
   and are not part of `prompt_tokens`. `cache_write_tokens` is a true cache
   creation charge billed at a premium.
+
+`cache_write_1h_tokens` is an optional **subset** of `cache_write_tokens`: the
+portion created with a one-hour TTL rather than the five-minute default, which
+Anthropic bills at a higher rate. It is never added to `cache_write_tokens`, so
+the five-minute portion is `cache_write_tokens - cache_write_1h_tokens`. A
+receiver that does not price the two TTLs separately can ignore it.
+
+The key is present whenever `cache_write_tokens` is non-zero, and `0` there means
+every write used the five-minute TTL. It is omitted entirely when the report
+carries no cache writes, which is every OpenAI and Gemini report. Absent means
+zero, so a report from an older gateway prices exactly as before.
 
 The platform must accept these additive keys with lenient parsing; a handler that
 rejects unknown fields would 422 the report (a non-retryable status), silently
--- a/docs/hybrid-mode-protocol.md
+++ b/docs/hybrid-mode-protocol.md
@@ -415,7 +415,7 @@ receiver that does not price the two TTLs separately can ignore it.
 
 The key is present whenever `cache_write_tokens` is non-zero, and `0` there means
 every write used the five-minute TTL. It is omitted entirely when the report
-carries no cache writes, which is every OpenAI and Gemini report — hence its
+carries no cache writes, which is every OpenAI and Gemini report; hence its
 absence from the example above. Absent means zero, so a report from an older
 gateway prices exactly as before. A cache-writing report looks like:
 
--- a/docs/hybrid-mode-protocol.md
+++ b/docs/hybrid-mode-protocol.md
@@ -340,9 +340,7 @@ Content-Type: application/json
     "completion_tokens": 7,
     "total_tokens": 20,
     "cache_read_tokens": 8,            // provider cache-read input tokens
-    "cache_write_tokens": 0,          // cache-write (creation) input tokens; Anthropic only
-    "cache_write_1h_tokens": 0        // subset of cache_write_tokens written with
-                                      // a 1-hour TTL; Anthropic only
+    "cache_write_tokens": 0           // cache-write (creation) input tokens; Anthropic only
   },
   "error_class": "http_401",           // optional on error; omitted when the
                                        // Otari can't classify the failure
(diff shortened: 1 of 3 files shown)


AloysJehwin:
Good catch, fixed. The zero-write example did contradict the omission rule two paragraphs down.

Dropped the field from that example and added a separate cache-writing one (30 written, 10 at 1h) so the subset relationship is still shown somewhere concrete.

docs/hybrid-mode-protocol.md
```diff
@@ -407,6 +407,32 @@ that in mind:
...
 
+`cache_write_1h_tokens` is an optional **subset** of `cache_write_tokens`: the
+portion created with a one-hour TTL rather than the five-minute default, which
+Anthropic bills at a higher rate. It is never added to `cache_write_tokens`, so
+the five-minute portion is `cache_write_tokens - cache_write_1h_tokens`. A
+receiver that does not price the two TTLs separately can ignore it.
+
+The key is present whenever `cache_write_tokens` is non-zero, and `0` there means
+every write used the five-minute TTL. It is omitted entirely when the report
+carries no cache writes, which is every OpenAI and Gemini report — hence its
```
khaledosman: Em dash. `AGENTS.md` (Writing style) rules them out in docs prose: use a comma or semicolon, or split the sentence.

tests/integration/test_hybrid_mode_messages.py
```diff
@@ -159,9 +184,7 @@ async def fake_post_platform(
         if url.endswith("/gateway/provider-keys/resolve"):
             return httpx.Response(
                 200,
-                json=_resolve_payload(
-                    [_attempt(0, attempt_id, "claude-3-5-sonnet-20241022", "sk-platform-key")]
-                ),
+                json=_resolve_payload([_attempt(0, attempt_id, "claude-3-5-sonnet-20241022", "sk-platform-key")]),
```
khaledosman: Unrelated reformatting (same at line 828, plus the blank line at 1142). `make lint` is `ruff check` only with `line-length = 120` and no CI job runs a formatter, so neither form is a lint fix. Revert these three so the diff is only the change.

tests/integration/test_hybrid_mode_messages.py
```diff
@@ -93,6 +93,31 @@ def _message_response(text: str = "hello") -> MessageResponse:
     )
 
 
+def _message_response_with_1h_cache_write() -> MessageResponse:
+    """A response whose cache creation splits across the 5m and 1h TTL buckets."""
+    from anthropic.types.cache_creation import CacheCreation
```
khaledosman: Move this to the module's existing anthropic imports at the top; nothing here needs a deferred import.

khaledosman (approved):
The conditional emission is sound, and worth recording why: the 1h bucket is a clamped subset everywhere it is produced (`metered_pricing.py:207`, `streaming._merge_usage`), so `cache_write_tokens == 0` alongside a non-zero 1h bucket is not reachable and "absent means zero" cannot drop a real value. Emitting only on a cache-writing report was the right read of the two acceptance criteria.

Three non-blocking cleanups inline.

---

njbrake (approved):
Approved.

On the conditional emission you flagged: keep it. The platform must tolerate an absent key regardless, since gateways older than this PR send none, so unconditional emission would not simplify the receiver. Conditional also fails loudly rather than only against old gateways, and a 422 here silently drops the report.

Pushed a commit reverting the unrelated `_resolve_payload` reformatting and the blank-line edit, moving the `CacheCreation` import to the top, and dropping the assertion that restated the two values above it.

This covers the Messages path. #987 tracks Anthropic cache writes on chat completions, which are dropped before they reach `GatewayUsage`.

PR #861, opened 2026-08-29
feat(dashboard): record existing telemetry in Mixpanel when a key is present

jigjigjig:
## Description

Record the existing dashboard telemetry catalog in Mixpanel when a project token is present. One codebase, gated by key presence: OSS users have no Mixpanel key, so this must do nothing.

- **No key:** load nothing and track nothing. No SDK fetch, no network, no console noise on a deployed host.
- **`make dev` without a key:** the browser logs exactly `Mixpanel not initialized` once (localhost or Vite DEV), then stops. It does not init.
- **With `VITE_MIXPANEL_TOKEN`:** dynamically import `mixpanel-browser` and forward the existing `useTelemetry` calls. Call sites are unchanged.
- Event names are unchanged from `TELEMETRY_EVENTS`. `CHECKOUT_CANCELLED` stays named and unfired (no wallet).
- This is not the HubSpot overlay-only pattern. Nathan floated "impl in otari-ai, interface in OSS" then said he might have been thinking of HubSpot (otari-ai#1749, which 404s). Toto's simpler path wins: events live in this dashboard; a missing key is a no-op.

The token is a Vite build-time env var (`VITE_MIXPANEL_TOKEN`). `make dashboard` sources the repo-root `.env` so the same file `make dev` uses can supply it. The Docker web stage accepts an optional `VITE_MIXPANEL_TOKEN` build ARG.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Hygiene/telemetry follow-up from Toto in #otari-oss-refactor and Fede's Aug 21 notes. No issue number. Started without Fede's supervision; not requesting reviewers.

## Checklist

- [x] I have added or updated tests that cover my change (dashboard Vitest: no-key load/track gate, local `Mixpanel not initialized` log, catalog pin, `CHECKOUT_CANCELLED` unfired).
- [x] I ran the Definition of Done checks locally (`pnpm --dir web run lint`, `pnpm --dir web run typecheck`, `pnpm --dir web test`: 1606 passed). Backend `make lint` / `make typecheck` / `make test` were not required (no `src/gateway/` Python change).
- [x] Documentation was updated where necessary (`web/AGENTS.md`, frontend-standards skill, `web/README.md`).
- [x] If the API contract changed, I regenerated the OpenAPI spec (`uv run python scripts/generate_openapi.py`). (N/A: no API change.)

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:** Cursor Grok 4.6 (cloud agent)

**Any additional AI details you'd like to share:** Product decisions were locked from Slack (#otari-oss-refactor) and Granola (Aug 21 Mixpanel notes). Draft only; do not request reviewers (especially not jigjigjig / Fede).

<div><a href="https://cursor.com/agents/bc-fc0a8c4a-3a34-4a7e-981d-a002e18efedc?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-fc0a8c4a-3a34-4a7e-981d-a002e18efedc&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

## Summary

Adds optional Mixpanel analytics to the dashboard.

- Records existing telemetry events only when `VITE_MIXPANEL_TOKEN` is configured.
- Keeps keyless builds free of Mixpanel loading, tracking, and production console output.
- Documents event coverage, configuration, and deployment behavior.
- Adds Docker and Vite configuration for build-time token handling.
- Adds comprehensive tests for tracking, identity, consent, loading, and failure behavior.

### DIFF
--- a/.github/skills/frontend-standards/SKILL.md
+++ b/.github/skills/frontend-standards/SKILL.md
@@ -7,9 +7,11 @@ description: Guidelines for the otari admin dashboard (`web/`), React 19 + TypeS
 
 `web/` is the dashboard shared by standalone, hosted, and hybrid deployments. It renders the
 management UI or hybrid landing page from `/v1/bootstrap` and uses an HttpOnly session for
-management calls. It is an operator tool, not a marketing surface. The base tracker under
-`src/shared/telemetry/` is a no-op seam an overlay may replace; do not give it an analytics
-implementation in this repository.
+management calls. It is an operator tool, not a marketing surface. Telemetry
+under `src/shared/telemetry/` is gated by `VITE_MIXPANEL_TOKEN`: with no key the Mixpanel SDK
+is never loaded and nothing is tracked, which is the OSS default. Do not invent events beyond
+`TELEMETRY_EVENTS`, do not fire `CHECKOUT_CANCELLED` (no wallet), and do not add a second
+analytics vendor.
 
 Stack: React 19 (with the React Compiler), TypeScript (`strict`), HeroUI v3 (`@heroui/react`),
 Tailwind CSS v4, TanStack Query, TanStack Router (file-based, `web/src/routes/`), Vite,
--- a/Makefile
+++ b/Makefile
@@ -26,6 +26,9 @@ dev:
 # degrades to the tutorial page) and before building a wheel. The Docker image
 # builds it in its own web stage and needs no local Node.
 dashboard: web/node_modules/.install-stamp
+	@set -a; \
+	if [ -f .env ]; then . ./.env; fi; \
+	set +a; \
 	pnpm --dir web run build
 
 # Reinstalling on every rebuild is the wrong price to pay now that the READMEs
--- a/Makefile
+++ b/Makefile
@@ -26,9 +26,6 @@ dev:
 # degrades to the tutorial page) and before building a wheel. The Docker image
 # builds it in its own web stage and needs no local Node.
 dashboard: web/node_modules/.install-stamp
-	@set -a; \
-	if [ -f .env ]; then . ./.env; fi; \
-	set +a; \
 	pnpm --dir web run build
 
 # Reinstalling on every rebuild is the wrong price to pay now that the READMEs
--- a/web/README.md
+++ b/web/README.md
@@ -66,10 +66,12 @@ filesystem events. Fall back to polling:
 VITE_USE_POLLING=1 pnpm run dev
 ```
 
-Dashboard analytics are gated by `VITE_MIXPANEL_TOKEN` (inlined at build time,
-also read from the repo-root `.env` by `make dashboard`). Without a token the
-SDK is never loaded. A local dashboard without a token logs `Mixpanel not
-initialized` once in the browser console.
+Dashboard analytics are gated by `VITE_MIXPANEL_TOKEN`, inlined at build time.
+Vite reads it from the repo-root `.env` (`envDir` in `vite.config.ts`), so
+`pnpm run dev` and `make dashboard` take it from the same file. Without a token
+the SDK is never loaded and nothing is recorded; the dev server logs `Mixpanel
+not initialized` once in the browser console. What a build with a token records
+is documented for operators in [docs/dashboard.md](../docs/dashboard.md).
 
 ## Build
 
--- a/web/README.md
+++ b/web/README.md
@@ -66,6 +66,11 @@ filesystem events. Fall back to polling:
 VITE_USE_POLLING=1 pnpm run dev
 ```
 
+Dashboard analytics are gated by `VITE_MIXPANEL_TOKEN` (inlined at build time,
+also read from the repo-root `.env` by `make dashboard`). Without a token the
+SDK is never loaded. A local dashboard without a token logs `Mixpanel not
+initialized` once in the browser console.
+
 ## Build
 
 ```bash
--- a/web/package.json
+++ b/web/package.json
@@ -26,6 +26,7 @@
     "@tanstack/react-query": "^5.101.4",
     "@tanstack/react-router": "^1.170.31",
     "clsx": "2.1.1",
+    "mixpanel-browser": "^2.82.1",
     "react": "^19.2.8",
     "react-aria-components": "1.20.0",
     "react-dom": "^19.2.8",
--- a/web/src/shared/telemetry/mixpanelToken.ts
+++ b/web/src/shared/telemetry/mixpanelToken.ts
@@ -1,0 +1,35 @@
+/**
+ * The Mixpanel project token the dashboard was built with, if any.
+ *
+ * Vite only exposes `VITE_*` to the client, and it inlines the value at build
+ * time (`make dashboard`, `pnpm run dev`, the Docker web stage). An empty or
+ * whitespace-only value is treated as absent: that is the OSS default, and it
+ * is what keeps the SDK from loading.
+ */
+export function readMixpanelToken(): string | undefined {
+  const raw = import.meta.env.VITE_MIXPANEL_TOKEN
+  if (typeof raw !== "string") {
+    return undefined
+  }
+  const token = raw.trim()
+  return token === "" ? undefined : token
+}
+
+const LOCAL_HOSTS = new Set(["localhost", "127.0.0.1", "[::1]"])
+
+/**
+ * Whether this page is a local dashboard, the case that may announce a missing
+ * Mixpanel key.
+ *
+ * `make dev` serves a production Vite build on localhost, so `import.meta.env.DEV`
+ * is false there and cannot be the only signal. A deployed OSS host is neither
+ * DEV nor a loopback name, and must stay silent.
+ */
+export function isLocalDashboard(
+  hostname: string = typeof window === "undefined"
+    ? ""
+    : window.location.hostname,
+  isDev: boolean = import.meta.env.DEV,
+): boolean {
+  return isDev || LOCAL_HOSTS.has(hostname)
+}
--- a/web/src/shared/telemetry/mixpanelToken.ts
+++ b/web/src/shared/telemetry/mixpanelToken.ts
@@ -6,8 +6,9 @@
  * whitespace-only value is treated as absent: that is the OSS default, and it
  * is what keeps the SDK from loading.
  */
-export function readMixpanelToken(): string | undefined {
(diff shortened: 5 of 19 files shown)


njbrake (approved):
Approving. Telemetry lives in this dashboard, gated on `VITE_MIXPANEL_TOKEN`, and a missing key stays a no-op.

Pushed three fixes in df0020f:

`createMixpanelTelemetry.recordEvent` now checks consent. `types.ts` puts that gate on the seam implementation and it was missing, so a stored refusal would have withheld the identity while events kept firing. Inert today, since token presence is the only consent source; not once anything can answer `"denied"`.

Corrected the consent docstrings in `types.ts` and `TelemetryIdentity.tsx`, which still claimed the base answers `"unknown"` and always will.

Listed `VITE_MIXPANEL_TOKEN` in `.env.example`, the file `make dashboard` sources.

Two to carry rather than fix. A keyless build still emits a 423 kB `mixpanelClient` chunk it never fetches, so the artifact ships the SDK. And Mixpanel's defaults leave `track_marketing`, `save_referrer` and a cross-subdomain cookie on, with no `api_host` for EU residency.

web/src/shared/telemetry/overlayTelemetry.ts
```diff
@@ -1,53 +1,120 @@
...
+    }
+    if (client) {
+      apply(client)
+      return
+    }
+    pending.push(apply)
+  }
+
+  return {
+    consent: "granted",
```
khaledosman, line 78: The queued wrapper hardcodes `consent: "granted"` rather than reporting what the loaded client answers, and `identify` is not consent-gated on either side of the queue.

`TelemetryIdentity.tsx:52` gates the identity on exactly this value, so it always reads granted. `mixpanelClient.ts:70` makes consent a parameter precisely so a build can pass something else, and `loadMixpanelClient` (line 39) never does. The moment anything does, or an overlay supplies its own `loadClient`, the membership id, organization id, organization name and role still reach `mixpanel.identify` and `people.set`: `createMixpanelTelemetry.identify` (lines 96-105) has no consent check, and this wrapper has already told `TelemetryIdentity` it may hand one over. `recordEvent` is gated (line 73 of `mixpanelClient.ts`); `identify` is the call `types.ts` calls "the one call that hands a tracker something durable about a person," and it is the one with no gate.

The wrapper cannot know consent before the chunk resolves, so the fix belongs in `createMixpanelTelemetry`: check `consent !== "granted"` in `identify` the way `recordEvent` already does. That closes it in one place with no async problem.

web/package.json
```diff
@@ -26,6 +26,7 @@
     "@tanstack/react-query": "^5.101.4",
     "@tanstack/react-router": "^1.170.31",
     "clsx": "2.1.1",
+    "mixpanel-browser": "^2.82.1",
```
khaledosman, line 29: `import mixpanel from "mixpanel-browser"` resolves to `dist/mixpanel.module.js`, which statically bundles the rrweb session recorder (38 references to it; `dist/mixpanel-core.cjs.js` has zero). So roughly 60% of the 423 kB chunk is replay code that `record_sessions_percent: 0` guarantees never runs, and the dependency drags in six `@mixpanel/rrweb*` packages pinned at `2.0.0-alpha.18.x` into an OSS product's tree.

Import a core/no-recorder entry point instead. `performance.md` ("Watch the bundle when you add a dependency") is the rung, and `vite.config.ts` already reasons this way about recharts at ~368 kB, which is smaller than this.

web/src/shared/telemetry/mixpanelToken.ts
```diff
@@ -0,0 +1,36 @@
...
+ * is false there and cannot be the only signal. A deployed OSS host is neither
+ * DEV nor a loopback name, and must stay silent.
+ */
+export function isLocalDashboard(
+  hostname: string = typeof window === "undefined"
+    ? ""
+    : window.location.hostname,
+  isDev: boolean = import.meta.env.DEV,
+): boolean {
+  return isDev || LOCAL_HOSTS.has(hostname)
```
khaledosman: The loopback branch makes `Mixpanel not initialized` fire for the documented OSS deployment, not just for a mozilla.ai developer. `README.md:101` and `docs/dashboard.md:51` both tell operators to open `http://localhost:8000/`, so a self-hosted gateway is a loopback hostname serving a production build: exactly the case the docstring describes as `make dev`. The two are indistinguishable from inside the page, so every self-hosted operator gets a console line implying the product wanted to track them, on a build that never will.

Drop the loopback set and keep `import.meta.env.DEV`. A developer who wants the reminder gets it from `pnpm run dev`; `make dev` losing it costs nothing next to telling every OSS operator that Mixpanel failed to start.

.github/skills/frontend-standards/SKILL.md
```diff
@@ -7,11 +7,12 @@ description: Guidelines for the otari admin dashboard (`web/`), React 19 + TypeS
 
 `web/` is the standalone admin dashboard: a React SPA that talks to the gateway's management
 API with the master key. It is an operator tool, not a general-purpose app; keep its footprint
-small and its conventions consistent with what is already there. There is no analytics and no
-marketing surface here; do not add either to match some other repo. `src/shared/telemetry/`
-is not a counterexample: it is a no-op seam an overlay build replaces, it declares no vendor
-dependency, and the base tracker records nothing. Wiring a call site to it is fine; giving it
-an implementation here is not.
+small and its conventions consistent with what is already there. There is no marketing
+surface here; do not add one to match some other repo. Telemetry is gated by
```
khaledosman: This is the standard that forbade the change in the same PR that makes it, and `overlayTelemetry.test.ts` (the test `web/AGENTS.md` named as the thing that fails "if either half of that stops being true") is edited in the same commit to allowlist the vendor. Both halves of the guard move together, so nothing is left that would have objected.

That may well be the right end state, but it needs an issue and a human owner before the rule is rewritten, not a description pointing at a 404.

web/README.md
```diff
@@ -70,6 +70,11 @@ filesystem events. Fall back to polling:
 VITE_USE_POLLING=1 pnpm run dev
 ```
 
+Dashboard analytics are gated by `VITE_MIXPANEL_TOKEN` (inlined at build time,
```
khaledosman: This is the only place a reader learns about Mixpanel, and it is developer-facing. `docs/dashboard.md` is the operator guide bundled into the running app and says nothing; with a token, this build asserts `consent: "granted"` from build configuration alone, has no consent UI, and sends membership id, organization id, organization name and role to a third party. Add a paragraph to `docs/dashboard.md` so the person using the dashboard can find out from the dashboard's own guide.

Makefile
```diff
@@ -26,6 +26,9 @@ dev:
 # degrades to the tutorial page) and before building a wheel. The Docker image
 # builds it in its own web stage and needs no local Node.
 dashboard: web/node_modules/.install-stamp
+	@set -a; \
+	if [ -f .env ]; then . ./.env; fi; \
+	set +a; \
 	pnpm --dir web run build
```
khaledosman: Sourcing the repo-root `.env` exports all of it, provider keys included, into the Vite build environment to deliver one `VITE_*` var. Vite only inlines `VITE_*`, so nothing leaks into the bundle, but Vite has a first-class way to do this: `envDir` in `web/vite.config.ts` pointed at the repo root. That also fixes the asymmetry this leaves behind, where `make dashboard` reads the root `.env` and the `pnpm run dev` path the README documents does not, so the token has to live in two files.

(The `dev` target above uses this shell pattern already, so it is consistent; it is just that `dev` needs `OTARI_HOST`/`OTARI_PORT` in the recipe itself, and this one does not need anything in the recipe at all.)

web/src/shared/telemetry/overlayTelemetry.ts
```diff
@@ -1,53 +1,120 @@
...
+      client = loaded
+      for (const apply of pending) {
+        apply(loaded)
+      }
+      pending.length = 0
+    })
+    .catch(() => {
+      failed = true
+      pending.length = 0
+      client = NO_TELEMETRY
```
khaledosman: Both guards here are unreachable. `failed` is only ever set inside this `.catch`, and `.then`/`.catch` on one chain are mutually exclusive, so line 52 never fires. `client = NO_TELEMETRY` is never read either, because `enqueue` returns on `failed` (line 68) before it looks at `client`. Drop both and the failure path is `failed = true; pending.length = 0`.

khaledosman (changes requested):
Verified locally on `df0020f1`: `pnpm --dir web run typecheck`, `run lint`, and `test` (1617 passed) are all green, and `pnpm run build` with no token succeeds.

Three things need resolving before this lands, and one of them is not a code question.

**The rule this PR rewrites was a deliberate one.** `web/AGENTS.md` and `frontend-standards/SKILL.md` said the telemetry directory "is a no-op seam an overlay build replaces... Wiring a call site to it is fine; giving it an implementation here is not," and `overlayTelemetry.test.ts` was the guard that made it stick. This PR inverts the standard, edits the guard to allow the vendor, and cites the justification as Slack plus a Granola note plus `otari-ai#1749`, which the description says 404s. The description also says it was started without supervision and asks for no reviewers. Whether the OSS dashboard gets an analytics implementation at all is a product decision that needs a named human owner and an issue; it should not arrive as a side effect of a hygiene follow-up. Everything below assumes the answer turns out to be yes.

**Measured bundle impact.** With no token set, the build emits `mixpanelClient-*.js` at 423.35 kB raw / 125.34 kB gzip. That is the largest chunk in the bundle, above `heroui` at 416 kB, and it ships in the wheel and the Docker image for every OSS user who can never load it. It is not preloaded and there is no service-worker precache, so the runtime claim in the description does hold: nothing fetches it. But the artifact carries it. Details inline.

The remaining findings are inline.

---

PR #869, opened 2026-08-31
fix(dashboard): name the person at the foot of the sidebar, not their role

### LINKED ISSUE #832: fix(dashboard): every signed-in user is labelled "Operator" at the foot of the sidebar
The account control that ends the sidebar names the session "Operator / OP" for every signed-in user, including an ordinary member of a hosted deployment who is not the operator.

`sessionIdentity` (`web/src/app/nav/AccountMenu.tsx:76`) returns `{ name: "Operator", initials: "OP" }` whenever `session_type === "local_operator"`, and `GET /v1/deployment/bootstrap` hardcodes that value for hosted mode as well as standalone (`src/gateway/api/routes/bootstrap.py:284`). No path returns `hosted_user` today, so the `"Signed in"` fallback is unreachable and a hosted member is told they are the operator. The otari.ai superset build ships this component unchanged, so it shows there too.

Two fixes, worth taking in this order:

1. Key the label on `deployment_type` as well, so a hosted deployment falls back to "Signed in". One line, no new endpoint, and it stops the false claim immediately.
2. Name the actual person. The shell has no caller identity today: `/v1/deployment/bootstrap` is public and pre-auth so it cannot name the caller, and there is no `GET /v1/auth/me`. Sign-in returns `user_id` (`src/gateway/api/routes/auth_session.py:165`) but no email. Either add an authenticated route reporting the current identity, or have the shell match its `user_id` against `GET /v1/organizations/me/members`.

Acceptance: a member of a hosted deployment is not labelled "Operator"; a standalone operator still is.

njbrake:
## Description

The little account button at the bottom of the sidebar called everybody "Operator". That is a job, not a name, and on a shared deployment it was not even true: the button was reading a value that describes what kind of sign-in the whole deployment uses, which says the same thing for every person, so an ordinary member logging in was told they run the place. The reason it showed a job title at all is that nothing the dashboard could ask told it who was signed in. Now the server includes the signed-in person's own name and address in an answer the dashboard already asks for on every page load, and the button shows the person: their name, or their email address if they have not set a name yet. Someone who signs in to a single-user gateway still sees "Operator", because that is the name the gateway gives that account when it first sets itself up.

`GET /v1/organizations/me` gains `caller` (`user_id`, `email`, `full_name`), beside the standing fields it already publishes. It goes there rather than on a route of its own for the reason `deployment_operator` did in #852: the shell reads that context before it paints, so an identity taken from it costs no request and cannot arrive a beat after the chrome it names. Both name fields are nullable, because a local operator identity has no address and a member added to the roster by address has no name until they claim it, so the trigger falls back address-then-"Signed in".

This closes both halves of #832, not just the first: the issue's suggested one-line fix (key the label on `deployment_type`) would have stopped the false claim while still showing a role, and would still have mislabeled a standalone member who does not operate the deployment.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes #832

## Checklist

Two integration cases pin the new field (per-identity, and the address-less local operator); five component cases pin what the trigger draws. Also smoke-tested in a browser against a real gateway serving the built bundle: the operator's session reads "Operator", and an ordinary rostered member's reads their own address.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

Implementation, tests and the smoke run were done by the model in back-and-forth with @njbrake, who directed the approach: show the person's name, since "Operator" is a role.

### DIFF
--- a/web/src/app/nav/AccountMenu.tsx
+++ b/web/src/app/nav/AccountMenu.tsx
@@ -83,7 +83,7 @@ const MENU_DIVIDER = "h-px shrink-0 bg-border"
 // name: first boot provisions the identity with it (`OPERATOR_FULL_NAME`), and
 // the roster shows the same word in the same place.
 //
-// Three fallbacks, each for a real identity this deployment can hold. A member
+// Three answers, each for a real identity this deployment can hold. A member
 // added to the roster by address has no name until they claim it, so the
 // address stands in and is a better answer than a role. An identity with
 // neither, and a context that has not landed or could not be read, get
@@ -94,14 +94,23 @@ function sessionIdentity(caller: OrganizationContext["caller"]): {
   initials: string
 } {
   const named = caller?.full_name?.trim()
-  const addressed = caller?.email?.trim()
-  const name = named || addressed
-  if (!name) {
-    return { name: "Signed in", initials: "··" }
+  if (named) {
+    return { name: named, initials: initialsFor(named) }
   }
-  // `||`, not `??`: a name that is only whitespace trims to "" and has already
-  // lost the `name` slot to the address, so it must not keep the initials.
-  return { name, initials: initialsFor(named || localPart(name)) }
+  const addressed = caller?.email?.trim()
+  if (addressed) {
+    // An address has no words to take initials from, so its local part stands
+    // in, and that part is split on its own punctuation as well: an address
+    // spells a name with dots where a name uses spaces, so `ada.lovelace@…`
+    // initials as AL. A name is split on spaces alone, because the same
+    // punctuation means something else in one: `Ada Lovelace-Byron` is two
+    // names, and splitting the hyphen would initial her as AB.
+    return {
+      name: addressed,
+      initials: initialsFor(localPart(addressed), true),
+    }
+  }
+  return { name: "Signed in", initials: "··" }
 }
 
 /** The part of an address that names a person, which is the part before the host. */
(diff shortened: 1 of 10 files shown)


web/src/app/nav/AccountMenu.tsx
```diff
@@ -69,18 +70,64 @@ const MENU_ROW_DISABLED = "cursor-not-allowed text-muted opacity-60"
...
+// first and last word for a name in parts, and the first two letters of a name
+// that is one word. Addresses are split on their punctuation as well as on
+// spaces, so `ada.lovelace@…` initials as AL rather than AD.
+//
+// Counted in characters rather than in the code units `[0]` and `slice` count,
+// because a name outside the basic plane (an extension-B CJK character, an
+// emoji) is two units per character: indexing one takes half a surrogate pair
+// and the avatar draws the replacement mark instead of the letter.
+function initialsFor(source: string): string {
+  const words = source.split(/[\s._-]+/).filter(Boolean)
```
khaledosman: The comment above says addresses are split on punctuation "as well as on spaces", but one regex serves both sources, so the address rule is applied to names too. `Ada Lovelace-Byron` splits to three words and initials as `AB` instead of `AL`; `Jean-Luc Picard` gives `JP` (right by luck) while `Mary-Jane Watson` gives `MW`. Hyphenated surnames are common enough to be worth the split:

```ts
function initialsFor(source: string, splitPunctuation = false): string {
  const words = source.split(splitPunctuation ? /[\s._-]+/ : /\s+/).filter(Boolean)
```

with the address path at line 104 passing `true`. Nothing else calls it.

web/src/app/nav/AccountMenu.tsx
```diff
@@ -229,9 +276,10 @@ function AppearanceControl() {
 
 export function AccountMenu({ collapsed }: { collapsed: boolean }) {
   const { logout } = useAuth()
-  const { session_type, management_url, docs_url } = useDeployment()
+  const { management_url, docs_url } = useDeployment()
+  const organization = useOrganizationContext()
   const [open, setOpen] = useState(false)
-  const identity = sessionIdentity(session_type)
+  const identity = sessionIdentity(organization.data?.caller)
```
khaledosman, line 291: The trigger's `aria-label="Account"` (line 292) is a static override, so it wins over the children and a screen reader still hears only "Account". The identity this PR exists to draw never reaches the accessible name, and on the collapsed rail that name is the only place it could appear at all. `AppearanceControl` already sets the precedent for folding the visible state into the name:

```tsx
aria-label={`Account: ${identity.name}`}
```

The test queries would move to `{ name: /^Account:/ }`, and #832's fix would then hold for a non-sighted operator too.

khaledosman (approved):
Two small points inline; neither blocks. Verified locally on the head revision: `ruff check src tests scripts`, `make openapi-check` and `make postman-check` all pass, and all three `_to_context` call sites are caller-scoped, so `caller` never carries another identity. `TelemetryIdentity` enumerates its fields rather than forwarding the context, so the new `email` does not reach the tracker seam by default; worth keeping that way.

Heads-up outside the diff: this conflicts with #870 in three places in `web/src/app/nav/AccountMenu.tsx` (`git merge-tree` on the two heads). #870 removes `management_url` and `EntitlementGate` from this file, which this PR still uses, so whichever lands second needs the rebase to drop them rather than take both sides.

<sub>Reviewed by Claude Opus 5 via Claude Code.</sub>

PR #870, opened 2026-08-31
feat(dashboard): publish the deployment's legal page addresses

njbrake:
## Description

The account menu has two rows for a deployment's legal documents, Terms of service and Data & Privacy, and neither could ever point anywhere. Terms built its address out of a field only a hybrid gateway sets, and a hybrid gateway shows the landing page instead of this menu; it also sat behind a licence capability nothing grants. Data & Privacy was hardcoded as a greyed-out row. So an operator whose dashboard sits beside a site holding their privacy notice and their terms had no way to link either one, and users of that dashboard had no way in.

This adds two settings, `terms_url` and `privacy_url`, configured the way `docs_url` already is and published by `GET /v1/bootstrap` in every mode. Each row now links to whichever document the deployment has published, independently. Nothing changes for a deployment that sets neither: Terms of service stays absent and Data & Privacy stays disabled with the reason it has always carried, which is honest for a gateway that stores its data locally and reports nothing outward. The three link fields share one validator, so a scheme that is not http(s) is a startup error rather than an anchor that goes nowhere.

## PR Type

- [x] New Feature
- [x] Bug Fix

## Relevant issues

Unblocks mozilla-ai/otari-ai#1945, which closes on the otari-ai side once the pin moves and the hosted deployment sets both values.

## Checklist

`make lint`, `make typecheck`, `tests/unit` (2614 passed), `pnpm --dir web` lint, typecheck and test (1729 passed), `make openapi-check` and `make postman-check` all pass. The Python integration suite and Playwright were not run locally, no Docker in this environment; two e2e bootstrap payloads are asserted with `toEqual` and were updated, so CI is the confirmation on those.

## AI Usage

**AI Model/Tool used:**

Written by Claude Opus 5 through back-and-forth with @njbrake, who chose the approach: publish both addresses rather than privacy alone, since the terms row was broken for the same reason.

## Summary

- Added optional `terms_url` and `privacy_url` settings.
- Published both URLs through `GET /v1/bootstrap` in all deployment modes.
- Updated the account menu to show each legal page independently.
- Kept Terms of Service hidden and Data & Privacy disabled when URLs are not configured.
- Added shared HTTP(S)-only validation for documentation and legal-page URLs.
- Updated signup wording, documentation, API schemas, fixtures, and tests.

### DIFF
--- a/src/gateway/core/config.py
+++ b/src/gateway/core/config.py
@@ -14,7 +14,7 @@ import yaml
 from any_llm import AnyLLM, LLMProvider
 from any_llm.exceptions import AnyLLMError
 from dotenv import load_dotenv
-from pydantic import BaseModel, Field, PrivateAttr, field_validator, model_validator
+from pydantic import BaseModel, Field, PrivateAttr, ValidationInfo, field_validator, model_validator
 from pydantic_settings import BaseSettings, SettingsConfigDict
 
 from gateway.core.addresses import normalized_address
@@ -452,6 +452,27 @@ class GatewayConfig(BaseSettings):
             "right default for a self-hosted deployment. Set it to retarget those links at "
             "a product documentation site instead; the bundled guide stays reachable at "
             "/#/docs either way."
+        ),
+    )
+    terms_url: str | None = Field(
+        default=None,
+        description=(
+            "Where this deployment's terms of service live, as an absolute http(s) URL "
+            "(e.g. 'https://otari.ai/terms'). Unset, the account menu shows no terms row: a "
+            "deployment nobody wrote terms for has none to point at, which is the ordinary "
+            "case for a self-hosted gateway. Set it and the row appears. A link target an "
+            "operator configured, held to the same bar as docs_url."
+        ),
+    )
+    privacy_url: str | None = Field(
+        default=None,
+        description=(
+            "Where this deployment's privacy notice lives, as an absolute http(s) URL "
+            "(e.g. 'https://otari.ai/privacy'). Unset, the account menu's Data & Privacy row "
+            "stays disabled with the reason it carries today, which is honest for a gateway "
+            "that stores its data locally and reports nothing outward. Set it and the row "
+            "becomes a link to the notice. A link target an operator configured, held to the "
+            "same bar as docs_url."
         ),
     )
     data_plane_url: str | None = Field(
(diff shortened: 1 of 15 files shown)


src/gateway/core/config.py
```diff
@@ -1741,23 +1762,28 @@ def _validate_stream_missing_usage_policy(cls, value: str) -> str:
...
+        consumer that builds a request from it. ``data_plane_url`` is held to a
+        stricter bar for exactly that reason and keeps its own.
         """
         normalized = (value or "").strip()
         if not normalized:
             return None
         parsed = urlsplit(normalized)
         if parsed.scheme not in {"http", "https"} or not parsed.netloc:
-            msg = f"docs_url must be an absolute http(s) URL, got '{value}'"
+            msg = f"{info.field_name} must be an absolute http(s) URL, got '{value}'"
```
khaledosman, line 1792: These two land in the same unauthenticated `GET /v1/bootstrap` payload that made `_validate_data_plane_url` refuse userinfo (`config.py:1841`, whose comment gives exactly this reason), but this validator only checks scheme and netloc, so `https://token:secret@example.com/terms` loads and is then served to any browser that asks. `docs/configuration.md:252` documents the gap ("do not put a credential in either") instead of closing it.

Since this PR is already consolidating the three link fields onto one validator, lift the same check in:

```python
if parsed.username is not None or parsed.password is not None:
    msg = f"{info.field_name} must carry no username or password"
    raise ValueError(msg)
```

That also picks up `docs_url`, which has had the same hole. The `configuration.md` sentence then becomes "credentials are refused", matching the data-plane section three headings down.

khaledosman (approved):
One hardening point inline. Verified locally on the head revision: `ruff check src tests scripts`, `make openapi-check` and `make postman-check` all pass (the collection is built from request bodies and operation descriptions, so a response-only schema change leaves it untouched, but confirmed rather than assumed).

Heads-up outside the diff: this conflicts with #869 in three places in `web/src/app/nav/AccountMenu.tsx` (`git merge-tree` on the two heads). #869 keeps `management_url` and `EntitlementGate`, which this PR removes, so whichever lands second needs the rebase to drop them rather than take both sides.

<sub>Reviewed by Claude Opus 5 via Claude Code.</sub>

PR #872, opened 2026-08-31
fix(dashboard): open Model pricing to organization admins

khaledosman:
## Description

The [11 Aug roles matrix](https://github.com/mozilla-ai/otari-ai/issues/1946) puts **Model pricing** at Hidden for members and **Edit for admins**. The nav row was `operatorOnly: "refused"`, so a hosted organization's admin could not reach the page that hosts their own rate overrides, even though the server already allowed them to write those overrides.

The page is two halves with two different owners, so the gating splits rather than the destination opening wholesale:

| Section | Server gate | Who sees it now |
| --- | --- | --- |
| Catalog policy banner (`GET /v1/settings`) | `require_deployment_operator` | operator |
| Default pricing catalog / refresh flow (`/v1/pricing/refresh*`) | `require_deployment_operator` | operator |
| Model prices table (`GET /v1/pricing`) | `verify_catalog_reader` (any session) | operator + admin |
| Rate overrides (`/v1/organizations/me/pricing`) | owner-or-admin, in `organization_pricing_service` | operator + admin |

The two operator-only sections are withheld from anyone else rather than fired into a 403 banner, which is how Models has withheld its operator-only reads since #867. The sentences pointing at the per-model editor on Models go with them, since setting a catalog rate is `POST /v1/pricing`.

**No new read is exposed.** Both halves the page keeps were already reachable by URL for any session; this only stops rendering the two that refuse. The row stays out of a member's sidebar because the organization rail opens only to a caller who manages the organization.

Second half of otari-ai#1943, the org-scoped Spend & budgets surface, follows in its own PR.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Refs mozilla-ai/otari-ai#1943, mozilla-ai/otari-ai#1946

## Checklist

Dashboard-only change, so the suites that matter here are the web ones: `pnpm --dir web run lint`, `pnpm --dir web run typecheck`, and `pnpm --dir web test` (107 files, 1745 tests, all passing). `make lint` was run and is clean. No API contract change, so the OpenAPI spec and the Postman collection are untouched.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

Implemented from the roles-matrix audit in otari-ai#1946. Reviewed by the author before opening.

### DIFF
--- a/web/src/app/AppShell.test.tsx
+++ b/web/src/app/AppShell.test.tsx
@@ -246,7 +246,7 @@ describe("AppShell responsive layout", () => {
     await renderShell()
 
     const overview = screen.getByRole("link", { name: "Overview" })
-    // Awaited: Providers is one of the eight rows gated `operatorOnly`, so it
+    // Awaited: Providers is one of the four rows gated `operatorOnly`, so it
     // arrives with the membership context rather than with the first paint.
     const providers = await screen.findByRole("link", { name: "Providers" })
     expect(overview).toHaveAttribute("aria-current", "page")
@@ -533,11 +533,11 @@ describe("AppShell surface gating", () => {
     // Same reasoning as above, for the other context: the organization rail is
     // its own registry, and nothing else compares it against a full list.
     await renderShell(bootstrap(), { url: "/organization/members" })
-    // Awaited, not assumed: four of these rows declare `operatorOnly`, so none
+    // Awaited, not assumed: three of these rows declare `operatorOnly`, so none
     // of them exists until `GET /v1/organizations/me` answers. Taking the
     // snapshot without waiting is a race that passes on a fast machine and fails
-    // on CI, which is what it did. One await covers all four, because the caller
-    // axis is one read and they appear in the same paint.
+    // on CI, which is what it did. One await covers all three, because the
+    // caller axis is one read and they appear in the same paint.
     await within(
       screen.getByRole("navigation", { name: "Sidebar" }),
     ).findByRole("link", { name: "Accounts" })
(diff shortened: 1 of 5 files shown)


njbrake (approved):
Rebased onto main and pushed. #866 removed `operatorOnly` from `/keys` after this branch forked, so the row count in `AppShell.test.tsx` is now four, and the shrunk `registry.test.ts` array plus the fixture line needed reformatting; without that, Biome would have failed CI. Also pinned the admin case to `role: "admin"` rather than the fixture's default owner, since the matrix row is about the admin.

Gating verified against the routes rather than the description: `GET /v1/settings` and the three `/v1/pricing/refresh` routes are `require_deployment_operator`, `GET /v1/pricing` is `verify_catalog_reader` which `deps.py` returns for any dashboard session, `POST /v1/pricing` is operator-only (so dropping the Models pointers is right), and `organization_pricing_service` gates the override writes on organization management. The split matches what `nav/types.ts` already documents for a partly deployment-wide page, and the page-level shape matches `ModelsPage` since #867.

One thing left alone: `isDeploymentOperator` returns false when the context read errors, so an operator whose `GET /v1/organizations/me` flakes loses the catalog controls with no explanation, while the nav's `"refused"` semantics deliberately fail open on the same failure. Pre-existing on `ModelsPage`, `RoutingPage`, `ToolsGuardrailsPage`, and `WorkspacesPage`; worth one shared decision later, not here.

Full web suite green on the rebased head (1752 passing), lint and typecheck clean.

khaledosman:
The `isDeploymentOperator` point is filed as #876, with the five pages that read the axis this way plus #875's `BudgetsPage`, which fails in the opposite direction and is the reason it is worth one decision rather than a per-page choice. Nothing changed here for it.

The org-scoped budget surface is #875, now retargeted to `main`.

(via Claude Opus 5 in Claude Code)

PR #874, opened 2026-08-31
docs: prune the comments #870 left around the legal URLs

njbrake:
## Description

Follow-up to #870. That PR added a lot of comment for the amount of code it added, and most of it was not carrying anything the code did not already say. This trims it back and sharpens the rule that was supposed to prevent it.

Deletes the seven-line block above the Terms of service row: it explained the `management_url` and `legal.terms` gating that #870 replaced, which is history about code that no longer exists, already stated in that PR's commit message and body, wrapped around a ternary that reads plainly. Shortens the header note, the bootstrap fixture, both e2e payload comments, three test comments, and one validator docstring paragraph. Kept in every case is the fact the code cannot state: that Data & Privacy stays disabled rather than vanishing, and that `data_plane_url` keeps a stricter validator.

`AGENTS.md` now says no restating the code, no narrating the change, no recording what the code used to do, and leave the comments around a change shorter than you found them. The previous wording, "add a comment only where the logic is not obvious", caught none of it, because narration and history can both claim to be intent.

36 of the 124 dashboard lines #870 added were comment. No behavior changes.

## PR Type

- [x] Refactor
- [x] Documentation

## Relevant issues

Follow-up to #870. No issue.

## Checklist

Comments only, so no test was added or changed and the contract is untouched. Run locally: `pnpm --dir web run lint`, `run typecheck`, `vitest run src/app/nav/AccountMenu.test.tsx` (15 passed), `uv run ruff check src tests scripts`, `pytest tests/unit/test_deployment_bootstrap.py` (71 passed).

## AI Usage

**AI Model/Tool used:**

Written by Claude Opus 5 through back-and-forth with @njbrake, who asked for the trim after reading #870 and pointed at agent-of-empires' `AGENTS.md` as the standard to hold it to. Its two clauses that otari's rule lacked, no preserved implementation history and leave the comments shorter than you found them, are what this adopts.

## Summary

- Removed redundant comments about legal URL handling across the application, tests, fixtures, and end-to-end tests.
- Kept comments that explain behavior not clear from the code, including the disabled Data & Privacy row and stricter `data_plane_url` validation.
- Tightened `AGENTS.md` guidance to favor concise comments and avoid repeating code or historical context.
- Preserved all application behavior and test logic.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -102,7 +102,7 @@ The per-request flow (auth → budget → dispatch → reconciliation) spans sev
 
 ## Repository Conventions
 - Prefer minimal, targeted edits over broad refactors, and match the import order and typing style of the file you are in (`TYPE_CHECKING` for type-only imports where it helps, as in `routes/_helpers.py`).
-- Add a comment only where the logic is not obvious; keep docstrings concise and meaningful on public functions and classes.
+- Add a comment only where the logic is not obvious; keep docstrings concise and meaningful on public functions and classes. Do not restate the code, narrate the change, or record what the code used to do: the commit message is where that belongs. Leave the comments around a change shorter than you found them.
 - Preserve security-relevant behavior: header parsing, auth checks, and the error-detail boundary. Do not leak internals in public error responses, and never log secrets, tokens, or raw API keys (the one-time bootstrap key print is the deliberate exception).
 - Keep test additions next to the behavior they cover: unit for pure logic, integration for route or database behavior.
 - CI runs Python 3.14 (`.github/workflows/otari-tests.yml`), matching the Docker image; the package still supports 3.13+ (`requires-python = ">=3.13"`).
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -102,7 +102,7 @@ The per-request flow (auth → budget → dispatch → reconciliation) spans sev
 
 ## Repository Conventions
 - Prefer minimal, targeted edits over broad refactors, and match the import order and typing style of the file you are in (`TYPE_CHECKING` for type-only imports where it helps, as in `routes/_helpers.py`).
-- Add a comment only where the logic is not obvious; keep docstrings concise and meaningful on public functions and classes. Do not restate the code, narrate the change, or record what the code used to do: the commit message is where that belongs. Leave the comments around a change shorter than you found them.
+- Add a comment only where the logic is not obvious; keep docstrings concise and meaningful on public functions and classes. Do not restate the code, narrate the change, or record what the code used to do: the commit message is where that belongs. Leave the comments around a change shorter than you found them: prune narration, repeated rationale, and implementation history as you touch them.
 - Preserve security-relevant behavior: header parsing, auth checks, and the error-detail boundary. Do not leak internals in public error responses, and never log secrets, tokens, or raw API keys (the one-time bootstrap key print is the deliberate exception).
 - Keep test additions next to the behavior they cover: unit for pure logic, integration for route or database behavior.
 - CI runs Python 3.14 (`.github/workflows/otari-tests.yml`), matching the Docker image; the package still supports 3.13+ (`requires-python = ">=3.13"`).
--- a/src/gateway/core/config.py
+++ b/src/gateway/core/config.py
@@ -1773,10 +1773,8 @@ class GatewayConfig(BaseSettings):
         for the same reason ``platform.management_url`` is: a typo should be a
         startup error, not a Documentation link that silently goes nowhere.
 
-        One validator for the three because they are one kind of value: an
-        address a person follows out of the dashboard, with no downstream
-        consumer that builds a request from it. ``data_plane_url`` is held to a
-        stricter bar for exactly that reason and keeps its own.
+        ``data_plane_url`` keeps its own, stricter validator: a client builds a
+        request from it rather than following it.
 
         Userinfo is the one refusal the three share with ``data_plane_url``, and
         they share it for its reason rather than theirs: ``GET /v1/bootstrap``
--- a/src/gateway/core/config.py
+++ b/src/gateway/core/config.py
@@ -1776,7 +1776,7 @@ class GatewayConfig(BaseSettings):
         ``data_plane_url`` keeps its own, stricter validator: a client builds a
         request from it rather than following it.
 
-        Userinfo is the one refusal the three share with ``data_plane_url``, and
+        Userinfo is the one refusal these three fields share with ``data_plane_url``, and
         they share it for its reason rather than theirs: ``GET /v1/bootstrap``
         is unauthenticated, so a credential written into any of these would
         reach every browser that asked, which no redaction in the operator-gated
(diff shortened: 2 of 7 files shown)


src/gateway/core/config.py
```diff
@@ -1773,10 +1773,8 @@ def _validate_link_url(cls, value: str | None, info: ValidationInfo) -> str | No
         for the same reason ``platform.management_url`` is: a typo should be a
         startup error, not a Documentation link that silently goes nowhere.
 
-        One validator for the three because they are one kind of value: an
-        address a person follows out of the dashboard, with no downstream
-        consumer that builds a request from it. ``data_plane_url`` is held to a
-        stricter bar for exactly that reason and keeps its own.
+        ``data_plane_url`` keeps its own, stricter validator: a client builds a
+        request from it rather than following it.
```
khaledosman, line 1777: The deleted paragraph was what introduced "the three", so line 1779's `Userinfo is the one refusal the three share` now has no antecedent in the docstring; the reader has to jump back to the decorator to resolve it. Change 1779 to `the one refusal these three fields share`.

Worth keeping the grouping rationale too: *why* one validator covers three fields is non-obvious intent rather than history, and it is now only implied by the `data_plane_url` contrast.

AGENTS.md
```diff
@@ -102,7 +102,7 @@ The per-request flow (auth → budget → dispatch → reconciliation) spans sev
 
 ## Repository Conventions
 - Prefer minimal, targeted edits over broad refactors, and match the import order and typing style of the file you are in (`TYPE_CHECKING` for type-only imports where it helps, as in `routes/_helpers.py`).
-- Add a comment only where the logic is not obvious; keep docstrings concise and meaningful on public functions and classes.
+- Add a comment only where the logic is not obvious; keep docstrings concise and meaningful on public functions and classes. Do not restate the code, narrate the change, or record what the code used to do: the commit message is where that belongs. Leave the comments around a change shorter than you found them.
```
khaledosman: `Leave the comments around a change shorter than you found them` contradicts the same bullet's first clause. A change that genuinely introduces non-obvious logic must add a comment, which cannot leave the region shorter, so the rule as written makes the correct action a violation, and an agent optimizing for it will delete a load-bearing comment to pay for a new one.

Make it directional rather than absolute: `never longer just to record the change; if the comments around it can be shorter, leave them shorter.`

khaledosman (commented):
Comments-only, no behavior change. Five notes: one dangling reference the trim created, one rule wording that now contradicts itself, and three comments that either overstate a fact or survive as pure restatement under the rule this PR is adding.

🤖 Written by Claude Opus 5 (1M context) via the Claude Code CLI as an agent, at @khaledosman's request.

web/src/tests/fixtures.ts
```diff
@@ -129,9 +129,8 @@ export function bootstrap(
     // Null, because a fixture describes a deployment whose documentation is the
     // bundled guide; the tests about an operator-configured docs site set it.
     docs_url: null,
-    // Null both, because a self-hosted gateway publishes no legal pages of its
-    // own: Terms of service is then absent from the account menu and Data &
-    // Privacy is the disabled row. The account-menu tests set them.
+    // A self-hosted gateway publishes no legal pages. The account-menu tests
```
khaledosman: `A self-hosted gateway publishes no legal pages` asserts as a general fact what is only this fixture's value. `_validate_link_url` accepts `terms_url`/`privacy_url` on any deployment, and `parity.hybrid.spec.ts:56` calls these fields deployment-wide. Dropping the original's `because` turned a justification into a false claim.

Suggest: `// Unset, like docs_url: this fixture's deployment publishes none. The account-menu tests set them.`

web/src/app/nav/AccountMenu.test.tsx
```diff
@@ -100,8 +100,7 @@ describe("AccountMenu", () => {
     await openMenu()
 
     expect(screen.queryByRole("link", { name: "Terms of service" })).toBeNull()
-    // Present but inert, which is the row a gateway with no notice to link
-    // still owes: the surface is coming, and omitting it reads as never.
+    // Present but inert rather than absent.
```
khaledosman: `Present but inert rather than absent` restates the two assertions directly under it (`findByRole("button")` plus `toBeDisabled`), which is what the new AGENTS.md clause forbids. The removed half was the only part carrying a reason, and that reason now lives in `AccountMenu.tsx`. Delete the comment rather than keeping the restating half.

web/src/app/nav/AccountMenu.tsx
```diff
@@ -41,13 +41,10 @@ import {
...
 // since otari#647, and the identity holds a password of its own since otari#649,
 // so the row this menu carried disabled from the start now has a destination.
-// The two legal rows are the deployment's own documents, and each appears only
-// where there is one to point at: `terms_url` and `privacy_url`, the addresses
-// an operator configures and `GET /v1/bootstrap` publishes beside `docs_url`.
-// Terms of service is absent without one, because a gateway nobody wrote terms
-// for has none to show. Data & Privacy falls back to the disabled row instead
-// of vanishing, because the settings surface it will become is coming and a
-// menu that silently lacks it reads as a menu that never will.
+// The legal rows link `terms_url` and `privacy_url` where the deployment set
```
khaledosman: `The legal rows link terms_url and privacy_url where the deployment set them` restates the two ternaries at lines 363-377. Only the second sentence, that Data & Privacy stays disabled rather than vanishing, says something the code cannot. Drop the first sentence and start at `Data & Privacy stays a disabled row when unset...`.

src/gateway/core/config.py
njbrake: Fixed in 8722760: `these three fields`. Kept the grouping rationale where the replacement sentence already carries it, that these are followed and `data_plane_url` is built from.

_Note: drafted by Claude Opus 5 via back-and-forth with @njbrake._

AGENTS.md
njbrake: Right, and the absolute form was my paraphrase. Taken agent-of-empires' own wording, which is directional: `prune narration, repeated rationale, and implementation history as you touch them`. Same fix applied to otari-ai's AGENTS.md, which had the same bullet.

_Note: drafted by Claude Opus 5 via back-and-forth with @njbrake._

web/src/tests/fixtures.ts
njbrake: Correct, dropping the `because` turned it into a claim the validator contradicts. Taken your wording.

_Note: drafted by Claude Opus 5 via back-and-forth with @njbrake._

web/src/app/nav/AccountMenu.test.tsx
njbrake: Deleted.

_Note: drafted by Claude Opus 5 via back-and-forth with @njbrake._

web/src/app/nav/AccountMenu.tsx
njbrake: Deleted the first sentence; the comment now opens at Data & Privacy.

_Note: drafted by Claude Opus 5 via back-and-forth with @njbrake._

PR #882, opened 2026-08-31
feat(web-search): call Tavily and Brave in-process, and serve them to a data plane

njbrake:
## Description

Until now, using a paid search API with Otari meant running a second container next to the gateway whose only job was to translate between Otari and that provider, and that container held the provider's key. There was one for Tavily and one for Brave. This change moves the translation into Otari itself: name the provider and give Otari the key, and it calls Tavily or Brave directly. The two containers, their compose profiles and their documentation are gone.

That works when one process does everything. It does not work when the machine answering user traffic is a different machine from the one holding the search key, which is how a hosted deployment is arranged. For that case Otari also gains a small search endpoint of its own, so the process that holds the key runs the search and hands back the results. Nothing it returns is a credential.

That endpoint spends the deployment's own search allowance and is reachable from the internet, so it is served only when the operator has configured a shared token for it, and it refuses anyone who does not present that token. A deployment that configures no token does not serve the endpoint at all, rather than serving an open one.

### The two entry points

`gateway/services/web_search_providers.py` is the one translation module, reached two ways:

- **In-process.** `WebSearchBackend` calls it directly when `web_search_provider` and `web_search_provider_api_key` are set, so `base_url` becomes optional and a configured provider wins over a configured URL.
- **Over HTTP.** `GET /v1/web-search/search`, in the SearXNG shape `WebSearchBackend` already speaks, mounted only when a provider *and* `web_search_backend_token` are both configured, and gated on that token as `X-Gateway-Token` under `compare_digest`. An upstream failure becomes a 502 whose detail names the provider and nothing else; the provider and the upstream status are logged, and the body never is, since a provider's error text can echo back what was sent to it.

### Notes for review

- `web_search_configured()` replaces four copies of `config.web_search_url or otari_env("WEB_SEARCH_URL")`, so the request path, the per-workspace page, the tool catalog and pricing init cannot disagree about whether this deployment can search.
- The gate in `prepare_gateway_tools` moves from "is there a URL" to "can this deployment search", which is what lets a provider-only deployment past it. The three `_build_web_search_backend` call sites lose their `web_search_url is not None` asserts accordingly.
- `web_search_provider_api_key` and `web_search_backend_token` are on `_DELIBERATELY_OMITTED` in `routes/settings.py`; `web_search_provider` is shown.
- The route is mounted only where a provider and a token are both configured, so `scripts/generate_openapi.py` sets placeholder values for both: the spec has to describe what the app can serve, not what a bare default config reaches. Both generated artifacts are regenerated, and `sdk-endpoints.txt` classifies the route as gateway-to-gateway, since its caller holds a deployment token no SDK holds.
- Bounded and observable like the other provider-calling scaffolds: `max_results` is capped at `MAX_RESULTS_CAP`, the search registers with `track_request`, and the upstream failure behind a 502 is logged with the provider and the status.
- The query rides in the request target, because that is the SearXNG shape `WebSearchBackend` already speaks. A control plane that must not retain queries turns access logging off or redacts the path; the module docstring says so.
- `demo/web-search/`'s `--brave` path now configures the provider instead of bringing up an adapter container, so it starts one fewer service.

Paired with mozilla-ai/otari-ai#1914, which serves this endpoint from the hosted control plane and migrates the per-workspace policy rows.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Refs mozilla-ai/otari-ai#1914

## Checklist

`make lint`, `mypy`, `make openapi-check` and `make postman-check` are clean. `tests/unit` is 2654 passed, 8 skipped. `pnpm --dir web run lint` is clean and the Tools page suite is 21 passed. The integration suite needs Docker, which this environment has none of, so it has not been run locally.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

Written by Claude Opus 5 through back-and-forth with @njbrake. The decisions, including where the credential sits and that the hosted endpoint answers the default gateway alone, are his; the prose and the code are the model's.

## Summary

Otari now supports direct Tavily and Brave web search without separate adapter containers. It also provides an authenticated `/v1/web-search/search` endpoint for deployments where search keys remain in another process.

The change removes the adapter services, compose profiles, and related documentation. It adds centralized search configuration, safer settings display, updated deployment guidance, and a simpler Brave demo setup.

## Technical notes

- Provider requests return normalized SearXNG-shaped results.
- The endpoint uses `X-Gateway-Token` authentication.
- Upstream failures return HTTP 502 without exposing provider response details.
- Unit, lint, type, OpenAPI, Postman, and web checks pass. Integration tests were not run because Docker was unavailable.


src/gateway/api/routes/_pipeline.py
```diff
@@ -2539,7 +2539,9 @@ async def prepare_gateway_tools(
                 # already trusts this token with for resolve). Never leak this
                 # high-privilege credential to a bundled SearXNG or a third-party
                 # adapter that an operator happened to point GATEWAY_WEB_SEARCH_URL at.
-                if url_targets_platform(web_search_url, ctx.config.platform.get("base_url")):
+                if web_search_url is not None and url_targets_platform(
+                    web_search_url, ctx.config.platform.get("base_url")
+                ):
                     web_search_auth_token = ctx.config.platform_token
```
khaledosman, line 2545: The two ends of the new endpoint don't share a secret. The route accepts only `web_search_backend_token`, but this is the only place anything sends `X-Gateway-Token`, and it sends `platform_token` — in the hybrid branch only, and only when the URL sits under the platform base URL. So a non-hybrid gateway pointed at `{control-plane}/v1/web-search` sends no token at all and gets 401 unconditionally, and a hybrid one works only if the operator set the control plane's `web_search_backend_token` to that gateway's platform token, which neither the field description nor `docs/tools.md` says. That also means rotating the platform token silently kills web search deployment-wide.

Reusing this path isn't the fix — the `url_targets_platform` guard exists precisely so a high-privilege credential isn't handed to a bundled SearXNG. Either give the client side its own forwarded token field, or state the platform-token coupling explicitly in the `web_search_backend_token` description and in `docs/tools.md`, and drop the route docstring's claim that pointing `web_search_url` at the control plane "is the whole of the client-side configuration".

src/gateway/services/web_search_providers.py
```diff
@@ -0,0 +1,238 @@
...
+            params=params,
+            headers=headers,
+            timeout=timeout_s,
+        )
+    except httpx.HTTPError as exc:
+        msg = f"{provider} search could not be reached: {exc}"
+        raise WebSearchProviderError(msg) from exc
+
+    if response.status_code >= httpx.codes.BAD_REQUEST:
+        msg = f"{provider} search returned HTTP {response.status_code}: {response.text[:_ERROR_BODY_CHARS]}"
```
khaledosman: `.github/instructions/security-review.instructions.md` is explicit: never log raw provider bodies. This message is what `_pipeline.py` hands to `logger.error` on the in-process path (`Web search backend unreachable ...: %s`), so 500 bytes of an arbitrary upstream body land in the deployment's log, and if a provider ever echoes the presented key the credential goes with it — `test_web_search_backend_route.py:148` models exactly that body (`invalid api key tvly-secret`) and only asserts it stays out of the *response*.

Keep the status, drop the body, or scrub `api_key` out of the slice before it reaches the message.

src/gateway/api/routes/web_search_backend.py
```diff
@@ -0,0 +1,126 @@
...
+    }
+    try:
+        results = await provider_search(
+            provider=provider,
+            api_key=api_key,
+            query=q,
+            options={key: value for key, value in options.items() if value is not None},
+            client=get_search_client(),
+        )
+    except WebSearchProviderError as exc:
```
khaledosman, line 156: This comment and the module docstring both say the provider's status and body stay in the log, but nothing on this path logs: the `WebSearchProviderError` only becomes the `__cause__` of an `HTTPException` that FastAPI renders and discards. Add the `logger.error` the sibling route already uses (`routes/search.py:424`) — status and provider name, not the body, per the note on `web_search_providers.py:227` — or correct the two comments. As it stands an operator debugging a 502 here has nothing at all.

src/gateway/core/config.py
```diff
@@ -1753,6 +1807,19 @@ def validate_search_tools(self) -> None:
             if provider in SEARCH_PROVIDERS_REQUIRING_API_BASE and not entry.get("api_base"):
                 validate_search_tool_transport(name, self.web_search_url, entry.get("api_key"))
 
+    @field_validator("web_search_provider")
+    @classmethod
+    def _validate_web_search_provider(cls, value: str | None) -> str | None:
```
khaledosman, line 1857: Nothing validates the pair at load. `web_search_provider` with no `web_search_provider_api_key` silently falls back to `web_search_url` (or reports web search unconfigured), so a typo'd `OTARI_WEB_SEARCH_PROVIDER_API_KEY` degrades instead of failing; `web_search_backend_token` with no provider silently doesn't mount the route it was set for. `src/gateway/AGENTS.md` ("Validate a new security or routing setting at config load") and the security-review rung on fail-closed configuration both want these rejected at startup, the way `SEARCH_PROVIDERS_REQUIRING_API_KEY` already is. A `@model_validator(mode="after")` beside the ones at 249/292.

src/gateway/api/routes/web_search_backend.py
```diff
@@ -0,0 +1,126 @@
...
+
+def _authorize(config: GatewayConfig, presented: str | None) -> None:
+    """Refuse anything but the deployment's own gateway.
+
+    Compared as digests through :func:`secrets.compare_digest`, so neither the
+    length nor the leading characters of the configured token leak through the
+    comparison's timing.
+    """
+    expected = config.web_search_backend_token
+    if not expected or not presented or not compare_digest(presented, expected):
```
khaledosman: `compare_digest` on `str` raises `TypeError: comparing strings with non-ASCII characters is not supported`, and Starlette latin-1-decodes headers, so `X-Gateway-Token: ü` is a 500 on an internet-reachable endpoint rather than a 401. Compare `presented.encode()` against `expected.encode()`.

src/gateway/services/web_search_providers.py
```diff
@@ -0,0 +1,238 @@
...
+    timeout_s: float,
+) -> list[dict[str, Any]]:
+    """Brave's web search, which returns snippets only.
+
+    ``extracted_content`` is deliberately left unset, so the caller fetches and
+    extracts each page itself exactly as it does behind SearXNG.
+    """
+    count = _BRAVE_DEFAULT_COUNT
+    max_results = options.get("max_results")
+    if isinstance(max_results, int) and max_results > 0:
```
khaledosman: Both providers ignore the backend's resolved ceiling. `WebSearchBackend._max_results` is applied only as a post-hoc slice, so a deployment that raises `web_search_max_results` to 20 still gets 10 hits from Brave and Tavily's own default from Tavily — and with domain filters on, fewer. The adapter containers had the same blind spot, but in-process the ceiling is right there: pass it through as the default `count` / `max_results` and let `provider_options` override it.

src/gateway/services/web_search_providers.py
```diff
@@ -0,0 +1,238 @@
...
+# forwarding it wholesale would let a workspace set Tavily request fields the
+# deployment never chose.
+_TAVILY_OPTION_KEYS = ("max_results", "search_depth", "topic", "time_range", "include_answer")
+
+
+class WebSearchProviderError(RuntimeError):
+    """The search provider could not be reached or returned malformed data."""
+
+
+def provider_configured(provider: str | None, api_key: str | None) -> bool:
```
khaledosman: `provider_configured` has no caller outside `test_web_search_providers.py:80` — `main.py:131` and `_tools.py:378` both go through `GatewayConfig.web_search_provider_configured()`. Delete it, or have the config method delegate to it: this is the stricter of the two, since it also checks membership in `WEB_SEARCH_PROVIDERS`.

src/gateway/services/web_search_backend.py
```diff
@@ -321,16 +336,21 @@ async def _search_tool(self, arguments: dict[str, Any]) -> str:
             query = (arguments.get("query") or "").strip()
             span.set_attribute("web_search.query", query)
             span.set_attribute("web_search.provider", ",".join(self._engines))
```
khaledosman: This still reports the SearXNG engine list as `web_search.provider` on a search that never touched SearXNG. Set it to `self._provider` on the provider path; as it is, `backend_provider` reads as a second attribute competing to answer the same question.

khaledosman (approved):
Findings inline. Two that have no diff line to hang on:

- `pyproject.toml:175-179` — the mypy `exclude` and its comment still describe `scripts/web-search-*-adapter/` and point at `tests/unit/test_*_adapter.py`, all deleted here. Drop both.
- `src/gateway/AGENTS.md` "Authentication and authority" enumerates four authorization shapes; `/v1/web-search/search` introduces a fifth (a shared operator token, no identity resolved, no dashboard cookie). Worth one line there, since that roster is what a reviewer checks a new route against.

---

njbrake:
All review comments are addressed across `202dc8cc`, `3e2e3fa1`, `fad27b8f` and `6aa77564`.

| Comment | Resolution |
|---|---|
| Brave drops `time_range` and `page_age`/`age` | Mapped onto `freshness`; `published_date` carried back, `page_age` preferred |
| Provider response body in the exception message | Provider and status only |
| `compare_digest` raises on a non-ASCII header | Hashed before comparing |
| Dead `provider_configured` | Removed |
| Platform-token coupling undocumented | Stated in `docs/tools.md`, the `web_search_backend_token` description and the route docstring, including that rotating the platform token stops search |
| Config pair unvalidated at load | A half-configured provider warns, and so does a backend token with no provider. Warned rather than refused: web search is one optional tool, and a compose file can default a provider name without defaulting a secret |
| `published_date` absent from the response model | Declared, so it survives the hop |
| `max_results` unbounded on the route | Capped at `MAX_RESULTS_CAP`, the ceiling `/v1/search` declares |
| No `track_request` | Registers before provider dispatch |
| Nothing logs on the 502 path | `logger.error` with the provider and the upstream status |
| Providers ignore the resolved ceiling | Passed through as the default `max_results`; an explicit `provider_options` value still wins |
| `web_search.provider` reports the SearXNG engine list | Names what served the search; `backend_provider` dropped |
| Route missing from the generated artifacts | The generator configures placeholder values, and the spec, Postman collection, SDK manifest and dashboard client are regenerated. The route now declares `GatewayTokenAuth` rather than inheriting the API-key schemes |

One is resolved differently than asked. `q` stays on the GET, because the route exists to speak the SearXNG shape `WebSearchBackend` already talks, and a POST body would make the client special-case its own control plane. The exposure is stated in the module docstring: a deployment whose access log records request targets records search queries, and a control plane that must not keep them turns that log off or redacts the path. Say if you want the POST instead.

PR #883, opened 2026-08-31
refactor(dashboard): gate the pricing warning on the operator signal

### LINKED ISSUE #834: refactor(dashboard): gate the pricing warning on the operator signal instead of a refused settings query
`PricingWarning` is only ever visible to a deployment operator, but nothing gates it on that. `AppShell.tsx:665` mounts it for anyone who reaches the shell, and what actually hides it is that `useSettings()` → `GET /v1/settings` is gated on `require_deployment_operator` (`src/gateway/api/routes/settings.py:412`): a non-operator's query 403s, `settings.data` is undefined, `needsPricing` is false, and the banner returns `null`. An organization owner or admin sees nothing either, since that dependency is deployment-wide and not a tenancy role.

The visible behavior is right. Two things follow from the gate being implicit:

- Every non-operator page load fires a `GET /v1/settings` that 403s, on a 60s `staleTime`, to feed a banner that cannot render for them.
- The banner's "Enable default pricing" button PATCHes the same operator-gated endpoint, so the whole affordance is operator-only and reads as though it is not.

Fix: gate `PricingWarning` on `useDeploymentAdminAccess`, the `/v1/admin/access` signal `useNavVisibility.ts:56` already uses for `operatorOnly` rows, and skip the settings query until it answers yes. Same visible behavior, stated rather than inferred, and no refused request per page load.

Not a security issue: a 403 is deliberately not the sign-out path (`web/src/shared/api/client.ts:411`) and is not retried (`web/src/app/provider.tsx:15`), so a non-operator sees no error and no loop today.

Acceptance: a signed-in non-operator issues no `GET /v1/settings`; an operator on a deployment with `require_pricing` on and `default_pricing` off still gets the banner.

njbrake:
## Description

The dashboard shows a gateway-wide alarm when the gateway is set to reject requests for models it has no price for. Only the person who runs the deployment can act on it, and only they could ever see it, but nothing in the code said so. What hid it from everyone else was the server refusing the request that feeds it: the banner asked for the deployment's settings on every page load, the server said no to anyone who is not a deployment operator, and with no answer the banner drew nothing. That worked, but it meant every ordinary user's browser fired a request that was always refused, once a minute per page, for a banner they could never be shown. This PR asks the question directly instead: the shell already knows whether the signed-in person operates the deployment, so the banner reads that first and only asks for settings when the answer is yes. Nothing looks different to anyone. There is just one refused request fewer on every page load.

The signal is `deployment_operator` off the organization context, passed as the `enabled` argument to `useSettings()`. Issue #834 names `useDeploymentAdminAccess` / `/v1/admin/access` instead, from before #836 moved the rail's operator answer onto the context; using it here would replace one refused request with one extra request, where the context is a read the shell already makes on every page. `ModelsPage.tsx` and `useProviderKeyEncryption` gate the same operator-only route the same way.

## PR Type

- [ ] New Feature
- [ ] Bug Fix
- [x] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes #834

## Checklist

Dashboard-only, so the checks that apply are `web/AGENTS.md`'s: `pnpm --dir web run lint`, `typecheck`, `test`, and `build`, all clean. No API contract change, so no regenerated spec. The new test asserts a non-operator issues no `GET /v1/settings` and sees no banner; it fails without the change.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

Implemented and tested by the agent from the issue; @njbrake reviewed before it was opened.

## Summary

- Gated the pricing warning to deployment operators.
- Prevented non-operators from requesting deployment settings or seeing cached pricing warnings.
- Preserved the existing operator warning and pricing update behavior.
- Added tests for non-operators, demoted sessions, and context read failures.

This avoids refused settings requests while keeping user-visible behavior unchanged.

### DIFF
--- a/web/src/features/models/PricingWarning.tsx
+++ b/web/src/features/models/PricingWarning.tsx
@@ -33,11 +33,17 @@ export function PricingWarning() {
   // `useProviderKeyEncryption` does: a second request to ask the same question
   // is the cost this removes.
   const organization = useOrganizationContext()
-  const settings = useSettings(isDeploymentOperator(organization.data))
+  const isOperator = isDeploymentOperator(organization.data)
+  const settings = useSettings(isOperator)
   const updateSettings = useUpdateSettings()
   const [dismissed, setDismissed] = useState(false)
 
+  // The render is gated on the same answer the request is, because a disabled
+  // query still serves whatever sits under its key and the query client outlives
+  // a sign-out: without this, a session that follows an operator's in the same
+  // tab renders their settings and a button that would be refused.
   const needsPricing =
+    isOperator &&
     settings.data?.require_pricing === true &&
     settings.data.default_pricing === false
   const showing = needsPricing && !dismissed
--- a/web/src/features/models/PricingWarning.tsx
+++ b/web/src/features/models/PricingWarning.tsx
@@ -33,7 +33,14 @@ export function PricingWarning() {
   // `useProviderKeyEncryption` does: a second request to ask the same question
   // is the cost this removes.
   const organization = useOrganizationContext()
-  const isOperator = isDeploymentOperator(organization.data)
+  // Fails open on a failed context read, which is what the rail does with the
+  // same class of gate: `/v1/settings` is `require_deployment_operator`, so it
+  // refuses with a 403 rather than a 404, and `nav/types.ts` settles what that
+  // means with no answer. Here it costs more than a hidden row, because the
+  // banner is the only thing reporting that traffic is being dropped right now.
+  // The ordinary tenant path, a resolved context saying no, still asks nothing.
+  const isOperator =
+    isDeploymentOperator(organization.data) || organization.isError
   const settings = useSettings(isOperator)
   const updateSettings = useUpdateSettings()
   const [dismissed, setDismissed] = useState(false)
--- a/web/src/features/models/PricingWarning.tsx
+++ b/web/src/features/models/PricingWarning.tsx
@@ -2,8 +2,10 @@ import { Button } from "@heroui/react"
 import { Link } from "@tanstack/react-router"
 import { useState } from "react"
 
+import { isDeploymentOperator } from "@/features/organization/roles"
 import {
   useFailureCount,
+  useOrganizationContext,
   useSettings,
   useUpdateSettings,
 } from "@/shared/api/hooks"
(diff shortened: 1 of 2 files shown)


web/src/features/models/PricingWarning.tsx
```diff
@@ -22,11 +24,25 @@ import { HOUR_S } from "@/shared/helpers/timeRange"
...
+  // Both the read behind the alarm and the button that clears it are
+  // deployment-operator-only (`require_deployment_operator` on `/v1/settings`),
+  // so the audience is stated here rather than left to be inferred from a
+  // refused query: without it every tenant page load fired a `GET /v1/settings`
+  // that 403s to feed a banner that could never render for them (#834). Off the
+  // organization context, which the shell reads anyway, for the reason
+  // `useProviderKeyEncryption` does: a second request to ask the same question
+  // is the cost this removes.
+  const organization = useOrganizationContext()
+  const isOperator = isDeploymentOperator(organization.data)
```
khaledosman: Fails closed when `/v1/organizations/me` itself fails, which silently removes the alarm from a real operator.

`/v1/settings` is `require_deployment_operator`-gated, so it is a **403**, and `app/nav/types.ts:79-85` settles what that class does with no answer: a `"refused"` surface is shown on a failed read, "because its existence is no secret ... hiding seven destinations because one query failed strands them". `useNavVisibility.ts:60` implements exactly that as `isOperator || answerUnavailable`.

Here the same failed read costs more than a hidden nav row: the banner is the only thing reporting that the gateway is dropping traffic right now, and nothing else surfaces it. Suggest mirroring the rail:

```ts
const isOperator =
  isDeploymentOperator(organization.data) || organization.isError
```

The ordinary tenant path (context resolved, `deployment_operator: false`) still asks nothing of `/v1/settings`, which is the request this PR is removing; only the error arm falls back to the old behavior, and the PATCH button behind it reports its own refusal the way those pages do.

khaledosman (approved):
One non-blocking finding inline: the gate fails closed on a failed context read, where this repo's own rule for a 403-refused surface is to fail open.

`make openapi-check` / Postman are not in play (dashboard-only, no contract change), and the new test does fail without the change.

---

PR #884, opened 2026-08-31
feat(growth): resolve GrowthSignalPort at signup and a member's first key

### LINKED ISSUE #796: Resolve GrowthSignalPort at the lifecycle moments this tree already has
#756 landed `GrowthSignalPort` with its Null adapter, and nothing calls it. The
port is otherwise fully wired: a binding in `container.py`, a resolver and an
exported `GrowthSignalPortDep` in `api/deps.py`, and container tests. No route or
service under `src/` resolves the dep, and none of its methods is called anywhere
in the tree.

Same shape as #757 for `ModelProviderPort`, and the same silent failure. In this
build nothing looks wrong, because the Null adapter records nothing by design. A
build that binds a real CRM or support messenger gets a fully resolved port that
is simply never called: no signup, activation, profile edit or deletion notifies
anything, and no test, log line or type error says so. That is the "contribution
vanishes with no error" mode the seam-specifier rule and #758 exist to prevent,
one layer down. The port reads as live infrastructure and is scaffolding.

## Four of the five methods, not all five

`record_onboarding_completed` is deliberately uncalled and should stay that way.
Its own docstring says why ("the questions and the answers they produce are not in
this tree"), `models/tenancy.py` says the reconciled schema carries no onboarding
columns, and #789 settled where those answers go: a hosted build's onboarding step
reaches its overlay's own contributed route, never this port. Wiring it would mean
growing a question registry in this tree, which is what all three of those say not
to do.

The other four have had their moments since #708 landed signup, so staging does
not explain them.

| method | site | how "first" is known |
|---|---|---|
| `record_signup` | `routes/auth_signup.py` `signup` | the call is the event |
| `record_activation(API_KEY_CREATED)` | `routes/keys.py` `create_key` | the owner's key count |
| `record_activation(FIRST_ROUTE_CONFIGURED)` | `routes/routing.py` `set_policy` | policy count |
| `record_activation(FIRST_REQUEST_ROUTED)` | `services/tenancy/workspace_activation_service.py`, where status becomes `activated` | already derived there |
| `record_activation(BUDGET_RULE_SET)` | `routes/budgets.py` `create_budget` | budget count |
| `record_profile_updated` | `routes/users.py` `update_user` | not a first-occurrence event |
| `record_account_deleted` | `routes/users.py` `delete_user` | not a first-occurrence event |

## The decision this needs, and most of the work

`record_activation`'s contract puts first-occurrence detection on the caller, and
this schema has nowhere to record "already notified" for the same reason it has no
CRM columns. So each activation site either derives "first" from data it is one
query away from anyway, or leans on the docstring's other sentence, that "an
adapter may additionally de-duplicate on its own end, so a redundant call is
safe", and notifies every time. The first writes more code and sends no redundant
traffic; the second is trivial here and moves the burden onto every adapter. Worth
picking one for all four rather than deciding per site.

One ambiguity to settle on the way: `GrowthActivationEvent.SIGNED_UP` overlaps the
dedicated `record_signup` method, and it is not obvious whether a signup should
fire one, the other, or both. Whichever it is belongs in the port docstring.

The fire-and-forget shape is already settled and adds nothing here: every method
takes the request's `BackgroundTasks` and returns nothing, so no call site gains a
failure path.

## If the answer is that this stays uncalled

Then close this by saying so in the port docstring, the way
`record_onboarding_completed` already does, so the next reader does not conclude
the port works.

_Note: this issue was drafted by Claude Opus 5 via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is the model's._

njbrake:
## Description

`POST /v1/auth/signup` calls `record_signup` only on a genuine claim, which `create_user_for_signup` now reports by returning the identity; the response is unchanged either way, so the seam does not undo the route's enumeration-safety. `POST /v1/organizations/me/keys` calls `record_activation(API_KEY_CREATED)`, with first occurrence derived from the owner's own keys before the insert.

The rest of the port stays uncalled, with the reason in each docstring: the two operator-route milestones would name whoever holds deployment-wide authority rather than the person a vendor watches, `FIRST_REQUEST_ROUTED` has no moment because workspace activation is derived on read, and profile edits and account deletion have no surface in this tree. `SIGNED_UP` is settled in `record_signup`'s favor.

Integration tests bind a recording adapter through the container and assert the silences as well as the calls.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Closes #796

## Checklist

## AI Usage

- [ ] No AI was used.

Implemented and self-reviewed by Claude Opus 5 through back-and-forth with @njbrake. The reasoning and decisions are his; the prose and code are Claude's.

## Summary

- Added growth notifications for genuine signups.
- Added activation tracking for a user’s first API key.
- Documented supported and intentionally unsupported lifecycle events.
- Shared the request database session with growth adapters.
- Added integration tests for notifications, duplicate actions, and excluded cases.

These changes make growth tracking active, accurate, and transaction-safe while preserving existing API responses.

### DIFF
--- a/src/gateway/api/routes/auth_signup.py
+++ b/src/gateway/api/routes/auth_signup.py
@@ -15,11 +15,11 @@ genuinely just claimed.
 
 from typing import Annotated
 
-from fastapi import APIRouter, Depends, Request
+from fastapi import APIRouter, BackgroundTasks, Depends, Request
 from pydantic import BaseModel, Field
 from sqlalchemy.ext.asyncio import AsyncSession
 
-from gateway.api.deps import get_config, get_db
+from gateway.api.deps import GrowthSignalPortDep, get_config, get_db
 from gateway.api.routes._public_auth import mail_unavailable, throttle_public_auth
 from gateway.core.config import GatewayConfig
 from gateway.services.mail import MailNotConfiguredError
@@ -85,8 +85,10 @@ _RESEND_MESSAGE = "If this address is registered and unverified, a verification 
 async def signup(
     body: SignupRequest,
     request: Request,
+    background_tasks: BackgroundTasks,
     db: Annotated[AsyncSession, Depends(get_db)],
     config: Annotated[GatewayConfig, Depends(get_config)],
+    growth: GrowthSignalPortDep,
 ) -> SignupResponse:
     """Claim a roster identity, or do nothing: the response never says which.
 
@@ -95,7 +97,7 @@ async def signup(
     """
     throttle_public_auth(request)
     try:
-        await create_user_for_signup(
+        claimed = await create_user_for_signup(
             db,
             config,
             email=body.email,
@@ -105,6 +107,19 @@ async def signup(
         )
     except MailNotConfiguredError as exc:
         raise mail_unavailable(exc) from None
+    # Only a genuine claim notifies, which is what keeps the seam from undoing
+    # the enumeration-safety above: the response is identical either way, and a
+    # signal an operator's own vendor receives is not something the caller can
+    # observe. ``email`` is set on any claimed identity, which is looked up by
+    # address in the first place.
+    if claimed is not None and claimed.email is not None:
+        await growth.record_signup(
+            background_tasks=background_tasks,
+            user_id=claimed.id,
+            email=claimed.email,
+            full_name=claimed.full_name,
+            created_at=claimed.created_at,
+        )
     return SignupResponse(message=_SIGNUP_MESSAGE)
 
 
--- a/src/gateway/api/routes/organization_keys.py
+++ b/src/gateway/api/routes/organization_keys.py
@@ -36,14 +36,20 @@ import uuid
 from datetime import datetime
 from typing import Annotated, Any
 
-from fastapi import APIRouter, Depends, HTTPException, Query, status
+from fastapi import APIRouter, BackgroundTasks, Depends, HTTPException, Query, status
 from pydantic import BaseModel, Field
 from sqlalchemy import select
 from sqlalchemy.exc import SQLAlchemyError
 from sqlalchemy.ext.asyncio import AsyncSession
 from sqlmodel import col
 
-from gateway.api.deps import CurrentIdentity, get_config, get_db, verify_master_key
+from gateway.api.deps import (
+    CurrentIdentity,
+    GrowthSignalPortDep,
+    get_config,
+    get_db,
+    verify_master_key,
+)
 from gateway.api.routes.keys import (
     _KEY_EXCEEDS_USER_DETAIL,
     CreateKeyResponse,
@@ -55,6 +61,7 @@ from gateway.core.config import GatewayConfig
 from gateway.models.entities import APIKey, User
 from gateway.models.tenancy import User as TenancyUser
 from gateway.models.tenancy import Workspace
+from gateway.ports.growth_signal_port import GrowthActivationEvent
 from gateway.repositories.users_repository import get_or_create_attribution_user
 from gateway.services.model_access import is_allowlist_subset, validate_allowed_models
 from gateway.services.tenancy import OrganizationService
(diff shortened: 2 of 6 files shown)


src/gateway/api/routes/auth_signup.py
```diff
@@ -85,8 +85,10 @@ class ResendVerificationResponse(BaseModel):
 async def signup(
     body: SignupRequest,
     request: Request,
+    background_tasks: BackgroundTasks,
     db: Annotated[AsyncSession, Depends(get_db)],
     config: Annotated[GatewayConfig, Depends(get_config)],
+    growth: GrowthSignalPortDep,
```
khaledosman, line 91: This PR gives `GrowthSignalPort` its first callers, which is the moment to settle the session it is resolved on. `get_growth_signal_port` still takes `PortSessionDep` (`api/deps.py:553`), and both new call sites are standalone-only routes that already take `get_db`, so FastAPI builds the adapter a **second, independent** session per request; `deps.py:534-539` states that consequence, and it is precisely why `get_identity_provider_port` (`:558`) and `get_telemetry_storage_port` (`:581`) name `get_db` instead, each with the comment saying so.

No bug today (the core adapter is a Null Object and no connection is checked out until something queries), but it is a latent one for the first overlay adapter that writes through the session it is handed: an outbox row or a `notified_at` stamp would land in a transaction nobody commits, and silently vanish. Switching `get_growth_signal_port` to `Annotated[AsyncSession, Depends(get_db)]` with the sibling comment costs nothing while both callers are standalone-only.

src/gateway/api/routes/organization_keys.py
```diff
@@ -229,6 +238,14 @@ async def create_own_key(
     if not is_allowlist_subset(allowed_models, owner.allowed_models):
         raise HTTPException(status_code=status.HTTP_400_BAD_REQUEST, detail=_KEY_EXCEEDS_USER_DETAIL)
 
+    # ``record_activation`` puts first-occurrence detection on the caller, and
+    # this schema has nowhere to record "already notified", so the milestone is
+    # derived from the owner's own keys: an index seek on ``api_keys.user_id``,
+    # read before the insert below because afterwards there is always one.
+    first_key = (
```
khaledosman, line 245: `first_key` is "holds no key row right now", not "never held one": `delete_own_key` (`:429`) hard-deletes, so create → delete → create fires `API_KEY_CREATED` a second time for the same person. Same for an operator pruning the member's keys through `DELETE /v1/keys/{id}`.

The port tolerates it ("a redundant call is safe", and an adapter may de-duplicate), so this is not a defect in the call. It is a mismatch with the docstring this PR adds to `record_activation`: "first occurrence is the owner holding no key yet" reads as a once-per-user guarantee an adapter author might rely on rather than de-duplicate against. Worth one clause there saying the derivation re-arms when every key is deleted, since that docstring is now the contract an overlay reads.

khaledosman (approved):
Two non-blocking findings inline. Verified locally on the head commit: `make lint` (architecture check, alembic heads, Ruff) and `make openapi-check` are clean, so no generated artifact is owed.

The transaction ordering is right in both routes (signup notifies after the service commits, the key route after `db.commit()`), and `expire_on_commit=False` (`core/database.py:131`) is what makes reading `identity.email` after the commit safe.

---

src/gateway/api/routes/auth_signup.py
njbrake: Done in a57d00a. `get_growth_signal_port` now takes `Annotated[AsyncSession, Depends(get_db)]` with the sibling comment, so an overlay adapter's writes land in the transaction the route commits rather than in a second session nobody commits.

src/gateway/api/routes/organization_keys.py
njbrake: Clause added in a57d00a. `record_activation` now says the `API_KEY_CREATED` derivation is a read of current state rather than a once-per-person guarantee: deleting every key re-arms it, and two concurrent creates can both read zero, so an adapter needing once-per-person de-duplicates on its own end.

PR #885, opened 2026-09-01
feat(dashboard): adopt the divided-surface design direction

jigjigjig:
## Description

Adopts a single visual direction across the dashboard: one flat surface, divided by hairline rules rather than assembled from cards. Square corners everywhere, no drop shadows, status carried by a square dot plus a word rather than by a tinted pill, and one data accent (`#0098A4`) that never leaks into chrome.

**Stacked on #871.** The base is `feat/dashboard-palette`, not `main`, so the palette commits are not counted twice. The plan is to land one PR in the end; the stack exists for reviewing a clean diff meanwhile.

### The token layer

`web/src/styles/globals.css` carries the systemic half, and most of it reaches every component at once:

- `--radius: 0px` declared unlayered, which zeroes the corners on HeroUI's internals as well as ours. It has to be unlayered: `@theme` is hoisted, so a `@theme` override of a HeroUI key compiles clean and loses.
- All four shadow tokens are `none`. On dark, several of those "shadows" were secretly 1px rings doing structural work; every control that lost its only edge that way was given a real border (the popover family, the toggle, the segmented controls). The pattern held four times, always on controls, never on content.
- One focus ring, `@utility otari-focus-ring`, plus a by-ground rule: the ring takes the theme's ink when it rings a control whose own fill is an accent or status color, `--color-focus` everywhere else.
- Two-tier dividers with a registered class for each tier, after finding that two of the three "subtle" tokens were declared but never registered, so their classes compiled to nothing.

### Pages

Every page is on the direction: the shell, Overview, Usage, Activity, API keys, Models, Providers, Routing, Tools and Guardrails, the organization suite (General, Members, Provider keys, Rate overrides, Workspaces, Model pricing, Settings, Account, Deployment accounts), Budgets, Docs, and the public auth pages (one shared layout, so they cannot drift). No `<Card>` remains; the last boxed grid (the docs guide's comparison table) is divided rows like every other table.

### The controls system

Sizes are named places rather than per-site numbers: form fields 36px, dense controls 32px (toolbars, table rows, pagination), with sign-in keeping its 44px submit and 40px key field. Buttons are filled with white ink on a darkened step of their own family's color (primary `#00808A`, danger split per theme because danger's ink differs per theme), with a size-tuned press transform carrying feedback and a reduced-motion fallback that never resurrects a failing color step. Field helpers and error messages share one caption role (13px), errors differ by color only, and every field reserves its caption line so validation does not shift the form. Selection chips, filter chips and status words follow a four-kind taxonomy (status, affirmative, category, identifier) where dismissability and the severity dot are modifiers, not kinds. Hit targets are 40px (44 coarse) behind invisible bleeds that never exceed the pitch they repeat within.

### Shared vocabulary

`web/src/shared/components/surface.tsx` collects what stopped being one screen's layout: the full-bleed band, the status dot, the KPI strip and its subgrid cells, the meter (three states, including a near-limit form that reads in grayscale), the tab row, the segmented control, the table frame with its held identity lane, the field piece that owns the label rung, description rung and reserved line together, and the page header.

### Guards added along the way

Tests now pin the things that failed silently at least once: token registration (a self-referential `@theme` entry that no theme block declares), the divider tiers, dot colors staying on text ramps, focus rings never spelled at call sites, comment nesting in the stylesheet, and the URL rewriting to the applied preset when an unrecognized range falls back.

## Still open, deliberately

- The four PWA icons and the `/welcome` page's wordmark still carry the old mark color (`#4E8295`); both need either new artwork or a decided regeneration path.
- The extracted field labels are `<span>`s not yet associated with their controls; giving them `aria-labelledby` is a behavior change deliberately kept out of a styling refactor and worth its own change.
- The degraded-provider status (a provider that answers but lists no models) has never been rendered against a real gateway; it is correct against the design and unverified live.
- The coarse-pointer pagination variant is verified in the emitted CSS and not rendered (no touch device in the loop); the five-column KPI strip has not been eyeballed at a native ~1440 viewport.
- `Login.test.tsx` has flaked three times under full-suite load and never reproduced in isolation; green in every recorded run since.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

- [x] Documentation was updated where necessary: `web/AGENTS.md` and the frontend-standards topic files carry the token-registration trap and its probe method, the divider tiers, the shared vocabulary module, and three measurement hazards (stale bundle in a hash-routed tab, transitions frozen in background tabs, ancestor-clipped elements reporting as rendered).
- [ ] If the API contract changed, I regenerated the OpenAPI spec. No API contract changed; this is `web/` plus one backend template (the favicon).

Tests are the dashboard's own suites rather than `tests/unit` and `tests/integration`: `pnpm --dir web test` is at 1609 passing across 109 files. `pnpm --dir web run lint`, `run typecheck` and `run build` are clean.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:** Claude (Claude Code).

Nearly everything here was verified by measuring the running page rather than by reading the source, which repeatedly caught changes that compiled, linted and passed every test while doing nothing (an unregistered token class, a tailwind-merge collision that ate a type role, a hover painting its own background color). The recurring instrument hazards that produced false findings the same way are documented in `web/AGENTS.md` so they are not relearned.


jigjigjig:
Handoff notes for review, verified against the tree at d8e60302 (or later brand-asset commits):

**Two conditional-render stragglers** found by a final grep; neither has ever rendered in a review environment because their conditions never fired, and both predate this branch's rules:
- `web/src/features/usage/UsagePage.tsx:1281` — the `groupingUnsupported` notice is a rounded, bordered, tinted panel (`rounded-md border border-warning bg-warning-subtle`). The direction bans all three; the shipped pattern for this is the flat notice strip (dot plus text, no fill, no border, no radius), and the copy is fine as is.
- `web/src/features/routing/RoutingPage.tsx:1191` — the `outgrewAlias` note is `text-xs text-warning` under the control it explains; per the caption rule in the codebase it belongs at `text-caption` with the warning ink kept.

**Known-open items, deliberate rather than missed** (also summarized in the PR description):
- `ControlField` labels are spans with no programmatic association; adding ids and `aria-labelledby` is a behavior change kept out of the styling refactor on purpose.
- The degraded-provider status (answers but lists no models) has never been rendered against a real gateway.
- The coarse-pointer pagination variant exists in the emitted CSS and has never been rendered on a touch device.
- Two `Login.test.tsx` tests have been intermittently slow under full-suite load; green in every recorded run.
- `font-medium` resolves to 400 at all call sites because the type scale pins `--font-weight-medium: 400` deliberately (documented at the token); flagging so it is not filed as a bug.
- `account/PasskeysCard.tsx:80` stays at 12px deliberately: it reports a listed item's state rather than helping operate a form control.

jigjigjig:
Superseded by #934, which lands this work rebased onto current main. Closing.

PR #888, opened 2026-09-01
fix(hybrid): map PLATFORM_HEALTH_PATH and trust only a 2xx from the platform probe

### LINKED ISSUE #877: Map PLATFORM_HEALTH_PATH so a hybrid gateway can probe a non-default platform health route
`_check_platform_reachability` builds its probe URL as `{platform.base_url}/{platform.health_path}`, with `health_path` defaulting to `/utils/health-check/` (`src/gateway/api/routes/health.py:21`).

`_apply_platform_env_overrides` maps `base_url`, `management_url` and the timeouts, and nothing else (`src/gateway/core/config.py:2235`), so `health_path` can only come from a config file. A deployment that sets `PLATFORM_BASE_URL` from the environment and passes no config file cannot move the probe off that default.

The default is a platform route. A peer that does not serve it answers 404, and the probe counts any status below 500 as reachable, so `GET /health` reports `platform_reachable: "yes"` and `/health/readiness` returns 200 while the resolve peer is unverified. The field stops meaning what it says.

Add `PLATFORM_HEALTH_PATH` to `env_mappings`.

Found pointing otari.ai's local gateway at its control plane, whose `/v1` surface does not serve `/utils/health-check/` (mozilla-ai/otari-ai#1950). Worked around there with a config file; deployed development still runs the default.

njbrake:
When Otari runs in hybrid mode, its health endpoint checks whether the platform it depends on is really up, by asking that platform a question over the network. Two things were wrong with that check. The address it asked could only be half set from environment variables, so anyone running Otari purely from the environment was stuck asking for a page specific to otari.ai. And the check accepted almost any reply as proof of life, including "no such page" and "please log in first". Together those meant Otari could report the platform as healthy when it had never actually reached it.

This fixes both. The address is now fully settable from the environment, and only a genuine success counts as healthy. A deployment pointed at the wrong place now finds out, instead of being told everything is fine.

## Description

`_check_platform_reachability` builds its probe URL from `platform.health_path`, defaulting to `/utils/health-check/`. Two defects, both reported in #877.

**The path could not be set from the environment.** `_apply_platform_env_overrides` mapped `base_url`, `management_url` and the timeouts, but not `health_path`, so only a config file could move it. Adds `PLATFORM_HEALTH_PATH` to `env_mappings`.

**Any status below 500 counted as reachable.** A peer that does not serve the configured path answered 404 and still reported `platform_reachable: "yes"`, with `/health/readiness` returning 200 against a peer nobody had verified. The probe now requires a 2xx: a 404, a 401 from an authenticated route, and a redirect to a login page all report unreachable. Redirects are left unfollowed, so the 200 at the end of one cannot stand in for the peer's own answer.

Two smaller things travel with that. The `except` clause caught only `TimeoutException` and `NetworkError`, leaving `ProtocolError`, `ProxyError`, `UnsupportedProtocol` and `InvalidURL` to surface as a 500 from `/health` rather than a `"no"`; letting an operator aim the probe at an arbitrary route is what makes those reachable in practice. And the default path is hoisted to `DEFAULT_PLATFORM_HEALTH_PATH`, so the route, the docs table and the tests stop keeping separate copies of it.

**Behavior change worth a release note.** A hybrid deployment whose configured health route answers non-2xx, an auth-protected endpoint returning 401 for instance, flips from `platform_reachable: "yes"` to `"no"`, and `/health/readiness` starts returning 503. That is what the issue asks for, but it can pull pods out of rotation on upgrade.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes #877

## Checklist

- [x] I ran the Definition of Done checks locally (`make lint`, `make typecheck`, tests).
- [x] If the API contract changed, I regenerated the OpenAPI spec. No route or docstring changed, so the spec is unaffected; `make openapi-check` and `make postman-check` both report no drift.

Probe tests went from 2 to 13. Six of them fail against the previous implementation (404, 401, 403, the redirect, `RemoteProtocolError`, `ProxyError`), so the new semantics are covered rather than assumed. Local runs: `make lint`, `make typecheck`, full `tests/unit` (2652 passed), `tests/integration` for hybrid surface, health and config env loading (53 passed), and the OSS-edition smoke, all clean against a local PostgreSQL, since this sandbox has no Docker for Testcontainers.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

Implemented and tested by the agent through back-and-forth with @njbrake; the reasoning and decisions are his.

### DIFF
--- a/tests/unit/test_platform_reachability_probe.py
+++ b/tests/unit/test_platform_reachability_probe.py
@@ -1,0 +1,70 @@
+"""The ``platform_reachable`` probe must hit the path the deployment configured.
+
+The default path is a platform route. A peer that does not serve it answers
+404, which the probe counts as reachable, so an unverified peer reports healthy.
+``PLATFORM_HEALTH_PATH`` is what moves the probe off that default in an
+env-only deployment.
+"""
+
+from typing import Any
+
+import httpx
+import pytest
+
+from gateway.api.routes.health import _check_platform_reachability
+from gateway.core.config import load_config
+
+
+class _RecordingTransport(httpx.AsyncBaseTransport):
+    def __init__(self) -> None:
+        self.requested: list[httpx.URL] = []
+
+    async def handle_async_request(self, request: httpx.Request) -> httpx.Response:
+        self.requested.append(request.url)
+        return httpx.Response(200)
+
+
+@pytest.fixture
+def transport(monkeypatch: pytest.MonkeyPatch) -> _RecordingTransport:
+    recorder = _RecordingTransport()
+    original_init = httpx.AsyncClient.__init__
+
+    def patched_init(self: httpx.AsyncClient, *args: Any, **kwargs: Any) -> None:
+        kwargs["transport"] = recorder
+        original_init(self, *args, **kwargs)
+
+    monkeypatch.setattr(httpx.AsyncClient, "__init__", patched_init)
+    return recorder
+
+
+@pytest.fixture(autouse=True)
+def _isolated_platform_env(monkeypatch: pytest.MonkeyPatch) -> None:
+    for name in ("OTARI_AI_TOKEN", "PLATFORM_BASE_URL", "PLATFORM_HEALTH_PATH", "OTARI_MODE"):
+        monkeypatch.delenv(name, raising=False)
+
+
+@pytest.mark.asyncio
+async def test_health_path_env_var_moves_the_probe(
+    monkeypatch: pytest.MonkeyPatch, transport: _RecordingTransport
+) -> None:
+    monkeypatch.setenv("OTARI_AI_TOKEN", "gw_test_token")
+    monkeypatch.setenv("PLATFORM_BASE_URL", "http://platform.test/v1")
+    monkeypatch.setenv("PLATFORM_HEALTH_PATH", "/healthz")
+
+    config = load_config()
+
+    assert await _check_platform_reachability(config) is True
+    assert [str(url) for url in transport.requested] == ["http://platform.test/v1/healthz"]
+
+
+@pytest.mark.asyncio
+async def test_probe_falls_back_to_the_platform_route(
+    monkeypatch: pytest.MonkeyPatch, transport: _RecordingTransport
+) -> None:
+    monkeypatch.setenv("OTARI_AI_TOKEN", "gw_test_token")
+    monkeypatch.setenv("PLATFORM_BASE_URL", "http://platform.test/v1")
+
+    config = load_config()
+
+    assert await _check_platform_reachability(config) is True
+    assert [str(url) for url in transport.requested] == ["http://platform.test/v1/utils/health-check/"]
--- a/tests/unit/test_platform_reachability_probe.py
+++ b/tests/unit/test_platform_reachability_probe.py
@@ -6,6 +6,7 @@ The default path is a platform route. A peer that does not serve it answers
 env-only deployment.
 """
 
+from pathlib import Path
 from typing import Any
 
 import httpx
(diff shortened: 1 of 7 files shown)


tests/unit/test_platform_reachability_probe.py
```diff
@@ -0,0 +1,77 @@
...
+async def test_probe_falls_back_to_the_platform_route(
+    monkeypatch: pytest.MonkeyPatch, transport: _RecordingTransport
+) -> None:
+    monkeypatch.setenv("OTARI_AI_TOKEN", "gw_test_token")
+    monkeypatch.setenv("PLATFORM_BASE_URL", "http://platform.test/v1")
+
+    config = load_config()
+
+    assert await _check_platform_reachability(config) is True
+    assert [str(url) for url in transport.requested] == ["http://platform.test/v1/utils/health-check/"]
```
khaledosman: The default `/utils/health-check/` now has four homes: this assertion, the literal in `health.py:21`, the Default column of the protocol table this PR adds, and the comment in `web/e2e/otari.hybrid.yml`. Now that it is a documented default, hoist it to `DEFAULT_PLATFORM_HEALTH_PATH` in `core/config.py` beside `DEFAULT_PLATFORM_BASE_URL` and `DEFAULT_PLATFORM_MANAGEMENT_URL`, read it at `health.py:21`, and import it here so the test moves with the default rather than pinning a copy of it.

khaledosman (approved):
Does what #877 asks, with the mapping, the docs row and both probe-URL cases covered. Two notes beyond the diff, neither blocking.

**`src/gateway/api/routes/health.py:29`** (unchanged, but this PR is what makes that function's target configurable): `except (httpx.TimeoutException, httpx.NetworkError)` leaves `httpx.ProtocolError` (including `RemoteProtocolError`), `ProxyError`, `UnsupportedProtocol` and `InvalidURL` uncaught. A peer that answers the probe with a malformed HTTP response therefore makes `GET /health` raise 500 rather than report `platform_reachable: "no"`, and `/health/readiness` 500 instead of 503. Aiming the probe at an operator-chosen route on an arbitrary peer is exactly what this adds, so it widens that surface. `except (httpx.HTTPError, httpx.InvalidURL)` covers it.

**Follow-up for the half of #877 left open.** The issue's third paragraph is the semantic defect (`404` counts as reachable, so the field "stops meaning what it says"), and scoping it out here is defensible, but `Fixes #877` closes the only record of it. Worth opening a follow-up. Related, in the same rule: the probe does not follow redirects, so a peer route that `302`s to a login page also reports reachable.

One thing worth the note: the `httpx.AsyncClient.__init__` patching in the new fixture matches the established pattern in `test_web_search_backend.py` and `test_sandbox_backend.py`, and the autouse env fixture is genuinely needed, since neither `tests/conftest.py` nor any unit conftest isolates env.

---
🤖 Reviewed by Claude Opus 5 via [Claude Code](https://claude.com/claude-code), run by @khaledosman.

tests/unit/test_platform_reachability_probe.py
njbrake: Done in bd23bf366. `DEFAULT_PLATFORM_HEALTH_PATH` now sits in `core/config.py` beside `DEFAULT_PLATFORM_BASE_URL` and `DEFAULT_PLATFORM_MANAGEMENT_URL`; `health.py` reads it and this test imports it, so neither pins a copy.

The same commit takes the other two notes: the probe now requires a 2xx (404, 401 and an unfollowed redirect all report unreachable), and the except clause is `(httpx.HTTPError, httpx.InvalidURL)`.

PR #897, opened 2026-09-01
feat(dashboard): add an organization-wide usage page for admins

khaledosman:
## Description

Gives org admins on a multi-tenant deployment a dedicated organization-wide Usage page. The organization rail gains an Observe section whose Usage destination renders the existing analytics pinned to `/v1/organizations/me/usage` (never the deployment-wide surface, even for an operator) and unnarrowed by default, with an explicit workspace filter that can name any workspace in the organization rather than only the caller's memberships.

The page is gated on a new `organization_usage` surface reported only by the hosted bootstrap: standalone's organization is the deployment, so `/usage` already answers it whole. Breakdown-row drill-down into Activity is withheld on the organization page because Activity still narrows to the sidebar's selected workspace and would silently show a subset.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes mozilla-ai/otari-ai#1963

## Checklist

- [x] I ran the Definition of Done checks locally (`make lint`, `make typecheck`, `make test`); the integration half of `make test` is left to CI because this environment has no Docker, and `tests/unit` (2790 passed) plus the full dashboard suite (2668 passed), `tsc`, and Biome all ran locally.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

Claude Fable 5 (Claude Code)

Surface list is runtime data, so regenerating the OpenAPI spec produced no diff.

## Summary

- Added an organization-wide Usage page for admins in hosted, multi-tenant deployments.
- Added an **Observe** section with a **Usage** destination in the organization navigation.
- Added workspace filtering for workspaces visible to the caller.
- Updated usage requests to use `/v1/organizations/me/usage` for organization scope.
- Kept organization usage unnarrowed by default and labeled shared data clearly.
- Disabled Activity drill-down for organization-wide breakdown rows.
- Added hosted-surface synchronization, route, navigation, filter, roster, API, and screenshot tests.
- Updated dashboard documentation.

### DIFF
--- a/docs/dashboard.md
+++ b/docs/dashboard.md
@@ -77,6 +77,7 @@ The workspace view contains day-to-day gateway operations:
 
 The organization view contains tenant-wide administration:
 
+- Organization-wide usage, in hosted mode
 - Workspaces and organization members
 - Spend and budgets
 - Organization pricing
--- a/src/gateway/api/routes/bootstrap.py
+++ b/src/gateway/api/routes/bootstrap.py
@@ -116,9 +116,17 @@ STANDALONE_SURFACES: tuple[str, ...] = (
 # Dropping the surface is not itself a guard over the table: ``config.yml``'s
 # ``providers:`` block and ``/v1/provider-credentials`` still populate it with no
 # page in front of them, which is #818's to close.
+#
+# ``organization_usage`` appears for a different reason: not a credential's
+# ownership but a question that only exists once tenants do. The router behind it
+# (``/v1/organizations/me/usage``) is mounted on both editions, but on standalone
+# the organization is the deployment and ``/usage`` already answers it whole, so
+# the dashboard's organization-wide Usage page is a destination only where "my
+# organization" is narrower than "everything" (otari-ai#1963).
 HOSTED_SURFACES: tuple[str, ...] = (
     *(surface for surface in STANDALONE_SURFACES if surface != "providers"),
     "organization_providers",
+    "organization_usage",
 )
 
 
--- a/src/gateway/api/routes/bootstrap.py
+++ b/src/gateway/api/routes/bootstrap.py
@@ -95,9 +95,10 @@ STANDALONE_SURFACES: tuple[str, ...] = (
 )
 
 # The same list for a *hosted* deployment: one control plane serving many
-# organizations, rather than one operator's own gateway. Two rows differ, and
-# both differences are the same fact seen from either side, that a credential
-# here belongs to a tenant rather than to the process.
+# organizations, rather than one operator's own gateway. Three rows differ. The
+# first two are the same fact seen from either side, that a credential here
+# belongs to a tenant rather than to the process; the third is not about
+# credentials at all.
 #
 # ``providers`` drops. It is the deployment-instance surface over
 # ``provider_credentials``, whose primary key is the instance name alone, so an
--- a/tests/unit/test_deployment_bootstrap.py
+++ b/tests/unit/test_deployment_bootstrap.py
@@ -219,7 +219,10 @@ def test_mail_ready_turns_on_only_with_a_transport_and_a_public_url(tmp_path: Pa
 # would collapse them into the roster surface, which is a different page with
 # different access. Listed here rather than in the surface tuple itself so the
 # tuple stays the plain list of names the dashboard gates on.
-SURFACE_ROUTE_PREFIXES = {"organization_providers": "/v1/organizations/me/provider-keys"}
+SURFACE_ROUTE_PREFIXES = {
+    "organization_providers": "/v1/organizations/me/provider-keys",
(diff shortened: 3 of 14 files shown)


njbrake (approved):
Approving. The scope claims in the description hold in the code, not only in the prose.

## Findings

**Stale fixture, worth closing before merge.** `web/src/tests/fixtures.ts:105` states it is kept in step with the backend `HOSTED_SURFACES`, and only the backend half moved. Any hosted-shell test built on that constant (`AccountMenu.test.tsx:170`, `OverviewPage.test.tsx:428`) renders an organization rail with no Usage row, so the gate this PR adds cannot be exercised from a hosted fixture. Nothing fails today, which is what makes it quiet.

**Stale comment.** `src/gateway/api/routes/bootstrap.py:98` still opens the hosted block with "Two rows differ, and both differences are the same fact seen from either side, that a credential here belongs to a tenant rather than to the process." Three rows differ now, and the third is not about credentials, as the new comment 20 lines below says.

**One-caller wrapper.** `web/src/features/usage/OrganizationUsagePage.tsx` exists only to pass one prop. The identically shaped case inlines it: `web/src/routes/tools.guardrails.tsx:6` is `component: () => <ToolsGuardrailsPage only="guardrails" />` against `ToolsGuardrailsPage({ only } = {})`. Optional.

**Copy overstates what a member sees.** A member reaching `/organization/usage` by URL on a hosted deployment reads "across every workspace in your organization" and an "All workspaces" filter option, while the server correctly returns only their own workspaces. Either gate the row and route on `canManage`, or soften the two strings (`web/src/features/usage/UsagePage.tsx:1063,1110`).

**Share card scope line.** `shareScopeSuffix` names the workspace when the filter is set and says nothing when it is not, so an organization-wide card reads identically to a workspace-scoped one. Pre-existing on `/usage`; the organization page is where it misleads. Optional.

## Scope, as enforced

Pinning holds by construction. `isDeploymentWide = scope === "caller" && ...`, and all five usage reads on the page pass `scope`. Removing the guard fails both new tests, so they are load-bearing.

The widened workspace filter is enforced server side. `resolve_workspace_in_organization` raises `WorkspaceNotFoundError` on an organization mismatch before the management-role arm is reached, so another tenant's workspace answers 404 rather than empty rows. Covered by `test_another_organizations_workspace_is_not_found_on_every_route`. The picker cannot name one either: `/v1/workspaces` is bounded by the caller's active organization even for a superuser.

The unnarrowed default is bounded by a subquery over the caller's own active organization, which is never a request parameter, and `UsageLog.workspace_id` is NOT NULL, so nothing falls out of the organization total.

The surface gate is a route gate, not only a hidden row: `AppShell` runs `useRouteVisibility()` over `navItemForPath`, so standalone answers the URL with its "not available here" panel. No server-side gate on `organization_usage` is needed, since the router is mounted on every edition and is independently organization-scoped.

The rail entry sits behind `managesOrganization`, so the new section is absent for a member or a viewer.

njbrake:
@jigjigjig may want to chime in here to ensure that this aligns with his UI/UX vision.

khaledosman:
Thanks. All five addressed in 1380900b.

**Stale fixture.** Added `organization_usage` to the frontend `HOSTED_SURFACES`, and pinned it so it cannot drift again: `registry.test.ts` now asserts every declared-but-gated-off row's surface is present in that fixture, so a name added to the backend list and not the fixture fails there instead of quietly rendering a rail the product does not have.

That assertion immediately earned itself. Written first as "all three are hosted surfaces" it failed on `organization_guardrails`, which is in neither list: no edition here serves that API. So the three rows are not one category, and the test now says which is which. The comment above it was already imprecise on that point before this PR (it called provider keys an absent endpoint, where the registry's own note calls it a choice), so I corrected it while I was there.

**Stale comment.** Fixed. It says three rows differ now, with the third called out as not being about credentials.

**One-caller wrapper.** Inlined and the file deleted, exactly as `tools.guardrails.tsx` does it.

**Copy.** Softened rather than gated. The description now reads "across the workspaces you can see in your organization", which is true for an admin and for a member who arrives by URL, and the page does not have to know which is reading it. Gating the route on `canManage` would refuse a member a page that works for them and shows them their own true figures, which is the trade the standards argue against.

I left the filter's empty option as "All workspaces". It follows the house convention for an unnarrowed filter (Activity's source select says "All"), so it reads as "not narrowed" rather than as a coverage claim, and the picker only ever lists workspaces the caller can see. Happy to change it if you read it the other way.

**Share card.** An unnarrowed organization card now says "organization-wide" instead of staying silent, so it no longer reads identically to a workspace-scoped one.

jigjigjig:
@nathanbrake  span up a local env and things looks good, also nothing should block migration atm. all the rest we can fix later.

PR #898, opened 2026-09-01
fix(api): refuse to cap a gateway user at an organization's budget

### LINKED ISSUE #881: POST /v1/users can assign a gateway user to an organization-owned budget
Follow-up from #875. The data-loss half of this was fixed there; the authority half is still open, and closing it changes an operator-facing contract, which is why it was held back.

## What is open

`GET /v1/budgets` is unfiltered, so since `b7e1c4a9d2f5` it lists organization-owned budgets alongside the deployment's own. `POST /v1/users` and `PATCH /v1/users/{id}` validate a `budget_id` for **existence**, not for ownership, so an operator can point a gateway user at a tenant's budget.

#875 closed the worst consequence: deleting such a budget used to *succeed silently*, because `Budget.users` is a plain relationship with no `passive_deletes`, so the ORM nulled `users.budget_id` out before the foreign key was ever reached and the gateway user lost their cap with nobody told. That path now refuses with a 409.

What remains is the assignment itself. While it stands, a tenant admin editing their own budget moves what a gateway user may spend, and neither party can see the coupling: the admin cannot see the users page, and the operator sees no sign that the budget they picked is a tenant's.

A `users` row is deployment-wide with no tenancy column, so there is no reading under which a tenant's budget is the right cap for one.

## Suggested fix

Reject an organization-owned budget at both assignment sites:

```python
def _require_assignable_budget(budget: Budget | None, budget_id: str) -> Budget:
    if budget is None or budget.organization_id is not None:
        raise HTTPException(
            status_code=status.HTTP_404_NOT_FOUND,
            detail=f"Budget with id '{budget_id}' not found",
        )
    return budget
```

404 rather than 403, matching the not-found this route already answers for an unknown id: from the deployment surface's point of view a tenant's budget is not one it may assign, and telling the two apart says nothing useful.

Optionally narrow `GET /v1/budgets`, or mark the tenant-owned rows in its response, so the assignment UI does not offer a choice the API then refuses.

## Why it was not folded into #875

It turns a previously accepted request into a 404 on a published operator endpoint. That is a small but real contract narrowing on a surface #875 did not own, and @njbrake had already reviewed the neighbouring code and chosen to refuse the *delete* rather than prevent the *assignment*. The change is written and tested and can be lifted straight in if wanted.

Related: #875, mozilla-ai/otari-ai#1943

khaledosman:
## Description

`POST /v1/users` and `PATCH /v1/users/{id}` validated `budget_id` for existence, not for ownership. Since `b7e1c4a9d2f5` a budget can belong to a tenant, so an operator could cap a gateway user at one. A `users` row is deployment-wide with no tenancy column, so while that stood a tenant admin editing their own figure moved what a gateway user may spend, and neither party could see the coupling: the admin cannot reach the users surface, and the operator saw no sign the budget they picked was a tenant's.

#875 closed the data-loss half (deleting such a budget used to null `users.budget_id` out silently; it now refuses with a 409). This is the authority half it deliberately held back.

**What changed**

- `_require_assignable_budget` at both assignment sites answers the 404 an id naming nothing already gets. From the deployment surface's point of view a tenant's budget is not one it may assign, and telling the two apart says nothing useful.
- `BudgetResponse` now carries `organization_id` (null for the deployment's own). The list stays unfiltered, so an operator still sees and can still refigure a tenant's budget; what changes is that the row now says whose it is.
- The operator's Budgets page marks such a row ("Owned by an organization", mirroring the tenant page's "Set at the deployment level") and withholds its assignment control with a line saying why, so the UI stops offering a save the API refuses.

**Contract narrowing**, stated plainly: a request that used to succeed now answers 404. That is the whole point of the issue, and it is why it was not folded into #875.

**Legacy rows.** Nothing migrates an assignment made before this. `b7e1c4a9d2f5` is a day old and unreleased, so a row in that shape can only exist in a tree that ran it unreleased, and rewriting live `users.budget_id` values on upgrade is the worse trade. Worth knowing if one does exist: the assignment multiselect is the only dashboard control that writes `users.budget_id`, and this PR withholds it on exactly the rows that would carry such an assignment, while the People column goes on counting them. `PATCH /v1/users/{id}` with `budget_id: null` detaches, and #875's 409 keeps the tenant from deleting the budget out from under the user in the meantime.

**Not in scope**, and reachable after this merges: `POST /v1/scoped-budgets` validates a `budget_id` for existence only (`scoped_budgets.py:173`), and the workspace-member ceiling picker offers the same unfiltered list (`OrganizationMembersPage.tsx:480`), so a ceiling on a tenant's budget still saves with a 200. And `DELETE /v1/budgets/{id}` still removes a tenant's budget and nulls its holders. Filed as #901 and #902, so that closing #881 does not read as closing the class.

The delete-refusal guard from #875 still matters for rows assigned before this, so its test moved to the service layer and seeds the assignment directly; the comments in `errors.py` and `organization_budget_service.py` that described the assignment as possible were updated with it.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes #881

## Checklist

Dashboard checks were run too (`pnpm --dir web run lint`, `typecheck`, `test`). `make test` is green: the two unit assertions on `BudgetResponse` needed the new field and were updated. The OSS-edition smoke gate fails identically on an unmodified tree in this environment (its local `/health` probe never answers), so it was not usable as a signal here.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

Claude Opus 5 (1M context), via Claude Code

Implemented from the issue's suggested fix, which the issue notes was already written and tested.

### DIFF
--- a/docs/public/openapi.json
+++ b/docs/public/openapi.json
@@ -1531,6 +1531,18 @@
             ],
             "title": "Name"
           },
+          "organization_id": {
+            "anyOf": [
+              {
+                "format": "uuid",
+                "type": "string"
+              },
+              {
+                "type": "null"
+              }
+            ],
+            "title": "Organization Id"
+          },
           "reset_alignment": {
             "anyOf": [
               {
@@ -1564,6 +1576,7 @@
         },
         "required": [
           "budget_id",
+          "organization_id",
           "name",
           "max_budget",
           "budget_duration_sec",
--- a/src/gateway/api/routes/budgets.py
+++ b/src/gateway/api/routes/budgets.py
@@ -1,3 +1,4 @@
+import uuid
 from decimal import Decimal
 from typing import Annotated
 
@@ -68,6 +69,11 @@ class BudgetResponse(BaseModel):
     """
 
     budget_id: str
+    # Null for the deployment's own, and a tenant's organization when set. Carried
+    # so a caller can tell the two apart before offering one: an organization's is
+    # listed here for the operator to see, and `POST /v1/users` refuses to cap a
+    # gateway user at it (mozilla-ai/otari#881).
+    organization_id: uuid.UUID | None
     name: str | None
     max_budget: float | None
     budget_duration_sec: int | None
@@ -90,6 +96,7 @@ class BudgetResponse(BaseModel):
         """Create a BudgetResponse from a Budget ORM model and its usage rollup."""
         return cls(
             budget_id=budget.budget_id,
+            organization_id=budget.organization_id,
             name=budget.name,
             # Narrowed on the way out: the cap is exact in the database, while
             # the wire contract and the dashboard client stay float.
--- a/src/gateway/api/routes/users.py
+++ b/src/gateway/api/routes/users.py
@@ -122,6 +122,26 @@ class UsageLogResponse(BaseModel):
         )
 
 
+def _require_assignable_budget(budget: Budget | None, budget_id: str) -> Budget:
+    """The budget a gateway user may be capped at, or the not-found an unknown id gets.
+
+    An organization's budget is not one. A ``users`` row is deployment-wide with
+    no tenancy column, so pointing one at a tenant's budget lets a tenant admin
+    editing their own figure move what a gateway user may spend, and neither side
+    can see the coupling: the admin cannot reach this surface, and the operator
+    sees no sign the budget they picked is a tenant's. Answered as the 404 this
+    route already gives an id naming nothing, because from the deployment
+    surface's point of view a tenant's budget is not one it may assign, and
+    telling the two apart says nothing useful.
+    """
+    if budget is None or budget.organization_id is not None:
+        raise HTTPException(
+            status_code=status.HTTP_404_NOT_FOUND,
+            detail=f"Budget with id '{budget_id}' not found",
+        )
+    return budget
+
+
 @router.post("")
 async def create_user(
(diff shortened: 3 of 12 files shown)


njbrake (approved):
Verified at main: both assignment sites answered 200 on a tenant-owned budget and `/v1/budgets` carried no ownership signal. Verified on this head: both answer the same 404, and the refused create leaves no user behind. There are three writers of `users.budget_id`, all in `users.py`, and all are covered.

Three follow-ups, none blocking.

`OrganizationMembersPage.tsx:482` offers every budget from the same unfiltered list when setting a workspace-member ceiling, and `scoped_budgets.py:173` validates existence only. A ceiling on an organization's budget saves with a 200, so this coupling is still reachable through `scoped_budgets`. Worth filing before #881 closes, since closing it will read as closing the class.

Legacy rows: no migration is the right call given `b7e1c4a9d2f5` is a day old and unreleased, but the PR should say so. The assignment multiselect is the only dashboard control that writes `users.budget_id`, and it is withheld on the rows that need it while the People column still counts them.

Separately, `DELETE /v1/budgets/{id}` removes a tenant's budget with a 204 and nulls its holders. Pre-existing, and the page now labels those rows while still offering Delete on them.

PR #905, opened 2026-09-02
fix(tools): enforce the web-search max_uses cap in every format

### LINKED ISSUE #594: [BUG] max_uses is accepted and silently ignored on a per-call billed tool
**Otari version:** `main` at 7274000 (read from source, not a released tag)
**Python version:** n/a (static reading of the tool loop)
**Run mode:** standalone
**How you're running Otari:** docker compose
**Provider and model:** any non-Anthropic model with a gateway-run web search

### What happened

A client can declare Anthropic's web-search tool with `max_uses` to bound how many searches a request performs. Otari accepts the declaration and ignores the cap: `max_uses` appears nowhere in `src/`, only in `docs/tools.md`, which states "`max_uses` on an Anthropic-native declaration is accepted but not enforced; `max_tool_iterations` bounds the loop instead."

That matters more than an unimplemented parameter usually would, because gateway-run search is billed per call at `otari:web_search`. A client that sets `max_uses: 1` specifically to bound spend can be charged for as many searches as `max_tool_iterations` permits, with no error and nothing in the response indicating the cap was dropped. Anthropic documents this field as the hard constraint for precisely this purpose, so a client has good reason to trust it.

### Steps to reproduce

1. Configure a web-search backend and turn on `web_search_intercept`.
2. Price `otari:web_search` so calls are billed.
3. `POST /v1/messages` against a non-Anthropic model with `tools: [{"type": "web_search_20250305", "name": "web_search", "max_uses": 1}]` and a prompt that invites several searches.
4. Watch the usage row: `web_search_calls` can exceed 1.

### What you expected

Either the cap is enforced, or the request is rejected so the client learns the gateway will not honour it. Silently accepting a spend limit and not applying it is the one option that can surprise someone at the invoice.

Enforcing looks straightforward given the loop already counts tool iterations: carry the declared `max_uses` per web-search tool into the loop and refuse further calls once it is reached. Anthropic's own behaviour there is to return a `web_search_tool_result` carrying `error_code: max_uses_exceeded`, which would keep an SDK's expectations intact.

### Relevant config (redacted)

```yaml
web_search_url: http://searxng:8080
web_search_intercept: true
pricing:
  otari:web_search:
    input_price_per_million: 10000
```

mikemikimike:
## Description

Enforce a web-search declaration's `max_uses` as a per-request cap on the searches Otari runs. A gateway-run search is billed per successful call, so a cap the gateway accepts and drops can surprise the caller at the invoice.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes #594

## Checklist

- [x] The API contract is unchanged, so OpenAPI regeneration is not required.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:** OpenAI Codex (original commit); Claude Opus 5 via Claude Code (maintainer follow-ups)

## Changes

- The cap is honored on `/v1/messages`, `/v1/chat/completions` and `/v1/responses`, and on every declaration shape including `otari_web_search`. Enforcement bounds spend, so it does not depend on a format's ability to describe a server-side tool call.
- `WebSearchBudget` is built once per request on `ToolContext`, beside `ToolUsageTally`. A multi-attempt request re-runs its searches on fallover and every one is billed, so the cap is spent by the request rather than refilled per attempt.
- Past the cap, an Anthropic-native declaration gets a `web_search_tool_result` carrying `error_code: max_uses_exceeded`; every other caller gets the `[tool error]` string a failed tool already produces. Responses emits no `web_search_call` item for a search that never ran.
- A failed search does not spend the cap, read off the same `[tool error]` sentinel that decides billable.
- `max_uses: 0` is a cap of zero searches rather than the absence of one. An invalid `max_uses` value is rejected with HTTP 400 rather than silently dropping the cap.
- The gateway mints its web-search `server_tool_use` ids with a reserved `otari_` prefix and scrubs an echoed pair by matching on it, the rule #795 established for MCP activity blocks. A provider cannot produce the prefix, so its own blocks round-trip untouched, including a `max_uses_exceeded` from Anthropic's own capped search. The empty `encrypted_content` check stays as a fallback for transcripts minted before the prefix.

## Compatibility

A declaration without `max_uses` is unchanged and stays bounded by `max_tool_iterations`. Provider passthrough is untouched. An MCP server's own tool named `web_search` is not capped, matched on the search backend's result buffer rather than on the name.

## Tests

`make lint`, `make typecheck`, `make openapi-check`, `make postman-check`, the OSS-edition smoke, and the full unit and integration suites pass. The budget, its provenance rule, and the per-request sharing across a streaming fallover are each mutation-tested, so the assertions covering them are known to fail when the behavior regresses.

Rebased onto main. #795 replaced `_execute_stream_owned` with `_execute_stream_owned_events`, so the streaming Messages enforcement is re-derived there: the cap is checked before any MCP activity block is emitted, and native web-search blocks are now gated on `is_error` rather than `transport_error` so a search returning the `[tool error]` sentinel without raising announces nothing.

## Summary

This change enforces `max_uses` limits for gateway-handled web searches across the Messages, Chat Completions, and Responses APIs.

- Applies valid limits, including `max_uses: 0`.
- Shares one search budget across fallback attempts.
- Does not count failed or empty searches.
- Returns format-appropriate `max_uses_exceeded` errors.
- Omits Responses `web_search_call` events for blocked searches.
- Preserves behavior for missing limits, zero limits, provider tools, and unrelated tools; invalid limits are rejected with HTTP 400.
- Adds documentation and tests for route handling, streaming, fallback behavior, mixed tool batches, and error handling.

This prevents unexpected search charges while preserving existing behavior where the limit does not apply.


mikemikimike:
The fork-triggered validation workflows are currently awaiting repository-admin approval (action_required) before they can run. The PR title and template checks pass; once the repository approves the fork run, the test, lint, and typecheck workflows should execute.

src/gateway/api/routes/messages.py
```diff
@@ -139,13 +140,21 @@ def _is_gateway_minted_result(block: Any) -> bool:
     An empty ``content`` list counts as ours: that is what a gateway search with no
     usable hits produces, and a provider reporting no results uses the error shape
     instead.
+
+    The error shape has no ``encrypted_content`` to reason about, so provenance is
```
khaledosman: This paragraph says the `max_uses_exceeded` error block is read as gateway-minted and "dropped anyway", but line 157 returns `False` for any non-list `content`, and the error shape is a dict. `test_a_max_uses_error_result_is_kept_without_gateway_provenance` pins it as kept, so the docstring, the test, and the PR description ("scrubbed from inbound messages with its paired `server_tool_use`") all disagree.

The kept behavior is the one that matters: `_strip_gateway_minted_blocks` runs only under interception, where the outbound `web_search` declaration is gone, so an echoed capped turn ships a fabricated `srvtoolu_<uuid4>` `server_tool_use` plus its result to Anthropic. That is precisely the hazard the function exists to prevent.

Pick one: match on `error_code == "max_uses_exceeded"` (what this paragraph argues for) and flip the test, or delete the paragraph and the description claim.

src/gateway/services/mcp_loop_responses.py
```diff
@@ -576,7 +638,17 @@ def terminal_events(self, state: _ResponsesStreamState, acc: dict[str, Any]) ->
         # client just accumulated, and hand it a call it cannot dispatch.
         hidden = _hidden_call_ids(state)
         folded = _without_output_items(state.deferred_completed, hidden) if hidden else state.deferred_completed
-        folded = _prepend_output_items(folded, acc["compactions"])
+        refusal_items = [
```
khaledosman: This mints a second client-visible gateway item type, `function_call_output`, but `responses.py:177` `_GATEWAY_MINTED_ITEM_TYPES` is still `frozenset({"web_search_call"})`. The matching `function_call` is removed from the same output (`_without_output_items` here, `owned_call_ids` in `filter_owned`), so a client following the documented continuation pattern (append `response.output` to the next `input`) sends back an orphaned `function_call_output` and `_strip_gateway_minted_items` passes it straight through to the provider. That is the hazard the comment at `responses.py:169-177` describes.

Either add `"function_call_output"` to `_GATEWAY_MINTED_ITEM_TYPES`, or do not put the refusal in client-visible output at all. The latter is more consistent with this module's stated position that Responses has no native vocabulary for a gateway-run search, and with `_web_search_items_for` already suppressing the refused search's item.

src/gateway/services/mcp_loop_responses.py
```diff
@@ -385,14 +445,14 @@ def filter_owned(self, result: Response, owned: list[Any], pool: ToolBackend) ->
         owned_call_ids = {item.call_id for item in owned}
         output = list(result.output or [])
         try:
-            result.output = [
+            result.output = cast(Any, [
```
khaledosman: Raw dicts go into `result.output` here, while the streaming path at line 641 builds a proper `ResponseFunctionToolCallOutputItem` for the same refusal. `cast(Any, ...)` is silencing that type error rather than fixing it, and pydantic will not validate the assignment, so the mismatch surfaces later as a serialization warning. Use the typed item in both paths and drop the cast.

src/gateway/services/mcp_loop_responses.py
```diff
@@ -376,7 +435,8 @@ async def execute_owned(
     ) -> list[dict[str, Any]]:
         # ``acc`` is accepted for interface parity and unused: this format has no
         # native vocabulary for a server-side tool call to report on a mixed batch.
-        return await _execute_function_calls(pool, owned)
+        self._mixed_outputs = await _execute_function_calls(pool, owned, budget=self._budget)
```
khaledosman: `self._mixed_outputs` is per-request mutable state, and this line writes it on the module-level `_RESPONSES_STRATEGY` for every uncapped request, contradicting `_strategy_for`'s own invariant ("keeps the shared instance free of per-request state"). `_execute_function_calls` awaits, so two concurrent uncapped requests interleave between `execute_owned` and `filter_owned`; it is latent only because an uncapped request never produces a refusal.

Guard the write on `self._budget is not None`, or better, track refused `call_id`s the way the streaming path already does with `state.refused_call_ids`. That also removes the exact-string comparisons at lines 455 and 630, which currently misread any tool that happens to return `[tool error] max_uses_exceeded` as a refusal.

src/gateway/api/routes/_pipeline.py
```diff
@@ -2044,6 +2053,29 @@ def emit_native_web_search(self) -> bool:
...
+        direction this must not fail; a caller who meant "do not search" is better
+        served by every search being refused than by a bill.
+
+        A ``bool`` is rejected outright rather than counted as its integer value,
+        and so is a negative: a cap of ``True`` or ``-1`` is a caller mistake, and
+        guessing at what it meant would silently answer a spend question they did
+        not ask.
+        """
+        value = (self.web_search_tool_entry or {}).get("max_uses")
+        return value if isinstance(value, int) and not isinstance(value, bool) and value >= 0 else None
```
khaledosman: `-1`, `True` and `"2"` all resolve to `None`, meaning uncapped. The docstring above argues the opposite direction for `0`: "answering a spend limit with unlimited spend ... is the one direction this must not fail". A malformed cap deserves the same treatment, and Anthropic itself 400s an invalid `max_uses`, so under interception the gateway silently accepts what the provider would reject and then spends without a bound. Reject it with a 400 instead of dropping the cap.

src/gateway/services/mcp_loop_messages.py
```diff
@@ -172,13 +207,19 @@ async def _execute_tool_uses(
...
+            continue
         try:
             text = await pool.call_tool(block.name, arguments)
         except Exception as exc:  # noqa: BLE001 — see docstring
             logger.warning("MCP tool %s execution failed: %s", block.name, exc)
             text = f"[tool error] {exc}"
         else:
-            if native_blocks is not None:
+            if capped and budget is not None and not is_tool_error(text):
+                budget.record(text)
```
khaledosman, line 251: `record` already returns early on the `[tool error]` sentinel, and `mcp_loop.py:219` calls it unguarded, so this `not is_tool_error(text)` (and its twin at line 355) is dead and splits one rule across two places. Drop the caller-side check.

Same line: gating `native_blocks.extend(...)` on `not is_tool_error(text)` is an unrelated behavior change. `take_last_results` clears on read, so `_native_blocks_for_call` already returns `[]` after a failed search; if the gate is meant as defense in depth it belongs in a separate PR, and either way the PR description does not mention it.

khaledosman (changes requested):
The cap itself is threaded correctly: all three web-search dispatch sites get the budget, the pool selection is exclusive so the `take_last_results` duck-type never mistakes an MCP `web_search` for the backend's, and spending it per request rather than per attempt matches what is billed.

Two findings block: an echoed capped turn forwards a fabricated `server_tool_use` to Anthropic, and Responses gains a client-visible item type that is not registered as gateway-minted. Both break multi-turn continuation, which is what the two stripping helpers exist to protect.

---

src/gateway/api/routes/messages.py
mikemikimike: Addressed in 9ee26f7. Gateway-minted max_uses_exceeded blocks are now recognized and stripped with their paired server_tool_use; the regression covers the echoed error shape.

src/gateway/services/mcp_loop_responses.py
mikemikimike: Addressed in 9ee26f7 by removing refused function_call_output items from mixed Responses batches; no new client-visible minted item type is emitted.

src/gateway/services/mcp_loop_responses.py
mikemikimike: Addressed in 9ee26f7; the raw refusal output and cast path were removed, keeping Responses output handling consistent.

src/gateway/services/mcp_loop_responses.py
mikemikimike: Addressed in 9ee26f7; shared _mixed_outputs state was removed and refused call IDs are tracked per request.

src/gateway/api/routes/_pipeline.py
mikemikimike: Addressed in 9ee26f7; present but invalid max_uses values now return HTTP 400. Missing remains uncapped and 0 remains valid.

src/gateway/services/mcp_loop_messages.py
mikemikimike: Addressed in 9ee26f7; budget.record() is called unconditionally and owns tool-error filtering, while caller-side is_tool_error and native-block gating were removed.

mikemikimike: Addressed in 9ee26f7; budget.record() is called unconditionally and owns tool-error filtering, while caller-side is_tool_error and native-block gating were removed.

src/gateway/services/mcp_loop_messages.py
```diff
@@ -172,12 +206,18 @@ async def _execute_tool_uses(
...
+            continue
         try:
             text = await pool.call_tool(block.name, arguments)
         except Exception as exc:  # noqa: BLE001 — see docstring
             logger.warning("MCP tool %s execution failed: %s", block.name, exc)
             text = f"[tool error] {exc}"
         else:
+            if capped and budget is not None:
+                budget.record(text)
             if native_blocks is not None:
```
khaledosman: This is the revert I asked for on bad reasoning; please put the gate back here and at line 355.

My claim was that `_native_blocks_for_call` returns `[]` for a failed search. It does not: `take_last_results()` returns `[]`, but `_native_web_search_blocks` then builds `citations = []` and still returns both blocks, so a sentinel-failed search now emits a `server_tool_use` + `web_search_tool_result(content=[])` pair. Per `messages._is_gateway_minted_result`, an empty `content` list means "a gateway search with no usable hits", so a client that speaks the native vocabulary is told the search ran and found nothing, when it actually failed.

The reason to restore it is consistency, not stale results: an *exception*-failing search takes the `except` branch and never reaches this `else`, so it contributes no native blocks at all (`test_failed_search_contributes_no_native_blocks` pins that). Suppressing the sentinel case makes the two failure modes agree.

Worth noting why `test_tool_error_search_contributes_no_native_blocks` was easy to delete: its docstring claimed it was about "stale structured results", and that cannot happen on the real backend, because `WebSearchBackend._search_tool` clears `_last_results` up front specifically so an early return cannot leave the previous call's hits behind. `_FakeSearchPool` did not model that clear, so the test passed for a reason production does not have. Restore the guard with a docstring that says what it actually protects.

The `budget.record(text)` half of my original comment stands, and your change there is right.

src/gateway/api/routes/_pipeline.py
```diff
@@ -1950,6 +1954,15 @@ async def resolve_request_context(
 # ---------------------------------------------------------------------------
 
 
+def _read_web_search_max_uses(entry: dict[str, Any] | None) -> int | None:
+    if entry is None or "max_uses" not in entry:
```
khaledosman: `"max_uses" not in entry` returns `None`, but `{"max_uses": None}` falls through to the raise, so an explicit JSON `null` is now a 400. Anthropic types `max_uses` as an optional `int | None`, and plenty of clients serialize unset optionals as `null` (a Pydantic model dumped without `exclude_none`, for one), so this rejects a request that works today and works against Anthropic directly.

Treat `None` as absent:

```python
if entry is None or entry.get("max_uses") is None:
    return None
```

and drop `None` from the invalid list in `test_a_nonsensical_max_uses_is_rejected_instead_of_becoming_uncapped` and the integration parametrize. The other five values there are genuine caller mistakes and rejecting them is exactly right.

Separately, and non-blocking: `max_web_search_uses` raising `HTTPException` from a property is unusual, though the `prepare_gateway_tools` guard at line 2560 runs before `ToolContext` is built at 2697 and does raise through `adapter.error`, so `/v1/messages` and `/v1/responses` do get their own error envelope. Scoping the check to `_extract_web_search_tool`'s result is the right call: a passthrough provider-named declaration is never validated, so the provider keeps owning its own contract.

khaledosman (approved):
All six findings from the previous review are resolved, and two of them well: dropping the refusal from client-visible output entirely (rather than registering a new minted item type) is the better of the two options I offered, and threading `refused_call_ids` as an explicit out-param removed the shared-singleton state and both `MAX_USES_EXCEEDED_ERROR` string comparisons at once. The `max_uses_exceeded` provenance match is sound: `_strip_gateway_minted_blocks` runs only under interception, where the provider never ran a search of its own, so the inference holds.

**One correction: my native-blocks comment was wrong, and it reverted something correct.** I claimed `_native_blocks_for_call` "already returns `[]` after a failed search". It does not: `_native_web_search_blocks` builds `citations = []` and still returns the full `server_tool_use` + `web_search_tool_result(content=[])` pair. So removing the `not is_tool_error(text)` gate does change behavior, and `2f0b813c` was doing real work. Please restore that half; the `budget.record` half of my comment was right and your fix there is correct. Details inline.

Also inline: `max_uses: null` now 400s, which I think overshoots what I asked for.

Note the test and lint workflows have not reported on this head yet; only the title and template checks have.

---
_Review by Claude Opus 5 via Claude Code._

src/gateway/services/mcp_loop_messages.py
mikemikimike: Addressed in 7e22241. Both synchronous and streaming Messages execution paths now gate native search block creation on a non-error result. Added regression coverage for the [tool error] sentinel; targeted MCP tests pass.

src/gateway/api/routes/_pipeline.py
mikemikimike: Addressed in 7e22241. Explicit max_uses: null is treated like an omitted optional field, while malformed non-null values remain rejected. Updated unit and integration expectations; targeted tests pass.

src/gateway/api/routes/_pipeline.py
```diff
@@ -2044,6 +2063,32 @@ def emit_native_web_search(self) -> bool:
...
+        direction this must not fail; a caller who meant "do not search" is better
+        served by every search being refused than by a bill.
+
+        Invalid values are rejected rather than treated as uncapped, matching the
+        provider contract and preventing malformed spend controls from failing open.
+        """
+        try:
+            return _read_web_search_max_uses(self.web_search_tool_entry)
+        except ValueError as exc:
+            raise HTTPException(
```
khaledosman: `max_uses` is validated twice with two different error envelopes, and this one is unreachable from any route. `prepare_gateway_tools` rejects at `_pipeline.py:2561` through `adapter.error(...)`, and the only production `ToolContext(` is `:2697` inside that same function. Nothing between them touches `max_uses`: the hybrid policy loop at `:2617` writes only `max_results`/`allowed_domains`/`blocked_domains`/`purpose_hint`, and `narrow_web_search_tool_entry` returns a copy rewriting only `max_results`, the domain lists and `provider_options`.

So the bare `HTTPException` can only fire for a direct constructor call, and if it ever did fire on a route it would answer `/v1/messages` and `/v1/responses` with `{"detail": …}` instead of that format's envelope, which is the thing `adapter.error` exists to prevent. Make the property pure (`return _read_web_search_max_uses(self.web_search_tool_entry)`) and leave the 400 to the single site in `prepare_gateway_tools`. `tests/unit/test_chat_request_helpers.py:406` currently pins the unreachable behavior, so it would move to asserting the value rather than the raise.

src/gateway/api/routes/messages.py
```diff
@@ -139,14 +140,20 @@ def _is_gateway_minted_result(block: Any) -> bool:
...
         return False
     hits = block.get("content")
     if not isinstance(hits, list):
-        # The error shape (``web_search_tool_result_error``) is a dict, and only a
-        # provider produces it. Never ours.
-        return False
+        return (
+            isinstance(hits, dict)
+            and hits.get("type") == "web_search_tool_result_error"
+            and hits.get("error_code") == "max_uses_exceeded"
```
khaledosman: This claims *any* `max_uses_exceeded` error block as gateway-minted, but Anthropic emits that same `error_code` when its own native web search hits the cap. The docstring's "the gateway emits only `max_uses_exceeded`" is true outbound; inbound the two are indistinguishable, and this is the inbound direction.

Concretely: a conversation whose earlier turns ran with interception off (or against Anthropic directly) records a genuine provider `max_uses_exceeded` pair. Replay that transcript through this gateway with interception on and the pair is stripped along with its `server_tool_use`, so the model no longer sees that a search was capped and can ask for one again, which the gateway then runs and bills. Low impact, but it is a real widening of the provenance rule the rest of this function is careful about, so worth an explicit decision (accept and say so here, or narrow it).

src/gateway/services/web_search_budget.py
```diff
@@ -0,0 +1,72 @@
...
+        returns the sentinel for an empty query, and neither is billed.
+        ``max_tool_iterations`` is what bounds a model that keeps retrying a broken
+        backend.
+        """
+        if is_tool_error(result):
+            return
+        self._remaining -= 1
+
+
+def is_capped_search(budget: WebSearchBudget | None, pool: ToolBackend | Any, name: str) -> bool:
```
khaledosman, line 60: This already returns `False` for `budget is None`, so the `and budget is not None` at every call site (`mcp_loop.py:208`, `mcp_loop_messages.py:210` and `:344`, `mcp_loop_responses.py:106` and `:280`) is there only to narrow the optional for mypy, which a `bool` return cannot do.

Returning `WebSearchBudget | None` (the active budget, or `None`) narrows and collapses each site to one condition:

```python
active = active_search_budget(budget, pool, name)
if active is not None and active.exhausted():
    ...
```

and the `record` call below becomes `active.record(text)` with no second check.

docs/tools.md
```diff
@@ -35,6 +35,24 @@ Interception is off by default because enabling it changes who performs searches
...
+### Bounding the searches one request runs
+
+A declaration may carry `max_uses` to cap how many searches Otari runs for that
+request. Because a gateway-run search is billed per successful call, the cap is
+honored on every endpoint (`/v1/messages`, `/v1/chat/completions`,
+`/v1/responses`) and on every declaration shape, including `otari_web_search`.
+A failed search does not count against it, matching what is billed.
+
+`max_uses: 0` is a cap of zero searches, not the absence of a cap, so every
+search is refused. A negative or non-numeric value is a caller mistake and is
```
khaledosman: "non-numeric" understates what is rejected: `1.5` and `5.0` both 400, since the check is `isinstance(value, int)`. `5.0` is the interesting case, because the same body passes through to Anthropic (which accepts it) when interception is off, so one request is valid or invalid depending on a deployment toggle it cannot see. Either accept an integral `float` in `_read_web_search_max_uses` or say "non-integer" here.

src/gateway/api/routes/responses.py
```diff
@@ -354,11 +360,16 @@ def open_tool_loop_stream(
         max_iterations: int,
         *,
         emit_native_web_search: bool = False,
+        web_search_budget: WebSearchBudget | None = None,
```
khaledosman, line 363: `test_messages_route_dispatch.py:1137` asserts the budget arrives at the loop with the caller's exact cap for `/v1/messages`, which is the right shape of test. Nothing does that for `/v1/responses`, and both streaming adapters (`chat.py:266` and here) are only reached through the loop functions the unit tests call directly, so this `extra` plumbing is untested. One test in the same shape as the messages one would cover the claim that the cap is honored on all three endpoints.

khaledosman (approved):
Verified on the head revision: `make check-architecture` clean, `ruff check` clean, `mypy` clean, and the 175 unit tests across the five touched files pass. The two structural claims hold up under tracing: the budget is one object per request across fallback attempts, and `advance_stream_transcript` runs before `synthetic_events` (`_tool_loop.py:340` then `:344`), so a refused Responses call cannot leak a `web_search_call` item. Skipping `_native_blocks_for_call` on a tool error does not strand the backend's result buffer either, since `call_tool` clears it on entry.

Nothing blocking. Findings inline: one duplicated validation with mismatched error envelopes, one provenance question on the inbound stripping, and a coverage gap on `/v1/responses`.

PR #909, opened 2026-09-02
fix(settings): prevent database_url value from overflowing Settings card

### LINKED ISSUE #809: [BUG] A long database_url overflows its Settings row and squeezes the description column

AloysJehwin:
## Description

A long startup-only value such as `database_url` overflowed the Settings page card and forced the description column to wrap to one word per line.

Root cause: `ConfigRow` wrapped `SettingControl` in `shrink-0`, preventing the column from ever narrowing below its max-content width. `break-all` on a descendant of a non-shrinkable column has no effect because `max-content` ignores soft wrap opportunities.

Fix: replace `shrink-0` with `min-w-0` on the value column so it can shrink, add `flex-wrap` on the row (matching sibling rows at lines 457 and 505), and break the `<span>` across lines per Biome's formatting rules. Removed the redundant `title` tooltip — `break-all` already makes the full value visible in place.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes #809.

## Checklist

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:** N/A

**Any additional AI details you'd like to share:** N/A

### DIFF
--- a/web/src/features/settings/SettingsPage.tsx
+++ b/web/src/features/settings/SettingsPage.tsx
@@ -190,10 +190,11 @@ function SettingControl({
   disabled: boolean
 }) {
   if (!field.settable) {
+    const display = formatValue(field)
     return (
-      <div className="flex items-center gap-2">
-        <span className="text-body tabular-nums">{formatValue(field)}</span>
-        <span className="rounded-full border border-border px-2 py-0.5 text-caption">
+      <div className="flex min-w-0 items-center gap-2">
+        <span className="min-w-0 break-all text-body tabular-nums" title={display}>{display}</span>
+        <span className="shrink-0 rounded-full border border-border px-2 py-0.5 text-caption">
           startup-only
         </span>
       </div>


web/src/features/settings/SettingsPage.tsx
```diff
@@ -190,10 +190,11 @@ function SettingControl({
   disabled: boolean
 }) {
   if (!field.settable) {
+    const display = formatValue(field)
     return (
-      <div className="flex items-center gap-2">
-        <span className="text-body tabular-nums">{formatValue(field)}</span>
-        <span className="rounded-full border border-border px-2 py-0.5 text-caption">
+      <div className="flex min-w-0 items-center gap-2">
```
khaledosman, line 195: This does not fix the overflow. `ConfigRow` wraps `SettingControl` in `<div className="shrink-0 pt-0.5">` at line 266, and with `flex-shrink: 0` that column keeps its max-content width. `max-content` is defined as the size with no line wrapping, so it ignores the break opportunities `break-all` introduces, and `min-w-0` on a descendant of a column that cannot shrink has nothing to act on.

Measured in Chromium with this exact class structure, a 640px card at a 900px viewport: before and after are identical, value column 1035px, description column 0px, card overflow 395px. Changing line 266 from `shrink-0` to `min-w-0` fixes it: value column 444px, description column 172px, overflow 0. Keep the `break-all` here and the badge's `shrink-0`; the settable controls survive a shrinkable column, since `NumberSetting` and `TextSetting` set `w-28` and `w-56` and `Toggle` carries its own `shrink-0`.

One more at 390px: even with the shrink fixed, the description column lands at ~91px, because `ConfigRow` does not wrap. Adding `flex-wrap` at line 259, as the sibling rows at lines 456 and 504 already do, gives both columns the full width instead. `SettingsPage.test.tsx` is the place for the narrow-viewport regression.

web/src/features/settings/SettingsPage.tsx
```diff
@@ -190,10 +190,11 @@ function SettingControl({
...
+        <span className="min-w-0 break-all text-body tabular-nums" title={display}>{display}</span>
```
khaledosman, line 196: CI blocker: `pnpm run lint` fails here, Biome wants the attributes broken across lines. `pnpm --dir web run lint:fix`.

khaledosman, line 196: Drop `title={display}`. Nothing truncates in this branch: `break-all` wraps the value in place, so the tooltip repeats text that is already fully visible. A native `title` is also hover-only, which makes it dead weight on the 390px touch viewport the screenshot matrix captures.

khaledosman (changes requested):
The `dashboard` check is red on a Biome format error, and the layout change does not fix #809. Details inline, with the measurement behind the second claim. CodeRabbit reached the same root cause at `:196`.

_Review written by Claude Code (Opus 5)._

AloysJehwin:
Fixed Biome format errors and corrected the layout per review feedback.

Changes in commit b37761e:
- Replaced `shrink-0` with `min-w-0` on the `ConfigRow` value column so long values like `database_url` can actually shrink (the previous `break-all` had no effect since `max-content` ignores soft wrap opportunities in a `shrink-0` column)
- Added `flex-wrap` to the `ConfigRow` row, matching sibling rows at lines 457 and 505
- Broke the read-only value `<span>` across lines to satisfy Biome's formatting check
- Removed the `title={display}` tooltip — with `break-all` the full value is already visible in place

The branch is up to date and Biome passes clean. Could a maintainer reopen this PR? GitHub is preventing me from reopening it programmatically after the force-push that followed the auto-close.

PR #911, opened 2026-09-02
fix(db): hand the request's connection back before the provider call

njbrake:
## Description

The standalone preamble reads on the request-scoped session from `get_db`, and nothing commits before dispatch, so that connection stayed checked out for the whole upstream call. `db_pool_size + db_max_overflow` was therefore a ceiling on concurrent provider calls rather than on concurrent database work. Past it a request waits `db_pool_timeout` and is then refused by the auth dependency, which reports a pool timeout as "Authentication temporarily unavailable, please retry". A gateway-run tool loop reaches the ceiling sooner, since one request spans several provider calls plus each tool call.

Reproduced against a stub upstream holding each call 20s: 80 concurrent requests leave 30 connections `idle in transaction`, and 50 are refused after 30s while the log shows `QueuePool limit of size 10 overflow 20 reached`. Also observed on a self-hosted deployment running the defaults, where it presented as an authentication outage.

Two changes follow from the same failure. Usage logging gets a small pool of its own, because a saturated request pool timed the writer out too and a dropped usage row is unrecoverable. And the database path gets the timeouts it had none of, plus a finite `db_pool_recycle` default: a managed database or the NAT in front of it drops idle connections without a FIN, and `pool_pre_ping` is meant to catch that, but the ping is itself a statement and on a half-open socket it blocks on TCP retransmission rather than failing.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

None filed.

## Checklist

- [ ] If the API contract changed, I regenerated the OpenAPI spec (`uv run python scripts/generate_openapi.py`). Ran it; the new settings are reported through the existing dynamic `/v1/settings` payload, so the spec is unchanged.

`make lint`, `make typecheck` and the full `tests/unit` suite pass locally. `tests/integration` could not finish locally (the dev machine's volume filled up, unrelated to the change), so CI is the gate for that suite; it is green here.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:** Claude Opus 5, via the Claude Code CLI.

The investigation started from a downstream service seeing multi-minute latency and 503s from this gateway. The diagnosis, the reproduction and this patch were produced by the model in discussion with @njbrake; the reasoning and the decision to fix it this way are his.

## Summary

- Release request database connections before upstream provider calls, including pass-through routes.
- Add a dedicated usage-log connection pool for PostgreSQL.
- Add connection, command, and statement timeout settings with validation.
- Normalize database errors across request, authentication, gateway-tool, and usage-log paths.
- Improve SQLite handling by reusing one engine.
- Add rollback-on-failure transaction handling.
- Set a finite default for connection recycling.
- Update configuration and data-layer documentation.
- Replace source-based metering checks with behavioral tests and add coverage for timeout handling, pool behavior, and connection release.

These changes reduce connection starvation and improve database reliability during long provider calls.


khaledosman (commented):
The connection release itself is right, and `expire_on_commit=False` plus the no-`relationship()` rule do make the commit safe. Two things need attention before merge: the new `command_timeout` changes what a database call raises and the codebase's ~81 `except SQLAlchemyError` arms do not catch it, and the fix covers `chat`/`messages`/`responses` but not the pass-through inference routes, which have the identical hold.

🤖 Review written by Claude Opus 5 via the Claude Code CLI, acting as an agent.

src/gateway/core/database.py
```diff
@@ -107,29 +109,86 @@ def _set_sqlite_pragma(dbapi_connection: Any, _: Any) -> None:  # noqa: ANN001,
...
+
+    kwargs["pool_size"] = config.db_pool_size if pool_size is None else pool_size
+    kwargs["max_overflow"] = config.db_max_overflow if max_overflow is None else max_overflow
+    kwargs["pool_timeout"] = config.db_pool_timeout
+    if config.db_pool_recycle >= 0:
+        kwargs["pool_recycle"] = config.db_pool_recycle
+
+    connect_args["timeout"] = config.db_connect_timeout
+    if config.db_command_timeout > 0:
+        connect_args["command_timeout"] = config.db_command_timeout
```
khaledosman: Setting `command_timeout` changes what a database call can raise, and nothing here widens the handlers that catch it. asyncpg raises a bare `asyncio.TimeoutError` on a command timeout, and SQLAlchemy's `_asyncpg_error_translate` maps only asyncpg exception classes, so it propagates untranslated and is not a `SQLAlchemyError`.

There are ~81 `except SQLAlchemyError` arms under `src/gateway/`, and the ones on this path now have a hole:

- `api/deps.py:138` exists to turn a database failure into 503 "Authentication temporarily unavailable"; a timed-out auth lookup now escapes it as an unhandled 500.
- `api/routes/_pipeline.py:2627` stops running, so `users.reserved` keeps the hold until the reservation sweep, which is exactly the leak that arm was written to prevent.
- `services/log_writer.py`'s two arms drop the usage row without the rollback or the failure metric.

Translate it once at the boundary, or widen the arms on the request path to `(SQLAlchemyError, TimeoutError)` in this PR.

src/gateway/core/database.py
```diff
@@ -107,29 +109,86 @@ def _set_sqlite_pragma(dbapi_connection: Any, _: Any) -> None:  # noqa: ANN001,
...
+    if config.db_statement_timeout_ms > 0:
```
khaledosman, line 242: The server-side backstop defaults to exactly the client-side timeout it backstops (60000 ms vs 60.0 s), so which one fires is a race, and the bare-`TimeoutError` path above is reached about as often as the translated one. Default `db_statement_timeout_ms` above `db_command_timeout * 1000` (e.g. 65000), or validate the ordering at config load the way the other cross-field settings are.

src/gateway/core/database.py
```diff
@@ -107,29 +109,86 @@ def _set_sqlite_pragma(dbapi_connection: Any, _: Any) -> None:  # noqa: ANN001,
...
+            connect_args=connect_args,
+            is_sqlite=is_sqlite,
+            pool_size=config.db_log_pool_size,
+            max_overflow=0,
+        ),
+    )
+    _LogSessionLocal = async_sessionmaker(_log_engine, expire_on_commit=False)
+
     if is_sqlite:
         _configure_sqlite_pragmas(_engine)
```
khaledosman, line 271: `_configure_sqlite_pragmas` is applied to `_engine` only, so `_log_engine`'s SQLite connections get `busy_timeout=0` and `foreign_keys` off. The writer now contends with the request engine on the same file with no busy timeout, which is the intermittent "database is locked" failure that function's own docstring says it exists to remove. SQLite is the default for a bare `otari` run and for `scripts/oss_edition_smoke.py`.

Either apply the pragmas to both engines, or skip the second engine entirely when `is_sqlite`: it gains nothing there, since NullPool ignores `db_log_pool_size`.

src/gateway/core/database.py
```diff
@@ -107,29 +109,86 @@ def _set_sqlite_pragma(dbapi_connection: Any, _: Any) -> None:  # noqa: ANN001,
...
+    )
     _SessionLocal = async_sessionmaker(_engine, expire_on_commit=False)
 
+    # Usage logging gets a pool of its own. It shares the request pool's
+    # database, but not its contention: a saturated request pool used to time
+    # the writer out too, and a dropped usage row is unrecoverable. Metering is
+    # how spend, budgets and the activity log are reconstructed afterwards, so
+    # it must not be the first thing a busy gateway loses. Small and with no
+    # overflow, because the writer batches and is never a source of bursts.
+    _log_engine = create_async_engine(
```
khaledosman: Both engines are handed the same `connect_args` object, and `engine_kwargs` mutates it after the first engine has already captured the reference. The values written are identical today so there is no bug, but pass a copy per engine so a future per-pool timeout cannot silently rewrite the request engine's args.

src/gateway/core/database.py
```diff
@@ -160,29 +219,52 @@ async def create_session() -> AsyncIterator[AsyncSession]:
         yield session
 
 
+@asynccontextmanager
+async def create_log_session() -> AsyncIterator[AsyncSession]:
+    """Session for the usage-log writer, on the metering pool.
+
+    Falls back to the request pool when :func:`init_db` has not run, which is
```
khaledosman, line 333: `src/gateway/AGENTS.md:189` still says request code gets a session through `get_db` and non-request code uses `create_session()`. There is a third factory now; add it there or that paragraph goes stale.

src/gateway/api/routes/_pipeline.py
```diff
@@ -3023,6 +3028,46 @@ async def release_reservation(ctx: RequestContext) -> None:
     await refund_reservation(ctx.db, ctx.reservation)
 
 
+async def release_request_connection(ctx: RequestContext) -> None:
```
khaledosman: This covers `chat`, `messages` and `responses`, but the pass-through inference routes have the identical hold and are untouched. `api/routes/_passthrough.py:409` reads `resolve_request_allowlist` on the request session, nothing commits, and `call_provider` runs at :491 with the connection still checked out. That keeps the whole `db_pool_size + db_max_overflow` ceiling in place for embeddings, images, audio, rerank, moderations, files and search, and as the comment above `track_request` already notes, an image generation routinely runs longer than a completion does.

Call this before `track_request` in `run_passthrough` too.

src/gateway/api/routes/_pipeline.py
```diff
@@ -3023,6 +3028,46 @@ async def release_reservation(ctx: RequestContext) -> None:
...
+
+    Best-effort: a request that has reached the provider must not fail because
+    the commit did. A poisoned session would already have been rolled back by
+    the caller's ``SQLAlchemyError`` arm.
+    """
+    if ctx.db is None:
+        return
+    try:
+        await ctx.db.commit()
+    except SQLAlchemyError:
```
khaledosman: `except SQLAlchemyError` does not catch the `TimeoutError` that this PR's own `db_command_timeout` introduces (see the note on `core/database.py:148`), so the docstring's "a request that has reached the provider must not fail because the commit did" is not true as written: a commit that hits the 60 s client-side timeout kills a request that already passed every gate. Catch `(SQLAlchemyError, TimeoutError)`.

tests/unit/test_usage_log_metering_pool.py
```diff
@@ -0,0 +1,48 @@
...
+from gateway.core.config import GatewayConfig
+from gateway.core.database import init_db, reset_db
+from gateway.services import log_writer as log_writer_module
+from gateway.services.log_writer import BatchLogWriter, SingleLogWriter
+
+
+def test_writers_use_the_metering_session_factory() -> None:
+    source = Path(log_writer_module.__file__).read_text()
+    # Both writers, and no path left on the request pool.
+    assert source.count("create_log_session()") == 2
```
khaledosman: This asserts on the module's source text: it breaks when someone adds a comment or a third call site, and it proves nothing about what the writers actually do at runtime. `tests/unit/test_log_writer.py` already has the pattern this wants: monkeypatch `create_log_session` and assert the writer drew from it. Do that for `BatchLogWriter._flush` and drop the grep.

tests/unit/test_usage_log_metering_pool.py
```diff
@@ -0,0 +1,48 @@
...
+        async with database.create_log_session() as db:
+            assert (await db.execute(text("SELECT 1"))).scalar_one() == 1
+    finally:
+        reset_db()
+
+
+@pytest.mark.asyncio
+async def test_writers_construct_without_init_db() -> None:
+    # The fallback in create_log_session: tests build these directly.
+    assert isinstance(SingleLogWriter(), SingleLogWriter)
```
khaledosman: The comment says this covers `create_log_session`'s fallback, but the test only constructs the writers and never calls `put` or `_flush`, so the fallback is never reached; `isinstance(SingleLogWriter(), SingleLogWriter)` holds unconditionally. Exercise the fallback: `reset_db()`, set `_SessionLocal` alone, then assert `create_log_session` yields from it.

src/gateway/core/database.py
njbrake: Confirmed and fixed. `TimeoutError` on 3.11+ is `builtins.TimeoutError`, whose MRO misses every key in `_asyncpg_error_translate`, so it propagated untranslated; verified against a live server that `command_timeout` yields a bare `builtins.TimeoutError` with `isinstance(exc, SQLAlchemyError)` false. Translated once per engine via a `handle_error` listener. Connect is not reachable that way, since SQLAlchemy only wraps DBAPI errors raised while opening a connection, so the arms on the request path (both auth arms, `_bump_last_used_at`, `prepare_gateway_tools`, and both writer arms) now catch `DATABASE_ERRORS`.

src/gateway/core/database.py
njbrake: Fixed. `db_statement_timeout_ms` defaults to 65000 against a 60s `db_command_timeout`, and a `model_validator` rejects a configuration where the server-side backstop would not fire after the client-side ceiling.

src/gateway/core/database.py
njbrake: Took the second option: no metering engine on SQLite. `_LogSessionLocal` aliases `_SessionLocal` there, so the writer runs on the pragma-configured engine and `NullPool` makes the separate pool pointless anyway. Covered by `test_sqlite_keeps_one_engine`.

src/gateway/core/database.py
njbrake: Fixed. `engine_kwargs` now copies `connect_args` and returns the copy, so neither engine can rewrite the other's arguments. `test_the_callers_connect_args_are_never_mutated` pins it.

src/gateway/core/database.py
njbrake: Updated, and it needed two lines rather than one: the third factory, and the fact that handlers on the request path catch `DATABASE_ERRORS` rather than `SQLAlchemyError` alone now that a bare `TimeoutError` is reachable.

src/gateway/api/routes/_pipeline.py
njbrake: Good catch, fixed. `run_passthrough` releases before `track_request`. The existing pass-through route tests exercise the new call; I did not add a unit test for it because the ordering assertion needs the whole preamble stubbed, which buys less than the integration coverage already there.

src/gateway/api/routes/_pipeline.py
njbrake: Fixed. The primitive moved to `core/database.py` as `release_session`, which catches `DATABASE_ERRORS` and, per the other bot's note on the same line, rolls the session back on a failed commit so settlement after dispatch does not meet a `PendingRollbackError`.

tests/unit/test_usage_log_metering_pool.py
njbrake: Agreed, dropped. Both writers are now driven through a stubbed `create_log_session` and asserted on what they sent it, following `test_log_writer.py`. `BatchLogWriter._flush` is covered too.

tests/unit/test_usage_log_metering_pool.py
njbrake: Right, it asserted nothing. Replaced with a test that resets, leaves `_SessionLocal` alone, clears `_LogSessionLocal`, and asserts `create_log_session` still yields a working session, plus one that a process with neither raises.

src/gateway/api/routes/_pipeline.py
```diff
@@ -3023,6 +3028,46 @@ async def release_reservation(ctx: RequestContext) -> None:
...
+        logger.warning("Could not release the request's database connection before dispatch", exc_info=True)
```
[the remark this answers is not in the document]
njbrake: Fixed. The primitive moved to `core/database.py` as `release_session`, which rolls the session back when the commit fails so the settlement writes after dispatch do not meet a `PendingRollbackError`. That also answers the layering nitpick: the commit boundary now sits in the database layer rather than in the route pipeline.

src/gateway/core/database.py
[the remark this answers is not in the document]
njbrake: Fixed by dropping the second engine on SQLite entirely rather than duplicating the pragmas. `_LogSessionLocal` aliases `_SessionLocal` there, so the writer uses the pragma-configured connection, and `NullPool` made the separate pool pointless on that backend anyway.

njbrake:
All nine review points are in `26a3307`, with a reply on each thread.

Two changed the shape of the fix. The `command_timeout` hole was real: against a live server, asyncpg raises `builtins.TimeoutError` and `isinstance(exc, SQLAlchemyError)` is false, so statement timeouts are now translated per engine and the request-path arms catch `DATABASE_ERRORS`, since connect timeouts stay outside the listener's reach. And the pass-through routes did carry the identical hold, so they release before `track_request`.

`make lint`, `make typecheck` and 2833 unit tests pass locally.

PR #912, opened 2026-09-03
fix(api): refuse a blank provider narrowing on POST /v1/scoped-budgets

### LINKED ISSUE #878: POST /v1/scoped-budgets accepts a blank provider narrowing, storing a ceiling that never binds
Adjacent finding from the PRs for mozilla-ai/otari-ai#1943. Two of the three sibling cases were fixed there (#875); this is the one left alone deliberately, because fixing it narrows a published operator contract.

## The bug

`CreateScopedBudgetRequest.provider_key_id` on `POST /v1/scoped-budgets` is length-checked but not constrained:

```python
provider_key_id: str | None = Field(
    default=None,
    max_length=255,
    description="Narrow the cap to one provider instance; null caps spend across every provider",
)
```

NULL means "cap spend across every provider" and a value means "cap only this instance". An empty or whitespace-only string is neither, and both halves of the design fail open at once:

- It is `NOT NULL`, so the partial unique index `uq_scoped_budgets_scope_no_key` (declared `WHERE provider_key_id IS NULL`) does not cover it, and the "one aggregate cap per scope" rule silently stops applying to that row.
- `applicable_budgets` matches `provider_key_id == provider_instance OR provider_key_id IS NULL`. A blank string satisfies neither, so the ceiling never binds.

The result is a ceiling that is created, returned by `GET /v1/scoped-budgets`, and enforced against nothing, with no error at any point — the same created-listed-and-silently-unenforced failure `_require_scope_exists` refuses an unresolvable scope for, and it fails in the permissive direction.

## Why it is lower priority than the two already fixed

This router is gated on `require_deployment_operator` at the router level, so only a deployment operator can reach it, and that caller can also list and delete the dud row. The two fixed in #875 were worse: `OrganizationScopedBudgetCreate` is reachable by an organization owner or admin, and `WorkspaceMemberBudgetPolicyCreate` is reachable by the same audience *and* fans the bad value out into one ceiling per workspace member.

## Suggested fix

Mirror what #875 applied to both tenant-facing models:

```python
provider_key_id: str | None = Field(
    default=None,
    min_length=1,
    max_length=255,
    pattern=r"^\S+$",
    ...
)
```

Refuse blanks rather than coercing them to NULL: NULL is the *wider* cap, so coercion would silently cap more than the caller asked for. Whitespace is excluded outright because an instance name cannot contain any — the model-key pattern in `organization_pricing` already assumes that in its provider half, since the allow-list is keyed on `"{instance}:{model}"`.

## The reason this is not in #875

The change turns a previously accepted input into a 422 on a public operator endpoint. That is a small but real contract narrowing, so it belongs in a change that owns that surface rather than riding along in one about the tenant-scoped surface. A test asserting the blank and whitespace-only cases are refused should come with it, since neither shows up as an error otherwise.

champ18ion:
## Description
`CreateScopedBudgetRequest.provider_key_id` on `POST /v1/scoped-budgets` was length-checked but not constrained, so an empty or whitespace-only value was accepted. NULL means "cap spend across every provider" and a value means "cap only this instance"; a blank string satisfies neither:

- It is `NOT NULL`, so the partial unique index `uq_scoped_budgets_scope_no_key` (declared `WHERE provider_key_id IS NULL`) does not cover it, and the "one aggregate cap per scope" rule silently stops applying to that row.
- `applicable_budgets` matches `provider_key_id == provider_instance OR provider_key_id IS NULL`. A blank string satisfies neither, so the ceiling never binds.

The result: a ceiling created, listed, and enforced against nothing, with no error anywhere — the same permissive-direction failure `test_a_ceiling_on_a_scope_that_does_not_exist_is_refused` already guards against for an unresolvable scope.

Mirrors the fix #875 applied to the two tenant-facing siblings (`OrganizationScopedBudgetCreate`, `WorkspaceMemberBudgetPolicyCreate`): `min_length=1` plus `pattern=r"^\S+$"`, refusing rather than coercing to null, since null is the *wider* cap and coercion would silently cap more than the caller asked for. This router is gated on `require_deployment_operator`, reached by a smaller audience than those two, which is why #878 filed it separately and lower priority rather than folding it into #875.

Also regenerated `docs/public/openapi.json` for the changed field (the Postman collection doesn't embed this field's schema, so it didn't need regenerating).

## PR Type

- [x] Bug Fix

## Relevant issues
Fixes #878

## Checklist

- [x] I ran the Definition of Done checks locally (`make lint`, `make typecheck`, `make test` — ran `ruff check` directly and confirmed it's clean; verified the Pydantic validation logic in isolation (blank/whitespace/tab all rejected, `None` and a real value accepted) since I couldn't run the full suite through `uv sync` on this machine, same platform constraint as #699).
- [x] If the API contract changed, I regenerated the OpenAPI spec (`uv run python scripts/generate_openapi.py`) — done by hand, matching the already-correct sibling schema exactly (confirmed identical shape/description text).

## AI Usage

**Any additional AI details you'd like to share:** Picked this up off the issue directly — it's fully specified with the exact fix and a pointer to the sibling pattern already in the codebase, so there was no design decision to make, just verifying the mirror was exact and adding the matching test and OpenAPI update. I confirmed the validation behavior directly before opening this rather than trusting the diff alone.

## Summary

- Reject blank or whitespace-only `provider_key_id` values for `POST /v1/scoped-budgets`.
- Preserve omitted or `null` values as aggregate provider caps.
- Add tests for invalid and aggregate inputs.
- Update the OpenAPI documentation and generated client schema.

This prevents ceilings that never apply to provider spending.

### DIFF
--- a/src/gateway/api/routes/scoped_budgets.py
+++ b/src/gateway/api/routes/scoped_budgets.py
@@ -47,10 +47,20 @@ class CreateScopedBudgetRequest(BaseModel):
         max_length=255,
         description="Id of the capped identity: an organization, workspace, membership row, or API key",
     )
+    # Absent means every provider; a present value must name a real instance.
+    # Resolution matches `provider_key_id == provider_instance OR IS NULL`, and a
+    # blank string is neither, so it would store, list, and never bind. Refused
+    # rather than folded into null, because null is the *wider* cap and coercing
+    # would silently cap more than the caller asked for.
     provider_key_id: str | None = Field(
         default=None,
+        min_length=1,
         max_length=255,
-        description="Narrow the cap to one provider instance; null caps spend across every provider",
+        pattern=r"^\S+$",
+        description=(
+            "Narrow the cap to one provider instance; omit or null to cap spend across every provider. "
+            "Must name a real instance: a blank value would store a ceiling that never binds"
+        ),
     )
     budget_id: str = Field(
         min_length=1,
--- a/src/gateway/api/routes/scoped_budgets.py
+++ b/src/gateway/api/routes/scoped_budgets.py
@@ -47,11 +47,13 @@ class CreateScopedBudgetRequest(BaseModel):
         max_length=255,
         description="Id of the capped identity: an organization, workspace, membership row, or API key",
     )
-    # Absent means every provider; a present value must name a real instance.
-    # Resolution matches `provider_key_id == provider_instance OR IS NULL`, and a
-    # blank string is neither, so it would store, list, and never bind. Refused
-    # rather than folded into null, because null is the *wider* cap and coercing
-    # would silently cap more than the caller asked for.
+    # Absent means every provider. Resolution matches `provider_key_id ==
+    # provider_instance OR IS NULL`, and a blank string is neither, so it would
+    # store, list, and never bind. Refused rather than folded into null, because
+    # null is the *wider* cap and coercing would silently cap more than the
+    # caller asked for. This only refuses the blank/whitespace shape, not an
+    # unresolvable value: unlike `scope_id` and `budget_id` above, a present
+    # `provider_key_id` naming no configured instance is not checked here (#918).
     provider_key_id: str | None = Field(
         default=None,
         min_length=1,
@@ -59,7 +61,8 @@ class CreateScopedBudgetRequest(BaseModel):
         pattern=r"^\S+$",
         description=(
             "Narrow the cap to one provider instance; omit or null to cap spend across every provider. "
-            "Must name a real instance: a blank value would store a ceiling that never binds"
+            "A blank value would store a ceiling that never binds, so it is refused; this does not check "
+            "that the value names a configured provider instance"
         ),
     )
     budget_id: str = Field(
--- a/docs/public/openapi.json
+++ b/docs/public/openapi.json
@@ -2942,13 +2942,15 @@
             "anyOf": [
               {
                 "maxLength": 255,
+                "minLength": 1,
+                "pattern": "^\\S+$",
                 "type": "string"
               },
               {
                 "type": "null"
               }
             ],
-            "description": "Narrow the cap to one provider instance; null caps spend across every provider",
+            "description": "Narrow the cap to one provider instance; omit or null to cap spend across every provider. Must name a real instance: a blank value would store a ceiling that never binds",
             "title": "Provider Key Id"
           },
           "scope_id": {
--- a/docs/public/openapi.json
+++ b/docs/public/openapi.json
@@ -2950,7 +2950,7 @@
                 "type": "null"
               }
             ],
-            "description": "Narrow the cap to one provider instance; omit or null to cap spend across every provider. Must name a real instance: a blank value would store a ceiling that never binds",
+            "description": "Narrow the cap to one provider instance; omit or null to cap spend across every provider. A blank value would store a ceiling that never binds, so it is refused; this does not check that the value names a configured provider instance",
             "title": "Provider Key Id"
           },
           "scope_id": {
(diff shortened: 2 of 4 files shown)


src/gateway/api/routes/scoped_budgets.py
```diff
@@ -47,10 +47,20 @@ class CreateScopedBudgetRequest(BaseModel):
...
+    # would silently cap more than the caller asked for.
     provider_key_id: str | None = Field(
         default=None,
+        min_length=1,
         max_length=255,
-        description="Narrow the cap to one provider instance; null caps spend across every provider",
+        pattern=r"^\S+$",
+        description=(
+            "Narrow the cap to one provider instance; omit or null to cap spend across every provider. "
+            "Must name a real instance: a blank value would store a ceiling that never binds"
```
khaledosman: The description promises more than the schema enforces. "Must name a real instance" reads as an existence guarantee, but only blanks are refused: `provider_key_id="gemini-typo"` still stores a ceiling that is created, listed, and binds to nothing. That is the same permissive-direction failure `_require_scope_exists` and `_require_budget` refuse for the scope and the budget in this very handler, which makes the gap more visible here than in the two #875 siblings.

Either resolve the name against the configured provider instances alongside those two checks, or drop the "Must name a real instance" clause so the wording matches the constraint. Since the same sentence is already in `organization_budget_service.py` and `workspace_budget_default_service.py`, a follow-up issue covering all three is a reasonable way to take it.

src/gateway/api/routes/scoped_budgets.py
```diff
@@ -47,10 +47,20 @@ class CreateScopedBudgetRequest(BaseModel):
         max_length=255,
         description="Id of the capped identity: an organization, workspace, membership row, or API key",
     )
+    # Absent means every provider; a present value must name a real instance.
```
khaledosman: Third verbatim copy of this comment plus field, after `organization_budget_service.py:248` and `workspace_budget_default_service.py:76`. Worth hoisting to one shared `ProviderNarrowing = Annotated[str | None, Field(...)]` and letting the three call sites name it, so the rationale cannot drift between copies (routes may import services, so a service-side home stays within the layer rules).

khaledosman (approved):
Verified the gates rather than the description: `make openapi-check` and `make postman-check` both pass on the head, and the collection genuinely does not embed this field's description, so the claim about Postman holds. `PATCH /v1/scoped-budgets/{id}` does not expose `provider_key_id`, so the create route is the whole surface. `SpendCeilingDialog.tsx:129` already folds a blank to null client-side, so the dashboard does not regress. The four new cases pass locally against Testcontainers PostgreSQL.

Two notes inline, neither blocking.

_Review written by Claude Code (Opus 5)._

champ18ion:
Both addressed.

Wording: reworded the comment and description to say what's actually enforced (blank/whitespace refused, existence not validated), rather than fix by adding real validation, since that's a bigger scope than this PR. Filed #918 to track adding it, or fixing the same overclaim in the two #875 siblings if it isn't, plus your hoist-to-a-shared-type suggestion.

Duplication: agreed it's worth doing, tracked in #918 too rather than touching organization_budget_service.py / workspace_budget_default_service.py from this PR, since neither is otherwise in scope here.

njbrake (approved):
Approving.

Every writer of `provider_key_id` is now constrained: this route, `organization_budget_service.py:253`, and `workspace_budget_default_service.py:85`. No update path exposes the field, so the create routes are the whole surface. The three blank cases fail on `origin/main` and pass here.

Non-blocking, and #918 covers it: the spec now publishes three descriptions of one field with identical validation, and only this copy is accurate about existence not being checked. Worth landing before the siblings' "Must name a real instance" misleads a client.

No dashboard change needed. The only caller of `POST /v1/scoped-budgets` omits the field, and `provider_key_id` stays optional, since null is the aggregate cap rather than an accident.

PR #914, opened 2026-09-03
feat(tenancy): give the invitee a membership inbox to list, accept and decline

khaledosman:
## Description

An organization invitation was reachable only through the emailed link, so losing that email left the invitee with nothing: no list of what was waiting on them, and no way to decline at all.

This adds the invitee's side of the flow, where `/me/member-invitations` is the admin's:

```
GET  /v1/organizations/me/pending-memberships
POST /v1/organizations/me/pending-memberships/{organization_member_id}/accept
POST /v1/organizations/me/pending-memberships/{organization_member_id}/decline
```

Plus an `/invitations` dashboard page, reached from the organization switcher and only while something is waiting.

Three things worth a reviewer's attention:

- **No token, addressed by membership id.** The invite path already lands an `OrganizationMember` against a resolved identity, so a signed-in caller needs no token; the membership (not the invitation) is the stable id, since a revoke and re-invite mints a fresh `Invitation` row each round.
- **Accepting shares one implementation with the token path.** Otherwise an inbox accept would quietly skip the workspace assignments parked at invite time and the attribution row.
- **Declining suspends the membership, matching revoke.** The emailed link is a separate credential, so leaving the membership `invited` would let it flip back to `active` and undo the decline. It is reversible: a later invite revives the suspended membership.

The commit message carries the fuller reasoning, including why the list filters on `expires_at` in the query rather than on status.

## PR Type

- [x] New Feature

## Relevant issues

Fixes mozilla-ai/otari-ai#1999. Refs mozilla-ai/otari-ai#1828.

## Checklist

- [x] I ran the Definition of Done checks locally (`make lint`, `make typecheck`, `make test`) — lint, typecheck, `check-architecture`, the OpenAPI/Postman `--check` gates and the full `web` vitest suite (2706 tests) all pass locally. The new `tests/integration` cases need Postgres, so they run in CI rather than on my machine.
- [x] If the API contract changed, I regenerated the OpenAPI spec (`uv run python scripts/generate_openapi.py`) — plus the Postman collection and `web/src/client/schema.ts`.

## AI Usage

**AI Model/Tool used:**

Claude Opus 5 (1M context) via Claude Code, at @khaledosman's request.

The scope questions the issue left open (where this lives, whether decline suspends, whether accepting needs the token) were settled with @khaledosman before implementation.

## Summary

- Added an invitations inbox for signed-in users.
- Users can view, accept, and decline organization invitations.
- Added an `/invitations` page with organization details, clear actions, and decline confirmation.
- Added an invitation link to the organization switcher when invitations are pending.
- Excluded expired invitations and preserved workspace assignments during acceptance.
- Made acceptance safe to repeat for memberships that are already active.
- Added API, schema, Postman, documentation, client, and integration test coverage.


src/gateway/services/tenancy/organization_service.py
```diff
@@ -1040,6 +1061,185 @@ async def accept_invitation(self, token: str) -> AcceptInvitationResultPublic:
...
+        decline so the two answer identically.
+
+        A membership that is not the caller's collapses into the same
+        ``InvitationNotFoundError`` as one that never existed, which is what
+        that error is for: the ids here are another tenant's roster rows, and a
+        distinguishable 404 would let a caller probe for which of them exist.
+        The caller's own membership in the wrong state collapses into it too,
+        since an ``active`` or ``suspended`` one has no invitation to act on.
+        """
+        membership = await self.members.get(organization_member_id)
```
njbrake: Fetch by primary key plus a Python ownership check, where the house rule is the tenant predicate in the query. #915's `OrganizationDomainRepository.get_by_id_and_organization` does it the other way for this reason. Same response either way, so this is consistency rather than a hole: add `OrganizationMemberRepository.get_by_id_and_user`.

src/gateway/services/tenancy/organization_service.py
```diff
@@ -1040,6 +1061,185 @@ async def accept_invitation(self, token: str) -> AcceptInvitationResultPublic:
...
+                await self.invitations.update_status(lapsed, {"status": "expired"})
+            await self.db.commit()
+            raise InvitationExpiredError
+
+        organization = await self.organizations.get(membership.organization_id)
+        if organization is None:
+            raise InvitationNotFoundError(organization_member_id)
+        return live[0], membership, organization
+
+    async def accept_pending_membership_for_user(
```
njbrake, line 1169: A repeat accept answers 404, so a double click before the list invalidates surfaces "not found" on an action that succeeded. Either return the same success shape when the caller's own membership is already `active`, or swallow the second in-flight mutation's 404 on the page.

src/gateway/repositories/tenancy/invitation_repository.py
```diff
@@ -46,6 +52,67 @@ async def get_pending_by_organization_members(
...
+            col(Invitation.expires_at) >= now,
+        ]
+        joined = (
+            select(Invitation, OrganizationMember, Organization)
+            .join(OrganizationMember, col(Invitation.organization_member_id) == col(OrganizationMember.id))
+            .join(Organization, col(Invitation.organization_id) == col(Organization.id))
+        )
+
+        count_result = await self.db.execute(
+            select(func.count())
```
njbrake: The count query omits the join to `Organization` that the page query makes, so an invitation whose organization row is absent would be counted and not returned, ending `fetchAllPaged`'s walk early on a short page. Unreachable today (no organization deletion, and the FK cascades), but the two should share their join set.

web/src/shared/api/hooks.ts
```diff
@@ -2281,6 +2282,68 @@ export function useRevokeOrganizationMemberInvitation() {
...
+// switcher, so the whole tenancy prefix goes rather than only the inbox key.
+// Not `invalidateQueries()` with no key, unlike `useSwitchOrganization`:
+// accepting does not move `active_organization_id`, so everything cached for
+// the organization the caller is still acting in stays valid.
+export function useAcceptPendingMembership() {
+  const queryClient = useQueryClient()
+  return useMutation({
+    mutationFn: (organizationMemberId: string) =>
+      apiFetch<AcceptInvitationResult>(
+        `/v1/organizations/me/pending-memberships/${organizationMemberId}/accept`,
```
njbrake: Missing `encodeURIComponent`, here and on the decline path below. Server-generated UUIDs, so no exploit, but the adjacent `me/members/${...}` and `me/member-invitations/${...}` hooks encode, as do #915's new ones.

njbrake (approved):
Approving. The no-token design holds: not-mine, nonexistent, and wrong-state all collapse into one 404, the only 400 is reachable after the caller's own membership already resolved, ids are `uuid4`, and a header master key resolves to the bootstrap operator rather than an arbitrary identity. Accept's extraction left the token path unchanged, and decline composing with #915's auto-join is correct, since a suspended membership is returned untouched there.

Nits inline, none blocking.

`docs/access-control.md` still ends its Invitations section with "The browser validates and accepts them through the public invitation endpoints." Add a clause for the authenticated path.

src/gateway/services/tenancy/organization_service.py
khaledosman: Addressed in 2eedd1a7: added `OrganizationMemberRepository.get_by_id_and_user`, alongside `get_by_id_and_organization` and documented as its counterpart for the surfaces reached as the subject of a membership rather than an administrator of it. The Python ownership check is gone. (via Claude Opus 5, at @khaledosman's request)

src/gateway/repositories/tenancy/invitation_repository.py
khaledosman: Addressed in 2eedd1a7: both statements now go through one `joined()` helper, so the count cannot drift to a narrower join set than the page. Kept your reasoning in the comment, including that nothing reaches it today. (via Claude Opus 5, at @khaledosman's request)

web/src/shared/api/hooks.ts
khaledosman: Addressed in 2eedd1a7: both mutation URLs now encode the membership id, matching the adjacent `me/members` and `me/member-invitations` hooks. (via Claude Opus 5, at @khaledosman's request)

src/gateway/services/tenancy/organization_service.py
khaledosman: Addressed in 2eedd1a7, server-side rather than on the page, so any client gets it. Accepting a membership the caller already holds active now answers that membership's organization and role.

Two details worth flagging. It is checked twice, before and under the organization lock: the first check races anything already in flight, so without the second the loser of that race is the exact 404 the check exists to avoid. And the branch is any active membership of theirs, not only one that reached active by accepting; telling those apart costs a lookup for an answer GET /me/memberships already gives, and the check is scoped to the caller's own rows either way. Both the route docstring and a test now say so, since it does mean accept answers for a membership that never had an invitation. Say the word if you would rather pay the lookup and keep that a 404. (via Claude Opus 5, at @khaledosman's request)

PR #915, opened 2026-09-03
feat(tenancy): email-domain auto-join, claimed and proven by DNS

khaledosman:
## Description

Email-domain auto-join: an organization claims a domain, proves control of it with a DNS TXT record, and anyone who signs in afterwards with a verified address at that domain becomes a member at the role the claim names.

The claim and the proof are deliberately separate. Claiming is free and grants nothing, so a claim on a domain someone else controls is inert; only the published record makes it act on anybody. The role it hands out is narrower than a membership's (member and viewer, never owner or admin), because a DNS record proves control of a domain rather than a decision about a person.

Ported from the platform's `OrganizationDomainService` plus the auto-join half of `OrganizationMembershipService`, converted to async. Two things changed on arrival: the resolver is `dns.asyncresolver`, since the blocking one would hold the event loop for the full timeout on a domain whose nameservers are gone, and auto-join now stages into the sign-in's own transaction instead of committing for itself and unwinding by hand. Mixpanel tracking did not travel.

Three commits, one per layer: the claim surface, the sign-in hook, the dashboard page.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Addresses mozilla-ai/otari-ai#1984, which tracks this capability having no control-plane home before the cutover.

## Checklist

`make lint` and `make typecheck` are green, as are `uv run pytest tests/unit` (2812), the whole web suite (2703) and `tsc -b`. `make test` was not run: the integration suite needs Docker, which is not available on this machine, so `tests/integration/test_organization_domains.py` has only ever run in CI. Worth a reviewer's attention for that reason.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

Written from the platform's implementation as the reference, against the attachment point `organization_service.py`'s own docstring names for this slice. The scope decision behind it (port rather than retire) is @khaledosman's, recorded on mozilla-ai/otari-ai#1984.

## Summary

Organizations can now claim email domains and verify ownership through DNS TXT records. Verified domains can automatically add users at the matching email domain as members or viewers.

The change adds:

- Domain claiming, verification, expiry, pausing, and removal.
- Punycode normalization and public email-provider protection.
- Automatic membership during password, OAuth, and passkey sign-in.
- Safe handling for existing, suspended, and active organization memberships.
- Organization dashboard controls and API support.
- Database migration, documentation, OpenAPI updates, and comprehensive unit, integration, web, and screenshot tests.

## Technical notes

- DNS checks use asynchronous resolution with bounded timeouts.
- Membership changes remain within the sign-in transaction.
- Verified domain claims are unique across organizations.
- Domain proofs expire after 90 days.
- Integration tests were not run locally because Docker was unavailable.


src/gateway/services/tenancy/organization_domain_service.py
```diff
@@ -0,0 +1,328 @@
...
+        await self.organizations.require_active_organization_management_access(user=user, organization=organization)
+
+        domain = self._validated_claim(request.domain)
+
+        # The pre-check answers the ordinary case (a domain another organization
+        # already holds) with a clean conflict; the caught IntegrityError closes
+        # the race where two admins claim the same domain at once. Both are
+        # needed: the first alone loses the race, the second alone would make a
+        # routine conflict indistinguishable from a bug.
+        if await self.domains.get_by_domain(domain) is not None:
```
njbrake: P1. An unproven claim permanently locks out the party that can actually prove control. This refuses on any existing row, verified or not, only the holder can delete it, and the 409 hides who that is. `create_organization_for_user` runs no role check, so any signed-in identity can mint an organization, become its owner, and claim any domain. On hosted that is a cross-tenant denial with no operator remedy short of SQL, and it contradicts the PR's "a claim on a domain someone else controls is inert": inert for membership, not for the owner's ability to claim.

Let a verified claim displace an unproven one, or expire unproven claims, or both.

src/gateway/services/tenancy/organization_domain_service.py
```diff
@@ -0,0 +1,328 @@
...
+
+        row = await self._require_domain(organization_domain_id, organization.id)
+        if row.verified_at is not None:
+            return self._to_public(row)
+
+        expected = f"{DOMAIN_VERIFICATION_TXT_PREFIX}{row.verification_token}"
+        if expected not in await resolve_txt_records(row.domain):
+            raise OrganizationDomainNotVerifiedError(row.domain)
+
+        verified = await self.domains.mark_verified(row, verified_at=datetime.now(UTC))
```
njbrake: P2. `verified_at` is stamped once and never re-read. Pull the TXT record, transfer the domain, or let it lapse, and the claim keeps admitting people at the role it names, including strangers who now own the domain.

Re-verify lazily on the auto-join read with a TTL, or add a staleness column and a sweeper. If proof is deliberately once-only, say so and drop the claim of re-checkability in the model docstring.

src/gateway/models/tenancy.py
```diff
@@ -764,6 +764,120 @@ class OrganizationMember(OrganizationMemberBase, PrimaryKeyMixin, CreatedAtMixin
...
+
+class OrganizationDomain(OrganizationDomainBase, PrimaryKeyMixin, CreatedAtMixin, UpdatedAtMixin, table=True):
+    """One organization's claim on an email domain.
+
+    A claim on its own grants nothing. Auto-join reads only rows that are both
+    ``enabled`` and verified, so an organization may name any domain it likes
+    and nothing happens until the DNS proof lands; that is what stops a claim on
+    a domain someone else controls from sweeping up that domain's people.
+
+    Kept after verification rather than collapsed into a boolean on the
```
njbrake: "revocable and re-checkable: the token is the same one the published record has to keep matching" describes behavior that does not exist. Nothing re-reads DNS after the first verify.

src/gateway/core/email_domains.py
```diff
@@ -0,0 +1,106 @@
...
+    }
+)
+
+# A registrable domain: dot-separated labels with an alphabetic TLD. Stricter
+# than `core.addresses`' address pattern on purpose. That one bounds what a
+# person may sign in as and leans permissive; this one bounds what an admin
+# stores as a claim, and the stored value has to match the normalizer's output
+# exactly on every later sign-in, so a scheme, a path, or a stray label would
+# store a claim that can never match.
+_DOMAIN_PATTERN = re.compile(r"^(?=.{1,253}$)([a-z0-9](?:[a-z0-9-]{0,61}[a-z0-9])?\.)+[a-z]{2,}$")
```
njbrake, line 90: P2. The two normalizers disagree for internationalized domains, which is the failure this module's docstring exists to prevent. ASCII-only here means `münchen.de` is refused as invalid, pushing an admin to `xn--mnchen-3ya.de`, which is claimable and DNS-provable; `email_domain("a@münchen.de")` then returns `münchen.de` and never matches. The result is a claim shown as verified and Active that admits nobody, with no error anywhere.

IDNA-encode in both `email_domain` and `normalized_domain`, or refuse an `xn--` claim with a message saying IDN is unsupported.

tests/integration/test_organization_domains.py
```diff
@@ -0,0 +1,570 @@
...
+    assert updated.verified_at is None
+    assert updated.domain == "acme.example"
+
+
+# =============================================================================
+# Auto-join
+# =============================================================================
+
+
+async def test_a_verified_address_joins_the_organization_that_proved_its_domain(
```
njbrake, line 328: P2. Every auto-join test calls `auto_join_for_user` directly; nothing in this file touches `/v1/auth/session`, `/v1/auth/oauth`, or `/v1/auth/webauthn`. Delete any one of those three call sites and the suite stays green while the feature dies on that route. The transaction claim (membership and session land together or not at all) also has nothing exercising it.

Add one route-level test per sign-in path, plus one asserting no membership survives a sign-in that fails after the stage.

src/gateway/api/routes/organizations.py
```diff
@@ -260,3 +272,94 @@ async def revoke_active_organization_member_invitation(
...
+    domain is a different claim and needs its own proof.
+    """
+    return await service.update_domain_for_user(
+        user=current_identity,
+        organization_domain_id=organization_domain_id,
+        request=body,
+    )
+
+
+@router.post("/me/domains/{organization_domain_id}/verify")
```
njbrake, line 412: P3. No cap on claims per organization, and this route performs a 5s outbound DNS lookup on a caller-chosen name with no throttle. An admin can create claims on domains they do not own and hammer verify, making the gateway a query relay against arbitrary nameservers. A `MAX_ORGANIZATION_DOMAINS` cap in the shape of `MAX_WORKSPACE_ASSIGNMENTS` is the cheap fix; there is no throttle precedent for authenticated management routes to follow.

src/gateway/models/tenancy.py
```diff
@@ -764,6 +764,120 @@ class OrganizationMember(OrganizationMemberBase, PrimaryKeyMixin, CreatedAtMixin
...
+    # nothing. Unique across the deployment, not per organization: the whole
+    # point is that one domain resolves to one organization at sign-in, and two
+    # organizations holding the same claim has no defensible answer.
+    domain: str = Field(max_length=255)
+    default_role: str = Field(default="member", max_length=32)
+    enabled: bool = Field(default=True)
+
+    @field_validator("default_role")
+    @classmethod
+    def validate_default_role(cls, value: str) -> str:
```
njbrake, line 831: P3. This validator never runs. SQLModel skips Pydantic validation on `table=True` models: `OrganizationMember(role="bogus-role")` constructs fine today, and the repository builds `OrganizationDomain(...)` directly while `update_domain` goes through `sqlmodel_update`, so `OrganizationDomainCreate` and `OrganizationDomainUpdate` are never routed through either. The request `Literal` is the only enforcement, which is adequate, but the commit message claims one layer more than exists.

njbrake (changes requested):
Changes needed. Nothing here grants membership to anyone unentitled: `auto_join_for_user` is the only reader of the claim table, and it needs a verified address plus an enabled, verified claim; the role narrowing holds at the request `Literal`; matching is equality on the registrable domain, so `evil-example.com` does not match `example.com`.

Blocking items are inline. Two have no line to sit on:

**Screenshot entry.** `/organization/domains` is a new page and `web/e2e/screenshots/authenticated.spec.ts` is untouched. Add an entry beside `organization-members`.

**Docs.** Nothing under `docs/` changed, and three places owe an update: a section in `access-control.md` beside Invitations, since this is a second way to become a member; the organization page list in `dashboard.md`; and `dashboard.md`'s "Authentication options", which still states that sign-in does not make an account a member.

src/gateway/services/tenancy/organization_domain_service.py
khaledosman: Addressed in 281ad7ec: uniqueness is now partial, over verified rows only, so any number of organizations may hold an unproven claim and the first to publish the record takes it. Verifying deletes the rival unverified rows, which the partial index had already made un-verifiable. Creating is refused only against a proven claim; the caller's own duplicate gets a named refusal instead, since that row is theirs to see. Covered by test_an_unproven_claim_does_not_lock_out_the_domains_real_owner and test_proving_a_domain_displaces_the_unproven_claims_on_it. (via Claude Opus 5)

src/gateway/services/tenancy/organization_domain_service.py
khaledosman: Addressed in 281ad7ec: DOMAIN_PROOF_TTL (90 days) bounds a proof. auto-join skips a claim whose stamp is older, so it fails closed with no sweeper, and verify no longer short-circuits once the proof has aged out, which is how a claim is renewed. Re-checking DNS on the auto-join read was the other option and puts a 5s outbound lookup in front of somebody signing in, so the read stays cheap and the staleness is what moves. proof_expires_at is computed server-side and published, so the dashboard shows Proof expired and offers the re-check without holding a second copy of the TTL. (via Claude Opus 5)

src/gateway/models/tenancy.py
khaledosman: Addressed in 281ad7ec: the claim is true now that proofs age out and are re-checked, and the docstring was rewritten anyway to say what verified_at actually means (the age of the evidence, not a flag) and why the row outlives verification. (via Claude Opus 5)

src/gateway/core/email_domains.py
khaledosman: Addressed in 281ad7ec: both normalizers punycode through a shared _ascii helper, so muenchen with an umlaut and its xn-- spelling are one claim and a sign-in at either matches. A value the codec refuses is passed through unchanged rather than raising, so a claim is rejected by the shape check a moment later and a sign-in simply matches nothing. Unit tests cover the round trip and assert the two normalizers agree. (via Claude Opus 5)

tests/integration/test_organization_domains.py
khaledosman: Addressed in 7de73fea: tests/integration/test_email_domain_auto_join_routes.py signs in over HTTP once per credential (password, OAuth, passkey) against a claim held by a second organization, so deleting a call site fails rather than going quiet. A fourth test makes the session write raise after auto-join has staged and reads the membership table on its own connection to show nothing committed, which is the transaction claim. (via Claude Opus 5)

src/gateway/api/routes/organizations.py
khaledosman: Addressed in 281ad7ec: MAX_ORGANIZATION_DOMAINS is 50, checked in the service, in the shape of MAX_WORKSPACE_ASSIGNMENTS. No throttle added, for the reason you gave: there is no precedent for one on an authenticated management route here, and inventing one in this PR would be the wrong place to set that pattern. (via Claude Opus 5)

src/gateway/models/tenancy.py
khaledosman: You are right on both counts. The comment now states it: the validator runs on the request and response schemas and not on the table class, the request Literal is the enforcement, and it is kept only as a last guard so a row that somehow held a management role could not be serialized as if fine. The overclaim is corrected in 281ad7ec's message rather than by rewriting the pushed history. (via Claude Opus 5)

khaledosman:
All seven inline threads are addressed and resolved, and the two without a line to sit on are done:

**Screenshot entry.** `/organization/domains` is in `web/e2e/screenshots/authenticated.spec.ts` now, beside `organization-members`, with its own capture.

**Docs.** `access-control.md` gains an Email-domain auto-join section beside Invitations: the TXT record, that claiming is not exclusive but proving is, that a proof expires, and every narrowing that keeps it safe. `dashboard.md` lists the page with the other organization pages, and its Authentication options section no longer says sign-in cannot make an account a member, which this feature made false. That sentence was the one most worth correcting, so thank you for naming it.

Two of the fixes changed the design rather than patching it, so they are worth a second look rather than a re-read of the diff:

- **Uniqueness is now partial**, over verified rows only, and verifying deletes the rival unverified claims. That is what stops an unproven claim from locking out the party who can actually prove control. It does mean one organization's row can disappear because another proved the domain, which is deliberate and documented, but it is a behaviour a reviewer should agree with rather than discover.
- **`DOMAIN_PROOF_TTL` is 90 days**, chosen rather than derived. Past it a claim admits nobody until an admin re-verifies. That number is the exposure window for a domain that changes hands, since nothing re-reads DNS in between, and it is a one-line change if you want it shorter.

Local: `make lint`, `make typecheck`, `pytest tests/unit` (2822), the web suite (2705) and `tsc -b` all green. The integration suite still needs Docker, which this machine does not have, so `test_organization_domains.py` and the new `test_email_domain_auto_join_routes.py` are CI-verified only.

_Replies and fixes drafted by Claude Opus 5 (1M context) via Claude Code, at @khaledosman's direction._

tests/integration/test_email_domain_auto_join_routes.py
```diff
@@ -0,0 +1,251 @@
...
+    created = client.post("/v1/organizations", json={"name": "Beta"}, headers=master_key_header)
+    assert created.status_code == 201, created.text
+    beta = str(created.json()["id"])
+
+    switched = client.post(
+        "/v1/organizations/me/switch",
+        json={"organization_id": beta},
+        headers=master_key_header,
+    )
+    assert switched.status_code == 200, switched.text
```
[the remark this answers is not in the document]
khaledosman: Good catch, and the most important one here: those tests were passing for the wrong reason. The fixture left the operator switched into the claiming organization, so POST /me/members seeded Ada straight into it and all three assertions held with auto-join deleted. Fixed in a19960be: the operator is switched back to the booted organization, _rostered_identity asserts the active organization is not the claiming one, and the shared assertion refuses to run when the two ids are equal. Two of them were also failing in CI for a plainer reason, that signup needs public_base_url as well as a transport; both now come from one mail_configured fixture. (via Claude Opus 5)

src/gateway/core/email_domains.py
```diff
@@ -0,0 +1,131 @@
...
+    is a claim shown as verified and active that admits nobody, with no error
+    raised anywhere to explain it.
+
+    A value the codec refuses (an empty label, an over-long one) is returned
+    unchanged rather than raised on. For a claim, ``is_registrable_domain``
+    rejects it a moment later with a message about the domain; for a sign-in it
+    simply matches nothing, which is what an unclaimable domain should do.
+    """
+    try:
+        return domain.encode("idna").decode("ascii")
```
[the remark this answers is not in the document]
khaledosman: Addressed in a19960be: switched to the idna library with uts46=True, transitional=False. Confirmed the collapse you describe: the stdlib codec maps the sharp-s spelling and fass.de onto the same fass.de, so a proof on one would admit addresses at the other. The two now normalize to xn--fa-hia.de and fass.de and stay apart, with a unit test naming the case. The dependency is declared explicitly rather than leaned on transitively, since this is now a correctness-critical import. (via Claude Opus 5)

src/gateway/services/tenancy/organization_domain_service.py
```diff
@@ -0,0 +1,374 @@
...
+        organization = await self.organizations.get_active_organization_for_user(user)
+        await self.organizations.require_active_organization_management_access(user=user, organization=organization)
+
+        domain = self._validated_claim(request.domain)
+
+        # Only a *proven* claim blocks a new one. An unproven claim by anyone
+        # else is no obstacle: it grants nothing, and refusing on it would let
+        # whoever claimed a domain first lock its real owner out for good.
+        if await self.domains.get_verified_by_domain(domain) is not None:
+            raise OrganizationDomainAlreadyClaimedError(domain)
```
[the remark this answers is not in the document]
khaledosman: Addressed in a19960be: get_verified_by_domain is now filtered through the model's proof_expired at both call sites. A live proof still blocks a new claim; an aged-out one does not, and when the new claimant proves the domain the stale holder is demoted to unproven so the partial index admits the new proof. Demoted rather than deleted, since the row is that organization's own. This was the same permanent lockout the partial index was introduced to fix, so thank you for spotting it wearing a different hat. (via Claude Opus 5)

src/gateway/services/tenancy/organization_domain_service.py
```diff
@@ -0,0 +1,374 @@
...
+
+        # Re-checked here rather than only at claim time: another organization
+        # may have proven this domain in the meantime, and the index would
+        # otherwise refuse the write with an error the caller cannot read.
+        holder = await self.domains.get_verified_by_domain(row.domain)
+        if holder is not None and holder.id != row.id:
+            raise OrganizationDomainAlreadyClaimedError(row.domain)
+
+        expected = f"{DOMAIN_VERIFICATION_TXT_PREFIX}{row.verification_token}"
+        if expected not in await resolve_txt_records(row.domain):
```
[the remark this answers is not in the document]
khaledosman: Addressed in a19960be: verify now calls release_session(self.db) before resolve_txt_records and re-reads the claim, the holder and the expiry afterwards, which is the pattern mozilla-ai/otari#911 established for the provider call on the inference path. The re-read is what makes it safe: another organization may prove the domain while the lookup is out, and that now surfaces as the conflict or a 404 rather than a constraint error. (via Claude Opus 5)

src/gateway/services/tenancy/organization_domain_service.py
```diff
@@ -0,0 +1,374 @@
...
+
+        # Only a *proven* claim blocks a new one. An unproven claim by anyone
+        # else is no obstacle: it grants nothing, and refusing on it would let
+        # whoever claimed a domain first lock its real owner out for good.
+        if await self.domains.get_verified_by_domain(domain) is not None:
+            raise OrganizationDomainAlreadyClaimedError(domain)
+        if await self.domains.get_by_domain_and_organization(domain, organization.id) is not None:
+            raise OrganizationDomainClaimedHereError(domain)
+        if await self.domains.count_for_organization(organization.id) >= MAX_ORGANIZATION_DOMAINS:
+            raise TooManyOrganizationDomainsError(MAX_ORGANIZATION_DOMAINS)
```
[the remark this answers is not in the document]
khaledosman: Not fixed, deliberately, and the code now says so rather than implying otherwise. The count and the insert are not serialized, so concurrent creates can both pass the check; the overshoot is bounded by how many requests one organization's admins have in flight, every later create still refuses, and the ceiling exists to bound outbound DNS names rather than to be exact. Locking the organization row on a management write is a heavier invariant than the value being protected. Happy to add it if you disagree. (via Claude Opus 5)

khaledosman:
Four of CodeRabbit's five are fixed in `a19960be` and resolved; the quota race is answered on its thread and left open, since it is a decision rather than an oversight.

The **test finding was the one that mattered**, and it was mine to have caught. The route tests I added last round were passing for the wrong reason: the fixture switched the operator into the claiming organization and never switched back, so `POST /me/members` seeded Ada straight into it and all three assertions held with auto-join deleted. Exactly the failure those tests existed to prevent. They now switch back, assert the two organizations differ, and refuse to run if they do not.

Two of them were also **failing in CI** on the previous head, for the plainer reason that `POST /v1/auth/signup` needs `public_base_url` as well as a transport. Both settings now come from one fixture instead of each test setting half.

The other three changed behaviour and are worth reading rather than skimming:

- **IDNA2003 folded `ß` to `ss`**, so a proof taken on `fass.de` admitted addresses at the sharp-s spelling, which is a different registrable domain with a different owner. That defeats the one guarantee this feature has. Now on the `idna` library with `transitional=False`, declared as a real dependency.
- **An expired proof held a domain hostage.** It blocked every new claim while admitting nobody, which is the same permanent lockout the partial index was added to prevent. A live proof still blocks; an aged-out one is demoted when someone else proves the domain.
- **`verify` held a pooled connection across the DNS lookup**, up to the resolver timeout. It releases first and re-reads after, the way #911 does before a provider call.

One thing I changed that nobody asked for: **proving a domain no longer deletes rival claims.** The partial index already makes a proof exclusive, so the sweep bought nothing, it deleted another tenant's row as a side effect, and it silently undid the demotion above one line later. A rival now survives and gets a readable conflict if it tries to verify.

On the screenshot nitpick: removed, though its stated reason was wrong. `/organization/members` does have a dedicated test as well as a matrix entry, as do general, workspaces, pricing and usage. But the substance holds, because `open()` is exactly `gotoRoute` plus a heading wait, so mine was pure duplication. I did not touch the pre-existing ones.

Local: lint, typecheck, `pytest tests/unit` (2853) and the web suite (2732) green. Integration is CI-only here.

_Fixes drafted by Claude Opus 5 (1M context) via Claude Code, at @khaledosman's direction._

Issue #565, opened 2026-08-12
Drop the hand-declared service_tier once any-llm models it on CompletionParams


njbrake:
_Note: this issue was drafted by Claude via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

Blocked on mozilla-ai/any-llm#1269.

#561 hand-declares `service_tier` on `ChatCompletionRequest` (`src/gateway/api/routes/chat.py`) because any-llm's `CompletionParams` does not model it, so the derived base dropped a caller's value before the provider call. That declaration is a stopgap. Once any-llm ships the param and `any-llm-sdk` is bumped here, remove it so the field comes from derivation like every other completion param.

The trigger is the version bump, which Dependabot lands on its own, so this issue is the thing that surfaces the cleanup at that moment.

### Not a plain revert

A subclass field fully overrides the derived one (see `_schema_derive`), so as long as the declaration stays, otari's `str | None` wins over whatever annotation any-llm chooses. If upstream types it as a literal, our looser type silently persists in `docs/public/openapi.json` and in every generated SDK, and no test flags it.

So the steps are:

1. Delete the `service_tier` field from `ChatCompletionRequest`, keeping the derived one.
2. Regenerate both artifacts (`uv run python scripts/generate_openapi.py`, then `make postman`) and confirm the annotation in `openapi.json` matches what any-llm declares, rather than staying `string | null` by accident.
3. Keep the tests in `tests/integration/test_provider_kwargs_override.py`. They assert the forwarding contract end to end and hold whether the field is derived or hand-declared.

HareeshBahuleyan:
Blocker cleared. any-llm-sdk 1.27.0 models `service_tier` on `CompletionParams` (`any_llm/types/completion.py:244`); 1.25.0 did not. mozilla-ai/any-llm#1269 was implemented by mozilla-ai/any-llm#1274, which added the param to both `CompletionParams` and `MessagesParams`.

#938 bumps the pin to `any-llm-sdk[all]>=1.27.0`, so this is actionable once that merges.

One step to add to the list above: drop the `| {"service_tier"}` union from `_pipeline._FORWARDED_PARAMS` as well. Derivation puts the name there via `CompletionParams.model_fields`, so the explicit entry is now redundant. #924 carries the equivalent step for `container`.

Nothing to do on the Messages side. `MessagesRequest` never hand-declared the param, so it just appears in the derived schema; #938 already carries that spec change.

[...]

PR #916, opened 2026-09-03
fix(messages): preserve Anthropic container field

HareeshBahuleyan:
## Description

Accept Anthropic's top-level `container` value on `/v1/messages` and forward it unchanged to the selected provider. The request field is declared locally as a compatibility measure until the corresponding any-llm release is available; a follow-up will derive it from `MessagesParams` after that dependency is updated.

The OpenAPI schema and route-level tests cover request validation, serialization, and provider forwarding.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Related to mozilla-ai/octonous#4913

## Checklist

`make lint`, `make typecheck`, the focused tests, generated-artifact checks, and the OSS smoke test pass. The full suite has one unrelated existing failure in `tests/unit/test_docs_style.py` because `docs/superpowers/plans/2026-08-20-inline-platform-response-cost.md` contains em dashes.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:** OpenAI GPT-5.6 SOL with Pi

**Any additional AI details you'd like to share:** The implementation and PR preparation were completed with an AI coding agent and reviewed against the repository guidance.

**NOTE:**
When responding to reviewer questions, please respond yourself rather than copy/pasting reviewer comments into an AI and pasting back its answer. We want to discuss with you, not our AI :)

## Summary

- Added support for the top-level `container` field in `/v1/messages`.
- Forwarded `container` for compatible provider requests and removed it when Otari manages code execution.
- Added validation for managed hybrid routes and conflicting code-execution tools.
- Updated the OpenAPI schema and generated client.
- Added route, hybrid-mode, schema, and error-classification tests.

### DIFF
--- a/src/gateway/api/routes/messages.py
+++ b/src/gateway/api/routes/messages.py
@@ -74,7 +74,9 @@ class MessagesRequest(derive_request_base(MessagesParams)):  # type: ignore[misc
 
     The wire fields are derived from any-llm's ``MessagesParams`` (see
     ``_schema_derive``) so the schema cannot silently drop a param any-llm
-    forwards. Gateway-internal fields (``mcp_servers``, ``mcp_server_ids``,
+    forwards. ``container`` is an Anthropic wire param ``MessagesParams`` does
+    not model, declared here and forwarded as an any-llm ``**kwargs`` param.
+    Gateway-internal fields (``mcp_servers``, ``mcp_server_ids``,
     ``guardrails``, ``tools_header``, ``max_tool_iterations``) opt the request
     into gateway-managed MCP / sandbox / web_search / guardrails without
     changing the upstream wire shape. They're stripped before the request is
@@ -82,7 +84,16 @@ class MessagesRequest(derive_request_base(MessagesParams)):  # type: ignore[misc
     """
 
     messages: list[dict[str, Any]] = Field(min_length=1)
-    # Compatibility field until any-llm's MessagesParams includes it.
+    # Anthropic's top-level container id, for continuing a code-execution
+    # container across turns. ``MessagesParams`` does not model it, so the
+    # derived base would drop a caller's value before the provider call. It
+    # rides any-llm's ``**kwargs``, which is why it is also registered in
+    # ``_pipeline._FORWARDED_PARAMS``: without that, a bridged (non-Anthropic)
+    # provider's rejection reads as an upstream outage instead of a 400.
+    #
+    # Stopgap: remove this declaration once the SDK pin carries the param
+    # (mozilla-ai/any-llm#1329, merged after 1.26.0). Until then it also shadows
+    # whatever annotation any-llm picks for it.
     container: str | None = None
     # any-llm types ``stream`` as ``bool | None``; keep the Anthropic wire
     # contract (a non-nullable boolean defaulting to false) for stable SDK
@@ -587,6 +598,12 @@ async def create_message(
         request_fields["tools"] = openai_to_anthropic_tools(request_fields["tools"])
     if tool_ctx.intercepts_web_search and request_fields.get("messages"):
         request_fields["messages"] = _strip_gateway_minted_blocks(request_fields["messages"])
+    if tool_ctx.use_sandbox:
+        # ``container`` addresses Anthropic's own code-execution container. The
+        # gateway is running the code in its sandbox instead, and the tool is
+        # gone from the forwarded ``tools``, so the provider would be asked to
+        # attach a container nothing in this request can reach.
+        request_fields.pop("container", None)
 
     # ------------------------------------------------------------------
     # Streaming path
--- a/src/gateway/api/routes/messages.py
+++ b/src/gateway/api/routes/messages.py
@@ -92,8 +92,8 @@ class MessagesRequest(derive_request_base(MessagesParams)):  # type: ignore[misc
     # provider's rejection reads as an upstream outage instead of a 400.
     #
     # Stopgap: remove this declaration once the SDK pin carries the param
-    # (mozilla-ai/any-llm#1329, merged after 1.26.0). Until then it also shadows
-    # whatever annotation any-llm picks for it.
+    # (mozilla-ai/any-llm#1329, merged after 1.26.0; tracked in #924). Until
+    # then it also shadows whatever annotation any-llm picks for it.
     container: str | None = None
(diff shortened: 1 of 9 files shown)


khaledosman (commented):
Generated artifacts check out at this head: `make openapi-check` and `make postman-check` both pass (the collection is unaffected because `/v1/messages` uses a curated body override), and `web/src/client/schema.ts` is regenerated.

One should-fix inline: a hand-declared param forwarded through any-llm's `**kwargs` also has to be registered in `_FORWARDED_PARAMS`, or its rejection by a bridged provider is misclassified as a 502.

🤖 Review written by Claude Code (Opus 5).

src/gateway/api/routes/messages.py
```diff
@@ -82,6 +82,8 @@ class MessagesRequest(derive_request_base(MessagesParams)):  # type: ignore[misc
     """
 
     messages: list[dict[str, Any]] = Field(min_length=1)
+    # Compatibility field until any-llm's MessagesParams includes it.
+    container: str | None = None
```
khaledosman, line 99: `container` rides any-llm's `**kwargs`, so it needs the registration `service_tier` has: it is missing from `_FORWARDED_PARAMS` (`_pipeline.py:376-386`).

Against any non-Anthropic model the Messages -> Completions bridge passes it straight into the provider SDK (`any_llm/providers/openai/base.py:97` does `converted_params.update(kwargs)`, and `AsyncCompletions.create` takes no `**kwargs`), so the call raises `TypeError: ...create() got an unexpected keyword argument 'container'`. `_rejected_param` gates on that set, so the name fails the gate and a caller-fixable request error surfaces as a generic 502 instead of the 400 that names the param. Verified at this head: `'container' in _FORWARDED_PARAMS` is `False`, and `_rejected_param(TypeError("AsyncCompletions.create() got an unexpected keyword argument 'container'"))` returns `None`.

Add `"container"` beside `service_tier` in that set, plus a case in `tests/unit/test_provider_error_classification.py`.

khaledosman, line 99: The type is narrower than the wire contract. Anthropic's container param is `BetaContainerParams | str` on the beta Messages endpoint (`anthropic/types/beta/message_create_params.py:331`), so `{"id": ..., "skills": [...]}` is valid. `str | None` rejects that with a 422 even for a caller who sets the matching `betas`, and because a subclass field redefinition fully overrides the derived one, this annotation keeps shadowing any-llm's after `MessagesParams` grows the field. `str | dict[str, Any] | None` matches the wire and mirrors how the other Anthropic passthrough objects (`cache_control`, `context_management`) are typed.

khaledosman, line 99: Non-blocking: a container ID is provider-side state (uploaded files, an execution environment) scoped to the upstream credential, so forwarding a caller-supplied one unmodified means whoever holds the ID resumes that container, across tenants wherever the upstream account is shared. `scope_prompt_cache_key`, called on `request_fields` a few lines down in the dispatch, exists for exactly that shared-account problem, and a container ID cannot be namespaced the same way. Worth a note in the comment or a follow-up issue rather than a change here.

src/gateway/api/routes/messages.py
```diff
@@ -82,6 +82,8 @@ class MessagesRequest(derive_request_base(MessagesParams)):  # type: ignore[misc
     """
 
     messages: list[dict[str, Any]] = Field(min_length=1)
+    # Compatibility field until any-llm's MessagesParams includes it.
```
khaledosman: `service_tier`, the other hand-declared `**kwargs` passthrough field, carries a comment (`chat.py:95-104`) that names the tracking issue for its removal and states that it shadows whatever annotation any-llm picks. This one-liner does neither, so the "follow-up will derive it" promise lives only in the PR body. Name the issue here, and add `container` to the `MessagesRequest` docstring's enumeration of hand-declared fields, the way `chat.py:75` lists `service_tier`.

src/gateway/api/routes/messages.py
njbrake: Done in 23357bd4. `"container"` is registered beside `service_tier`, with a case in `tests/unit/test_provider_error_classification.py`.

One correction to the original claim: the resulting 400 covers standalone and single-attempt hybrid. A multi-attempt hybrid route keeps the generic 502 by design, since `run_platform_attempts` does not attribute one provider's status to a heterogeneous set of failures.

njbrake: Not changing this. Correct that Anthropic's beta endpoint types it `BetaContainerParams | str`, so the skills shape now 422s where it was previously accepted and dropped.

Upstream any-llm#1329 also declares `container: str | None`, so widening here would diverge from the pin and silently re-narrow when the SDK bump lands. Tracked in #924, which lists checking the width as a step at that point.

njbrake: Gated in 533e2ea6. A caller-supplied `container` is refused with a 400 when any attempt on a hybrid route is `managed`, before dispatch. Any attempt rather than the first, because fallback picks the attempt past that point.

BYO routes still forward it, since there the account is the caller's own. Standalone is not gated: those credentials belong to the deployment operator, and the managed rung answers `None` in every build that mounts this route.

src/gateway/api/routes/messages.py
njbrake: Done in 23357bd4 and b6973909. The comment names the upstream PR, states that the declaration shadows any-llm's annotation, and points at #924, the container equivalent of #565. `container` is also named in the `MessagesRequest` docstring, matching `chat.py`.

njbrake (approved):
Approving. `container` now carries the four things `service_tier` carries: the `_FORWARDED_PARAMS` registration, a tracking issue (#924), the shadowing note, and a line in the docstring.

Two behavior changes landed on top of the original fix, worth naming for anyone reading the squash later.

`otari_code_execution` alongside a provider-native code-execution tool is now a 400. It previously resolved silently: the gateway ran the code, forwarded the native tool, and dropped the container the caller named. The check sits in `prepare_gateway_tools`, so it applies to chat and responses as well.

A caller-supplied `container` is refused on a hybrid route with any managed attempt. That account is shared across workspaces, and a provider-minted id cannot be namespaced the way `scope_prompt_cache_key` namespaces a cache key.

Full suites and all 16 checks green.

PR #919, opened 2026-09-03
fix(migrations): run otari's chain only against otari's own database

khaledosman:
## Description

Otari's Alembic chain stamps the default `alembic_version` table, the same table every other Alembic application
stamps, so a URL naming a different application's database is accepted as readily as otari's own. All Alembic says
then is `Can't locate revision identified by ...`, which reads as a corrupt database rather than as a URL naming the
wrong one.

Three changes:

- Before migrating, refuse a database stamped with a revision this chain does not own, naming the target (password
  masked) and the revision. That covers both readings: another application's database, or one written by a newer
  otari than this process, which is the shape a rolled-back deploy leaves behind.
- Let `alembic stamp` through the check. Rewriting the version table is how an operator repairs the state the
  check refuses, and `alembic stamp --purge` works on such a database today, so refusing to run it would take the
  fix away along with the failure.
- Stop falling back to a bare `DATABASE_URL`, and stop defaulting to `sqlite:///./otari.db` behind it. `otari
  serve` and `otari migrate` both take that name as `--database-url` and pass the resolved URL on explicitly, so
  the only invocation the fallback ever served was a bare `alembic upgrade head` in a process holding another
  application's `DATABASE_URL`. Keeping the SQLite default under it would have turned that run into a silent
  success: a throwaway file migrated in the working directory, exit 0, the operator's database untouched. A run
  that resolves no URL now raises before it connects.

The refusal names host and database rather than the URL: `render_as_string(hide_password=True)` masks the password
field but not credentials passed as query parameters, and `otari migrate` echoes that string (#922).

Found while diagnosing a deploy where this chain met another application's `alembic_version` table. The general
case, a neighbor whose history lives anywhere but this `alembic_version` table, is out of reach of any
revision-set check and needs a version table of otari's own: #921.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

None. Filed from an incident in `mozilla-ai/otari-ai`.

## Checklist

`make lint`, `make typecheck`, `tests/unit` (2834 passed) and the OSS-edition smoke ran locally. The integration
suite needs PostgreSQL and runs in CI. No API contract or documentation change: the refusal message is what a
misconfigured deployment reads.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

Claude Opus 5 (1M context) via Claude Code.

Written at @khaledosman's request while diagnosing the deploy failure this fixes. Both tests were confirmed to fail
without the change.

**NOTE:**

### DIFF
--- a/alembic/env.py
+++ b/alembic/env.py
@@ -3,7 +3,11 @@ import os
 from logging.config import fileConfig
 
 from alembic import context
+from alembic.runtime.migration import MigrationContext
+from alembic.script import ScriptDirectory
+from alembic.util import CommandError
 from sqlalchemy import engine_from_config, pool
+from sqlalchemy.engine import Engine, make_url
 
 from gateway.core.database import to_sync_url
 
@@ -28,11 +32,15 @@ if config.config_file_name is not None:
 # for 'autogenerate' support
 target_metadata = Base.metadata
 
-# Get database URL from config (if already set programmatically) or environment variables
-# Priority: Programmatically set URL -> OTARI_DATABASE_URL -> DATABASE_URL -> default
-database_url = config.get_main_option("sqlalchemy.url") or (
-    os.getenv("OTARI_DATABASE_URL") or os.getenv("DATABASE_URL") or "sqlite:///./otari.db"
-)
+# Get database URL from config (if already set programmatically) or the environment.
+# Priority: Programmatically set URL -> OTARI_DATABASE_URL -> default
+#
+# A bare `DATABASE_URL` is deliberately not read. `otari serve` and `otari
+# migrate` both take that name as `--database-url` and pass the resolved URL on
+# explicitly, so the only invocation this fallback ever served was a bare
+# `alembic upgrade head` in a process holding another application's
+# `DATABASE_URL`, which aims this chain at that application's database.
+database_url = config.get_main_option("sqlalchemy.url") or os.getenv("OTARI_DATABASE_URL") or "sqlite:///./otari.db"
 # Normalized here rather than at each call site so every entry point is covered:
 # `otari migrate`, auto_migrate on startup, and a bare `alembic upgrade head`
 # reading OTARI_DATABASE_URL. Migrations run on a sync engine, so an async URL
@@ -43,6 +51,41 @@ config.set_main_option("sqlalchemy.url", to_sync_url(database_url))
 # can be acquired:
 # my_important_option = config.get_main_option("my_important_option")
 # ... etc.
+
+
+def _reject_foreign_history(engine: Engine) -> None:
+    """Refuse a database whose stamped history is not this chain's.
+
+    Otari stamps the default `alembic_version` table, and so does every other
+    Alembic application, so a URL naming someone else's database is accepted as
+    readily as otari's own. Alembic then reports only that it cannot locate the
+    revision it read, which reads as a corrupt database rather than as the URL
+    being wrong, and the next `upgrade` would write otari's schema into that
+    database. Naming the target and the foreign revision is what turns it back
+    into a configuration error.
+
+    On a connection of its own, so that reading the version table leaves no
+    transaction open on the one the migrations run in: Alembic commits its work
+    inside `begin_transaction`, which is a no-op when a transaction is already
+    under way, and the run would then be rolled back when the connection closes.
+    """
+    with engine.connect() as connection:
+        heads = MigrationContext.configure(connection).get_current_heads()
+    if not heads:
+        return
+
+    known = {script.revision for script in ScriptDirectory.from_config(config).walk_revisions()}
+    foreign = sorted(set(heads) - known)
+    if not foreign:
+        return
+
+    target = make_url(config.get_main_option("sqlalchemy.url") or "").render_as_string(hide_password=True)
+    msg = (
+        f"{target} is stamped with alembic revision(s) {', '.join(foreign)}, which are not otari's. "
+        "The database holds another application's migration history, or one written by a newer otari "
+        "than this one. Point OTARI_DATABASE_URL at otari's own database."
+    )
+    raise CommandError(msg)
 
 
 def run_migrations_offline() -> None:
@@ -81,6 +124,8 @@ def run_migrations_online() -> None:
         prefix="sqlalchemy.",
         poolclass=pool.NullPool,
     )
+
+    _reject_foreign_history(connectable)
 
     with connectable.connect() as connection:
         context.configure(connection=connection, target_metadata=target_metadata)
--- a/alembic/env.py
+++ b/alembic/env.py
@@ -33,14 +33,26 @@ if config.config_file_name is not None:
 target_metadata = Base.metadata
 
 # Get database URL from config (if already set programmatically) or the environment.
-# Priority: Programmatically set URL -> OTARI_DATABASE_URL -> default
+# Priority: Programmatically set URL -> OTARI_DATABASE_URL, and nothing else.
 #
 # A bare `DATABASE_URL` is deliberately not read. `otari serve` and `otari
 # migrate` both take that name as `--database-url` and pass the resolved URL on
-# explicitly, so the only invocation this fallback ever served was a bare
+# explicitly, so the only invocation that fallback ever served was a bare
 # `alembic upgrade head` in a process holding another application's
 # `DATABASE_URL`, which aims this chain at that application's database.
-database_url = config.get_main_option("sqlalchemy.url") or os.getenv("OTARI_DATABASE_URL") or "sqlite:///./otari.db"
+#
+# Nor is there a default. The `sqlite:///./otari.db` that stood here turned the
+# same run into a silent success: the chain migrated a throwaway file in the
+# working directory, reported "Running upgrade" and exited 0, while the database
+# the operator believed they were migrating went untouched.
+database_url = config.get_main_option("sqlalchemy.url") or os.getenv("OTARI_DATABASE_URL")
+if not database_url:
+    no_url = (
+        "No database URL to migrate. Set OTARI_DATABASE_URL, or go through `otari migrate`, which "
+        "resolves one and passes it on. A bare DATABASE_URL is not read here: it belongs to whichever "
+        "application put it in this environment."
+    )
+    raise CommandError(no_url)
 # Normalized here rather than at each call site so every entry point is covered:
 # `otari migrate`, auto_migrate on startup, and a bare `alembic upgrade head`
 # reading OTARI_DATABASE_URL. Migrations run on a sync engine, so an async URL
@@ -81,6 +93,15 @@ def _reject_foreign_history(engine: Engine) -> None:
     transaction open on the one the migrations run in: Alembic commits its work
     inside `begin_transaction`, which is a no-op when a transaction is already
     under way, and the run would then be rolled back when the connection closes.
+
+    Scoped to a neighbor that keeps its history in this same `alembic_version`
+    table. One that keeps it anywhere else, under a custom `version_table` or in
+    `django_migrations` or `flyway_schema_history`, leaves nothing here to read
+    and is accepted. Refusing those too would take a positive signal that the
+    database is otari's, and "has tables, none of them otari's" is the shape of a
+    legitimate first migration into a database an edition's own chain populated
+    first. Stamping a version table of otari's own is the structural answer, and
+    it is a migration for every deployment that exists.
     """
     if _is_stamp_command():
         return
@@ -95,12 +116,19 @@ def _reject_foreign_history(engine: Engine) -> None:
     if not foreign:
         return
 
+    # Host and database, not the URL: `render_as_string(hide_password=True)` masks
+    # the password field and not credentials passed as query parameters, which is
+    # how libpq takes `password` and `sslpassword`, and this string is echoed by
+    # `otari migrate`.
+    url = engine.url
+    target = f"{url.host}/{url.database}" if url.host else str(url.database)
     msg = (
-        f"{engine.url.render_as_string(hide_password=True)} is stamped with alembic revision(s) "
-        f"{', '.join(foreign)}, which are not otari's. The database holds another application's migration "
-        "history, or one written by a newer otari than this one. Point OTARI_DATABASE_URL at otari's own "
-        "database, or, where this database is otari's and the row is wrong, rewrite it with "
-        "`alembic stamp --purge <revision>`."
+        f"{target} is stamped with alembic revision(s) {', '.join(foreign)}, which are not otari's. "
+        "The database holds another application's migration history, or one written by a newer otari "
+        "than this one. Point OTARI_DATABASE_URL at otari's own database first. Once it is established "
+        "that this database is otari's and the row is the wrong one, `alembic stamp --purge <revision>` "
+        "rewrites it, replacing whatever the version table holds: run against a neighbor's database it "
+        "destroys that application's history."
     )
     raise CommandError(msg)
 
--- a/alembic/env.py
+++ b/alembic/env.py
@@ -7,7 +7,7 @@ from alembic.runtime.migration import MigrationContext
 from alembic.script import ScriptDirectory
 from alembic.util import CommandError
 from sqlalchemy import engine_from_config, pool
-from sqlalchemy.engine import Engine, make_url
+from sqlalchemy.engine import Engine
 
 from gateway.core.database import to_sync_url
 
@@ -53,6 +53,19 @@ config.set_main_option("sqlalchemy.url", to_sync_url(database_url))
 # ... etc.
 
 
+def _is_stamp_command() -> bool:
+    """Whether this run is `alembic stamp`, which the check below has to let through.
+
+    Rewriting the version table is how an operator repairs the state that check
+    refuses (`alembic stamp --purge <revision>`), so refusing to run it would
+    take away the fix along with the failure. `config.cmd_opts` is set by the
+    CLI and absent when alembic is driven programmatically, which is the
+    gateway's own path and never a stamp.
+    """
+    cmd = getattr(getattr(config, "cmd_opts", None), "cmd", None)
+    return bool(cmd) and getattr(cmd[0], "__name__", "") == "stamp"
+
+
 def _reject_foreign_history(engine: Engine) -> None:
     """Refuse a database whose stamped history is not this chain's.
 
@@ -69,6 +82,9 @@ def _reject_foreign_history(engine: Engine) -> None:
     inside `begin_transaction`, which is a no-op when a transaction is already
     under way, and the run would then be rolled back when the connection closes.
     """
+    if _is_stamp_command():
+        return
+
     with engine.connect() as connection:
         heads = MigrationContext.configure(connection).get_current_heads()
     if not heads:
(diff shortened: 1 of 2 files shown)


alembic/env.py
```diff
@@ -45,6 +53,58 @@
...
+    being wrong, and the next `upgrade` would write otari's schema into that
+    database. Naming the target and the foreign revision is what turns it back
+    into a configuration error.
+
+    On a connection of its own, so that reading the version table leaves no
+    transaction open on the one the migrations run in: Alembic commits its work
+    inside `begin_transaction`, which is a no-op when a transaction is already
+    under way, and the run would then be rolled back when the connection closes.
+    """
+    if _is_stamp_command():
```
khaledosman, line 106: Blocking, fixed in d79c4f62. Without this exemption the check ran for every online command, so `alembic stamp head --purge` was refused on exactly the database it is the repair for, and the PR would have removed the fix along with the failure. Confirmed both ways against the CLI on a foreign-stamped SQLite: the purge succeeds on `main` and failed with the first draft here. `tests/unit/test_alembic_target_database.py::test_stamp_still_rewrites_a_foreign_version_table` pins it and fails without these two lines.

Known limitation: `config.cmd_opts` is None when alembic is driven programmatically, so an in-process `command.stamp(...)` is still refused. Nothing in this repository calls it; a repair script that did would need its own exemption.

alembic/env.py
```diff
@@ -45,6 +53,58 @@
...
+        return
+
+    with engine.connect() as connection:
```
khaledosman, line 109: Fixed before this PR was opened, recorded because it is easy to reintroduce by simplifying the check onto the caller's connection. Reading the version table on the *migration* connection leaves a transaction open, alembic's `begin_transaction` is then a no-op, and the migration's own work is rolled back when the connection closes. The only symptom was `tests/unit/test_budget_consolidation_chain.py` failing with `no such table: organization`.

alembic/env.py
```diff
@@ -45,6 +53,58 @@
...
+    if not heads:
+        return
+
+    known = {script.revision for script in ScriptDirectory.from_config(config).walk_revisions()}
+    foreign = sorted(set(heads) - known)
+    if not foreign:
+        return
+
+    msg = (
+        f"{engine.url.render_as_string(hide_password=True)} is stamped with alembic revision(s) "
```
khaledosman: Fixed in d79c4f62. This read `make_url(config.get_main_option("sqlalchemy.url") or "")`: the `or ""` defends against a missing option with a value `make_url` rejects outright, and the URL is already on the engine we just connected with. Take it from there.

khaledosman (commented):
Self-review, written by me as an agent (Claude Opus 5 (1M context) via Claude Code) at @khaledosman's request, since I wrote this PR too.

Three findings, all fixed in d79c4f62. Comments are inline at the lines that carry the fixes.

Verified against the real CLI on a foreign-stamped SQLite database: `alembic current` already failed the same way before this PR, so the only change there is the message, and `alembic stamp head --purge` succeeded, which is what the first finding is about.

alembic/env.py
```diff
@@ -45,6 +53,58 @@
...
+    transaction open on the one the migrations run in: Alembic commits its work
+    inside `begin_transaction`, which is a no-op when a transaction is already
+    under way, and the run would then be rolled back when the connection closes.
+    """
+    if _is_stamp_command():
+        return
+
+    with engine.connect() as connection:
+        heads = MigrationContext.configure(connection).get_current_heads()
+    if not heads:
```
peteski22, line 111: **medium: the guard is bypassed for exactly the neighbor it most needs to catch.**

`if not heads: return` accepts any database with no default `alembic_version` table, so only a neighbor that happens to stamp *that* table is detected. Reproduced on this branch against a SQLite database seeded with `django_migrations` and `auth_user`:

```
$ OTARI_DATABASE_URL=sqlite:///django.db alembic upgrade head
INFO  [alembic.runtime.migration] Running upgrade c9f2a6b4e8d7 -> d5b7f9a1c3e6, ...
EXIT=0
table count: 44
django_migrations still there: True
otari tables present: ['alembic_version', 'api_keys', 'usage_logs', 'users']
```

No warning, exit 0, otari's whole schema written alongside the neighbor's tables. The same holds for Flyway (`flyway_schema_history`), Prisma (`_prisma_migrations`), and an Alembic application configured with a custom `version_table`.

Since the stated property is running otari's chain only against otari's own database, catching the general case needs a positive identity signal (for example: refuse when the database has tables but none of otari's) rather than a foreign-revision check.

alembic/env.py
```diff
@@ -28,11 +32,15 @@
...
-)
+# Get database URL from config (if already set programmatically) or the environment.
+# Priority: Programmatically set URL -> OTARI_DATABASE_URL -> default
+#
+# A bare `DATABASE_URL` is deliberately not read. `otari serve` and `otari
+# migrate` both take that name as `--database-url` and pass the resolved URL on
+# explicitly, so the only invocation this fallback ever served was a bare
+# `alembic upgrade head` in a process holding another application's
+# `DATABASE_URL`, which aims this chain at that application's database.
+database_url = config.get_main_option("sqlalchemy.url") or os.getenv("OTARI_DATABASE_URL") or "sqlite:///./otari.db"
```
peteski22: **medium: dropping the `DATABASE_URL` fallback turns a wrong-database run into a silent success.**

Removing the fallback is the right call and nothing depended on it: the Railway and Render templates set `OTARI_DATABASE_URL`, and `cli.py` reads `DATABASE_URL` itself (`cli.py:158`) and re-exports it (`cli.py:199`). The problem is what an operator lands on instead. Reproduced:

```
$ cd /tmp/x && DATABASE_URL=postgresql://u:p@prod-host.internal:5432/prod alembic upgrade head
INFO  [alembic.runtime.migration] Running upgrade c9f2a6b4e8d7 -> d5b7f9a1c3e6, ...
$ ls
otari.db
```

The chain runs to completion against a throwaway file in the cwd and exits 0 with a clean "Running upgrade" log. The operator reads a successful migration; `prod` is untouched and unmigrated. It is also the one wrong-database path no guard can catch, because an empty version table hits the early return above.

`test_a_bare_alembic_run_ignores_a_foreign_database_url` currently documents this behavior rather than the intended one: `assert not foreign.exists()` passes because the run silently created `otari.db` in `tmp_path`. Raising when no URL is resolvable would make the failure legible, and would let that test assert a refusal instead.

alembic/env.py
```diff
@@ -45,6 +53,58 @@
...
+    known = {script.revision for script in ScriptDirectory.from_config(config).walk_revisions()}
+    foreign = sorted(set(heads) - known)
+    if not foreign:
+        return
+
+    msg = (
+        f"{engine.url.render_as_string(hide_password=True)} is stamped with alembic revision(s) "
+        f"{', '.join(foreign)}, which are not otari's. The database holds another application's migration "
+        "history, or one written by a newer otari than this one. Point OTARI_DATABASE_URL at otari's own "
+        "database, or, where this database is otari's and the row is wrong, rewrite it with "
```
peteski22: **low: this remedy destroys the neighbor's history if run at the moment it is printed.**

When this fires, the configured URL *is* the foreign database, so copy-pasting the command without first changing the URL replaces the other application's version row. Reproduced end to end:

```
$ alembic upgrade head
FAILED: sqlite:////tmp/foreign.db is stamped with alembic revision(s) neighbor_rev_abc123, ...

$ alembic stamp --purge d5b7f9a1c3e6      # the suggested command, verbatim
INFO  [alembic.runtime.migration] Running stamp_revision  -> d5b7f9a1c3e6

$ select version_num from alembic_version
[('d5b7f9a1c3e6',)]                        # neighbor_rev_abc123 is gone
```

The "where this database is otari's" qualifier is doing all the work. Putting the fix-the-URL step first, or naming the destructive effect on `--purge`, keeps the repair without the footgun.

Worth noting the non-destructive variant is already safe, so only `--purge` needs the warning:

```
$ alembic stamp head
FAILED: Can't locate revision identified by 'neighbor_rev_abc123'
$ select version_num from alembic_version
[('neighbor_rev_abc123',)]                 # intact
```

alembic/env.py
peteski22: **low: `hide_password=True` masks only the password field, not credentials in query parameters.**

```
in:  postgresql+psycopg://u:sekret@h:5432/db?password=alsosekret&sslpassword=pem123
out: postgresql+psycopg://u:***@h:5432/db?password=alsosekret&sslpassword=pem123
```

libpq accepts `password` and `sslpassword` as URI query parameters, so that form does reach deployments. This string goes into a `CommandError` that propagates through `init_db` on the `auto_migrate` path and is echoed to stderr by `otari migrate`, against the repo's never-log-secrets rule. Rendering only the host and database name carries the same diagnostic weight.

(Out of scope here, but `cli.py:194` already echoes the full URL unmasked on the `otari migrate` path, so that rule is bent on this path either way. Worth a separate issue.)

peteski22 (commented):
**Review produced by an agent, posted on behalf of @peteski22.**

Every finding below was reproduced locally against real SQLite databases; the evidence is inline with each one. `make lint` and the three new tests pass on `9069824`.

The guard does what it says for the case it targets: a database stamped `neighbor_rev_abc123` is refused before any write, with the target and the foreign revision both named, and `stamp` is correctly let through. Two gaps are worth settling before merge, both about how far "otari's own database" is actually established.

**Summary**

| | Where | Point |
|---|---|---|
| medium | `alembic/env.py:90` | Guard is bypassed when the foreign database is not Alembic-managed |
| medium | `alembic/env.py:43` | Wrong-database run now silently succeeds against `./otari.db` |
| low | `alembic/env.py:102` | Suggested remedy destroys the neighbor's history if run as printed |
| low | `alembic/env.py:99` | `hide_password` misses credentials passed as query parameters |
| note | `alembic/env.py:93` | Open-core: accepted revision set has no configuration seam |
| nit | `alembic/env.py:89` | Duplicates Alembic's context INFO output on every online run |
| test | test file | No coverage for a non-Alembic foreign database; one assertion passes for the wrong reason |

**Checked and cleared**

- `_is_stamp_command` reads Alembic 1.18.4's `cmd_opts.cmd` shape correctly (`(fn, positional, kwarg)`, `fn.__name__ == "stamp"`), and `cmd_opts` is set on the `Config` before the command runs.
- The separate connection genuinely avoids leaving a transaction open on the migration connection. A full `upgrade head` and a second idempotent run both work.
- No offline-mode gap: `run_migrations_offline` never connects.
- `tests/integration/conftest.py` runs `command.upgrade` on a fresh database per worker, so the empty version table hits the early return and the guard does not affect the integration suite.
- One candidate finding I dropped after checking: the remedy's `OTARI_DATABASE_URL` **does** take effect on the `auto_migrate` and `otari migrate` paths, even though `sqlalchemy.url` outranks the env var inside `env.py`. `_apply_otari_env_overrides` layers `OTARI_DATABASE_URL` over YAML into the init dict, so it reaches `config.database_url`, which `_run_migrations` then passes as `sqlalchemy.url`. Only an explicit `--database-url` outranks it, and there the operator supplied the URL directly. Verified by resolving `load_config(None).database_url` with the variable set.
- Nothing in `Dockerfile`, `docker-compose*.yml`, `.github/workflows/`, or the Railway and Render templates ran a bare `alembic` relying on `DATABASE_URL`.

Posting as a comment rather than a change request; the two mediums are judgment calls about intended scope, not defects in what is here.

alembic/env.py
```diff
@@ -45,6 +53,58 @@
...
+    """
+    if _is_stamp_command():
+        return
+
+    with engine.connect() as connection:
+        heads = MigrationContext.configure(connection).get_current_heads()
+    if not heads:
+        return
+
+    known = {script.revision for script in ScriptDirectory.from_config(config).walk_revisions()}
```
peteski22, line 114: **note, open-core alignment, not a blocker.**

`known` is derived solely from this checkout's `script_location`, and the version table is the hardcoded default, so the accepted set has no configuration seam and `env.py` is a core file. `ARCHITECTURE.md` rule 6: "An overlay never edits an Otari source file [...] If extending an overlay *requires* editing an Otari file, that is a missing seam, and the seam belongs in Otari."

Nothing in `ARCHITECTURE.md` establishes that an overlay ships migrations of its own today, so this is forward-looking rather than a violation, and no port or adapter boundary is crossed. I am flagging it because the structural alternative, scoping otari's chain to its own `version_table`, would deliver the commit's stated property by construction and would close both mediums above at the same time, at the cost of a one-time migration for existing deployments. Worth an explicit decision either way rather than arriving there later.

alembic/env.py
```diff
@@ -45,6 +53,58 @@
...
+    On a connection of its own, so that reading the version table leaves no
+    transaction open on the one the migrations run in: Alembic commits its work
+    inside `begin_transaction`, which is a no-op when a transaction is already
+    under way, and the run would then be rolled back when the connection closes.
+    """
+    if _is_stamp_command():
+        return
+
+    with engine.connect() as connection:
+        heads = MigrationContext.configure(connection).get_current_heads()
```
peteski22, line 110: **nit: this duplicates Alembic's context INFO output on every online run.**

`alembic current` against the same database, `main` vs this branch:

```
main:   Context impl SQLiteImpl.    x1
branch: Context impl SQLiteImpl.    x2
        Will assume non-transactional DDL.  x2
```

So every migration log now reads as though the context was built twice. Reading `alembic_version` directly on the separate connection would avoid it. Cosmetic, noting it rather than asking for a change.

tests/unit/test_alembic_target_database.py
```diff
@@ -0,0 +1,130 @@
...
+    picked up the value some other application left in the environment.
+    """
+    foreign = tmp_path / "platform.db"
+    env = {key: value for key, value in os.environ.items() if key != "OTARI_DATABASE_URL"}
+    env["DATABASE_URL"] = f"sqlite:///{foreign}"
+
+    result = _run_alembic(["current"], cwd=tmp_path, env=env)
+
+    assert result.returncode == 0, result.stderr
+    assert not foreign.exists(), "DATABASE_URL selected the database this chain connected to"
```
peteski22, line 134: **test coverage.** Two things on this file:

1. Nothing covers a foreign database that is **not** Alembic-managed, which is the case the guard silently accepts (see my comment on `alembic/env.py:90`). A test seeding `django_migrations` and asserting the refusal would pin the property the module docstring claims.

2. This assertion passes for the wrong reason. The run does not refuse: it falls back to `sqlite:///./otari.db` and creates `otari.db` in `tmp_path`, so `foreign` was never going to exist. Verified:

```
$ cd $tmp && DATABASE_URL=sqlite:///$tmp/platform.db alembic current
INFO  [alembic.runtime.migration] Context impl SQLiteImpl.
$ ls $tmp
otari.db
```

Asserting on the resolved target, or that the run failed, would make the test hold for the reason its docstring gives.

alembic/env.py
khaledosman: Fixed in 19f532ae, and thank you for the repro: this was a regression the PR introduced. Dropping the fallback moved the wrong-database run from "migrates the wrong database" to "migrates a throwaway file and reports success", which is the worse of the two.

There is no default now. A run with no resolvable URL raises before it connects:

```
$ cd /tmp/x && DATABASE_URL=postgresql://u:p@prod-host.internal:5432/prod alembic upgrade head
FAILED: No database URL to migrate. Set OTARI_DATABASE_URL, or go through `otari migrate`, which
resolves one and passes it on. A bare DATABASE_URL is not read here: it belongs to whichever
application put it in this environment.
$ ls -A
$
```

Nothing depended on that default: `otari serve` and `otari migrate` both set `sqlalchemy.url` programmatically (`GatewayConfig.database_url` keeps the SQLite default for the OSS path, so that is unchanged), `revision_environment` is off so `alembic revision -m` never runs `env.py`, and nothing in the Makefile, CI, docs or the deploy templates runs a bare alembic. The test asserts the refusal now, and additionally that the working directory is left empty.

🤖 Reply written by Claude Opus 5 (1M context) via Claude Code, on @khaledosman's behalf.

alembic/env.py
khaledosman: Confirmed, and the docstring now says it: the check reads this one `alembic_version` table, so a neighbor keeping its history under a custom `version_table`, or in `django_migrations` or `flyway_schema_history`, leaves nothing to read and is accepted.

I am not taking the positive-identity remedy here, because "has tables, none of them otari's" is also the shape of a legitimate first migration. Downstream, otari-ai's enterprise overlay registers its tables in this same database and now ships its own chain on `overlay_alembic_version` (mozilla-ai/otari-ai#2014), and its `OVERLAY_DB_AUTO_CREATE` boot step can run before otari's chain ever has. A refusal on "no otari tables" would fail that boot.

So the general case needs the structural answer rather than a wider heuristic, which is your note on line 93. Replied there.

🤖 Reply written by Claude Opus 5 (1M context) via Claude Code, on @khaledosman's behalf.

alembic/env.py
khaledosman: Fixed in 19f532ae. The remedy now leads with the URL and names what `--purge` does to whatever row is there:

> Point OTARI_DATABASE_URL at otari's own database first. Once it is established that this database is otari's and the row is the wrong one, `alembic stamp --purge <revision>` rewrites it, replacing whatever the version table holds: run against a neighbor's database it destroys that application's history.

Your observation that plain `alembic stamp head` is refused by alembic itself and leaves the row intact is what makes this safe to keep in the message at all.

🤖 Reply written by Claude Opus 5 (1M context) via Claude Code, on @khaledosman's behalf.

alembic/env.py
khaledosman: Fixed in 19f532ae. The message renders host and database only, so there is no URL string to leak a query parameter through:

```python
url = engine.url
target = f"{url.host}/{url.database}" if url.host else str(url.database)
```

Agreed that the `cli.py` echo is the same rule bent on the same path, and out of scope here. Filed as #922, which covers both call sites (`:168` and `:195`) and carries your query-parameter evidence, since `hide_password=True` alone would not be enough there either.

🤖 Reply written by Claude Opus 5 (1M context) via Claude Code, on @khaledosman's behalf.

alembic/env.py
khaledosman: Agreed, and I think this is the real answer to your line 90 finding rather than a forward-looking note: a `version_table` of otari's own delivers the property by construction, and no revision-set heuristic gets there.

What it costs is a one-time migration for every deployment that exists, prod included: the new table has to be seeded from the current `alembic_version` at first boot, and a deployment that rolls back to an image predating that seed reads an empty table and tries to migrate from base. That is a bigger decision than this PR, and it is yours rather than mine.

Worth noting the downstream edition already went that way (`overlay_alembic_version`, mozilla-ai/otari-ai#2014), so `otari_alembic_version` would make the pattern uniform across the two chains that share a database. Filed as #921 with that scope written up, including the rollback shape, which is the part that needs deciding rather than implementing: an image predating the seed reads an empty table of its own and migrates from base, which is a worse failure than the one it replaces unless the old row stays readable.

🤖 Reply written by Claude Opus 5 (1M context) via Claude Code, on @khaledosman's behalf.

alembic/env.py
khaledosman: Noted, and I am leaving it. `MigrationContext.configure` is the same read alembic performs, with the table-existence handling that goes with it; replacing it with a raw `SELECT` to save one INFO line would hardcode `alembic_version` in a second place and lose that. The duplicate line is worth less than the drift.

🤖 Reply written by Claude Opus 5 (1M context) via Claude Code, on @khaledosman's behalf.

tests/unit/test_alembic_target_database.py
khaledosman: Point 2 is fixed in 19f532ae: the test asserts a refusal now (non-zero exit, `OTARI_DATABASE_URL` named in stderr) and adds `assert list(tmp_path.iterdir()) == []`, so it can no longer pass because a default quietly migrated something else.

Point 1 I have not added, deliberately. A test seeding `django_migrations` would pin the behavior as it is, which is acceptance, and I did not widen the guard to refuse it (reasons on line 90). Codifying the limitation in a test reads as intent; the docstring now states it as a limitation instead.

🤖 Reply written by Claude Opus 5 (1M context) via Claude Code, on @khaledosman's behalf.

PR #936, opened 2026-09-04
fix(dashboard): let the pricing review and share frames be dismissed

### LINKED ISSUE #917: Form modals cannot be dismissed with Esc or a backdrop click
**Otari version:** current `main` (dashboard)
**Run mode:** standalone
**How you're running Otari:** from source / dashboard

### What happened

On Model pricing, **Add rate override** (and **Edit rate override**) cannot be dismissed with Escape or by clicking the dimmed backdrop. The only way out is **Cancel**.

This is not unique to that dialog. Most dashboard overlays are HeroUI `AlertDialog`. In HeroUI v3 (`@heroui/react` 3.2.4) `AlertDialog.Backdrop` defaults to `isDismissable={false}` and `isKeyboardDismissDisabled={true}` on purpose for real alerts. Form and share overlays inherit those defaults even though they are not alerts.

The dashboard already has the dismissable primitive: Create workspace in `WorkspaceSwitcher` uses `Modal` (`isDismissable` defaults to `true`). It also sets `Modal.Backdrop className="bg-backdrop/50"` because Modal's dim is otherwise fully opaque.

### Steps to reproduce

1. Open the dashboard, go to Model pricing.
2. Click Add override.
3. Press Escape, or click the dimmed area outside the dialog.

### What you expected

Form overlays should close on Esc and on backdrop click, the same way Create workspace already does. Destructive confirms should still require an explicit Cancel or Confirm.

### Audit (other overlays)

**Same bug (form / share overlays built as `AlertDialog`):**

- `web/src/features/organization/PricingOverrideDialog.tsx` — Add / Edit rate override (reported)
- `web/src/features/models/SetPriceDialog.tsx` — Set price / bulk price (Models, Activity)
- `web/src/features/usage/ShareDialog.tsx` — Share this view as an image (Usage)
- `web/src/features/pricing/ModelPricingPage.tsx` — Review default price updates (`PricingRefreshDialog`). Esc / backdrop never fire `onOpenChange(false)`, so the existing reject path cannot run that way.

**Correct as `AlertDialog` (explicit action is the point):**

- `web/src/shared/components/ConfirmDialog.tsx` and every delete / remove confirm that uses it (keys, budgets, workspaces, members, rate override delete, activity)
- `web/src/features/settings/SettingsPage.tsx` — Regenerate master key. After the new key is shown, `onOpenChange` already refuses to close until **I've saved this key**.

**Already dismissable:**

- `web/src/app/nav/WorkspaceSwitcher.tsx` — Create workspace (`Modal`)

### Suggested fix

Move the form and share overlays onto `Modal` (matching Create workspace, including the `bg-backdrop/50` backdrop). Leave `ConfirmDialog` and master-key rotation on `AlertDialog`. Cover Esc and backdrop dismiss in the colocated tests; Cancel-click coverage alone does not catch this.

mikemikimike:
## Description

Fixes #917. The pricing review and image share dialogs could not be closed: no Escape, no backdrop press, no close control.

Both hand-rolled a HeroUI `AlertDialog`, the shell the design system retired. They now use `Dialog`, the plain frame for something to read and act on, which is dismissable by default. So the behavior comes from the frame rather than from something each call site has to remember.

Two things a reviewer should weigh:

Dismissing the pricing review no longer rejects it. That dialog's close handler was wired to `reject()`, which discards the fetched prices on the server. It was unreachable while the dialog could not be closed; making it closable would have meant a stray Escape throwing the update away. Closing now just puts the dialog away and leaves the update pending, so the page's notice offers it again. Rejecting stays the footer's own labeled button, and neither exit works while a request is in flight.

The share dialog was already narrower than intended: its 940px width was a class the stylesheet discarded, capping it at 672px. It takes a new `xl` size in the shared dialog sizes instead, which is the only layer that rule can be beaten from. Its footer `Close` is gone, since the frame now supplies one.

The branch was force-pushed. Its previous tip was a bad merge that resurrected a test file deleted on `main`, dropped three unrelated pricing tests and two imports, which is what CI was failing on. `main` has also grown `FormDialog` since, and that already fixed `SetPriceDialog`, `PricingOverrideDialog` and `RateOverridesCard`, so those are no longer part of this change.

## How to test it locally

Open the model pricing page as an operator, press "Check for price updates", and press Escape. The dialog closes and the pending update is still offered. Then open the usage page's share dialog and press Escape or click outside it.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes #917

## Checklist

- [x] I ran the Definition of Done checks locally (lint, typecheck, test, build).
- [x] If the API contract changed, I regenerated the OpenAPI spec (no API change here).

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:** Claude Opus 5 via Claude Code (this revision); GPT-5 Codex (original revision)


web/src/features/usage/ShareDialog.tsx
```diff
@@ -250,16 +250,19 @@ export function ShareDialog(props: ShareDialogProps) {
...
+          <Modal.Dialog aria-label="Share this view as an image" className="w-[92vw] max-w-[940px]">
+            <Modal.Header>
+              <Modal.Heading>
                 Share this view as an image
-              </AlertDialog.Heading>
-            </AlertDialog.Header>
-            <AlertDialog.Body className="flex max-h-[70vh] flex-col gap-4 overflow-y-auto">
+              </Modal.Heading>
+            </Modal.Header>
+            <Modal.Body className="flex max-h-[70vh] flex-col gap-4 overflow-y-auto">
```
khaledosman: `overflow-y-auto` is inert here: `globals.css:916` pins `.modal__body { overflow: visible }` unlayered, so it outranks the utility. With `max-h-[70vh]` still applied the body no longer scrolls and content past the fold spills out of the dialog.

`KeysPage.tsx:174` hit exactly this and fixed it with `Modal.Container scroll="outside"`. Do the same on line 258 and drop the dead `overflow-y-auto`.

web/src/features/usage/ShareDialog.tsx
```diff
@@ -250,16 +250,19 @@ export function ShareDialog(props: ShareDialogProps) {
...
-          <AlertDialog.Dialog className="w-[92vw] max-w-[940px]">
-            <AlertDialog.Header>
-              <AlertDialog.Heading>
+    <Modal isOpen onOpenChange={(open) => (open ? undefined : onClose())}>
+      <Modal.Trigger aria-hidden="true" className="hidden">
+        Share this view as an image
+      </Modal.Trigger>
+      <Modal.Backdrop className="bg-backdrop/50">
+        <Modal.Container placement="center" size="lg">
+          <Modal.Dialog aria-label="Share this view as an image" className="w-[92vw] max-w-[940px]">
```
khaledosman: `w-[92vw] max-w-[940px]` is inert for the same reason: `globals.css:911` sets `.modal__dialog { width: fit-content; min-width: min(28rem, …); max-width: min(42rem, …) }` unlayered. `AlertDialog` had no counterpart rule, so this dialog rendered up to 940px and now caps at 672px with a fit-content width, squeezing the side-by-side preview and controls below.

This needs an escape the cascade respects (an `.otari-*` wide modifier next to that rule, or relaxing the rule) rather than a class the unlayered rule drops.

web/src/features/pricing/ModelPricingPage.test.tsx
```diff
@@ -272,15 +272,15 @@ describe("ModelPricingPage", () => {
     await user.click(
       screen.getByRole("button", { name: "Check for price updates" }),
     )
-    await screen.findByRole("alertdialog", {
+    await screen.findByRole("dialog", {
       name: "Review default price updates",
     })
 
-    await user.click(screen.getByRole("button", { name: "Reject changes" }))
+    await user.click(document.querySelector('[data-slot="modal-backdrop"]')!)
```
khaledosman: This is the only click on "Reject changes" in the file, and the test is still named `rejects a reviewed default price update`. Swapping it for a backdrop click leaves that button with no coverage while the name no longer describes what runs. Keep the button click here and add the backdrop/Escape dismissal as its own test.

web/src/features/models/SetPriceDialog.tsx
```diff
@@ -153,15 +153,18 @@ export function SetPriceDialog({
...
+      </Modal.Trigger>
       {isOpen ? (
-        <AlertDialog.Backdrop>
-          <AlertDialog.Container placement="center" size="lg">
-            <AlertDialog.Dialog>
-              <AlertDialog.Header>
-                <AlertDialog.Heading>{title}</AlertDialog.Heading>
-              </AlertDialog.Header>
-              <AlertDialog.Body className="flex flex-col gap-4">
+        <Modal.Backdrop className="bg-backdrop/50">
```
khaledosman: Cancel is deliberately `isDisabled={isPending}` (line 218), but this backdrop is unconditionally dismissible, so Escape or a backdrop click now does exactly what the disabled button prevents: closing the form mid-mutation, losing the error banner the request may still produce.

`Modal.Backdrop isDismissable={!isPending} isKeyboardDismissDisabled={isPending}` keeps the two agreeing; `KeysPage.tsx:172` uses those props.

web/src/features/pricing/ModelPricingPage.tsx
```diff
@@ -162,13 +162,13 @@ function PricingRefreshSection() {
           <ErrorBanner error={previewRefresh.error} />
         </Card.Content>
       </Card>
-      <AlertDialog
+      <Modal
         isOpen={preview !== undefined}
         onOpenChange={(isOpen) => (!isOpen ? reject() : undefined)}
```
khaledosman, line 171: Dismissal now runs `reject()`, which POSTs `/v1/pricing/refresh/reject`, so a stray Escape discards the fetched preview server-side. This is the only one of the four converted dialogs where dismissing has a server side effect; worth confirming that is intended rather than inherited from an `onOpenChange` that was previously unreachable.

It is also unguarded while `isPending`: `reject()` returns early, but the modal still closes on the accept mutation in flight.

web/src/features/organization/PricingOverrideDialog.tsx
```diff
@@ -191,17 +191,20 @@ export function PricingOverrideDialog({
...
+      <Modal.Trigger aria-hidden="true" className="hidden">
+        {editing ? "Edit rate override" : "Add rate override"}
+      </Modal.Trigger>
       {isOpen ? (
-        <AlertDialog.Backdrop>
-          <AlertDialog.Container placement="center" size="lg">
-            <AlertDialog.Dialog>
-              <AlertDialog.Header>
-                <AlertDialog.Heading>
+        <Modal.Backdrop className="bg-backdrop/50">
```
khaledosman: `bg-backdrop/50` is dead on all four new backdrops: `globals.css:933` already dims `.modal__backdrop--opaque` to 50% and, being unlayered, wins over this class. The comment there says it was centralized precisely so "a future one would start out wrong again" could not happen, so new call sites should not carry it.

web/src/features/organization/PricingOverrideDialog.tsx
```diff
@@ -191,17 +191,20 @@ export function PricingOverrideDialog({
...
+          <Modal.Container placement="center" size="lg">
+            <Modal.Dialog aria-label={editing ? "Edit rate override" : "Add rate override"}>
```
khaledosman: `aria-label` duplicates `Modal.Heading` below, which already supplies the accessible name (the pre-existing `findByRole("alertdialog", { name: … })` queries passed with no `aria-label`, and `KeysPage` passes a Heading and no label). `WorkspaceSwitcher` uses `aria-label` because its dialog has no Heading. Drop it here and in `SetPriceDialog`, `ModelPricingPage`, `ShareDialog`.

web/src/features/organization/RateOverridesCard.test.tsx
```diff
@@ -194,6 +194,25 @@ describe("RateOverridesCard", () => {
...
+    mockApi({ overrides: [] })
+    const user = userEvent.setup()
+
+    await renderPage()
+    await user.click(
+      await screen.findByRole("button", { name: /add override/i }),
+    )
+    await screen.findByRole("dialog", { name: "Add rate override" })
+
+    await user.click(document.querySelector('[data-slot="modal-backdrop"]')!)
```
khaledosman: `document.querySelector('[data-slot="modal-backdrop"]')` couples the test to a HeroUI internal slot name, repeated in four files. A backdrop has no role to query, so the escape hatch is fair, but it belongs behind one helper in `web/src/tests/` so a HeroUI rename is a one-line fix.

khaledosman (changes requested):
The `AlertDialog` -> `Modal` swap crosses a cascade boundary this repo has already documented: `globals.css` carries **unlayered** rules against `.modal__*` and none against `.alert-dialog__*`, and unlayered beats `@layer utilities` (`globals.css:1024`: "a Tailwind class at a call site cannot override one of them"). Two `className`s on `ShareDialog` that worked under `AlertDialog` are now discarded, which is a visible layout regression the new dismissal tests cannot see.

Also: `pnpm --dir web run lint` fails. `biome check` includes the formatter and 5 of the 8 changed files would be reformatted (`ModelsPage.test.tsx`, `PricingOverrideDialog.tsx`, `RateOverridesCard.test.tsx`, `ModelPricingPage.tsx`, `ShareDialog.tsx`). The PR ran `biome lint`, which skips formatting; `pnpm --dir web run lint:fix` settles it.

web/src/features/usage/ShareDialog.tsx
khaledosman: Fixed. The frame is now `Dialog`, which owns its own scrolling body, so there is no `overflow-y-auto` or `max-h-[70vh]` at this call site to be discarded.

web/src/features/usage/ShareDialog.tsx
khaledosman: Fixed, and you were right that it needed an escape the cascade respects. `w-[92vw] max-w-[940px]` was already losing to the unlayered `.modal__dialog` rule, so the frame takes a new `xl` (928px) in the shared dialog size family in `globals.css`, which is the layer that rule can be beaten from. `DialogSize`/`FormDialogSize` and `web/design/feedback.md` carry it.

web/src/features/pricing/ModelPricingPage.test.tsx
khaledosman: Fixed. `rejects a reviewed default price update` is back and still clicks the button. The dismissal cases are their own parametrized test, and they assert the opposite: that Escape and a backdrop press send **no** reject.

web/src/features/models/SetPriceDialog.tsx
khaledosman: Moot on current main: `SetPriceDialog` uses `FormDialog` now, which already gates dismissal on `isPending` (`requestClose` returns early, and the backdrop takes `isDismissable && !isPending`). Out of this PR's diff.

web/src/features/pricing/ModelPricingPage.tsx
khaledosman: Fixed, and it was not intended. Dismissing now calls `close()`, which is local only; the preview stays pending server-side and the page's notice offers it again. `reject()` is reachable from the footer's labeled control alone. The frame is also `isDismissable={!isPending}`, so neither dismissal lands mid-mutation, and `close()` now resets both mutations so a dismissed error does not reappear on reopen.

web/src/features/organization/PricingOverrideDialog.tsx
khaledosman: Moot on current main: `PricingOverrideDialog` uses `FormDialog`, which carries that exact note over its backdrop and spells no `bg-backdrop/50`. Out of this PR's diff.

web/src/features/organization/PricingOverrideDialog.tsx
khaledosman: Moot on current main for the same reason: `FormDialog` and `Dialog` both supply the name through `Modal.Heading` and neither takes an `aria-label` from a call site. Out of this PR's diff.

web/src/features/organization/RateOverridesCard.test.tsx
khaledosman: Fixed. `getModalBackdrop()` in `web/src/tests/modal.ts` is the one place the slot name is spelled, with a comment saying why the escape hatch is fair.

web/src/features/models/ModelsPage.test.tsx
```diff
@@ -0,0 +1,1480 @@
...
+  DiscoverableModelsResponse,
+  GatewaySettings,
+  ModelListResponse,
+  ModelMetadata,
+  ModelMetadataResponse,
+  ModelObject,
+  OrganizationContext,
+  PricingResponse,
+} from "@/client"
+import { ModelsPage } from "@/features/models/ModelsPage"
```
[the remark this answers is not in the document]
khaledosman: Fixed. That file should not have existed: `ModelsPage.tsx` and its test were both deleted on main in #1015, and a bad merge resolution resurrected the test alone. It is gone.

web/src/features/pricing/ModelPricingPage.tsx
```diff
@@ -1,6 +1,5 @@
-import { AlertDialog, Button } from "@heroui/react"
-import { Link, useNavigate } from "@tanstack/react-router"
-import { useState } from "react"
+import { Button, Modal } from "@heroui/react"
+import { Link } from "@tanstack/react-router"
```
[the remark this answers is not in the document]
khaledosman: Fixed. Both imports were casualties of the same bad merge. `ModelPricingPage.tsx` no longer imports `AlertDialog` at all and keeps `useState` and `useNavigate`; `pnpm --dir web run typecheck` passes.

web/src/features/pricing/ModelPricingPage.test.tsx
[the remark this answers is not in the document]
khaledosman: Re-addressed after the rebase: Escape and backdrop dismissal are both covered for the pricing review, asserting that neither sends a reject.

khaledosman:
@coderabbitai full review

The branch was rewritten onto current `main`, so an incremental pass would be against a history that no longer exists. Previous findings are all resolved; the change now uses the design system's `Dialog` rather than hand-rolled `Modal.*`.

khaledosman (approved):

PR #938, opened 2026-09-04
fix(messages): adopt any-llm container support

### LINKED ISSUE #924: Drop the hand-declared container once the any-llm pin models it on MessagesParams
Blocked on the any-llm release carrying mozilla-ai/any-llm#1329 (merged, not in 1.26.0).

#916 hand-declares `container` on `MessagesRequest` (`src/gateway/api/routes/messages.py`) because any-llm 1.25.0 does not model it on `MessagesParams`, so the derived base dropped a caller's value before the provider call. That declaration is a stopgap. Once the SDK pin carries the param, remove it so the field comes from derivation like every other Messages param.

The trigger is the version bump, which Dependabot lands on its own, so this issue is the thing that surfaces the cleanup at that moment.

### Not a plain revert

A subclass field fully overrides the derived one (see `_schema_derive`), so as long as the declaration stays, otari's `str | None` wins over whatever annotation any-llm chooses. Upstream picked the same type today, so the shadow is invisible; if it later widens to Anthropic's beta shape (`BetaContainerParams | str`, which is what the Agent Skills flow sends), our narrower type silently persists in `docs/public/openapi.json` and in every generated SDK, and no test flags it.

Steps:

1. Delete the `container` field from `MessagesRequest`, keeping the derived one, and drop `container` from `_pipeline._FORWARDED_PARAMS`: derivation puts it there via `MessagesParams.model_fields`.
2. Regenerate `docs/public/openapi.json` and `make postman`, then confirm the annotation matches what any-llm declares rather than staying `string | null` by accident. The `MessagesRequest` docstring names the field, so regenerate `web/src/client/schema.ts` too.
3. Keep the route tests in `tests/integration/test_messages_route_dispatch.py`. They assert the forwarding contract end to end and hold whether the field is derived or hand-declared, including the drop when `otari_code_execution` puts the sandbox in charge.
4. Re-check `tests/unit/test_request_schema_derivation.py::test_messages_openapi_request_schema_includes_container`, which pins the exact `anyOf` shape in the published spec and will fail if the type widens.

Mirrors #565 for `service_tier`.

HareeshBahuleyan:
### Why

Fixes #924. any-llm 1.27 models `container` directly on `MessagesParams`, so Otari no longer needs to declare the field locally or register it as an extra forwarded parameter.

any-llm 1.27.1 constrains Anthropic to the compatible `>=0.119.0,<1` range. This preserves the typed-refusal, container, and service-tier functionality introduced upstream without resolving to the incompatible Anthropic 1.x API.

### What changed

- Upgraded any-llm to 1.27.1.
- Removed the temporary local `container` declaration and its `_FORWARDED_PARAMS` entry.
- Preserved the managed-credential refusal, sandbox filtering, and conflicting-sandbox guards.
- Preserved the bridged-provider rejection path as a caller-facing 400.
- Updated usage fixtures for the upstream per-iteration `model` field.
- Updated the generated OpenAPI and TypeScript schemas with upstream `container` and `service_tier` support.
- Removed the interim httpx2 provider-error classification changes after the 1.27.1 lock resolved Anthropic to 0.125.0.

Anthropic 1.x support remains tracked by mozilla-ai/any-llm#1347.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes #924

## Checklist

- [ ] I ran the full Definition of Done suite on the final 1.27.1 diff. Focused unit tests, lint, type checking, OpenAPI, and Postman checks passed; the full suite was not rerun.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:** OpenAI GPT-5.6 SOL with Pi; Claude Opus 5; OpenAI Codex

**Any additional AI details you would like to share:** GPT-5.6 SOL implemented the container cleanup and dependency update. Claude Opus 5 reviewed the change and added the replacement container-rejection test. OpenAI Codex audited the Anthropic SDK compatibility boundary, prepared the upstream pre-v1 constraint, and updated this description. The interim httpx2 classifier work was removed after any-llm 1.27.1 became available.

**NOTE:**
When responding to reviewer questions, please respond yourself rather than copy/pasting reviewer comments into an AI and pasting back its answer. We want to discuss with you, not our AI :)

## Summary

- Updated any-llm support for `container` and `service_tier`.
- Removed Otari’s temporary local `container` declaration and forwarding configuration.
- Updated generated schemas and test fixtures for upstream field support.
- Updated timeout coverage for the supported SDK transport.
- Preserved sandbox, credential, and bridged-provider safeguards.

## Technical notes

- The dependency now requires any-llm `>=1.27.1`.
- Anthropic resolves to `0.125.0`.
- Anthropic 1.x support remains out of scope.

### DIFF
--- a/src/gateway/api/routes/_pipeline.py
+++ b/src/gateway/api/routes/_pipeline.py
@@ -389,11 +389,6 @@ _FORWARDED_PARAMS: frozenset[str] = frozenset(
         # Declared on the chat request rather than derived, and forwarded through
         # any-llm's ``**kwargs`` (see ``chat.ChatCompletionRequest.service_tier``).
         | {"service_tier"}
-        # Same shape on the Messages request (see
-        # ``messages.MessagesRequest.container``). A provider that bridges
-        # Messages through Chat Completions has no parameter for it, so it
-        # arrives here as the unknown-keyword TypeError.
-        | {"container"}
     )
     - SENSITIVE_PARAM_FIELDS
 )
--- a/src/gateway/api/routes/_pipeline.py
+++ b/src/gateway/api/routes/_pipeline.py
@@ -386,9 +386,6 @@ _FORWARDED_PARAMS: frozenset[str] = frozenset(
         set(CompletionParams.model_fields)
         | set(MessagesParams.model_fields)
         | set(ResponsesParams.model_fields)
-        # Declared on the chat request rather than derived, and forwarded through
-        # any-llm's ``**kwargs`` (see ``chat.ChatCompletionRequest.service_tier``).
-        | {"service_tier"}
     )
     - SENSITIVE_PARAM_FIELDS
 )
--- a/tests/unit/test_provider_error_classification.py
+++ b/tests/unit/test_provider_error_classification.py
@@ -14,7 +14,6 @@ upstream text.
 import asyncio
 
 import httpx
-import httpx2
 import pytest
 from anthropic import APITimeoutError as AnthropicAPITimeoutError
 from any_llm.exceptions import UnsupportedParameterError
@@ -96,21 +95,9 @@ def test_sdk_wrapped_timeout_maps_to_504() -> None:
     as a 504, not fall through to the generic 502."""
     for exc in (
         OpenAIAPITimeoutError(request=httpx.Request("POST", "http://upstream")),
-        AnthropicAPITimeoutError(request=httpx2.Request("POST", "http://upstream")),
+        AnthropicAPITimeoutError(request=httpx.Request("POST", "http://upstream")),
     ):
         assert classify_provider_error(exc) == (504, PROVIDER_TIMEOUT_DETAIL)
-
-
-@pytest.mark.parametrize("exc", [httpx2.ReadTimeout("slow"), httpx2.ConnectTimeout("slow")])
-def test_raw_httpx2_timeout_maps_to_504(exc: BaseException) -> None:
-    """Both SDKs wrap only the initial send, so a read timeout while the SSE body
-    is already streaming reaches the classifier as the HTTP client's own raw
-    exception. The Anthropic SDK's client is httpx2, whose exceptions share no
-    base class with httpx's and match neither ``*TimeoutError`` nor
-    ``*ConnectionError`` by name, so matching httpx alone would settle a real
-    provider timeout as the generic 502 an outage gets."""
-    assert classify_provider_error(exc) == (504, PROVIDER_TIMEOUT_DETAIL)
-    assert failure_status_code(exc) == 504
 
 
 def test_unified_any_llm_wrapped_timeout_maps_to_504() -> None:
--- a/tests/unit/test_provider_error_classification.py
+++ b/tests/unit/test_provider_error_classification.py
@@ -98,6 +98,7 @@ def test_sdk_wrapped_timeout_maps_to_504() -> None:
         AnthropicAPITimeoutError(request=httpx.Request("POST", "http://upstream")),
     ):
         assert classify_provider_error(exc) == (504, PROVIDER_TIMEOUT_DETAIL)
+        assert failure_status_code(exc) == 504
 
 
 def test_unified_any_llm_wrapped_timeout_maps_to_504() -> None:
--- a/tests/unit/test_provider_error_classification.py
+++ b/tests/unit/test_provider_error_classification.py
@@ -14,6 +14,7 @@ upstream text.
 import asyncio
 
 import httpx
+import httpx2
 import pytest
 from anthropic import APITimeoutError as AnthropicAPITimeoutError
 from any_llm.exceptions import UnsupportedParameterError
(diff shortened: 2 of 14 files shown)


src/gateway/api/routes/_pipeline.py
```diff
@@ -389,11 +389,6 @@ def _redacted_caller_fault_detail(message: str, fallback: str) -> str:
         # Declared on the chat request rather than derived, and forwarded through
         # any-llm's ``**kwargs`` (see ``chat.ChatCompletionRequest.service_tier``).
         | {"service_tier"}
```
khaledosman: The `service_tier` twin of the entry removed just below is now dead, and its comment is false. any-llm 1.27 declares `service_tier` on both `CompletionParams` and `MessagesParams` (verified against the 1.27.0 wheel), so this union member adds nothing to the derived set, and "Declared on the chat request rather than derived, and forwarded through any-llm's ``**kwargs``" no longer describes what happens. Drop these three lines alongside `container`.

tests/unit/test_provider_error_classification.py
```diff
@@ -359,19 +362,6 @@ def test_rejected_param_maps_to_400_naming_the_param() -> None:
...
-    misses the caller-fault gate and a permanent, caller-fixable rejection is
-    reported as a generic upstream failure."""
-    exc = TypeError("AsyncCompletions.create() got an unexpected keyword argument 'container'")
-    mapping = classify_provider_error(exc)
-    assert mapping is not None
-    assert mapping.status_code == 400
-    assert "'container'" in mapping.detail
-
-
 def test_rejected_param_detail_does_not_name_the_sdk_internals() -> None:
```
khaledosman, line 380: The deleted `test_rejected_container_maps_to_400_naming_the_param` guarded a guarantee that still holds, just through a different branch, and nothing asserts it now. Under 1.27 a bridged provider never reaches the `TypeError`: `BaseProvider._amessages` raises `NotImplementedError("container requires a provider with a native Anthropic Messages API")` before the Messages -> Completions conversion, which `_is_unsupported_feature_error` maps to a 400 carrying that message as the detail.

Re-point the test at that exception rather than deleting it, otherwise `container` against a non-Anthropic model is unguarded against regressing to a generic 502.

khaledosman (approved):
Verified the upgrade's premises against a scratch install of any-llm 1.27.0 / anthropic 1.3.0: `MessagesParams` gains exactly `container: str | None` and `service_tier: str | None`, `CompletionParams` gains `service_tier: str | None`, `ResponsesParams` is unchanged, so the spec and `schema.ts` delta is complete and `container`'s wire shape does not change. `any-llm 1.27` forces `anthropic>=0.119.0`, so the 0.88 -> 1.3.0 (httpx -> httpx2) jump is driven by the bump rather than incidental lock churn; both anthropic import sites still hold (`APITimeoutError`/`APIConnectionError` for `upstream_exception_shape`, and the `ServerToolUseBlock`/`WebSearchToolResultBlock` constructions in `mcp_loop_messages`, whose new `caller` field stays off the wire under `model_dump(exclude_none=True)`). `make postman-check` passes at head as well: the collection does not embed component-schema descriptions, so the spec change leaves no Postman diff.

Two notes that have no line in this diff to hang on:

**`src/gateway/api/routes/chat.py:104`** - this PR satisfies #565's stated exit condition. The comment above that field reads "remove this declaration once `CompletionParams` models the param and the SDK pin is bumped (any-llm#1269, tracked in #565)", and 1.27's `CompletionParams.service_tier` is `str | None`, identical to the hand declaration, so removal changes neither the spec nor the wire contract, and `test_service_tier_reaches_provider` already guards forwarding. Either fold it in and close #565 alongside #924, or note on #565 that it is deliberately deferred.

**Commit message body** - it carries literal `\n\n` rather than newlines ("...tenant-isolation protections.\n\nRefresh generated schemas..."). Squash-merge concatenates commit bodies, so that can land in main's history.

---
🤖 Review by [Claude Code](https://claude.com/claude-code) (Opus 5), posted by @khaledosman

src/gateway/api/routes/_pipeline.py
HareeshBahuleyan: Fixed in 2a10769c. Removed the redundant service_tier union member and its stale comment; the any-llm parameter models remain the source of the derived set.

tests/unit/test_provider_error_classification.py
HareeshBahuleyan: Addressed in 9c94118e. The replacement regression test covers the NotImplementedError emitted for container on the bridged Messages path, and it remains valid with any-llm 1.27.1.

tests/unit/test_provider_error_classification.py
```diff
@@ -93,8 +93,10 @@ def test_sdk_wrapped_timeout_maps_to_504() -> None:
     ``APITimeoutError`` (no ``status_code``, not an httpx exception instance).
     any-llm surfaces that wrapped type directly, so it must still classify
     as a 504, not fall through to the generic 502."""
-    request = httpx.Request("POST", "http://upstream")
-    for exc in (OpenAIAPITimeoutError(request=request), AnthropicAPITimeoutError(request=request)):
+    for exc in (
+        OpenAIAPITimeoutError(request=httpx.Request("POST", "http://upstream")),
+        AnthropicAPITimeoutError(request=httpx.Request("POST", "http://upstream")),
```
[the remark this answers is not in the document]
HareeshBahuleyan: Fixed in 2a10769c. The SDK timeout loop now also asserts that failure_status_code returns 504 for both OpenAI and Anthropic timeout exceptions.

PR #934, opened 2026-09-03
feat(dashboard): land the divided-surface redesign on current main

jigjigjig:
**Redesigns the dashboard onto one flat surface: no cards, square corners, hairline rules,
one teal accent.** Twelve pages rebuilt on a shared set of primitives, plus the palette and
type scale underneath them.

**And, merged in from its own branch, the three-variant button work:** seven button variants
cut to three across 176 call sites, one press value everywhere, and the pressed fill reaching
a pointer press for the first time. It has its own section below, including what it costs.

It replaces two stacked draft PRs, #871 (palette) and #885 (structure), which were 34 commits
behind `main` and could not be rebased separately. Those stay open for their author to close.

| | |
| --- | --- |
| Scope | 148 files, 46 commits, `web/` plus two `src/gateway/templates/` brand assets |
| Behavior | Visual only. No API contract change, no new dependency |
| Risk | 15 defects were introduced *by the rebase itself* and fixed. See below |
| CI | All 12 checks green, including the Python jobs the template assets select |

## What a reviewer should check

**1. The rebase produced 15 real defects, and none was caught by a compiler or a linter.**
Six were lost authorization gates, two of which exposed another member's data. All are fixed;
the table below records what caught each, because that is the argument for how carefully the
rest was done.

| Caught by | Defects | Worst example |
| --- | --- | --- |
| Identifier accounting | 4 | A budget-exemption control kept while its operator-only gate was dropped: a member could exempt their own key from their own budget |
| Unit tests | 3 | The scoped-budget picker rendered for every caller |
| Typecheck, once it could run | 2 | Price write controls rendered for non-operators |
| Full suite | 1 | Three model filters ungated; their own test passed because the redesign renamed the control it asserted on |
| Touch-target sweep | 1 | A 44px pan rail became a 10px stripe |
| Re-reading one element | 3 | A focus ring, described twice as an improvement, never rendered |
| A Python job | 1 | The favicon served starting with an XML comment |

**2. Three checks were changed rather than satisfied.** Two type-scale rules could not express
a distinction the design record makes, and a heading rule rejected a role that exists to head a
group. None changes anything rendered. The exemption list carries a guard that fails if an
exempted site stops existing.

**3. The dashboard is not uniformly converted.** 23 surfaces carry the new page header, 4 still
use the old one, 6 components still use `Card`. `OverviewPage` renders both, so it is
half-converted on the page most people open first.

## The three-variant button work, merged in

[...]

PR #939, opened 2026-09-04
refactor(dashboard): cut seven button variants to three

jigjigjig:
## Description

Cuts the dashboard's seven button variants to three (primary, ghost, danger), makes the press one value everywhere, and makes the pressed fill reach a pointer press for the first time. 176 Button call sites, ten commits, and every number below was measured on the running dashboard rather than derived from the stylesheet.

Stacked deliberately on `feat/dashboard-redesign` (#934) rather than on `main`, and kept separate from it for one reason: this change touches controls on every page, and it has to be revertible on its own. Because this repository squash-merges, a single sha is what makes that possible.

**To undo this entire change:** `git revert <squash sha of this PR>`. That restores the seven variants, the size-stepped press ladder, the filled danger, ghost's edgeless rest state, the focus-ring list and the old confirm labels, and removes the place rules, the new place class and the enforcement scan. It does not touch #934's work, which is why the two are separate PRs.

**The revert is rehearsed, not asserted.** On a scratch branch I squash-merged this PR onto the base the way GitHub will, reverted that single commit, and checked what came back:

- `git diff feat/dashboard-redesign` against the reverted tree is **empty**. Nothing is left behind.
- `lint`, `typecheck` and `test` are all green on the reverted state, at 3057 tests (the pre-scan count, since the scan reverts with the change).
- The bundle rebuilds from the reverted tree to **`index-CNLUdEUK.js` / `index-5YVrOLEX.css`**, byte-identical filenames to the bundle before this work started.
- That build boots and renders with **zero page errors and zero console errors**: `variant="outline"` is back on the pager with its 1px `rgb(90, 95, 107)` edge, ghost is back to `border-width: 0px`, the `.button--lg` press is back to `scale(.96)`, and our unlayered pressed-fill rule is absent from the emitted CSS.
- No generated artifact moves in either direction: `git status` is clean after the revert and after the rebuild.

### The three variants

| from | count | to |
|---|---|---|
| `primary` | 73 | `primary`, unchanged |
| `ghost` | 64 | `ghost`, new edge and press ramp |
| `outline` | 26 | `ghost` |
| `secondary` | 2 | `ghost` |
| `danger` | 6 | `danger`, unfilled |
| `danger-soft` | 4 | `danger` |
| `tertiary` | 0 | declared, never used |

**176 Button call sites, not 178.** `variant="…"` returns 177 matches, of which two are `<Chip variant="secondary">` (`features/budgets/SpendCeilingsCard.tsx:119`, `features/budgets/BudgetsPage.tsx:642`) and not Buttons at all; the one a prop grep misses is `features/settings/SettingsPage.tsx:482`, which calls `buttonVariants({ variant: "danger-soft" })` to build a className.

**Outline to ghost is an identity outside a place, and that is a fact about the tokens rather than luck.** `--color-control-border` IS `--color-text-subtle`, in both theme blocks, so the edge outline drew and the edge ghost now draws are the same value. Measured before and after on the three converted sites that render outside any place (/settings' "Re-encrypt provider credentials" and "Send test email", /providers' "Re-check all"): 1px `rgb(90, 95, 107)`, padding 11px, height 32px, both ways. A reviewer looking at 26 converted call sites should read them as an identity rather than as 26 chances to drift; the ones that change are the ones inside a place, where the edge is suppressed on purpose.

### Ghost's edge needed three properties where the design named one

HeroUI gives `.button--outline` `border-width: 1px` and gives ghost **none**. Measured on the running page before any edit, a ghost button's computed `border-top-width` was `0px` while its `border-color` was already set to the divider alpha. So setting only the colour, which is what the design asked for, would have compiled, linted, passed every gate, shipped and painted nothing. The rule sets `border-style`, `border-width` and `border-color` together for that reason.

The colour is `--color-text-subtle`, opaque: **6.01** against the page in light and **6.13** in dark, where the divider alpha it replaces measured **1.14** and **1.18**. No new token; the same one in a job it was not being given.

### Ghost's press ramp, and the coupling that justifies the flat press

Ghost gains a third step, from the pair `.list-box-item` already uses: transparent, `--color-surface-alt` on hover, `--color-surface-subtle` pressed. Rest to pressed goes 1.05 to **1.22** in dark and 1.06 to **1.14** in light. The light half is marginal and stated rather than trimmed: no neutral fill in this palette does better, the whole ramp spans 1.05 to 1.22.

**The press is 0.98 at every rung**, replacing a ladder that deepened with size, and **`.button--full-width` now stops pressing at all** (HeroUI's own `fullWidth` prop emits that class, which the existing exclusion did not name). Measured on the sign-in button at 471px: `transform: none` under a real press, where the ladder moved it 7.07px per edge.

**These two are coupled and neither is defensible alone.** The flat 0.98 is justified by the pressed fill existing, because a flat transform is a smaller acknowledgement than the ladder it replaces (1.27px at md against 1.90px). And the pressed fill has never rendered under a pointer: HeroUI declares the pressed and hover backgrounds in one layer at equal specificity with hover last (byte 74094 against 74220 in the emitted stylesheet), so a pointer that is down is also hovering and hover won. The step itself was never broken, which is what makes this a cascade fix rather than a state fix: pressing without hovering (focus a control, hold Space) already painted `--color-surface-subtle`, with `data-pressed="true"` and matching neither `:hover` nor `:active`. Restating the declaration unlayered fixes it, and this is the first time anyone has seen that fill paint under a pointer:

| control | rest | hover | pressed |
|---|---|---|---|
| primary | `rgb(0, 128, 138)` | `rgb(0, 109, 118)` | `rgb(0, 91, 98)` |
| ghost, in a place | transparent | `rgb(240, 241, 242)` | `rgb(232, 233, 235)` |
| ghost, no place | transparent | `rgb(240, 241, 242)` | `rgb(232, 233, 235)` |
| danger | transparent | `rgb(255, 236, 233)` | `rgb(255, 236, 233)` |

All measured with `:active`, `:hover` and `data-pressed="true"` all true at the moment of reading. **Danger is the asymmetry:** the palette has one subtle danger value, so danger's press is carried by the ink and the transform rather than by a third fill. That is a consequence of the palette rather than an oversight, and it is why the row above reads the same twice.

Consequence worth stating plainly: the pressed fill now renders for pointer users on **all** variants including primary, which is a state nobody has seen in the shipped app. It matches the approved prototype, so it is restoration, but restoration describes its relation to the approved artifact and not to what anyone's eyes are used to.

### Danger, once, unfilled

Transparent ground, danger ink, danger edge, filling to `--color-danger-subtle` on hover. A filled destructive button read as a second primary, which is what absorbing `danger-soft` is for. Measured at rest: transparent with a `rgb(203, 37, 38)` edge and ink; `--color-danger` on the page ground is 5.15 in light and 5.54 in dark, both clearing 4.5 as text.

It also **leaves the filled-control focus-ring list**, which that rule's own comment asks for since it names fills rather than components. Measured on the button, `--focus` now reads `#00848f` where the list would have made it near-black in light and near-white in dark.

### Places, and why one class name is new

Where an edge would be too loud it is suppressed by naming the place, not by remembering a modifier at each call site: `.otari-toolbar`, `.otari-pagination`, `.otari-table` and the per-page table classes already work this way for inputs and pagination controls, and `.otari-bulk-bar` and `.otari-breakdown` already existed.

`.otari-rail` is new, and it is a **correction rather than an addition**: the approved prototype declared four places (`place-toolbar`, `place-row`, `place-cell`, `place-nav`) and this stylesheet had a word for only three. `WorkspaceSwitcher`'s own comment records that the artboard removed its border because it "made the rail open with an outlined card sitting above a list of flat rows", so an opaque edge there would undo a decision already taken.

`place-row` deliberately has no counterpart here. Every candidate for it was already covered or wanted the edge: a table row's ghosts are inside `.otari-table`; the prototype demonstrates `place-row` in exactly one spot, as the container for its four icon-only ghosts, which the rule below covers by what the control is; a form row's single trailing action reads better WITH the edge, compared both ways on /routing; and a table row's Edit and Delete are `RowAction`, a plain caption button outside this family. A place nobody consumes is a place that goes stale.

One rule is not a place, deliberately: **an icon-only ghost never takes the edge.** `CopyButton` renders inside tables, inside the models detail panel, inside banners and standalone on a page, so no container can reach it, and a rule keyed on what the control *is* cannot be forgotten by a call site either.

### The reduction is enforced, and the scan is mutation-checked

`Button` is HeroUI's, re-exported directly, so its `variant` prop is typed by HeroUI's union and still accepts every retired name: deleting our CSS for `secondary` does not make `variant="secondary"` a type error, it makes it a silently unstyled button.

The scan in `foundation.test.ts` is keyed on the **host** rather than on the word, which is load-bearing: `Chip` has its own variant union that shares `secondary`, and two legitimate `<Chip variant="secondary">` render on the Budgets page, so a scan for the word alone would have failed on correct code the first time anyone ran it. A gate that fails on correct code is one people learn to suppress.

Three mutation checks, because a scan that cannot fail is worse than no scan:

1. One real `variant="ghost"` in `features/keys/KeysPage.tsx` changed to `variant="secondary"`: the sweep fails and names the file and the line.
2. Restored: green, 124 files and 3201 tests.
3. Green with the two Chips in place, and the host keying is pinned by two assertions in the block, so a future edit that loosened it to a word match fails its own test.

### The two-step destructive confirm

`ConfirmButton`'s resting trigger was `danger-soft` and its armed confirm `danger`; with danger-soft absorbed they became the same variant at the same size. The first click is safe, so the resting trigger is now neutral and the danger hue marks the irreversible step. Measured at rest: resting is a ghost with `rgb(8, 9, 10)` ink and a `rgb(90, 95, 107)` edge; armed is danger, ink and edge `rgb(203, 37, 38)`, with the ghost Cancel beside it.

**What gives, and it is a regression against what ships today.** The escalation is hue-only and slightly quieter rather than louder: the ghost edge measures 6.01/6.13 and the danger edge 5.15/5.54, so arming *changes* edge contrast by -0.86 in light and -0.59 in dark. Ghost edge against danger edge is 1.17 light and 1.11 dark, which is no luminance difference at all, and under simulated protanopia in light it is **1.03**. Today's escalation is a pale filled box to a saturated filled box, a fill and luminance change that survives colour deficiency; ours replaces it with a hue change that does not. The user chose this with those figures in front of them.

What carries it instead is structural, and the component's docstring now says so: **the Cancel appearing when armed is load-bearing, not decorative.** A second control appearing and the layout changing is the one cue no colour deficiency hides. Anybody simplifying the armed state by dropping it takes the escalation to zero for those operators.

The labels are part of that answer rather than a tidy-up. All five call sites passed the same word twice, so the two-step pattern carried no information in its labels anywhere it was used; the armed label is what an affected operator reads to learn what the second click does. Now: `Remove domain` then `Remove claim`; `Remove tool` and `Remove guardrail` then `Remove permanently`; `Reset price` then `Reset to default`. Domains is deliberately not "permanently": `delete_domain_for_user` drops a claim row that can be re-added, and its docstring records that memberships already auto-joined through the claim are left alone on purpose, so "permanently" would overstate it in two directions at once.

### The sign-in screen

`e2e/otari.yml` configures no passkey relying party and no OAuth client, so "Use a passkey" and the per-provider buttons never render against this fixture and were the only surface here shipping on a CSS-level claim. `e2e/screenshots/public.spec.ts` now stubs the bootstrap the way its OAuth callback captures already do and takes one capture of that state.

Measured on it: each of the three is a full-width ghost at 471x44, transparent, ink `rgb(8, 9, 10)`, edge 1px `rgb(90, 95, 107)`, padding 15px, beside the filled primary at 471x44 and padding 16px. A press hovers, presses and does not scale, which is the full-width exclusion demonstrated on the widest buttons in the product. **One filled primary above three edged alternatives reads as a hierarchy, where three filled greys would have read as four primaries** — that is the argument for the conversion rather than a consolation for it.

### Regressions and costs, in one place

- **The pager's disabled arrows lose a channel.** With the boxes gone, "which arrows can I press" rests on opacity alone (0.4 against 1.0) at glyph size, where it used to be a box and opacity. Put to the user with that cost stated; they chose bare.
- **The refresh control loses its box everywhere.** It is icon-only, so the icon rule reaches it regardless of place. Same trade as the pager, on three pages.
- **The confirm's escalation is hue-only** and measurably quieter, carried structurally by the Cancel (above).
- **Danger has no third press fill**, because the palette has one subtle danger value (above).
- **The pressed fill now paints for pointer users on every variant**, primary included, which no shipped build has done (above).

### Not in this change

Loading is a capability rather than a repair and ships as its own PR afterwards, so a revert of this one cannot take it out. No colour value moves; roles are reassigned and not one hex changes. Sizes stay at 32/36/40 and `features/auth/Login.tsx`'s `h-11` override stays, flagged separately. Brand teal on primary is measured, unapplied and left with the user. The two-layer focus ring, square corners, both typefaces and the 100ms timing are untouched. The disabled cursor is its own item. A filled armed confirm was offered and declined. `RowActions` is noted as the latent home if a row place is ever wanted.

`globals.css` stays recognisably `otari-ai/frontend/src/index.css`: every change here is a rule in a variant block or the utilities layer, not a token value, so the cross-repo cost is low.

## PR Type

- [x] Refactor
- [ ] New Feature
- [ ] Bug Fix
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

None filed; this comes from a design review of the shipped button set.

## Checklist

- [x] I have added or updated tests that cover my change. The enforcement scan in `src/styles/foundation.test.ts`, mutation-checked three ways above; a screenshot capture of the sign-in screen's alternatives; and two component tests updated to the new confirm labels. What is deliberately NOT tested is the appearance itself: jsdom renders none of this CSS, so a component test asserting it would pin class strings rather than behaviour, which is a test that cannot fail for the reason we care about.
- [x] I ran the Definition of Done checks locally. The dashboard's four, which are the ones this touches: `pnpm --dir web run lint`, `run typecheck`, `test` (124 files, 3201 tests) and `run build`. No Python changed, so `make lint` and `make test` cover nothing in this diff.
- [ ] Documentation was updated where necessary. Nothing to update: no user-facing behaviour beyond the appearance described here, and no contract change.
- [ ] If the API contract changed, I regenerated the OpenAPI spec. No API change; `client/schema.ts` and `routeTree.gen.ts` both come back clean.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

Every figure here was read off the running dashboard or the emitted stylesheet, and three of them corrected the plan this started from: ghost's zero border-width, a padding compensation danger needed that only appeared beside a ghost of the same size, and the pressed fill that no pointer press has ever painted. The place audit that produced `.otari-rail` and deleted `.otari-row` was done by reading all 64 ghost call sites and their rendered containers rather than by grep.

### DIFF
--- a/web/e2e/screenshots/public.spec.ts
+++ b/web/e2e/screenshots/public.spec.ts
@@ -13,6 +13,23 @@ test("sign-in screen", async ({ page }) => {
   await captureScreenshot(page, "sign-in")
 })
 
+test("sign-in screen offering a passkey and OAuth", async ({ page }) => {
+  // The one sign-in state this fixture cannot reach on its own: `e2e/otari.yml`
+  // configures no passkey relying party and no OAuth client, so the buttons
+  // below never render against it, and they are the two controls on this screen
+  // that carry a variant rather than being the primary. Stubbed on the one
+  // response the shell reads before it mounts, the way the OAuth callback
+  // captures below already do it, rather than changing the fixture underneath
+  // `parity.bootstrap.spec.ts`, which asserts the empty lists as part of what
+  // this deployment IS.
+  await withPasskeyAndOauth(page)
+  await page.goto("/")
+  await expect(
+    page.getByRole("button", { name: "Use a passkey" }),
+  ).toBeVisible()
+  await captureScreenshot(page, "sign-in-alternatives")
+})
+
 test("sign-in screen with a rejected key", async ({ page }) => {
   await page.goto("/")
   const key = page.locator('input[type="password"]')
@@ -49,6 +66,25 @@ async function withMailReady(page: Page): Promise<void> {
     await route.fulfill({
       response,
       json: { ...bootstrap, mail_ready: true },
+    })
+  })
+}
+
+// The sign-in screen's alternatives: a passkey needs `passkeys_ready` and a
+// relying party, and each OAuth button needs its provider configured. Same
+// treatment and same reason as the two stubs below it.
+async function withPasskeyAndOauth(page: Page): Promise<void> {
+  await page.route("**/v1/bootstrap", async (route) => {
+    const response = await route.fetch()
+    const bootstrap = await response.json()
+    await route.fulfill({
+      response,
+      json: {
+        ...bootstrap,
+        sign_in_methods: [...bootstrap.sign_in_methods, "passkey"],
+        passkeys_ready: true,
+        oauth_providers: ["google", "github"],
+      },
     })
   })
 }
--- a/web/src/app/AppShell.tsx
+++ b/web/src/app/AppShell.tsx
@@ -696,7 +696,10 @@ function AppShellChrome() {
           tabIndex={-1}
           inert={isMobile && !mobileNavOpen ? true : undefined}
           className={clsx(
-            "flex flex-col border-r border-border bg-background focus:outline-none",
+            // `otari-rail`: the rail is a named place, the way the toolbar and
+            // the tables already are, and a ghost control inside it drops its
+            // edge rather than each of them remembering to.
+            "otari-rail flex flex-col border-r border-border bg-background focus:outline-none",
             isMobile
               ? clsx(
                   // Full width, starting below the top bar: `top-14` pairs with
--- a/web/src/features/activity/ActivityPage.tsx
+++ b/web/src/features/activity/ActivityPage.tsx
@@ -198,7 +198,7 @@ function InFlightControl({
   if (data.total === 0 && !isOpen) return null
   return (
     <Popover isOpen={isOpen} onOpenChange={setIsOpen}>
-      <Button size="sm" variant="outline">
+      <Button size="sm" variant="ghost">
         {/* Decorative: the count beside it carries the same meaning in text, so
             nothing is encoded in motion alone. That is also why it can stop
             outright under `prefers-reduced-motion`, being the one element on the
@@ -1072,7 +1072,7 @@ function RequestDetail({
           {onPriceModel ? (
             <Button
               size="sm"
-              variant="outline"
+              variant="ghost"
               onPress={() => onPriceModel(pricingKey)}
             >
               Price this model
@@ -2068,7 +2068,7 @@ export function ActivityPage() {
               {newRows > 0 ? (
                 <Button
                   size="sm"
-                  variant="outline"
+                  variant="ghost"
                   onPress={refresh}
                   isDisabled={usage.isFetching}
                 >
--- a/web/src/features/auth/Login.tsx
+++ b/web/src/features/auth/Login.tsx
@@ -780,7 +780,7 @@ export function Login() {
             {offersPasskey ? (
               <Button
                 type="button"
-                variant="secondary"
+                variant="ghost"
                 fullWidth
                 isDisabled={
                   isSubmitting || isSigningOut || pendingProvider !== null
@@ -800,7 +800,7 @@ export function Login() {
                 <Button
                   key={provider}
                   type="button"
-                  variant="secondary"
+                  variant="ghost"
                   fullWidth
                   isDisabled={
                     isSubmitting ||
--- a/web/src/features/invitations/PendingInvitationsPage.tsx
+++ b/web/src/features/invitations/PendingInvitationsPage.tsx
@@ -129,7 +129,7 @@ export function PendingInvitationsPage() {
                     Accept
                   </Button>
                   <Button
-                    variant="outline"
+                    variant="ghost"
                     aria-label={`Decline invitation to ${invitation.organization_name}`}
                     onPress={() => setDeclining(invitation)}
                   >
--- a/web/src/features/keys/KeysPage.tsx
+++ b/web/src/features/keys/KeysPage.tsx
@@ -1258,7 +1258,7 @@ export function KeysPage() {
         >
           <Button
             size="sm"
-            variant="outline"
+            variant="ghost"
             isDisabled={bulkPending}
             onPress={() =>
               void runBulk(selectedKeys, (k) =>
@@ -1273,7 +1273,7 @@ export function KeysPage() {
           {isDeploymentWide ? (
             <Button
               size="sm"
-              variant="outline"
+              variant="ghost"
               isDisabled={bulkPending}
               onPress={() =>
                 void runBulk(selectedKeys, (k) =>
--- a/web/src/features/models/ModelsPage.tsx
+++ b/web/src/features/models/ModelsPage.tsx
@@ -438,7 +438,7 @@ function PricingTierEditor({
             request.
           </p>
         </div>
-        <Button size="sm" variant="outline" onPress={add}>
+        <Button size="sm" variant="ghost" onPress={add}>
           Add tier
         </Button>
       </div>
@@ -844,7 +844,7 @@ function PanelPriceEditor({
       />
       {readOnly ? null : (
         <div className="flex items-center gap-2 pt-1">
-          <Button size="sm" variant="outline" onPress={startEdit}>
+          <Button size="sm" variant="ghost" onPress={startEdit}>
             {row.source === "configured" ? "Edit price" : "Set price"}
           </Button>
           {row.source === "configured" ? (
@@ -1603,7 +1603,7 @@ function DiscoveredErrors({
               the wrong provider without saying so. */}
             <Button
               size="sm"
-              variant="outline"
+              variant="ghost"
               className="mt-2"
               onPress={() =>
                 onPriceModel(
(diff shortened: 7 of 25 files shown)


jigjigjig:
Folded into #934 rather than reviewed separately, at the user's call and with the trade stated: that branch squash-merges, so the button work no longer reverts independently of the redesign. Merged into `feat/dashboard-redesign` as 2033b298, a real merge so the ten commits `eb35a968..8d4e61cc` stay readable on the branch.

Everything this description carried now lives in #934's, including the token identity behind the 26 `outline` conversions, the zero-border-width finding, the coupling between the flat press and the pointer fill with its measurements, the five regressions in one place, the three mutation checks, and what a full revert was rehearsed to restore.

Closing as merged into the branch, not as abandoned.

PR #942, opened 2026-09-05
fix(cli): redact database URL password in migrate and init-db output

### LINKED ISSUE #922: `otari migrate` and `otari init-db` echo the database URL with its password
`src/gateway/cli.py:168` and `src/gateway/cli.py:195` echo `gateway_config.database_url` verbatim:

```python
click.echo(f"Initializing database: {gateway_config.database_url}")
click.echo(f"Running migrations on: {gateway_config.database_url}")
```

A production DSN carries its password, so `otari init-db` and `otari migrate` print it to stdout and into whatever
collects that output. `AGENTS.md` and the backend standards say never log secrets, and these are the paths that do.

Worth noting for whoever fixes it that `URL.render_as_string(hide_password=True)` is not sufficient on its own:
libpq accepts `password` and `sslpassword` as URI query parameters, which that flag does not mask.

```
in:  postgresql+psycopg://u:sekret@h:5432/db?password=alsosekret&sslpassword=pem123
out: postgresql+psycopg://u:***@h:5432/db?password=alsosekret&sslpassword=pem123
```

Host and database name carry the same diagnostic weight and cannot hold a credential, which is what #919 does in
its refusal message.

Refs #919.

_Filed by Claude Opus 5 (1M context) via Claude Code at @khaledosman's request._

AloysJehwin:
## Description
Redact the database URL password before echoing in `otari migrate` and `otari init-db` using SQLAlchemy's `make_url(...).render_as_string(hide_password=True)`.

## PR Type
- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues
Fixes #922

## Checklist

## AI Usage
- [ ] No AI was used.

### DIFF
--- a/src/gateway/cli.py
+++ b/src/gateway/cli.py
@@ -7,6 +7,7 @@ import sys
 
 import click
 import uvicorn
+from sqlalchemy.engine.url import make_url
 from uvicorn.config import logger
 
 from gateway.core.config import load_config
@@ -165,7 +166,8 @@ def init_db(config: str | None, database_url: str | None) -> None:
     if database_url:
         gateway_config.database_url = database_url
 
-    click.echo(f"Initializing database: {gateway_config.database_url}")
+    safe_url = make_url(gateway_config.database_url).render_as_string(hide_password=True)
+    click.echo(f"Initializing database: {safe_url}")
 
     db_init(gateway_config)
 
@@ -192,7 +194,8 @@ def migrate(config: str | None, database_url: str | None, revision: str) -> None
         click.echo("alembic command not found in PATH", err=True)
         sys.exit(1)
 
-    click.echo(f"Running migrations on: {gateway_config.database_url}")
+    safe_url = make_url(gateway_config.database_url).render_as_string(hide_password=True)
+    click.echo(f"Running migrations on: {safe_url}")
     click.echo(f"Target revision: {revision}")
 
     env = os.environ.copy()


src/gateway/cli.py
```diff
@@ -165,7 +166,8 @@ def init_db(config: str | None, database_url: str | None) -> None:
     if database_url:
         gateway_config.database_url = database_url
 
-    click.echo(f"Initializing database: {gateway_config.database_url}")
+    safe_url = make_url(gateway_config.database_url).render_as_string(hide_password=True)
```
khaledosman, line 169: Two problems with `make_url(...).render_as_string(hide_password=True)`, both fixed by the same swap.

**The password still prints when it travels as a query parameter.** `hide_password` masks the authority field only. Verified against this tree:

```
in : postgresql+psycopg://u:sekret@h:5432/db?password=alsosekret&sslpassword=pem123
out: postgresql+psycopg://u:***@h:5432/db?password=alsosekret&sslpassword=pem123
```

libpq accepts both `password` and `sslpassword` as URI parameters, so a production DSN that carries its credential there is echoed intact.

**A malformed URL becomes a traceback.** `database_url` is a plain `str` field (`src/gateway/core/config.py:395`) with no URL validation, so anything reaches this line. `make_url("")` and `make_url("not a url")` raise `ArgumentError`; `make_url("postgresql://u:p@h:bad/db")` raises `ValueError: invalid literal for int()`. Either aborts the command with an uncaught traceback ahead of the clearer failure `db_init` would have produced.

Use the helper the repo already has. `redact_url_secrets` in `src/gateway/services/url_safety.py` masks the userinfo password and every query value, and returns the input unchanged when it cannot parse. `src/gateway/api/routes/settings.py:353` already applies it to `database_url` for exactly this reason. Same input through it:

```
postgresql+psycopg://u:***@h:5432/db?password=***&sslpassword=***
```

```python
from gateway.services.url_safety import redact_url_secrets

safe_url = redact_url_secrets(gateway_config.database_url)
```

The import is clean: `cli.py` already imports from `gateway.services` (`secret_box`, line 230), and `gateway/` top level is unconstrained by `scripts/check_architecture.py`. Drop the `make_url` import on line 10 once both call sites move.

src/gateway/cli.py
```diff
@@ -192,7 +194,8 @@ def migrate(config: str | None, database_url: str | None, revision: str) -> None
         click.echo("alembic command not found in PATH", err=True)
         sys.exit(1)
 
-    click.echo(f"Running migrations on: {gateway_config.database_url}")
+    safe_url = make_url(gateway_config.database_url).render_as_string(hide_password=True)
```
khaledosman, line 197: Same change as line 169: `redact_url_secrets(gateway_config.database_url)`. This is the call site `alembic/env.py:120-123` names in its comment, so leaving it on `hide_password=True` keeps the leak that comment was written to route around.

khaledosman (commented):
`render_as_string(hide_password=True)` masks only the authority password, so the DSN shape this fix targets is the one it leaves in the clear. #922 says so in its body, and `alembic/env.py:120-123` carries a comment saying the same thing and naming `otari migrate` as the caller it was written for. The repo already has the helper that covers it. Details inline.

Also missing: `tests/unit/test_gateway_cli.py` exists and covers neither command, and the PR's test checkbox is unticked. A case asserting the password is absent from the output for a DSN carrying both `user:pass@` and `?password=` is what keeps this from regressing.

CI is green on all 11 checks.

_Reviewed by Claude Opus 5 (1M context) via [Claude Code](https://claude.com/claude-code) at @khaledosman's request._

PR #943, opened 2026-09-06
fix(settings): prevent database_url value from overflowing Settings card

### LINKED ISSUE #809: [BUG] A long database_url overflows its Settings row and squeezes the description column

AloysJehwin:
## Description
Replace `shrink-0` with `min-w-0` on the value column and add `flex-wrap` to the row so long values like `database_url` wrap instead of overflowing.

## PR Type
- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues
Fixes #809

## Checklist
- [x] I ran the Definition of Done checks locally.

## AI Usage
- [ ] No AI was used.

## Summary

Updated the Settings layout so long database URLs and paths stay within the card. The description column keeps usable space, and the page avoids horizontal overflow.

## Technical notes

- Added row wrapping and flexible value-column sizing.
- Applied `break-all` to long read-only values.
- Removed fixed-width behavior from the settings controls.

### DIFF
--- a/web/src/features/settings/SettingsPage.tsx
+++ b/web/src/features/settings/SettingsPage.tsx
@@ -191,15 +191,20 @@ function SettingControl({
 }) {
   if (!field.settable) {
     return (
-      <div className="flex items-center gap-2">
-        <span className="text-sm tabular-nums text-foreground">
+      <div className="flex min-w-0 items-center gap-2">
+        {/* `break-all` because the values that reach this branch are paths and
+            URLs, which hold no space to wrap at: without it the string's
+            min-content width is its full length, which is what pushed the row
+            past the card and squeezed the description column to a word a line
+            (otari#809). */}
+        <span className="min-w-0 break-all text-sm tabular-nums text-foreground">
           {formatValue(field)}
         </span>
         {/* A label, not a chip. A bordered capsule next to a value, on a
             surface that has no boxes on it, puts the shape to work where the
             type should be. `text-overline` is exactly that role and already
             carries the size, weight, tracking, uppercase and muted color. */}
-        <span className="text-overline">startup-only</span>
+        <span className="shrink-0 text-overline">startup-only</span>
       </div>
     )
   }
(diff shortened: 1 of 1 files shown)


njbrake (approved):
Approved. Rebased onto main: the original diff targeted markup that the dashboard redesign replaced, so both hunks were reauthored against the current file.

The overflow is a min-content problem, not a wrapping one: a path has no space to break at, so the value's min-content width was its whole length, and `shrink-0` on the control column held the row at that width. `break-all` drops it to one character, `min-w-0` lets the column shrink to it, and `flex-wrap` gives the control its own line rather than a squeezed share.

Measured on the built bundle with a long SQLite path: the card overflowed its own width by 214px at 1280 and 832px at 390, and by nothing at either afterwards. The settable row is pixel-identical at 1280 and stops clipping its key at 390.

PR #944, opened 2026-09-06
fix(budgets): guard DELETE /v1/budgets/{id} against nulling referenced users

### LINKED ISSUE #902: DELETE /v1/budgets/{id} deletes an organization-owned budget and silently uncaps its users
Noted in review on #898. Pre-existing, and a sibling of #881 rather than part of it.

## What is open

`DELETE /v1/budgets/{id}` (`src/gateway/api/routes/budgets.py`) resolves the budget by id with no ownership predicate, so a deployment operator can delete an organization-owned budget. The route refuses with a 409 while a workspace default or a spend ceiling names it, which are the two RESTRICT foreign keys, but neither check is about ownership and neither covers the tenant's other rows:

- `users.budget_id` is not checked at all. `Budget.users` is a plain relationship, so the ORM nulls the column out and any gateway user on that budget loses their cap with nobody told. This is the same shape #875 fixed on the organization-scoped delete, which counts those rows and refuses; the deployment route never got the matching guard.
- `budget_reset_logs.budget_id` is NOT NULL behind a plain relationship, so a budget that has ever reset fails the null-out at the commit and the caller gets a 500 "Database error" rather than a refusal naming anything.

From the tenant's side, a budget they defined disappears from their Spend page with no record of who removed it, and any ceiling that named it is gone with it.

The dashboard reaches it: since #898 the operator's Budgets page labels a tenant's row "Owned by an organization" and still offers Delete on it, individually and through the bulk bar.

## Suggested fix

Two separable decisions:

1. Whether the deployment surface may delete a tenant's budget at all. Refusing is the consistent answer with #881 and #875, but note that `PATCH /v1/budgets/{id}` deliberately still reaches those rows (it retimes their ceilings), so "the operator may edit but not delete" needs saying out loud if that is the choice.
2. The `users` and `budget_reset_logs` holds, which are a bug on this route regardless of ownership: a delete that silently uncaps a user, or answers 500 on a budget that has reset, is wrong on the deployment's own budgets too.

(2) is worth doing whichever way (1) goes.

AloysJehwin:
## Description

`DELETE /v1/budgets/{id}` did not guard `users.budget_id` or `budget_reset_logs.budget_id`: the ORM would null a gateway user's cap out silently, and a budget that has ever reset would fail the null-out at commit as an opaque 500. This mirrors the `users` / reset-log holds that otari#875 already guards on the organization-scoped delete, refusing with a clean 409 instead.

## PR Type
- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues
Fixes #902

## Checklist

## AI Usage
- [ ] No AI was used.

## Summary

- Prevents deletion of organization-owned budgets through the deployment endpoint.
- Removes reset history when a deployment-owned budget is deleted.
- Clears users’ budget assignments when deletion succeeds.
- Adds tests and updates API documentation for the new behavior.


src/gateway/api/routes/budgets.py
```diff
@@ -435,6 +459,13 @@ async def delete_budget(
...
         )
 
+    # ``budget_reset_logs.budget_id`` is NOT NULL behind a plain relationship, so
+    # deleting a budget that has ever reset would fail the null-out at the commit
+    # as an opaque 500. Clear its rows in the same transaction first. ``users``
+    # is left to the ORM: nulling a deleted own budget's caps is the intended
+    # behavior, since a budget that no longer exists caps no one.
+    await db.execute(delete(BudgetResetLog).where(BudgetResetLog.budget_id == budget_id))
+
     await db.delete(budget)
```
khaledosman, line 469: Letting the ORM null `users.budget_id` is now the stated design, but it does not do what the explicit detach does, and the difference leaves the row inconsistent.

`PATCH /v1/users/{id}` with `budget_id: null` clears `budget_started_at` and `next_budget_reset_at` along with the assignment (`src/gateway/api/routes/users.py:293-295`), and `test_an_explicit_null_budget_detaches_and_clears_the_reset_clock` is the contract for it. This path clears neither, so after the delete the user carries a future `next_budget_reset_at` with no budget attached. `reserve_budget` only rolls the window when a budget is attached (`src/gateway/services/budget_service.py:563-572`), so that clock never fires again: `GET /v1/users/{id}` reports a reset date for an uncapped user, and `users.spend` stays frozen at whatever was recorded until something reassigns a budget.

Clear them in the same transaction, before the delete:

```python
await db.execute(
    update(User)
    .where(User.budget_id == budget_id)
    .values(budget_started_at=None, next_budget_reset_at=None)
)
```

`test_deleting_an_own_budget_with_an_assigned_user_succeeds` asserts only `budget_id is None`, so it passes either way; it is worth extending to the two clock columns.

src/gateway/api/routes/budgets.py
```diff
@@ -396,6 +410,16 @@ async def delete_budget(
             detail=f"Budget with id '{budget_id}' not found",
         )
 
+    if budget.organization_id is not None:
```
khaledosman, line 413: The refusal names neither a route nor an actor, which the two refusals below it both do ("Organization > Workspaces > Edit", `/v1/scoped-budgets`), and which this file's own docstring gives as the reason for checking here at all: "so the refusal can say which, and where".

"manage it through the organization that owns it" is also not necessarily actionable by the reader: `DELETE /api/v1/organizations/{id}/budgets/{budget_id}` resolves a `CurrentIdentity` and requires that identity to manage that organization, which a deployment operator holding a master key need not be. Name the route, and say that an administrator of the owning organization is who can call it.

src/gateway/api/routes/budgets.py
```diff
@@ -379,13 +377,29 @@ async def delete_budget(
...
     db: Annotated[AsyncSession, Depends(get_db)],
 ) -> None:
-    """Delete a budget.
-
-    Refused with 409 while anything still names this budget: a workspace handing
-    it to its members, or a scoped ceiling enforcing it. Both foreign keys are
-    ``RESTRICT``, so the database would refuse either anyway, but as an
-    ``IntegrityError`` reported as "Database error" with nothing naming what to
-    go and change. Checked here so the refusal can say which, and where.
+    """Delete a budget the deployment owns.
```
khaledosman, line 380: This docstring is published API copy: it goes verbatim into `docs/public/openapi.json`, the Postman collection's request description, and `web/src/client/schema.ts`. "behind a plain relationship", the NOT NULL null-out and the "opaque 500" are notes about this file's ORM mechanics, and the last paragraph is repeated a second time as the comment at line 462-466.

Keep the wire behavior here (which two cases answer 409, what happens to assigned users and to reset history) and leave the mechanics to the inline comment alone.

Also `deployment-own budget`, twice, reads as a typo for `deployment-owned`, and it ships into all three artifacts.

src/gateway/api/routes/budgets.py
```diff
@@ -333,9 +333,7 @@ async def update_budget(
             if "budget_duration_sec" in request.model_fields_set
             else budget.budget_duration_sec
         )
-        alignment = (
-            request.reset_alignment if "reset_alignment" in request.model_fields_set else budget.reset_alignment
-        )
+        alignment = request.reset_alignment if "reset_alignment" in request.model_fields_set else budget.reset_alignment
```
khaledosman, line 336: Unrelated reformatting, here and at line 488 (and `tests/integration/test_budget_dashboard.py:87`). `make lint` runs `ruff check` only; there is no `ruff format` gate in the Makefile or in CI, so these are churn in a bugfix diff and they are part of what makes the test module conflict with `main`. Revert them.

tests/integration/test_budget_dashboard.py
```diff
@@ -176,6 +174,98 @@ def test_deleting_a_budget_a_workspace_hands_out_is_refused_by_name(
...
+    assert refused.status_code == 409, refused.text
+    assert "organization" in refused.json()["detail"].lower()
+
+    # Untouched, and an edit still goes through.
+    assert client.get(f"/v1/budgets/{budget_id}", headers=master_key_header).status_code == 200
+    edited = client.patch(f"/v1/budgets/{budget_id}", json={"max_budget": 50.0}, headers=master_key_header)
+    assert edited.status_code == 200, edited.text
+
+
+def test_deleting_an_own_budget_with_reset_history_succeeds_without_500(
```
khaledosman, line 208: This seeds `user_id=None`, which is the case where no user row is involved at all. A reset log is only written by `_cas_reset_user_budget` (`src/gateway/services/budget_service.py:89-95`), which always sets a `user_id` and runs on a user currently assigned to the budget, so the shape that actually occurs has both a reset log with a `user_id` and a `users` row pointing at the budget. That is the case where the explicit reset-log delete and the ORM's `users` null-out meet, and it is the one not covered. Worth seeding both in one test.

khaledosman (changes requested):
The final commit reverses what the title and the description say this PR does: `1e955517` implemented the `users` guard #902 asked for, and `eb458af` replaced it with an ownership refusal plus a reset-log wipe. Three things follow from that, none of which are anchorable to a line.

**Title and description.** The title ("guard `DELETE /v1/budgets/{id}` against nulling referenced users") and the body ("refusing with a clean 409 instead") describe the reversed commit. Squash-merge means the title is the line git-cliff puts in the changelog, so it would ship describing a guard that is not there. Retitle to what the change does, and rewrite the body around the two decisions actually made.

**`Fixes #902`.** #902 calls the `users` hold "a bug on this route regardless of ownership" and says it is "worth doing whichever way (1) goes". Declining it may well be right: `web/src/features/budgets/BudgetsPage.tsx:1022` already promises the operator that users on a deleted budget "lose this limit, so nothing caps them until another budget does". But that is a reversal of what the issue asked for, and it belongs in the PR body as an argument, not in a docstring aside. Either implement the guard or say `Refs #902` so the decision stays open.

**The dashboard still offers Delete on organization-owned rows.** `BudgetsPage.tsx:787` (row action) and `:964` (bulk bar) now offer an action the API refuses, which is the thing #898 fixed for assignment ("the UI stops offering a save the API refuses"). The page already withholds the assignment control on exactly these rows at `:1106-1108`. Worse in bulk: `onBulkDelete` (`:662-673`) awaits sequentially and breaks on the first rejection, so one organization-owned row in a selection aborts the loop and the remaining own budgets are silently not deleted. Withhold Delete on `isOrganizationOwned(row)` and keep those rows out of the bulk selection, or file it and name the issue here.

**Rebase.** The branch conflicts with `main` in `docs/public/openapi.json` and `tests/integration/test_budget_dashboard.py`. Two things to carry through: that test module now builds paths from `API_ROOT` (`from gateway.core.config import API_ROOT`, `f"{API_ROOT}/budgets"`) while the three new tests spell `/v1/budgets` literally, and the spec plus the Postman collection need regenerating after the rebase (`uv run python scripts/generate_openapi.py`, then `make postman`) rather than resolving by hand.

---

PR #946, opened 2026-09-07
feat(sdk): type Batch API responses in full client generation

### LINKED ISSUE #899: [FEATURE] Add typed response schemas for the Batch API
### Problem or use case

The five Batch API operations are declared `response_model=None` and return `dict[str, Any]`, so each advertises an empty `application/json` successful response. API documentation and generated clients cannot show any response fields.

Affected endpoints:
- `POST /v1/batches`
- `GET /v1/batches`
- `GET /v1/batches/{batch_id}`
- `POST /v1/batches/{batch_id}/cancel`
- `GET /v1/batches/{batch_id}/results`

### Proposed solution

Three response shapes back these five operations:

- Create, retrieve, and cancel return a serialized `any_llm.types.batch.Batch` extended with an otari-added `provider` string (`src/gateway/api/routes/batches.py`, lines 509-511, 536-538, 578-580).
- List returns `{"data": [<that same shape>]}` (line 646).
- Results returns `{"results": [{"custom_id": string, "result": object | null, "error": {"code", "message"} | null}]}` (lines 851-861).

Work:

- Generate a component schema from `any_llm.types.batch.Batch` in serialization mode, so the schema matches the bytes actually sent rather than the looser validation shape.
- Extend it with the `provider` field the gateway adds, and add the two wrapper schemas for list and results.
- Reference the appropriate component from each operation's successful `application/json` response.
- Add a test confirming all five response schemas are non-empty and that every reference resolves.

This belongs alongside the existing `set_json_200` calls that already type chat, messages, rerank, embeddings, and images, so the generated SDK cores benefit from the same change.

### Alternatives considered

Hand-write the batch fields in the route signatures. That duplicates the upstream any-llm model and drifts whenever any-llm changes, which is exactly why the existing enrichment derives the other inference responses from any-llm rather than restating them.

Ovvomii:
## Description

Generate typed responses for all five Batch API operations in the full SDK. Successful result items contain a `ChatCompletion`; failed items retain their error and null result. Batch lifecycle responses include Otari's provider field.

The bindings use the current API prefix. Tests cover response schemas, invalid payloads, and an actual batch result response with reasoning and provider-specific fields. Gateway runtime responses are unchanged.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [x] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes #899

## How to test it locally

With PostgreSQL available through `TEST_DATABASE_URL` or Docker:

```sh
uv run pytest tests/unit/test_sdk_codegen.py tests/unit/test_sdk_endpoint_coverage.py tests/integration/test_batches_endpoint.py -q
```

With a JRE and OpenAPI Generator installed:

```sh
uv run python scripts/sdk_codegen/generate.py --mode full --language all --out-dir dist
```

The generated batch result model should reference `ChatCompletion` in each language.

Validation:

- `make test` passed: 6,717 passed, 24 skipped (Python 3.13, four workers, isolated PostgreSQL, `NO_PROXY=*`).
- OpenAPI and Postman drift checks passed.
- All four full SDKs generated. The generated Python models passed round-trip checks for reasoning, provider extras, successful and failed results, and rejected invalid completions.
- Go and Rust formatting was skipped locally because `gofmt` and `rustfmt` are unavailable.

## Checklist

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:** Codex (GPT-6).

**Any additional AI details you'd like to share:** Implementation and local validation by Codex.

## Summary

Added typed successful responses for all five Batch API operations.

The schemas now derive Batch fields from `any-llm` serialization and include the gateway’s `provider` field. List and results endpoints now use dedicated response wrappers. Tests verify schema references, serialized payloads, endpoint bindings, and invalid field types.

This improves API documentation and generated SDK support without changing runtime response data.

## Validation

Lint, type checks, unit tests, integration tests, code-generation tests, OpenAPI checks, and SDK generation passed. Go and Rust formatting were skipped because the required formatters were unavailable.

### DIFF
--- a/scripts/sdk_codegen/generate.py
+++ b/scripts/sdk_codegen/generate.py
@@ -249,6 +249,7 @@ def enrich_spec(spec: dict[str, Any]) -> dict[str, Any]:
     Streaming still cannot be generated (OpenAPI Generator emits no SSE); the SDK
     shell hand-writes the stream iterator over the generated client.
     """
+    from any_llm.types.batch import Batch
     from any_llm.types.completion import (
         ChatCompletion,
         ChatCompletionChunk,
@@ -320,8 +321,49 @@ def enrich_spec(spec: dict[str, Any]) -> dict[str, Any]:
         "IMG",
     )
 
-    def set_json_200(path: str, schema_name: str, description: str) -> None:
-        op = spec["paths"][path]["post"]
+    schemas["BatchResponse"] = absorb(
+        Batch.model_json_schema(mode="serialization", ref_template="#/components/schemas/BATCH_{model}"),
+        "BATCH",
+    )
+    schemas["BatchResponse"]["properties"]["provider"] = {
+        "type": "string",
+        "description": "Provider instance that owns the batch.",
+    }
+    schemas["BatchResponse"].setdefault("required", []).append("provider")
+    schemas["BatchListResponse"] = {
+        "type": "object",
+        "required": ["data"],
+        "properties": {"data": {"type": "array", "items": {"$ref": "#/components/schemas/BatchResponse"}}},
+    }
+    schemas["BatchResultItem"] = {
+        "type": "object",
+        "required": ["custom_id", "result", "error"],
+        "properties": {
+            "custom_id": {"type": "string", "description": "Identifier supplied for this request in the batch."},
+            "result": {
+                "description": "Serialized provider result, or null when the request failed.",
+                "anyOf": [{"type": "object", "additionalProperties": True}, {"type": "null"}],
+            },
+            "error": {
+                "anyOf": [
+                    {
+                        "type": "object",
+                        "required": ["code", "message"],
+                        "properties": {"code": {"type": "string"}, "message": {"type": "string"}},
+                    },
+                    {"type": "null"},
+                ]
+            },
+        },
+    }
+    schemas["BatchResultsResponse"] = {
+        "type": "object",
+        "required": ["results"],
+        "properties": {"results": {"type": "array", "items": {"$ref": "#/components/schemas/BatchResultItem"}}},
+    }
+
+    def set_json_200(path: str, schema_name: str, description: str, method: str = "post") -> None:
+        op = spec["paths"][path][method]
         op["responses"]["200"] = {
             "description": description,
             "content": {"application/json": {"schema": {"$ref": f"#/components/schemas/{schema_name}"}}},
@@ -332,6 +374,11 @@ def enrich_spec(spec: dict[str, Any]) -> dict[str, Any]:
     set_json_200("/v1/rerank", "RerankResponse", "Rerank result")
     set_json_200("/v1/embeddings", "CreateEmbeddingResponse", "Embeddings")
     set_json_200("/v1/images/generations", "ImagesResponse", "Generated images")
+    set_json_200("/v1/batches", "BatchResponse", "Created batch")
+    set_json_200("/v1/batches", "BatchListResponse", "Batches", method="get")
+    set_json_200("/v1/batches/{batch_id}", "BatchResponse", "Batch", method="get")
+    set_json_200("/v1/batches/{batch_id}/cancel", "BatchResponse", "Canceled batch")
+    set_json_200("/v1/batches/{batch_id}/results", "BatchResultsResponse", "Batch results", method="get")
 
     schemas["ChatCompletionRequest"]["properties"]["messages"] = {
         "type": "array",
--- a/tests/unit/test_sdk_codegen.py
+++ b/tests/unit/test_sdk_codegen.py
@@ -1,6 +1,7 @@
 """Unit tests for the control-plane SDK codegen spec filter."""
 
 import importlib.util
+import json
 import shutil
 import sys
 from pathlib import Path
@@ -210,6 +211,10 @@ def _full_spec_stub() -> dict[str, Any]:
             "/v1/rerank": {"post": {"responses": {}}},
             "/v1/embeddings": {"post": {"responses": {}}},
             "/v1/images/generations": {"post": {"responses": {}}},
+            "/v1/batches": {"post": {"responses": {}}, "get": {"responses": {}}},
+            "/v1/batches/{batch_id}": {"get": {"responses": {}}},
+            "/v1/batches/{batch_id}/cancel": {"post": {"responses": {}}},
+            "/v1/batches/{batch_id}/results": {"get": {"responses": {}}},
         },
         "components": {"schemas": {"ChatCompletionRequest": {"type": "object", "properties": {}}}},
     }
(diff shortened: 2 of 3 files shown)


Ovvomii:
@coderabbitai review

scripts/sdk_codegen/generate.py
```diff
@@ -332,6 +374,11 @@ def set_json_200(path: str, schema_name: str, description: str) -> None:
     set_json_200("/v1/rerank", "RerankResponse", "Rerank result")
     set_json_200("/v1/embeddings", "CreateEmbeddingResponse", "Embeddings")
     set_json_200("/v1/images/generations", "ImagesResponse", "Generated images")
+    set_json_200("/v1/batches", "BatchResponse", "Created batch")
+    set_json_200("/v1/batches", "BatchListResponse", "Batches", method="get")
+    set_json_200("/v1/batches/{batch_id}", "BatchResponse", "Batch", method="get")
+    set_json_200("/v1/batches/{batch_id}/cancel", "BatchResponse", "Canceled batch")
+    set_json_200("/v1/batches/{batch_id}/results", "BatchResultsResponse", "Batch results", method="get")
```
khaledosman, line 381: Blocking. These five paths are hard-coded `/v1/...`, but `main` has since moved `enrich_spec` onto `f"{API_ROOT}/..."` (`API_ROOT = "/api/v1"`, imported from `gateway.core.config` at line 40), and the spec's paths are now `/api/v1/batches`. This PR is already `CONFLICTING`; once it is rebased, `spec["paths"]["/v1/batches"]` raises `KeyError` and full-mode codegen dies before it generates anything.

Use `f"{API_ROOT}/batches"`, `f"{API_ROOT}/batches/{{batch_id}}"`, `f"{API_ROOT}/batches/{{batch_id}}/cancel"` and `f"{API_ROOT}/batches/{{batch_id}}/results"`, matching lines 372-376 after the rebase.

scripts/sdk_codegen/generate.py
```diff
@@ -320,8 +321,49 @@ def absorb(js: dict[str, Any], prefix: str) -> dict[str, Any]:
...
+    }
+    schemas["BatchResultItem"] = {
+        "type": "object",
+        "required": ["custom_id", "result", "error"],
+        "properties": {
+            "custom_id": {"type": "string", "description": "Identifier supplied for this request in the batch."},
+            "result": {
+                "description": "Serialized provider result, or null when the request failed.",
+                "anyOf": [{"type": "object", "additionalProperties": True}, {"type": "null"}],
+            },
```
khaledosman, line 346: `result` is typed as a bare free-form object, but upstream it is `ChatCompletion | None` (`any_llm/types/batch.py`, `BatchResultItem`), and `enrich_spec` injects `schemas["ChatCompletion"]` twenty lines above. Full mode exists so the generated core returns typed models, and this leaves the one payload a caller actually reads as an untyped dict in all four SDKs.

```suggestion
            "result": {
                "description": "Serialized provider result, or null when the request failed.",
                "anyOf": [{"$ref": "#/components/schemas/ChatCompletion"}, {"type": "null"}],
            },
```

Safe on the validation side: that schema comes out with `additionalProperties: true`, so provider extras still pass. One thing to confirm first, though: the route dumps in python mode (`item.result.model_dump()`, `src/gateway/api/routes/batches.py:857`) against a serialization-mode schema. `jsonable_encoder` and `Reasoning`'s `model_serializer` should cover the gap, but it is worth a round-trip check, and if it does not hold that is a route bug rather than a reason to keep `result` untyped.

tests/unit/test_sdk_codegen.py
```diff
@@ -210,6 +211,10 @@ def _full_spec_stub() -> dict[str, Any]:
             "/v1/rerank": {"post": {"responses": {}}},
             "/v1/embeddings": {"post": {"responses": {}}},
             "/v1/images/generations": {"post": {"responses": {}}},
+            "/v1/batches": {"post": {"responses": {}}, "get": {"responses": {}}},
+            "/v1/batches/{batch_id}": {"get": {"responses": {}}},
+            "/v1/batches/{batch_id}/cancel": {"post": {"responses": {}}},
+            "/v1/batches/{batch_id}/results": {"get": {"responses": {}}},
```
khaledosman, line 217: Same rebase break as `generate.py:377-381`: these stub keys need `API_ROOT` interpolation, the way lines 210-214 already do on `main`.

tests/unit/test_sdk_codegen.py
```diff
@@ -249,6 +254,76 @@ def ref(path: str) -> str:
...
 
 
+@pytest.mark.parametrize(
+    ("path", "method"),
+    [
+        ("/v1/batches", "post"),
+        ("/v1/batches", "get"),
+        ("/v1/batches/{batch_id}", "get"),
+        ("/v1/batches/{batch_id}/cancel", "post"),
+        ("/v1/batches/{batch_id}/results", "get"),
```
khaledosman, line 264: Needs `API_ROOT` interpolation too, and this one fails sooner than the rest: `test_enrich_types_batch_operations` reads the real `generate.DEFAULT_SPEC` (line 268) rather than the stub, so a literal `/v1/batches` raises `KeyError` against `main`'s committed spec as soon as this branch is rebased, independent of whether codegen ever runs.

khaledosman (changes requested):
Two findings that cannot be anchored inline, because they sit on lines this PR does not touch:

**`scripts/sdk_codegen/generate.py:8` and `:51`.** Both still say batch responses are "untyped in the spec" and hand-written in every SDK. The README is updated in this PR; these two are not. Line 51's exclusion of batches from `CONTROL_PLANE_TAGS` is still correct, but the reason it gives no longer is.

**PR title.** `fix(sdk):` puts this under Bug Fixes in the generated changelog, and squash-merge means the title is what git-cliff parses. #899 is `[FEATURE] Add typed response schemas for the Batch API`, and nothing here changes runtime behavior, so `feat(sdk):` is the honest classification.

Gates checked and clean: no `docs/public/openapi.json` or Postman drift (`enrich_spec` enriches in memory only, it does not write the committed spec), `sdk-endpoints.txt:33-37` already classifies all five operations, the four new schema names collide with nothing in the spec, `jsonschema` is already a dev dependency, and the `status` enum is safe because every any-llm provider normalizes to the OpenAI literal set before constructing `Batch`. `tests/unit/test_sdk_codegen.py` passes as-is at this head (41 passed); the blocking finding below is entirely about what happens on the rebase this PR already needs.

---

scripts/sdk_codegen/generate.py
Ovvomii: Rebased onto current main in 323bf361. All five batch bindings now use API_ROOT; full generation completed for Python, TypeScript, Go, and Rust.

scripts/sdk_codegen/generate.py
Ovvomii: 5bea654c references ChatCompletion for non-null results. Tests cover reasoning strings, provider extras, and invalid completions. The actual batch route and generated Python models both passed round-trip checks.

tests/unit/test_sdk_codegen.py
Ovvomii: Updated the batch stub paths to use API_ROOT in 323bf361.

tests/unit/test_sdk_codegen.py
Ovvomii: Updated the real-spec batch cases to use API_ROOT in 323bf361. All five endpoint cases pass against the current committed spec.

Ovvomii:
Local validation passed: 6,717 tests passed, 24 skipped. All four SDK generation jobs are green, and CodeRabbit finished reviewing 5bea654c with no actionable comments.

The remaining CI failure is in [test-integration (4/4)](https://github.com/mozilla-ai/otari/actions/runs/35299978573/job/105460373271): `test_concurrent_claims_of_one_address_leave_exactly_one_holder` expects three integrity-error mappings but observed two. The assertions for one successful claim and three `EmailAlreadyInUseError` results passed before that assertion. The test and `user_service.py` are unchanged from main; this looks timing-sensitive, with one loser rejected by the preflight check instead of the unique index.

Could a maintainer rerun the failed shard? My account cannot rerun Actions on this repository.

PR #948, opened 2026-09-07
fix(mcp): stop managed transport redirects

HareeshBahuleyan:
## Description

Restricts managed MCP redirects to the validated origin or a same-host HTTP-to-HTTPS upgrade, preventing requests to unvalidated hosts while preserving common endpoint canonicalization. Adds regression coverage and documents the constraint.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

None.

## Checklist

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:** Claude Opus 5 and OpenAI Codex GPT-5.6 Sol

**Any additional AI details you'd like to share:** Implemented and verified under user direction.

**NOTE:**

### DIFF
--- a/src/gateway/services/mcp_client.py
+++ b/src/gateway/services/mcp_client.py
@@ -18,8 +18,10 @@ from contextlib import AsyncExitStack
 from dataclasses import dataclass, field
 from typing import TYPE_CHECKING, Any
 
+import httpx
 from mcp import ClientSession
 from mcp.client.streamable_http import streamablehttp_client
+from mcp.shared._httpx_utils import create_mcp_http_client
 
 from gateway.log_config import logger
 from gateway.services.tool_usage import ToolUsageTally
@@ -28,6 +30,35 @@ if TYPE_CHECKING:
     from mcp.types import Tool as MCPTool
 
     from gateway.models.mcp import McpServerConfig
+
+
+def _no_redirect_http_client(
+    headers: dict[str, str] | None = None,
+    timeout: httpx.Timeout | None = None,
+    auth: httpx.Auth | None = None,
+) -> httpx.AsyncClient:
+    """The MCP SDK's own HTTP client, with redirects turned off.
+
+    ``services/url_safety.validate_mcp_url`` vets a server's configured URL
+    before anything connects, refusing private, loopback, link-local and
+    reserved addresses. The SDK's default client sets ``follow_redirects=True``,
+    so a server answering with a redirect had that check bypassed entirely: the
+    redirected request that actually left the process was never vetted. For a
+    status such as 307, httpx also re-sends the method and request body to the
+    redirect destination.
+
+    Delegates to the SDK for everything else, deliberately. Its timeout defaults
+    are what the managed tool loop has always run with, and narrowing them here
+    would turn this into a behavior change for anyone whose MCP tool is slower
+    than a bound picked for a different purpose.
+
+    Following redirects safely means validating each destination before sending
+    anything to it. That is a larger change than this, and out of scope: no
+    caller needs a redirecting MCP server today.
+    """
+    client = create_mcp_http_client(headers, timeout, auth)
+    client.follow_redirects = False
+    return client
 
 
 def mcp_tool_to_openai(tool: MCPTool) -> dict[str, Any]:
@@ -91,7 +122,9 @@ class MCPClientPool:
         if cfg.authorization_token:
             headers = {"Authorization": f"Bearer {cfg.authorization_token}"}
 
-        transport = await self._stack.enter_async_context(streamablehttp_client(cfg.url, headers=headers))
+        transport = await self._stack.enter_async_context(
+            streamablehttp_client(cfg.url, headers=headers, httpx_client_factory=_no_redirect_http_client)
+        )
         read, write, _ = transport
         session = await self._stack.enter_async_context(ClientSession(read, write))
         await session.initialize()
--- a/src/gateway/services/mcp_client.py
+++ b/src/gateway/services/mcp_client.py
@@ -21,7 +21,6 @@ from typing import TYPE_CHECKING, Any
 import httpx
 from mcp import ClientSession
 from mcp.client.streamable_http import streamablehttp_client
-from mcp.shared._httpx_utils import create_mcp_http_client
 
 from gateway.log_config import logger
 from gateway.services.tool_usage import ToolUsageTally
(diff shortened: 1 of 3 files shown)


khaledosman (approved):

src/gateway/services/mcp_client.py
```diff
@@ -30,6 +32,35 @@
...
+    are what the managed tool loop has always run with, and narrowing them here
+    would turn this into a behavior change for anyone whose MCP tool is slower
+    than a bound picked for a different purpose.
+
+    Following redirects safely means validating each destination before sending
+    anything to it. That is a larger change than this, and out of scope: no
+    caller needs a redirecting MCP server today.
+    """
+    client = create_mcp_http_client(headers, timeout, auth)
+    client.follow_redirects = False
```
khaledosman: Turning redirects off outright will break MCP endpoints that work today. `create_mcp_http_client` sets `follow_redirects=True` as a deliberate SDK-wide default (its docstring lists it first) because streamable-HTTP endpoints redirect in practice: a Starlette/FastMCP mount 307s `/mcp` to `/mcp/`, an apex host 301s to `www`, an `http://` endpoint upgrades to `https://`. `docs/mcp.md:53` and `:103` both advertise `https://mcp.example.com/github` with no trailing slash.

After this, such a server fails at `raise_for_status()` inside the SDK's `_handle_post_request` during `session.initialize()`, and `MCPClientPool._connect` surfaces that as a generic pool-construction failure naming nothing the operator can act on.

Two asks:

1. Allow the hop that is provably safe: same scheme, host and port, or a plain `http` to `https` upgrade of the same host. That host is the one `validate_mcp_url` already vetted, so the hop needs no new check, and it covers the trailing-slash and TLS-upgrade cases that account for most real redirects here.
2. Add a line to `docs/mcp.md`'s "Limits and safety" list. Not following redirects is now a requirement on the operator's URL, and that list is where every other MCP URL constraint is documented.

The fully general fix is the pattern already in this repo at `web_search_backend._fetch_capped` (`src/gateway/services/web_search_backend.py:509-545`): walk hops manually, re-validate each against the SSRF guard, bound the walk. It is harder here because the SDK owns the request loop, which is why the same-origin allowance is the cheap correct middle rather than a full walk.

Also worth flagging that the docstring's "no caller needs a redirecting MCP server today" is an assertion about deployments this repo cannot see.

src/gateway/services/mcp_client.py
```diff
@@ -18,8 +18,10 @@
 from dataclasses import dataclass, field
 from typing import TYPE_CHECKING, Any
 
+import httpx
 from mcp import ClientSession
 from mcp.client.streamable_http import streamablehttp_client
+from mcp.shared._httpx_utils import create_mcp_http_client
```
khaledosman: This is the second module-scope `mcp` import in the gateway, and it is from an underscore module with no stability promise inside the `<2.0.0` ceiling.

`tests/integration/test_mcp_dependency_ceiling.py` exists because exactly that shape, a module-scope `mcp` import removed upstream, killed every gateway command in #689. It resolves `mcp` fresh while ignoring `uv.lock` and then proves the import still works, and its probe (`tests/integration/test_mcp_dependency_ceiling.py:80-90`) still names only `streamablehttp_client`.

Add `from mcp.shared._httpx_utils import create_mcp_http_client` to that subprocess probe. `tests/unit/test_mcp_client.py` imports it too, but only under the 1.28.1 that `uv.lock` pins, which is precisely the gap that integration test was written to close.

src/gateway/services/mcp_client.py
```diff
@@ -30,6 +32,35 @@
...
+
+    Delegates to the SDK for everything else, deliberately. Its timeout defaults
+    are what the managed tool loop has always run with, and narrowing them here
+    would turn this into a behavior change for anyone whose MCP tool is slower
+    than a bound picked for a different purpose.
+
+    Following redirects safely means validating each destination before sending
+    anything to it. That is a larger change than this, and out of scope: no
+    caller needs a redirecting MCP server today.
+    """
```
khaledosman: 19 lines of docstring on a 3-line function, and two of the three paragraphs defend what the change does *not* do. AGENTS.md: "Do not restate the code, narrate the change, or record what the code used to do: the commit message is where that belongs."

Keep the first paragraph's first two sentences, which say what a reader of this function needs. The timeout paragraph and the out-of-scope paragraph are both already in the PR body, and the same rationale appears a third time in the test docstrings at `tests/unit/test_mcp_client.py:155-160` and `179-184`. Three copies means two that go stale.

tests/unit/test_mcp_client.py
```diff
@@ -104,3 +107,118 @@ async def test_call_tool_sanitizes_transport_failure(
...
+    captured = _capture_transport(monkeypatch)
+    config = McpServerConfig(name="tools", url="https://93.184.216.34/mcp", authorization_token="ghp_token")
+
+    async with MCPClientPool([config]):
+        pass
+
+    client = captured["httpx_client_factory"]({"Authorization": "Bearer ghp_token"}, None, None)
+    client._transport = httpx.MockTransport(handler)  # noqa: SLF001
+    async with client:
+        response = await client.post("https://93.184.216.34/mcp", json={"method": "tools/list"})
```
khaledosman: This asserts httpx's behavior rather than the pool's: the request is hand-rolled and never goes through the MCP transport, so the test name over-promises.

Its real value is narrower and worth naming instead, because it is the one non-obvious thing about this implementation: mutating `follow_redirects` after `create_mcp_http_client` has returned is honored at send time, since `AsyncClient.send` resolves it from `self.follow_redirects` on every call. Say that, and the test earns its place; as written a reader cannot tell why it is not redundant with `test_connect_opens_the_transport_with_redirects_disabled`.

khaledosman (approved):
The mechanism is right and I verified it against the installed SDK: mcp 1.28.1's streamable transport never passes `follow_redirects` per-request, and httpx 0.28.1 reads `self.follow_redirects` in `AsyncClient.send`, so mutating it after `create_mcp_http_client` returns takes effect on every POST, GET-SSE and DELETE the transport makes. `mypy --strict` clean, the 8 unit tests pass at this head.

One calibration note, not a finding: httpx already strips `Authorization` on cross-origin redirects (`Client._redirect_headers`), and `validate_mcp_url` rejects `http` plus a token, so a followed redirect never carried the bearer token. The SSRF exposure, a 307 replaying the JSON-RPC body to an unvetted address, is real.

Four inline items. The redirect-interop one is worth resolving before this ships, or at least deciding deliberately.

<sub>Reviewed with [Claude Code](https://claude.com/claude-code) (Opus 5)</sub>

src/gateway/services/mcp_client.py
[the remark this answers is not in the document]
HareeshBahuleyan: Addressed in 49dfafc1: removed the private SDK import and construct the client directly from the transport factory arguments.

src/gateway/services/mcp_client.py
HareeshBahuleyan: Addressed in c80c8b40: same-origin and default-port HTTP-to-HTTPS redirects remain allowed; host changes, port changes, and downgrades are blocked before sending. Added docs and regression coverage.

src/gateway/services/mcp_client.py
HareeshBahuleyan: Addressed in 49dfafc1: removed the private helper import entirely, so the fresh-resolution probe does not need another private import target.

src/gateway/services/mcp_client.py
HareeshBahuleyan: Addressed in c80c8b40: reduced the client factory docstring to one sentence and moved the operator-facing constraint to docs/mcp.md.

tests/unit/test_mcp_client.py
HareeshBahuleyan: Addressed in c80c8b40: replaced the broad test with policy-level cases covering allowed same-origin and HTTPS-upgrade redirects plus blocked host, port, and downgrade redirects.

PR #950, opened 2026-09-08
fix(auth): default oauth providers to empty array so sign-in survives an older gateway

### LINKED ISSUE #806: [BUG] An older gateway's bootstrap white-screens the dashboard sign-in page

AloysJehwin:
## Description
`renderableOAuthProviders` now defaults its `configured` parameter to an empty array, so a gateway whose `/v1/bootstrap` omits `oauth_providers` no longer throws in `.filter()` and white-screens the sign-in page.

## PR Type
- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues
Fixes #806

## Checklist

## AI Usage
- [ ] No AI was used.

## Summary

- Default `renderableOAuthProviders` to an empty list when `oauth_providers` is missing.
- Prevent the sign-in page from failing during rendering with older gateways.

### DIFF
--- a/web/src/features/auth/oauthProviders.ts
+++ b/web/src/features/auth/oauthProviders.ts
@@ -55,7 +55,7 @@ export function isOAuthProvider(value: string): value is OAuthProvider {
 
 /** The providers from a bootstrap that this dashboard can render, in its order. */
 export function renderableOAuthProviders(
-  configured: readonly string[],
+  configured: readonly string[] = [],
 ): OAuthProvider[] {
   return configured.filter(isOAuthProvider)
 }


njbrake:
Closing in favor of #962, which fixes this at the boundary instead.

The one-line default is correct and does not regress anything, but it does not fix #806, and merging it would close that issue behind a bug that is still live. `Login.tsx` reads `sign_in_methods` at lines 226, 232 and 246, before it reaches `renderableOAuthProviders` at 251, so on a gateway old enough to omit `oauth_providers` sign-in dies 25 lines above this guard and never reaches it. `PublicAuthPage.tsx:51` and `AcceptInvitationPage.tsx:78` read `oauth_providers` again on their own; both still throw with this patch applied, verified.

#962 completes the payload once where it enters the app and adds the error boundary #806 asks for, which covers all four pre-router pages rather than one field on one of them.

Nothing wrong with the change itself, and thanks for the report of it. The layer is the only problem.

PR #951, opened 2026-09-08
fix(a11y): stop sidebar nav disclosures from emitting document headings

### LINKED ISSUE #808: [BUG] Sidebar disclosure groups are <h3> elements, so they sit in every page's document outline

AloysJehwin:
## Description
The sidebar `NavGroup` disclosure toggles rendered via `Disclosure.Heading` (an `<h3>`), putting "Routing" and "Tools" into every page's document outline; this swaps that wrapper for a plain `<div>` so the labels leave the outline while the `Disclosure.Trigger` button keeps the toggle working.

## PR Type
- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues
Fixes #808

## Checklist

## AI Usage
- [ ] No AI was used.

Replaced the sidebar group heading wrapper with a neutral container. `Routing` and `Tools` no longer appear in page heading outlines, while their expand and collapse controls continue to work.

### DIFF
--- a/web/src/app/AppShell.tsx
+++ b/web/src/app/AppShell.tsx
@@ -325,7 +325,13 @@ function NavGroup({
       onExpandedChange={setOpen}
       className={clsx("flex flex-col", open && "gap-0.5")}
     >
-      <Disclosure.Heading>
+      {/* A plain wrapper rather than `Disclosure.Heading`: HeroUI renders that
+          as an <h3>, which put every group's label ("Routing", "Tools") into the
+          document outline of every page, so a screen-reader user navigating by
+          heading landed on nav chrome between a page's own sections. The toggle
+          is the `Disclosure.Trigger` button below, which carries its own
+          `aria-expanded` and needs no heading role to work. */}
+      <div>
         {/* `ancestor`, not `isActive`: the selected row is the child below,
               which is visible whenever this trigger is expanded. The collapsed
               rail's trigger a few lines up keeps the full selected marker,
@@ -340,7 +346,7 @@ function NavGroup({
             className={navIndicatorClass({ open })}
           />
         </Disclosure.Trigger>
-      </Disclosure.Heading>
+      </div>
       {/* The rows sit straight in the panel rather than in a `Disclosure.Body`,
           which wraps its children in a div carrying 0.5rem of padding that no
           className can reach: it would inset these rows from the lane their


njbrake (approved):
Approved.

`Disclosure.Heading` renders a bare react-aria `Heading` carrying an unstyled `disclosure__heading` hook class that no rule in HeroUI or `web/src` matches, so the swap is structural only, with no visual change. `Disclosure.Trigger` takes `aria-expanded` and its panel wiring from the disclosure context rather than from the wrapper, so the toggle is unaffected; the sidebar nav now emits no heading elements at all.

#808 asked for a decision rather than a styling fix. Taking nav chrome out of the outline is the right one, and the comment records why.

Existing AppShell coverage keys on the `Routing` and `Tools` buttons and their expanded state, so the toggle stays guarded. Nothing guards the heading itself: a test asserting the sidebar contributes no headings would stop the wrapper coming back.

PR #952, opened 2026-09-08
refactor(dashboard): one module per component and per data domain

khaledosman:
## Description

`shared/components/surface.tsx` (876 lines, 22 exports), `shared/components/ui.tsx`
(1,075 lines, 20) and `shared/api/hooks.ts` (3,638 lines, 172) held 214 exports
between them. None of the three names said what was inside, and `surface` was worse
than uninformative: `bg-surface` is a token naming one of four neutral levels, so the
word carried two unrelated meanings in one tree.

The tree had already half-decided the split. `ui.tsx` and a `ui/` directory coexisted,
`surface.tsx` had no `surface.test.tsx` but eleven per-component test files, and
`design/buttons.md` documented one action family (`RowAction`, `RowActionRow`,
`RowActions`, `ConfirmButton`, `ConfirmRowAction`, `RefreshButton`, `CopyButton`) that
lived in three modules.

### A directory is named for the design topic that documents it

So `web/design/`'s topic list is the directory list, and the doc-to-code mapping is
one-to-one rather than a second naming of something the design system already names:

```
shared/components/
├── layout/     ← design/layout.md        ├── navigation/  ← design/navigation.md
├── metrics/    ← design/metrics.md       ├── indicators/  coined: the status marks
├── feedback/   ← design/feedback.md      ├── access/      coined: deployment states
├── forms/      ← design/forms.md         ├── deprecated/  the review-blocked set
├── actions/    ← design/actions.md       └── ProductMark.tsx
└── data/       ← design/data.md
```

`buttons.md` becomes `actions.md`: `CopyField` and `RowAction` are not buttons, and
that file's own closing section was already titled "Other action shapes".
`indicators/` and `access/` are the only two names this invents, and they are called
out as such in the record below.

`shared/api/` drops the meaningless `hooks` segment (under `api/` every export is a
hook) and is one module per domain: `usage.ts`, `apiKeys.ts`, `organizations.ts`,
`budgets.ts`, and so on. **All 43 query keys stay in `queryKeys.ts`**, which is the
one rule this split must not break: 109 invalidations reach 39 of them across
domains, so a key declared next to its own hooks would either be duplicated or
imported back across domains, which is the circular import the no-barrel rule exists
to prevent. The resulting graph is a DAG (`client` and `queryKeys` are leaves).

### How it was done

Mechanically, because 200 files of import rewriting is not hand work:

- Blocks were lifted **by line range**, so every comment moved byte for byte rather
  than being retyped. `surface.tsx` reports 855 of 855 lines covered after its header.
- Each new file's import header was **computed** from the identifiers the block
  actually references, which is what kept 34 headers free of both an unused import
  and a missing one.
- Call sites were rewritten by a codemod, not `sed`: one grouped statement becomes
  several pointing at several directories, and the specifier list spans lines. It
  refuses a statement it cannot fully map rather than dropping a name, which is how
  the one import carrying a comment inside its braces was caught instead of silently
  losing `StatCard`.

### `deprecated/` makes the review block mechanical

DESIGN.md kept a prose table of components that must not be used in new code, with
"adding a new one is a review block" enforced by a human noticing.
`deprecated/deprecated.test.ts` now fails on a call site that is not already in its
list, naming what to use instead. A gate test rather than a Biome rule because every
existing call site violates such a rule, so it would need a `biome-ignore` in each of
five feature files and the ignores would become the thing to maintain.

It also **derives** the counts, which is how two stale ones were found: `StatCard` was
listed on Usage after Usage stopped using it, and `RowActions` on two call sites when
it had one. And it records **`SettingsSection` at zero call sites** — dead code rather
than a deprecation, shadowing `layout/SettingsGroup` (7 call sites) while diverging
from this tree's `export function` convention. Deleting a component is a maintainer's
call, not a side effect of moving files, so it is pinned at zero by a test instead.

### One live gate was silently repointed

`dotRamp.test.ts` carried a hardcoded `file.endsWith("surface.tsx")` exception, which
is `RowAction`'s own definition and the one file its rule must skip. It now names
`actions/RowAction.tsx`. It also gains the "covers the source tree" guard its
neighbors have, since both its rules assert only that an offender list is empty, so a
path resolved to the wrong directory would scan nothing and pass. Same for
`foundation.test.ts`'s allowlist entry.

### Two tests stopped sleeping through real timers

Together 10.6s of a 24s suite:

| Test | Before | After | Why |
| --- | --- | --- | --- |
| `CopyButton` "clears the confirmation on its own" | 1,515ms | **44ms** | waited out the component's 1.5s dismissal |
| `ActivityPage` "drops the live control when the in-flight poll starts failing" | 9,118ms | **85ms** | `useInFlightRequests` declares its own `retry`, which **overrides the harness's `retry: false`**, so reaching the error arm meant a 2s poll plus TanStack's 1s/2s/4s backoffs |

Both use the fake-timer pattern this tree already had in three places, and
`advanceTimers` is the answer to the deadlock the old `CopyButton` comment cited as
its reason for using real timers. **Both were checked against a mutated component**
(reset removed; `inFlight.isError` arm removed) to prove they still fail when the
behavior goes, because a fake-timer test that passes vacuously is worse than a slow
one. The `{ timeout: 20000 }` and `}, 30_000)` ceilings that existed only to survive
the sleeping are gone.

No timeouts were raised. `web/vite.config.ts` and `src/tests/setup.ts` are untouched.

### Reading the diff

It is 200 files but readable by **shape**: 40 renames, and the rest are moves plus
import rewrites with no logic change. `web/design/module-map.md` is the record, and
it is deliberately absent from DESIGN.md's topics table because it describes work
rather than a rule; delete it once its follow-ups are closed.

**Follow-ups it deliberately leaves:** delete `deprecated/SettingsSection`; convert
the remaining deprecated call sites (`PageHeader` on 4 pages, `StatCard` on 4 uses in
`OverviewPage`, `RowActions` on 1 in `PasskeysCard`), one page at a time; move the
five poll cadences out of `queryKeys.ts` (they are query options, not keys).

## PR Type

- [ ] New Feature
- [ ] Bug Fix
- [x] Refactor
- [x] Documentation
- [ ] Infrastructure / CI

## Relevant issues

None.

## Checklist

Notes on the two boxes that need qualifying, since this change is dashboard-only and
nothing under `src/gateway/` or `tests/` is touched:

- **Tests:** the new gate is `web/src/shared/components/deprecated/deprecated.test.ts`,
  and the two timer conversions above are the updated ones. The Python suites are
  unchanged because no Python is.
- **Definition of Done:** `make lint` and `make typecheck` pass (Ruff, the
  architecture check, mypy over 567 files). `make test` was **not** run: it needs
  PostgreSQL and covers a backend this diff does not reach. The dashboard checks are
  what apply, and all four pass: `lint`, `typecheck`, **135 test files / 3,838 tests**,
  and `build`. Both generated artifacts are in sync
  (`git diff --exit-code` on `routeTree.gen.ts` and `client/schema.ts`).
- **OpenAPI:** unchecked because no route or schema changed.

The behavioral Playwright suite has **not** been run against this; it needs a built
gateway and is the one check still owed.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

Claude Code (Opus 5)

The splits and the call-site rewrites were done by scripts rather than by hand, for
the reasons in "How it was done" above. Three traps are recorded in
`web/design/module-map.md` because each cost a cycle: zsh does not word-split an
unquoted variable, so a mapping loop ran once and `${pair##*:}` resolved to the last
colon in the whole string, landing `DataTable` at `components/access/DataTable`; a
specifier rewrite has to be anchored on its closing quote, or `components/Field"`
also matches `components/FieldMessages"`; and a block boundary can be off by one
declaration and still compile, which is how `optionKey`/`optionValue` went with
`Badge` instead of `FilterSelect`.


web/src/shared/api/queryKeys.ts
```diff
@@ -0,0 +1,107 @@
+/**
+ * Every query key the dashboard uses, and the cadences and options that ride
+ * with them.
+ *
+ * One module, deliberately, and the one rule this split must not break. The
```
khaledosman, line 5: The split made 43 previously file-private identifiers public: all 40 keys, the three poll cadences, `NO_RETRY`, and `fetchAllPaged`.

This docstring argues the keys must live in one module because 109 invalidations reach 39 of them across domains. The monolith enforced that *structurally*; nothing does now. A feature can `import { USAGE } from "@/shared/api/queryKeys"` and hand-roll an invalidation, bypassing the hooks that own the graph.

No call site does it today, so this is a rule to add rather than a cleanup: a `noRestrictedImports` group in `web/biome.jsonc` confining `@/shared/api/queryKeys` and `@/shared/api/paging` to `src/shared/api/**`, spelled the way the three existing layer rules are (alias and relative form, each with and without a trailing segment).

One wrinkle that makes a naive rule wrong: `features/providers/ProvidersPage.test.tsx:17` already imports `PROVIDER_HEALTH_REFRESH_MS` from here, and legitimately, since that constant was public before the split. Either exempt tests or move that one constant to `providers.ts`, where its only consumer is.

web/design/module-map.md
```diff
@@ -0,0 +1,365 @@
+# Module map: giving `shared/` a shape
+
+**This is a migration record, not a design topic.** It is deliberately absent
+from DESIGN.md's topics table, because it describes work rather than a rule.
+Delete it once the follow-ups at the bottom are closed; the durable half (where a
+component lives) is already in DESIGN.md's "Where things come from" table.
+
+**Status: executed.** `surface.tsx` (848 lines, 22 exports), `ui.tsx` (1,075
```
khaledosman: Stale figures, left over from the pre-rebase run. `surface.tsx` was **876** lines on main, not 848, and the suite is **135 files / 3,838 tests**, not 132 / 3,573. Same numbers again at lines 11, 23, 330 and 333.

The commit message says 876, so this record contradicts the commit that carries it. Worth fixing before merge in a repo whose stated doctrine is that the stale copy is the one someone believes.

web/design/DESIGN.md
```diff
@@ -98,10 +118,30 @@ artboards (foundations, components with every state, page archetypes).
...
 | --- | --- | --- |
-| `PageHeader` (`shared/components/ui.tsx`) | `PageIntro` (`shared/components/surface.tsx`) | 4 pages |
-| `StatCard` (`shared/components/ui.tsx`) | `KpiStrip` + `KpiCell` | Overview, Usage |
+| `deprecated/PageHeader` | `layout/PageIntro` | 4 pages, one use each |
+| `deprecated/StatCard` | `metrics/KpiStrip` + `KpiCell` | Overview only, 4 uses |
+| `deprecated/RowActions` | `actions/RowActionRow` | 1 use, in `PasskeysCard` |
+| `deprecated/SettingsSection` | `layout/SettingsGroup` | **nothing. Dead code** |
 | HeroUI `Card` | `Section`, or a bare band | 6 components |
 
+The counts are the tree's, not this table's memory of it: the gate test derives
```
khaledosman: "the gate test derives them" overstates what `deprecated.test.ts` covers. Its `KNOWN` map is file to component names and never counts usages, so it gates the *importing-file* counts but not the per-file figures in the same column ("4 pages, one use each", "Overview only, 4 uses").

Those figures are therefore exactly as ungated as the two that already drifted, which is the drift this sentence claims to have closed. All four rows are correct right now (verified against the tree), so this is about the guarantee, not the data: either narrow the sentence, or have the gate assert usage counts as well as importers.

web/src/shared/components/navigation/tabs.test.tsx
```diff
@@ -1,7 +1,8 @@
 import { render, screen } from "@testing-library/react"
```
khaledosman, line 1: The only camelCase test file left, against `naming-conventions.md:11` (`<Name>.test.ts(x)`). The reason is real (one header docstring argues the ARIA departures of `Tab`, `TabRow` and `Segmented` together, so splitting would duplicate that rationale), but it is recorded only in `web/design/module-map.md`, whose own header says to delete it once its follow-ups close.

So the file outlives its explanation. Put the exception in `naming-conventions.md` beside the rule it bends, or split the docstring and the file.

web/src/shared/components/actions/RowAction.tsx
```diff
@@ -0,0 +1,69 @@
...
+ * A row's action: 13px of muted text, no border, no fill, no box.
+ *
+ * A row has three or four of these and boxing each one turned the last lane
+ * into a control panel, louder than the row it acts on.
+ *
+ * `isDanger` is for the ARMED state and nothing else, which is worth spelling
+ * out because it was got wrong at six sites on the first pass: a destructive
+ * action at rest is the same muted text as its neighbors, and the ink arrives
+ * only once the next click commits something. A row that paints Remove red
+ * before anybody has touched it spends the colour on a state nothing is in, and
```
khaledosman: "colour" here and on line 15, against AGENTS.md's US-English rule for doc comments. The text moved byte-for-byte from `surface.tsx` so the spelling is pre-existing, but this PR is what creates the file.

khaledosman (commented):
Self-review of the module-map split. Five findings, none blocking: one boundary that the split removed and nothing replaced, three docs that overstate or misstate what they describe, and one spelling nit on a new file.

Posted as a comment rather than an approval only because GitHub refuses an approval from the PR author.

Checked and clean, since these were the plausible failure modes: no export lost or duplicated (214 old exports, each present exactly once), `surface.tsx` fully accounted for (855 of 855 lines after its header), the dissolved module docstring's rationale preserved in `components.md`, both generated artifacts in sync, no import cycles in `shared/api/` (`client` and `queryKeys` are leaves), and `TableScrollFrame` keeping main's `[]` deps and MutationObserver rather than reverting to the no-deps version.

PR #956, opened 2026-09-08
fix(dashboard): correct two sidebar defects and inline the record copy control

jigjigjig:
## Description

Three changes to the dashboard, all reported or approved from use: two visual
defects in the sidebar, and the verify block on the organization email-domain
page rearranged so the record and the action it feeds sit on one row.

**The account monogram rendered at the wrong size.** The initials span at
`AccountMenu.tsx` carried `text-chrome-initials`, a class that is defined
nowhere. It was defined in #638, renamed to `text-shell-monogram` with identical
values in #892 (the old definition being deleted), and #934 moved the call site
back to the old name. Since #934 the span has had no font size of its own and
has inherited the nav row's, so a monogram sized for a 26px tile rendered at
14px instead of 9px. The fix is the current name. Verified on a running
instance: computed `font-size` is 9px, and removing the fix live in the DOM puts
it back to 14px, so the rename is load bearing rather than cosmetic.

**The monogram tile was not centered on the collapsed rail.** `navRowClass`
contributes `justify-center` when the rail is collapsed, and the call site
appended an unconditional `justify-start` on top of it. Both landed on the
element with equal specificity, so the later emitted one won: in the built
stylesheet `.justify-center` sits at byte 424928 and `.justify-start` at 425005.
The tile centered at x=14.0 in a 72px rail whose center, and whose every other
collapsed icon, sits at 35.5 to 36.5. That is 21.5px off, and it is what a
reader sees as the icon being off center.

Rather than make the appended utility conditional at the call site, this gives
`navRowClass` an optional `expandedJustify` parameter, because the missing
affordance is the cause: a row that needs a justification other than the
collapsed default had no way to say so through the helper, so it had to append
and then lose. The parameter is additive. It is unset by default, all seven
existing call sites and the helper's own tests are untouched, and only
`AccountMenu` passes it. Confirmed by importing `navRowClass` from the base
commit alongside the new one and comparing output across ten call site shapes:
byte identical on all ten.

After the fix the tile centers at x=35.5, matching the rail center and the other
collapsed icons. Expanded, the row is still start aligned, with the tile's left
edge at x=25 alongside the other rows' icons at 25 and 26.

**Two dividers at the foot of the rail on the organization pages.** The footer
block and the rail's closing band each draw a rule, and the only row between
them is the way onto the organization rail, which is gated out once you are on
it. With that row absent the two rules were separated by `pt-1` alone, so 4px of
gap and 6px from outer edge to outer edge, which reads as one doubled hairline.
Reproduced in the DOM before fixing: two painted rules, both
`rgba(255, 255, 255, 0.08)`, with tops at y=784 and y=789.

The band's rule is deliberately unconditional, since it mirrors the scope band
at the head of the rail so that both ends of the rail are the same shape, so the
footer's rule is the one that gives way. The condition is now a single named
constant shared with the row's own gate, so the two cannot drift apart: them
drifting apart is the defect, not this one instance of it.

After the fix, one rule on the organization rail. In the workspace context,
still a rule, then the row, then a rule.

**The verify block on the email-domain page put the record and its action on
separate rows.** The TXT record an operator has to publish took a full-width
field, the copy affordance sat above it in the label row, and the Verify button
sat in a row of its own below. The record is one line that is copied and then
acted on, so the copy control moves inside the field and the button moves beside
it.

`CopyField` takes one optional prop, `action`, and its presence selects the
arrangement. Absent, the field renders exactly what it always has. The prop is
excluded from the multiline variant at the type level rather than guarded at
runtime, through a discriminated union on the props type, so a call site cannot
pass both: the multiline field is a `<textarea>`, where the right padding an
in-field control needs indents every line instead of clearing the first.

In the inline arrangement the control is `CopyButton`, the same component the
models table uses. That is an improvement beyond the layout, and it is the
reason the swap is worth making rather than restyling the existing button:
`CopyField`'s own copy path only ever tried `navigator.clipboard`, which does
not exist on the plain-HTTP origins this dashboard is routinely served from, so
it could not actually copy there and instead selected the value and asked for
Ctrl/Cmd-C. `CopyButton` goes through `copyToClipboard`, which falls back to
`execCommand`, the only clipboard write such an origin has. So the common
non-secure case now copies.

The one case that would have got worse is when `execCommand` also fails, where
the old field at least left the value selected. `CopyButton` therefore takes an
optional `selectOnFailure` ref, defaulted off so `CopyableValue` and the models
table are untouched, and focuses and selects the field on a failed copy. It does
that **after** the attempt resolves, not before: `copyToClipboard`'s legacy path
deliberately restores the selection and focus it found on its way out
(`clipboard.ts:55-59`), so a selection made first is undone by the very fallback
it exists to serve. Selecting first would look correct and do nothing on exactly
the path that needs it.

Scope: the shared `CopyField` has seven call sites, and this changes one. The
master key field on Settings is a separate local component declared at
`SettingsPage.tsx:281`, not this one. The
other six, including the four multiline ones, render byte-identically, and that
is proved rather than eyeballed: `CopyField` was imported from the base commit
alongside the new one and both were rendered for every distinct prop shape those
six use, comparing markup with only `useId` values normalized away. Identical on
all of them, and the comparison was itself checked by confirming it fails on a
single added class.

### Numbers, measured on the running instance

| Quantity | Value |
| --- | --- |
| Record length | 74 characters, all three blocks |
| Fira Code advance | 7.20px at the field's 12px, so 0.600 em |
| Glyph run | 532.79px |
| Field, xl and up | 600px, height 36px, right padding 44px |
| Control, xl and up | 32 x 32px, 4px inset, vertically centered to 0.00px |
| Whole value visible | yes, `scrollWidth` 598 against `clientWidth` 598 |
| Row content, Verify domain | 735.65px |
| Row content, Re-verify domain | 757.13px |
| Field, below md | height 46px, right padding 56px |
| Control, below md | 44 x 44px, fits inside the field's borders |

The control renders 32 x 32, not the 36 x 32 its declared width suggests, which
is why the right padding is 44px and not 46px.

### What this does not do

It does not touch the `!` appends in the nav, `navBandRowClass`, or the two
findings below, all of which are real and each of which is going to its own
issue:

- `border-border-strong` matches no utility in the bundle. The variable itself is
  fine: `--color-border-strong` is emitted in both theme blocks and is consumed
  at `globals.css:2062`. Only the utility class is missing, and nothing in
  `web/src` writes that name, so it is a latent gap rather than a live defect.
- `.border-emerald-200` reaches the bundle because Tailwind scans it out of a
  comment in `foundation.test.ts`.
- The `!` appends elsewhere in the nav are a related shape but not this bug.
  `AppShell.tsx:278` is the same helper versus call site collision, resolved with
  a bang instead of a parameter. `navBandRowClass`'s own bang may simply be
  unnecessary. Whether the nav helpers should own width the way they now own
  justification is worth deciding on its own, not in passing here.

`text-shell-monogram` is now the only type role in the file with a call site
pointing at it, and `text-emphasis` remains without one by design.

The organization domains page has an existing screenshot entry and its baseline
changes. Those PNGs are gitignored and the suite runs on demand, so there is
nothing to commit, but the baseline should be retaken.

### Verification

Measured on a running dashboard, confirming the served asset hash before each
measurement, and reading numbers only after a full reload. Two things about that
are worth stating, because both nearly produced a wrong number: an automated tab
reports `visibilityState: "hidden"`, so the rail's `transition-[width]` never
advances and a width read mid transition is the frozen start value rather than
the collapsed width; and navigating to a `#/` route from the same origin does not
reload, so it can be measured against a stale bundle.

**One thing could not be verified here, and it is the press itself.** An
automated tab reports `document.hasFocus()` as false and `visibilityState` as
hidden, so react-aria never registers the press: clicking the in-field control
opened no tooltip and wrote to no live region. The tooltip's strings, its 1500ms
and 5000ms timings and its self-clearing are covered by the existing
`CopyButton` cases, and the failure string was checked against the two candidate
stylesheet rules for the tooltip box (`.tooltip`, capped at `--container-xs` with
12px text, and `.tooltip__content`, capped at `min(22rem, 100vw - 2rem)` with
13px text). It overruns its content box under both, 309px against 302px and
334.8px against 328px, so it wraps to two lines either way. Which rule the
rendered box uses could not be settled without opening it.

Checks on this exact tree: `pnpm --dir web run lint` clean over 387 files,
`pnpm --dir web run typecheck` (`tsc -b --noEmit`) clean,
`pnpm --dir web test` 3858 passed across 135 files. The baseline was measured on
a pristine checkout of the base commit rather than assumed: 3844 passed across
the same 135 files, so the delta is exactly the fourteen added cases and no test
moved from pass to fail. `make lint` clean (architecture, alembic heads, Ruff),
`make typecheck` clean over 567 files, `make test-unit` 2881 passed and 3
skipped.

Use `pnpm --dir web run typecheck` and not a bare `tsc --noEmit` to reproduce
this. The package script is `tsc -b --noEmit`, and the root `tsconfig.json`
carries `"files": []` with project references, so a bare `tsc --noEmit` resolves
that root config and checks almost nothing, including none of the test files. It
reports success on a tree the real typecheck rejects.

The narrow-width numbers are from a 606px viewport, which is the narrowest this
browser window would go, and every rule below `md` is confirmed there: 46px
field, 56px right padding, a 44 x 44 control inside the borders, the action
wrapping to its own line, and the value scrolling rather than truncating. The
390px case is computed from the measured 7.20px advance and the live 16px page
padding, giving a 358px field, a 288px content box and 40 of the 74 characters
visible. Computed, not measured, because the window would not reach 390px.

### About the tests

Fourteen cases are added and nine of them are checks. Each fix was removed and the
suite re-run to see which cases actually fail without it, because a case that
passes either way is a regression guard and not evidence:

- `rowStyles.test.ts`: "start-aligns an expanded row that asks for it" fails
  without the fix. "still centers that row when the rail is collapsed" and "adds
  no justification when it is left unset" pass either way.
- `AppShell.test.tsx`: both cases are checks, and neither covers the other.
  Suppressing the band's rule instead of the footer's fails "drops it on the
  organization rail, where that row is gone". Inverting
  `footerHasRowAboveBand` fails "keeps the footer rule where a row sits below
  it" at its own assertion. Neither break fails the other case, so together they
  cover both regressions and separately they cover one each.
- `actions/CopyField.test.tsx`: the three inline-arrangement cases fail without
  the `action` branch. "selects only after the copy attempt, not before it"
  fails when the selection is moved back before the attempt, printing
  `expected [ 'focus', 'attempt' ] to deeply equal [ 'attempt', 'focus' ]`. The
  two default-arrangement cases pass either way by design: they were written
  against `main` first, so they guard the six untouched call sites rather than
  describe the new code.
- `actions/CopyButton.test.tsx`: "selects a field it was given only after a
  failed attempt" fails the same way, and "leaves focus alone when no field was
  given" guards the default.
- The `@ts-expect-error` case for `action` with `multiline` is a check too, and
  a compile-time one: relaxing the union to allow both produces
  `TS2578: Unused '@ts-expect-error' directive` from
  `pnpm --dir web run typecheck`. It reports nothing under a bare
  `tsc --noEmit`, for the reason given above.

Worth knowing about one of them. "selects the value when a copy fails" asserts
the end state, and it passes whether the selection happens before or after the
copy, because the legacy fallback needs a `document.execCommand` that jsdom does
not define and so never reaches the code that would undo an early selection.
That is why the ordering is asserted separately and directly. The end-state case
was kept, but on its own it would have let the ordering bug through.

The `AppShell` pair asserts a class name rather than a rendered rule, which is
the weaker kind of test; there is precedent for that in the file, and the
rendered result was confirmed by measuring the page instead.

A note on how the classification was arrived at, because it changed one answer.
Removing a fix and re-running is only informative if the break is aimed at the
regression the case claims to prevent. The first break tried on the `AppShell`
pair was aimed at something neither case claims, and it produced a pass that
looked exactly like the pass it was meant to test. The same trap sits under
"selects the value when a copy fails": it asserts the end state and passes
whether the selection happens before or after the copy, because jsdom never
reaches the code that would undo an early one. Both were only settled by
breaking the specific thing each case names.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [x] Documentation
- [ ] Infrastructure / CI

## Relevant issues

None. Both defects were reported directly from use rather than filed.

## Checklist

Notes on the checklist:

- Tests are in `web/`, not `tests/unit` or `tests/integration`, because the
  change is entirely in the dashboard. They are `rowStyles.test.ts` and
  `AppShell.test.tsx`, next to the behavior they cover.
- `make lint` and `make typecheck` are clean and `make test-unit` passes.
  `make test` as a whole was not run: the integration suite needs PostgreSQL,
  either Docker or `TEST_DATABASE_URL`, and neither was available here. Nothing
  outside `web/` changed, so there is no `src/gateway/` behavior for it to cover,
  but the box is left unticked rather than ticked on a partial run.
- `.github/skills/frontend-standards/design-tokens.md` is corrected here. It
  claimed this repo aliases none of HeroUI's non-color variables, which is wrong
  about `--radius`, `--disabled-opacity` and `--field-border-width`, and it now
  also records that our variable block has to stay unlayered, with what breaks
  if it is wrapped in `@layer`. Verified against `globals.css` before applying:
  `--radius: 0px` at 877, `--disabled-opacity: 0.4` at 894, and
  `--field-border-width: 1px` at 425 and 634, once per theme block. Beyond that,
  the mechanism behind each fix is explained in a comment at its site, which is
  where this repo asks for it.
- No API contract changed and no route or docstring was touched, so no generated
  artifact went stale. The dashboard bundle is generated and not committed, and
  the committed generated files (`web/src/client/schema.ts` and
  `web/src/routeTree.gen.ts`) are unaffected.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

The diagnosis, the fixes, the tests and this description were all written by the
model. Several claims made along the way were wrong and were corrected by
measuring rather than by reasoning, which is worth stating because it is the
reason the numbers above are quoted rather than described: a supposed vertical
offset in the monogram did not reproduce and measures 0.000px; a first set of
collapsed-rail measurements was invalid because it was read while a CSS width
transition sat frozen in an automated tab; one measurement was taken against a
stale bundle because a `#/` route change does not reload the page; a sweep for
dead utility classes returned three different wrong answers before it was
checked against known positives; and a claim that the repo had three alembic
heads was wrong, as `scripts/check_alembic_heads.py` reports one.

**NOTE:**

## Summary

- Fixed account navigation alignment and monogram styling.
- Removed the duplicate organization sidebar divider.
- Placed email-domain verification controls on one row.
- Added clipboard fallback behavior for plain-HTTP origins and copy failures.
- Updated design-token and dashboard typecheck guidance.

## Technical notes

- Added `expandedJustify` to `navRowClass`.
- Added a typed `action` slot to `CopyField`.
- Added selection fallback support to `CopyButton`.
- Added tests for navigation, sidebar dividers, copy behavior, layout states, and shell role usage.
- Dashboard and backend lint, typecheck, and unit tests pass. Integration tests were not run because PostgreSQL was unavailable.

### DIFF
--- a/.github/skills/frontend-standards/design-tokens.md
+++ b/.github/skills/frontend-standards/design-tokens.md
@@ -172,8 +172,12 @@ share is the rule that matters: the value lives in one named place, and a compon
 Never a hex, never a numbered Tailwind palette class, never `bg-white`, because those three do
 not adapt and nothing downstream can make them.
 
-The ones we have not aliased are a gap rather than a decision, and the gap decides how
-much work a visual fix is. A value
+The ones we have not aliased are a gap rather than a decision **where
+`@heroui/styles` still declares and reads them**, and the gap decides how much work a
+visual fix is. Not every unaliased name is one: `globals.css:361` records that upstream's
+`--content1` through `--content4` are deliberately left out because HeroUI v3 neither
+declares nor reads them, so aliasing those would restore four inert lines. Check that a
+variable is live upstream before treating its absence here as a gap. A value
 computed from a variable is one alias away from being ours; the same value chased through the
 rules that read it is a selector to keep in sync with somebody else's internals, forever. A
 table whose body corners are drawn at `min(32px, var(--radius-2xl))`, which is 16px, inside a
--- a/.github/skills/frontend-standards/design-tokens.md
+++ b/.github/skills/frontend-standards/design-tokens.md
@@ -140,9 +140,28 @@ through `--radius-4xl` calculated from it in `dist/themes/shared/theme.css`, so 
 is `calc(var(--radius) * 2)`), `--field-radius`, `--spacing`, `--border-width`,
 `--field-border-width`, `--ring-offset-width`, `--disabled-opacity`, `--cursor-interactive` /
 `--cursor-disabled` and the `--scrollbar-*` family, all declared in
-`dist/themes/default/variables.css`. This repo aliases none of them, so a component's corner
-radius, its disabled dimming and its pointer all come from HeroUI's defaults rather than from
-anything we name.
+`dist/themes/default/variables.css`. This repo declares three of them, all in `globals.css`:
+`--radius` (`0px`, the square corners), `--disabled-opacity` (`0.4`) and `--field-border-width`
+(`1px`, once per theme block). The rest are not aliased, so a component's pointer, its ring
+offset and its scrollbars still come from HeroUI's defaults rather than from anything we name.
+
+**Our variable block must stay unlayered.**
+
+`@heroui/styles` declares many of the same variables we do, from inside `@layer`. Ours are
+declared in a bare `:root` at the end of `globals.css`, and being unlayered is what makes them
+win: an unlayered declaration outranks a layered one regardless of source order, so it is not
+position in the file that is holding this up. Wrap the theme blocks in `@layer` and `--radius`
+goes from `0px` to `8px` (every corner in the app rounds), the weight ladder shifts up a step
+(`medium` 400 to 500, `semibold` 550 to 600, `bold` 600 to 700), the four `--shadow-*` tokens
+stop being `none` so elevation returns on top of the hairline edges that replaced it,
+`--tracking-tight` doubles, and the 12px and 18px line-heights move by 2px each way. The font
+sizes are the trap: `--text-xs` and `--text-sm` revert to `.75rem` and `.875rem`, which compute
+to the same 12px and 14px at our 16px root, so the sizes hold and only the leading around them
+moves. Nothing fails either way: Ruff is silent, the dashboard builds, `foundation.test.ts`
+passes, and the only place the change is visible is the screen.
+
+`--spacing` is the one variable in that set we deliberately do not override; the 4px step is
+Tailwind's own.
 
 **Only the color half belongs in both theme blocks.** A color token is declared twice because
 its whole job is to hold a light value and a dark one, which is what lets a component adapt
@@ -153,7 +172,8 @@ share is the rule that matters: the value lives in one named place, and a compon
 Never a hex, never a numbered Tailwind palette class, never `bg-white`, because those three do
 not adapt and nothing downstream can make them.
 
-That is a gap rather than a decision, and it decides how much work a visual fix is. A value
+The ones we have not aliased are a gap rather than a decision, and the gap decides how
+much work a visual fix is. A value
 computed from a variable is one alias away from being ours; the same value chased through the
 rules that read it is a selector to keep in sync with somebody else's internals, forever. A
 table whose body corners are drawn at `min(32px, var(--radius-2xl))`, which is 16px, inside a
--- a/web/src/app/nav/AccountMenu.tsx
+++ b/web/src/app/nav/AccountMenu.tsx
@@ -303,9 +303,16 @@ export function AccountMenu({ collapsed }: { collapsed: boolean }) {
         // place it could reach anybody there. `AppearanceControl` folds its own
         // visible state in for the same reason.
         aria-label={`Account: ${identity.name}`}
-        className={`${navRowClass({ collapsed, band: true })} justify-start`}
+        // `expandedJustify` rather than a `justify-start` appended here: this is
+        // a HeroUI `Button`, which arrives centered, and the collapsed rail
+        // wants the monogram in the icon column instead.
+        className={navRowClass({
+          collapsed,
+          band: true,
+          expandedJustify: "start",
+        })}
       >
-        <span className="flex h-[1.625rem] w-[1.625rem] shrink-0 items-center justify-center border border-control-border bg-surface-alt text-chrome-initials font-semibold text-muted">
+        <span className="flex h-[1.625rem] w-[1.625rem] shrink-0 items-center justify-center border border-control-border bg-surface-alt text-shell-monogram font-semibold text-muted">
           {identity.initials}
         </span>
         {collapsed ? null : (
--- a/web/src/shared/components/actions/CopyField.tsx
+++ b/web/src/shared/components/actions/CopyField.tsx
@@ -1,5 +1,5 @@
 import { Button } from "@heroui/react"
-import type { ReactNode } from "react"
+import type { ReactElement, ReactNode } from "react"
 import { useEffect, useId, useRef, useState } from "react"
 import { CopyButton } from "@/shared/components/actions/CopyButton"
 
@@ -89,8 +89,12 @@ type CopyFieldProps = {
        * A control to sit beside the field, which moves the copy affordance
        * inside the field and drops the button from the label row. Absent, the
        * field renders the arrangement it always has.
+       *
+       * `ReactElement` rather than `ReactNode`: the latter admits `false`, so
+       * `action={enabled && <Button />}` compiled and then silently fell back
+       * to the default arrangement, which is the one this exists to replace.
        */
-      action: ReactNode
+      action: ReactElement
     }
 )
 
@@ -140,7 +144,7 @@ export function CopyField({
   const shared =
     "w-full rounded-lg border border-border bg-surface-alt px-3 py-2 font-mono text-xs text-foreground"
 
-  if (action) {
+  if (action !== undefined) {
     return (
       <div className="flex flex-col gap-1">
         {/* The label keeps its own row and stays a real `<label>`: it is what
--- a/web/src/shared/components/actions/CopyField.tsx
+++ b/web/src/shared/components/actions/CopyField.tsx
@@ -68,17 +68,39 @@ export function CopyableValue({
 // values are handed over in pairs and threes (a key and two snippets), so
 // "which field is this" has to be answerable by a screen reader and by a test
 // that queries the way an operator reads.
+type CopyFieldProps = {
+  label: string
+  value: string
+  fieldRef?: React.RefObject<HTMLInputElement | HTMLTextAreaElement | null>
+} & (
+  | {
+      multiline?: boolean
+      /**
+       * Excluded here rather than guarded at runtime, so a call site that passes
+       * both fails to compile. The multiline field is a `<textarea>`, where the
+       * right padding an in-field control needs indents every line instead of
+       * making room on the first, which is a different design.
+       */
+      action?: never
+    }
+  | {
+      multiline?: false
+      /**
+       * A control to sit beside the field, which moves the copy affordance
+       * inside the field and drops the button from the label row. Absent, the
+       * field renders the arrangement it always has.
+       */
+      action: ReactNode
+    }
+)
+
 export function CopyField({
   label,
   value,
   multiline = false,
   fieldRef,
-}: {
-  label: string
-  value: string
-  multiline?: boolean
-  fieldRef?: React.RefObject<HTMLInputElement | HTMLTextAreaElement | null>
-}) {
+  action,
+}: CopyFieldProps) {
   const internalRef = useRef<HTMLInputElement | HTMLTextAreaElement | null>(
     null,
   )
@@ -117,6 +139,47 @@ export function CopyField({
 
   const shared =
     "w-full rounded-lg border border-border bg-surface-alt px-3 py-2 font-mono text-xs text-foreground"
+
+  if (action) {
+    return (
+      <div className="flex flex-col gap-1">
+        {/* The label keeps its own row and stays a real `<label>`: it is what
+            names the in-field control to a screen reader ("Copy TXT record for
+            example.com"), not just a caption over the value. */}
+        <label htmlFor={fieldId} className="text-caption">
+          {label}
+        </label>
+        {/* Wraps below `xl`, where 757px of content does not fit beside the
+            rail, so `action` drops to its own line rather than squeezing the
+            field below the width the whole record needs. */}
+        <div className="flex flex-wrap items-center gap-2">
+          <div className="relative w-full xl:w-[37.5rem] xl:shrink-0">
+            <input
+              id={fieldId}
+              ref={ref as React.RefObject<HTMLInputElement>}
+              readOnly
+              value={value}
+              onFocus={(e) => e.currentTarget.select()}
+              // Right padding clears the control rather than the value running
+              // under it, and the field grows below `md` so the 44px touch
+              // floor fits between its borders. The value is never truncated:
+              // it scrolls inside the input and stays wholly selectable.
+              className={`${shared} min-h-[2.875rem] pr-14 md:min-h-0 md:pr-11`}
+            />
+            {/* `CopyButton` rather than this component's own button: it falls
+                back to `execCommand`, so it actually copies on the plain-HTTP
+                origins where the async Clipboard API does not exist. It owns the
+                acknowledgement too, which is why the field's own `aria-live`
+                line and select hint are absent from this arrangement. */}
+            <span className="absolute top-1/2 right-1 -translate-y-1/2">
+              <CopyButton value={value} label={label} selectOnFailure={ref} />
+            </span>
+          </div>
+          {action}
+        </div>
+      </div>
+    )
+  }
 
   return (
     <div className="flex flex-col gap-1">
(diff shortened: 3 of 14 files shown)


web/src/app/nav/AccountMenu.tsx
```diff
@@ -303,9 +303,16 @@ export function AccountMenu({ collapsed }: { collapsed: boolean }) {
...
+        // a HeroUI `Button`, which arrives centered, and the collapsed rail
+        // wants the monogram in the icon column instead.
+        className={navRowClass({
+          collapsed,
+          band: true,
+          expandedJustify: "start",
+        })}
       >
-        <span className="flex h-[1.625rem] w-[1.625rem] shrink-0 items-center justify-center border border-control-border bg-surface-alt text-chrome-initials font-semibold text-muted">
+        <span className="flex h-[1.625rem] w-[1.625rem] shrink-0 items-center justify-center border border-control-border bg-surface-alt text-shell-monogram font-semibold text-muted">
```
khaledosman, line 315: `text-chrome-initials` reached `main` because nothing checks that a shell type role is actually referenced. The shell-roles block in `foundation.test.ts` says "Both halves are asserted: the roles exist, and the chrome uses them", but only the declaration half is there — `it.each(SHELL_ROLES)("declares %s in globals.css")` is the only case.

Add the promised half: assert each `SHELL_ROLES` entry appears in a non-test `.tsx` under `src/app`. All three pass at this head, and it fails on `main` for `text-shell-monogram`. That is the guard that stops a fourth rename of this class.

web/src/shared/components/actions/CopyField.tsx
```diff
@@ -68,17 +68,39 @@ export function CopyableValue({
...
+      action?: never
+    }
+  | {
+      multiline?: false
+      /**
+       * A control to sit beside the field, which moves the copy affordance
+       * inside the field and drops the button from the label row. Absent, the
+       * field renders the arrangement it always has.
+       */
+      action: ReactNode
```
khaledosman: `ReactNode` does not force an element, so the union still admits a falsy action and `if (action)` then silently renders the default arrangement — the one with the label-row button and the `navigator.clipboard`-only path this change exists to get away from.

Confirmed against this head with `pnpm --dir web run typecheck`: `action={can && <button type="button" />}`, `action={undefined}` and `action={null}` all compile.

`action: ReactElement` rejects the first and routes the other two to the default branch by type rather than by truthiness.

web/src/shared/components/actions/CopyField.tsx
```diff
@@ -118,6 +140,47 @@ export function CopyField({
...
+              // it scrolls inside the input and stays wholly selectable.
+              className={`${shared} min-h-[2.875rem] pr-14 md:min-h-0 md:pr-11`}
+            />
+            {/* `CopyButton` rather than this component's own button: it falls
+                back to `execCommand`, so it actually copies on the plain-HTTP
+                origins where the async Clipboard API does not exist. It owns the
+                acknowledgement too, which is why the field's own `aria-live`
+                line and select hint are absent from this arrangement. */}
+            <span className="absolute top-1/2 right-1 -translate-y-1/2">
+              <CopyButton value={value} label={label} selectOnFailure={ref} />
```
khaledosman, line 179: Follow-up rather than this PR: `copy` at line 120 still reaches for `navigator.clipboard` directly, so the five other shared call sites, including the Keys page one-time secret reveal, keep the degraded select-and-hint path on exactly the plain-HTTP origins this swap was made for.

`const copied = await copyToClipboard(value)` in that handler carries the fix to all of them, and the markup you proved byte-identical does not change.

.github/skills/frontend-standards/design-tokens.md
```diff
@@ -140,9 +140,28 @@ through `--radius-4xl` calculated from it in `dist/themes/shared/theme.css`, so
 is `calc(var(--radius) * 2)`), `--field-radius`, `--spacing`, `--border-width`,
 `--field-border-width`, `--ring-offset-width`, `--disabled-opacity`, `--cursor-interactive` /
 `--cursor-disabled` and the `--scrollbar-*` family, all declared in
-`dist/themes/default/variables.css`. This repo aliases none of them, so a component's corner
-radius, its disabled dimming and its pointer all come from HeroUI's defaults rather than from
-anything we name.
+`dist/themes/default/variables.css`. This repo declares three of them, all in `globals.css`:
```
khaledosman, line 143: The same claim is restated at `.github/instructions/frontend-standards.instructions.md:28` ("`--cursor-interactive`, `--disabled-opacity` and the rest are in `@heroui/styles/dist/themes/`, none of them aliased in `globals.css` yet") and is not corrected here.

That file is the copy Copilot review loads, so the stale half is the one that gets acted on. Correct it in the same PR.

khaledosman (approved):
Verified at this head: `pnpm --dir web run typecheck` clean, and the five relevant test files pass. No generated artifacts owed. Four findings inline, none blocking.

web/src/shared/components/actions/CopyField.tsx
jigjigjig: Addressed in 7295bb60: `action` is `ReactElement`, and the branch selects on `action !== undefined` rather than truthiness. Added a case asserting the falsy form is rejected, and confirmed it is a real check by reverting the type to `ReactNode` and watching `pnpm --dir web run typecheck` report `TS2578: Unused @ts-expect-error directive`. Your three cases were reproduced before the fix.

web/src/app/nav/AccountMenu.tsx
jigjigjig: Addressed in 7295bb60. Confirmed the comment at foundation.test.ts:1040 does claim both halves while only the declaration case existed. Added `it.each(SHELL_ROLES)("uses %s somewhere in the chrome")`, scanning non-test `.tsx` under `src/app`. Verified it fails on the bug that shipped rather than only passing here: restoring `text-chrome-initials` at the call site gives "text-shell-monogram is declared but never used". This was the most useful comment on the PR.

.github/skills/frontend-standards/design-tokens.md
jigjigjig: Addressed in 7295bb60. The line is 29 rather than 28, and it was the only remaining site: `grep` finds that claim in just those two files and this PR had already fixed ours. Corrected in that file directly rather than by a link, since `.github/instructions/` is the one place the repo expects restatement. It now names the three variables `globals.css` does declare, says the rest are still upstream only, and carries the unlayered rule.

.github/skills/frontend-standards/design-tokens.md
```diff
@@ -153,7 +172,8 @@ share is the rule that matters: the value lives in one named place, and a compon
 Never a hex, never a numbered Tailwind palette class, never `bg-white`, because those three do
 not adapt and nothing downstream can make them.
 
-That is a gap rather than a decision, and it decides how much work a visual fix is. A value
+The ones we have not aliased are a gap rather than a decision, and the gap decides how
+much work a visual fix is. A value
```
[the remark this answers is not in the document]
jigjigjig: Addressed in 7295bb60. Verified the premise: globals.css:361 records that `--content1` through `--content4` are deliberately not carried over because HeroUI v3 neither declares nor reads them, ending "Do not restore". The sentence is now scoped to variables `@heroui/styles` still declares and reads, and names the `--content1` case as the counter-example so the distinction is stated rather than implied. Wording is close to your suggestion rather than identical.

web/src/shared/components/actions/CopyField.tsx
jigjigjig: Filed as #957 rather than fixed here, honoring your own framing. Agreed on the mechanism and confirmed it: `copy()` still reaches for `navigator.clipboard` directly, and `copyToClipboard` is the drop-in.

One correction: it is six other shared call sites, not five. KeysPage:200, :215 and :216, and SetupGuideCard:244, :262 and :267. The master key field on Settings is a separate local component at SettingsPage.tsx:281 and is unaffected.

The issue carries two traps for whoever takes it: `legacyCopy` restores the selection and focus it found on its way out (clipboard.ts:55-59), so `copy()` selecting before the attempt would be undone by the fallback it exists for; and jsdom defines no `document.execCommand`, so a test asserting only the end state passes either way and the order has to be asserted directly. Both bit this PR.

PR #963, opened 2026-09-08
fix(gateway): let the platform health probe reach a route outside base_url's own path

macaab26:
Superseded — closed without merge. See the replacement PR for the actual change and description.


macaab26:
Closing — the description referenced internal deployment details that don't belong in this public repo. Reopening clean in a new PR.

PR #966, opened 2026-09-08
feat(cli): add otari import claude-code to backfill transcript usage

shoemoney:
## Description

`docs/use-with-claude-code.md` shows how to send Claude Code usage to Otari over OTLP. That export only carries sessions which run *after* it is configured, so for anyone who installs Otari today, every token they have already burned stays invisible. For a daily Claude Code user that is most of the usage there has ever been.

Claude Code has already written it all down. It keeps one JSONL transcript per session under `~/.claude/projects`, and each assistant line carries the API response id, the model, and the Anthropic usage block. This adds `otari import claude-code`, which reads those transcripts and posts them to the existing `POST /v1/usage/external-events`. No new endpoint, no schema change, no new dependency.

```bash
otari import claude-code --dry-run                 # counts only, sends nothing
otari import claude-code --user-id alice           # backfill everything
otari import claude-code --since 7d                # top up, e.g. from cron
```

Three details drove the shape of it:

- **A response id repeats.** One reply spans several transcript lines, one per content block, each carrying the same id and the same usage block. Counting lines multiplies a session's tokens by however many blocks its replies happened to contain, so the scan deduplicates on the response id. Using that id as `source_event_id` also means the endpoint's existing `(source, source_event_id)` uniqueness does the rest: re-running imports only what is new and reports the remainder as duplicates.
- **Claude Code appends a context tag to the model**, e.g. `claude-sonnet-4-5[1m]`. Brackets are outside the endpoint's identifier grammar, so an untouched value 422s the whole batch, and the tag names a context window rather than a separately priced model, so keeping it would leave those events unpriced under a model id no price list has. It is stripped.
- **Cache writes are split by TTL.** `cache_creation_input_tokens` is the total and the 1h share is priced differently, so the 1h part is reported separately and subtracted rather than counted twice.

Provider is mapped only for model families that can be identified and is left `unknown` otherwise, since `ANTHROPIC_BASE_URL` points Claude Code at anything speaking the Messages API. An unpriced row is visibly unpriced; one priced as the wrong provider is silently wrong.

Parsing lives in `services/claude_code_import.py` as a pure module with no network and no database, so all of the above is unit-testable. The click command only wires options, calls it, and posts in batches. Prompt, completion, and tool payloads are never read out of a transcript, matching the content-free contract the endpoint already enforces.

A refused batch exits non-zero and prints what the server said, because a silent success on a failed import leaves a gap nobody goes looking for.

The docs section repeats the existing double-count warning, which matters more for a backfill than for live export: do not import sessions that were already routed through Otari.

### Validation

Beyond the unit tests, this was run against a real corpus of 14,820 transcripts spanning eight weeks. It parsed 271,506 API calls, and a second run reported the previously imported events as duplicates and created no new rows, which is the behaviour the uniqueness constraint promises.

Happy to split the docs out or open an issue first if you would rather align on the approach.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [x] Documentation
- [ ] Infrastructure / CI

## Relevant issues

None open that I could find. This extends the workflow documented in `docs/use-with-claude-code.md` and `docs/external-usage.md`.

## Checklist

- [x] If the API contract changed, I regenerated the OpenAPI spec (`uv run python scripts/generate_openapi.py`). Not applicable: no route or schema changed, this is a client of the existing endpoint.

`make lint` and `make typecheck` (mypy --strict, 569 files) are clean. `tests/unit` is 2918 passed, 2 skipped.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

**NOTE:**

## Summary

- Added `otari import claude-code` to backfill Claude Code usage from local transcripts.
- Added dry-run mode, date filtering, user IDs, labels, batching, and safe reruns.
- Handles malformed records, duplicate events, unreadable files, and transport errors.
- Excludes content payloads and documents the workflow.
- Added comprehensive parser and CLI tests.

## Technical notes

- Splits cache usage by TTL.
- Maps recognized model families to providers.
- Leaves unknown models unpriced.
- Supports budget-exempt API keys and master keys.
- Reports undecodable transcript lines and stops safely when the first batch fails.

### DIFF
--- a/docs/external-usage.md
+++ b/docs/external-usage.md
@@ -32,6 +32,13 @@ analytics would count the session twice.
 Prefer a dedicated API key for each importer. It limits attribution mistakes and
 keeps the master key out of collectors.
 
+## Backfill Claude Code from local transcripts
+
+For Claude Code specifically, the bundled `otari import claude-code` reads the
+transcripts already on disk and posts them here, which is how history from before
+any exporter was configured gets in. See
+[Use with Claude Code](use-with-claude-code.md#backfill-history-from-local-transcripts).
+
 ## Import normalized events
 
 ```bash
--- a/docs/use-with-claude-code.md
+++ b/docs/use-with-claude-code.md
@@ -87,3 +87,34 @@ Imported events are priced for analytics and never count toward budgets. Do not
 both route and export one session, or its cost will appear twice. See
 [Importing external usage](external-usage.md) for attribution, privacy,
 idempotency, and pricing behavior.
+
+## Backfill history from local transcripts
+
+The exporter above carries sessions that run after it is configured. Everything
+Claude Code did before that is already on disk, one JSONL transcript per session
+under `~/.claude/projects`. `otari import claude-code` reads those transcripts and
+posts them to the same import endpoint:
+
+```bash
+export OTARI_URL="https://otari.example.com"
+export OTARI_API_KEY="gw-your-import-key"
+otari import claude-code --user-id alice
+```
+
+The key must be budget-exempt (`exclude_from_budget: true`), the same requirement
+every import has. Use `--dry-run` first to see the event and token counts without
+sending anything, and `--since 7d` (or an ISO date) to read only recently modified
+transcripts.
+
+Only token counts and identifiers are read. Prompts, completions, and tool
+payloads are never opened out of a transcript, and the endpoint rejects them.
+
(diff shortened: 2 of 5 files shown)


daavoo (approved):
Thanks @shoemoney ! I have tested this locally and it works.

However, I will push a couple small followups on top of your work

Issue #947, opened 2026-09-07
Filter chip's dismiss button is a 24px touch target


jigjigjig:
**Type:** accessibility, needs a design decision.

`DismissChip` (`web/src/shared/components/surface.tsx`) draws its close button at
`h-6 w-6`, a 24px square. The repo's floor is 44px (`responsiveness.md`, and the rule in
`globals.css`), and that rule is keyed on `[data-slot="button"]`, which a bare `<button>`
is not, so nothing lifts it.

`FilterChips` puts `DismissChip` on Activity, Models and Usage, all three captured at
390px by the screenshot matrix.

Two more on the same mechanism in the same file: `Tab` and `Segmented`'s label both come
out around 30px from `py-[5px] text-sm`.

## Why it was not fixed in #934

Neither of the two obvious repairs is free, so this wants a design call rather than a
patch:

- **A vertical-only invisible bleed** (the compromise #927 settled on for row actions)
  would reach 24x44 with no visual change. But the chip row is `flex flex-wrap gap-2`, so
  when chips wrap onto a second line the bleeds overlap by around 12px, and adjacent X
  buttons remove *different* filters. A press landing wrong dismisses the wrong one.
- **A real 44px target at the 767px boundary** cannot overlap, but chip rows, tabs and
  segmented controls all grow visibly taller on mobile, so the filter area expands and
  pushes content down on three pages.

`Toggle` in the same file carries its target on a `before:absolute before:-inset-*`
pseudo-element, which is the pattern worth copying if the overlap can be ruled out (a
non-wrapping row, or a larger gap).

## Related

- #927, row actions, same floor and also unreachable without a layout change. Different
  cause: that one is capped by a 16px horizontal pitch between Edit and Delete.

Issue #818, opened 2026-08-26
Document organization-scoped provider keys


njbrake:
Add a docs section for the feature #670 shipped. It currently appears in `docs/` only inside the generated `openapi.json`, and nothing in `configuration.md`, `dashboard.md`, or `models.md` states any of the rules a user needs:

- **Addressing.** A bare `provider:model` selector reaches them. An `instance:model` selector never does, and a key's name is not an address, so `my-key:gpt-4o-mini` answers 400.
- **Precedence.** Workspace pin, then organization default, then oldest non-archived key.
- **Shadowing.** A `config.providers` entry, from config.yml or a stored provider instance, wins outright for the same name. A deployment instance named `openai` therefore makes every organization's BYO `openai` key unreachable, silently and with no log line. Worth a startup warning on the collision as well as a doc line.
- **Convergence.** A write refreshes the serving worker immediately; sibling workers and replicas converge within `ORG_PROVIDER_CACHE_TTL_SECONDS` (30s).

Which organization's keys a request can reach follows the workspace on its API key, which is worth stating here alongside #817.

_Note: this issue was drafted by Claude Opus 5 via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

PR #970, opened 2026-09-09
feat(dashboard): extract a design system layer with a published catalog

khaledosman:
## Description

Moves the dashboard's presentational components into `web/src/design-system/`, a
layer that may import nothing else under `src/`, and publishes every one of them
as a Storybook catalog.

This builds on @jigjigjig's `ee20d1d2`, which set up Storybook and wrote the
first 35 stories. That commit is part of this branch and this PR.

### The boundary is the point

`biome.jsonc` rejects `@/app`, `@/features`, `@/routes`, `@/shared` and
`@/client` from the new layer, and `src/architecture.test.ts` proves each
rejection by planting a module and reading Biome's answer. The rule is the
inverse of every other one in that file: the others name the few layers a layer
may not reach, this one names all of them, because the question is not "does
this import point the wrong way" but **"would this directory still compile with
the rest of `src/` deleted"**. It is meant to leave as a package, and extraction
should stay a folder move plus a specifier rewrite.

Two modules needed untangling to cross that line:

- **`errorMessage`** tested `ApiError` before `Error` and returned
  `error.message` from both. `ApiError extends Error`, so the first branch could
  never change an answer, and naming `ApiError` was the only thing coupling the
  error banner family to the transport.
- **`formatPct` and `formatRelative`** moved into the layer, with
  `@/shared/helpers/format` re-exporting them. One implementation, one
  documented import path, no call site moved. The domain formatters
  (`formatUsd`, `formatTokens`) stayed, because a spend figure is this product's
  vocabulary rather than a design system's.

`shared/components/` is now 14 files in two directories: `access/`, which reads
the deployment, and `deprecated/`.

### Components

Seventeen added or relocated, each presentational, stateless and variant driven,
with a story per axis:

`Button`, `IconButton`, `TextArea`, `Select`, `RadioGroup`, `SearchField`,
`Toggle`, `Chip`, `Kbd`, `Avatar`, `Dialog`, `Skeleton`, `Divider`, `Tooltip`,
`Popover`, `Disclosure`, `ProgressBar`, `Markdown`.

Three worth calling out:

- **`Button`** narrows HeroUI's seven variant names to the three that exist
  here. A retired variant (`outline`, `secondary`, `tertiary`, `danger-soft`) is
  now a compile error at the call site, where it used to compile, lint, ship and
  paint an unstyled button. The source scan in `foundation.test.ts` stays, since
  it still covers the direct imports this tree has not converted.
- **`Toggle`** moves out of `features/settings/` and its props now match
  `Checkbox` (`isSelected` / `isDisabled`), closing the inconsistency
  `forms.md` had flagged as real rather than as a typo in the docs.
- **`Avatar`** takes `initials` rather than a name. Deriving two letters from an
  identity is application logic (this product falls back from a full name to an
  email and splits the two differently, and counts characters rather than code
  units), and `AccountMenu`'s `initialsFor` already carries all of it.

Five plausible primitives were declined for having no consumer. DESIGN.md
records which and why, so the next person does not re-derive it.

### Four classes that emitted no CSS

Checked against the real built stylesheet, which is the only trustworthy check
here:

| Class | Why it was dead |
| --- | --- |
| `border-border-strong` | Declared in both theme blocks, never registered. `colors.md` documents it as one of three border tiers. The new `Kbd` and `Divider` were the first to name it, so their edges drew in `currentColor`. |
| `bg-chart-cat-1` through `-8`, `-other`, `ramp-1` through `-4` | All thirteen declared, none registered. `colors.md` shows one as its worked example. Charts never noticed, because `metrics/charts` passes a `var()` to recharts as a fill prop, which reads the declaration directly. |
| `text-chrome-meta` | Names no token and no utility at all. Ten places across seven story files. Replaced with `text-caption`, the role those paragraphs wanted. |

This is the same latent bug `--color-border-subtle` and `--color-text-subtle`
were each registered to fix, and globals.css documents both incidents. A gate
now fails on any `--color-*` declared in the theme blocks, named as a class in
`src/`, and absent from `@theme`. It reads HeroUI's own `@theme` from
`node_modules` so its roughly forty tokens do not read as missing.

### Three groups of CSS stop reaching into other components

- **Field density is a variable handoff.** A place declares `--field-height`
  rather than restyling its descendants. A descendant selector at (0,2,0) is
  unconditional and invisible from the call site; a custom property inherits, so
  a nested subtree can reset it.
- **The table's base is the flat plane it already claimed to be.** It painted
  `--color-surface` and an accent tinted header while its own docstring said "a
  table is not a box on the page", so all sixteen per-table classes carried the
  same three declarations to undo it. Forty-eight copies of one decision, each
  commented "Same terms as the others". The base now owns them.
- **Rendered Markdown is a component.** Thirty descendant selectors on an
  `.otari-markdown` wrapper become renderers wearing their own utilities. The
  headings now wear the type role they were reaching through `@apply`, and
  `h3`/`h4`, hand spelling 16/24 at 550, turned out to be `text-title` exactly.

### Verification

The screenshot suite is workflow dispatch only with gitignored baselines, so
this is verified by measurement. `.storybook/__tableGeometry.mjs` reads the
computed geometry of every table context in both themes at three viewports;
`_TableGeometry.stories.tsx` renders those contexts, derived from globals.css
rather than transcribed from the feature files, so it measures what the CSS
actually reaches.

It earned its keep: a first attempt moved the row separator to the base, looked
correct, and applied nothing, because an existing pair owned the same property
at the same specificity later in the file. Every wrapped table had silently gone
from the 0.03 tier to 0.06. The final run is **zero drift across all fourteen
wrapped contexts**.

One intended change remains, and it is measured rather than hidden. The three
cards that render a bare `DataTable` (`OrganizationBudgetsCard`,
`SpendCeilingsCard`, `OrganizationRosterCard`) were the ones still on the old
base, so correcting it converts them: their headers lose an accent tint and
their row separators move to the faint tier, which is the treatment every other
table already has.

### Catalog and CI

`.github/workflows/otari-design-system.yml` publishes the catalog to GitHub
Pages from `main`. **Pages has to be enabled for the repository (Settings,
Pages, source "GitHub Actions") before the publish job can claim a
deployment.**

Three jobs, split by cost. A pull request is gated on the catalog *building*,
which is 36 seconds and is where a story that stops compiling fails. The sweep
that renders all 327 stories in both themes runs on main and on demand: it takes
about ten seconds locally and over twenty minutes on a hosted runner, which is
six times the slowest existing job, and the cause is not pinned down. `publish`
waits on the sweep rather than on the build alone.

The gap that leaves is stated rather than papered over: a story that compiles
but throws at render reaches main, and the sweep reports it there. Running it
locally before pushing is the cheap mitigation
(`pnpm --dir web exec node .storybook/smoke.mjs`).

### What reviewing this branch found

Nine defects, all in the new code, all fixed on this branch:

- **`Select` offered a blank option whenever nothing was selected.** The empty
  string means "no choice made" there, not a value to preserve, but it fell into
  the same branch as a withdrawn option and was synthesized into the list.
  Confirmed against the built catalog: five options where the caller passed four,
  the first with no label. Three tests pin it.
- **The Pages concurrency key cancelled the wrong runs.** Its comment named the
  risk as two publishes racing to leave the site half-updated, then set
  `cancel-in-progress: true`, which cancels a deployment in flight and is
  precisely how that happens.
- **`publish` deployed ahead of its own verification**, needing only the build.
- **The sweep waited on `raf` polling.** It drains from a pool of pages, so all
  but one page is backgrounded, and a backgrounded page gets no animation
  frames. web/AGENTS.md records that exact mechanism.
- **The render sweep was a 20 minute blocking gate**, now split off.
- Two invented action SHAs and one duplicate action version (see the commits).
- The geometry harness said it was temporary after being kept as an instrument.
- `content/Markdown` was missing from DESIGN.md's module table, and
  web/AGENTS.md still counted ten design topic files where overlays.md made
  eleven.

Three things were considered and deliberately left, so they are decisions rather
than oversights:

- **`Meter` and `SpendMeter` use `style={{ width: … }}`.** Pre-existing and
  unchanged here, but the standards say never to inline a style and a dynamic
  percentage has no Tailwind spelling. Worth a ruling; not a silent change in
  this PR.
- **`reserveMessage` is not phrased as a question** like every other boolean prop
  in the layer. It is propagated to the new `TextArea` and `Select` for
  consistency with the existing `Field`, since two names for one thing inside a
  family reads worse than one imperfect name.
- **`.storybook/` is neither linted nor typechecked**, which `main.ts`'s own
  docstring already calls open work. That gap matters more now that a committed
  instrument lives there.

### Notes for review

- The diff is 264 files, but roughly 200 of those are a `git mv` plus an import
  rewrite. `git log -p --follow` on a moved component shows the content is
  unchanged.
- `tsc` was already failing on this branch before this PR:
  `TrendChip.stories.tsx` imported a `TrendVariant` that does not exist and
  passed a `variant` prop the component never had. Storybook's build does not
  typecheck, so it went green. Replaced with the real polarity matrix.
- Two story mocks built `models` as plain strings while the component reads
  `model.key`, so `undefined.toLowerCase()` threw on any story with a non empty
  value. That surfaced as a different failing subset on each catalog run, which
  is what prompted looking rather than writing it off as flake.
- Three gates I added first tripped on my own explanatory comments. Each now
  reads declarations rather than prose, which is the accommodation
  `FieldMessages`' docstring already documents for this reason.

## PR Type

- [x] New Feature
- [x] Refactor
- [x] Documentation
- [x] Infrastructure / CI
- [ ] Bug Fix

## Relevant issues

None filed. Related follow-ups this leaves open, unchanged from before:
`#818` (hosted provider table scoping) and `#947` (the one control under the
touch floor) are untouched here.

## Checklist

Precisely what ran, since two of those boxes deserve a qualifier:

- **Tests**: this change is `web/` only, so the tests are the frontend suites
  rather than `tests/unit` and `tests/integration`. `pnpm --dir web test` is
  green at 4833 tests across 135 files, which includes the new boundary probes
  in `src/architecture.test.ts` and the new gates in
  `src/styles/foundation.test.ts`. All 327 stories render clean in both themes
  through the same path CI uses.
- **Definition of Done**: `make lint` and `make typecheck` both pass. `make test`
  is the Python suite and **was not run**, because Docker is unavailable in this
  environment (AGENTS.md: without it the integration suite cannot start) and no
  Python file changed in this PR. Left unchecked rather than ticked, since the
  box says all three.
- **OpenAPI**: no API contract changed, so nothing to regenerate. Both committed
  generated artifacts (`web/src/client/schema.ts`, `web/src/routeTree.gen.ts`)
  are unchanged and `git diff --exit-code` on them is clean.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

Claude Opus 5, through Claude Code.

The commit and this description were written by the agent. The scoping decisions
were the repository owner's: where the layer lives and what it is called, how the
boundary is enforced, how far to take the CSS de-globalization, how broad the
component set should be, and that the table change be verified by measurement
rather than by turning the screenshot suite into a gate.


jigjigjig:
### Heads-up: #985 landed on main first, so this branch will conflict in a handful of places

#985 (Web search / Code execution redesign) was originally stacked on this
branch for its Storybook. It has been rebased onto `main` and merges on its own,
so **nothing here is blocked by it** — but it touches nine files this branch
also rewrites. Writing down what it did and why, so resolving is mechanical
rather than archaeological.

#### What #985 added, in this branch's terms

Five new components, all of which belong in `design-system/` once this lands,
not in `shared/components/` where #985 had to put them:

| #985 put it at | Should end up at |
| --- | --- |
| `shared/components/layout/SettingRow.tsx` (+ test) | `design-system/layout/` |
| `shared/components/navigation/DisclosureRow.tsx` (+ test) | `design-system/navigation/` |
| `shared/components/navigation/DocsLink.tsx` | `design-system/navigation/` |

Plus `shared/hooks/useAutosave.ts`, `shared/helpers/docs.ts` and
`features/tools/PolicyRow.tsx`, which stay where they are — none of them is a
design-system primitive.

**They have no stories.** #985 wrote `SettingRow.stories.tsx`,
`DisclosureRow.stories.tsx`, `DocsLink.stories.tsx` and a `Bounded` story on
`SettingsGroup.stories.tsx` while it was stacked here, then dropped them in the
rebase because main has no Storybook. They are recoverable verbatim from
`feat/web-search-page-redesign@2374bb13` if you want them back when this merges;
that commit is the last one before the rebase.

#### The conflicts, and how each resolves

**`src/styles/globals.css`** — the interesting one. #985 needed a third dense
place (`.otari-settings`, 32px controls beside a label, 36px at 16px below `md`
so iOS does not zoom) and wrote it in **main's descendant-selector idiom**:

```css
.otari-settings .input,
.otari-settings .select__trigger { height: 32px; min-height: 32px; padding-block: 4px; }
```

This branch replaces that idiom with `--field-height` / `--field-padding-block`
declared by the place. **Take this branch's version and re-express the settings
place as variables**, exactly as `.otari-toolbar` and `.otari-pagination` are
here:

```css
.otari-toolbar, .otari-pagination, .otari-settings { --field-height: 32px; --field-padding-block: 4px; }
@media (width < 48rem) { .otari-settings { --field-height: 36px; --field-padding-block: 6px; } }
```

The mobile font size is the part worth care. `.otari-settings` needs 16px below
`md`, and `.field-machine` (the mono lane for URLs, engine lists, domain lists
and numbers) needs 12px on a desk and 16px below `md`. If you add a
`--field-font-size` variable for that, **note that HeroUI ships `.input` at 16px
with a `@media (width>=40rem)` step down to 14px, and an unlayered flat rule
beats both.** #985 shipped that bug briefly and it silently removed the iOS
zoom guard from every text field in the product; the fix was to give the root
default HeroUI's own two values and keep the variable off `.select__trigger`,
which is never typed into. `d1e46a4b` on the pre-rebase branch has it.

**`src/styles/foundation.test.ts`** — #985 deletes the
`WorkspaceCodeExecutionPolicyCard.tsx` entry from `CAPTION_SIZE_IS_RULED`,
because that card no longer hand-rolls the caption role. Keep the deletion.

Separately, and **not** in #985 on main: while it was stacked here, its three
extra test files changed worker scheduling enough to expose a race between this
branch's `foundation.test.ts` (which sweeps `src/**` at module load, in eight
places) and `architecture.test.ts` (which plants a Biome probe under
`src/features/` for the length of one assertion). The suite went red
intermittently with the probe reported as an offender.
`overlaySeams.test.ts` already skips it; this branch's `foundation.test.ts` does
not. Worth fixing here with a single `sourceFiles()` helper — `ba75cab0`'s
parent `d1e46a4b`..`c5f30396` range on the pre-rebase branch has that patch.
main's `foundation.test.ts` is a smaller file and does not hit it.

**`layout/SettingsGroup.tsx`** (+59/-12) — a `bounded` variant (unruled heading
over a framed, non-bleeding block of rows, carrying the `.otari-settings` class)
and a `docsHref` slot. Both are additive; the existing full-width path is
refactored only to share the heading/blurb between the two branches. Take both
sides' changes.

**`layout/PageIntro.tsx`** (+16/-1) — a `docsHref` slot that trails the
description. Additive.

**`actions/CopyField.tsx`** (+19/-8) — a new `inFieldCopy` flag. Previously the
in-field copy arrangement was selected by passing `action`; it is now selected
explicitly, and `action` became optional so a value with no trailing action can
use the same field. `OrganizationDomainsPage` and `CopyField.test.tsx` gain
`inFieldCopy` at their call sites.

**`actions/CopyButton.tsx`** (+6) — a 44px hit area via
`before:absolute before:-inset-1.5`. It is 32px at every call site, under the
floor `motion-and-access.md` sets. Verified across 7 pages and 18 buttons that
the 6px bleed overlaps no other interactive element.

**`features/tools/*`** — `ToolsGuardrailsPage`, `SearchToolsCard`,
`WorkspaceWebSearchCard`, `WorkspaceCodeExecutionPolicyCard` and their tests are
rewritten wholesale by #985. **Take #985's version and rewrite the imports**
(`@/shared/components/*` → `@/design-system/*`); the only other adjustments are
`Button` (#985 imports it from `@heroui/react` because main has no wrapper — use
`design-system/actions/Button` here) and a local `Placeholder` in
`ToolsGuardrailsPage`, which should become `design-system/feedback/Skeleton`.

#### Why it went this way

The design brief for #985 asked for Storybook stories, so it started here. Fede
then wanted it independent of this branch, which is currently `CONFLICTING`
against main and therefore an uncertain merge date. Rebasing cost the stories
and one re-expression of the dense place; keeping the stack would have meant
#985 waiting on this. Happy to do the reverse-port (move the three components
into `design-system/`, restore the stories, convert the settings place to
variables) as a follow-up once this merges — say the word and I will.

Issue #786, opened 2026-08-25
Four list endpoints have no pagination bound


khaledosman:
Four list endpoints select without a `limit`, so each returns every matching row in one response and grows unbounded with the table behind it.

- `GET /v1/files` (`src/gateway/api/routes/files.py`, `list_files`): every live file for the caller.
- `GET /v1/aliases` (`src/gateway/api/routes/aliases.py`, `list_aliases`): every stored alias, plus config ones.
- `GET /v1/routing/policies` (`src/gateway/api/routes/routing.py`, `list_policies`): every stored policy. Each row also carries a JSON `spec`, so the payload grows faster than the row count suggests.
- `GET /v1/batches` (`src/gateway/api/routes/batches.py`, `list_batches`): the provider page is bounded, but the ownership filter runs after it, so a page can come back short and there is no way to ask for the next one deterministically.

This violates the repo's own rule in [.github/instructions/performance-review.instructions.md](https://github.com/mozilla-ai/otari/blob/main/.github/instructions/performance-review.instructions.md): "Every list endpoint has a hard upper bound" and "Pagination (`limit`/`skip`) with a sane default and maximum on every list endpoint". `GET /v1/keys` and `GET /v1/usage` already do this, so the shape to copy is in the codebase.

## Suggested shape

`limit`/`skip` query parameters with a sane default and a server-side maximum, matching `GET /v1/usage`. The dashboard pages should be updated alongside, since a default limit changes what an unparameterized call returns.

## Why a separate issue

Found during review of #784, which scopes these endpoints to a workspace. The missing bounds predate that change: it narrows what each returns and adds no new unbounded read. Fixing four endpoints' pagination inside an already-large tenancy PR would be scope creep, and the dashboard change that has to land with it belongs in its own review.

peteski22:
`GET /v1/files` is bounded since #1366: `limit` defaults to 100, with a maximum of 1000, and it pages with a cursor (`after` for the OpenAI flavor, `after_id` for the Anthropic flavor). #1473 moves the Anthropic flavor to `page`. The cursor style is the one the OpenAI and Anthropic SDKs send, as `GET /v1/batches` already does with `after`. Tracked in #1470.

PR #976, opened 2026-09-09
feat(files): serve both SDKs, seed the sandbox with uploads, collect its outputs, sweep expired files

daavoo:
## Description

Octonous talks to provider files APIs only through the official Anthropic and OpenAI SDKs pointed at a base URL, references uploads with `image`/`document` `file` sources, `container_upload`, `input_file`, and expects code execution to see attachments and hand back downloadable files. Otari's files API already stored and resolved files, but only in the OpenAI shape, never into the sandbox, and never out of it. This closes those gaps so an open-source model behind Otari can take the same requests a frontier model does.

- **Anthropic SDK compatibility.** A request carrying `anthropic-version` (the SDK sends it on every call) gets `FileMetadata` (`type`, `size_bytes`, `mime_type`, `downloadable`, RFC 3339 `created_at`) and `file_deleted` on delete. OpenAI callers are unchanged.
- **Cursor pagination on the file listing:** `limit`, `after`/`after_id`, `order`, `has_more`, `first_id`, `last_id` (part of #786).
- **Uploads reach the sandbox.** With `otari_code_execution` in the request, every referenced upload is seeded into the session via the protocol's `PutFile`. `container_upload` blocks are staged and replaced by a marker for the model; without a sandbox they are read as documents. A refused seed fails the request rather than running over a missing input.
- **Sandbox outputs become files.** File references in the result block are fetched with `GetFile`, stored as `code_execution_output` files owned by the same user and workspace, and named with their `file_id` in the tool result. Storage failures never fail the run.
- **Bare Responses `input` items** (`input_file`, `input_image` outside a message) are now normalized.
- **Retention sweep** reclaims bytes and rows of expired and deleted files (`files_sweep_interval_sec`, hourly, `0` disables). It is one entry in the lifespan worker registry, and commits through the Unit of Work like every other service.

Docs (`docs/files.md`, protocol doc), OpenAPI spec, and Postman collection are regenerated.

### Not in this PR

- Files in hybrid mode: hybrid has no database, so that is an architecture decision, not a gap.
- Container-file ids (`cfile_`) and code-interpreter annotations on the Responses wire: needs the Anthropic-content-block lift the sandbox backend already notes.
- The reference `otari-sandbox-container` does not populate the file-reference list yet, so output collection is wired but idle until it does.
- The fsspec storage backend is #977, stacked on this one.

## How to test it locally

1. Upload a file with the OpenAI SDK and with the Anthropic SDK, both pointed at the gateway; each gets its own response shape, and `GET /api/v1/files?limit=2` pages with `has_more` and `last_id`.
2. Send a chat request with `otari_code_execution` and a `file_id` reference against a sandbox; the file is on disk in the session, and any file the run writes comes back as a `file_id` in the tool result and is fetchable from the files API.
3. Set `files_retention_hours: 1`, wait past expiry (or set `files_sweep_interval_sec` low), and watch the sweep log line reclaim the bytes and rows.

Covered by `tests/integration/test_files_endpoint.py` (Anthropic shapes, paging, sweep), `tests/unit/test_sandbox_backend.py` (seeding and output collection), `tests/unit/test_content_normalizer.py` (bare Responses items), and `tests/unit/test_gateway_lifespan_shutdown.py` (the sweep worker starts and can be switched off). `make lint`, `make typecheck`, the unit suite, the OSS smoke gate, and the files, messages-dispatch, hybrid-chat and settings integration tests pass locally.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Part of #786.

## Checklist

## AI Usage

- [ ] No AI was used.

Rebased onto main after the models package split, the lifespan worker registry, the derived settings view, and the Unit of Work transaction rule landed; the change was adapted to each.

### DIFF
--- a/docs/public/openapi.json
+++ b/docs/public/openapi.json
@@ -17106,7 +17106,7 @@
     },
     "/v1/files": {
       "get": {
-        "description": "List the authenticated user's uploaded files in the request's workspace.\n\n``workspace_id`` narrows a master-key listing to one workspace; a keyed\nrequest is already confined to its key's own and cannot widen or move it.",
+        "description": "List the authenticated user's uploaded files in the request's workspace.\n\n``workspace_id`` narrows a master-key listing to one workspace; a keyed\nrequest is already confined to its key's own and cannot widen or move it.\n\nPages are cursor-based: ``after`` (OpenAI) or ``after_id`` (Anthropic) names\nthe last file of the previous page, and ``has_more`` says whether to ask\nagain. A cursor the caller cannot see (another user's file, a deleted one)\nis a 404, the same answer a direct read of it gets.",
         "operationId": "list_files_v1_files_get",
         "parameters": [
           {
@@ -17157,6 +17157,64 @@
               ],
               "title": "Workspace Id"
             }
+          },
+          {
+            "in": "query",
+            "name": "limit",
+            "required": false,
+            "schema": {
+              "default": 100,
+              "maximum": 1000,
+              "minimum": 1,
+              "title": "Limit",
+              "type": "integer"
+            }
+          },
+          {
+            "in": "query",
+            "name": "after",
+            "required": false,
+            "schema": {
+              "anyOf": [
+                {
+                  "type": "string"
+                },
+                {
+                  "type": "null"
+                }
+              ],
+              "title": "After"
+            }
+          },
+          {
+            "in": "query",
+            "name": "after_id",
+            "required": false,
+            "schema": {
+              "anyOf": [
+                {
+                  "type": "string"
+                },
+                {
+                  "type": "null"
+                }
+              ],
+              "title": "After Id"
+            }
+          },
+          {
+            "in": "query",
+            "name": "order",
+            "required": false,
+            "schema": {
+              "default": "desc",
+              "enum": [
+                "asc",
+                "desc"
+              ],
+              "title": "Order",
+              "type": "string"
+            }
           }
         ],
         "responses": {
@@ -17197,7 +17255,7 @@
         ]
       },
       "post": {
-        "description": "OpenAI-compatible file upload endpoint.",
+        "description": "Upload a file. Answers in the OpenAI or Anthropic file shape, following the caller's headers.",
         "operationId": "create_file_v1_files_post",
         "requestBody": {
           "content": {
--- a/src/gateway/api/routes/files.py
+++ b/src/gateway/api/routes/files.py
@@ -12,18 +12,23 @@ request-plane row does (``services/workspace_scope``). A keyed request is confin
 to its own key's workspace on every verb; a master-key request is the operator
 acting deployment-wide and sees every workspace, narrowable on the listing with
 ``workspace_id``, matching ``GET /v1/keys``.
+
+The same five routes serve two SDKs. OpenAI's and Anthropic's Files APIs share
+their paths and verbs and differ only in the JSON they return, so the response
+shape follows the caller: a request carrying Anthropic's ``anthropic-version``
+header (which its SDK sends on every call) gets ``FileMetadata``, everything
+else gets the OpenAI file object.
 """
 
-import mimetypes
 import uuid
 from collections.abc import AsyncGenerator, AsyncIterator
-from datetime import UTC, datetime, timedelta
-from typing import Annotated, Any
+from datetime import UTC, datetime
+from typing import Annotated, Any, Literal
 from urllib.parse import quote
 
-from fastapi import APIRouter, Depends, File, Form, HTTPException, UploadFile, status
+from fastapi import APIRouter, Depends, File, Form, HTTPException, Query, Request, UploadFile, status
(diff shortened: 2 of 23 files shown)


HareeshBahuleyan (commented):
Review by ChatGPT (AI assistant in pi), posted with user approval. Five findings from the PR/base-relative review are attached inline.

src/gateway/services/sandbox_backend.py
```diff
@@ -208,8 +247,73 @@ async def __aenter__(self) -> SandboxBackend:
...
+                response = await self._client.get(
+                    f"{self._sandbox_url}/sessions/{self._session_id}/files",
+                    params={"path": ref.filename},
+                )
+                response.raise_for_status()
+            except httpx.HTTPError as exc:
+                logger.warning("sandbox output %r could not be fetched: %s", ref.filename, exc)
+                continue
+            data = response.content
+            if not data or len(data) > self._files.max_output_bytes:
```
HareeshBahuleyan, line 308: **[P1] Bound sandbox downloads before buffering**

`httpx.get()` downloads the entire output before checking `max_output_bytes`. Large sandbox files can exhaust gateway memory despite the configured limit. A 4-byte-limit probe consumed all 40 bytes.

**Fix:** Stream downloads and abort when the limit is exceeded.

src/gateway/services/sandbox_backend.py
```diff
@@ -208,8 +247,73 @@ async def __aenter__(self) -> SandboxBackend:
...
+        for staged in self._files.inputs:
+            try:
+                data = await self._files.read_input(staged)
+            except OSError as exc:
+                raise SandboxNotReachableError(f"could not read attachment {staged.file_id} for the sandbox") from exc
+            try:
+                response = await self._client.post(
+                    f"{self._sandbox_url}/sessions/{self._session_id}/files",
+                    files={"file": (staged.filename, data, staged.mime_type)},
+                    data={"path": staged.filename},
```
HareeshBahuleyan, line 278: **[P2] Prevent same-name attachments from overwriting**

Distinct file IDs named `data.csv` are uploaded to the same sandbox path. Reproduction left only the second file’s contents, silently corrupting multi-file analysis.

**Fix:** Allocate unique paths and expose those paths consistently to the model.

src/gateway/api/routes/files.py
```diff
@@ -263,16 +282,51 @@ async def list_files(
...
-    stmt = stmt.order_by(FileObject.created_at.desc())
 
-    result = await db.execute(stmt)
-    records = result.scalars().all()
-    return {"object": "list", "data": [r.to_dict() for r in records]}
+    cursor_id = after or after_id
+    if cursor_id is not None:
+        cursor = await fetch_file(db, cursor_id, user_id, workspace_id=scope)
+        if cursor is None:
+            raise HTTPException(status_code=status.HTTP_404_NOT_FOUND, detail="File not found")
```
HareeshBahuleyan, line 290: **[P2] Keep pagination and cursor expiry rules consistent**

Listings include expired files, but cursor lookup rejects them. Reproduction returned an expired `last_id` with `has_more=true`; following that cursor returned 404, preventing access to subsequent live files.

**Fix:** Exclude expired files from listings and define consistent cursor-expiry behavior.

src/gateway/services/content_normalizer.py
```diff
@@ -354,6 +417,21 @@ async def _normalize_block(
...
+
+def _wrap_bare_responses_item(item: Any) -> Any:
+    """Put a bare item the normalizer turned into text back into a valid position.
+
+    A file or image item is valid at the top level of a Responses ``input``, but
+    the text it extracts to is not: ``input_text`` only lives inside a message.
+    """
+    if isinstance(item, dict) and item.get("type") == "input_text":
+        return {"role": "user", "content": [item]}
+    return item
```
HareeshBahuleyan, line 432: **[P2] Wrap native Responses content parts in messages**

The new bare-input handling wraps extracted text only. Native PDFs/images remain top-level `input_file`/`input_image` items, which fail OpenAI’s `ResponseInputItemParam` schema validation. Wrapping the same result in a user message validates.

**Fix:** Wrap all supported bare content parts, including native passthrough results.

docs/public/openapi.json
```diff
@@ -17157,6 +17157,64 @@
               ],
               "title": "Workspace Id"
             }
+          },
+          {
```
HareeshBahuleyan, line 17161: **[P2] Regenerate the dashboard API client**

The spec adds pagination parameters, but `web/src/client/schema.ts` remains unchanged. Regeneration produces a diff, failing the dashboard’s explicit client-drift gate.

**Fix:** Run `pnpm --dir web client:generate`.

daavoo:
Continued in https://github.com/mozilla-ai/otari/pull/1364: GitHub refuses to reopen a PR after its branch is force-pushed, and the branch was rebased onto current main.

PR #977, opened 2026-09-09
feat(files): fsspec backend so any filesystem can hold uploaded files

daavoo:
Stacked on #976 (`feat/files-api-parity`); only the last commit is this PR.

## What

`files_backend: fsspec` with `files_url` (`gcs://bucket/prefix`, `abfs://container/prefix`, `s3://bucket/prefix`, `sftp://host/path`, `file:///path`, `memory://`) and `files_storage_options` for the implementation's own keyword arguments (credentials, endpoints, projects). `FsspecFileStore` implements the same `FileStore` protocol as the local and boto3 S3 backends: writes and reads stream through worker threads, a partial upload is removed on failure or cancellation, backend exceptions are translated to the `OSError` family callers already handle, and a ref that could leave the root is refused.

fsspec was already in the dependency tree through any-llm and is now declared; the lock gains only that entry. `files_storage_options` is kept out of the settings view as a credential. Operators install the implementation package for their protocol (`gcsfs`, `adlfs`, `s3fs`, `paramiko`).

## Validation

Unit tests run on fsspec's built-in `memory://` and `file://` filesystems, so no cloud package or network is needed. `make lint`, `make typecheck`, unit suite, OSS smoke gate, and the files/config/settings integration tests pass.


daavoo:
Folded into https://github.com/mozilla-ai/otari/pull/1364 together with #976, rebased onto current main.

Issue #813, opened 2026-08-26
Add gateway-managed Web Fetch tool


HareeshBahuleyan:
## Desired behavior

Otari exposes `otari_web_fetch` on `/v1/chat/completions`, `/v1/messages`, and `/v1/responses`. The model supplies a URL, and the gateway downloads it and returns bounded, LLM-ready content from HTML, text, or PDF documents.

`otari_web_fetch` and `otari_web_search` must work together in the same gateway tool loop. Fetching must work in standalone and hybrid modes without requiring Tavily or another commercial fetch provider.

## Why

mozilla-ai/octonous#4950 needs its Web Search sub-agent to use Otari-managed Search and Fetch. Otari currently has `otari_web_search` but no callable URL-fetch tool, so the sub-agent must retain a direct-provider credential path.

Otari already fetches and extracts the top Web Search results in `src/gateway/services/web_search_backend.py`. The missing work is to make that behavior reusable and expose it as a first-class managed tool.

## Notes

- **Tool contract.** Add the canonical `{"type": "otari_web_fetch"}` declaration and a model-facing function that accepts a URL. Advertise it through `GET /v1/tools`.
- **Provider-native tools.** Anthropic declarations such as `web_fetch_20250910` and `web_fetch_20260209` remain provider-native and pass through unchanged. Do not intercept them or claim Anthropic-specific dynamic-filtering semantics in this issue. Octonous should translate its Web Fetch tool to `otari_web_fetch` when using its Otari model route.
- **Shared backend.** Extract the existing bounded HTTP fetch and extraction logic from `WebSearchBackend` into a shared service. Use that service for both search-result enrichment and `otari_web_fetch`. A single web-retrieval backend should own both tool names so Search and Fetch can run together without introducing another mutually exclusive tool backend.
- **Content extraction.** Preserve trafilatura extraction for HTML and reuse `src/gateway/services/file_extractors.py` for PDF and text/document extraction where applicable. Return bounded text or Markdown to the model.
- **Security.** Reuse `validate_outbound_fetch_url`. Validate every redirect, reject unsafe schemes and private or reserved destinations by default, and enforce redirect, timeout, response-size, and extracted-content limits. Do not accept caller-supplied request headers or credentials.
[...]

PR #979, opened 2026-09-09
feat(web-fetch): Part 3 activate managed Fetch

HareeshBahuleyan:
## Description

Activates `otari_web_fetch` across Chat Completions, Messages, and Responses on top of the bounded retrieval foundation and Search migration in #959.

This stack layer:

- adds exact managed Fetch declaration validation and provider-native Fetch pass-through
- runs Fetch through the shared bounded Web Retrieval backend with display-safe output and sanitized failures
- enforces one request-scoped 10-call allowance across Search, Fetch, streaming continuations, and routing attempts
- resolves standalone and hybrid Web Access policy once, requiring explicit `web_fetch` authorization from the capability-aware control-plane contract
- meters successful calls as `otari:web_fetch` / `web_fetch_calls`, leaves failures unbilled, and exposes Fetch in Usage and `/v1/tools`
- updates the workspace Web Access policy presentation and public documentation
- regenerates OpenAPI, Postman, and the dashboard API client

Stack base: #959
Upstream contract dependency: mozilla-ai/otari-ai#2078 must be merged and deployed before this PR is merged or deployed.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [x] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Part of #813

## Checklist

Validation completed locally:

- `make lint`
- `make typecheck`
- `make openapi-check`
- `make postman-check`
- full backend unit suite: 3,053 passed, 9 skipped
- focused backend unit suite after final edits: 92 passed
- relevant integration suites passed across route dispatch, hybrid policy, workspace policy, billing, Usage, startup pricing, and provider-native pass-through coverage
- dashboard lint and typecheck
- dashboard tests: 3,845 passed
- dashboard production build

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:** GPT-5 via Codex

**Any additional AI details you'd like to share:** Implemented and verified in an isolated git worktree against the Part 3 boundaries in the repository design document and the concrete capability contract in mozilla-ai/otari-ai#2078.

**NOTE:**

## Summary

- Added managed `otari_web_fetch` support across Chat Completions, Messages, and Responses.
- Unified Search and Fetch authorization, domain controls, billing, usage tracking, and request limits.
- Added bounded Fetch results and sanitized errors.
- Kept managed Fetch disabled by default, with YAML and `OTARI_WEB_FETCH_ENABLED` opt-in support.
- Exposed Fetch through `/v1/tools`, documentation, dashboard settings, Postman collections, and generated clients.
- Added backend, integration, dashboard, and contract test coverage.

## Technical notes

- Provider-native Fetch tools continue to pass through unchanged.
- Web Access policy resolution supports Search-only, Fetch-only, and combined requests.
- Invalid declarations and malformed policies fail before network or policy work.
- MCP iteration-limit errors now propagate correctly.


src/gateway/api/routes/_pipeline.py
```diff
@@ -2566,105 +2717,101 @@ async def prepare_gateway_tools(
...
                 try:
                     workspace_search = await resolve_workspace_web_search_config(ctx.db, ctx.workspace_id)
                 except InvalidStoredWebSearchDomainError as exc:
                     raise adapter.error(503, WEB_SEARCH_CONFIG_INVALID_DETAIL, ErrorKind.API) from exc
                 if workspace_search is not None:
                     if not workspace_search.enabled:
-                        raise adapter.error(403, WEB_SEARCH_NOT_ENABLED_DETAIL, ErrorKind.PERMISSION)
+                        detail = WEB_ACCESS_NOT_ENABLED_DETAIL if use_web_fetch else WEB_SEARCH_NOT_ENABLED_DETAIL
+                        raise adapter.error(403, detail, ErrorKind.PERMISSION)
+                    mandatory_policy = _policy_from_domain_values(
```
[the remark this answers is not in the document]
HareeshBahuleyan: Fixed in `8565fd45`. Standalone admission now composes combined Search request domain filters into the Fetch policy even when no workspace policy row exists. Added regression coverage for both allowed and blocked request domains in the no-row case.

src/gateway/api/routes/_pipeline.py
```diff
@@ -2750,6 +2901,8 @@ async def _require_tool_pricing(
     tools = [CODE_EXECUTION_TOOL_NAME] if use_sandbox else []
     if use_web_search:
         tools.append(WEB_SEARCH_TOOL_NAME)
+    if use_web_fetch:
+        tools.append(WEB_FETCH_TOOL_NAME)
```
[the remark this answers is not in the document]
HareeshBahuleyan: Fetch intentionally uses the same reservation and settlement design as managed Web Search. Both verify pricing at admission and add successful tool charges during settlement. We are keeping those semantics aligned in this activation PR rather than introducing Fetch-only budget behavior.

web/src/client/schema.ts
```diff
@@ -14601,8 +14601,8 @@ export interface operations {
                 api_key_id?: string[] | null;
                 /** @description Filter by token-pricing state: true = only rows whose model tokens were priced, false = only rows that still need pricing (no cost at all, or tokens that were never metered because the model had no rate). A row charged only for gateway-run tool calls still counts as needing pricing. */
                 priced?: boolean | null;
-                /** @description Filter to requests that ran a gateway-run tool. 'any' matches any tool; a tool name (web_search, code_execution) matches that tool specifically. */
-                tool?: ("any" | "web_search" | "code_execution") | null;
+                /** @description Filter to requests that ran a gateway-run tool. 'any' matches any tool; a tool name (web_search, web_fetch, code_execution) matches that tool specifically. */
+                tool?: ("any" | "web_search" | "web_fetch" | "code_execution") | null;
```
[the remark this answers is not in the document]
HareeshBahuleyan: This is intentionally deferred to the follow-up dashboard PR. This activation PR adds `web_fetch` to the backend filters and API schema; the follow-up adds the Activity option and tests, as specified in PR 4.

HareeshBahuleyan:
@coderabbitai review

src/gateway/api/routes/_pipeline.py
```diff
@@ -2566,105 +2720,102 @@ async def prepare_gateway_tools(
...
+                    raise adapter.error(403, detail, ErrorKind.PERMISSION)
+                if not set(requested_tools) <= authorized_tools:
+                    raise adapter.error(403, WEB_ACCESS_TOOL_NOT_AUTHORIZED_DETAIL, ErrorKind.PERMISSION)
+                try:
+                    web_fetch_policy = _combined_fetch_policy(
+                        mandatory_policy,
+                        web_search_tool_entry if use_web_fetch else None,
                     )
+                except DisjointDomainAllowListsError as exc:
+                    raise adapter.error(403, WEB_ACCESS_DOMAINS_EXCLUDED_DETAIL, ErrorKind.PERMISSION) from exc
```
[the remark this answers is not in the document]
HareeshBahuleyan: Agreed. This rollout dependency affects hybrid mode only; standalone mode resolves Web Access policy locally and is unaffected. The strict hybrid validation is intentional so Fetch cannot be authorized by a legacy response. `mozilla-ai/otari-ai#2078` is now merged, and this PR’s description records that its contract must be deployed before this gateway change is merged or deployed. We will keep this thread open until that deployment is confirmed.

Drafted with Pi.

src/gateway/api/routes/_pipeline.py
```diff
@@ -2466,6 +2514,109 @@ def _canonicalize_web_search_request_domains(tool_entry: dict[str, Any]) -> None
...
+    workspace_allowed = workspace_policy.allowed or None
+    allowed = intersect_domain_allow_lists(workspace_allowed, request_allowed) or ()
+    blocked = union_domain_block_lists(workspace_policy.blocked or None, request_blocked)
+    return DomainPolicy(allowed=allowed, blocked=blocked)
+
+
+def _validated_hybrid_web_policy(payload: dict[str, Any]) -> tuple[bool, set[str], DomainPolicy]:
+    """Strictly validate the authorization fields supplied by the control plane."""
+    enabled = payload.get("enabled")
+    authorized = payload.get("authorized_tools")
```
daavoo: Requiring `authorized_tools` on every response makes this PR break Search, not just gate Fetch. The deployed contract has no such field, so a hybrid gateway that upgrades before otari-ai#2078 ships gets a 502 on a plain `otari_web_search` request: a feature it already had, removed by a field added for a tool it never enabled.

That coupling is fine when one team deploys both sides together. Here self-hosters point their own gateway at otari.ai and upgrade on their own schedule, so the ordering cannot be enforced.

Default the absent case to Search and keep the 502 for a field that is present and malformed:

```python
authorized = payload.get("authorized_tools")
if authorized is None:
    # A platform that predates the capability contract can only mean Search:
    # Fetch ships with the field, so its absence is never an authorization.
    authorized = [WEB_SEARCH_TOOL_NAME]
elif not isinstance(authorized, list) or any(not isinstance(v, str) for v in authorized):
    raise ValueError("authorized_tools must be a list of strings")
```

Fetch stays fail-closed, because no old platform can name it. This also drops the "upstream must be merged and deployed before this PR" dependency from the description, so this can merge on its own.

tests/fixtures/web_access_resolution_contract.json
```diff
@@ -0,0 +1,69 @@
...
+      "tools": [{"type": "otari_web_fetch"}],
+      "platform_response": {
+        "enabled": "true",
+        "authorized_tools": ["web_fetch"]
+      },
+      "expected_requested_tools": ["web_fetch"],
+      "expected_status": 502
+    },
+    {
+      "name": "legacy_response",
```
daavoo, line 60: The legacy case is only covered for Fetch. Add the Search-only one, which is where the behavior change actually bites:

```json
{
  "name": "legacy_response_search_only",
  "tools": [{"type": "otari_web_search"}],
  "platform_response": {"enabled": true, "allowed_domains": ["example.com"]},
  "expected_requested_tools": ["web_search"],
  "expected_status": 200
}
```

With the default suggested on `_pipeline.py`, the existing `legacy_response` case becomes 403 rather than 502: the policy parses, Fetch is simply not authorized.

docs/hybrid-mode-protocol.md
```diff
@@ -337,15 +342,19 @@ The request body is empty; the workspace is identified by `X-User-Token`.
...
-its own value: `max_results`, `allowed_domains`, `blocked_domains`, and
-`purpose_hint` fill in when the per-request tool entry omits them (an empty list
-or empty string reads as "no preference" and does not clear the workspace
-value), and `provider_options` is shallow-merged with per-request keys winning.
-`provider` is informational: the active web-search backend is configured on the
-gateway itself via `OTARI_WEB_SEARCH_URL`, so Otari does not switch backends based on
-this field.
+Otari requires `enabled` to be a boolean and `authorized_tools` to be a list of
+strings. Every requested tool must be explicitly authorized; in particular, a
+legacy response that omits `authorized_tools` never permits Fetch. Recognized
```
daavoo: This paragraph describes compatibility in one direction and leaves out the one that regresses. "`{}` means a legacy Search-only request" is an old gateway talking to a new platform, which is a rule for otari-ai and is genuinely compatible. The response side is a new gateway talking to an old platform, and today that fails for Search too, not only Fetch.

If you take the default suggested on `_pipeline.py`, this text becomes true as written. If you keep the strict check, say plainly here that a response without `authorized_tools` is rejected and the platform must deploy first or Search stops working.

daavoo (commented):
Reviewed against the repo's scoped guidance (`src/gateway/AGENTS.md`, backend-standards, frontend-standards, the security and performance instruction files). CI is green and all three drift-checked artifacts (openapi.json, the Postman collection, `web/src/client/schema.ts`) are in the diff.

One item is a regression in an existing feature rather than a problem with the new tool: the strict `authorized_tools` check breaks Search on any hybrid gateway that upgrades before the platform does. Details inline. The rest is smaller.

Skipping the points Copilot and CodeRabbit already raised and resolved.

_Reviewed by Claude Opus 5 via Claude Code._

web/src/features/tools/WorkspaceWebSearchCard.tsx
```diff
@@ -168,15 +168,17 @@ export function WorkspaceWebSearchCard({ docsHref }: { docsHref: string }) {
         <div className="px-4 py-3">
           <InfoBanner>
             This deployment has no in-loop search backend configured, so
-            otari_web_search is unavailable here whatever this workspace allows.
-            Blocking still takes effect on POST /api/v1/search.
+            otari_web_search is unavailable here. otari_web_fetch remains
```
daavoo: This tells the admin Fetch is available even when it is not. The card reads `config.web_search_configured` only; it never sees `web_fetch_enabled`, which defaults to false. On a deployment with no search backend and Fetch left at its default (a fresh install), the banner promises Fetch while every Fetch request is rejected with the 400 from `WEB_FETCH_NOT_ENABLED_DETAIL`.

Either surface the deployment flag in this endpoint's public shape and condition the sentence on it, or make the wording conditional: "otari_web_fetch remains available if this deployment has enabled it".

src/gateway/api/routes/_tools.py
```diff
@@ -466,4 +485,27 @@ def _build_web_search_backend(
...
+
+def _build_web_search_backend(
+    *,
+    base_url: str | None,
+    tool_entry: dict[str, Any],
+    auth_token: str | None = None,
+    config: GatewayConfig | None = None,
+    tally: ToolUsageTally | None = None,
+) -> WebRetrievalBackend:
+    """Compatibility wrapper for Search-only callers and tests."""
```
daavoo: Nothing in `src/` calls this any more. The only callers are `tests/unit/test_tool_settings_read_sites.py` and a docstring reference in `workspace_web_search_service.py`. Drop it and have those tests call `_build_web_retrieval_backend(search_tool_entry=...)` directly, so the API layer does not carry a shim that exists for tests.

src/gateway/api/routes/_pipeline.py
```diff
@@ -2138,6 +2168,7 @@ def __init__(
         # refilled per attempt.
         cap = self.max_web_search_uses
         self.web_search_budget = WebSearchBudget(cap) if cap is not None else None
+        self.web_retrieval_counter = WebRetrievalCounter()
```
daavoo, line 2217: The two limits directly above and below each other behave differently and nothing says why. `web_search_budget` is refilled per attempt (the comment three lines up says so); this counter is not. On a streaming fallback the second provider re-runs the tool calls, so a request that spent 6 calls on provider A has 4 left on provider B and can hit the cap mid-fallback.

`docs/tools.md` says the allowance spans routing attempts, so this looks deliberate. Worth one line here saying so, since the neighbor sets the opposite expectation.

Separately: `WebRetrievalLimitExceededError` subclasses `MaxToolIterationsExceeded`, so hitting the cap returns 422 and the caller loses the whole response rather than getting a `[tool error]` the model could wrap up from. That is a reasonable choice but `docs/tools.md` does not mention it; the Fetch section should.

src/gateway/api/routes/_pipeline.py
```diff
@@ -396,11 +424,7 @@ def _redacted_upstream_detail(message: str, fallback: str) -> str:
 # either: the two definitions of "settable by a caller" are one definition, and
 # spelling it twice is how they drift.
 _FORWARDED_PARAMS: frozenset[str] = frozenset(
```
daavoo, line 470: Unrelated reflow (and the same at the `_validate_mcp_server_urls` call further down). No lint config changed in this PR, so this looks like a different local Ruff version. Please revert both; they add noise to an already-large diff.

src/gateway/api/routes/_pipeline.py
HareeshBahuleyan: Clarified both counters: neither resets between routing attempts. Search’s `max_uses` counts successful calls; the shared retrieval cap counts attempted calls. Also documented that exceeding the shared cap aborts the loop: HTTP 422 for non-streaming requests, or an error event once streaming has started.

src/gateway/api/routes/_pipeline.py
HareeshBahuleyan: Reverted both unrelated formatting changes to keep the diff focused. The pinned Ruff 0.15.9 reproduces that formatting, so this wasn’t a version mismatch. `make lint` passes with the original layout restored.

web/src/features/tools/WorkspaceWebSearchCard.tsx
HareeshBahuleyan: Changed the banner to say Fetch is available only if the deployment has enabled it and the workspace policy allows it. Updated the component test to assert both conditions.

src/gateway/api/routes/_tools.py
HareeshBahuleyan: Removed `_build_web_search_backend`. The tests now call `_build_web_retrieval_backend(search_tool_entry=...)` directly, and the stale docstring references are updated.

src/gateway/api/routes/_pipeline.py
HareeshBahuleyan: Added in d97f3a323cef948159df24cb28eebfc64109c485: a Search-only fallback when `authorized_tools` is absent, matching the compatibility contract in mozilla-ai/otari-ai#2078. Explicit `null` or malformed values still return 502; Fetch requires explicit authorization. Regression tests cover legacy Search and Fetch denial.

docs/hybrid-mode-protocol.md
HareeshBahuleyan: Updated in d97f3a323cef948159df24cb28eebfc64109c485. The docs now distinguish both compatibility directions: older gateways can send an empty request, and newer gateways treat a missing `authorized_tools` field as Search-only authorization. Fetch remains denied without explicit authorization, and malformed responses return 502.

tests/fixtures/web_access_resolution_contract.json
HareeshBahuleyan: Addressed in d97f3a323cef948159df24cb28eebfc64109c485. Added the legacy Search-only case expecting 200 and updated legacy Fetch to expect 403 instead of 502.

daavoo (approved):
@HareeshBahuleyan tests are failing

PR #985, opened 2026-09-09
feat(dashboard): rebuild the Web search page on an autosaving row grammar

jigjigjig:
## Description

Rebuilds the Web search view of `ToolsGuardrailsPage` to the Paper design
(`Nice grove`, artboards `01 · Web search — BUILD THIS` and `04 · Tool status
panel — opened`, direction 2 approved in review). The same component renders
Code execution and Guardrails, so the row grammar is built once and those two
follow it.

Targets `main` and is self-contained: every file it touches is under `web/`, and
it merges into `main` without conflict. It was opened against
`design-system/storybook` for the Storybook stories the brief asked for, but the
branch was later rebased onto `main` and now carries none of that work, so the
base is `main`. The stories move to #970, which owns Storybook.

## What changed

The page was a three-track grid of rows labeled by config key, each with its own
Save button and a page-level toast. It is now one row grammar:

- **A bounded group per topic** (Backend, Behavior, Search tools, This
  workspace) rather than full-width bands, which run together into one striped
  field when a page stacks several small groups.
- **A row names the setting in prose**, with its config key as a mono caption
  beside it, its help under it, and its control in a lane every row shares.
- **No Save anywhere, and no confirmation.** Text commits on blur and on Enter,
  only when the value changed; a select commits on change. The control disables
  while the write is in flight and that is the whole acknowledgment: a mark on
  every row of a page where everything saves itself is one the reader learns to
  ignore. Only a refusal gets a line, and it keeps the value that caused it.
- **A tool status row heads each service**, opening into two indented rows in
  the same grammar: why the tool is unavailable (with a jump to the URL field),
  and the exact `tools[].type` a client declares, as selectable text beside a
  copy button. This replaces the request-JSON card, whose body repeated the
  manual.
- **The named tools behind `POST /v1/search` drill in** from a row instead of
  occupying a group, since most deployments have none.

Data, endpoints and mutations are untouched: the same one-field PATCH, the same
reachability check, the same `/v1/tools` read.

## New in the design system

| Component | What it is |
| --- | --- |
| `layout/SettingRow` | The row: label + key left, control right. `nested` indents one that belongs to the row above it |
| `layout/SettingsGroup` `bounded` | Frames rows inside the page column; declares `.otari-settings` |
| `navigation/DisclosureRow` | A settings row that opens in place |
| `navigation/DocsLink` | The external "Docs ↗" link, plus a `docsHref` slot on `PageIntro` and `SettingsGroup` |
| `shared/hooks/useAutosave` | Per-control save state, and the Enter-commits rule |

`.otari-settings` is the third dense place after `.otari-toolbar` and
`.otari-pagination`: it declares `--field-height` and the new
`--field-font-size` rather than restyling its descendants, which is what lets a
full-width control on a phone take 16px so iOS does not zoom on focus. No row
picks a height or a font size.

Two shared controls also changed. `actions/CopyField` gained `inFieldCopy`, so
the in-field copy arrangement is reachable without inventing a trailing action.
`actions/CopyButton` gained a 44px hit area through a pseudo-element: it is 32px
at every call site in the product, which is under the floor `motion-and-access.md`
sets, so this fixes an existing violation rather than adding a feature.

Docs updated: `design/layout.md` (the autosaving page recipe, `SettingRow`, the
bounded group, `nested`), `design/forms.md` (validation timing under autosave),
`design/DESIGN.md` (the component registry).

## Two deviations from the brief, called out

1. **Docs links point at `docs/tools.md` on GitHub, not `/docs`.** The brief says
   to anchor them into the bundled guide, but only `docs/dashboard.md` is
   bundled and it has no web-search sections, so every anchor would be dead.
   GitHub is where `DocsPage` already sends its own sibling links; the anchors
   are verified against the file.
2. **A row label is `text-emphasis`, not 13px/500.** `foundation.test.ts`
   refuses a font weight beside a type role, and there is no 13/500 role. The
   label/help pair is 14/550 over 12/subtle instead of 13/500 over 12.

`SaveToast` survives for one card only: `OrganizationGuardrailsCard` still
submits per entry with a Save of its own and has no row to report into. It is a
list of things rather than a set of settings, which is why it keeps that shape.
Neither the Web search nor the Code execution page renders a toast any more.

## Also in here, and separable

- **A layout fix the bounded column exposed.** `.otari-bleed` escapes to the
  width of `<main>` and centers that escape, so the two workspace cards nested
  in the new column drew bands starting 152px left of `<main>`, painting under
  the sidebar. Both are bounded groups now, and `WorkspaceCodeExecutionPolicyCard`
  follows the same autosaving row grammar as the rest of the page.
- **A test-suite race, pre-existing.** `foundation.test.ts` swept `src/**` at
  module load while `architecture.test.ts` had its Biome boundary probe planted
  under `src/features/`, so the two reported the probe as an offender whenever
  they overlapped. `overlaySeams.test.ts` already skipped it; foundation has
  eight sweeps, so the skip now lives in one helper. The base branch runs clean
  (the three test files this PR adds changed the worker scheduling enough to
  surface it).

Worth a reviewer's eye: on the combined `/tools` page a group title is prefixed
with its service (`Web search · Backend`), because three groups called "Backend"
say nothing about which service each configures. On a narrowed page the prefix
is dropped, since the page title already names the service.

## Checks

- `pnpm --dir web run lint`, `typecheck`, `test` (139 files, 4950 tests) clean,
  and the suite run twice to confirm the race above is gone
- Verified in a running gateway in light and dark: 32px dense fields, 12px mono
  for machine values, no control under the 44px row floor, no Save button, no
  save confirmation, both disclosures working from pointer and keyboard, and no
  band escaping the column on any of the three Tools pages
- New screenshot entry for `/tools/web-search`; the narrowed view is no longer
  the combined page with two services hidden
- No generated-artifact drift (`schema.ts`, `routeTree.gen.ts`)

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [x] Refactor
- [x] Documentation
- [ ] Infrastructure / CI

## Relevant issues

No driving issue; this implements the approved Paper direction for the Web search
page. Follow-up filed from the review of this PR: #994 (setting-dense pages still
show too many fields at once).

Not stacked on anything. #970 (the design system layer) owns the Storybook
stories from the same brief and lands separately.

## Checklist

Two boxes are left unchecked deliberately, because this change is dashboard-only
and the literal commands they name do not apply to it:

- **Definition of Done.** `make lint` does not cover `web/`. What ran instead is
  the dashboard counterpart, `pnpm --dir web run lint`, `typecheck` and `test`,
  all clean, with the suite run twice to confirm the test-file race described
  above is gone.
- **OpenAPI.** No API contract changed, and there is no generated-artifact drift
  in `schema.ts` or `routeTree.gen.ts`.

The tests box is checked, but they are the dashboard's own suites rather than
`tests/unit` / `tests/integration`: three new test files, and
`pnpm --dir web run test` clean at 139 files and 4950 tests.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

The visual direction was chosen by a human from Paper artboards before any code
was written, and the result was reviewed against a running gateway in light and
dark. The two deviations from the brief are called out in the description above
rather than silently resolved.

**NOTE:**

## Summary

- Rebuilt Tools settings pages around autosaving rows for Web search, Code execution, Guardrails, and combined Tools.
- Removed Save buttons and success toasts from these views.
- Added shared setting rows, disclosure rows, documentation links, and autosave handling.
- Added inline validation, loading states, error handling, and accessible control associations.
- Improved tool status, configuration, copy controls, and dense-field layouts.
- Updated workspace policy editing and added tests for autosave, validation, concurrency, failed reads, accessibility, and disclosures.

## Technical notes

- Text fields save on blur or Enter. Selects save on change.
- Added reusable `PolicyRow`, `ToolSettingRow`, `ToolStatusGroup`, `useAutosave`, and `usePolicyWriter` utilities.
- Preserved existing endpoints, mutations, reachability checks, and `/v1/tools` reads.
- Successful policy writes now update React Query caches directly.

### DIFF
--- a/web/src/features/tools/SearchToolsCard.tsx
+++ b/web/src/features/tools/SearchToolsCard.tsx
@@ -17,7 +17,6 @@ import { ConfirmButton } from "@/shared/components/actions/ConfirmButton"
 import { ErrorBanner } from "@/shared/components/feedback/ErrorBanner"
 import { errorMessage } from "@/shared/components/feedback/errorMessage"
 import { INPUT_CLASS } from "@/shared/components/forms/inputClass"
-import { Dot } from "@/shared/components/indicators/Dot"
 import { SettingsGroup } from "@/shared/components/layout/SettingsGroup"
 import { DisclosureRow } from "@/shared/components/navigation/DisclosureRow"
 import { FilterSelect } from "@/shared/components/navigation/FilterSelect"
@@ -34,7 +33,7 @@ import { useAutosave } from "@/shared/hooks/useAutosave"
 
 /** The lanes every line in the panel shares, so the columns read down. */
 const NAME_LANE = "w-full shrink-0 font-mono text-xs md:w-[7.5rem]"
-const PROVIDER_LANE = "w-full shrink-0 text-xs text-subtle md:w-[5.5rem]"
+const PROVIDER_LANE = "w-full shrink-0 text-caption text-subtle md:w-[5.5rem]"
 
 // The endpoint a tool with no api_base of its own will actually call, so a blank
 // box reads as "inherits X" rather than as "unconfigured".
@@ -118,7 +117,7 @@ function StoredToolLine({
           isDisabled={busy || !changed}
           onPress={commit}
         >
-          Save
+          {update.isPending ? "Saving…" : "Save"}
         </Button>
         <ConfirmButton
           confirmLabel="Remove permanently"
@@ -129,17 +128,17 @@ function StoredToolLine({
         </ConfirmButton>
       </div>
       {tool.decryptable ? null : (
-        <p className="text-xs text-warning">
+        <p className="text-caption text-warning">
           Key unreadable: check OTARI_SECRET_KEY
         </p>
       )}
       {tool.shadows_config ? (
-        <p className="text-xs text-warning">
+        <p className="text-caption text-warning">
           Overrides the config-file tool of this name
         </p>
       ) : null}
       {save.error ? (
-        <p role="alert" className="break-words text-xs text-danger">
+        <p role="alert" className="break-words text-caption text-danger">
           {save.error}
         </p>
       ) : null}
@@ -154,16 +153,15 @@ function ConfigToolLine({ tool }: { tool: ConfigSearchTool }) {
     <div className="flex flex-col gap-2 px-4 py-3 md:flex-row md:flex-wrap md:items-center">
       <code className={NAME_LANE}>{tool.name}</code>
       <span className={PROVIDER_LANE}>{tool.provider}</span>
-      <span className="min-w-0 flex-1 text-xs text-subtle">
+      <span className="min-w-0 flex-1 text-caption text-subtle">
         {tool.api_base ?? "no api_base declared"} · config file, editable where
         it is defined
       </span>
-      <span className="flex shrink-0 items-center gap-2.5 text-mono-overline text-subtle">
-        <Dot className="bg-text-subtle" />
+      <span className="shrink-0 text-mono-overline text-subtle">
         {tool.has_api_key ? "Key set" : "No key"}
       </span>
       {tool.shadowed ? (
-        <p className="text-xs text-warning">
+        <p className="text-caption text-warning">
           Overridden by the stored tool of this name
         </p>
       ) : null}
@@ -271,12 +269,12 @@ function AddToolForm({ providers }: { providers: SearchProviderInfo[] }) {
           {create.isPending ? "Adding…" : "Add"}
         </Button>
       </div>
-      <p className="text-xs text-subtle">
+      <p className="text-caption text-subtle">
         Storing an API key needs{" "}
         <code className="font-mono">OTARI_SECRET_KEY</code> set on the gateway.
       </p>
       {error ? (
-        <p role="alert" className="break-words text-xs text-danger">
+        <p role="alert" className="break-words text-caption text-danger">
           {error}
         </p>
       ) : null}
@@ -301,6 +299,9 @@ export function SearchToolsCard({ docsHref }: { docsHref: string }) {
   const stored = tools.data?.stored ?? []
   const fromConfig = tools.data?.config ?? []
   const count = stored.length + fromConfig.length
+  // Nothing is known until the read answers, and "0 tools · refuses every
+  // request" is a claim, not a placeholder.
+  const answered = !tools.isLoading
 
   return (
     <SettingsGroup
@@ -312,15 +313,17 @@ export function SearchToolsCard({ docsHref }: { docsHref: string }) {
       <DisclosureRow
         label="Configure search tools"
         help={
-          count === 0
-            ? "None configured, so POST /v1/search refuses every request."
-            : "Callers name one in search_tool_name, or in the /v1/search/{tool} path."
+          !answered
+            ? "Reading the tools this deployment serves."
+            : count === 0
+              ? "None configured, so POST /v1/search refuses every request."
+              : "Callers name one in search_tool_name, or in the /v1/search/{tool} path."
         }
         isOpen={isOpen}
         onToggle={() => setIsOpen((open) => !open)}
         trailing={
           <span className="text-caption text-subtle tabular-nums">
-            {toolCount(count)}
+            {answered ? toolCount(count) : ""}
           </span>
         }
       >
--- a/web/src/features/tools/SearchToolsCard.tsx
+++ b/web/src/features/tools/SearchToolsCard.tsx
@@ -1,5 +1,6 @@
 import { Button } from "@heroui/react"
 import { useEffect, useState } from "react"
+
 import type {
   ConfigSearchTool,
   SearchProviderInfo,
@@ -16,15 +17,24 @@ import { ConfirmButton } from "@/shared/components/actions/ConfirmButton"
 import { ErrorBanner } from "@/shared/components/feedback/ErrorBanner"
 import { errorMessage } from "@/shared/components/feedback/errorMessage"
 import { INPUT_CLASS } from "@/shared/components/forms/inputClass"
-import { Badge } from "@/shared/components/indicators/Badge"
+import { Dot } from "@/shared/components/indicators/Dot"
 import { SettingsGroup } from "@/shared/components/layout/SettingsGroup"
+import { DisclosureRow } from "@/shared/components/navigation/DisclosureRow"
 import { FilterSelect } from "@/shared/components/navigation/FilterSelect"
+import { useAutosave } from "@/shared/hooks/useAutosave"
 
 // Search tools are what POST /v1/search dispatches against. They used to be
 // declarable only in a config file, so a deployment configured entirely through
-// the dashboard could not use that endpoint at all. This card is the route in:
+// the dashboard could not use that endpoint at all. This is the route in:
 // stored tools are editable here, config-file tools are shown read-only so the
 // operator can see every tool a caller could name.
+//
+// A drill-in rather than a fifth group of rows, because it is a list of things
+// rather than a set of settings, and it is empty on most deployments.
+
+/** The lanes every line in the panel shares, so the columns read down. */
+const NAME_LANE = "w-full shrink-0 font-mono text-xs md:w-[7.5rem]"
+const PROVIDER_LANE = "w-full shrink-0 text-xs text-subtle md:w-[5.5rem]"
 
 // The endpoint a tool with no api_base of its own will actually call, so a blank
 // box reads as "inherits X" rather than as "unconfigured".
@@ -37,35 +47,35 @@ function inheritedBase(
   )
 }
 
-function StoredToolRow({
+function toolCount(count: number): string {
+  return `${count} ${count === 1 ? "tool" : "tools"}`
+}
+
+function StoredToolLine({
   tool,
   providers,
-  onSaved,
 }: {
   tool: StoredSearchTool
   providers: SearchProviderInfo[]
-  onSaved: (message: string) => void
 }) {
   const update = useUpdateSearchTool()
   const remove = useDeleteSearchTool()
+  const save = useAutosave()
   const [apiBase, setApiBase] = useState(tool.api_base ?? "")
   // Blank means "keep the stored key". The field is write-only, so it never
-  // shows what is stored, only the last four of it in the badge above.
+  // shows what is stored, only the last four of it beside the row.
   const [apiKey, setApiKey] = useState("")
-  const [error, setError] = useState("")
-
-  useEffect(() => {
-    setApiBase(tool.api_base ?? "")
-  }, [tool.api_base])
+  // Re-synced from the server's answer, so a change made elsewhere lands in the
+  // box rather than leaving a stale draft over it.
+  useEffect(() => setApiBase(tool.api_base ?? ""), [tool.api_base])
 
   const inherited = inheritedBase(providers, tool.provider)
   const changed = apiBase.trim() !== (tool.api_base ?? "") || apiKey !== ""
   const busy = update.isPending || remove.isPending
 
-  const save = () => {
-    setError("")
-    update.mutate(
-      {
+  const commit = () =>
+    void save.run(async () => {
+      await update.mutateAsync({
         name: tool.name,
         body: {
           api_base: apiBase.trim() === "" ? null : apiBase.trim(),
@@ -73,33 +83,15 @@ function StoredToolRow({
           ...(apiKey === "" ? {} : { api_key: apiKey }),
           expected_updated_at: tool.updated_at,
         },
-      },
-      {
-        onSuccess: () => {
-          setApiKey("")
-          onSaved(`${tool.name} saved`)
-        },
-        onError: (err) => setError(errorMessage(err)),
-      },
-    )
-  }
+      })
+      setApiKey("")
+    })
 
   return (
-    <div className="flex flex-col gap-2 py-4">
-      <div className="flex flex-wrap items-center gap-2">
-        <code className="font-mono text-body">{tool.name}</code>
-        <Badge tone="muted">{tool.provider}</Badge>
-        {tool.last4 ? (
-          <Badge tone="muted">{`key ····${tool.last4}`}</Badge>
-        ) : null}
-        {tool.decryptable ? null : (
-          <Badge tone="warn">Key unreadable: check OTARI_SECRET_KEY</Badge>
-        )}
-        {tool.shadows_config ? (
-          <Badge tone="warn">Overrides the config-file tool of this name</Badge>
-        ) : null}
-      </div>
-      <div className="flex flex-wrap items-center gap-2">
+    <div className="flex flex-col gap-2 px-4 py-3">
+      <div className="flex flex-col gap-2 md:flex-row md:flex-wrap md:items-center">
+        <code className={NAME_LANE}>{tool.name}</code>
+        <span className={PROVIDER_LANE}>{tool.provider}</span>
         <input
           type="text"
           inputMode="url"
@@ -108,7 +100,7 @@ function StoredToolRow({
           disabled={busy}
           placeholder={inherited ? `inherits ${inherited}` : "backend URL"}
           onChange={(event) => setApiBase(event.target.value)}
-          className={`w-full sm:w-72 ${INPUT_CLASS}`}
+          className={`field-machine w-full md:w-[15rem] ${INPUT_CLASS}`}
         />
         <input
           type="password"
@@ -116,65 +108,70 @@ function StoredToolRow({
           aria-label={`New API key for ${tool.name}`}
           value={apiKey}
           disabled={busy}
-          placeholder={tool.last4 ? "replace key" : "add key (optional)"}
+          placeholder={tool.last4 ? `replace key ····${tool.last4}` : "add key"}
           onChange={(event) => setApiKey(event.target.value)}
-          className={`w-full sm:w-52 ${INPUT_CLASS}`}
+          className={`field-machine w-full md:w-[10rem] ${INPUT_CLASS}`}
         />
         <Button
           size="sm"
-          variant="primary"
           aria-label={`Save ${tool.name}`}
           isDisabled={busy || !changed}
-          onPress={save}
+          onPress={commit}
         >
-          {update.isPending ? "Saving…" : "Save"}
+          Save
         </Button>
         <ConfirmButton
           confirmLabel="Remove permanently"
           isPending={busy}
-          onConfirm={() => {
-            setError("")
-            remove.mutate(tool.name, {
-              onSuccess: () => onSaved(`${tool.name} removed`),
-              onError: (err) => setError(errorMessage(err)),
-            })
-          }}
+          onConfirm={() => void save.run(() => remove.mutateAsync(tool.name))}
         >
-          Remove tool
+          Remove
         </ConfirmButton>
       </div>
-      {error ? (
-        <span className="break-words text-caption text-danger">{error}</span>
+      {tool.decryptable ? null : (
+        <p className="text-xs text-warning">
+          Key unreadable: check OTARI_SECRET_KEY
+        </p>
+      )}
+      {tool.shadows_config ? (
+        <p className="text-xs text-warning">
+          Overrides the config-file tool of this name
+        </p>
+      ) : null}
+      {save.error ? (
+        <p role="alert" className="break-words text-xs text-danger">
+          {save.error}
+        </p>
       ) : null}
     </div>
   )
 }
 
-function ConfigToolRow({ tool }: { tool: ConfigSearchTool }) {
+// A config-file tool is a fact, not a control: it is editable only where the
+// file is defined, so it reads as a line of values.
+function ConfigToolLine({ tool }: { tool: ConfigSearchTool }) {
   return (
-    <div className="flex flex-wrap items-center gap-2 py-4">
-      <code className="font-mono text-body">{tool.name}</code>
-      <Badge tone="muted">{tool.provider}</Badge>
-      <Badge tone="muted">Config file</Badge>
-      {tool.has_api_key ? <Badge tone="muted">key set</Badge> : null}
-      {tool.shadowed ? (
-        <Badge tone="warn">Overridden by the stored tool of this name</Badge>
-      ) : null}
-      <span className="text-caption">
-        {tool.api_base ?? "no api_base declared"} · editable only where the
-        config file is defined
+    <div className="flex flex-col gap-2 px-4 py-3 md:flex-row md:flex-wrap md:items-center">
+      <code className={NAME_LANE}>{tool.name}</code>
+      <span className={PROVIDER_LANE}>{tool.provider}</span>
+      <span className="min-w-0 flex-1 text-xs text-subtle">
+        {tool.api_base ?? "no api_base declared"} · config file, editable where
+        it is defined
       </span>
+      <span className="flex shrink-0 items-center gap-2.5 text-mono-overline text-subtle">
+        <Dot className="bg-text-subtle" />
+        {tool.has_api_key ? "Key set" : "No key"}
+      </span>
+      {tool.shadowed ? (
+        <p className="text-xs text-warning">
+          Overridden by the stored tool of this name
+        </p>
+      ) : null}
     </div>
   )
 }
 
-function AddToolForm({
-  providers,
-  onSaved,
-}: {
-  providers: SearchProviderInfo[]
-  onSaved: (message: string) => void
-}) {
+function AddToolForm({ providers }: { providers: SearchProviderInfo[] }) {
   const create = useCreateSearchTool()
   const [name, setName] = useState("")
   const [provider, setProvider] = useState(providers[0]?.id ?? "")
@@ -197,10 +194,9 @@ function AddToolForm({
 
   const submit = () => {
     setError("")
-    const created = name.trim()
     create.mutate(
       {
-        name: created,
+        name: name.trim(),
         provider,
         api_base: apiBase.trim() === "" ? null : apiBase.trim(),
         api_key: apiKey === "" ? null : apiKey,
@@ -210,25 +206,23 @@ function AddToolForm({
           setName("")
           setApiBase("")
           setApiKey("")
-          onSaved(`${created} added`)
         },
-        onError: (err) => setError(errorMessage(err)),
+        onError: (cause) => setError(errorMessage(cause)),
       },
     )
   }
 
   return (
-    <div className="flex flex-col gap-2 py-4">
-      <span className="text-body">Add a search tool</span>
-      <div className="flex flex-wrap items-end gap-2">
+    <div className="flex flex-col gap-2 px-4 py-3">
+      <div className="flex flex-col gap-2 md:flex-row md:flex-wrap md:items-center">
         <input
           type="text"
           aria-label="Search tool name"
           value={name}
-          placeholder="name, e.g. local"
+          placeholder="local"
           disabled={create.isPending}
           onChange={(event) => setName(event.target.value)}
-          className={`w-full sm:w-44 ${INPUT_CLASS}`}
+          className={`field-machine w-full md:w-[7.5rem] ${INPUT_CLASS}`}
         />
         <FilterSelect
           ariaLabel="Search provider"
@@ -254,7 +248,7 @@ function AddToolForm({
                 : "backend URL"
           }
           onChange={(event) => setApiBase(event.target.value)}
-          className={`w-full sm:w-72 ${INPUT_CLASS}`}
+          className={`field-machine w-full md:w-[15rem] ${INPUT_CLASS}`}
         />
         <input
           type="password"
@@ -266,7 +260,7 @@ function AddToolForm({
             keyRequired ? "API key (required)" : "API key (optional)"
           }
           onChange={(event) => setApiKey(event.target.value)}
-          className={`w-full sm:w-52 ${INPUT_CLASS}`}
+          className={`field-machine w-full md:w-[10rem] ${INPUT_CLASS}`}
         />
         <Button
           size="sm"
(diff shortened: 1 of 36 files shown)


jigjigjig:
Filed #994 for the field-count reservation from the [review thread](https://mzlai.slack.com/archives/C0BNZGFQ7TJ/p1788950152320059) (Khaled's context-overload point, Javier's foldable-sections suggestion). The row grammar this PR establishes is settled and not in scope there; #994 covers only how a setting-dense page handles field volume, and is blocked on this landing.

jigjigjig (commented):
Reviewed against `web/AGENTS.md`, the frontend-standards skill and its topic guides, `web/design/*`, and `.github/instructions/frontend-standards`. Ten inline findings; five are correctness, five are consistency inside the PR's own new grammar.

Four more that did not earn an anchor:

- `LoadingGroups` (`ToolsGuardrailsPage.tsx:259`) says "at the real row height, so settings arriving does not move the page", but it uses `gap-1.5` where `SettingRow` uses `gap-0.5`, and it never stacks below `md`. Below 768px the skeleton is one line where the real row is two, so the page does move.
- `DisclosureRow`'s `aria-controls` names an id that is not in the DOM while the row is collapsed.
- "acknowledgement" in `design/layout.md` and `ToolsGuardrailsPage.tsx:228`. `AGENTS.md` sets US English, so "acknowledgment".
- The `__boundary_probe__` skip fixes eight readers rather than the one writer. Any sweep added in another test file re-acquires the flake. Writing the probe outside `src/` (or into a fixture the sweeps already exclude by path convention) closes it once.

The row grammar itself is the right call, and `SettingsGroup bounded` plus `.otari-settings` as a place rather than a set of descendant rules is the correct shape for this. The findings below are about what the layer above it still does by hand.

Review by Claude Opus 5 via Claude Code.

web/src/styles/globals.css
```diff
@@ -1323,13 +1323,30 @@ body {
 :root {
   --field-height: 36px;
   --field-padding-block: 6px;
+  /* HeroUI's own two values for `.input`, reproduced here so a dense place can
+     move them. 16px on a phone is not a style choice: iOS zooms the page when a
+     field under 16px takes focus, and the zoom has nowhere to go on a
+     full-width control. Overriding this with a flat 14px is exactly that bug,
+     product-wide, which is what a first cut of this rule did. */
+  --field-font-size: var(--text-base);
```
jigjigjig: The other two dense places take the 16px without taking the height.

`INPUT_CLASS` dropped `text-sm`, so every `.input` in the product now reads `--field-font-size`, which `:root` sets to `--text-base` below `40rem`. `.otari-settings` compensates: its `<48rem` block raises `--field-height` to 36px alongside the 16px. `.otari-toolbar` and `.otari-pagination` (line 1370) do not, and they stay at 32px with `--field-padding-block: 4px`.

So on a phone every toolbar search box and pagination field in the product becomes 16px text in a 32px box. `ModelsPage.tsx:1035` and `SettingsPage.tsx:653` are both `INPUT_CLASS` inside an `.otari-toolbar`.

Either give those two the same `<48rem` treatment `.otari-settings` gets, or pin them to `--text-sm` since neither is a full-width control and neither has the iOS zoom problem the variable exists for.

web/src/features/tools/ToolSettingRows.tsx
```diff
@@ -0,0 +1,456 @@
...
+            aria-label={`Price per call for ${pricingKey}`}
+            aria-invalid={message ? true : undefined}
+            aria-describedby={message ? errorId : undefined}
+            value={draft}
+            disabled={disabled || save.isSaving}
+            placeholder="0.00"
+            onChange={(event) => setDraft(event.target.value)}
+            onKeyDown={commitOnEnter}
+            onBlur={() => {
+              if (!valid || trimmed === committed) return
```
jigjigjig: A cleared or invalid price saves nothing and says nothing.

`valid` requires `trimmed !== ""`, so `if (!valid || trimmed === committed) return` swallows both cases, and `message` only ever carries `loadError || save.error`. Clear the field, tab out, and the row looks saved.

On the base branch the Save button was `isDisabled={... || !changed}`, which was the signal. Removing the button removed it. Every sibling row in this PR reports instead: `NumberRow` at line 155 has "A whole number, 1 or more.", and `parseCeiling` / `ceilingParser` in the two workspace cards both return a message. Give this row one too, and decide what a blank field means (unpriced is a real state the help line names).

web/src/features/tools/ToolSettingRows.tsx
```diff
@@ -0,0 +1,456 @@
...
+  /** Set when the current rate could not be read, so nothing may overwrite it. */
+  loadError?: string
+}) {
+  const committed = configured === null ? "" : String(configured / PER_MILLION)
+  const [draft, setDraft] = useDraft(committed)
+  const save = useAutosave()
+  const errorId = useId()
+
+  const trimmed = draft.trim()
+  const parsed = Number(trimmed)
```
jigjigjig: The money row is the one numeric row that skipped the guard the others document.

`NumberRow` twelve lines up, `parseCeiling` in `WorkspaceWebSearchCard`, and `ceilingParser` in `WorkspaceCodeExecutionPolicyCard` all use a digits-only regex, each with a comment saying `Number` reads `0x10` as 16 and `1e1` as 10. Here it is bare `Number(trimmed)`, so `1e3` is accepted as 1000 and `0x10` as 16. Decimals mean the regex differs, but it should still be an explicit one (`/^\d*\.?\d+$/` or similar) rather than `Number`.

Separately, line 446: `parsed * PER_MILLION` is unrounded, so `0.07` stores `70000.00000000001`. `Math.round` it before it goes on the wire.

web/src/features/tools/WorkspaceWebSearchCard.tsx
```diff
@@ -138,168 +109,133 @@ export function WorkspaceWebSearchCard({
...
+  // rows sitting at "Deployment default" over a workspace that may well have a
+  // stored row, and one blur issues the write that drops it.
+  const unreadable = query.isLoading || query.isError || !config
+  const narrowingDisabled = unreadable || stance === "default"
+
+  // A PUT replaces the row, so one field's save carries the rest of it, and
+  // `provider_options` with them: it is a per-backend bag with no form that
+  // could validate it, and omitting it would silently drop a value set over the
+  // API.
+  const commitField = (patch: Partial<UpdateWorkspaceWebSearchConfigRequest>) =>
```
jigjigjig, line 150: Lost update between two rows of the same card.

`commitField` rebuilds the whole PUT body from the last-read `config`, and `useAutosave` is per control by design, so nothing serializes two rows. Tab out of "Allowed domains" (PUT 1 in flight), type in "Blocked domains", blur before the invalidation lands: PUT 2 carries the pre-PUT-1 `config` and reverts the allowed list.

The previous shape had one Save per group, which made this unreachable. Autosave is what opens it. Two ways out: take the mutation's response as the base for the next body, or hold one save state for the card so a second row's commit queues behind the first. Same code and same reachability in `WorkspaceCodeExecutionPolicyCard.tsx:135`.

web/src/features/tools/SearchToolsCard.tsx
```diff
@@ -277,65 +271,70 @@ function AddToolForm({
...
-        </p>
-      ) : null}
-      {known.length > 0 ? (
-        <AddToolForm providers={known} onSaved={onSaved} />
-      ) : null}
+      <DisclosureRow
+        label="Configure search tools"
+        help={
+          count === 0
+            ? "None configured, so POST /v1/search refuses every request."
```
jigjigjig: The collapsed row asserts this before the read lands.

`count` is `stored.length + fromConfig.length`, which is 0 while `tools.isLoading`, so the row reads "None configured, so POST /v1/search refuses every request" with "0 tools" beside it on every first paint. The old empty state was guarded on `!tools.isLoading` for exactly this. Guard the help line and the count on the query having answered.

web/src/features/tools/SearchToolsCard.tsx
```diff
@@ -108,73 +100,78 @@ function StoredToolRow({
...
         <Button
           size="sm"
-          variant="primary"
           aria-label={`Save ${tool.name}`}
           isDisabled={busy || !changed}
-          onPress={save}
+          onPress={commit}
         >
-          {update.isPending ? "Saving…" : "Save"}
+          Save
```
jigjigjig: This lost its pending label: it was `{update.isPending ? "Saving…" : "Save"}`, and is now always "Save", disabled through `busy`. `AddToolForm` in the same file kept "Adding…", and the `ConfirmButton` beside this one kept `isPending`. This row still submits with a button rather than autosaving, so it still owes the in-flight word.

web/src/features/tools/WorkspaceCodeExecutionPolicyCard.tsx
```diff
@@ -153,251 +129,229 @@ export function WorkspaceCodeExecutionPolicyCard({
   // What the checkboxes cover: what is served, plus whatever stale kinds the
   // policy still names, so unticking one is how an operator retires it.
   const listedTools = [...availableTools, ...staleTools]
+  const storedTools = policy?.tools
+
+  // A PUT replaces the policy, so one field's save carries the rest of it.
+  const commitField = (
```
jigjigjig, line 156: Five hand-rolled copies of the same autosaving text row across this PR: `TextRow` and `NumberRow` (with the `useDraft` effect) in `ToolSettingRows`, `NarrowingRow` in `WorkspaceWebSearchCard`, and `CeilingRow` and `HintRow` here.

They are the same draft/commit/blur/error shape with four different rules: `useDraft`'s `useEffect` rehydrate versus the render-phase `synced` pair used here and in `NarrowingRow`; error precedence `save.error || parsed.error` here versus `loadError || save.error` in `ToolPriceRow`; and trim-compare versus parse-compare on the blur guard. Finding 2 above (the silent no-op on an invalid price) is that drift already producing a bug.

This PR adds `SettingRow` because "spelled by hand the label size, the key caption and the control lane drift between groups". The same argument applies one layer up: an `AutosaveTextRow` taking `committed`, `parse` and `commit` would collapse all five and make the validation rule one decision instead of four.

web/src/shared/components/actions/CopyField.tsx
```diff
@@ -86,15 +87,19 @@ type CopyFieldProps = {
...
       multiline?: false
       /**
-       * A control to sit beside the field, which moves the copy affordance
-       * inside the field and drops the button from the label row. Absent, the
-       * field renders the arrangement it always has.
+       * Put the copy control inside the field and drop the button from the
+       * label row. Right for a value an operator copies and then acts on, where
+       * a separate Copy button competes with the action that follows it.
+       */
+      inFieldCopy: true
```
jigjigjig: `inFieldCopy` is added, documented, tested, and reached by nothing new.

The PR description gives its reason as making "the in-field copy arrangement reachable without inventing a trailing action", but the only production call site is `OrganizationDomainsPage`, mechanically converted from the old rule where passing `action` chose the arrangement, and it still passes an `action`. `DeclarationChip` in `ToolStatusGroup` uses `CopyableValue`, not this.

So the flag adds a required prop to an existing union arm and buys nothing in this diff. Either land it with the call site that needs it, or drop it and keep `action !== undefined` until one exists.

web/src/shared/components/layout/SettingRow.tsx
```diff
@@ -0,0 +1,81 @@
...
+          {configKey ? (
+            <code
+              id={labelId && `${labelId}-key`}
+              className="text-mono-micro text-subtle"
+            >
+              {configKey}
+            </code>
+          ) : null}
+        </div>
+        {help ? <p className="text-xs text-subtle">{help}</p> : null}
```
jigjigjig: Two secondary lines in the same column at two different sizes: `text-xs` (12px) here, `text-caption` (13px) on the error three lines down.

`design/typography.md` gives `text-caption` as "a secondary line under a value", which is what this is, and `DESIGN.md`'s reach-for table names `text-xs` as the thing `text-caption` replaced: "a size spelled inline is a scale nobody can see the whole of". A component whose stated job is stopping label and caption drift between groups is the wrong place to spell one.

Same mix in `DisclosureRow`'s help, `UrlRow`'s note, and the `note` blocks in both workspace cards.

web/src/shared/components/layout/SettingRow.tsx
```diff
@@ -0,0 +1,81 @@
...
+}) {
+  return (
+    <div
+      className={`flex min-h-11 flex-col gap-2.5 py-3 pr-4 md:flex-row md:items-center md:gap-6 ${
+        nested ? "pl-8" : "pl-4"
+      }`}
+    >
+      <div className="flex min-w-0 flex-1 flex-col gap-0.5">
+        <div className="flex flex-wrap items-baseline gap-x-2">
+          <span id={labelId} className="text-emphasis">
```
jigjigjig: A `<span>`, so clicking a row's label does not focus its control, on a page that is nothing but labelled rows. That is a real target on a phone, where the label sits directly above the stacked field.

The id is already there: `settingInputId(field.key)` is on the input in every `ToolSettingRows` case. Make this a `<label htmlFor>` and keep the `aria-labelledby` wiring for the label-plus-key name, which they compose fine (the `for` association gives the click target, `aria-labelledby` still wins for the accessible name).

web/src/features/tools/ToolStatusGroup.tsx
```diff
@@ -0,0 +1,128 @@
...
+  return (
+    <SettingsGroup bounded>
+      <DisclosureRow
+        label={<code className="text-mono-caption">{tool.id}</code>}
+        help={tool.description}
+        isOpen={isOpen}
+        onToggle={() => setIsOpen((open) => !open)}
+        trailing={
+          <span className="flex items-center gap-2.5 text-mono-overline text-subtle">
+            <Dot className={tool.available ? "bg-accent" : "bg-text-subtle"} />
```
jigjigjig: `bg-accent` for the healthy state is off the status vocabulary. The frontend-standards skill and `design/colors.md` name `bg-success` for healthy and reserve the accent for the brand's five jobs, so a green-adjacent brand teal here reads as a status color that is not one and will not follow a retheme of the status family.

Related, in `SearchToolsCard`'s `ConfigToolLine`: that dot is `bg-text-subtle` whether the tool has a key or not, so it distinguishes nothing and the word beside it is carrying the whole meaning. Either color it or drop it.

web/src/styles/globals.css
jigjigjig: Withdrawing this: the rebase onto `main` removed the thing it was about.

The `--field-height` / `--field-font-size` variable system only existed on the storybook branch. This PR now scopes the density with descendant rules (`.otari-settings .input { height: 32px }`) and leaves `INPUT_CLASS` alone, so `.otari-toolbar` and `.otari-pagination` never see the 16px and the phone regression I described cannot happen.

Same for the `__boundary_probe__` note in the review summary: that change is gone too, and `foundation.test.ts` is now only the `WorkspaceCodeExecutionPolicyCard` exemption removal.

web/src/features/tools/WorkspaceCodeExecutionPolicyCard.tsx
jigjigjig: Addressed. `PolicyRow` collapses `NarrowingRow`, `CeilingRow` and `HintRow` into one row with `ceilingParser` and `parsePhrase` shared, and the docstring's reason for keeping `ToolSettingRows` separate (a config key, a reachability note, a Test button, none of which a policy row has) is a fair line to draw.

One thing left, not worth its own thread: `useDraft` in `ToolSettingRows` rehydrates through a `useEffect` while `PolicyRow` does it in render through `synced`. Two idioms for the same job, one per file. Worth settling on the render-phase one next time either is touched.

web/src/features/tools/ToolSettingRows.tsx
jigjigjig: Fixed in 75c6f431. Blank and invalid are now separate answers: an invalid amount gets "An amount in dollars, such as 0.01." in the row, and a cleared field puts the stored value back rather than looking saved. Blank cannot mean unpriced on the wire, since `/v1/pricing` only writes a rate and has no delete, so restoring the value is the honest read of what the field can express. Comment in the code says so.

web/src/features/tools/ToolSettingRows.tsx
jigjigjig: Both fixed in 75c6f431. The guard is `/^\d+(\.\d+)?$/`, explicit rather than `Number`, and it admits the decimal point the other numeric rows do not because this one is money. `Math.round` on the way to the wire, so 0.07 stores 70000. Regression tests cover `1e3`, `0x10`, `abc`, and the rounding.

web/src/features/tools/WorkspaceWebSearchCard.tsx
jigjigjig: Fixed in 75c6f431, and this was the one worth catching. `usePolicyWriter` chains commits and bases each body on the previous write's own response rather than on the query, so the second row carries the first row's value. Both cards use it.

It also surfaced a second thing while fixing the first: spreading the stored row into the body sent `workspace_id`, `configured`, the server-owned `allowed_images` / `available_tools` and two timestamps on every PUT. The writer takes an explicit `toBody` projection now, and a test pins the exact key set.

web/src/features/tools/SearchToolsCard.tsx
jigjigjig: Fixed in 75c6f431. The row reads "Reading the tools this deployment serves." until the query answers, and the count is blank rather than "0 tools".

web/src/features/tools/SearchToolsCard.tsx
jigjigjig: Fixed in 75c6f431 — back to `{update.isPending ? "Saving…" : "Save"}`.

web/src/shared/components/actions/CopyField.tsx
jigjigjig: Agreed, dropped in 75c6f431. `CopyField`, its test and `OrganizationDomainsPage` are back to what main has; `DeclarationChip` uses `CopyableValue`, so nothing in this diff needed the flag.

web/src/shared/components/layout/SettingRow.tsx
jigjigjig: Fixed in 75c6f431. `SettingRow`'s help is `text-caption`, and the same swap went through `DisclosureRow`, `UrlRow`'s note and the note blocks in both workspace cards. One secondary size in the column now. The three `text-xs` left are mono machine values, which is the 12px lane on purpose.

web/src/shared/components/layout/SettingRow.tsx
jigjigjig: Fixed in 75c6f431. `SettingRow` takes `controlId` and renders a real `<label for>`; `ToolSettingRows` passes `settingInputId(field.key)` and `PolicyRow` generates one. `aria-labelledby` still decides the accessible name, so the label-plus-key name is unchanged. Test asserts the association.

web/src/features/tools/ToolStatusGroup.tsx
jigjigjig: Both fixed in 75c6f431. The available dot is `bg-success`. The `ConfigToolLine` dot is gone rather than recoloured: "Key set" / "No key" is not a health state, so the word carries it alone.

web/src/features/tools/WorkspaceCodeExecutionPolicyCard.tsx
jigjigjig: The last note here is done too: `useDraft` in `ToolSettingRows` rehydrates in render now, the same way `PolicyRow` does, so there is one idiom rather than one per file.

jigjigjig:
@coderabbitai full review

Base moved from `design-system/storybook` to `main`. The diff is now this PR's own work only: 31 files, all under `web/`. Please review against `main`.

web/src/features/tools/usePolicyWriter.ts
```diff
@@ -0,0 +1,65 @@
...
+  put: (body: Body) => Promise<Stored>
+}) {
+  const carried = useRef<Stored | undefined>(undefined)
+  const queue = useRef<Promise<unknown>>(Promise.resolve())
+  const [seenKey, setSeenKey] = useState(resetKey)
+
+  if (resetKey !== seenKey) {
+    setSeenKey(resetKey)
+    carried.current = undefined
+  }
```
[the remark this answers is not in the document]
jigjigjig: Fixed in 361b3f09, and the reachability is right: the card rerenders without remounting the writer, so an in-flight write from the old workspace could repopulate the base after the switch.

`commit` now stamps itself with the row it started in and stores the answer only while that still matches, and a `resetKey` change starts a fresh queue rather than making the new row wait behind the old row's write.

Regression test is `usePolicyWriter.test.tsx`, on the workspace-switch path specifically rather than leaning on the existing lost-update one. Mutation-checked it: with the era guard removed that test fails and the other three still pass, so it is testing the guard rather than the happy path.

web/src/features/tools/SearchToolsCard.tsx
```diff
@@ -277,65 +269,75 @@ function AddToolForm({
...
   const providers = useSearchProviders()
+  const [isOpen, setIsOpen] = useState(false)
+
   const known = providers.data ?? []
   const stored = tools.data?.stored ?? []
   const fromConfig = tools.data?.config ?? []
+  const count = stored.length + fromConfig.length
+  // Nothing is known until the read answers, and "0 tools · refuses every
+  // request" is a claim, not a placeholder.
+  const answered = !tools.isLoading
```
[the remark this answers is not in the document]
jigjigjig: Fixed in 361b3f09. `isLoading` going false with no data behind it was being read as an empty deployment, so the row asserted that nothing was configured and that POST /v1/search refuses every request, on a read that never answered. It reports the failed read instead, and the count stays blank. Regression test covers a 500 on the initial GET.

web/src/features/tools/usePolicyWriter.ts
```diff
@@ -0,0 +1,75 @@
...
+  // old row's values and PUT it to the new one.
+  const era = useRef(0)
+  const [seenKey, setSeenKey] = useState(resetKey)
+
+  if (resetKey !== seenKey) {
+    setSeenKey(resetKey)
+    carried.current = undefined
+    era.current += 1
+    // A commit for the new row must not queue behind the old row's write.
+    queue.current = Promise.resolve()
```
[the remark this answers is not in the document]
jigjigjig: Fixed in c97f7d1f, and the trace is right: guarding only the store was too late. A commit queued behind an in-flight write on the old row still ran after the reset, read `carried.current` (by then holding the new row's values) and PUT them through the old row's own request. Declining to store the answer afterwards is exactly why it left no trace.

It bails out before the read and before the send now, and resolves rather than rejects: the operator navigated away, which is not a failed save.

On the queue reset, agreed and noted in the code: replacing `queue.current` is what lets an orphaned callback run beside the new row's chain, so the queue stops being a single serialization point across a reset and the early bail-out is what carries that weight. The comment says so rather than only explaining the why.

Regression test is the sequence you described: A1 in flight, A2 chained behind it, row switch, B write completes, then A1 released. It asserts nothing carrying B's values was ever sent through A's `put`. Mutation-checked: removing the early bail-out fails that test and leaves the other four passing.

web/src/features/tools/SearchToolsCard.tsx
```diff
@@ -108,73 +99,77 @@ function StoredToolRow({
...
-          className={`w-full sm:w-52 ${INPUT_CLASS}`}
+          className={`field-machine w-full md:w-[10rem] ${INPUT_CLASS}`}
         />
         <Button
           size="sm"
-          variant="primary"
           aria-label={`Save ${tool.name}`}
           isDisabled={busy || !changed}
-          onPress={save}
+          onPress={commit}
```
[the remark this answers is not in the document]
jigjigjig: Implemented in c97f7d1f. Fede's call was autosave, so the stored tool row now commits on blur and on Enter and has no Save button of its own.

Scope is the stored row only. `AddToolForm` keeps its explicit create, since that is a new entry rather than an edit to an existing one, and `ConfigToolLine` is untouched because it is already a line of values with nothing to edit.

Two things the conversion had to answer, both worth stating since neither is visible from the diff alone.

Every write carries `expected_updated_at`. Under a Save button that was one write per click; under autosave, leaving the URL and then the key fires two, and the second would still carry the timestamp captured before the first, which optimistic concurrency refuses. The row goes through `usePolicyWriter` now, the same thing the workspace policy groups use: commits chain, and each body takes the fresh `updated_at` from the previous write's own response. Verified against a running gateway, not just in tests: two writes on one row, the second carrying the timestamp the first returned.

The key field is write-only, and blank there means "keep the stored key" rather than "clear it". So an empty blur writes nothing at all and focusing the field and leaving costs nothing; only a typed value goes on the wire. A refused write keeps what was typed so it can be corrected, marks the field `aria-invalid`, and shows the server's reason in the row. On success the field clears and the placeholder becomes `replace key ····abcd`, which is the visible confirmation that the key landed.

Tests cover blur-saves-with-no-Save-button, empty-key-writes-nothing, and the fresh `expected_updated_at` on a second write.

web/src/styles/globals.css
```diff
@@ -1931,6 +1931,41 @@ main {
...
+   zoom has nowhere to go. */
+.otari-settings .input,
+.otari-settings .select__trigger {
+  height: 32px;
+  min-height: 32px;
+  padding-block: 4px;
+}
+@media (width < 48rem) {
+  .otari-settings .input,
+  .otari-settings .select__trigger {
```
khaledosman, line 1957: Settings controls come out 36px below `md`, under the 44px floor. `design/motion-and-access.md:55` is "44px is the floor, everywhere", and :68 says a control under it needs the argument in writing or it is a bug. Both dense places this rule cites as precedent take their controls *up* to 44px at ≤767px (`.otari-pagination` and `.otari-toolbar` further down this file). The comment above argues the 16px font size, not the height.

It is also two classes, so `.otari-settings .select__trigger` outranks the global `[data-slot="button"] { min-height: 2.75rem }` floor: the same `FilterSelect` is 44px in a toolbar and 36px in a settings row.

Raise this block to 44px like its two siblings, or write the exception down the way `DismissChip`'s is.

web/src/features/tools/WorkspaceWebSearchCard.tsx
```diff
@@ -139,168 +109,133 @@ export function WorkspaceWebSearchCard({
...
+    setConfig.mutateAsync({
+      workspaceId: selected.workspace_id,
+      body: {
+        enabled: stance !== "blocked",
+        max_results: config?.max_results ?? null,
```
khaledosman: Autosave over a whole-body PUT can revert a row that just saved. `commitField` fills the body from `config`, which only updates once `useSetWorkspaceWebSearchConfig`'s `onSuccess` invalidation has **refetched** (`shared/api/tools.ts:309`), and each row disables only itself.

Blur "Max results" (PUT #1 in flight) → type into "Prompt hint" → blur before the refetch lands: PUT #2 carries the pre-#1 `max_results` and undoes it. The old single-Save card could not reach this state.

`onSuccess` already receives the fresh config the PUT returned and throws it away. `queryClient.setQueryData([WORKSPACES, workspaceId, "web-search"], data)` there narrows the window to the PUT's own round trip and drops one GET per save (four rows currently cost eight requests). Same shape in `WorkspaceCodeExecutionPolicyCard.tsx:135`.

web/src/features/tools/SearchToolsCard.tsx
khaledosman: A failed or in-flight read reads as "None configured". `count` is 0 whenever `tools.data` is undefined, so this row asserts "None configured, so POST /v1/search refuses every request" during the first load and after an error.

The `ErrorBanner` below is inside the disclosure panel, so on an error nothing visible contradicts it until the operator opens the row. Before this PR the banner sat in the group description and was always visible.

Hoist the banner out of `DisclosureRow`, and gate the "None configured" wording on a settled read (the old code gated its empty state on `!tools.isLoading`).

web/src/features/tools/ToolStatusGroup.tsx
```diff
@@ -0,0 +1,128 @@
...
+                      const input = document.getElementById(
+                        settingInputId(urlFieldKey),
+                      )
+                      input?.scrollIntoView({
+                        behavior: "smooth",
+                        block: "start",
+                      })
+                      input?.focus({ preventScroll: true })
+                    }}
+                    className="text-caption text-link transition-colors duration-150 ease-out hover:text-link-hover focus-visible:otari-focus-ring motion-reduce:transition-none"
```
khaledosman: "Set backend URL ↓" is the only control in this row's lane and it is unpadded inline text, so its target is ~20px on every viewport, under the floor `motion-and-access.md:55` sets. `min-h-11` plus `inline-flex items-center`.

Also `cursor-pointer`: it is a bare `<button>`, so Tailwind's reset leaves it on the default arrow (the frontend-standards instructions call that out as the correct call-site class for a native button, not a finding against one).

web/src/shared/components/layout/SettingsGroup.tsx
```diff
@@ -33,26 +41,65 @@ export function SettingsGroup({
...
+  if (bounded) {
+    return (
+      <section className="flex flex-col gap-3">
+        {heading === null && blurb === null ? null : (
+          <div className="flex flex-col gap-1">
+            {heading}
+            {blurb}
+          </div>
+        )}
+        {/* `otari-settings` is the dense place, the way `otari-toolbar` is:
```
khaledosman, line 81: This comment describes a mechanism the CSS does not use. It says `.otari-settings` "declares `--field-height` and `--field-font-size` for everything inside" and points at globals.css for "why a place declares a variable rather than restyling its descendants" — but the rule sets `height` / `min-height` / `padding-block` / `font-size` directly on HeroUI's `.input` and `.select__trigger`, and globals.css:1252 states outright that HeroUI ships no `--field-height`.

Reword to what the rule actually does, or the next reader goes hunting for a variable to override.

web/src/styles/globals.css
```diff
@@ -1931,6 +1931,41 @@ main {
...
+  }
+  .otari-settings .input {
+    font-size: var(--text-base);
+  }
+}
+/* A field holding a machine value (a URL, an engine list, a domain list, a
+   number) takes the mono family a step down, so a long value fits the lane and
+   reads as data rather than as prose. The rule above hands it 16px back where
+   the field goes full width. */
+.field-machine {
```
khaledosman: `.field-machine` is a per-field, unnamespaced, unlayered class. This file argues the opposite for the same kind of look a few hundred lines up: "Dense is opted into by PLACE, not per field ... no site picks a height." Here five call sites pick a font.

The mono type roles already exist as `@utility text-mono-caption` / `text-mono-micro`, and every escape-hatch rule in this file is `.otari-*` namespaced. If it has to stay unlayered to beat `INPUT_CLASS`'s `text-sm`, say so in the comment; otherwise it is a `@utility` or one of the existing mono roles.

web/src/features/tools/ToolSettingRows.tsx
```diff
@@ -0,0 +1,456 @@
...
+  placeholder: string
+  /** Mono, for a value a machine reads. Off for a sentence a model reads. */
+  machine?: boolean
+}
+
+function useDraft(committed: string) {
+  const [draft, setDraft] = useState(committed)
+  // Re-hydrated from the server's answer, so a save or a clear lands in the
+  // field instead of leaving a stale draft over it.
+  useEffect(() => setDraft(committed), [committed])
```
khaledosman: `useDraft` syncs via effect where the rest of this PR syncs during render. `UrlRow` (line 265 of this file) and `PolicyRow.tsx:59` both use the render-phase compare; this one copies query data into state and reconciles in `useEffect`, which is the "derive, don't duplicate ... instead of copying them into `useState` and syncing with effects" rung (`frontend-standards/typescript-and-react.md:41`). Same shape, one paint later. `SearchToolsCard.tsx:70` has the same effect.

web/src/shared/components/navigation/DisclosureRow.tsx
```diff
@@ -0,0 +1,60 @@
...
+  isOpen: boolean
+  onToggle: () => void
+  children: ReactNode
+}) {
+  const panelId = useId()
+  return (
+    <>
+      <button
+        type="button"
+        aria-expanded={isOpen}
```
khaledosman, line 38: `aria-controls` names an id that does not exist while the row is collapsed, which is what the panel is unmounted at line 57. Render the panel with `hidden` instead of unmounting it, or drop the attribute when closed.

khaledosman (approved):
Verified at head: `typecheck`, `lint`, and the 24 touched test files (2216 tests) pass; no generated-artifact drift owed; the `/tools/web-search` screenshot entry is present.

Eight inline findings, none blocking. The first three are worth acting on before merge: a phone-viewport touch target under the floor, a lost update the autosaving PUT makes reachable, and a failed read that reads as "None configured".

web/src/styles/globals.css
jigjigjig: Fixed in 31341934. Raised to 44px, not written down as an exception: the two places this rule is modelled on both take their controls up at 767px, and a third one quietly not doing so is drift rather than a decision. Same `max-width: 767px` boundary as the `[data-slot="button"]` floor, so a settings row's fields and its buttons arrive together, which also closes the `FilterSelect`-is-44-in-a-toolbar-but-36-here gap you spotted. The 16px stays and the comment now separates the two rules, since only the font size is about iOS zoom.

web/src/features/tools/WorkspaceWebSearchCard.tsx
jigjigjig: Fixed in 31341934, and thank you for the better fix. `onSuccess` now does `setQueryData` with the row the PUT returned, on both cards.

Worth noting the revert you describe was already unreachable at the head you reviewed: `usePolicyWriter` carries the base forward from each write's own response rather than from `config`, so PUT #2 already had #1's values. But your point stands on its own merits and is the better shape: the cache was holding the pre-write row until a GET landed, and every save cost two requests where one would do. Both are now one.

web/src/features/tools/SearchToolsCard.tsx
jigjigjig: Fixed in 31341934. The banner is hoisted out of `DisclosureRow` so a failed read is visible with the row collapsed, which is the state it is nearly always in.

The wording was already gated at the head you reviewed: `answered` is `!isLoading && !error`, and the row reads "Reading the tools this deployment serves." or "Could not read the tools this deployment serves." rather than "None configured", with the count blank in both. That part landed in 361b3f09 with a regression test for the 500. The banner placement is the half that was still wrong.

web/src/features/tools/ToolStatusGroup.tsx
jigjigjig: Fixed in 31341934: `inline-flex min-h-11 items-center` plus `cursor-pointer`, so the only control in that lane carries the floor itself.

web/src/shared/components/layout/SettingsGroup.tsx
jigjigjig: Fixed in 31341934. That comment was describing the `--field-height` / `--field-font-size` mechanism from the design-system branch this work was originally stacked on; the rebase onto main replaced it with descendant rules and the comment did not follow. It now says what the CSS does, and drops the pointer at a variable that does not exist here.

web/src/styles/globals.css
jigjigjig: Fixed in 31341934. Renamed to `.otari-machine-field`, namespaced like every other escape hatch in the file, and the comment now carries both reasons rather than leaving them to be rediscovered: it has to stay unlayered because `INPUT_CLASS` carries `text-sm` and a `@utility` would be settled by Tailwind's emit order rather than by the call site; and it is opted into per field rather than per place on purpose, because the family belongs to the value in the box and not to the row, so a prose hint and a mono domain list legitimately differ inside one row.

The existing mono roles do not fit: `text-mono-caption` is 13px and `text-mono-micro` 11px, where the lane wants 12px, and neither carries the phone-size override.

web/src/features/tools/ToolSettingRows.tsx
jigjigjig: Already fixed at head, so nothing changed here. `useDraft` moved to the render-phase compare in 75c6f431, and `SearchToolsCard`'s `apiBase` sync went the same way in c97f7d1f when that row was converted to autosave. Both now match `UrlRow` and `PolicyRow`, so there is one idiom rather than two. Worth a look at the current head to confirm I have read you right.

web/src/shared/components/navigation/DisclosureRow.tsx
jigjigjig: Fixed in 31341934. The panel stays mounted with `hidden` rather than being unmounted, so `aria-controls` always names an element that exists. Verified in the browser rather than only in jsdom, since a reset can defeat `[hidden]`: collapsed, the panel resolves to `display: none` with its id present. The test asserts visibility now instead of presence, and checks the id resolves while the row is still closed.

PR #995, opened 2026-09-09
fix(dashboard): remove bulk price editing from the models page

jigjigjig:
## Description

The models page let an operator select several models and apply one price to all of them. That write is blunt: it replaces each selected model's price and clears its pricing tiers and 1h cache rate, across as many rows as the current filter matches, since "select all matching" expands past the visible page. Mass-editing prices is not a capability the page should offer, so it is removed.

The multi-select existed only to feed that action, so the checkbox column goes with it. Confirmed before deleting it: nothing else on this page consumed the selection. `useTableSelection`, `resolveSelectedIds` and `BulkActionBar` are shared and stay, because Activity, Budgets and Keys still use them.

Per-model pricing is untouched:

- the inline editor opened by clicking a price cell
- the detail panel editor
- the "Price a model" dialog, for a selector the catalog does not list

### The sticky lane needs no CSS change, and deliberately gets none

`globals.css` has one shared sticky rule that pins `:first-child` for four tables, models among them. On `main` that first child is the checkbox lane, so the models table freezes its checkboxes and lets the model name scroll away, which is the opposite of the comment on that rule and of what its three sibling tables do. Removing the selection column makes `model` the first child, so the pin lands on the identity lane by itself.

That is why this diff touches no CSS, and the absence is intentional rather than an oversight. Adding a `data-key="model"` sticky rule would pin the lane twice and take models back out of step with its siblings.

Measured on the running branch build, with the scroll container narrowed to force horizontal overflow: the model cell computes `position: sticky; left: 0`, holds at x=536 while its neighbor moves from 756 to 551.5 under a 204.5px scroll, and `data-scrolled="true"` lands on the table so the cell's `border-right` goes from 0px to 1px. The lane and its border both retarget with no rule change.

Verified on a local build against a seeded database. With the branch bundle served, the models table renders its rows with zero checkboxes and no bulk bar; the same seed on a `main` bundle renders the same rows with the checkbox column present. Clicking a price cell still opens the inline editor with its Save and Reset price controls.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes mozilla-ai/otari-ai#2096

The issue is filed on `otari-ai`, but the models page lives here and reaches that repo through its `otari` submodule, so the fix belongs in this repo. GitHub does not auto-close across repositories, so that issue needs closing by hand once this ships.

## Checklist

Notes on the boxes above.

Tests: the change is dashboard-only, so the tests are the dashboard's (`pnpm --dir web test`), not `tests/unit` or `tests/integration`. Three tests changed. Removed: "bulk-sets pricing on the selected models", which exercised the capability being removed, and "still offers an operator the selection that feeds it", which asserted the checkbox column exists. Replaced: "offers a non-operator no row selection, and so no bulk pricing" becomes "offers no row selection and no bulk pricing, even to an operator". Two of those are deletions rather than a test cleanup: the behavior they covered is gone on purpose.

That new case is a check, not a guard. Reverting the source change and re-running it fails it on the checkbox assertion, which is the regression it claims to catch. Its last two assertions are weaker guards and the comment on the test says so: with no checkbox there is no way to make a selection, so they would only bite on a bulk bar reintroduced without one.

Documentation: no user-facing doc describes the bulk action, so nothing was left stale. Searched for it rather than assuming.

OpenAPI: no route, schema or docstring changed, so there was nothing to regenerate and no artifact to commit. Ticked to record that it was checked, not skipped.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

The removal was written mechanically and the resulting diff was read line by line before committing. The seam was audited first: a repo-wide search established that the shared selection helpers have other consumers and must not be deleted, and that the second `SetPriceDialog` mount on this page is the "Price a model" dialog rather than a second bulk one, so it stays.

One judgment call worth a reviewer's attention. Removing selection also removes the role gate that made it operator-only, so the security property that gate protected is now satisfied by the capability not existing at all rather than by an authority check. That is a stronger position, but it does mean the two tests that documented the gate are gone, and a future reviewer will not find them.

## Summary

- Removed bulk price editing from the dashboard Models page.
- Removed row selection and the bulk action bar.
- Kept single-model pricing options unchanged.
- Updated tests to confirm that operators cannot select rows or use bulk pricing.

### DIFF
--- a/web/src/features/models/ModelsPage.test.tsx
+++ b/web/src/features/models/ModelsPage.test.tsx
@@ -1061,39 +1061,6 @@ describe("ModelsPage", () => {
     ).not.toBeInTheDocument()
   })
 
-  it("bulk-sets pricing on the selected models", async () => {
-    const fetchMock = mockApi()
-    const user = userEvent.setup()
-
-    renderWithClient(<ModelsPage />)
-    await screen.findByText("openai:gpt-4o")
-
-    const row = tableRow("openai:gpt-4o")
-    await user.click(within(row).getByRole("checkbox"))
-
-    const bar = (await screen.findByText("1 selected")).closest("div")!
-    await user.click(within(bar).getByRole("button", { name: "Set pricing" }))
-
-    const dialog = await screen.findByRole("alertdialog")
-    await user.type(within(dialog).getByLabelText("Input $ / 1M"), "5")
-    await user.type(within(dialog).getByLabelText("Output $ / 1M"), "12")
-    await user.click(within(dialog).getByRole("button", { name: "Set price" }))
-
-    await vi.waitFor(() => {
-      const call = fetchMock.mock.calls.find(
-        ([u, init]) =>
-          String(u).includes("/v1/pricing") &&
-          (init?.method ?? "").toUpperCase() === "POST",
-      )
-      expect(call).toBeTruthy()
-      expect(JSON.parse(String(call![1]!.body))).toMatchObject({
-        model_key: "openai:gpt-4o",
-        input_price_per_million: 5,
-        output_price_per_million: 12,
-      })
-    })
-  })
-
   // -- pricing a model the catalog does not list --------------------------
 
   it("prices the searched selector when nothing matches, seeded from the search box", async () => {
@@ -1404,21 +1371,18 @@ describe("ModelsPage", () => {
     expect(urls.some((url) => url.includes("/v1/models/metadata"))).toBe(false)
   })
 
-  // Selection is the route into bulk pricing, so it is an operator control.
-  // Left ungated, a member could select rows, press "Set pricing" and fire the
-  // bulk write: the price-write class of defect, reached by another door.
-  it("offers a non-operator no row selection, and so no bulk pricing", async () => {
-    mockApi({
-      context: organizationContext({
-        deployment_operator: false,
-        role: "member",
-      }),
-    })
+  // Bulk price editing was removed (otari-ai#2096): mass-editing prices was
+  // never a capability the models page should have offered. Selection existed
+  // only to feed it, so the checkbox column goes with it. Asserted against an
+  // operator, the role that used to have both. The checkbox assertions are the
+  // real check (they fail if selection comes back); the last two are weaker
+  // guards that would only bite on a bulk bar reintroduced without selection,
+  // since with no checkbox there is no way to make a selection here.
+  it("offers no row selection and no bulk pricing, even to an operator", async () => {
+    mockApi()
     renderWithClient(<ModelsPage />)
     await screen.findByText("openai:gpt-4o")
 
-    // No checkbox on the row and none in the header, so there is nothing to
-    // select and the bulk bar has no way to appear.
     const row = tableRow("openai:gpt-4o")
     expect(within(row).queryByRole("checkbox")).toBeNull()
     expect(screen.queryByRole("checkbox")).toBeNull()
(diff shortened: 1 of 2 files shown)


web/src/features/models/ModelsPage.tsx
```diff
@@ -2354,29 +2293,6 @@ export function ModelsPage() {
           onPriceModel={setCustomPriceKey}
         />
```
jigjigjig, line 2295: The block deleted just above here was a role gate: the bulk bar rendered only for an operator, and the comment on it called itself the second of two locks against a member reaching the pricing write.

That gate is gone rather than relaxed. The write it guarded no longer exists on this page, so the property is now satisfied by absence instead of by an authority check. Worth naming explicitly because the diff shows a security control being deleted and does not show what replaced it.

Per-model pricing keeps its own gate: `onEditPricing` is still passed only when `isOperator`, and the detail panel still takes `canEditPricing={isOperator}`.

web/src/features/models/ModelsPage.test.tsx
```diff
@@ -1404,21 +1371,18 @@ describe("ModelsPage", () => {
...
-      }),
-    })
+  // Bulk price editing was removed (otari-ai#2096): mass-editing prices was
+  // never a capability the models page should have offered. Selection existed
+  // only to feed it, so the checkbox column goes with it. Asserted against an
+  // operator, the role that used to have both. The checkbox assertions are the
+  // real check (they fail if selection comes back); the last two are weaker
+  // guards that would only bite on a bulk bar reintroduced without selection,
+  // since with no checkbox there is no way to make a selection here.
+  it("offers no row selection and no bulk pricing, even to an operator", async () => {
```
jigjigjig, line 1377: This case replaces two that asserted the opposite: one that a member gets no checkbox, one that an operator does. Both are deleted on purpose, not tidied away.

It is a check rather than a guard. Restoring the old ModelsPage.tsx and running this file fails it on the checkbox assertion, which is the regression it claims to catch. The two assertions after it are weaker, and the comment in the test says so: with no checkbox there is no way to make a selection, so they only bite on a bulk bar reintroduced without one.

jigjigjig (commented):
Self-review before anyone else reads it. No correctness findings; two notes on decisions a reviewer would reasonably stop at. Checks are green: build, dashboard, e2e, serving, title and template all pass, screenshots and close-stale skipping.

web/src/features/models/ModelsPage.tsx
jigjigjig: Resolving so the PR has nothing outstanding. This note is explanatory and needs no change; it stays readable here for anyone reviewing.

web/src/features/models/ModelsPage.test.tsx
jigjigjig: Resolving so the PR has nothing outstanding. This note is explanatory and needs no change; it stays readable here for anyone reviewing.

jigjigjig:
Skipping the docstring coverage warning deliberately. The two functions it counted are a Vitest `it()` callback and a change inside a React component body, neither of which is public API.

This repository's guidance is to keep docstrings concise and meaningful on public functions and classes, and to add a comment only where the logic is not obvious rather than to restate the code. Adding docstrings here would work against that, so the warning is left as it stands.

khaledosman (approved):

PR #860, opened 2026-08-29
docs: simplify project documentation

njbrake:
## Description

Simplifies Otari's documentation after the recent gateway and dashboard upgrades. This removes 4,461 net lines while keeping the current operational contracts discoverable.

- Replace duplicated API, configuration, provider, and dashboard inventories with shorter guidance and pointers to runtime or generated sources.
- Remove stale migration-era and speculative material, including `docs/routing-scaling.md`.
- Condense repository, backend, frontend, and review guidance around Otari-specific invariants.
- Correct hosted and hybrid behavior, authorization, budget enforcement, Render deployment details, and the `gw_` platform-token prefix.
- Update dashboard documentation tests to protect the new concise structure.

Related context: https://github.com/mozilla-ai/otari-ai/pull/1921

## PR Type

- [ ] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [x] Documentation
- [ ] Infrastructure / CI

## Relevant issues

None.

## Checklist

- [x] I understand the code I am submitting.
- [x] I have added or updated tests that cover my change (`tests/unit`, `tests/integration`).
- [ ] I ran the Definition of Done checks locally (`make lint`, `make typecheck`, `make test`). The full integration suite could not start because this sandbox has no Docker socket or configured PostgreSQL server; details are below.
- [x] Documentation was updated where necessary.
- [x] No API contract changed, so OpenAPI regeneration is not required.

## Validation

- `make lint`
- `make typecheck`
- `pnpm --dir web run lint`
- `pnpm --dir web run typecheck`
- `pnpm --dir web test` (107 files, 1,588 tests)
- `pnpm --dir web run build`
- `uv run pytest -v tests/unit/test_dashboard_doc.py` (7 tests)
- `git diff --check origin/main`
- Local Markdown target and anchor validation across 72 files
- Fenced JSON, YAML, TOML, and `config.example.yml` parsing

`make test` recorded 2,592 passes before 1,974 integration setup errors. Every integration error has the same environment cause: Testcontainers cannot find the Docker socket, and `TEST_DATABASE_URL` is unset. The only branch-owned failures were stale dashboard documentation assertions; those were updated, and both affected Python and dashboard suites pass.

## AI Usage

- [ ] No AI was used.
- [ ] AI was used for drafting/refactoring.
- [x] This is fully AI-generated.

**AI Model/Tool used:** OpenAI Codex

**Any additional AI details you'd like to share:**

[...]

PR #999, opened 2026-09-09
fix(auth): background verification and reset mail sends off the request path

### LINKED ISSUE #720: Awaited inline mail send leaks address validity by response timing
Decide between moving the mail send off the request path and softening the enumeration-safety claim in the docs. Right now the docs promise more than the code delivers.

`create_user_for_signup`, `resend_verification_email`, and `request_password_reset` all `await mailer.send(...)` before returning (`user_service.py:307`, `:383`, `:417`). `Mailer.send` runs `transport.deliver` in a thread that the request coroutine waits on, and `SmtpTransport` allows `SMTP_TIMEOUT_SECONDS = 10` per step.

So the eligible path costs a bcrypt hash, a commit, and a full SMTP round trip, while the ineligible path pays one bcrypt verify and returns. That difference is large enough to walk an organization's roster by timing alone, which is the attack the generic response bodies exist to prevent.

The service docstrings are honest about this ("Not a full equalization"). `docs/access-control.md` is not: it calls signup and reset-request enumeration-safe without qualification.

Either fix works. Backgrounding the send is the stronger one, and it also stops a slow SMTP host from holding a request open for tens of seconds.

Found during review of #708.

AmirF194:
## Description

Backgrounded the send, per your recommendation in the issue.

`create_user_for_signup`, `resend_verification_email`, and `request_password_reset` now take a `background_tasks: BackgroundTasks` (same shape `record_signup` already uses) and hand `mailer.send` to `background_tasks.add_task` after the commit instead of awaiting it. The mailer holds no db handle, so scheduling it once the request's session is otherwise done is safe.

The docs half looks already handled: `docs/access-control.md` no longer makes the enumeration-safe claim. It was trimmed by an unrelated doc simplification (#860) a couple weeks ago, so there's nothing left to soften there.

Added a test per function that intercepts `BackgroundTasks.add_task` and checks the console transport hasn't logged anything by the time the response comes back. A wall-clock assertion would be flaky; this one just checks the send got scheduled instead of run inline. All three are new failures on main, since the mail gets logged during the request there instead of after it.

Ran the full unit suite (2886 passed) and, for integration, the signup and password-reset files directly against a real Postgres rather than the full `tests/integration` directory, since most of it is unrelated to this change. `make lint` and `make typecheck` are both clean.

## PR Type

- [x] Bug Fix

## Relevant issues

Fixes #720

## Checklist

`background_tasks` is a dependency-injected parameter, not part of any request or response schema, so `make openapi-check` and `make postman-check` both pass unchanged (checked).

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

Claude Sonnet 5, via an autonomous coding agent operating in a terminal.

Written from your issue and its stated fix preference.

## Summary

Verification and password-reset emails now send after the request completes. This prevents slow email delivery from delaying responses and reduces response-timing leaks.

## Technical notes

- Added `BackgroundTasks` to signup, verification-resend, and password-reset flows.
- Updated tests to verify that email delivery is scheduled instead of executed during the request.

### DIFF
--- a/src/gateway/api/routes/auth_password_reset.py
+++ b/src/gateway/api/routes/auth_password_reset.py
@@ -9,7 +9,7 @@ the same "the request's own token is the whole proof" shape as
 
 from typing import Annotated
 
-from fastapi import APIRouter, Depends, Request, status
+from fastapi import APIRouter, BackgroundTasks, Depends, Request, status
 from pydantic import BaseModel, Field
 from sqlalchemy.ext.asyncio import AsyncSession
 
@@ -52,13 +52,14 @@ class ResetPasswordRequest(BaseModel):
 async def request_reset(
     body: RequestPasswordResetRequest,
     request: Request,
+    background_tasks: BackgroundTasks,
     db: Annotated[AsyncSession, Depends(get_db)],
     config: Annotated[GatewayConfig, Depends(get_config)],
 ) -> RequestPasswordResetResponse:
     """Mail a password-reset link, or do nothing: the response never says which."""
     throttle_public_auth(request)
     try:
-        await request_password_reset(db, config, email=body.email)
+        await request_password_reset(db, config, background_tasks=background_tasks, email=body.email)
     except MailNotConfiguredError as exc:
         raise mail_unavailable(exc) from None
     return RequestPasswordResetResponse(message=_RESET_REQUEST_MESSAGE)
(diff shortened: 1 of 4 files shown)


AmirF194:
Checking in since this has been open a week: CI is green (lint, typecheck, unit, integration, e2e all passing) and it's just waiting on review.

Issue #879, opened 2026-08-31
[FEATURE] Generic OIDC login for the standalone dashboard


arthuursantos:
> *Refines #564, the follow-up ticket @njbrake asked for: how a self-hosted IdP (Keycloak) fits the Otari design. Drafted by Claude through back-and-forth with @arthuursantos; the decisions are his*

### Problem or use case

In standalone mode the dashboard can authenticate with the shared master key, or with Google or GitHub SSO. It cannot authenticate against a self-hosted OIDC identity provider such as Keycloak.

I run my company on Keycloak. My colleagues therefore either cannot sign in to the dashboard under their company identity at all, or we share the master key, which the master key's own drawbacks make unacceptable for a team.

This is standalone-only. Hybrid mode has no local management API, and auth there is the platform's job.

### Proposed solution

Let an operator configure a single **generic OIDC provider** (Keycloak, Entra, Okta, or any OIDC issuer) by entering its authorization, token, and userinfo endpoints, a client ID and secret, and a display name. A "Sign in with &lt;name&gt;" button then appears on the login screen alongside the existing options. Signing in mints the same **dashboard session** cookie the master-key and password paths already issue, so session TTL and revocation are unchanged.

Access stays governed by the **roster**: only a person a master-key admin has already added by email may sign in. A stranger from the company IdP is refused, not auto-created. The **break-glass master key** remains available for first-run bootstrap and for access when the IdP is unreachable.

LiteLLM's admin-UI SSO ([https://docs.litellm.ai/docs/proxy/admin_ui_sso](https://docs.litellm.ai/docs/proxy/admin_ui_sso)) is the reference for the config surface and scope.

#### User stories

1. As an operator, I want to configure one OIDC provider through the existing config layering (`config.yml` plus env), so that setup matches how I already configure providers and settings.
2. As an operator, I want to supply the authorization, token, and userinfo endpoints manually (copied from my IdP's `.well-known/openid-configuration`), so that dashboard login does not depend on runtime discovery.
3. As an operator, I want to set a display name for the provider, so that the button reads "Sign in with Acme SSO" rather than a generic label.
4. As an operator, I want to optionally set the `issuer`, so that the gateway validates the `iss` on the token response.
5. As an operator, I want the OIDC button to appear only once the provider is fully configured, so that a half-configured provider never shows a broken button.
6. As an operator, I want to register `<base_url>/auth/oidc/callback` as the redirect URI in my IdP, so that the callback lands where the gateway expects.
[...]

PR #1000, opened 2026-09-09
fix(auth): check the OAuth state server-side and turn PKCE on

daavoo:
## What

The dashboard's Google and GitHub sign-in exchanged any authorization code posted to `POST /v1/auth/oauth/{provider}/callback`. The CSRF `state` was compared only in the browser, against a value in `sessionStorage` that no server ever saw, so the gateway could not tell a callback belonging to a flow it started from a forged or replayed one. No PKCE, and no OIDC nonce.

Reported privately by @pepunfold, who audited it before sending. Thank you.

## Why it was off

#765 cleared PKCE on the grounds that authorize and callback are independent requests with no store between them, so a verifier minted at authorize time had nowhere to live.

That was not accurate:

- apron-auth 0.15.1, the version #765 pinned and the one still installed, ships a `StateStore` protocol, generates the verifier and challenge inside `get_authorization_url`, reads them back in `exchange_code`, and defaults `use_pkce` to `True`. The flag was cleared, not absent.
- #751 had merged `webauthn_challenge` the day before: a single-use row for a value minted in one request and consumed by the next, across workers, with a migration and a docstring explaining why a process dictionary will not do.

The hand-built authorization URL followed from the same premise and was not needed either. `get_authorization_url` only emits a challenge when `use_pkce` is true, which #765 had already cleared.

## What this does

**`oauth_pending_state`**, one row per authorization in flight. Written by `/authorize`, consumed by the callback.

- Keyed by the SHA-256 of the state, not the state. This departs from `webauthn_challenge`, which stores its nonce in the clear: nothing is stored *under* that key, while a row here holds the PKCE `code_verifier`. A reader of this table gets digests; the wire carries the preimage.
- Claimed with one conditional `DELETE ... RETURNING`, the rule `webauthn_service.consume_challenge` documents. A read-then-write cannot provide single use.
- Staged on the request's transaction, so an exchange that fails afterwards rolls the claim back and the person can retry.
- Provider is compared after the row is claimed, so a state minted for Google and returned to GitHub's callback is refused as itself. The refusal keeps the request from committing, so the claim rolls back and the Google callback it was minted for still works.
- Swept on write, on an indexed column. `/authorize` is the only thing that grows the table.
- Bound to the browser, not only to the flow. `/authorize` sets an HttpOnly, SameSite=Lax cookie scoped to `/v1/auth/oauth` carrying a random flow secret; the row keeps its SHA-256; the callback refuses without a matching one, before any outbound call (RFC 9700, section 4.7.1). One secret per browser rather than per flow: a second tab reuses the cookie it finds, so neither tab's callback breaks the other.

**The callback requires `state` and the flow cookie** and refuses before any outbound call when either matches no pending row. Never minted, already spent, expired, wrong-provider and wrong-browser all answer identically.

**Two preset defaults are now pinned explicitly.** While the URL was built by hand neither field was read, so both departures were invisible:

- Scopes are held to the exact set in `_PROVIDERS`. Each preset merges its own `BASE_SCOPES` over what it is given, which for Google adds the long-form `userinfo.email` beside the `email` already asked for.
- Google's `access_type=offline` and `prompt=consent` are cleared. The preset's own `extra_params` argument merges *over* those rather than replacing them, so `model_copy` is the only place to drop them.

**The browser keeps its `sessionStorage` check.** It binds a callback to the tab that started the flow, which the server cannot do. The row binds it to a flow this deployment started, which the browser cannot do.

## Notes for review

- **Both providers verify the challenge.** [GitHub shipped PKCE support in July 2025](https://github.blog/changelog/2025-07-14-pkce-support-for-oauth-and-github-app-authentication/), a year before #765.
- **`/authorize` now writes**, and it is unauthenticated. Bounded rather than unbounded: `throttle_public_auth` limits per IP, rows expire in 600s, and every save sweeps. Steady state is capped at roughly request rate x TTL. Worth a second opinion.
- **A slow sign-in can now fail.** Ten minutes from consent screen to callback and the state is gone. New failure mode; 600s matches apron-auth's own default.
- **No OIDC nonce.** apron-auth has no support for it, and PKCE covers the same threat for a confidential client. It belongs with #879, which @pepunfold is taking.
- `test_oauth_live_provider.py` now needs the deployment's database and the `state` from the redirect, because the verifier lives in the row `/authorize` wrote. A verifier the test invented would prove the provider accepts *a* verifier, not the one we would have sent.

## Severity

Medium. Before this PR an attacker needed a victim's redirect URL, code and state together, which the access log and browser history both record, and a window in which the victim's browser had not spent it; the `sessionStorage` check blocks the tab from posting whenever the state is not the one it stored, which leaves that window open for the full TTL. The `sessionStorage` check already blocked naive forced-login. Not the account-takeover-of-any-identity the original report framed, and worth saying plainly if we publish an advisory. With the flow cookie, the redirect URL alone no longer completes a sign-in from anywhere but the browser that started it.

Separately: `SECURITY.md` links to `/security/advisories/new` and private advisories are not enabled on this repo, so our stated primary reporting channel 404s. Not fixed here.

## Checks

`make lint`, mypy over 569 files, 42 unit and 32 integration tests, Biome, `pnpm typecheck`, and the web suite. OpenAPI, Postman and `schema.ts` regenerated; `make openapi-check` and `make postman-check` clean.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Follows up on #765 and #751. Reported privately by @pepunfold.

## Checklist

## AI Usage

- [ ] No AI was used.

## Summary

- Added server-side OAuth state validation for Google and GitHub sign-in.
- Added PKCE and browser-bound, HttpOnly flow cookies.
- Rejected missing, expired, replayed, wrong-provider, and wrong-browser callbacks.
- Updated API schemas, frontend callback handling, documentation, and database migrations.
- Added unit and integration tests for the new OAuth flow.

## Benefits

These changes reduce CSRF risk and prevent unauthorized or replayed OAuth callbacks while preserving safe retries after failed exchanges.

### DIFF
--- a/src/gateway/api/routes/auth_oauth.py
+++ b/src/gateway/api/routes/auth_oauth.py
@@ -22,7 +22,13 @@ compares the returned state against the stored one and, only then, posts the
 code and the state here, where the state is checked again against the row
 ``/authorize`` wrote. Two checks that fail in different directions: the
 browser's binds a callback to the tab that started the flow, and this one binds
-it to a flow this deployment started. See ``gateway.services.oauth_service``.
+it to a flow this deployment started.
+
+**And to the browser that started it.** ``/authorize`` also sets an HttpOnly
+flow cookie whose digest the row keeps, and the callback refuses without it.
+The code and the state share one redirect URL that the access log and browser
+history both record; the cookie is the half of the flow that neither does. See
+``gateway.services.oauth_service``.
 
 **What this route decides, and what it does not.** It proves the person holds
 the provider account. Who that makes them *here* is behind
@@ -35,7 +41,7 @@ import uuid
 from datetime import datetime
 from typing import Annotated
 
-from fastapi import APIRouter, Depends, HTTPException, Path, Request, Response, status
+from fastapi import APIRouter, Cookie, Depends, HTTPException, Path, Request, Response, status
 from pydantic import BaseModel, Field
 from sqlalchemy.exc import SQLAlchemyError
 from sqlalchemy.ext.asyncio import AsyncSession
@@ -57,8 +63,11 @@ from gateway.services.dashboard_session_service import (
 )
 from gateway.services.maintenance_mode_service import is_maintenance_mode
 from gateway.services.oauth_service import (
+    FLOW_COOKIE_NAME,
+    apply_flow_cookie,
     authorization_url,
     exchange_code,
+    flow_secret_for,
     provider_label,
     require_configured,
 )
@@ -81,6 +90,12 @@ _MAX_SUBMITTED_STATE = 512
 # answers at all, so an unknown segment is the framework's own 422 rather than a
 # handler deciding what to do with it. Spelled from the config vocabulary so the
 # two cannot drift.
+# The browser's flow cookie, when it sent one. Bounded for the same reason the
+# state is: it is hashed before anything looks at it.
+FlowCookie = Annotated[
+    str | None, Cookie(alias=FLOW_COOKIE_NAME, max_length=_MAX_SUBMITTED_STATE, include_in_schema=False)
+]
+
 ProviderPath = Annotated[
     str,
     Path(
@@ -104,7 +119,9 @@ class AuthorizeResponse(BaseModel):
             "'state' the provider returns, and send back with the authorization code. A callback "
             "whose state does not match the one held by the browser that started the flow should "
             "be abandoned by the client rather than sent here; one that does is checked again "
-            "against this deployment's own record of it."
+            "against this deployment's own record of it. The response also sets an HttpOnly "
+            "cookie that the callback requires, so the exchange can only be completed from the "
+            "browser this call was made from."
         )
     )
 
@@ -120,7 +137,8 @@ class OAuthCallbackRequest(BaseModel):
     ``state`` is required, and is what binds this callback to an authorization
     request this deployment actually made: it is claimed from
     ``oauth_pending_state`` before the code is sent anywhere, and the row it
-    claims is what carries the PKCE verifier the exchange needs.
+    claims is what carries the PKCE verifier the exchange needs. The flow
+    cookie ``/authorize`` set travels alongside and binds it to the browser.
     """
 
     code: str = Field(
@@ -180,24 +198,29 @@ def require_oauth_provider(
 async def authorize(
     provider: ProviderPath,
     request: Request,
+    response: Response,
     config: Annotated[GatewayConfig, Depends(get_config)],
     db: Annotated[AsyncSession, Depends(get_db)],
+    flow_cookie: FlowCookie = None,
 ) -> AuthorizeResponse:
     """Start an OAuth sign-in: where to send the browser, and the state to keep.
 
     A GET that writes, which is the one thing to know about it. It records the
-    authorization it is about to start (the state's hash, and the PKCE verifier
-    the exchange will need) so the callback has something to check against, and
-    that record is the whole reason the callback can refuse a code this
-    deployment never asked for.
+    authorization it is about to start (the state's hash, the PKCE verifier the
+    exchange will need, and the digest of a flow secret it sets as an HttpOnly
+    cookie) so the callback has something to check against, and that record is
+    the whole reason the callback can refuse a code this deployment never asked
+    for, or one presented from a browser other than the one that asked.
 
     Still safe to repeat: each call mints its own state, and only the one the
     browser kept is the one it sends back. The rows the others leave expire on
-    their own and are swept by the next call.
+    their own and are swept by the next call. The cookie is reused when the
+    browser already holds one, so a second tab does not break the first.
     """
     throttle_public_auth(request)
+    flow_secret = flow_secret_for(flow_cookie)
     try:
-        url, state = await authorization_url(config, provider, db=db)
+        url, state = await authorization_url(config, provider, db=db, flow_secret=flow_secret)
         await db.commit()
     except SQLAlchemyError:
         await db.rollback()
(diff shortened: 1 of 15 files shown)


src/gateway/api/routes/auth_oauth.py
```diff
@@ -171,18 +181,32 @@ async def authorize(
...
+    browser kept is the one it sends back. The rows the others leave expire on
+    their own and are swept by the next call.
     """
     throttle_public_auth(request)
-    state = new_state()
-    return AuthorizeResponse(
-        authorization_url=authorization_url(config, provider, state=state), state=state
-    )
+    try:
+        url, state = await authorization_url(config, provider, db=db)
```
njbrake: `/authorize` sets no cookie, and the callback accepts `code` plus `state` from any client, so a caller holding the provider's redirect query completes the flow and receives the session. Both values travel in one URL: uvicorn's access log records `GET /auth/{provider}/callback?code=...&state=...` verbatim, and browser history holds the same pair. No race is needed, because the dashboard refuses without posting whenever the state is not in `sessionStorage` (redirect landing in another tab, storage cleared, URL reopened), leaving both halves unspent for the row's full 600s.

PKCE does not cover it. The verifier is keyed by the state, so a replayer holding the state gets the gateway to complete the exchange with the stored verifier.

Bind the row to the user agent (RFC 9700 4.7.1): set an HttpOnly, Secure, SameSite=Lax cookie here carrying a random flow secret, store its SHA-256 on the row, require it at the callback. Concurrent tabs sharing one cookie is the wrinkle; a per-provider cookie name, or a row per secret, settles it.

If it stays out of scope, the PR body and any advisory should say the replay vector is unchanged, rather than that a callback is now bound to a flow.

src/gateway/services/oauth_service.py
```diff
@@ -163,50 +156,143 @@ def callback_landing_target(config: GatewayConfig, provider: str, query: str) ->
...
+        self._provider = provider
+
+    async def save(self, state: PendingState) -> None:
+        """Stage one pending authorization, and sweep whatever has expired."""
+        now = datetime.now(UTC)
+        # The sweep rides here rather than on a scheduler because this is the
+        # only write the table takes, so it is the only place that can grow it.
+        # Same reason ``create_dashboard_session`` prunes on the way in.
+        await self._db.execute(delete(OAuthPendingState).where(col(OAuthPendingState.expires_at) <= now))
+        self._db.add(
```
njbrake, line 210: `save` drops `state.metadata` and `consume` returns the default empty dict, so the round trip apron-auth documents (carried through save/consume, surfaced on `TokenSet.context`) silently loses it. Nothing passes metadata today, which is what makes it easy to miss later. Persist it, or raise in `save` on a non-empty value so the next caller finds out at the write.

src/gateway/services/oauth_service.py
```diff
@@ -163,50 +156,143 @@ def callback_landing_target(config: GatewayConfig, provider: str, query: str) ->
...
+        if provider != self._provider:
+            logger.warning("OAuth state minted for %s was returned to %s", provider, self._provider)
+            return None
+        if expires_at <= datetime.now(UTC):
+            return None
+        return PendingState(
+            state=state_key,
+            redirect_uri=redirect_uri,
+            code_verifier=code_verifier,
+            created_at=(expires_at - timedelta(seconds=OAUTH_STATE_TTL_SECONDS)).timestamp(),
```
njbrake: `created_at` is written to the row and never selected, and this recomputes it from `expires_at` minus the current TTL, so rows in flight across a TTL change report a creation time they never had. Add `created_at` to the `RETURNING` list and use it.

njbrake (changes requested):
One blocker: the pending row binds a callback to a flow this deployment started, not to the browser that started it, so the leaked-authorization-response vector from the report survives. Details inline. PKCE end to end, single use, the cross-provider refusal, and the rollback that leaves a state spendable for the person's own retry all hold up, and the tests cover them.

src/gateway/api/routes/auth_oauth.py
daavoo: Done in 33b974e9, the cookie route. `/authorize` sets `otari_oauth_flow` (HttpOnly, Secure mirroring the request scheme, SameSite=Lax, Path=/v1/auth/oauth, max-age 600), the row stores its SHA-256 in a new `flow_hash` column (migration edited in place, it is unreleased), and `exchange_code` refuses a callback with no cookie before touching the database and one with a mismatched cookie after claiming the row, the same way the provider check works. Compared with `hmac.compare_digest`.

Concurrent tabs: one secret per browser rather than per flow. `/authorize` reuses a well-formed cookie it finds and refreshes the max-age, so a second tab does not rotate the first out. Tests cover no cookie, another browser's cookie, two tabs sharing one, and that all three refusals read the same as an unknown state. PR body and severity section rewritten to match.

_Reply drafted by Claude Code (Fable 5.1) on behalf of @daavoo._

src/gateway/services/oauth_service.py
daavoo: Done in 33b974e9: `save` raises `ValueError` on a non-empty `state.metadata` rather than adding a column nobody reads yet.

_Reply drafted by Claude Code (Fable 5.1) on behalf of @daavoo._

src/gateway/services/oauth_service.py
daavoo: Done in 33b974e9: `created_at` is in the `RETURNING` list and is what the returned `PendingState` carries.

_Reply drafted by Claude Code (Fable 5.1) on behalf of @daavoo._

njbrake (approved):
Verified at 33b974e: the replay is refused before any outbound call, another browser's flow cookie is refused, and the browser that started the flow still completes.

PR #1002, opened 2026-09-09
fix(dashboard): claim a deployment for an operator who already has an address

### LINKED ISSUE #992: Claim form on a migrated deployment refuses every address but the hidden right one
On an unclaimed deployment the account page shows "Claim this deployment" and asks for an email and a new password. The form was written for a first-boot operator with no address. A deployment whose operator identity already has one, which is every migrated deployment, refuses any other address with `EmailChangeNotSupportedError` (400), and the detail says an address change is not supported. Nothing on the page says what address it wants.

Seen on the otari-ai development deployment on 9 September: master-key sign-in, then two `PUT /v1/auth/password` calls at 09:03:53Z and 09:04:19Z, both 400, because the tester typed their own address and the operator identity holds a different one.

Fix: when the signed-in identity already has an address, prefill it and make the field read-only, or drop it from the request. Keep the current behaviour only when the identity has none.

Refs #702, mozilla-ai/otari-ai#1716.

_Note: this issue was drafted by Claude Fable 5.1 via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

tbille:
## Description

On a deployment that has not been claimed yet, an operator who came over from an existing tenancy already has a sign-in address. The gateway refuses any other address, but the claim form still asked for one. Nothing on the page said which address it wanted, so the claim failed.

#1214 changed the password card on `main`, so it no longer asks for an address the person already holds. That fixed the refusal. But the card could not tell that this operator's password is the claim. So it showed the words meant for someone who signed in with Google, GitHub or a passkey. The button said "Set a password", and after saving it never said the master key had stopped working as a dashboard login.

The signed-in person's details from `GET /api/v1/organizations/me` now include `claims_deployment`. It is true for the deployment's operator until they set a password, and false for everyone else. The card uses it to say "Claim this deployment" and to name the address the operator will sign in with. After saving, it says the master key no longer signs in. A normal member who sets a first password on an unclaimed deployment still sees the normal words, because their password does not retire the master key.

This PR was rebased onto `main`, and the new change follows #1214's approach. The old commits changed a form that #1214 replaced, so they were dropped.

## How to test it locally

1. Start a fresh standalone gateway and sign in to the dashboard with the master key.
2. Give the operator an address without a password, for example: `UPDATE "user" SET email = 'operator@example.com';`
3. Reload **Account**. The password card should say "Set a password to sign in as operator@example.com from now on", and its button should say **Claim this deployment**.
4. Open the dialog. It should ask for no email. Set a password. The saved line should say the master key no longer signs in.

Automated coverage:
- `tests/unit/test_password_sign_in.py::test_the_membership_context_reports_whether_a_password_claims_the_deployment` (adopted operator: true before the claim, false after it)
- `tests/integration/test_oauth_api.py::test_a_provider_sign_in_reports_that_it_holds_no_password_and_can_set_one` (a member is never the claimant)
- `web/src/features/account/PasswordCard.test.tsx` (claim copy and request body for the adopted operator; normal copy for a member on an unclaimed deployment)
- Storybook: `Dashboard/Account/PasswordCard/UnclaimedWithAddress`

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes #992

## Checklist

- [ ] I ran the Definition of Done checks locally (`make lint`, `make typecheck`, `make test`). `make lint`, `make openapi-check`, `make postman-check`, dashboard lint, typecheck and account tests, and the touched backend unit tests pass locally. `make typecheck` and the integration suite are left to CI.

## AI Usage

- [ ] No AI was used.

**Any additional AI details you'd like to share:** A maintainer chose the scope: rebase onto `main`, follow #1214's pattern, and fix the claim copy in this PR. Claude did the rebase, the implementation and the tests under that direction.

**NOTE:**

## Summary

- Updated the password card for migrated deployments.
- Existing operator email addresses are shown as read-only and omitted from password requests.
- Submission waits for the organization lookup when the address is unknown.
- Failed lookups keep the form usable and show a notice.
- Added deployment-claim status to identity context and updated related UI, stories, and tests.

These changes prevent rejected email changes while preserving the existing flow for identities without an email address.

### DIFF
--- a/web/src/features/account/PasswordCard.test.tsx
+++ b/web/src/features/account/PasswordCard.test.tsx
@@ -69,6 +69,51 @@ function putCall(fetchMock: ReturnType<typeof vi.spyOn>) {
   )
 }
 
+/**
+ * Holds `GET /v1/organizations/me` pending forever, for the loading-window
+ * gap (otari#992): while it holds, `existingEmail` cannot yet distinguish a
+ * migrated identity from a bare one.
+ */
+function mockOrganizationContextPending() {
+  return vi.spyOn(globalThis, "fetch").mockImplementation(async (input) => {
+    const url = String(input)
+    if (url === "/v1/organizations/me") {
+      return new Promise<Response>(() => {})
+    }
+    return jsonResponse({ count: 0, data: [] })
+  })
+}
+
+/**
+ * Fails `GET /v1/organizations/me`, everything else answers normally. 403,
+ * not 500: the app's `QueryClient` retries a failed query up to twice by
+ * default and only skips that for 401/403, and a status this test would
+ * retry through is a status this test would also have to wait through.
+ */
+function mockOrganizationContextFailing(put: unknown = CLAIMED) {
+  return vi.spyOn(globalThis, "fetch").mockImplementation(async (input) => {
+    const url = String(input)
+    if (url === "/v1/organizations/me") {
+      return jsonResponse({ detail: "boom" }, 403)
+    }
+    if (url === "/v1/auth/password") {
+      return jsonResponse(put)
+    }
+    return jsonResponse({ count: 0, data: [] })
+  })
+}
+
+/**
+ * Wait past the organization-context query the claim form gates on: the
+ * email field is disabled while this identity's address is still unknown
+ * (otari#992's loading-window gap), so a test that types and submits before
+ * it settles would exercise a state real typing speed never reaches rather
+ * than the form itself.
+ */
+async function identityResolved() {
+  await waitFor(() => expect(screen.getByLabelText("Email")).not.toBeDisabled())
+}
+
 describe("PasswordCard on an unclaimed deployment", () => {
   afterEach(() => {
     vi.restoreAllMocks()
@@ -88,6 +133,7 @@ describe("PasswordCard on an unclaimed deployment", () => {
     const user = userEvent.setup()
     renderCard(["master_key"])
 
+    await identityResolved()
     await user.type(screen.getByLabelText("Email"), "operator@example.com")
     await user.type(screen.getByLabelText("New password"), "a-real-password")
     await user.type(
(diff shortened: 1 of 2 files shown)


tbille:
Ran this diff through a 3-round star-chamber debate review (openai/gpt-5.6, anthropic/claude-opus-5, gemini/gemini-3.6-flash).

**Converged, and fixed in 85383d4:** the migrated-identity guard only took effect after `GET /v1/organizations/me` resolved — during the loading window (or permanently, on a failed fetch) `existingEmail` was indistinguishable from "resolved, no address," so the card fell back to the old unlocked field and an operator could submit a mismatched address, reproducing the exact `EmailChangeNotSupportedError` this PR exists to remove. Fixed by gating submission on `organization.isPending` only — deliberately *not* on `isError`, since blocking on a failed fetch too would make an unreachable `/v1/organizations/me` turn an unclaimed deployment permanently unclaimable through its only UI. Added tests for both the pending and the failed-query states.

**Converged, left as tracked follow-up (not fixed here):**
- Client-side validation messages (`problem`, `unchanged`) aren't wired to their fields via `isInvalid`/`aria-describedby`, unlike the advisory copy elsewhere in the same file.
- `isClaimed` is a deployment-wide fact applied to a per-identity decision; the existing doc comment names one direction of the resulting mismatch (a member on an unclaimed deployment sees an unsatisfiable claim form) but not the symmetric one (an OAuth/passkey-only member on a *claimed* deployment sees a change form requiring a password they don't have). One reviewer's suggested real fix: publish `has_password` on `caller` in `GET /v1/organizations/me`, which would collapse this and the loading-window issue onto one authoritative source.
- Test maintainability nits: exact-string JSON body assertions (order-coupled) instead of `toEqual`/`not.toHaveProperty`; the shared fetch mock's catch-all quietly answers unrecognized URLs instead of failing loudly.

Happy to file the above as separate follow-up issues if useful — didn't want to scope-creep this PR further.

peteski22 (approved):

PR #1003, opened 2026-09-09
fix(dashboard): align trailing actions to the input line in routing control rows

jigjigjig:
## Description

The condition row on Routing, Policy showed its threshold field sitting a line above the model dropdown beside it. The row is laid out `items-end`, which bottom-aligns each child's whole box, and a field's box includes the caption line it reserves under its input. A control that reserves no line therefore lands with its input level with its neighbor's caption instead of with its neighbor's input: 19px of reserve plus the row's 4px gap, so 23px.

`FieldAction` gives a trailing action the same reserve a field has, which puts the input lines on one line. It reserves rather than nudging: no `self-*`, no padding, and the height reads the caption role's own line-height variable, so a retune of the caption carries the action with it.

Applied to the Remove button on all four control rows on this page, and to the safe-choice radio label, which carried a hand-tuned `pb-2`.

That `pb-2` was not load-bearing spacing, which is worth saying because a deleted padding usually is. Measured on `main`, the radio label already sat level with the model input and the Remove button at 804, while the Share field sat alone at 781. So the 8px was compensating for a missing reserve by matching the two children that had none, rather than reaching for the 23px the fields hold. With the rung in place all four children agree.

Separately, and this is a second fix rather than part of the alignment one: `ModelComboBox` now routes its hint through `FieldMessages`. Under `FieldAction` the rows align whether or not the combo box reserves, so this is not load-bearing for alignment. It is here because without it the row jumps 23px when discovery resolves and "Loading models from your providers…" disappears, which is exactly what the reserve exists to prevent.

`forms.md` gains the rule none of it stated: the trailing caption rung is a property of the flex line, not of the control. That scopes the two existing statements rather than declaring either wrong, and it gives the right answer at any child count.

### Three things a reviewer should know before reading the diff

**This changes the appearance of two rows nobody reported.** The pool row and the guardrails row were both misaligned before this change. Fixing only the reported row would have left a known misalignment in place for a reason no reviewer could see, so they ride along.

**One of the five rows was never rendered, by anyone.** `+ Add guardrails` reports `disabled` on both environments here, because no guardrails service is configured, so that row exists only on a deployment that has one. Everything said about it, including the change in this PR, is read from markup. It ships changed and unobserved. The other four were measured on a running build.

**Uniform trailing reserve holds only while every message is one line, and per flex line rather than per row.** These rows are `flex-wrap`, and `items-end` aligns each flex line to its own cross-end, so after a wrap the action aligns within whatever line it lands on. A message long enough to wrap makes its field 38px and drops it back out of line, and no reserve can answer that. Both limits are in the component's docstring. The two longest captions in the app, 70 and 85 characters, are on the guardrails row, so the assumption is untested exactly where it is most likely to fail.

This branch points at `5ed641ca` while the PR's base reads `25217c71`, one commit ahead. That commit is a dependency bump: `git diff --stat 5ed641ca..origin/main -- web/` is empty, so nothing under `web/` moved and the measurements below describe current `main`. No rebase is owed.

The built CSS hash does not move across this change, `pb-2` leaving a call site included: the class is still emitted, so another consumer in the tree holds it.

### Measured

Control-edge bottoms, same seeded policy, `main` build against this branch's build. Spread is the distance between the highest and lowest control edge in the row.

| Row | Children | Before | After |
| --- | --- | --- | --- |
| Condition (tier down) | threshold, model, Remove | spread 23 | spread 0 |
| Pool, weighted, model 1 | model, share, radio, Remove | spread 23 | spread 0 |
| Pool, weighted, model 2 | model, share, radio, Remove | spread 23 | spread 0 |
| Fallback chain | model, Remove | spread 0 | spread 0 |
| Guardrails | profile, mode, on-unavailable, Remove | not rendered | not rendered |

The fallback row is the regression check. It is correct today, an earlier draft of this fix broke it, and it is still correct here. That row has one reserving control and one button, so it is the case where "align to the row's cross-end" and "align to the input line" give different answers from the same rule.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes mozilla-ai/otari-ai#2091

The issue is filed on `otari-ai`, but the routing page lives here and reaches that repo through its `otari` submodule. GitHub does not auto-close across repositories, so that issue needs closing by hand once this ships.

The issue body is one line plus two screenshots that are not readable from here, so this matches the described symptom on the named section rather than the pictures.

## Checklist

Notes on the boxes.

Tests: dashboard-only, so `pnpm --dir web test`, not `tests/unit` or `tests/integration`. Two new files, four cases, and all four are checks rather than guards: removing the reserve from `FieldAction` fails its two on `expected null not to be null` and on the wrapper having one child instead of two, and reverting `ModelComboBox` fails its two, one because no element renders at all when the hint is empty and one because the element renders with the caption role but without the reserve class. I removed each fix and read the failures rather than inferring them.

What those tests do not cover, said plainly: jsdom does no layout, so nothing here asserts the alignment itself, and a regression that unwrapped a call site would pass all four. The alignment claim rests on the measured table above, not on the suite.

Documentation: `web/design/forms.md` gains a Control rows section. It is the file the whole finding is about, since nothing in the design system specified cross-axis alignment for a control row.

OpenAPI: unticked because nothing changed. No route, schema or docstring moved, so there was no spec to regenerate and no artifact to commit.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

The first version of this fix reserved a line on the combo box alone. It aligned the reported row and broke the fallback row, which is correct today. That was caught by measuring all four rows on a running build rather than the one in the issue, and it is why the fix is a named component applied at five call sites instead of one local change.

Two things worth a reviewer's attention. `FieldAction` passes `null` as the child of `FieldMessages` because that prop is required; widening the signature for one caller seemed worse than one explicit `null`, and there is a comment saying so. And `text-xs` on the radio label is left alone: it is off the type scale and that is a real problem, but it is a type offense rather than an alignment one and it does not belong in this PR.

## Summary

- Added consistent spacing for routing action buttons and the safe-choice option.
- Updated `ModelComboBox` to keep rows stable as hints appear or disappear.
- Documented the spacing rule and added tests for alignment and hint behavior.

These changes prevent visible alignment shifts and improve form consistency.

### DIFF
--- a/web/src/features/models/ModelComboBox.tsx
+++ b/web/src/features/models/ModelComboBox.tsx
@@ -2,6 +2,7 @@ import { ComboBox, Input, Label, ListBox, ListBoxItem } from "@heroui/react"
 import { type ReactNode, useMemo } from "react"
 import type { DiscoverableModel } from "@/client"
 import { useDiscoverableModels } from "@/shared/api/models"
+import { FieldMessages } from "@/shared/components/forms/FieldMessages"
 
 // How many matches to render at once. A single provider can report a few hundred
 // models, and past this the popover is a wall of text nobody scrolls; typing one
@@ -126,7 +127,10 @@ export function ModelComboBox({
           )}
         </ListBox>
       </ComboBox.Popover>
-      {hint ? <span className="text-caption">{hint}</span> : null}
+      {/* Reserved even when silent, so this control matches a `Field` beside it
+          in a row: `Field` always renders its own reserve, and a bare hint here
+          left the two roots a caption line apart under `items-end`. */}
+      <FieldMessages>{hint}</FieldMessages>
     </ComboBox.Root>
   )
 }
--- a/web/src/features/models/ModelComboBox.tsx
+++ b/web/src/features/models/ModelComboBox.tsx
@@ -1,4 +1,11 @@
-import { ComboBox, Input, Label, ListBox, ListBoxItem } from "@heroui/react"
+import {
+  ComboBox,
+  Description,
+  Input,
+  Label,
+  ListBox,
+  ListBoxItem,
+} from "@heroui/react"
 import { type ReactNode, useMemo } from "react"
 import type { DiscoverableModel } from "@/client"
 import { useDiscoverableModels } from "@/shared/api/models"
@@ -128,9 +135,13 @@ export function ModelComboBox({
         </ListBox>
       </ComboBox.Popover>
       {/* Reserved even when silent, so this control matches a `Field` beside it
-          in a row: `Field` always renders its own reserve, and a bare hint here
-          left the two roots a caption line apart under `items-end`. */}
-      <FieldMessages>{hint}</FieldMessages>
+          in a row. The hint goes through HeroUI's `Description`, which is what
+          wires it to the input via aria-describedby; a bare node here leaves
+          the combo box reporting `aria-describedby: null`. Same reasoning as
+          `Field`. */}
+      <FieldMessages>
+        {hint ? <Description>{hint}</Description> : null}
+      </FieldMessages>
     </ComboBox.Root>
   )
 }
--- a/web/src/features/routing/RoutingPage.tsx
+++ b/web/src/features/routing/RoutingPage.tsx
@@ -898,7 +898,7 @@ function PolicyForm({
                   </div>
                 ) : null}
                 <FieldAction>
-                  <label className="flex items-center gap-2 text-body text-foreground">
+                  <label className="flex items-center gap-2 text-body">
                     <input
                       type="radio"
                       name="router-safe-choice"
--- a/web/src/features/routing/RoutingPage.tsx
+++ b/web/src/features/routing/RoutingPage.tsx
@@ -39,6 +39,7 @@ import {
 import { EmptyState } from "@/shared/components/feedback/EmptyState"
 import { ErrorBanner } from "@/shared/components/feedback/ErrorBanner"
 import { Field } from "@/shared/components/forms/Field"
+import { FieldAction } from "@/shared/components/forms/FieldAction"
 import { ControlField } from "@/shared/components/forms/FieldMessages"
 import { Dot } from "@/shared/components/indicators/Dot"
 import { PageIntro } from "@/shared/components/layout/PageIntro"
@@ -822,14 +823,18 @@ function PolicyForm({
                     isRequired
                   />
                 </div>
-                <Button
-                  variant="ghost"
-                  onPress={() =>
-                    setConditions((prev) => prev.filter((_, i) => i !== index))
-                  }
-                >
-                  Remove
-                </Button>
+                <FieldAction>
+                  <Button
+                    variant="ghost"
+                    onPress={() =>
+                      setConditions((prev) =>
+                        prev.filter((_, i) => i !== index),
+                      )
+                    }
+                  >
+                    Remove
+                  </Button>
+                </FieldAction>
               </div>
             ))}
           </div>
@@ -892,29 +897,35 @@ function PolicyForm({
                     />
                   </div>
                 ) : null}
-                <label className="flex items-center gap-2 pb-2 text-xs text-foreground">
-                  <input
-                    type="radio"
-                    name="router-safe-choice"
-                    checked={safeIndex === index}
-                    onChange={() => setSafeIndex(index)}
-                  />
-                  {weighted ? "Serves on opt-out" : "Serves when unsure"}
-                </label>
-                <Button
-                  variant="ghost"
-                  onPress={() => {
-                    setCandidates((prev) => prev.filter((_, i) => i !== index))
-                    setWeights((prev) => prev.filter((_, i) => i !== index))
-                    // Keep the mark on the same model where possible; if the marked
-                    // one went, fall back to the first, never to nothing.
-                    setSafeIndex((prev) =>
-                      index < prev ? prev - 1 : index === prev ? 0 : prev,
-                    )
-                  }}
-                >
-                  Remove
-                </Button>
+                <FieldAction>
+                  <label className="flex items-center gap-2 text-xs text-foreground">
+                    <input
+                      type="radio"
+                      name="router-safe-choice"
+                      checked={safeIndex === index}
+                      onChange={() => setSafeIndex(index)}
+                    />
+                    {weighted ? "Serves on opt-out" : "Serves when unsure"}
+                  </label>
+                </FieldAction>
+                <FieldAction>
+                  <Button
+                    variant="ghost"
+                    onPress={() => {
+                      setCandidates((prev) =>
+                        prev.filter((_, i) => i !== index),
+                      )
+                      setWeights((prev) => prev.filter((_, i) => i !== index))
+                      // Keep the mark on the same model where possible; if the marked
+                      // one went, fall back to the first, never to nothing.
+                      setSafeIndex((prev) =>
+                        index < prev ? prev - 1 : index === prev ? 0 : prev,
+                      )
+                    }}
+                  >
+                    Remove
+                  </Button>
+                </FieldAction>
               </div>
             ))}
             <p className="text-caption">
@@ -1011,14 +1022,16 @@ function PolicyForm({
                     isRequired
                   />
                 </div>
-                <Button
-                  variant="ghost"
-                  onPress={() =>
-                    setChain((prev) => prev.filter((_, i) => i !== index))
-                  }
-                >
-                  Remove
-                </Button>
+                <FieldAction>
+                  <Button
+                    variant="ghost"
+                    onPress={() =>
+                      setChain((prev) => prev.filter((_, i) => i !== index))
+                    }
+                  >
+                    Remove
+                  </Button>
+                </FieldAction>
               </div>
             ))}
             <div className="flex flex-wrap items-baseline gap-2">
@@ -1103,16 +1116,18 @@ function PolicyForm({
                     }
                     hint="block fails closed, so a guardrails outage refuses every request through this policy."
                   />
-                  <Button
-                    variant="ghost"
-                    onPress={() =>
-                      setGuardrails((prev) =>
-                        prev.filter((_, i) => i !== index),
-                      )
-                    }
-                  >
-                    Remove
-                  </Button>
+                  <FieldAction>
+                    <Button
+                      variant="ghost"
+                      onPress={() =>
+                        setGuardrails((prev) =>
+                          prev.filter((_, i) => i !== index),
+                        )
+                      }
+                    >
+                      Remove
+                    </Button>
+                  </FieldAction>
                 </div>
                 {guardrail.mode === "block" &&
                 (guardrail.on_unavailable ?? "block") === "block" ? (
--- a/web/src/features/routing/RoutingPage.tsx
+++ b/web/src/features/routing/RoutingPage.tsx
@@ -898,7 +898,7 @@ function PolicyForm({
                   </div>
                 ) : null}
                 <FieldAction>
-                  <label className="flex items-center gap-2 text-xs text-foreground">
+                  <label className="flex items-center gap-2 text-body text-foreground">
                     <input
                       type="radio"
                       name="router-safe-choice"
--- a/web/src/shared/components/forms/FieldAction.test.tsx
+++ b/web/src/shared/components/forms/FieldAction.test.tsx
@@ -17,8 +17,11 @@ import { FieldAction } from "./FieldAction"
  * caption role rather than from a pixel.
  */
 describe("FieldAction", () => {
+  // Structural rather than by class: a token gets renamed, and a test that
+  // breaks on a restyle teaches everyone to stop trusting the suite.
   const reserve = (container: HTMLElement) =>
-    container.querySelector(".text-caption") as HTMLElement | null
+    (container.firstElementChild as HTMLElement)
+      .lastElementChild as HTMLElement | null
 
   it("holds a caption line under its child, so the action meets the input line", () => {
     const { container } = render(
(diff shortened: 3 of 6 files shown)


web/src/features/routing/RoutingPage.tsx
```diff
@@ -892,29 +897,35 @@ function PolicyForm({
...
-                    // one went, fall back to the first, never to nothing.
-                    setSafeIndex((prev) =>
-                      index < prev ? prev - 1 : index === prev ? 0 : prev,
-                    )
-                  }}
-                >
-                  Remove
-                </Button>
+                <FieldAction>
+                  <label className="flex items-center gap-2 text-xs text-foreground">
```
jigjigjig: Two things about this one child, both deliberate.

The `pb-2` is gone. It was 8px of hand-tuned padding standing in for the 23px the fields hold, so this label was already misaligned before, just less visibly than the others. `FieldAction` gives it the real rung, and measuring after the change puts its box bottom on the same 761 as the model input and the Remove button.

The `text-xs` stays. It is off the type scale and it should be `text-caption`, but that is a type offense rather than an alignment one, and changing it here would put a visible type change into a PR about geometry. Worth its own issue.

The same `pb-2` idiom exists at `SettingsPage.tsx:681`. Left alone on purpose: different page, different issue, and it says there is a pass to do.

web/src/shared/components/forms/FieldAction.tsx
```diff
@@ -0,0 +1,37 @@
...
+ * whatever line it lands on, which after a wrap may hold nothing else. Correct
+ * per line, not per row.
+ */
+export function FieldAction({ children }: { children: ReactNode }) {
+  return (
+    <div className="flex flex-col gap-1">
+      {children}
+      {/* `children` is required on FieldMessages; an empty reserve passes null
+          rather than widening that signature for this one caller. */}
+      <FieldMessages>{null}</FieldMessages>
```
jigjigjig, line 36: `FieldMessages` declares `children: ReactNode` as required, so `<FieldMessages />` does not typecheck. The options were to pass an explicit `null` here or to make that prop optional on a shared component for the sake of one caller. I took the local one.

If a second caller ever wants an empty reserve, making the prop optional there is the better answer and this comment should go with it.

web/src/features/models/ModelComboBox.tsx
```diff
@@ -126,7 +127,10 @@ export function ModelComboBox({
           )}
         </ListBox>
       </ComboBox.Popover>
-      {hint ? <span className="text-caption">{hint}</span> : null}
+      {/* Reserved even when silent, so this control matches a `Field` beside it
+          in a row: `Field` always renders its own reserve, and a bare hint here
+          left the two roots a caption line apart under `items-end`. */}
+      <FieldMessages>{hint}</FieldMessages>
```
jigjigjig: This is the second fix in the PR and it is not the alignment one. Under `FieldAction` the rows align whether or not this control reserves, so nothing here is load-bearing for the reported bug.

It is here because `hint` is "Loading models from your providers…" while discovery is in flight and resolves to nothing on these rows, so without a reserve the row jumped 23px on load. That is the same motion the reserve exists to prevent, in the same component, found while measuring the fix for the other thing.

Said separately in the PR body for the same reason: folding it into the alignment claim would overstate what the alignment change does.

web/src/shared/components/forms/FieldAction.test.tsx
```diff
@@ -0,0 +1,52 @@
...
+ * bare button's does not, which puts the button a caption line below the input
+ * it acts on. This holds the same line under the action so the two input lines
+ * agree.
+ *
+ * jsdom does no layout, so the reserve is asserted as the class that carries
+ * it, the same limit `FieldMessages.test.tsx` works under. What a test can say
+ * here is that the reserve is present, empty, and reads its height from the
+ * caption role rather than from a pixel.
+ */
+describe("FieldAction", () => {
```
jigjigjig, line 19: What this file and `ModelComboBox.test.tsx` do not cover, so nobody reads four green cases as covering the bug.

jsdom does no layout, so none of these four asserts the alignment itself. They assert the reserve is rendered, empty, and sized from the caption role's variable. A regression that unwrapped a call site in `RoutingPage.tsx` would leave all four passing.

They are checks rather than guards: removing the reserve from `FieldAction` fails both of these, and reverting `ModelComboBox` fails both of its own, on two different messages. I read each failure rather than inferring it. The alignment claim itself rests on the measured table in the PR body.

jigjigjig (commented):
Self-review before anyone else reads it. No correctness findings. Three notes on decisions a reviewer would stop at, and one on what the tests do not cover.

web/src/features/routing/RoutingPage.tsx
jigjigjig: Resolving so the PR has nothing outstanding. This note is explanatory and needs no change; it stays readable here for anyone reviewing.

web/src/shared/components/forms/FieldAction.tsx
jigjigjig: Resolving so the PR has nothing outstanding. This note is explanatory and needs no change; it stays readable here for anyone reviewing.

web/src/features/models/ModelComboBox.tsx
jigjigjig: Resolving so the PR has nothing outstanding. This note is explanatory and needs no change; it stays readable here for anyone reviewing.

web/src/shared/components/forms/FieldAction.test.tsx
jigjigjig: Resolving so the PR has nothing outstanding. This note is explanatory and needs no change; it stays readable here for anyone reviewing.

web/src/features/models/ModelComboBox.tsx
khaledosman: The hint is still not announced on the input. A bare node inside `FieldMessages` leaves the combo box with `aria-describedby: null`; HeroUI's `Description` is what wires it. Verified in jsdom against this tree: rendering a `ComboBox.Root` with a bare child in `FieldMessages` reports `aria-describedby: null`, and with a `Description` reports an id resolving to the hint text.

That matters more here than on a decorative caption, because this line is where "Could not list models for openai. Check that provider's credentials, or type the model key directly." and "Showing 50 of 300 matches" are said. `Field.tsx:60` and the sibling `UserComboBox.tsx:157` (also a `ComboBox.Root`) already use the wired shape:

```tsx
<FieldMessages>
  {hint ? <Description className="text-muted">{hint}</Description> : null}
</FieldMessages>
```

The reserve and both new tests hold unchanged under that.

web/src/shared/components/forms/FieldAction.tsx
```diff
@@ -0,0 +1,37 @@
...
+ *
+ * It reserves rather than nudging: no `self-*`, no padding, and nothing that
+ * asks `align-items` for what it cannot do. The height is the caption role's
+ * own line height, so a retune of the caption carries the action with it.
+ *
+ * Two limits, both real. Every field in these rows must reserve exactly one
+ * caption line: a message long enough to wrap makes that one field taller and
+ * drops it back out of line, and no reserve here can answer that. None of these
+ * messages wraps at the widths they are used, so this is sound now rather than
+ * permanently.
```
khaledosman: "None of these messages wraps at the widths they are used" does not hold for the guardrails row, which is the one row the description says was never rendered.

Measured at 13px Mozilla Text with this caption's letter-spacing: the two `ModeToggle` hints are 430px and 509px, the Profile description 281px. With the Remove button and three 12px gaps, a wrap-free line needs roughly 1500px. `PolicyForm` renders as inline page content rather than in a modal, so at 1280px (the desktop screenshot project) the row wraps and both hints go to two lines, which is exactly the case the two sentences above name as unanswerable by a reserve.

Suggest scoping the claim to the four rows that were measured and naming guardrails as the row where it does not hold, rather than asserting it for all of them. `ModeToggle` (`RoutingPage.tsx:387`) is also the one control in these rows with no width cap and its hint outside `FieldMessages`, if that is worth a follow-up.

web/src/features/routing/RoutingPage.tsx
khaledosman: `text-xs` is off the type scale, on a line this diff rewrites anyway. `text-body` is the role for a control label, or `text-caption` if it should read as metadata; design-tokens.md, "Type scale": the scale is the whole set of sizes.

Safe to change inside this PR rather than after it: the row bottom-aligns, so a taller label grows upward and the measured spread stays 0.

web/src/shared/components/forms/FieldAction.tsx
```diff
@@ -0,0 +1,37 @@
+import type { ReactNode } from "react"
+import { FieldMessages } from "@/shared/components/forms/FieldMessages"
+
+/**
+ * A trailing action in a control row, aligned to the input line.
```
khaledosman: "A trailing action in a control row" is narrower than the component's own use: `RoutingPage.tsx:900` wraps a mid-row radio label, and the rule `forms.md` gains in this PR is per-child, not per-trailing-child. Widening this first line keeps the radio call site from reading as a misuse of the component.

web/src/shared/components/forms/FieldAction.test.tsx
```diff
@@ -0,0 +1,52 @@
...
+  const reserve = (container: HTMLElement) =>
+    container.querySelector(".text-caption") as HTMLElement | null
```
khaledosman: `container.querySelector(".text-caption")` is the class selector `testing.md` rules out ("never a class selector": a token that will be renamed, and a test that breaks on a restyle teaches everyone to stop trusting the suite). The reserve is reachable structurally, which the second test in this file already does with `wrapper.firstElementChild`, so `wrapper.lastElementChild` works here. Same in `ModelComboBox.test.tsx:50`.

khaledosman (approved):
Checked out `cd4c2bb` in a clean worktree: `pnpm lint` and `pnpm typecheck` clean, `pnpm test` 3937/3938. The one failure is `WorkspaceCodeExecutionPolicyCard.test.tsx` in an untouched feature, and it passes when run alone, so it is the contention flake `testing.md` describes rather than anything here.

Gates: no route, schema or docstring moved, so no OpenAPI or Postman artifact is owed; `schema.ts` and `routeTree.gen.ts` untouched; no new page, so no screenshot entry; no dependency; `shared/` imports only `shared/`; conventional title; no em dashes or `--` separators in the prose.

Five inline notes, none blocking. The first is the one worth acting on before merge.

---

web/src/features/models/ModelComboBox.tsx
jigjigjig: Fixed in 5e4a27ee. The hint now goes through HeroUI's `Description`, same as `Field.tsx:60`. Verified in jsdom on this tree: with the bare node the input reported `aria-describedby: null`; with `Description` it reports an id resolving to the hint text. A test covers it, and it fails with the `Description` removed.

Worth recording that this is not a regression this PR introduced: the bare `<span className="text-caption">` it replaced was equally unannounced, so the hint has never been wired. It is worth fixing here because the line was being rewritten anyway, and because the strings on it are "Could not list models for openai. Check that provider's credentials, or type the model key directly." and "Showing 50 of 300 matches".

web/src/shared/components/forms/FieldAction.tsx
jigjigjig: Fixed in 5e4a27ee. You are right, and this is the worse half of the finding: I asserted "None of these messages wraps at the widths they are used" about the one row nobody can render.

The docstring now states it as known to break rather than as sound, and attributes the measurement to review rather than presenting it as mine: the two hints at 430px and 509px against roughly 1500px for a wrap-free line, so both wrap at 1280px and the row misaligns again. I have not re-derived your pixel math and do not claim to have. I did confirm the part I could check cheaply, that `PolicyForm` is inline page content rather than a modal, so it gets page width.

The PR body now says the guardrails row ships changed, unobserved, and with a documented case where the treatment does not hold.

web/src/shared/components/forms/FieldAction.tsx
jigjigjig: Fixed in 5e4a27ee. The first line now reads "A child of a control row that has no caption of its own, aligned to the input line: a trailing action, or any mid-row control without one", so the radio call site no longer reads as a misuse.

web/src/shared/components/forms/FieldAction.test.tsx
jigjigjig: Fixed in 5e4a27ee, in both files. `FieldAction.test.tsx` reaches the reserve as the last element child of the wrapper. `ModelComboBox.test.tsx` anchors on the label instead of the container, because the container's first element child is a react-aria `<template>` rather than the combo box root; the reserve is then the root's last element child.

Catching a real thing on the way: swapping the selector silently turned both of those tests into passes with the fix reverted, and my first re-check missed it because `git checkout --` restores to HEAD, which by then was my own commit rather than `main`. Re-verified against `git show origin/main:...`, where both fail on the missing reserve class.

Not fixed here, deliberately: `FieldMessages.test.tsx` does the same thing at lines 104, 112 and 129. Those predate this PR and are being raised separately rather than swept in.

web/src/features/routing/RoutingPage.tsx
jigjigjig: Changed to `text-body` in 1e627b74. It names a control, so it is the control-label role rather than metadata.

Your prediction verified rather than accepted, measured on the running branch build with all four rows open. Control-edge spread on the pool row stays 0: threshold 761, share 761, radio label 761, Remove 761, and the same on the other three rows. The label grew from 18px to 20px and grew upward, as you said it would.

One number you did not ask for but which follows from a taller label: the pool row's wrap-free width moves from 700px to 720px, measured at 20px granularity by narrowing the container until the row gains a line. So it wraps 20px earlier, and both figures sit far below the 1280px desktop viewport, so nothing changes at any width the page is actually used at.

PR #1006, opened 2026-09-09
fix(pipeline): map status-less InvalidRequestError to HTTP 400

### LINKED ISSUE #989: A status-less any-llm InvalidRequestError collapses to an opaque 502
### Problem

`classify_provider_error` returns `None` for any exception carrying no HTTP status, so a status-less `any_llm.exceptions.InvalidRequestError` falls through to `PROVIDER_ERROR_DETAIL = "LLM provider error"` (502, `src/gateway/api/routes/_pipeline.py:232`). `InvalidRequestError` defines no `status_code`, yet it means the request to the provider was invalid, which is a caller fault that deserves a 400 carrying the provider's own explanation.

any-llm raises it across gemini, bedrock and anthropic for request-shape problems: `"<field> must be a data URI"`, `"<field> is missing a MIME type"`, `"Tool name '...' is reserved by any-llm"`, and the Anthropic non-streaming timeout guard translation (`_translating_nonstreaming_guard`, `any_llm/providers/anthropic/base.py`). Each of those currently reaches the caller as an unactionable 502 with nothing billed, telling them a permanent failure was transient.

### Proposed solution

Add a branch to `classify_provider_error` mapping a status-less `InvalidRequestError` found anywhere on `upstream_exception_chain(exc)` to `HTTP_400_BAD_REQUEST` with `_caller_fault_detail(exc, PROVIDER_BAD_REQUEST_DETAIL)`. That mirrors how `_is_unsupported_feature_error`, immediately above it, already handles the same shape, and `_caller_fault_detail` redacts and length-caps, so the error-detail boundary holds.

Key on the exception type, not on message text. #799 matched the anthropic SDK's English sentence, which couples the classifier to a vendored string and covers one raise site out of many.

Unit coverage in `tests/unit/test_provider_error_classification.py`: the bare error, the `_WrappedError` path so it survives `ANY_LLM_UNIFIED_EXCEPTIONS=1`, `failure_status_code` so the usage row records 400, and a negative case pinning that an error which does carry a status is still classified by status.

### Notes

This does not close #533. It turns that opaque 502 into an actionable 400; the request itself still fails. Unblocking it is #988.

No dependency change and no config surface, so this is independent of #988 and can land first.

AloysJehwin:
## Description

`classify_provider_error` in `src/gateway/api/routes/_pipeline.py` returned `None` for any exception with no HTTP status. A status-less `any_llm.exceptions.InvalidRequestError` — raised by gemini, bedrock, and anthropic for bad request shapes — therefore fell through to a 502 "LLM provider error". It should instead map to `HTTP_400_BAD_REQUEST` with the provider's own explanation passed through `_caller_fault_detail`.

This PR adds a branch immediately before the `if status_code is None: return None` early exit that checks whether any exception in `upstream_exception_chain(exc)` is an `InvalidRequestError` with `status_code is None`, and maps it to 400. The pattern mirrors `_is_unsupported_feature_error` and the `_rejected_param` branches which handle the same "no HTTP status, but the exception type is the whole signal" shape.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes #989

## Checklist

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

**NOTE:**

## Summary

- Classify status-less `InvalidRequestError` exceptions as HTTP 400 caller errors.
- Preserve provider explanations with existing redaction and length limits.
- Preserve status-based classification when an error already has an HTTP status.
- Add coverage for wrapped errors, status preservation, and usage status reporting.
- Replace opaque provider errors with actionable caller errors.

### DIFF
--- a/src/gateway/api/routes/_pipeline.py
+++ b/src/gateway/api/routes/_pipeline.py
@@ -50,7 +50,7 @@ from typing import Any, Generic, Literal, NamedTuple, NoReturn, Protocol, TypeVa
 from urllib.parse import ParseResult, urlparse
 
 from any_llm import LLMProvider
-from any_llm.exceptions import AnyLLMError, UnsupportedParameterError
+from any_llm.exceptions import AnyLLMError, InvalidRequestError, UnsupportedParameterError
 from any_llm.types.completion import (
     ChatCompletion,
     ChatCompletionChunk,
@@ -495,6 +495,16 @@ def classify_provider_error(exc: BaseException) -> ProviderErrorMapping | None:
     # provider's SDK has no parameter for.
     if (param := _rejected_param(exc)) is not None:
         return ProviderErrorMapping(status.HTTP_400_BAD_REQUEST, _provider_rejected_param_detail(param))
+    # A status-less InvalidRequestError means the provider (gemini, bedrock,
+    # anthropic) rejected the request shape before assigning an HTTP status.
+    # Mirror the UnsupportedParameterError / NotImplementedError branches: the
+    # exception type is the whole signal, and the provider's own message names
+    # what was wrong.
+    if any(
+        isinstance(candidate, InvalidRequestError) and candidate.status_code is None
+        for candidate in upstream_exception_chain(exc)
+    ):
+        return ProviderErrorMapping(status.HTTP_400_BAD_REQUEST, _caller_fault_detail(exc, PROVIDER_BAD_REQUEST_DETAIL))
     if status_code is None:
         return None
     # Account billing exhaustion, which several providers report as a 400/422
--- a/tests/unit/test_provider_error_classification.py
+++ b/tests/unit/test_provider_error_classification.py
@@ -16,7 +16,7 @@ import asyncio
 import httpx
 import pytest
 from anthropic import APITimeoutError as AnthropicAPITimeoutError
-from any_llm.exceptions import UnsupportedParameterError
+from any_llm.exceptions import InvalidRequestError, UnsupportedParameterError
 from openai import APITimeoutError as OpenAIAPITimeoutError
 
 from gateway.api.routes._pipeline import (
@@ -619,3 +619,47 @@ def test_failure_status_code_keeps_the_upstream_status_for_billing() -> None:
     of my error rate is an empty wallet" stays answerable even though the caller
     saw a 502."""
     assert failure_status_code(_ParamError(400, None, _ANTHROPIC_BILLING_MSG)) == 400
+
+
+# ---------------------------------------------------------------------------
+# InvalidRequestError with no HTTP status maps to 400 (Fixes #989)
+# ---------------------------------------------------------------------------
+
+
+def test_status_less_invalid_request_error_maps_to_400() -> None:
+    """A bare InvalidRequestError with no HTTP status (as raised by gemini,
+    bedrock, and anthropic for bad request shapes) must map to HTTP 400 with
+    the provider's own explanation, not fall through to the generic 502."""
+    exc = InvalidRequestError("max_tokens exceeds context window for this model")
+    mapping = classify_provider_error(exc)
+    assert mapping is not None
+    assert mapping.status_code == 400
+    assert "max_tokens" in mapping.detail
+
+
+def test_status_less_invalid_request_error_survives_wrapped_error() -> None:
+    """The InvalidRequestError type check fires when it lives on
+    ``original_exception`` (the ANY_LLM_UNIFIED_EXCEPTIONS=1 shape)."""
+    original = InvalidRequestError("invalid message role: 'system'")
+    wrapped = _WrappedError(500, original)
+    mapping = classify_provider_error(wrapped)
+    assert mapping is not None
+    assert mapping.status_code == 400
+    assert "invalid message role" in mapping.detail
+
+
+def test_status_less_invalid_request_error_is_recorded_as_400() -> None:
+    """failure_status_code records 400 for a status-less InvalidRequestError so
+    the usage log reflects the real classification, not the generic 502."""
+    exc = InvalidRequestError("unknown parameter 'response_format'")
+    assert failure_status_code(exc) == 400
+
+
+def test_invalid_request_error_with_status_is_classified_by_status() -> None:
+    """An InvalidRequestError that already carries an HTTP status must not be
+    rerouted by the type-based branch; the status branch handles it as usual."""
+    exc = InvalidRequestError("bad request shape")
+    exc.status_code = 400
+    mapping = classify_provider_error(exc)
+    assert mapping is not None
+    assert mapping.status_code == 400


src/gateway/api/routes/_pipeline.py
```diff
@@ -495,6 +495,16 @@ def classify_provider_error(exc: BaseException) -> ProviderErrorMapping | None:
     # provider's SDK has no parameter for.
     if (param := _rejected_param(exc)) is not None:
         return ProviderErrorMapping(status.HTTP_400_BAD_REQUEST, _provider_rejected_param_detail(param))
+    # A status-less InvalidRequestError means the provider (gemini, bedrock,
+    # anthropic) rejected the request shape before assigning an HTTP status.
+    # Mirror the UnsupportedParameterError / NotImplementedError branches: the
+    # exception type is the whole signal, and the provider's own message names
+    # what was wrong.
+    if any(
```
khaledosman, line 503: `candidate.status_code is None` only says *that link* carried no status, and this branch runs before `if status_code is None`. So any chain that does carry an upstream status, but has a status-less `InvalidRequestError` anywhere on `original_exception`, is now classified 400 with the upstream message echoed:

```
AuthenticationError(status_code=401, original_exception=InvalidRequestError(...)) -> (400, "<upstream text>")   # was 502 / PROVIDER_CREDENTIALS_DETAIL
ProviderError(status_code=429, ...)                                              -> (400, "<upstream text>")   # was 429 / PROVIDER_RATE_LIMITED_DETAIL
ProviderError(status_code=500, ...)                                              -> (400, "<upstream text>")   # was 502
```

It jumps `is_provider_billing_error` for the same reason. It also breaks an invariant `redact_upstream_message` documents about itself: "Statuses where secrets concentrate (a rejected credential, a 5xx) never reach here at all, because `classify_provider_error` keeps a fixed detail for them." Redaction did catch an `sk-ant-` key in the 401 case above, but that is the second line of defense by design, not the first.

Gate on the failure's status, which is what "status-less" means here:

```python
if status_code is None:
    if any(isinstance(candidate, InvalidRequestError) for candidate in upstream_exception_chain(exc)):
        return ProviderErrorMapping(
            status.HTTP_400_BAD_REQUEST, _caller_fault_detail(exc, PROVIDER_BAD_REQUEST_DETAIL)
        )
    return None
```

Every new test still passes with that, except the `_WrappedError(500, ...)` one (see the comment on it).

src/gateway/api/routes/_pipeline.py
```diff
@@ -495,6 +495,16 @@ def classify_provider_error(exc: BaseException) -> ProviderErrorMapping | None:
     # provider's SDK has no parameter for.
     if (param := _rejected_param(exc)) is not None:
         return ProviderErrorMapping(status.HTTP_400_BAD_REQUEST, _provider_rejected_param_detail(param))
+    # A status-less InvalidRequestError means the provider (gemini, bedrock,
```
khaledosman, line 498: Separate concern on the same block: keying on the type is only as narrow as whatever assigns the type. Once `ANY_LLM_UNIFIED_EXCEPTIONS=1` (the flag this branch's own test targets), `convert_exception` picks `InvalidRequestError` from `re.search(r"invalid|badrequest|validation", type_name + message)` whenever there is no status, and that pattern sits ahead of the `timeout|connection|network` catch-all. Gateway-side faults then land here:

```
httpx.InvalidURL("Invalid URL component 'scheme'")   -> (400, "Invalid URL component 'scheme' ...")   # operator base_url typo
ParamValidationError("Parameter validation failed")  -> (400, "Parameter validation failed ...")       # botocore
```

That is an operator config fault reported to the caller as "your request was invalid", with no remedy they can apply. Not reachable today (the flag is unset repo-wide), so this is a decision rather than a bug: either narrow the branch, or pin the shape with a test so the flag flip does not land it silently.

tests/unit/test_provider_error_classification.py
```diff
@@ -619,3 +619,47 @@ def test_failure_status_code_keeps_the_upstream_status_for_billing() -> None:
...
+    assert mapping is not None
+    assert mapping.status_code == 400
+    assert "max_tokens" in mapping.detail
+
+
+def test_status_less_invalid_request_error_survives_wrapped_error() -> None:
+    """The InvalidRequestError type check fires when it lives on
+    ``original_exception`` (the ANY_LLM_UNIFIED_EXCEPTIONS=1 shape)."""
+    original = InvalidRequestError("invalid message role: 'system'")
+    wrapped = _WrappedError(500, original)
```
khaledosman, line 644: This pins the preemption above as the intended contract: it asserts a failure carrying a 500 becomes a client 400. The `ANY_LLM_UNIFIED_EXCEPTIONS=1` shape it claims to model does not produce that, because `convert_exception` returns an `AnyLLMError` unchanged, so no any-llm wrapper ever sits above an `InvalidRequestError`.

Use a status-less wrapper to model the real shape, and add the negative case the branch is missing: a 429 or 401 outer with a status-less `InvalidRequestError` on `original_exception` has to stay 429 / 502. `test_invalid_request_error_with_status_is_classified_by_status` below only covers the status sitting on the same exception, which is the case the current guard already handles.

khaledosman (changes requested):
Three findings inline. The first is the blocker: the new branch sits ahead of the whole status ladder and its guard is per-link rather than per-failure, so a chain that does carry an upstream status (401, 429, 5xx) is now reclassified as a client 400 with the upstream message echoed.

Checked and clean: `make lint` (architecture check + ruff) passes, the 133 related unit tests pass, no route or schema change so no OpenAPI/Postman debt, and the PR title is conventional. (`ruff format --check` flags `_pipeline.py`, but it does so on `main` too and `make lint` does not run it, so that one is not yours.)

One nit with no line to anchor it to: the PR description uses em dashes, which `AGENTS.md` rules out for PR descriptions.

---

AloysJehwin:
You're right on all three, and the first one is a real bug I introduced. Fixed in 24cd51b9.

**The blocker.** The branch tested `candidate.status_code is None` per link and ran ahead of `if status_code is None`, so any failure carrying its own status was reclassified whenever a status-less `InvalidRequestError` sat anywhere on `original_exception`. Reproduced against the ladder:

```
case                            BEFORE                          AFTER
401 wrapping status-less IRE    (400, '<upstream text>')        (502, 'PROVIDER_CREDENTIALS_DETAIL')
429 wrapping status-less IRE    (400, '<upstream text>')        (429, 'PROVIDER_RATE_LIMITED_DETAIL')
500 wrapping status-less IRE    (400, '<upstream text>')        (502, 'PROVIDER_UPSTREAM_DETAIL')
bare status-less IRE            (400, '<upstream text>')        (400, '<upstream text>')   <- the case the PR is for
```

Moved inside the `status_code is None` gate, exactly as you suggested. Your point about `redact_upstream_message`'s invariant is the part I had not considered: a 401's detail is fixed precisely so the upstream text never reaches redaction, and my branch routed it there. Redaction catching the `sk-ant-` key was the second line of defense doing work the first line should have made unnecessary.

**The `_WrappedError(500, ...)` test.** Also correct, and worse than a wrong assertion: it pinned the preemption as the intended contract, so it would have defended the bug. `_WrappedError` requires a status, so it cannot model the shape the PR targets. Replaced with a `_StatusLessWrapper` helper, and added the negative case as a parametrized test over 401 and 429.

Both new cases are genuine regression tests. Against the previous commit:

```
FAILED test_a_carried_status_wins_over_an_invalid_request_error_in_the_chain[401-expected0]
FAILED test_a_carried_status_wins_over_an_invalid_request_error_in_the_chain[429-expected1]
2 failed, 1 passed
```

and at this head, `88 passed in 34.28s`, `make lint` clean.

**On `ANY_LLM_UNIFIED_EXCEPTIONS=1` narrowing the branch** — I've left this as-is deliberately and want to flag the reasoning rather than quietly skip it. With the gate moved, the flag's failure mode is narrower than it was: `httpx.InvalidURL` and `ParamValidationError` carry no status, so they would land on the 400 branch, but that is now the same fall-through they already take today (`return None` → generic 502 at the call site) rather than a preemption of a real status. It is still an operator-config fault described to the caller as their fault, which I agree is wrong. It seemed better as its own change with a decision behind it than a guess bolted onto this one. Happy to either narrow the isinstance check or add the pinning test here if you'd prefer it not wait.

Em dashes are out of the PR description.

PR #1008, opened 2026-09-09
fix(dashboard): wrap response.json() SyntaxError as ApiError in apiFetch

### LINKED ISSUE #993: A non-JSON 2xx body reaches the dashboard as a raw SyntaxError
`apiFetch` rethrows whatever `response.json()` throws on a 2xx body that is not JSON. The comment above it says every caller expects an `ApiError`, and a raw `SyntaxError` "would reach the UI as an unrecognized failure". That is what a dashboard behind an edge that answers a JSON path with an HTML fallback shows: "Unexpected token '<' ... is not valid JSON", with nothing naming the request. Seen on the otari-ai development deployment on 9 September, from `/dashboard-build.json` being served by the wrong origin.

Wrap it as an `ApiError` with a message that says the gateway answered with something other than JSON, and have `useDashboardBuild` treat that as unknown rather than an error, since a failed build poll should never interrupt anyone.

_Note: this issue was drafted by Claude Fable 5.1 via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

AloysJehwin:
## Description

An edge proxy can answer `200` with an HTML interstitial instead of JSON. Every caller in the dashboard's API client expects an `ApiError`, so the raw `SyntaxError` from parsing that body escaped as an unrecognized failure: the sign-in screen recorded no status at all and showed the browser's own `Unexpected token '<'`. The same body now produces an ordinary `ApiError` saying the gateway answered with something other than JSON, so the screens that already handle a failed request handle this one too.

The response's own status is kept rather than `0`, because the gateway did answer. The dashboard treats status `0` as "cannot reach the gateway", and a non-JSON answer is a different fault that should not raise that banner.

## How to test it locally

1. Put any proxy that returns an HTML page in front of the gateway, or stub `fetch` to resolve a `200` with a `text/html` body.
2. Sign in from the dashboard.
3. The screen reports that the gateway answered with something other than JSON, rather than `Unexpected token '<'`, and the "Can't reach the gateway" banner stays down.

Automated coverage: `pnpm --dir web exec vitest run src/shared/api/client.test.ts` adds one case per entry point. Both fail with the source change reverted and pass at this head. `pnpm --dir web exec vitest run src/shared/api src/app` is 292 passing, `tsc --noEmit` and Biome are clean.

## PR Type

- [x] Bug Fix

## Relevant issues

Fixes #993

## Checklist

No documentation or API contract change: this is dashboard-internal error handling, with no route or schema touched.

## AI Usage

**AI Model/Tool used:**

Claude Code (Opus 5)

The review on this PR was acted on with Claude Code, and doing so surfaced that the branch was built on a stale base: `main` had moved the build poll onto `siteFetch`, so the original diff patched a function that path no longer uses. The branch is rebuilt on current `main` as a result, and covers four unguarded reads rather than one. Each claim above was checked by running it, including reverting the source change to confirm the new tests actually fail without it.

---

## What changed since the last review

@khaledosman's three findings, and what each turned into:

**Incomplete fix.** Correct, and wider than reported. `publicGet` and `publicPost` were unguarded as flagged, and the rebase found two more: `siteFetch` (the build poll, which runs once a minute against the gateway's own root and so meets an interstitial first) and `apiFetch`, which still had a bare `throw error`. All four now go through one `parseJsonBody` helper.

**Message wording.** Taken: it now matches its neighbours in the file, which are sentences ending in a period and say "the gateway".

**Dropped `cause`.** Taken: `ApiError` accepts an optional `{ cause }`, so the underlying `SyntaxError` is still recorded.

**The two `deployment.ts` findings resolve by deletion.** Both were caused by the `.catch(() => null)` in the old diff. Current `main` has no catch on the build poll, so react-query keeps the last successful data on an error, the update pill survives a failed poll on its own, and the poll stays in the status-0 scan because nothing swallows its errors. I reproduced the retraction you described and then found it only reachable through my own catch, so there is no latch to add. `deployment.ts` and `UpdatePrompt.tsx` are out of this PR entirely.

https://claude.ai/code/session_01CFRv5kvesfnKYgHmNw9Vpf

### DIFF
--- a/web/src/shared/api/client.ts
+++ b/web/src/shared/api/client.ts
@@ -436,11 +436,12 @@ export async function apiFetch<T>(
   try {
     return (await response.json()) as T
   } catch (error) {
-    // Every caller expects an ApiError; a raw DOMException here would reach the
-    // UI as an unrecognized failure. A malformed body is still its own error.
+    // Every caller expects an ApiError; a raw SyntaxError or DOMException here
+    // would reach the UI as an unrecognized failure. A malformed body is still
+    // its own error, but it must be wrapped so callers can handle it uniformly.
     if (isTimeout(error)) {
       throw new ApiError(0, timeoutMessage)
     }
-    throw error
+    throw new ApiError(response.status, "Gateway returned a non-JSON response")
   }
 }
--- a/web/src/shared/api/deployment.ts
+++ b/web/src/shared/api/deployment.ts
@@ -21,14 +21,16 @@ import {
 export function useDashboardBuild() {
   return useQuery({
     queryKey: [BUILD],
-    queryFn: () => apiFetch<DashboardBuild>("/dashboard-build.json"),
+    queryFn: () =>
+      apiFetch<DashboardBuild>("/dashboard-build.json").catch(() => null),
     refetchInterval: BUILD_POLL_MS,
     // A tab left open in the background is the one most likely to be stale, so
     // check again the moment someone comes back to it.
     refetchOnWindowFocus: true,
     staleTime: 0,
     // A failed check is not worth reporting: the tab keeps working, and the next
-    // poll retries anyway.
+    // poll retries anyway. A non-JSON body (ApiError from apiFetch) is treated as
+    // unknown rather than an error so build polling never interrupts the user.
     retry: false,
   })
 }


web/src/shared/api/deployment.ts
```diff
@@ -21,14 +21,16 @@ import {
 export function useDashboardBuild() {
   return useQuery({
     queryKey: [BUILD],
-    queryFn: () => apiFetch<DashboardBuild>("/dashboard-build.json"),
+    queryFn: () =>
+      apiFetch<DashboardBuild>("/dashboard-build.json").catch(() => null),
```
khaledosman, line 25: Regression: `.catch(() => null)` replaces the last-known build with `null`, so one failed poll retracts an **already-shown** update pill for up to `BUILD_POLL_MS` (60s), until the next poll restores it. Before this change react-query kept the last successful `data` on an error, so the pill survived.

Reproduced with the existing `UpdatePrompt.test.tsx` harness on both revisions:

| | data after a failed poll | pill still shown |
|---|---|---|
| base | `{build: "build-b", …}` | yes |
| this head | `null` | no, returns on the next poll |

The tab already knows a new build exists; one bad poll should not retract that. Latch `useUpdateAvailable` once it is true (`UpdatePrompt.tsx:19`), or keep the previous value rather than resolving `null`.

web/src/shared/api/deployment.ts
```diff
@@ -21,14 +21,16 @@ import {
...
     refetchInterval: BUILD_POLL_MS,
     // A tab left open in the background is the one most likely to be stale, so
     // check again the moment someone comes back to it.
     refetchOnWindowFocus: true,
     staleTime: 0,
     // A failed check is not worth reporting: the tab keeps working, and the next
-    // poll retries anyway.
+    // poll retries anyway. A non-JSON body (ApiError from apiFetch) is treated as
+    // unknown rather than an error so build polling never interrupts the user.
```
khaledosman, line 33: The comment names the non-JSON body, but the catch on line 25 swallows every failure. An unreachable gateway (`ApiError` status 0) now resolves as a success, so this poll drops out of the cache-wide scan in `ConnectionStatus.tsx:12` that looks for status-0 errors to raise "Can't reach the gateway".

Rethrow `status === 0` and let the rest be unknown, so the code matches the comment. Keeping `response.status` on the new `ApiError` rather than 0 is what makes that distinction possible, which is worth a word here.

web/src/shared/api/client.ts
```diff
@@ -436,11 +436,12 @@ export async function apiFetch<T>(
...
-    // Every caller expects an ApiError; a raw DOMException here would reach the
-    // UI as an unrecognized failure. A malformed body is still its own error.
+    // Every caller expects an ApiError; a raw SyntaxError or DOMException here
+    // would reach the UI as an unrecognized failure. A malformed body is still
+    // its own error, but it must be wrapped so callers can handle it uniformly.
     if (isTimeout(error)) {
       throw new ApiError(0, timeoutMessage)
     }
-    throw error
+    throw new ApiError(response.status, "Gateway returned a non-JSON response")
```
khaledosman, line 445: This message is the odd one out among the user-facing strings in this file, which are sentences ending in a period and say "the gateway" ("Network error: could not reach the gateway.", "The gateway did not respond in time."). #993 asks for "a message that says the gateway answered with something other than JSON"; use that wording.

Separately, the underlying `SyntaxError` is dropped with no `cause`, so nothing now records what the intermediary actually returned. Worth widening `ApiError` to carry one.

khaledosman (changes requested):
The `apiFetch` wrap is the right shape. Two things below it need changing before merge, and both are outside the diff hunks so they are here rather than inline.

**`web/src/shared/api/client.ts:285` and `:330`** — incomplete fix. `publicGet` and `publicPost` still `await response.json()` unguarded, and `POST /v1/auth/session` is the likeliest path in this file to meet an edge-proxy HTML interstitial. `Login.tsx` then records `status: undefined` and renders the same "Unexpected token '<'" that #993 is about. Same one-line wrap as the one in this diff.

**No test.** The checklist and the summary both claim coverage for non-JSON successful responses, but the diff has no test file. `client.test.ts:54` ("bounds a response whose body stalls after the headers arrive") is the sibling case and the pattern to copy, and frontend-standards requires a Vitest test for any helper whose behavior changes. Owed: a `client.test.ts` case for a 200 with an HTML body, and an `UpdatePrompt.test.tsx` case for a failed poll. The second is what would have caught the inline finding on `deployment.ts:25`.

Gates that are clear: conventional title, no route or schema change so no OpenAPI or Postman artifact owed, no `schema.ts` / `routeTree.gen.ts` drift, no dependency change. Typecheck and the three affected suites pass at this head.

AloysJehwin:
All three findings addressed, in 98504e59. Acting on the first one turned up that this branch was built on a stale base, so it is rebuilt on current `main` rather than patched.

**The incomplete fix.** You were right that `publicGet`/`publicPost` were left unguarded, and the rebase found two more: `main` moved the build poll onto `siteFetch` (for #1026), which has the same unguarded read, and `apiFetch` still had a bare `throw error`. Four sites in total, now behind one `parseJsonBody` helper.

`siteFetch` is the one worth calling out: it polls the gateway's own root once a minute for the life of every open tab, so it is the likeliest of the four to meet an edge-proxy interstitial. It was not in scope before the rebase.

**Wording and the dropped cause.** Both taken. The message now matches its neighbours ("The gateway answered with something other than JSON."), and `ApiError` takes an optional `{ cause }` so the underlying `SyntaxError` survives.

**The two `deployment.ts` findings resolve by deletion.** Both were caused by the `.catch(() => null)` in the old diff. Current `main` has no catch on the build poll at all, so react-query keeps the last successful `data` on an error, the pill survives a failed poll on its own, and the poll stays in `ConnectionStatus`'s status-0 scan because nothing swallows its errors. I confirmed the retraction you reproduced (`data: null`, `pill: false`) and then found it only reachable through my own catch. No latch needed; `deployment.ts` and `UpdatePrompt.tsx` are out of the PR entirely.

I kept the response's own status rather than 0 on the new `ApiError`, which is the distinction that makes the above work: `siteFetch` already throws status 0 for a genuinely unreachable gateway, and a non-JSON answer is a different fault.

**Tests**, verified as real regression tests rather than assumed. With the source change reverted:

```
 ❯ src/shared/api/client.test.ts (22 tests | 2 failed)
     × wraps a non-JSON body as an ApiError carrying the parse failure
     × reports a 200 whose body is not JSON as an ApiError
```

at this head, `292 passed (292)` across `src/shared/api` and `src/app`, `tsc --noEmit` clean, Biome clean.

PR #1014, opened 2026-09-09
security(auth): stop is_superuser from bypassing tenant role checks

### LINKED ISSUE #1011: is_superuser bypasses organization- and workspace-role checks, not just the deployment's own surfaces
## What is open

Every tenant-scoped authorization check that gates budgets, provider keys, guardrails, spend ceilings, domains, and workspace visibility/management treats `user.is_superuser` as an automatic pass, regardless of that identity's actual organization or workspace role:

- `services/tenancy/organization_service.py:219`, `_enforce_management_role`:
  ```python
  if membership.role not in MANAGEMENT_ROLES and not user.is_superuser:
      raise NotAuthorizedError
  ```
  Gates `require_active_organization_management_access`, which every one of `/v1/organizations/me/budgets`, `/provider-keys`, `/guardrails`, `/spend-ceilings`, `/domains` (and their writes) goes through.
- `services/tenancy/authorization.py:67` (`resolve_visible_workspace`), `:104` (`has_workspace_management_access`), and `:183` (`visible_workspace_scope`) each open with `if user.is_superuser: return ...` ahead of the organization- and workspace-role checks. `visible_workspace_scope` is what `WorkspaceService.list_workspaces` (`GET /v1/workspaces`) resolves through, so it decides whether the caller sees every workspace in the organization or only the ones they belong to.

An identity that is a plain **viewer** in an organization, but happens to be a deployment operator, therefore:

- sees every workspace in that organization via `GET /v1/workspaces` and the Workspaces page, not just ones they are a member of
- gets `200` (not `403`) from `GET /v1/organizations/me/budgets`, `/provider-keys`, `/guardrails`, `/spend-ceilings`, `/domains`
- would have those same routes honor a write (`PATCH`/`POST`/`DELETE`) if called directly, since the same `is_superuser` check gates management access, not only listing

Verified live against a seeded viewer-role operator identity: `GET /v1/workspaces` returned all three of the organization's workspaces despite zero workspace memberships, and the five management endpoints above all returned `200` where a genuine viewer or member correctly gets `403`. The Workspaces page itself renders all three rows (with Edit/Delete present in the DOM, though disabled client-side on that one page - the server-side check would still honor the request).

This contradicts what the dashboard's own Members & roles page states as the model (`OrganizationMembersPage.tsx`'s copy): *"members and viewers read the workspaces they belong to... no role set here reaches the deployment's own pages."* An operator's organization-level role is exactly the "role set here," and it is not what decides their access.

Not the same shape as #838 (which is about an org role conferring nothing deployment-wide): this is the inverse - deployment-operator status conferring organization-owner-equivalent access inside a tenant, past whatever role that identity actually holds there.

## Suggested fix

Two separable decisions:

1. **Whether an operator should have a tenant-data support path at all.** If yes, it should be an explicit, auditable "act as this organization" path (a distinct capability/log trail), not an implicit fallthrough on every organization-role check. If no, drop the `is_superuser` arm from the three call sites in `authorization.py` and from `_enforce_management_role`, so a caller's own active membership and role decide organization- and workspace-scoped access exactly as they would for anyone else.
2. **Scope the fix to the tenant-facing surfaces only.** The genuinely deployment-wide routes (`/v1/admin/*`, `/v1/users`, `/v1/keys`, `/v1/budgets`, `/v1/scoped-budgets`, `/v1/routing/policies`, `/v1/aliases`, `/v1/settings`, `/v1/provider-credentials`, `/v1/hosted-backends`, `/v1/credit-grants`) are already correctly gated on `require_deployment_operator` at the router level and read/write across every organization on purpose - that is the intended "manage the whole platform" exception and should not change.

Acceptance: an identity holding only a viewer or member role in an organization sees and can act on exactly what that role permits there, whether or not they separately operate the deployment; the deployment-wide admin surfaces are unaffected.

tbille:
## Description

Every tenant-scoped authorization check that gates budgets, provider keys, guardrails, spend ceilings, domains, and workspace visibility/management treated `user.is_superuser` as an automatic pass, regardless of that identity's actual organization or workspace role. An identity that was a plain viewer/member in an organization but also a deployment operator could see every workspace in that organization via `GET /v1/workspaces` and get `200` (not `403`) from `GET /v1/organizations/me/{budgets,provider-keys,guardrails,spend-ceilings,domains}` and their writes.

Per the issue's decision point, this drops the `is_superuser` arm entirely with no replacement mechanism: a caller's own active membership and role decide organization- and workspace-scoped access exactly as they would for anyone else. An explicit, auditable "act as this organization" support path is a separate feature, out of scope here.

### What changed

Removed the bypass from all five call sites (one of them a previously undocumented duplicate the live repro actually goes through):
- `authorization.py::resolve_workspace_in_organization` (backs `resolve_visible_workspace`)
- `authorization.py::has_workspace_management_access`
- `authorization.py::resolve_visible_workspace_scope`
- `organization_service.py::_enforce_management_role`
- `workspace_service.py::list_workspaces`, a fifth, hand-rolled copy of the same rule that `GET /v1/workspaces` goes through

`require_deployment_operator` (`/v1/admin/*`, `/v1/keys`, `/v1/budgets`, etc.) is untouched, since that is the correct, separate gate for genuinely deployment-wide surfaces.

While the module was open, also:
- Recorded the invariant this change encodes in `authorization.py`'s module docstring and `VisibleWorkspaceScope`'s, so a future contributor fixing an operator-access complaint doesn't reach for the arm again.
- Standardized `has_workspace_management_access` onto `get_active_by_workspace_and_user`, matching its neighbor, now that this read is the entire authorization decision rather than one branch of it.
- Dropped `VisibleWorkspaceScope.role`, which no caller read.

### Review

Went through a 3-round multi-model council review (OpenAI/Anthropic/Gemini). All three converged on: the bypass is fully closed at the right layer, 404-vs-403 existence-oracle discipline is preserved, and the deployment-wide surfaces are unaffected. The council's two unanimous "before merge" asks (the docstring rationale and the active-membership read consistency) are included above. I also verified their one open, lower-confidence concern, that the dashboard's `deployment_operator` context flag might gate a tenant-scoped affordance instead of the caller's role, with a dedicated audit; it does not. Every consumer either gates a genuinely deployment-wide surface or selects between two parallel API surfaces, never a tenant-scoped one in place of a role check.

Deferred as follow-up (not blocking, per unanimous council agreement): `list_workspaces` still re-derives its visibility boolean rather than delegating to `resolve_visible_workspace_scope`, because that resolver returns an unpaged workspace-id list while `list_workspaces` needs `skip`/`limit` pushed into the query, a real extraction rather than a one-line delegation. Left documented with a cross-reference for now.

### Testing

- New/updated regression tests in `test_tenancy_authorization.py` and `test_org_provider_keys.py`: an `is_superuser=True` identity holding only a viewer/member role is refused exactly as a plain member would be, across `resolve_visible_workspace`, `has_workspace_management_access`, `resolve_visible_workspace_scope`, and organization provider-key management.
- `tests/integration/` full organization/workspace/tenancy sweep: 607 passed.
- `make lint`: architecture check, alembic heads, Ruff all clean.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes #1011

## Checklist

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

Claude Code (Claude Sonnet 5), including a `/star-chamber` multi-model council review (OpenAI, Anthropic, Gemini) as part of self-review.

Implemented end-to-end by Claude Code from the linked issue: identified the affected call sites, made the fix, wrote/updated regression tests, ran a 3-round adversarial multi-model review against the acceptance criteria, and applied the review's unanimous "before merge" findings.

**NOTE:**

## Summary

Tenant-scoped access now depends on active organization and workspace roles. `is_superuser` no longer bypasses checks for budgets, provider keys, guardrails, spend ceilings, domains, workspace visibility, or workspace management.

The update standardizes active-membership checks, removes unused scope data, and aligns documentation and regression tests with the new rules. Deployment-wide operator routes remain unchanged.

### DIFF
--- a/src/gateway/services/tenancy/authorization.py
+++ b/src/gateway/services/tenancy/authorization.py
@@ -8,6 +8,13 @@ active organization and organization-level role) but nothing here reaches
 back, which is what lets ``workspace_budget_default_service`` sit between
 ``workspace_service`` and ``organization_service`` in the import graph without
 closing a cycle.
+
+Every check below reads only active organization and workspace membership.
+``User.is_superuser`` is deployment-operator status, not an organization or
+workspace role, and is deliberately never consulted here: an operator's
+tenant-scoped access is decided by their own membership exactly as anyone
+else's is. Deployment-wide surfaces are gated separately, by
+``require_deployment_operator`` at the router level. See mozilla-ai/otari#1011.
 """
 
 import uuid
@@ -30,11 +37,11 @@ async def resolve_visible_workspace(
 ) -> Workspace:
     """Resolve a workspace the caller may see, in their active organization, or raise not-found.
 
-    Visible to a superuser, an organization owner/admin (who see every
-    workspace in it), or an active member of the workspace itself. Every "may
-    not see" case answers 404 rather than 403: another organization's
-    workspace, and a workspace in this organization the caller is not a member
-    of, must be indistinguishable from one that does not exist.
+    Visible to an organization owner/admin (who see every workspace in it), or
+    an active member of the workspace itself. Every "may not see" case answers
+    404 rather than 403: another organization's workspace, and a workspace in
+    this organization the caller is not a member of, must be indistinguishable
+    from one that does not exist.
     """
     organization = await organizations.get_active_organization_for_user(user)
     return await resolve_workspace_in_organization(
@@ -64,9 +71,6 @@ async def resolve_workspace_in_organization(
     if workspace is None or workspace.organization_id != organization.id:
         raise WorkspaceNotFoundError(workspace_id)
 
-    if user.is_superuser:
-        return workspace
-
     organization_membership = await organizations.members.get_active_by_organization_and_user(
         organization.id,
         user.id,
@@ -87,13 +91,7 @@ async def has_workspace_management_access(
     workspace: Workspace,
     organizations: OrganizationService,
 ) -> bool:
-    """Whether the caller is a superuser, an organization owner/admin, or an owner/admin here.
-
-    The superuser arm is what makes this agree with the two checks either side
-    of it: ``resolve_visible_workspace`` grants a superuser read on every
-    workspace, and organization-level management access grants them
-    organization management, so without it a superuser could delete a
-    workspace but not rename it.
+    """Whether the caller is an organization owner/admin, or an owner/admin of this workspace.
 
     The predicate form exists for the one caller that has to *report* the answer
     rather than act on it: the first-request setup guide tells the dashboard
@@ -101,9 +99,6 @@ async def has_workspace_management_access(
     managing it is told "not for you" instead of being refused. Everything else
     wants :func:`require_workspace_management_access`.
     """
-    if user.is_superuser:
-        return True
-
     organization_membership = await organizations.members.get_active_by_organization_and_user(
         workspace.organization_id,
         user.id,
@@ -111,8 +106,8 @@ async def has_workspace_management_access(
     if organization_membership is not None and organization_membership.role in MANAGEMENT_ROLES:
         return True
 
-    membership = await WorkspaceMemberRepository(db).get_by_workspace_and_user(workspace.id, user.id)
-    return membership is not None and membership.status == "active" and membership.role in MANAGEMENT_ROLES
+    membership = await WorkspaceMemberRepository(db).get_active_by_workspace_and_user(workspace.id, user.id)
+    return membership is not None and membership.role in MANAGEMENT_ROLES
 
 
 async def require_workspace_management_access(
@@ -122,7 +117,7 @@ async def require_workspace_management_access(
     workspace: Workspace,
     organizations: OrganizationService,
 ) -> None:
-    """Allow a superuser, an organization owner/admin, or an owner/admin of this workspace."""
+    """Allow an organization owner/admin, or an owner/admin of this workspace."""
     if not await has_workspace_management_access(db, user=user, workspace=workspace, organizations=organizations):
         raise NotAuthorizedError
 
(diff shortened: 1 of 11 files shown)


src/gateway/services/tenancy/authorization.py
```diff
@@ -8,6 +8,13 @@
 back, which is what lets ``workspace_budget_default_service`` sit between
 ``workspace_service`` and ``organization_service`` in the import graph without
 closing a cycle.
+
+Every check below reads only active organization and workspace membership.
+``User.is_superuser`` is deployment-operator status, not an organization or
+workspace role, and is deliberately never consulted here: an operator's
+tenant-scoped access is decided by their own membership exactly as anyone
+else's is. Deployment-wide surfaces are gated separately, by
+``require_deployment_operator`` at the router level. See mozilla-ai/otari#1011.
```
khaledosman, line 17: The invariant is recorded here, but its client half still records the opposite, and that half is the one a contributor acts on:

* `web/src/features/organization/roles.ts:30-34`: "**Superuser is not on the contract.** The server grants organization management, and deletion, to `user.is_superuser` whatever their role ... which is the divergence to close when the context grows the field." As of this PR the server does not, and "close the divergence" now reads as an instruction to add a superuser check to the client, which is the arm this PR removes. It should say that tenant-scoped access reads membership only (#1011), and that otari.ai's `useIsOrgAdmin` is now the side that diverges.
* `web/src/features/organization/OrganizationProviderKeysPage.tsx:287-293`: "the server also admits a superuser whatever their organization role" is no longer true. `canEdit = canManage(context.data)` is now exactly the server's rule, not a deliberate narrowing in the safe direction.
* `web/src/features/overview/OverviewPage.tsx:381-382`: same, "`canManage` is narrower than the server on one point, a superuser with no management role".

Behavior is unaffected: every `isDeploymentOperator` call site in `web/src` gates a genuinely deployment-wide read, so this is comments only. It matters because #1011 opened partly on the dashboard's own copy contradicting the server, and three files now contradict it in the other direction.

khaledosman (approved):
`is_superuser` is gone from every tenant-scoped path: `git grep is_superuser src/gateway` at head leaves only the deployment-user admin surface, the model, and the repository. The five call sites are each covered by a new test, and the 404-vs-403 split is preserved (outsider raises `WorkspaceNotFoundError`, visible-but-unprivileged raises `NotAuthorizedError`). Generated artifacts (spec, Postman, `schema.ts`) are in sync, and CI is green on all 17 jobs.

No first-boot lockout follows from dropping the arm: `provisioning_service._provision` gives the operator an `owner` organization membership *and* an `owner` workspace membership, and `create_organization_for_user` makes the creator an owner, so the two identities that could have depended on the bypass do not.

**PR title.** `security(auth):` rather than `fix(auth):`. `security` is a recognized prefix in `otari-pr-title.yml` and `cliff.toml`, `RELEASE.md` names it for security fixes, and there is precedent (`security: refuse an organization rate override...` #1164, `security(guardrails): ...` #1088). This is a tenant-isolation fix that also removes access an operator holds today, so the release line is the only place a deployment finds that out.

One inline finding below. It is anchored here rather than on the three `web/` files it names, because those files are not in the diff and GitHub takes no inline comment on them.

*Review by Claude Code (Claude Opus 5), run through the repo's `/review` skill.*

src/gateway/services/tenancy/authorization.py
peteski22: Fixed in 4d1d87a1

Issue #998, opened 2026-09-09
feat(pricing): poll genai-prices upstream on a schedule instead of waiting for an operator to click refresh


daavoo:
## What

Automate the upstream half of the genai-prices refresh. The fetch machinery already exists and is fully operator-driven: nothing checks for new pricing data unless a human posts to `/v1/pricing/refresh`.

## Why

Prices resolve through three layers (`src/gateway/services/pricing_service.py`):

1. explicit `model_pricing` rows (config or `POST /v1/pricing`),
2. an accepted snapshot in `pricing_snapshots`, applied via `set_custom_snapshot`,
3. the dataset bundled in the installed `genai-prices` wheel.

A deployment that never clicks refresh runs on layer 3 forever, and the bundled dataset lags upstream `data.json` by construction, because the data changes faster than releases are cut. Bumping the dependency (#996) raises the floor but does not close this: the wheel is a snapshot too.

This is a silent failure by nature. Wrong prices look exactly like right prices, and the only symptom is mis-billing. GPT-5.6 Luna is currently priced from the bundled `0.1.0` dataset at `$1/M` in and `$6/M` out, against upstream's post-2026-07-30 `$0.20/M` and `$1.20/M`. That is a 5x overcharge that nothing in the product surfaces.

## What exists today

- `prepare_price_refresh` fetches `DEFAULT_UPDATE_URL`, diffs it against the active snapshot, and stores it as a `genai-prices-pending` row. Only reachable from `POST /v1/pricing/refresh`.
- `confirm_price_refresh` promotes pending to active, calls `set_custom_snapshot`, and resets the price cache.
- `run_price_snapshot_refresher` (30s, wired in `main.py`) converges sibling workers and replicas on the *accepted* snapshot. It never touches the network.
- genai-prices ships its own `UpdatePrices` auto-update thread. It is deliberately unused: it would reprice in memory with no diff, no DB record, no cross-replica agreement, and no audit trail.

So the only missing piece is something that calls `prepare_price_refresh` on a schedule.

## Proposal

A poller task alongside the existing refreshers, gated by config:

```yaml
pricing:
  auto_refresh: notify      # off | notify | auto
  refresh_interval_hours: 24
```

- `notify` (suggested default): fetch, stage the pending snapshot, surface "N price changes available" in the dashboard. A money-affecting change still needs a human, which is the property the current preview/confirm design protects.
[...]

PR #1015, opened 2026-09-09
feat(catalog): group the model catalog by model, with one offering per provider

### LINKED ISSUE #700: Per-million rates round to cents above $0.01, so a rate list can show a figure nobody entered
`formatCost` uses four fraction digits only below `$0.01` and two above it, which is right for spend and wrong for a per-million rate. A rate of `0.075` renders as `$0.08`, `0.035` as `$0.04`, `0.0125` as `$0.01`. Those are real per-million rates, so a rate table can display a figure the operator did not enter.

## Where

Three surfaces format the same stored rates through `formatCost`:

- `web/src/features/pricing/ModelPricingPage.tsx` — the deployment catalog columns
- `web/src/features/models/ModelsPage.tsx` — nine per-million call sites
- `web/src/features/organization/RateOverridesCard.tsx` — the organization's own rates

## Why it needs all three at once

`ModelPricingPage.test.tsx` states the invariant in its own comment: "`formatCost` is what Models formats the same stored numbers with, so the two pages cannot print different figures for one price." Fixing one surface breaks that, so a partial change trades a rounding bug for an inconsistency bug.

## Suggested shape

A `formatRate` beside `formatCost` in `shared/helpers/format.ts` with `minimumFractionDigits: 2` and `maximumFractionDigits: 4`, used by every per-million column, leaving `formatCost` to spend. That fixes the rounding and strengthens the shared-formatter invariant rather than weakening it.

`ModelPricingPage.test.tsx`'s "keeps a sub-cent rate legible" case currently pins `$0.08` for an input of `0.075`. That assertion is incidental to the test's stated intent (it protects sub-cent legibility), so extending the precision extends the intent rather than contradicting it, but the literal needs updating.

## Scope note

Raised out of review on #677, where the overrides column adopted `formatCost` for consistency with the catalog. That was the right call for consistency and inherits this defect; it is deliberately not fixed there because the full change is roughly thirteen call sites across three pages and does not belong in a review-fix PR.

njbrake:
## Description

Models is grouped by model: a list of cards with a filter rail, and a page per model listing every provider's offering with its price and where that price came from. Screenshots are in the comment below.

**Decisions**

- **Model ids are vendor-qualified with dotted versions**, `z-ai/glm-5.3-flash`, `openai/gpt-oss-120b`. The name comes from models.dev, else from the provider's id with path, org and version pins removed (Fireworks' `glm-5p3` reads as `glm-5.3`). A model whose vendor is unknown keeps a bare name.
- **The model id is the selector to advertise.** Sending it reaches the model's cheapest offering by the deployment's rates. `instance:<cleaned id>` (`fireworks:glm-5.3`) pins a provider; the full `instance:model` is still accepted. Resolution sits after aliases and static policies, never rewrites a selector that names an offering, keeps the legacy `provider/model` spelling meaning that provider's offering, and relabels the response like an alias. The index rebuilds every minute and on `POST /v1/catalog/selectors/refresh`. Known limit: a bare id resolves before a key's allow-list is checked.
- **The catalog is read-only.** Rates are edited on Model pricing; an org admin sets its own overrides there. The model page is its facts and its providers; "Use this model" opens a drawer with the provider choice, the key step and the request. Pricing roll-ups, per-row detail and compare-at were cut from the UI for the first cut (`?at_context` stays on the API).
- **Pricing provenance and refresh.** Rates carry `unit` and `origin`; `pricing_refresh: manual|review|auto` with a poller, a history of accepted snapshots (newest 30) and `GET /v1/pricing/drift`. All operator-only.
- **`public_catalog`** serves the list and pages to visitors, off by default, on its own rate limit.

**Not built, for follow-ups:** routing with failover across a model's offerings (a bare id picks once, by price); organization keys as named instances; the hosted offerings adapter.

`demo/catalog-smoke/` seeds a platform admin, an org admin and a member for a LAN smoke; whether it belongs upstream is the reviewer's call. Delete the `screenshots/pr-1015` branch after merge.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Closes #700. Refs #998, mozilla-ai/otari-ai#2109, mozilla-ai/otari-ai#2112, mozilla-ai/otari-ai#2095, mozilla-ai/otari-ai#2096, mozilla-ai/otari-ai#1833.

## Checklist

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:** Claude Fable 5.1 via Claude Code.

**Any additional AI details you'd like to share:** Design and scope decided with @njbrake; reviewed by a separate Claude Opus pass before requesting review.

## Summary

- Added a grouped, read-only model catalog with model detail pages and provider offerings.
- Added vendor-qualified model identities, provider pricing, pricing sources, usage data, and selector resolution.
- Added public catalog access, rate limiting, responsive views, and a LAN smoke demo.
- Added precise per-million rate formatting, pricing units, refresh policies, snapshot history, and drift reporting.
- Added catalog and pricing APIs, dashboard workflows, documentation, and tests.
- Moved catalog construction and selector indexing into shared services.
- Replaced the editable Models page with separate catalog and model-detail routes.


njbrake:
Screenshots from the `demo/catalog-smoke` seed (fake keys, real models.dev metadata, 120 seeded usage rows), taken at e1c14962. Images live on the `screenshots/pr-1015` branch; delete it after merge.

**Models: filter rail, search, sort, one card per model (1440px, operator)**
models list

**Filtered to Fireworks and open weights**
models filtered

**A model's page: facts and offerings**
model page

**"Use this model": which provider answers, the key, the request**
use this model

**Public catalog, signed out**
public model page

**Phone, dark**
 

_Note: this comment was drafted by Claude Fable 5.1 via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

tbille:
just did some testing with Anthropic key, nebius and fireworks keys. This is awesome 💙 

I see some minor UI details that can be fixed on follow up PRs.

@peteski22 can we get some 👀 to validate that the architecture is well respected, then let's go for this.

src/gateway/services/catalog_selectors.py
```diff
@@ -0,0 +1,126 @@
...
+    for one the index does not know, and while the index is empty.
+    """
+    index = _index
+    if model_selector in index.full:
+        return None
+    # ``openai/gpt-oss-120b`` is both the legacy spelling of an offering on the
+    # ``openai`` instance and, read as vendor/model, a catalog id. Where the
+    # instance really serves that id, the caller meant the offering.
+    prefix, slash, rest = model_selector.partition("/")
+    if slash and f"{prefix}:{rest}" in index.full:
```
peteski22, line 80: **High.** This exempts the legacy `provider/model` spelling only when that exact instance is configured. A vendor slug (`openai`, `anthropic`, `deepseek`, and others) is the same word as the any-llm provider name, so on a deployment that configures only `azure`, `openai/gpt-4o` resolves to `azure:gpt-4o` and runs on Azure credentials at Azure's price. Reproduced with `build_selector_index` and `resolve_catalog_selector` alone, no routes involved.

This contradicts `docs/models.md` line 21 and this module's own docstring ("a real selector is never rewritten").

Invariant to restore: a selector that names a provider reaches that provider or fails; it is never redirected to another one. Please add the failing unit test for the unconfigured-provider case first, then make it pass. I am not prescribing the shape of the fix.

src/gateway/main.py
```diff
@@ -15,6 +15,7 @@
 
 from gateway.api.deps import set_config
 from gateway.api.main import register_routers
+from gateway.api.routes.catalog import run_selector_index_refresher
```
peteski22, line 18: **Medium, layering.** The app lifespan now imports a route module to start a background task. No other refresher needs that; they all live in `services/`. `rebuild_selector_index` and `run_selector_index_refresher` (`api/routes/catalog.py` 435 to 511) open `create_session()` and hold the "what does this deployment serve" logic, while `services/provider_kwargs.py` on the dispatch path reads an index only the API layer can fill.

The cause is that `build_merged_catalog` already sits in `routes/models.py`, so this extends an existing smell rather than creating one. Two asks:
- Move the merged-catalog build and the index rebuild into a service module, and have both routers call it.
- Add a rule to `scripts/check_architecture.py` so `gateway.main` and `services/` cannot import `api.routes`. ARCHITECTURE.md says the check and the boundary move together, and the check did not catch this.

peteski22 (commented):
Reviewed locally against ARCHITECTURE.md, src/gateway/AGENTS.md, the backend and frontend standards skills, and the review instructions under .github/instructions. The findings were worked through in a conversation between me and Claude Code before posting: Claude did the reading, the checks and the reproduction, and I decided what to raise and how.

**Overall.** The branch follows the documented seam. No new port, no adapter named outside the composition root, the catalog routers registered in the core list with the right gates, money kept decimal in the database, both migrations chained and reversible, and every generated artifact current. The inline comments are where it diverges, most serious first.

**Blocking**
- A legacy `provider/model` selector can be rerouted to a different provider (`catalog_selectors.py`). Reproduced with the pure functions.
- Catalog business logic and a background task live in the API layer, and the app lifespan now imports a route module (`main.py`, `routes/catalog.py`).
- The price lookup order and the tier selection are re-implemented in routes, so the catalog can show a price settlement does not charge (`routes/catalog.py`, `routes/pricing.py`).

**Fix, or state as deliberate**
- The pricing poller runs on every worker with no coordination.
- The public rate limit is per worker and keyed on the raw socket address.
- What `public_catalog` means in hosted mode is untested and undocumented.
- `origin` is written but never read.
- An `if instance == "hosted"` branch in a core route.
- Dashboard: the public catalog is statically imported into the entry chunk; card links swallow modified clicks; the model page swaps to an error banner on a failed refetch; rail controls are under the touch floor.

**Not anchored to a line.** Deleting `ModelsPage.tsx` drops the "Could not list X; models are missing from the list below" notice from Models. It now appears only per provider on Providers. Say whether that is intended.

**On size.** I would have preferred this as a series: move the merged-catalog build into a service first, then identity and the selector index, the pricing columns, history and policy, drift, the grouped API, the public catalog, the dashboard, the demo. Noting it for next time rather than asking for it now.

**Checks run on the head commit**

| Check | Result |
| --- | --- |
| `make lint` (architecture check, Ruff, single Alembic head) | pass |
| `make typecheck` | pass |
| `make openapi-check`, `make postman-check` | pass |
| Backend unit tests for the new modules | 256 passed |
| Catalog, pricing refresh, alias integration tests (Docker) | 98 passed |
| Web Biome, tsc, build, client and route-tree drift | pass, no drift |
| Web vitest over the changed features | 3840 passed |
| Not run | full `make test`, full `pnpm test`, e2e, screenshots |

src/gateway/api/routes/catalog.py
```diff
@@ -0,0 +1,808 @@
...
+        else {}
+    )
+    configured_types = {models_dev_provider_id(config.provider_instance_type(i)) for i in config.providers}
+    seeds: list[OfferingSeed] = []
+    offerings: dict[str, _Offering] = {}
+    for obj in real:
+        seed, metadata, instance, model_id, provider_type = _seed(config, catalog, obj)
+        seeds.append(seed)
+
+        override = resolve_organization_override(overrides, [obj.id], now)
```
peteski22, line 539: **Medium.** `src/gateway/AGENTS.md`: price lookup order belongs to `services/pricing_service.py` and is not duplicated in a route. This block spells override, then deployment row, then default, itself. It also resolves the override on `[obj.id]` only, where settlement (`pricing_service.py` around line 591) also tries the legacy key, so an override stored under `provider/model` is billed but shown here as "deployment" priced.

The page and the invoice should come from the same function. A "price with source" helper in `pricing_service` that both call would do it.

src/gateway/api/routes/catalog.py
```diff
@@ -0,0 +1,808 @@
...
+    if instance == "hosted":
+        return "hosted"
+    return "deployment" if instance in config.providers else "organization"
+
+
+def _first(values: Iterable[str | None]) -> str | None:
+    return next((value for value in values if value), None)
+
+
+def _rates_at_context(pricing: ModelPricingInfo, at_context: int | None) -> tuple[float, float]:
```
peteski22, line 611: `core/metered_pricing.py` already owns tier selection (`_tier_threshold`, `_applicable_tier`). A second copy here means the compare-at rate can drift from what settlement charges. Expose a rate-at-tokens helper from `metered_pricing` and call it.

src/gateway/api/routes/pricing.py
```diff
@@ -205,18 +277,102 @@ async def preview_pricing_refresh(
...
+    A stored row shadows the genai-prices default silently and forever, whether
+    it was a deliberate override or a copy of a then-current default. This is
+    what makes the difference visible: a row that matches the default is a row
+    that could be deleted, and a row far from it is one worth a second look.
+    Tool rows (``otari:``) are per request and have no default to drift from,
+    so they are left out.
+    """
+    now = normalize_effective_at(None)
+    latest_effective = (
+        select(ModelPricing.model_key.label("model_key"), func.max(ModelPricing.effective_at).label("effective_at"))
```
peteski22, line 336: Same concern as the catalog route: "the row in force at now" is re-implemented as a subquery instead of going through the service lookup (`pricing_service._find_by_model_key`), and it skips the legacy-key canonicalization that lookup applies. Reuse the service lookup so drift reports on the row settlement would actually pick.

src/gateway/services/pricing_refresh_service.py
```diff
@@ -210,6 +227,125 @@ async def confirm_price_refresh(session: AsyncSession) -> bool:
...
+async def run_price_update_poller(config: GatewayConfig) -> None:
+    """Check upstream genai-prices on a schedule, forever.
+
+    The policy is read on every tick rather than captured, because it is a
+    runtime setting: an operator who switches from ``manual`` to ``review`` on
+    the dashboard gets the next check without a restart. The interval is
+    config-only and re-read for symmetry. Every error is swallowed and retried
+    on the next tick, for the reason the snapshot refresher gives.
+    """
+    while True:
```
peteski22, line 335: **Medium.** Every worker runs this loop and nothing coordinates them. `poll_price_updates` compares the fetch against this worker's in-memory snapshot, which siblings converge to only on the 30s refresher. Workers that boot together tick together, so under `auto` each can confirm the same snapshot and append its own `PricingSnapshotHistory` row with `accepted_by="schedule"`.

An in-process lock cannot help across processes or replicas. Suggest a conditional `UPDATE` claim on a row before fetching (dialect neutral, and the same "the claim decides" shape `budget_reservation_ledger.py` already uses), and skip the tick when the claim did not land.

src/gateway/services/pricing_refresh_service.py
```diff
@@ -210,6 +227,125 @@ async def confirm_price_refresh(session: AsyncSession) -> bool:
...
+
+
+async def poll_price_updates(session: AsyncSession, policy: str) -> str:
+    """One tick of the scheduled refresh: fetch, then hold or apply per ``policy``.
+
+    Returns what happened, for the log: ``unchanged`` (the fetch matched what is
+    active, and nothing is left pending), ``pending`` (held for review) or
+    ``applied``. A ``manual`` policy never reaches here.
+    """
+    preview = await prepare_price_refresh(session)
```
peteski22, line 316: Under `review`, a snapshot the operator rejected is fetched and re-pended on the next tick, because nothing records the rejected payload. Either remember the rejected hash until upstream changes, or say in the docs that reject means "not now".

src/gateway/api/deps.py
```diff
@@ -475,6 +475,38 @@ async def verify_catalog_reader(
...
+    if session_identity is not None:
+        return None, True
+    if _header_credentials_present(request) or not config.public_catalog:
+        return await verify_api_key_or_master_key(request, db, config)
+    limiter = getattr(request.app.state, "public_catalog_rate_limiter", None)
+    if limiter is not None:
+        # The limiter raises its own 429; the key is the address, since a
+        # visitor has no other identity.
+        limiter.check(request.client.host if request.client is not None else "unknown")
+    return None
```
peteski22, line 507: Two things to state beside `public_catalog_rate_limit_per_minute`. The key is `request.client.host`, and the CLI starts uvicorn without proxy headers, so behind a reverse proxy every visitor shares one bucket and one scraper 429s the whole catalog. And the limiter is in-process, so N workers give N times the configured budget. Either document both, or key on a configured trusted-proxy header.

src/gateway/api/routes/models.py
```diff
@@ -505,6 +505,11 @@ async def _catalog_scope(
     tenant's, and the workspace-scoped rows only where that workspace is theirs
     (otari-ai#1969).
     """
+    # A visitor, while the catalog is public: the deployment's configured
+    # instances and nothing that belongs to a tenant. Not a member of anything,
+    # so no BYO key, no workspace's aliases or policies.
+    if anonymous:
+        return _CatalogScope(allowlist=[f"{instance}:*" for instance in config.providers], reads_workspace_layer=False)
```
peteski22, line 512: In hosted mode `config.providers` is process-global, so an anonymous read with `public_catalog` on lists every shared instance, its type, its discovered models, and whether the provider answered. No tenant rows leak, and a hosted operator may well want a public price list. But the flag is runtime-settable and `test_catalog_endpoint.py` covers standalone only. Please add a hosted-mode test that pins what a visitor sees, and one line in the docs saying what public means there.

alembic/versions/c7e9a1b3d5f7_add_pricing_unit_and_origin.py
```diff
@@ -0,0 +1,50 @@
...
+``input_price_per_million`` has been read four ways: per million tokens for a
+model, per million requests for a gateway-run tool or a moderation call, and per
+image for image generation. Nothing on the row said which, so every reader that
+renders a rate had to know the convention by the key's spelling, and a generic
+table put a tool's ``10000.0`` under a column labeled "$ / 1M". ``unit`` is that
+convention made a column. Existing rows default to ``tokens``; the one family
+the key spelling identifies for certain, the reserved ``otari:`` prefix of a
+gateway-run tool, is backfilled to ``requests``.
+
+``origin`` says which path wrote a deployment price: the config file, the API, or
```
peteski22, line 12: This docstring says `origin` is what lets the config loader tell a stale copy of its own entry from a deliberate rate. `pricing_init_service.py` still keeps any existing row and only stamps `origin="config"` on insert, and nothing reads `origin` anywhere. Either wire the loader's decision to it, or shrink the docstring to what is true today.

src/gateway/api/routes/catalog.py
```diff
@@ -0,0 +1,808 @@
...
+        return [member.wire for member in members]
+    usage = await _usage_by_selector(
+        db,
+        grouped.organization_id,
+        ((member.wire.selector, member.wire.provider, member.model_id) for member in members),
+    )
+    return [member.wire.model_copy(update={"usage_30d": usage.get(member.wire.selector)}) for member in members]
+
+
+def _credential(config: GatewayConfig, instance: str) -> Credential:
```
peteski22, line 600: This branch fires only for one overlay. Reserving the name in config is a seam ("this name is yours"). An `if instance == "hosted"` in a core route is the overlay's knowledge living in core, which is what ARCHITECTURE.md rule 6 guards against; the next hosted feature adds a second branch. Cleaner: the credential label comes from the offering the adapter contributes, so core never spells the word.

src/gateway/api/routes/catalog.py
```diff
@@ -0,0 +1,808 @@
...
+    session_identity: Annotated[TenancyUser | None, Depends(get_session_identity)],
+) -> CatalogModelDetail:
+    """One model and every offering of it this caller may use.
+
+    A model the caller may not see answers 404, the same as one that does not
+    exist, so the route cannot be used to probe the catalog behind an allow-list.
+    A signed-in caller's offerings also carry their organization's own usage of
+    each over the last 30 days.
+    """
+    merged = await _merged_for(db, config, caller, session_identity)
```
peteski22, line 788: The detail read builds and groups the whole merged catalog (full `model_pricing` load, all overrides, models.dev scan) to answer one model. Query count is constant, so this is CPU per page view growing with catalog size. Fine if deliberate; say so here, or narrow the build to the model's selectors.

src/gateway/api/routes/catalog.py
```diff
@@ -0,0 +1,808 @@
...
+    get_session_identity,
+    require_deployment_operator,
+    verify_catalog_reader_or_public,
+)
+from gateway.api.routes.models import (
+    ALIAS_OWNED_BY,
+    MergedCatalog,
+    ModelObject,
+    ModelPricingInfo,
+    _pricing_info,
```
peteski22, line 51: Nit: a private `_pricing_info` imported across route modules. Make it public in `models.py`, or move it with the catalog build when that moves to a service.

src/gateway/services/model_catalog_service.py
```diff
@@ -80,6 +80,19 @@ class _CatalogCache:
 _MISS = object()
 
 
+def cached_models_dev_catalog(config: GatewayConfig) -> dict[str, Any] | None:
+    """The catalog as last fetched, at any age, without fetching or locking.
+
+    For a reader that runs off the request path and must not bind the fetch
```
peteski22, line 86: Nit: `src/gateway/AGENTS.md` says a docstring in `services/` does not name what calls it. "such as the selector index refresher" is a caller. Same at `catalog_selectors.py` lines 15 to 20 ("the catalog route feeds it").

demo/catalog-smoke/README.md
```diff
@@ -0,0 +1,41 @@
+# Catalog smoke
```
peteski22, line 1: This is a LAN demo, not a smoke gate: nothing runs it in CI, and it needs `uv`, `pnpm` and `make dashboard`, unlike `scripts/oss_edition_smoke.py`. Say that in the first line so nobody reads "smoke" as a gate.

Also: `run.sh` prints the master key and uses a fixed password (`smoke-test-1234`, `run.sh` line 87, `seed.py` line 38) on a `0.0.0.0` bind with `public_catalog: true`. Fine for a demo, but the README should say not to leave it running.

web/src/app/App.tsx
```diff
@@ -13,6 +13,8 @@ import {
   publicAuthPath,
 } from "@/features/auth/publicAuthPaths"
 import { AcceptInvitationPage } from "@/features/invitations/AcceptInvitationPage"
+import { PublicCatalogPage } from "@/features/models/PublicCatalogPage"
```
peteski22, line 16: **Medium.** A static import puts the whole Models feature (`ModelCatalogView`, `ModelDetailView`, `UseModelDrawer`, `DataTable`) in the entry chunk. Build output: `index-*.js` at 266 kB carries the catalog copy, while `models.index-*.js` is 0.07 kB and `models._-*.js` is 0.26 kB. Every visitor of every deployment downloads it, including with the default `public_catalog: false`, and the route splitting `routes.test.ts` protects is undone from this side. `lazy(() => import(...))` here, plus `<Suspense>` at the branch below.

web/src/app/App.tsx
```diff
@@ -143,6 +145,15 @@ function DeploymentRoot({ hash }: { hash: string }) {
     )
   }
 
+  // The catalog, where the deployment has opened it to visitors. Only for a
+  // visitor: a signed-in caller reaches the same pages through the router,
+  // priced for their organization. Keyed on the hash so a second model opened
+  // in the tab remounts the view with its own selection.
+  const publicCatalog = publicCatalogPath(hash)
+  if (public_catalog && !isAuthenticated && publicCatalog !== null) {
+    return <PublicCatalogPage key={hash} modelId={publicCatalog.modelId} />
```
peteski22, line 154: The other public roots wrap in `<PublicPageTitle page=...>` (lines 103, 125, 142); this one does not, so a visitor's tab has no title. Same family: `PublicCatalogPage.tsx` line 18 uses `/favicon.svg` where `LoginPageShell` and `HybridLanding` use `${import.meta.env.BASE_URL}favicon.svg`, which breaks under a non-root base.

web/src/features/models/ModelCatalogPage.tsx
```diff
@@ -0,0 +1,699 @@
...
+          {publicView ? (
+            <a href={publicCatalogHref(model.id)} className={titleClass}>
+              {title}
+            </a>
+          ) : (
+            <Link
+              to="/models/$"
+              params={{ _splat: model.id }}
+              className={titleClass}
+              onClick={(event) => {
```
peteski22, line 348: **Medium.** `preventDefault()` runs on every click, so Cmd/Ctrl-click and middle-click no longer open a new tab, and `onOpen` then navigates the current one. `Link` already navigates on a plain click. Drop this handler and keep `onOpen` for the table's `onRowAction`.

web/src/features/models/ModelDetailPage.tsx
```diff
@@ -0,0 +1,625 @@
...
+  const catalog = useCatalog()
+  const selected = useCatalogModel(modelId)
+  const [quantization, setQuantization] = useState("all")
+  const [useModel, setUseModel] = useState(false)
+  const [sort, setSort] = useState<{
+    column: OfferingSortColumn
+    direction: "asc" | "desc"
+  }>({ column: "input", direction: "asc" })
+
+  if (selected.error) return <ErrorBanner error={selected.error} />
```
peteski22, line 354: **Medium.** `[CATALOG]` is invalidated by every provider, alias, policy, settings and pricing mutation, and `staleTime` is 60s, so a failed background refetch replaces a rendered page with a banner (layout-stability.md). Guard on `selected.isError && !model`, and render the banner above the page when data is present.

web/src/features/models/ModelCatalogPage.tsx
```diff
@@ -0,0 +1,699 @@
...
+  children: ReactNode
+}) {
+  const [open, setOpen] = useState(defaultOpen || count > 0)
+  return (
+    <div className="flex flex-col border-b border-border-subtle py-2">
+      <button
+        type="button"
+        aria-expanded={open}
+        onClick={() => setOpen((prev) => !prev)}
+        className="flex min-h-9 w-full items-center justify-between gap-2 text-left text-sm text-foreground"
```
peteski22, line 108: **Medium.** The rail is what a phone opens from "Filters", and its controls are under the 44px touch floor (responsiveness.md): this toggle is `min-h-9`, the radio rows at line 151 are `min-h-8`, and the `Checkbox` rows have no minimum. `RadioList` (line 135) also hand-rolls native radios where `design-system/forms/RadioGroup.tsx` takes the same props. `min-h-11 lg:min-h-8`, use `RadioGroup`, and add `cursor-pointer` to this button.

web/e2e/screenshots/authenticated.spec.ts
```diff
@@ -17,6 +17,12 @@ const WORKSPACE_ROUTES: ReadonlyArray<{
 }> = [
   { route: "/", name: "overview", heading: /overview/i },
   { route: "/models", name: "models", heading: /models/i },
+  {
+    route: "/models/openai/gpt-parity-priced",
+    name: "model-detail",
```
peteski22, line 22: Screenshot entries are owed for each page a PR adds (testing.md). `model-detail` is here, but `public.spec.ts` has no `#/models` or `#/models/<id>` capture for the visitor view. That file already fulfills the bootstrap through `route.fulfill`, so `public_catalog: true` plus a mocked `/api/v1/catalog/models` is enough.

web/src/features/models/ModelCatalogPage.tsx
```diff
@@ -0,0 +1,699 @@
...
+   * Ahead of a session: no organization to price for, and plain hash links
+   * because there is no router to link through.
+   */
+  publicView?: boolean
+  /** A provider instance to start filtered on. */
+  initialProvider?: string
+}) {
+  const catalog = useCatalog()
+
+  const [filters, setFilters] = useState<CatalogFilters>({
```
peteski22, line 485: `initialProvider` is read into state once, so a navigation to `/models?provider=x` while the page is mounted, or a changed param, does not re-filter. `key={providerParam}` on the view in the route, or derive the filter from the param.

web/src/features/models/ModelCatalogPage.tsx
```diff
@@ -0,0 +1,699 @@
...
+            >
+              {railOpen
+                ? "Hide filters"
+                : activeCount > 0
+                  ? `Filters (${activeCount})`
+                  : "Filters"}
+            </Button>
+          </div>
+
+          {catalog.isPending && !catalog.data ? (
```
peteski22, line 628: With `catalog.error` and no data this falls through to "No models yet. Configure a provider..." under the banner, which reads as an empty deployment rather than a failed load. Short-circuit on `catalog.isError && !catalog.data`.

web/src/features/models/ModelDetailPage.tsx
```diff
@@ -0,0 +1,625 @@
...
+  publicView = false,
+}: {
+  modelId: string
+  /** Ahead of a session: nothing to link through, and the rates are the deployment's list. */
+  publicView?: boolean
+}) {
+  const organization = useOrganizationContext(!publicView)
+  const canPrice = !publicView && isDeploymentOperator(organization.data)
+  const canOverride = !publicView && !canPrice && canManage(organization.data)
+  const catalog = useCatalog()
```
peteski22, line 345: The whole list is fetched here for `default_pricing` alone (line 380). A deep link pays for every summary, and a visitor spends a second request per minute against the public rate limit. Carry `default_pricing` on the detail response, or note that this is accepted.

web/src/shared/api/models.ts
```diff
@@ -1,17 +1,52 @@
...
 
+// The catalog folded by model, priced for the caller. Any session may read it,
+// like `/v1/models`; the detail is keyed under the list so a pricing write that
+// invalidates CATALOG takes every open detail with it.
+// Keyed beside the model id so a detail read and a list read never share a
+// cache entry, while both still fall under the CATALOG prefix invalidations use.
+export function useCatalog() {
+  return useQuery({
+    ...NO_RETRY,
+    queryKey: [CATALOG, "list"],
```
peteski22, line 26: Nit: `[CATALOG, "list"]` and `[CATALOG, modelId]` collide for a model whose id is `list`. `[CATALOG, "model", modelId]`.

demo/catalog-smoke/run.sh
```diff
@@ -0,0 +1,116 @@
...
+
+cd "$ROOT"
+# A wide console for the log file: the logger wraps at 80 columns when it is
+# not a terminal, which cuts a verification link in two.
+COLUMNS=400 uv run otari serve --config "$STATE/config.yml" --host 0.0.0.0 --port "$PORT" > "$STATE/gateway.log" 2>&1 &
+GATEWAY=$!
+echo "$GATEWAY" > "$STATE/gateway.pid"
+trap 'kill "$GATEWAY" 2>/dev/null; wait "$GATEWAY" 2>/dev/null' EXIT INT TERM
+
+for _ in $(seq 1 90); do
```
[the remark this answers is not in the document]
daavoo: Fixed in d07fb28: the loop sets `READY` and the script exits non-zero with a timeout message when no health check succeeded.

demo/catalog-smoke/run.sh
```diff
@@ -0,0 +1,116 @@
...
+  curl -sf "http://127.0.0.1:$PORT/health" >/dev/null 2>&1 && break
+  if ! kill -0 "$GATEWAY" 2>/dev/null; then
+    echo "the gateway exited before answering; see $STATE/gateway.log" >&2
+    exit 1
+  fi
+  sleep 1
+done
+
+PASSWORD="${OTARI_SMOKE_PASSWORD:-smoke-test-1234}"
```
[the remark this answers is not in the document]
daavoo: Fixed in d07fb28. `run.sh` draws a random password per `.state/` directory (`chmod 600`, cleared by `--reset`), `OTARI_SMOKE_PASSWORD` still overrides it, and `seed.py` no longer carries a default at all: the variable is required. The README now opens by saying not to leave the demo running and why.

demo/catalog-smoke/seed.py
```diff
@@ -0,0 +1,172 @@
...
+
+def sign_in_works(email: str) -> bool:
+    status, _ = call("POST", "/v1/auth/session", {"email": email, "password": PASSWORD}, auth=False)
+    return status == 200
+
+
+def main() -> None:
+    claim_operator()
+    acme, home = ensure_acme()
+    call("POST", "/v1/organizations/me/switch", {"organization_id": acme})
```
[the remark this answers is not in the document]
daavoo: Fixed in d07fb28: both switches go through a `switch_to` helper and are checked with `must(...)`, so a failed switch stops the seed instead of putting both identities in the previous organization.

src/gateway/api/routes/pricing.py
```diff
@@ -205,18 +277,102 @@ async def preview_pricing_refresh(
...
+            id=str(row.id), accepted_at=row.accepted_at, accepted_by=row.accepted_by, model_count=row.model_count
+        )
+        for row in await list_accepted_snapshots(db, limit=limit)
+    ]
+
+
+@operator_router.get("/drift", response_model=list[PricingDriftRow])
+async def list_pricing_drift(
+    db: Annotated[AsyncSession, Depends(get_db)],
+) -> list[PricingDriftRow]:
```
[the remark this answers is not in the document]
daavoo: Fixed in d07fb28, with the lookup consolidation @peteski22 asked for in the same route: the endpoint takes `limit` (1..500, default 200) and reads through the new `pricing_service.rates_in_force`, which resolves the row in force as settlement would and reports a model stored under both key spellings once.

src/gateway/services/organization_pricing_service.py
```diff
@@ -221,6 +222,8 @@ async def create_for_caller(
             pricing_tiers=override.pricing_tiers,
             effective_from=effective_from,
             effective_to=effective_to,
+            unit=override.unit,
+            origin="api",
```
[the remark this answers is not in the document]
daavoo: Not changed here, deliberately. `OrganizationPricingService` and its route predate this PR, which only adds the `unit`/`origin` writes to them; the transaction boundary is the route's by design there and documented on `organization_pricing._commit`, including the conflict mapping that depends on the service flushing first. Moving commit ownership would change that error mapping, so it belongs in its own change against the whole file rather than riding on a catalog PR.

web/src/features/models/catalog.ts
```diff
@@ -0,0 +1,396 @@
...
+    if (
+      filters.maxInput > 0 &&
+      (model.min_input_price_per_million == null ||
+        model.min_input_price_per_million > filters.maxInput)
+    ) {
+      return false
+    }
+    if (
+      releasedAfter !== null &&
+      (model.release_date == null || model.release_date < releasedAfter)
```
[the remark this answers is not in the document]
daavoo: Fixed in d07fb28: `filterModels` now bounds the window at both ends (`release_date > today` is excluded), with a test that a future-dated model is dropped from every window.

web/src/features/models/ModelCatalogPage.tsx
```diff
@@ -0,0 +1,699 @@
...
+              params={{ _splat: model.id }}
+              className={titleClass}
+              onClick={(event) => {
+                event.preventDefault()
+                onOpen(model.id)
+              }}
+            >
+              {title}
+            </Link>
+          )}
```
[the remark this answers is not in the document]
daavoo: Fixed in d07fb28, the way @peteski22 put it on the same line: the `onClick` is gone entirely. `Link` navigates on a plain click by itself, and `onOpen` stays for the table's `onRowAction`.

web/src/features/models/publicCatalog.ts
```diff
@@ -0,0 +1,26 @@
...
+): { modelId: string | undefined } | null {
+  const path = hash.replace(/^#/, "").split("?")[0] ?? ""
+  if (path === "/models" || path === "/models/") {
+    return { modelId: undefined }
+  }
+  // A model id carries its vendor, `z-ai/glm-5.3`, so the rest of the path is
+  // the id whole.
+  const match = /^\/models\/(.+)$/.exec(path)
+  if (!match) return null
+  return { modelId: decodeURIComponent(match[1] ?? "") }
```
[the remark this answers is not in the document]
daavoo: Fixed in d07fb28: the decode is wrapped and a `URIError` returns `null`, with `#/models/%` as a test case.

web/src/features/models/UseModelDrawer.tsx
```diff
@@ -0,0 +1,216 @@
...
+  const cheapest = model.offerings.find((o) => o.selector === model.resolves_to)
+  // The gateway's own pick is the model id, where the gateway has indexed it;
+  // until then the cheapest offering's selector stands in.
+  const gatewayModel =
+    model.selector ??
+    (model.offerings[0] ? selectorFor(model.offerings[0]) : "")
+  const pinned = model.offerings.find((o) => o.selector === choice)
+  const sendAs = pinned ? selectorFor(pinned) : gatewayModel
+  const options = [
+    { value: GATEWAY_PICKS, label: "Let the gateway choose" },
```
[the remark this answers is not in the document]
daavoo: Fixed in d07fb28: "Let the gateway choose" is only offered when `model.selector` is set. Without one the selection defaults to the first offering, so the request always names the provider it will actually reach.

src/gateway/services/catalog_selectors.py
daavoo: Fixed in d07fb28, and thank you for the reproduction.

The invariant is now enforced where the index is built rather than where it is read. `build_selector_index` takes `OfferingRow`s carrying the instance *and* its provider type, and for a slug whose vendor is also a provider name (any-llm member, a configured instance, or a provider type in the deployment) it considers only that provider's own offerings. So on an azure-only deployment `openai/gpt-4o` is not in the index at all and falls through to the ordinary split, which fails as an unconfigured provider; on a deployment that has both, it reaches the OpenAI offering even when the Azure one is cheaper.

The failing test came first: `test_a_slug_whose_vendor_is_a_provider_resolves_only_on_that_provider` in `tests/unit/test_catalog_selectors.py` is your case (azure-only, then both), and `test_a_real_selector_is_never_rewritten` now asserts `openai/gpt-oss-120b` resolves to nothing rather than to Cerebras. `docs/models.md` states the rule and that a model's `selector` is null where it has no such spelling; the module docstring says the same.

src/gateway/main.py
daavoo: Both done in d07fb28.

`services/merged_catalog_service.py` now owns `build_merged_catalog`, the catalog DTOs (`ModelObject`, `ModelPricingInfo`) and the scope/pricing helpers around it; `services/selector_index_service.py` owns the offering seeds, `rebuild_selector_index` and `run_selector_index_refresher`. `routes/models.py` and `routes/catalog.py` both call the services, and `main.py`'s lifespan imports the refresher from `services/`. `PricingTier` moved to `models/pricing_schemas.py` so a service can carry it without importing a route.

`scripts/check_architecture.py` grows a `FILE_RULES` table (rule 8 in its docstring) with `gateway/main.py` forbidden from importing `gateway.api.routes`; verified it fires on a planted import, and `AGENTS.md` lists it beside the other rules.

src/gateway/api/routes/catalog.py
daavoo: Partly, in d07fb28.

The override lookup bug is fixed at the root: `pricing_service.pricing_key_forms` is now the single statement of which key spellings a lookup offers and in what order, and `find_model_pricing`, the catalog override index and its resolution all go through it, so an override stored under `provider/model` is found here exactly as settlement finds it.

The ladder itself moved out of the route into `merged_catalog_service.viewer_price`, whose docstring names `find_model_pricing` as the authority on the order and says why only the override rung is left to apply: the merged catalog has already folded the stored row and the default in (its phases 2 and 3). What I did not do is make the page and the invoice literally one call, because settlement resolves one model with queries while the catalog prices the whole list from one bulk load; a shared function would have to be the pure selection over already-loaded rows, which is what `viewer_price` plus `pricing_key_forms` amounts to. Happy to go further if you want the bulk path to be the one `find_model_pricing` is built on.

src/gateway/api/routes/catalog.py
daavoo: Fixed in d07fb28: `_rates_at_context` calls `core.metered_pricing.effective_rates`, which already owned tier selection, after normalizing tiers to mappings. No second copy.

src/gateway/api/routes/pricing.py
daavoo: Fixed in d07fb28: the route reads `pricing_service.rates_in_force`, which resolves the rows in force at `as_of` and reports a model stored under both the canonical and the legacy key once, under the key a lookup would return. It also takes a bounded `limit` (CodeRabbit raised the same endpoint).

src/gateway/services/pricing_refresh_service.py
daavoo: Fixed in d07fb28 with the claim shape you suggested. `claim_poll_tick` is one conditional `UPDATE` on a row in `pricing_snapshots` under a reserved source (`genai-prices-poll-claim`), taken only when the stored timestamp is older than one interval; the poller skips the tick when the claim does not land, and the first-ever tick is decided by the insert instead, where losing that race is a sibling's claim. A worker that dies holding the claim costs one interval. Documented in `docs/configuration.md`.

src/gateway/services/pricing_refresh_service.py
daavoo: Documented rather than remembered, which is the smaller of the two options you named. `reject_price_refresh`'s docstring and `docs/configuration.md` now say that reject means "not now": nothing records the rejected payload, so the next check re-offers the same update while upstream still differs from what is active.

src/gateway/api/deps.py
daavoo: Documented, both of them: the dependency's docstring and `docs/configuration.md` now say that the key is the socket address and that the CLI starts uvicorn without proxy headers (so a proxied deployment should throttle at the proxy), and that the counter is per worker, so N workers serve N times the configured number. Keying on a trusted-proxy header would be a new config field, which this PR does not add; happy to open one if you would rather have that than the note.

src/gateway/api/routes/models.py
daavoo: Both added in d07fb28. `tests/integration/test_catalog_endpoint.py::test_a_visitor_in_hosted_mode_reads_the_deployments_own_instances` pins what a visitor sees on `mode: hosted` with `public_catalog: true`: the process-wide instances at the deployment's rates, `credential: deployment`, and no usage. `docs/configuration.md` says the same under "A public catalog". I could not run the integration suite locally (no Docker here), so CI is the first run of that test.

alembic/versions/c7e9a1b3d5f7_add_pricing_unit_and_origin.py
daavoo: Shrunk to what is true, in d07fb28: `origin` is written on insert and reported on a read (Model pricing, beside the drift report), so a config-seeded rate is distinguishable from one an operator set. The docstring no longer claims the config loader decides on it, and says the loader still keeps any existing row.

src/gateway/api/routes/catalog.py
daavoo: Moved to the seam in d07fb28, though core still knows the word once. `HOSTED_OFFERING_INSTANCE` now lives in `core/config.py` beside the reservation that gives it meaning (`RESERVED_PROVIDER_INSTANCE_NAMES`), and `_credential` reads it from there with a docstring saying why the name is readable at all: config refuses it in `providers:`, so an offering carrying it came from an overlay. What I did not do is let the adapter contribute the label, because `Credential` is a wire enum and the value has to be spelled somewhere in the schema; if you want the offering itself to carry it, that is a shape change I would rather do with the hosted offerings adapter than ahead of it.

src/gateway/api/routes/catalog.py
daavoo: Said so, in the route docstring: the whole catalog is built and grouped because the identity a model is found by is a property of the group, so narrowing the build would need the grouping first. Query count is constant and the cost is CPU per page view, growing with catalog size rather than with readers. One of the follow-ups if that stops being acceptable.

src/gateway/api/routes/catalog.py
daavoo: Resolved by the move: it is now `merged_catalog_service.pricing_info`, public, and imported by both route modules from the service rather than across routes.

src/gateway/services/model_catalog_service.py
daavoo: Fixed in d07fb28: that clause is gone from `cached_models_dev_catalog`, and the two new service modules' docstrings name the surfaces ("the flat listing", "the catalog grouped by model") rather than the routes. The route paths inherited by the moved code are gone too.

demo/catalog-smoke/README.md
daavoo: Fixed in d07fb28. The README now opens with "A hand-run demo, not a CI gate: nothing runs this, and unlike `scripts/oss_edition_smoke.py` it wants `uv`, `pnpm` and a dashboard build", followed by a paragraph saying not to leave it running and why (0.0.0.0, public catalog, the printed master key). The fixed password is gone: `run.sh` draws a random one per state directory.

web/src/app/App.tsx
daavoo: Fixed in d07fb28: `PublicCatalogPage` is `lazy()`-imported with a `<Suspense>` fallback at the branch, so the Models feature leaves the entry chunk and a deployment with the default `public_catalog: false` never downloads it.

web/src/app/App.tsx
daavoo: Both fixed in d07fb28: the branch is wrapped in `<PublicPageTitle page="Models">`, and `PublicCatalogPage` reads the mark from `${import.meta.env.BASE_URL}favicon.svg` like `LoginPageShell` and `HybridLanding`.

web/src/features/models/ModelCatalogPage.tsx
daavoo: Fixed in d07fb28: the handler is gone. The `Link` navigates on a plain click by itself, modified and middle clicks open a tab again, and `onOpen` stays for the table's `onRowAction`. CodeRabbit raised the same line.

web/src/features/models/ModelDetailPage.tsx
daavoo: Fixed in d07fb28: the early return is now `selected.isError && !model`, and when data is present the banner renders above the page (under the breadcrumb) with the model still on screen.

web/src/features/models/ModelCatalogPage.tsx
daavoo: Fixed in d07fb28. The group toggle is `min-h-11 lg:min-h-8` with `cursor-pointer`, and the group's content wrapper raises every row in it (checkbox and radio alike) to the same floor, so a control added later inherits it. `RadioList` is now a thin wrapper over `design-system/forms/RadioGroup`; I added a `hideLabel` prop to it, since the rail already prints the group name and would otherwise show each label twice, with a story for it.

web/e2e/screenshots/authenticated.spec.ts
daavoo: Added in d07fb28: `public.spec.ts` now captures `public-catalog` and `public-model-detail`, with the bootstrap fulfilled to `public_catalog: true` and `/v1/catalog/models` stubbed for both shapes, in the same style as the mail and OAuth stubs above them. Not run here: the suite is workflow-dispatch and its baselines are gitignored.

web/src/features/models/ModelCatalogPage.tsx
daavoo: Fixed in d07fb28 with the first of your two: the route passes `key={providerParam}`, so a navigation to `/models?provider=x` or a changed param remounts the view and re-seeds the filter, while the filter stays the reader's to edit from there.

web/src/features/models/ModelCatalogPage.tsx
daavoo: Fixed in d07fb28: `catalog.isError && !catalog.data` short-circuits to "The catalog could not be loaded." above the rows, so the failure no longer reads as an empty deployment.

web/src/features/models/ModelDetailPage.tsx
daavoo: Carried on the detail response, in d07fb28: `CatalogModelDetail` now has `default_pricing`, the page reads it from the model and no longer calls `useCatalog`, so a deep link (and a visitor) costs one request. OpenAPI, the Postman collection and `web/src/client/schema.ts` are regenerated.

web/src/shared/api/models.ts
daavoo: Fixed in d07fb28: `[CATALOG, "model", modelId]`, with the comment above it noting the `list` collision it avoids.

daavoo:
Review addressed in d07fb28. Thread-by-thread replies are inline; three things that had no line to sit on:

**The "Could not list X" notice.** Not intended as a silent drop, and not restored either. The notice said a provider could not be listed, which is a fact about a provider and not about a model, and the grouped catalog has no row to hang it on: a provider that answered nothing contributes no offering to group. Providers is now the one place that state is reported, per instance, with the error. If you would rather the catalog said something too, the honest shape is a banner naming the instances that did not answer, which I would rather add with the `also_available_from` work than fold into this PR.

**`RateOverridesCard` and units** (CodeRabbit, outside the diff). Fixed: the table gains a "Per 1M" column reading the override's own `unit`, the same column and labels Model pricing shows, so a per-request or per-image override is no longer presented under token headings. `UNIT_LABELS` moved to `features/pricing/units.ts` rather than being copied.

**On size.** Taken, and agreed. The service extraction in this commit (`merged_catalog_service`, `selector_index_service`) is the first commit of the series you described, arriving last instead of first; next time it goes first.

**Checks on d07fb28**

| Check | Result |
| --- | --- |
| `make lint` (architecture check, single head, Ruff) | pass |
| `make typecheck` | pass |
| `make openapi-check`, `make postman-check` | pass |
| `uv run pytest tests/unit` | 3281 passed, 3 skipped |
| Web Biome, `tsc -b`, client regenerated from the spec | pass |
| `pnpm --dir web test` | 5995 passed |
| Not run | integration tests (no Docker on this machine, so the new hosted-mode test runs first in CI), e2e, screenshots |

PR #1019, opened 2026-09-10
test: boot one FastAPI app per worker in the integration suite

daavoo:
## Description

The integration job spent about 940s in pytest for 2249 tests, a median of 1.5s per test with no real outliers. Profiling the `client` fixture showed the cost was `create_app` itself: FastAPI re-analyzes 453 route signatures and rebuilds their pydantic adapters on every call, about half a second locally and over a second on a shared CI core. The database reset is 15ms and the lifespan startup under 100ms.

Two commits:

**Boot one app per worker for the shared config.** The `client` fixture builds the app once and only enters the lifespan per test. Before each boot it redoes the part of `create_app` that the lifespan does not reset and that a test could observe across boots: the two rate limiters (the login limiter counts per client IP, and the test client always has the same one) and the DI container (one fixture rebinds a port on it).

**Share the app across equal configs, and close the engine at shutdown.** About 430 tests build a client on a config of their own. Everything `create_app` derives from the config it derives from the config's content, so the cache matches by value against a snapshot taken before the build, and points the app at the caller's config the way `create_app` would have (`app.state.config` and the `set_config` fallback). For that to hold, the lifespan now reads its config from `app.state.config` rather than a closure, which is also what `get_config` hands every request. Two production changes come with it, both small:

- `_create_lifespan()` takes no config; the lifespan reads `app.state.config`.
- The lifespan disposes the engines `init_db` opened. Each boot replaced the module-level engine without closing the previous one, so its pooled connections lived until garbage collection on a closed event loop and held their server slots meanwhile. In production that is process exit; under a parallel suite it showed up as engine disposals taking ten seconds and more.

Full suite locally, same flags:

| | 18 workers, wall | CPU-seconds | 4 workers, wall |
|---|---|---|---|
| main | 251s | 2173 | 316s (measured at first commit) |
| first commit | 100s | 810 | |
| both commits | 109s | 680 | 141s |

The 18-worker wall time stops improving after the first commit because 18 Postgres containers on one Docker VM become the bottleneck; the 4-worker column is closer to the CI runner. All 2238 tests pass; the one failure in every run is `test_mcp_dependency_ceiling`, which installs from PyPI and is the known egress case in AGENTS.md. `make lint`, `make typecheck`, the unit suite, and the OSS-edition smoke script pass.

## PR Type

- [ ] New Feature
- [ ] Bug Fix
- [x] Refactor
- [ ] Documentation
- [x] Infrastructure / CI

## Relevant issues

None. Came out of profiling the `test-integration` job.

## Checklist

## AI Usage

- [ ] No AI was used.

**Any additional AI details you'd like to share:** Profiled from the CI job log timestamps and local cProfile runs; the change and its measurements were reviewed by the author before opening.

## Summary

- Reuse FastAPI apps for integration tests with equivalent configurations.
- Reset application state before each test startup.
- Preserve each test’s configuration references.
- Dispose database engines during application shutdown.
- Route hybrid-mode tests through the shared app helper.
- Update lifecycle tests and integration guidance.

These changes reduce integration test setup time and prevent pooled database connections and stale state from persisting between test runs.


tests/integration/conftest.py
```diff
@@ -281,9 +284,51 @@ def test_config(postgres_url: str) -> GatewayConfig:
...
+    and the test client always has the same one, and a fixture may rebind a
+    port on the container.
+    """
+    app.state.rate_limiter = RateLimiter(config.rate_limit_rpm) if config.rate_limit_rpm is not None else None
+    app.state.login_rate_limiter = (
+        RateLimiter(config.dashboard_login_rate_limit_per_minute)
+        if config.dashboard_login_rate_limit_per_minute is not None
+        else None
+    )
+    app.state.container = build_container(config.bootstrap)
```
khaledosman, line 346: `app.state.inflight` is a third piece of per-app state a test observes across boots, and it is not reset here. `test_in_flight_requests.py` reads the shared app's registry directly: `(entry,) = seen[0]` (line 169) requires the snapshot to hold exactly one entry, and lines 179/188/222/258/298 assert `len(registry) == 0`. On a per-test app those were statements about the test; on the reused app they are statements about every test that ran before it on the worker, and `snapshot()` sorts oldest first, so a stale entry also becomes the one that gets unpacked.

Clear it in place, do not reassign. `InFlightMiddleware` is constructed with `registry=app.state.inflight` at `create_app` time and calls `finish()` on that object, while `begin()` and `GET /v1/usage/in-flight` resolve through `request.app.state.inflight` (`gateway/inflight.py:131`). Assigning a fresh registry would send inserts to the new one and removals to the old, and nothing would ever be dropped. Add a `clear()` to `InFlightRegistry` and call it here.

tests/integration/conftest.py
```diff
@@ -281,9 +284,51 @@ def test_config(postgres_url: str) -> GatewayConfig:
...
+    route's signature and builds its pydantic adapters on each call, which is
+    most of a second on a CI runner, for hundreds of tests. The lifespan still
+    runs per test (``TestClient`` enters it), so the database-backed startup
+    and the cache resets at shutdown are as fresh as they were.
+
+    Keyed on the config's identity, not a session fixture: a module that
+    overrides ``test_config`` per test gets an app per test, as it always did.
+    """
+    global _APP_FOR_CONFIG
+    if _APP_FOR_CONFIG is None or _APP_FOR_CONFIG[0] is not config:
```
khaledosman: One slot means last-write-wins, and two modules evict the shared app on every one of their tests: `test_search_endpoint.py:37` and `test_search_tools_api.py:24` override `test_config` at function scope, 49 tests between them. CI runs `-n auto` with xdist's default `--dist load`, which interleaves modules on a worker, so each of those costs the next shared-config test a full `create_app` rebuild: on the order of 100 extra boots of the kind this change exists to remove.

Pin the shared app instead. `GatewayConfig` is unhashable so a dict cache is out, and a session-scoped `shared_app` that depends on `test_config` raises `ScopeMismatch` in those two modules. Splitting the config fixture works:

```python
@pytest.fixture(scope="session")
def _shared_config(postgres_url: str) -> GatewayConfig: ...  # today's body

@pytest.fixture
def test_config(_shared_config: GatewayConfig) -> GatewayConfig:
    return _shared_config

@pytest.fixture(scope="session")
def shared_app(_shared_config: GatewayConfig) -> FastAPI:
    return create_app(_shared_config)

@pytest.fixture
def client(test_config, shared_app, clean_database):
    app = shared_app if test_config is shared_app.state.config else None
    yield from build_test_client(test_config, app=app)
```

It also drops the mutable module global.

tests/integration/conftest.py
```diff
@@ -325,7 +376,7 @@ async def override_get_db() -> AsyncGenerator[AsyncSession, None]:
 @pytest.fixture
 def client(test_config: GatewayConfig, clean_database: None) -> Generator[TestClient]:
     """Create a test client for the FastAPI app."""
-    yield from build_test_client(test_config)
+    yield from build_test_client(test_config, app=_app_for(test_config))
```
khaledosman: The docstring is still "Create a test client for the FastAPI app", and nothing at the fixture says the app is now shared for the worker's lifetime. That is the invariant the next test author breaks: an `app.state` value the lifespan does not re-set, or an entry added to `client.app.dependency_overrides`, now outlives the test (`build_test_client` never clears the overrides dict; it is harmless today only because `get_db` is the sole entry and each boot reassigns it). Say it here, and alongside the sentence AGENTS.md already carries under Test Notes, "The schema is built once per worker, not once per test".

khaledosman (approved):
Three notes, all in `tests/integration/conftest.py`: one piece of per-app state the refresh misses, the cache evicting the shared app, and the invariant the fixture no longer states.

tests/integration/conftest.py
daavoo: Addressed in ac166bd: InFlightRegistry gained clear(), and _refresh_process_state calls it in place before each boot, for the reason you give (the middleware holds the registry create_app built).

tests/integration/conftest.py
daavoo: Superseded in 2a18925 (now f6812a6): the cache is a list of up to eight (config snapshot, app) pairs matched by value, so the shared app is never evicted by the two modules that override test_config, and those 49 tests now hit the cache too. GatewayConfig being unhashable is why it is a list compared with ==.

tests/integration/conftest.py
daavoo: Addressed in ac166bd: the client fixture docstring now states that the app object outlives the test and only the lifespan is per test, AGENTS.md carries a sentence next to the once-per-worker schema note, and build_test_client assigns the overrides dict instead of adding to it.

tests/integration/conftest.py
```diff
@@ -281,9 +285,67 @@ def test_config(postgres_url: str) -> GatewayConfig:
...
+    """
+    for snapshot, app in _APPS:
+        if snapshot == config:
+            break
+    else:
+        snapshot = config.model_copy(deep=True)
+        app = create_app(config)
+        _APPS.append((snapshot, app))
+        del _APPS[:-_APP_CACHE_SIZE]
+    app.state.config = config
```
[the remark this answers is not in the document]
daavoo: Addressed in ac166bd: oauth_callback_landing was the one route reading config from the create_app closure at request time; it now reads request.app.state.config. The other config reads inside create_app are build-time (docs, CORS, mode, metrics, limiters, bootstrap), and equal content is what the cache matches on.

PR #1026, opened 2026-09-10
feat(api)!: mount the API under /api/v1 and OTLP ingest under /otlp

peteski22:
## Description

Otari's API moves from `/v1/...` to `/api/v1/...`. Health checks move with it. OpenTelemetry ingest moves to its own door at `/otlp/...`. The old addresses stop being served in the same release, so anyone pointing a client, a probe or an exporter at Otari has to change the address they use. The dashboard moves in the same change, because it is served by the gateway and tested against a real one.

Why now: three URL shapes were in use at once and none of them owned the root. Otari's own default control-plane address already ends in `/api/v1`, so a self-hosted gateway that does not override it already expects that shape, and leaving the API at `/v1` meant that default was on a collision course. `/api/v1` is now the single answer for everything Otari defines and versions. Anything belonging to somebody else's protocol gets its own door and keeps its own path shape behind it, which is why telemetry sits at `/otlp` rather than being folded into ours.

What a caller changes:

| Client | After this |
| --- | --- |
| OpenAI-style SDKs | base URL `https://<host>/api/v1` |
| Anthropic-style SDKs and Claude Code | base URL `https://<host>/api` (the SDK appends the rest) |
| OpenTelemetry exporters | `OTEL_EXPORTER_OTLP_ENDPOINT=https://<host>/otlp` |
| Container and platform health probes | `/api/v1/health` |
| Prometheus dashboards and alerts | see below, this one is easy to miss |
| Otari's own SDKs | nothing, once they regenerate |

Two addresses deliberately stay put: the OAuth provider callback, because it is registered with the providers themselves and we cannot move it unilaterally, and the metrics scrape target.

### Operators: your Prometheus series are renamed

The request metrics used to carry the whole path in their `endpoint` label, so moving the API would have silently retired every existing series and started a new one beside it. Panels flatten, and a rate alert spanning the upgrade reads as a traffic collapse rather than a rename.

The root is now a label of its own, so this is the last time these move:

```
before   gateway_requests_total{endpoint="/v1/chat/completions"}
after    gateway_requests_total{endpoint="/chat/completions", api_version="v1"}
```

A dashboard that does not care about versions sums across `api_version` and never notices another move. One that does can now ask. **This break happens once and cannot be avoided**: keeping the old label would hardcode a path that no longer exists and mislabel every route added later.

### Four things that would have broken quietly

None of these announces itself, and none was in the original plan.

**Signing in with Google or GitHub would have stopped working.** The cookie that ties a sign-in to the browser that started it was pinned to the old address, so the browser would simply never send it back. The failure reads as an expired sign-in, not a routing fault. The cookie path and the router prefix now come from one constant, and a test reads the path off the response and compares it against the routing table, so neither side is a literal that can drift.

**Importing usage from the command line would have failed** against any upgraded gateway. The URL was built by string concatenation, so no search for a route would ever have found it.

**Every generated SDK would have lost its typed responses** on chat, messages, rerank, embeddings and images, with nothing to show for it. The generator looks each path up in the spec to attach the response type, and those paths were written out by hand.

**A future sibling route would have been served to shared caches without `no-store`.** The public-path exemption compared bytes rather than path segments, so a later `/api/v1/health-internal` would have inherited it and shipped an authenticated response cacheable by anyone. It fails open, which is the wrong direction.

### The published document names the new addresses

Every description in the spec, and therefore in the Postman collection and the generated SDKs, now names an address that answers. A contract check walks the generated document and fails on any that does not, which is how the endpoints that landed on `main` mid-review were caught.

### Reviewing this

Commits are grouped by area and meant to be read in order. They are not individually green; the branch is.

The one thing worth knowing before reading the diff: three kinds of string in this codebase look identical and need opposite treatment. A **request path** moves. A **usage-log label** is an identifier written to the database and compared by string equality, so it does not move, and neither does the in-flight registry value beside it. A path named in **someone else's contract** is not ours to change. The same test file often contains two of the three on adjacent lines. Every classification here was made by reading what the value feeds, never by matching the string.

---

**BREAKING CHANGE:** the API is served at `/api/v1/*` instead of `/v1/*`, health probes at `/api/v1/health*`, and OTLP ingest at `/otlp/v1/*`. `/v1/*` is no longer served. Set OpenAI-style base URLs to `{origin}/api/v1`, Anthropic-style base URLs to `{origin}/api`, and `OTEL_EXPORTER_OTLP_ENDPOINT` to `{origin}/otlp`. Prometheus request metrics drop the root from their `endpoint` label and gain an `api_version` label, so existing dashboards, recording rules and alerts keyed on the old label need updating. Every generated SDK method named after an operation id is renamed, because those ids derive from the path.

## How to test it locally

Boot the gateway against a throwaway database:

```bash
DB=$(mktemp -d)/dev.db
OTARI_DATABASE_URL="sqlite:///$DB" OTARI_MASTER_KEY=local-dev-master-key \
OTARI_MODEL_DISCOVERY=false OTARI_REQUIRE_PRICING=false \
OTARI_BOOTSTRAP_API_KEY=false OTARI_ENABLE_METRICS=true \
uv run uvicorn gateway.dev:create_dev_app --factory --app-dir src --port 8123
```

(`make dev` does the same if you keep a `.env`.) Then, in another shell:

```bash
# the new addresses answer
for p in /api/v1/health /api/v1/health/readiness /api/v1/bootstrap \
         /api/v1/docs /api/v1/openapi.json /metrics; do
  printf '%-28s %s\n' "$p" "$(curl -s -o /dev/null -w '%{http_code}' localhost:8123$p)"
done

# the old ones are gone
for p in /v1/health /v1/bootstrap /docs /openapi.json; do
  printf '%-28s %s\n' "$p" "$(curl -s -o /dev/null -w '%{http_code}' localhost:8123$p)"
done

# telemetry has its own door, and is not under the API root
curl -s -o /dev/null -w '%{http_code}\n' -X POST localhost:8123/otlp/v1/traces \
  -H 'content-type: application/json' -d '{}'
curl -s -o /dev/null -w '%{http_code}\n' -X POST localhost:8123/api/v1/traces \
  -H 'content-type: application/json' -d '{}'

# the metric labels carry the version separately
curl -s localhost:8123/metrics | grep gateway_requests_total | head -3
```

Expect `200` on the first group, `404` on the second, `401` for telemetry (the route is there and wants a credential) and `404` for telemetry under the API root. The API reference should render at `/api/v1/docs` rather than `/docs`. The metric lines should read `endpoint="/health",api_version="v1"`.

For the dashboard: `make dashboard` then reload the gateway root, or `pnpm --dir web run dev`. Sign in, and check the Keys page hands you a `curl` snippet pointing at `/api/v1/chat/completions`.

Already proven automatically, so you need not eyeball it:

- `make lint`, `make typecheck`, `make test`, `make openapi-check`, `make postman-check`, and the dashboard's own type-check, lint and test suites.
- A contract check that every address the app serves and every address it publishes sits under one of the two roots, that the two frozen addresses are still there, and that the security allowlists name addresses that exist.
- A check that the sign-in cookie reaches the routes that read it, reading the scope off the response rather than off a constant.
- A check that nothing in the published document's prose names an address that moved.
- A check that a near-miss of a public path is not treated as public.
- A check that mount order still puts fixed routes ahead of the catch-alls that would swallow them, and the mode stubs behind the routes an overlay contributes.
- A check that the labels written to usage rows have not moved, and that the metric label carries no root.
- Every one of these was confirmed to fail when the thing it guards is deliberately broken.

The packaged-edition smoke gate is worth a run too, since it walks the whole flow as a subprocess and is not part of `make test`: `uv run --frozen --no-dev python scripts/oss_edition_smoke.py`.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

None. Part of the API prefix alignment series; follows #1005, #1010 and #1012.

## Checklist

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

Claude Code (Opus 5)

The design, the written plan and the review are human. An agent carried out the mechanical edits and ran the checks, and verified the plan's cited line numbers, counts and command recipes against the code before running any of them; six mismatches were corrected in the plan before work started. The four silent breakages above, the decision to stop spelling the address out in either test suite, and the metric label's shape all came out of human review during the work rather than from the plan.

**NOTE:**


tbille (approved):
Pre-approve. please QA

PR #1027, opened 2026-09-10
docs(dashboard): put the settings-place rationale back beside its rules

jigjigjig:
## Description

Restore the settings density rationale beside the CSS rules it explains. Correct the documented form and dense heights to 36px and 32px, distinguish the pager's coarse-pointer sizing from the toolbar and settings viewport sizing, and document both variables needed to reset a subtree to form size.

This changes comments only. The rebase preserves the current heading and action markup.

## PR Type

- [x] Documentation

## Checklist

- [x] Local lint checks passed.
- [x] No behavior changed: CSS and TSX match main after stripping comments and whitespace.
- [ ] API contract changed (not applicable).

## How to test it locally

Run `make lint` and `pnpm --dir web run lint`. Compare both changed files against main after stripping block comments and whitespace; the remaining source is identical. No generated artifacts or new tests are needed for this comment-only change.

Validation: `make lint` passed with the existing Python environment and `UV_NO_SYNC=1`. Biome checked all 630 dashboard files successfully through the installed binary because the pnpm launcher stalled locally.

## AI Usage

Originally written with Claude Opus 5 via Claude Code. Review corrections and rebase by GPT-6 via Codex.


web/src/styles/globals.css
```diff
@@ -2060,12 +2050,16 @@ main {
...
-/* The dense height above is a desktop size, so the phone viewport takes it
-   back off. The `[data-slot="button"]` floor further down cannot reach these:
-   a native `input[type="search"]` and a HeroUI select trigger are not buttons,
-   so without this the toolbar is the one place on the phone layout still
-   handing a finger a 32px target. Same 767px boundary as that floor, so the
-   toolbar's fields and its buttons come up to 44px together. */
+/* The dense height is a desktop size, so the phone viewport takes it back off.
+   44px is the floor everywhere, and `design/motion-and-access.md` makes a place
+   that keeps its desktop height here the one thing needing an argument in
+   writing, so all three take it rather than two of them.
```
khaledosman, line 2056: "so all three take it rather than two of them" sits on a block whose selector lists two, and it is wrong on the axis it names. The pager reaches 44px only through `@media (pointer: coarse)` further up (line 2006), a different query: a narrow non-touch window (`max-width: 767px`, fine pointer) leaves `.otari-pagination` at 32px, and a coarse-pointer tablet above 767px raises the pager while the toolbar and the settings row stay dense.

`foundation.test.ts:1370` pins the two-selector form, so the code is the intended shape and the sentence is the error. This is the same defect the PR exists to remove: a reader trusts the comment over the selector.

Fix: say the pager comes up on the coarse-pointer query above, and keep the `motion-and-access.md` argument for the two places that are actually here.

web/src/styles/globals.css
```diff
@@ -1345,11 +1345,16 @@ textarea.input {
...
+   their controls sit on a table's own header row; a settings row is dense
+   because its control sits in a lane beside the label it explains, and at the
+   form height every row comes out taller than its own sentence, so a column of
+   them reads as a stack of boxes.
+
+   `input[type="search"]` needs its own rule because it is a native element
+   carrying none of HeroUI's classes, and it stays scoped to the toolbar: the
+   only two in the tree are toolbar search boxes, and pulling a bare search
+   input into the family would restyle a control nothing here renders. */
 .otari-toolbar,
```
khaledosman, line 1358: Adjacent and pre-existing, but it is the loudest stale comment left in the exact block this PR audits, so it belongs in this pass: the section banner at line 1291 still says "40px is the form size... 38px is the dense size" against `--field-height: 36px` at 1324 and `32px` on this rule, and it is contradicted by the comment at 1302 ("36 in a form, 32 in a dense place").

38 also does double duty in that same paragraph as HeroUI's own natural height, which is what makes the stale pair read as deliberate rather than as a leftover.

Fix: make the banner 36/32 and drop "a 40px control outgrows it".

web/src/design-system/layout/SettingsGroup.tsx
```diff
@@ -78,10 +78,12 @@ export function SettingsGroup({
...
         )}
-        {/* `otari-settings` is the dense place, the way `otari-toolbar` is:
-            globals.css sizes `.input` and `.select__trigger` inside it, so a
-            row's control is 32px beside its label on a desk and 44px at 16px
-            where it stacks full width on a phone. A row never picks a height. */}
+        {/* `otari-settings` is the dense place, the way `otari-toolbar` is: it
+            declares `--field-height`, which globals.css reads on `.input` and
+            `.select__trigger`, so a row's control is 32px beside its label on
+            a desk and 44px at 16px where it stacks full width on a phone. A
+            row never picks a height; a subtree that needs the form size resets
```
khaledosman, line 85: "resets the variable" is singular, but a place declares two: `--field-height` and `--field-padding-block` (`globals.css:1358-1362`, and `6px` again for `.otari-settings` in the 767px block, whose own comment explains why the padding is part of the size decision rather than incidental).

Resetting `--field-height` alone lands the right height on `.input` and `.select__trigger` but keeps the dense 4px padding, and `globals.css:1302` tells a reader the opposite in as many words: "padding-block alongside the floor is what actually moves them... setting the floor alone changes nothing."

Fix: name both variables.

khaledosman (approved):
Verified the "no rule changed" claim: `git show` on both revisions with comments stripped gives an identical hash, so the emitted stylesheet really is untouched.

Three findings inline. None blocks merge, though the first is a factual error in a comment-only PR, so it is worth fixing before this lands.

One non-blocking note that has no line to sit on: the PR body has an em dash ("worse than no comment, because a reader trusts it over the code &mdash; and a stale comment..."), which AGENTS.md's prose gate covers for PR descriptions. The commit message and every added comment line are clean.

Other gates check out: no route, schema, or docstring touched, so nothing is owed under Generated Artifacts; no `schema.ts` or `routeTree.gen.ts`; no dependency, so no lockfile or `allowBuilds`; no new page, so no screenshot entry; the title is conventional.

web/src/styles/globals.css
jigjigjig: Addressed in 02f69035: the comment now distinguishes the toolbar/settings viewport rule from the pager's coarse-pointer rule.

web/src/styles/globals.css
jigjigjig: Addressed in 02f69035: the banner now says 36px/32px, and the stale 40px explanation is removed.

web/src/design-system/layout/SettingsGroup.tsx
jigjigjig: Addressed in 02f69035: SettingsGroup now names both --field-height and --field-padding-block and says to reset both.

Issue #1031, opened 2026-09-10
One placement and one surface for every create action in the dashboard


jigjigjig:
## Context

Creating something in the dashboard works differently depending on the page, along two axes:

1. **Where the create action sits.** On `/routing`, `/keys`, `/budgets`, `/providers`, `/workspaces` and `/organization/domains` it is the top-right `PageIntro` action. On `/tools/mcp-servers` it is a small button on a card heading in the middle of the page. On `/tools` and `/tools/web-search` there is no trigger at all: the add form is permanently expanded at the foot of a `SettingsGroup`. Three pages (`McpServersPage`, `WorkspaceMembersPage`, `OrganizationBudgetsPage`) still render the deprecated `PageHeader`, which has no action slot, so they cannot place a header action even if they wanted to.
2. **What pressing it opens.** Page-level creates (keys, budgets, routing, providers, org provider keys, workspaces, domains, org members) reveal an inline `<Section>` form under the header and hide the header button while it is open. Card-level creates (MCP servers, rate overrides, org budgets, spend ceilings, set price) open a HeroUI `AlertDialog`. The workspace switcher opens the same create-workspace form in a HeroUI `Modal`. No drawer exists.

Labels drift with the surfaces: `Create key`, `Create budget`, `New policy`, `Add provider`, `Add MCP server` (whose dialog then says `Add server`), `Claim domain`.

An operator learns where to look and what will happen on one page and is wrong on the next. The rule should be: the placement is always the same, and the surface is always the same.

## Decision

Made 2026-09-10 after comparing four directions (centered dialog, right side panel, dedicated route, in-table expanding row) against the same API keys page.

- **Placement:** the heading row of the collection being created into. `PageIntro`'s `action` slot when the page is one collection; the section heading row (right-aligned, same button) when a page holds several. An empty state may repeat the CTA and opens the same dialog. The header button never hides while the form is open.
[...]

PR #1032, opened 2026-09-10
feat(dashboard): add FormDialog, the one surface every create form opens in

jigjigjig:
## Description

Every create flow in the dashboard opens in the same surface now, or will: this
adds the surface, and nothing is migrated onto it yet.

The problem it settles is one an operator hits on their second page. The action
that starts a create sits in a page's top right on `/routing` and halfway down
the page on `tools/mcp-servers`; pressing it opens a modal on
`tools/mcp-servers` and appends a whole new section to the page on `/keys`. Two
inconsistencies, so a page has to be learned rather than recognized.
`FormDialog` is the surface half. The placement and label rules are the other
half, and they are written down in `web/design/actions.md` beside it.

It is built on HeroUI's `Modal` rather than `AlertDialog`, which is the
difference between it and `ConfirmDialog` rather than a detail of it: an alert
interrupts to ask one question, and a form is a place to work. `ConfirmDialog`
stays exactly as it is, for "are you sure".

Three things it owns because call sites kept getting them wrong on their own:

- **The footer never changes height, and the primary never changes width.**
  `isPending` replaces the label with a spinner *in place*. That is the
  component's doing, not the library's: HeroUI's `isPending` draws no spinner
  (it sets `data-pending`, which the stylesheet answers with
  `pointer-events: none` and nothing else), so today's forms swap the label for
  "Creating…" and the button changes width mid-submit.
- **`isDirty` arms a guard in the footer, not a second dialog.** Escape and a
  click outside swap the actions for "Unsaved changes · Keep editing ·
  Discard". A dialog never opens a dialog.
- **The widths live in `globals.css`, not at the call site.** `.modal__dialog`
  is pinned there unlayered (`width: fit-content`, `min-width: min(28rem, …)`),
  which outranks `@layer utilities`: a `w-[32.5rem]` on the dialog compiles,
  lints, ships and loses, and the 440px size is unreachable under the 448px
  floor whatever you write. So the size travels as an inline
  `--form-dialog-width` and the geometry is settled beside the rule it has to
  beat. Both facts were read out of the emitted stylesheet rather than reasoned
  about.

Also here: `design/overlays.md` and `design/DESIGN.md` are corrected, because
they told a reader to reach for the bare `Dialog` shell when a dialog holds a
form. That shell has one importer in the tree, its own story, and no product
call site; it is left in place rather than deleted, since removing it is a
separate decision.

## How to test it locally

    pnpm --dir web run storybook

`Design system → Feedback → FormDialog`, ten stories: one per size, one per
state (submitting, request failed, dirty guard, success step), the scrolling
`lg` body, and a phone sheet on the 414px viewport preset.

Worth eyeballing, because no test asserts geometry: that the primary does not
change width between **Medium** and **Submitting**, that **ScrollingBody**
keeps its header and footer put and grows a hairline under the header only once
content is under it, and that **PhoneSheet** fills the frame.

Measured on the rendered page rather than asserted, for the record: widths 440 /
520 / 640, dialog top 120px, header 20/24/16/24, footer 12/24, title 18px at
weight 550, submit button 105.72px both resting and pending, footer 61px in both
states, phone sheet 414x800 at (0,0) with 16px inputs and 44px controls.

- `pnpm --dir web test` — 5288 pass, 152 files, of which 17 are
  `FormDialog.test.tsx`.
- `pnpm --dir web run lint`, `pnpm --dir web run typecheck`,
  `pnpm --dir web run build` — clean.
- `pnpm --dir web run storybook:build`, then
  `pnpm --dir web exec node .storybook/smoke.mjs` — 376 stories x 2 themes, all
  rendered clean.

Each of the 17 cases was checked by removing the behavior it covers and
watching it fail: deleting the dirty guard fails 4, sending the spinner beside
the label instead of over it fails 1, dropping the Cmd/Ctrl+Enter handler fails
1, swapping `Modal` for `AlertDialog` fails 5, un-disabling the close control
while pending fails 1, moving the error banner below the fields fails 1, and
removing the close control's `aria-label` fails 7. The keyboard case needed
that pass to be honest: written against a single-line input it passed with the
handler deleted, because the browser's own implicit form submission fires
there, so it is written against a textarea and asserts that plain Enter inserts
a newline first.

Nothing about backend behavior changes, so no OpenAPI artifact moves.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [x] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Part of #1031

## Checklist

      17 cases in `web/src/design-system/feedback/FormDialog.test.tsx`, plus ten
      Storybook stories. The dashboard's tests live under `web/src`, next to the
      component, rather than in the repo-root `tests/` the template names.
      The dashboard's counterparts, which is what this diff touches: `make lint`
      does not reach `web/`. Commands and results above.
      The API contract did not change; no route, schema or docstring is touched.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

The design was specified in Paper and read out of that file with the MCP server
rather than off screenshots. Two places the specification and this codebase
disagreed are called out in the review comments on the diff rather than settled
quietly here.

## Summary

- Added the shared `FormDialog` for create and edit forms.
- Added loading, error, success, scrolling, mobile sheet, and unsaved-changes states.
- Added 17 component tests and 10 Storybook stories.
- Updated design documentation and form styling.
- Kept existing flows, `Dialog`, and `ConfirmDialog` unchanged.

### DIFF
--- a/web/design/overlays.md
+++ b/web/design/overlays.md
@@ -75,18 +75,21 @@ says the same about React Aria popovers under "Checks".
 
 ## Dialog
 
-Modal and centered. `Dialog` is the shell; `ConfirmDialog` is its
-specialization for a destructive action, with the two buttons and the error line
-built in.
+Modal and centered. Three of them, and the question sorts them:
 
-Reach for `ConfirmDialog` when the dialog's whole job is "are you sure", which
-includes every delete of a record, and for `Dialog` when it holds a form.
+- **`FormDialog`** when the operator is creating or editing an object. Every
+  create flow in the product, no exceptions. See [feedback.md](feedback.md).
+- **`ConfirmDialog`** when the dialog's whole job is "are you sure", which
+  includes every delete of a record.
+- **`Dialog`** is the bare `AlertDialog` shell the other pattern was built from.
+  It has no call sites; a form wants `FormDialog`, which is a `Modal`, because
+  an alert interrupts to ask one question and a form is a place to work.
 
-Both are controlled only, because a dialog opens from something elsewhere on the
-page (a row's Edit, a toolbar's Add) rather than from a trigger inside itself.
-Both mount their body only while open, which is not an optimization: the body of
-a form dialog holds controlled inputs, and leaving them mounted carries one
-row's draft into the next row's dialog.
+All three are controlled only, because a dialog opens from something elsewhere
+on the page (a row's Edit, a heading row's Create) rather than from a trigger
+inside itself. All three mount their body only while open, which is not an
+optimization: the body of a form dialog holds controlled inputs, and leaving
+them mounted carries one row's draft into the next row's dialog.
 
 `isDismissable` is on by default. Turning it off takes away Escape and the
 outside click, and the only honest reason is unsaved work that would be lost;
--- a/web/design/overlays.md
+++ b/web/design/overlays.md
@@ -11,7 +11,7 @@ Does the operator need to interact with what appears?
       ├── Is it about the control that opened it?
       │    └── Yes -> Popover      (anchored, takes focus, not modal)
       └── Does it want the whole screen's attention?
-           └── Yes -> Dialog       (modal, centered, dismissed deliberately)
+           └── Yes -> Dialog       (modal, dismissed deliberately)
 ```
 
 `Dialog` lives in [feedback.md](feedback.md)'s directory rather than this one,
--- a/web/design/overlays.md
+++ b/web/design/overlays.md
@@ -75,7 +75,7 @@ says the same about React Aria popovers under "Checks".
 
 ## Dialog
 
-Modal and centered. Three of them, and the question sorts them:
+Modal. Three of them, and the question sorts them:
 
 - **`FormDialog`** when the operator is creating or editing an object. Every
   create flow in the product, no exceptions. See [feedback.md](feedback.md).
--- a/web/src/design-system/feedback/FormDialog.stories.tsx
+++ b/web/src/design-system/feedback/FormDialog.stories.tsx
@@ -53,7 +53,6 @@ function KeyFields() {
           { value: "workspace", label: "This workspace" },
           { value: "organization", label: "Whole organization" },
         ]}
-        reserveMessage
       />
     </>
   )
@@ -251,7 +250,9 @@ function ManyFields() {
               prev.map((item, at) => (at === index ? next : item)),
             )
           }
-          reserveMessage
+          // No description, so no reserved line: the line exists for an error to
+          // replace a description in, and eight of them reserved against nothing
+          // added 8 empty rungs to a body that already scrolls.
         />
       ))}
     </>
--- a/web/src/design-system/feedback/FormDialog.stories.tsx
+++ b/web/src/design-system/feedback/FormDialog.stories.tsx
@@ -1,0 +1,300 @@
+import type { Meta, StoryObj } from "@storybook/react-vite"
+import { useState } from "react"
+
+import { Button } from "../actions/Button"
+import { CopyField } from "../actions/CopyField"
+import { Checkbox } from "../forms/Checkbox"
+import { Field } from "../forms/Field"
+import { SecretField } from "../forms/SecretField"
+import { Select } from "../forms/Select"
+import { Tab, TabRow } from "../navigation/TabRow"
+import { FormDialog } from "./FormDialog"
+
+const meta = {
+  title: "Design system/Feedback/FormDialog",
+  component: FormDialog,
+  parameters: { layout: "fullscreen" },
+  args: {
+    isOpen: true,
+    onOpenChange: () => {},
+    title: "New key",
+    description: "The secret is shown once, right after you create it.",
+    submitLabel: "Create key",
+    onSubmit: () => {},
+    isPending: false,
+    children: null,
+  },
+} satisfies Meta<typeof FormDialog>
+
+export default meta
+
+type Story = StoryObj<typeof meta>
+
+/** The two fields every create form has at least one of. */
+function KeyFields() {
+  const [name, setName] = useState("")
+  const [scope, setScope] = useState("workspace")
+  return (
+    <>
+      <Field
+        label="Key name"
+        value={name}
+        onChange={setName}
+        placeholder="checkout-service"
+        description="Lowercase, hyphens, no spaces."
+        autoFocus
+        reserveMessage
+      />
+      <Select
+        label="Scope"
+        value={scope}
+        onChange={setScope}
+        options={[
+          { value: "workspace", label: "This workspace" },
+          { value: "organization", label: "Whole organization" },
+        ]}
+        reserveMessage
+      />
+    </>
+  )
+}
+
+/**
+ * `md`, the default, and the size all but three call sites want: a form of two
+ * to five fields with nothing to choose between.
+ */
+export const Medium: Story = {
+  render: (args) => (
+    <FormDialog {...args}>
+      <KeyFields />
+    </FormDialog>
+  ),
+}
+
+/** `sm`, for one or two fields. Narrower than HeroUI's own floor; see the CSS. */
+export const Small: Story = {
+  args: {
+    size: "sm",
+    title: "Claim domain",
+    description: undefined,
+    submitLabel: "Claim domain",
+  },
+  render: (args) => (
+    <FormDialog {...args}>
+      <DomainField />
+    </FormDialog>
+  ),
+}
+
+function DomainField() {
+  const [domain, setDomain] = useState("mozilla.ai")
+  return (
+    <Field
+      label="Domain"
+      value={domain}
+      onChange={setDomain}
+      description="You will verify it with a DNS record next."
+      autoFocus
+      reserveMessage
+    />
+  )
+}
+
+/** `lg`, the size a form earns by having two shapes or six or more fields. */
+export const LargeWithTabs: Story = {
+  args: {
+    size: "lg",
+    title: "Add provider",
+    description:
+      "Credentials are encrypted with the server secret key and never leave this gateway.",
+    submitLabel: "Add provider",
+    footerStart: (
+      <p className="text-caption">A test request runs before saving.</p>
+    ),
+  },
+  render: (args) => <ProviderForm {...args} />,
+}
+
+function ProviderForm(args: React.ComponentProps<typeof FormDialog>) {
+  const [tab, setTab] = useState("known")
+  const [secret, setSecret] = useState("")
+  const [isDefault, setIsDefault] = useState(false)
+  return (
+    <FormDialog
+      {...args}
+      tabs={
+        <TabRow>
+          <Tab isActive={tab === "known"} onPress={() => setTab("known")}>
+            Known provider
+          </Tab>
+          <Tab isActive={tab === "custom"} onPress={() => setTab("custom")}>
+            Custom endpoint
+          </Tab>
+        </TabRow>
+      }
+    >
+      <SecretField
+        label="API key"
+        value={secret}
+        onChange={setSecret}
+        description="Stored encrypted. Autofill is off for this field."
+        reserveMessage
+      />
+      <Checkbox isSelected={isDefault} onChange={setIsDefault}>
+        Make this the default route for Anthropic models
+      </Checkbox>
+    </FormDialog>
+  )
+}
+
+/**
+ * The mutation is in flight. The primary keeps its resting width, because the
+ * spinner replaces the label in place; Cancel and the close control are
+ * disabled for the same duration, so the footer's height never moves.
+ */
+export const Submitting: Story = {
+  args: { isPending: true },
+  render: (args) => (
+    <FormDialog {...args}>
+      <KeyFields />
+    </FormDialog>
+  ),
+}
+
+/**
+ * The gateway refused. The dialog stays open holding its own fields, and the
+ * banner mounts at the top of the body, where `ConfirmDialog` already puts it.
+ */
+export const RequestFailed: Story = {
+  args: {
+    error: new Error(
+      "Could not claim mozilla.ai. Another organization already verified it.",
+    ),
+  },
+  render: (args) => (
+    <FormDialog {...args}>
+      <KeyFields />
+    </FormDialog>
+  ),
+}
+
+/**
+ * `isDirty`, after Escape or a click outside. The guard is in the footer rather
+ * than in a second dialog: a dialog never opens a dialog.
+ */
+export const DirtyGuard: Story = {
+  args: { isDirty: true },
+  render: (args) => (
+    <FormDialog {...args}>
+      <KeyFields />
+    </FormDialog>
+  ),
+}
+
+/**
+ * The success step is not a prop. The caller swaps the children and the submit
+ * label once the mutation resolves; the frame stays mounted, so the secret
+ * appears where the form was rather than in a second surface.
+ */
+export const SuccessStep: Story = {
+  args: {
+    title: "Key created",
+    description: "Copy it now. Otari stores a hash and cannot show it again.",
+    submitLabel: "Done",
+    footerStart: <Button>Create another</Button>,
+  },
+  render: (args) => (
+    <FormDialog {...args}>
+      <CopyField
+        label="checkout-service"
+        value="otari-sk-9f3a1c77b0e244d1e8a972fc0a3bc5d86e10f4b2"
+      />
+      <p className="text-caption">
+        Owner Léa Fontaine-Whitaker · $500.00 per month · never expires
+      </p>
+    </FormDialog>
+  ),
+}
+
+/**
+ * A body taller than the height budget. The header and the footer stay put and
+ * the body scrolls between them, and the header gains its rule only once
+ * content is under it.
+ */
+export const ScrollingBody: Story = {
+  args: {
+    size: "lg",
+    title: "Add MCP server",
+    description: "Otari calls it on the operator's behalf.",
+    submitLabel: "Add server",
+  },
+  render: (args) => (
+    <FormDialog {...args}>
+      <ManyFields />
+    </FormDialog>
+  ),
+}
+
+function ManyFields() {
+  const [values, setValues] = useState<string[]>(() => Array(8).fill(""))
+  return (
+    <>
+      {values.map((value, index) => (
+        <Field
+          // The list is fixed-length and positional, so the index is the
+          // identity: there is nothing to reorder and nothing to key on.
+          key={index}
+          label={`Header ${index + 1}`}
+          value={value}
+          onChange={(next) =>
+            setValues((prev) =>
+              prev.map((item, at) => (at === index ? next : item)),
+            )
+          }
+          reserveMessage
+        />
+      ))}
+    </>
+  )
+}
+
+/**
+ * Below 640px every size is a full-screen sheet: the dialog fills the frame,
+ * the inputs are 16px so iOS does not zoom, and the footer clears the home
+ * indicator.
+ *
+ * `mobile2` is Storybook's own 414px preset, the nearest one to the 390px frame
+ * this was drawn at. The width is what matters: the sheet is a media query on
+ * the preview frame, so a wrapper `<div>` would not produce it. The dialog
+ * portals to the frame's `<body>` and never sees a wrapper at all.
+ */
+export const PhoneSheet: Story = {
+  globals: { viewport: { value: "mobile2", isRotated: false } },
+  render: (args) => (
+    <FormDialog {...args}>
+      <KeyFields />
+    </FormDialog>
+  ),
+}
+
+/** Driven from a trigger, which is how a heading row actually opens it. */
+export const FromTrigger: Story = {
+  render: (args) => {
+    const [open, setOpen] = useState(false)
+    return (
+      <div className="p-6">
+        <Button variant="primary" onPress={() => setOpen(true)}>
+          Create key
+        </Button>
+        <FormDialog
+          {...args}
+          isOpen={open}
+          onOpenChange={setOpen}
+          onSubmit={() => setOpen(false)}
+        >
+          <KeyFields />
+        </FormDialog>
+      </div>
+    )
+  },
+}
--- a/web/src/design-system/feedback/FormDialog.stories.tsx
+++ b/web/src/design-system/feedback/FormDialog.stories.tsx
@@ -53,6 +53,7 @@ function KeyFields() {
           { value: "workspace", label: "This workspace" },
           { value: "organization", label: "Whole organization" },
         ]}
+        reserveMessage={false}
       />
     </>
   )
@@ -250,9 +251,16 @@ function ManyFields() {
               prev.map((item, at) => (at === index ? next : item)),
             )
           }
-          // No description, so no reserved line: the line exists for an error to
-          // replace a description in, and eight of them reserved against nothing
-          // added 8 empty rungs to a body that already scrolls.
+          // No description, so no reserved line: the line exists for an error
+          // to replace a description in, and eight of them reserved against
+          // nothing added eight empty rungs to a body that already scrolls.
+          //
+          // Spelled `={false}` rather than omitted, which reads as redundant
+          // and is not: `FieldMessages` defaults `reserve` to true and `Field`
+          // forwards the prop undefined, so a field that says nothing reserves
+          // a line anyway. Measured, omitting it leaves the row at 83px where
+          // this brings it to 60.
+          reserveMessage={false}
         />
       ))}
     </>
--- a/web/src/design-system/feedback/FormDialog.tsx
+++ b/web/src/design-system/feedback/FormDialog.tsx
@@ -1,5 +1,5 @@
 import { Modal, Spinner } from "@heroui/react"
-import { type ReactNode, useEffect, useState } from "react"
+import { type ReactNode, useEffect, useRef, useState } from "react"
 import { FiX } from "react-icons/fi"
 
 import { Button } from "../actions/Button"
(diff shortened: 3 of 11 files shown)


web/src/design-system/feedback/FormDialog.tsx
```diff
@@ -0,0 +1,259 @@
...
+          return
+        }
+        requestClose()
+      }}
+    >
+      {/* HeroUI renders a press responder for the trigger slot and warns when
+          nothing fills it. These dialogs are driven from state, so the slot is
+          hidden rather than absent; `WorkspaceSwitcher` does the same. */}
+      <Modal.Trigger className="hidden">{submitLabel}</Modal.Trigger>
+      {/* No `bg-backdrop/50` here, despite the two dialogs that carry it:
```
jigjigjig: The spec I built from says to carry the two Modal fixes `WorkspaceSwitcher` already has, and this is me deliberately not carrying one of them.

`bg-backdrop/50` is a no-op. globals.css dims `.alert-dialog__backdrop--opaque, .modal__backdrop--opaque, .drawer__backdrop--opaque` to 50% in an **unlayered** rule, and the utility is in `@layer utilities`, so the unlayered rule wins whatever the call site says. Read out of the built stylesheet: the override sits at offset 472541 with an empty layer stack, `.bg-backdrop\/50` at 431637 inside `layer:utilities`. Measured on the rendered page, the backdrop paints `color(srgb 0 0 0 / 0.5)` with the class absent.

Both spellings land on the same value, which is why nobody noticed. The class on `WorkspaceSwitcher` is dead in the same way and I have not touched it here; happy to file it.

web/src/design-system/feedback/FormDialog.tsx
```diff
@@ -0,0 +1,259 @@
...
+          globals.css dims `.modal__backdrop--opaque` to 50% in an unlayered
+          rule, which outranks the utility. Measured in the built stylesheet,
+          the class changes nothing. */}
+      <Modal.Backdrop isDismissable={!isPending}>
+        <Modal.Container
+          placement="top"
```
jigjigjig, line 124: `placement="top"` plus a 120px pad is my reading of a line in the spec that reads two ways: "placement center, top offset 120px on >=md".

Centered and offset 120 from the top cannot both hold, so I took the offset as the operative half, which is also the behavior a growing form wants: fields appearing push the dialog down rather than creeping it up the screen. Measured, the dialog's top edge is at 120px on a 1280x900 viewport. One line to change if the other reading was meant.

web/src/design-system/feedback/FormDialog.tsx
```diff
@@ -0,0 +1,259 @@
...
+          // 120px from the top on a desktop viewport: a dialog optically
+          // centered sits high, and a form that grows as fields appear should
+          // grow downward rather than creep up the screen. Below `sm` the
+          // container is the sheet's own frame and takes no padding.
+          className="p-0 sm:px-4 sm:pt-[7.5rem] sm:pb-10"
+        >
+          {/* No edge spelled here. `globals.css` gives every floating surface
```
jigjigjig: Worth stopping at. The spec draws the frame with `border-border` at 1px, and I am not spelling any edge here, because the product already ruled this the other way and its rule wins.

globals.css gives every floating surface (`.popover`, `.tooltip`, `.toast`, `.modal__dialog`, `.alert-dialog__dialog`, the six popovers) `border: 1px solid var(--color-control-border)` in an unlayered rule, and its comment argues the tier by name: the 0.06-alpha divider tier is for separating two surfaces of nearly the same value, which is the job it fails at on a floating one.

So `border border-border` here was a silent no-op, which I only found by measuring: the frame computed `rgb(90, 95, 107)` (`--color-control-border`) while a probe div carrying the identical classes, appended inside the same dialog, computed `rgba(0, 0, 0, 0.06)`. I removed the dead classes rather than leaving a call site claiming something it does not do.

The header, footer and tab dividers **are** `--color-border`, as drawn: measured `1px rgba(0, 0, 0, 0.06)` on each.

web/src/design-system/feedback/FormDialog.tsx
```diff
@@ -0,0 +1,259 @@
...
+              compile, lint, ship and lose to it, and it argues against that
+              tier by name. */}
+          <Modal.Dialog
+            className="otari-form-dialog flex flex-col p-0"
+            style={
+              { "--form-dialog-width": WIDTH[size] } as React.CSSProperties
+            }
+          >
+            <header
+              className={`flex shrink-0 items-start justify-between gap-4 px-6 pt-5 ${
```
jigjigjig, line 140: `pt-5` (20px), not the 24px the spec's ANATOMY block names.

The spec disagrees with itself here. Its prose says "Header 24 / 24 / 16 padding"; every dialog actually built in the file, all five of them across both artboards, is `padding-top: 20px` with `padding-inline: 24px` and `padding-bottom: 16px`, and the body under it opens at `padding-top: 4px`. I took the built frames, since they are consistent five ways and the prose is one line. Trivial to flip if the prose is the ruled one.

web/src/design-system/feedback/FormDialog.tsx
```diff
@@ -0,0 +1,259 @@
...
+                <FiX aria-hidden className="text-muted size-3.5" />
+              </Button>
+            </header>
+            {tabs ? (
+              <div className="border-border shrink-0 border-b px-6 pt-4 pb-3">
+                {tabs}
+              </div>
+            ) : null}
+            <form
+              // A real form, so a label associates, a password manager sees a
```
jigjigjig, line 174: A real `<form>` rather than a keydown handler alone, which gives one gesture the spec does not name: plain Enter from a single-line field submits, the way it does in every other form on the web.

Cmd/Ctrl+Enter is there too and is the only gesture that works from a textarea, where Enter belongs to the control. The test asserts that split explicitly, because a Cmd+Enter case written against a single-line input passes with the handler deleted.

jigjigjig (commented):
Self-review. The component and its docs only; no call site is migrated, and the 15 of them land per page on top of this.

Five decisions below are places I departed from the specification I was building to, or where the specification and this codebase disagreed. Each says what I did and what it would take to reverse it. The two that matter are the frame's edge color and the header's top padding.

One thing that is **not** covered by a test and needs eyes: the geometry. The numbers in the description were measured on the rendered page in headless Chromium, not asserted in CI, so a later change to `.modal__dialog` or to the overlay border rule would move them silently.

web/design/overlays.md
```diff
@@ -75,18 +75,20 @@ says the same about React Aria popovers under "Checks".
...
-page (a row's Edit, a toolbar's Add) rather than from a trigger inside itself.
-Both mount their body only while open, which is not an optimization: the body of
-a form dialog holds controlled inputs, and leaving them mounted carries one
-row's draft into the next row's dialog.
+Modal and centered. Three of them, and the question sorts them:
+
+- **`FormDialog`** when the operator is creating or editing an object. Every
+  create flow in the product, no exceptions. See [feedback.md](feedback.md).
+- **`ConfirmDialog`** when the dialog's whole job is "are you sure".
+- **`Dialog`** is the bare `AlertDialog` shell the other pattern was built from.
```
jigjigjig, line 84: Beyond the brief I was given, and worth a decision rather than my judgment.

`design-system/feedback/Dialog.tsx` is a shell built on `AlertDialog` whose docstring and this file both said "reach for this one when it holds a form". It has exactly one importer in the tree, its own story, and no product call site; nine feature files still hand-roll `AlertDialog` or `Modal` instead. Since `FormDialog` is now the answer for a form and is a `Modal` by design, that sentence would have become false in two places, so I corrected both rather than leaving a stale copy where the next reader would believe it.

I have **not** deleted the shell. Nine hand-rolled dialogs may still want it, and removing a component is not mine to decide here.

web/src/design-system/feedback/FormDialog.tsx
```diff
@@ -0,0 +1,259 @@
...
+              // fields where Enter is a newline.
+              onSubmit={(event) => {
+                event.preventDefault()
+                if (!isPending) onSubmit()
+              }}
+              onKeyDown={(event) => {
+                if (event.key !== "Enter") return
+                if (!event.metaKey && !event.ctrlKey) return
+                event.preventDefault()
+                if (!isPending) onSubmit()
```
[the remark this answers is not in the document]
jigjigjig: Fixed in 10ededbd, and this thread was marked resolved a commit early: 9d13df8b changed the pending button, not the shortcut, so the direct `onSubmit()` was still there when the bot closed it.

Taking the suggestion as written, `event.currentTarget.requestSubmit()`. The reasoning holds beyond validation: it made the shortcut a second path to the mutation where the button had one, and two paths drift.

Covered by a case that required a textarea, pressed Cmd+Enter against it empty, and asserted the callback did not fire, then typed one character and asserted it did. Reverting the line to `onSubmit()` fails exactly that case and nothing else.

web/src/design-system/feedback/FormDialog.tsx
```diff
@@ -0,0 +1,259 @@
...
+                        variant="primary"
+                        isPending={isPending}
+                        className="relative"
+                      >
+                        <span className={isPending ? "invisible" : undefined}>
+                          {submitLabel}
+                        </span>
+                        {isPending ? (
+                          <span className="absolute inset-0 flex items-center justify-center">
+                            <Spinner size="sm" aria-hidden="true" />
```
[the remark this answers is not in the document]
jigjigjig: Right on both halves, and already fixed in 9d13df8b, which crossed with this review.

The label is `opacity-0` now, for exactly the reason given: `visibility: hidden` takes it out of the accessibility tree and the spinner is `aria-hidden`, so the button had no accessible name mid-submit.

One correction to the remedy. `aria-busy` on the button does not reach the DOM: react-aria filters unrecognized ARIA off `Button`, so the attribute renders as `null`. I asserted it, watched the assertion fail, and moved it to the `<form>`. The test now asserts it there and says why in a comment.

The button also no longer takes `isPending` at all. That prop paints it at the product`&#39;`s disabled 0.4, which is the treatment for *denied*, and a submit in flight is working; it keeps its fill and blocks its own press instead. Which surfaced a third thing neither of us had: HeroUI`&#39;`s `Spinner` defaults to `accent`, so it was drawing teal on the teal fill and was invisible until the 0.4 came off. It is `color="current"` now, measured white on the fill.

web/src/design-system/feedback/FormDialog.stories.tsx
```diff
@@ -0,0 +1,300 @@
...
+    footerStart: (
+      <p className="text-caption">A test request runs before saving.</p>
+    ),
+  },
+  render: (args) => <ProviderForm {...args} />,
+}
+
+function ProviderForm(args: React.ComponentProps<typeof FormDialog>) {
+  const [tab, setTab] = useState("known")
+  const [secret, setSecret] = useState("")
```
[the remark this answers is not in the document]
jigjigjig: Not taking this one, and leaving the thread open for a human rather than resolving a disagreement myself.

There is no API key here to store. `ProviderForm` is a Storybook fixture whose state starts at `""` and is only ever typed into by whoever is looking at the catalog; nothing is committed, sent or logged. `SecretField`'s signature is `{ label, value, onChange, … }`, so a controlled `useState` is the only way to render the control at all, and a "display-only redacted value" would demonstrate a component the product does not have.

The existing convention agrees: `SecretField.stories.tsx` on main opens with `useState("sk-proj-4f8a2c9e1b7d3a6f5e0c8b2d")`, a fabricated string in the same shape.

I did check the adjacent risk this points at, which is a fake credential in a story tripping a scanner: there is no secret-scanning workflow in `.github/workflows`, and the one literal I added (`otari-sk-…` in the SuccessStep story) is fabricated and matches no real key format the gateway issues.

If the guideline behind this is that a `SecretField` should never be controlled in a story at all, that is a design-system decision rather than a change to this file, and it would apply to `SecretField.stories.tsx` first.

web/src/design-system/feedback/FormDialog.tsx
jigjigjig: Settled, and in favor of what this call site already does: the spec rules for the offset.

Artboard 06's ANATOMY block now reads `placement top, 120px offset on ≥md so a growing form pushes down instead of creeping up`. The "center" half of the line this thread was reading two ways is gone, and the rationale the spec gives for the offset is the one this comment gives. The sheet's AS BUILT paragraph cites this PR and records `top 120`, so a later reader can see which came first.

No change: `sm:pt-[7.5rem]` stays.

web/src/design-system/feedback/FormDialog.tsx
jigjigjig: Settled in favor of the built frames, which is to say in favor of this code.

The ANATOMY block now reads `Header 20 / 24 / 16 padding, body opens at pt 4 (matches every built frame on this sheet)`. The 24px prose this thread was arguing with is gone, and AS BUILT records `header 20/24/16/24`.

I checked the frames rather than trusting the prose's claim about them. All five dialog heads on the sheet compute `padding-top: 20px`, `padding-inline: 24px`, `padding-bottom: 16px`, and the `lg` one omits the bottom pad because a TabRow sits under it, which is exactly what `tabs ? "" : "pb-4"` does here. The body opens at `padding-top: 4px`, and all five footers are `padding-block: 12px` / `padding-inline: 24px`.

No change: `pt-5` stays.

web/src/design-system/feedback/FormDialog.tsx
jigjigjig: Confirmed, and settled the same way: the spec no longer draws this edge at the call site.

The ANATOMY block now spells it as `1px edge from the product's shared floating-surface rule (control-border, the same edge every popover carries)`.

I verified the mechanism at the source rather than taking the measurement on faith. In this branch's `globals.css`, `.modal__dialog` takes `border: 1px solid var(--color-control-border)` at line 1657, and a brace-depth scan puts that rule at depth 0, unlayered, so it outranks `@layer utilities` and a call-site `border-border` could not have applied. Removing the dead classes was right.

One related thing this turned up, which is not a question about this file: the canvas still draws the `md` frame with `border-color: var(--color-border)` while its own prose says control-border. That is a stale node on the sheet, and it is being corrected there.

web/src/design-system/feedback/FormDialog.tsx
jigjigjig: The spec names the gesture now, so this is no longer an addition beyond it.

The ANATOMY block reads `Enter submits from a single-line field, Cmd/Ctrl+Enter from anywhere including a textarea`. A real `<form>` is what delivers both halves rather than one, and the split the test asserts is the reason to keep it that way.

No change.

web/src/design-system/feedback/FormDialog.tsx
jigjigjig: Already filed, so the offer in this comment is closed without a new issue: #1033 covers this dead class and lists both `WorkspaceSwitcher` call sites, the create-workspace modal and the create-organization one.

The spec agrees with the omission here too. The ANATOMY block now reads `backdrop 50% (the product's .modal__backdrop--opaque rule)`, naming the rule rather than a class a call site can spell.

Verified at the source: the `.alert-dialog__backdrop--opaque, .modal__backdrop--opaque, .drawer__backdrop--opaque` rule sits at line 2197 of this branch's `globals.css` at brace depth 0, unlayered. It mixes `--color-backdrop`, which is `#000000` here, to 50%, so it lands at `rgba(0, 0, 0, 0.5)` and agrees with the measurement in this comment.

#1033 also raises whether that rule should stay unlayered, which is the right place for that question rather than this thread.

web/design/overlays.md
jigjigjig: Ruled, and not deleting it here was correct: the shell stays until the last PR of this series, which removes it once the five hand-rolled `AlertDialog` forms (MCP server, rate override, org budget, spend ceiling, set price) have moved to `FormDialog`. That is already on #1031's checklist, which reads `delete the unused design-system/feedback/Dialog.tsx shell once nothing hand-rolls a form dialog`, so this needed no separate decision.

Correcting the sentence in both places was right too, and it turned out the claim lived in three more places than the one this thread found. The section opened `Modal and centered.` above three dialogs, one of which is placement top by ruling, so this PR had made its own first two words false. Fixed in ca04f170. A repo-wide grep then turned up two more: the decision tree at the top of this file, where `Dialog` is the category rather than the component, and the same sentence duplicated in `Popover.tsx` and `Popover.stories.tsx`. Fixed in 45a5fe1a, since the contrast a popover draws is modal against not-modal and "centered" was never what carried it.

What the grep leaves standing is all correct: eleven `placement="center"` call sites that really are centered (`ConfirmDialog`, the `Dialog` shell, and the nine hand-rolled dialogs), Storybook's own `layout: "centered"` parameters, and the rationale comment in this component that argues against centering by name.

web/src/design-system/feedback/FormDialog.tsx
```diff
@@ -0,0 +1,288 @@
...
+              </div>
+            ) : null}
+            <form
+              // A real form, so a label associates, a password manager sees a
+              // submit, and Enter in a single-line field submits the way it
+              // does everywhere else. Cmd/Ctrl+Enter is added on top, for the
+              // fields where Enter is a newline.
+              onSubmit={(event) => {
+                event.preventDefault()
+                if (!isPending) onSubmit()
```
khaledosman: The dirty guard does not block submission. With `isDirty`, pressing Close and then Enter in a field calls `onSubmit`, and Cmd+Enter calls it again (probed on this commit: 1 call, then 2). The mutation runs from a footer whose only actions are Keep editing and Discard, and `isPending` never becomes visible because the guard owns the footer for its whole duration.

Guard both paths on `isGuarding` as well: `if (!isPending && !isGuarding) onSubmit()` here, and the same on line 193.

web/src/design-system/feedback/FormDialog.tsx
```diff
@@ -0,0 +1,288 @@
...
+          // grow downward rather than creep up the screen. Below `sm` the
+          // container is the sheet's own frame and takes no padding.
+          className="p-0 sm:px-4 sm:pt-[7.5rem] sm:pb-10"
+        >
+          {/* No edge spelled here: globals.css gives every floating surface one
+              opaque control-border hairline, unlayered, and argues that tier by
+              name, so a `border-border` here would lose to it. */}
+          <Modal.Dialog
+            className="otari-form-dialog flex flex-col p-0"
+            style={
```
khaledosman: `frontend-standards/SKILL.md`'s Never list: "Inline `style={{}}` or `<style>` tags. Use Tailwind utilities or a token", with the `index.html` pre-paint block as the one exception.

The cascade argument in the docstring is about `width:` losing to the unlayered `.modal__dialog`; it does not reach *setting* `--form-dialog-width`, which no other rule declares, so a class has nothing to lose to. Three modifier rules beside the base one and above the phone block (`.otari-form-dialog--sm { width: 27.5rem }`, and so on) keep the geometry in `globals.css` with the rest of it and drop the `as React.CSSProperties` cast with them; `className="[--form-dialog-width:32.5rem]"` is the call-site form if you prefer the value visible there. components.md rung 1 sanctions the variable on a component's own root, not the inline delivery.

web/src/styles/globals.css
```diff
@@ -2133,6 +2133,54 @@ main {
...
+ *
+ * `--visual-viewport-height` is HeroUI's own, already on the container, so the
+ * height budget follows a phone's keyboard rather than the layout viewport.
+ * 15rem is the 120px offset the dialog sits at, twice: what is left is where
+ * the body starts scrolling and the header and footer stay put. */
+.otari-form-dialog {
+  width: var(--form-dialog-width);
+  min-width: 0;
+  max-width: calc(100vw - 2rem);
+  max-height: calc(var(--visual-viewport-height, 100dvh) - 15rem);
```
khaledosman, line 2152: `- 15rem` disagrees with the container's own `sm:pt-[7.5rem] sm:pb-10` (10rem). The dialog caps 80px above where the container would let it grow, so `sm:pb-10` never binds and on a 900px viewport the body starts scrolling at 660px instead of 740px.

Pick one: `- 10rem` here, or `sm:pb-[7.5rem]` on `Modal.Container` if the symmetric 120px gap the comment describes ("the 120px offset ... twice") is the intent.

web/src/styles/globals.css
```diff
@@ -2133,6 +2133,54 @@ main {
...
+@media (max-width: 639px) {
+  .otari-form-dialog {
+    width: 100%;
+    max-width: none;
+    height: var(--visual-viewport-height, 100dvh);
+    max-height: none;
+    border: 0;
+  }
+  .otari-form-dialog > form > footer {
+    padding-bottom: calc(0.75rem + env(safe-area-inset-bottom));
```
khaledosman: `env(safe-area-inset-bottom)` is 0 unless the viewport is `viewport-fit=cover`, and `index.html:6` is `width=device-width, initial-scale=1.0` deliberately (its own comment: "the shell has no safe-area insets to sit out of the way"). So this reduces to `padding-bottom: 0.75rem` and the footer does not clear the home indicator as the comment claims.

Drop the `env()` and the claim, or make it real with the shell change. `AppShell.tsx:885`'s `pb-[env(safe-area-inset-bottom)]` is inert for the same reason.

web/src/design-system/feedback/FormDialog.tsx
```diff
@@ -0,0 +1,288 @@
...
+              {/* `min-h-0` above and here is what lets this scroll rather than
+                  push the footer off the viewport: a flex child's default
+                  `min-height: auto` refuses to shrink below its content. */}
+              <div
+                onScroll={(event) =>
+                  setIsScrolled(event.currentTarget.scrollTop > 0)
+                }
+                className="flex min-h-0 flex-col gap-4 overflow-y-auto px-6 pt-1 pb-6"
+              >
+                <ErrorBanner error={error} />
```
khaledosman, line 214: In a scrolled `lg` body (the ScrollingBody story's shape) a failed submit mounts the banner above the fold: the AT path is covered by `role="alert"`, but a sighted operator sees the submit finish and nothing change. Reset the body's `scrollTop` when `error` turns truthy.

khaledosman (approved):
Reviewed against AGENTS.md, web/AGENTS.md, frontend-standards (SKILL plus components.md, motion-and-access.md, responsiveness.md), and the three `.github/instructions/` files. No backend, route, schema, or generated artifact is touched, so no OpenAPI/Postman/schema.ts work is owed. Typecheck, Biome and the full Vitest suite are green on 7dc6aa2 (152 files, 5290 tests).

One confirmed defect (the dirty guard does not block Enter/Cmd+Enter) and one standards violation (inline `style`), both one-line fixes. Nothing is migrated onto this surface yet, so neither is blocking.

AGENTS.md's prose rule names PR descriptions: three em dashes as separators in the body (the three `pnpm` result lines).

web/src/design-system/feedback/FormDialog.tsx
jigjigjig: Real bug, fixed as you suggested in d6fd9ee5: both paths now read `if (!isPending && !isGuarding)`.

Your probe is what makes it obvious in hindsight. The guard takes the footer, so the submit control is not on screen at all, and the mutation ran from a footer offering only Keep editing and Discard, with `isPending` never becoming visible because the guard owns the footer for its whole duration.

Kept the keydown gate as well as the `onSubmit` one, even though every path funnels through `onSubmit` and that gate alone would block the call. `requestSubmit` runs the form's constraint validation first, so without it a shortcut the guard is refusing would still focus an invalid field and answer a gesture that is going nowhere.

Covered by a case that arms the guard, presses Enter and then both Cmd+Enter and Ctrl+Enter against a focused field, asserts `onSubmit` was never called, then presses Keep editing and asserts Enter submits again. That last half matters: it is the guarded state that is blocked, not the gesture. Removing `&& !isGuarding` from both lines fails exactly that case and nothing else.

web/src/design-system/feedback/FormDialog.tsx
jigjigjig: You are right, and the docstring was arguing a point that does not reach this. What the unlayered `.modal__dialog` rule beats is a `width`; nothing declared `--form-dialog-width`, so a class setting it had nothing to lose to and the inline style bought exactly nothing.

Fixed in d6fd9ee5 with the three modifiers, beside the geometry and above the phone block as you suggested: `.otari-form-dialog--sm { width: 27.5rem }` and so on. The `style` prop, the `as React.CSSProperties` cast and the sentence that justified them are gone, and the call site is `otari-form-dialog otari-form-dialog--${size}`.

I took the modifier form rather than `className="[--form-dialog-width:32.5rem]"` because it keeps all three widths in one place next to the `min-width` and `max-height` they have to sit with, and the size a call site asks for stays a name rather than a number.

Compiled and read out of the emitted stylesheet rather than reasoned about, since the whole question is cascade order. All four rules are unlayered, in this order: `.modal__dialog{width:fit-content;min-width:min(28rem,…)}` at offset 472671, `.otari-form-dialog{min-width:0;…}` at 472802, the three modifiers at 472924, 472961 and 472998, and the phone block's `.otari-form-dialog{width:100%…}` at 473111. So `min-width: 0` still clears the 448px floor that put `sm` out of reach, the modifiers still beat `width: fit-content`, and a sheet below 640px still ignores all three.

web/src/styles/globals.css
jigjigjig: Taking the second option: `sm:pb-[7.5rem]` on `Modal.Container` in d6fd9ee5, and the cap stays at `- 15rem`.

The symmetric gap is the intent. The design is the viewport minus 240px with 120px above and below, so the container's own lower bound should agree with the cap rather than the cap being loosened to meet a `pb-10` that never bound. The comment says so now instead of leaving "the 120px offset ... twice" to be inferred.

Your arithmetic was the useful part: `sm:pb-10` never binding is why nothing looked wrong, and 660px against 740px on a 900px viewport is the number that shows it.

web/src/styles/globals.css
jigjigjig: Correct, and the rule reduced to nothing rather than to less than it claimed. Once the `env()` resolves to 0 what was left was `padding-bottom: 0.75rem`, which is the `py-3` the footer already carries, so in d6fd9ee5 the whole rule is gone rather than rewritten. The block comment says why there is no safe-area pad here instead of claiming the footer clears the home indicator.

Not touching the viewport meta, which is deliberate and documented where it is.

`AppShell.tsx:885` is the same mistake and I have left it alone: it is outside this PR and reverting it here would put an unrelated change in a design-system diff. Worth its own issue, along with the question of whether the shell should adopt `viewport-fit=cover` at all, which is the only thing that would make either of them real.

web/src/design-system/feedback/FormDialog.tsx
jigjigjig: Fixed in d6fd9ee5: the body's `scrollTop` returns to 0 when `error` turns truthy, so the banner is where the operator is looking rather than above the fold.

You had already credited the AT path, and that is the asymmetry worth naming: `role="alert"` announced a failure that a sighted operator could not see, which is the worse of the two ways round.

jsdom lays nothing out, so the test's `scrollTop` is only what the test itself sets. It still answers the question the fix is about, which is whether the effect writes it: the case scrolls an `lg` body to 240, rerenders with an error, and asserts 0. Removing the effect fails exactly that case. I said so in a comment on the test rather than letting a later reader take it for a layout assertion.

PR #1037, opened 2026-09-10
feat(dashboard): move the keys page's one-time secret into FormDialog

jigjigjig:
## Description

Creating an API key used to append a form to the page, and then a second full-width strip holding the secret. Regenerating one appended that same strip from a table row. Both now open the dialog every create flow in the dashboard is moving to.

**They moved together on purpose.** They hand over the same thing, they are eight lines apart in the same file, and migrating one alone would have left a single page showing a one-time secret in a dialog after Create and in a strip after Regenerate, which is the inconsistency this series exists to remove. `RevealSecretStrip` is gone; those were its only two consumers.

**The secret step keeps everything the strip carried**, in the strip's own order and its own words: the "shown only once" sentence, the concealed key, the curl and Python snippets concealed the same way, and the notice for a deployment that never published a gateway address. An operator creating their first key arrives from the empty state and "make your first call" is the next thing they need, so leaving it behind would have made this dialog the one place the flow got worse. Its footer action stays **"I've saved this key"** rather than the "Done" the pattern would give it: this control is an acknowledgement rather than a dismissal.

**Two things the strip's own docstring argued for are kept rather than lost to the surface change.** It refused a modal over a focus trap, a swallowed Escape and a backdrop that ignores clicks. `FormDialog` has none of those, except here, where `isDismissable={false}` *is* the strip's "there is one control and it is the acknowledgement", because a stray backdrop click loses the key forever. And focus still returns to the page's create action, which is why that button no longer hides while the dialog is open: a dialog sits over the page, so there was never a reason to take the heading's action away, and it is where focus lands when the empty state's own button has been unmounted by the row it created.

**The caption under the key** names the owner, the access and the expiry. It deliberately carries no spend figure, though the frame is where you would look for one: a key has no budget of its own, only its owner's, which the paragraph above the table already says and links to. The owner is a name or nothing, never an id.

**The create form's "Create another" Checkbox is gone, and that is a bug fix rather than a copy change.** It was the only control that skipped the secret step, and skipping it discarded a live credential: the server cannot return `result.key` again, so a key created that way was unusable until it was regenerated. A second reviewer found the same defect from the code independently. The only "Create another" left is the ghost button on the secret step, which resets the form after the key has been handed over.

`FormDialog` grows three changes for this, all in `feedback/FormDialog.tsx`:

- **`isSubmitDisabled`.** The create form has three conditions the browser cannot check (a model scope that is valid, an owner, a resolved workspace). The old form disabled its button on them; a submit that fails silently teaches nothing.
- **`isDismissable`.** It also drops the close control and Cancel, since both are dismissals and a dialog with neither has exactly one way out.
- **A footer that stays at the foot of a phone sheet.** Below 640px the dialog is the height of the viewport rather than of its content, and nothing claimed the free space, so the footer sat halfway up an empty sheet. Found by capturing the phone viewport, which is the only place that shape exists.

## How to test it locally

    pnpm --dir web run storybook

`Design system → Feedback → FormDialog` has `SubmitBlocked`, `NotDismissable`, `PhoneSheet` and `PhoneSheetLandscape`; `Actions → Copy` has `RevealedOnArrival` and `CoupledReveal`.

For the page itself:

    pnpm --dir web run build
    pnpm --dir web exec playwright test --project=seed --project=screenshots-desktop-small-light --project=screenshots-mobile-light --grep "create dialog"

Captures land in `web/e2e/screenshots/authenticated.spec.ts-snapshots/` (gitignored, per the repo's convention while the dashboard is mid-migration). Two new ones, each across all three viewports and both themes: the dialog on its form step, and the same dialog with **Advanced** open, which is what makes the body scroll with the header and footer pinned.

The success step is deliberately **not** in that suite. Those projects read pages and write nothing, which is what lets them retry safely; reaching the secret means creating a key. It is covered by the vitest suite and by the `onboarding` project's existing end-to-end create flow instead.

- `pnpm --dir web test` — 5298 pass across 153 files. 41 in `KeysPage.test.tsx`, 5 in the new `secretCaption.test.ts`.
- `pnpm --dir web run storybook:build` then `node .storybook/smoke.mjs` — 378 stories x 2 themes, all rendered clean.
- Playwright `onboarding` + `seed` + two screenshot projects: 98 passed, 4 failed. **Those 4 fail identically on the base commit** (`45a5fe1a`), verified by running them there in a separate worktree; they are `organization member › workspace members` and `organization admin › organization spend and budgets`, neither of which touches keys. Delta from this branch: zero.

Each was checked by removing the behavior and watching the right case go red:

| Remove | Fails |
| --- | --- |
| the trigger's visibility while the dialog is open (restore `addOpen ? undefined :`) | `keeps the page's create action visible while the dialog is open` |
| the `isOpen` gate on the owner roster | `does not walk /v1/users until the create dialog is opened` |
| either setter `resetForm` was missing | `clears the owner and the budget exemption when it reopens` |
| the unified dirty predicate, back to the three-field check | `guards a field the old dirty check did not know about` |
| the `isSubmitDisabled` gate on either submit path | `blocks the button, plain Enter, and Cmd/Ctrl+Enter alike` |
| the absence of a "Create another" slot on the regenerate dialog | `shows a regenerated secret in the same dialog a created one arrives in` |
| the owner-name guard, printing `user_id` when no name resolves | `drops the owner rather than printing an id` **and** `drops the owner for a virtual user` |

The caption's date is asserted against `formatDate` rather than a literal: it is `toLocaleDateString()`, so a literal would pin the suite to whichever locale CI runs under.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Part of #1031

Stacked on #1032, which adds `FormDialog`. Review that one first; this branch contains its commits.

## Checklist

      41 cases in `KeysPage.test.tsx` (3 new, 12 rewritten for the new surface),
      5 in the new `secretCaption.test.ts`, 2 new Storybook stories, and 2 new
      screenshot captures. The dashboard's tests live under `web/src` and
      `web/e2e` rather than the repo-root `tests/` the template names.
      The dashboard's counterparts, which is what this diff touches; `make lint`
      does not reach `web/`. Commands and results above.
      No design-doc change: the placement, label and surface rules this page now
      follows were written down in #1032, and this is the first page to follow
      them rather than a new rule.
      No API change. No route, schema or docstring is touched, and the request
      bodies this form sends are byte-for-byte what it sent before.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

Three premises this change started from turned out to be wrong about the page and were corrected before any code was written: a spend figure for the caption (a key has no budget), the belief that the strip was only a key and a copy field (it also carries the snippets), and the belief that its other consumer was elsewhere in the app (it was this page's own regenerate). Each is called out in the review comments on the diff.

## Summary

- Moved one-time secrets for key creation and regeneration into `FormDialog`.
- Preserved secret details, code snippets, captions, and acknowledgement actions in the dialog.
- Added controlled secret reveal support to `CopyField`.
- Added dialog controls for disabled submission, dismissal behavior, mobile footer layout, and accessibility.
- Removed the create-form “Create another” checkbox and kept the action in the secret step.
- Updated key-flow tests, end-to-end helpers, Storybook stories, and screenshot coverage.

## Benefits

- Keeps sensitive key information in a focused, non-dismissible step.
- Provides consistent behavior for key creation and regeneration.
- Improves keyboard, mobile, and accessibility behavior.
- Delays user loading until the create dialog opens.
- Reduces stale secret reveal behavior during asynchronous copy operations.

### DIFF
--- a/web/e2e/helpers.ts
+++ b/web/e2e/helpers.ts
@@ -180,18 +180,23 @@ export async function dismissComboBox(box: Locator): Promise<void> {
   await expect(box).not.toHaveAttribute("aria-expanded", "true")
 }
 
-// The same, for a combobox inside a dialog, which is the only difference that
-// matters: a modal contains focus, so the `blur()` above lands straight back on
-// the box and the box re-opens on focus. Escape alone closes the popover.
+// The same, for a combobox inside a dialog. Two differences, and the second one
+// is why this is a separate helper rather than a flag.
 //
-// The wait stays, because it is what makes either of these deterministic, and
-// dropping it here is worse than dropping it on a page. If the popover is
-// already closed when the key lands (react-aria closes it on an empty filtered
-// list) the Escape reaches the dialog instead, and a dirty form answers that by
-// arming its unsaved-changes guard, which takes the submit out of the footer.
-// The next lookup then fails on a missing button rather than on the real cause.
+// No `blur()`: a modal contains focus, so it lands straight back on the box and
+// the box re-opens on focus. Escape alone closes the popover.
+//
+// And the Escape is sent only when there is a popover to close. Otherwise it
+// reaches the dialog, and a dirty form answers that by arming its
+// unsaved-changes guard, which takes the submit out of the footer, so the next
+// lookup fails on a missing button rather than on the real cause. The wait
+// cannot prevent that: a closed popover satisfies it instantly, by which point
+// the keystroke has already landed. It is kept for what it does cover, a
+// popover that closes asynchronously.
 export async function dismissComboBoxInDialog(box: Locator): Promise<void> {
-  await box.press("Escape")
+  if ((await box.getAttribute("aria-expanded")) === "true") {
+    await box.press("Escape")
+  }
   await expect(box).not.toHaveAttribute("aria-expanded", "true")
 }
 
--- a/web/e2e/helpers.ts
+++ b/web/e2e/helpers.ts
@@ -180,6 +180,21 @@ export async function dismissComboBox(box: Locator): Promise<void> {
   await expect(box).not.toHaveAttribute("aria-expanded", "true")
 }
 
+// The same, for a combobox inside a dialog, which is the only difference that
+// matters: a modal contains focus, so the `blur()` above lands straight back on
+// the box and the box re-opens on focus. Escape alone closes the popover.
+//
+// The wait stays, because it is what makes either of these deterministic, and
+// dropping it here is worse than dropping it on a page. If the popover is
+// already closed when the key lands (react-aria closes it on an empty filtered
+// list) the Escape reaches the dialog instead, and a dirty form answers that by
+// arming its unsaved-changes guard, which takes the submit out of the footer.
+// The next lookup then fails on a missing button rather than on the real cause.
+export async function dismissComboBoxInDialog(box: Locator): Promise<void> {
+  await box.press("Escape")
+  await expect(box).not.toHaveAttribute("aria-expanded", "true")
+}
+
 // Pick an option from a FilterSelect. It is a HeroUI Select, so the control is a
 // button that opens a listbox popover: there is no native <select> to
 // `selectOption` on, and the option is named by its visible label. React-aria
--- a/web/e2e/parity.management.spec.ts
+++ b/web/e2e/parity.management.spec.ts
@@ -147,9 +147,9 @@ test.describe("api keys", () => {
     // anonymous virtual users an omitted id would.
     const ownerBox = dialog.getByPlaceholder("Pick a user, or type a new id…")
     await ownerBox.fill(PARITY.users.heavy)
-    // A bare Escape rather than `dismissComboBox`: that helper waits for
-    // `aria-expanded` to clear on the box, and inside a dialog the key closes
-    // the popover without the box reporting it.
+    // A bare Escape rather than `dismissComboBox`: that helper follows the key
+    // with a `blur()`, and a modal contains focus, so it lands back on the box
+    // and the box re-opens on focus. Escape alone puts the popover away.
     await page.keyboard.press("Escape")
     await dialog.getByRole("button", { name: "Create key" }).click()
 
--- a/web/e2e/parity.management.spec.ts
+++ b/web/e2e/parity.management.spec.ts
@@ -138,22 +138,32 @@ test.describe("api keys", () => {
     await openPage(page, "API keys", "API keys")
 
     await page.getByRole("button", { name: "Create key" }).click()
-    await page.getByLabel("Name").fill(KEY_NAME)
+    // Scoped from here on, because "Create key" is now on screen twice: the
+    // heading's trigger stays visible while the dialog is open, and the labels
+    // rule makes the dialog's submit say the same words.
+    const dialog = page.getByRole("dialog")
+    await dialog.getByLabel("Name").fill(KEY_NAME)
     // Owner is required: this is what keeps the dashboard from minting the
     // anonymous virtual users an omitted id would.
-    const ownerBox = page.getByPlaceholder("Pick a user, or type a new id…")
+    const ownerBox = dialog.getByPlaceholder("Pick a user, or type a new id…")
     await ownerBox.fill(PARITY.users.heavy)
-    await dismissComboBox(ownerBox)
-    await page.getByRole("button", { name: "Create key" }).click()
+    // A bare Escape rather than `dismissComboBox`: that helper waits for
+    // `aria-expanded` to clear on the box, and inside a dialog the key closes
+    // the popover without the box reporting it.
+    await page.keyboard.press("Escape")
+    await dialog.getByRole("button", { name: "Create key" }).click()
 
-    // The secret is shown exactly once, behind an explicit acknowledgement that
-    // Esc deliberately does not dismiss. A strip with `role="alert"` rather than
-    // a dialog: the modal's focus trap and swallowed Esc went with the redesign,
-    // and the acknowledgement is still the only thing that dismisses it.
+    // The secret is shown exactly once, behind an acknowledgement that is the
+    // only way out: the dialog is `isDismissable={false}`, so Escape and the
+    // close control are gone. The reveal keeps its own `role="alert"` inside
+    // the dialog body, which is what the first two assertions read.
     const reveal = page.getByRole("alert", { name: /API key created/ })
     await expect(reveal).toBeVisible()
     await expect(reveal).toContainText("shown only once")
-    await reveal.getByRole("button", { name: /saved this key/i }).click()
+    // Scoped to the dialog rather than to the reveal: the acknowledgement is
+    // the dialog's own submit, in the footer, so it sits outside the alert
+    // region the announcement covers.
+    await dialog.getByRole("button", { name: /saved this key/i }).click()
 
     const key = row(page, "API keys", KEY_NAME)
     await expect(key).toContainText("Active")
--- a/web/e2e/parity.management.spec.ts
+++ b/web/e2e/parity.management.spec.ts
@@ -2,6 +2,7 @@ import { expect, type Locator, type Page, test } from "@playwright/test"
 
 import {
   dismissComboBox,
+  dismissComboBoxInDialog,
   gotoRoute,
   login,
   nav,
@@ -147,10 +148,7 @@ test.describe("api keys", () => {
     // anonymous virtual users an omitted id would.
     const ownerBox = dialog.getByPlaceholder("Pick a user, or type a new id…")
     await ownerBox.fill(PARITY.users.heavy)
-    // A bare Escape rather than `dismissComboBox`: that helper follows the key
-    // with a `blur()`, and a modal contains focus, so it lands back on the box
-    // and the box re-opens on focus. Escape alone puts the popover away.
-    await page.keyboard.press("Escape")
+    await dismissComboBoxInDialog(ownerBox)
     await dialog.getByRole("button", { name: "Create key" }).click()
 
     // The secret is shown exactly once, behind an acknowledgement that is the
--- a/web/src/design-system/actions/CopyField.tsx
+++ b/web/src/design-system/actions/CopyField.tsx
@@ -197,6 +197,16 @@ export function CopyField({
   const resetTimer = useRef<ReturnType<typeof setTimeout> | undefined>(
     undefined,
   )
+  // The value on screen right now, which is not what `copy` below closes over:
+  // that is the value of the render the press happened in, and a credential can
+  // arrive while the attempt is in flight. Controlled mode needs it because
+  // `onRevealChange` carries a bare boolean and cannot say which value a reveal
+  // was asked for; uncontrolled mode keys on the value itself through
+  // `revealedValue`.
+  const latestValue = useRef(value)
+  useEffect(() => {
+    latestValue.current = value
+  }, [value])
   // The value a failed copy asked to have selected, once revealing it has put
   // it on the field: `select()` in the failure's own tick would span the
   // stand-in, and the value React writes on the revealing render discards a
@@ -246,7 +256,7 @@ export function CopyField({
       if (ref.current?.value === copying) {
         ref.current.focus()
         ref.current.select()
-      } else {
+      } else if (latestValue.current === copying) {
         selectOnReveal.current = copying
         setRevealed(true)
       }
--- a/web/src/design-system/actions/CopyField.tsx
+++ b/web/src/design-system/actions/CopyField.tsx
@@ -112,6 +112,13 @@ type CopyFieldProps = {
        * `onRevealChange` and shows what the caller says. Several fields sharing
        * one credential share one of these, so they reveal and conceal together
        * rather than one at a time.
+       *
+       * **A controlled field does not re-conceal on a new value.** Uncontrolled,
+       * the reveal is keyed to the value it was asked for (see the comment on
+       * `revealedValue` below), so a rotation arriving into a revealed field
+       * conceals itself. A controlled caller owns that instead: conceal on a
+       * value change, or the replacement is on screen without anyone having
+       * asked to see it.
        */
       isRevealed?: boolean
       /** The toggle's press, for a controlled field. */
--- a/web/src/design-system/actions/CopyField.tsx
+++ b/web/src/design-system/actions/CopyField.tsx
@@ -1,6 +1,6 @@
 import { Button } from "@heroui/react"
 import type { ReactElement, ReactNode } from "react"
-import { useEffect, useId, useRef, useState } from "react"
+import { useEffect, useId, useLayoutEffect, useRef, useState } from "react"
 import { FiEye, FiEyeOff } from "react-icons/fi"
 import { CopyButton } from "@/design-system/actions/CopyButton"
 import { copyToClipboard } from "@/design-system/helpers/clipboard"
@@ -204,7 +204,12 @@ export function CopyField({
   // was asked for; uncontrolled mode keys on the value itself through
   // `revealedValue`.
   const latestValue = useRef(value)
-  useEffect(() => {
+  // `useLayoutEffect`, not `useEffect`: the reader is a rejected-promise
+  // handler, so it runs as a microtask, and a passive effect is scheduled
+  // rather than run at commit. A rejection landing after the render that
+  // carries the new credential but before that effect would read the old one
+  // here and pass the gate below, which is the case this exists to refuse.
+  useLayoutEffect(() => {
     latestValue.current = value
   }, [value])
   // The value a failed copy asked to have selected, once revealing it has put
--- a/web/src/design-system/actions/CopyField.tsx
+++ b/web/src/design-system/actions/CopyField.tsx
@@ -95,11 +95,32 @@ type CopyFieldProps = {
        * has been asked for, while Copy copies the real value either way, so a
        * key can be handed over without being read off the screen (otari-ai#2111).
        *
+       * The one-time secret step is the exception and opens revealed, by product
+       * ruling: that screen exists to hand the key over, and there is no second
+       * chance to read it. The toggle and copying without reading survive there,
+       * which is what otari-ai#2111 asked for.
+       *
        * A whole-value field passes `CONCEALED_SECRET`. A snippet passes the same
        * snippet built around that stand-in, so what is hidden is the key rather
        * than the request that explains it.
        */
       concealed?: string
+      /**
+       * Whether the value is revealed, for a caller that owns the state.
+       *
+       * Passing it makes the field controlled: the toggle reports through
+       * `onRevealChange` and shows what the caller says. Several fields sharing
+       * one credential share one of these, so they reveal and conceal together
+       * rather than one at a time.
+       */
+      isRevealed?: boolean
+      /** The toggle's press, for a controlled field. */
+      onRevealChange?: (next: boolean) => void
+      /**
+       * Whether an uncontrolled field starts revealed. Concealed by default,
+       * and a later value arrives concealed whatever this said.
+       */
+      defaultRevealed?: boolean
       /**
        * Excluded here rather than guarded at runtime, so a call site that passes
        * both fails to compile. The multiline field is a `<textarea>`, where the
@@ -112,6 +133,9 @@ type CopyFieldProps = {
       multiline?: false
       /** Nothing hands out a credential beside a domain-verification action. */
       concealed?: never
+      isRevealed?: never
+      onRevealChange?: never
+      defaultRevealed?: never
       /**
        * A control to sit beside the field, which moves the copy affordance
        * inside the field and drops the button from the label row. Absent, the
@@ -132,6 +156,9 @@ export function CopyField({
   fieldRef,
   action,
   concealed,
+  isRevealed,
+  onRevealChange,
+  defaultRevealed = false,
 }: CopyFieldProps) {
   const internalRef = useRef<HTMLInputElement | HTMLTextAreaElement | null>(
     null,
@@ -146,12 +173,16 @@ export function CopyField({
   // copying: a rotation with the previous key still on screen would otherwise
   // publish the replacement without anyone asking to see it.
   const [revealedValue, setRevealedValue] = useState<string | undefined>(
-    undefined,
+    defaultRevealed ? value : undefined,
   )
   const [selectHintFor, setSelectHintFor] = useState<string | undefined>(
     undefined,
   )
-  const revealed = revealedValue === value
+  const revealed = isRevealed ?? revealedValue === value
+  const setRevealed = (next: boolean) => {
+    onRevealChange?.(next)
+    if (isRevealed === undefined) setRevealedValue(next ? value : undefined)
+  }
   const selectHint = selectHintFor === value
   // Same shape as CopyButton's below: the acknowledgement clears itself on a
   // timer, so the timer has to die with the component (and be replaced rather
@@ -174,10 +205,10 @@ export function CopyField({
     // Only once the reveal has landed on the value the copy was for. Another
     // credential arriving in the meantime is concealed, and selecting its
     // stand-in is exactly what this is here to avoid.
-    if (revealedValue !== wanted || value !== wanted) return
+    if (!revealed || value !== wanted) return
     ref.current?.focus()
     ref.current?.select()
-  }, [revealedValue, value, ref])
+  }, [revealed, value, ref])
 
   const isConcealed = concealed !== undefined && !revealed
   const shown = isConcealed ? concealed : value
@@ -210,7 +241,7 @@ export function CopyField({
         ref.current.select()
       } else {
         selectOnReveal.current = copying
-        setRevealedValue(copying)
+        setRevealed(true)
       }
       setSelectHintFor(copying)
       return
@@ -303,7 +334,7 @@ export function CopyField({
       variant="ghost"
       isIconOnly
       aria-label={`${revealed ? "Hide" : "Show"} ${label}`}
-      onPress={() => setRevealedValue(revealed ? undefined : value)}
+      onPress={() => setRevealed(!revealed)}
     >
       {revealed ? (
         <FiEyeOff aria-hidden="true" className="h-3.5 w-3.5" />
--- a/web/src/design-system/feedback/FormDialog.tsx
+++ b/web/src/design-system/feedback/FormDialog.tsx
@@ -30,6 +30,25 @@ export interface FormDialogProps {
    * so Escape and a click outside ask before discarding.
    */
   isDirty?: boolean
+  /**
+   * Whether Escape, a backdrop click and the close control can dismiss it.
+   *
+   * On by default, and `isDirty` is the answer for a form holding unsaved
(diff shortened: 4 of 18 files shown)


web/src/design-system/feedback/FormDialog.tsx
```diff
@@ -0,0 +1,336 @@
...
+              }}
+              onKeyDown={(event) => {
+                if (event.key !== "Enter") return
+                if (!event.metaKey && !event.ctrlKey) return
+                event.preventDefault()
+                // `requestSubmit`, not `onSubmit` directly: it runs the form's
+                // constraint validation first, so the shortcut and the button
+                // are one path rather than two, and a required field left empty
+                // cannot reach the mutation through the keyboard alone.
+                if (!isPending) event.currentTarget.requestSubmit()
```
khaledosman: **Cmd/Ctrl+Enter bypasses `isSubmitDisabled`.** `requestSubmit()` ignores the submit button's `disabled` state (it only runs constraint validation), and the `onSubmit` handler on line 216 guards `isPending` alone.

Verified against this head: with `isSubmitDisabled`, a click and a plain Enter are both blocked (react-aria does render a native `disabled`), and Cmd+Enter fires `onSubmit` once.

`CreateKeyDialog` happens to be safe because `submit()` re-checks `blocked` at `KeysPage.tsx:508`, so this is latent today, but the prop's own docstring makes it the guard and the next caller will trust it. Both handlers:

```ts
if (!isPending && !isSubmitDisabled) onSubmit()
// and
if (!isPending && !isSubmitDisabled) event.currentTarget.requestSubmit()
```

web/src/features/keys/KeysPage.tsx
```diff
@@ -1242,15 +1344,13 @@ export function KeysPage() {
...
-      {addOpen ? (
-        <CreateKeyForm
-          isDeploymentWide={isDeploymentWide}
-          onClose={() => setAddOpen(false)}
-          onCreated={(result) =>
-            setRevealed({ title: "API key created", result })
-          }
-        />
-      ) : null}
+      <CreateKeyDialog
```
khaledosman, line 1331: **The `/v1/users` walk now runs on every visit to this page.** Mounting the dialog unconditionally means `useUsers(isDeploymentWide)` (line 436) is gated only on the deployment, not on `isOpen`, and `fetchAllUsers` walks up to 100 pages of 1000 (`shared/api/users.ts:9`).

Confirmed by probe: rendering the page with the dialog closed issues `/v1/users?skip=0&limit=1000` on this head and nothing on the base commit, where `CreateKeyForm` only mounted on `addOpen`.

`useUsers(isDeploymentWide && isOpen)` keeps the mount, and so the exit animation the `close()` comment relies on, while deferring the fetch; `staleTime: 60_000` makes a reopen free.

web/src/design-system/feedback/FormDialog.tsx
```diff
@@ -0,0 +1,336 @@
...
+  isDismissable?: boolean
+  /**
+   * Whether the form is not yet in a state that can be submitted.
+   *
+   * The submit is shown disabled rather than allowed to fail: a press that does
+   * nothing teaches nothing, and the reason belongs beside the control that is
+   * missing. Reserve it for what the browser cannot check itself; `required` on
+   * a field is better than a boolean here, because it also says which field.
+   */
+  isSubmitDisabled?: boolean
```
khaledosman, line 51: **Two new props on this primitive and three on `CopyField`, with no colocated cases.** `FormDialog.test.tsx` and `CopyField.test.tsx` both exist and neither is in this diff, while the change adds `isDismissable`, `isSubmitDisabled`, `isRevealed`, `onRevealChange` and `defaultRevealed`. Coverage is only indirect, through `KeysPage.test.tsx`.

The frontend standards make the colocated test the rule ("Add a Vitest test for any component, hook, or helper whose behavior you change"), and it is not academic here: `FormDialog.test.tsx:110` already owns the case next door ("runs the form's validation on Cmd/Ctrl+Enter rather than going around it"), which is exactly where the `isSubmitDisabled` hole above would have surfaced.

web/src/design-system/actions/CopyField.tsx
```diff
@@ -95,11 +95,32 @@ type CopyFieldProps = {
...
       concealed?: string
+      /**
+       * Whether the value is revealed, for a caller that owns the state.
+       *
+       * Passing it makes the field controlled: the toggle reports through
+       * `onRevealChange` and shows what the caller says. Several fields sharing
+       * one credential share one of these, so they reveal and conceal together
+       * rather than one at a time.
+       */
+      isRevealed?: boolean
```
khaledosman, line 123: **The controlled reveal hands away a security invariant without saying so.** `revealed = isRevealed ?? revealedValue === value` (line 181) means a controlled caller loses the keyed-to-value behavior the block comment at line 174 argues for: "a rotation with the previous key still on screen would otherwise publish the replacement without anyone asking to see it." `CopyField.test.tsx:329` asserts that ("conceals a second value, rather than inheriting the first one's reveal") for the uncontrolled path only.

`defaultRevealed`'s docstring does state it ("a later value arrives concealed whatever this said"); this one does not. No live exposure, since both consumers remount `KeySecretStep` between secrets. Worth saying on the prop, so the next controlled caller re-conceals on a value change rather than inheriting one.

web/e2e/parity.management.spec.ts
```diff
@@ -137,22 +137,32 @@ test.describe("api keys", () => {
...
-    await page.getByRole("button", { name: "Create key" }).click()
-
-    // The secret is shown exactly once, behind an explicit acknowledgement that
-    // Esc deliberately does not dismiss. A strip with `role="alert"` rather than
-    // a dialog: the modal's focus trap and swallowed Esc went with the redesign,
-    // and the acknowledgement is still the only thing that dismisses it.
+    // A bare Escape rather than `dismissComboBox`: that helper follows the key
+    // with a `blur()`, and a modal contains focus, so it lands back on the box
+    // and the box re-opens on focus. Escape alone puts the popover away.
+    await page.keyboard.press("Escape")
```
khaledosman: **The bare Escape dropped the wait along with the `blur()`, and its failure mode is confusing.** Only the blur is unsafe in a modal; `await expect(box).not.toHaveAttribute("aria-expanded", "true")` was the part that made `dismissComboBox` deterministic.

If the popover is not open when the key lands (react-aria closes it on an empty filtered list), Escape reaches the dialog, and because the form is dirty by then it arms the unsaved-changes guard and swaps the submit out of the footer, so the next `getByRole("button", { name: "Create key" })` fails on a missing element rather than on the real cause. `KeysPage.test.tsx:891` documents that hazard precisely and drops the Escape; the specs keep it. Same at `dashboard.spec.ts:189`.

Keeping the `aria-expanded` wait also gives the four-line comment now duplicated in both files one home, which matters since the rest of the series migrates providers, budgets and users into the same dialog.

khaledosman (approved):
Gates clean: no route, schema or docstring touched, so no OpenAPI or Postman regeneration owed; no `schema.ts` / `routeTree.gen.ts` drift; no dependency or lockfile change; conventional title; no em dashes or `--` separators in the added prose; two screenshot entries for the new dialog states.

Five findings inline, no blockers. Two are confirmed by running the code at this head: the `isSubmitDisabled` hole on the Cmd/Ctrl+Enter path, and the `/v1/users` walk that now fires on page load (base commit issues none).

---

web/src/design-system/feedback/FormDialog.tsx
jigjigjig: Fixed in efa3698d, both handlers, exactly as written. Same shape as the guard hole on the parent PR, in the prop this one added, and your reason for fixing it despite `CreateKeyDialog` being safe is the right one: the docstring makes the prop the guard, so the next caller will trust it rather than re-checking.

The comment now says the general rule instead of naming one condition, since that is twice this has been the bug: every reason the button is not pressable has to hold for the keyboard too. `requestSubmit` consults constraint validation and nothing else.

Covered colocated, in the file you pointed at: `isSubmitDisabled` with a textarea, a click and then both Cmd+Enter and Ctrl+Enter, asserting `onSubmit` was never called, plus a case without the prop asserting Cmd+Enter does submit, so the gate is the prop and not the gesture. Removing `&& !isSubmitDisabled` fails the first and nothing else.

web/src/features/keys/KeysPage.tsx
jigjigjig: Confirmed and fixed in efa3698d: `useUsers(isDeploymentWide && isOpen)`. The mount stays, so the exit animation `close()` relies on is untouched, and `staleTime` makes a reopen free, as you said.

Your probe against the base commit is what makes this a regression rather than a pre-existing cost, and I had not looked: the dialog moving from conditionally mounted to always mounted took its queries with it, which is a consequence of the surface change that is invisible in the diff of the surface change.

The assertion is on the page rather than on the dialog, deliberately: it asserts no `/v1/users` request fires while the dialog is closed, then that one fires once it opens. Written that way because you are right that the next page repeats the shape, and organization members is the one with an owner picker of its own. Reverting the gate fails exactly that case.

web/src/design-system/feedback/FormDialog.tsx
jigjigjig: Right, and I had taken the wrong thing for coverage. The catalog gate (`foundation.test.ts`, "the catalog shows every prop") made me add stories for the `CopyField` three, and a story renders a prop without asserting anything about it. That is a rendering, not a test, and I treated the gate going green as the box being ticked.

Added in efa3698d, colocated:

- `FormDialog.test.tsx`: `isSubmitDisabled` blocking the button, plain Enter and Cmd/Ctrl+Enter alike; a case without it proving the gate is the prop; the submit rendering disabled rather than failing on press. Plus `isDismissable` dropping Close and Cancel together, and offering both by default.
- `CopyField.test.tsx`: `defaultRevealed` opening revealed and still concealing on the toggle; the controlled pair taking its state from the caller and reporting the press without acting on it; and the value-change behavior in the thread below.

Your point about where it would have surfaced is exact: the neighbouring case at `FormDialog.test.tsx:110` is about `requestSubmit` going around validation, and the `isSubmitDisabled` hole is the same sentence with a different condition. It is now the case immediately after it.

web/src/design-system/actions/CopyField.tsx
jigjigjig: Correct, and worth more than a docstring on its own, so efa3698d does both.

`isRevealed`'s docstring now says it: a controlled field does not re-conceal on a new value, uncontrolled keying is what a controlled caller gives up, and conceal on a value change or the replacement is on screen without anyone having asked to see it. It points at the `revealedValue` comment rather than restating its argument.

And a case pins it beside the uncontrolled one you cited, asserting that a controlled revealed field shows a second value. That reads as asserting the hazard, which is why the comment on it says so in as many words: it is there so the behavior is a decision the next controlled caller can read, rather than something discovered by a rotation.

Agreed there is no live exposure. Both consumers remount `KeySecretStep` between secrets, and the reveal state lives in that component, so it cannot survive into a second key.

web/e2e/parity.management.spec.ts
jigjigjig: You are right, and this corrects my own fix rather than the original code. The conclusion I worked to was "Escape alone is enough", which was right about the `blur()` and wrong to take the wait with it. Only the blur is unsafe in a modal; the wait was the part that made the dismissal deterministic.

Fixed in efa3698d with `dismissComboBoxInDialog` in `helpers.ts`, so the comment has one home as you asked, and both call sites use it: `parity.management.spec.ts` and `dashboard.spec.ts:189`.

```ts
export async function dismissComboBoxInDialog(box: Locator): Promise<void> {
  await box.press("Escape")
  await expect(box).not.toHaveAttribute("aria-expanded", "true")
}
```

The failure mode is in its comment, because it is the part nobody would reconstruct from the symptom: with the popover already closed, the Escape reaches the dialog, a dirty form answers it by arming the unsaved-changes guard, the guard takes the submit out of the footer, and the next lookup fails on a missing button rather than on the real cause.

Parity and onboarding both pass on this head, including the two tests that use it.

web/src/features/keys/KeysPage.tsx
```diff
@@ -469,10 +478,38 @@ function CreateKeyForm({
...
+    // document.body means nothing took it.
+    requestAnimationFrame(() => {
+      if (document.activeElement === document.body) {
+        returnFocusRef.current?.focus()
+      }
+    })
+    // The frame keeps its content while it animates out, so the reset waits for
+    // the next open rather than blanking the dialog as it goes.
+    setCreated(null)
+    resetForm()
```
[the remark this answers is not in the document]
jigjigjig: Right, and fixed, though not by deferring the reset.

`close()` no longer resets anything: it only sets `isOpen` false. The page keys the dialog on a counter it bumps where the trigger opens it, so the draft is remounted fresh on the way **in** and is untouched through the exit. Your note that one `requestAnimationFrame` would not cover it is what ruled out deferring: the exit is an animation, not a frame, and any timer tuned to it would be a second copy of the animation's duration.

The comment three lines above the old reset had claimed exactly the behavior the code did not have ("the reset waits for the next open rather than blanking the dialog as it goes"), so this thread found a comment and its code disagreeing, not only a flash.

The rule is now in `web/design/feedback.md` for the pages still migrating, and three more of them were fixed the same way on the workspaces PR: "a draft is fresh on every open and untouched through the exit". The edit dialog beside this one already worked this way, keyed on the row id.

web/src/styles/globals.css
```diff
@@ -2189,22 +2189,70 @@ main {
...
   }
+  /* The footer stacks. Two 36px buttons and a caption sharing one row fit a
+   * desktop and not a 390px sheet: the routing dialog's caption wrapped to three
+   * lines beside them. Primary on top, Cancel under it, each the sheet's full
+   * width at the 44px touch floor. Drawn on artboard 07.
+   *
+   * Cancel is first in the DOM, where it belongs beside the primary on a
+   * desktop, so `order` moves it rather than the markup: reading order on a
+   * desktop and stacking order on a phone are different questions. */
+  .otari-form-dialog__footer {
```
khaledosman, line 2236: **The footer stack fires on a short desktop window, not just a landscape phone.** `@media (width <= 639px), (height <= 639px)` at line 2207 is an `or`, so a 1280x600 viewport matches: `.otari-form-dialog` goes `width: 100%; height: 100%; border: 0`, and these rules stack two buttons full-width across 1280px. That is not a phone sheet, it is a maximized dialog with a broken footer.

Not a rare viewport. A 1366x768 laptop with a bookmarks bar lands near 630, and any docked DevTools or half-height window gets there on a larger display. The width-only query this replaced was correct on the desktop; the landscape-phone case (844x390) is the only thing that needed the second axis.

Split the two concerns: keep the sheet sizing on either axis, and gate the footer stack on width, since its whole argument is the 390px column.

```css
@media (width <= 639px), (height <= 639px) {
  .otari-form-dialog__container { padding: 0 }
  .otari-form-dialog { width: 100%; max-width: none; height: 100%; max-height: none; border: 0 }
}
@media (width <= 639px) {
  .otari-form-dialog__footer { … }
  .otari-form-dialog__actions { … }
}
```

web/src/features/keys/KeysPage.tsx
```diff
@@ -494,46 +545,88 @@ function CreateKeyForm({
...
+        setCreated(result)
       },
     })
   }
 
+  if (created) {
+    return (
+      <FormDialog
+        isOpen={isOpen}
+        onOpenChange={close}
```
khaledosman: **The secret step takes `onOpenChange={close}` bare, where the form step guards the `open === true` call.** Line 594 spells it `(open) => (open ? onOpenChange(true) : close())`; here any `onOpenChange(true)` runs `close()`, which sets `isOpen` false. `FormDialog` forwards a truthy open straight through (`FormDialog.tsx:129-131`), and `isDismissable={false}` does not cover that path, only `requestClose`.

Not reachable in a browser today: the only source is the hidden `Modal.Trigger`, and `.hidden` is `display: none`. But this is the one dialog where being wrong is unrecoverable, and the guard is already written twelve lines up. Use the same shape here.

web/e2e/helpers.ts
```diff
@@ -180,6 +180,21 @@ export async function dismissComboBox(box: Locator): Promise<void> {
...
+// the box and the box re-opens on focus. Escape alone closes the popover.
+//
+// The wait stays, because it is what makes either of these deterministic, and
+// dropping it here is worse than dropping it on a page. If the popover is
+// already closed when the key lands (react-aria closes it on an empty filtered
+// list) the Escape reaches the dialog instead, and a dirty form answers that by
+// arming its unsaved-changes guard, which takes the submit out of the footer.
+// The next lookup then fails on a missing button rather than on the real cause.
+export async function dismissComboBoxInDialog(box: Locator): Promise<void> {
+  await box.press("Escape")
```
khaledosman: **The wait does not prevent the hazard the comment above credits it with.** Lines 187-192 argue that dropping the wait is worse here because an Escape landing on an already-closed popover reaches the dialog and arms the dirty guard. But if the popover is already closed, `aria-expanded` is not `"true"`, so line 195 passes instantly and the guard is armed either way; the next `getByRole("button", { name: … })` still fails on a missing submit rather than on the real cause.

The wait is worth keeping for the case it does cover (the popover closing asynchronously). Press Escape only when there is something to dismiss:

```ts
if ((await box.getAttribute("aria-expanded")) === "true") await box.press("Escape")
await expect(box).not.toHaveAttribute("aria-expanded", "true")
```

and shorten the comment to what survives.

khaledosman (approved):
All five findings from the earlier review are fixed and verified at this head: `isSubmitDisabled` now gates `onSubmit` and the Cmd/Ctrl+Enter path alike (`FormDialog.tsx:218,232`), `useUsers(isDeploymentWide && isOpen)` defers the 100-page walk (`KeysPage.tsx:440`), `FormDialog.test.tsx` and `CopyField.test.tsx` cover all five new props by name, the `isRevealed` docstring states the invariant a controlled caller takes over, and `dismissComboBoxInDialog` has one home in `helpers.ts` with the `aria-expanded` wait restored.

Gates clean: no route, schema or docstring touched, so nothing owed on OpenAPI or Postman; no `schema.ts` / `routeTree.gen.ts` drift; no dependency or lockfile change; conventional title; no em dashes or `--` separators in the added prose. CI green.

One thing to fix before merge, at `globals.css:2226`: the `(height <= 639px)` half of the query takes the whole sheet treatment onto a short desktop window.

---

web/src/styles/globals.css
jigjigjig: Taking this, and removing the reordering rather than compensating for it. Fixed in acef9226.

`order: 1` is gone and nothing reorders the footer, so **DOM, visual and focus order agree at both widths**. That is the part worth stating, because the obvious alternative does not manage it: reversing the row on desktop instead would have kept the phone's visual order and put Tab right-then-left under a left-to-right pair, which is the same divergence turned sideways rather than removed.

It also changes the stacking, and for a reason independent of focus order: on a phone sheet the control nearest the thumb is the **lowest** one, so the primary belongs at the bottom with Cancel above it. Cancel-above-primary is the design now, not a consequence of dropping the rule; the artboard that drew it the other way is being corrected.

Two facts I gathered while deciding, both of which counted against what I had written: the design docs say nothing about focus or visual order, and that `order: 1` was the only use of CSS `order` in the tree. So the mechanism was novel here and had no house rule behind it, which is a poor position for a call about keyboard behavior.

Measured on a build at 390x844: Cancel's top is 742, the primary's is 796, both at the default `order: 0`. A test pins the DOM half, which is the half jsdom can see, and says in as many words that the stacking itself is a media query measured on a build rather than there.

web/src/features/keys/KeysPage.tsx
jigjigjig: Taking it, ec240efd. The secret step reads `(open) => (open ? onOpenChange(true) : close())`, the same shape as the form step.

Your reasoning for fixing an unreachable path is the part I want to agree with out loud: the guard was already written twelve lines up, so the cost of consistency here is nothing and the cost of being wrong is a key nobody can recover. The `.hidden` trigger is `display: none` today, which is a property of a workaround for a HeroUI warning rather than a decision anyone made about safety.

I also took the second half of your note into the comment: `isDismissable={false}` guards `requestClose` and not a forwarded open, which is exactly why it read as covering this and did not.

web/e2e/helpers.ts
jigjigjig: You are right, and this corrects my own comment rather than the code around it. Fixed in ec240efd.

The wait cannot prevent that hazard. If the popover is already closed, `aria-expanded` is not `"true"`, so the assertion passes instantly, and by then the keystroke has already reached the dialog and armed the guard. My comment credited the wait with a protection only the *press* can provide, which is the inverse of what it does.

So the press is now conditional and the wait stays for what it does cover:

```ts
if ((await box.getAttribute("aria-expanded")) === "true") {
  await box.press("Escape")
}
await expect(box).not.toHaveAttribute("aria-expanded", "true")
```

The comment is shortened to two facts that survive: no `blur()`, because a modal contains focus and the box reopens on it; and the Escape only when there is a popover to close, because otherwise it reaches the dialog. The sentence claiming the wait was the protection is gone.

Worth noting for the rest of the stack: that wrong sentence had already been quoted elsewhere as the explanation for this helper, so this thread caught a mistake that was propagating rather than sitting still.

web/src/styles/globals.css
```diff
@@ -2189,22 +2189,69 @@ main {
...
+ * The container's own `sm:` paddings still apply at that width, so they are
+ * overridden here too; without that the sheet would sit inside a 240px vertical
+ * inset and none of this would help.
+ *
+ * The footer's bottom padding is a plain 1rem and not a safe-area inset:
+ * `env(safe-area-inset-bottom)` is 0 without `viewport-fit=cover`, which
+ * `index.html` leaves off deliberately (the shell has no insets to sit out of
+ * the way), so adding it would read as clearing the home indicator while
+ * changing nothing. */
+@media (width <= 639px), (height <= 639px) {
```
khaledosman, line 2207: **Still one block, so the phone sheet still fires on a short desktop window.** `(width <= 639px), (height <= 639px)` is an `or`: a 1280x600 viewport matches, and everything inside applies. The dialog goes `width: 100%; max-width: none; height: var(--visual-viewport-height, 100dvh); border: 0`, and `:2228-2244` stacks two buttons full-width across 1280px. That is not a phone sheet, it is an edge-to-edge desktop dialog with a broken footer.

Reachable on ordinary hardware: a 1366x768 or 1280x720 laptop, any window that is not maximized, a split-screen pane, or DevTools docked to the bottom.

The height axis is right for the case it was added for (a landscape phone at 844x390 could not afford the 120px insets). It is only the footer stack that needs the narrow-width argument, since its whole rationale is the 390px column. Split them:

```css
@media (width <= 639px), (height <= 639px) {
  .otari-form-dialog__container { padding: 0 }
  .otari-form-dialog { width: 100%; max-width: none;
    height: var(--visual-viewport-height, 100dvh); max-height: none; border: 0 }
}
@media (width <= 639px) {
  /* the footer stack, unchanged */
}
```

If you would rather keep one block, `(height <= 639px) and (width <= 1023px)` on the second half also does it; the point is that nothing above 639px wide should be stacking a full-width footer.

web/src/design-system/actions/CopyField.tsx
```diff
@@ -146,19 +180,33 @@ export function CopyField({
...
     undefined,
   )
+  // The value on screen right now, which is not what `copy` below closes over:
+  // that is the value of the render the press happened in, and a credential can
+  // arrive while the attempt is in flight. Controlled mode needs it because
+  // `onRevealChange` carries a bare boolean and cannot say which value a reveal
+  // was asked for; uncontrolled mode keys on the value itself through
+  // `revealedValue`.
+  const latestValue = useRef(value)
+  useEffect(() => {
```
khaledosman: **`useEffect` leaves a window where `latestValue.current` is the value the gate is meant to reject.** Passive effects are scheduled rather than run at commit, and the reader here (`:259`) is a rejected-promise handler, which is a microtask. A rejection that lands after the render carrying the new value but before the passive effect runs sees the ref still holding the old one, `latestValue.current === copying` passes, and `setRevealed(true)` puts the replacement on screen. That is precisely the case this commit exists to prevent.

The new test does not cover it: `rerender` in Testing Library flushes effects synchronously, so the ref is always current by the time `refuse()` runs.

`useLayoutEffect` runs at commit, before any microtask queued after it, and is the standard construct for a latest-value ref for this reason. One word, and the gate holds in the window it was written for.

khaledosman (changes requested):
Two of the three fixed exactly as asked, and both are better than what I suggested: `dismissComboBoxInDialog` now sends the Escape only when there is a popover to close and the comment claims only what the wait actually covers, and the secret step's `onOpenChange` is the same guarded shape as the form step's.

The third is not fixed. The reply on that thread is about `order: 1`, which is CodeRabbit's finding and is indeed gone; mine was about the media query's height axis, and `globals.css:2207` is unchanged. Flagging that rather than approving again, because last round I called it "one thing to fix before merge" and it went out with the round.

One new thing on `99a1da60`, which is otherwise a good catch and a good test.

Gates clean: no route, schema or docstring; no generated-artifact drift; no dependency change; conventional title; no em dashes or `--` separators in the added prose. CI green. `feedback.md`'s correction in `fee56c2f` is worth having on its own, since the page-level mutation it now warns about is exactly what #1041 hit.

---

web/src/styles/globals.css
jigjigjig: Right, and fixed as you wrote it in 08499247: geometry on either axis, footer stack on width alone.

The height axis stays for the reason it was added (a landscape phone at 844x390 cannot afford the 120px insets), and the stack moves to `@media (width <= 639px)` because its whole rationale was the 390px column. So a 1280x600 window now gets the full-bleed sheet with a normal row footer, and a landscape phone keeps the sheet and gets a row footer too, which its 844px affords. That last part is a deliberate change from what the artboard note said, and the note is being corrected rather than worked around.

Checked in the emitted CSS rather than the source, since that is where a media query either landed or did not: the built `index-*.css` carries `@media (width<=639px),(height<=639px)` holding only `.otari-form-dialog` and `.otari-form-dialog__container`, and a second `@media (width<=639px)` holding `.otari-form-dialog__footer` and `.otari-form-dialog__actions` with the column direction. The landscape story's doc comment now says the footer stays a row there.

web/src/design-system/actions/CopyField.tsx
jigjigjig: Correct, and changed to `useLayoutEffect` in 08499247, with the reason in the comment.

One thing to be plain about: I could not build a jsdom case that distinguishes the two, so this rests on your mechanism argument and not on a test. Three shapes tried, all of which reveal nothing with the old `useEffect` in place: `rerender` (your point, effects flush synchronously), the rotation and the rejection inside one `await act`, and both outside `act` with only a microtask awaited afterwards. React flushes passive effects at every one of those boundaries, so the window is real in a browser and unreachable here.

I would rather say that than add a case that passes either way and call it coverage. The existing controlled test still pins the gate itself; what no test pins is the ordering, and the comment now names it.

PR #1038, opened 2026-09-10
feat(dashboard): move the routing policy form into FormDialog

jigjigjig:
## Description

Creating a routing policy appended a form to the page; editing one appended the same form again, lower down. Both open the dialog now.

**Edit comes with create**, because they are one component. The same dialog opens on an existing row with the title "Edit policy" and "Save" as its submit. The old heading carried the policy's name and its user scope inline (`Edit policy `fast` for user `alice``); `FormDialog`'s `title` is a string, so those moved into the description, where the mono face says they are a value rather than prose.

**The trigger is now "Create policy"**, not "New policy", so the control that starts the flow and the one that completes it say the same words. The heading's action no longer hides while the dialog is open, and the empty state gained the same action while keeping its five-step explanation.

**`lg` on every routing dialog.** This form grows a fallback chain, a condition tier, a weighted split and a guardrail list as they are asked for, and a frame that changed width while an operator built one up would read as a different dialog each time. `PolicyForm.stories.tsx` shows the longest policy this form can author: measured, 1849px of content in a 511px body, with the header and footer pinned.

**The old action row carried four sentences after its buttons**, which a footer has no room for. The one that is about the save, "In effect for new requests within 30s.", is the footer's caption. The other three each explain a mode chosen above them (a learned router, a weighted split, an alias that has outgrown itself) and stay beside that choice in the body. No wording changed.

`FormDialog` itself is untouched. The three props this needed (`size`, `isSubmitDisabled`, `isDirty`) all exist. The dirty guard compares every piece of editable state against what the form was seeded with, rather than watching the name alone: this is the form where a stray Escape can land ten minutes into building a chain.

## How to test it locally

    pnpm --dir web run storybook

`Features → Routing → PolicyForm`: `New`, `LongestPolicy` (the scrolling body), and `WithoutGuardrailsService`.

    pnpm --dir web run build
    pnpm --dir web exec playwright test --project=seed --project=screenshots-desktop-small-light --project=screenshots-mobile-light --grep "routing policy dialog"

Two new captures across three viewports and both themes: the dialog at rest, and the dialog once a fallback chain is summoned, which is the shortest route to a body taller than the frame.

- `pnpm --dir web test` — 5318 pass across 153 files; 62 in `RoutingPage.test.tsx`, 4 of them new.
- `storybook:build` then `node .storybook/smoke.mjs` — 381 stories x 2 themes, all rendered clean.

| Remove | Fails |
| --- | --- |
| the trigger's visibility while the dialog is open | `keeps the page's create action visible while the dialog is open` |
| the empty state's action | `offers the same dialog from the empty state` |
| the name and scope from the description | `names the object in the title and the policy in the description when editing` |
| `isDirty`, or narrow it to the name alone | `guards a half-built policy against a stray Escape` |

### On the unslop pass

Run over this diff before opening. It removed: a comment left describing the wrong declaration after an insertion split it from its code; three comments that described the change rather than the code ("the other half of the old heading", "stay in the body rather than following the buttons", "the trigger no longer hides"); the same four-line comment repeated three times across two e2e specs, cut to two lines; and an `as never` on the story's policy fixture, replaced with a real `PolicySpec` built from `select` entries, which is also what made the fixture a true one.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Part of #1031

Stacked on #1037, which is stacked on #1032. Review those first; this branch contains their commits.

## Checklist

      62 cases in `RoutingPage.test.tsx` (4 new, the rest updated for the new
      surface), 3 Storybook stories, 2 screenshot captures. The dashboard's tests
      live under `web/src` and `web/e2e` rather than the repo-root `tests/`.
      No design-doc change: the placement, label and surface rules were written
      down in #1032 and this page follows them.
      No API change. The four write paths send exactly what they sent before.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

One thing to look at rather than take on trust: on a 390px viewport the footer's caption wraps to three lines beside the two buttons, which is tight. It is legible and both controls are reachable, and it is noted rather than fixed because the remedy is either hiding information from phone users or changing the shared footer.

## Summary

- Moved routing policy creation and editing into a shared `FormDialog`.
- Replaced inline API-key creation and secret reveal flows with dialogs.
- Kept create actions visible while dialogs are open and added create actions to empty states.
- Added clearer edit labels, descriptions, scrolling layouts, and unsaved-change protection.
- Added shared `CopyField` reveal controls and API-key caption handling.
- Updated tests, Storybook stories, screenshots, and Playwright coverage.
- No API contract changes.

## Technical notes

- Extracted routing form logic into `PolicyForm` and shared policy helpers.
- Added `FormDialog` support for non-dismissable dialogs and disabled submission.
- Added controlled and initially revealed states to `CopyField`.


web/src/design-system/feedback/FormDialog.tsx
```diff
@@ -0,0 +1,332 @@
...
+              </div>
+            ) : null}
+            <form
+              // A real form, so a label associates, a password manager sees a
+              // submit, and Enter in a single-line field submits the way it
+              // does everywhere else. Cmd/Ctrl+Enter is added on top, for the
+              // fields where Enter is a newline.
+              onSubmit={(event) => {
+                event.preventDefault()
+                if (!isPending) onSubmit()
```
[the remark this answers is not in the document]
jigjigjig: Correct, and already fixed upstream rather than here. This PR carries the keys PR's commits, so `FormDialog.tsx` in this diff is the inherited copy and the fix landed on #1037: both handlers read `if (!isPending && !isGuarding && !isSubmitDisabled)`, and it reaches this branch on its next rebase.

The same finding arrived from a human reviewer on #1037 independently, which is worth saying because you were both right about the mechanism: `requestSubmit()` consults constraint validation and not the submit button's `disabled` state, so the shortcut reached the mutation while a click and a plain Enter were refused.

Regression tests exist for both states, colocated in `FormDialog.test.tsx`: `isSubmitDisabled` with a click and then both Cmd+Enter and Ctrl+Enter asserting `onSubmit` was never called, a case without the prop asserting Cmd+Enter does submit, and the guard case asserting the same for `isGuarding` with a return to submitting once Keep editing is pressed. Removing either condition fails exactly its own case.

web/src/features/keys/KeysPage.tsx
```diff
@@ -469,10 +463,39 @@ function CreateKeyForm({
...
+
+  const resetForm = () => {
+    setKeyName("")
+    setExpiresAt("")
+    setShowAdvanced(false)
+    setAllowedModels(null)
+    setRejectUserMismatch(null)
+    setScopeValid(true)
+    create.reset()
+  }
```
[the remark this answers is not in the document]
jigjigjig: Real, and confirmed against the current head rather than this branch's inherited copy. Fixed on #1037 in 47d3af58; it reaches here on the next rebase.

Your diff is right except for one line: `createAnother` no longer exists. The checkbox on the create form was removed by product ruling, so there is no repeated-creation state to retain, and the only way to create another is the button on the secret step, which resets the form after the key has been shown.

There is a second symptom you did not name, and it is the one an operator would have hit first: because `userId` survived and the dirty check included it, a reopened dialog was dirty on arrival, so Escape on a form nobody had touched armed the unsaved-changes guard.

Rather than adding two setters, `resetForm` and the dirty check now come off one predicate declared beside it. They are the same list of what the form owns read two ways, and they had already drifted twice, which is the actual defect behind both this thread and the next one.

web/src/features/keys/KeysPage.tsx
```diff
@@ -494,46 +517,103 @@ function CreateKeyForm({
...
+      title="New key"
+      description="The secret is shown once, right after you create it."
+      submitLabel="Create key"
+      onSubmit={submit}
+      isPending={create.isPending}
+      isSubmitDisabled={blocked}
+      error={create.error}
+      isDirty={
+        keyName.trim() !== "" || expiresAt !== "" || userId.trim() !== ""
+      }
```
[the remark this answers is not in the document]
jigjigjig: Real, and the same root cause as the thread above: the check listed three of the seven values the form owns, so a backdrop press or Escape discarded model restrictions, the budget exemption and the mismatch behavior without asking.

Fixed on #1037 in 47d3af58, and deliberately not by extending the boolean. There is one `isPristine` predicate beside `resetForm` naming every field, and `isDirty` is its negation, so the reset list and the dirty list cannot drift apart again. That drift is what produced both findings.

One field is in neither, with the reason stated in the code: `showAdvanced` reveals fields rather than holding a value, and a disclosure left open is not unsaved work.

Covered by a case that ticks the budget exemption, which the old check could not see, presses Escape, and asserts the guard armed. Reverting to the three-field check fails exactly that case.

web/src/styles/globals.css
```diff
@@ -2133,6 +2133,54 @@ main {
...
+ * HeroUI's own `data-slot` because that is the element carrying the cap. */
+.otari-form-dialog [data-slot="textfield"] {
+  max-width: none;
+}
+
+/* Below 640px every size is a full-screen sheet: a centered box with a phone's
+ * margins around it wastes the only axis a form needs, and the keyboard takes
+ * half of what is left. The footer clears the home indicator. */
+@media (max-width: 639px) {
+  .otari-form-dialog {
```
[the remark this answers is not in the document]
jigjigjig: Right, and the sharpest of the four. Fixed on #1037 in 47d3af58; it reaches here on the next rebase.

Your arithmetic holds and the cause is slightly worse than you had it: the container's lower padding was `sm:pb-10` until a review on the parent PR ruled the 120px gap symmetric, which made it `sm:pb-[7.5rem]`. So the cap and the container agree now and both are wrong at a short viewport, where the pre-ruling value would at least have left 40px more.

The ruling is that the sheet keys on either dimension rather than gaining a third layout: `@media (width <= 639px), (height <= 639px)`. A viewport that cannot afford the 120px gap in either axis gets the full-screen sheet, with the same stacked footer. The container's `sm:` paddings are overridden inside that same query, which is the part your comment is right to insist on: without it the sheet would sit inside a 240px vertical inset and none of it would help.

Measured on a build at both orientations rather than quoted from the rule:

| viewport | dialog | container padding | header | footer | body |
| --- | --- | --- | --- | --- | --- |
| 844x390 | 844x390, `max-height: none` | 0 / 0 | 86px | 133px | **171px** |
| 390x844 | 390x844 | 0 / 0 | 106px | 133px | **605px** |

Landscape was a 150px dialog holding about 137px of header and footer before this. There is a `PhoneSheetLandscape` story at 844x390 so the case is in the catalog rather than only in a comment.

web/src/features/keys/KeysPage.tsx
```diff
@@ -494,46 +517,103 @@ function CreateKeyForm({
...
-        // Capture the secret into the reveal BEFORE closing the form, so a render
-        // hiccup can never drop the one-time key.
-        onCreated(result)
-        onClose()
+        // "Create another" trades the secret step for a cleared form. The row is
+        // already in the table either way; what is given up is the one chance to
+        // read this key, which is why the checkbox says so.
+        if (createAnother) {
+          resetForm()
+          return
```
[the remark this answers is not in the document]
jigjigjig: On the record, since this thread is resolved and outdated and the reason is not in it.

The branch is gone. The "Create another" Checkbox on the create form was removed at the product owner's request, and `onSuccess` no longer has a path that returns before `setCreated(result)`, so there is no longer a way to create a key and not be shown its secret.

Worth stating plainly rather than leaving as a copy change: your reading is that the path did not merely skip the secret step but discarded a live credential, since the server cannot return `result.key` again and the operator never saw it. That makes the removal a bug fix. The only "Create another" left is the ghost button on the secret step, which resets the form *after* the key has been handed over.

web/design/layout.md
```diff
@@ -102,6 +103,27 @@ select on change. Reach for it when the settings are independent of one another,
 so no single button could say what it is about to write.
 `features/tools/ToolsGuardrailsPage` is the worked example.
 
+A list-and-detail page, which is the shape for a set of records where reading
+one is most of the work (`features/routing/RoutingPage` is the worked example):
+
+```
```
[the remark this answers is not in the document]
jigjigjig: Declining this one on this PR, and it is worth saying why precisely, because the finding is not wrong so much as pointed at the wrong tree.

`web/design/layout.md` is not in this PR's diff. Line 109 of it on `main` *is* a bare opening fence, so the location was real; it is simply not something this change touches, and fixing it here would put an unrelated docs edit in a routing PR.

The class is closed elsewhere: https://github.com/mozilla-ai/otari/pull/1067 labels every bare fenced block in the repo, seventeen of them across twelve files, `layout.md`'s four included. It also records what this PR cannot: there is no markdownlint config and no CI job running it, so nothing was failing and MD040 reached us through your linters rather than the repo's.

web/src/features/routing/PolicyForm.stories.tsx
```diff
@@ -0,0 +1,127 @@
...
+  fields: [{ key: "guardrails_url", value: "https://guardrails.internal" }],
+}
+
+const meta = {
+  title: "Features/Routing/PolicyForm",
+  component: PolicyForm,
+  parameters: {
+    layout: "fullscreen",
+    api: {
+      [`${API_ROOT}/models/discoverable`]: CATALOG,
```
khaledosman, line 51: **These mocks cannot resolve until #1052 lands, so both stories render the same thing.** They key on `${API_ROOT}`, which is `/api/v1` (`shared/api/client.ts:28`), but `.storybook/apiRouting.ts:14` on this branch still has `const GATEWAY_PREFIX = "/v1/"` and line 94 returns `{ kind: "network" }` for anything outside it. So `/tool-settings` never answers, `guardrails_.configured` is false in `New` as well as in `WithoutGuardrailsService`, and the two stories are identical.

`dd4e074e` on `fix/storybook-api-prefix` (#1052) is the fix. Nothing here is wrong once that is in the base; it just must not land ahead of it. Worth saying in the PR description so the ordering is not implicit.

While you are there: both create stories render `ScopePicker` (`RoutingPage.tsx:820`), which calls `useUsers()` unconditionally. Once the prefix is fixed that becomes a deliberate 501 ("No story mock for …") on every render. Add `[`${API_ROOT}/users`]: []` to both tables.

web/src/features/routing/RoutingPage.tsx
```diff
@@ -658,621 +676,619 @@ function PolicyForm({
...
+          <div>
+            <span className="text-body">Always check</span>
+            <p className="text-caption">
+              Runs on every request through this policy. Callers can add their
+              own guardrails but cannot weaken these.
+            </p>
+            {guardrails_.configured ? null : (
+              <p className="mt-1 text-caption text-warning">
+                No guardrails service is configured, so these cannot run. With
+                `if the service is down` set to block, every request through
```
khaledosman: **A fourth literal backtick pair, inside the form.** "With \`if the service is down\` set to block" renders the backticks on screen. `ee780166` fixed three of these and scoped itself to the form; this one is in the form. Use `<code>` or plain quotes, whichever the other three settled on.

web/src/features/routing/RoutingPage.tsx
```diff
@@ -394,7 +394,7 @@ function ModeToggle({
  *  until summoned rather than collapsed-and-empty, which keeps naming one model a
  *  three-field task.
  */
-function PolicyForm({
+export function PolicyForm({
```
khaledosman: **`PolicyForm` now has a public export and a catalog entry but still lives in the page module.** `PolicyForm.stories.tsx:5` imports `./RoutingPage`, which drags `DataTable`, `ConfirmDialog`, `RouterReadiness` and the page's whole hook graph into every story render. component-architecture.md is explicit that a page composes rather than implements, and the story's own filename already claims `features/routing/PolicyForm.tsx`. Giving a 900-line dialog body its own catalog entry is the moment to move it.

Related, and cheaper to decide now than later: `:1624-1640` mounts the form only while open, so this dialog cannot play the exit animation the rest of the stack gets. HeroUI drives the exit off `data-exiting`, which needs the component mounted while `isOpen` flips false, and `CreateKeyDialog` stays mounted for exactly that (`KeysPage.tsx:522-526`). Either hold `isOpen` as a prop, or say in the comment at `:684` that the pop-out is deliberate here.

web/src/features/routing/RoutingPage.test.tsx
```diff
@@ -328,6 +328,27 @@ afterEach(() => {
...
 })
 
+/**
+ * The page's own create action, scoped to the heading's own header.
+ *
+ * Scoped rather than resolved by name, because "Create policy" is on screen up
+ * to three times: this one, the dialog's submit, and the empty state's action,
+ * which repeats the trigger's words where every sibling page deliberately does
+ * not ("Create your first key", "Create a workspace").
+ *
```
khaledosman, line 361: **`createTrigger()` is now ambiguous on an empty list, and passes only by winning a race.** The docstring says resolving while the dialog is shut is what makes it unambiguous, but `:1642-1650` puts a second "Create policy" on the page whenever `rows.length === 0`, with no `isAdding` gate. `findByRole` throws on multiple matches, so every `mockApi([])` test here (`:378`, `:432`, `:457`) depends on the heading button appearing one query ahead of the empty state.

Scope it, the way the e2e specs handled the same collision: resolve inside the `PageIntro` band, or `getAllByRole(...)[0]`.

web/src/features/routing/RoutingPage.tsx
```diff
@@ -1621,9 +1639,16 @@ export function RoutingPage() {
...
       ) : null}
 
-      {rows.length === 0 && !isListLoading && !isAdding ? (
+      {rows.length === 0 && !isListLoading ? (
         canEdit ? (
-          <EmptyState title="No routing policies yet">
+          <EmptyState
+            title="No routing policies yet"
+            actionLabel="Create policy"
+            onAction={() => {
```
khaledosman: **The new empty-state action drops keyboard focus once the first policy exists.** Creating from here fills the table, so the empty state unmounts and the node react-aria stored for focus restoration is gone; focus resets to `<body>` and the next Tab starts at the top of the document.

`KeysPage.tsx:515-527` solves the identical case for its empty-state CTA with a `returnFocusRef` plus a `requestAnimationFrame` check for `document.activeElement === document.body`, and its comment names this cause. Same pattern here.

web/src/features/routing/RoutingPage.tsx
```diff
@@ -658,621 +676,619 @@ function PolicyForm({
...
+      }
+      submitLabel={editing ? "Save" : "Create policy"}
+      onSubmit={submit}
+      isPending={pending}
+      isSubmitDisabled={!canSubmit || outgrewAlias}
+      isDirty={isDirty}
+      // All four writers, not two: an organization-scoped save fails through its
+      // own mutation, and with those two missing the refusal was swallowed and
+      // the panel just sat there.
+      error={
```
khaledosman: **The save error moved into the dialog with no test.** The delete equivalent has one (`RoutingPage.test.tsx:808`, "reports a refused delete inside the dialog, leaving the row"), and the failure this wiring guards against is the silent one: a refused save showing nothing at all. The comment says that has already happened for the two organization writers. A `$status: 400` on `POST /routing/policies` asserted inside `getByRole("dialog")` covers it.

web/src/features/routing/PolicyForm.stories.tsx
```diff
@@ -0,0 +1,127 @@
+import type { Meta, StoryObj } from "@storybook/react-vite"
+
+import type { PolicySpec } from "@/client/local"
```
khaledosman: `PolicySpec` from `@/client/local` is the only such import in the tree. `@/client` re-exports it (`client/index.ts:573`) and the component this stories imports it from there (`RoutingPage.tsx:5-10`). web/AGENTS.md names `@/client` as the entry point.

khaledosman (approved):
Gates clean: no route, schema or docstring touched; no `schema.ts` / `routeTree.gen.ts` drift; no dependency change; conventional title; no em dashes or `--` separators. CI green, including the two routing e2e tests (`parity.management.spec.ts:250` and the `dashboard.spec.ts` policy flows) at this head.

The two things worth checking across this stack are both right here and not findings: the form needs no open-counter key, because `PolicyForm` mounts only while open and holds its whole draft and all four mutation hooks inside itself (`:1624-1640`, `:416-419`); and `canSubmit` (`:532-542`) carries no pending term, so the primary keeps its fill while the mutation runs.

One merge-order item and five smaller things below. Nothing blocking the code.

---

web/src/features/routing/RoutingPage.tsx
jigjigjig: Both halves taken, in two commits: b55ac15b for the move, 1efb22ac for the mount.

**The move.** `features/routing/PolicyForm.tsx` holds the form and the six helpers only it uses. The seam is not quite where it looks, so I measured it rather than cutting: the form reads **fifteen** things declared outside itself, of which **nine** are also read by the page (`RoutingRow` ten times, `defaultTargetOf` six). Importing those back out of `RoutingPage` would invert the dependency into exactly what this thread objects to, so they went to `features/routing/policyModel.ts`, imported by both sides. Two more joined them once the first pass compiled, `candidatesOf` and `normalizedBackend`, because the shared derivations call them.

Verified as a pure move by inventory rather than by eye: **26 top-level declarations before, 26 after**, none lost and none invented, 7 left on the page, 11 in the model, 8 in the form.

One thing a move like this can break quietly, and the design gate caught it: the caption-size exemption is a named ruling about consequence notes beside a button, and four of its five sites moved. So the ruling names both files now, rather than the visuals changing inside a refactor.

**The mount.** You were right that this was the one dialog that could not play its exit. `isOpen` is a prop, the page keys the form on an open counter, and both triggers go through one `openCreate`.

And staying mounted reproduced the defect found on the keys page from the other direction: the form's own reads landed on the page. The roster and the tool settings are the form's, not the table's, so both are gated on `isOpen`, and `ScopePicker` takes the flag since it holds the roster query. Worth saying that your own neighbouring test is what caught it: "never fires the operator reads for a non-operator" went red the moment the form mounted unconditionally, and removing the gating fails it again alongside the new case.

**One thing I did not change, with the reasoning.** `isDirty` here is already `draft !== seeded.current`, a serialization of all ten fields, and the form owns exactly ten `useState`. That is stronger than a predicate of empties, and it is what makes the edit case correct: dirty has to mean "changed from what was loaded", not "not blank". A predicate would report an edit form dirty on arrival.

web/src/features/routing/RoutingPage.tsx
jigjigjig: Both halves taken, in two commits: b55ac15b for the move, 1efb22ac for the mount.

**The move.** `features/routing/PolicyForm.tsx` holds the form and the six helpers only it uses. The seam is not quite where it looks, so I measured it rather than cutting: the form reads **fifteen** things declared outside itself, of which **nine** are also read by the page (`RoutingRow` ten times, `defaultTargetOf` six). Importing those back out of `RoutingPage` would invert the dependency into exactly what this thread objects to, so they went to `features/routing/policyModel.ts`, imported by both sides. Two more joined them once the first pass compiled, `candidatesOf` and `normalizedBackend`, because the shared derivations call them.

Verified as a pure move by inventory rather than by eye: **26 top-level declarations before, 26 after**, none lost and none invented, 7 left on the page, 11 in the model, 8 in the form.

One thing a move like this can break quietly, and the design gate caught it: the caption-size exemption is a named ruling about consequence notes beside a button, and four of its five sites moved. So the ruling names both files now, rather than the visuals changing inside a refactor.

**The mount.** You were right that this was the one dialog that could not play its exit. `isOpen` is a prop, the page keys the form on an open counter, and both triggers go through one `openCreate`.

And staying mounted reproduced the defect found on the keys page from the other direction: the form's own reads landed on the page. The roster and the tool settings are the form's, not the table's, so both are gated on `isOpen`, and `ScopePicker` takes the flag since it holds the roster query. Worth saying that your own neighbouring test is what caught it: "never fires the operator reads for a non-operator" went red the moment the form mounted unconditionally, and removing the gating fails it again alongside the new case.

**One thing I did not change, with the reasoning.** `isDirty` here is already `draft !== seeded.current`, a serialization of all ten fields, and the form owns exactly ten `useState`. That is stronger than a predicate of empties, and it is what makes the edit case correct: dirty has to mean "changed from what was loaded", not "not blank". A predicate would report an edit form dirty on arrival.

web/src/features/routing/RoutingPage.test.tsx
jigjigjig: Fixed, and confirmed in this branch's tree rather than assumed: the helper resolves inside the header that holds the heading, and the inner query awaits, because the action is gated on a query and arrives after the heading.

Your diagnosis was exact, including the part I would not have guessed: the thirteen callers were passing by winning a race. They ran while the list was still loading, so the empty state had not rendered and the name was momentarily unique. Adding `await screen.findByText("No routing policies yet")` before any one of them turns it red with "Found multiple elements", which is how I confirmed it rather than taking the green at face value.

The duplicate name is gone as well, which removes the collision rather than scoping around it: the empty state says "Create your first policy" now, matching the convention its siblings already follow ("Create your first key", "Create your first budget", "Create a workspace"). Probed on the empty state afterwards: one control named "Create policy" and one named "Create your first policy", where there had been two of the former.

The scoping stays regardless. It is the right query whether or not a name happens to be unique, and it is what stops these tests depending on when anything renders.

web/src/features/routing/RoutingPage.tsx
jigjigjig: Both halves taken, in two commits: b55ac15b for the move, 1efb22ac for the mount.

**The move.** `features/routing/PolicyForm.tsx` holds the form and the six helpers only it uses. The seam is not quite where it looks, so I measured it rather than cutting: the form reads **fifteen** things declared outside itself, of which **nine** are also read by the page (`RoutingRow` ten times, `defaultTargetOf` six). Importing those back out of `RoutingPage` would invert the dependency into exactly what this thread objects to, so they went to `features/routing/policyModel.ts`, imported by both sides. Two more joined them once the first pass compiled, `candidatesOf` and `normalizedBackend`, because the shared derivations call them.

Verified as a pure move by inventory rather than by eye: **26 top-level declarations before, 26 after**, none lost and none invented, 7 left on the page, 11 in the model, 8 in the form.

One thing a move like this can break quietly, and the design gate caught it: the caption-size exemption is a named ruling about consequence notes beside a button, and four of its five sites moved. So the ruling names both files now, rather than the visuals changing inside a refactor.

**The mount.** You were right that this was the one dialog that could not play its exit. `isOpen` is a prop, the page keys the form on an open counter, and both triggers go through one `openCreate`.

And staying mounted reproduced the defect found on the keys page from the other direction: the form's own reads landed on the page. The roster and the tool settings are the form's, not the table's, so both are gated on `isOpen`, and `ScopePicker` takes the flag since it holds the roster query. Worth saying that your own neighbouring test is what caught it: "never fires the operator reads for a non-operator" went red the moment the form mounted unconditionally, and removing the gating fails it again alongside the new case.

**One thing I did not change, with the reasoning.** `isDirty` here is already `draft !== seeded.current`, a serialization of all ten fields, and the form owns exactly ten `useState`. That is stronger than a predicate of empties, and it is what makes the edit case correct: dirty has to mean "changed from what was loaded", not "not blank". A predicate would report an edit form dirty on arrival.

web/src/features/routing/RoutingPage.tsx
jigjigjig: Both halves taken, in two commits: b55ac15b for the move, 1efb22ac for the mount.

**The move.** `features/routing/PolicyForm.tsx` holds the form and the six helpers only it uses. The seam is not quite where it looks, so I measured it rather than cutting: the form reads **fifteen** things declared outside itself, of which **nine** are also read by the page (`RoutingRow` ten times, `defaultTargetOf` six). Importing those back out of `RoutingPage` would invert the dependency into exactly what this thread objects to, so they went to `features/routing/policyModel.ts`, imported by both sides. Two more joined them once the first pass compiled, `candidatesOf` and `normalizedBackend`, because the shared derivations call them.

Verified as a pure move by inventory rather than by eye: **26 top-level declarations before, 26 after**, none lost and none invented, 7 left on the page, 11 in the model, 8 in the form.

One thing a move like this can break quietly, and the design gate caught it: the caption-size exemption is a named ruling about consequence notes beside a button, and four of its five sites moved. So the ruling names both files now, rather than the visuals changing inside a refactor.

**The mount.** You were right that this was the one dialog that could not play its exit. `isOpen` is a prop, the page keys the form on an open counter, and both triggers go through one `openCreate`.

And staying mounted reproduced the defect found on the keys page from the other direction: the form's own reads landed on the page. The roster and the tool settings are the form's, not the table's, so both are gated on `isOpen`, and `ScopePicker` takes the flag since it holds the roster query. Worth saying that your own neighbouring test is what caught it: "never fires the operator reads for a non-operator" went red the moment the form mounted unconditionally, and removing the gating fails it again alongside the new case.

**One thing I did not change, with the reasoning.** `isDirty` here is already `draft !== seeded.current`, a serialization of all ten fields, and the form owns exactly ten `useState`. That is stronger than a predicate of empties, and it is what makes the edit case correct: dirty has to mean "changed from what was loaded", not "not blank". A predicate would report an edit form dirty on arrival.

web/src/features/routing/RoutingPage.tsx
jigjigjig: Correction: my previous reply on this thread was posted in error, by a script that matched threads on filename and put the module-move answer on every `RoutingPage.tsx` thread. Ignore it here; the answer to this finding is below.

Fixed in 8f71a74d. It reads `<code>if the service is down</code>` now, matching what `ee780166` settled on for the other three rather than inventing a fourth spelling.

You were right that its scoping is what let this survive: that commit scoped itself to the form, and this one is in the form. I checked the rest of the file afterwards, and every remaining backtick pair is inside a comment, which is where they belong.

web/src/features/routing/RoutingPage.tsx
jigjigjig: Correction: my previous reply on this thread was posted in error, by a script that matched threads on filename and put the module-move answer on every `RoutingPage.tsx` thread. Ignore it here; the answer to this finding is below.

Fixed in 8f71a74d, with the pattern you pointed at.

A ref on the page's own trigger, and the close checks a frame later whether `document.activeElement` is still `document.body`, so anything with a real claim has had its turn first. Your diagnosis of the cause is what made it a two-line fix rather than a hunt: creating the first policy fills the table, the empty state unmounts, and the node react-aria stored for restoration goes with it.

Covered by a case that creates from the empty state and asserts the page's trigger has focus once the dialog closes. Removing the restoration fails exactly that case.

web/src/features/routing/RoutingPage.tsx
jigjigjig: Correction: my previous reply on this thread was posted in error, by a script that matched threads on filename and put the module-move answer on every `RoutingPage.tsx` thread. Ignore it here; the answer to this finding is below.

Test added in 8f71a74d, and `mockApi` gained a `saveBody` beside its `deleteBody` to make it possible.

The case posts a policy against a refusing endpoint and asserts the detail inside `getByRole("dialog")`, with the form still filled so the operator can correct the name rather than retype the policy. Modelled on the delete equivalent you named.

The negative control is worth stating, because the wiring is a four-way `??` chain and a test can pass while watching the wrong branch: replacing `save.error` with `undefined` in that chain fails exactly this case and nothing else. So it is watching the branch the comment above it claims.

web/src/features/routing/PolicyForm.stories.tsx
jigjigjig: Both halves fixed in 8f71a74d.

The ordering is real and I have not left it implicit: `#1052` is what makes these mocks resolve, and until it is in the base both stories render the same thing for the reason you give. It is open and green.

And the roster is mocked in both tables now, with the reason in a comment: `ScopePicker` asks for it on every render, so once the prefix is fixed the mock would answer its deliberate 501 rather than the network's 404. That is the failure your note predicted one step further on, and it would have looked like a story bug rather than a missing mock.

web/src/features/routing/PolicyForm.stories.tsx
jigjigjig: Fixed in 8f71a74d: the story imports `PolicySpec` from `@/client`, which is the entry point `web/AGENTS.md` names, and which is where the component itself takes it from.

I checked whether the type should move rather than just the import, since the obvious reading is that a `local.ts` shape belongs in the generated schema. It does not: `PolicySpec` appears **zero** times in `schema.ts`, and `local.ts` carries ten lines above the declaration explaining why OpenAPI cannot own it. The request needs a permissive form or FastAPI's generic 422 swallows the detail, and the real model carries defaults that would add a router entry to every policy on the way out.

So the type stays where it is and only the import path was wrong.

web/src/features/routing/RoutingPage.tsx
```diff
@@ -1337,6 +226,28 @@ export function RoutingPage() {
...
+  const createButtonRef = useRef<HTMLButtonElement | null>(null)
+  const [createCount, setCreateCount] = useState(0)
+  const openCreate = () => {
+    setEditing(null)
+    setCreateCount((n) => n + 1)
+    setAdding(true)
+  }
+  const closeCreate = () => {
+    setAdding(false)
+    requestAnimationFrame(() => {
```
khaledosman: **Keeping the form mounted for the exit is what makes this fallback unreachable.** The rAF fires about 16ms after `setAdding(false)`. By then `react-aria-components` has *not* unmounted the overlay subtree, because it keeps it while exiting (`Modal.mjs:96`, `if (!state.isOpen && !isExiting || isSSR) return null`), and HeroUI's exit runs `duration-100` on `[data-exiting="true"]` (`modal.css:56`, `:125`). The dialog still holds focus, `document.activeElement === document.body` is false, the fallback is skipped, and ~85ms later the `FocusScope` unmounts onto a `nodeToRestore` that no longer exists.

Failure: on the empty state, press "Create your first policy", fill the form, submit. The dialog fades out and the next Tab starts at the top of the document. That is the bug the fallback was written for.

The suite cannot catch it: jsdom reports no animation, `useExitAnimation` returns false, the subtree unmounts in the same commit, focus is on `body` at the rAF, and `RoutingPage.test.tsx:848` passes.

Run the check where the content actually goes away rather than on a timer: an effect-cleanup child inside `FormDialog`'s children (the shape `KeysPage.tsx:150-162` uses for `RevealSecretStrip`), or give `FormDialog` a `returnFocusRef` and do the rAF in its modal body's cleanup. The latter is worth more, since every page in this series has the same empty-state CTA.

web/src/features/routing/RoutingPage.tsx
```diff
@@ -1603,14 +514,17 @@ export function RoutingPage() {
...
+      <PolicyForm
+        key={createCount}
+        existing={null}
+        initialTarget={initialTarget}
+        isOpen={isAdding}
+        workspaceId={writeWorkspaceId}
+        onClose={closeCreate}
+      />
       {editing !== null ? (
         <PolicyForm
```
khaledosman, line 527: **The edit dialog has no `key`, though `PolicyForm`'s own prop docstring says it does.** `PolicyForm.tsx:220-222`: "The edit path mounts on the row it is editing and keyed on its id, the way the keys page does". `KeysPage.tsx:1205` really does carry `key={editingKey.id}`; this render site passes none, and the fields seed from props through mount-only `useState`.

Latent today, because the modal backdrop makes a second row's Edit unpressable so `setEditing` cannot fire while one is mounted. It stops being latent the moment anything reaches a second row: a keyboard path, an Edit from the readiness panel, a non-modal edit surface. Then row A's draft opens in row B's form and the submit writes A's spec under B's name.

Either add the key or drop the claim from both docstrings; a docstring describing code that is not there is the worse of the two.

web/src/features/routing/PolicyForm.tsx
```diff
@@ -0,0 +1,1103 @@
...
+/** Who a policy applies to. Same control and wording as the alias scope picker,
+ *  because it is the same decision. */
+function ScopePicker({
+  userId,
+  onChange,
+  enabled,
+}: {
+  userId: string | null
+  onChange: (userId: string | null) => void
+  /**
```
khaledosman, line 103: **Two comments describe mechanisms that cannot occur.**

Here: "which would otherwise leave the roster fetching on every render of the page" is not what `enabled` prevents. `ScopePicker` renders inside `FormDialog`'s children and `ModalOverlay` returns `null` while closed, so `useUsers` never mounts with the dialog shut. What `enabled={isOpen}` actually buys is suppressing a refetch during the ~100ms exit. The gate that *is* load-bearing for "the form's own reads" is `useGuardrailsConfigured(isOpen)` at `:236`, which sits in `PolicyForm`'s body outside the modal; move the rationale there.

And `RoutingPage.tsx:430-435`: "The table stays mounted while the create form is open, so Edit is still reachable from it" was true of the inline form. The create surface is now a `Modal` with a full-screen backdrop (`FormDialog.tsx:150`) plus `ariaHideOutside`, so a row's Edit is neither clickable nor in the accessibility tree. The `setAdding(false)` there is harmless; the reason given for it is wrong.

web/src/features/routing/PolicyForm.tsx
```diff
@@ -0,0 +1,1103 @@
+/**
+ * The routing policy form, in the dialog every create and edit opens in.
+ *
+ * Its own module rather than the page's, because it has a catalog entry: a
+ * story importing it from `RoutingPage` dragged the table, the confirm dialog
+ * and the page's whole hook graph into every render. The model both sides read
+ * lives in `policyModel.ts`.
+ */
+
+import { Button } from "@heroui/react"
```
khaledosman: `Button` from `@heroui/react` in a new file with four `variant="ghost"` call sites. `actions.md` and the SKILL both name `@/design-system/actions/Button` as the one to reach for: it narrows the variant union so a retired variant is a compile error rather than a silently unstyled button. Carried over from `RoutingPage.tsx:1` rather than invented, but the split re-authored this import block, which is the moment to change it.

web/src/features/routing/policyModel.ts
```diff
@@ -0,0 +1,91 @@
+/**
```
khaledosman, line 1: A new pure-derivation module with no colocated test. `sharesOf`, `initialPool`, `normalizedBackend` and `defaultTargetOf` are what component-architecture.md calls the part worth unit testing directly, and they are currently reached only through a 70-case page test. `sharesOf`'s zero-total branch and `initialPool`'s append-the-fallthrough branch are a line each. Extracting them was the right call; this is the other half of it.

khaledosman (changes requested):
Six of seven fixed cleanly, and the module split is a real improvement: `PolicyForm.tsx` and `policyModel.ts` carry their own weight, the create form is keyed on an open counter that both triggers bump, all four mutations sit below the key, and the save-error test does what the delete one already did.

The seventh is the problem. The `returnFocusRef` plus `requestAnimationFrame` fallback landed in `fa0b54e4` and then `b41eed06` made it inert in a browser: keeping the form mounted so the frame can play its exit is exactly what stops the rAF from ever seeing an empty `document.body`. So the focus bug I raised is still there, behind a fix that reads correct and that jsdom cannot fail.

On finding 1, only the `/users` half is in your hands and it is done. The prefix half still needs #1052 first: `apiRouting.ts:14` is `GATEWAY_PREFIX = "/v1/"` at this head and #1052 is still open, so the three stories render an empty catalog and `WithoutGuardrailsService` is indistinguishable from `New`. Fair that this is tree-wide rather than yours (`RouterReadiness.stories.tsx` on main has it too), but the merge order is worth a line in the PR body.

Gates clean, CI green (70/70 routing tests).

---

web/src/features/routing/PolicyForm.stories.tsx
```diff
@@ -0,0 +1,131 @@
...
+            // owner's budget is most of the way spent.
+            when: { budget_used_pct: { gte: 80 } },
+            target: "anthropic:claude-3-5-haiku-latest",
+          },
+          {
+            router: "weighted",
+            candidates: ["openai:gpt-4o", "anthropic:claude-sonnet-4-5"],
+            weights: { "openai:gpt-4o": 70, "anthropic:claude-sonnet-4-5": 30 },
+            default: "openai:gpt-4o",
+          },
```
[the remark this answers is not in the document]
jigjigjig: Fixed in 4b73d2c3: the fixture now writes `{ router: "weighted", … }` and then `{ default: "openai:gpt-4o" }` as its own last entry, with a comment saying why the fallthrough is last. That is the shape `PolicyForm` writes, so the fixture no longer teaches a shape the form never produces, and nothing built on it later reasons about the `entry.default !== undefined` branch by accident.

web/src/features/routing/PolicyForm.tsx
```diff
@@ -0,0 +1,1103 @@
...
+                    prev.map((c, i) =>
+                      i === index ? { ...c, threshold: Number(value) || 0 } : c,
+                    ),
+                  )
+                }
+                description={
+                  condition.threshold >= 100 ? (
+                    <span className="text-danger">Must be under 100.</span>
+                  ) : undefined
+                }
```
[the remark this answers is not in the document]
jigjigjig: Fixed in 4b73d2c3. Clearing the box coerces to 0 through the `onChange`, `conditionsReady` goes false, and the submit died with nothing on screen. The description now reads "Must be over 0." below zero, beside its existing sibling "Must be under 100.", same line and same tone.

web/src/features/routing/RoutingPage.tsx
jigjigjig: You are right, and the mechanism you name is the whole of it: the rAF fired while the overlay subtree was still mounted for its exit and still holding focus, so the fallback could never see `<body>`.

Fixed in 4b73d2c3 with your second option, since every page in the series has this shape. `FormDialog` takes an optional `returnFocusRef` and renders a child inside the modal subtree whose cleanup does the check, so it runs when the frame is actually gone rather than 16ms after it was asked to close. Routing passes the heading's action, which survives the create; `closeCreate` is back to `setAdding(false)`.

It is also testable now, which the timer version was not: in jsdom the subtree unmounts in the same commit, so the cleanup runs, and `RoutingPage.test.tsx:848` fails when the mechanism is removed. It passed either way before, exactly as you said.

Documented in feedback.md beside the signature, and `returnFocusRef` is listed in `foundation.test.ts`'s `CANNOT_BE_SHOWN` with the reason: a story could pass the prop and would demonstrate nothing, because the point is where focus goes after the trigger has unmounted with the frame.

web/src/features/routing/RoutingPage.tsx
jigjigjig: Fixed in 4b73d2c3 rather than dropping the claim, since the claim is right about what the code should do. One correction to it: a policy has no `id`. It is unique by name within a scope, and the same name can exist deployment-wide and per user, which the page already encodes in its own `rowKeyOf` for the accordion. The edit dialog is keyed on that, and I did not add a second helper for it. Both docstrings now say what identifies a row instead of naming an id that does not exist.

web/src/features/routing/PolicyForm.tsx
jigjigjig: Both rewritten in 4b73d2c3.

`ScopePicker`'s now says what the gate actually buys: the picker renders inside the modal, which is `null` while closed, so the roster never fetches with the dialog shut whatever the prop says; what the gate prevents is a refetch landing during the ~100ms exit. The page-renders rationale moved to `useGuardrailsConfigured(isOpen)`, which is the read that sits in the form's body outside the modal and is load-bearing for it.

`RoutingPage`'s Edit comment no longer claims the table is reachable while the create form is open. It says what is true: the backdrop and `ariaHideOutside` make that impossible, and the `setAdding(false)` is belt and braces for a future surface that reaches a row without going through the modal.

web/src/features/routing/PolicyForm.tsx
jigjigjig: Changed in 4b73d2c3: `Button` now comes from `@/design-system/actions/Button`. You are right that the split re-authored the import block, which is the moment to fix a carried-over import rather than carry it once more. All four `variant="ghost"` call sites compile against the narrowed union.

web/src/features/routing/policyModel.ts
jigjigjig: Added in 4b73d2c3: `policyModel.test.ts`, 16 cases over the seven exports, including the three branches you named. `sharesOf`'s zero-total (a drained policy has to read 0% rather than NaN), `initialPool`'s append-the-fallthrough (a spec written through the API can omit it and the server appends it, so a form that did not would show a plan that is not the one that runs), and `normalizedBackend`'s strip-and-lower.

Each was checked by breaking the implementation: dropping the zero guard, returning the candidates unchanged, and returning the raw name fail four cases between them. The page's 70 cases still pass, so nothing moved to make room for these.

PR #1040, opened 2026-09-10
feat(dashboard): move the budget form into FormDialog

### LINKED ISSUE #1047: Form-grade MultiSelect: keep the search field still and show selection in the list
## Context

The people picker in the budget form (`UserMultiSelect`, used by `BudgetForm` for "Assign to people") stacks selected people as chips *above* its search field. Every pick grows the chip block and pushes the field down under the pointer, and a selected person disappears from the list, so the list never shows who is in. Seen while testing the budget dialog on a seeded environment with 14 users.

The dialog made it visible, but it is not the dialog's bug: the design system has no form-grade multi-select (`FilterMultiComboBox` is for toolbars), so the feature hand-rolled one.

**Design approved 2026-09-10** (Paper artboard 08, three states plus spec). Owner ruling: built into #1040 (budgets → FormDialog) as its own commits, not a separate PR; #1040 closes this issue.

## What to do

Add `design-system/forms/MultiSelect` and make `UserMultiSelect` its first consumer. Design: Paper file `01M25A5MEHAH18XD64PGWXFE7S`, artboard `08 · MultiSelect — form-grade people picker` (three states and the spec block).

- The search field stays at the top and never moves. Selected people render **below** it as `DismissChip`s, so growth pushes down.
- Options are toggles: a selected row stays in the list with a leading check; pressing it again deselects. Row order never changes on a pick.
- Closed field text: `Search people…` when empty, `{n} people assigned` otherwise. Open: the query, with `x of y matches selected · n assigned` in the popover footer.
- Popover capped at 6 rows and scrolls; Escape closes the popover only, never the enclosing dialog.
- Keyboard: arrows move, Space/Enter toggle, Backspace on an empty query removes the last chip.
- `role="listbox"` with `aria-multiselectable`, `aria-selected` per option, chip row is a list named by the field label.
- Field label, description and error go through the same rungs as `Field` (`forms.md`).

## Notes

- Lands in #1040 with the budgets migration; not part of #1031 itself.
- Stories: closed empty, closed with selection, open with query, error state. New design-system props need a story (the catalog gate enforces it).
- Follow `.github/skills/frontend-standards/SKILL.md` and `web/design/forms.md`.

jigjigjig:
## Description

Creating a budget appended a form to the page; editing one appended the same form again. Both open the dialog now.

`BudgetForm` already took its title and submit label as props, so create and edit needed one change between them rather than two.

- **The trigger keeps its words**, "Create budget"; the dialog's title is "New budget".
- **Editing splits the old heading in two.** It was one string, `Edit budget team-free-tier`. "Edit budget" is the title and the budget it names is the description, because a dialog title is a noun phrase rather than a sentence. Its submit is **"Save"** rather than "Save changes", which is what the routing dialog already says: two edit dialogs disagreeing about that word is the inconsistency this series exists to remove.
- **The heading's action no longer hides**, and no longer hides behind the empty state either. It used to do both. A dialog sits over the page, so there is nothing for hiding it to prevent, and it is where focus returns when one closes. The empty state keeps its own call to action beside it.
- **`md`, not `lg`.** Four controls, and the assignment multiselect is the only one that grows; the body scrolls when it does.

One cleanup the migration forced rather than invited: `canSubmit` was a single expression, and passing it straight to `isSubmitDisabled` would have painted the submit as *refused* while a save was in flight, which is the treatment for denied. It is two readings of one rule now. `blocked` is what disables the submit; `canSubmit` adds the in-flight check that `submit` itself refuses on.

`FormDialog` is untouched.

## How to test it locally

    pnpm --dir web run build
    pnpm --dir web exec playwright test --project=seed --project=screenshots-desktop-small-light --project=screenshots-mobile-light --grep "budget dialog"

One new capture across three viewports and both themes.

- `pnpm --dir web test` — 5320 pass across 153 files; 63 in `BudgetsPage.test.tsx`, 2 of them new.

| Remove | Fails |
| --- | --- |
| the trigger's visibility while the dialog is open | `keeps the page's create action visible while the dialog is open` |
| the budget name from the description | `names the object in the title and the budget in the description when editing` |

Two existing tests changed rather than being added to, and both assert the new contract on purpose: `shows onboarding when there are no budgets` used to assert the heading's action was **absent** while the empty state showed, and now asserts it is present; the edit tests moved from "Save changes" to "Save".

### On the unslop pass

Run over this diff before opening. It removed two comments that described the change rather than the code ("The old heading was one string…", "It used to hide here, on the reasoning that…"), and it is what caught the duplicated submit rule described above: the same condition was spelled once in `submit` and again inline in `isSubmitDisabled`, which is two places to update and one to forget.

## MultiSelect for the people picker (#1047)

Two commits beyond the dialog migration, on the owner's ruling that they land here rather than as a PR of their own.

The picker stacked chips **above** its search field, so every pick grew the block and pushed the field down under the pointer, and a picked person left the list, so the list only ever showed who was left rather than who was in. The dialog made it visible; it is not the dialog's bug.

`design-system/forms/MultiSelect` is the form-grade half of the pair `forms.md` already describes, with `FilterMultiComboBox` staying the toolbar one. Chips render below the field. A selected option stays listed with a check and toggles off when pressed again, and the order never re-sorts, so a second press lands where the first one did. `features/users/UserMultiSelect` is its first consumer and keeps only what a design-system component cannot know: which rows are people and what a person is called.

Written to the APG combobox-with-listbox pattern rather than on HeroUI's `ComboBox`, which selects one value and closes. Focus stays in the input and the active option travels by `aria-activedescendant`. Escape closes the popover and stops there, because the same key reaches the dialog and would otherwise arm its guard.

Seven cases in `MultiSelect.test.tsx`, and the three the issue names were each checked by breaking them: moving the chips above the field fails the position test, filtering selected options out of the list fails the stays-listed test, and letting Escape bubble fails the guard test.

One user-visible name change: the field's visible label is now its accessible name. It carried a hidden "Add a person" beside a heading that labelled nothing, so one control had two names; three references in the suites moved with it.

Closes #1047

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Part of #1031

Stacked on #1038 → #1037 → #1032. Review those first; this branch contains their commits.

## Checklist

      63 cases in `BudgetsPage.test.tsx` (2 new, 3 updated to the new contract),
      1 screenshot capture. The dashboard's tests live under `web/src` and
      `web/e2e` rather than the repo-root `tests/`.
      No design-doc change: the rules this page now follows were written down in
      #1032.
      No API change. The create and update bodies are unchanged.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

"Save changes" becoming "Save" is the one user-visible word this changes beyond the title split, and it is a deliberate alignment with the routing dialog rather than a rewrite.

## Summary

- Moved budget create and edit forms into `FormDialog`.
- Kept page actions visible while dialogs and empty states are active.
- Added a form-grade `MultiSelect` and migrated `UserMultiSelect` to it.
- Added stable search, selected chips, visible checked options, keyboard support, accessibility semantics, and capped scrolling.
- Improved dialog behavior for validation, submission, dismissal, mobile layouts, and draft reset.
- Updated related keys and routing dialogs to use the shared dialog patterns.
- Added stories, unit tests, end-to-end tests, and screenshots.

These changes provide consistent modal forms, clearer interactions, and a more usable people picker.

### DIFF
--- a/web/e2e/dashboard.spec.ts
+++ b/web/e2e/dashboard.spec.ts
@@ -1,7 +1,6 @@
 import { expect, test } from "@playwright/test"
 import { API_ROOT } from "@/shared/api/client"
 import {
-  dismissComboBox,
   dismissComboBoxInDialog,
   login,
   MASTER_KEY,
@@ -163,7 +162,7 @@ test.describe("dashboard core flows", () => {
     })
     await owners.fill("alice@example.com")
     await page.getByRole("option", { name: /alice@example\.com/ }).click()
-    await dismissComboBox(owners)
+    await dismissComboBoxInDialog(owners)
     await editDialog.getByRole("button", { name: "Save" }).click()
 
     // The budget now reports one holder in its People column, which is the
--- a/web/e2e/dashboard.spec.ts
+++ b/web/e2e/dashboard.spec.ts
@@ -154,7 +154,7 @@ test.describe("dashboard core flows", () => {
     const budgetRow = page.getByRole("row", { name: /e2e-budget/ })
     await budgetRow.getByRole("button", { name: "Edit" }).click()
     // The field's visible label is its accessible name now: the picker used to
-    // carry a hidden "Add a person" beside a heading that labelled nothing, so
+    // carry a hidden "Add a person" beside a heading that labeled nothing, so
     // one control had two names.
     const editDialog = page.getByRole("dialog")
     const owners = editDialog.getByRole("combobox", {
--- a/web/e2e/dashboard.spec.ts
+++ b/web/e2e/dashboard.spec.ts
@@ -118,9 +118,12 @@ test.describe("dashboard core flows", () => {
     await openOrganization(page)
     await nav(page).getByRole("link", { name: "Spend & budgets" }).click()
     await page.getByRole("button", { name: "Create your first budget" }).click()
-    await page.getByLabel("Name (optional)").fill("e2e-budget")
-    await page.getByLabel("Spending limit (USD)").fill("100")
-    await page.getByRole("button", { name: "Create budget" }).click()
+    // Scoped: the heading's trigger and the dialog's submit both say "Create
+    // budget", so an unscoped press is ambiguous.
+    const dialog = page.getByRole("dialog")
+    await dialog.getByLabel("Name (optional)").fill("e2e-budget")
+    await dialog.getByLabel("Spending limit (USD)").fill("100")
+    await dialog.getByRole("button", { name: "Create budget" }).click()
 
     // The shared table renders on react-aria, so non-row-header cells are gridcells.
     await expect(page.getByRole("gridcell", { name: "$100.00" })).toBeVisible()
@@ -155,11 +158,12 @@ test.describe("dashboard core flows", () => {
     // the section heading beside it and labels nothing. Same shape as
     // `addFilterValue` otherwise: the popover aria-hides the rest of the page, so
     // it has to be put away before the submit button is reachable.
-    const owners = page.getByRole("combobox", { name: "Add a person" })
+    const editDialog = page.getByRole("dialog")
+    const owners = editDialog.getByRole("combobox", { name: "Add a person" })
     await owners.fill("alice@example.com")
     await page.getByRole("option", { name: /alice@example\.com/ }).click()
     await dismissComboBox(owners)
-    await page.getByRole("button", { name: "Save changes" }).click()
+    await editDialog.getByRole("button", { name: "Save" }).click()
 
     // The budget now reports one holder in its People column, which is the
     // assignment landing.
--- a/web/e2e/dashboard.spec.ts
+++ b/web/e2e/dashboard.spec.ts
@@ -154,12 +154,13 @@ test.describe("dashboard core flows", () => {
     await nav(page).getByRole("link", { name: "Spend & budgets" }).click()
     const budgetRow = page.getByRole("row", { name: /e2e-budget/ })
     await budgetRow.getByRole("button", { name: "Edit" }).click()
-    // "Add a person" is the input's own accessible name; "Assign to people" is
-    // the section heading beside it and labels nothing. Same shape as
-    // `addFilterValue` otherwise: the popover aria-hides the rest of the page, so
-    // it has to be put away before the submit button is reachable.
+    // The field's visible label is its accessible name now: the picker used to
+    // carry a hidden "Add a person" beside a heading that labelled nothing, so
+    // one control had two names.
     const editDialog = page.getByRole("dialog")
-    const owners = editDialog.getByRole("combobox", { name: "Add a person" })
+    const owners = editDialog.getByRole("combobox", {
+      name: "Assign to people (optional)",
+    })
     await owners.fill("alice@example.com")
     await page.getByRole("option", { name: /alice@example\.com/ }).click()
     await dismissComboBox(owners)
--- a/web/src/design-system/forms/MultiSelect.stories.tsx
+++ b/web/src/design-system/forms/MultiSelect.stories.tsx
@@ -1,0 +1,123 @@
+import type { Meta, StoryObj } from "@storybook/react-vite"
+import { useState } from "react"
+
+import { MultiSelect } from "./MultiSelect"
+
+const PEOPLE = [
+  { id: "operator", label: "Operator" },
+  { id: "alice@example.com", label: "alice@example.com" },
+  { id: "parity-heavy@example.com", label: "parity-heavy@example.com" },
+  { id: "parity-light@example.com", label: "parity-light@example.com" },
+  { id: "pat@example.com", label: "pat@example.com (Pat Okafor)" },
+  { id: "priya@example.com", label: "priya@example.com (Priya Raghavan)" },
+  { id: "sam@example.com", label: "sam@example.com (Sam Iyer)" },
+  { id: "wei@example.com", label: "wei@example.com (Wei Zhang)" },
+]
+
+/**
+ * Pick several without the field moving.
+ *
+ * Controlled, like every other control here, so each story owns the selection it
+ * shows. The interesting states are what the selection does to the layout, so
+ * they differ by how much is picked rather than by a prop.
+ */
+const meta = {
+  title: "Design system/Forms/MultiSelect",
+  component: MultiSelect,
+  parameters: { layout: "padded" },
+  args: {
+    label: "Assign to people (optional)",
+    description:
+      "Everyone selected is held to this budget, each with their own allowance rather than a shared pool.",
+    options: PEOPLE,
+    value: [],
+    onChange: () => {},
+    searchPlaceholder: "Search people…",
+    countNoun: "people assigned",
+  },
+} satisfies Meta<typeof MultiSelect>
+
+export default meta
+
+type Story = StoryObj<typeof meta>
+
+function Live({
+  initial,
+  ...args
+}: React.ComponentProps<typeof MultiSelect> & { initial?: string[] }) {
+  const [value, setValue] = useState<string[]>(initial ?? [])
+  return <MultiSelect {...args} value={value} onChange={setValue} />
+}
+
+/** Nothing picked. The field reads as a search box, because that is all it is. */
+export const Empty: Story = {
+  render: (args) => <Live {...args} />,
+}
+
+/**
+ * Six picked. The field has not moved: the chips are below it, so the block
+ * grew downward and the control the operator is pointing at stayed put.
+ */
+export const WithSelection: Story = {
+  render: (args) => (
+    <Live
+      {...args}
+      initial={[
+        "operator",
+        "alice@example.com",
+        "parity-heavy@example.com",
+        "parity-light@example.com",
+        "pat@example.com",
+        "priya@example.com",
+      ]}
+    />
+  ),
+}
+
+/**
+ * The list open. A picked person stays listed with a check rather than
+ * disappearing, so the list answers "who is in" and not only "who is left", and
+ * the row order does not change on a pick.
+ *
+ * Open it and type to see the footer count follow the query. Stories cannot
+ * carry an open popover on their own: it opens on focus, which a static render
+ * has not had.
+ */
+export const Open: Story = {
+  render: (args) => (
+    <Live {...args} initial={["parity-heavy@example.com", "operator"]} />
+  ),
+}
+
+/** A refusal, on the same rung a `Field`'s sits on. */
+export const Invalid: Story = {
+  render: (args) => (
+    <Live
+      {...args}
+      isInvalid
+      errorMessage="Pick at least one person, or leave the budget unassigned."
+      reserveMessage
+    />
+  ),
+}
+
+/**
+ * Nothing matches the query, which is a different sentence from having nothing
+ * to offer: one is about what was typed, the other about the deployment.
+ */
+export const NoMatches: Story = {
+  args: {
+    noMatchesMessage: "Nobody matches what you typed.",
+  },
+  render: (args) => <Live {...args} />,
+}
+
+/** Nothing to offer, which is a different sentence from nothing matching. */
+export const NoOptions: Story = {
+  args: {
+    options: [],
+    emptyMessage:
+      "Nobody to assign yet. Add people under Members & roles, or issue a key, and they can be assigned here.",
+  },
+  render: (args) => <Live {...args} />,
+}
--- a/web/src/design-system/forms/MultiSelect.stories.tsx
+++ b/web/src/design-system/forms/MultiSelect.stories.tsx
@@ -1,5 +1,5 @@
 import type { Meta, StoryObj } from "@storybook/react-vite"
-import { useState } from "react"
+import { useEffect, useRef, useState } from "react"
 
 import { MultiSelect } from "./MultiSelect"
 
@@ -33,7 +33,7 @@ const meta = {
     value: [],
     onChange: () => {},
     searchPlaceholder: "Search people…",
-    countNoun: "people assigned",
+    countNoun: { one: "person assigned", other: "people assigned" },
   },
 } satisfies Meta<typeof MultiSelect>
 
@@ -43,10 +43,32 @@ type Story = StoryObj<typeof meta>
 
 function Live({
   initial,
+  query,
   ...args
-}: React.ComponentProps<typeof MultiSelect> & { initial?: string[] }) {
+}: React.ComponentProps<typeof MultiSelect> & {
+  initial?: string[]
+  /** Typed into the field on mount, for the story that needs a query behind it. */
+  query?: string
+}) {
   const [value, setValue] = useState<string[]>(initial ?? [])
-  return <MultiSelect {...args} value={value} onChange={setValue} />
+  const ref = useRef<HTMLDivElement>(null)
+  useEffect(() => {
+    if (!query) return
+    const input = ref.current?.querySelector("input")
+    if (!input) return
+    // Through the native setter, so React sees the change rather than only the
+    // DOM: the same reason the e2e helpers fill controlled inputs this way.
+    Object.getOwnPropertyDescriptor(
+      window.HTMLInputElement.prototype,
+      "value",
+    )?.set?.call(input, query)
+    input.dispatchEvent(new Event("input", { bubbles: true }))
+  }, [query])
+  return (
+    <div ref={ref}>
+      <MultiSelect {...args} value={value} onChange={setValue} />
+    </div>
+  )
 }
 
 /** Nothing picked. The field reads as a search box, because that is all it is. */
@@ -75,17 +97,20 @@ export const WithSelection: Story = {
 }
 
 /**
- * The list open. A picked person stays listed with a check rather than
- * disappearing, so the list answers "who is in" and not only "who is left", and
- * the row order does not change on a pick.
+ * The list open, which `autoFocus` is what makes a static render able to show:
+ * the popover opens on focus. A picked person stays listed with a check rather
+ * than disappearing, so the list answers "who is in" and not only "who is
+ * left", and the row order does not change on a pick.
  *
- * Open it and type to see the footer count follow the query. Stories cannot
- * carry an open popover on their own: it opens on focus, which a static render
- * has not had.
+ * Type to see the footer count follow the query.
  */
 export const Open: Story = {
   render: (args) => (
-    <Live {...args} initial={["parity-heavy@example.com", "operator"]} />
+    <Live
+      {...args}
+      autoFocus
+      initial={["parity-heavy@example.com", "operator"]}
+    />
   ),
 }
 
@@ -108,14 +133,17 @@ export const Invalid: Story = {
 export const NoMatches: Story = {
   args: {
     noMatchesMessage: "Nobody matches what you typed.",
+    // The story is the popover saying it, so it has to be open and querying.
+    autoFocus: true,
   },
-  render: (args) => <Live {...args} />,
+  render: (args) => <Live {...args} query="zzz" />,
 }
 
 /** Nothing to offer, which is a different sentence from nothing matching. */
 export const NoOptions: Story = {
   args: {
     options: [],
+    autoFocus: true,
     emptyMessage:
       "Nobody to assign yet. Add people under Members & roles, or issue a key, and they can be assigned here.",
   },
--- a/web/src/design-system/forms/MultiSelect.stories.tsx
+++ b/web/src/design-system/forms/MultiSelect.stories.tsx
@@ -114,7 +114,11 @@ export const Open: Story = {
   ),
 }
 
-/** A refusal, on the same rung a `Field`'s sits on. */
+/**
+ * A refusal, on the same rung a `Field`'s sits on, and in the description's
+ * place rather than under it: the error replaces that line, so going invalid
+ * moves nothing.
+ */
 export const Invalid: Story = {
   render: (args) => (
     <Live
--- a/web/src/design-system/forms/MultiSelect.tsx
+++ b/web/src/design-system/forms/MultiSelect.tsx
@@ -224,11 +224,6 @@ export function MultiSelect({
       <label htmlFor={inputId} className="text-body">
         {label}
       </label>
-      {description ? (
-        <p id={descriptionId} className="text-caption">
-          {description}
-        </p>
-      ) : null}
       {/* `relative` so the popover hangs off the field rather than off the
           dialog, and the chips below it stay in flow. */}
       <div ref={wrapperRef} className="relative flex flex-col gap-1.5">
@@ -246,13 +241,14 @@ export function MultiSelect({
             // Wired by hand, because this does not go through HeroUI's
             // Description and FieldError slots: without it a screen reader
             // says "invalid" and never says why.
+            // Whichever line is actually rendered, since the error replaces
+            // the description rather than joining it.
             aria-describedby={
-              [
-                description ? descriptionId : null,
-                isInvalid && errorMessage ? errorId : null,
-              ]
-                .filter(Boolean)
-                .join(" ") || undefined
+              isInvalid && errorMessage
+                ? errorId
+                : description
+                  ? descriptionId
+                  : undefined
             }
             className={`${INPUT_CLASS} w-full pr-9`}
             // The closed field says how many are in rather than staying blank,
@@ -420,10 +416,19 @@ export function MultiSelect({
       <span aria-live="polite" className="sr-only">
         {counted(value.length)}
       </span>
+      {/* One rung, below the control, the way `Field` does it: the error
+          replaces the description rather than adding a row. Below the chips
+          rather than above the field, because that is where a `Field` puts it
+          and a dialog full of `Field`s should not have one control speaking
+          from somewhere else. */}
       <FieldMessages reserve={reserveMessage}>
         {isInvalid && errorMessage ? (
           <span id={errorId} className="text-danger">
             {errorMessage}
+          </span>
+        ) : description ? (
+          <span id={descriptionId} className="text-muted">
+            {description}
           </span>
         ) : null}
       </FieldMessages>
--- a/web/src/design-system/forms/MultiSelect.tsx
+++ b/web/src/design-system/forms/MultiSelect.tsx
@@ -1,4 +1,4 @@
-import { type ReactNode, useId, useRef, useState } from "react"
+import { type ReactNode, useEffect, useId, useRef, useState } from "react"
 import { FiCheck, FiChevronDown } from "react-icons/fi"
 
 import { DismissChip } from "../indicators/DismissChip"
@@ -86,12 +86,38 @@ export function MultiSelect({
   const [query, setQuery] = useState("")
   const [isOpen, setIsOpen] = useState(false)
   const [activeIndex, setActiveIndex] = useState(0)
-  const inputRef = useRef<HTMLInputElement>(null)
+  // The control's own box, which is what "focus left this control" means. Held
+  // as a ref rather than walked up from the input, so wrapping the input in one
+  // more div cannot silently close the popover before an option press lands.
+  const wrapperRef = useRef<HTMLDivElement>(null)
+  // The list, so the active option can be scrolled into view: the popover shows
+  // six rows and the cap renders fifty, so the highlight walks off the bottom.
+  const listRef = useRef<HTMLDivElement>(null)
   const listId = useId()
   const inputId = useId()
   const descriptionId = useId()
   const errorId = useId()
   const optionId = (index: number) => `${listId}-option-${index}`
+
+  // The popover shows six rows and renders up to `maxVisible`, so the active
+  // row walks past the fold on its own. Scrolled here rather than by the row,
+  // because only the list knows which one is active.
+  useEffect(() => {
+    if (!isOpen) return
+    // The id is spelled here rather than through `optionId`, which is a fresh
+    // arrow every render and so cannot be a dependency.
+    const active = listRef.current?.querySelector(
+      `[id="${listId}-option-${activeIndex}"]`,
+    )
+    // Guarded because jsdom implements no layout and so no `scrollIntoView`;
+    // the tests below drive the arrows and would throw on the first press.
+    if (
+      active instanceof HTMLElement &&
+      typeof active.scrollIntoView === "function"
+    ) {
+      active.scrollIntoView({ block: "nearest" })
+    }
+  }, [isOpen, activeIndex, listId])
 
   const needle = query.trim().toLowerCase()
   // Filtered, never re-ordered: a selected row keeps its place, so a second
@@ -156,8 +182,12 @@ export function MultiSelect({
       if (!isOpen) return
       // Before the row lookup, not after: an open combo box owns Enter whether
       // or not it has a row to give, and this input sits inside a real form, so
-      // falling through submits it.
+      // falling through submits it. Stopped as well as prevented, because
+      // `FormDialog` submits on Cmd/Ctrl+Enter from anywhere inside the form:
+      // without this, one gesture toggles a row and posts the selection from
+      // before that toggle.
       event.preventDefault()
+      event.stopPropagation()
       const option = matches[activeIndex]
       if (!option) return
       toggle(option.id)
@@ -201,10 +231,9 @@ export function MultiSelect({
       ) : null}
       {/* `relative` so the popover hangs off the field rather than off the
           dialog, and the chips below it stay in flow. */}
-      <div className="relative flex flex-col gap-1.5">
+      <div ref={wrapperRef} className="relative flex flex-col gap-1.5">
         <div className="relative">
           <input
-            ref={inputRef}
             id={inputId}
             role="combobox"
             aria-expanded={isOpen}
@@ -245,11 +274,7 @@ export function MultiSelect({
               // Only when focus actually left the control: a press on an option
               // blurs the input and must not close the list before the press
               // lands.
-              if (
-                !event.currentTarget.parentElement?.parentElement?.contains(
-                  event.relatedTarget,
-                )
-              ) {
+              if (!wrapperRef.current?.contains(event.relatedTarget)) {
                 setIsOpen(false)
               }
             }}
@@ -267,6 +292,7 @@ export function MultiSelect({
                 both a lint error and a second, conflicting announcement. The
                 chip row below IS a real list, because that one is a list. */}
             <div
+              ref={listRef}
               id={listId}
               role="listbox"
               aria-multiselectable
@@ -274,11 +300,13 @@ export function MultiSelect({
               // carrying one name is three things a screen reader cannot tell
               // apart, and it is what a query for the field then finds.
               aria-label={`${label}, options`}
-              // Six rows before it scrolls, each at the 44px touch floor
-              // (motion-and-access.md: "44px is the floor, everywhere"), so
-              // 6 x 44px = 264px = `max-h-66`. A class rather than an inline
-              // style: the arithmetic is a compile-time constant.
-              className="max-h-66 overflow-y-auto"
+              // Six rows before it scrolls. Each is 44px, the touch floor
+              // (motion-and-access.md: "44px is the floor, everywhere"), and
+              // all but the first carry a 1px rule, so six of them are
+              // 6 x 44 + 5 = 269px. `max-h-[16.8125rem]` rather than the 264px
+              // the rows alone would suggest, which clipped the sixth by 5px.
+              // A class rather than an inline style: it is a constant.
+              className="max-h-[16.8125rem] overflow-y-auto"
             >
               {matches.length === 0 ? (
                 <p className="text-caption px-2.5 py-2">
@@ -353,9 +381,13 @@ export function MultiSelect({
               )}
             </div>
             <div className="border-border text-mono-caption text-subtle flex h-8 items-center justify-between border-t px-2.5">
+              {/* Two facts, two clauses. Nested, they read as one garbled
+                  sentence: "0 of 3 matches selected of 8". */}
               <span>
-                {selectedMatches} of {matches.length} matches selected
-                {capped ? ` of ${reached.length}` : ""} ·{" "}
+                {capped
+                  ? `Showing ${matches.length} of ${reached.length} · `
+                  : ""}
+                {selectedMatches} of {matches.length} selected here ·{" "}
                 {counted(value.length)}
               </span>
               <span>ESC closes</span>
--- a/web/src/design-system/forms/MultiSelect.tsx
+++ b/web/src/design-system/forms/MultiSelect.tsx
@@ -10,8 +10,8 @@ export interface MultiSelectOption {
   label: string
 }
 
-/** Six rows at 36px, which is where the list starts scrolling. */
-const MAX_VISIBLE_ROWS = 6
+/** How many matches are rendered at once, however many the query reaches. */
+const DEFAULT_MAX_VISIBLE = 50
 
 /**
  * Pick several from a searchable list, without the field moving.
@@ -43,7 +43,9 @@ export function MultiSelect({
   searchPlaceholder = "Search…",
   emptyMessage = "Nothing to choose from.",
   noMatchesMessage = "Nothing matches what you typed.",
-  countNoun = "selected",
+  countNoun = { one: "selected", other: "selected" },
+  maxVisible = DEFAULT_MAX_VISIBLE,
+  autoFocus,
 }: {
   label: string
   description?: ReactNode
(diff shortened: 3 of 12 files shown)


web/src/design-system/forms/MultiSelect.tsx
```diff
@@ -0,0 +1,303 @@
...
+      )
+      return
+    }
+    if (event.key === "Enter" || event.key === " ") {
+      // Space only toggles when the query is empty; otherwise it is a character
+      // somebody is typing into a name.
+      if (event.key === " " && query !== "") return
+      if (!isOpen) return
+      const option = matches[activeIndex]
+      if (!option) return
```
khaledosman, line 192: **Enter with the list open but no active option submits the form.** `if (!option) return` returns without `preventDefault()`, and this is a bare single-line `<input>` inside `FormDialog`'s real `<form>` (`feedback/FormDialog.tsx:206`), so the browser's implicit submission fires.

Failure: open "New budget", fill name and limit, type `zzz` into the people picker, press Enter expecting the empty list to do nothing. The budget is created with no assignees and the dialog closes. Same path any time `matches[activeIndex]` is out of range, which happens whenever `options` changes under a stale `activeIndex` (the roster resolving while the picker is open).

An open combobox owns Enter whether or not it has a row to give. Hoist the `preventDefault()`:

```ts
if (event.key === "Enter" || event.key === " ") {
  if (event.key === " " && query !== "") return
  if (!isOpen) return
  event.preventDefault()
  const option = matches[activeIndex]
  if (!option) return
  toggle(option.id)
  return
}
```

web/src/features/budgets/BudgetsPage.tsx
```diff
@@ -823,31 +838,34 @@ function DeploymentBudgetsPage() {
...
-          assignUsers={users.data ?? []}
-          onSubmit={createAndAssign}
-          onClose={() => {
-            setAssignmentError(null)
-            setPendingAssignments(null)
-            setAddOpen(false)
-          }}
-        />
-      ) : null}
+      <BudgetForm
```
khaledosman: **The create dialog's draft survives close and reopen.** `BudgetForm` went from `{addOpen ? … : null}` to an unconditional mount. `FormDialog` unmounts its body, but `name` / `limit` / `durationSec` / `userIds` (`:263-271`) live in `BudgetForm`, one level outside the dialog, and their initializers run once at page mount.

Failure: create "team-a / $100 / 3 people". Reopen "Create budget": prefilled with all of it, and one press creates a duplicate. Worse, `seeded` (`:282`) is a ref captured on the first render, so `isDirty` is permanently true afterwards and Escape on a freshly-opened untouched dialog arms the unsaved-changes guard.

overlays.md names this exactly: "leaving them mounted carries one row's draft into the next row's dialog."

#1041 already fixes it, with `key={addOpenCount}` at its `BudgetsPage.tsx:852` and the counter bumped in `openCreate`. Move that hunk down into this PR: the two merge in order, so as it stands this lands the defect on `main` and takes it back out one PR later.

web/src/design-system/forms/MultiSelect.tsx
```diff
@@ -0,0 +1,303 @@
...
+              isOpen && matches[activeIndex] ? optionId(activeIndex) : undefined
+            }
+            aria-invalid={isInvalid}
+            className={`${INPUT_CLASS} w-full pr-9`}
+            // The closed field says how many are in rather than staying blank,
+            // because the chips below it can be scrolled past in a long form.
+            placeholder={
+              value.length === 0
+                ? searchPlaceholder
+                : `${value.length} ${countNoun}`
```
khaledosman: **`countNoun` is never pluralized, and the test locks the wrong grammar in.** The JSDoc says "Pluralized into the closed field and the footer"; both sites interpolate it verbatim (here and `:271`), and `UserMultiSelect.tsx:59` passes `"people assigned"`. One selection renders `1 people assigned`. It is already in CI output: the #1045 e2e log shows `placeholder="1 people assigned"`.

`MultiSelect.test.tsx:129` asserts that string, so the test would not have caught it. Take `{ one, other }` or `(n: number) => string`, pass `person assigned` / `people assigned`, and fix the test.

web/src/features/users/UserMultiSelect.tsx
```diff
@@ -36,111 +30,35 @@ export function UserMultiSelect({
...
-      (user: User): string => {
-        const member = memberLabels.get(user.user_id)
-        if (member) return member
-        return user.alias ? `${user.user_id} (${user.alias})` : user.user_id
-      },
-    [memberLabels],
-  )
-
-  const options = useMemo<Option[]>(
+  const options = useMemo(
```
khaledosman: **The 50-row render cap was dropped in the rewrite.** The old implementation ended in `.slice(0, MAX_VISIBLE)`; `MultiSelect` filters (`MultiSelect.tsx:76-81`) and renders every match, with no `maxVisible`. `useUsers` walks `fetchAllUsers`, 100 pages of 1000 (`shared/api/users.ts:5-20`).

Failure: on a 20,000-user deployment, focusing the picker mounts 20,000 `role="option"` nodes inside a modal and every keystroke re-renders them. `FilterMultiComboBox` still caps at 50 for this reason (`navigation/FilterMultiComboBox.tsx:13,46`).

Give `MultiSelect` a `maxVisible` (default 50) applied after the filter, and say "showing 50 of 300" in the footer that already reports counts.

While here: the `useMemo` on this line has no measurement behind it and the React Compiler is on. The file was rewritten in this commit, so this is the moment to drop it.

web/src/design-system/forms/MultiSelect.tsx
```diff
@@ -0,0 +1,303 @@
...
+    <div className="flex flex-col gap-1.5">
+      {/* A real label with `htmlFor`, not HeroUI's `Label`: that one resolves
+          its target through a field context this control is not inside, so it
+          would render the words and name nothing. The visible text IS the
+          accessible name here, which is why there is no `aria-label` beside
+          it. */}
+      <label htmlFor={inputId} className="text-body">
+        {label}
+      </label>
+      {description ? <p className="text-caption">{description}</p> : null}
```
khaledosman: **The description and the error are not announced on the control.** Neither the description `<p>` nor the `errorMessage` span (`:296-300`) carries an `id`, and the input has no `aria-describedby` / `aria-errormessage`. `aria-invalid` is set (`:168`), so a screen reader says "invalid" and never says why.

forms.md's form-vs-filter split turns on precisely this: the form half "owns a description and an error announced on the control", which is what separates this from `FilterMultiComboBox`. `Field` and `ComboBoxField` get it from HeroUI's `Description` / `FieldError` slots, which this deliberately does not use (`:145-149`), so it has to be wired by hand: two more `useId()`s, and `aria-describedby` on the input set to whichever are present.

web/src/design-system/forms/MultiSelect.tsx
```diff
@@ -0,0 +1,303 @@
...
+                      // stop for a keyboard that is already driving the list
+                      // from the input.
+                      onMouseDown={(event) => {
+                        // Before blur, so the press is not lost to the list
+                        // closing under it.
+                        event.preventDefault()
+                        toggle(option.id)
+                      }}
+                      onMouseEnter={() => setActiveIndex(index)}
+                      className={`flex h-9 cursor-pointer items-center gap-2.5 border-border-subtle px-2.5 text-sm not-first:border-t ${
```
khaledosman: **Option rows are 36px, under the 44px floor, with no written argument.** motion-and-access.md:55: "44px is the floor, everywhere", and :69: "A new control under 44px needs the same kind of argument, in writing, or it is a bug." `DismissChip` is the one documented exception and says so in its own docstring; this row says nothing. On the 390px viewport the screenshot matrix captures, these are six adjacent 8px-under targets and a near-miss assigns the wrong person.

`min-h-11`, and derive the popover's pixel budget from the same number rather than the hardcoded `36` at `:219`.

Also on this line: a row that is both selected and active carries `bg-surface-alt` and `bg-surface-subtle` at once, and which wins is Tailwind's emitted order rather than the class string. `forms/inputClass.ts:16-19` documents that hazard as the reason `w-full` came out of `INPUT_CLASS`. It resolves the way you want today only because `alt` sorts before `subtle`. Pick one with an if/else chain.

web/src/design-system/forms/MultiSelect.tsx
```diff
@@ -0,0 +1,303 @@
...
+            <div
+              id={listId}
+              role="listbox"
+              aria-multiselectable
+              // Named apart from the field and from the chip row: three things
+              // carrying one name is three things a screen reader cannot tell
+              // apart, and it is what a query for the field then finds.
+              aria-label={`${label}, options`}
+              className="overflow-y-auto"
+              style={{ maxHeight: `${MAX_VISIBLE_ROWS * 36}px` }}
```
khaledosman: `style={{ maxHeight: `${MAX_VISIBLE_ROWS * 36}px` }}` is on the SKILL's Never list. The two precedents in `design-system/` (`metrics/Meter.tsx:16`, `metrics/SpendMeter.tsx:92-103`) carry a runtime percentage no class can express; this is a compile-time constant. `className="max-h-[13.5rem] overflow-y-auto"`, and keep the row arithmetic in the comment that already explains it.

web/src/design-system/forms/MultiSelect.tsx
```diff
@@ -0,0 +1,303 @@
...
+                        ) : null}
+                      </span>
+                      {option.label}
+                    </div>
+                  )
+                })
+              )}
+            </div>
+            <div className="border-border text-mono-caption text-subtle flex h-8 items-center justify-between border-t px-2.5">
+              <span aria-live="polite">
```
khaledosman: **The live region is created together with its content, so it announces nothing.** `aria-live="polite"` sits on a span that only exists while `isOpen`. A region inserted with content already in it is not announced, so the first count after opening is silent; and with the list closed there is no region at all, so removing a chip announces nothing. Render the counter (or a visually-hidden mirror) unconditionally, outside the `isOpen` branch.

web/src/design-system/forms/MultiSelect.tsx
```diff
@@ -0,0 +1,303 @@
...
+
+  const toggle = (id: string) => {
+    onChange(
+      value.includes(id)
+        ? value.filter((current) => current !== id)
+        : [...value, id],
+    )
+    // The query survives a pick. Clearing it would refill the list under the
+    // pointer, which is the same movement the chips were moved to avoid.
+    inputRef.current?.focus()
```
khaledosman: **Removing a chip reopens the popover.** `toggle()` unconditionally focuses the input (`:184`), whose `onFocus` is `open`. The chips render outside the popover, so they are pressable while it is closed: pick four people, Escape to put the list away, press "Remove Priya" and the list springs open over the chip row. Four removals cost four extra Escapes.

Refocus only from an option press, or drop the refocus entirely: the option's `onMouseDown` already calls `preventDefault()`, so focus never leaves the input on that path.

Separately, `Tab` has no branch in `onKeyDown` and `onBlur` keeps the popover open while `relatedTarget` is inside the wrapper (`:189-195`), which the chip `<ul>` is. With the list open, Tab moves focus to a chip's Remove button drawn underneath the opaque popover panel: a focus ring on a control the operator cannot see. The APG combobox pattern this docstring claims (`:27`) closes the popup on Tab; `if (event.key === "Tab") setIsOpen(false)`, no preventDefault.

web/src/design-system/forms/MultiSelect.stories.tsx
```diff
@@ -0,0 +1,123 @@
...
+/**
+ * The list open. A picked person stays listed with a check rather than
+ * disappearing, so the list answers "who is in" and not only "who is left", and
+ * the row order does not change on a pick.
+ *
+ * Open it and type to see the footer count follow the query. Stories cannot
+ * carry an open popover on their own: it opens on focus, which a static render
+ * has not had.
+ */
+export const Open: Story = {
```
khaledosman, line 107: **`Open` and `NoMatches` render nothing the other two stories don't.** The popover opens on focus only, as the docstring at `:83-85` concedes, so `Open` is `WithSelection` and `NoMatches` is `Empty`. The catalog is the PR gate and the both-themes sweep, and neither ever shows the listbox, the check mark, the footer count, or either empty sentence.

`ComboBoxField.stories.tsx:116-121` solves this with `autoFocus` plus `menuTrigger="input"`. `MultiSelect` has no `autoFocus` prop at all, which it needs anyway for a form primitive (feedback.md: "The first field takes `autoFocus`"). Add it and set it on these two.

web/src/design-system/forms/MultiSelect.tsx
```diff
@@ -0,0 +1,303 @@
...
+  label: string
+}
+
+/** Six rows at 36px, which is where the list starts scrolling. */
+const MAX_VISIBLE_ROWS = 6
+
+/**
+ * Pick several from a searchable list, without the field moving.
+ *
+ * The form-grade half of the pair `forms.md` describes: `FilterMultiComboBox` is
```
khaledosman, line 26: **The primitive is in no design doc, and forms.md still routes to the wrong sibling.** `git grep MultiSelect -- web/design` at this head returns nothing: forms.md's decision tree answers "Is it many of a set?" with `FilterMultiComboBox` (`forms.md:33-34`), which is the toolbar-grade one this docstring says is wrong for a form; there is no `MultiSelect` row in the Signatures block, and none in DESIGN.md's module table.

That is the whole cost of the miss: the next agent building a form with a many-of-a-set field reads forms.md, reaches for `FilterMultiComboBox`, and ships a caption-labelled filter control that speaks no validation message inside a `FormDialog`, which is the bug this PR exists to fix. Add the branch to the tree, the signature beside `ComboBoxField`, a line on how the two split (forms.md:139-141 already has the shape for `FilterSelect`), and the `forms/MultiSelect` row to DESIGN.md.

khaledosman (changes requested):
Gates clean: no route or schema touched; no generated-artifact drift; no dependency change; conventional title; no em dashes or `--` separators. CI green.

Two blockers. The Enter fall-through in `MultiSelect` submits the enclosing form, and the create dialog's draft survives a close (that second one is fixed by #1041's `key={addOpenCount}`, which should move down here so it never lands on main). The rest of the new primitive's findings are the ones worth taking before it becomes the pattern the next six forms copy: it is a `design-system/` component with no entry in `forms.md`, so whatever ships here is what gets copied.

---

web/src/design-system/forms/MultiSelect.tsx
jigjigjig: Fixed at 995ffc06. The `preventDefault()` is hoisted above the row lookup, so an open list owns Enter whether or not `matches[activeIndex]` resolves.

Covered by a new case that renders the picker inside a real `<form>` with a spy on submit, types a query matching nobody, and presses Enter. It fails with the hoist undone.

web/src/design-system/forms/MultiSelect.tsx
jigjigjig: Fixed at 995ffc06. `countNoun` is now `{ one, other }`, and `UserMultiSelect` passes `person assigned` / `people assigned`.

You were right that the test was locking the grammar in: the assertion on `1 people assigned` is gone, and a new case picks exactly one person and asserts the singular.

web/src/design-system/forms/MultiSelect.tsx
jigjigjig: Fixed at 995ffc06. Two more `useId()`s, the description and the error message carry them, and the input's `aria-describedby` names whichever are present.

A new case reads the attribute, resolves each id, and asserts both texts are reachable through it.

web/src/design-system/forms/MultiSelect.tsx
jigjigjig: Fixed at 995ffc06. `max-h-66` (6 x 44px = 264px), with the arithmetic in the comment. The inline style is gone.

web/src/design-system/forms/MultiSelect.tsx
jigjigjig: Fixed at 995ffc06. The counter is mirrored into an `sr-only` `aria-live` span rendered unconditionally, outside the `isOpen` branch, so it exists before it has anything to announce.

Covered by a case asserting the region is present with the list closed and reads `0 people assigned`.

web/src/design-system/forms/MultiSelect.tsx
jigjigjig: Both fixed at 995ffc06. Rows are `min-h-11`, and the popover's budget is derived from the same number (`max-h-66`, 6 x 44px) rather than from a separate 36.

The double background is gone too: the row picks one in an if/else, active winning over selected, rather than leaving `bg-surface-alt` and `bg-surface-subtle` on one element for Tailwind's emit order to settle. Thank you for that one, it resolved correctly today by accident of alphabetical order.

web/src/design-system/forms/MultiSelect.tsx
jigjigjig: Both fixed at 995ffc06. `toggle` no longer focuses the input at all: an option press keeps focus anyway, because its `onMouseDown` prevents the default that would have moved it, and the chips' Remove is the only other caller. And `Tab` now closes the popup, no `preventDefault`, as the APG pattern says.

A new case picks two people, Escapes the list shut, removes a chip and asserts the listbox is still absent.

web/src/features/users/UserMultiSelect.tsx
jigjigjig: Fixed at 995ffc06. `MultiSelect` takes `maxVisible` (default 50), applied after the filter so the query still sees every option, and the footer says "n of m matches selected of N" when it is capping.

The `useMemo` is gone as well.

A new case renders 8 options with `maxVisible={3}` and asserts three rows plus the footer naming the eight.

web/src/design-system/forms/MultiSelect.stories.tsx
jigjigjig: Fixed at 995ffc06. `MultiSelect` gains `autoFocus`, which a form primitive needs anyway per feedback.md, and both stories set it, so the catalog and the theme sweep now actually show the listbox, the check, the footer and the empty sentence.

`NoMatches` also needed a query behind it, so the story helper types one through the native value setter and an `input` event, the way the e2e helpers fill a controlled input.

web/src/design-system/forms/MultiSelect.tsx
jigjigjig: Fixed at 995ffc06. `forms.md` gets the decision-tree branch (many of a set, in a form vs in a toolbar), the signature beside `ComboBoxField`, and a `## MultiSelect` section covering the form/filter split and the two behaviors that are its own. `DESIGN.md` gets the `forms/MultiSelect` module row.

This was the right one to insist on: the routing was pointing the next form at the toolbar control, which is the bug this PR exists to fix.

web/src/features/budgets/BudgetsPage.tsx
jigjigjig: Fixed at 995ffc06, and taken further than moving the hunk down.

The rule this now follows is that everything which should reset between opens lives below the dialog's key, the create mutation included: `CreateBudgetDialog` owns `useCreateBudget` and the create-and-assign path, the page keys it on an open counter, and the page keeps only what outlives an open (whether it is open, how many times it has been, and the list). Moving just `key={addOpenCount}` down would have left a refusal from one attempt sitting on the mutation to greet the next open, which is the same defect one layer up. The page's own ErrorBanner no longer lists `createBudget.error` either, so a failed create is rendered once rather than twice.

`seeded` is no longer a snapshot string. `isDirty` is the negation of one `isPristine` predicate naming name, limit, period and the assigned people, so the fields it compares cannot drift from the fields the form has.

web/src/features/budgets/BudgetsPage.tsx
```diff
@@ -1002,6 +1017,80 @@ function DeploymentBudgetsPage() {
...
+}) {
+  const createBudget = useCreateBudget()
+
+  // Create the budget, then (optionally) attach it to the chosen users. The
+  // per-user PATCH sets each user's reset clock. Failed assignments stay in the
+  // form so a retry never creates a duplicate budget.
+  const createAndAssign = (body: CreateBudgetRequest, userIds: string[]) => {
+    if (pendingAssignments) {
+      void assignUsers(pendingAssignments.budgetId, pendingAssignments.userIds)
+      return
```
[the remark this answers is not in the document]
jigjigjig: Real, and fixed in 3b909d3d. The retry branch discarded the `assignUsers` promise, so a retry that succeeded left the dialog open with `pendingAssignments` cleared, the label back to "Create budget", and the next press creating a second budget for the same people.

It is awaited now, and `onClose()` runs only when every assignment succeeded. New test scripts the first user PATCH to refuse and the second to succeed, then asserts the form closed and exactly one budget was posted. Restoring the discarded `void assignUsers(...)` fails it.

web/src/styles/globals.css
```diff
@@ -2189,22 +2189,70 @@ main {
...
+    width: 100%;
+  }
+  /* Full width earns the same press suppression `.button` already gives
+   * `.w-full` and HeroUI's `fullWidth` class: the transform displaces a wide
+   * button by 7.07px, and this one spans the sheet. */
+  .otari-form-dialog__actions > button:is(:active, [data-pressed="true"]) {
+    transform: none;
+  }
+  .otari-form-dialog__cancel {
+    order: 1;
```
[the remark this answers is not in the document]
jigjigjig: Already fixed, and by the route you asked for rather than by keeping `order: 1`. There is no `order` property left on the dialog footer anywhere in `globals.css` at this head (3b909d3d); the phone rule stacks `.otari-form-dialog__actions` with `flex-direction: column` and nothing reorders it.

So visual order is DOM order in both layouts, which is the alignment this asks for. Cancel is first in the DOM and first visually, and the primary follows: the decision was that Cancel goes with the dismiss it is, and that the artboard's thumb-reach argument for putting the primary first did not survive the keyboard-order cost.

One note on my own check, since it nearly went the other way: `grep -c "order: 1"` reports two hits in that file, and both are `border: 1px solid`. The count is a substring artifact, not a survival.

web/src/design-system/actions/CopyField.tsx
```diff
@@ -210,7 +248,7 @@ export function CopyField({
         ref.current.select()
       } else {
         selectOnReveal.current = copying
-        setRevealedValue(copying)
+        setRevealed(true)
```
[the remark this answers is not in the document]
jigjigjig: Real, and fixed on #1037 at 99a1da60, which is where the controlled-reveal API lives.

Measured before changing anything, because only one of the two modes has this. Uncontrolled mode is safe: `setRevealed(true)` records `revealedValue = A` and `revealed` is `revealedValue === value`, which is false once the value is B, and there is already a test for it ("does not reveal a credential that arrived while a copy was failing"). Controlled mode throws that keying away, because `onRevealChange?.(next)` carries a bare boolean and cannot say which value the reveal was for; the caller flips `isRevealed` and B is on screen unasked. `selectOnReveal` only stops the selection.

The fix keys the request rather than the API: the latest value is tracked in a ref and the reveal is requested only when the value the copy was for is still the one on screen. A ref rather than the `value` in scope, because `copy` closes over the value of the render the press happened in, so comparing `value === copying` inside it is comparing a variable with itself.

New test is the controlled twin of the uncontrolled one: rotate A to B while the clipboard promise is pending, refuse it, and assert `onRevealChange` was never called and B never reached the DOM. It fails with the gate removed.

It landed on #1037 rather than here because #1037 owns `CopyField`, and a design-system fix does not belong in the budgets PR. This thread will resolve once that lands; the code you are looking at in this diff is #1037's.

web/src/features/budgets/BudgetsPage.tsx
```diff
@@ -818,36 +825,44 @@ function DeploymentBudgetsPage() {
...
+        onClose={() => {
+          setAssignmentError(null)
+          setPendingAssignments(null)
+          setAddOpen(false)
+        }}
+      />
       {/* Key on the row id so switching which budget is edited remounts the form,
           its fields seed from `initial` on mount only. */}
       {editingBudget ? (
         <BudgetForm
```
khaledosman: **The edit dialog still greets a fresh form with the last row's refusal.** This is the create blocker's twin, left standing. `updateBudget` and `assignmentError` live on the page, above `key={editingBudget.budget_id}`, and `error={updateBudget.error ?? assignmentError}` (`:871`) reads them from up there. Neither `onOpenChange` (`:860-862`) nor `onClose` (`:903`) clears either; both only `setEditing(null)`.

Failure: edit budget A, rename it onto a collision, take the 409, press Cancel, press Edit on budget B. B's dialog opens with "A budget named … already exists" in its banner. Worse for the assignment half: `assignmentError` reads "The budget was saved, but these people were not updated: …. Retry to try again." on a budget that never failed, and the edit form's submit says "Save", so the retry it names does not exist on that surface.

The delete confirm at `:967-971` already does the right thing. Either mirror it (`updateBudget.reset()`, `setAssignmentError(null)`, `setPendingAssignments(null)` in the edit close path) or wrap the edit form the way `CreateBudgetDialog` wraps the create one, which is the shape `feedback.md:193-196` now prescribes.

web/src/features/budgets/BudgetsPage.tsx
```diff
@@ -1002,6 +1017,93 @@ function DeploymentBudgetsPage() {
...
+  const createBudget = useCreateBudget()
+
+  // Create the budget, then (optionally) attach it to the chosen users. The
+  // per-user PATCH sets each user's reset clock. Failed assignments stay in the
+  // form so a retry never creates a duplicate budget.
+  const createAndAssign = async (
+    body: CreateBudgetRequest,
+    userIds: string[],
+  ) => {
+    if (pendingAssignments) {
```
khaledosman, line 1109: **The retry throws away everything the operator changed, then closes as though it worked.** When `pendingAssignments` is set, `createAndAssign` ignores both its arguments and replays the original failed set (`pendingAssignments.userIds`), while the form's fields stay editable and the submit reads "Retry assignments".

Failure: the create's assignment half fails; the operator adds a second person to the picker, or corrects the limit, and presses Retry. The PATCH goes out for the originally-failed ids only, `assignUsers` returns true, `onClose()` fires. The added person is never assigned and the limit edit is gone, with no error and no trace.

Either retry against the form's current selection and what already landed, `assignUsers(pendingAssignments.budgetId, userIds, alreadyAssignedIds)`, or render the fields read-only while `pendingAssignments` is set so there is nothing to lose. The first is better: the retry is the operator's chance to fix the selection that failed.

web/src/design-system/forms/MultiSelect.tsx
```diff
@@ -0,0 +1,400 @@
...
+    )
+  }
+
+  const open = () => {
+    setIsOpen(true)
+    setActiveIndex(0)
+  }
+
+  const onKeyDown = (event: React.KeyboardEvent<HTMLInputElement>) => {
+    if (event.key === "ArrowDown" || event.key === "ArrowUp") {
```
khaledosman, line 165: **Nothing scrolls the active option into view, and the cap made that reachable.** `maxVisible` is 50 while the box stays six rows tall (`max-h-66`, `:281`), and the component has no effect and no ref read, so `aria-activedescendant` walks off the bottom with the list unmoved.

Failure: open the picker on a 40-person roster and press ArrowDown seven times. The highlight is below the fold, the operator sees nothing change, and Enter toggles someone they cannot see. ArrowUp from the top wraps to row 50, which is worse.

An effect on `activeIndex` while `isOpen`: `document.getElementById(optionId(activeIndex))?.scrollIntoView({ block: "nearest" })`.

web/src/design-system/forms/MultiSelect.tsx
```diff
@@ -0,0 +1,400 @@
...
+    }
+    if (event.key === "Enter" || event.key === " ") {
+      // Space only toggles when the query is empty; otherwise it is a character
+      // somebody is typing into a name.
+      if (event.key === " " && query !== "") return
+      if (!isOpen) return
+      // Before the row lookup, not after: an open combo box owns Enter whether
+      // or not it has a row to give, and this input sits inside a real form, so
+      // falling through submits it.
+      event.preventDefault()
```
khaledosman, line 189: **Cmd/Ctrl+Enter toggles a row and submits the pre-toggle selection in the same gesture.** The `preventDefault()` here stops implicit submission, which was the fix, but the event still bubbles to `FormDialog.tsx:220-234`, which calls `requestSubmit()` synchronously on a modified Enter. `BudgetForm.submit` then reads `userIds` from the render before the toggle, so the person whose check the operator just watched appear is not in the POST body.

`event.stopPropagation()` beside the `preventDefault()`. The Escape branch at `:178` already does exactly this, for the same reason.

web/src/design-system/forms/MultiSelect.tsx
```diff
@@ -0,0 +1,400 @@
...
+                        ) : null}
+                      </span>
+                    </div>
+                  )
+                })
+              )}
+            </div>
+            <div className="border-border text-mono-caption text-subtle flex h-8 items-center justify-between border-t px-2.5">
+              <span>
+                {selectedMatches} of {matches.length} matches selected
```
khaledosman: **The capped footer reads as a garbled sentence.** With 8 options and `maxVisible` 3 it renders `0 of 3 matches selected of 8 · 0 people assigned`, and `MultiSelect.test.tsx:171` locks that string in.

Split the two facts rather than nesting them: `Showing {matches.length} of {reached.length} · {selectedMatches} selected here · {counted(value.length)}`, and update the assertion. The count is accurate; it is the sentence that is not.

web/src/design-system/forms/MultiSelect.tsx
```diff
@@ -0,0 +1,400 @@
...
+              setActiveIndex(0)
+              setIsOpen(true)
+            }}
+            onFocus={open}
+            onBlur={(event) => {
+              // Only when focus actually left the control: a press on an option
+              // blurs the input and must not close the list before the press
+              // lands.
+              if (
+                !event.currentTarget.parentElement?.parentElement?.contains(
```
khaledosman: **A dead ref beside a hand-walked DOM path.** `inputRef` is assigned at `:207` and read nowhere since the refocus came out of `toggle`, while this `onBlur` decides whether focus left the control by walking `currentTarget.parentElement?.parentElement`. Wrap the input in one more div and option presses break silently (the popover closes before `onMouseDown` lands), with no type error and no lint error.

Retarget the ref to the `:204` wrapper and test `wrapperRef.current?.contains(event.relatedTarget)`. That spends the ref that is already there and removes the positional coupling.

web/src/design-system/forms/MultiSelect.tsx
```diff
@@ -0,0 +1,400 @@
...
+                chip row below IS a real list, because that one is a list. */}
+            <div
+              id={listId}
+              role="listbox"
+              aria-multiselectable
+              // Named apart from the field and from the chip row: three things
+              // carrying one name is three things a screen reader cannot tell
+              // apart, and it is what a query for the field then finds.
+              aria-label={`${label}, options`}
+              // Six rows before it scrolls, each at the 44px touch floor
```
khaledosman: The comment computes 6 x 44px = 264px, but the rows carry `not-first:border-t` (`:323`), so six of them occupy 269px and the sixth is clipped by 5px inside `max-h-66`. Either put the borders in the arithmetic or pick a value that includes them.

web/src/features/budgets/BudgetsPage.tsx
```diff
@@ -1002,6 +1017,93 @@ function DeploymentBudgetsPage() {
  * deployment-wide reads as an admin and paint their refusals; the shell already
  * waits on this query, so the cost is a spinner that is usually already over.
  */
+/**
```
khaledosman, line 1067: `CreateBudgetDialog` and its JSDoc were inserted between `BudgetsPage`'s docstring (`:1005-1019`) and `BudgetsPage` itself, which is now at `:1107` with nothing above it. Two consecutive JSDoc blocks means tooling attaches this one and drops the other. Move `:1005-1019` down to sit above `export function BudgetsPage()`.

web/src/design-system/forms/MultiSelect.tsx
```diff
@@ -0,0 +1,400 @@
...
+    <div className="flex flex-col gap-1.5">
+      {/* A real label with `htmlFor`, not HeroUI's `Label`: that one resolves
+          its target through a field context this control is not inside, so it
+          would render the words and name nothing. The visible text IS the
+          accessible name here, which is why there is no `aria-label` beside
+          it. */}
+      <label htmlFor={inputId} className="text-body">
+        {label}
+      </label>
+      {description ? (
```
khaledosman: **The description and the error render as two rows, where forms.md says one replaces the other.** `forms.md:118`: "An error message **replaces** the description line rather than adding a row." Here the description is a `<p>` above the control and the error is in `FieldMessages` below it, so going invalid adds a row and moves the footer, which is the thing `reserveMessage` exists to prevent.

The chips and the popover sitting between the two rungs is a real argument for the exception, but it is not written anywhere. Either follow the rule or say why this control does not, in the new `forms.md` section at `:156-179`.

While in that section: the signature writes `options: MultiSelectOption[]` where the prop is `readonly MultiSelectOption[]`. It gets `value: readonly string[]` right two lines up.

web/e2e/dashboard.spec.ts
```diff
@@ -150,15 +153,17 @@ test.describe("dashboard core flows", () => {
     await nav(page).getByRole("link", { name: "Spend & budgets" }).click()
     const budgetRow = page.getByRole("row", { name: /e2e-budget/ })
     await budgetRow.getByRole("button", { name: "Edit" }).click()
-    // "Add a person" is the input's own accessible name; "Assign to people" is
-    // the section heading beside it and labels nothing. Same shape as
-    // `addFilterValue` otherwise: the popover aria-hides the rest of the page, so
-    // it has to be put away before the submit button is reachable.
-    const owners = page.getByRole("combobox", { name: "Add a person" })
+    // The field's visible label is its accessible name now: the picker used to
+    // carry a hidden "Add a person" beside a heading that labelled nothing, so
```
khaledosman: "labelled" on a line this PR adds. US English per AGENTS.md, and `:59` in the same file already has "labeled".

khaledosman (changes requested):
All eleven fixed, and several past what I asked: the row cap applied after the filter so the query still searches everything, `min-h-11` spelled the way motion-and-access.md prescribes, the live region mirrored `sr-only` and unconditional, `forms.md` and `DESIGN.md` both carrying the primitive now, and `useCreateBudget` moved below the key rather than reset by hand.

Two new ones I would want before merge, both on the budget page rather than the primitive. The first is the fixed blocker's twin on the edit path; the second loses operator input and then reports success.

Gates clean, CI green.

---

web/src/features/budgets/BudgetsPage.tsx
jigjigjig: Fixed at 8c7ccff8, by the second route you named rather than the first.

`EditBudgetDialog` now wraps the edit form the way `CreateBudgetDialog` wraps the create one, keyed on the row, with `useUpdateBudget` inside it. Mirroring the delete confirm's resets would have worked for the 409 but left the same shape one edit away from the next person to touch this page; the wrapper makes the reset structural.

The page's ErrorBanner no longer lists `updateBudget.error` either, so a refused save renders once rather than behind the dialog as well.

Your assignment half is the sharper case and it is fixed by the same move: `assignmentError` and `pendingAssignments` are cleared on the edit close path, so an edit surface whose submit says Save cannot show a message naming a Retry it does not offer.

New case: take a 409 on budget A, discard, press Edit on B, assert the banner is gone. It fails with the mutation hoisted back to the page.

web/src/features/budgets/BudgetsPage.tsx
jigjigjig: Fixed at 8c7ccff8, the first way, for the reason you give: the retry is the operator's chance to fix the selection, and fields that stay editable have to mean something.

It now reconciles against the form's current selection, passing what the roster says already holds the budget as `previousUserIds`, so an addition made before pressing Retry lands and a removal is honored too.

Worth flagging how this interacted with a parallel fix: 3b909d3d on this branch had just made the retry `await` and close on success. I force-pushed over it while working from an older base, restored it, and merged the two by hand: theirs is the await and the close, mine is what the retry sends. The current head has both, verified by merge-base rather than by reading the log.

New case: the first PATCH refuses, the operator adds a second person, presses Retry, and the test asserts the second person was PATCHed and exactly one budget was posted. It fails against the replay.

web/src/design-system/forms/MultiSelect.tsx
jigjigjig: Fixed at 8c7ccff8. An effect scrolls the active row into view with `block: "nearest"`, keyed on the open state and the active index.

You are right that the cap is what made it reachable rather than theoretical: six visible rows against fifty rendered is thirty-odd rows of blind travel, and the wrap to row 50 on ArrowUp is the worst of it.

One implementation note, since it is the kind of thing that reads as a workaround: the call is guarded on `scrollIntoView` existing, because jsdom implements no layout and the arrow-key cases below would throw on the first press.

web/src/design-system/forms/MultiSelect.tsx
jigjigjig: Fixed at 8c7ccff8: `event.stopPropagation()` beside the `preventDefault()`, exactly as the Escape branch does.

The failure you describe is the worse half of it: not that the shortcut is ignored, but that it posts a selection one toggle out of date, so the operator watches a check appear and the request goes without it.

web/src/design-system/forms/MultiSelect.tsx
jigjigjig: Fixed at 8c7ccff8, split rather than reworded.

It reads `Showing 3 of 8 · 0 of 3 selected here · 0 people assigned` when the cap bites, and drops the first clause when it does not. Both assertions are updated, including the one at :230 that shared the old phrasing.

Your last line is the right diagnosis: the numbers were all correct, which is why nothing caught it. It was one sentence built out of two facts by nesting.

web/src/design-system/forms/MultiSelect.tsx
jigjigjig: Fixed at 8c7ccff8. The ref is on the wrapper at :204 and the blur test is `wrapperRef.current?.contains(event.relatedTarget)`.

The dead ref and the parent walk were the same mistake seen from two sides: the ref stopped being read when the refocus came out of `toggle`, and rather than noticing that, I left it and reached for the DOM. Spending it is the fix and it removes the positional coupling in the same edit.

web/src/design-system/forms/MultiSelect.tsx
jigjigjig: Fixed at 8c7ccff8. `max-h-[16.8125rem]`, which is 6 x 44 + 5, and the comment now does the arithmetic with the separators in it.

I had written the comment as though the rows were the only thing with height, so the clipping was invisible from the code that caused it.

web/src/features/budgets/BudgetsPage.tsx
jigjigjig: Fixed at 8c7ccff8: the page docstring sits directly above `export function BudgetsPage()` again, and `CreateBudgetDialog` and `EditBudgetDialog` carry their own above themselves.

I inserted a component between a docstring and its function without noticing the docstring was there, which is exactly the failure two adjacent JSDoc blocks produce: silent, and only visible to whatever attaches them.

web/src/design-system/forms/MultiSelect.tsx
jigjigjig: Both fixed at 8c7ccff8, the first by writing the exception down rather than following the rule.

`forms.md`'s MultiSelect section now says why this control keeps the two on separate rungs: what lies between them is the field, its popover and its chip row, so an error rendered up at the description would be separated from the control it is about by everything the operator is looking at, and one that replaced the description would move the field on the way in and out. `reserveMessage` holds the lower line open, so going invalid still moves nothing, which is the property the rule exists to protect.

If you would rather it followed the rule instead, say so and I will move it; the argument is written down now either way, which is the part that was missing.

The signature says `readonly MultiSelectOption[]`.

web/e2e/dashboard.spec.ts
jigjigjig: Fixed at 8c7ccff8: "labeled".

Note one nearby that I did not touch: `OrganizationMembersPage.tsx:522` has the same word in prose, but that file belongs to #1068 and is not in this diff. The `aria-labelledby` occurrences stay as they are, being the attribute's own spelling.

web/src/design-system/forms/MultiSelect.tsx
jigjigjig: Changed rather than argued: at 9efceb0a the description and the error share one rung below the control, the way `Field` does it, and the error replaces that line.

Your framing is what settled it. My argument was about this control in isolation, where the chips and the popover really do sit between the two rungs; but the control lives in a dialog full of `Field`s, and one of them speaking from a different place is the inconsistency the rule exists to stop. `reserveMessage` was already holding the line open, so the property I was protecting, that going invalid moves nothing, survives the move.

The forms.md exception is gone. `aria-describedby` names whichever line is rendered rather than both. The case asserts the replacement now: one id before and after, the error's text in it, and the description gone from the DOM.

The signature says `readonly MultiSelectOption[]` as of the earlier commit.

PR #1041, opened 2026-09-10
feat(dashboard): move both workspace creates and the organization one into FormDialog

jigjigjig:
## Description

The workspace form had **two frames**: a bleeding band between the page's header and its table, and a `Modal` hand-rolled in the scope switcher around the same component. Both are the dialog now, and `WorkspaceSwitcher` no longer imports `Modal` at all.

"Create and open" stays as the switcher's submit. It is a different verb because it navigates, and the label is the only warning an operator gets that the page is about to change under them.

**Create organization comes with it.** It is the switcher's other menu entry and was the twin of the workspace `Modal` sitting right beside it, so leaving it hand-rolled would have migrated one of two identical things. It was also built on HeroUI's `Card`, which `design/DESIGN.md` lists as "must not be used in new code", so that list drops from six call sites to five.

**Deleting the hand-rolled submit is this change's own argument.** That button had independently reached the same two conclusions `FormDialog` did: `opacity-0` rather than `invisible`, so the label survives in the accessibility tree and the button is never nameless mid-submit; and a spinner at `color="current"`, because the default `accent` paints it in the fill's own color. Its comment also knew something the component did not, that HeroUI 3.2.4's `Spinner` is its own live region, which is why the spinner is `aria-hidden`.

### Three test contracts changed

Each is rewritten where it is asserted, with a sentence saying what it used to hold. None was deleted.

1. **`frames the create form as a band of the page`** (the page) and **`frames the create form without the page's bleeding band`** (the switcher). These existed because the same form had two frames and each entry point had to pin its own. There is one frame now and neither owns it, so what is worth asserting is that both reach the shared dialog.
2. **`does not enter a workspace the operator dismissed the form over`.** The form disabled Cancel while a create was in flight and left Escape as a second answer, so an operator could create a workspace and suppress the navigation into it. `FormDialog` gives one answer: nothing dismisses a dialog while its submit is in flight. A create that cannot be cancelled cannot be half-cancelled either. This is a **user-visible behavior change** and the test now asserts the new contract by name.
3. **`data-pending="true"` on the submit** became `aria-busy` on the form. `FormDialog` withholds `isPending` from that button on purpose: the prop paints it at the product's disabled 0.4, and a submit in flight is working rather than refused.

### One defect found

The form kept its own `ErrorBanner` while the dialog rendered one from the same `error`, so a single failure showed as two `role="alert"` nodes. Caught by a test rather than by eye. Worth knowing for the pages still to migrate: any form with its own banner will do the same, and the other three migrated pages were checked and have none.

The create form takes `returnFocusRef`, pointing at the heading's own trigger, for the case where a successful first create retires the empty state whose CTA opened the dialog and react-aria has nothing left to restore focus to. It carries no test of its own here, deliberately: every version I wrote either passed while the empty state was still mounted, where the prop has no work to do, or turned into a fight with the mock rather than with the behavior. #1040 covers the same prop through the same component with a case that fails when the prop is removed, so the behavior is checked once where the check is trustworthy rather than twice where one of them proves nothing.

## How to test it locally

    pnpm --dir web run build
    pnpm --dir web exec playwright test --project=seed --project=screenshots-desktop-small-light --project=screenshots-mobile-light --grep "workspace dialogs"

- `pnpm --dir web test` — 5320 pass across 153 files; 273 in the three directories this touches.
- `storybook:build` then `node .storybook/smoke.mjs` — 383 stories x 2 themes, all rendered clean.

### What would have to break for each changed test to fail

| Change | Fails |
| --- | --- |
| render the page's form in a band again | `opens the create form in the dialog rather than as a band of the page` |
| give the switcher a frame of its own | `opens the same dialog the workspaces page opens` |
| let Escape dismiss while the submit is in flight | `cannot be dismissed mid-create, so the entry it promised happens` |
| pass `isPending` to the submit | `enters the workspace it just created` |

The `hold` gate keeps its own tests untouched: it is an injected wall-clock beat so the acknowledged-press window is entered and left deterministically, and it survives the migration unchanged.

### On the unslop pass

Run over this diff before opening. It tightened three test comments that explained the change at more length than the fact needed, and confirmed the deletions: 8 lines of hand-rolled spinner markup and its rationale go, along with the duplicated banner.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Part of #1031

Stacked on #1040 → #1038 → #1037 → #1032. Review those first; this branch contains their commits.

## Checklist

      Three contracts rewritten with their reasons, one screenshot capture, and
      the existing 273 cases in the touched directories kept green. The
      dashboard's tests live under `web/src` and `web/e2e`.
      The design system's deprecated-`Card` table names six call sites and this
      leaves five. Left for the change that owns that table rather than edited
      from here, since its own note says the counts are prose and should be
      checked against the tree.
      No API change.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

The Escape-during-create change is the one thing here that alters what the product does rather than where it draws it, and it was raised and ruled on before it landed rather than discovered in review.

## Summary

Workspace and organization creation now use the shared `FormDialog`. This removes custom page sections, cards, and modal layouts.

The dialogs now manage submission state, errors, dismissal rules, and responsive layout. Users cannot dismiss a dialog while creation is in progress. The workspace switcher keeps its “Create and open” behavior.

Tests cover dialog opening, fresh forms, blocked dismissal, pending state, retries, and visual snapshots for create dialogs.

### DIFF
--- a/web/src/app/nav/WorkspaceSwitcher.test.tsx
+++ b/web/src/app/nav/WorkspaceSwitcher.test.tsx
@@ -204,6 +204,32 @@ describe("the organization half of the scope switcher", () => {
     expect(screen.getByRole("dialog")).toBeInTheDocument()
   })
 
+  it("offers a fresh draft on each open of the organization form", async () => {
+    // Reset on the way in, not on the way out: the dialog keeps its content
+    // while it animates out, so clearing on close blanks the body in front of
+    // the operator. The switcher keys the form on an open counter instead.
+    mockApi()
+    await renderSwitcher()
+
+    const { user, menu } = await openMenu()
+    await user.click(
+      within(menu).getByRole("button", { name: /Create organization/ }),
+    )
+    await user.type(await screen.findByLabelText(/Name/), "half-typed")
+
+    // Out through the guard, which is the only way out of a dirty form.
+    await user.keyboard("{Escape}")
+    await user.click(screen.getByRole("button", { name: "Discard" }))
+
+    const reopened = await openMenu()
+    await reopened.user.click(
+      within(reopened.menu).getByRole("button", {
+        name: /Create organization/,
+      }),
+    )
+    expect(await screen.findByLabelText(/Name/)).toHaveValue("")
+  })
+
   it("creates an organization and moves into it", async () => {
     const requests = mockApi()
     await renderSwitcher()
--- a/web/src/app/nav/WorkspaceSwitcher.test.tsx
+++ b/web/src/app/nav/WorkspaceSwitcher.test.tsx
@@ -295,7 +295,9 @@ async function fillCreateForm(user: ReturnType<typeof userEvent.setup>) {
       name: "Create workspace",
     }),
   )
-  const form = await screen.findByRole("dialog", { name: "Create workspace" })
+  // The dialog's title names the object and its submit names the action, so
+  // this is "New workspace" and the button below is "Create workspace".
+  const form = await screen.findByRole("dialog", { name: "New workspace" })
   await user.type(within(form).getByLabelText(/^Name/), "Staging")
   return form
 }
@@ -348,13 +350,18 @@ describe("the workspace half of the scope switcher", () => {
   // `<main>`, so a band that travelled here measured a viewport wide and the
   // dialog's `overflow-clip` cropped it to an empty modal (otari-ai#2107). jsdom
   // computes no layout, so what is pinned is the class that causes it.
-  it("frames the create form without the page's bleeding band", async () => {
+  it("opens the same dialog the workspaces page opens", async () => {
+    // The same form once had two frames, a bleeding band on the page and a
+    // Modal here, and this asserted it was not the band. There is one frame and
+    // neither entry point owns it, so the fact worth holding is that this is
+    // the shared dialog.
     mockApi({ context: startedInAWorkspace })
     const user = userEvent.setup()
     await renderSwitcherOnAPage({})
 
     const form = await fillCreateForm(user)
 
+    expect(form).toHaveClass("otari-form-dialog")
     expect(form.querySelector(".otari-bleed")).toBeNull()
     expect(within(form).getByLabelText(/^Name/)).toBeInTheDocument()
   })
@@ -374,7 +381,13 @@ describe("the workspace half of the scope switcher", () => {
     // The press is acknowledged before the page moves, rather than the create
     // landing them somewhere else with nothing in between. The button keeps its
     // name through the beat, so it is still the control it was.
-    expect(submit).toHaveAttribute("data-pending", "true")
+    //
+    // On the form, not the button: `FormDialog` withholds `isPending` from the
+    // submit because the prop paints it at the disabled 0.4, and a submit in
+    // flight is working rather than refused. React-aria filters `aria-busy` off
+    // a Button anyway.
+    expect(submit.closest("form")).toHaveAttribute("aria-busy", "true")
+    expect(submit).toHaveAccessibleName(/Create and open/)
     expect(screen.queryByText("OVERVIEW PAGE")).toBeNull()
 
     beat.release()
@@ -389,7 +402,7 @@ describe("the workspace half of the scope switcher", () => {
     ).toBeInTheDocument()
   })
 
-  it("does not enter a workspace the operator dismissed the form over", async () => {
+  it("cannot be dismissed mid-create, so the entry it promised happens", async () => {
     // The beat is a gate this test opens, so the window the guard exists for is
     // entered and left on purpose rather than by sleeping long enough to have
     // been inside it. Nothing here waits on a duration.
@@ -402,9 +415,13 @@ describe("the workspace half of the scope switcher", () => {
     await user.click(
       within(form).getByRole("button", { name: /Create and open/ }),
     )
-    // Escape rather than Cancel: Cancel is disabled while the create is in
-    // flight, so dismissal is what is left, and it is the path that bypasses
-    // every button.
+    // This asserted the opposite until the form moved into `FormDialog`, and the
+    // change is deliberate rather than incidental. Escape used to dismiss the
+    // form mid-flight and suppress the navigation, which was the one path that
+    // bypassed a Cancel the form had already disabled for the same window.
+    // `FormDialog` closes that path: while a submit is in flight neither
+    // Escape, the backdrop nor the close control dismisses it, so there is one
+    // answer to "can I abandon this" rather than two that disagree.
     await user.keyboard("{Escape}")
     beat.release()
 
@@ -418,16 +435,10 @@ describe("the workspace half of the scope switcher", () => {
     const staging = await within(menu).findByRole("button", { name: /Staging/ })
     expect(staging).toBeVisible()
 
-    // The workspace was created and the switcher offers it; what must not
-    // happen is being taken there after saying not to.
-    expect(screen.queryByText("OVERVIEW PAGE")).toBeNull()
-    expect(screen.getByText("USAGE PAGE")).toBeInTheDocument()
-    // Read from the menu rather than the trigger, which the open menu hides
-    // from the accessibility tree: the scope never moved.
-    expect(
-      within(menu).getByRole("button", { name: /Default Workspace/ }),
-    ).toHaveTextContent("Selected")
-    expect(staging).not.toHaveTextContent("Selected")
+    // The workspace was created and the operator is in it, because the create
+    // they started was never abandoned.
+    expect(staging).toHaveTextContent("Selected")
+    expect(screen.getByText("OVERVIEW PAGE")).toBeInTheDocument()
   })
 
   it("still completes after StrictMode's development remount", async () => {
--- a/web/src/app/nav/WorkspaceSwitcher.test.tsx
+++ b/web/src/app/nav/WorkspaceSwitcher.test.tsx
@@ -257,6 +257,43 @@ describe("the organization half of the scope switcher", () => {
     })
   })
 
+  it("retries only the switch when the switch after a create was refused", async () => {
+    // The create succeeded, so pressing the button again has to send the second
+    // call and not the first: a retry that repeats the pair leaves a second
+    // organization behind, and nothing in this menu removes one.
+    const requests = mockApi({ switchFails: true })
+    await renderSwitcher()
+
+    const { user, menu } = await openMenu()
+    await user.click(
+      within(menu).getByRole("button", { name: /Create organization/ }),
+    )
+    await user.type(await screen.findByLabelText(/Name/), "Research")
+    const form = await screen.findByRole("dialog")
+    await user.click(
+      within(form).getByRole("button", { name: "Create organization" }),
+    )
+    expect(await within(form).findByRole("alert")).toHaveTextContent(
+      "Organization not found",
+    )
+
+    // The label is the retry: it names the one call that is left.
+    const retry = within(form).getByRole("button", {
+      name: "Switch to organization",
+    })
+    await user.click(retry)
+
+    const posts = requests.filter((request) => request.method === "POST")
+    expect(posts.map((request) => request.url)).toEqual([
+      "/organizations",
+      "/organizations/me/switch",
+      "/organizations/me/switch",
+    ])
+    expect(posts[2]?.body).toEqual({
+      organization_id: SECOND_ORGANIZATION_ID,
+    })
+  })
+
   it("offers Create organization whatever the caller's role in the one they are in", async () => {
     // No role in an organization gates creating one: it is not an action inside
     // a tenant, which is also why the server checks only the credential. Create
@@ -390,6 +427,34 @@ describe("the workspace half of the scope switcher", () => {
     expect(form).toHaveClass("otari-form-dialog")
     expect(form.querySelector(".otari-bleed")).toBeNull()
     expect(within(form).getByLabelText(/^Name/)).toBeInTheDocument()
+  })
+
+  it("offers a fresh draft on each open of the workspace form", async () => {
+    // Nothing here unmounts the form, so the remount on the way in is the only
+    // thing that clears it. Reset on the way out would blank the body while the
+    // dialog is still animating away.
+    mockApi({ context: startedInAWorkspace })
+    const user = userEvent.setup()
+    await renderSwitcherOnAPage({})
+
+    await fillCreateForm(user)
+
+    // Out through the guard, which is the only way out of a dirty form.
+    await user.keyboard("{Escape}")
+    await user.click(screen.getByRole("button", { name: "Discard" }))
+
+    await user.click(
+      await screen.findByRole("button", { name: /^Switch workspace/ }),
+    )
+    await user.click(
+      within(await screen.findByRole("dialog")).getByRole("button", {
+        name: "Create workspace",
+      }),
+    )
+    const reopened = await screen.findByRole("dialog", {
+      name: "New workspace",
+    })
+    expect(within(reopened).getByLabelText(/^Name/)).toHaveValue("")
   })
 
   it("enters the workspace it just created", async () => {
--- a/web/src/app/nav/WorkspaceSwitcher.test.tsx
+++ b/web/src/app/nav/WorkspaceSwitcher.test.tsx
@@ -53,6 +53,9 @@ function mockApi(
     memberships?: CallerOrganizationMembership[]
     context?: Parameters<typeof organizationContext>[0]
     switchFails?: boolean
+    // Holds the switch in flight, so the created step's pending line can be
+    // read before the refusal lands.
+    switchGate?: Promise<unknown>
     pendingInvitations?: PendingOrganizationInvitation[]
     pendingInvitationsFail?: boolean
   } = {},
@@ -104,13 +107,23 @@ function mockApi(
       return { data: memberships, count: memberships.length } as never
     }
     if (url === "/organizations/me/switch") {
+      if (options.switchGate) await options.switchGate
       if (options.switchFails) {
         throw new apiClient.ApiError(404, "Organization not found")
       }
       return organizationContext() as never
     }
     if (url === "/organizations") {
-      return organization({ id: SECOND_ORGANIZATION_ID }) as never
+      // Echoes the posted name, as the server does: the created step names the
+      // organization back to the operator, so a fixture name would let that
+      // copy pass while showing the wrong one.
+      const body = init?.body
+        ? (JSON.parse(String(init.body)) as { name?: string })
+        : {}
+      return organization({
+        id: SECOND_ORGANIZATION_ID,
+        ...(body.name ? { name: body.name } : {}),
+      }) as never
     }
     return organizationContext(options.context) as never
   })
@@ -255,6 +268,49 @@ describe("the organization half of the scope switcher", () => {
     expect(posts[1]?.body).toEqual({
       organization_id: SECOND_ORGANIZATION_ID,
     })
+  })
+
+  it("says the switch is running before it says the switch failed", async () => {
+    // Two states, one step: the organization exists the moment the create
+    // lands, so the body has to say what is happening rather than announcing a
+    // failure that has not happened yet.
+    let release = () => {}
+    const gate = new Promise<void>((resolve) => {
+      release = resolve
+    })
+    mockApi({ switchFails: true, switchGate: gate })
+    await renderSwitcher()
+
+    const { user, menu } = await openMenu()
+    await user.click(
+      within(menu).getByRole("button", { name: /Create organization/ }),
+    )
+    await user.type(await screen.findByLabelText(/Name/), "Research")
+    const form = await screen.findByRole("dialog")
+    await user.click(
+      within(form).getByRole("button", { name: "Create organization" }),
+    )
+
+    // In flight: the create is done and the switch is not.
+    expect(
+      await within(form).findByText("Research was created. Switching into it…"),
+    ).toBeVisible()
+    expect(
+      await within(form).findByRole("heading", {
+        name: "Organization created",
+      }),
+    ).toBeVisible()
+
+    release()
+
+    expect(
+      await within(form).findByText(
+        "Research was created. Switching into it failed.",
+      ),
+    ).toBeVisible()
+    expect(await within(form).findByRole("alert")).toHaveTextContent(
+      "Organization not found",
+    )
   })
 
   it("retries only the switch when the switch after a create was refused", async () => {
--- a/web/src/app/nav/WorkspaceSwitcher.test.tsx
+++ b/web/src/app/nav/WorkspaceSwitcher.test.tsx
@@ -491,6 +491,29 @@ describe("the workspace half of the scope switcher", () => {
         name: /^Switch workspace, currently Staging/,
       }),
     ).toBeInTheDocument()
+
+    // And the next open is usable. This is the path that leaves `holding` set:
+    // the create navigated, so nothing unmounted the form and nothing cleared
+    // the flag, and an unkeyed mount reopens spinning with its Cancel and Close
+    // disabled and no way out. The remount on the way in is what clears it.
+    await user.click(
+      await screen.findByRole("button", { name: /^Switch workspace/ }),
+    )
+    await user.click(
+      within(await screen.findByRole("dialog")).getByRole("button", {
+        name: "Create workspace",
+      }),
+    )
+    const reopened = await screen.findByRole("dialog", {
+      name: "New workspace",
+    })
+    expect(reopened.querySelector("form")).not.toHaveAttribute(
+      "aria-busy",
+      "true",
+    )
+    expect(
+      within(reopened).getByRole("button", { name: "Cancel" }),
+    ).toBeEnabled()
   })
 
   it("cannot be dismissed mid-create, so the entry it promised happens", async () => {
--- a/web/src/app/nav/WorkspaceSwitcher.tsx
+++ b/web/src/app/nav/WorkspaceSwitcher.tsx
@@ -88,6 +88,16 @@ export function WorkspaceSwitcher({
   const [open, setOpen] = useState(false)
   const [creating, setCreating] = useState(false)
   const [creatingOrganization, setCreatingOrganization] = useState(false)
+  // Bumped on each open, and the create form is keyed on it, so the draft is
+  // fresh every time and untouched through the exit: the dialog keeps its
+  // content while it animates out, so clearing on the way out would blank the
+  // body in front of the operator. See feedback.md, "A draft is fresh on every
+  // open and untouched through the exit".
+  const [creatingOrganizationCount, setCreatingOrganizationCount] = useState(0)
+  const openCreateOrganization = () => {
+    setCreatingOrganizationCount((n) => n + 1)
+    setCreatingOrganization(true)
+  }
   // Owners and admins only, which is what the server says of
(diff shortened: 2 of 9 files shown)


web/src/design-system/feedback/FormDialog.tsx
```diff
@@ -262,6 +309,7 @@ export function FormDialog({
                       <Button
                         type="submit"
                         variant="primary"
+                        isDisabled={isSubmitDisabled}
```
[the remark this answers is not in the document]
jigjigjig: Fixed, and confirmed present in this branch's tree rather than assumed from the rebase: I grepped the string at this head before resolving, because a rebase moves a line without carrying a fix.

Both submit paths read `if (!isPending && !isGuarding && !isSubmitDisabled)`. `requestSubmit` consults constraint validation and not the button's `disabled` state, so the shortcut reached the mutation while a click and a plain Enter were refused. A human reviewer raised the same thing independently on the keys PR, which is where it was fixed. Colocated cases cover both conditions, and removing either fails only its own case.

web/src/features/budgets/BudgetsPage.tsx
```diff
@@ -823,31 +838,34 @@ function DeploymentBudgetsPage() {
...
-          onSubmit={createAndAssign}
-          onClose={() => {
-            setAssignmentError(null)
-            setPendingAssignments(null)
-            setAddOpen(false)
-          }}
-        />
-      ) : null}
+      <BudgetForm
+        isOpen={addOpen}
```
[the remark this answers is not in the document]
jigjigjig: Fixed, and confirmed present in this branch's tree rather than assumed from the rebase: I grepped the string at this head before resolving, because a rebase moves a line without carrying a fix.

Not by resetting on close, though. The frame keeps its content while it animates out, so clearing there blanks the body in front of the operator. This page keys its create form on a counter it bumps where the trigger opens the dialog, so the draft is remounted fresh on the way **in**. `feedback.md` carries the rule, and the edit form beside this one already worked that way keyed on the row id. A case asserts the reopened form is empty, and removing the key fails exactly that case.

web/src/features/keys/KeysPage.test.tsx
```diff
@@ -586,10 +618,87 @@ describe("KeysPage", () => {
...
+    const user = userEvent.setup()
+    renderPage(<KeysPage />)
+
+    await screen.findByText("No API keys yet")
+    await user.click(
+      screen.getByRole("button", { name: "Create your first key" }),
+    )
+    const dialog = await screen.findByRole("dialog")
+    expect(
+      within(dialog).queryByLabelText("Create another"),
```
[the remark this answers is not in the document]
jigjigjig: Fixed, and confirmed present in this branch's tree rather than assumed from the rebase: I grepped the string at this head before resolving, because a rebase moves a line without carrying a fix.

It queries the button by role now. You were right that the old locator named a labelled control that no longer exists anywhere, so it would have passed whatever happened.

web/src/features/keys/KeysPage.tsx
```diff
@@ -469,10 +474,38 @@ function CreateKeyForm({
...
+
+  const resetForm = () => {
+    setKeyName("")
+    setExpiresAt("")
+    setShowAdvanced(false)
+    setAllowedModels(null)
+    setRejectUserMismatch(null)
+    setScopeValid(true)
+    create.reset()
+  }
```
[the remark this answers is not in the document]
jigjigjig: Fixed, and confirmed present in this branch's tree rather than assumed from the rebase: I grepped the string at this head before resolving, because a rebase moves a line without carrying a fix.

One predicate rather than two lists: `isPristine` sits beside `resetForm` and names every field the form owns, and `isDirty` is its negation, so the reset list and the dirty list cannot drift apart again. That drift is what produced both halves of your finding. `showAdvanced` is in neither, with the reason in the code: it reveals fields rather than holding a value. Two cases cover it, each failing only its own half.

web/src/features/organization/CreateOrganizationForm.tsx
```diff
@@ -13,47 +12,51 @@ import {
...
+        label="Name"
+        value={name}
+        onChange={setName}
+        placeholder="Research"
+        isRequired
+        autoFocus
+        description="You become its owner, and it starts with a default workspace. Names do not have to be unique."
+        reserveMessage
+      />
+    </FormDialog>
```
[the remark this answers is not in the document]
jigjigjig: Fixed, and confirmed present in this branch's tree rather than assumed from the rebase: I grepped the string at this head before resolving, because a rebase moves a line without carrying a fix.

Fixed where the form is mounted rather than inside it: `WorkspaceSwitcher` keys it on an open counter, so the draft is remounted fresh on each open and untouched through the exit. Resetting on close would blank the body while it animates out. A case in the switcher's own tests asserts the reopened form is empty, and removing the key fails exactly that case.

web/src/features/workspaces/WorkspacesPage.tsx
```diff
@@ -354,16 +358,73 @@ export function CreateWorkspaceForm({
...
+          const finish = async () => {
+            await held
+            // Dismissed while it was held: the workspace exists, and
+            // the list and the switcher will both show it, but the
+            // operator said not to go there.
+            if (!active.current) return
+            // No setHolding here: the form unmounts on close, and the
+            // failure paths below are what release the button.
+            onClose()
+            onCreated?.(workspace)
```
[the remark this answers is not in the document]
jigjigjig: Fixed, and confirmed present in this branch's tree rather than assumed from the rebase: I grepped the string at this head before resolving, because a rebase moves a line without carrying a fix.

Taking the second of your two options, and for the reason the first one is wrong here: unmounting on close would blank the frame while it animates out. The page keys the form on an open counter instead, so it remounts fresh on the way in. A case asserts the reopened form is empty, and removing the key fails exactly that case.

web/src/app/nav/WorkspaceSwitcher.tsx
```diff
@@ -348,77 +358,32 @@ export function WorkspaceSwitcher({
...
-        </Modal.Backdrop>
-      </Modal>
-      {/* Its own modal rather than one dialog switching on which row opened it:
-          the two forms share no field, and a single state would have to encode
-          "which" as well as "open". */}
-      <Modal
+      {/* Both entry points from this menu open the dialog every create in the
+          dashboard opens in, rather than a Modal wrapped by hand here: the two
+          forms were already the page's, and what was local was the frame. */}
+      <CreateWorkspaceForm
```
khaledosman, line 372: **This is the fourth create form and it did not get the key, which leaves a dialog that cannot be closed.**

`CreateOrganizationForm` three lines down has `key={creatingOrganizationCount}` (`:383`); this one has neither a counter nor a key, and it is now an always-mounted sibling where it used to be wrapped in the switcher's own `<Modal>` (which react-aria unmounted on close). `WorkspacesPage.tsx:382` still carries the assumption that went with that: "No setHolding here: the form unmounts on close."

It does not unmount any more, and this is the only caller that passes `onCreated`, so it is the only one where `setHolding(true)` runs (`WorkspacesPage.tsx:367`). On the success path `finish()` calls `onClose()` and `onCreated()` and never clears `holding` (`:376-386`). The switcher stays mounted across the `navigate({ to: "/" })`, so `holding` stays `true` for the rest of the session.

Reopen "Create workspace" from the scope menu and `FormDialog` mounts with `isPending={true}`: `requestClose` early-returns (`FormDialog.tsx:117`), the backdrop is non-dismissable (`:150`), the close control and Cancel are disabled (`:194`, `:285`), and the submit is `pointer-events-none` (`:314`). A permanently spinning dialog with no way out but a page reload, prefilled with the previous workspace's name and still carrying the previous `create.error`.

Fix: the same shape as the organization form. Bump a `creatingCount` in the `Create workspace` row's `onClick` (`:330-333`) and pass `key={creatingCount}` here. Then correct `WorkspacesPage.tsx:382-383` to say the remount is what releases the button.

web/src/features/budgets/BudgetsPage.tsx
```diff
@@ -818,36 +843,40 @@ function DeploymentBudgetsPage() {
...
-          assignUsers={users.data ?? []}
-          onSubmit={createAndAssign}
-          onClose={() => {
-            setAssignmentError(null)
-            setPendingAssignments(null)
-            setAddOpen(false)
-          }}
-        />
-      ) : null}
+      <BudgetForm
```
khaledosman: **The key remounts the form but not the mutation, so a failed create's banner survives into a fresh one.** `createBudget` is a page-level hook (`:512`), above the key, and `openCreate` (`:522-525`) does not reset it. A create fails, the operator discards and reopens: the draft is fresh, and `error={createBudget.error ?? assignmentError}` (`:858`) still renders the previous attempt's `ErrorBanner` over an empty form.

`WorkspacesPage` and `CreateOrganizationForm` avoid this by holding their mutations inside the keyed component. Here the cheap fix is `createBudget.reset()` in `openCreate`.

web/src/features/workspaces/WorkspacesPage.tsx
```diff
@@ -868,7 +854,7 @@ export function WorkspacesPage() {
           title="No workspaces yet"
           description="A workspace groups the work inside this organization and carries its own members and roles. Every organization is created with one, so an empty list usually means the default was deleted."
           actionLabel={manages ? "Create a workspace" : undefined}
-          onAction={manages ? () => setCreating(true) : undefined}
+          onAction={manages ? openCreate : undefined}
```
khaledosman, line 883: **`showOnboarding` still carries `&& !creating` (`:707`), which was right for an inline band and is wrong for a modal.** An operator with no workspaces presses this button; `creating` flips true, the `EmptyState` unmounts, the empty `DataTable` swaps in behind the backdrop, and the node react-aria would restore focus to is gone. Closing the dialog drops focus to `<body>`.

`5dfb010a` removed the identical `creating ||` gate from `PageIntro`'s action for this reason, and `BudgetsPage.test.tsx` pins it as "keeps the page's create action visible while the dialog is open". This is its twin, left behind. Drop `&& !creating`.

web/src/features/workspaces/WorkspacesPage.tsx
```diff
@@ -354,16 +358,73 @@ export function CreateWorkspaceForm({
...
+      isOpen={isOpen}
+      onOpenChange={(open) => {
+        if (!open) onClose()
+      }}
+      title="New workspace"
+      submitLabel={entersWorkspace ? "Create and open" : "Create workspace"}
+      onSubmit={submit}
+      isPending={pending}
+      isSubmitDisabled={trimmed === ""}
+      isDirty={trimmed !== "" || description.trim() !== ""}
```
khaledosman: `isDirty` ignores `budgetId`. Name nothing, pick a default member budget, press Escape: the selection is discarded with no guard, which is the one thing `isDirty` is for. `|| budgetId !== NO_DEFAULT`.

web/src/features/workspaces/WorkspacesPage.tsx
```diff
@@ -79,7 +81,7 @@ function DefaultBudgetPicker({
   onChange: (budgetId: string) => void
 }) {
   return (
-    <FilterSelect
+    <Select
```
khaledosman, line 85: **`DefaultBudgetPicker` reserves a message line it cannot use.** The `Select` has no `description`, no `isInvalid` / `errorMessage` and no `reserveMessage`, so `FieldMessages` defaults `reserve` to `true` (`FieldMessages.tsx:24`) and holds ~23px open under it, in the create dialog and in `EditWorkspaceForm`.

This is the defect `953850e7` fixed for the description `Field` two commits earlier, reintroduced by `698e4824`'s `FilterSelect` → `Select` swap: `FilterSelect` reserved nothing. `reserveMessage={false}`.

web/src/features/organization/CreateOrganizationForm.tsx
```diff
@@ -13,47 +12,51 @@ import {
...
+    <FormDialog
+      isOpen={isOpen}
+      onOpenChange={(open) => {
+        if (!open) onClose()
+      }}
+      // One field, which is what `sm` is for.
+      size="sm"
+      title="New organization"
+      submitLabel="Create organization"
+      onSubmit={() =>
```
khaledosman: **A failed switch after a successful create makes the obvious retry create a second organization.** The chained `create` → `switchTo` keeps no record that the create landed. When `switchTo` fails (the case this file's own header comment says it handles), the dialog stays open showing the switch error with the primary still reading "Create organization", so pressing it again POSTs another one, and the copy at `:57` says names do not have to be unique, so the server will not stop it.

Hold the created organization in state and, once it exists, retry only the switch: swap `onSubmit` and `submitLabel` over to it, the way feedback.md's "the success step is not a prop" describes.

khaledosman (changes requested):
Gates clean: no route or schema touched; no generated-artifact drift; no dependency change; conventional title; no em dashes or `--` separators. CI green.

One blocker: `fcf47828` keyed three of the four create forms this PR mounts, and the one it missed is the one that can strand an operator in a dialog with no way out. Four smaller findings below.

Worth knowing for the rest of the stack, since this PR establishes the pattern: the counter has to be bumped at every trigger, not just the toolbar one, and the key only resets state living below it, so a mutation hook held on the page still needs an explicit `reset()` (finding on `BudgetsPage.tsx:851`).

---

web/src/app/nav/WorkspaceSwitcher.tsx
jigjigjig: Fixed in d86ac503. `openCreateWorkspace` bumps a `creatingCount` and the form is keyed on it, the same shape as `CreateOrganizationForm` beside it.

Added a case in `WorkspaceSwitcher.test.tsx` that opens the form from the menu, types a name, leaves through the discard guard, and reopens: the name is empty. Removing the `key` fails it.

web/src/features/workspaces/WorkspacesPage.tsx
jigjigjig: Fixed in d86ac503. `showOnboarding` is now `!workspaces.isLoading && rows.length === 0`, with a comment naming the focus-restore reason rather than the band the gate was written for.

web/src/features/workspaces/WorkspacesPage.tsx
jigjigjig: Fixed in d86ac503, and as a snapshot rather than a longer predicate: `const draft = JSON.stringify({ name, description, budgetId })` compared against a `useRef` seeded on mount. A field added to the form is added to the snapshot or the guard cannot see it, which is what the hand-listed version could not promise.

web/src/features/workspaces/WorkspacesPage.tsx
jigjigjig: Fixed in d86ac503: `reserveMessage={false}` on the picker, with the reason stated on the prop (no description, nothing that can be invalid, so no message it could ever hold).

web/src/features/organization/CreateOrganizationForm.tsx
jigjigjig: Fixed in d86ac503. The created organization is held in state, and once it exists the submit is the switch alone and its label reads "Switch to organization"; the name field goes read-only, since editing it would change nothing the retry sends. `isDirty` and `isSubmitDisabled` both drop out at that point: the organization exists, so closing discards nothing.

The test drives it against a transport whose switch always 404s and asserts the POST sequence is `/organizations`, `/organizations/me/switch`, `/organizations/me/switch`. Restoring the chained submit fails it.

web/src/features/budgets/BudgetsPage.tsx
jigjigjig: Correct, and the fix is on #1040 at 995ffc06, which moves `useCreateBudget` into `BudgetForm` so the key resets the mutation with the draft. This branch drops its own open-counter hunk for the budgets page rather than carrying a second, weaker version of it, so the fix arrives here with the rebase. Leaving this open until it does.

web/src/features/routing/RoutingPage.test.tsx
```diff
@@ -327,6 +327,16 @@ afterEach(() => {
...
 
+/**
+ * The page's own create action, resolved while the dialog is shut.
+ *
+ * That is the only time it is unambiguous: the dialog's submit says "Create
+ * policy" too, and both are on screen together once it is open, which is why
+ * every reference to the submit below is scoped to the dialog.
+ */
+const createTrigger = () =>
+  screen.findByRole("button", { name: "Create policy" })
```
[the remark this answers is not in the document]
jigjigjig: Fixed, and now present in this tree. The helper was rewritten on #1038 and reached #1041 with the chain rebase onto main 057d715d; confirmed by reading `web/src/features/routing/RoutingPage.test.tsx` at this PR's head c2b5fd6f rather than by trusting the rebase:

```
const createTrigger = async () => {
  const heading = await screen.findByRole("heading", { name: "Routing" })
  const header = heading.closest("header")
  if (!header) throw new Error("PageIntro's header is gone")
  return within(header).findByRole("button", { name: "Create policy" })
}
```

Scoped to `PageIntro`'s own `<header>`, and the inner query is `findByRole` rather than `getByRole` because the header's action can arrive a tick after the heading.

Worth recording why the thirteen callers were green before the fix: they were, but only because none of them raced the list. Awaiting the empty state first turns one of them red with "Found multiple elements", which is the failure this describes.

web/src/features/budgets/BudgetsPage.tsx
jigjigjig: Fixed, and fixed the way you described rather than with `.reset()`. `useCreateBudget()` now lives inside the keyed form component (`:1059` at this PR's head c2b5fd6f), below the `key={addOpenCount}` at `:848`, so the remount clears the mutation with the draft. The two openers (`:798`, `:837`) both clear `assignmentError` and `pendingAssignments` before calling `openCreate()`, which covers the other half of `error={createBudget.error ?? assignmentError}`: that half is page state and a remount cannot reach it.

The fix landed on #1040 and arrived here with the chain rebase onto main 057d715d. Confirmed by reading this tree, not by assuming the rebase carried it: a rebase that moves a line is not a fix, and a thread on a stacked PR can describe a file state that only exists further up the stack.

web/src/app/nav/WorkspaceSwitcher.tsx
```diff
@@ -348,77 +361,38 @@ export function WorkspaceSwitcher({
...
+      {/* Both entry points from this menu open the dialog every create in the
+          dashboard opens in, rather than a Modal wrapped by hand here: the two
+          forms were already the page's, and what was local was the frame. */}
+      {/* Keyed on an open counter like its sibling below: nothing here unmounts
+          either form, so the remount on the way in is what clears the draft,
+          and with it the `holding` flag a create that navigates leaves set. An
+          unkeyed mount leaves the next open spinning, its Cancel and Close
+          disabled. */}
+      <CreateWorkspaceForm
+        key={creatingCount}
```
khaledosman: **The two sibling forms now carry the same React key.** `creatingCount` and `creatingOrganizationCount` both start at `0`, so this and `:392` are duplicate-keyed children of the same fragment on first render, and they collide again at 1/1, 2/2, and so on.

Measured, not inferred: **204** `Encountered two children with the same key, \`0\`` errors in this PR's dashboard job (`102957006771`), all from `src/app/*` test files, against 0 in main's run and 0 at `221e3e61`.

React builds its sibling key map last-write-wins, so opening the organization dialog while the counters match looks up key *N* and finds the organization fiber in the workspace slot; the type mismatch makes it delete and recreate `CreateWorkspaceForm`, remounting a form nothing asked to remount. React documents this as unsupported rather than as a defined behavior, so it is not something to rely on staying benign.

Namespace them: `key={`workspace-${creatingCount}`}` and `key={`organization-${creatingOrganizationCount}`}`.

web/src/features/organization/CreateOrganizationForm.tsx
```diff
@@ -13,47 +13,66 @@ import {
...
-    </Card>
+    <FormDialog
+      isOpen={isOpen}
+      onOpenChange={(open) => {
+        if (!open) onClose()
+      }}
+      // One field, which is what `sm` is for.
+      size="sm"
+      title="New organization"
+      submitLabel={created ? "Switch to organization" : "Create organization"}
```
khaledosman, line 44: **The second step changes the label and nothing else, so it never says the organization was created.** The title is still "New organization", the body is still the same field, and the description under it still reads "You become its owner, and it starts with a default workspace." The banner carries only the switch's error.

Failure: the switch is refused twice. The operator reads an error above a form that looks unsubmitted, closes it (`isDirty` is false at `:61`, so nothing asks), reopens from the menu, retypes "Research" and creates the duplicate this fix exists to prevent. Same outcome as before, one step further along.

`feedback.md:261-263` prescribes the shape and it is the half that is missing: "the caller swaps the **children** and the submit label", not the label alone. When `created !== null`, swap the body for a line naming what happened, "Research was created. Switching into it failed.", and drop the field description that is now false.

web/src/app/nav/WorkspaceSwitcher.test.tsx
```diff
@@ -348,17 +413,50 @@ describe("the workspace half of the scope switcher", () => {
...
     await renderSwitcherOnAPage({})
 
     const form = await fillCreateForm(user)
 
+    expect(form).toHaveClass("otari-form-dialog")
     expect(form.querySelector(".otari-bleed")).toBeNull()
     expect(within(form).getByLabelText(/^Name/)).toBeInTheDocument()
   })
 
+  it("offers a fresh draft on each open of the workspace form", async () => {
```
khaledosman, line 488: The test added for the blocker pins the draft, not the stuck spinner. It reaches the reopen through Escape then Discard, a path on which `setHolding` never runs, so it proves the key exists without ever exercising the symptom: create, navigate, reopen, and find `isPending` stuck true with Cancel and Close disabled.

Extend "enters the workspace it just created" (`:460-494`) instead: after `beat.release()` and the navigation, reopen the switcher and the form and assert the form has no `aria-busy="true"` and Cancel is enabled. That case fails against the unkeyed version; this one does not.

web/src/features/workspaces/WorkspacesPage.tsx
```diff
@@ -334,6 +343,13 @@ export function CreateWorkspaceForm({
     }
   }, [])
   const trimmed = name.trim()
+  // One snapshot of everything the form owns, seeded on mount: dirty means
+  // "differs from what was seeded", and a field added to the form is added here
+  // or the guard cannot see it. A predicate of empties had already forgotten
+  // `budgetId`, so picking a default member budget and pressing Escape
+  // discarded it with no guard.
+  const draft = JSON.stringify({ name, description, budgetId })
```
khaledosman: Two behavior changes in this commit landed untested: `isDirty` now covers `budgetId`, and the empty state stays mounted through the dialog (`:724`). The `budgetId` one is cheap and is the one a future edit is most likely to undo: as an operator, pick a default member budget with the name left empty, press Escape, assert the "Unsaved changes" guard appears rather than the dialog closing.

Related, in a file this PR already edits: `BudgetsPage.tsx:596` still reads `showOnboarding = !loading && rows.length === 0 && !addOpen`, which is exactly the defect finding 3 fixed here, and your new comment at `:721-723` now states the rule it breaks. `KeysPage.tsx:842` and `ProvidersPage.tsx:931` have it too; those are out of scope, but worth an issue.

khaledosman (changes requested):
Five of six fixed, and the blocker properly: `creatingCount` bumped in `openCreateWorkspace`, `key={creatingCount}` on the form, and the comment that said "the form unmounts on close" corrected to name the remount as what releases the button. I traced the `holding` lifecycle rather than taking the key on trust: `finish()` still leaves it set, both failure paths release it, and the remount is what clears it on the way in. The `draft` / `seededDraft` snapshot is seeded per open rather than per mount, because both call sites key the component, so it avoids the permanently-dirty shape `BudgetsPage` hit earlier in this stack.

Two things below. The duplicate key is new and this PR introduced it: **204** `Encountered two children with the same key` errors in this PR's own dashboard job (`102957006771`), against 0 on main and 0 at `221e3e61`. And finding 6 is half done: the retry can no longer POST a second organization, but the operator can still reach a duplicate by the other route.

**Merge ordering.** This branch sits on `221e3e61`, which is behind #1040's head `3b909d3d`. Your `BudgetsPage.tsx` is therefore the pre-fix copy, without "close the budget form when a retried assignment lands". You do not touch that region, so a rebase resolves cleanly, but merging this onto a main that already has #1040 without rebasing first would revert that fix and the test guarding it.

Gates clean, CI green, both new tests pass by name.

---

web/src/app/nav/WorkspaceSwitcher.tsx
jigjigjig: Fixed in ff73b4ef: `key={`workspace-${creatingCount}`}` and `key={`organization-${creatingOrganizationCount}`}`, with the reason on the first one.

Thank you for measuring it rather than describing it. I added the second counter without noticing that a bare number is only unique within one slot, and 204 errors in one job against 0 in main is not something I would have argued with.

web/src/features/organization/CreateOrganizationForm.tsx
jigjigjig: You are right, and it is the half of feedback.md I quoted at myself and then did not do. Fixed in ff73b4ef:

- title becomes "Organization created", because nothing is left to create and the frame was the last thing still saying otherwise (the keys page's secret step is the precedent)
- the body is one line, "{name} was created. Switching into it failed.", and the field is gone rather than disabled, along with the description that describes a create which has already happened
- the submit stays "Switch to organization", and the banner still carries the switch's error

One thing I want on the record rather than buried, because I have not fixed it: the swap is keyed on `created !== null`, which is true from the instant the create succeeds, so during the first switch attempt (a few hundred ms) the body says the switch failed while it is still in flight. The footer's spinner is the live signal in that window. Wording it accurately in both states needs a second string, so it is a copy decision rather than a code one and it is with the owner; if you would rather it read differently, say so and I will change it here.

web/src/app/nav/WorkspaceSwitcher.test.tsx
jigjigjig: Correct, and fixed the way you suggested rather than by adding a third case. "enters the workspace it just created" now carries the symptom: after `beat.release()` and the navigation, it reopens the switcher and the form and asserts the form has no `aria-busy="true"` and Cancel is enabled.

It is a check now: removing the `key` fails it, along with the draft case. Before, only the draft case failed, which is exactly what you said.

web/src/features/workspaces/WorkspacesPage.tsx
jigjigjig: Both covered in ff73b4ef, and both fail with their fix removed.

The `budgetId` one: pick a default member budget with the name empty, dismiss, and the guard has to appear; taking `budgetId` back out of the snapshot fails it. It dismisses through Cancel rather than Escape, because in jsdom focus lands on `<body>` after picking from a `Select`, so a keystroke reaches nothing. That is worth knowing on its own and I have reported it separately; it may be a jsdom artifact or a real focus drop in every `Select` in the product, and settling it needs a browser.

The empty state: it stays mounted under the dialog and focus returns to "Create a workspace" on close; restoring `&& !creating` fails it.

On the sibling finding, thank you, and it is being handled rather than noted: `BudgetsPage.tsx:596` goes to whoever holds that branch, and `KeysPage.tsx:842` gets its own issue. `ProvidersPage.tsx:931` I had already fixed on #1045 for the same reason.

PR #1045, opened 2026-09-10
feat(dashboard): move the add-provider form into FormDialog

jigjigjig:
## Description

Adding a provider appended a panel to the page with a tab row at its top. It is the dialog now, at `lg`, with the tab row under the header, which is the shape artboard 06 draws for a form that has two of them. This is the first dialog in the product with tabs.

**Each tab keeps its own mutation, validity and submit.** The two write the same five fields but derive them from different questions: one asks which provider and fills the rest from its built-in detail, the other asks for a name, an API flavor and a base URL. A shared submit would be a switch on the tab, which is the same code with one more place to look.

What is shared is the frame. Both halves render the same `FormDialog` with the same title, size and tab row, so switching tabs changes the fields and nothing else, and a half-filled tab still does not survive a switch away from it, which is what it did as a panel.

**"Test connection" is the footer's left slot.** It belongs beside the submit rather than in the body: it is about the thing being submitted, and it was already in that action row.

**The heading's action stays visible** while the dialog is open and beside the first-run panel's own copy of it, and stays **disabled rather than hidden** when the server has no secret key. That is the rule for a control carrying its own reason nearby, and it was already how this button behaved; what changed is that it no longer also hides itself while the panel is open.

## How to test it locally

    pnpm --dir web run build
    pnpm --dir web exec playwright test --project=seed --project=screenshots-desktop-small-light --grep "provider dialog"

Two captures, one per tab, so the tab row and the two field sets are both in the baseline.

- `pnpm --dir web test` — 5352 pass across 154 files; 65 in `ProvidersPage.test.tsx`.

### Tests changed rather than added

Three, each to a contract the migration changes, and each saying so where it is asserted:

| Was | Is |
| --- | --- |
| the heading's action is absent while the first-run panel shows | it is present, beside the panel's own copy |
| `getByRole("button", { name: "Add provider" })` unscoped | scoped to the dialog, because the trigger says the same words |
| `{Escape}` after picking a provider, to close the combo box | removed: the popover closes on selection, so the key reached the dialog and armed its unsaved-changes guard |

That third one is the same interaction found on `/keys` and `/routing`. It is now three pages, which is worth saying plainly: **a combo box inside one of these dialogs turns a habitual `{Escape}` in a test into a dialog dismissal.**

### On the unslop pass

Run over this diff before opening, and it caught a false claim I had written into the new docstring: that the two payloads "have three fields in common and six that are not". They have the same five fields; only the derivation differs. Corrected to what the code does. Nothing else survived worth removing.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Part of #1031

Stacked on #1041 → #1040 → #1038 → #1037. Review those first; this branch contains their commits.

## Checklist

      65 cases in `ProvidersPage.test.tsx`, three updated to the new contract,
      plus two screenshot captures. The dashboard's tests live under `web/src`
      and `web/e2e` rather than the repo-root `tests/`.
      No design-doc change: the placement, label and surface rules were written
      down in #1032 and this page follows them.
      No API change. Both tabs send exactly the bodies they sent before.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

One copy question is open and named in the report to the reviewer rather than settled here: the dialog's title is the bare noun "Provider", where the other five dialogs in the series are "New key", "New budget", "New workspace", "New organization" and "Edit policy". It reads flatter than its siblings and is one string to change.

## Summary

- Moved provider, workspace, and organization creation forms into reusable `FormDialog` components.
- Added separate provider tabs for known providers and custom endpoints.
- Preserved validation, dirty-state protection, draft reset, focus restoration, and submit behavior.
- Improved connection-test result display and provider picker focus behavior.
- Updated tests, screenshots, and end-to-end flows for the new dialogs.
- Added coverage for pending actions, retries, navigation states, and fresh drafts on each open.

## Technical notes

- Added dialog open counters to remount forms with fresh drafts.
- Added `scrollIntoView` test support.
- No API changes.
- Unit tests, lint, typecheck, build, and Playwright tests pass.

### DIFF
--- a/web/src/features/providers/ProvidersPage.tsx
+++ b/web/src/features/providers/ProvidersPage.tsx
@@ -57,28 +57,39 @@ import {
   parseClientArgs,
 } from "./providerFields"
 
-// A "Test connection" button + inline result, testing the form's credentials
-// before they are saved. `getPayload` returns null when the minimum fields for a
-// test are not filled in yet, which disables the button.
-function ConnectionTest({
+// Testing the form's credentials before they are saved, in two nodes because
+// they belong in two places: the button sits in the footer beside the submit,
+// its outcome at the end of the body. The unverified case below is four lines
+// plus the provider's own reply, and a footer that grows shoves the form up
+// under the operator's hands; in the body it scrolls with everything else.
+type ConnectionTestState = ReturnType<typeof useTestProviderCredentials>
+
+// `getPayload` returns null when the minimum fields for a test are not filled
+// in yet, which disables the button.
+function ConnectionTestButton({
+  test,
   getPayload,
 }: {
+  test: ConnectionTestState
   getPayload: () => CreateStoredProviderRequest | null
 }) {
-  const test = useTestProviderCredentials()
   const payload = getPayload()
+  return (
+    <Button
+      variant="ghost"
+      isDisabled={payload === null || test.isPending}
+      onPress={() => {
+        if (payload) test.mutate(payload)
+      }}
+    >
+      {test.isPending ? "Testing…" : "Test connection"}
+    </Button>
+  )
+}
 
+function ConnectionTestResult({ test }: { test: ConnectionTestState }) {
   return (
     <div className="flex flex-col gap-1.5">
-      <Button
-        variant="ghost"
-        isDisabled={payload === null || test.isPending}
-        onPress={() => {
-          if (payload) test.mutate(payload)
-        }}
-      >
-        {test.isPending ? "Testing…" : "Test connection"}
-      </Button>
       {/* aria-live so the connection outcome is announced to assistive tech. */}
       <span role="status" aria-live="polite">
         {test.isPending ? null : test.error ? (
@@ -130,6 +141,7 @@ function KnownProviderForm({
   tabs: ReactNode
 }) {
   const create = useCreateStoredProvider()
+  const test = useTestProviderCredentials()
   const [providerId, setProviderId] = useState("")
   const [apiKey, setApiKey] = useState("")
   const [showAdvanced, setShowAdvanced] = useState(false)
@@ -168,8 +180,7 @@ function KnownProviderForm({
     !nameHasDelimiter &&
     (!needsKey || apiKey.trim() !== "") &&
     clientArgs.ok &&
-    Object.keys(credentialErrors).length === 0 &&
-    !create.isPending
+    Object.keys(credentialErrors).length === 0
   // Hold the section open while something inside it is what's blocking submit,
   // so collapsing it can't leave a disabled button with its reason off screen.
   // A hide requested meanwhile is remembered and applies once the field is fixed.
@@ -195,7 +206,9 @@ function KnownProviderForm({
 
   const submit = () => {
     const payload = buildPayload()
-    if (!canSubmit || payload === null) return
+    // `create.isPending` is a reason not to send twice, not a reason to draw
+    // the primary as refused, so it guards the call rather than `canSubmit`.
+    if (!canSubmit || create.isPending || payload === null) return
     create.mutate(payload, { onSuccess: onClose })
   }
 
@@ -214,9 +227,13 @@ function KnownProviderForm({
       isSubmitDisabled={!canSubmit}
       isDirty={providerId !== "" || apiKey.trim() !== ""}
       error={create.error}
-      footerStart={<ConnectionTest getPayload={buildPayload} />}
+      footerStart={
+        <ConnectionTestButton test={test} getPayload={buildPayload} />
+      }
     >
       <ProviderComboBox
+        // The known tab is the dialog's default, so this is its first field.
+        autoFocus
         label="Provider"
         value={providerId}
         onChange={(id) => {
@@ -305,6 +322,7 @@ function KnownProviderForm({
           />
         </div>
       ) : null}
+      <ConnectionTestResult test={test} />
     </FormDialog>
   )
 }
@@ -321,6 +339,7 @@ function CustomProviderForm({
   tabs: ReactNode
 }) {
   const create = useCreateStoredProvider()
+  const test = useTestProviderCredentials()
   const [name, setName] = useState("")
   const [providerType, setProviderType] = useState("openai-compatible")
   const [apiBase, setApiBase] = useState("")
@@ -333,11 +352,10 @@ function CustomProviderForm({
     name.trim() !== "" &&
     !nameHasDelimiter &&
     apiBase.trim() !== "" &&
-    clientArgs.ok &&
-    !create.isPending
+    clientArgs.ok
 
   const submit = () => {
-    if (!canSubmit || !clientArgs.ok) return
+    if (!canSubmit || create.isPending || !clientArgs.ok) return
     create.mutate(
       {
         instance: name.trim(),
@@ -363,10 +381,13 @@ function CustomProviderForm({
       onSubmit={submit}
       isPending={create.isPending}
       isSubmitDisabled={!canSubmit}
-      isDirty={name.trim() !== "" || apiBase.trim() !== ""}
+      isDirty={
+        name.trim() !== "" || apiBase.trim() !== "" || apiKey.trim() !== ""
+      }
       error={create.error}
       footerStart={
-        <ConnectionTest
+        <ConnectionTestButton
+          test={test}
           getPayload={() =>
             name.trim() === "" || apiBase.trim() === "" || !clientArgs.ok
               ? null
@@ -430,6 +451,7 @@ function CustomProviderForm({
         onChange={setClientArgsText}
         error={clientArgs.ok ? null : clientArgs.error}
       />
+      <ConnectionTestResult test={test} />
     </FormDialog>
   )
 }
@@ -953,9 +975,21 @@ export function ProvidersPage() {
   const updateSettings = useUpdateSettings()
 
   const [addOpen, setAddOpen] = useState(false)
+  const [addOpenCount, setAddOpenCount] = useState(0)
   const [editing, setEditing] = useState<string | null>(null)
   const [pendingDelete, setPendingDelete] = useState<string>()
   const [tests, setTests] = useState<Record<string, TestState>>({})
+  // `addOpenCount` above is bumped on each open, and the add form is keyed on
+  // it, so the draft (a pasted provider key included) is fresh every time and
+  // untouched through the exit: the dialog keeps its content while it animates
+  // out, so clearing on the way out would blank the body in front of the
+  // operator. See feedback.md, "A draft is fresh on every open and untouched
+  // through the exit". Both openers on this page go through here.
+  const openAdd = () => {
+    setEditing(null)
+    setAddOpenCount((n) => n + 1)
+    setAddOpen(true)
+  }
 
   const rows = buildRows(meta.data?.providers, stored.data)
   const healthByInstance = new Map(
@@ -971,7 +1005,10 @@ export function ProvidersPage() {
   // membership context reports and `/settings` no longer answers for every
   // caller who reaches this page (#839).
   const secretKeyConfigured = useProviderKeyEncryption()
-  const showOnboarding = !loading && rows.length === 0 && !addOpen
+  // Not gated on `addOpen`: unmounting the first-run panel when the dialog
+  // opens takes away the node react-aria restores focus to, so closing drops
+  // focus to `<body>`. The heading's action is ungated for the same reason.
+  const showOnboarding = !loading && rows.length === 0
 
   // Which test run each row is currently showing. A row's result is only worth
   // recording while it is still the answer to the newest thing the operator asked
@@ -1153,11 +1190,6 @@ export function ProvidersPage() {
       <PageIntro
         title="Providers"
         action={
-          /* The first-run strip supplies its own focused call to action, and the
-             form has its own Close, so the header action is redundant while
-             either is open. Disabled rather than hidden when the server has no
-             secret key: an operator who cannot add a provider still needs to see
-             that adding one is the thing they are being denied. */
           <Button
             // Visible while the dialog is open and beside the first-run
             // panel's own copy of it: the dialog is over the page. Disabled
@@ -1165,10 +1197,7 @@ export function ProvidersPage() {
             // rule for a control that carries its own reason nearby.
             variant="primary"
             isDisabled={!secretKeyConfigured}
-            onPress={() => {
-              setEditing(null)
-              setAddOpen(true)
-            }}
+            onPress={openAdd}
           >
             Add provider
           </Button>
@@ -1231,10 +1260,7 @@ export function ProvidersPage() {
 
       {showOnboarding ? (
         <OnboardingPanel
-          onAddProvider={() => {
-            setEditing(null)
-            setAddOpen(true)
-          }}
+          onAddProvider={openAdd}
           needsPricing={needsPricing}
           onEnablePricing={() =>
             updateSettings.mutate({ default_pricing: true })
@@ -1252,6 +1278,7 @@ export function ProvidersPage() {
           out to be unavailable, retract it so its submit can never reach the create
           mutation. The banner above explains why. */}
       <AddProviderForm
+        key={addOpenCount}
         isOpen={addOpen && secretKeyConfigured}
         onClose={() => setAddOpen(false)}
       />
--- a/web/src/features/providers/ProvidersPage.tsx
+++ b/web/src/features/providers/ProvidersPage.tsx
@@ -88,44 +88,54 @@ function ConnectionTestButton({
 }
 
 function ConnectionTestResult({ test }: { test: ConnectionTestState }) {
+  const answered = test.data ?? test.error
+  const ref = useRef<HTMLSpanElement | null>(null)
+  // The body scrolls, and this is its last child: on an `lg` dialog in an 800px
+  // window the custom tab's own fields already fill it, so a verdict rendered
+  // here can land below the fold with the footer button back to "Test
+  // connection" and nothing else, to a sighted operator, having happened.
+  useEffect(() => {
+    if (answered) ref.current?.scrollIntoView({ block: "nearest" })
+  }, [answered])
   return (
-    <div className="flex flex-col gap-1.5">
-      {/* aria-live so the connection outcome is announced to assistive tech. */}
-      <span role="status" aria-live="polite">
-        {test.isPending ? null : test.error ? (
-          <span className="text-xs text-danger">
-            {errorMessage(test.error)}
+    // No wrapper, and `empty:hidden` rather than a conditional render: the
+    // parent's `gap-4` would otherwise reserve a row before a test is ever run,
+    // and the live region has to exist before it has something to say.
+    <span
+      ref={ref}
+      role="status"
+      aria-live="polite"
+      className="flex flex-col gap-1.5 empty:hidden"
+    >
+      {test.isPending ? null : test.error ? (
+        <span className="text-xs text-danger">{errorMessage(test.error)}</span>
+      ) : test.data ? (
+        test.data.ok ? (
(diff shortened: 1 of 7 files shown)


web/src/features/providers/ProvidersPage.tsx
```diff
@@ -1206,9 +1251,10 @@ export function ProvidersPage() {
           if it was opened while settings were still loading and the key then turns
           out to be unavailable, retract it so its submit can never reach the create
           mutation. The banner above explains why. */}
-      {addOpen && secretKeyConfigured ? (
-        <AddProviderForm onClose={() => setAddOpen(false)} />
-      ) : null}
+      <AddProviderForm
```
khaledosman, line 1312: **The create dialog keeps its draft, including the pasted API key, across a close.** `AddProviderForm` went from `{addOpen && secretKeyConfigured ? … : null}` to an unconditional mount with no counter and no `key`, so `KnownProviderForm` / `CustomProviderForm` never unmount and `providerId` / `apiKey` (`:133-134`) and `name` / `apiKey` (`:324-327`) survive. `EditProviderForm` nine lines down already keys on `editingProvider.instance` (`:1263`), so the file demonstrates the shape this one is missing.

Failure: pick OpenAI, paste `sk-live-…`, submit. `onSuccess: onClose` (`:199`) closes the dialog and clears nothing. Reopen "Add provider": the provider is still selected, the secret is rendered back into the `SecretField`'s DOM after save, and a still-enabled submit re-POSTs the same credential. Same on cancel. `create.error` survives too, so a previous failure's `ErrorBanner` renders over the reopened form.

Fix: `const [addOpenCount, setAddOpenCount] = useState(0)` beside `addOpen` (`:955`), an `openCreate()` that bumps it, and `key={addOpenCount}` here. **Both triggers must call it** — the `PageIntro` action (`:1170`) and the `OnboardingPanel`'s `onAddProvider` (`:1236`); this is the one page in the stack with two openers. No explicit `.reset()` needed: `useCreateStoredProvider()` is held inside the form components (`:132`, `:323`), below the key.

web/src/features/providers/ProvidersPage.tsx
```diff
@@ -191,8 +200,22 @@ function KnownProviderForm({ onClose }: { onClose: () => void }) {
...
+      onOpenChange={(open) => {
+        if (!open) onClose()
+      }}
+      size="lg"
+      title="New provider"
+      tabs={tabs}
+      submitLabel="Add provider"
+      onSubmit={submit}
+      isPending={create.isPending}
+      isSubmitDisabled={!canSubmit}
```
khaledosman, line 251: **`isSubmitDisabled` includes `isPending`, so the primary reads as refused while it is working.** `canSubmit` folds in `!create.isPending` (`:172`, and `:337` for the custom tab), and `FormDialog` renders `isDisabled={isSubmitDisabled}` — the product's one disabled treatment at 0.4 opacity — with the spinner on top of it.

feedback.md is explicit: "The primary is **not** disabled while it runs… a submit in flight is working rather than refused, so it keeps its fill and blocks its own press." `BudgetsPage.tsx:276-279` spells the same rule out in a comment.

Drop `!create.isPending` from both `canSubmit`s and keep it in `submit`'s own guard.

web/src/features/providers/ProvidersPage.tsx
```diff
@@ -191,8 +200,22 @@ function KnownProviderForm({ onClose }: { onClose: () => void }) {
...
+      isDirty={providerId !== "" || apiKey.trim() !== ""}
+      error={create.error}
+      footerStart={<ConnectionTest getPayload={buildPayload} />}
```
khaledosman: **`ConnectionTest` in `footerStart` puts a variable-height live region in a footer feedback.md says never changes height.** `footerStart` is documented for a "Create another" checkbox or a caption.

Failure: test a backend with no `/v1/models`; the `discovery_unsupported` branch (`:99-109`) renders a four-line `max-w-md` warning plus the provider's reply into the footer, which grows ~100px and shoves the body up mid-form. On the phone sheet, where the footer is already stacked, it takes most of the frame.

Keep the button in `footerStart` and render its outcome as the last child of the dialog body, where a growing element scrolls with everything else. Same at `:368`.

web/src/features/providers/ProvidersPage.tsx
```diff
@@ -329,8 +351,36 @@ function CustomProviderForm({ onClose }: { onClose: () => void }) {
...
+        if (!open) onClose()
+      }}
+      size="lg"
+      title="New provider"
+      tabs={tabs}
+      submitLabel="Add provider"
+      onSubmit={submit}
+      isPending={create.isPending}
+      isSubmitDisabled={!canSubmit}
+      isDirty={name.trim() !== "" || apiBase.trim() !== ""}
```
khaledosman: `isDirty` omits `apiKey`. Paste a key into the custom tab's API key field and nothing else, press Escape: the dialog closes with no guard and the secret is gone. `|| apiKey.trim() !== ""`.

web/src/features/providers/ProvidersPage.tsx
```diff
@@ -191,8 +200,22 @@ function KnownProviderForm({ onClose }: { onClose: () => void }) {
...
+      tabs={tabs}
+      submitLabel="Add provider"
+      onSubmit={submit}
+      isPending={create.isPending}
+      isSubmitDisabled={!canSubmit}
+      isDirty={providerId !== "" || apiKey.trim() !== ""}
+      error={create.error}
+      footerStart={<ConnectionTest getPayload={buildPayload} />}
+    >
       <ProviderComboBox
```
khaledosman, line 258: The known-provider tab is the default and its first field takes no `autoFocus`, where the custom tab's Name field does (`:391`). feedback.md: "The first field takes `autoFocus`." Thread it through `ProviderComboBox` (`providerFields.tsx:178`) into `ComboBoxField`, which already accepts one.

web/src/features/providers/ProvidersPage.tsx
```diff
@@ -1115,18 +1158,20 @@ export function ProvidersPage() {
...
-              onPress={() => {
-                setEditing(null)
-                setAddOpen(true)
-              }}
-            >
-              Add provider
-            </Button>
-          )
+          <Button
+            // Visible while the dialog is open and beside the first-run
```
khaledosman, line 1226: The comment block directly above (`:1156-1160`) states the behavior this PR deleted — "the header action is redundant while either is open" — and this new one contradicts it two lines later. Delete the old block.

khaledosman (changes requested):
Gates clean: no route or schema touched; no generated-artifact drift; no dependency change; conventional title; no em dashes or `--` separators. The `e2e` failure on the last run was the pre-rebase `dismissComboBox` regression and is gone at this head.

One blocker: the new create dialog does not adopt `9487bb1e`'s draft pattern for itself, and here the surviving draft holds a pasted provider API key. Five smaller findings below.

---

web/src/features/providers/ProvidersPage.tsx
jigjigjig: Fixed in 8cd4a6a2. `addOpenCount` beside `addOpen`, one `openAdd()` that bumps it, and `key={addOpenCount}` on the form. Both openers go through `openAdd`: the heading action and the panel's `onAddProvider`.

The test drives both, in both orders, and asserts a blank picker and a blank API key on each reopen. Removing the `key` fails it.

web/src/features/providers/ProvidersPage.tsx
jigjigjig: Fixed in 8cd4a6a2: `!create.isPending` is out of both `canSubmit`s and into each `submit`'s own guard, with a comment saying which question each answers.

The test holds a create in flight behind a gate, asserts the primary is still enabled, presses it again, and asserts exactly one POST went out.

web/src/features/providers/ProvidersPage.tsx
jigjigjig: Fixed in 8cd4a6a2. `ConnectionTest` is now two nodes: `ConnectionTestButton` in `footerStart` and `ConnectionTestResult` as the last child of the body, in both tabs, sharing one `useTestProviderCredentials()` lifted into each form.

The test runs a `discovery_unsupported` verdict and asserts the footer does not contain the outcome and does contain the button. Putting the outcome back in `footerStart` fails it, so it pins the placement rather than just the text.

web/src/features/providers/ProvidersPage.tsx
jigjigjig: Fixed in 8cd4a6a2: `|| apiKey.trim() !== ""` added to the custom tab's `isDirty`.

web/src/features/providers/ProvidersPage.tsx
jigjigjig: Done in 8cd4a6a2, with one change to the mechanism you named. `ProviderComboBox` does not wrap `ComboBoxField`; it builds `ComboBox.Root` itself, and it carries `menuTrigger="focus"` with a comment resting on the field not being autofocused.

Measured before changing it: rendered autofocused in jsdom, the input mounts `aria-expanded="true"` with a listbox in the tree, so the catalog would be down over the form before anything had been asked. So the autofocused instance opens on input instead (`menuTrigger={autoFocus ? "input" : "focus"}`), typing filters, and the chevron still shows the whole catalog. Every other instance keeps focus-to-open.

One existing case pressed the field to open the list and now types instead; that is the cost, stated in the comment. New test asserts the picker has focus on open and no listbox is rendered. Both halves are checked: dropping `autoFocus` fails the focus assertion, and restoring `menuTrigger="focus"` fails the closed-list one (and eight other cases, whose clicks the open popover swallows).

web/src/features/providers/ProvidersPage.tsx
jigjigjig: Deleted in 8cd4a6a2. The block below it, which describes what the control does now, is the one that stays.

web/src/features/providers/ProvidersPage.tsx
```diff
@@ -186,14 +206,34 @@ function KnownProviderForm({ onClose }: { onClose: () => void }) {
...
+        if (!open) onClose()
+      }}
+      size="lg"
+      title="New provider"
+      tabs={tabs}
+      submitLabel="Add provider"
+      onSubmit={submit}
+      isPending={create.isPending}
+      isSubmitDisabled={!canSubmit}
+      isDirty={providerId !== "" || apiKey.trim() !== ""}
```
khaledosman: **The known tab's `isDirty` still omits four fields the form owns:** `name`, `apiBase`, `clientArgsText` and `credentials`. The Advanced toggle at `:283` renders with no provider selected, so: open the dialog, press Advanced, type a rename or a `client_args` blob, press Escape. The dialog closes with no guard and the work is gone.

`feedback.md:236-254` calls the hand-listed predicate "the create-only degenerate form… it drifts, because a field added to the form has to be remembered in a second place", and prescribes the snapshot, which #1062's fix commit in this same series adopted. `const draft = JSON.stringify({ providerId, apiKey, name, apiBase, clientArgsText, credentials })` against a `useRef(draft)`; `key={addOpenCount}` already reseeds it per open, so nothing else is needed.

web/src/features/providers/ProvidersPage.tsx
```diff
@@ -329,8 +369,39 @@ function CustomProviderForm({ onClose }: { onClose: () => void }) {
...
+        if (!open) onClose()
+      }}
+      size="lg"
+      title="New provider"
+      tabs={tabs}
+      submitLabel="Add provider"
+      onSubmit={submit}
+      isPending={create.isPending}
+      isSubmitDisabled={!canSubmit}
+      isDirty={
```
khaledosman: Same shape one field narrower: the custom tab omits `clientArgsText` and `providerType`. Type only a `client_args` blob and press Escape and it goes unguarded. Same snapshot fix.

web/src/features/providers/ProvidersPage.tsx
```diff
@@ -282,23 +322,24 @@ function KnownProviderForm({ onClose }: { onClose: () => void }) {
...
-        <Button variant="primary" isDisabled={!canSubmit} onPress={submit}>
-          {create.isPending ? "Adding…" : "Add provider"}
-        </Button>
-        <Button variant="ghost" onPress={onClose}>
-          Cancel
-        </Button>
-        <ConnectionTest getPayload={buildPayload} />
-      </div>
-    </div>
+      <ConnectionTestResult test={test} />
```
khaledosman, line 349: **The connection-test outcome can render below the fold with nothing scrolling it in.** Moving it out of the footer put it after `ClientArgsField`, a four-row textarea. `.otari-form-dialog` caps at `calc(viewport - 15rem)` (`globals.css:2164`), so an `lg` dialog on an 800px viewport has about 386px of body and the custom tab's five controls already overflow it.

Failure: fill the custom form, press Test connection, get the four-line `discovery_unsupported` verdict. The footer button goes from "Testing…" back to "Test connection" and, to a sighted operator, nothing else happens. A screen reader is fine, since the result keeps `role="status"`.

`FormDialog` already solves this for `error` by scrolling the body to top (`:110-114`). Either give `ConnectionTestResult` a ref and `scrollIntoView({ block: "nearest" })` in an effect keyed on `test.data ?? test.error`, or hoist that into `FormDialog` beside the error scroll. Same at `:454`.

web/src/features/providers/ProvidersPage.tsx
```diff
@@ -56,28 +57,39 @@ import {
...
+      }}
+    >
+      {test.isPending ? "Testing…" : "Test connection"}
+    </Button>
+  )
+}
 
+function ConnectionTestResult({ test }: { test: ConnectionTestState }) {
   return (
     <div className="flex flex-col gap-1.5">
```
khaledosman: `ConnectionTestResult`'s wrapper `<div className="flex flex-col gap-1.5">` renders unconditionally as a child of the body's `flex flex-col gap-4`, so every provider dialog now carries 16px of dead space before a test is ever run. The wrapper is also vestigial: it held a button and a result and now holds one child. Drop it and render the `role="status"` span directly with `empty:hidden`.

khaledosman (approved):
All six fixed, and two of them better than what I proposed: the `autoFocus` went to the real control inside `ProviderComboBox` rather than to a wrapper, with `menuTrigger` paired so the chevron still opens the full catalog and a test asserting `aria-expanded="false"`; and the connection test split into a button in `footerStart` and a `role="status"` result in the body, which keeps the announcement without keeping the height change.

Four non-blocking things below. The two `isDirty` ones are the same class as the `apiKey` gap I raised, which you fixed exactly as named; the rest of the fields on both tabs still have it, and this is now the only form in the series still hand-listing that predicate.

Gates clean, CI green.

---

web/src/features/providers/ProvidersPage.tsx
jigjigjig: Right, and worse than four fields: the list was two of six. Fixed in a564459e with the snapshot, `JSON.stringify({ providerId, apiKey, name, apiBase, clientArgsText, credentials })` against a `useRef`, reseeded per open by `key={addOpenCount}`.

I had found the same thing myself while surveying every `isDirty` in the series and had it queued as a conversion; your thread is the reason it landed here rather than in a later sweep, and the Bedrock region is the case that makes it more than tidiness.

Test: open the dialog with no provider chosen, press Advanced, type a rename, press Escape, and the guard has to appear. It fails with the old predicate restored. The custom tab has its own case for the client-options-only path.

One note for the rebase: this is a hand-rolled snapshot, not `useDirtySnapshot`, because that hook landed on main with #1037 and is not in this tree yet. It converts to the hook when this branch rebases.

web/src/features/providers/ProvidersPage.tsx
jigjigjig: Fixed in the same commit, same shape: `JSON.stringify({ name, providerType, apiBase, apiKey, clientArgsText })`. The case that pins it types a `client_args` blob and nothing else, then presses Escape.

web/src/features/providers/ProvidersPage.tsx
jigjigjig: Fixed in a564459e, locally rather than in `FormDialog`. `ConnectionTestResult` carries a ref and calls `scrollIntoView({ block: "nearest" })` in an effect keyed on `test.data ?? test.error`, in both tabs, since the verdict is the only thing on these forms that appears at the bottom without the operator having moved.

I left `FormDialog` alone deliberately: the error scroll it already owns is about a banner it renders itself, and hoisting "scroll my last child into view" into the shared surface would give six other dialogs a behavior none of them asked for. If a second form grows the same need, that is the moment to hoist it.

jsdom implements no scrolling at all, so `scrollIntoView` is absent on every element and the effect threw. It is stubbed in `src/tests/setup.ts` beside `ResizeObserver` and `window.scrollTo`, with the same "only when absent" guard, which also makes the call assertable: the test spies on it and checks the instance is the live region. Removing the effect fails it.

web/src/features/providers/ProvidersPage.tsx
jigjigjig: Fixed in a564459e: the wrapper is gone and the `role="status"` span carries the layout classes with `empty:hidden`, so the parent's `gap-4` reserves nothing before a test is run.

One thing I decided rather than skipped: the span still renders unconditionally rather than being rendered only when there is a verdict, because a live region has to exist before it has something to say. `empty:hidden` is `display: none`, which is not ideal for a region that is about to be announced, so if the first verdict is ever missed by a screen reader the fix is to take it out of flow rather than hide it. Not worth the extra classes today, and worth knowing where to look if it shows up.

web/src/features/providers/ProvidersPage.tsx
```diff
@@ -380,61 +483,55 @@ function CustomProviderForm({ onClose }: { onClose: () => void }) {
...
-      {tab === "known" ? (
-        <KnownProviderForm onClose={onClose} />
-      ) : (
-        <CustomProviderForm onClose={onClose} />
-      )}
-    </Section>
+  return tab === "known" ? (
+    <KnownProviderForm isOpen={isOpen} onClose={onClose} tabs={tabs} />
+  ) : (
+    <CustomProviderForm isOpen={isOpen} onClose={onClose} tabs={tabs} />
```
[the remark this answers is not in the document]
jigjigjig: Real as a product observation, and not a regression from this PR: switching tabs has always discarded the half-filled one, because each tab was its own panel with its own state before this series and is its own component with its own state now. `AddProviderForm`'s docstring says so in as many words ("a half-filled tab still does not survive a switch away from it, which is what it did as a panel"), so this PR inherits the behavior rather than introducing it.

It is still a loss the rest of this series exists to prevent, so I am not arguing it away. What I am not doing is fixing it inside this PR, and the reason is the size of the fix rather than the merit:

- Keeping both drafts alive means hoisting every field of both forms into `AddProviderForm` (a dozen `useState` pairs, two mutations, two validity derivations), which is a refactor of both halves rather than a guard.
- Guarding the tab press means `FormDialog` growing a way to gate an arbitrary action on its own dirty state. Today `isDirty` gates dismissal only, and the guard lives inside the footer. That is a shared-surface API decision, and it would land in the same file six other dialogs render through.

Either one is a change I would want reviewed on its own diff, not folded into a PR whose threads are otherwise closed and whose CI is green. So this is going to a follow-up issue with your two options recorded as the two candidates, and I am resolving the thread here so the merge queue is not held by a decision that belongs in its own change.

One thing worth adding for whoever picks it up: the tab row is also where a pasted provider key can be lost, which makes it the same class as the `isDirty` gaps fixed in this PR rather than a cosmetic annoyance.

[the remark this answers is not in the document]
jigjigjig: Filed as #1081, with both of your options recorded as the two candidates and the check each one implies.

PR #1046, opened 2026-09-10
feat(dashboard): reorganize the routing page as list and detail

khaledosman:
> **Stacked on #1038.** That PR merges first. This one deliberately keeps its
> dialog as the create-and-edit surface: the editor stays in a `FormDialog` over
> the page and nothing is edited in a column.
>
> #1038 itself owes a rebase onto `main`, because its parent (#1032, which added
> `FormDialog`) was squash-merged and left the child conflicted. This PR
> therefore owes a rebase of its own once #1038 is rebased or merged. Two
> protections are also missing here, both a consequence of stacking:
> `.coderabbit.yaml` sets `base_branches: ["main"]`, so CodeRabbit skips a child
> and has to be summoned by comment; and `protect-main` covers the default branch
> only, so this can report `CLEAN` and be mergeable with no approval. **Please do
> not merge it on that basis.**

## Description

The Routing page listed every policy in one wide table and put the editor above
it, so nothing said which row you were changing and the facts about a policy were
spread across six columns you had to scroll sideways to read.

It is now two columns. The left one lists the policies and aliases, with the
"Create policy" control at its top, and each row shows a policy's name and one
line saying what it serves. Click a row and the right column shows that policy on
its own: what serves it, what is tried if that fails, which guardrails always run,
who it applies to, where it came from, and whether it is an alias. That column is
also where the policy's own controls live, so Edit, Delete and Examples all sit
beside the facts they act on rather than in a lane at the far right of a table.

With no policies yet, the left column is one big "Create your first policy"
button and the right one carries the getting-started steps.

On a phone the two columns become one: the list, until you open a policy, and
then the policy with a way back to the list. Every control is at the 44px touch
floor there and none of them depends on hover.

Nothing about what a policy *is* changed, and nothing about who may write one: an
operator still writes deployment-wide, an admin still writes into the selected
workspace, and a member still gets a page with no controls on it at all. A policy
defined in `config.yml` still says so instead of offering an edit, and a policy
using options the form cannot represent is still read-only with the sentence
explaining why.

### Where the Examples panel went

It used to open as an accordion row underneath its own row in the table. It is now
a panel at the foot of the detail column, opened by a control in that column's
header. The detail column is the better home for it: the panel is about the one
policy that column is already showing, so it no longer has to be tucked between
two unrelated rows, and it gets the width to lay out a candidate pool and a
warmth bar without a table's lanes squeezing it. It stays operator-only, because
it reads the deployment-wide `/v1/routing/status`.

### The new primitive

`ListDetail` and `ListDetailRow` are in `design-system/layout/`, because the shape
will be reused. They own the 1:1.75 ratio, the hairline between the columns, the
row separators, and the single-column swap below `md`; which row is selected is a
prop and a callback, so the page keeps being the one place that knows what is
open. They import nothing under `src/` beyond their own layer, which
`architecture.test.ts` proves.

Two things worth knowing about the deliberate choices in it. There is no
per-row Edit or Delete: both would have to be repeated in the detail column
anyway, since below `md` the list is not on screen while a policy is being read,
and one home per action beats two. And the empty column's "click anywhere" is a
real `<button>`, not a div with a click handler, so a keyboard reaches it.

One known limitation: below `md` with no policies, the column that shows is the
list, so the getting-started steps in the detail column are not visible on a
phone. The action is what a phone gets, which seems the right way round, but it
is a trade rather than an oversight.

## How to test it locally

```
make dashboard && otari serve      # or however you run the gateway
```

Then sign in and go to Routing → Policies.

1. With no policies, click anywhere in the left column. The create dialog opens.
   Tab to it instead of clicking: it takes focus as a button.
2. Create two policies, one plain and one with a fallback chain. Both appear as
   rows with a summary line. Click each: the right column fills in, and the row
   you clicked wears the selection tint.
3. From the right column, Edit opens the same dialog with the policy loaded, and
   Delete opens a confirm dialog naming the policy. Cancel it, then confirm it:
   the row goes and the column returns to "pick a policy".
4. Create a policy with a learned router (`+ Let a router pick…`), open it, and
   press Examples. The readiness panel opens at the foot of the column.
5. Narrow the window to 390px. One column at a time, and the Back control appears
   above an open policy.
6. Switch themes. The selection tint and the hairlines are both tokens, so both
   follow.

Automated, and run locally:

- `pnpm --dir web run lint` (Biome, including the layer boundaries): clean.
- `pnpm --dir web run typecheck`: clean.
- `pnpm --dir web test --run src/features/routing src/design-system src/styles`:
  43 files, 3828 tests, all passing. That covers the rewritten
  `RoutingPage.test.tsx` (69 tests: the three role readings, the config.yml and
  not-representable read-only cases, the four delete surfaces, the confirm
  dialog's error handling, and the `?target=` deep link), the new
  `ListDetail.test.tsx`, and the foundation gates, which include the one that
  fails when a design-system prop no story passes.
- `pnpm --dir web test --run src/architecture.test.ts`: 30 tests, passing. Run on
  its own, because it plants probe files that race the tree walk in
  `monoScale.test.ts` when the two run together.

Not run locally, and why:

- **The full Vitest suite and the Storybook sweep.** Deliberate: an earlier
  attempt at this task ran several suites at once and froze the machine. CI runs
  the whole suite and builds the catalog, so both are covered there.
- **The Playwright suites.** They need a live gateway. `dashboard.spec.ts` and
  `parity.management.spec.ts` are updated for the new list shape (two new helpers,
  `listRow` and `detailColumn`, replace the `rowheader` locators), and the roles
  and accessible names were kept coherent with what the Vitest tests query, but
  the e2e runs have not been executed here.
- Screenshot baselines are gitignored and `/routing` already has an entry, so
  there is nothing to commit and no new entry owed.

Nothing on the Python side changed, so no OpenAPI or Postman artifact is stale,
and no route moved, so `routeTree.gen.ts` is unchanged.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [x] Refactor
- [x] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes mozilla-ai/otari-ai#2109

## Checklist

On the third box: this change is entirely under `web/`, so the dashboard's own
checks are the applicable ones and they are listed above. No Python changed, so
the API contract did not move and there was no spec to regenerate.

Docs updated: `web/design/layout.md` gains the band-component row and a
list-and-detail page recipe, `web/design/DESIGN.md` gains the module row, and
`web/design/data.md` loses `otari-routing-table` from its closed list of table
places, since that class and its CSS went with the table.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

The design-system primitive was drafted in an earlier interrupted run and reviewed
rather than taken as finished: its docstring justified the 1:1.75 ratio partly
with "the right one a form", which #1038 made untrue, and the row was drawing its
own bottom border where the frame should divide its children the way
`SettingsGroup` does. Both are corrected here, along with the touch floor on the
controls the detail column adds.

## Summary

- Replaced the Routing table with a responsive list-and-detail layout.
- Added policy selection, detail actions, mobile navigation, and improved empty states.
- Added reusable `ListDetail` and `ListDetailRow` design-system components.
- Updated documentation, styles, tests, and end-to-end helpers for the new layout.

## Benefits

- Improves routing policy navigation on small and large screens.
- Keeps permissions, read-only behavior, and policy semantics unchanged.
- Provides reusable layout primitives for similar pages.

## Technical notes

- No Python, API, or route-generation changes were made.
- Routing, design-system, architecture, lint, and type checks pass.

### DIFF
--- a/web/design/DESIGN.md
+++ b/web/design/DESIGN.md
@@ -90,6 +90,7 @@ for a component, not for those.
 | `indicators/Dot` · `/Badge` · `/DismissChip` · `/Kbd` · `/Avatar` | one each |
 | `indicators/Chip` | `Chip`, and the `ChipTone` type |
 | `layout/Divider` | `Divider` |
+| `layout/ListDetail` | `ListDetail`, `ListDetailRow`. See [layout.md](layout.md) |
 | `content/Markdown` | `Markdown` |
 | `ProductMark` | `ProductMark`. At the top level: it belongs to no topic |
 | `@/shared/components/access/EntitlementGate` · `/UnavailableHere` · `/MissingGatewayAddressNotice` | one each |
--- a/web/design/data.md
+++ b/web/design/data.md
@@ -111,7 +111,7 @@ The lane does the aligning; see [layout.md](layout.md).
 tracks horizontal scroll so a first column can pin, and the per-page class is what
 `globals.css` hangs that page's column widths and pinning off. The list is closed:
 `otari-keys-table`, `otari-models-table`, `otari-providers-table`,
-`otari-routing-table`, `otari-domains-table`, `otari-members-table`,
+`otari-domains-table`, `otari-members-table`,
 `otari-provider-keys-table`, `otari-workspaces-table`, `otari-accounts-table`,
 `otari-activity-table`, `otari-budgets-table`, `otari-pricing-table`,
 `otari-mcp-table`, `otari-rate-overrides-table`. Inventing one at the call site
--- a/web/design/layout.md
+++ b/web/design/layout.md
@@ -41,6 +41,7 @@ constants for these; `Section` is the one place the pair is named.
 | `SettingRow` | `label`, `configKey?`, `help?`, `control`, `nested?`, `error?` | One setting inside a group |
 | `KpiStrip` + `KpiCell` | see [metrics.md](metrics.md) | The metrics band |
 | `TableScrollFrame` | `className`, children | Around a wide table |
+| `ListDetail` + `ListDetailRow` | `listLabel`, `listAction?`, `list`, `empty?`, `onEmptyPress?`, `detail`, `detailLabel?`, `isDetailShown`, `onShowList`, `backLabel?` | A set of records where reading one is most of the work |
 
 **`PageIntro` renders its own `<h1 className="text-display">`**, so never put a
 heading inside it and never spell `text-display` on a page yourself. The sentence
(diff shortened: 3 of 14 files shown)


khaledosman:
@coderabbitai review

PR #1048, opened 2026-09-10
feat(dashboard): move the organization provider-key form into FormDialog

jigjigjig:
## Description

Adding or editing an organization's own upstream credential appended a band to the page. Both open the dialog now, at `md`.

The edit heading was one string, `Edit <name>`. A dialog title is a noun phrase, so it is **"Edit provider key"** with the key it names in the description, the same split the routing and budget dialogs took. Create is **"New provider key"** with the trigger and submit both saying "Add provider key", per the labels rule: a provider key exists upstream and is attached here.

The heading's action stays visible while the dialog is open, and stays **disabled rather than hidden** when the server has no secret key, which is unchanged and is the rule for a control carrying its own reason nearby.

## No screenshot capture, and why

This is the first page in the series that the harness cannot reach. Its page is gated on the `organization_providers` surface, which only a **hosted** deployment publishes (`HOSTED_SURFACES` in `src/gateway/api/routes/bootstrap.py`), and the e2e suite boots a **standalone** gateway. Driven there, the route answers with "Providers is not available here", which is the correct behavior and not something a capture can work around.

Covering it would mean adding a third gateway to the harness beside the standalone and hybrid ones, which is infrastructure rather than a test. The reason is recorded in `authenticated.spec.ts` where a capture would otherwise go, so the next person does not rediscover it.

Component coverage is unaffected: the dialog is exercised by 19 cases in `OrganizationProviderKeysPage.test.tsx`.

## How to test it locally

    pnpm --dir web test -- OrganizationProviderKeysPage

- `pnpm --dir web test` — 5353 pass across 154 files; 19 in `OrganizationProviderKeysPage.test.tsx`, one of them new.

### What would have to break for the new test to fail

Restoring the `!adding` condition on the heading's action fails `keeps the page's add action visible while the dialog is open`.

The other 18 needed no change, which is worth one sentence rather than being passed over: they already disambiguate the two "Add provider key" buttons with `getByRole(..., { hidden: false })`, because the provider combo box's popover `aria-hides` the rest of the page. The modal now hides the page behind it for the same reason, so the same query keeps selecting the dialog's submit. That is luck rather than design, and `within(dialog)` would be the deliberate way to write it.

### On the unslop pass

Run over this diff before opening. Nothing to remove: the change is a frame swap, the deleted action row took its own comment with it, and the one comment added states why the description carries the key's name.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Part of #1031

Stacked on #1045 → #1041 → #1040 → #1038 → #1037. Review those first; this branch contains their commits.

## Checklist

      19 cases in `OrganizationProviderKeysPage.test.tsx`, one new. No screenshot
      capture is possible; see above. The dashboard's tests live under
      `web/src` and `web/e2e` rather than the repo-root `tests/`.
      No design-doc change: the rules were written down in #1032.
      No API change. Both the create and update bodies are unchanged, including
      the deliberate omission of `api_key` when the secret was not retyped.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

The missing capture was found by writing one and watching it fail, not by reasoning about surfaces beforehand.

### DIFF
--- a/web/src/features/organization/OrganizationProviderKeysPage.test.tsx
+++ b/web/src/features/organization/OrganizationProviderKeysPage.test.tsx
@@ -26,6 +26,8 @@ interface MockOpts {
   keys?: OrgProviderKey[]
   context?: OrganizationContext
   catalog?: { id: string; name: string }[]
+  // Refuse the create, so a test can read what a refusal leaves behind.
+  createFails?: boolean
 }
 
 function mockApi(opts: MockOpts = {}) {
@@ -43,6 +45,9 @@ function mockApi(opts: MockOpts = {}) {
     if (url.includes("/provider-keys")) {
       if (method === "GET") {
         return jsonResponse({ count: keys.length, data: keys })
+      }
+      if (method === "POST" && opts.createFails) {
+        return jsonResponse({ detail: "Provider key already exists" }, 409)
       }
       // Every write answers with a key-shaped body the page only re-reads
       // through the invalidated list, so one row is enough for all of them.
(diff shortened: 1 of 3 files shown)


web/src/features/organization/OrganizationProviderKeysPage.tsx
```diff
@@ -491,14 +491,17 @@ export function OrganizationProviderKeysPage() {
         </InfoBanner>
       ) : null}
 
-      {adding && secretKeyConfigured ? (
-        <KeyForm editing={null} onClose={() => setAdding(false)} />
-      ) : null}
+      <KeyForm
```
khaledosman, line 505: **The create `KeyForm` is mounted unconditionally with no key, so its draft and its mutation survive the close.** The edit `KeyForm` nine lines down carries `key={editing.id}` (`:503`); this one has neither a counter nor a key, and `draft` (`:126`) includes `apiKey`.

Failure: add "Production" for OpenAI with a pasted secret. On success the dialog closes. Press "Add provider key" again: the name, the provider and the plaintext secret are all still there. `seeded` (`:142`) still holds the empty draft, so the form is already `isDirty` and arms the unsaved-changes guard before anything is typed. A failed create leaves `create.error` (`:198`) mounted, so the next open shows the previous refusal's banner.

Fix: the `9487bb1e` pattern. `const [addOpenCount, setAddOpenCount] = useState(0)`, bumped at the trigger (`:448`), and `key={addOpenCount}` here. One trigger only, so nothing to miss; and `useCreateOrgProviderKey()` is inside `KeyForm` (`:124`), below the key, so the remount clears the mutation too.

web/src/features/organization/OrganizationProviderKeysPage.test.tsx
```diff
@@ -77,6 +77,21 @@ afterEach(() => {
 })
 
 describe("OrganizationProviderKeysPage", () => {
+  it("keeps the page's add action visible while the dialog is open", async () => {
```
khaledosman, line 85: The only test added covers the trigger staying visible, which is not the behavior most likely to break here. #1062 in this same stack added the one that would have caught the finding above ("opens on a blank draft after a close"); copy that shape.

khaledosman (changes requested):
Gates clean: no route or schema touched; no generated-artifact drift; no dependency change; conventional title; no em dashes or `--` separators. The `e2e` failure on the last run was the pre-rebase `dismissComboBox` regression and is gone at this head.

One blocker, the same shape as #1045: the create form keeps its draft, and here the draft holds a provider API key. The screenshot note at `authenticated.spec.ts:467` checks out, `organization_providers` is in `HOSTED_SURFACES` only (`bootstrap.py:127-129`) and the harness boots standalone.

Correct here and worth copying back into #1045: `canSubmit` (`:144`) deliberately excludes `pending`, so the submit keeps its fill in flight.

---

web/src/features/organization/OrganizationProviderKeysPage.tsx
jigjigjig: Fixed in 164ea0bb, in the `9487bb1e` shape: `addOpenCount` beside `adding`, one `openAdd()` that bumps it (the page has the one trigger), and `key={addOpenCount}` on the create form. `useCreateOrgProviderKey()` is inside `KeyForm`, so the remount clears the refusal banner along with the draft.

web/src/features/organization/OrganizationProviderKeysPage.test.tsx
jigjigjig: Added in 164ea0bb, two of them. "opens on a blank draft after a create" walks the full flow (pick Anthropic, name it, paste a secret, submit, the dialog closes), reopens, and asserts the name, the secret and the provider are all empty and that Escape closes without arming the guard, which is the second half of the finding above. "does not carry a refused create's banner into the next open" refuses the POST with a 409, reads the banner, discards, reopens and asserts no alert. Both fail with the `key` removed.

The trigger-visibility case stays; it is cheap and it pins where focus returns.

khaledosman (approved):
Both fixed, in the shape the rest of the series settled on: `addOpenCount` bumped by the single `openAdd()` at `:315-319`, `key={addOpenCount}` at `:506`, and both mutations at `:124-125` inside `KeyForm` so the refusal banner resets with the draft.

The tests are the part worth calling out. `:100` walks a full create with a real secret, reopens, asserts all three fields blank, and *then* presses Escape to prove the dialog closes rather than arming the guard, which is the `seeded`/`isDirty` half I did not think to ask for. `:177` covers the error carryover. Both fail against the pre-fix code.

Three things I checked and cleared rather than found:

- The JSON snapshot is safe here: it covers every field of `KeyDraft` (`:71-82`), holds no unstable value, and key-order drift is unreachable because `ProviderCredentialFields` writes `{ ...values, [key]: next }` (`providerFields.tsx:93`), which keeps position for existing keys.
- `draft.apiKey` living in JS memory between close and next open is what the rung prescribes, not a leak: the DOM node is gone (`FormDialog` mounts its body only while open) and nothing puts the key in a query key, a log, or a mutation key.
- `submit` (`:150`) has no `create.isPending` guard where #1045's now does. That is fine rather than inconsistent, since `FormDialog` blocks all three paths (`:218`, `:232`, `:310`). Worth knowing that #1045's guards are belt-and-braces, not a hole this page still has.

Gates clean, CI green. Nothing to change.

---

PR #1052, opened 2026-09-10
fix(dashboard): key Storybook's API mock on the API root

jigjigjig:
## Description

Every story in the catalog has been reaching the network and 404ing since #1026, and nothing reported it.

#1026 moved the API to `/api/v1`. It updated every story's own mock table, but `.storybook/apiRouting.ts` held two more copies of the old path: `GATEWAY_PREFIX` matched `/v1/`, so no gateway request was recognized as one, and `BASELINE` keyed the organization context at `/v1/organizations/me`, which every story needs because `SelectedWorkspaceProvider` wraps all of them. A request that misses the mock goes to the network, and the static catalog answers 404.

Both now derive from `API_ROOT`. Its own comment promises that moving the API is "a change here and nowhere else", which only holds if nothing else writes the path down.

**Why the per-PR gate did not catch it.** A dead mock still builds and still paints, so `Otari Design System` stays green: the catalog compiles and every story renders something. Only the console-clean sweep asks whether a story rendered *without errors*, and that runs on main.

**The tests could not have caught it either, which is the part worth fixing properly.** All seven routing cases spelled `/v1/` themselves, so they were checking the mock against the same stale literal the mock had. The suite was self-consistent with the bug and would have gone on passing forever. They build their paths from `API_ROOT` now, and one added case pins both halves of the contract: a request at the current root is intercepted, and one at the old prefix is not. Reverting `GATEWAY_PREFIX` to `"/v1/"` fails eight of the ten.

The mock's usage docs in `apiMock.tsx` and one comment in `appContext.tsx` taught the old key as well, which is how the next story written would have inherited it.

This is a regression from #1026, not a fault in the design system work above it.

## How to test it locally

    pnpm --dir web run storybook:build
    python3 -m http.server 6006 --directory web/storybook-static &
    pnpm --dir web exec node .storybook/smoke.mjs

Measured on a local build, before and after this change:

| | renders | result |
| --- | --- | --- |
| before | 387 stories x 2 themes = 774 | **771 failing**: 757 "console errors", 14 timeouts |
| after | the same 774 | **`ALL RENDERED CLEAN`** |

The failures before were all the same 404, either reported as a console error or as a 20s timeout where the component waited on the request.

`pnpm --dir web exec vitest run .storybook` covers the routing unit, ten cases.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

None. Found while working the dashboard series; it is a regression from #1026 rather than an issue anybody had filed, and it is small enough to stand on its own.

## Checklist

- [x] I have added or updated tests that cover my change (`tests/unit`, `tests/integration`). `.storybook/apiRouting.test.ts`: every case now derives its path from `API_ROOT`, plus one that pins the old prefix as *not* the gateway's. Verified by reverting the fix and watching eight of ten fail.
- [x] I ran the Definition of Done checks locally (`make lint`, `make typecheck`, `make test`). `make lint` passes. The dashboard's own three are the relevant ones here and all pass: `pnpm --dir web run lint`, `run typecheck`, `test`.
- [x] Documentation was updated where necessary. The mock's usage examples and one decorator comment named the old path.
- [ ] If the API contract changed, I regenerated the OpenAPI spec (`uv run python scripts/generate_openapi.py`). No API contract change; this follows one.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

The two counts in the table are from running the sweep against a local static build before and after the change, not inferred from the rule. The second stale copy (`BASELINE`) was found by the new test rather than by reading: with only `GATEWAY_PREFIX` fixed, the organization-context case failed with a 501.

**NOTE:**

### DIFF
--- a/web/.storybook/appContext.tsx
+++ b/web/.storybook/appContext.tsx
@@ -33,7 +33,7 @@ export const withAppContext: Decorator = (Story, context) => {
     <DeploymentProvider value={bootstrap(overrides)}>
       {/* Inside the deployment, and inside the query client `apiMock` provides:
           it seeds itself from `useOrganizationContext()`, so a story that cares
-          which workspace is selected mocks `${API_ROOT}/organization/context`. */}
+          which workspace is selected mocks {API_ROOT}/organizations/me. */}
       <SelectedWorkspaceProvider>
         <Story />
       </SelectedWorkspaceProvider>
--- a/web/.storybook/appContext.tsx
+++ b/web/.storybook/appContext.tsx
@@ -33,7 +33,7 @@ export const withAppContext: Decorator = (Story, context) => {
     <DeploymentProvider value={bootstrap(overrides)}>
       {/* Inside the deployment, and inside the query client `apiMock` provides:
           it seeds itself from `useOrganizationContext()`, so a story that cares
-          which workspace is selected mocks /v1/organization/context. */}
+          which workspace is selected mocks `${API_ROOT}/organization/context`. */}
       <SelectedWorkspaceProvider>
         <Story />
       </SelectedWorkspaceProvider>
--- a/web/.storybook/apiMock.tsx
+++ b/web/.storybook/apiMock.tsx
@@ -27,14 +27,14 @@ import { type ApiMocks, pathOf, route } from "./apiRouting"
  *
  *   parameters: {
  *     api: {
- *       "/v1/settings/mail": { configured: true, from_address: "otari@example.com" },
- *       "/v1/organizations/me": organizationContext(),
+ *       [`${API_ROOT}/settings/mail`]: { configured: true, from_address: "otari@example.com" },
+ *       [`${API_ROOT}/organizations/me`]: organizationContext(),
  *     },
  *   }
  *
  * To exercise a failure, give the path a `$status` envelope instead of a body:
  *
- *   api: { "/v1/settings/mail": { $status: 503, $body: { detail: "No transport." } } }
+ *   api: { [`${API_ROOT}/settings/mail`]: { $status: 503, $body: { detail: "No transport." } } }
  *
  * The `$` prefix is what makes that unambiguous, and it is worth the ugliness. An
  * earlier version guessed instead -- "has a `status` or `body` key, therefore it is
--- a/web/.storybook/apiRouting.test.ts
+++ b/web/.storybook/apiRouting.test.ts
@@ -1,19 +1,25 @@
 import { describe, expect, it } from "vitest"
 
+import { API_ROOT } from "@/shared/api/client"
 import { type ApiMocks, pathOf, route } from "./apiRouting"
+
+// Every path here is built from `API_ROOT` rather than written out. These cases
+// used to spell `/v1/`, which is why they went on passing when the API moved:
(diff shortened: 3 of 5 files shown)


web/.storybook/appContext.tsx
```diff
@@ -33,7 +33,7 @@ export const withAppContext: Decorator = (Story, context) => {
     <DeploymentProvider value={bootstrap(overrides)}>
       {/* Inside the deployment, and inside the query client `apiMock` provides:
           it seeds itself from `useOrganizationContext()`, so a story that cares
-          which workspace is selected mocks /v1/organization/context. */}
+          which workspace is selected mocks `${API_ROOT}/organization/context`. */}
```
khaledosman: The path named here does not exist. `SelectedWorkspaceProvider` queries through `useOrganizationContext`, whose endpoint is `/organizations/me` — the single key in `BASELINE` (`apiRouting.ts:57`) and the only spelling in `web/src/client/schema.ts:1584`. `git grep 'organization/context'` across `web/` matches this comment and nothing else, so a reader following it declares a key that is never hit and gets a 501 they cannot explain.

The wording predates the PR, but this is the line the PR edits. `mocks {API_ROOT}/organizations/me` — plain, since this file does not import `API_ROOT` and the backticks read as an unresolved template reference.

khaledosman (approved):
The fix is right and the test pins it. Verified rather than assumed:

- `apiRouting.ts:24` derives `GATEWAY_PREFIX` from `API_ROOT`, the same constant `apiUrl()` uses (`shared/api/client.ts:27-31`), so there is no second literal. `git grep '/v1/'` over `web/.storybook/` at this head returns only prose, and every `parameters.api` key across the eight story files is already built from `` `${API_ROOT}/…` ``.
- Matching is exact-key then query-stripped pathname (`:84-88`), so `/api/v1/keys` cannot swallow `/api/v1/keys/{id}`, and an unmatched gateway path returns 501 with the path in `detail` (`:114-120`) rather than `undefined` or the network.
- The test genuinely regresses: with `apiRouting.ts` swapped back to main's copy, 8 of 10 fail including `apiRouting.test.ts:76`; unmodified, all 10 pass. `vite.config.ts:178-181` already includes `.storybook/**` in the vitest run, so it is a real PR gate.

Worth landing ahead of #1038: its two new `PolicyForm` stories key on `${API_ROOT}` and cannot resolve until this is in their base.

One stale reference below.

---

web/.storybook/appContext.tsx
jigjigjig: Confirmed and fixed in 785d88b0. `git grep 'organization/context' -- web` matched that comment and nothing else, and `BASELINE` (`apiRouting.ts:57`) keys on `${API_ROOT}/organizations/me`, which is also the only spelling in the generated client. So the comment was sending a reader to declare a key nothing hits.

It now reads `{API_ROOT}/organizations/me`, plain rather than in backticks, for the reason you gave: this file does not import `API_ROOT`, so a template literal reads as a reference to something that is not there.

The grep is empty afterwards, which is the check that matters here: the wrong path existed in exactly one place, and it does not any more.

PR #1054, opened 2026-09-10
feat(dashboard): add ListDetail, the list-and-detail page frame

khaledosman:
## Description

Adds a reusable page frame to the design system: a list of records on the left, the one that is
open on the right, at a 1:1.75 ratio, divided by a hairline. Below the tablet breakpoint one
column shows at a time, with a way back to the list, because two columns on a phone give neither
one a readable width and stacking them buries the record you opened under the whole list.

The ratio, the divider and the column swap live in the component rather than at each call site,
so the next screen that wants this shape does not have to work any of it out again.

**Nothing uses it yet, and that is a deliberate exception worth naming.** The design system's own
rules decline a component with no consumer, on the grounds that it is an API to maintain and a
claim about the product that is not true. This one ships ahead of its first consumer because it
was asked for as a reusable shape rather than for one screen. The routing policies page was the
candidate and stays a table: a flat list of policies has no grouping or hierarchy for a detail
column to earn, so a table fits it better. `layout.md` records that reasoning next to the
component, so the next reader finds the decision rather than re-opening it.

It is presentational and holds no state of its own. Which record is open, and which column shows
on a phone, are both passed in, which is what lets a page keep one source of truth for what the
reader has selected.

## How to test it locally

1. `pnpm --dir web run storybook`, then `Design system → Layout → ListDetail`. The stories cover
   the two-column frame, a selected row, per-row actions, the empty column with and without a
   press target, and the phone view where one column shows at a time.
2. Resize below 768px in the two-column story: the list is what shows, opening a record swaps to
   it, and the back control returns. The back control and the row actions are at least 44px tall
   on a phone, since they are the only way through the page there.
3. Switch the catalog between light and dark: the selected row and the divider are both token
   driven, so neither needs a per-theme rule.

Automated coverage: `ListDetail.test.tsx` beside the component, plus the design-system layer gate
in `src/architecture.test.ts` (this layer may import nothing else under `src/`) and the token
gates in `src/styles/foundation.test.ts` (no raw hex, no numbered palette class, no `bg-white`).

Checks run locally: `pnpm --dir web run lint`, `pnpm --dir web run typecheck`, Vitest over
`src/design-system/layout` and `src/styles` (3543 passing) and `src/architecture.test.ts`
(30 passing, run on its own because its probe files race another suite's tree walk). The full
suite and the catalog build are left to CI. Nothing under `src/gateway/` changes, so there is no
OpenAPI or Postman artifact to regenerate.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [x] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Part of mozilla-ai/otari-ai#2109. That issue asked for the routing policies page to be
reorganized into two columns. The page reorganization was reverted by decision after review, on
the grounds above, and the issue stays open pending a call on the page itself. This PR keeps the
one durable piece of that work.

## Checklist

- [x] I have added or updated tests that cover my change (`ListDetail.test.tsx`, plus the layer and token gates).
- [x] I ran the Definition of Done checks locally (the `web/` equivalents: `pnpm --dir web run lint`, `run typecheck`, and scoped Vitest; `make lint` and the Python suites cover no part of this change).
- [x] Documentation was updated where necessary (`web/design/DESIGN.md`, `web/design/layout.md`).
- [ ] If the API contract changed, I regenerated the OpenAPI spec (no API change).

## AI Usage

- [ ] No AI was used.

**Any additional AI details you'd like to share:** The component was written for the routing page
reorganization, reviewed, then separated from it when that page was kept as a table.

### DIFF
--- a/web/src/design-system/layout/ListDetail.test.tsx
+++ b/web/src/design-system/layout/ListDetail.test.tsx
@@ -1,0 +1,186 @@
+import { render, screen } from "@testing-library/react"
+import userEvent from "@testing-library/user-event"
+import { describe, expect, it, vi } from "vitest"
+
+import { ListDetail, ListDetailRow } from "./ListDetail"
+
+describe("ListDetail", () => {
+  it("names both columns, so neither half is anonymous", () => {
+    render(
+      <ListDetail
+        listLabel="Policies"
+        list={null}
+        detail={<p>Nothing open.</p>}
+        detailLabel="Policy detail"
+        isDetailShown={false}
+        onShowList={() => {}}
+      />,
+    )
+
+    expect(screen.getByRole("region", { name: "Policies" })).toContainElement(
+      screen.getByRole("heading", { name: "Policies" }),
+    )
+    expect(
+      screen.getByRole("region", { name: "Policy detail" }),
+    ).toHaveTextContent("Nothing open.")
+  })
+
+  it("puts the create control in the column it adds to", () => {
+    render(
+      <ListDetail
+        listLabel="Policies"
+        listAction={<button type="button">Create policy</button>}
+        list={null}
+        detail={null}
+        isDetailShown={false}
+        onShowList={() => {}}
+      />,
+    )
+
+    expect(screen.getByRole("region", { name: "Policies" })).toContainElement(
+      screen.getByRole("button", { name: "Create policy" }),
+    )
+  })
+
+  it("offers the way back only while a column is hidden", async () => {
+    // The control is `md:hidden`, and jsdom applies no media query, so what is
+    // asserted here is the other half: it exists while the detail column is
+    // the one on screen and not while both are.
+    const onShowList = vi.fn()
+    const { rerender } = render(
+      <ListDetail
+        listLabel="Policies"
+        list={null}
+        detail={<p>fast</p>}
+        isDetailShown
+        onShowList={onShowList}
+        backLabel="All policies"
+      />,
+    )
+
+    await userEvent.click(screen.getByRole("button", { name: "All policies" }))
+    expect(onShowList).toHaveBeenCalledOnce()
+
+    rerender(
+      <ListDetail
+        listLabel="Policies"
+        list={null}
+        detail={<p>fast</p>}
+        isDetailShown={false}
+        onShowList={onShowList}
+      />,
+    )
+    expect(
+      screen.queryByRole("button", { name: /All policies|Back to the list/ }),
+    ).not.toBeInTheDocument()
+  })
+
+  it("makes the empty column a real button, reachable by keyboard", async () => {
+    // The whole column is the target, which a div with an onClick would also
+    // manage while being unreachable without a pointer.
+    const onEmptyPress = vi.fn()
+    render(
+      <ListDetail
+        listLabel="Policies"
+        list={null}
+        empty="Create your first policy"
+        onEmptyPress={onEmptyPress}
+        detail={null}
+        isDetailShown={false}
+        onShowList={() => {}}
+      />,
+    )
+
+    await userEvent.tab()
+    await userEvent.keyboard("{Enter}")
+    expect(onEmptyPress).toHaveBeenCalledOnce()
+    expect(
+      screen.getByRole("button", { name: "Create your first policy" }),
+    ).toBeInTheDocument()
+  })
+
+  it("states the empty column without offering it to a reader who cannot write", () => {
+    render(
+      <ListDetail
+        listLabel="Policies"
+        list={null}
+        empty="No policies yet."
+        detail={null}
+        isDetailShown={false}
+        onShowList={() => {}}
+      />,
+    )
+
+    expect(screen.getByText("No policies yet.")).toBeInTheDocument()
+    expect(screen.queryByRole("button")).not.toBeInTheDocument()
+  })
+
+  it("shows the list when nothing is empty about it", () => {
+    render(
+      <ListDetail
+        listLabel="Policies"
+        list={<p>fast</p>}
+        detail={null}
+        isDetailShown={false}
+        onShowList={() => {}}
+      />,
+    )
+
+    expect(screen.getByText("fast")).toBeInTheDocument()
+  })
+})
+
+describe("ListDetailRow", () => {
+  it("presses its label to open the record, not the whole row", async () => {
+    const onSelect = vi.fn()
+    render(
+      <ListDetailRow label="fast" isSelected={false} onSelect={onSelect}>
+        openai:gpt-5-mini
+      </ListDetailRow>,
+    )
+
+    await userEvent.click(screen.getByRole("button", { name: /fast/ }))
+    expect(onSelect).toHaveBeenCalledOnce()
+    expect(screen.getByText("openai:gpt-5-mini")).toBeInTheDocument()
+  })
+
+  it("says which record is the one being shown", () => {
+    render(
+      <>
+        <ListDetailRow label="fast" isSelected onSelect={() => {}} />
+        <ListDetailRow label="cheap" isSelected={false} onSelect={() => {}} />
+      </>,
+    )
+
+    expect(screen.getByRole("button", { name: "fast" })).toHaveAttribute(
+      "aria-current",
+      "true",
+    )
+    expect(screen.getByRole("button", { name: "cheap" })).not.toHaveAttribute(
+      "aria-current",
+    )
+  })
+
+  it("keeps a row's actions out of the press target that opens it", async () => {
+    // Nested buttons are neither valid nor operable, so the label and the
+    // actions are siblings. Pressing an action must not also open the record.
+    const onSelect = vi.fn()
+    const onDelete = vi.fn()
+    render(
+      <ListDetailRow
+        label="cheap"
+        isSelected={false}
+        onSelect={onSelect}
+        actions={
+          <button type="button" onClick={onDelete}>
+            Delete cheap
+          </button>
+        }
+      />,
+    )
+
+    await userEvent.click(screen.getByRole("button", { name: "Delete cheap" }))
+    expect(onDelete).toHaveBeenCalledOnce()
+    expect(onSelect).not.toHaveBeenCalled()
+  })
+})
--- a/web/src/design-system/layout/ListDetail.tsx
+++ b/web/src/design-system/layout/ListDetail.tsx
@@ -1,0 +1,186 @@
+import type { ReactNode } from "react"
+import { FiChevronLeft } from "react-icons/fi"
+
+import { Button } from "../actions/Button"
+import { EmptyMessage } from "../feedback/EmptyMessage"
+
+/**
+ * A list of records beside the one that is open: the page shape for a set of
+ * things where reading one is most of the work.
+ *
+ * The ratio is 1:1.75 and lives here rather than at a call site, because it is
+ * the whole reason the frame exists. The left column holds a name and a line
+ * about it; the right one holds every fact a table would have spread across
+ * lanes, plus the controls that act on the record. Two columns at equal width
+ * give the reading half a page and the naming half more than it needs.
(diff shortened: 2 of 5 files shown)


web/src/design-system/layout/ListDetail.tsx
```diff
@@ -0,0 +1,186 @@
...
+      </button>
+    )
+
+  return (
+    <div className="grid border border-border md:grid-cols-[1fr_1.75fr]">
+      {/* Both halves are regions, so a reader can move between them rather
+          than through one to reach the other. */}
+      <section
+        aria-label={listLabel}
+        className={`min-w-0 flex-col ${isDetailShown ? "hidden md:flex" : "flex"}`}
```
jigjigjig, line 96: **Blocking: focus is lost on every mobile column swap.**

Below `md`, pressing a `ListDetailRow` gives this column `hidden` (`display:none`) while the pressed button still holds focus. I confirmed in Chromium that this drops `document.activeElement` to `<body>`: a keyboard, switch-control or screen-reader user is thrown to the top of the document at the moment they open a record, and again when they press Back. Nothing announces the swap either.

The argument for the component is that the ratio, the divider and the column swap live here rather than at each call site. Focus management is part of that swap and is the half that did not come along. On a change of `isDetailShown` under `MOBILE_QUERY` (`max-width: 767px`, already a constant in `app/AppShell.tsx`), move focus to the Back control when entering the detail column and to the previously open row when returning.

jsdom cannot catch this, since it applies no CSS: a probe of exactly this case passes there and fails in a browser.

web/src/design-system/layout/ListDetail.tsx
```diff
@@ -0,0 +1,186 @@
...
+  detailLabel?: string
+  /** Below `md`: whether the detail column is the one on screen. */
+  isDetailShown: boolean
+  /** Below `md`: the way back, since only one column is reachable at a time. */
+  onShowList: () => void
+  /** Names the list on that control, where "the list" is not what to call it. */
+  backLabel?: string
+}) {
+  const rows =
+    empty === undefined ? (
```
jigjigjig, line 72: `empty` doubles as the is-empty flag, so `empty={null}` silently blanks the list.

The check is `empty === undefined`. Verified with a probe: one real row plus `empty={null}` renders no row at all. The natural call-site idioms, `empty={rows.length === 0 ? "No policies yet." : null}` or `empty={error ?? null}`, therefore hide every record with no error anywhere.

Take an explicit `isEmpty: boolean` beside the `empty` node, or at minimum test `empty == null` so `null` behaves the way `undefined` does.

web/src/design-system/layout/ListDetail.tsx
```diff
@@ -0,0 +1,186 @@
...
+ * **The action lane is there whether or not a row has actions**, so the
+ * trailing controls form a vertical lane down the list instead of landing
+ * wherever their row's label ends. Its width is the touch floor, which is also
+ * the width of the one control it is sized for.
+ *
+ * `aria-current` rather than `aria-pressed`: this is the record being shown out
+ * of a set, not a control that stays down. The selected row wears the accent as
+ * a tint, which is the accent spent on selection the way an active nav row
+ * spends it, and keeps the page's ink: the accent's own ink is under AA on its
+ * tint.
```
jigjigjig, line 143: The selected-row rationale here is wrong on both halves, and the treatment is a third one.

An active nav row is `bg-surface-subtle text-foreground border-foreground` (`app/nav/rowStyles.ts:185`), a neutral fill plus an ink edge, documented there as deliberately not an accent tint. So the tint is not "the accent spent on selection the way an active nav row spends it".

The ink argument then rules out `--color-primary` without reaching `--color-primary-subtle-foreground`, which `globals.css:195-200` exists for precisely "chips and active nav items" on this tint. `nav/WorkspaceSwitcher` uses tint plus that ink; `features/models/ModelsPage.tsx:1520` uses tint plus default ink, which is what this matches.

Pick one and correct the comment. Contrast is not the issue: muted 13px on the tint measures 9.2:1 to 11.5:1 across both themes.

web/src/design-system/layout/ListDetail.tsx
```diff
@@ -0,0 +1,186 @@
...
+        className={`flex min-h-11 min-w-0 flex-1 flex-col justify-center gap-0.5 px-4 py-2 text-left transition-colors motion-reduce:transition-none ${
+          isSelected ? "" : "hover:bg-surface-alt"
+        }`}
+      >
+        <span className="truncate text-body">{label}</span>
+        {children === undefined ? null : (
+          <span className="truncate text-caption">{children}</span>
+        )}
+      </button>
+      <div className="flex min-w-11 shrink-0 items-center justify-end pr-2">
```
jigjigjig, line 181: The lane is 44px in an action-less row and 52px in one holding an `IconButton`: `min-w-11` is a floor, and the 44px control plus `pr-2` overruns it.

The docstring above claims its width is the touch floor, and `layout.md`'s own rule asks for a fixed-width slot, "even when the slot is empty in some rows". `w-11` with the padding inside the control makes every row's label end at the same x.

web/src/design-system/layout/ListDetail.tsx
```diff
@@ -0,0 +1,186 @@
...
+    empty === undefined ? (
+      // The frame divides its children, so a row draws no rule of its own and
+      // the last one leaves none hanging over the space below it.
+      <div className="flex flex-1 flex-col divide-y divide-border-subtle">
+        {list}
+      </div>
+    ) : onEmptyPress === undefined ? (
+      <EmptyMessage>{empty}</EmptyMessage>
+    ) : (
+      <button
```
jigjigjig, line 81: `EmptyMessage` renders a `<div>`, so this press target is a `<div>` inside a `<button>`, outside `<button>`'s phrasing-content model.

Tolerated by every browser, but this file cites validity as its reason for keeping row actions out of the label button, so relying on tolerance here is inconsistent.

web/src/design-system/layout/ListDetail.tsx
```diff
@@ -0,0 +1,186 @@
...
+  detailLabel = "Details",
+  isDetailShown,
+  onShowList,
+  backLabel = "Back to the list",
+}: {
+  /** What the left column holds, as its heading and as its region's name. */
+  listLabel: string
+  /** The control that adds a record, at the top of the column it adds to. */
+  listAction?: ReactNode
+  /** The rows, which are `ListDetailRow`s. */
```
jigjigjig, line 43: This one restates the prop. Every other docstring in the file states something the code cannot, which is what makes them worth reading.

web/src/design-system/layout/ListDetail.test.tsx
```diff
@@ -0,0 +1,186 @@
...
+        isDetailShown={false}
+        onShowList={() => {}}
+      />,
+    )
+
+    expect(screen.getByRole("region", { name: "Policies" })).toContainElement(
+      screen.getByRole("button", { name: "Create policy" }),
+    )
+  })
+
```
jigjigjig, line 44: This comment is the whole coverage story for the responsive behavior, which is the component's main claim.

The screenshot matrix covers pages at 390px; this component has no page, so the one-column-at-a-time swap is exercised at no width by anything. Combined with shipping ahead of a consumer, the riskiest behavior is the untested one. A Storybook interaction test on `DetailShown` would close it, and would also cover the focus finding on `ListDetail.tsx`.

jigjigjig (commented):
Post-merge review. One accessibility blocker, three should-fixes, three nits, all inline.

The PR's stated checks hold: `vitest src/design-system/layout src/styles` gives 3558 passing here, and the artifact gates are correctly owed nothing (no `src/gateway/`, no `schema.ts` or `routeTree.gen.ts`, no new dependency, and no page so no screenshot entry). Title is conventional, prose gate clean, the design-system layer imports nothing outside itself, and both `DESIGN.md` and `layout.md` rows are present.

The "ships without a consumer" exception is left alone: `layout.md` names it in the right place and it was a merged decision.

Reviewed by Claude Opus 5 via Claude Code, following `.github/skills/review`.

web/src/design-system/layout/ListDetail.tsx
khaledosman: Fixed in fe4dedb5 (#1072). Focus follows the swap now: the Back control on the way in, the row carrying `aria-current` on the way back, and the list section as the fallback when no record is current so focus never lands on `<body>`.

Not gated on the media query, though, and that is a deliberate departure from the suggestion. `MOBILE_QUERY` lives in `app/AppShell.tsx`, which this layer may not import, so reading it here would mean a second copy of the breakpoint that only one side would remember to change. The Back control's wrapper is already `md:hidden`, so from `md` up the focus call lands on a `display:none` element and does nothing. CSS stays the only place the breakpoint is written.

Three tests cover the mechanism, and they say in comments what jsdom cannot prove, which is your point about the probe: no CSS there, so the column is never actually hidden and the original bug is invisible. A browser-level check wants Storybook interaction tests, and this repo has none (only `addon-docs`, no `@storybook/test`, no play function anywhere), so that is its own PR rather than a rider on this.

web/src/design-system/layout/ListDetail.tsx
khaledosman: Fixed in fe4dedb5 (#1072). There is an explicit `isEmpty` now and `empty` is only the node, which is the first of your two options. A regression test covers the exact idiom you named: one real row plus `empty={null}` renders the row.

web/src/design-system/layout/ListDetail.tsx
khaledosman: Fixed in fe4dedb5 (#1072), by correcting the comment rather than the treatment, since the treatment already had a precedent: `bg-primary-subtle` with the page's own ink is what `features/models/ModelsPage` gives its selected row. The docstring now names that, drops the false analogy to the active nav row, and says why the accent's own ink is not used and why `--color-primary-subtle-foreground` is not the answer either (it is for the small text of a chip or nav item, not a row's label). Thanks for the contrast measurements; they are why this stayed a comment fix.

web/src/design-system/layout/ListDetail.tsx
khaledosman: Fixed in fe4dedb5 (#1072). `w-11` with no padding on the lane, so the control's own inset is the only inset and every row's label ends at the same x. The docstring said the width was the touch floor, which was the claim that was false; it now says fixed at it rather than floored at it, and why.

web/src/design-system/layout/ListDetail.tsx
khaledosman: Fixed in fe4dedb5 (#1072), by removing the button rather than changing what is inside it. The column-sized press target is gone and the action is a slot beside the message, the way `listAction` already is, so it is a named button. That also drops the other half of the problem, which you did not raise: the old target's accessible name was whatever prose happened to sit in it.

web/src/design-system/layout/ListDetail.tsx
khaledosman: Fixed in fe4dedb5 (#1072). It now says why `list` is a node rather than data plus a row renderer, which is the part the type cannot say.

web/src/design-system/layout/ListDetail.test.tsx
khaledosman: Addressed in fe4dedb5 (#1072), though not with an interaction test. Three tests now cover the swap, and each says what it does and does not prove: jsdom applies no CSS, so they pin the mechanism and the width gating stays a stylesheet concern.

On the interaction test itself: this repo has no infrastructure for one. `main.ts` lists only `addon-docs`, there is no `@storybook/test` dependency, and no story anywhere has a play function, so it would mean a new dependency plus addon config. AGENTS.md asks for tooling changes in their own PR, so I left it out and named the gap in the PR body instead.

PR #1055, opened 2026-09-10
docs: say when Escape closes the wrong thing in a dialog's combo box

jigjigjig:
## Description

A form inside a dialog gives one keystroke two meanings, and the habit that works everywhere else is the one that breaks.

Pressing Escape after picking an option is how a spec puts a react-aria popover away, because that popover hides the rest of the page from the queries that come next. Inside a dialog the option press has already closed the popover, so the key travels on to the dialog, which either closes it or, if the form has been typed into, swaps its footer for the unsaved-changes guard. What the run then reports is a missing submit button, several lines below the line that actually caused it.

Three pages have paid for this: keys, routing and providers. In each the Escape was vestigial once the option click closed the popover on its own, and deleting it was the whole fix. This adds a section to the testing guide so the fourth page does not rediscover it.

The rule it states is a state rather than a sequence: send Escape only while the box reports `aria-expanded="true"`. That is what `dismissComboBoxInDialog` checks, and it is worth saying explicitly that waiting for the attribute is not the same thing, because a popover that is already closed satisfies the wait instantly and the keystroke has landed by then.

Three facts travel with it. Blur cannot dismiss a box inside a dialog, since focus containment returns focus to it and these boxes open on focus, which is why the state check is the rule and not an optimization. The submit has to be queried through the dialog, because the label rule deliberately puts the same words on the page's trigger, the dialog's submit, and an empty state's call to action. And a Vitest flow needs one `userEvent` instance throughout: mixing a fresh `setup()` with the bare default silently drops the press, which reads exactly like a validation guard refusing to submit.

It also records the side a component can take, which `forms/MultiSelect` does: stopping the Escape it handles from propagating, so dismissing its list cannot arm the dialog's guard.

## How to test it locally

Documentation only. The behavior it describes is covered by the specs on the pages it names.

## PR Type

- [x] Documentation

## Relevant issues

None. Written from three failures in the #1031 series.

## Checklist

- [ ] I have added or updated tests that cover my change. Documentation only; the three pages it draws on carry the tests.
- [x] Documentation was updated where necessary. That is the change.

## AI Usage

**AI Model/Tool used:**

Claude Code (Opus 5)

The `userEvent` fact was added after the guidance was first written, because I hit it, misdiagnosed it as the browser's own validation pre-empting the app's message, and was one commit from shipping a weakened test with a convincing explanation on it. It is in the guide as the thing it looks like rather than the thing it is.

## Summary

- Added testing guidance for combo boxes inside dialogs.
- Documented when to use Escape, how to move focus, and how to scope dialog queries.
- Standardized `userEvent` usage in Vitest flows.
- Clarified behavior for different `menuTrigger` values and dialog Escape guards.

This helps prevent tests from closing or guarding the wrong element.

### DIFF
--- a/.github/skills/frontend-standards/testing.md
+++ b/.github/skills/frontend-standards/testing.md
@@ -162,6 +162,52 @@ done: `user.type` is one line and it is the library's own API. Revisit only if
 the suite gets slow enough to be worth the churn, and measure one file before
 committing to the rest.
 
+## A combo box inside a dialog: Escape is not the way to close its popover
+
+A form that lives in a `FormDialog` gives one keystroke two meanings, and the
+habit that works everywhere else is the one that breaks. `{Escape}` after
+picking an option is how a spec puts a react-aria popover away, because that
+popover `aria-hidden`s the rest of the page and hides the control the spec
+presses next. Inside a dialog the key travels on to the dialog, which either
+closes it or, where the form is dirty, swaps the footer for its unsaved-changes
+guard. The failure reads as a missing submit button, several lines below the
+line that caused it.
+
+This has cost a run on three pages: keys, routing and providers.
+
+**What to do after a pick depends on `menuTrigger`, and the default is the
+awkward one.** `ComboBoxField` defaults to `menuTrigger="focus"`, which
+`UserComboBox` and `providerFields` also set explicitly. Selecting an option
+hands focus back to the input, and a box that opens on focus reopens: the
+popover is open again, the page is `aria-hidden` again, and the dialog's submit
+is out of reach. So:
+
+- **A `menuTrigger="focus"` box needs focus moved off it**, by clicking a named
+  control inside the dialog: `await user.click(screen.getByLabelText("Name"))`
+  is what the keys dialog does. Not a bare `blur()`, which lands straight back
+  on the box because a modal contains focus, and reopens the popover for the
+  same reason.
+- **A `menuTrigger="input"` box needs nothing.** Its popover opens on typing,
+  so the pick leaves it closed and there is nothing for a keystroke to do.
+- **Where a spec has to dismiss without picking, send Escape only while the box
+  reports `aria-expanded="true"`**, and check the attribute rather than waiting
+  on it: a popover that is already closed satisfies a wait instantly, by which
+  point the keystroke has landed on the dialog.
+- **Query the submit through the dialog** (`within(dialog)` in Vitest,
+  `page.getByRole("dialog").getByRole(...)` in Playwright). The labels rule puts
+  the same words on the page trigger and on the dialog's submit, so an unscoped
+  query is ambiguous while the dialog is open, and ambiguous again on an empty
+  list where the empty state offers the same words a third time.
+- **In Vitest, use one `userEvent` instance for the whole flow.** Mixing a
+  fresh `userEvent.setup()` with the bare default drops the press: the submit
+  reports as enabled, inside the form, and nothing happens. It reads exactly
+  like a validation guard refusing, and it cost an afternoon being mistaken for
+  one.
+
+None of this applies to a combo box in a toolbar, which is what
+`e2e/helpers.ts`'s `dismissComboBox` was written for: it presses Escape and
+blurs unconditionally, and both are right outside a modal.
+
 ## Playwright: behavioral
 
 `pnpm --dir web run e2e` builds the bundle and boots a real gateway against a throwaway SQLite


.github/skills/frontend-standards/testing.md
```diff
@@ -162,6 +162,53 @@ done: `user.type` is one line and it is the library's own API. Revisit only if
...
+deleting it was the whole fix.
+
+- **Send Escape only while the box says it has a popover open**, which is
+  `aria-expanded="true"` on the box itself. That is the whole rule, and it is
+  what `e2e/helpers.ts`'s `dismissComboBoxInDialog` does; the toolbar helper
+  beside it, `dismissComboBox`, is the one that may press unconditionally.
+  Waiting for the attribute instead of checking it does not help: a popover that
+  is already closed satisfies the wait instantly, by which point the keystroke
+  has landed on the dialog.
+- **After clicking an option, that means press nothing.** The option press
```
khaledosman: Wrong for the default combo box, and this is the rule the section is built on. `ComboBoxField.tsx:73` defaults `menuTrigger = "focus"`, and `UserComboBox.tsx:120` and `providerFields.tsx:241` set it explicitly, so selecting an option hands focus back to the input and the popover **reopens**.

`b5837623` diagnosed exactly this on the keys dialog and deleted the comment that said what line 187 now says:

> `menuTrigger="focus"` survived the move of the combo box's selection handling into the shared `ComboBoxField`, so selecting an option hands focus back to the input and the popover reopens. React-aria marks the rest of the page `aria-hidden` while it is open, which is what put the dialog's submit out of reach.

So "press nothing" leaves the popover open and the submit unreachable, which is the failure this section exists to prevent. Split the rule by `menuTrigger`: an `"input"` box stays closed after a pick and needs nothing; a `"focus"` box needs focus moved to a named control. Line 177's "deleting it was the whole fix" needs the same correction.

.github/skills/frontend-standards/testing.md
```diff
@@ -162,6 +162,53 @@ done: `user.type` is one line and it is the library's own API. Revisit only if
...
+  closed the popover, so there is nothing left for Escape to do.
+- **Never `blur()` inside a dialog.** A modal contains focus, so it lands back
+  on the box, and these boxes open on focus, so the popover reopens. This is the
+  half that has no workaround: moving focus is not available to you here, which
```
khaledosman: "Moving focus is not available to you here" is false, and it is the sentence that talks a reader out of the fix that works. A bare `blur()` does land back on the box, but clicking a specific control inside the dialog does not: the landed keys test is `await usr.click(screen.getByLabelText("Name"))`.

Keep "never bare `blur()`" and replace the generalization with "move focus to a named control in the dialog".

.github/skills/frontend-standards/testing.md
```diff
@@ -162,6 +162,53 @@ done: `user.type` is one line and it is the library's own API. Revisit only if
...
+is dirty, swaps the footer for its unsaved-changes guard. The failure reads as a
+missing submit button, several lines below the line that caused it.
+
+This has cost a run on three pages: keys, routing and providers. In each, the
+Escape was vestigial once the option click closed the popover on its own, and
+deleting it was the whole fix.
+
+- **Send Escape only while the box says it has a popover open**, which is
+  `aria-expanded="true"` on the box itself. That is the whole rule, and it is
+  what `e2e/helpers.ts`'s `dismissComboBoxInDialog` does; the toolbar helper
```
khaledosman: `dismissComboBoxInDialog` does not exist at this head or on `main`; `git grep` finds it only in this doc. `web/e2e/helpers.ts:177` has only `dismissComboBox`, which presses Escape *and* blurs unconditionally. The helper is added by `353043aa` on the unmerged `feat/dashboard-*` stack.

This PR is based on `main` and can land first, so a reader following the rule has no helper to call. Either land the helper here or describe the `aria-expanded` check inline and drop the name.

.github/skills/frontend-standards/testing.md
```diff
@@ -162,6 +162,53 @@ done: `user.type` is one line and it is the library's own API. Revisit only if
...
+  the same words on the page trigger and on the dialog's submit, so an unscoped
+  query is ambiguous while the dialog is open, and ambiguous again on an empty
+  list where the empty state offers the same words a third time.
+- **In Vitest, use one `userEvent` instance for the whole flow.** Mixing a
+  fresh `userEvent.setup()` with the bare default drops the press: the submit
+  reports as enabled, inside the form, and nothing happens. It reads exactly
+  like a validation guard refusing, and it cost an afternoon being mistaken for
+  one.
+
+A component can take the other side of this. `forms/MultiSelect` calls
```
khaledosman: `forms/MultiSelect` does not exist at this head. `web/src/design-system/forms/` holds no `MultiSelect`, and the component plus its Escape `stopPropagation` arrive with `4eb27634` on the unmerged stack. The people picker here is `features/users/UserMultiSelect.tsx`, which handles no Escape at all.

Drop the paragraph or gate it on that component landing.

khaledosman (changes requested):
The section's central rule is the belief this stack later disproved, and two of its concrete references name code that is not in the tree this PR lands on.

`b5837623` (`test(dashboard): blur the owner picker rather than closing it with Escape`, on `feat/dashboard-keys-form-dialog`) was authored after this branch and says the opposite of lines 171 and 187: "`menuTrigger="focus"` survived the move of the combo box's selection handling into the shared `ComboBoxField`, so selecting an option hands focus back to the input and the popover reopens." It deleted the very comment this doc turns into a rule and replaced it with `await usr.click(screen.getByLabelText("Name"))`.

Prose gates are clean: no em dashes or `--` separators, US spelling throughout, no new fences, and the title is a valid conventional commit.

.github/skills/frontend-standards/testing.md
jigjigjig: You are right, and this was the load-bearing sentence. Fixed at 0cfb0b10.

I checked before rewriting rather than taking it on your word, and it holds all the way: `ComboBoxField.tsx:92` defaults `menuTrigger = "focus"`, `UserComboBox.tsx:100` and `providerFields.tsx:241` set it explicitly, and b5837623's landed fix is `await usr.click(screen.getByLabelText("Name"))`, which is moving focus rather than pressing nothing.

The section now splits on `menuTrigger` and says the default is the awkward one: a "focus" box needs focus moved off it to a named control in the dialog, an "input" box needs nothing because its popover opens on typing. "Deleting it was the whole fix" is gone with it; the sentence now just names the three pages that paid for this.

.github/skills/frontend-standards/testing.md
jigjigjig: Fixed at 0cfb0b10. "Moving focus is not available to you here" is gone.

The rule now reads as the positive one, with the counterexample attached: click a named control inside the dialog, `await user.click(screen.getByLabelText("Name"))` as the keys dialog does, and not a bare `blur()`, which lands back on the box because a modal contains focus. Worth having both halves in one bullet, since the bare-blur failure is what sends a reader looking for a different approach when the right one is two characters away.

.github/skills/frontend-standards/testing.md
jigjigjig: Fixed at 0cfb0b10 by describing the check and dropping the name.

You are right that this PR has to stand on its own: it is based on main and can land first, so naming a helper that arrives with an unmerged stack would send a reader to a function that is not there. The bullet now says to send Escape only while the box reports `aria-expanded="true"`, and to check the attribute rather than wait on it. `dismissComboBox` is still named at the end, correctly: it is the toolbar helper that exists on main, and the closing line now says what it does (presses and blurs unconditionally) and why both are right outside a modal.

.github/skills/frontend-standards/testing.md
jigjigjig: Fixed at 0cfb0b10, paragraph dropped.

Same reason as the helper: `forms/MultiSelect` arrives with the unmerged stack, so a reader at this head would go looking for a component that is not there. The component-side point is worth making once that lands, and I would rather it be made in the PR that adds the component than promised here.

PR #1061, opened 2026-09-10
feat(core): accept several bootstrap modules in OTARI_BOOTSTRAP

### LINKED ISSUE #1060: OTARI_BOOTSTRAP loads exactly one module, so two independent extensions cannot coexist
\`build_container\` accepts a single \`module:callable\` selector. That fits one overlay per deployment, which is what the seam was designed for, but not a deployment that installs two independent extensions (say, a budget-alerts plugin and a custom identity adapter from different authors): one of them has to wrap the other.

**Proposed shape**

- \`OTARI_BOOTSTRAP\` (and the \`bootstrap\` config field) accept a comma-separated list of selectors, applied in order after the core defaults are bound. A later bind wins, as it does today.
- A blank entry anywhere in the list is refused the way a blank selector is now.
- The container summary names every selector and what each rebound or contributed.
- Explicit and ordered, so it stays within the "no entry-point auto-discovery" rule in ARCHITECTURE.md: the whole wiring is still readable from one config value.

Refs #973.

daavoo:
## Description

Otari can be extended by pointing `OTARI_BOOTSTRAP` at a trusted module that rebinds ports and adds routes at startup. Until now that setting took exactly one module, which fits one overlay per deployment but not a deployment that installs two independent extensions from different authors: one of them had to wrap the other.

`OTARI_BOOTSTRAP` (and the `bootstrap` config field) now accepts a comma-separated list of `module:callable` selectors, applied in the order written after the core defaults are bound. A later bind wins, as it did already. A blank entry anywhere in the list, a trailing comma included, refuses to start and names the position, and a second selector that cannot be loaded fails startup rather than quietly running with only the first. The startup summary names each selector with the ports it rebound and the routers it contributed.

The list is explicit and ordered, so the whole wiring is still readable from one config value and nothing is discovered from installed packages, which keeps the no-auto-discovery rule in ARCHITECTURE.md intact. A single selector behaves exactly as before, including every error message.

This is one of the four seams needed so that budget alerts (#973) can ship as a plugin rather than in core. Refs #973.

## How to test it locally

Automated:

- `make lint`, `make typecheck`, `make test-unit`: the new cases live in `tests/unit/test_container.py` (two selectors applied in order with a later bind winning and both routers recorded, blank middle entry, trailing comma, leading comma, and a failing second selector).
- `uv run --frozen --no-dev python scripts/oss_edition_smoke.py`: boots with no bootstrap at all, confirming the plain build is untouched.

The integration suite (`make test-integration`, which includes `tests/integration/test_bootstrap_overlay.py`) was **not** run locally: the machine this was developed on has neither Docker nor a PostgreSQL server, so Testcontainers cannot start. CI runs it on this PR.

By hand, from a checkout:

```bash
mkdir -p /tmp/boot && cat > /tmp/boot/one.py <<'EOF'
from gateway.ports.billing_port import BillingPort
class OneBilling:
    def __init__(self, session): ...
def register(container):
    container.bind(BillingPort, OneBilling)
EOF
cat > /tmp/boot/two.py <<'EOF'
from gateway.ports.entitlement_port import EntitlementPort
class TwoEntitlements:
    def __init__(self, session): ...
def register(container):
    container.bind(EntitlementPort, TwoEntitlements)
EOF
PYTHONPATH=/tmp/boot OTARI_BOOTSTRAP="one:register, two:register" uv run otari serve
```

The startup log reads `Composition root: one:register rebound BillingPort; two:register rebound EntitlementPort`. Then try `OTARI_BOOTSTRAP="one:register,"`: the gateway refuses to start with `OTARI_BOOTSTRAP entry 2 of 2 is blank; remove the extra comma`.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [x] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes #1060
Refs #973

## Checklist

- [x] I ran the Definition of Done checks locally (`make lint`, `make typecheck`, `make test`). `make test-unit` and the smoke gate ran locally; `make test-integration` could not (no Docker or PostgreSQL on the machine) and is left to CI.
- [x] If the API contract changed, I regenerated the OpenAPI spec (`uv run python scripts/generate_openapi.py`). Not applicable: no route or schema changed.

## AI Usage

- [ ] No AI was used.

**Any additional AI details you'd like to share:** Implementation, tests, docs and this description were written by the agent following the repo's `pr-cycle`, `backend-standards` and `review` skills. A human requested the change and set its scope.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -158,6 +158,8 @@ def register(container: Container) -> None:
 
 With nothing configured nothing is imported, the defaults stand, and Otari boots standalone. A selector that is set but cannot be loaded fails startup rather than quietly falling back, because a build nobody chose is worse than a gateway that will not start.
 
+Several independent extensions (say, a budget-alerts plugin and a custom identity adapter from different authors) coexist without one wrapping the other: `OTARI_BOOTSTRAP` accepts a comma-separated list of selectors, applied in that order after the core defaults, and a later bind wins. This stays inside the no-auto-discovery rule above, because the list is explicit and ordered: the whole wiring is still readable from one config value, and a blank entry in it fails startup the way a blank selector does.
+
 A contributed router is the additive half of the seam, and it is gated rather than swapped: Otari mounts it behind `require_capability(...)`, which resolves `EntitlementPort` and answers a request for an unentitled capability with the same 404 a path nothing serves gets. That gate is the server-side half of the entitlement axis; hiding a nav item in the dashboard is not authorization.
 
 Entitlement is not authentication either, and the mount point adds none. A capability names no caller, so on an entitled deployment a contributed route is reachable by anyone unless the router says otherwise. A contribution declares the credential each of its routes needs on the route, the way Otari's own routers do; there is no router-level default to mount, because the right answer differs per route, a contributed route may be deliberately public, and the header check resolves a database session a hybrid gateway does not have.
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -309,6 +309,21 @@ usage-report timeout and retries, and first-chunk fallback timeout. See
 inside the gateway process. The callable can rebind extension ports and
 contribute capability-gated routers. Most deployments should leave it unset.
 
+Several independent extensions coexist by listing their selectors in that one
+value, comma-separated, in the order they should apply:
+
+```bash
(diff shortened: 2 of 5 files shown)


daavoo:

PR #1062, opened 2026-09-10
feat(dashboard): move the domain claim form into FormDialog

jigjigjig:
## Description

Claiming an email domain used to append a form section under the page heading, which is one of the two creation patterns the dashboard is collapsing into one. It now opens the shared form dialog: press "Claim domain" in the heading row and a small centered dialog titled "New domain" appears, with "Claim domain" as its submit.

Two behaviors change beyond the surface. The heading button stays on screen while the dialog is open, because the dialog sits over the page rather than replacing the action, so the control never vanishes from under the pointer. And a half-typed draft is cleared when the dialog opens rather than when it closes: clearing on the way out would blank the fields while the dialog is still animating away.

The "They join as" picker moves from the toolbar filter control to the form one. The toolbar control draws a label of its own, so beside a labeled field inside a dialog it would ship two label rungs.

Part of #1031

## How to test it locally

1. `make dashboard && make dev`, sign in, and open Organization, Email domains as an owner or admin.
2. Press "Claim domain" in the heading row. The dialog opens over the page and the heading button is still there.
3. Type a domain, press Escape, and discard the unsaved changes. Press "Claim domain" again: the field is empty.
4. Claim a domain and confirm the row and its pending-proof card appear as before.

Covered automatically by `web/src/features/organization/OrganizationDomainsPage.test.tsx` (14 cases, including the two above). Both new cases were checked by removing the fix and watching them fail: dropping the remount key makes the draft survive, and hiding the trigger while the dialog is open fails the visibility case.

## PR Type

- [x] Refactor

## Relevant issues

Part of #1031

## Checklist

  Dashboard counterparts: `pnpm --dir web run lint`, `run typecheck`, and the vitest suite, all green. The Playwright behavioral projects (onboarding, seed, parity, hybrid) were run before and after this change and produce the same result: one pre-existing failure, `parity.management.spec.ts` budgets, which is unrelated to this page and fails identically on the base commit.
- [ ] Documentation was updated where necessary. No documentation covers this form's placement; the design system's dialog guidance already describes the target pattern.

## AI Usage

**AI Model/Tool used:**

Claude Code (Opus 5)

The reset-on-open behavior and the two new tests were checked by deliberately reintroducing each defect and reading the failure, rather than by the tests passing on first write.

### DIFF
--- a/web/e2e/screenshots/authenticated.spec.ts
+++ b/web/e2e/screenshots/authenticated.spec.ts
@@ -475,7 +475,7 @@ test.describe("the domain dialog", () => {
     await page.getByRole("button", { name: "Claim domain" }).first().click()
     const dialog = page.getByRole("dialog", { name: "New domain" })
     await expect(dialog.getByLabel(/^Domain/)).toBeVisible()
-    await captureScreenshot(page, "domains-claim-dialog")
+    await captureScreenshot(page, "organization-domains-claim-dialog")
   })
 })
 
--- a/web/e2e/screenshots/authenticated.spec.ts
+++ b/web/e2e/screenshots/authenticated.spec.ts
@@ -464,6 +464,21 @@ test.describe("the provider dialog", () => {
   })
 })
 
+test.describe("the domain dialog", () => {
+  // The suite's only `sm` dialog, which is the width nothing else here covers.
+  test("the claim dialog", async ({ page }) => {
+    await login(page)
+    await gotoRoute(page, "/organization/domains")
+    await expect(
+      page.getByRole("heading", { name: /email domains/i }).first(),
+    ).toBeVisible()
+    await page.getByRole("button", { name: "Claim domain" }).first().click()
+    const dialog = page.getByRole("dialog", { name: "New domain" })
+    await expect(dialog.getByLabel(/^Domain/)).toBeVisible()
+    await captureScreenshot(page, "domains-claim-dialog")
+  })
+})
+
 // No capture for the organization provider-key dialog. Its page is gated on the
 // `organization_providers` surface, which only a hosted deployment publishes
 // (bootstrap.py's HOSTED_SURFACES), and this suite boots a standalone gateway,
--- a/web/src/features/organization/OrganizationDomainsPage.test.tsx
+++ b/web/src/features/organization/OrganizationDomainsPage.test.tsx
@@ -220,6 +220,40 @@ describe("OrganizationDomainsPage", () => {
     })
   })
 
+  it("keeps the header trigger on screen while the dialog is open", async () => {
+    // The dialog sits over the page rather than replacing the action, so the
+    // control that opened it does not vanish from under the pointer.
+    mockApi()
+    const user = userEvent.setup()
+    renderPage(<OrganizationDomainsPage />)
+
+    const trigger = await screen.findByRole("button", { name: "Claim domain" })
+    await user.click(trigger)
+
+    expect(await screen.findByRole("dialog")).toBeInTheDocument()
+    expect(trigger).toBeVisible()
+  })
+
+  it("opens on a blank draft after a close, not on the last one typed", async () => {
+    // Reset on the way in: clearing on the way out would blank the fields
+    // while the dialog is still animating away.
+    mockApi()
+    const user = userEvent.setup()
+    renderPage(<OrganizationDomainsPage />)
+
+    await user.click(
+      await screen.findByRole("button", { name: "Claim domain" }),
+    )
+    await user.type(screen.getByLabelText(/Domain/), "acme.example")
+    // A draft this far along is dirty, so the way out is through the guard.
+    await user.keyboard("{Escape}")
+    await user.click(screen.getByRole("button", { name: "Discard" }))
+    await waitFor(() => expect(screen.queryByRole("dialog")).toBeNull())
+
+    await user.click(screen.getByRole("button", { name: "Claim domain" }))
+    expect(await screen.findByLabelText(/Domain/)).toHaveValue("")
+  })
+
   it("never offers a management role, because a DNS record must not mint admins", async () => {
     mockApi()
     renderPage(<OrganizationDomainsPage />)
--- a/web/src/features/organization/OrganizationDomainsPage.test.tsx
+++ b/web/src/features/organization/OrganizationDomainsPage.test.tsx
@@ -281,6 +281,18 @@ describe("OrganizationDomainsPage", () => {
     expect(
       screen.getByRole("button", { name: /They join as/ }),
     ).toHaveTextContent("Viewer")
+
+    // And the seed is per open, not per page: the ref lives below the key, so
+    // the next open starts clean. Held above it, or with the key dropped, the
+    // dialog would arrive already dirty and Escape would ask before closing an
+    // untouched form.
+    await user.click(screen.getByRole("button", { name: "Cancel" }))
+    await user.click(screen.getByRole("button", { name: "Discard" }))
+    await waitFor(() => expect(screen.queryByRole("dialog")).toBeNull())
+    await user.click(screen.getByRole("button", { name: "Claim domain" }))
+    await screen.findByRole("dialog")
+    await user.keyboard("{Escape}")
+    await waitFor(() => expect(screen.queryByRole("dialog")).toBeNull())
   })
 
   it("never offers a management role, because a DNS record must not mint admins", async () => {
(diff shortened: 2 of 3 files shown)


web/src/features/organization/OrganizationDomainsPage.tsx
```diff
@@ -79,21 +86,24 @@ function ClaimForm({ onClose }: { onClose: () => void }) {
...
+      onOpenChange={(open) => {
+        if (!open) onClose()
+      }}
+      size="sm"
+      title="New domain"
+      submitLabel="Claim domain"
+      onSubmit={submit}
+      isPending={create.isPending}
+      isSubmitDisabled={domain.trim() === ""}
+      isDirty={domain.trim() !== ""}
```
khaledosman: `isDirty` omits `role`. Open the dialog, change "They join as" to Viewer, press Escape: the choice is discarded with no guard. `domain.trim() !== "" || role !== "member"`.

web/e2e/screenshots/authenticated.spec.ts
```diff
@@ -331,3 +331,141 @@ test.describe("organization admin", () => {
...
+    await login(page)
+    await gotoRoute(page, "/providers")
+    await expect(
+      page.getByRole("heading", { name: /providers/i }).first(),
+    ).toBeVisible()
+    await page.getByRole("button", { name: "Add provider" }).first().click()
+    const dialog = page.getByRole("dialog", { name: "New provider" })
+    await dialog.getByRole("button", { name: "Custom endpoint" }).click()
+    await expect(dialog.getByLabel("API base")).toBeVisible()
+    await captureScreenshot(page, "providers-add-dialog-custom")
```
khaledosman, line 463: **No capture for the claim dialog, and no note saying why.** Every other conversion in this stack added one — `keys-create-dialog`, `routing-create-dialog`, `budgets-create-dialog`, `workspaces-create-dialog`, `providers-add-dialog` — and unlike #1048's case this page is reachable from this harness: it is gated on `organizations`, which is in `STANDALONE_SURFACES`, and the suite already captures the page at rest (`organization-domains`, `:42`).

It would also be the suite's first `sm` (440px) dialog, which is the one size nothing else covers. A `test.describe("the domain dialog")` that logs in, goes to `/organization/domains`, presses "Claim domain", waits on `dialog.getByLabel(/Domain/)` and captures `domains-claim-dialog`.

khaledosman (approved):
Gates clean: no route or schema touched; no generated-artifact drift; no dependency change; conventional title; no em dashes or `--` separators.

This is the reference implementation of the seven conversions. `key={openCount}` (`:322`) bumped at the one and only trigger (`:293`), `close()` reduced to `setAdding(false)` (`:324`), `useCreateOrganizationDomain()` held inside `ClaimForm` (`:79`) below the key so a remount resets the mutation, `autoFocus` on the first field, `reserveMessage={false}` on the description-less `Select`, `isSubmitDisabled` free of `isPending`, and a colocated test for the fresh draft. #1045 and #1048 each miss the first of those; this one is what they should look like.

Two small things below.

---

web/src/features/organization/OrganizationDomainsPage.tsx
jigjigjig: Fixed in 12b84806, though as a snapshot rather than the longer predicate: `const draft = JSON.stringify({ domain, role })` against a `useRef` seeded on mount. Same shape as the workspace and provider-key forms in this series, and the reason is this thread: a hand-listed predicate is one edit behind the form, and `role` is what it was behind.

Test: pick Viewer with nothing typed, dismiss, and the guard has to appear; Keep editing returns to the choice. It fails with the old `domain.trim() !== ""` restored.

One thing found while writing it, and reported as a jsdom observation rather than a claim about the product: after picking from that `Select`, `document.activeElement` is `<body>`, so a keyboard Escape reaches nothing and the case has to dismiss through Cancel. If react-aria restores focus to the trigger in a real browser this is only a test detail; if it does not, it is a focus defect worth its own issue. Not verified in a browser here.

web/e2e/screenshots/authenticated.spec.ts
jigjigjig: Added in 12b84806: a `test.describe("the domain dialog")` that logs in, goes to `/organization/domains`, presses the header's "Claim domain", waits on the dialog's Domain field and captures `domains-claim-dialog`.

Stated plainly: the capture itself has not been run here. The suite is workflow-dispatch only with gitignored baselines and needs a built gateway, so what I have checked is the spec's typecheck and lint and that its selectors are the ones the unit tests resolve against the same markup (`New domain` as the dialog name, the Domain field's label). The `.first()` on the trigger matters, since the dialog's submit carries the same name.

web/e2e/screenshots/authenticated.spec.ts
```diff
@@ -331,3 +331,156 @@ test.describe("organization admin", () => {
...
+  test("the claim dialog", async ({ page }) => {
+    await login(page)
+    await gotoRoute(page, "/organization/domains")
+    await expect(
+      page.getByRole("heading", { name: /email domains/i }).first(),
+    ).toBeVisible()
+    await page.getByRole("button", { name: "Claim domain" }).first().click()
+    const dialog = page.getByRole("dialog", { name: "New domain" })
+    await expect(dialog.getByLabel(/^Domain/)).toBeVisible()
+    await captureScreenshot(page, "domains-claim-dialog")
```
khaledosman: The baseline name breaks the file's page-prefix convention. Every organization-rail capture is prefixed `organization-` (`organization-general`, `organization-members`, `organization-model-pricing`, `organization-usage`, `organization-spend-budgets`), including this page's own at `:41-45` (`organization-domains`). `domains-claim-dialog` sorts away from the page it belongs to in the artifact listing a reviewer reads. `organization-domains-claim-dialog`.

web/src/features/organization/OrganizationDomainsPage.test.tsx
```diff
@@ -220,6 +220,69 @@ describe("OrganizationDomainsPage", () => {
...
+    // A draft this far along is dirty, so the way out is through the guard.
+    await user.keyboard("{Escape}")
+    await user.click(screen.getByRole("button", { name: "Discard" }))
+    await waitFor(() => expect(screen.queryByRole("dialog")).toBeNull())
+
+    await user.click(screen.getByRole("button", { name: "Claim domain" }))
+    expect(await screen.findByLabelText(/Domain/)).toHaveValue("")
+  })
+
+  it("guards a changed role on the way out, with nothing typed", async () => {
```
khaledosman, line 257: The test proves `role` is in `isDirty`, but not that the ref reseeds per open, which is the half this stack has already got wrong once and the half a refactor is most likely to break: move `seededDraft` above the key, or drop `key={openCount}`, and this stays green while the dialog goes permanently dirty. Three more lines: Discard, reopen, press Escape, assert the dialog closes rather than showing "Unsaved changes".

khaledosman (approved):
Both fixed, and the dirty check is the one in the series I would point others at: `JSON.stringify({ domain, role })` against a ref that is seeded **per open** rather than per mount, because `key={openCount}` at `:328` is bumped at the sole trigger. That is the distinction `BudgetsPage` got wrong earlier in this stack. The test at `:257` picks a role with nothing typed, presses Cancel, and asserts the guard arms and Keep editing returns to "Viewer".

The screenshot block matches the others exactly, and I confirmed the selector resolves rather than assuming it: "Claim domain" is the `PageIntro` action at `:295-304`, `canEdit` is true for the harness's master-key identity, the empty state is a `DataTable` `emptyContent` string with no competing button, and "New domain" is the dialog title at `:105`.

Two small things below. Gates clean, CI green.

---

web/e2e/screenshots/authenticated.spec.ts
jigjigjig: Fixed in ecb1c297: `organization-domains-claim-dialog`. You are right that the artifact listing is the reader here, and a name that sorts away from the page it belongs to costs a reviewer the comparison they came for.

web/src/features/organization/OrganizationDomainsPage.test.tsx
jigjigjig: Right, and it is the half worth pinning: fixed in ecb1c297. The case now discards, reopens and presses Escape, which has to close an untouched form. Dropping `key={openCount}` fails it, along with the existing fresh-draft case, so the reseed is covered from both directions rather than assumed.

PR #1063, opened 2026-09-10
feat(core): let a bootstrap contribute lifespan background tasks

### LINKED ISSUE #1057: A bootstrap cannot contribute a lifespan background task
The composition root (\`src/gateway/container.py\`) lets a bootstrap module rebind ports and contribute routers, but the periodic workers Otari runs are a hand-kept list inside \`_create_lifespan\` in \`src/gateway/main.py\`. A feature that needs a timer (a budget alert evaluator, a sync job, a purge) can only get one by editing that file, which ARCHITECTURE.md rule 6 rules out for anything outside the core.

Surfaced while assessing whether #973 (budget alerts) could ship as a plugin rather than core: its evaluator is a lifespan refresher and has nowhere to register.

**Proposed shape**

- \`Container.contribute_background_task(BackgroundTaskContribution(name, start))\`, where \`start\` builds the coroutine the lifespan schedules. Names are unique per container.
- The lifespan starts contributed tasks after Otari's own refreshers, in every mode, and stops them under the same shared cancellation bound \`_stop_refreshers\` already applies, so a plugin task that ignores cancellation cannot hang shutdown.
- A task that dies is logged with the contribution's name, as a core refresher is, and never takes the process down.
- The container summary line names the contributed tasks, the way it names contributed routers.

Refs #973, #410.

daavoo:
## Description

A deployment that extends Otari with a bootstrap module could already swap an adapter or add API routes, but it could not run a periodic worker of its own: the timers Otari runs (cache refreshers, the budget hold sweeper) were a fixed list inside the gateway's startup code, and adding one meant editing Otari itself, which the architecture rules forbid for anything outside the core.

This adds a third thing a bootstrap can contribute: a named background task. Otari starts it beside its own refreshers when the process boots, in both standalone and hybrid mode, and shuts it down under the same bounded wait it already applies to its own workers. A contributed task that ignores cancellation is abandoned after that bound instead of holding shutdown open, and one that crashes is logged under its name rather than taking the gateway down. Two tasks registered under the same name are refused at startup, and the composition-root summary line names the tasks a build contributed, the way it already names contributed routers.

This is one of four seams needed so that budget alerts (#973) can ship as a plugin instead of core: its evaluator is a periodic worker and now has a place to register. Refs #973.

## How to test it locally

Automated coverage:

- `uv run pytest tests/unit/test_container.py tests/unit/test_gateway_lifespan_shutdown.py -q -p no:cacheprovider` covers the contribution being recorded in order, a duplicate name being refused, the summary naming the tasks, a contributed task being started with the config and cancelled at shutdown in both modes, a task that ignores cancellation being abandoned after the shared bound, and a task that raises being logged by name without breaking shutdown.
- `make lint`, `make typecheck`, `uv run pytest tests/unit -q -x -p no:cacheprovider` and the OSS smoke gate `uv run --frozen --no-dev python scripts/oss_edition_smoke.py` all pass locally.
- The integration suite (`make test-integration`) was **not** run locally: the machine this was written on has neither Docker nor a PostgreSQL server, so Testcontainers cannot start. CI runs it. No route, schema, model, or migration changed, so no generated artifact needed regenerating.

To see it by hand, write a bootstrap module on the gateway's `PYTHONPATH`:

```python
# probe_bootstrap.py
import asyncio
from gateway.container import BackgroundTaskContribution, Container
from gateway.log_config import logger

async def tick(config) -> None:
    while True:
        logger.info("probe tick")
        await asyncio.sleep(5)

def register(container: Container) -> None:
    container.contribute_background_task(BackgroundTaskContribution(name="probe", start=tick))
```

Run `OTARI_BOOTSTRAP=probe_bootstrap:register uv run otari serve` and check the startup log: the composition-root line ends in `contributed background tasks probe`, `probe tick` appears every five seconds, and Ctrl+C shuts the process down promptly. Register the same name twice and startup fails with `A background task named 'probe' is already contributed`.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [x] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes #1057
Refs #973

## Checklist

- [x] I ran the Definition of Done checks locally (`make lint`, `make typecheck`, `make test`). (`make test-unit` and the smoke gate locally; the integration half runs in CI, see above.)
- [ ] If the API contract changed, I regenerated the OpenAPI spec (`uv run python scripts/generate_openapi.py`). (Not applicable: no route or schema changed.)

## AI Usage

- [ ] No AI was used.

**Any additional AI details you'd like to share:** Implemented, tested, and self-reviewed by the agent from the issue's proposed shape. A human should confirm the seam matches what #973 needs before merging.

**NOTE:**

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -154,6 +154,7 @@ An overlay (or your own deployment) rebinds ports to its own adapters **without 
 def register(container: Container) -> None:
     container.bind(BillingPort, _wallet_billing_adapter)
     container.contribute_router(RouterContribution(capability="billing", router=wallet_router))
+    container.contribute_background_task(BackgroundTaskContribution(name="budget alerts", start=run_alert_evaluator))
 ```
 
 With nothing configured nothing is imported, the defaults stand, and Otari boots standalone. A selector that is set but cannot be loaded fails startup rather than quietly falling back, because a build nobody chose is worse than a gateway that will not start.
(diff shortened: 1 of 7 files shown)


daavoo:

PR #1065, opened 2026-09-10
feat(core): grant and publish the capabilities a bootstrap contributes

### LINKED ISSUE #1059: A contributed router's capability is never granted in the base build, and the browser cannot learn it exists
\`_register_contributed_routers\` mounts every contributed router behind \`require_capability(contribution.capability)\`, and \`BaseEntitlementAdapter\` grants \`BASE_CAPABILITIES\`, which is empty. So in the base build every contributed route answers 404 unless the bootstrap also rebinds \`EntitlementPort\`, just to unlock its own surface. That conflates two axes ARCHITECTURE.md keeps apart: whether a capability is *installed* in this process, and whether the deployment is *entitled* to it.

On the browser side, \`GET /v1/bootstrap\` publishes \`surfaces\` (the topology axis) but nothing about capabilities, and ARCHITECTURE.md records that "no endpoint serves entitlements to the browser". A dashboard therefore has no way to learn that a plugin is present.

Surfaced while assessing whether #973 (budget alerts) could ship as a plugin.

**Proposed shape**

- The base entitlement adapter grants \`BASE_CAPABILITIES\` plus every capability the container's contributions declare. An overlay that binds a real resolver is unaffected: it replaces the adapter.
- \`GET /v1/bootstrap\` gains a \`capabilities\` field: the capabilities this deployment resolves as entitled, sorted, so the browser's entitlement axis can be answered by the server rather than a constant.
- Regenerate the OpenAPI spec, the Postman collection and \`web/src/client/schema.ts\`. Consuming the field in \`useEntitlements\` can follow separately.

Refs #973.

daavoo:
## Description

A deployment can be extended by a bootstrap that contributes routers for a capability of its own. Until now those routes answered 404 in the very build that mounted them, unless the bootstrap also replaced the gateway's entitlement resolver just to unlock its own surface. That mixed up two different questions: whether a capability is installed in this process, and whether the deployment is licensed for it. The base build now grants every capability it has actually installed, on top of the (currently empty) base set. A deployment running a real entitlement resolver is unaffected, because that resolver replaces the base one entirely.

The dashboard also had no way to learn that a plugin was present: the bootstrap payload it reads first published which management surfaces the deployment hosts, but nothing about capabilities. It now carries a sorted `capabilities` list with what the server resolves as entitled, so the browser's entitlement axis can be answered by the server rather than by a constant. Consuming it in the dashboard's `useEntitlements` hook is left as a follow-up; this PR only publishes the field and gives it a safe default in the normalizer for a gateway too old to send it.

This is one of the four seams needed for budget alerts (#973) to ship as a plugin rather than as core code.

Fixes #1059. Refs #973.

## How to test it locally

Automated:

- `make lint`, `make typecheck`, `make test-unit` (3169 passed, 10 skipped).
- `make openapi-check` and `make postman-check` pass on the regenerated spec, collection and `web/src/client/schema.ts`.
- `pnpm --dir web run lint`, `pnpm --dir web run typecheck`, `pnpm --dir web test` (5376 passed).
- `uv run --frozen --no-dev python scripts/oss_edition_smoke.py` passes.
- New unit tests: `tests/unit/test_container.py` (the contributed set, the base adapter granting it, and the set being read at resolve time rather than bind time) and `tests/unit/test_deployment_bootstrap.py` (the field is present and empty in the base build, lists a bootstrap's contributed capabilities sorted, and degrades to `[]` on a resolver's database failure). `web/src/shared/helpers/bootstrap.test.ts` covers the normalizer default.
- The integration suite (`tests/integration/test_bootstrap_overlay.py` in particular) was **not** run locally: this machine has neither Docker nor PostgreSQL. CI runs it. The overlay there rebinds `EntitlementPort` itself, so its "unentitled contributed route is 404" case is unaffected by design; no existing test asserts a contributed-but-unentitled route in the *base* build (the base-build 404 test asserts that no router was contributed at all), so none needed updating.

By hand:

1. `uv run otari` (any mode) and `curl -s localhost:8000/api/v1/bootstrap | jq .capabilities`: `[]`.
2. Write a module `probe.py` that calls `container.contribute_router(RouterContribution(capability="probe", router=<an APIRouter with a GET /probe>))` in a `register(container)` function, start with `OTARI_BOOTSTRAP=probe:register`, and repeat: `["probe"]`, and `GET /api/v1/probe` answers 200 rather than 404.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [x] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes #1059

## Checklist

- [x] I ran the Definition of Done checks locally (`make lint`, `make typecheck`, `make test`). Integration tests excepted, see above.

## AI Usage

- [ ] No AI was used.

**Any additional AI details you'd like to share:** Implemented, tested and self-reviewed by the agent from the issue text; a human scoped the change.


daavoo:

Issue #921, opened 2026-09-03
Otari's Alembic chain has no version table of its own


khaledosman:
`alembic/env.py` stamps the default `alembic_version`, the same table every other Alembic application stamps, so
nothing in a database establishes that the history in it is otari's.

#919 makes the resulting failure legible: a database stamped with a revision this chain does not own is refused by
name rather than reported as `Can't locate revision identified by ...`. It cannot do more than that. The check reads
that one table, so a neighbor keeping its history under a custom `version_table`, or in `django_migrations` or
`flyway_schema_history`, leaves nothing to read. Reproduced in review on #919: `alembic upgrade head` against a
Django database exits 0 and writes otari's whole schema alongside the neighbor's tables.

Detection cannot close that gap. A positive signal, refusing when the database has tables and none of them are
otari's, would refuse a legitimate first migration: an edition's chain may populate the same database first, which
is what `mozilla-ai/otari-ai` does today on `overlay_alembic_version`.

A version table of otari's own delivers the property by construction, and is the only thing that does.

**Scope to settle when picked up**

- The name. `otari_alembic_version` would match the pattern the downstream edition already chose, so two chains
  sharing a control-plane database would each be self-describing.
- How a deployment that exists moves onto it: the new table seeded from `alembic_version` at first boot, or a
  revision that seeds it, and whether the old row is left in place or removed.
- The rollback shape, which is the part that needs deciding rather than implementing. An image predating the seed
  reads an empty table of its own and tries to migrate from base, which is a worse failure than the one this
  replaces unless the old row stays readable to it.

Refs #919.

_Filed by Claude Opus 5 (1M context) via Claude Code at @khaledosman's request._

PR #1066, opened 2026-09-10
feat(core): let a bootstrap contribute an Alembic migration chain

### LINKED ISSUE #1058: A bootstrap cannot contribute an Alembic migration chain
\`init_db\` (\`src/gateway/core/database.py\`) upgrades one hardcoded script directory (\`alembic/\`) to \`head\`, and \`alembic/env.py\` targets \`Base.metadata\`. A module that a bootstrap loads has no way to own tables of its own: its revisions cannot join Otari's chain without editing the repo, and running a second \`alembic upgrade\` against the same database would fight over the shared \`alembic_version\` row.

Surfaced while assessing whether #973 (budget alerts, two new tables) could ship as a plugin rather than core.

**Proposed shape**

- \`Container.contribute_migrations(MigrationContribution(name, script_location, version_table))\`. The version table must not be \`alembic_version\` and must be unique among contributions; a collision is refused at contribution time.
- \`init_db\` runs Otari's own chain first, then each contribution's chain to \`head\` on the same URL, each stamping its own version table so the histories never interleave. This is the same shape the downstream edition already uses (\`overlay_alembic_version\`, see #921).
- A contributed chain must not reference core tables by foreign key in a way that would block a core migration; document the contract.
- Hybrid mode skips \`init_db\` and therefore contributed chains too, which is correct: there is no local database.
- Whether \`otari migrate\` also runs contributed chains, or only startup auto-migration does, is part of the scope to settle.

Refs #921, #973.

daavoo:
## Description

A module loaded through `OTARI_BOOTSTRAP` can now own database tables of its own. Until now, a bootstrap could rebind ports and add routers, but its tables had nowhere to go: its migrations could not join Otari's chain without editing this repo, and running a second Alembic upgrade against the same database would have fought Otari's over the shared `alembic_version` row.

A bootstrap now records a migration contribution: a name, its own Alembic script directory, and its own version table. With auto-migration on, startup upgrades Otari's own chain first and then each contributed chain, on the same database, each stamping only the version table it named, so the two histories never interleave. Claiming `alembic_version`, or a version table or name another contribution already holds, is refused while the container is built, before anything touches the database. Hybrid mode skips database initialization and therefore contributed chains too, which is correct: it has no local database.

This is one of four seams needed for budget alerts (#973, two new tables) to ship as a plugin rather than in core. Refs #973.

The contract for the contributed `env.py` (read `config.attributes["version_table"]` and pass it to `context.configure`) and a caution against foreign keys into core tables are documented in `docs/configuration.md` and `ARCHITECTURE.md`.

**Known gap, left to settle:** `otari migrate` still runs the core chain only, as a subprocess. Contributed chains run on startup auto-migration only. A deployment that migrates with the CLI has to run each contributed chain itself for now; the docs say so.

Core keeps stamping `alembic_version` (#921 is not addressed here). Refs #921.

## How to test it locally

Automated:

- `uv run pytest tests/unit/test_container.py tests/unit/test_contributed_migrations.py -q`: the container refuses the reserved and duplicate version tables and names contributed chains in its summary; `init_db` on a throwaway SQLite file against the fixture chain under `tests/fixtures/plugin_alembic/` leaves the core schema, a `plugin_demo` table, the contributed head in `plugin_demo_alembic_version`, and only core's head in `alembic_version`. A second boot is idempotent, and `auto_migrate=False` runs nothing.
- `make lint`, `make typecheck`, `uv run pytest tests/unit -q` (3173 passed locally) and the OSS-edition smoke gate `uv run --frozen --no-dev python scripts/oss_edition_smoke.py` all pass locally.
- The integration suite (`make test-integration`) was **not** run locally: the machine this was developed on has no Docker and no PostgreSQL, so it cannot start. CI runs it.

By hand, if you want to see it in a running gateway: write a module with a `register(container)` that calls `container.contribute_migrations(MigrationContribution(name=..., script_location=..., version_table=...))` pointing at `tests/fixtures/plugin_alembic`, put it on `PYTHONPATH`, set `OTARI_BOOTSTRAP=module:register`, start `otari serve` against a fresh SQLite file, and check that the startup log names the chain and that the database holds `plugin_demo` and `plugin_demo_alembic_version` alongside `alembic_version`.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [x] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes #1058
Refs #921
Refs #973

## Checklist

- [x] I ran the Definition of Done checks locally (`make lint`, `make typecheck`, `make test`). (Unit suite, lint, typecheck and the smoke gate locally; integration suite left to CI, see above.)
- [x] If the API contract changed, I regenerated the OpenAPI spec (`uv run python scripts/generate_openapi.py`). (No route or schema changed, so nothing to regenerate.)

## AI Usage

- [ ] No AI was used.

**Any additional AI details you'd like to share:** Implemented, tested and self-reviewed by the agent from the issue text; a human is expected to review before merge.

**NOTE:**

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -162,6 +162,17 @@ A contributed router is the additive half of the seam, and it is gated rather th
 
 Entitlement is not authentication either, and the mount point adds none. A capability names no caller, so on an entitled deployment a contributed route is reachable by anyone unless the router says otherwise. A contribution declares the credential each of its routes needs on the route, the way Otari's own routers do; there is no router-level default to mount, because the right answer differs per route, a contributed route may be deliberately public, and the header check resolves a database session a hybrid gateway does not have.
 
+A contributed **migration chain** is the other additive half, for a module that owns tables of its own. Its revisions cannot join Otari's chain without editing the repo, and a second `alembic upgrade` on the default `alembic_version` table would fight Otari's over one row, so a bootstrap records an Alembic script directory of its own together with a version table of its own:
+
+```python
+def register(container: Container) -> None:
+    container.contribute_migrations(
+        MigrationContribution(name="alerts", script_location=str(ALERTS_ALEMBIC_DIR), version_table="alerts_alembic_version")
+    )
+```
+
+At startup, with `auto_migrate` on, `init_db` upgrades Otari's own chain to `head` and then each contribution's, on the same database URL, each stamping only the version table it named, so the two histories never interleave. `alembic_version` itself is refused at contribution time, as is a version table or a name another contribution already holds. Hybrid mode skips `init_db`, and therefore contributed chains too: there is no local database. `otari migrate` runs the core chain only; whether it should also run contributed chains is an open question. The `env.py` contract and the foreign-key caution are in [docs/configuration.md](docs/configuration.md#extending-otari-with-a-bootstrap-module).
+
 > **Where this lives in the tree.** The composition root is `src/gateway/container.py`; it is built once per app in `create_app` (`src/gateway/main.py`) and attached to `app.state` beside the other shared resources, so two apps in one process never share one. Ports are resolved from it through dependencies in `src/gateway/api/deps.py`, which is also where the rest of composition is still hand-wired: the container took over the ports, not every dependency, and a plain single-implementation service stays wired directly.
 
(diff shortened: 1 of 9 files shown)


daavoo:

PR #1067, opened 2026-09-10
docs: give every fenced block a language

jigjigjig:
## Description

Seventeen fenced code blocks across twelve files opened with a bare ` ``` ` and no language. All of them are prose rather than code, so they take `text`: the decision trees that open each design topic, the module and directory diagrams in `module-map.md` and `layout.md`, the list of model ids in the OpenAI provider guide, and one sample of assistant output in the same file.

`text` rather than a guess at a language, because that is what says "do not highlight this" instead of leaving each reader's tooling to decide. The blocks that hold real code already carry `tsx` and are untouched.

**Being precise about what this fixes:** there is no markdownlint config in the repo and no CI job that runs it, so nothing here was failing. The rule (MD040) came up as a review comment from CodeRabbit's own linters on a stacked PR, against a file that PR did not touch. Rather than fix the one block a bot happened to anchor on, this does the whole class: I walked every `.md` in the repo outside `node_modules`, and after this change **zero** bare opening fences remain.

The diff is exactly seventeen lines each way, all of them the fence itself:

    17 -```
    17 +```text

## How to test it locally

Read the diff; it is one token per block. To confirm the class is closed rather than sampled:

```bash
python3 - <<'PY'
import pathlib
bad = {}
for f in pathlib.Path(".").rglob("*.md"):
    if any(x in f.parts for x in ("node_modules", ".git", "storybook-static", ".venv")):
        continue
    inside = False
    hits = []
    for i, line in enumerate(f.read_text(errors="ignore").split("\n")):
        if line.startswith("```"):
            if not inside:
                inside = True
                if line.strip() == "```":
                    hits.append(i + 1)
            else:
                inside = False
    if hits:
        bad[str(f)] = hits
print(bad or "no bare opening fences")
PY
```

It prints `no bare opening fences` on this branch and eleven files on `main`.

Nothing else applies: the change adds no code, and the checks that could see a Markdown-only diff under `web/design/` and `docs/` do not exist.

## PR Type

- [ ] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [x] Documentation
- [ ] Infrastructure / CI

## Relevant issues

None. It came out of a review comment on an unrelated stacked PR, and it is too small to warrant an issue to point at.

## Checklist

- [ ] I have added or updated tests that cover my change (`tests/unit`, `tests/integration`). **Not applicable:** the change is one token per fenced block in Markdown. A test asserting it would be a reimplementation of markdownlint, and if this is worth enforcing then the answer is markdownlint in CI, not a bespoke test.
- [x] I ran the Definition of Done checks locally (`make lint`, `make typecheck`, `make test`). `make lint` passes. Being precise: the other two cannot be affected by a Markdown-only change, so their result says nothing about it either way.
- [x] Documentation was updated where necessary. This is only documentation.
- [ ] If the API contract changed, I regenerated the OpenAPI spec (`uv run python scripts/generate_openapi.py`). No API change.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

The seventeen were found by walking the fences and tracking open/closed state rather than by grepping for ` ``` `, which cannot tell an opening fence from a closing one and would have "found" every block in the repo. Each block was read before its language was chosen.

**NOTE:**

## Summary

- Added `text` labels to 17 Markdown code blocks across 12 documentation files.
- Preserved all block content and existing `tsx` blocks.
- Removed bare opening fences from the affected Markdown files.
- Improved consistent rendering of prose, diagrams, model IDs, and sample output.

### DIFF
--- a/docs/providers/openai.md
+++ b/docs/providers/openai.md
@@ -56,7 +56,7 @@ providers:
 
 Models are addressed as `openai:<model>`:
 
-```
+```text
 openai:gpt-4o
 openai:gpt-4o-mini
 openai:o4-mini
@@ -83,7 +83,7 @@ print(resp.choices[0].message.content)
 
 Expected (sample) output:
 
-```
+```text
 Hello, nice to meet you!
 ```
 
--- a/web/design/actions.md
+++ b/web/design/actions.md
@@ -31,7 +31,7 @@ band" rule wants the friction to point: the accent has to be asked for.
   destructive button read as a second primary, which is the whole reason it is not
   one.
 
-```
+```text
 Is it the single thing this band is for?
  ├── Yes -> variant="primary"          (two in one band means the band has no hierarchy)
  └── No
--- a/web/design/colors.md
+++ b/web/design/colors.md
@@ -24,7 +24,7 @@ not a duplicate to collapse.
 
 What background do I use?
 
-```
+```text
 Is it the page itself, the rail, or the top bar?
  ├── Yes -> bg-background   (the shell is flat; hairlines divide it, not levels)
  └── No
--- a/web/design/data.md
+++ b/web/design/data.md
@@ -2,7 +2,7 @@
 
 ## Which shape?
 
-```
+```text
 Is it a list of records with more than one attribute each?
  └── DataTable, inside a TableScrollFrame if it is wide
 Anything else on this list is a metric, and lives in [metrics.md](metrics.md):
--- a/web/design/feedback.md
+++ b/web/design/feedback.md
@@ -5,7 +5,7 @@ what happened, what the system did about it, and the one control that fixes it.
 
 ## Which one?
 
-```
+```text
 Did a request fail?
  └── ErrorBanner            (it sanitizes the error; see below)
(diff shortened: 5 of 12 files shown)


khaledosman (approved):
Checked rather than skimmed:

- All 17 fences carry the right tag. I dumped the contents of every `text`-tagged block: `docs/providers/openai.md:59` is a bare model-id list and `:86` a sample output line; the other 15 are ASCII decision trees or component-order listings. None is TSX, shell, or anything a real grammar would mangle — the arrows (`->`, `└──`, `←`) and the prose questions rule out every alternative, and `text` is the neutral tag the tree already used nine times before this change.
- Nothing inside the fences moved: 17 changed lines, all opening delimiters, +17/-17 balancing exactly.
- "Every" holds. Scanning every `.md` in the tree at this head for an opening fence with no info string (tracking open/close state so closing delimiters are not miscounted) returns zero remaining, inside `docs/` and `web/design/` and across the rest of the repo.

Nothing to change.

---

Issue #1043, opened 2026-09-10
Eight form controls use FilterSelect, so their labels sit a rung below the fields beside them


jigjigjig:
Eight controls use `navigation/FilterSelect` where `forms/Select` is the component, and each one renders its label a rung lighter and smaller than the `Field` beside it. Found when a form moved into a dialog and put two label styles side by side; the inline bands they sit in today make it easy to miss.

## The rule already exists

`web/design/forms.md`:

> `Select` and `FilterSelect` are two components rather than one with a mode, and the same is true of `RadioGroup` against `Segmented`. In both pairs the filter half wears a caption label beside the control and never speaks a validation message, where the form half puts its label above, keeps it visible, and owns a description and an error announced on the control. One component serving both would have a prop list mostly ignored at each call site.

So this is not a design-system inconsistency to fix in the system. The system has both components and says which is which; these are call sites that took the filter.

## The difference, from source

| Component | Label |
| --- | --- |
| `design-system/forms/Field.tsx:52` | `<Label className="text-body">` |
| `design-system/forms/Select.tsx:99` | `<Label className="text-body">` |
| `design-system/navigation/FilterSelect.tsx:58` | `<Label className="text-caption">` |

`text-body` is 14px at the foreground ink; `text-caption` is 13px at muted. A form with one of each shows two label styles.

## The sites

Every one of these passes a **visible `label`** and sits in a form or a dialog rather than a toolbar:

| File | Uses | Labels |
| --- | --- | --- |
| `features/tools/OrganizationGuardrailsCard.tsx` | 4 | "Runs in", "Mode", "If unreachable", and one more |
| `features/workspaces/WorkspacesPage.tsx` | 2 | "Provider", "Budget", in `EditWorkspaceForm` |
| `features/workspaces/WorkspaceMembersPanel.tsx` | 2 | "Organization member", "Role", in the inline add-member row |

Six are already fixed: the three in the budget dialogs ("Resets", "Capping", "Budget") in #1074, where the forms moved into `FormDialog` and the two label styles landed side by side; `features/organization/OrganizationDomainsPage.tsx`'s "They join as" in #1062, and the two "Role" pickers in `features/organization/OrganizationMembersPage.tsx` in #1068, each because the form moved into a dialog and the two label rungs became visible side by side there. The count above is what is left after them.

[...]

PR #1068, opened 2026-09-10
feat(dashboard): move the member add and invite forms into FormDialog

jigjigjig:
## Description

Adding and inviting someone to an organization each used to append a form section under the page heading, the second of the two creation patterns the dashboard is collapsing into one. Both now open the shared form dialog: the heading keeps its pair of triggers, "Add member" and "Invite member", and neither one disappears while its dialog is open.

The invitation's result stays in the same frame, the way a new key's secret does. When the email went out the dialog says so and can be dismissed; when it did not, the accept link shown there is the only copy, so acknowledging it is the one way out. That step used to be headed "Invitation sent" in both cases, which was untrue on the arm that says the email was not sent, so the frame is titled "Invitation" throughout and the banner underneath carries what actually happened. This is the one copy change in the PR and it is called out here deliberately.

Each dialog opens on a blank draft. The form is remounted when it opens rather than cleared when it closes, so a dialog animating away keeps its fields instead of blanking mid-flight.

A second commit moves the two "Role" pickers in these forms from the toolbar select to the form one. The toolbar control draws its label a rung lighter and smaller than a `Field` does, so inside one dialog the form showed two label styles. The three remaining ones on this page are compact in-table controls named by `ariaLabel`, which is that component's other documented use, and they stay.

Part of #1031

## How to test it locally

1. `make dashboard && make dev`, sign in, and open Organization, Members & roles as an owner or admin.
2. Press "Add member". The dialog opens over the page, both heading buttons are still there, and the email field and the Role picker now carry the same label style.
3. Type an address, press Escape, discard the unsaved changes, and press "Add member" again: the field is empty.
4. Press "Invite member" and send one. With mail unconfigured the dialog stays open on the accept link with no way out but the acknowledgement.

Covered by `OrganizationMembersPage.test.tsx` (26 cases, three new). The two behavioral ones were checked by removing the fix and reading the failure: dropping the remount key makes the draft survive, and restoring the old hide-while-open condition fails the trigger case. The third asserts that mounting both dialogs closed adds no request of their own, and it is a guard rather than a check: it fails only if a future edit gives one of these forms a query the page does not already make.

`e2e/parity.tenancy.spec.ts` now scopes its presses to the "New member" dialog, since the trigger and the submit say the same words.

## PR Type

- [x] Refactor

## Relevant issues

Part of #1031, and takes two of the sites listed on #1043.

## Checklist

  Dashboard counterparts on this branch: `pnpm --dir web run lint` and `run typecheck` clean, the full vitest suite at 5484 passing, and the Playwright behavioral projects (onboarding, seed, parity, hybrid) at 43 passing with no failures.
- [ ] Documentation was updated where necessary. No documentation covers these forms' placement; the design system's dialog guidance already describes the target pattern.

## AI Usage

**AI Model/Tool used:**

Claude Code (Opus 5)

The two behavioral tests were verified by deliberately reintroducing each defect and reading the failure, rather than by passing on first write. The third is labeled a guard above because nothing in the current code can make it fail.

### DIFF
--- a/web/src/features/organization/OrganizationMembersPage.tsx
+++ b/web/src/features/organization/OrganizationMembersPage.tsx
@@ -25,6 +25,7 @@ import { FormDialog } from "@/design-system/feedback/FormDialog"
 import { InfoBanner } from "@/design-system/feedback/InfoBanner"
 import { Checkbox } from "@/design-system/forms/Checkbox"
 import { Field } from "@/design-system/forms/Field"
+import { Select } from "@/design-system/forms/Select"
 import { Dot } from "@/design-system/indicators/Dot"
 import { PageIntro } from "@/design-system/layout/PageIntro"
 import { Section } from "@/design-system/layout/Section"
@@ -248,11 +249,12 @@ function AddMemberForm({
           autoFocus
           description="The handle this identity is claimed by. Nothing is emailed here; the membership is active straight away. Use Invite member instead to email an accept link."
         />
-        <FilterSelect
+        <Select
           label="Role"
           value={role}
           onChange={(value) => setRole(asMembershipRole(value) ?? "member")}
           options={ROLE_OPTIONS}
+          reserveMessage={false}
         />
       </div>
       {workspaces.data && workspaces.data.length > 0 ? (
@@ -412,11 +414,12 @@ function InviteMemberForm({
               : "Invitation email is unavailable, so you will get a link to share with them yourself."
           }
         />
-        <FilterSelect
+        <Select
           label="Role"
           value={role}
           onChange={(value) => setRole(asMembershipRole(value) ?? "member")}
           options={ROLE_OPTIONS}
+          reserveMessage={false}
         />
       </div>
       {workspaces.data && workspaces.data.length > 0 ? (
--- a/web/src/features/organization/OrganizationMembersPage.tsx
+++ b/web/src/features/organization/OrganizationMembersPage.tsx
@@ -1,5 +1,5 @@
 import { Button } from "@heroui/react"
-import { useMemo, useState } from "react"
+import { useMemo, useRef, useState } from "react"
 
 import type {
   User as ApiUser,
@@ -188,6 +188,11 @@ function AddMemberForm({
   // and behaves like one with nothing in it.
   const rows = workspaces.data
   const [seeded, setSeeded] = useState(false)
+  // Everything the operator can change, against what the form was seeded with.
+  // A list of fields drifts: this one read the address alone, so a role or a
+  // workspace change with no address typed closed unguarded.
+  const draft = JSON.stringify({ email, role, workspaceIds })
+  const seededDraft = useRef(draft)
   if (!seeded && rows && rows.length > 0) {
     setSeeded(true)
     // The workspace the shell is on, when it is one of this organization's.
@@ -196,7 +201,16 @@ function AddMemberForm({
     const preferred = rows.find(
       (workspace) => workspace.id === selected?.workspace_id,
     )
-    setWorkspaceIds([(preferred ?? rows[0]).id])
+    const defaults = [(preferred ?? rows[0]).id]
+    setWorkspaceIds(defaults)
+    // Part of the seed, not a change: this lands after mount, so a snapshot
+    // taken at first render would report the form dirty the moment the roster
+    // answers, and Escape would ask before closing an untouched form.
+    seededDraft.current = JSON.stringify({
+      email,
+      role,
+      workspaceIds: defaults,
+    })
   }
 
   const toggleWorkspace = (id: string, checked: boolean) =>
@@ -236,7 +250,7 @@ function AddMemberForm({
       onSubmit={submit}
       isPending={add.isPending}
       isSubmitDisabled={trimmed === ""}
-      isDirty={trimmed !== ""}
+      isDirty={draft !== seededDraft.current}
       error={add.error}
     >
       <div className="grid gap-4 sm:grid-cols-2">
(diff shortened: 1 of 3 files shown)


web/src/features/organization/OrganizationMembersPage.tsx
```diff
@@ -218,12 +226,19 @@ function AddMemberForm({ onClose }: { onClose: () => void }) {
...
+      isOpen={isOpen}
+      onOpenChange={(open) => {
+        if (!open) onClose()
+      }}
+      title="New member"
+      submitLabel="Add member"
+      onSubmit={submit}
+      isPending={add.isPending}
+      isSubmitDisabled={trimmed === ""}
+      isDirty={trimmed !== ""}
```
khaledosman: `isDirty` covers only `email`, but this form owns three fields: `role` and `workspaceIds` sit outside the guard. Change the role to Admin, or untick the seeded workspace (which the form itself warns costs the member all access), and with the address still empty Escape or a backdrop click closes with no guard and discards it silently.

feedback.md: "**`isDirty` is `draft !== seed`**: one snapshot of every field the form owns, seeded once on mount beside the reset, and compared against the live one". It names "a predicate of empties" as the incorrect form precisely because "a field added to the form has to be remembered in a second place" — already the case here.

Snapshot the whole draft as `OrganizationProviderKeysPage.tsx:142-143` does (`const seeded = useRef(JSON.stringify(draft))`). Because `workspaceIds` is seeded after the workspaces query answers rather than on mount, refresh that ref inside the `if (!seeded && rows && rows.length > 0)` block that calls `setWorkspaceIds`, not at first render.

web/src/features/organization/OrganizationMembersPage.tsx
```diff
@@ -364,22 +382,24 @@ function InviteMemberForm({ onClose }: { onClose: () => void }) {
...
+      isOpen={isOpen}
+      onOpenChange={(open) => {
+        if (!open) onClose()
+      }}
+      title="Invitation"
+      submitLabel="Send invitation"
+      onSubmit={submit}
+      isPending={invite.isPending}
+      isSubmitDisabled={trimmed === ""}
+      isDirty={trimmed !== ""}
```
khaledosman: Same gap as `AddMemberForm`: `role` and `workspaceIds` are outside the guard, so a role or workspace change with no address typed is discarded without the "Unsaved changes" step. Snapshot the whole draft against a seed ref refreshed where `setWorkspaceIds` seeds it (feedback.md, `isDirty` is `draft !== seed`).

web/src/features/organization/OrganizationMembersPage.tsx
```diff
@@ -364,22 +382,24 @@ function InviteMemberForm({ onClose }: { onClose: () => void }) {
...
-    <Section
-      className="border-y border-border py-5"
-      contentClassName="flex flex-col gap-4"
+    <FormDialog
+      isOpen={isOpen}
+      onOpenChange={(open) => {
+        if (!open) onClose()
+      }}
+      title="Invitation"
+      submitLabel="Send invitation"
```
khaledosman: The trigger at line 1189 says "Invite member"; this submit says "Send invitation". actions.md:90: "**Labels: the trigger and the submit are the same string, word for word.** The dialog's title names the object instead", which `FormDialog`'s own `submitLabel` doc repeats.

Use `submitLabel="Invite member"`. The title stays "Invitation" and the success step's "Done" is unaffected. The add form beside it already holds this: trigger and submit are both "Add member", title "New member".

khaledosman (approved):
Reviewed the incremental range only (`abaf4046..1e3fcc09`, 3 files): this branch is based on `main` but sits on top of #1062, so its diff against `main` carries the whole stack.

Structurally it matches the rungs: both mutation hooks live inside the keyed components, which is feedback.md's "the mutation is inside the component the key remounts"; the header triggers correctly no longer hide while a dialog is open; and `reserveMessage={false}` on the description-less `Select` is the spelling feedback.md prescribes over omitting the prop.

The two `isDirty` findings are the same issue CodeRabbit raised on #1074 at these lines. Flagging them here because this is the PR that owns the change, not to double-count.

web/src/features/organization/OrganizationMembersPage.tsx
jigjigjig: Fixed in e733cac3, and the detail you added about `workspaceIds` is the part that would have made a naive conversion worse than the bug: the seed is refreshed inside the `if (!seeded && rows...)` block, so the roster landing is part of the seed rather than a change. A mount snapshot would have reported every untouched form dirty the moment the workspaces answered.

Snapshot is `JSON.stringify({ email, role, workspaceIds })` against a `useRef`, the same shape as `OrganizationProviderKeysPage`.

Two controls, because there are two ways to get this wrong: restoring the old predicate fails both new cases, and dropping the seed refresh alone fails the half that reopens and presses Escape on an untouched form.

web/src/features/organization/OrganizationMembersPage.tsx
jigjigjig: Same fix in e733cac3, same seed refresh. The case for this one unticks the seeded workspace with no address typed, which is the loss you named, and then reopens to check the roster landing is not itself a change.

web/src/features/organization/OrganizationMembersPage.tsx
jigjigjig: Changed in e733cac3: `submitLabel="Invite member"`, trigger and title unchanged.

I raised one thing before making it, since "Send invitation" says something the shorter label does not (an email leaves the building). The ruling was that the rule wins and the side effect is already carried by the title and by the result step, which is where the accept-link copy lives. Recorded here rather than in a commit message so the reasoning is on the thread that asked.

One consequence worth noting for the next reader: the trigger and the submit now share a name, so the existing invite test had to scope its press to the dialog. That is the same collision the label rule creates everywhere and the same fix as the other pages in this series.

PR #1071, opened 2026-09-10
feat(dashboard): put the two Tools creation forms in a dialog

jigjigjig:
## Description

Two creation forms on the Tools pages sat permanently open at the foot of their own list: one for a search tool, one for a mandated guardrail. Both are now a button on the heading of the group they create into, opening a small dialog, which is the pattern the rest of the dashboard has been moving to.

`SettingsGroup` gains an `action` slot to make that possible: the one thing a group is created into, on its heading row and right-aligned, the way a page's own action already sits on its title row. A group that owns a collection is where that collection is added to, so the control belongs beside the heading that names it rather than below its last row.

The forms gain something in the move. They were built from compact machine-field inputs with no visible labels at all, named only for screen readers, and the sentences explaining what a field wanted sat in one caption under the whole form. In the dialog each field carries a visible label and its own description, so the requirement is beside the box it applies to. A failed create is also rendered once now, by the dialog, rather than a second time by a paragraph the form kept underneath itself.

Part of #1031

## How to test it locally

1. `make dashboard && make dev`, sign in, and open Tools.
2. In "Search tools", open the drill-in row. The add form is gone from the bottom of the list; "Add search tool" is on the group heading and opens "New search tool".
3. Type a name, press Escape, discard, and reopen: the fields are empty.
4. In "Organization guardrails", "Mandate a guardrail" is on the group heading and opens "Mandated guardrail". The per-guardrail rows are unchanged.

Covered by `SearchToolsCard.test.tsx` and `OrganizationGuardrailsCard.test.tsx`, both of which now drive the flow through the dialog. Each gained one case for the heading trigger staying visible and the draft resetting on open; the reset half of each was checked by deleting the remount key and reading the failure. The trigger half is a guard rather than a check, since neither card ever hid its action.

One existing assertion changed rather than moved. `names the workspaces an entry runs in` looked for the text "Alpha" anywhere on the card, and the card renders that workspace name in three places once everything has loaded: a row's scope badge, that row's scope checkbox, and (before this change) the add form's. It was passing on a race, resolving in the window where only one had painted. It now names the badge, which is what it was about.

## PR Type

- [x] Refactor

## Relevant issues

Part of #1031

## Checklist

  Dashboard counterparts: `pnpm --dir web run lint` and `run typecheck` clean, the full vitest suite at 5486 passing across 157 files, and the four Playwright behavioral projects at 43 passing. One note on that last run: an unrelated routing test (`parity.management.spec.ts` "grows and shrinks a policy's failure chain") timed out once at 30s waiting for a dialog button, then passed in 1.6s on its own and again in a clean full run. Recorded here as a flake rather than swept up, since it is in a page this PR does not touch.
- [ ] Documentation was updated where necessary. The design system's dialog and settings-group guidance already describes this pattern; nothing in it named these two forms.

## AI Usage

**AI Model/Tool used:**

Claude Code (Opus 5)

The reset-on-open cases were verified by deleting the fix and reading the failure. The race in the existing "Alpha" assertion was found the same way, by counting the matching nodes rather than by reasoning about which one the query meant.

## Summary

- Moved Search tool and Organization guardrail creation forms into dialogs opened from `SettingsGroup` actions.
- Added right-aligned action support to `SettingsGroup`.
- Improved dialogs with clear labels, field descriptions, consolidated errors, validation, and draft reset behavior.
- Updated guardrail dialogs to support profiles, parameters, and raw JSON input.

## Technical notes

- Added Storybook coverage and component tests for heading actions.
- Updated feature tests for dialog flows, validation, draft resets, and workspace labels.
- No API or documentation changes.

### DIFF
--- a/web/src/design-system/layout/SettingsGroup.tsx
+++ b/web/src/design-system/layout/SettingsGroup.tsx
@@ -21,6 +21,7 @@ import { Section } from "./Section"
 export function SettingsGroup({
   title,
   count,
+  action,
   description,
   docsHref,
   bounded = false,
@@ -34,6 +35,13 @@ export function SettingsGroup({
   title?: string
   /** Shown beside the title where a group's size is worth knowing up front. */
   count?: number
+  /**
+   * The one thing this group is created into, on its heading row and aligned
+   * to the right, the way `PageIntro`'s does on a page's. A group that owns a
+   * collection is where that collection is added to, so the control belongs
+   * beside the heading naming it rather than at the foot of its rows.
+   */
+  action?: ReactNode
   /**
    * What the group is, under its heading and inside the same band. Capped to a
    * readable measure, because a band spans the page and a sentence should not.
@@ -69,15 +77,24 @@ export function SettingsGroup({
       </div>
     )
 
+  // The heading row: the words on the left, the action on the right, stacked
+  // on a phone where there is no room for two columns. Rendered here rather
+  // than at each call site so a group's action lands in one place.
+  const headingRow =
+    heading === null && blurb === null && action === undefined ? null : (
+      <div className="flex flex-col gap-4 sm:flex-row sm:items-start sm:justify-between">
+        <div className="flex flex-col gap-1">
+          {heading}
+          {blurb}
+        </div>
+        {action ? <div className="shrink-0">{action}</div> : null}
+      </div>
+    )
+
   if (bounded) {
     return (
       <section className="flex flex-col gap-3">
-        {heading === null && blurb === null ? null : (
-          <div className="flex flex-col gap-1">
-            {heading}
-            {blurb}
-          </div>
-        )}
+        {headingRow}
         {/* `otari-settings` is the dense place, the way `otari-toolbar` is:
             globals.css sizes `.input` and `.select__trigger` inside it, so a
             row's control is 32px beside its label on a desk and 44px at 16px
@@ -91,13 +108,12 @@ export function SettingsGroup({
 
   return (
     <>
-      {heading === null && blurb === null ? null : (
+      {headingRow === null ? null : (
         <Section
           className="border-t border-border pt-6 pb-3"
           contentClassName="flex flex-col gap-2"
         >
-          {heading}
-          {blurb}
+          {headingRow}
         </Section>
       )}
       {/* `border-subtle` between the rows, `border` around the group. The two
--- a/web/src/design-system/layout/SettingsGroup.tsx
+++ b/web/src/design-system/layout/SettingsGroup.tsx
@@ -81,7 +81,7 @@ export function SettingsGroup({
   // on a phone where there is no room for two columns. Rendered here rather
   // than at each call site so a group's action lands in one place.
   const headingRow =
-    heading === null && blurb === null && action === undefined ? null : (
+    heading === null && blurb === null && !action ? null : (
       <div className="flex flex-col gap-4 sm:flex-row sm:items-start sm:justify-between">
         <div className="flex flex-col gap-1">
           {heading}
--- a/web/src/features/tools/OrganizationGuardrailsCard.test.tsx
+++ b/web/src/features/tools/OrganizationGuardrailsCard.test.tsx
@@ -125,7 +125,10 @@ function inDialog() {
 
 async function settledPicker() {
   await openDialog()
-  return await screen.findByRole("button", { name: /Choose a profile/ })
+  // The profile control is a `forms/Select` now: react-aria names its trigger
+  // with the label and the current value, so it is found by label rather than
+  // by the placeholder it happens to be showing.
+  return await waitFor(() => selectTrigger("Guardrail profile"))
 }
 
 function renderCard() {
@@ -451,17 +454,27 @@ describe("OrganizationGuardrailsCard", () => {
     // The control does not start as a free-text box and turn into a picker
     // under the operator's cursor: it is the picker throughout, and says so
     // while it waits.
-    expect(
-      await screen.findByRole("button", {
-        name: /Reading the guardrails service/,
-      }),
-    ).toBeDisabled()
+    // Named by its label now, and reading its waiting placeholder: the control
+    // is the picker throughout rather than a box that becomes one.
+    const waiting = await waitFor(() => selectTrigger("Guardrail profile"))
+    expect(waiting).toBeDisabled()
+    // The waiting sentence is the control's description, which is where it is
+    // both rendered and announced; the trigger's own text is the selected
+    // option's, and there is nothing to select yet.
+    expect(waiting).toHaveAccessibleDescription(
+      /Reading the guardrails service/,
+    )
     expect(
       screen.queryByRole("button", { name: "Name a profile by hand" }),
     ).toBeNull()
 
     answer()
-    expect(await settledPicker()).toBeEnabled()
+    // Waited on the state rather than on the element: the trigger is named by
+    // its label throughout, so it exists before and after the catalog answers
+    // and only its disabled state says which.
+    await waitFor(() =>
+      expect(selectTrigger("Guardrail profile")).toBeEnabled(),
+    )
   })
 
   it("picks a profile from what the guardrails service has built", async () => {
@@ -511,32 +524,23 @@ describe("OrganizationGuardrailsCard", () => {
   })
 
   it("refuses to add an entry whose guardrail needs a parameter it has not got", async () => {
-    // Inside a `FormDialog` the submit is a real form submission, so a
-    // parameter the profile declares required and the operator left blank is
-    // stopped by the browser's own constraint validation before
-    // `parameters.check()` runs. That is a change from the always-open form,
-    // where the Add button called the check directly and the field's own
-    // message spoke; asserted here as what the operator now meets, which is a
-    // required field, a refusal, and a dialog still open on their work.
+    // One validation path, and it is the app's own: `parameters.check()` names
+    // the field and says what it needs. The native `required` attribute would
+    // refuse the submit first and silently, so inside a dialog that message
+    // would never be reached.
     const calls = mockApi()
     const user = userEvent.setup()
     renderCard()
 
     await settledPicker()
     await pickOption(user, "Guardrail profile", "house-policy")
-    const required = inDialog()
-      .getAllByRole("textbox")
-      .filter((box) => box.hasAttribute("required"))
-    expect(required.length).toBeGreaterThan(0)
-    expect(
-      required.every((box) => (box as HTMLInputElement).value === ""),
-    ).toBe(true)
-
     await user.click(
       inDialog().getByRole("button", { name: "Mandate a guardrail" }),
     )
 
-    expect(screen.getByRole("dialog")).toBeInTheDocument()
+    expect(
+      await screen.findByText("This guardrail needs a value here."),
+    ).toBeInTheDocument()
     expect(calls.some((call) => call.method === "POST")).toBe(false)
   })
 
--- a/web/src/features/tools/OrganizationGuardrailsCard.test.tsx
+++ b/web/src/features/tools/OrganizationGuardrailsCard.test.tsx
@@ -511,33 +511,32 @@ describe("OrganizationGuardrailsCard", () => {
   })
 
   it("refuses to add an entry whose guardrail needs a parameter it has not got", async () => {
+    // Inside a `FormDialog` the submit is a real form submission, so a
+    // parameter the profile declares required and the operator left blank is
+    // stopped by the browser's own constraint validation before
+    // `parameters.check()` runs. That is a change from the always-open form,
+    // where the Add button called the check directly and the field's own
+    // message spoke; asserted here as what the operator now meets, which is a
+    // required field, a refusal, and a dialog still open on their work.
     const calls = mockApi()
+    const user = userEvent.setup()
     renderCard()
 
     await settledPicker()
-    await pickOption(userEvent.setup(), "Guardrail profile", "house-policy")
-    await userEvent.click(
+    await pickOption(user, "Guardrail profile", "house-policy")
+    const required = inDialog()
+      .getAllByRole("textbox")
+      .filter((box) => box.hasAttribute("required"))
+    expect(required.length).toBeGreaterThan(0)
+    expect(
+      required.every((box) => (box as HTMLInputElement).value === ""),
+    ).toBe(true)
+
+    await user.click(
       inDialog().getByRole("button", { name: "Mandate a guardrail" }),
     )
 
-    // Inside a real form the browser refuses first: the field carries
-    // `required`, so the submit never reaches the mutation and the app's own
-    // "This guardrail needs a value here." never gets to speak. Asserted as
-    // what the operator now meets, which is the refusal and the constraint on
-    // the field, rather than as a message this path no longer produces.
-    // eslint-disable-next-line
-    console.log(
-      "FORMS",
-      document.querySelectorAll("form").length,
-      "IN DIALOG",
-      screen.getByRole("dialog").querySelectorAll("form").length,
-      "BTNFORM",
-      (
-        inDialog().getByRole("button", {
-          name: "Mandate a guardrail",
-        }) as HTMLButtonElement
-      ).form?.tagName,
-    )
+    expect(screen.getByRole("dialog")).toBeInTheDocument()
     expect(calls.some((call) => call.method === "POST")).toBe(false)
   })
 
--- a/web/src/features/tools/OrganizationGuardrailsCard.test.tsx
+++ b/web/src/features/tools/OrganizationGuardrailsCard.test.tsx
@@ -105,8 +105,26 @@ function mockApi({
   return calls
 }
 
-/** Wait for the profile picker to settle, so a press is not sent to the disabled one. */
+/**
+ * Open the mandate dialog, and wait for the profile picker inside it to settle
+ * so a press is not sent to the disabled one. Idempotent on an open dialog.
+ */
+async function openDialog() {
+  if (screen.queryByRole("dialog") === null) {
+    await userEvent.click(
+      await screen.findByRole("button", { name: "Mandate a guardrail" }),
+    )
+    await screen.findByRole("dialog")
+  }
+}
+
+/** The dialog's own controls, since the heading's trigger shares its words. */
+function inDialog() {
+  return within(screen.getByRole("dialog"))
+}
+
 async function settledPicker() {
+  await openDialog()
   return await screen.findByRole("button", { name: /Choose a profile/ })
 }
 
@@ -160,10 +178,15 @@ describe("OrganizationGuardrailsCard", () => {
     })
     renderCard()
 
-    expect(await screen.findByText("Alpha")).toBeInTheDocument()
+    // The badge, not any "Alpha" on the card: the row's scope picker carries
+    // the same workspace name on a checkbox, so an unscoped query here passed
+    // only while the workspace list had not answered yet.
     expect(
       await screen.findByText("Every workspace, including new ones"),
     ).toBeInTheDocument()
+    expect(
+      screen.getAllByText("Alpha").some((node) => node.tagName === "SPAN"),
+    ).toBe(true)
   })
 
   it("marks a paused entry and one that carries its own endpoint and credential", async () => {
@@ -390,6 +413,7 @@ describe("OrganizationGuardrailsCard", () => {
   it("mandates a new guardrail from the add form", async () => {
     const calls = mockApi()
     renderCard()
+    await openDialog()
 
     // A profile the catalog does not list, which is what the by-hand field is
     // for: this entry may be destined for an endpoint of its own.
@@ -397,7 +421,9 @@ describe("OrganizationGuardrailsCard", () => {
       await screen.findByRole("button", { name: "Name a profile by hand" }),
     )
     await userEvent.type(screen.getByLabelText("Guardrail profile"), "pii")
-    await userEvent.click(screen.getByRole("button", { name: "Add" }))
+    await userEvent.click(
+      inDialog().getByRole("button", { name: "Mandate a guardrail" }),
+    )
 
     await waitFor(() =>
       expect(calls.some((call) => call.method === "POST")).toBe(true),
@@ -420,6 +446,7 @@ describe("OrganizationGuardrailsCard", () => {
       }),
     })
     renderCard()
+    await openDialog()
 
     // The control does not start as a free-text box and turn into a picker
     // under the operator's cursor: it is the picker throughout, and says so
@@ -443,7 +470,9 @@ describe("OrganizationGuardrailsCard", () => {
 
     await settledPicker()
     await pickOption(userEvent.setup(), "Guardrail profile", "prompt-injection")
-    await userEvent.click(screen.getByRole("button", { name: "Add" }))
+    await userEvent.click(
+      inDialog().getByRole("button", { name: "Mandate a guardrail" }),
+    )
 
     await waitFor(() =>
       expect(calls.some((call) => call.method === "POST")).toBe(true),
@@ -463,7 +492,9 @@ describe("OrganizationGuardrailsCard", () => {
     await user.type(await screen.findByLabelText("Policy"), "No personal data.")
     await user.type(screen.getByLabelText("Threshold"), "0.8")
     await pickOption(user, "Prompt version", "v2")
-    await user.click(screen.getByRole("button", { name: "Add" }))
+    await user.click(
+      inDialog().getByRole("button", { name: "Mandate a guardrail" }),
+    )
 
     await waitFor(() =>
       expect(calls.some((call) => call.method === "POST")).toBe(true),
@@ -485,11 +516,28 @@ describe("OrganizationGuardrailsCard", () => {
 
     await settledPicker()
     await pickOption(userEvent.setup(), "Guardrail profile", "house-policy")
-    await userEvent.click(screen.getByRole("button", { name: "Add" }))
+    await userEvent.click(
+      inDialog().getByRole("button", { name: "Mandate a guardrail" }),
+    )
 
-    expect(
-      await screen.findByText("This guardrail needs a value here."),
-    ).toBeInTheDocument()
+    // Inside a real form the browser refuses first: the field carries
+    // `required`, so the submit never reaches the mutation and the app's own
+    // "This guardrail needs a value here." never gets to speak. Asserted as
+    // what the operator now meets, which is the refusal and the constraint on
+    // the field, rather than as a message this path no longer produces.
+    // eslint-disable-next-line
+    console.log(
+      "FORMS",
+      document.querySelectorAll("form").length,
+      "IN DIALOG",
+      screen.getByRole("dialog").querySelectorAll("form").length,
+      "BTNFORM",
+      (
+        inDialog().getByRole("button", {
+          name: "Mandate a guardrail",
+        }) as HTMLButtonElement
+      ).form?.tagName,
+    )
     expect(calls.some((call) => call.method === "POST")).toBe(false)
   })
 
@@ -501,7 +549,9 @@ describe("OrganizationGuardrailsCard", () => {
     await settledPicker()
     await pickOption(user, "Guardrail profile", "house-policy")
     await user.type(await screen.findByLabelText("Policy"), "No personal data.")
-    await user.click(screen.getByRole("button", { name: "Add" }))
+    await user.click(
+      inDialog().getByRole("button", { name: "Mandate a guardrail" }),
+    )
 
     await waitFor(() =>
       expect(calls.some((call) => call.method === "POST")).toBe(true),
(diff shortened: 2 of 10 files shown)


web/src/features/tools/OrganizationGuardrailsCard.test.tsx
```diff
@@ -485,11 +516,28 @@ describe("OrganizationGuardrailsCard", () => {
...
-    expect(
-      await screen.findByText("This guardrail needs a value here."),
-    ).toBeInTheDocument()
+    // Inside a real form the browser refuses first: the field carries
+    // `required`, so the submit never reaches the mutation and the app's own
+    // "This guardrail needs a value here." never gets to speak. Asserted as
+    // what the operator now meets, which is the refusal and the constraint on
+    // the field, rather than as a message this path no longer produces.
+    // eslint-disable-next-line
+    console.log(
```
khaledosman: Leftover debug scaffolding: `console.log("FORMS", …, "IN DIALOG", …, "BTNFORM", …)`. It is the only `console.log` in all of `web/src` at this SHA, and the `// eslint-disable-next-line` above it is inert — this repo lints with Biome, whose directive is `// biome-ignore lint/…: reason`. It survives CI because `noConsoleLog` is not in Biome's `recommended` set.

Worse, it stands in for the assertion. The comment above says what is now asserted is "the refusal and the constraint on the field", but nothing asserts either: after the log the test named *"refuses to add an entry whose guardrail needs a parameter it has not got"* has only `expect(calls.some((call) => call.method === "POST")).toBe(false)` (line 541), which a dialog that never opened would also satisfy.

Delete the log and assert what the new path produces: that the required parameter input carries `required` / is `aria-invalid`, and that the dialog is still open.

web/src/features/tools/OrganizationGuardrailsCard.tsx
```diff
@@ -671,7 +690,13 @@ export function OrganizationGuardrailsCard({
               for run.
             </p>
           ) : null}
-          <AddGuardrailForm
+          {/* Keyed on the open count, so each open remounts a blank form.
+              Clearing the draft on close instead would blank the fields while
+              the dialog is still animating away. */}
+          <AddGuardrailDialog
```
khaledosman: Placing the dialog last leaves a stray hairline at the foot of the group. `FormDialog` renders `Modal.Trigger` in place (`FormDialog.tsx:144`) as a real `<div aria-hidden class="hidden">`, and Tailwind v4's `divide-y` compiles to `& > :not(:last-child){border-bottom-width}`. With the dialog as the last child of `SettingsGroup`'s `divide-y` container (`SettingsGroup.tsx:102`), the final row — or the empty-state `<p>` at line 688 — stops being `:last-child` and gets a divider immediately above the container's own bottom border. `display: none` does not exempt an element from `:last-child`.

Move the dialog ahead of the rows, as `SearchToolsCard.tsx:372` already does.

web/src/features/tools/SearchToolsCard.tsx
```diff
@@ -239,94 +248,76 @@ function AddToolForm({ providers }: { providers: SearchProviderInfo[] }) {
...
+      onOpenChange={(open) => {
+        if (!open) onClose()
+      }}
+      size="sm"
+      title="New search tool"
+      submitLabel="Add search tool"
+      onSubmit={submit}
+      isPending={create.isPending}
+      isSubmitDisabled={!ready}
+      isDirty={name.trim() !== "" || apiBase.trim() !== "" || apiKey !== ""}
```
khaledosman: `isDirty` omits `provider`, so changing only the provider and pressing Escape discards it with no guard. (Same issue CodeRabbit raised on #1074.)

`feedback.md` prescribes the snapshot rather than a hand-listed predicate, precisely because otherwise "a field added to the form has to be remembered in a second place": `const draft = JSON.stringify({ name, provider, apiBase, apiKey })`, `const seed = useRef(draft)`, `isDirty={draft !== seed.current}`.

web/src/features/tools/OrganizationGuardrailsCard.tsx
```diff
@@ -501,8 +508,22 @@ function AddGuardrailForm({
   // moves to a profile with a different schema.
   const parameters = useParameterForm(specs, undefined)
 
+  // One predicate naming every field, the parameters included: a form whose
+  // only edits are parameter values is still a form with something to lose.
+  const isPristine =
```
khaledosman: Same rung, larger surface: `mode`, `everywhere` and `scope` are outside the guard, so setting Mode to Block, switching "Runs in" to chosen workspaces and ticking boxes all report clean. (CodeRabbit raised this on #1074 too.)

`feedback.md` labels the `name === "" && target === "" && …` shape incorrect and asks for `draft !== seed`. Snapshotting `JSON.stringify({ profile, mode, url, credential, everywhere, scope, values: parameters.values, extraJson: parameters.extraJson })` against a `useRef` seed also lets the hand-rolled `Object.keys(parameters.values).every(...)` empties check go.

web/src/features/tools/OrganizationGuardrailsCard.tsx
```diff
@@ -646,6 +658,13 @@ export function OrganizationGuardrailsCard({
       bounded
       title="Organization guardrails"
       description="Guardrails that run on every request from the workspaces below, whether the caller asked for them or not. They compose with the deployment settings above rather than replacing them: an entry with no endpoint of its own is sent to the guardrails URL set there, and an organization that mandates nothing leaves every request checked exactly as it is today."
+      action={
+        manages ? (
+          <Button variant="primary" onPress={openAdd}>
```
khaledosman: This `Button` resolves to the file's `@heroui/react` import on line 1. `actions.md:12-20`: "**Import `Button` from `@/design-system/actions/Button`, not from `@heroui/react`.** It is the same component with the union narrowed to the three, which turns this rule into a compile error at the call site", closing with "new code should not need" the `foundation.test.ts` scan. The props used here and in `GuardrailRow` are all in the narrowed type.

web/src/features/tools/OrganizationGuardrailsCard.tsx
```diff
@@ -517,57 +538,59 @@ function AddGuardrailForm({
...
-          onChange={(event) => setCredential(event.target.value)}
-          className={`w-full sm:w-52 ${INPUT_CLASS}`}
-        />
-      </div>
+      <Select
+        label="Mode"
+        value={mode}
+        onChange={(next) => setMode(next as Mode)}
+        options={MODE_OPTIONS}
+        reserveMessage={false}
```
khaledosman: This pair is inverted. This field has a `description` but passes `reserveMessage={false}`; the Endpoint field at line 582 has no description and omits the prop, which reserves anyway because `FieldMessages` defaults `reserve` to true.

`feedback.md`: a field reserves its message line "only where it has a description … so a field with nothing to say under it holds nothing", and "Spell that as `reserveMessage={false}` rather than by leaving the prop off, which currently reserves anyway … Measured, a bare field in a dialog is 83px against the 60px it should be."

Drop the prop here; add it at 582.

web/src/features/tools/OrganizationGuardrailsCard.tsx
```diff
@@ -517,57 +538,59 @@ function AddGuardrailForm({
       },
       {
         onSuccess: () => {
-          setProfile("")
-          setUrl("")
-          setCredential("")
-          setScope([])
-          parameters.reset()
           onSaved(`${named} added`)
+          onClose()
```
khaledosman, line 548: Removing this `onSuccess` branch's `parameters.reset()` leaves `useParameterForm`'s `reset` (line 218) with no caller anywhere in `web/src` — the remount does that job now. Neither Biome nor `tsc` flags an unused object member. Drop `reset` and the docstring above it, whose stated case ("for the caller that clears the rest of its fields itself") no longer exists; AGENTS.md asks for the comments around a change to come out shorter than you found them.

web/src/design-system/layout/SettingsGroup.tsx
```diff
@@ -34,6 +35,13 @@ export function SettingsGroup({
   title?: string
   /** Shown beside the title where a group's size is worth knowing up front. */
   count?: number
+  /**
+   * The one thing this group is created into, on its heading row and aligned
+   * to the right, the way `PageIntro`'s does on a page's. A group that owns a
+   * collection is where that collection is added to, so the control belongs
+   * beside the heading naming it rather than at the foot of its rows.
+   */
+  action?: ReactNode
```
khaledosman, line 44: New prop and a new heading-row wrapper with `SettingsGroup.test.tsx` untouched; a story alone is not coverage. frontend-standards: "Add a Vitest test for any component, hook, or helper whose behavior you change (`Foo.tsx` → `Foo.test.tsx`, colocated)."

Two cases worth pinning: the action renders in the heading row in both the `bounded` and unbounded shapes, and an `action` with no `title`/`description` still produces the heading row — that `headingRow === null` guard is the branch this change moved.

khaledosman (changes requested):
Reviewed the incremental range only (`1e3fcc09..b0782a57`, 6 files): this branch is based on `main` but sits on top of #1068, so its diff against `main` carries the whole stack.

The `action` slot on `SettingsGroup` is optional and mirrors `PageIntro`'s heading row, so no untouched call site is affected; the `key={openCount}` fresh-draft shape and the mutation-below-the-key rule are both honored, and both create mutations invalidate their lists.

Blocker is a leftover debug block that also removed the assertion its test exists to make. The two `isDirty` findings are the same issue CodeRabbit raised on #1074; noted here because this is the PR that owns the code.

web/src/features/tools/OrganizationGuardrailsCard.test.tsx
jigjigjig: Caught a bad commit of mine, and the assertion you asked for settles a question I had already answered wrongly in the open.

Fixed at f7442cfb. The log is gone and the case asserts the required field, the refusal and a dialog still open.

The substance: reading it properly shows the native validation *is* what stops this. `form.checkValidity()` is false at the press and the blank required input is what `:invalid` matches, so the submit never reaches `parameters.check()` and #972's own message never speaks. I had diagnosed exactly that earlier, then talked myself out of it after finding a second bug in my own probe (two `userEvent` instances, which drops the press) and reported that the native theory was wrong. Both were true; the second one hid the first. The case now says in a comment that this is a change from the always-open form, where the Add button called the check directly.

You are right about the eslint directive too. Biome takes `// biome-ignore lint/…: reason`, so that line was inert, and `noConsoleLog` not being in recommended is why CI stayed green over it.

web/src/features/tools/OrganizationGuardrailsCard.tsx
jigjigjig: Fixed at f7442cfb. The dialog is ahead of the rows in both cards now, and the comment on it says why, since the reason is invisible from the call site: the trigger slot is a real element, so a dialog mounted last takes `:last-child` off the final row and puts a divider above the container's own edge.

SearchToolsCard already had it in the right place, which is what made this worth fixing rather than arguing: two cards, one pattern, and only one of them wrong.

web/src/features/tools/SearchToolsCard.tsx
jigjigjig: Fixed at f7442cfb. `const draft = JSON.stringify({ name, provider, apiBase, apiKey })` against a `useRef` seed, with the predicate gone.

Your framing is the right one: the provider was not an oversight I could have avoided by being careful, it is what a hand-listed predicate costs. The snapshot is what stops the next field being forgotten in the same way.

web/src/features/tools/OrganizationGuardrailsCard.tsx
jigjigjig: Fixed at f7442cfb, all eight fields including the parameter values and the raw JSON, and the hand-rolled empties check over `parameters.values` went with it.

Worth naming what that predicate was missing, since it is more than the review says: mode, everywhere and scope, and also that a parameter typed into the section reported clean. A form whose only edits were parameters was one Escape from silence.

web/src/features/tools/OrganizationGuardrailsCard.tsx
jigjigjig: Fixed at f7442cfb: the import is `@/design-system/actions/Button` and both this call site and `GuardrailRow`'s go through it.

web/src/features/tools/OrganizationGuardrailsCard.tsx
jigjigjig: Fixed at f7442cfb: the prop is off the Mode select, which has a description, and on the Endpoint field, which has nothing to say under it.

I had it exactly inverted, and the reason is worth recording: leaving the prop off reads like the neutral choice and is not, because `FieldMessages` reserves by default. Both of mine were wrong in the direction that costs height.

web/src/design-system/layout/SettingsGroup.tsx
jigjigjig: Fixed at f7442cfb, two cases.

The first renders an action in both the bounded and unbounded shapes and asserts it shares the heading's row rather than following the band, so the two returns cannot drift. The second is the branch this PR actually moved: an action with no title still renders the row. That one fails against the old `heading === null && blurb === null` guard, which I checked by restoring it.

web/src/features/tools/OrganizationGuardrailsCard.tsx
jigjigjig: Fixed at f7442cfb: `reset` and its docstring are gone.

You are right that nothing flags it. The docstring was the tell and I left it standing: it argued for a caller that clears the rest of its fields itself, and the remount is what does that now, so the comment was describing a shape the PR had removed.

web/src/features/tools/OrganizationGuardrailsCard.tsx
```diff
@@ -517,57 +525,59 @@ function AddGuardrailForm({
...
+      size="lg"
+      title="Mandated guardrail"
+      submitLabel="Mandate a guardrail"
+      onSubmit={submit}
+      isPending={create.isPending}
+      isSubmitDisabled={profile.trim() === ""}
+      isDirty={draft !== seed.current}
+      error={create.error}
+    >
       <GuardrailProfileField
```
jigjigjig, line 572: `GuardrailProfileField` renders a `FilterSelect` (lines 47 and 100 of that file), and this move puts it inside a `FormDialog`. forms.md makes the two halves distinct on purpose: `FilterSelect` wears a caption label beside the control and never speaks a validation message, where the form half puts the label above, keeps it visible, and owns a description and an error announced on the control. Here it sits between `Select` and `Field` that do exactly that, so the first control in the dialog is the one that reads differently, and it is the only required one.

The component has a single call site (this dialog), so the fix does not reach the rows. Replace both `FilterSelect`s with `Select`: `label="Guardrail profile"`, `placeholder="Choose a profile"` (which retires the `{ value: "", label: "Choose a profile" }` sentinel option), the current caption as `description` (which retires the hand-rolled `FieldMessages` + `span`), `isDisabled={pending || disabled}` for the reading state with `placeholder="Reading the guardrails service…"`, and `isRequired`. Give the by-hand `Field` `isRequired` too. With `required` on the control, `isSubmitDisabled={profile.trim() === ""}` at line 549 can go: `FormDialog`'s docstring asks to reserve that prop for what the browser cannot check, because `required` also says which field.

web/src/features/tools/OrganizationGuardrailsCard.tsx
```diff
@@ -517,57 +525,59 @@ function AddGuardrailForm({
...
+        placeholder="blank uses the guardrails URL above"
+        reserveMessage={false}
+      />
+      <SecretField
+        label="Credential"
+        value={credential}
+        onChange={setCredential}
+        description="Needs an https endpoint of its own, since the URL above may be a plain-http sidecar, and OTARI_SECRET_KEY set on the gateway."
+      />
       <WorkspaceScope
```
jigjigjig, line 598: Same rule, second instance: `WorkspaceScope`'s "Runs in" is a `FilterSelect` (line 109), now inside the form with a caption label beside it while Mode, Endpoint and Credential above carry theirs above the control. Unlike the profile field this component is shared with `GuardrailRow`, whose controls this PR leaves alone by design, so the fix is a shape rather than a swap: either give `WorkspaceScope` a prop that renders the form half (`Select label="Runs in"`) and pass it here, or render the `Select` and the checkbox `fieldset` directly in `AddGuardrailDialog` and keep `WorkspaceScope` for the row. The checkbox group itself is fine; it is the picker that is the toolbar component.

web/src/features/tools/SearchToolsCard.tsx
```diff
@@ -233,100 +242,88 @@ function AddToolForm({ providers }: { providers: SearchProviderInfo[] }) {
...
+        placeholder={inherited ?? "https://…"}
+        description={
+          inherited
+            ? `Leave blank to inherit ${inherited}.`
+            : baseRequired
+              ? "This provider has no endpoint of its own, so it needs one here."
+              : "Optional for this provider."
+        }
+      />
+      <SecretField
```
jigjigjig, line 322: The API key is the one field whose requirement the disabled submit is hiding. Name has `isRequired`, Backend URL has `isRequired={baseRequired}`, but this `SecretField` (which supports `isRequired`) never says it is required when `keyRequired` is true, so an operator adding an exa tool sees a dimmed "Add search tool" with nothing on the form explaining which field is missing. `FormDialog.isSubmitDisabled`'s docstring: "`required` on a field is better than a boolean here, because it also says which field."

Add `isRequired={keyRequired}` here. Once every requirement is on its field, `isSubmitDisabled={!ready}` only needs to carry what `required` cannot (a whitespace-only name), or can go entirely and let the form's own validation refuse.

web/src/design-system/layout/SettingsGroup.tsx
```diff
@@ -69,15 +77,24 @@ export function SettingsGroup({
       </div>
     )
 
+  // The heading row: the words on the left, the action on the right, stacked
+  // on a phone where there is no room for two columns. Rendered here rather
+  // than at each call site so a group's action lands in one place.
+  const headingRow =
+    heading === null && blurb === null && action === undefined ? null : (
```
jigjigjig: The guard tests `action === undefined` while the render at line 90 tests `action ?`. Both new callers pass `cond ? <Button/> : null`, so a `null` action reaches this guard as present: with no `title` and no `description`, `headingRow` renders a flex row holding an empty inner `div`, and in the unbounded shape that becomes a whole `Section` with `border-t` and `pt-6 pb-3` framing nothing. Today both callers always pass a title so it is latent, but the new test "renders the heading row for an action alone" pins this branch, and the primitive should not depend on the caller spelling absence as `undefined`.

Make the guard match the render: `heading === null && blurb === null && !action`.

web/src/features/tools/OrganizationGuardrailsCard.tsx
```diff
@@ -517,57 +525,59 @@ function AddGuardrailForm({
...
+      isPending={create.isPending}
+      isSubmitDisabled={profile.trim() === ""}
+      isDirty={draft !== seed.current}
+      error={create.error}
+    >
       <GuardrailProfileField
         catalog={catalog}
         pending={catalogPending}
         value={profile}
         disabled={create.isPending}
```
jigjigjig: Half the dialog dims during a submit and half does not. `disabled={create.isPending}` is passed to the profile field (here), Endpoint (571), `WorkspaceScope` (586) and the parameters section (608), but not to the Mode `Select` or the Credential `SecretField`, so mid-submit four controls drop to the 0.4 disabled treatment while two keep their fill. No other `FormDialog` consumer in `web/src/features` disables its fields on `create.isPending`, and `SearchToolsCard`'s dialog beside this one does not either: `FormDialog` already blocks the submit, disables Cancel and the close control, and says `aria-busy` on the form, and actions.md is explicit that 0.4 has to read as denied rather than as working.

Drop the four `disabled={create.isPending}` props so this dialog matches its sibling and the dialog's own pending model.

web/src/features/tools/SearchToolsCard.tsx
```diff
@@ -233,100 +242,88 @@ function AddToolForm({ providers }: { providers: SearchProviderInfo[] }) {
...
-          {error}
-        </p>
-      ) : null}
-    </div>
+    <FormDialog
+      isOpen={isOpen}
+      onOpenChange={(open) => {
+        if (!open) onClose()
+      }}
+      size="sm"
```
jigjigjig: `size="sm"` is 440px and feedback.md reserves it "for one or two fields"; this dialog has four (Name, Provider, Backend URL, API key), three of them carrying a description line, and the Backend URL description interpolates an inherited URL that will wrap at 440. `md` (520) is the default for this shape. Drop the `size` prop.

jigjigjig (commented):
Reviewed the full head (45c56f56) against `main` from a fresh worktree: read all seven files, `pnpm --dir web run lint` and `run typecheck` clean, the three touched vitest files plus `architecture.test.ts` and `foundation.test.ts` at 3277 passing, and `.storybook/smoke.mjs` exit 0 with the new story in it. Series rules verified as holding: trigger and submit share their label in both dialogs, draft and mutation sit below `key={openCount}`, `isDirty` is a seeded snapshot, one `ErrorBanner` per dialog, no `returnFocusRef` owed (neither trigger disappears after a create), and `reserveMessage` follows the description in every field. I also checked that the refusal test's negative assertions are load-bearing by filling the required field in a scratch copy, which fails the test as it should.

Eight threads below, correctness first. The two that matter most are the `FilterSelect`s that ended up inside the form when the guardrail fields moved into `FormDialog`.

Review by Claude Code (Fable 5.1).

web/src/features/tools/SearchToolsCard.tsx
```diff
@@ -359,7 +358,29 @@ export function SearchToolsCard({ docsHref }: { docsHref: string }) {
...
+        ) : null
+      }
     >
+      {/* Keyed on the open count, so each open remounts a blank form. Clearing
+          the draft on close instead would blank the fields while the dialog is
+          still animating away. */}
+      <AddToolDialog
+        key={openCount}
+        isOpen={adding}
+        onClose={() => setAdding(false)}
```
jigjigjig, line 373: A successful create now lands its row inside a `DisclosureRow` that is collapsed by default, while the trigger sits on the group heading outside it. feedback.md's copy rule is "Nothing. The row changed; that is the confirmation", but with the drill-in closed the only thing that changes on screen is the trailing count ("1 tool" to "2 tools"), so an operator who has never opened the row sees the dialog close and nothing else. The old in-list form did not have this problem because it lived inside the disclosure.

Open the disclosure on success: have `AddToolDialog` take an `onCreated` alongside `onClose` and pass `() => { setAdding(false); setIsOpen(true) }` here, keeping `onClose` for cancel and the dirty-guard discard.

web/src/features/tools/OrganizationGuardrailsCard.tsx
```diff
@@ -656,6 +662,27 @@ export function OrganizationGuardrailsCard({
       {manages ? (
         <>
           <ErrorBanner error={guardrails.error ?? workspaces.error} />
+          {/* Ahead of the rows, not after them. `FormDialog` renders its
```
jigjigjig: Two call sites now carry a positional constraint ("ahead of the rows") and a comment explaining `divide-y` and `:last-child`, because `FormDialog`'s hidden `Modal.Trigger` is a real child of `SettingsGroup`'s `divide-y` container. The constraint is only needed while the dialog lives inside that container. `KeysPage` and `OrganizationDomainsPage` mount theirs in the page column instead, where a `display: none` div takes part in nothing.

Render `AddGuardrailDialog` (and `AddToolDialog` at `SearchToolsCard.tsx:378`) as a sibling of `SettingsGroup` inside a fragment, or in the `action` slot beside the button that opens it, which also mounts it exactly when it can be opened. Both cards then lose the ordering rule and this nine-line comment, which AGENTS.md's "leave the comments around a change shorter than you found them" would rather not have to carry.

web/src/features/tools/OrganizationGuardrailsCard.tsx
jigjigjig: Done at 8ef801d0, and it went further than a swap in one place.

Both `FilterSelect`s in `GuardrailProfileField` are `forms/Select` now, with `label`, `isRequired`, `isDisabled`, the caption as `description` (the hand-rolled `FieldMessages` and its span are gone), and the `{ value: "", label: "Choose a profile" }` sentinel replaced by `placeholder`. The by-hand `Field` takes `isRequired` too.

One deviation, measured rather than chosen. The waiting state could not use `placeholder="Reading the guardrails service…"`: HeroUI's trigger renders the selected option's text through `selectedText ?? placeholder`, and a `value` of `""` is a selection of the empty string rather than no selection, so the placeholder never reaches it and the trigger came out with no text at all. I found that by reading the trigger's accessible name, which was just "Guardrail profile". The waiting sentence is the `description` instead, where it renders and is announced. Worth knowing for the next `Select` that wants a placeholder while empty.

`isSubmitDisabled={profile.trim() === ""}` stays for now: with the parameter fields' `required` removed (see the parameters thread) the form's own validation no longer refuses a blank profile on its own, and I would rather the submit be visibly refused than the press do nothing. Say if you would rather it go.

web/src/features/tools/OrganizationGuardrailsCard.tsx
jigjigjig: Done at 8ef801d0, by the first of your two shapes.

`WorkspaceScope` takes a `variant` of `"filter"` or `"form"`, defaulting to filter so the rows are untouched, and the dialog passes `"form"`, which renders `Select label="Runs in"`. The checkbox fieldset is shared unchanged, since as you say it was only the picker that was the toolbar component.

I took the prop rather than lifting the controls into the dialog because the fieldset's `scopeName` labelling and the toggle logic would have had to be duplicated to do it the other way, and that is the part that is easy to get subtly different between the two copies.

web/src/features/tools/SearchToolsCard.tsx
jigjigjig: Done at 8ef801d0: `isRequired={keyRequired}`.

I left `isSubmitDisabled={!ready}` in place rather than reducing it to the whitespace case. It still carries something `required` cannot: a name of only spaces, and the backend-URL rule that depends on which provider is picked and whether one is inherited. Happy to narrow it if you would rather, but it is no longer the only thing saying a field is missing, which was the substance here.

web/src/design-system/layout/SettingsGroup.tsx
jigjigjig: Done at 8ef801d0: the guard is `heading === null && blurb === null && !action`.

You are right that it is latent today and right that it should not depend on how a caller spells absence. Both of mine pass `cond ? <Button/> : null`, which is the spelling that would have hit it.

web/src/features/tools/OrganizationGuardrailsCard.tsx
jigjigjig: Done at 8ef801d0: all four are gone.

Worth recording why it was inconsistent rather than wrong-by-halves: I passed `disabled` to the components that already had the prop and not to the two that would have needed it added, so the split was an artifact of the existing signatures rather than a decision. Removing them made `disabled` optional on `WorkspaceScope`, `GuardrailProfileField` and `GuardrailParametersSection`, which is the right shape anyway now that nothing dims on pending.

web/src/features/tools/SearchToolsCard.tsx
jigjigjig: Done at 8ef801d0: the `size` prop is off, so it takes the `md` default.

Your reading of the spec is the one I should have applied when I wrote it: I picked `sm` from the shape of the old inline row, which was four compact machine fields, rather than from what the dialog became once each field carried a visible label and a description.

web/src/features/tools/SearchToolsCard.tsx
jigjigjig: Done at 8ef801d0. `AddToolDialog` takes `onCreated` alongside `onClose`; the card passes one that closes the dialog and opens the drill-in, and `onClose` still just closes, so a cancel or a discard leaves the disclosure as it was.

New case asserts the disclosure's own `aria-expanded` after a create, not a row's text. My first version asserted the row and passed with the fix removed, because the disclosure's content is in the DOM whether it is open or not; that is the sort of green this would have shipped as.

web/src/features/tools/OrganizationGuardrailsCard.tsx
jigjigjig: Done at 8ef801d0. Both dialogs are siblings of `SettingsGroup` inside a fragment now, and the nine-line comment and the ordering rule are gone from both cards.

I mounted them outside rather than in the `action` slot. Either satisfies the constraint; the fragment keeps the slot holding one control, which is what its own docstring says it is for, and it reads the way `KeysPage` and `OrganizationDomainsPage` already do.

web/src/features/tools/OrganizationGuardrailsCard.tsx
jigjigjig: One more on this thread, for the parameters half that the #1073 review raised against the same files: the parameter fields no longer pass `isRequired`.

Inside a `FormDialog` the submit is a real form submission, so a natively required field refuses it before `parameters.check()` runs, and the app's own message never speaks. That is what I had measured earlier and then documented as the behavior; the ruling since is that there should be one path and it should be the app's. So `required` comes off those two fields, `parameterErrors` is the only judge in both the row and the dialog, and the case asserts the message rather than the refusal.

The cost, stated plainly: those inputs lose the `required` marker. The refusal still lands on the field and is announced, through `isInvalid` and `errorMessage`, and there is a comment at the top of `GuardrailParameterFields` saying why it is that way.

PR #1073, opened 2026-09-10
refactor(dashboard): open every page with PageIntro and retire PageHeader

jigjigjig:
## Description

Three pages still opened with the retired page header, which put their action on a left-aligned row underneath the title instead of beside it. They now use `PageIntro` like every other page, and `shared/components/deprecated/PageHeader.tsx` is deleted with them: it was the last of its rows on the deprecation gate.

MCP servers gets the more interesting version of that move. The card is the whole page there, so it renders the page's opening itself under a `variant` prop, and keeps its own heading at the foot of `/tools` where it is one section among several. Either way the register control sits on the heading beside the words naming what it registers, rather than in a row of its own.

Workspace members had an add form permanently open under the roster: two pickers and a button, always there whether anyone was being added or not. That is now "Add member" on the page heading, opening a small dialog. Its two pickers move from the toolbar select to the form one, so they carry the same label style as a field rather than a caption a rung lighter. The sentence explaining that everyone in the organization is already in this workspace is said inside the dialog now, rather than standing in for a form that is not there.

Spend & budgets keeps its two Add controls where they are. The page holds two collections, budgets and ceilings, and one action slot on the page title could only name one of them, so each stays on its own section heading.

Part of #1031

## How to test it locally

1. `make dashboard && make dev`, then open Tools, MCP servers. The title row carries "Add MCP server"; the same card at the foot of `/tools` still has its own heading and button.
2. Open the workspace Members page. "Add member" is in the heading row and opens "New workspace member". On a standalone deployment, where the one identity is already in the workspace, the dialog says so instead of offering an empty picker.
3. Open Organization, Spend & budgets. Both "Add budget" and "Add ceiling" are still on their own section headings.

Covered by `WorkspaceMembersPage.test.tsx`, which now drives the add flow through the dialog and gained a case for the trigger staying visible and the draft resetting on open. That reset half was checked by deleting the remount key and reading the failure. `parity.tenancy.spec.ts` follows the "already in this workspace" sentence into the dialog.

## PR Type

- [x] Refactor

## Relevant issues

Part of #1031, and takes two more of the sites listed on #1043.

## Checklist

  Dashboard counterparts: `pnpm --dir web run lint` and `run typecheck` clean, the full vitest suite at 5477 passing across 157 files, and the four Playwright behavioral projects at 43 passing with no failures.
- [ ] Documentation was updated where necessary. The deprecation gate is the documentation here, and it is updated: `PageHeader`'s three call sites and its replacement row are gone from `deprecated.test.ts` along with the file.

## AI Usage

**AI Model/Tool used:**

Claude Code (Opus 5)

The reset-on-open case was verified by deleting the fix and reading the failure rather than by passing on first write.

## Summary

- Migrated the remaining pages from `PageHeader` to `PageIntro`.
- Moved page actions into headings and dialogs.
- Added dialog-based flows for workspace members, search tools, and guardrails.
- Added MCP page-mode headings and preserved separate budget actions.
- Added tests, stories, and documentation updates.
- Removed the deprecated `PageHeader` component and its deprecation checks.

## Benefits

- Keeps page actions close to their headings.
- Provides cleaner, less crowded pages.
- Resets forms when dialogs reopen.
- Handles unsaved changes and unavailable member candidates more clearly.
- Removes the final `PageHeader` migration debt.


web/src/design-system/layout/PageIntro.stories.tsx
```diff
@@ -10,7 +10,7 @@ import { PageIntro } from "./PageIntro"
  * be larger than the page's own name. That is why a 30px KPI figure does not
  * fight a 28px title, and why a second `text-display` on one route is a bug.
  *
- * Replaces `deprecated/PageHeader`, which is still on four pages.
+ * Replaced `deprecated/PageHeader`, which is gone: every page opens with this.
```
khaledosman, line 13: This docstring was updated; the normative docs were not, and they are what the next reader reaches for. Still live at head, all describing a component this PR deletes:

- `web/design/DESIGN.md:214` — the row `| deprecated/PageHeader | layout/PageIntro | 3 pages, one use each |` in the "must not be used in new code" table
- `web/design/layout.md:65-66` — "`PageHeader` is the pre-redesign card version still live on 4 pages"
- `web/design/module-map.md:163` — maps `PageHeader` to `deprecated/PageHeader.tsx`, a file that no longer exists
- `web/design/module-map.md:237` — "`PageHeader` | 4 pages"
- `web/design/module-map.md:361` — "Convert the remaining deprecated call sites. `PageHeader` on 4 pages"
- `.github/skills/frontend-standards/components.md:169` — "**Not** `PageHeader`, which is in `deprecated/`"
- `components.md:153` and `component-architecture.md:64` — both say `deprecated/` is "the four that must not be used"; it is now three (`RowActions`, `SettingsSection`, `StatCard`)
- `web/src/design-system/feedback/EmptyState.tsx:19` — "Kept to a string (like `PageHeader`)"

Rungs: `component-architecture.md:66-68`, "`web/design/DESIGN.md` maps every export to its module and is the inventory ... add it to that table in the same change" — deleting one is the symmetric case. And root `AGENTS.md`: "a fact told in two layers is a fact that will go stale in one of them, and the stale copy is the one someone believes."

The PR body's "the deprecation gate is the documentation here" does not cover this: the gate derives call sites, it does not remove a row for a module that no longer exists.

(Anchored here because those doc files are outside this PR's diff.)

web/src/features/workspaces/WorkspaceMembersPage.tsx
```diff
@@ -1,14 +1,20 @@
+import { Button } from "@heroui/react"
```
khaledosman: New direct HeroUI `Button` import. Use `import { Button } from "@/design-system/actions/Button"`.

`actions.md:12-20`: "**Import `Button` from `@/design-system/actions/Button`, not from `@heroui/react`.** It is the same component with the union narrowed to the three, which turns this rule into a compile error at the call site" — and it closes with "That gate stays, because it still covers the direct imports this tree has not converted; new code should not need it." This is new code.

web/src/features/workspaces/WorkspaceMembersPage.tsx
```diff
@@ -28,6 +34,13 @@ export function WorkspaceMembersPage() {
   const { selected, isLoading } = useSelectedWorkspace()
   const context = useOrganizationContext()
   const orgMembers = useOrganizationMembers()
+  // The same query the panel below reads, so this costs no second request. It
+  // is here because who is *already* in the workspace is what makes the rest of
+  // the organization a candidate list, and the control that uses that list now
+  // sits in the heading row.
+  const members = useWorkspaceMembers(selected?.workspace_id ?? null)
+  const [adding, setAdding] = useState(false)
```
khaledosman: `adding` → `isAdding` (and `setIsAdding`). `naming-conventions.md:22-26`: booleans "read as a predicate, a yes/no question in English ... a bare noun phrase reads as the thing itself rather than as a question about it". `feedback.md`'s own `FormDialog` example uses `isCreating` for exactly this state.

web/src/features/workspaces/WorkspaceMembersPanel.tsx
```diff
@@ -55,72 +66,70 @@ function AddWorkspaceMember({
...
-  if (candidates.length === 0) {
-    return (
-      <InfoBanner>
-        Every active member of this organization is already in this workspace. A
-        workspace's members are always a subset of the organization's, so add
-        someone there first, on the Members page.
-      </InfoBanner>
-    )
-  }
+  const nobodyLeft = rosterResolved && candidates.length === 0
```
khaledosman: `nobodyLeft` → `isNobodyLeft` (or `hasNobodyLeft`). Same rung: `naming-conventions.md:22-26`, booleans read as a predicate, "including props and the boolean fields of a hook's return value".

web/src/features/tools/WorkspaceMcpServersCard.tsx
```diff
@@ -55,11 +56,21 @@ function tokenChip(server: WorkspaceMcpServer) {
...
-  /** Suppressed on the page whose own title already says "MCP servers". */
-  showHeading?: boolean
+  /**
+   * Where this is rendering. As a `card` it is one section at the foot of
+   * `/tools`, with its own `h2` and the register control beside it. As a
+   * `page` it *is* `/tools/mcp-servers`, so it renders that page's opening
+   * instead and the control sits in the heading row there, which is where
+   * every other page in the dashboard keeps the one thing it creates.
+   */
+  variant?: "card" | "page"
```
khaledosman, line 73: The `page` variant is the only thing that renders `/tools/mcp-servers`' `h1`, its description, and the Add control in the heading row, and nothing covers it: all 14 cases in `WorkspaceMcpServersCard.test.tsx` render the default `card` variant, and there is no `McpServersPage.test.tsx`.

Add one case rendering `variant="page"` asserting `getByRole("heading", { level: 1, name: "MCP servers" })` and that "Add MCP server" is present, so a later edit cannot silently drop that page's only `h1`. Rung: frontend-standards SKILL.md, "Add a Vitest test for any component, hook, or helper whose behavior you change".

khaledosman (changes requested):
Reviewed the incremental range only (`b0782a57..6361b7bf`, 11 files): this branch is based on `main` but sits on top of #1071, so its diff against `main` carries the whole stack.

The retirement itself is clean. No code, story, or test references `PageHeader` any more, there is no barrel to clean up, `deprecated.test.ts` still gates three real modules so its "name a replacement for every module" assertion is not vacuous, and the `h1` survives on all three converted pages through `PageIntro`.

What is missing is the other half of the deletion: eight places still document `PageHeader` as an existing component on live pages, two of them by counting `deprecated/` as four modules when it now holds three.

web/src/design-system/layout/PageIntro.stories.tsx
jigjigjig: Fixed at 2e51a581, all eight, and the count too.

Gone: DESIGN.md's do-not-use row, layout.md's page recipe sentence, module-map's export map row, its conversion table row and its remaining-work bullet, components.md's page-title row, and EmptyState's aside. `component-architecture.md:64` and `components.md:153` now say `deprecated/` holds three, which is what `ls` says: RowActions, SettingsSection, StatCard.

You are right that the PR body's line did not cover it, and the reason is worth stating: the gate derives call sites from the tree, so it removes a row when a caller converts and says nothing when a module goes. Those are different events and it only sees one of them.

web/src/features/workspaces/WorkspaceMembersPage.tsx
jigjigjig: Fixed at 2e51a581. It imports from `@/design-system/actions/Button` now.

New code, so no exemption applies; the direct import was mine reaching for the nearest thing rather than the right one.

web/src/features/workspaces/WorkspaceMembersPage.tsx
jigjigjig: Fixed at 2e51a581: `isAdding` and `setIsAdding`.

web/src/features/workspaces/WorkspaceMembersPanel.tsx
jigjigjig: Fixed at 2e51a581: `isNobodyLeft`.

While in that function I also took the dirty check the survey flagged: it watched only `userId`, so a role picked on its own was thrown away by Cancel with no guard. It is one snapshot naming both fields now, with a case that presses only the Role and asserts the guard appears. That case fails against the old predicate.

web/src/features/tools/WorkspaceMcpServersCard.tsx
jigjigjig: Fixed at 2e51a581, two cases rather than one.

`renderCard` takes the variant now. The first renders `variant="page"` and asserts the `h1`, the description and the register control; the second renders the default and asserts an `h2` and no `h1` at all, so the two cannot drift into each other. You were right that nothing covered this: fourteen cases, all on the default, and no test file for the page.

While there I also fixed a functional one the survey found on this page, since it is the same dialog: the candidate list was the organization roster minus a workspace roster that had not necessarily answered, so mid-load it offered exactly the people it was about to exclude. It waits for both reads now. The case for it holds the workspace read open and asserts the person already in the workspace is not offered; it fails on the old code.

web/src/features/workspaces/WorkspaceMembersPanel.tsx
```diff
@@ -55,72 +66,75 @@ function AddWorkspaceMember({
...
-        workspace's members are always a subset of the organization's, so add
-        someone there first, on the Members page.
-      </InfoBanner>
-    )
-  }
+  const isNobodyLeft = rosterResolved && candidates.length === 0
+  // Every field the operator can change, against what the form was seeded
+  // with, rather than the one that gates the submit: a role picked on its own
+  // is work, and a guard that only watches the person loses it silently.
+  const seeded = useRef(JSON.stringify({ userId: "", role: "member" }))
```
jigjigjig: This hand-rolls what `useDirtySnapshot` (`design-system/forms/useDirtySnapshot.ts`) already is, and `KeysPage`, `BudgetsPage` and `WorkspacesPage` reach for the hook. feedback.md: "`isDirty` comes from `useDirtySnapshot`: hand it every field the form owns and it snapshots them on mount." It also spells the seed a second time (`{ userId: "", role: "member" }`) instead of snapshotting the state's own initial values, so the two drift the day a default changes.

Fix: `const { isDirty } = useDirtySnapshot({ userId, role })`, drop `useRef`, and pass `isDirty` through.

web/src/features/workspaces/WorkspaceMembersPanel.tsx
```diff
@@ -55,72 +66,75 @@ function AddWorkspaceMember({
...
+          add someone there first, on the Members page.
+        </InfoBanner>
+      ) : (
+        <>
+          <Select
+            label="Organization member"
+            value={userId}
+            onChange={setUserId}
+            options={[
+              { value: "", label: "Select a member…" },
```
jigjigjig: `Select` treats `value === ""` as "nothing selected" and shows `placeholder` for it (see its docstring on why it deliberately does not synthesize a `""` row). Carrying `{ value: "", label: "Select a member…" }` as an option puts that sentinel in the popover as a pickable row, which is the exact case the component's comment says it was written to avoid. It was correct on `FilterSelect`, where `""` is a real choice; it is not here.

Fix: drop the sentinel from `options` and pass `placeholder="Select a member…"`.

web/src/features/workspaces/WorkspaceMembersPanel.tsx
```diff
@@ -55,72 +66,75 @@ function AddWorkspaceMember({
...
+    >
+      {isNobodyLeft ? (
+        <InfoBanner>
+          Every active member of this organization is already in this workspace.
+          A workspace's members are always a subset of the organization's, so
+          add someone there first, on the Members page.
+        </InfoBanner>
+      ) : (
+        <>
+          <Select
```
jigjigjig, line 101: feedback.md: "The first field takes `autoFocus`." Nothing in this dialog does, so on open focus rests on the frame rather than in the form, and the first Tab reaches the header's Close control before any field. `Select` does not expose the prop yet; `ComboBoxField` shows the shape (forward `autoFocus` to `HeroSelect.Root`, line 211 there).

Fix: add `autoFocus?: boolean` to `Select` and set it on "Organization member". `AddGuardrailDialog` has the same gap on its profile field.

web/src/features/tools/SearchToolsCard.tsx
```diff
@@ -233,100 +242,88 @@ function AddToolForm({ providers }: { providers: SearchProviderInfo[] }) {
   const baseRequired =
     Boolean(selected?.requires_api_base) && inherited === null
   const keyRequired = Boolean(selected?.requires_api_key)
+  // One snapshot rather than a hand-listed predicate: the provider is a field
+  // like the others, and a guard that forgot it discarded a changed provider
+  // with no question asked.
+  const draft = JSON.stringify({ name, provider, apiBase, apiKey })
+  const seed = useRef(draft)
+
```
jigjigjig, line 257: Same as the members dialog: this is `useDirtySnapshot` reimplemented inline. The comment gives the hook's own rationale, so the hook is the shape to use: `const { isDirty } = useDirtySnapshot({ name, provider, apiBase, apiKey })`, and drop the `useRef` import.

web/src/features/tools/OrganizationGuardrailsCard.tsx
```diff
@@ -501,8 +495,22 @@ function AddGuardrailForm({
...
+    profile,
+    mode,
+    url,
+    credential,
+    everywhere,
+    scope,
+    values: parameters.values,
+    extraJson: parameters.extraJson,
+  })
+  const seed = useRef(draft)
```
jigjigjig: Third copy of `useDirtySnapshot` in this PR. Hand the whole object to the hook (`const { isDirty } = useDirtySnapshot({ profile, mode, url, credential, everywhere, scope, values: parameters.values, extraJson: parameters.extraJson })`) and drop the `useRef`.

web/src/features/tools/OrganizationGuardrailsCard.tsx
```diff
@@ -517,57 +525,59 @@ function AddGuardrailForm({
...
+        placeholder="blank uses the guardrails URL above"
+        reserveMessage={false}
+      />
+      <SecretField
+        label="Credential"
+        value={credential}
+        onChange={setCredential}
+        description="Needs an https endpoint of its own, since the URL above may be a plain-http sidecar, and OTARI_SECRET_KEY set on the gateway."
+      />
       <WorkspaceScope
```
jigjigjig, line 598: `WorkspaceScope` renders a `FilterSelect` ("Runs in"), and this call site now puts it inside a `FormDialog`. forms.md keeps `FilterSelect` for a toolbar filter (caption label beside the control, no validation line) and `Select` for a form field; the PR body makes the same argument for moving the member pickers off `FilterSelect`, so the dialog ends up with one field labeled a rung lighter than the four above it.

Fix: give `WorkspaceScope` the form-style control (a `Select` with `label="Runs in"`), either always or behind a prop the row keeps on the filter shape.

web/src/features/tools/OrganizationGuardrailsCard.tsx
```diff
@@ -517,57 +525,59 @@ function AddGuardrailForm({
...
+      size="lg"
+      title="Mandated guardrail"
+      submitLabel="Mandate a guardrail"
+      onSubmit={submit}
+      isPending={create.isPending}
+      isSubmitDisabled={profile.trim() === ""}
+      isDirty={draft !== seed.current}
+      error={create.error}
+    >
       <GuardrailProfileField
```
jigjigjig, line 572: `GuardrailProfileField` renders a `FilterSelect` in both its pending and listed states, so the dialog's first and gating field is the toolbar control while Mode, Endpoint and Credential are form fields. Same rule as the `WorkspaceScope` thread below; the by-hand branch already uses `Field`, so the picker branch is the odd one.

Fix: swap the two `FilterSelect`s in `GuardrailProfileField` for `Select` (its `FieldMessages` caption becomes `description`), and give it `autoFocus` as the dialog's first field.

web/src/features/tools/OrganizationGuardrailsCard.tsx
```diff
@@ -517,57 +525,59 @@ function AddGuardrailForm({
...
+        value={mode}
+        onChange={(next) => setMode(next as Mode)}
+        options={MODE_OPTIONS}
+        description="A caller can tighten a mandated guardrail but never weaken it."
+      />
+      <Field
+        label="Endpoint"
+        value={url}
+        onChange={setUrl}
+        isDisabled={create.isPending}
```
jigjigjig: Half the form dims while the create is in flight (profile, Endpoint, `WorkspaceScope`, the parameters section take `create.isPending`) and half does not (Mode, Credential). feedback.md: disabled is the one 0.4 treatment and reads as *denied*; the `FormDialog` already blocks the submit and Cancel during `isPending`, and neither `AddToolDialog` nor `AddWorkspaceMemberDialog` disables fields for it.

Fix: drop the per-field `isPending` disabling here, or apply it to every field. Same either way, but not a mix.

web/src/features/tools/OrganizationGuardrailsCard.test.tsx
```diff
@@ -480,16 +511,32 @@ describe("OrganizationGuardrailsCard", () => {
   })
 
   it("refuses to add an entry whose guardrail needs a parameter it has not got", async () => {
+    // Inside a `FormDialog` the submit is a real form submission, so a
```
jigjigjig: This comment documents a divergence rather than a decision. `GuardrailParameterFields` sets `isRequired={spec.required}`, so inside the dialog's `<form>` a blank required text parameter is stopped by the browser's bubble and `parameters.check()` never runs, while `GuardrailRow`'s Save (a plain `onPress`) still reaches `check()` and shows "This guardrail needs a value here." The same omission now gets two different messages depending on whether the operator is adding or editing, and the assertion that used to pin the design's in-form message is gone.

Fix: pick one path. Either drop `isRequired` on the parameter fields so `check()` speaks in both places (and restore the message assertion), or keep native validation and remove the now-unreachable required branch from `parameterErrors` for text fields.

web/src/features/tools/WorkspaceMcpServersCard.tsx
```diff
@@ -195,16 +208,24 @@ export function WorkspaceMcpServersCard({
     })
   }
 
+  const addButton = manages ? (
+    <Button variant="primary" onPress={openAdd}>
```
jigjigjig, line 212: This `Button` is still `@heroui/react`'s (line 1). frontend-standards: reach for `@/design-system/actions/Button`, which narrows `variant` to the three that exist so a retired one is a compile error. `OrganizationGuardrailsCard` in this same PR made that swap; do it here too, since the file is already being touched and this is the one primary on the page.

web/design/module-map.md
```diff
@@ -234,7 +233,6 @@ Four members, two of which this spec is the first to name:
 
 | Component | Call sites | Replacement | Why |
```
jigjigjig, line 234: The sentence above this table still says "Four members, two of which this spec is the first to name"; with the `PageHeader` row removed the table has three. Change it to "Three members".

jigjigjig (commented):
Reviewed at 1d05478 from a fresh worktree against web/AGENTS.md, frontend-standards, and web/design (feedback.md, forms.md, actions.md, layout.md). `pnpm --dir web run lint`, `run typecheck`, the eight touched or gating vitest files (3300 passing), and `.storybook/smoke.mjs` against this branch's own Storybook (409 stories x 2 themes, all rendered) all pass locally; the Playwright change in `parity.tenancy.spec.ts` was not run here. Findings are inline, one thread each.

Review written by Claude (Fable 5.1) via Claude Code.

web/src/features/workspaces/WorkspaceMembersPanel.tsx
jigjigjig: Fixed at cda2a759: `useDirtySnapshot({ userId, role })`, with the hand-rolled `useRef` snapshot gone.

The case that pins it picks only the Role, presses Cancel and asserts the guard appears; it fails against a predicate watching `userId` alone.

web/src/features/workspaces/WorkspaceMembersPanel.tsx
jigjigjig: Fixed at cda2a759: the sentinel option is gone and the control takes `placeholder="Select a member…"`.

Worth noting what that also fixes, since it was invisible from the call site: an option keyed `""` is a real selectable row, so the list offered "Select a member…" as a choice and picking it set the value to empty.

web/src/features/workspaces/WorkspaceMembersPanel.tsx
jigjigjig: Fixed at cda2a759. `forms/Select` takes `autoFocus` now, wired the way `ComboBoxField` already had it, with a story in the catalog and a case asserting the picker holds focus when the dialog opens.

The case fails with the prop removed; I checked that rather than assuming, because focus assertions in jsdom have burned me twice today.

web/src/features/tools/SearchToolsCard.tsx
jigjigjig: Fixed on #1071 at 8ef801d0, which is merged into main as 5855af35, and this branch is rebased onto it.

Both hand-rolled snapshots there are `useDirtySnapshot` now, the guardrail one carrying the parameter values and the raw JSON as well as the six fields it had.

web/src/features/tools/OrganizationGuardrailsCard.tsx
jigjigjig: Fixed on #1071 at 8ef801d0, merged as 5855af35. The guardrail dialog's snapshot is the hook's, and it names every field including `parameters.values` and `parameters.extraJson`.

web/src/features/tools/OrganizationGuardrailsCard.tsx
jigjigjig: Fixed on #1071 at 8ef801d0, merged as 5855af35. `WorkspaceScope` takes a `variant`, the dialog passes `"form"` and gets `Select label="Runs in"`, and the rows keep the filter half unchanged.

web/src/features/tools/OrganizationGuardrailsCard.tsx
jigjigjig: Fixed on #1071 at 8ef801d0, merged as 5855af35. `GuardrailProfileField`'s two `FilterSelect`s are `forms/Select` with label, description, `isRequired` and `isDisabled`, and the `{ value: "", label: "Choose a profile" }` sentinel is a `placeholder`.

web/src/features/tools/OrganizationGuardrailsCard.tsx
jigjigjig: Fixed on #1071 at 8ef801d0, merged as 5855af35. All four `disabled={create.isPending}` props are off the dialog's fields, which also made `disabled` optional on the three components that took it.

web/src/features/tools/OrganizationGuardrailsCard.test.tsx
jigjigjig: Fixed on #1071 at 8ef801d0, merged as 5855af35, to the ruling that there is one validation path and it is the app's.

`required` comes off the parameter fields, so `parameterErrors` is the only judge in both the row and the dialog, and the case asserts the message rather than the refusal. The cost is stated in a comment at the top of `GuardrailParameterFields`: those inputs lose the required marker, and the refusal still arrives on the field through `isInvalid` and `errorMessage`.

web/src/features/tools/WorkspaceMcpServersCard.tsx
jigjigjig: Fixed at cda2a759: it imports from `@/design-system/actions/Button`.

web/design/module-map.md
jigjigjig: Fixed at cda2a759: "Three members".

Same slip as the two "four" counts in the skills earlier in this PR, and from the same cause: the sentence counted the directory as it was when it was written rather than as the change left it.

PR #1074, opened 2026-09-10
feat(dashboard): move the last six forms out of AlertDialog

### LINKED ISSUE #1031: One placement and one surface for every create action in the dashboard

jigjigjig:
## Description

An alert dialog interrupts to ask one question. A form is a place to work. Six forms in the dashboard were still built on the alert dialog, and they are the last ones: registering an MCP server, an organization's budget, a spend ceiling, a model's price, a rate override, and renaming the organization. All six now open in the shared form dialog, which is what the rest of this series has been moving every form to.

The shared alert-dialog shell they used to reach through goes with them. The rename was its last caller, so the component is deleted rather than left with none.

Each of the six also changes how its draft is cleared. They all seeded their fields from an effect that ran whenever the dialog opened, because the dialog stayed mounted between opens; now each seeds on mount alone and its caller remounts it on open. The visible difference is small but real: a form cleared on the way out blanks its own fields while the dialog is still animating away, and a form cleared on the way in does not. Six effects go with it.

Each form also states one predicate naming every field it holds, and the unsaved-changes guard is the negation of that. Written as a snapshot or a list assembled by hand, that check drifts from the form as fields are added; written once, naming them, it cannot.

One label changes: the MCP server dialog's submit reads "Add MCP server", the same words as the control that opens it, where it used to say "Add server".

A second commit takes three more of the pickers on #1043 to the form select. All three sit inside dialogs this PR moved, so each of those forms was showing two label styles side by side.

Closes #1031

## How to test it locally

1. `make dashboard && make dev`. Each of these opens a centered form dialog rather than an alert: Tools → MCP servers → "Add MCP server"; Organization → Spend & budgets → "Add budget" and "Add ceiling"; Organization → Rate overrides → the add control; Models → "Price a model by hand"; Organization → Org settings → "Change organization name".
2. In any of them, type something, press Escape, and discard. Reopen: the fields are empty, and they do not blank while the dialog is closing.
3. The budget and ceiling dialogs' pickers now carry the same label style as the fields beside them.

Covered by the existing suites, which drive all six flows and are updated to find a dialog rather than an alert dialog. The rename's reseed test now keys the dialog the way its page does, since remounting is the contract that replaced the effect.

Worth calling out for review: keying a dialog on an open counter is only correct if **every** opener bumps it. One of the price dialogs has two openers, and missing the second left a stale mount that seeded its model key from the previous open. `ModelsPage.test.tsx` caught it, which is the argument for having that assertion at all.

## PR Type

- [x] Refactor

## Relevant issues

Closes #1031. Takes three more of the sites listed on #1043.

## Checklist

  Dashboard counterparts: `pnpm --dir web run lint` and `run typecheck` clean, the full vitest suite at 5456 passing across 157 files, and the four Playwright behavioral projects at 43 passing.
- [ ] Documentation was updated where necessary. `overlays.md` already names the form dialog as where a form belongs and the alert dialog as where a confirmation does; this PR makes the code match it, and removes a component the docs describe as having no call sites of its own.

## AI Usage

**AI Model/Tool used:**

Claude Code (Opus 5)

No new tests were written for behavior the existing suites already covered; the ones that changed are the queries that named the surface, and the rename's reseed case, which had to be rewritten because the contract it tested moved from the component to its caller.

## Summary

- Migrated six dashboard forms from `AlertDialog` to the shared `FormDialog`.
- Added consistent dialog labels, validation, error handling, and unsaved-change protection.
- Reset form drafts on each open to prevent stale values and errors.
- Moved relevant save logic into the form dialogs.
- Added shared picker focus support and updated picker usage.
- Replaced remaining `PageHeader` usage with `PageIntro` and removed deprecated dialog and header components.
- Updated tests and stories for the new dialog behavior.

These changes provide a consistent creation and editing experience across the dashboard.


web/src/design-system/feedback/FormDialog.test.tsx
```diff
@@ -219,6 +219,87 @@ describe("FormDialog", () => {
...
+        </FormDialog>,
+      )
+
+      await user.click(screen.getByRole("button", { name: "Create key" }))
+      expect(onSubmit).not.toHaveBeenCalled()
+
+      await user.click(screen.getByLabelText("Instructions"))
+      await user.keyboard("{Meta>}{Enter}{/Meta}")
+      await user.keyboard("{Control>}{Enter}{/Control}")
+      expect(onSubmit).not.toHaveBeenCalled()
```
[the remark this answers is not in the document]
jigjigjig: Declining this one, with the measurement, and leaving the thread open rather than resolving it.

Two things are already true. A plain-Enter case from a single-line field exists at `FormDialog.test.tsx:132`, "submits on Enter from a single-line field, the way every form does", in the positive direction: click `Key name`, `{Enter}`, `onSubmit` called once. And the disabled direction cannot fail in jsdom, so adding it would be a case that looks like coverage and is not.

I wrote the suggested case and then tried to make it fail. Removing `isSubmitDisabled` from the submit funnel (`FormDialog.tsx:218`) leaves the entire suite green, 61 of 61, including the new case. The reason is that the clause is unreachable from both keyboard paths here: with the submit button rendered `disabled`, the browser performs no implicit submission at all, so plain Enter never reaches the form's submit event; and Cmd/Ctrl+Enter is gated before `requestSubmit` at `:232`, which is the gate the existing "blocks the button, plain Enter, and Cmd/Ctrl+Enter alike" case actually exercises. Neutering the handler outright fails seven cases, so the suite does test the submit path; it is this clause specifically that no case can reach.

So the clause stays as defense in depth, for a future path that reaches the funnel without passing the keydown gate, and its comment says that. What I am not willing to add is a test whose green means nothing: a case that passes with the fix deliberately removed is worse than no case, because the next reader takes it for a guard.

If you would rather see the disabled direction pinned at all, the honest place is the keydown gate, and that is the case that already exists.

web/src/features/budgets/OrganizationBudgetsCard.tsx
```diff
@@ -140,6 +146,7 @@ export function OrganizationBudgetsCard() {
       </Card>
 
       <OrganizationBudgetDialog
+        key={openCount}
```
khaledosman, line 133: The key remounts the fields but not the mutation, so a failed create's banner survives into the next fresh open: `create` and `update` are declared at lines 31-32, above the key, and `error={editing ? update.error : create.error}` (line 154) is never cleared on the way in.

`feedback.md` names this shape with this feature as its example — its Incorrect block is literally `<BudgetForm key={openCount} error={createBudget.error} … />`:

> A mutation above the key is the shape to watch for, because the draft looks right and the error does not: the key remounts the fields, the mutation keeps its state, and a failed create's banner survives into a fresh form.

Same in `SpendCeilingsCard.tsx:211` and `RateOverridesCard.tsx:270`. `OrganizationGeneralPage.tsx:133` and `WorkspaceMcpServersCard.tsx:123` in this same PR do it right, so either follow them with a `create.reset()`/`update.reset()` in `openAdd`/`openEdit`, or move the mutation hook into the dialog below the key as the rung prescribes.

web/src/features/budgets/SpendCeilingDialog.tsx
```diff
@@ -58,25 +57,31 @@ export function SpendCeilingDialog({
...
-
-  useEffect(() => {
-    if (!isOpen) return
-    setTarget(ORGANIZATION_SCOPE)
-    setBudgetId(editing?.budget_id ?? budgets[0]?.budget_id ?? "")
-    setProvider(editing?.provider_key_id ?? "")
-    setName(editing?.name ?? "")
-  }, [isOpen, editing, budgets])
+  // Seeded on mount only, because the caller remounts this on each open.
+  const seed = {
```
khaledosman, line 84: Adding to CodeRabbit's thread at line 66 rather than restating it: `seed` is recomputed every render, so it is a moving baseline rather than a snapshot. `feedback.md` asks for "one snapshot of every field the form owns, **seeded once on mount**" via `const seed = useRef(draft)`, and warns that "a guard that lies lets Escape discard work it cannot see." Snapshot it here (`useRef({…}).current`) and in the four other dialogs built the same way.

The part not yet raised: the old effect had `budgets` in its deps and re-seeded `budgetId` when the list arrived. Nothing does now, so a dialog opened before that query resolves leaves the Budget select unset rather than defaulting to the first budget. If that default is meant to survive the migration, gate the dialog on the budgets query instead of seeding from an empty array.

web/src/features/models/ModelsPage.tsx
```diff
@@ -2171,7 +2174,10 @@ export function ModelsPage() {
         <Button
           size="sm"
           variant="ghost"
-          onPress={() => setCustomPriceKey(searchedSelector ?? "")}
+          onPress={() => {
```
khaledosman, line 2172: This opener bumps `priceOpenCount` but does not clear `customError`, which is reset only at the start of `priceCustomModel`. So a rejected price stays banner-visible over the blank form the new `key={priceOpenCount}` remounts. Add `setCustomError(undefined)` and `setCustomPending(false)` here and in the second opener at line 2296.

`ActivityPage.tsx` has the same gap with page-level mutations: `setPrice.error` and `setModelPrice.error` are passed into keyed dialogs (2281, 2293) with no `.reset()` in either opener.

Rung: `feedback.md`, "The component that renders the `FormDialog` owns everything that resets between opens: the draft *and* its mutation … Both live below the key."

web/src/features/organization/RenameOrganizationDialog.tsx
```diff
@@ -1,8 +1,6 @@
-import { useEffect, useState } from "react"
+import { useState } from "react"
 
-import { Button } from "@/design-system/actions/Button"
-import { Dialog } from "@/design-system/feedback/Dialog"
-import { ErrorBanner } from "@/design-system/feedback/ErrorBanner"
+import { FormDialog } from "@/design-system/feedback/FormDialog"
```
khaledosman, line 3: Anchored here because a deleted file cannot carry a comment. Removing `design-system/feedback/Dialog.tsx` leaves three normative rungs describing a component that no longer exists:

- `web/design/DESIGN.md:66` still lists `/Dialog` in the `feedback/` module row
- `web/design/DESIGN.md:189` still advertises `Dialog` in the overlays topic row
- `web/design/overlays.md:14` routes "modal, dismissed deliberately" to it, and `:84` documents it as "the bare `AlertDialog` shell the other pattern was built from"

Delete those, and give the decision tree an answer for the non-form, non-confirm modal — `features/usage/ShareDialog.tsx` is a live instance of that shape hand-rolling `AlertDialog` today, so the branch still needs a destination.

AGENTS.md: "a fact told in two layers is a fact that will go stale in one of them, and the stale copy is the one someone believes." (Same class of gap as the `PageHeader` docs on #1073.)

web/src/features/organization/PricingOverrideDialog.tsx
```diff
@@ -191,127 +200,98 @@ export function PricingOverrideDialog({
...
+          value={modelKey}
+          onChange={setModelKey}
+          placeholder="provider:model"
+          isRequired
+          autoFocus
+          description="For example openai:gpt-4o. A provider instance name works too."
+        />
+      )}
+      <div className="grid grid-cols-1 gap-4 sm:grid-cols-2">
+        <RateField
```
khaledosman, line 270: On the edit path nothing takes focus: the `Field` that owns `autoFocus` is replaced by the read-only model block, and this local `RateField` has no `autoFocus` prop, so focus lands on the frame's Close control. `feedback.md`'s FormDialog rules: "The first field takes `autoFocus`." Add the prop to `RateField` — `SetPriceDialog`'s copy already has it — and pass it here when `editing`.

`SpendCeilingDialog` has the same gap with no field taking it at all; its first control is a `Select`, whose signature in forms.md carries no `autoFocus`, so that one needs a call on which control should receive it.

web/src/features/models/SetPriceDialog.tsx
```diff
@@ -153,83 +151,62 @@ export function SetPriceDialog({
...
-            </AlertDialog.Dialog>
-          </AlertDialog.Container>
-        </AlertDialog.Backdrop>
+    <FormDialog
+      isOpen={isOpen}
+      onOpenChange={onOpenChange}
+      size="lg"
+      title={title}
+      description={description(targetCount)}
+      submitLabel="Set price"
```
khaledosman: `submitLabel` is hardcoded while `title` and `description` are props, so no call site's submit repeats its own trigger: `ModelsPage` opens from "Price a model by hand" and "Price <selector>", `ActivityPage`'s row detail from "Price this model", and all three submit "Set price". `feedback.md`'s prop contract: "The submit button's label, and the trigger's, word for word: the control that opened this dialog and the one that completes it say the same thing." Make it a prop defaulting to "Set price".

While there: `ActivityPage`'s toolbar instance takes the default `title`, so title, submit and trigger all read "Set price" — `feedback.md`'s Incorrect example ("the title restates the button, and nothing says what was made"). The title should name what is being priced.

khaledosman (approved):
Reviewed the incremental range only (`6361b7bf..2f8f1132`, 23 files): this branch is based on `main` but sits at the top of the stack behind #1073, #1071 and #1068, so its diff against `main` carries all of them.

The deletion is clean: `feedback/Dialog.tsx` has no surviving code, story, test, or e2e reference, and `feedback/` has no barrel to update. Nothing that is actually a confirmation got moved — the three remaining `AlertDialog` call sites are the master-key regenerate confirm, the pricing-preview confirm, and `usage/ShareDialog`. Both budget pickers are now the form-half `Select`, which is what `9427253e` claimed.

Approving; none of the below blocks. The one worth landing before merge is the error-banner reset, because `feedback.md` uses this exact feature as its Incorrect example. CodeRabbit already has the `SpendCeilingDialog` seed race at line 66; the finding here adds the second half of it.

web/src/features/budgets/OrganizationBudgetsCard.tsx
jigjigjig, line 133: The key remounts the draft but not the mutation it reports. `create` and `update` live in this card, above the key, and neither `openAdd` nor `openEdit` calls `reset()`. Refused create, Cancel, Discard, Add again: the fresh form opens with the previous attempt's `ErrorBanner` already in it, and a failed edit of row A is what row B's dialog shows. feedback.md: "The component that renders the FormDialog owns everything that resets between opens: the draft *and* its mutation".

Fix: move `useCreateOrganizationBudget` / `useUpdateOrganizationBudget` into `OrganizationBudgetDialog` so the card keeps only `isOpen`, the counter and `editing`; or at minimum `create.reset()` / `update.reset()` in both openers, as `WorkspaceMcpServersCard` does.

web/src/features/budgets/SpendCeilingsCard.tsx
```diff
@@ -203,6 +208,7 @@ export function SpendCeilingsCard({
       </Card>
 
       <SpendCeilingDialog
+        key={openCount}
```
jigjigjig, line 187: Same defect as `OrganizationBudgetsCard`: `create` and `update` sit above the key and the openers never `reset()` them, so a refused save's banner survives into the next fresh open. Move the mutations below the key or reset both in `openAdd` / `openEdit`.

web/src/features/organization/RateOverridesCard.tsx
```diff
@@ -260,7 +265,9 @@ export function RateOverridesCard() {
         />
       </TableScrollFrame>
 
+      {/* Keyed on the open count, so each open remounts a blank form. */}
       <PricingOverrideDialog
+        key={openCount}
```
jigjigjig, line 240: Same defect as `OrganizationBudgetsCard`: `create` and `replace` sit above the key and the openers never `reset()` them, so a refused save's banner is still in the body when the next dialog opens blank. Move the mutations below the key or reset both in `openAdd` / `openEdit`.

web/src/features/models/ModelsPage.tsx
```diff
@@ -2360,7 +2369,9 @@ export function ModelsPage() {
         </div>
       </div>
 
+      {/* Keyed on the open count, so each open remounts a blank form. */}
       <SetPriceDialog
+        key={priceOpenCount}
```
jigjigjig, line 2375: `customError` is page state above the key and is cleared only when the next submit starts (`priceCustomModel`), so a refused hand-price followed by Cancel reopens, from either opener, a blank form under the old error. Clear it in both openers, or move the pending/error pair into `SetPriceDialog`; this is also the one caller still hand-rolling `customPending` / `customError` instead of using the mutation's own state.

web/src/features/activity/ActivityPage.tsx
```diff
@@ -2257,7 +2272,9 @@ export function ActivityPage() {
         onConfirm={onDeleteConfirm}
       />
 
+      {/* Keyed on the open count, so each open remounts a blank form. */}
       <SetPriceDialog
+        key={`repricing-${priceOpenCount}`}
```
jigjigjig, line 2279: `setPrice.error` and `setModelPrice.error` are never reset (no `.reset()` anywhere in this file) and both mutations sit above the keys, so a refused price is still in the banner when either dialog reopens blank. Reset the matching mutation in each opener (the `Set price` button and `RequestDetail`'s `onPriceModel`), or move the mutations under the key.

web/src/features/tools/McpServerDialog.tsx
```diff
@@ -118,6 +114,16 @@ export function McpServerDialog({
   const urlReason = urlProblem(url, willHaveToken)
   const invalid =
     name.trim() === "" || url.trim() === "" || urlReason !== undefined
+  // One predicate naming every field the operator can change, so "is there
+  // anything to lose" cannot drift from what the form actually holds.
+  const isPristine =
```
jigjigjig: This is the hand-listed predicate feedback.md rules out: "`isDirty` comes from `useDirtySnapshot`: hand it every field the form owns and it snapshots them on mount". The hook has been in `design-system/forms` since #1037 and `BudgetsPage`, `KeysPage` and `WorkspacesPage` use it. The five dialogs here (this one, `OrganizationBudgetDialog:77`, `SpendCeilingDialog:73`, `PricingOverrideDialog:125`, `SetPriceDialog:113`) each restate their field list a second time, which is the drift the hook was written to remove, and comparing against a `seed` recomputed on every render makes the guard track the props rather than what was seeded (see the `SpendCeilingDialog` thread).

Fix, in all five: `const { isDirty } = useDirtySnapshot({ name, url, token, clearToken, hint, allowedTools, enabled })` and pass `isDirty` through; drop `isPristine`.

jigjigjig (commented):
Review of the three #1074 commits (a7b76b70..617f0732) against web/AGENTS.md, frontend-standards, and web/design (feedback.md, forms.md, actions.md, overlays.md). Locally on the PR head: `pnpm --dir web run lint` and `run typecheck` clean; the seven touched vitest files plus the architecture and foundation probes pass (9 files, 3401 tests).

Twelve inline findings. The recurring one: every caller keeps its mutation above the remount key and (except the MCP and rename callers) never resets it, so a refused save's banner survives into the next fresh open. The guard is a hand-listed predicate where `useDirtySnapshot` already exists, and `SpendCeilingDialog` seeds its budget from a list that can land after mount, which is the case feedback.md names. The rest are label rules #1031 asks this PR to reconcile, the docs that still describe the deleted `Dialog`, and one dropped `<code>`.

Review written by Claude Code (claude-fable-5-1) at the request of the PR's reviewer.

web/src/features/budgets/SpendCeilingDialog.tsx
```diff
@@ -58,25 +57,31 @@ export function SpendCeilingDialog({
...
+    target: ORGANIZATION_SCOPE,
+    budgetId: editing?.budget_id ?? budgets[0]?.budget_id ?? "",
```
jigjigjig, line 86: The default budget is part of the seed but can arrive after mount: `budgets` is `budgets.data ?? []`, `Add ceiling` is enabled before that query resolves, and the effect this replaces had `budgets` in its deps for exactly that reason. Opened before the list lands, `budgetId` seeds to `""` and now stays there (placeholder in the Select, `blockedReason` reads "Choose the budget"). Worse, because `seed` is recomputed every render, `seed.budgetId` becomes `budgets[0]` once the list arrives while state is still `""`, so `isPristine` flips false and Escape asks to discard a form nobody typed in. A refetch that reorders the list while the dialog is open does the same. feedback.md names this case: a field whose default arrives after mount is part of the seed; seed from the resolved value or call the hook's `reset` when it lands.

Fix: `useDirtySnapshot` for the guard, and either gate the opener on `budgets.data !== undefined` in `SpendCeilingsCard`, or set `budgetId` and call `reset()` when `budgets` first becomes non-empty.

web/src/features/budgets/SpendCeilingDialog.tsx
```diff
@@ -134,100 +139,74 @@ export function SpendCeilingDialog({
...
+            label="Provider instance"
+            value={provider}
+            onChange={setProvider}
+            placeholder="openai-eu"
+            description="Optional. Narrows the cap to one provider; leave blank to cap spend across every provider."
+          />
+        </>
+      )}
+      <Select
+        label="Budget"
```
jigjigjig, line 230: Two of the three `blockedReason`s are about this control ("Choose the budget this ceiling enforces", "holds a budget set at the deployment level"), yet they render as a `<p>` at the foot of the body, below every other field. The reason to swap to `forms/Select` (its docstring) is that it owns an error announced on the control; `McpServerDialog` in this same PR does that for the URL and says why. Pass `isInvalid` and `errorMessage` here for those two reasons (with `reserveMessage` on, so the footer does not move), and keep the paragraph only for `noBudgets`, which is about the list rather than the choice.

web/src/features/models/SetPriceDialog.tsx
jigjigjig: #1031 lists label reconciliation for this PR, and actions.md's rule is that trigger and submit are the same string while the title names the object. This dialog's four openers read "Price a model by hand" / "Price <selector>" (ModelsPage) and "Price this model" / "Set price" (ActivityPage); the submit is always "Set price", and the bulk-reprice case also takes the default title "Set price", so there the title and the button restate each other. Make `submitLabel` a prop (default "Set price") that each caller passes with the same string as its trigger, and give the bulk dialog a title that names the object rather than the verb.

web/src/features/organization/RenameOrganizationDialog.tsx
```diff
@@ -34,38 +32,26 @@ export function RenameOrganizationDialog({
...
-            isDisabled={isUnchanged || trimmed === ""}
-            isPending={isPending}
-            onPress={() => onSubmit(trimmed)}
-          >
-            Change name
-          </Button>
-        </>
-      }
+      size="sm"
+      title="Change organization name"
```
jigjigjig: Title restates the trigger and the submit drops half of it: trigger "Change organization name" (`OrganizationGeneralPage:80`), title "Change organization name", submit "Change name". feedback.md's FormDialog example marks exactly this shape as incorrect ("the title restates the button"). Fix: `title="Organization name"`, `submitLabel="Change organization name"`, and update the three places that query these strings (this dialog's test, `OrganizationGeneralPage.test.tsx`, and `e2e/parity.tenancy.spec.ts`, whose `getByRole("dialog", { name })` follows the title).

web/src/design-system/feedback/Dialog.tsx
```diff
@@ -1,79 +0,0 @@
-import { AlertDialog } from "@heroui/react"
```
jigjigjig, line 1: The design docs still describe this component. `web/design/overlays.md` says there are three dialogs and that `Dialog` "has no call sites" (the Dialog section), and refers to it twice more (line 17, and the Popover comparison at line 62); `web/design/DESIGN.md:189` indexes overlays.md as covering `Dialog`. With the file gone those read as a component the reader cannot find, and this is the change that removes it, so it owes the edit: drop the `Dialog` bullet (two dialogs, not three), point the Popover comparison at `FormDialog` / `ConfirmDialog`, and fix the DESIGN.md index row.

web/src/features/tools/McpServerDialog.tsx
```diff
@@ -136,134 +142,99 @@ export function McpServerDialog({
...
-          </AlertDialog.Container>
-        </AlertDialog.Backdrop>
-      ) : null}
-    </AlertDialog>
+    <FormDialog
+      isOpen={isOpen}
+      onOpenChange={onOpenChange}
+      size="lg"
+      title={editing ? "Edit MCP server" : "New MCP server"}
+      description="An MCP endpoint this workspace's requests can reach by naming its id in mcp_server_ids. The gateway connects to it while a request runs, so it has to be reachable from the gateway rather than from this browser."
```
jigjigjig: `description` is a `ReactNode`; the old body set `mcp_server_ids` in `<code className="font-mono">`, which flattening to a string drops. Keep the identifier in code: `description={<>An MCP endpoint ... naming its id in <code className="font-mono">mcp_server_ids</code>. ...</>}`.

web/src/features/budgets/OrganizationBudgetsCard.tsx
jigjigjig: Fixed in 7c9d736a, by the rule rather than by `reset()` in the openers: `useCreateOrganizationBudget` and `useUpdateOrganizationBudget` now live inside `OrganizationBudgetDialog`, below the card's key, so the remount that clears the draft clears the refusal with it. The card keeps `isOpen`, the counter and `editing`.

Test: refuse the create with a 409, read the banner, discard, reopen, and assert no alert and a blank name. Checked by putting an error back above the key (a module-level one that survives the remount), which fails it.

web/src/features/budgets/SpendCeilingsCard.tsx
jigjigjig: Same fix in 7c9d736a: both mutations moved into `SpendCeilingDialog`. Its own case is beside the budgets one, refusing the create and asserting the reopened dialog carries no alert; the same control fails it.

web/src/features/organization/RateOverridesCard.tsx
jigjigjig: Same fix in 7c9d736a: `useCreateOrganizationPricing` and `useReplaceOrganizationPricing` moved into `PricingOverrideDialog`, along with the replace path's own narrowing of `effective_from`. `RateOverridesCard.test.tsx` gained "does not greet the next open with the last attempt's refusal", which fails when an error is allowed to outlive the remount.

web/src/features/models/ModelsPage.tsx
jigjigjig: Fixed in 7c9d736a, and the hand-rolled pair is gone with it. `SetPriceDialog` cannot own the mutation (its four callers each have their own endpoint, and one of them also moves the table's selection afterwards), so it owns the pending and error **state** instead: it awaits what `onSubmit` returns and reports a rejection itself, below the caller's key. `priceCustomModel` is now just `mutateAsync` plus the two things it does on success, and it rethrows for the dialog to report.

Test: refuse the POST, read the banner, discard, reopen from the same opener, assert no alert and an empty rate. Making the failure outlive the remount fails it.

web/src/features/activity/ActivityPage.tsx
jigjigjig: Fixed the same way in 7c9d736a: both callers hand the dialog a promise (`setPrice.mutateAsync`, `setModelPrice.mutateAsync`) and it owns the pending and error state below each key. Nothing in the page reports a price refusal any more, so there is no `.reset()` to forget.

The bulk dialog also gained a title that names its object ("Imported row costs"), since its submit and trigger are both "Set price" and the default title was a third copy of that string.

web/src/features/tools/McpServerDialog.tsx
jigjigjig: All five converted in 7c9d736a: this dialog, `OrganizationBudgetDialog`, `SpendCeilingDialog`, `PricingOverrideDialog` and `SetPriceDialog`. Each passes the whole draft and drops its `isPristine`.

The hook needed two things to take the second caller, both in this commit: `reset` keeps one identity (one that changed every render would re-seed on every render through an effect that depended on it, and no form would ever read dirty), and `reset(next)` seeds an explicit draft for the after-mount default case you describe in the ceiling thread. Three cases added on the hook for those.

Worth knowing: the same conversion for seven other call sites is sitting on an open PR of mine that is frozen pending a ruling, so the hook change here may shrink or replace it.

web/src/features/budgets/SpendCeilingDialog.tsx
jigjigjig: All of it fixed in 7c9d736a, and you were right that the recomputed `seed` was the worse half. The dialog now seeds the budget default when the list first arrives, in the render that applies it, with `reseed({ ... budgetId: landed ... })`: the explicit value matters because that render still holds the empty one.

The case holds the budgets GET behind a gate the test releases after the dialog is open, which is the only way the after-mount default is observable; opened against a resolved list, the old code passes. Two controls, and both bite: dropping the reseed fails it, and reseeding with a bare `reset()` instead of the explicit draft fails it too.

web/src/features/budgets/SpendCeilingDialog.tsx
jigjigjig: Done in 7c9d736a: the two reasons about the choice are the Budget select's `isInvalid` and `errorMessage`, with `reserveMessage` on so announcing one does not move the footer, and `noBudgets` stays a paragraph because it is about the list rather than the choice.

One correction to the premise, measured rather than assumed, because it changes what this buys. `forms/Select` does **not** announce its error on the control. Rendered invalid with a message, the trigger carries `aria-labelledby`, `aria-haspopup` and `aria-expanded` and nothing else: no `aria-invalid`, no `aria-describedby`, no `data-invalid`. The message is a plain span inside the select's group, and the component's own comment ("the control is marked invalid through `isInvalid` above, which is what assistive tech reads") is wrong about that.

So the change is real but smaller than the docstring promises: the refusal is now beside the control it is about rather than five fields below it, which is the visual half. The programmatic association a `Field` gives does not exist here. My test asserts what is true (the message renders inside the Budget control's own group) rather than an `aria-describedby` that is not there. Fixing `Select` is a design-system change and its own diff; flagged for whoever takes it.

web/src/features/models/SetPriceDialog.tsx
jigjigjig: Done in 7c9d736a. `submitLabel` is a prop defaulting to "Set price", and every caller passes its own trigger's string: ActivityPage's bulk case "Set price" and its request case "Price this model", and ModelsPage tracks which of its two openers was used ("Price <selector>" or "Price a model by hand" from the empty state, "Price a model" from the no-discovery banner) so the submit says what the operator just pressed. The bulk dialog's title is now "Imported row costs" rather than the verb.

Four tests had to be rescoped as a result, which is the cost of the rule: trigger and submit now share a name, so each press is scoped to the dialog.

web/src/features/organization/RenameOrganizationDialog.tsx
jigjigjig: Done in 7c9d736a: `title="Organization name"`, `submitLabel="Change organization name"`, matching the trigger word for word. Updated all three places that query the old strings: this dialog's test, `OrganizationGeneralPage.test.tsx`, and `e2e/parity.tenancy.spec.ts`, whose `getByRole("dialog", { name })` follows the title and now looks for "Organization name".

web/src/design-system/feedback/Dialog.tsx
jigjigjig: Done in 7c9d736a. `overlays.md` now says two dialogs rather than three, the decision tree's leaf points at `FormDialog` or `ConfirmDialog`, the Popover comparison refers to the dialogs generally, and the paragraph that described the shell says it is gone and why. `DESIGN.md`'s index row for overlays.md no longer lists `Dialog`. `grep` for a `Dialog` reference in `web/design` comes back empty.

web/src/features/tools/McpServerDialog.tsx
jigjigjig: Done in 7c9d736a: the description is a fragment again with `mcp_server_ids` in `<code className="font-mono">`.

web/src/features/budgets/OrganizationBudgetsCard.tsx
jigjigjig: Fixed in 7c9d736a. Both mutations moved into `OrganizationBudgetDialog`, below the card's key, so the remount clears the refusal with the draft; the card keeps `isOpen`, the counter and `editing`. Test: refuse the create, read the banner, discard, reopen, assert no alert and a blank name. It fails if an error is allowed to outlive the remount.

web/src/features/budgets/SpendCeilingDialog.tsx
jigjigjig: Right about the moving baseline, and it was the half that bit: fixed in 7c9d736a with `useDirtySnapshot`, in this dialog and the four others built the same way.

The ceiling needed one thing more than a snapshot, because its default arrives with the budget list rather than at mount: the hook gained `reset(next)` and the dialog seeds the landed default in the render that applies it. Both halves are checked, and the second one is easy to get wrong: the case holds the budgets GET behind a gate released after the dialog is open, and it fails both when the reseed is dropped and when it is done with a bare `reset()` instead of the explicit draft.

web/src/features/models/ModelsPage.tsx
jigjigjig: Fixed in 7c9d736a, and the pair is gone rather than cleared in the openers. `SetPriceDialog` owns the pending and error state now, below each caller's key, and awaits what `onSubmit` returns; `customPending` and `customError` are deleted and `priceCustomModel` is `mutateAsync` plus the two things it does on success. Test refuses the POST, discards, reopens from the same opener and asserts no banner; making the failure outlive the remount fails it.

web/src/features/organization/RenameOrganizationDialog.tsx
jigjigjig: Done in 7c9d736a: `overlays.md` says two dialogs, its decision tree points at `FormDialog` or `ConfirmDialog`, the paragraph about the shell says it is gone and why, and `DESIGN.md`'s index row drops `Dialog`. A grep for a `Dialog` reference under `web/design` is empty.

web/src/features/organization/PricingOverrideDialog.tsx
jigjigjig: Real, and the one finding in your round that our own review missed. Fixed in f940681d: `RateField` takes `autoFocus` and the input rate asks for it while editing, since the model key's field (which carries it on the add path) is replaced by the read-only block there.

Test asserts focus lands in "Input, per 1M tokens" after pressing Edit, and fails with the prop removed.

web/src/features/models/SetPriceDialog.tsx
jigjigjig: Done in 7c9d736a: `submitLabel` is a prop defaulting to "Set price", and each caller passes its own trigger's string. ModelsPage tracks which of its two openers was used, so the submit reads "Price <selector>", "Price a model by hand" or "Price a model" to match what was pressed; ActivityPage passes "Set price" for the bulk case and "Price this model" for the row detail. The bulk dialog's title is now "Imported row costs", since with the submit matching its trigger the old title was a third copy of the same words.

Four existing cases had to scope their press to the dialog as a result, which is the cost of the rule rather than a surprise.

web/src/features/budgets/BudgetsPage.test.tsx
```diff
@@ -159,6 +159,53 @@ describe("BudgetsPage", () => {
...
+    const user = userEvent.setup()
+    renderPage(<BudgetsPage />)
+
+    const row = (await screen.findByText("team-free-tier")).closest("tr")!
+    await user.click(within(row).getByRole("button", { name: "Edit" }))
+    // A dialog title is a noun phrase, so the budget it names is in the
+    // description rather than folded into the title.
+    const dialog = await screen.findByRole("dialog")
+    expect(dialog).toHaveAccessibleName("Edit budget")
+    expect(dialog).toHaveTextContent("team-free-tier")
```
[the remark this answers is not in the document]
jigjigjig: Fixed in e3dfcfe6. You were right that the assertion could pass while a screen reader omitted the name: `toHaveTextContent` reads the DOM, and the description was a paragraph with nothing tying it to the dialog.

The wiring landed in the meantime (`FormDialog` gives the description a `useId` and sets `aria-describedby` on `Modal.Dialog`, verified in the DOM rather than assumed, since react-aria filters unrecognized ARIA off some components), so this case now asserts `toHaveAccessibleDescription`. It fails when that attribute is taken off the dialog.

web/src/features/users/UserMultiSelect.tsx
```diff
@@ -36,111 +30,35 @@ export function UserMultiSelect({
...
-    </div>
+    <MultiSelect
+      label={label}
+      description={description}
+      options={options}
+      value={value}
+      onChange={onChange}
+      searchPlaceholder="Search people…"
+      countNoun="people assigned"
+      emptyMessage="Nobody to assign yet. Add people under Members & roles, or issue a key, and they can be assigned here."
```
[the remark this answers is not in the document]
jigjigjig: Already fixed, and not by this PR. The empty message on `main` reads "Nobody to assign yet. Add people under Members & roles and they can be assigned here." The API-key instruction you flagged is gone, which is the right answer for the reason you gave: line 38 excludes every `apikey-*` identity, so issuing a key could never add an option here.

Resolving as obsolete rather than claiming credit for it.

web/src/features/tools/SearchToolsCard.tsx
```diff
@@ -239,94 +248,76 @@ function AddToolForm({ providers }: { providers: SearchProviderInfo[] }) {
...
+      onOpenChange={(open) => {
+        if (!open) onClose()
+      }}
+      size="sm"
+      title="New search tool"
+      submitLabel="Add search tool"
+      onSubmit={submit}
+      isPending={create.isPending}
+      isSubmitDisabled={!ready}
+      isDirty={name.trim() !== "" || apiBase.trim() !== "" || apiKey !== ""}
```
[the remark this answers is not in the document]
jigjigjig: Real, and not this PR's to fix: `SearchToolsCard` belongs to #1071, which is being worked now. Filed here because this branch is the top of the stack and its diff carries the file.

Leaving this open deliberately, so it is not resolved by a rebase that merely moves the line. It should close on #1071, and this branch will carry the fix once it rebases over that head. The fix itself is the one you named: `provider` belongs in the snapshot with the other four fields, and `useDirtySnapshot` is now the shape for it.

web/src/features/workspaces/WorkspaceMembersPage.tsx
```diff
@@ -45,16 +58,51 @@ export function WorkspaceMembersPage() {
...
 
+  const present = new Set((members.data ?? []).map((member) => member.user_id))
+  const candidates = orgMembers.data
+    ? orgMembers.data.filter(
+        (member) =>
+          member.user_id &&
+          member.status === "active" &&
+          !present.has(member.user_id),
+      )
+    : []
```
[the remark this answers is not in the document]
jigjigjig: Real, and it belongs to #1073, which owns `WorkspaceMembersPage` and is being worked now. `rosterResolved` checking only the organization roster is live on `main` today, so it is not obsolete: if the organization list resolves first, the dialog offers people who are already members.

Left open rather than resolved, for the same reason as its sibling: a thread on a stacked PR should not be closed by a rebase that moves its line. It closes on #1073.

web/src/features/workspaces/WorkspaceMembersPanel.tsx
```diff
@@ -55,72 +66,70 @@ function AddWorkspaceMember({
...
+      }}
+      size="sm"
+      title="New workspace member"
+      submitLabel="Add member"
+      onSubmit={() =>
+        add.mutate({ workspaceId, userId, role }, { onSuccess: onClose })
+      }
+      isPending={add.isPending}
+      isSubmitDisabled={userId === ""}
+      isDirty={userId !== ""}
```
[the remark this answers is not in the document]
jigjigjig: Real, same owner: `WorkspaceMembersPanel` is #1073's. `isDirty={userId !== ""}` omits the role, so choosing a non-default role and dismissing discards it with nothing asked, and that is live on `main`.

Left open for #1073 to close. The fix there should be `useDirtySnapshot({ userId, role })` rather than the two-field predicate, since the hook is on `main` now.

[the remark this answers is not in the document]
jigjigjig: Closed on #1073 and now in this tree: `WorkspaceMembersPanel` reads `const { isDirty } = useDirtySnapshot({ userId, role })` at 2e0f9c74, which is the shape I described rather than the two-field predicate.

Resolving from this branch, since the code the thread points at is the fixed code here. Credit is #1073's.

web/src/features/workspaces/WorkspaceMembersPage.tsx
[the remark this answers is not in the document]
jigjigjig: Closed on #1073 and now in this tree. `WorkspaceMembersPage` gates the candidate list on `bothRostersAnswered = orgMembers.isSuccess && members.isSuccess`, so it no longer offers people the other read was about to exclude, and "everyone is already here" is only said once both have landed.

One note on the shape, since it is slightly better than what I suggested on this thread: the guard also drives the panel's `rosterResolved` prop, so the empty-state copy and the candidate list cannot disagree. Resolving; credit is #1073's.

web/src/features/tools/SearchToolsCard.tsx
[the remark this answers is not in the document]
jigjigjig: Closed by #1071, which merged as 5855af35, and now in this tree through the rebase: `SearchToolsCard`'s guard covers `provider` with the other four fields.

Resolving from this branch because the code is here; the fix is #1071's.

PR #1082, opened 2026-09-11
refactor(dashboard): derive isDirty from the seeded snapshot in every form

jigjigjig:
## Description

Seven create and edit forms each carried their own copy of the same four lines:
stringify the whole draft, keep the first one in a ref, compare the two. That is
the shape `feedback.md` prescribes and `useDirtySnapshot` already implements, and
each of the seven predates the hook by a few hours. This converts them: the
domain claim form, both organization member forms, the organization provider-key
form, both add-provider tabs, and the routing policy form.

Nothing an operator can see changes. What changes is that the rule lives in one
place, so the next form to grow a field cannot forget to tell its guard.

The hook gained two things the second caller needed:

- `reset` keeps one identity across renders, so an effect may depend on it. One
  that changed every render would re-seed on every render and no form would ever
  read dirty.
- `reset(next)` seeds an explicit draft. A form whose own default arrives after
  mount (the member forms seed a workspace from the roster) has to say "this is
  the seed" in the render that applies it. Doing it from an effect does not work
  and the PR proves why: a ref written after the commit re-seeds nothing already
  rendered, so the form arrived armed and Escape asked before closing something
  nobody had touched. The reopened-clean test caught it.

## How to test it locally

`pnpm --dir web test` covers it, and the point of this PR is that the existing
guards are the proof rather than new tests.

The check I ran, and the one to repeat if you want to see the refactor held:
make the hook report `isDirty: false` and run the suite. **22 cases fail** across
the seven forms and the hook's own file, among them

- providers: "guards an Advanced rename on the known tab", "guards client options typed on the custom tab"
- members: "guards a role change on the add form", "guards an unticked workspace on the invite form, and reopens clean"
- domains: "guards a changed role on the way out, with nothing typed"
- organization provider keys: "opens on a blank draft after a create", "does not carry a refused create's banner into the next open"
- routing: "guards a half-built policy against a stray Escape"
- budgets, keys, workspaces: the converted callers already on `main`, which this PR does not touch

One flake to disclose rather than have you find: `App.test.tsx`'s "shows a
loading state while the current route loads" failed once in a full run here and
passed on its own and on a second full run (5654 passing both times). It touches
none of these files.

## PR Type

- [ ] New Feature
- [ ] Bug Fix
- [x] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

None. Follow-up to the form-dialog series (#1031), deliberately not linked to it:
that issue tracks pages moving into `FormDialog`, and this is the shape of the
guard they all ended up needing.

## Checklist

      Three cases on the hook for what it grew (stable `reset`, `reset()` to the
      latest draft, `reset(next)` to an explicit one). The conversions are covered
      by the guards listed above, which is the point.
      `pnpm --dir web run lint`, `pnpm --dir web run typecheck`, `pnpm --dir web test`.
      `feedback.md` already names the hook as the shape; nothing to change.
      No API change.

## AI Usage

- [ ] No AI was used.

### DIFF
--- a/web/src/design-system/forms/useDirtySnapshot.ts
+++ b/web/src/design-system/forms/useDirtySnapshot.ts
@@ -1,4 +1,4 @@
-import { useCallback, useRef } from "react"
+import { useCallback, useRef, useState } from "react"
 
 /**
  * Whether a form's draft still matches what it was seeded with.
@@ -18,10 +18,14 @@ import { useCallback, useRef } from "react"
  * explicit draft instead, which is what a form whose own default lands after
  * mount needs: that default is part of the seed rather than a change, and the
  * code that computes it holds the new value while the render it is in still
- * holds the old one. Seed it there, during the render that applies it, and not
- * from an effect: a ref write after the commit re-seeds nothing that has
- * already been rendered, so the guard stays armed until something else
- * re-renders.
+ * holds the old one. Call it there, during the render that applies the default:
+ * the seed is state, so a reset during render re-runs the component and the
+ * `isDirty` the caller receives is the corrected one. From an effect it also
+ * works, one render later.
+ *
+ * A reset during render has to be guarded by whatever made the default arrive
+ * (`if (!seeded && rows.length > 0)`), the same as any other set-state-while-
+ * rendering: unconditional, it is an endless render.
  *
  * `reset`'s identity is stable, so an effect may depend on it; a `reset` that
  * changed every render would re-seed on every render and the form would never
@@ -32,13 +36,22 @@ export function useDirtySnapshot(draft: unknown): {
   reset: (next?: unknown) => void
 } {
   const snapshot = JSON.stringify(draft)
-  const seeded = useRef(snapshot)
+  // State rather than a ref, so a reset during render corrects that render.
+  // Held in a ref it corrected nothing already rendered, and the two callers
+  // that appeared to work did so only because they set other state in the same
+  // block, which is what re-ran the component.
+  const [seed, setSeed] = useState(snapshot)
   // Read by `reset`, which has no dependencies and so cannot close over the
   // current render's snapshot.
   const latest = useRef(snapshot)
   latest.current = snapshot
+  // Not reflexive memoization, which the frontend standards rule out: the
+  // compiler memoizes nothing in this hook (it bails with "Cannot access refs
+  // during render" on the `latest` read above, measured with
+  // babel-plugin-react-compiler 1.0.0 against both this version and the one
+  // before it), so this `useCallback` is what actually keeps `reset` stable.
   const reset = useCallback((next?: unknown) => {
-    seeded.current = next === undefined ? latest.current : JSON.stringify(next)
+    setSeed(next === undefined ? latest.current : JSON.stringify(next))
   }, [])
-  return { isDirty: snapshot !== seeded.current, reset }
+  return { isDirty: snapshot !== seed, reset }
 }
--- a/web/src/features/organization/OrganizationMembersPage.tsx
+++ b/web/src/features/organization/OrganizationMembersPage.tsx
@@ -1,5 +1,5 @@
 import { Button } from "@heroui/react"
-import { useMemo, useRef, useState } from "react"
+import { useMemo, useState } from "react"
 
 import type {
   User as ApiUser,
@@ -26,6 +26,7 @@ import { InfoBanner } from "@/design-system/feedback/InfoBanner"
 import { Checkbox } from "@/design-system/forms/Checkbox"
 import { Field } from "@/design-system/forms/Field"
 import { Select } from "@/design-system/forms/Select"
+import { useDirtySnapshot } from "@/design-system/forms/useDirtySnapshot"
 import { Dot } from "@/design-system/indicators/Dot"
 import { PageIntro } from "@/design-system/layout/PageIntro"
 import { Section } from "@/design-system/layout/Section"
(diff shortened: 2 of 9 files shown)


web/src/features/organization/OrganizationMembersPage.tsx
```diff
@@ -205,12 +209,9 @@ function AddMemberForm({
...
     // taken at first render would report the form dirty the moment the roster
-    // answers, and Escape would ask before closing an untouched form.
-    seededDraft.current = JSON.stringify({
-      email,
-      role,
-      workspaceIds: defaults,
-    })
+    // answers, and Escape would ask before closing an untouched form. Seeded
+    // with the explicit value, because this render still holds the old list.
+    reseed({ email, role, workspaceIds: defaults })
```
khaledosman, line 214: The seed captures `email` and `role` as they stand at the render the roster answers, not the form's mount values. `useWorkspaces()` walks every page through `fetchAllPaged`, so on a cold cache this gate can first fire after the operator has already typed an address or changed the role. That typed value then becomes part of the seed, `isDirty` reads false, and Escape closes without asking, which is the failure the guard exists to prevent.

Pre-existing (`main` had the same fields in the `seededDraft.current` write), but this PR rewrites the line, so it is cheap to fix here. Minimal:

```js
reseed({ email: "", role: "member", workspaceIds: defaults })
```

Same on line 349 in `InviteMemberForm`.

Cleaner shape if you want the reseed gone entirely: hold `workspaceIds` as `string[] | undefined` and render `workspaceIds ?? defaults`, snapshotting the same expression. The roster arriving then changes nothing about dirtiness, and both the `seeded` state and the render-phase reseed disappear.

web/src/design-system/forms/useDirtySnapshot.ts
```diff
@@ -13,19 +13,32 @@ import { useRef } from "react"
...
-    },
-  }
+  // Read by `reset`, which has no dependencies and so cannot close over the
+  // current render's snapshot.
+  const latest = useRef(snapshot)
+  latest.current = snapshot
+  const reset = useCallback((next?: unknown) => {
+    seeded.current = next === undefined ? latest.current : JSON.stringify(next)
+  }, [])
+  return { isDirty: snapshot !== seeded.current, reset }
```
khaledosman, line 43: `isDirty` is computed here, at the hook call, so a `reset(next)` executed later in the same render body cannot affect the value that render already holds.

The docstring above prescribes seeding during render and explains why an effect fails, but omits the precondition that makes the render version work: both member forms get away with it only because they also call `setSeeded` / `setWorkspaceIds` in the same block, which makes React discard and re-run the render. A caller that reseeds during render without a state update in the same block stays armed until something else re-renders, which is the failure the docstring attributes to effects.

Either state that precondition in the docstring, or compare lazily so `isDirty` reflects a same-render reset.

web/src/design-system/forms/useDirtySnapshot.ts
```diff
@@ -13,19 +13,32 @@ import { useRef } from "react"
...
-    isDirty: snapshot !== seeded.current,
-    reset: () => {
-      seeded.current = snapshot
-    },
-  }
+  // Read by `reset`, which has no dependencies and so cannot close over the
+  // current render's snapshot.
+  const latest = useRef(snapshot)
+  latest.current = snapshot
+  const reset = useCallback((next?: unknown) => {
```
khaledosman, line 40: Not a finding, worth recording: this `useCallback` is not reflexive memoization against the repo's no-reflex-memo rung, and a comment saying so would stop a later cleanup removing it on that rung's authority.

I ran `babel-plugin-react-compiler` over both the `main` and PR versions of this hook. Both already bail with `CompileError: Cannot access refs during render`, from the pre-existing `seeded.current` read on line 43. The compiler memoizes nothing in this hook, so the explicit `useCallback` is what actually makes `reset` stable.

khaledosman (approved):
Verified at `838d5ea`: `pnpm --dir web run typecheck` clean, `pnpm --dir web run lint` clean, 273 tests pass across the touched areas. No generated artifacts owed, no dependency change.

Two findings inline, plus one note worth keeping so a later cleanup does not undo the `useCallback`.

---

web/src/features/organization/OrganizationMembersPage.tsx
jigjigjig: Fixed in fff3dc38, exactly as you wrote it: `reseed({ email: "", role: "member", workspaceIds: defaults })` in both forms, with the reason named on the line (the roster pages through `fetchAllPaged`, so on a cold cache it can answer after an address has been typed).

The case that pins it holds the workspaces GET behind a gate, types an address while it is in flight, releases it, and asserts the guard still appears on Cancel. Reseeding the fields as they stand fails it.

web/src/design-system/forms/useDirtySnapshot.ts
jigjigjig: You are right, and the precondition is gone rather than documented: the seed is `useState` now, so a `reset(next)` during render re-runs the component on its own and the `isDirty` the caller receives is the corrected one. Held in a ref it corrected nothing already rendered, and the two forms that appeared to work did so only because they set other state in the same block, exactly as you say.

What the docstring says now is the one precondition that remains and is real: a reset during render has to be guarded by whatever made the default arrive, the same as any other set-state-while-rendering, or it is an endless render.

The new case renders a component that calls `reset(next)` during render with no other state update in the block and asserts the last `isDirty` it was handed is false. It fails with the seed back in a ref, which is the shape you are describing.

web/src/design-system/forms/useDirtySnapshot.ts
jigjigjig: Recorded on the line, and I re-ran your probe rather than taking it: babel-plugin-react-compiler 1.0.0 still bails with "Cannot access refs during render" on this hook after the change, because the `latest` ref read survives, and the emitted code carries no `useMemoCache`. So the compiler memoizes nothing here and the `useCallback` is what keeps `reset` stable.

The comment says that, with the version measured, so a later cleanup on the no-reflex-memo rung has the reason in front of it.

PR #1084, opened 2026-09-11
fix(dashboard): line the passkey button up with the field it submits

### LINKED ISSUE #1085: [BUG] The passkey button sits below the field it submits on Account settings
### What happened

On **Account settings → Passkeys**, the "Add a passkey" button does not line up with the **Name** input it submits: it sits roughly one text line lower, so the row reads as broken.

The button is bottom-aligned against the whole field and then lifted by a fixed offset that assumes the field's description renders on a single line. At the width the card gives it, "Optional. It is only a label, so you can tell this passkey from the others." wraps to two lines, and the button falls short by the extra line.

### Steps to reproduce

1. Run a gateway with `public_base_url` set, so passkeys are offered.
2. Sign in to the dashboard and open **Account settings**.
3. Scroll to the **Passkeys** section, on a viewport at or above the `sm` breakpoint.

### What you expected

The button to share the input's row: same height, same vertical center, with the field's description below both.

### Notes

Dashboard-only, no API or data involved. Fixed by #1084.

jigjigjig:
## Description

On Account settings, the button that registers a passkey did not line up with the name box it submits: it sat about one text line lower, so the row read as broken. It now shares the box's row, the same height and the same vertical center, with the field's hint below both.

Nothing else changes: same form, same wording, same behavior on a phone, where the button still stacks under the box.

## How to test it locally

1. Run a gateway whose `public_base_url` is set to the address the dashboard is served on, so passkeys are offered (an easy one: `bash web/e2e/serve.sh` with `public_base_url: "http://127.0.0.1:8000"` added to `web/e2e/otari.yml`).
2. Sign in and open **Account settings**, on a window wider than the `sm` breakpoint.
3. Scroll to **Passkeys**: "Add a passkey" should sit beside the **Name** box, tops and bottoms flush, and the hint should read on one line under both.
4. Narrow the window below `sm`: the button drops under the box, as before.

Already proven by automation: the dashboard Vitest suite (5655 tests, including `PasskeysCard.test.tsx`), Biome, and `tsc`. The alignment itself is not covered by a test, because jsdom does not lay out; it was checked in a browser instead, where the box and the button both measure 36px tall with an identical `top`.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes #1085

## Checklist

The tests box is deliberately unchecked: this is a layout change, and jsdom has no layout to assert against. The existing `PasskeysCard.test.tsx` covers the form's behavior and still passes. Nothing in the API contract or the docs is touched.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

Claude Code (Claude Opus 5)

The fix was verified in a real browser against a locally served gateway, not only by reading the CSS.

**NOTE:**

### DIFF
--- a/web/src/features/account/PasskeysCard.tsx
+++ b/web/src/features/account/PasskeysCard.tsx
@@ -293,21 +293,19 @@ export function PasskeysCard() {
               startRegistration()
             }}
           >
-            {/* The button sits beside the input rather than beside the whole
-              field, because the description below it is a paragraph whose
-              height depends on the width it is given. Aligned against the
-              field, the button would have to be pushed up by however tall
-              that paragraph happened to render, which is a number no class
-              can know. */}
+            {/* The button is inside the field, beside the input, because this
+              field's description wraps: the width a wrapped message costs is
+              not a number `FieldAction` can reserve. See forms.md, "Control
+              rows". */}
             <TextField
               value={newName}
               onChange={setNewName}
-              className="flex max-w-xl flex-col gap-1"
+              className="flex max-w-2xl flex-col gap-1"
             >
               <Label className="text-body">Name</Label>
               <div className="flex flex-col gap-3 sm:flex-row sm:items-center">
                 <Input
-                  className="flex-1"
+                  className="w-full max-w-md"
                   placeholder="Work laptop"
                   maxLength={MAX_PASSKEY_NAME_LENGTH}
                 />
--- a/web/src/features/account/PasskeysCard.tsx
+++ b/web/src/features/account/PasskeysCard.tsx
@@ -288,22 +288,37 @@ export function PasskeysCard() {
           is exactly when somebody has orphans to clear out. */}
         {passkeys_ready && canUsePasskeys && !passkeys.isError ? (
           <form
-            className="flex flex-col gap-3 sm:flex-row sm:items-end"
             onSubmit={(event) => {
               event.preventDefault()
               startRegistration()
             }}
           >
+            {/* The button sits beside the input rather than beside the whole
+              field, because the description below it is a paragraph whose
+              height depends on the width it is given. Aligned against the
+              field, the button would have to be pushed up by however tall
+              that paragraph happened to render, which is a number no class
+              can know. */}
             <TextField
               value={newName}
               onChange={setNewName}
-              className="flex max-w-md flex-1 flex-col gap-1"
+              className="flex max-w-xl flex-col gap-1"
             >
               <Label className="text-body">Name</Label>
-              <Input
-                placeholder="Work laptop"
-                maxLength={MAX_PASSKEY_NAME_LENGTH}
-              />
+              <div className="flex flex-col gap-3 sm:flex-row sm:items-center">
+                <Input
+                  className="flex-1"
+                  placeholder="Work laptop"
+                  maxLength={MAX_PASSKEY_NAME_LENGTH}
+                />
+                <Button
+                  type="submit"
+                  variant="primary"
+                  isPending={register.isPending}
+                >
+                  Add a passkey
+                </Button>
+              </div>
               <FieldMessages>
                 <Description className="text-muted">
                   Optional. It is only a label, so you can tell this passkey
@@ -311,15 +326,6 @@ export function PasskeysCard() {
                 </Description>
               </FieldMessages>
             </TextField>
-            <div className="sm:pb-6">
-              <Button
-                type="submit"
-                variant="primary"
-                isPending={register.isPending}
-              >
-                Add a passkey
-              </Button>
-            </div>
           </form>
         ) : null}
       </Section>
--- a/web/src/features/account/PasskeysCard.tsx
+++ b/web/src/features/account/PasskeysCard.tsx
@@ -303,7 +303,7 @@ export function PasskeysCard() {
               className="flex max-w-2xl flex-col gap-1"
             >
               <Label className="text-body">Name</Label>
-              <div className="flex flex-col gap-3 sm:flex-row sm:items-center">
+              <div className="flex flex-col items-start gap-3 sm:flex-row sm:items-center">
                 <Input
                   className="w-full max-w-md"
                   placeholder="Work laptop"
--- a/web/design/forms.md
+++ b/web/design/forms.md
@@ -317,7 +317,7 @@ beside the input, and let the message sit under both.
 // Correct: one row for the input line, the message under the whole field
 <TextField className="flex max-w-2xl flex-col gap-1">
   <Label className="text-body">Name</Label>
-  <div className="flex flex-col gap-3 sm:flex-row sm:items-center">
+  <div className="flex flex-col items-start gap-3 sm:flex-row sm:items-center">
     <Input className="w-full max-w-md" />
     <Button type="submit" variant="primary">Add a passkey</Button>
   </div>
@@ -328,4 +328,8 @@ beside the input, and let the message sit under both.
 ```
 
 The input keeps `max-w-md`, the house field width; the wrapper is widened to hold
-the input, the gap and the button, so the field does not narrow to make room.
+the input, the gap and the button, so the field does not narrow to make room. The
+`items-start` is what keeps the button its own width once the row stacks: a column
+stretches its children by default, and a full-width button does not press (see
+[actions.md](actions.md)). The input carries `w-full` so the stretch it wanted is
+still the width it gets.
--- a/web/design/forms.md
+++ b/web/design/forms.md
@@ -307,3 +307,25 @@ guesses a number, and the guess is wrong the moment the caption is retuned.
 // Wrong: a hand-tuned nudge that does not track the caption
 <Button variant="ghost" className="pb-2" onPress={remove}>Remove</Button>
 ```
+
+A field whose action submits it is the one row `FieldAction` cannot answer: its
+message is a sentence long enough to wrap, and a wrapped message makes that field
+taller than the reserve the action holds. Put the button inside the field instead,
+beside the input, and let the message sit under both.
+
+```tsx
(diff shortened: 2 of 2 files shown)


web/src/features/account/PasskeysCard.tsx
```diff
@@ -288,38 +288,44 @@ export function PasskeysCard() {
...
-                placeholder="Work laptop"
-                maxLength={MAX_PASSKEY_NAME_LENGTH}
-              />
+              <div className="flex flex-col gap-3 sm:flex-row sm:items-center">
+                <Input
+                  className="flex-1"
+                  placeholder="Work laptop"
+                  maxLength={MAX_PASSKEY_NAME_LENGTH}
+                />
+                <Button
```
khaledosman, line 314: The Button is now a direct child of a `flex flex-col` container, so below `sm` it stretches to the full field width: `align-items` defaults to `stretch`, and the `<div>` wrapper this replaced is what kept it auto-width (same reason `PasswordCard.tsx:275` wraps its submit).

`web/design/actions.md`, "Sizes": a full-width button does not press at all, because the 0.98 press scale displaces too far on a wide element. Add `className="self-start sm:self-auto"`.

Related: `globals.css` gives `[data-slot="button"]` `min-height: 2.75rem` below 767px, so the 36px/36px match verified in the browser holds at >=768px only; in the 640-767px band it is a 44px button centered against a 36px input.

web/src/features/account/PasskeysCard.tsx
```diff
@@ -288,38 +288,44 @@ export function PasskeysCard() {
           is exactly when somebody has orphans to clear out. */}
         {passkeys_ready && canUsePasskeys && !passkeys.isError ? (
           <form
-            className="flex flex-col gap-3 sm:flex-row sm:items-end"
             onSubmit={(event) => {
               event.preventDefault()
               startRegistration()
             }}
           >
+            {/* The button sits beside the input rather than beside the whole
```
khaledosman, line 296: This is exactly the limit `FieldAction`'s docstring documents ("a message long enough to wrap makes that one field taller and drops it back out of line, and no reserve here can answer that"), and this is the shape that answers it. But `web/design/forms.md` "Control rows" still prescribes `FieldAction` and offers nothing for the wrapping case, so the next person hits the same wall, and `PolicyForm`'s guardrails row is the case `FieldAction` names as still broken.

Put the rule in `forms.md` and cut this comment to a pointer; six lines restating a design doc is a fact in two places, and one of them will go stale.

khaledosman (approved):
`check-template` is failing: the body dropped `## PR Type`, `## Checklist` and `## AI Usage`. The PR already carries the `missing-template` label, and `pr-template-check.yml` closes it 24 hours after that label lands (opened 08:42 today). Restore the three sections.

`FieldAction` genuinely cannot fix this one, and its own docstring says why, so the restructure is the right call.

Verified: `dashboard`, `e2e`, `build`, `serving` and `catalog` pass; `PasskeysCard.test.tsx:219` clicks "Add a passkey", so submit still works through the new nesting (RAC's `TextField` provides no `ButtonContext`, so a nested button cannot be hijacked). No generated artifacts owed.

---

web/src/features/account/PasskeysCard.tsx
```diff
@@ -288,38 +288,44 @@ export function PasskeysCard() {
...
               onChange={setNewName}
-              className="flex max-w-md flex-1 flex-col gap-1"
+              className="flex max-w-xl flex-col gap-1"
             >
               <Label className="text-body">Name</Label>
-              <Input
-                placeholder="Work laptop"
-                maxLength={MAX_PASSKEY_NAME_LENGTH}
-              />
+              <div className="flex flex-col gap-3 sm:flex-row sm:items-center">
```
jigjigjig: `FieldAction` (`@/design-system/forms/FieldAction`) is the documented answer to a trailing action landing level with the caption instead of the input, and `web/design/forms.md` ("Control rows") names it as the rule. This lands a second, undocumented pattern for the same problem without saying why.

The why is real: `FieldAction`'s own docstring records that it fails when a field's message wraps to two lines, which is exactly this case. So keep this shape, but put the pattern where the next person will find it: a short entry under "Control rows" in `web/design/forms.md` for a field whose action submits it, and cut the six-line comment here down to a pointer.

web/src/features/account/PasskeysCard.tsx
```diff
@@ -288,38 +288,44 @@ export function PasskeysCard() {
...
+              field, because the description below it is a paragraph whose
+              height depends on the width it is given. Aligned against the
+              field, the button would have to be pushed up by however tall
+              that paragraph happened to render, which is a number no class
+              can know. */}
             <TextField
               value={newName}
               onChange={setNewName}
-              className="flex max-w-md flex-1 flex-col gap-1"
+              className="flex max-w-xl flex-col gap-1"
```
jigjigjig: `max-w-md` on the field wrapper is the house width, repeated in `providerFields.tsx`, `ModelScopeControl.tsx` and the two guardrail forms. Widening the wrapper to `max-w-xl` keeps the box roughly the old size but leaves the input itself at 418px, so this one text box renders narrower than every other one in the dashboard.

Cap the input rather than the wrapper: `w-full max-w-md` on the `Input`, and a wrapper wide enough to hold input + gap + button (`max-w-2xl`). The input then measures the same 448px it does everywhere else, and the description still reads on one line.

web/src/features/account/PasskeysCard.tsx
jigjigjig: Six lines narrating what the previous layout did wrong. `AGENTS.md` asks for a comment only where the logic is not obvious, and specifically not for what the code used to do, which is the commit message's job. Two lines saying why the button is inside the field is enough once `forms.md` carries the pattern.

web/src/features/account/PasskeysCard.tsx
```diff
@@ -288,38 +288,44 @@ export function PasskeysCard() {
...
+                  maxLength={MAX_PASSKEY_NAME_LENGTH}
+                />
+                <Button
+                  type="submit"
+                  variant="primary"
+                  isPending={register.isPending}
+                >
+                  Add a passkey
+                </Button>
+              </div>
```
jigjigjig, line 319: Below `sm` this reorders the stack: the hint used to sit under the input and now sits under the button, so on a phone the sentence describing the name box is separated from it by a button. Worth a deliberate answer rather than a side effect.

jigjigjig (commented):
Self-review, by Claude Opus 5 in Claude Code. The alignment itself is right (input and button measure 36px with an identical `top` in a browser), so these are about which idiom the repo already has for this and about a width that drifts off the house value.

web/src/features/account/PasskeysCard.tsx
jigjigjig: Both right, and the stretch was a real regression: below the breakpoint the button filled the field. Fixed in 35be63d6 with `items-start` on the row rather than `self-*` on the button, so any future child of that row keeps its own width too; the input carries `w-full`, so it still fills the line. Measured with the row forced to its stacked classes: button 130px, input 448px.

On the 640-767px band: leaving it. The 44px floor is the touch target the rule intends, and `items-center` puts the 36px input on the button's center line, which is the treatment that band wants. A 36px button there would be the bug.

web/src/features/account/PasskeysCard.tsx
jigjigjig: Agreed, and done in fdacc986: `forms.md` "Control rows" now carries the wrapping case beside the `FieldAction` rule, with the shape and why the input keeps `max-w-md`. The comment in the card is down to three lines pointing there.

Left `PolicyForm`'s guardrails row alone: that is the case the docstring names as still broken and it is not this PR's.

web/src/features/account/PasskeysCard.tsx
jigjigjig: Done in fdacc986. `forms.md` "Control rows" now documents the field-with-submit shape after the `FieldAction` rule, and the comment here is a pointer to it.

web/src/features/account/PasskeysCard.tsx
jigjigjig: Done in fdacc986: `w-full max-w-md` on the `Input`, `max-w-2xl` on the wrapper. Measured in the browser at 448px, level with the email and password boxes above it on the same page.

web/src/features/account/PasskeysCard.tsx
jigjigjig: Done in fdacc986: three lines, naming the wrapping case and pointing at `forms.md`.

web/src/features/account/PasskeysCard.tsx
jigjigjig: Deliberate, and keeping it. The hint is about the passkey being named, not about the box specifically, and the label above the input is what names the control. Splitting it (hint under the input when stacked, under the row otherwise) would mean two copies of the sentence or an order swap at the breakpoint, and neither is worth 36px of separation on a phone.

jigjigjig:
@jigjigjig ready for you.

- Issue #1085 is open, assigned to you, and linked as a sub-issue of the UI-improvements epic (mozilla-ai/otari-ai#2086).
- Body now follows the PR template; the title check and the template check both pass.
- Six review comments, all answered and resolved: two of them were real (a wrapped field caption is the case `FieldAction` cannot reserve, so the pattern is now written into `web/design/forms.md`; and the stacked row was stretching the button to full width, fixed with `items-start`).
- One decision left as-is on purpose: below the breakpoint the hint sits under the button rather than under the box. Say the word and I will split it.
- CI green. The e2e run failed once on the routing policy dialog, which this PR does not touch, and passed on a rerun.

PR #1093, opened 2026-09-11
feat(dashboard): add animated background to login and signup

### LINKED ISSUE #991: Sign-in screen reads empty: center the auth column and use the ground beside it
## Context

Improvement request from Slack ([thread](https://mzlai.slack.com/archives/C0BNZGFQ7TJ/p1788945096434489)):

> Hey guys I know we're going simple but is this maybe too simple? The page looks empty
>
> Desktop, regular size. This is a looooot of white space with nothing, not even an animation, and i have to look to the right. [...] really not engaging imho
>
> I wouldn't say it's the same. Let's at least put it in the centre. [...] this left alignment looks just off

The sign-in screen (and every other page in front of a session) renders through `AuthPageShell` in `web/src/features/auth/PublicAuthLayout.tsx`: a 56px bar with the mark, a `max-w-[520px]` column pinned to the page's left gutter with a full-height rule down its right, and deliberately empty ground to the right of it. On a desktop window that leaves most of the viewport blank, and the one thing to do on the page sits off to one side.

`Login.tsx` composes the same shell itself for its three states, so both files change together.

## What to do

Two levels, and the first is the floor:

1. **Center the column.** The minimum Adrian asked for. Note that `AuthPageShell`'s doc comment currently argues *for* pinning left (a centered column has no edge to align to once elevation is zeroed, so the page's own gutter supplies one). Overriding that means the centered column needs its own real edge, and the comment needs rewriting to state the decision that actually holds, not left describing the old one.
2. **Give the empty half something to do.** Fede posted an alternative in the thread with the ground used rather than left blank. Whatever lands there should stay operator-tool appropriate: no marketing chrome, no autoplaying motion, and it must respect `prefers-reduced-motion` and the operator's theme choice (the shell deliberately does not pin the auth flow to light).

Constraints to preserve while moving the column:

- `Login.tsx` centers its card on a **constant** half-height (17.5rem), not the card's measured height, so that opening the first-run disclosure or wrapping a refusal grows the card downward instead of walking the submit button out from under the pointer. Do not replace that with `items-center`; the file's comment records exactly why it was tried and reverted.
- `vh`, not `dvh`, for the same reason recorded there: a phone's soft keyboard must not re-center the card mid-typing.
- The shell is shared by sign-in, signup, check-email, verify-email, resend-verification, recover-password, reset-password and the OAuth callback. Check all of them, plus each of Login's three states (master key, password, unavailable), and the narrow/mobile breakpoint.
- Existing test suites cover these pages (`Login.test.tsx`, `PublicAuthPage.test.tsx`, `PublicAuthLayout.stories.tsx`, and the per-page tests); update them with the layout.

## Notes

- Adrian said "hold for now, bigger issues at hand" in the thread, so this is queued, not urgent. Filed so it is not lost.
- Follow `.github/skills/frontend-standards/SKILL.md` for tokens and component conventions.

jigjigjig:
## Description

Replace the public authentication presentation with the saved background from the login background studio: slow waves of blue bars, a centered form panel, Otari branding, and a footer. The background follows the existing light and dark design tokens and aligns its bar grid to the form.

Reduced-motion preferences render a static frame, and animation stops while the tab is hidden. Existing credential handling, signup validation, and deployment gates remain intact.

All public auth and invitation pages share the same horizontally centered card and animated background. The card grows downward from a viewport-based top offset, so disclosures and errors do not move its heading. The appearance control cycles through system, light, and dark. The full background is painted to avoid blank strips while scrolling. Painting is limited to 24 FPS and slows further on large grids using a 24,000-cell-per-second budget; cells hidden by the form are skipped.

The selected studio design deliberately uses a square auth card, slightly rounded decorative bars, accent corner marks, and a Mozilla AI footer, including on standalone deployments. These auth-only exceptions are documented in the design system; dashboard pages keep the flat-band rules.

## How to test it locally

1. Check out `jigjigjig/auth-animated-background` and run your configured gateway.
2. Run `pnpm --dir web install --frozen-lockfile`, then `pnpm --dir web run dev` (set `OTARI_DEV_API` if the gateway is not on port 8000).
3. Open the signed-out dashboard. Check master-key and password sign-in on deployments that offer them. Toggle the theme and check the background animation.
4. On a mail-enabled deployment, open `/#/signup`. Check the account-claim form and invitation email prefill.
5. Try a phone-width viewport, scrolling to the footer, and the system reduced-motion setting.

Validation completed locally: 5,700 dashboard tests; TypeScript; Biome; `make lint` (including architecture and migration-head checks); production build; Storybook build. The installed dependency binaries were used directly for the dashboard checks. The production bundle was also inspected in a browser with mock bootstrap data in both themes and at 390px width. Backend tests were not run because this change touches only dashboard presentation.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes https://github.com/mozilla-ai/otari/issues/991

## Checklist

The applicable dashboard checks are listed above. No API contract changed.

## AI Usage

- [ ] No AI was used.

**Any additional AI details you'd like to share:** Adapted the existing local studio renderer and saved settings, integrated the production auth screens, and ran local validation. Local testing is complete; the PR is ready for review.

## Summary

- Added a shared, centered authentication layout for public login, signup, verification, recovery, reset, invitation, and OAuth pages.
- Added an animated blue-bar background with light, dark, system, and reduced-motion support.
- Preserved stable card positioning across errors, disclosures, responsive layouts, and mobile keyboard use.
- Added rendering bounds, visibility handling, frame-rate limits, and focused tests for the background animation.
- Updated authentication tests and design documentation to reflect the shared layout and theme requirements.

## Benefits

- Public authentication pages use space more effectively.
- Authentication flows now have a consistent visual treatment.
- Users can control appearance and motion behavior without changing authentication logic.


web/src/features/auth/login.css
```diff
@@ -0,0 +1,101 @@
...
+  border-bottom: 1px solid var(--color-border);
+}
+
+.login-footer {
+  border-top: 1px solid var(--color-border);
+  font-size: 0.75rem;
+  color: var(--color-text-subtle);
+}
+
+.login-main {
```
jigjigjig: **Centering reintroduces the layout jump the old shell was built to avoid.**

`.login-main` is `justify-content: center` and `.login-panel` is `align-items: center`, so the content's top edge sits at `mainCenter - contentHeight/2`. `Login.tsx:677` renders the "First run? Where to find your key" `<details>` directly above the submit button; opening it grows the content by ~70px, which moves the submit button ~35px upward under a pointer that is already resting near it.

The old `AuthPageShell` was top-anchored (`py-10`, content grows downward), and commit 1 deletes the long comment in `Login.tsx` that documented exactly this defect while removing the behavior it described.

The panel-height quantization does not cancel it: with `P = ceil((H+2)/pitch)*pitch`, `mainCenter - P/2 + (P-H)/2 = mainCenter - H/2`.

web/src/features/auth/login.css
```diff
@@ -0,0 +1,101 @@
...
+  z-index: 1;
+  display: flex;
+  flex: 1;
+  flex-direction: column;
+  justify-content: center;
+  align-items: center;
+  padding: 3rem 1rem;
+}
+
+.login-panel {
```
jigjigjig: Second half of the centering above: `align-items: center` here is what makes the panel's own growth symmetric about the midpoint. Top-anchoring the panel (and `justify-content: flex-start` on `.login-main`) keeps the background while restoring the grows-downward behavior.

web/src/features/auth/PublicAuthLayout.tsx
```diff
@@ -50,33 +27,24 @@ export function AuthPageShell({ children }: { children: ReactNode }) {
...
- * self-hosted dashboard is an operator tool whose theme is the operator's
- * choice, so what survives the port is a heading, a body, and where to go next.
- */
+/** Shared auth content, with a replaceable frame for login and signup. */
 export function PublicAuthLayout({
   title,
   description,
   children,
   footer,
+  shell: Shell = AuthPageShell,
```
jigjigjig: **Only four of the pre-session pages get the new shell**, so the layout flips mid-flow.

Still on `AuthPageShell`: `RecoverPasswordPage`, `ResetPasswordPage`, `VerifyEmailPage`, `OAuthCallbackPage`, both "Not available on this gateway" panels (`PublicAuthPage.tsx:83,115`), and `features/invitations/AcceptInvitationPage.tsx:88`.

Concretely: sign-in (centered card, animated background) -> "Forgot your password?" (left-pinned column, rule, no background) -> check-email (centered card, animated background again). Two flips inside one recovery flow. An invited user hits the old shell on the invitation page and the new one on sign-in.

web/src/features/auth/LoginPageShell.tsx
```diff
@@ -0,0 +1,91 @@
...
+          <span className="text-title">Otari</span>
+        </div>
+        <div className="flex items-center gap-1">
+          <Button
+            variant="ghost"
+            isIconOnly
+            aria-label={
+              resolved === "dark" ? "Use light theme" : "Use dark theme"
+            }
+            onPress={() =>
```
jigjigjig: **This toggle permanently discards a `"system"` theme preference.**

It is two-state and writes `"light"` or `"dark"` straight into `localStorage`, but `app/nav/AccountMenu.tsx:268-280` cycles all three (`THEME_PREFERENCES`). An operator who has never signed in and clicks this once out of curiosity is pinned to a fixed theme with no way back to follow-the-OS from the signed-out screen: they have to sign in and find the account menu. The pre-paint script in `index.html` then honors the pinned value on every load.

Either cycle all three states here, or drop the control from the signed-out header.

web/src/features/auth/background/renderBars.ts
```diff
@@ -0,0 +1,65 @@
...
+  const offsetX = (pitchX - barWidth) / 2
+  const offsetY = (pitchY - barHeight) / 2
+  const radius = barWidth * config.rounding
+
+  ctx.globalAlpha = 1
+  ctx.fillStyle = palette.background
+  ctx.fillRect(0, 0, width, height)
+  ctx.fillStyle = `color-mix(in oklab, ${palette.foreground}, ${palette.accent} ${config.tint * 100}%)`
+
+  for (
```
jigjigjig, line 40: **The draw loop is sized to the whole canvas, not to the visible band.**

With `columns: 20` and a 28rem panel, `pitchX ~ 22.4px` and `pitchY ~ 47.6px`; on a 2560x1440 display that is ~114 x 30 = ~3,400 `roundRect` fills plus ~3,400 `globalAlpha` state changes every animation frame, at full rAF rate even though `speed` is 0.2.

It includes the rows behind `.login-header` and `.login-footer`, which are opaque (`background: var(--color-background)`) and can never show them. Commit 2 removed the pause control, so a sign-in screen left open on a kiosk or an ops monitor burns this indefinitely; `prefers-reduced-motion` is the only escape.

web/src/features/auth/LoginPageShell.tsx
```diff
@@ -0,0 +1,91 @@
...
+      if (!pitch) return
+      const rows = Math.ceil(
+        (content.getBoundingClientRect().height + 2) / pitch,
+      )
+      // Geometry is measured once per resize; animation never changes layout.
+      panel.style.setProperty("--login-panel-height", `${rows * pitch}px`)
+    }
+    align()
+    let frame = 0
+    const resize = new ResizeObserver(() => {
```
jigjigjig: `align()` is deferred one frame (`ResizeObserver` -> `requestAnimationFrame(align)`), so any content height change paints twice: once at the natural height while `min-height` is still the stale multiple, and again a frame later at the new quantized multiple. Opening the disclosure on the sign-in card makes it settle in two visible steps rather than one.

web/src/features/auth/background/renderBars.ts
```diff
@@ -0,0 +1,65 @@
+import { barLuminance } from "./barWave"
```
jigjigjig, line 1: **`renderBars.ts` has no test at all.** `LoginBackground.test.tsx` mocks `drawBars` and `LoginPageShell.test.tsx` mocks `LoginBackground`, so the loop bounds, the `luminance ** contrast * intensity` alpha, and the `color-mix(...)` `fillStyle` string never execute.

Two failure modes that would ship silently: an alpha outside `[0,1]` throws on the canvas call at runtime, and if `palette.accent` ever resolves empty the `fillStyle` assignment is ignored and the bars draw in the background color (invisible, no error).

web/src/features/auth/background/LoginBackground.tsx
```diff
@@ -0,0 +1,120 @@
...
+    }
+    const update = () => {
+      paint()
+      if (canAnimate()) {
+        if (!frame) frame = requestAnimationFrame(animate)
+      } else {
+        cancelAnimationFrame(frame)
+        frame = 0
+        lastFrame = 0
+      }
```
jigjigjig: Dead after commit 2 removed the pause button: `paused={false}` is hardcoded at `LoginPageShell.tsx:46`, so the `paused` prop, `settings.current.paused`, and the `!settings.current.paused` term in `canAnimate` are all unreachable.

`LoginPageShell.test.tsx` still asserts `getByText("Motion playing")`, which passes against the local mock only and exercises none of the shipped component, so the dead branch reads as covered.

jigjigjig (commented):
Review at high effort. Nine findings; the first two are the ones I would fix before merge.

Verified as **not** problems, so they do not need re-checking: `useTheme` outside a provider (`app/provider.tsx:35` wraps `App`, and both touched suites go through `AppProviders`), the canvas under jsdom (`getContext("2d")` returns null so the effect early-returns, and `ResizeObserver` is stubbed in `src/tests/setup.ts`), the `barWave.ts` numerics (no division by zero, all four `frequency` values non-zero multiples of the fundamental, output bounded to `[0.06, 0.92]`), the pause/resume/visibility state machine in `LoginBackground`, and the `Button` swap on `SignupPage`/`ResendVerificationPage` (both pass `variant="primary"` explicitly).

web/src/features/auth/login.css
```diff
@@ -0,0 +1,101 @@
+.login-page {
```
jigjigjig: A standalone 101-line stylesheet side-imported from a component (`import "./login.css"`) departs from `.github/skills/frontend-standards/components.md`, which puts custom CSS at rung 4 and in `globals.css` under the `.otari-*` namespace.

The practical cost: the file ships unlayered, so it outranks `@layer utilities`. A later `className="min-h-0"` or `className="items-start"` on `.login-panel` silently loses to these rules, which is the "takes the call site's ability to restyle it away" problem that document calls out.

web/src/features/auth/LoginPageShell.tsx
```diff
@@ -0,0 +1,91 @@
...
+  return (
+    <div className="login-page">
+      <LoginBackground
+        panelRef={panelRef}
+        config={savedBackground}
+        paused={false}
+      />
+      <header className="login-header">
+        <div className="flex items-center gap-3">
+          <img
```
jigjigjig, line 22: `className="h-6 w-auto"` with no `width`/`height` attributes: `favicon.svg` is 273x250, so the rendered width is 0 until the file loads and the "Otari" wordmark shifts right on first paint. The shell this replaces used a fixed `w-[26px]` (`PublicAuthLayout.tsx:15`) for that reason.

web/src/features/auth/PublicAuthLayout.tsx
khaledosman: Half the flow keeps the old shell. `RecoverPasswordPage`, `ResetPasswordPage`, `VerifyEmailPage`, `OAuthCallbackPage`, `PublicAuthPage` and `AcceptInvitationPage` still render `AuthPageShell`.

The redesigned login's own footer links split three ways: `#/signup` and `#/resend-verification` get the new shell, `#/recover-password` does not. And `CheckEmailPage` is new while `VerifyEmailPage`, where the mailed link actually lands, is old.

Either pass `shell={LoginPageShell}` at every `PublicAuthLayout` call site and move `AcceptInvitationPage` with them, or keep the new shell off `CheckEmailPage`/`ResendVerificationPage` too so the split is at least a whole flow.

web/src/features/auth/login.css
khaledosman: `web/design/layout.md:4`: "no card anywhere: two bands are ...". `DESIGN.md`: "Radius is `0px`. One key in `globals.css` squares every corner in the app, HeroUI's internals included. Never add `rounded-*`."

`.login-panel` is a bordered surface box floating on a ground, which is a card. And `renderBars.ts:60` draws every bar with `ctx.roundRect(..., radius)` at `rounding: 0.11`, which reintroduces radius in the one place no lint rule and no `foundation.test.ts` sweep can see it.

Both are product decisions rather than mistakes, so say which one was made. If the flat-plane rule still holds, this is a square panel and `rounding: 0`; if it does not, `DESIGN.md` and `layout.md` need to change in this PR.

web/src/features/auth/background/renderBars.ts
```diff
@@ -0,0 +1,65 @@
...
+    ) {
+      const x = left + col * pitchX
+      const y = top + row * pitchY
+      const luminance = barLuminance(
+        (x + pitchX / 2) / width,
+        (y + pitchY / 2) / height,
+        time,
+        config.waveScale,
+      )
+      ctx.globalAlpha = luminance ** config.contrast * config.intensity
```
khaledosman, line 66: `web/design/colors.md:83`: "`bg-accent/…` is not how you get a tint. Use `--color-primary-subtle`."

This paints `--color-primary` across the whole viewport at a per-bar `globalAlpha`, which is that rule in canvas form. On the same screen the accent is also spent on two corner marks (`login.css:74-86`) and the primary Sign in button, against `DESIGN.md`'s "one teal accent spent once per screen".

If the background is meant to carry the accent, the button and the corner marks are the two that have to give.

web/src/features/auth/login.css
khaledosman: This is the only hand-written stylesheet under `src/` besides `globals.css`, and standards rung #2 puts it last: a variable, then a shared utility, then the component's own prop, and only then a rule.

Nearly every rule here already has a utility. `.login-header` is `relative z-1 flex min-h-14 shrink-0 items-center justify-between gap-4 border-b border-border bg-background px-6`; `.login-main` is `relative z-1 flex flex-1 flex-col items-center justify-center px-4 py-12`; `.login-panel-content` is `flex w-full min-w-0 flex-col gap-6 p-8`.

Two costs beyond the style rung. `foundation.test.ts` sweeps only `.tsx`/`.ts`, so the raw `font-size: 0.75rem` on line 30 is invisible to the gate that exists to catch exactly that. And the file is unlayered while Tailwind's utilities sit in `@layer utilities`, so every rule here outranks a class a child sets on the same element.

web/src/features/auth/LoginPageShell.tsx
```diff
@@ -0,0 +1,91 @@
...
+      cancelAnimationFrame(frame)
+      resize.disconnect()
+    }
+  }, [])
+  return (
+    <div className="login-page">
+      <LoginBackground
+        panelRef={panelRef}
+        config={savedBackground}
+        paused={false}
```
khaledosman: `paused={false}` is the only call site, and `config` is a static JSON import that never changes either.

So the `paused` prop, `settings.current.paused` in `LoginBackground`, and the whole first `useEffect` that syncs the ref and calls `redraw` are dead: they exist for the studio control removed in `eeef4be`. Three of the four tests in `LoginBackground.test.tsx` ("pauses without advancing time", "keeps the animation clock moving while controls are adjusted", and the `paused` half of the lifecycle test) cover a surface that does not ship, and `LoginPageShell.test.tsx` asserts on a mock rendering that prop.

Drop the prop and the ref sync, or land the control that justified them.

web/src/features/auth/background/renderBars.ts
khaledosman, line 40: The bar field scales with viewport area and nothing caps it. `columns` fixes the pitch relative to the panel, not the screen: at the panel's 28rem cap, `pitchX` is 22.4px and `pitchY` about 47.6px whatever the window is.

So 1920x1080 is roughly 2,000 `roundRect` + `fill` calls and 8,000 `Math.sin` per frame; 2560x1440 is about 3,600 and 14,000. That runs at 60fps indefinitely on a page people leave open in a tab, and the only stops are reduced motion and a hidden tab.

Either cap the field (a maximum row and column count, letting the pitch grow past it), or batch: quantize `luminance ** contrast * intensity` into a dozen buckets and emit one path per bucket, which turns thousands of alpha changes and fills into a dozen.

web/src/features/auth/LoginPageShell.tsx
```diff
@@ -0,0 +1,91 @@
...
+          <span aria-hidden className="login-corner login-corner-start" />
+          <span aria-hidden className="login-corner login-corner-end" />
+          <div ref={contentRef} className="login-panel-content">
+            {children}
+          </div>
+        </div>
+      </main>
+      <footer className="login-footer">
+        <span>One gateway. Every model.</span>
+        <span className="font-mono text-xs">Mozilla AI</span>
```
khaledosman: The docstring this PR deletes from `PublicAuthLayout.tsx` argued this footer away by name: the platform's `PreloginLayout` byline and marketing footer "are hosted-product chrome that has no counterpart here", because a self-hosted dashboard is an operator tool. This adds both back with the argument removed rather than answered. If the decision was reversed, that belongs in the PR description; otherwise drop the footer.

Separately, `font-mono text-xs` on line 87 hand-rolls a mono type role. `text-mono-caption` (13px) and `text-mono-micro` (11px) both exist. `foundation.test.ts` only catches `text-xs` paired with `text-muted`, so this one slips the gate rather than being allowed by it.

web/src/features/auth/login.css
```diff
@@ -0,0 +1,101 @@
...
+  color: var(--color-text-subtle);
+}
+
+.login-main {
+  position: relative;
+  z-index: 1;
+  display: flex;
+  flex: 1;
+  flex-direction: column;
+  justify-content: center;
```
khaledosman: `justify-content: center` on a panel whose height tracks its content. `LoginPageShell.tsx:27` sets `--login-panel-height` from the measured content height, so opening the `<details>` disclosure at `Login.tsx:677`, or a gateway refusal wrapping onto a second line, grows the panel and moves the submit button up by half the growth, out from under a pointer already resting on it.

Anchor the panel's top edge instead (`justify-content: flex-start` with a padding-top, or a top offset computed from a constant rather than from the content's own height), so it grows downward.

web/src/features/auth/login.css
```diff
@@ -0,0 +1,101 @@
+.login-page {
+  position: relative;
+  isolation: isolate;
+  display: flex;
+  flex-direction: column;
+  min-height: 100vh;
```
khaledosman: `100vh` is the viewport without the phone's browser chrome subtracted, so the footer sits below the visible area on iOS Safari and the panel centers in a box taller than the screen. Use `100dvh`, or `100svh` if you want the tallest-chrome case to be the stable one.

(The `vh`-over-`dvh` argument in the code this PR replaces was about a fixed *top offset* re-centering when the soft keyboard opens; it does not carry to a `min-height` on the page.)

web/src/features/auth/background/renderBars.ts
```diff
@@ -0,0 +1,65 @@
...
+  const barWidth = pitchX * (1 - config.spacing)
+  const barHeight = pitchY * (1 - config.spacing * (0.134 / 0.248))
+  const offsetX = (pitchX - barWidth) / 2
+  const offsetY = (pitchY - barHeight) / 2
+  const radius = barWidth * config.rounding
+
+  ctx.globalAlpha = 1
+  ctx.fillStyle = palette.background
+  ctx.fillRect(0, 0, width, height)
+  ctx.fillStyle = `color-mix(in oklab, ${palette.foreground}, ${palette.accent} ${config.tint * 100}%)`
```
khaledosman: At the shipped `tint: 1` this mix resolves to pure `--color-primary`, so `palette.foreground` is read from the computed style on every `measure()` and never used.

It is also the one silent-failure path in the renderer: if a browser cannot parse `color-mix()` in a canvas color, the assignment is a no-op rather than a throw, `fillStyle` stays at `palette.background` from line 36, and every bar paints invisible against the ground with nothing in the console.

Use `palette.accent` directly and drop `tint` and `foreground`.

khaledosman (changes requested):
Reviewed at `21c7a50` against `AGENTS.md`, the frontend standards skill, the three `.github/instructions/` files, and `web/design/`.

Four of the ten points below are design-system decisions rather than code nits (the card, the radius, the accent budget, the marketing footer) and need an answer rather than an edit. The rest are code: a half-migrated auth flow, a dead prop with tests over it, an uncapped per-frame bar count, a growing panel that is vertically centered, `100vh`, and a `color-mix()` in a canvas color.

Checked and not findings: reduced-motion and `visibilitychange` handling is correct, and the screenshot projects pin `reducedMotion: "reduce"`, so the canvas renders a deterministic t=0 frame and baselines stay stable; the 44px touch floor comes from the global `[data-slot="button"]` rule in `globals.css`, so the header toggle is covered; the master-key toggle recentering is a real fix.

web/src/features/auth/login.css
jigjigjig: Addressed in a9848280: The card now uses a viewport-based top offset and grows downward. Browser verification kept the heading at y=169 before and after opening the disclosure.

web/src/features/auth/login.css
jigjigjig: Addressed in a9848280: Removed both vertical-centering layers. The panel is a normal top-aligned flex column.

web/src/features/auth/PublicAuthLayout.tsx
jigjigjig: Addressed in a9848280: PublicAuthLayout now uses LoginPageShell by default, and invitation acceptance uses it directly. Recovery, reset, verification, OAuth callbacks, and unavailable states all share the frame.

web/src/features/auth/LoginPageShell.tsx
jigjigjig: Addressed in a9848280: The appearance control now cycles system, light, and dark using THEME_PREFERENCES. Its test verifies returning to the system preference in storage.

web/src/features/auth/background/renderBars.ts
jigjigjig: Addressed in a9848280: Moved the canvas into the main band, capped animation painting at 24 FPS, restricted rows to the viewport, and skipped cells hidden by the opaque form. Bitmap resizing now occurs only when its dimensions change.

web/src/features/auth/LoginPageShell.tsx
jigjigjig: Addressed in a9848280: Removed panel-height quantization and its ResizeObserver/rAF layout writes entirely. Content now grows in one normal layout pass.

web/src/features/auth/background/renderBars.ts
jigjigjig: Addressed in a9848280: Added renderBars.test.ts exercising the real renderer: visible bounds, panel occlusion, wave-derived opacity, theme color mixing, alpha range, and empty geometry.

web/src/features/auth/background/LoginBackground.tsx
jigjigjig: Addressed in a9848280: Removed the paused prop, settings ref, and pause branches. The shell test now checks the real theme control; lifecycle tests cover frame limiting, reduced motion, hidden tabs, and cleanup.

web/src/features/auth/login.css
jigjigjig: Addressed in a9848280: Deleted login.css and its side import. The frame now uses Tailwind layout utilities and existing semantic tokens, so utilities retain normal cascade behavior.

web/src/features/auth/LoginPageShell.tsx
jigjigjig: Addressed in a9848280: Added intrinsic width/height attributes and an explicit width matching the mark aspect ratio, reserving space before the SVG loads.

web/src/features/auth/PublicAuthLayout.tsx
jigjigjig: Addressed in a9848280. PublicAuthLayout now uses LoginPageShell by default, and AcceptInvitationPage uses it directly. Recovery, reset, verification, OAuth callbacks, and unavailable states share the same frame.

web/src/features/auth/login.css
jigjigjig: Addressed in 4c8cdaa8. The card and decorative bar shapes are intentional parts of the design selected by the product owner. DESIGN.md and layout.md now document the narrow public-auth exception, while retaining the flat-band and square-control rules for the dashboard.

web/src/features/auth/background/renderBars.ts
jigjigjig: Addressed in 4c8cdaa8. The selected design intentionally combines the decorative field, corner marks, and primary action. The auth-only exception is now explicit in colors.md and layout.md; content remains on an opaque semantic surface.

web/src/features/auth/login.css
jigjigjig: Addressed in a9848280. Deleted login.css and converted the shell to Tailwind utilities and semantic tokens, including the existing mono caption role.

web/src/features/auth/LoginPageShell.tsx
jigjigjig: Addressed in a9848280. Removed the paused prop, settings ref, and redraw synchronization. Tests now exercise the shipped theme control and animation lifecycle.

web/src/features/auth/background/renderBars.ts
jigjigjig: Updated after subsequent revisions: the grid-density cap in 4c8cdaa8 was removed to preserve the product-approved studio bar sizes. Current head 71e7f039 instead limits animation to 24 FPS and reduces frequency further using a conservative 24,000-cell-per-second budget. The full canvas is painted to avoid scroll gaps, and cells hidden by the form are skipped. Regression tests cover preserved sizes and reduced frequency on large grids.

web/src/features/auth/LoginPageShell.tsx
jigjigjig: Addressed in a9848280 and 4c8cdaa8. The footer now uses text-mono-caption. The product owner selected this branding for public auth, including standalone deployments; the decision is recorded in the PR description and design-system layout guidance.

web/src/features/auth/login.css
jigjigjig: Addressed in a9848280. Removed content-height quantization and both vertical-centering layers. The card grows downward from a viewport-based constant top offset. Browser verification confirmed that opening the disclosure leaves the heading position unchanged.

web/src/features/auth/login.css
jigjigjig: Addressed in 4c8cdaa8. The page now uses min-h-svh so mobile browser chrome does not inflate the minimum page height.

web/src/features/auth/background/renderBars.ts
jigjigjig: Addressed in 4c8cdaa8. The renderer assigns palette.accent directly; removed tint and the foreground style read. The renderer test verifies the direct accent assignment.

web/src/features/auth/background/renderBars.ts
```diff
@@ -0,0 +1,91 @@
...
+  const pitchY = pitchX * BAR_ROW_RATIO
+  const barWidth = pitchX * (1 - config.spacing)
+  const barHeight = pitchY * (1 - config.spacing * (0.134 / 0.248))
+  const offsetX = (pitchX - barWidth) / 2
+  const offsetY = (pitchY - barHeight) / 2
+  const radius = barWidth * config.rounding
+
+  ctx.globalAlpha = 1
+  ctx.fillStyle = palette.background
+  ctx.clearRect(0, 0, width, height)
```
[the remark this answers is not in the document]
jigjigjig: Fixed: clearing now validates and uses only the visible band. The tall-canvas regression test checks the exact clear rectangle, and empty bands do not clear. Renderer tests, TypeScript, Biome, and production build pass.

web/src/features/auth/background/renderBars.ts
```diff
@@ -0,0 +1,86 @@
...
+    height,
+    left,
+    top,
+    panelWidth,
+    panelHeight,
+    visibleTop,
+    visibleBottom,
+  } = geometry
+  if (panelWidth <= 0 || width <= 0 || height <= 0) return
+  const pitchX = panelWidth / config.columns
```
khaledosman, line 27: The density cap was added in `4c8cdaa8` and removed again in `f66f7331`, so it is not in the branch. The reply on the earlier thread still reads "Grid pitch now grows to cap visible density at 64 columns and 32 rows plus edge cells. A real-renderer test verifies the bound at 7680 by 4320" — that is no longer true of head and should be corrected either way.

The test that replaced it asserts the opposite. `renderBars.test.ts:80`, "preserves studio bar dimensions at viewport width %i", pins `barWidth` to `pitch * 0.95` where `pitch = panelWidth / 20` at every width including 7680, which is exactly the statement that the grid does *not* coarsen. That row runs about 343 columns.

With rows band-limited, 2560x1440 is still roughly 114 columns by 28 rows, about 3,000 `roundRect` + `fill` calls per paint. If the studio sizing has to be preserved exactly, then bound the work somewhere else (a hard cell budget that stops the loop, or a lower frame cap on large viewports); it cannot be unbounded in both dimensions at once.

web/src/features/auth/background/LoginBackground.tsx
```diff
@@ -0,0 +1,137 @@
...
+      if (geometryDirty) measure()
+      if (paletteDirty) readPalette()
+      geometryDirty = false
+      paletteDirty = false
+      const moving = canAnimate()
+      if (moving && lastFrame)
+        time += Math.min((now - lastFrame) / 1000, 0.1) * config.speed
+      lastFrame = moving ? now : 0
+      // Invalidation and animation share a frame, including the initial resize observation.
+      if (invalidated || now - lastPaint >= 1000 / 24) {
```
khaledosman: `invalidated ||` short-circuits the 24 FPS cap, and every scroll event sets `geometryDirty`, so the throttle is off for the whole duration of a scroll.

Measured with this file's own harness: six rAF frames with no scrolling produce **2** `drawBars` calls; the same six frames each preceded by a `scroll` event produce **6**. Three times the paint work at the moment the main thread is already busiest, and it is the heaviest paint on the page.

The coalescing test at `LoginBackground.test.tsx:85` looks like it covers this but cannot: it builds `animationEnvironment(true)`, so reduced motion is on, `moving` is false, and there is no rAF chain for a scroll to ride. The bypass only exists in the animating path, which no test enters.

Letting the throttle win for the animating case is enough: `if ((invalidated && !moving) || now - lastPaint >= 1000 / 24)`.

web/src/features/auth/background/renderBars.ts
```diff
@@ -0,0 +1,86 @@
...
+  const barHeight = pitchY * (1 - config.spacing * (0.134 / 0.248))
+  const offsetX = (pitchX - barWidth) / 2
+  const offsetY = (pitchY - barHeight) / 2
+  const radius = barWidth * config.rounding
+
+  ctx.globalAlpha = 1
+  ctx.fillStyle = palette.background
+  if (visibleBottom <= visibleTop) return
+  ctx.clearRect(0, visibleTop, width, visibleBottom - visibleTop)
+  ctx.fillRect(0, visibleTop, width, visibleBottom - visibleTop)
```
khaledosman: Band-limited filling leaves the rest of the canvas permanently unpainted, and scrolling reveals it.

`main` is taller than the viewport whenever the card is tall (the login card with four footer links, the signup form). At `innerHeight: 800` with a 939px `main`, `visibleBottom` is 744, so canvas rows 744 to 939 are never filled. `canvas.width` is correctly guarded against reassignment at `LoginBackground.tsx:52`, which is the right call for other reasons but means nothing else ever clears those rows either. Scroll down and that strip composites as unpainted canvas — flat `--color-background` with no bars — until the rAF that follows the scroll event repaints. It chases the scroll by at least a frame.

Dropping the band and filling the whole canvas removes this, the throttle bypass above, and the scroll listener in one move, and leaves the cap from the other thread as the thing that actually bounds the work. If the band is kept instead, it needs a scroll margin of at least a viewport in each direction.

web/src/features/auth/background/renderBars.ts
```diff
@@ -0,0 +1,86 @@
...
+  const barWidth = pitchX * (1 - config.spacing)
+  const barHeight = pitchY * (1 - config.spacing * (0.134 / 0.248))
+  const offsetX = (pitchX - barWidth) / 2
+  const offsetY = (pitchY - barHeight) / 2
+  const radius = barWidth * config.rounding
+
+  ctx.globalAlpha = 1
+  ctx.fillStyle = palette.background
+  if (visibleBottom <= visibleTop) return
+  ctx.clearRect(0, visibleTop, width, visibleBottom - visibleTop)
```
khaledosman: `clearRect` here is dead work. `--color-background` is `#f7f8f8` light and `#08090a` dark, both fully opaque, and `globalAlpha` is 1 from line 46, so the `fillRect` on the next line overwrites every pixel this just cleared. It is a second full rasterization of the largest rectangle on the page, on every paint.

Drop the line. `renderBars.test.ts:43` pins it with `toHaveBeenCalledExactlyOnceWith`, so that assertion goes with it.

web/src/features/auth/background/LoginBackground.tsx
```diff
@@ -0,0 +1,137 @@
...
+    resize.observe(parent)
+    resize.observe(panel)
+    const theme = new MutationObserver(invalidatePalette)
+    theme.observe(document.documentElement, {
+      attributes: true,
+      attributeFilter: ["class", "data-theme"],
+    })
+    reduced.addEventListener("change", update)
+    document.addEventListener("visibilitychange", update)
+    document.addEventListener("scroll", invalidateGeometry, {
```
khaledosman: `passive: true` is right. `capture: true` on `document` is the part to reconsider: it fires for scrolls in every scrollable descendant, not just the page scroller, and each one schedules a measure and (per the throttle thread above) a paint.

If the band goes away, this listener goes away with it. If it stays, bind it to the actual scrolling element instead.

web/src/features/auth/LoginPageShell.tsx
```diff
@@ -0,0 +1,64 @@
...
+      {/* The viewport sets the top offset; disclosures and errors grow downward. */}
+      <main className="relative isolate flex flex-1 flex-col items-center px-4 pt-8 pb-8 md:pt-[max(3rem,calc(50vh-17.5rem))] md:pb-12">
+        <LoginBackground panelRef={panelRef} config={savedBackground} />
+        <div
+          ref={panelRef}
+          className="relative z-10 flex w-full max-w-md flex-col gap-6 border border-border-strong bg-surface p-6 md:p-8"
+        >
+          <span
+            aria-hidden
+            className="pointer-events-none absolute -top-px -left-px size-2 border-t-2 border-l-2 border-[var(--color-primary)]"
```
khaledosman: `border-accent` is the registered utility for this token, and is what `UpdatePrompt.tsx:49`, `AppShell.tsx:686` and `SettingsPage.tsx:619` use. Same on line 53.

Worth saying why the arbitrary value looked necessary: `--color-primary` is deliberately not registered in the `@theme` block, so `border-primary` genuinely does not exist. The accent reaches Tailwind through HeroUI's `--accent: var(--color-primary)` alias, under the name `accent`. `foundation.test.ts` does not catch this spelling, so it is rung-2 drift rather than a gate failure.

web/src/features/auth/background/LoginBackground.tsx
```diff
@@ -0,0 +1,137 @@
...
+    return () => {
+      cancelAnimationFrame(frame)
+      resize.disconnect()
+      theme.disconnect()
+      reduced.removeEventListener("change", update)
+      document.removeEventListener("visibilitychange", update)
+      document.removeEventListener("scroll", invalidateGeometry, true)
+      window.removeEventListener("resize", invalidateGeometry)
+    }
+  }, [panelRef, config])
```
khaledosman: `config` in the dependency array is stable only because `savedBackground` is a module-scope JSON import. The prop is public, so an inline object at any future call site tears down and rebuilds both observers and all four listeners on every render.

Either destructure the values the effect actually reads (`config.speed`, `config.columns`, and the rest) into the array, or hold `config` in a ref and keep the array at `[panelRef]`.

khaledosman (changes requested):
Re-reviewed at `f66f7331`. All ten earlier points are addressed: `login.css` is gone and the shell is utilities on semantic tokens, every public auth page and invitations share one shell (`AuthPageShell` no longer exists), `paused` and the settings ref are gone, the panel's top offset is content-independent, `min-h-svh`, a direct `palette.accent`, intrinsic `width`/`height` on the mark, and `text-mono-caption`. The design-doc change is the right resolution to the three I could not settle: scoped, linked to #991, and explicit that it does not extend to dashboard pages.

**Layout shift is clean**, since that was asked for specifically. The mark carries intrinsic dimensions matching its `1.638rem` box; `min-h-svh` uses the small viewport so retracting mobile chrome never reflows; the `md:pt-[max(3rem,calc(50vh-17.5rem))]` offset no longer depends on content height; the canvas is `absolute inset-0` and out of flow; the theme icons are same-size inside an `isIconOnly` button. The rAF handling is otherwise right: one chain, a `frame` guard against double-scheduling, timestamp throttling rather than `setTimeout`, cancel on unmount, `canvas.width` guarded against needless reassignment, `getComputedStyle` only on theme mutation, and layout events coalesced into a single `measure()`.

Seven code points below, three of them blocking (the missing cap, the throttle bypass, and the unpainted scroll band). Plus one that is not yours:

**CI has not run on any of these four commits.** `21c7a506` ran `build`, `catalog`, `dashboard`, `e2e` and `serving`. `f66f7331` has only `Lint PR title`, `check-template` and `close-stale`, and so do `a9848280`, `4c8cdaa8` and `d9485157`. Every `pull_request` workflow is missing, so the green ticks on this page cover the PR title and the template and nothing else. I ran the dashboard checks locally at `f66f7331` and they are fine: Biome clean over 558 files, `tsc -b` clean, 5,705 tests in 166 files passing, production build succeeding. Worth a re-trigger before merge so the record exists.

web/src/features/auth/background/renderBars.ts
jigjigjig: Addressed in 71e7f039. Preserved the studio dimensions and instead lowered the paint frequency with grid size: at most 24 FPS, further limited by a conservative 24,000-cell-per-second budget. The large-grid lifecycle test checks the lower frequency. Corrected the old reply about the removed density cap.

web/src/features/auth/background/LoginBackground.tsx
jigjigjig: Addressed in 71e7f039. Invalidation now bypasses the interval only for static rendering. The animating-path regression test sends scroll and resize events over six frames and asserts two paints.

web/src/features/auth/background/renderBars.ts
jigjigjig: Addressed in 71e7f039. The renderer now fills the entire canvas and draws all rows, skipping cells behind the form. A tall-canvas test verifies bars beyond the initial viewport. Large grids are budgeted through a lower paint frequency.

web/src/features/auth/background/renderBars.ts
jigjigjig: Addressed in 71e7f039. Removed clearRect. The opaque background fill is the only reset; the renderer test asserts that no clear occurs.

web/src/features/auth/background/LoginBackground.tsx
jigjigjig: Addressed in 71e7f039. Removed the scroll listener entirely because rendering no longer depends on the viewport band.

web/src/features/auth/LoginPageShell.tsx
jigjigjig: Addressed in 71e7f039. Both corner marks now use border-accent.

web/src/features/auth/background/LoginBackground.tsx
jigjigjig: Addressed in 71e7f039. The effect depends on scalar configuration values and constructs its settings inside the effect. A rerender test with a fresh equivalent config object verifies that the context is not recreated.

PR #1097, opened 2026-09-11
fix(dashboard): use Otari branding and page-specific tab titles

jigjigjig:
## Description
Use Otari as the website and installed app name. Browser tabs now show the current page first, such as “Usage · Otari” or “Providers · Otari”, so multiple open tabs are easier to distinguish. Sign-in and public authentication pages also receive page-specific titles.

## How to test it locally
- Run `pnpm --dir web run build` and open the dashboard.
- Confirm the sign-in heading says “Otari” and its tab says “Sign in · Otari”.
- Sign in and navigate from Overview to Providers or Usage. Confirm the tab follows the page using “Page · Otari”.
- Check the installed app manifest uses “Otari”.

Validation: architecture and migration checks plus Ruff (`make lint` with the existing environment), dashboard Biome lint, TypeScript build-mode typecheck, production build, dashboard Vitest suite (5,697 tests), and the gateway root-page/build serving tests (22 tests). Dashboard binaries were invoked directly because pnpm's version bootstrap stalled in this environment. No Python behavior changed.

## PR Type
- [x] Bug Fix

## Relevant issues
Product naming decision; no linked issue.

## Checklist
- [x] Documentation was updated where necessary (no documentation changes needed).

## AI Usage
- [ ] No AI was used.

### DIFF
--- a/tests/unit/test_gateway_root_page.py
+++ b/tests/unit/test_gateway_root_page.py
@@ -267,8 +267,8 @@ def test_hybrid_mode_serves_no_install_manifest(tmp_path: Path, monkeypatch: pyt
     """Only the standalone dashboard is an app worth installing.
 
     A hybrid gateway's root is a status page for a control plane that lives
-    elsewhere, so an installed icon named "Otari" would promise
-    management this deployment does not have. The index still links the manifest,
+    elsewhere, so an installed icon named "Otari" would promise management this
+    deployment does not have. The index still links the manifest,
     so what stops the install offer is this 404, which must not be cached: /pwa/
     carries a day of caching for the icons it usually serves, and a day-long 404
     would outlive a switch to standalone and keep the prompt away from a
--- a/tests/unit/test_gateway_root_page.py
+++ b/tests/unit/test_gateway_root_page.py
@@ -68,7 +68,7 @@ def test_dashboard_is_served_at_root(tmp_path: Path) -> None:
     assert response.status_code == 200
     assert "text/html" in response.headers["content-type"]
     assert '<div id="root">' in response.text
-    assert "Otari Dashboard" in response.text
+    assert "<title>Otari</title>" in response.text
 
 
 @requires_dashboard_bundle
@@ -105,7 +105,7 @@ def test_pwa_manifest_and_icons_are_served(tmp_path: Path) -> None:
     assert '<link rel="manifest" href="/pwa/manifest.webmanifest" />' in index
     assert '<link rel="apple-touch-icon" href="/pwa/apple-touch-icon.png" />' in index
 
-    assert payload["name"] == "Otari Dashboard"
+    assert payload["name"] == "Otari"
     assert payload["short_name"] == "Otari"
     assert payload["display"] == "standalone"
     # Android offers an install only with both a 192 and a 512 icon; the maskable
@@ -231,7 +231,7 @@ def _fake_bundle(tmp_path: Path) -> Path:
     bundle = tmp_path / "dashboard"
     (bundle / "assets").mkdir(parents=True)
     (bundle / "pwa").mkdir()
-    (bundle / "pwa" / "manifest.webmanifest").write_text('{"name": "Otari Dashboard"}')
+    (bundle / "pwa" / "manifest.webmanifest").write_text('{"name": "Otari"}')
     (bundle / "index.html").write_text('<html><body><div id="root"></div></body></html>')
     return bundle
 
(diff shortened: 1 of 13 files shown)


tests/unit/test_gateway_root_page.py
```diff
@@ -267,7 +267,7 @@ def test_hybrid_mode_serves_no_install_manifest(tmp_path: Path, monkeypatch: pyt
     """Only the standalone dashboard is an app worth installing.
 
     A hybrid gateway's root is a status page for a control plane that lives
-    elsewhere, so an installed icon named "Otari Dashboard" would promise
+    elsewhere, so an installed icon named "Otari" would promise
```
khaledosman: `src/gateway/main.py:718` still carries the other half of this mirrored sentence and still names a manifest `name` that no longer exists:

```python
# installed icon named "Otari Dashboard" would promise the wrong thing. The
```

Update it to "Otari" with this one. Also reflow this docstring; the shorter string leaves line 270 ragged against its neighbors.

web/src/app/AppShell.tsx
```diff
@@ -425,6 +426,14 @@ function AppShellChrome() {
   const isVisible = useNavVisibility()
   const recordNavigation = useRecordNavigation()
   const { pathname } = useLocation()
+  useDocumentTitle(
```
khaledosman: Nested ternary where this PR's own idiom is a table. `App.tsx` introduces `PUBLIC_AUTH_TITLES` for exactly this shape; mirror it here so the next chrome destination is one entry rather than another branch:

```ts
const CHROME_TITLES: Record<string, string> = {
  "/docs": "Documentation",
  "/account": "Account",
}
```

then `useDocumentTitle(navLabelForPath(pathname) ?? CHROME_TITLES[pathname])`.

web/src/app/App.tsx
```diff
@@ -93,7 +97,11 @@ function DeploymentRoot({ hash }: { hash: string }) {
   // is a link this deployment cannot honor, and the landing page's own
   // explanation is more useful here than a page that would just 404.
   if (deployment_type === "hybrid") {
-    return <HybridLanding />
+    return (
+      <PublicPageTitle>
```
khaledosman: No-op on this branch: with no `page` the hook writes `"Otari"`, which is what `web/index.html` already renders and nothing has overwritten before a hybrid gateway reaches here. Either name it (the page's own heading is "Otari gateway", so `page="Gateway"`) or drop the wrapper.

web/src/shared/hooks/useDocumentTitle.ts
```diff
@@ -0,0 +1,10 @@
+import { useEffect } from "react"
+
+export function useDocumentTitle(page?: string) {
+  useEffect(() => {
+    document.title = page ? `${page} · Otari` : "Otari"
+    return () => {
```
khaledosman, line 7: The cleanup is the non-obvious half and carries no explanation. Every unmount here is followed by another owner mounting and setting its own title, so the restore is never observable, yet `useDocumentTitle.test.ts:19` pins it as contract. Add a line saying why it exists, or drop it along with that assertion.

Same for `PublicPageTitle.tsx`: a component whose entire purpose is a side effect around a passthrough, in a tree where everything else states its rationale.

khaledosman (approved):
Four findings, none blocking. Checks are green and the affected suites pass locally (85 dashboard tests, 13 gateway tests); the dashboard job's console-warning count matches main.

Worth keeping: typing `PUBLIC_AUTH_TITLES` as `Record<PublicAuthPath, string>` makes a new public auth path a compile error, the same guard `OAuthCallbackPath` already puts on `PUBLIC_AUTH_PAGES`.

tests/unit/test_gateway_root_page.py
jigjigjig: Addressed in f66c5da8: updated the mirrored backend comment and reflowed the test docstring.

web/src/app/AppShell.tsx
jigjigjig: Addressed in f66c5da8: replaced the nested ternary with CHROME_TITLES, including undefined for unregistered paths.

web/src/app/App.tsx
jigjigjig: Addressed in f66c5da8: the hybrid landing page now sets Gateway · Otari, covered by the App test.

web/src/shared/hooks/useDocumentTitle.ts
jigjigjig: Addressed in f66c5da8: explained cleanup when an error boundary unmounts a page without another title owner, and documented the public wrapper.

PR #1099, opened 2026-09-11
feat(gateway): record time-to-first-token on the streaming path

### LINKED ISSUE #431: No time-to-first-token or throughput metrics, though first-chunk timing already drives failover

Related: #337 (provider-reported `provider_latency_ms`), #291.

## Summary

Otari records only total elapsed time. There is no time-to-first-token anywhere in `src/` (no `ttft`, `time_to_first`, or `first_token`), and no throughput measure. For interactive chat — the dominant Messages/Responses use case — TTFT is the number users actually feel, and total latency hides it completely.

Two requests can have identical `latency_ms` while one streams its first token in 200ms and the other stalls for 4 seconds and then dumps the response. Those are very different products, and otari currently reports them as the same.

## Distinction from #337

#337 captures *provider-reported* timing as a best-effort `provider_latency_ms` enrichment, keeping otari's self-timed `latency_ms` as canonical. This issue is about a different measure: TTFT is **otari-observable on the streaming path** — the gateway sees the first chunk arrive and does not need the provider to report anything. The two are complementary; neither subsumes the other.

## Why it's more than a nice-to-have

TTFT is already load-bearing inside otari's own routing logic and simply isn't recorded. `_platform.py:61` documents "streaming first-chunk timeouts (hybrid-mode fallback)" — a first-chunk timeout is already a failover trigger, and `_platform.py:72` describes a per-attempt budget that "abandons a slow" provider. So the gateway makes routing decisions on first-chunk timing while never persisting it, which means:

- an operator can't see how close to the timeout threshold a provider normally runs
- they can't tune the threshold from data
- they can't tell which provider is fast to first token versus fast overall

## Proposal

- Record `ttft_ms` (nullable, streaming path only) when the first content chunk is emitted.
- Derive output tokens/sec from `completion_tokens` and the streaming duration; optionally input tokens/sec.
- Surface TTFT alongside "Total time" in the Activity detail panel, and as a percentile (p50/p95) per model and per provider in the usage summary — a mean is the wrong statistic for latency.
- Feed it into the `by_provider` breakdown so provider comparison covers responsiveness, not just cost.

Prior art: Claude Code computes `ttft_ms`, `itps`, and `otps` internally for its own telemetry, so the shape of the measure is well established.

AmirF194:
## Description

The streaming path already makes a first-chunk-timeout routing decision on
time-to-first-token (the hybrid-mode fallback), but nothing recorded the value
that decision depends on. This adds `ttft_ms` to `UsageLog`, recorded on the
streaming path the same way `latency_ms` already is: `streaming_generator`
gains an `on_first_chunk` hook that fires once, on the first non-keepalive
chunk, and `build_streaming_response` uses it to compute and persist
`ttft_ms` on every settlement path (success, no-usage, error, disconnect).
Non-streaming requests and streams that fail before yielding anything record
no `ttft_ms`.

Scoped to backend capture and persistence, not the dashboard surface: #401
also asks for percentile stats and an Activity detail panel, which is a
separate change on top of this one.

Also fixes `test_the_revision_round_trips`: it pins a database to an older
migration and reads a row back through the `UsageLog` model, which now also
declares `ttft_ms` (added by a later revision than the one under test), so
the ORM's `SELECT` named a column that revision does not have yet. Switched
that read to raw SQL, matching the pattern the test already used one
revision earlier for the same reason.

## How to test it locally

`uv run pytest tests/unit/test_streaming_generator.py tests/integration/test_streaming_ttft.py -v`

- `test_on_first_chunk_fires_once_on_the_first_real_chunk` /
  `test_on_first_chunk_is_not_triggered_by_keepalives` pin the hook's
  contract in `streaming_generator`.
- `test_streaming_request_records_ttft_before_latency` sends a real request
  through the FastAPI app with a delayed first chunk and asserts the
  persisted `ttft_ms` reflects that delay and never exceeds `latency_ms`.
- `test_non_streaming_request_records_no_ttft` covers the null case.

`make lint`, `make typecheck`, and the full `make test` (unit + integration)
all pass locally, including the pre-existing suite unaffected by this change.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes #431

## Checklist

## AI Usage

**AI Model/Tool used:**

## Summary

- Added nullable `ttft_ms` recording for streaming requests.
- Recorded TTFT when the first valid response chunk arrives.
- Preserved TTFT for successful, incomplete, usage-free, failed, and disconnected streams.
- Kept TTFT unset for non-streaming requests and streams without a valid chunk.
- Added the database migration and tests.
- Updated migration tests for compatibility with later schema changes.

This gives operators a direct measure of streaming responsiveness without changing dashboard or activity views.


src/gateway/streaming.py
```diff
@@ -222,6 +228,10 @@ def _format(chunk: Any) -> str:
                 if chunk is _KEEPALIVE_DUE:
                     yield fmt.keepalive
                     continue
+                if not seen_first_chunk:
+                    seen_first_chunk = True
+                    if on_first_chunk is not None:
+                        on_first_chunk()
```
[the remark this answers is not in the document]
AmirF194: Confirmed and fixed: moved the mark into `_format_and_mark_first`, called at every yield site including the terminal-buffer flush, so it only fires once a chunk is actually formatted and about to be sent, not when it is merely pulled off the upstream. Added two unit tests pinning both cases you called out (extract_usage raising on the first chunk, and a buffered carrier), both fail against the prior code. Pushed as 28c6df89.

src/gateway/api/routes/_pipeline.py
```diff
@@ -3550,6 +3568,11 @@ def build_streaming_response(
       reservation does not leak.
     """
     platform_active = platform_correlation_id is not None
+    first_chunk_at: float | None = None
+
+    def _on_first_chunk() -> None:
+        nonlocal first_chunk_at
+        first_chunk_at = time.monotonic()
```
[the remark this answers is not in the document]
AmirF194: You're right, `_report_platform_usage` has no ttft_ms field so it's discarded for every hybrid stream. This diff only wired the metric through the standalone `log_usage` path. Extending the platform contract touches a shared interface with several other call sites (success, no-usage, and error all report through it), so I'd rather do that as its own PR than fold it into this one.

src/gateway/models/entities.py
```diff
@@ -811,6 +811,11 @@ class UsageLog(Base):
     # provider-never-reached rejections) have no meaningful request duration.
     latency_ms: Mapped[int | None] = mapped_column()
 
+    # Milliseconds from request start to the first streamed chunk. Nullable:
+    # non-streaming requests have no first chunk, historical rows predate the
+    # column, and a stream that failed before yielding anything never reached one.
+    ttft_ms: Mapped[int | None] = mapped_column()
```
khaledosman, line 823: The comment undersells what the number contains. `started_at` is taken in the handler preamble, so on a routing plan the serving row's `ttft_ms` also carries every earlier attempt's setup time, and on the hybrid path each abandoned attempt's first-chunk timeout on top of that. #401's percentile follow-up will key this by model and attribute failover time to the model that actually served, with nothing in the column saying so. Add the clause here, the way `status` and `status_code` above already document their traps.

tests/integration/test_streaming_ttft.py
```diff
@@ -0,0 +1,125 @@
...
+        )
+        assert response.status_code == 200, response.text
+        response.read()
+
+    log = db_session.query(UsageLog).filter(UsageLog.user_id == "ttft-user").first()
+    assert log is not None
+    assert log.ttft_ms is not None
+    assert log.ttft_ms >= round(_FIRST_CHUNK_DELAY_SECONDS * 1000)
+    assert log.latency_ms is not None
+    assert log.ttft_ms <= log.latency_ms
```
khaledosman, line 84: Only `_on_complete` is covered end to end. `_on_error`, `_on_no_usage` and `_on_incomplete` each gained a `_ttft_ms(...)` argument with no test behind it, and the error path is where the two clocks actually diverge (`ttft_ms` is fixed at the first chunk, `latency_ms` is measured at settlement), which is the reason `_ttft_ms` exists instead of reusing `_elapsed_ms`. Add a case: the stream yields one chunk and then raises, and the row comes back `status == "error"` with `ttft_ms` set and no greater than `latency_ms`.

tests/unit/test_pricing_provenance_schema_chain.py
```diff
@@ -181,8 +181,13 @@ def test_the_revision_round_trips(sqlite_at_head: tuple[Config, Engine]) -> None
...
-        row = session.get(UsageLog, "settled-1")
-    assert row is not None
+    with engine.connect() as connection:
+        # Raw SQL again: a revision after this one adds another usage_logs
+        # column (ttft_ms), so the mapped class now tracks columns this
+        # revision does not have yet and session.get(UsageLog, ...) would
+        # select one that is not there.
+        row = connection.execute(
+            text(f"SELECT {', '.join(_EXPECTED_TYPES)} FROM usage_logs WHERE id = 'settled-1'")
+        ).mappings().one()
```
khaledosman: `ruff format` rewrites this chain (it wants the `connection.execute(...)` call parenthesized with `.mappings()` and `.one()` on their own lines). The file was format-clean before this PR and is not now; `make lint` and pre-commit both run only `ruff check`, so nothing else will catch it. Run `uv run ruff format` on the file.

khaledosman (approved):
No blockers. Verified locally at `28c6df89`: `mypy` clean, `ruff check` clean, a single alembic head (`23fbd409eeac` -> `f1c4a8e2d6b9`), and the migration-chain unit tests, `tests/unit/test_streaming_generator.py` and the new `tests/integration/test_streaming_ttft.py` all pass. No route or schema change, so no OpenAPI/Postman regeneration is owed. Three non-blocking notes inline.

AmirF194:
All three addressed in 1bbf124: the ttft_ms comment now documents the routing-plan and hybrid-attempt cases, added the _on_error and _on_no_usage integration tests (status == "error", ttft_ms <= latency_ms), and ran ruff format on the schema-chain test. CI is green.

alembic/versions/23fbd409eeac_add_usage_logs_ttft_ms.py
```diff
@@ -0,0 +1,31 @@
...
+"""
+
+from collections.abc import Sequence
+
+import sqlalchemy as sa
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision: str = "23fbd409eeac"
+down_revision: str | Sequence[str] | None = "f1c4a8e2d6b9"
```
khaledosman, line 16: `f1c4a8e2d6b9` was head when this branch was cut, but main has since chained `a1d4f7c2e8b3` (playground tables) and then `c4e7a9b2d6f8` (api key suffix) onto it. Merged as-is the graph has two heads, `alembic upgrade head` cannot resolve a target, and `make lint` (via `check-migrations`), `tests/unit/test_alembic_heads.py` and `tests/unit/test_migration_single_head.py` all fail.

Verified by merging `origin/main` into this head and running `scripts/check_alembic_heads.py`:

```
2 Alembic heads. `alembic upgrade head` cannot resolve a target:
  23fbd409eeac  down_revision=f1c4a8e2d6b9
  c4e7a9b2d6f8  down_revision=a1d4f7c2e8b3
```

The green checks on this PR are not evidence against this: they ran 2026-09-11, and both of those migrations landed on main 2026-09-14.

Rebase and re-point `down_revision` at `c4e7a9b2d6f8`, then re-run the migration round trip, since the ordering changed.

src/gateway/models/entities.py
```diff
@@ -811,6 +811,17 @@ class UsageLog(Base):
     # provider-never-reached rejections) have no meaningful request duration.
     latency_ms: Mapped[int | None] = mapped_column()
 
+    # Milliseconds from request start to the first streamed chunk. Nullable:
+    # non-streaming requests have no first chunk, historical rows predate the
+    # column, and a stream that failed before yielding anything never reached one.
+    #
+    # ``started_at`` is taken in the handler preamble, so on a routing plan the
+    # serving row's value also carries every earlier attempt's setup time, and on
+    # the hybrid path each abandoned attempt's first-chunk timeout on top of that.
```
khaledosman, line 820: This clause describes a row that is never written. In hybrid mode every settlement callback in `build_streaming_response` returns before reaching `log_usage` (`if platform_active: ... return`, then `if db is None or log_writer is None: return`), and `run_streaming_with_fallback` passes `db=None` explicitly. `iterate_streaming_attempts`, the only caller of the first-chunk timeout, is on that same hybrid path, so no `ttft_ms` value ever carries an abandoned attempt's timeout.

The routing-plan sentence above it is right (standalone `walk_attempts` absorbs failed attempts, and their setup time is inside `started_at`). Drop the hybrid clause, or replace it with the fact that actually matters to a reader: hybrid streams record no `ttft_ms` at all, which is the gap CodeRabbit raised and you deferred.

src/gateway/api/routes/_pipeline.py
```diff
@@ -3731,6 +3758,7 @@ async def _on_incomplete() -> None:
                 user_id=user_id,
                 error="client disconnected before the stream completed",
                 latency_ms=_elapsed_ms(started_at),
+                ttft_ms=_ttft_ms(started_at, first_chunk_at),
```
khaledosman, line 3761: `_on_incomplete` is the one settlement path that gained a `ttft_ms` argument with nothing behind it. The latest commit covered `_on_complete`, `_on_no_usage` and `_on_error`; the disconnect path is still uncovered, and it is the one where the row is written only under a non-empty `tool_tally`, so it is the easiest to break silently. `tests/unit/test_pipeline_settlement.py::test_client_disconnect_refunds` already has the harness.

src/gateway/api/routes/_pipeline.py
```diff
@@ -3773,6 +3801,7 @@ def _attach_inline_cost(value: Any, settlement: SettledCost) -> bool:
             settle_before_done=platform_active,
             is_cost_carrier=adapter.is_stream_cost_carrier if platform_active else None,
             attach_settlement=_attach_inline_cost if platform_active else None,
+            on_first_chunk=_on_first_chunk,
```
khaledosman, line 3804: Add `on_first_chunk` to the callback list in `tests/unit/test_pipeline_settlement.py:173`. That test exists to pin exactly this: every streaming response, for every format and on both the single-attempt and platform-fallback paths, wires the full callback set. `on_first_chunk` is part of that contract now, and nothing catches it being dropped in a later refactor.

khaledosman (changes requested):
One blocker (the migration no longer chains onto main's head), one inaccurate comment, and two coverage gaps. The rest checks out: no route or schema change, so the OpenAPI and Postman artifacts are correctly untouched; the migration is nullable with a real downgrade and is dialect-neutral; `ruff format` does flag `_pipeline.py`, `entities.py` and `streaming.py`, but all three are already unformatted at the base commit, so that is pre-existing and not this PR's to fix.

AmirF194:
Rebased onto main and repointed down_revision at c4e7a9b2d6f8. check_alembic_heads.py reports a single head now, and the round trip and check-migrations tests pass.

Thanks for tracing the hybrid path all the way through on the entities.py comment, that was a real error on my part. I walked all four settlement callbacks again: on_complete, on_no_usage and on_error return through the platform_active branch before log_usage, and on_incomplete returns on `db is None` since run_streaming_with_fallback passes db=None explicitly. So no hybrid stream ever writes a ttft_ms row, full stop. Replaced the abandoned-attempt claim with that.

Added a unit test for the on_incomplete path with a non-empty tool_tally in test_pipeline_settlement.py, using test_client_disconnect_refunds's own fixtures, asserting ttft_ms gets set and stays no greater than latency_ms. Reverting the ttft_ms argument locally makes it fail, so it's exercising the real thing.

Also added on_first_chunk to the wired-callbacks list in test_all_settlement_callbacks_wired_for_every_format_and_path. Same check: pulling the wiring out of build_streaming_response fails it across all six format/path combinations.

Ran ruff check, mypy strict on the whole project (598 files, clean), the alembic head and migration tests, the full test_pipeline_settlement.py file (70 passed) and test_pricing_provenance_schema_chain.py (5 passed), all in a container off this branch. The broader tests/unit run was still going when I pushed, so I can't say the whole directory is green yet, just that nothing in this diff touches code outside those files. Left the ruff format flags in _pipeline.py, entities.py and streaming.py alone since they're on lines outside this diff, confirmed against the same file on main.

tests/unit/test_pipeline_settlement.py
```diff
@@ -495,6 +500,41 @@ async def stream() -> AsyncIterator[ChatCompletionChunk]:
     assert settlement.reconciled == []
 
 
+@pytest.mark.asyncio
+async def test_client_disconnect_with_tool_work_records_ttft(monkeypatch: pytest.MonkeyPatch) -> None:
+    """_on_incomplete only reaches log_usage when there is tool work to bill
+    for (a non-empty tool_tally). That is also the one settlement path this
+    PR's own coverage missed: it should still record ttft_ms from the same
```
khaledosman, line 507: "That is also the one settlement path this PR's own coverage missed" records review history, which `AGENTS.md` puts in the commit message rather than the code. In six months the reader has no PR in view and the sentence just reads as noise.

The first sentence is the part that earns its place: `_on_incomplete` only reaches `log_usage` when there is tool work to bill for, which is exactly why this test has to set a non-empty `tool_tally`. Keep that and the closing clause, drop the middle.

khaledosman (approved):
All four findings from the previous review are addressed, and each verified rather than taken on trust:

- `scripts/check_alembic_heads.py` on the rebased branch reports `Single head: 23fbd409eeac`.
- The `UsageLog.ttft_ms` comment now states the hybrid fact correctly.
- `test_client_disconnect_with_tool_work_records_ttft` exercises `_on_incomplete` through the real `build_streaming_response` on the branch that actually writes a row (non-empty `tool_tally`).
- `on_first_chunk` is in the wired-callback parity list.

Ran `test_pipeline_settlement.py`, `test_streaming_generator.py`, `test_pricing_provenance_schema_chain.py`, `test_alembic_heads.py` and `test_migration_single_head.py` locally on this head: 99 passed.

One non-blocking nit inline.

AmirF194:
Appreciate the re-verification against a fresh clone rather than trusting the diff, that's exactly the coverage this migration change needed. Noted on the test comment nit for next time.

Issue #1094, opened 2026-09-11
Redesign the /welcome data-plane landing page


jigjigjig:
## Context

Asked in Slack whether `/welcome` should stay or be trimmed. Nathan confirmed it stays and what it is for:

> It is for gateways running in "hybrid mode" so they don't have a dashboard or control plane. It's the landing page to say "this is just the data plane, no control plane on this host".

So the page keeps its role. What it needs is a design pass: today it reads as vibe-coded rather than as an intentional surface, and it is the only Otari page a self-hosted hybrid operator ever sees in a browser.

## Where it lives

- Route: `src/gateway/main.py` (`GET /welcome`, `root_tutorial`), registered in both modes.
- Markup and styles: `src/gateway/templates/root_tutorial.html` (a single hand-written 500+ line file with an inline `<style>` block, not part of the `web/` dashboard build).
- Loader: `src/gateway/root_page.py`, which injects `QUICKSTART_URL` at import time.
- Test: `tests/unit/test_gateway_root_page.py`.

Because it sits outside `web/`, it does not pick up the dashboard's design tokens or components: its palette is a hand-copied snapshot of them, which is exactly the drift this should close.

## Scope

1. Make the page say what Nathan said it says. The first thing a reader should understand is "this host is a data plane; the control plane is at otari.ai", with the link to get there. The quickstart material is secondary to that.
2. Redesign the surface against the current dashboard direction (the accent, square corners, divided-surface treatment) so it does not look like a page from a different product than the one it links to.
3. Check it at phone width; the page has never had a responsiveness pass.
4. Keep it self-contained (no bundle dependency, no network fetches): a hybrid gateway must be able to serve it with the dashboard bundle absent.

## Out of scope

Whether `/welcome` exists at all, and any change to what a standalone deployment serves at its root.

jigjigjig:
fixed

PR #1104, opened 2026-09-11
feat(welcome): redesign the welcome page in the dashboard's divided-surface language

### LINKED ISSUE #1122: Redesign the /welcome page in the dashboard's divided-surface language
## Context

The `/welcome` page is rendered by the gateway itself, not by the React dashboard: it is `src/gateway/templates/root_tutorial.html`, served at `/welcome` and also at `/` when no dashboard bundle has been built. It still carries its own visual language, with teal accents, rounded corners and drop shadows, from before the dashboard adopted the divided-surface direction. Anyone who lands on it after setting up a gateway sees a page that does not look like the product it introduces.

The design is in Paper: https://app.paper.design/file/01M28ESMV1XY5CXH5S8Z7W3DPV/1-0. Artboard R1 is the page at 1440, R2 is the copy control in its pressed state, and R3 is the page at 390.

## What to do

Rewrite `src/gateway/templates/root_tutorial.html` to match the design:

- A flat `#F7F8F8` ground, with a white top bar separated by a hairline.
- No cards, no border radii and no shadows anywhere.
- Step rows divided by hairlines, each with a monospaced `01` / `02` / `03` lane.
- Code blocks on `#20282D`, each with a monospaced language label and an icon copy control.
- Square 36px buttons.
- Mozilla Headline for the title, Mozilla Text for body copy, and Fira Code for labels and code, loaded with `@font-face` from `/fonts/` and each given a system fallback in the same declaration. The `/fonts/` mount only exists when the dashboard bundle is present, so the fallbacks have to carry an installation that has not built it.

Two constraints on the page itself:

- Keep it dependency-free, with no build step. It is a single HTML file with inline CSS and a few lines of vanilla JavaScript for the copy buttons, and that is what lets it be served from a source checkout.
- Keep the strings `tests/unit/test_gateway_root_page.py` asserts on, keep a real `<h1>`, and keep `QUICKSTART_URL` as the placeholder `root_page.py` substitutes.

## Notes

Related: #1094 covers the broader data-plane versus control-plane messaging of this page; this issue is the visual redesign only. PR #1104 implements it.

jigjigjig:
## Description
Redesign `/welcome` around the linked [Paper desktop and mobile designs](https://app.paper.design/file/01M28ESMV1XY5CXH5S8Z7W3DPV/1-0): a compact header, flat surfaces, numbered setup steps, and dark code panels. On phones, actions stack and code scrolls within the page width.

Copy controls show a checkmark and “Copied,” announce completion to screen readers, and explain how to copy manually when clipboard access is unavailable or denied. The tutorial remains available without a dashboard build and uses system font fallbacks when the design fonts are not installed.

## How to test it locally
1. Start the gateway with `uv run otari serve --config config.yml` and open `/welcome`.
2. Check the page at desktop and phone widths, scroll the Python example horizontally, and use Tab/Enter on each copy button.
3. Run the browser coverage with `pnpm --dir web exec playwright test e2e/welcome.spec.ts --project=parity` against the normal built E2E setup.

Validated locally:
- `uv run pytest tests/unit/test_gateway_root_page.py -q`: 13 passed, 4 bundle-dependent tests skipped.
- Four Playwright tests passed against a real local gateway, using a temporary configuration without unrelated seed projects.
- Dashboard Biome lint passed.
- `uv run --frozen --no-dev python scripts/oss_edition_smoke.py` passed.
- Desktop (1440px), mobile (390px), and copied-state screenshots reviewed.

The full integration suite was not run. No API contract or generated artifacts changed.

## PR Type
- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues
Fixes #1122

Refs #1094. This PR implements the Paper visual redesign and responsive copy controls. The issue remains open for its broader data-plane/control-plane messaging scope.

## Checklist

## AI Usage
- [ ] No AI was used.

**AI Model/Tool used:** GPT-6 / Codex.

**Any additional AI details you'd like to share:** Implemented from the user-provided Paper design; browser tests cover clipboard success, failure, keyboard access, and mobile overflow.

## Summary

- Redesigned `/welcome` to match the Paper desktop and mobile designs.
- Added a flat layout, compact header, numbered steps, dark code panels, and responsive mobile behavior.
- Improved copy controls with success feedback, screen reader announcements, keyboard access, and manual-copy guidance when clipboard access fails.
- Added browser coverage for copy behavior, keyboard access, touch targets, and mobile code scrolling.

## Technical notes

- The tutorial remains dependency-free and works without a dashboard build.
- Added dashboard font loading with system fallbacks and accessible focus states.
- Preserved `QUICKSTART_URL` and the required page heading.
- No API contracts or generated artifacts changed.

### DIFF
--- a/src/gateway/templates/root_tutorial.html
+++ b/src/gateway/templates/root_tutorial.html
@@ -7,25 +7,26 @@
     <link rel="icon" type="image/svg+xml" href="/favicon.svg" />
     <style>
       :root {
+        color-scheme: light;
+        /* Shared with the dashboard's light semantic palette. */
         --brand: #0098a4;
-        /* Both derived from the accent above rather than from the blue-grey it
-           replaced, and both taken from the dashboard's own light palette so
-           this page and the app cannot drift: `--brand-dark` is
-           `--color-link`, the accent's darker step, and `--brand-tint` is
-           `--color-primary-subtle` composited on white, since this page uses
-           opaque fills. Measured: the ink reads 5.53 on white and 4.70 on the
-           tint, both clearing 4.5. */
         --brand-dark: #00747e;
-        --brand-tint: #dbf1f2;
-        --ink: #14242c;
-        --muted: #5b6b73;
-        --line: #dbe5e8;
+        --primary-button: #00808a;
+        --ink: #08090a;
+        --text: #2c2e33;
+        --muted: #5a5f6b;
+        --line: rgb(0 0 0 / 6%);
         --surface: #ffffff;
-        --code-bg: #15242b;
-        --code-fg: #e7eef0;
-        --code-muted: #8fb0ba;
-        --radius: 14px;
-        --shadow: 0 1px 2px rgba(20, 36, 44, 0.04), 0 12px 32px -12px rgba(20, 36, 44, 0.18);
+        --surface-muted: #f0f1f2;
+        --background: #f7f8f8;
+        --success: #21783d;
+        --code-bg: #20282d;
+        --code-line: #354148;
+        --code-control: #2c363c;
+        --code-fg: #e8edf0;
+        --code-muted: #a1a8b2;
+        --code-success: #4fbf95;
+        --mono: "Fira Code", ui-monospace, SFMono-Regular, Menlo, Consolas, monospace;
       }
       * {
         box-sizing: border-box;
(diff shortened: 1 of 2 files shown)


src/gateway/templates/root_tutorial.html
```diff
@@ -7,397 +7,386 @@
...
-      .brand svg {
-        height: 30px;
-        width: auto;
-        display: block;
+      .brand img {
+        width: 1.375rem;
+        height: 1.375rem;
+      }
+      .brand, h1 {
+        font-family: "Mozilla Headline", system-ui, sans-serif;
```
jigjigjig, line 91: `Mozilla Headline` and `Fira Code` are named here but nothing loads them, so the h1 and code render in system fonts for every real deployment (Docker ships the bundle, and its fonts sit at `/fonts/`, but this page never asks for them). The Paper design is set in both. Fix: add two `@font-face` rules pointing at the bundle's woff2 files with `font-display: swap`; without a bundle the request 404s and the existing system fallback stays. Visual change, so this goes to Fede as a question before it is applied.

src/gateway/templates/root_tutorial.html
```diff
@@ -7,397 +7,386 @@
...
         -webkit-font-smoothing: antialiased;
       }
       a {
         color: var(--brand-dark);
       }
-      header.nav {
-        max-width: 880px;
-        margin: 0 auto;
-        padding: 22px 24px;
+      a:focus-visible, button:focus-visible, pre:focus-visible {
```
jigjigjig, line 61: The `.copy` button is 32px tall inside a 36px `figcaption`, and this ring is 2px at a 3px offset: 42px, so on keyboard focus the ring crosses the caption's bottom border into the code area (visible in a focus screenshot at 1440). Fix: `.copy:focus-visible { outline-offset: -2px }` so the ring sits inside the control. Visual change, question for Fede.

src/gateway/templates/root_tutorial.html
```diff
@@ -406,77 +395,86 @@ <h1>Your gateway is running.</h1>
...
               <span class="cap-label">bash</span>
-              <button class="copy" type="button">Copy</button>
+              <button class="copy" type="button" aria-label="Copy key export command">
+                <svg class="copy-icon" aria-hidden="true" viewBox="0 0 24 24" fill="none" stroke="currentColor" stroke-width="2" stroke-linecap="round" stroke-linejoin="round"><rect x="9" y="9" width="13" height="13"/><path d="M5 15H4a2 2 0 0 1-2-2V4a2 2 0 0 1 2-2h9a2 2 0 0 1 2 2v1"/></svg>
+                <svg class="check-icon" aria-hidden="true" viewBox="0 0 24 24" fill="none" stroke="currentColor" stroke-width="2" stroke-linecap="round" stroke-linejoin="round"><path d="m20 6-11 11-5-5"/></svg>
+                <span class="copy-label" aria-hidden="true"></span>
+              </button>
             </figcaption>
-            <pre><code>export OTARI_API_KEY=YOUR_BOOTSTRAP_OTARI_KEY</code></pre>
+            <pre tabindex="0" aria-label="Key export command"><code>export OTARI_API_KEY=YOUR_BOOTSTRAP_OTARI_KEY</code></pre>
```
jigjigjig: `aria-label` on a `<pre>` is a name on a generic-role element, which ARIA 1.2 prohibits and screen readers ignore, so the focusable code block announces as nothing. Add `role="region"` to the three `<pre tabindex="0">` blocks (here, line 456, line 477); `getByLabel` in the spec keeps working. No pixel changes.

src/gateway/templates/root_tutorial.html
```diff
@@ -497,25 +495,46 @@ <h3>Send a chat completion</h3>
...
     </footer>
+    <span id="copy-status" class="sr-only" role="status"></span>
 
     <script>
       document.querySelectorAll(".copy").forEach(function (btn) {
-        btn.addEventListener("click", function () {
-          var code = btn.closest(".code").querySelector("code").innerText;
-          navigator.clipboard.writeText(code).then(function () {
-            btn.textContent = "Copied";
+        var resetTimer;
```
jigjigjig: `var` next to `async`/`await`, `?.` and a bare `catch {` is two eras in one script; the old file used `var` because it had nothing newer. Switch to `const`/`let` throughout the handler.

src/gateway/templates/root_tutorial.html
```diff
@@ -7,397 +7,386 @@
     <link rel="icon" type="image/svg+xml" href="/favicon.svg" />
     <style>
       :root {
+        color-scheme: light;
+        /* Shared with the dashboard's light semantic palette. */
```
jigjigjig: Thirteen of these fifteen values are in `web/src/styles/globals.css`; `--code-muted` (#a1a8b2) and `--code-success` (#4fbf95) are not, so the comment overstates the sharing. Say which two are this page's own, or add them to the dashboard tokens if the dashboard's code panel wants them too.

jigjigjig (commented):
Self-review (design engineer pass: repo review, unslop-code, unslop-design, ui-review). Render matches the Paper R1/R2/R3 artboards at 1440 and 390; CI, CodeRabbit clean. Five threads below, two of them visual and going to Fede as questions.

src/gateway/templates/root_tutorial.html
jigjigjig: Addressed in c75fb63f: two @font-face rules load Mozilla Headline and Fira Code from the bundle at /fonts/ with font-display swap; verified loaded with the bundle and falling back to system fonts on a 404. Approved by Fede.

src/gateway/templates/root_tutorial.html
jigjigjig: Addressed in c75fb63f: .copy:focus-visible gets outline-offset -2px; the ring now sits inside the 32px control (measured 2px inside each edge of the caption). Approved by Fede.

src/gateway/templates/root_tutorial.html
jigjigjig: Addressed in c75fb63f: role="region" on the three pre blocks; getByLabel in the spec still resolves each one.

src/gateway/templates/root_tutorial.html
jigjigjig: Addressed in c75fb63f: const/let throughout the handler.

src/gateway/templates/root_tutorial.html
jigjigjig: Addressed in c75fb63f: the comment now names --code-muted and --code-success as this page's own.

web/e2e/welcome.spec.ts
```diff
@@ -0,0 +1,83 @@
...
+    ).not.toHaveText("Copied")
+  })
+}
+
+test("welcome fits a phone while keeping long code scrollable", async ({
+  page,
+}) => {
+  await page.setViewportSize({ width: 390, height: 844 })
+  await page.goto("/welcome")
+  expect(await page.evaluate(() => document.documentElement.scrollWidth)).toBe(
```
khaledosman: Asserting the literal viewport width makes this depend on whether Chromium paints a classic scrollbar. I measured it here: on macOS (overlay scrollbars) `document.documentElement.scrollWidth` is 390 and this passes, but with a classic scrollbar both `clientWidth` and `scrollWidth` are 375 and it fails on layout that is perfectly fine. The page is tall enough at 390x844 to always take a vertical scrollbar, so there is no case where it does not matter.

Compare the two instead, which is also what `dashboard.spec.ts:437-439` does:

```ts
expect(
  await page.evaluate(() => ({
    scroll: document.documentElement.scrollWidth,
    client: document.documentElement.clientWidth,
  })),
).toMatchObject({ scroll: expect.any(Number) })
```

or more directly, `expect(scrollWidth).toBeLessThanOrEqual(clientWidth)`.

web/e2e/welcome.spec.ts
```diff
@@ -0,0 +1,83 @@
...
+    const button = page.getByRole("button", { name: `Copy ${name}` })
+    const snippet = await page
+      .getByLabel(name, { exact: false })
+      .filter({
+        has: page.locator("code"),
+      })
+      .textContent()
+    await button.focus()
+    await page.keyboard.press("Enter")
+    await expect(button).toHaveText("Copied")
```
khaledosman: This polls a label that clears itself after 1500ms (`root_tutorial.html`'s `resetTimer`). If the first poll lands after the timer has fired, `toHaveText` keeps retrying against a value that will never come back and the test fails; the `parity` project runs at `retries: 0`, so that is a red run rather than a retry.

The live region is the stable signal, since it is only cleared on the next click. Assert it first and let the label check that follows it stand as the reset check:

```ts
await expect(page.getByRole("status")).toHaveText(`Copied ${name}.`)
await expect(button).toHaveText("Copied")
```

src/gateway/templates/root_tutorial.html
```diff
@@ -7,397 +7,404 @@
...
-        <a href="/docs">API docs</a>
-        <a href="https://otari.ai">otari.ai</a>
-        <a class="gh" href="https://github.com/mozilla-ai/otari">GitHub</a>
-      </nav>
+    <header class="nav-bar">
+      <div class="nav">
+        <span class="brand"><img src="/favicon.svg" alt="" />Otari</span>
+        <nav class="nav-links" aria-label="Main navigation">
+          <a href="/docs">API docs</a>
+          <a href="https://otari.ai">otari.ai ↗</a>
```
khaledosman: The arrow is part of the link text, so it is announced: "otari.ai north east arrow", and the same on lines 400, 519 and 521. Wrap it so it stays decorative:

```html
<a href="https://otari.ai">otari.ai <span aria-hidden="true">↗</span></a>
```

khaledosman (commented):
Reviewed at `c75fb63`. Two findings I had drafted against `35a23b6` (the brand faces named but never `@font-face`d, and `aria-label` on a role-less `<pre>`) are fixed by that commit, so what is left is the e2e suite plus one markup nit.

One note on the fonts fix: the `Mozilla Headline` face is declared without `font-stretch: 75% 125%`, which `web/src/styles/globals.css` does carry for the same file. Nothing on this page sets `font-stretch`, so it renders identically today; it only matters if someone later reaches for a width.

---
Reviewed by Claude Opus 5 via [Claude Code](https://claude.com/claude-code).

web/e2e/welcome.spec.ts
jigjigjig: Addressed in 2550c6d3: the spec now compares scrollWidth with clientWidth (390/390 here, and it holds with a classic scrollbar too).

web/e2e/welcome.spec.ts
jigjigjig: Addressed in 2550c6d3: the live region is asserted first, the label second; the status text is present within a few milliseconds of the click so the label check always runs inside the 1500ms window.

src/gateway/templates/root_tutorial.html
jigjigjig: Addressed in 2550c6d3: all four arrows sit in aria-hidden spans; the links announce as otari.ai, GitHub, Quickstart. Also took the font-stretch note from your summary: the Mozilla Headline face now declares 75% 125% like globals.css.

src/gateway/templates/root_tutorial.html
```diff
@@ -7,397 +7,405 @@
...
+          padding: 1rem 1rem 1.5rem;
+        }
+      }
+      @media (pointer: coarse) {
+        .btn, .copy {
+          min-height: 2.75rem;
+        }
+        .copy {
+          min-width: 2.75rem;
+        }
```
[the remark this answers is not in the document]
jigjigjig: Addressed in c0aed944: use any-pointer: coarse so hybrid touch devices retain the 44px copy targets.

web/e2e/welcome.spec.ts
```diff
@@ -0,0 +1,85 @@
...
+    scrollWidth: document.documentElement.scrollWidth,
+    clientWidth: document.documentElement.clientWidth,
+  }))
+  expect(scrollWidth).toBeLessThanOrEqual(clientWidth)
+  const snippet = page.getByLabel("Chat completion example", { exact: true })
+  await snippet.focus()
+  await page.keyboard.press("End")
+  expect(
+    await snippet.evaluate((node) => node.scrollWidth > node.clientWidth),
+  ).toBe(true)
```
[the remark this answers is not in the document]
jigjigjig: Addressed in c0aed944: assert computed overflowX is auto or scroll alongside excess content width, and remove the ineffective End key press. All four welcome browser tests pass against the local template server.

src/gateway/templates/root_tutorial.html
```diff
@@ -7,397 +7,405 @@
...
         margin: 0;
         color: var(--ink);
-        background:
-          radial-gradient(1100px 520px at 50% -8%, rgba(78, 130, 149, 0.16), transparent 60%),
-          linear-gradient(180deg, #f6f9fa 0%, #eef3f5 100%);
-        background-attachment: fixed;
-        font-family: ui-sans-serif, system-ui, -apple-system, "Segoe UI", Roboto, sans-serif;
-        line-height: 1.6;
+        background: var(--background);
+        font-family: system-ui, -apple-system, "Segoe UI", sans-serif;
```
khaledosman: The page loads two of the three brand faces and then renders every paragraph in `system-ui`, while the dashboard one click away renders the same copy in Mozilla Text (`--font-sans`, `web/src/styles/globals.css:699`). The bundle already serves `MozillaText-Variable.woff2` out of the same `/fonts/` directory the two `@font-face` rules above point at, so as written the headings land in Mozilla Headline and the text under them does not.

Add the third face and lead the stack with it:

```css
@font-face {
  font-family: "Mozilla Text";
  src: url("/fonts/MozillaText-Variable.woff2") format("woff2");
  font-weight: 200 700;
  font-display: swap;
}
```
```css
body { font-family: "Mozilla Text", system-ui, -apple-system, "Segoe UI", sans-serif; }
```

`.copy` repeats the bare `system-ui` stack at line 267 and needs the same edit. It is a visual change, so worth confirming against the Paper artboards before applying, the way the other two were.

src/gateway/templates/root_tutorial.html
```diff
@@ -497,25 +514,46 @@ <h3>Send a chat completion</h3>
...
-            btn.textContent = "Copied";
+        let resetTimer;
+        btn.addEventListener("click", async function () {
+          const figure = btn.closest(".code");
+          const status = document.getElementById("copy-status");
+          const label = btn.querySelector(".copy-label");
+          clearTimeout(resetTimer);
+          btn.classList.remove("ok");
+          label.textContent = "";
+          status.textContent = "";
```
khaledosman, line 544: This clear never reaches a screen reader on the failure path. On an insecure origin (`http://<host>:8000/welcome`, how most self-hosted gateways are reached) `navigator.clipboard` is undefined, so line 541 throws a TypeError synchronously and the `catch` runs in the same task as this line. The accessibility tree is diffed once the task ends, so it only ever sees the final value: the first click announces "Could not copy...", and every click after it announces nothing, because the live region already holds that exact string.

Put the set a task away from the clear:

```js
status.textContent = ""
setTimeout(function () { status.textContent = message }, 0)
```

The success path is unaffected (a real `writeText` settles in a later task), and so is the spec, which clicks once per case.

khaledosman (approved):
Reviewed at `c0aed944`. The token block checks out against `web/src/styles/globals.css` value for value, and the font-face, ARIA and reset-timer threads above are resolved at this head. Two non-blocking findings inline.

src/gateway/templates/root_tutorial.html
jigjigjig: Addressed in b0359d47: load Mozilla Text and use it for body and copy controls, matching the dashboard and the welcome Paper file font-sans token. All four welcome browser tests pass with the actual bundled fonts served.

src/gateway/templates/root_tutorial.html
jigjigjig: Addressed in b0359d47: defer the failure announcement to a separate task. Both clipboard failure tests now click twice and assert the live region clears before the repeated message is restored. All four welcome browser tests pass.

PR #1105, opened 2026-09-11
fix(dashboard): preserve add-provider dialog state across tab switches

### LINKED ISSUE #1081: Add-provider dialog: switching tabs discards the half-filled tab, including a pasted key
Switching tabs in the add-provider dialog discards the half-filled tab, with no
guard and nothing said. Pick OpenAI, paste a key, press "Custom endpoint": the
key is gone, and pressing "Known provider" again opens an empty form.

The behavior predates the dialog. Each tab was its own panel with its own state,
and is now its own component with its own state, so `AddProviderForm`'s
conditional render unmounts the active one on every switch;
`AddProviderForm`'s docstring states it. `FormDialog`'s `isDirty` guard gates
dismissal only, and the tab buttons call `setTab` directly, so nothing asks.

It matters because a pasted provider key is exactly what is lost, which puts it
in the same class as the dirty-check gaps fixed across #1031's series rather
than in the cosmetic pile.

Two candidate fixes, from the review thread on #1045
(https://github.com/mozilla-ai/otari/pull/1045#discussion_r0):

1. **Hoist both drafts into `AddProviderForm`** so each tab's state survives a
   switch. Honest to the operator's expectation, and the larger change: a dozen
   `useState` pairs, two mutations and two validity derivations move up, and the
   two halves stop being independently readable, which is the property their
   docstring argues for today.
2. **Guard the tab press.** `FormDialog` would need a way to gate an arbitrary
   action on its own dirty state, since the guard currently lives in the footer
   and answers dismissal alone. Smaller in the page and larger in the shared
   surface: six other dialogs render through that file.

Whichever is chosen, the check is the same: fill one tab, switch, switch back,
and assert the field still holds what was typed (option 1) or that the switch
asked first (option 2).

Not urgent, and deliberately not folded into #1045: its threads are otherwise
closed and its CI is green, and this is a decision that wants its own diff.

AloysJehwin:
## Description

Switching tabs in the add-provider dialog discarded whatever was half-filled, including a pasted API key: each tab's fields lived in the tab component, and only the active tab is mounted.

The fields now live one level up, in `AddProviderForm`, so they survive while the component that renders them comes and goes. Only the active tab still mounts.

Two things scoped to the draft's lifetime had to move up with it, or the fix would have traded one loss for two others:

- **The unsaved-changes guard.** `useDirtySnapshot` seeds on mount, so a snapshot taken inside a tab reseeded against already-filled values after a switch and read clean. Escape then closed the dialog with the pasted key and asked nothing. It is now taken once over both drafts.
- **The API base prefill.** With the provider lifted, its detail query answers from cache on the first render after a switch, so a mount-keyed effect refired and overwrote a hand-edited base with the provider's default. It is now keyed on the provider already seeded.

## How to test it locally

1. Providers → **Add provider**.
2. On **Known provider**, pick a provider and paste an API key. Open **Advanced** and replace the API base with something of your own, e.g. `https://proxy.internal/v1`.
3. Switch to **Custom endpoint**, then back.
4. The provider, the key and your API base are all still there, and pressing Escape now offers **Discard** rather than closing silently.

Automated: `pnpm --dir web exec vitest run src/features/providers/ProvidersPage.test.tsx` covers all three (the surviving draft, the armed guard, the preserved base) in one case. It fails before this change on the API base. 46 passing at this head; `tsc --noEmit` and Biome clean.

## PR Type

- [x] Bug Fix

## Relevant issues

Fixes #1081

## Checklist

Dashboard-only: no route or schema touched, so no docs or OpenAPI change owed.

## AI Usage

**AI Model/Tool used:**

Claude Code (Opus 5)

The review on this PR was acted on with Claude Code. Both findings were reproduced against the reviewed commit before being fixed, and the new test was checked against that commit to confirm it actually fails without the change rather than passing for an unrelated reason.

---

**Note on an earlier version of this description:** it said both panels render with `isOpen={false}` on the inactive one. That was never what the diff did, and @khaledosman was right to flag it. Only the active tab mounts (`{tab === "known" && …}`); what survives is the state, lifted above them.

## Summary

- Preserves partially completed provider forms when users switch tabs.
- Keeps API keys, API base edits, and other draft fields intact.
- Maintains the unsaved-changes prompt across both drafts.
- Prevents provider defaults from overwriting manual API base edits.
- Adds regression tests for draft persistence, discard protection, and API base reseeding.

### DIFF
--- a/web/src/features/providers/ProvidersPage.tsx
+++ b/web/src/features/providers/ProvidersPage.tsx
@@ -502,8 +502,9 @@ type ProviderTab = "known" | "custom"
  *
  * What is shared is the frame. Each half renders the same `FormDialog` with the
  * same title, size and tab row, so switching tabs changes the fields and
- * nothing else, and a half-filled tab still does not survive a switch away from
- * it, which is what it did as a panel.
+ * nothing else. Both forms stay mounted at all times; the inactive one receives
+ * `isOpen={false}`, which hides its dialog while keeping its React state alive
+ * so a half-filled tab survives a switch away from it.
  */
 function AddProviderForm({
   isOpen,
@@ -528,10 +529,19 @@ function AddProviderForm({
     </TabRow>
   )
 
-  return tab === "known" ? (
-    <KnownProviderForm isOpen={isOpen} onClose={onClose} tabs={tabs} />
-  ) : (
-    <CustomProviderForm isOpen={isOpen} onClose={onClose} tabs={tabs} />
+  return (
+    <>
+      <KnownProviderForm
+        isOpen={isOpen && tab === "known"}
+        onClose={onClose}
+        tabs={tabs}
+      />
+      <CustomProviderForm
+        isOpen={isOpen && tab === "custom"}
+        onClose={onClose}
+        tabs={tabs}
+      />
+    </>
   )
 }
 
--- a/web/src/features/providers/ProvidersPage.tsx
+++ b/web/src/features/providers/ProvidersPage.tsx
@@ -147,20 +147,41 @@ function KnownProviderForm({
   isOpen,
   onClose,
   tabs,
+  providerId,
+  setProviderId,
+  apiKey,
+  setApiKey,
+  showAdvanced,
+  setShowAdvanced,
+  apiBase,
+  setApiBase,
+  name,
+  setName,
+  clientArgsText,
+  setClientArgsText,
+  credentials,
+  setCredentials,
 }: {
   isOpen: boolean
   onClose: () => void
   tabs: ReactNode
+  providerId: string
+  setProviderId: (v: string) => void
+  apiKey: string
+  setApiKey: (v: string) => void
+  showAdvanced: boolean
+  setShowAdvanced: (v: boolean) => void
+  apiBase: string
+  setApiBase: (v: string) => void
+  name: string
+  setName: (v: string) => void
+  clientArgsText: string
+  setClientArgsText: (v: string) => void
+  credentials: CredentialFieldValues
+  setCredentials: (v: CredentialFieldValues) => void
 }) {
   const create = useCreateStoredProvider()
   const test = useTestProviderCredentials()
-  const [providerId, setProviderId] = useState("")
-  const [apiKey, setApiKey] = useState("")
-  const [showAdvanced, setShowAdvanced] = useState(false)
-  const [apiBase, setApiBase] = useState("")
-  const [name, setName] = useState("")
-  const [clientArgsText, setClientArgsText] = useState("")
-  const [credentials, setCredentials] = useState<CredentialFieldValues>({})
   const clientArgs = parseClientArgs(clientArgsText)
   const credentialFields = credentialFieldsFor(providerId)
   const credentialErrors = validateCredentialFields(
@@ -178,7 +199,7 @@ function KnownProviderForm({
   // provider so it fires once per selection and does not clobber later edits.
   useEffect(() => {
     if (selected) setApiBase(selected.default_api_base ?? "")
-  }, [selected])
+  }, [selected, setApiBase])
   const envKeyPresent = selected?.env_key_present ?? false
   // The key is only mandatory when the provider needs one and its env var is not
   // already set on the server; any-llm falls back to that env var otherwise.
@@ -308,7 +329,7 @@ function KnownProviderForm({
       <button
         type="button"
         className="self-start text-xs font-medium text-link hover:text-link-hover"
-        onClick={() => setShowAdvanced((v) => !v)}
+        onClick={() => setShowAdvanced(!showAdvanced)}
       >
         {advancedOpen
           ? "Hide advanced"
@@ -358,18 +379,33 @@ function CustomProviderForm({
   isOpen,
   onClose,
   tabs,
+  name,
+  setName,
+  providerType,
+  setProviderType,
+  apiBase,
+  setApiBase,
+  apiKey,
+  setApiKey,
+  clientArgsText,
+  setClientArgsText,
 }: {
   isOpen: boolean
   onClose: () => void
   tabs: ReactNode
+  name: string
+  setName: (v: string) => void
+  providerType: string
+  setProviderType: (v: string) => void
+  apiBase: string
+  setApiBase: (v: string) => void
+  apiKey: string
+  setApiKey: (v: string) => void
+  clientArgsText: string
+  setClientArgsText: (v: string) => void
 }) {
   const create = useCreateStoredProvider()
   const test = useTestProviderCredentials()
-  const [name, setName] = useState("")
-  const [providerType, setProviderType] = useState("openai-compatible")
-  const [apiBase, setApiBase] = useState("")
-  const [apiKey, setApiKey] = useState("")
-  const [clientArgsText, setClientArgsText] = useState("")
   const clientArgs = parseClientArgs(clientArgsText)
 
   const nameHasDelimiter = /[:/]/.test(name)
@@ -502,9 +538,8 @@ type ProviderTab = "known" | "custom"
  *
  * What is shared is the frame. Each half renders the same `FormDialog` with the
  * same title, size and tab row, so switching tabs changes the fields and
- * nothing else. Both forms stay mounted at all times; the inactive one receives
- * `isOpen={false}`, which hides its dialog while keeping its React state alive
- * so a half-filled tab survives a switch away from it.
+ * nothing else. Only the active tab's `FormDialog` is mounted; field values for
+ * both tabs are lifted into this component so they survive a switch away.
  */
 function AddProviderForm({
   isOpen,
@@ -514,6 +549,25 @@ function AddProviderForm({
   onClose: () => void
 }) {
   const [tab, setTab] = useState<ProviderTab>("known")
+
+  // Known-tab field values — lifted so they survive when the tab is inactive.
+  const [knownProviderId, setKnownProviderId] = useState("")
+  const [knownApiKey, setKnownApiKey] = useState("")
+  const [knownShowAdvanced, setKnownShowAdvanced] = useState(false)
+  const [knownApiBase, setKnownApiBase] = useState("")
+  const [knownName, setKnownName] = useState("")
+  const [knownClientArgsText, setKnownClientArgsText] = useState("")
+  const [knownCredentials, setKnownCredentials] =
+    useState<CredentialFieldValues>({})
+
+  // Custom-tab field values — lifted for the same reason.
+  const [customName, setCustomName] = useState("")
+  const [customProviderType, setCustomProviderType] =
+    useState("openai-compatible")
+  const [customApiBase, setCustomApiBase] = useState("")
+  const [customApiKey, setCustomApiKey] = useState("")
+  const [customClientArgsText, setCustomClientArgsText] = useState("")
+
   const tabs = (
     <TabRow>
       {(
@@ -531,16 +585,44 @@ function AddProviderForm({
 
   return (
     <>
-      <KnownProviderForm
-        isOpen={isOpen && tab === "known"}
-        onClose={onClose}
-        tabs={tabs}
-      />
-      <CustomProviderForm
-        isOpen={isOpen && tab === "custom"}
-        onClose={onClose}
-        tabs={tabs}
-      />
+      {tab === "known" && (
+        <KnownProviderForm
+          isOpen={isOpen}
+          onClose={onClose}
+          tabs={tabs}
+          providerId={knownProviderId}
+          setProviderId={setKnownProviderId}
+          apiKey={knownApiKey}
+          setApiKey={setKnownApiKey}
+          showAdvanced={knownShowAdvanced}
+          setShowAdvanced={setKnownShowAdvanced}
+          apiBase={knownApiBase}
+          setApiBase={setKnownApiBase}
+          name={knownName}
+          setName={setKnownName}
+          clientArgsText={knownClientArgsText}
+          setClientArgsText={setKnownClientArgsText}
+          credentials={knownCredentials}
+          setCredentials={setKnownCredentials}
+        />
+      )}
+      {tab === "custom" && (
+        <CustomProviderForm
+          isOpen={isOpen}
+          onClose={onClose}
+          tabs={tabs}
+          name={customName}
+          setName={setCustomName}
+          providerType={customProviderType}
+          setProviderType={setCustomProviderType}
+          apiBase={customApiBase}
+          setApiBase={setCustomApiBase}
+          apiKey={customApiKey}
+          setApiKey={setCustomApiKey}
+          clientArgsText={customClientArgsText}
+          setClientArgsText={setCustomClientArgsText}
+        />
+      )}
     </>
   )
 }


web/src/features/providers/ProvidersPage.tsx
```diff
@@ -502,8 +538,8 @@ type ProviderTab = "known" | "custom"
  *
  * What is shared is the frame. Each half renders the same `FormDialog` with the
  * same title, size and tab row, so switching tabs changes the fields and
- * nothing else, and a half-filled tab still does not survive a switch away from
- * it, which is what it did as a panel.
+ * nothing else. Only the active tab's `FormDialog` is mounted; field values for
```
khaledosman, line 541: The unsaved-changes guard is disarmed by a tab switch, so the form now loses the pasted key on Escape instead of on the switch.

`useDirtySnapshot` seeds from the draft **on mount** (its docstring: "the component that owns the draft has to be the one that remounts per open"). The draft moved up here, but the hook stayed in `KnownProviderForm`/`CustomProviderForm`, which remount on every tab switch. So the reseed happens against already-filled values, `isDirty` reads false, and `FormDialog.requestClose` closes without arming the footer guard.

Two paths, both reproduced against 2695134 in `ProvidersPage.test.tsx`:

```
// 1. pick OpenAI, type a key, switch to Custom, switch back
expect(screen.getByLabelText("API key")).toHaveValue("sk-live-aaaa")  // passes
await user.keyboard("{Escape}")
await screen.findByRole("button", { name: "Discard" })
// TestingLibraryElementError: Unable to find role="button" and name "Discard"

// 2. fill the Custom tab, switch to Known, Escape
// same failure: the held-but-unrendered draft arms nothing
```

Case 2 does not even need the remount: the guard only ever sees the active tab's snapshot, so the other tab's draft is unguarded by construction. This is exactly the class the comment at line 209-215 records ("went on Escape with nothing asked"), and it is a regression, because on main there was no surviving draft to guard.

Fix: call `useDirtySnapshot` once in `AddProviderForm` over both drafts and pass `isDirty` down, rather than once per tab component.

web/src/features/providers/ProvidersPage.tsx
```diff
@@ -178,7 +199,7 @@ function KnownProviderForm({
   // provider so it fires once per selection and does not clobber later edits.
   useEffect(() => {
     if (selected) setApiBase(selected.default_api_base ?? "")
-  }, [selected])
+  }, [selected, setApiBase])
```
khaledosman, line 202: This effect now clobbers a hand-edited API base on every tab switch, which is the one field the PR silently rewrites rather than preserves.

`providerId` is lifted, so on remount `useProviderDetail` answers from the TanStack Query cache and `selected` is truthy on the first render; the effect then fires as a mount effect and overwrites `apiBase` with `default_api_base`. The comment two lines up ("fires once per selection and does not clobber later edits") no longer holds.

Reproduced against 2695134: pick OpenAI, open Advanced, set API base to `https://proxy.internal/v1`, switch to Custom, switch back:

```
Expected the element to have value:
  https://proxy.internal/v1
Received:
  https://api.openai.com/v1
```

An operator routing through a proxy loses that and submits against the provider's own endpoint. Fix: make the seed conditional on the provider rather than on mount, with the marker lifted alongside the rest of the draft, for example a lifted `knownApiBaseSeededFor` compared against `selected.id` and set in the same guarded render write `useDirtySnapshot`'s docstring describes.

Adding `setApiBase` to the dep array is harmless on its own (a `useState` setter is stable), but it will stop being harmless if the setter prop is ever wrapped.

web/src/features/providers/ProvidersPage.tsx
```diff
@@ -513,6 +549,25 @@ function AddProviderForm({
   onClose: () => void
 }) {
   const [tab, setTab] = useState<ProviderTab>("known")
+
+  // Known-tab field values — lifted so they survive when the tab is inactive.
+  const [knownProviderId, setKnownProviderId] = useState("")
+  const [knownApiKey, setKnownApiKey] = useState("")
```
khaledosman, line 555: No test for the behavior this PR changes, and the standards require one ("Add a Vitest test for any component, hook, or helper whose behavior you change"). `ProvidersPage.test.tsx` already has the harness and the neighbouring guard tests at lines 744 and 765; a case that fills the known tab, switches away and back, and asserts both the surviving key *and* that Escape still offers Discard covers the fix and both findings above in one.

web/src/features/providers/ProvidersPage.tsx
```diff
@@ -147,20 +147,41 @@ function KnownProviderForm({
...
+  setCredentials,
 }: {
   isOpen: boolean
   onClose: () => void
   tabs: ReactNode
+  providerId: string
+  setProviderId: (v: string) => void
+  apiKey: string
+  setApiKey: (v: string) => void
+  showAdvanced: boolean
```
khaledosman, line 172: Nit: booleans read as a question in this tree, props included (`naming-conventions.md`: "including props and the boolean fields of a hook's return value"). `isAdvancedShown` as the prop, with `setIsAdvancedShown`. Moot if the setters collapse into one draft object.

Related, line 332: `setShowAdvanced(!showAdvanced)` replaces the functional updater because the prop is typed `(v: boolean) => void`. Typing these as `Dispatch<SetStateAction<T>>` keeps the updater form available.

web/src/features/providers/ProvidersPage.tsx
```diff
@@ -513,6 +549,25 @@ function AddProviderForm({
   onClose: () => void
 }) {
   const [tab, setTab] = useState<ProviderTab>("known")
+
+  // Known-tab field values — lifted so they survive when the tab is inactive.
```
khaledosman, line 553: Em dash in a doc comment; AGENTS.md's writing style rules them out of comments (a colon reads the same here). Same on line 563.

khaledosman (changes requested):
Lifting the draft above the tab components fixes the reported loss but moves it: the draft now outlives the mount, while the three things keyed to the draft's lifetime (`useDirtySnapshot`, the `api_base` prefill effect, and the create mutation) stayed below it. Two of those are user-visible losses, both reproduced against 2695134 with the existing `ProvidersPage.test.tsx` harness; details inline.

`web/design/feedback.md` ("The component that renders the `FormDialog` owns everything that resets between opens: the draft *and* its mutation") is the rung: the draft and the hooks scoped to it have to share one mount. Moving `useDirtySnapshot` up into `AddProviderForm` over both drafts, and seeding `api_base` off the selected provider rather than off mount, keeps the current structure. Collapsing the two tabs onto one `FormDialog` that swaps only its fields is the other way, and larger than this PR.

The description also describes a different implementation than the diff: it says both panels render with `isOpen={false}` on the inactive one, but the code mounts only the active tab (`{tab === "known" && …}`). Worth correcting, since it is what a reader evaluates.

AloysJehwin:
Both findings confirmed and fixed in 3217e24d, rebased onto current `main` first (the branch was 68 commits behind, though `ProvidersPage.tsx` itself had not moved).

**The disarmed guard.** Correct, and your diagnosis of the mechanism was exact: `useDirtySnapshot` seeds with `useState(snapshot)` on mount, the tab components remount on every switch, so the reseed happened against already-filled values and `isDirty` read false. Case 2 was the sharper one — the inactive tab's draft was unguarded by construction, not just after a remount.

Computed once in `AddProviderForm` over both drafts now and passed down as a prop. Both tab components take `isDirty` rather than calling the hook.

**The clobbered API base.** Also confirmed. With `providerId` lifted, `useProviderDetail` answers from the TanStack Query cache, so `selected` is truthy on the first render after a switch and the mount-keyed effect fired again. The comment two lines above it ("fires once per selection and does not clobber later edits") had stopped being true.

Keyed on the provider already seeded, with the marker lifted alongside the rest of the draft, as you suggested:

```ts
useEffect(() => {
  if (selected && apiBaseSeededFor !== selected.id) {
    setApiBaseSeededFor(selected.id)
    setApiBase(selected.default_api_base ?? "")
  }
}, [selected, apiBaseSeededFor, setApiBaseSeededFor, setApiBase])
```

**The test.** One case covering both findings plus the draft survival the PR is for: fill the known tab, edit the API base to a proxy, switch away and back, then assert the key, the base, and that Escape still offers Discard. Verified it is a real regression test — against the reviewed commit it fails on the API base with exactly the mismatch you reported:

```
Expected the element to have value: https://proxy.internal/v1
Received:                           https://api.openai.com/v1
```

46 passed at this head, `tsc --noEmit` and Biome clean.

**The description.** You were right that it described a different implementation. Rewritten to match the diff: only the active tab mounts (`{tab === "known" && …}`), and the state that survives is lifted rather than both panels rendering with `isOpen={false}`.

I left the two nits: `showAdvanced` naming and the `Dispatch<SetStateAction<T>>` typing are both worth doing, but they touch every prop on these two components, and the diff is already large enough that separating them seemed better for review. Happy to fold them in here if you'd rather not have a follow-up.

web/src/features/providers/ProvidersPage.tsx
```diff
@@ -173,12 +200,18 @@ function KnownProviderForm({
...
+  // Prefill the (editable) API base with the provider's built-in default once
+  // its detail loads, so Advanced shows what will be used. Keyed on the provider
+  // already seeded, held above this component, rather than on mount: the draft
+  // now outlives a tab switch while this component does not, so `selected` is
+  // truthy on the first render after a switch (TanStack Query answers from
+  // cache) and a mount-keyed effect overwrote a hand-edited base every time.
   useEffect(() => {
-    if (selected) setApiBase(selected.default_api_base ?? "")
-  }, [selected])
+    if (selected && apiBaseSeededFor !== selected.id) {
```
khaledosman, line 217: `apiBaseSeededFor` is never cleared, so re-picking a provider after clearing the picker leaves the API base empty.

`ProviderComboBox` emits `onChange("")` when the selection is cleared (`providerFields.tsx:272`). The handler below does `setApiBase("")` but leaves `apiBaseSeededFor` on the old id, so re-selecting that same provider hits `apiBaseSeededFor === selected.id` and this effect skips. Advanced then shows a blank base where it showed the built-in default, which is the one thing this effect's own comment says it is for.

Verified: a test that selects OpenAI, clears the picker, re-selects OpenAI and asserts the base passes at the merge base `2f93c332` and fails at this head (`Expected: https://api.openai.com/v1, Received: ""`).

One line beside the existing clear in the combobox handler (~line 294) fixes it, with all 46 existing tests plus the new one still green:

```suggestion
          setApiBase("")
          setApiBaseSeededFor(null)
```

(that suggestion belongs on the `setApiBase("")` line in `onChange`, not here)

web/src/features/providers/ProvidersPage.tsx
```diff
@@ -188,18 +221,10 @@ function KnownProviderForm({
   // Require the key when the chosen provider says it needs one; keyless local
   // backends (Ollama, llama.cpp) can submit without it.
   // One snapshot of everything the form owns, seeded on mount, rather than a
```
khaledosman: Orphaned half-sentence left from the removed `useDirtySnapshot` call: it survived as diff context and now runs into the replacement comment below it. Delete this line.

web/src/features/providers/ProvidersPage.tsx
```diff
@@ -513,6 +548,50 @@ function AddProviderForm({
   onClose: () => void
 }) {
   const [tab, setTab] = useState<ProviderTab>("known")
+
+  // Known-tab field values — lifted so they survive when the tab is inactive.
```
khaledosman: Em dash in a doc comment, here and again on line 567. AGENTS.md ("Writing style") bans em dashes as separators in prose including doc comments; a colon reads the same:

```suggestion
  // Known-tab field values: lifted so they survive when the tab is inactive.
```

khaledosman (changes requested):
One reproducible regression from the lifted state, plus two comment cleanups. The lift itself holds up: `key={addOpenCount}` still remounts the draft owner per open, and the combobox reseeds its display text from the preserved `providerId`, so the picker is not blank after a switch.

CodeRabbit's "do not retain API keys in React state across tab switches" was written against the earlier commit's claim that both panels stay mounted; the premise is stale and the key's lifetime is still bounded by the per-open remount.

web/src/features/providers/ProvidersPage.tsx
```diff
@@ -502,8 +502,9 @@ type ProviderTab = "known" | "custom"
  *
  * What is shared is the frame. Each half renders the same `FormDialog` with the
  * same title, size and tab row, so switching tabs changes the fields and
- * nothing else, and a half-filled tab still does not survive a switch away from
- * it, which is what it did as a panel.
+ * nothing else. Both forms stay mounted at all times; the inactive one receives
+ * `isOpen={false}`, which hides its dialog while keeping its React state alive
+ * so a half-filled tab survives a switch away from it.
```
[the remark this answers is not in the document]
khaledosman: Premise no longer holds: this was written against the earlier commit, which claimed both panels stay mounted. Only the active tab mounts (`{tab === "known" && …}`); what was lifted is the state.

The key's lifetime is unchanged in the way that matters: `AddProviderForm` is keyed on `addOpenCount` and remounts per open, so the draft, pasted key included, is discarded on every close. A controlled `SecretField` has to hold its value in React state to be typed into at all, and it is still never rendered back or logged.

web/src/features/providers/ProvidersPage.tsx
```diff
@@ -528,10 +529,19 @@ function AddProviderForm({
...
+        isOpen={isOpen && tab === "known"}
+        onClose={onClose}
+        tabs={tabs}
+      />
+      <CustomProviderForm
+        isOpen={isOpen && tab === "custom"}
+        onClose={onClose}
+        tabs={tabs}
+      />
+    </>
```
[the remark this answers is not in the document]
khaledosman: Both run clean at 46e08b67.

`make lint`: architecture check (no violations), single alembic head, Ruff all checks passed.
`pnpm --dir web run lint`: Biome, 660 files, no fixes applied.
Also `pnpm --dir web run typecheck` (clean) and `pnpm --dir web test` (6657 tests, 202 files, all passing).

web/src/features/providers/ProvidersPage.tsx
khaledosman: Fixed in 3217e24d: the snapshot moved up to `AddProviderForm`, taken once over both drafts, so it is seeded on the component that actually remounts per open rather than on one that remounts per tab switch. Covered by the test at ProvidersPage.test.tsx.

web/src/features/providers/ProvidersPage.tsx
khaledosman: Fixed in 3217e24d: the effect is keyed on `apiBaseSeededFor`, lifted with the draft, rather than on mount.

That fix had a gap of its own, now closed in 46e08b67: the marker was never cleared, so emptying the picker and choosing the same provider again skipped the reseed and left the base blank. Both paths have a test.

web/src/features/providers/ProvidersPage.tsx
khaledosman: Added. 3217e24d covers the surviving draft, the armed guard and the preserved API base in one case; 46e08b67 adds the clear-then-repick path. `pnpm --dir web test` is green at 6657 tests.

web/src/features/providers/ProvidersPage.tsx
khaledosman: Done in 46e08b67: `isAdvancedShown` / `setIsAdvancedShown` for the prop and the lifted state, and every lifted setter is now typed `Dispatch<SetStateAction<T>>`, so line 343 is back to the functional updater `setIsAdvancedShown((v) => !v)`.

web/src/features/providers/ProvidersPage.tsx
khaledosman: Fixed in 46e08b67: both are colons now.

web/src/features/providers/ProvidersPage.tsx
khaledosman: Fixed in 46e08b67: `setApiBaseSeededFor(null)` sits beside `setApiBase("")` in the picker's `onChange`. The regression test added with it fails without that line.

web/src/features/providers/ProvidersPage.tsx
khaledosman: Removed in 46e08b67. The paragraph that followed it had also come to sit between the `canSubmit` comment and `canSubmit`, so it moved above both.

web/src/features/providers/ProvidersPage.tsx
khaledosman: Fixed in 46e08b67: both colons.

khaledosman:
Pushed 46e08b67 to this branch (`maintainerCanModify`), addressing every open thread.

**The regression from the last round.** Emptying the provider picker clears the API base but left `apiBaseSeededFor` on the old id, so picking that same provider back read as already seeded and the field stayed blank instead of showing the built-in default. `setApiBaseSeededFor(null)` now sits beside `setApiBase("")` in the picker's `onChange`. The test added with it passes at the merge base and fails at 3217e24d, so it pins the regression rather than the fix.

**From the earlier review.** The boolean prop reads as a question (`isAdvancedShown`), and the lifted setters are typed `Dispatch<SetStateAction<T>>`, which puts the functional updater back on the Advanced toggle. Comments lose a stranded half-sentence left by diff context, two em dashes, and a paragraph that had come to sit between `canSubmit` and its own note.

**Checks at this head.** `make lint` clean (architecture, single alembic head, Ruff), `pnpm --dir web run lint` clean, `pnpm --dir web run typecheck` clean, `pnpm --dir web test` 6657 passing across 202 files. Dashboard-only, so no OpenAPI, Postman, schema or route-tree artifact is owed.

CodeRabbit's API-key thread was written against the earlier commit's claim that both panels stay mounted; only the active tab mounts, and the draft is still discarded per open by `key={addOpenCount}`.

khaledosman (approved):

Issue #1114, opened 2026-09-13
Add the dashboard card that defines a guardrail


dpoulopoulos:
Add the dashboard card that defines a guardrail, modeled on `web/src/features/providers/ProvidersPage.tsx`.

Pick a guardrail from the catalog and typed fields render from its create and validate parameter specs.

## Scope

- Reuse `GuardrailParameterFields.tsx`, `GuardrailParametersSection.tsx` and `guardrailParameters.ts` as they are. They already handle every `ParameterType` plus a raw-JSON fallback, and already pass the `***` mask (`REDACTED_SECRET` from `shared/helpers/redaction.ts`) through untouched at `:93`, `:197`, `:227`.
- Secrets use `SecretField` (`design-system/forms/SecretField.tsx`), a never-prefilled password input. It has no stored-value preview today, so an edit form shows "Set" or "Not set" per secret from the `***` map and asks for a new value only when the operator types one.
- A secret whose `storable` is false renders disabled with a note that it cannot be stored.
- A guardrail whose `runnable` is false renders disabled with the extra to install, taken from `missing_extra`.
- Gate the add flow the way `ProvidersPage.tsx:1039` does: on `useProviderKeyEncryption()`, which reads `provider_key_encryption_available` from the membership context (`shared/api/organizations.ts:56-61`). Do not read the operator-only `secret_key_configured` settings field; #839 explains why.
- Lives beside the existing `OrganizationGuardrailsCard` in `ToolsGuardrailsPage` (`:464-466`). That card is unchanged; its profile picker gains the locally defined names.
- All hooks and cache invalidation go in `web/src/shared/api/tools.ts`, next to `useGuardrailProfiles` (`:179`). Biome (`web/biome.jsonc:105`) forbids a file under `src/features/**` from importing `@/shared/api/queryKeys`.

## Success criteria

- An operator can add, edit, disable and delete a guardrail without leaving the browser.
- A stored secret is never prefilled and never echoed back.
- The three web test suites pass.

Issue #1108, opened 2026-09-13
[EPIC] Run guardrails inside Otari and remove the sidecar


dpoulopoulos:
## Problem

Running a guardrail needs a second container. `docker compose --profile guardrails up` starts `anyguardrails` (and usually `encoderfile` with it), and Otari calls it over `POST /validate` at `OTARI_GUARDRAILS_URL`. Which guardrails exist is fixed by a YAML file mounted into that container at boot, so adding one means editing a file on disk and restarting a container.

## Outcome

An operator adds a guardrail in the dashboard the way they add a provider: pick it, fill typed fields, paste the secrets it needs. No second container, no YAML, no restart. Otari then runs the guardrail itself through `any-guardrail`, and the sidecar is removed.

The caller's request contract does not change. `{"profile": "...", "mode": "block"}` still works, and `organization_guardrails.profile` keeps pointing into the same namespace, so organization and routing-policy mandates need no migration.

## Why this is cheap

Verified against the installed `any_guardrail` 0.7.7:

- Base `any-guardrail` (no extras) needs only `any-llm-sdk`, `pydantic` and `requests`, all already installed. `import any_guardrail` loads neither torch nor transformers.
- 9 of its 40 guardrails are `backend=hosted_api, requires_api_key=True`: a plain HTTP call with a key, no model download. `AnyGuardrail.create(GuardrailName.LAKERA_GUARD, api_key="x")` constructs today with nothing extra installed.
- The other 31 need local models and go behind one opt-in extra.

`any-guardrail` is already a direct dependency (`pyproject.toml:24`), imported at one site (`src/gateway/services/guardrail_catalog.py:38-39`) for parameter metadata only. Nothing in this repo constructs or calls a guardrail today.

## Out of scope

- Metering the provider call an LLM-judge guardrail makes. Otari does not meter it today either.
- Guardrail latency. Guardrails run one at a time, each with its own 30s timeout, so N guardrails can take N x 30s. Giving the whole set one shared deadline is a separate change.
- Output-direction enforcement. `on: ["output"]` stays parsed and unenforced.
- A second Docker image with local-model support baked in.

PR #1116, opened 2026-09-13
feat(guardrails): publish a catalog of the guardrails otari can run itself

### LINKED ISSUE #1109: Publish a catalog of the guardrails Otari can run itself
Publish the guardrails this gateway can run itself, so a form can be built from them.

`services/guardrail_catalog.py` reads `any_guardrail.parameter_registry` today, but only to describe profiles the sidecar reported, and only their `validate`-stage parameters. Add a second source next to it: the full built-in list from `GUARDRAIL_METADATA`, with **both** stages.

The `create` stage is the point. It is where secrets live (`ParameterSpec(name='api_key', stage=CREATE, secret=True, env_var='LAKERA_API_KEY')`), and today's catalog never publishes it because the sidecar's constructor is fixed by its YAML at boot. That reason goes away once Otari runs the guardrail. Only 22 of the 40 guardrails have any validate-stage parameter, so the existing catalog renders an empty form for 18 of them.

## Scope

- A built-in spec per guardrail: `guardrail_name`, `display_name`, `description`, `primary_category`, `backend` (`hosted_api` / `local_encoder` / `local_decoder` / `library_wrapped`), `requires_api_key`, `runnable`, `missing_extra`, and the typed parameters split by stage.
- Each secret parameter also carries `storable`, false when its type is `json`. `boto3_session` and `api_client` are live objects and cannot be written to a database.
- Reuse `GuardrailParameterSpec` and the degrade-to-json rule already in the file (`:58-65`, `:131-162`). Factor `_parameter_specs` to take a stage filter instead of hardcoding `"validate"` at `:160`.
- New route on `reader_router` in `api/routes/tool_settings.py`, beside `GET /tool-settings/guardrails/profiles` (`:187-215`). Same auth, same router.
- `runnable` comes from a per-guardrail `importlib.util.find_spec` probe of the top-level packages its extra brings. Never construct a guardrail, never import torch. One table maps `GuardrailName` to its extra and probe packages; a guardrail missing from the table reports `runnable: false` with `missing_extra: null` rather than a guess.

The existing `/guardrails/profiles` endpoint and the dashboard are untouched.

## Success criteria

- The new endpoint lists all 40 guardrails with create and validate parameters.
- `runnable` is correct with and without the local extra installed.
- OpenAPI and Postman regenerated and committed.

dpoulopoulos:
## Description

Otari can only use a guardrail today if a second container is running. That container decides which guardrails exist, from a YAML file mounted into it at boot. Adding one means editing a file on disk and restarting the container.

The guardrail library itself is already installed here. We just never used it to run anything, only to look up what a guardrail's settings are called.

This PR adds one read-only endpoint that lists every guardrail that library ships, and what each one needs to be set up. Nothing reads it yet. Nothing an operator does today changes. It is the list a form will later be built from, so adding a guardrail becomes the same job as adding a provider: pick it from a menu, fill in the fields, paste your API key.

Two details worth calling out:

- The list now includes **setup settings**, not just per-call settings. That is where an API key lives, so it is the part that matters. The old endpoint could not include them, because the sidecar's setup is fixed by its own YAML.
- Each guardrail says whether it can actually run here (`runnable`). Some need a local model and so need extra packages. We answer that by checking whether the packages are installed, never by loading a model, so listing the menu stays fast and light.

### 1. Start Postgres only

```bash
docker compose up -d postgres
docker compose ps postgres     # wait for (healthy)
```

Postgres only. The `otari` service pulls the published image from Docker Hub, which does not have this branch. The container maps host port `5433` by default, or whatever `OTARI_POSTGRES_PORT` is set to in your `.env`.

### 2. Point the app at it

`make dev` runs `load_config()` with no file path, so `config.yml` is never read. That is why its `postgres:5432` does not apply here: that hostname only resolves inside Docker's network, and you are on the host. Put the settings in `.env` instead (gitignored):

```bash
cat >> .env <<'EOF'
OTARI_DATABASE_URL=postgresql://otari:otari@localhost:5433/otari
OTARI_MASTER_KEY=local-dev-master-key
EOF
```

If your database already generated a master key on first run, setting `OTARI_MASTER_KEY` to a different value logs out open dashboard sessions. That is `revoke_sessions_on_master_key_change` doing its job. Sign in again with the new key.

### 3. Run the gateway from source

```bash
make dev
```

Just that, with no `OTARI_PORT=` prefix: the Makefile sources `.env` afterwards, so `.env` always wins. It binds `8000` unless `.env` says otherwise.

### 4. Call the new endpoint

```bash
curl -s -H "Authorization: Bearer local-dev-master-key" \
  http://localhost:8000/api/v1/tool-settings/guardrails/catalog | jq '{
    total: (.guardrails | length),
    runnable: [.guardrails[] | select(.runnable)] | length,
    backends: (.guardrails | group_by(.backend) | map({(.[0].backend): length}) | add),
    lakera_api_key: (.guardrails[] | select(.guardrail_name=="lakera_guard")
                     | .create_parameters[] | select(.name=="api_key")
                     | {secret,required,storable}),
    llama_guard: (.guardrails[] | select(.guardrail_name=="llama_guard")
                  | {runnable,missing_extra})}'
```

Expected, and what it returns here:

```json
{
  "total": 40,
  "runnable": 8,
  "backends": { "hosted_api": 9, "library_wrapped": 4,
                "local_decoder": 15, "local_encoder": 12 },
  "lakera_api_key": { "secret": true, "required": true, "storable": true },
  "llama_guard": { "runnable": false, "missing_extra": "guardrails-local" }
}
```

8 runnable are the hosted APIs. The ninth, `azure_content_safety`, needs an SDK that is not in the base install, so it reports `runnable: false` correctly.

### 5. Confirm nothing broke

The old endpoint still answers the same way. With no guardrails container running it reports unavailable with a reason, not an error:

```bash
curl -s -H "Authorization: Bearer local-dev-master-key" \
  http://localhost:8000/api/v1/tool-settings/guardrails/profiles | jq
```

A 401 means the key does not match `.env`; restart `make dev`, which reads it only at startup. A 404 means you are on `main`, not this branch.

**Take those two lines back out of `.env` before running the test suite.** `load_config` reads `.env`, so a set `OTARI_MASTER_KEY` or `OTARI_DATABASE_URL` fails 14 unit tests across `test_master_key_service`, `test_settings_endpoint`, `test_dashboard_session`, `test_deps_db_error` and `test_gateway_cli`. They cover first-run key generation and the SQLite default, which cannot happen when the environment already pins both. This is not new here, but these instructions are the first thing that asks you to set them.

### Already proven by automated checks

- `tests/unit/test_guardrail_catalog.py` (30 tests) covers the list contents, both setting stages, the `runnable` probe in both directions, and a guard that building the list never loads `torch` or `transformers`.
- `tests/unit/test_tool_settings_endpoint.py` covers the route: auth, shape, and that it stays unmounted in hybrid mode.
- Run locally and green: `make lint`, `make typecheck`, `tests/unit` (3221), `tests/integration -k guardrail` (49), `tests/integration/test_api_prefix_contract.py`, `make openapi-check`, `make postman-check`, and the dashboard's `lint`, `typecheck`, `build` and `test` (5706). The full `tests/integration` suite was not run locally; CI covers it.

## Known limitation

`susfactor` can also run over a hosted API with no extra packages. One `runnable` flag per guardrail cannot say "no locally, yes hosted", so it reports `false`. The response publishes `alternate_backends`, so the dashboard work in #1114 can surface that path. It is the only guardrail with a second backend.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Closes #1109. Part of #1108.

## Checklist

- [x] Documentation was updated where necessary. No user-facing docs change: the endpoint backs a form that does not exist yet, so `docs/guardrails.md` still describes the current flow correctly.
- [x] If the API contract changed, I regenerated the OpenAPI spec (`uv run python scripts/generate_openapi.py`). Postman and `web/src/client/schema.ts` regenerated with it.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:** Claude Opus 5, via Claude Code

Planned against the issue, then implemented test-first. The full unit suite caught one thing the plan missed: a new endpoint has to be classified in `scripts/sdk_codegen/sdk-endpoints.txt`, so it is listed there as not yet wrapped, beside the other tool-settings routes.

## Summary

- Added an authenticated catalog endpoint for all built-in guardrails.
- Included setup parameters, metadata, backend information, licensing, and availability details.
- Added safe checks for missing package extras without loading model libraries.
- Preserved the existing guardrail profiles endpoint and dashboard behavior.
- Updated tests, OpenAPI client schemas, and Postman documentation.
- Added `storable` metadata for secret parameters so forms can identify values that can be saved safely.

## Benefits

The catalog allows configuration forms to list supported guardrails and generate their setup fields automatically. Operators can also see which guardrails are available in the current installation.


src/gateway/api/routes/tool_settings.py
```diff
@@ -215,6 +223,27 @@ async def list_guardrail_profiles(
...
+    state. ``runnable`` reports whether the modules a guardrail's backend needs
+    are installed here, probed rather than imported, and ``missing_extra`` names
+    the Otari extra that would fix it.
+
+    On the reader router, and for the same reason: a guardrail name is what a
+    caller puts in a request body, and no endpoint address appears in the answer.
+    """
+    return build_builtin_guardrail_catalog()
+
+
```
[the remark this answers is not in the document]
dpoulopoulos: Good catch on the gate, though I went the other way with it.

You're right that the rationale I wrote didn't hold. I said "same reason as the profiles read", but the profiles read sits on `verify_master_key` because fetching it dials the deployment's guardrails service, and this one dials nothing. So the justification was borrowed and wrong.

I didn't move it to `verify_catalog_reader` because that would loosen it. That plane admits a data-plane API key on top of a dashboard session, and it's a deliberately closed set of three deployment-describing reads (`/models`, `/pricing`, `/tools`). Two reasons this isn't one of them:

- It's the picker behind a form that stores a vendor API key deployment-wide. That's an operator's action, not a workspace credential's.
- `runnable` reports which packages are installed on this host. That's infrastructure detail, not something a tenant is owed about their own requests.

The nearest precedent agrees: `GET /api/v1/providers/catalog`, the add-a-provider picker this one is modeled on, is on `require_deployment_operator`. So I moved this to `operator_router` to match, which is stricter than where it was.

A profile *name* is the one thing a caller genuinely needs, and the profiles read next door already publishes that set. It stays on the reader router, which also restores `api/main.py`'s comment that the reader is "the one route a tenant may reach".

The docstring now states this outright instead of borrowing it, and `test_guardrail_catalog_is_an_operator_read` pins the placement.

On the lint note: `make lint` was run, along with `make typecheck`, and it passes.

src/gateway/services/guardrail_catalog.py
```diff
@@ -253,3 +281,254 @@ async def fetch_guardrail_catalog(base_url: str | None) -> GuardrailCatalog:
             "Guardrail catalog from %s held %d rows this gateway could not read", shown, len(body) - len(profiles)
         )
     return GuardrailCatalog(available=True, profiles=sorted(profiles, key=lambda spec: spec.profile))
+
+
+# ---------------------------------------------------------------------------
```
javiermtorres, line 302: Can this info and models be already obtained from any-guardrails?

dpoulopoulos: Hey, @javiermtorres, very good catch. Yes we can and it seems much cleared. Could you take a look to the last fixup commit I added here?

javiermtorres (approved):

PR #1117, opened 2026-09-13
fix(usage): capture provider-reported latency for Groq and Ollama

### LINKED ISSUE #337: Capture provider-reported timing as a best-effort provider_latency_ms enrichment (Groq/Ollama first)

## Summary

Capture provider-reported timing (best-effort) as a separate, nullable `provider_latency_ms` enrichment on the usage log, surfaced alongside the existing "Total time". This is an addition, not a replacement: otari's self-timed `latency_ms` stays the canonical end-to-end measure.

## Background

Today "Total time" is otari's own server-side measurement: `started_at = time.monotonic()` at the top of the handler preamble (`src/gateway/api/routes/_pipeline.py:542`), settled as `latency_ms = _elapsed_ms(ctx.started_at)` when the `UsageLog` row is written (`_elapsed_ms` at `_pipeline.py:954`; column at `src/gateway/models/entities.py:311`; rendered as "Total time" in `web/src/pages/ActivityPage.tsx:161`, `:399`).

That is the right canonical number: it is uniform across every provider and reflects what the caller experiences (gateway overhead + network to/from the provider + streaming duration).

Some providers also report their own server-side timing, which measures only provider compute. The gap between that and otari's `latency_ms` is essentially our gateway + network overhead, which is a useful diagnostic.

## What providers expose (inconsistent, hence best-effort)

- Groq (OpenAI-compatible body): `usage.queue_time`, `usage.prompt_time`, `usage.completion_time`, `usage.total_time` (seconds), plus an `x_groq` block.
- Ollama: `total_duration`, `load_duration`, `prompt_eval_duration`, `eval_duration` (nanoseconds).
- OpenAI: nothing in the body; only an `openai-processing-ms` response header.
- Anthropic (messages): nothing timing-related in body or headers.
- Most others: nothing.

## Feasibility notes

- Body-level timing is reachable: any-llm returns openai-typed objects, and openai's models are `extra="allow"` (`openai/_models.py`), so unknown fields survive into `result.model_extra` / `result.usage.model_extra`. Survival is adapter-dependent though: OpenAI-compatible passthrough providers (Groq) tend to keep the `usage` extras; adapters that hand-map fields (Ollama) may drop them. Verify per adapter.
- Header-level timing (OpenAI's `openai-processing-ms`) is NOT reachable through any-llm's normalized return; it hands back the parsed object, not the raw httpx response, and does not surface `.with_raw_response`. Out of scope unless/until any-llm exposes headers.
- Streaming: provider timing, when present, typically only lands in the final chunk's usage; must be read at settlement, tolerate absence.

## Proposed scope

- Add a nullable `provider_latency_ms` (int, ms) column to `UsageLog` (Alembic migration).
- Populate best-effort from `model_extra` at settlement, normalizing units to ms (Groq seconds, Ollama nanoseconds). Leave NULL when the provider reports nothing or the adapter dropped it. Never let extraction failure affect billing or the request.
- Surface it in the usage/activity API (`api/routes/usage.py`, `users.py`) and show it next to "Total time" in `ActivityPage` when present (e.g. "Provider time"), making clear it is provider compute, not end-to-end.
- Start with Groq and Ollama; treat everything else as "not reported".

## Explicitly out of scope

- Replacing `latency_ms`. It stays canonical.
- Reading response headers / patching any-llm for `openai-processing-ms`.

## Open questions

- One generic `provider_latency_ms`, or keep the richer breakdown (queue / prompt / completion) where a provider gives it? Leaning toward a single normalized total first, with room to add a JSON detail blob later.

AmirF194:
## Description

`GatewayUsage.from_completion_usage` and the chat adapter's streaming usage extraction both rebuild a fresh usage object naming only their own explicit fields, so any extras any-llm left on the source object (Groq's `total_time`, Ollama's `total_duration`) never survived to settlement. Forwarded those extras through in both places and added `provider_latency_ms_of` to normalize whatever lands there into a new nullable `provider_latency_ms` column on the usage log, shown next to "Total time" on the activity page as "Provider time". `latency_ms` stays canonical; this is a diagnostic alongside it, null for every other provider.

One thing worth flagging: this keys off the resolved instance name (`groq`/`ollama`), which is also the provider type unless someone sets an explicit `provider_type:` override on a differently-named instance. That case falls back to "not reported" rather than the enrichment, same as an unmapped provider.

## How to test it locally

`uv run pytest tests/unit/test_provider_latency.py tests/unit/test_pipeline_settlement.py::test_log_usage_writes_provider_latency_ms_onto_the_row -v` covers the normalization (seconds vs. nanoseconds, absent field, wrong type), both forwarding paths, and the actual write onto the row inside `log_usage` (that line has no other unit coverage, so it's tested directly rather than through the two extraction functions alone).

Ran the full `tests/unit` suite, `ruff`, `mypy`, `check_architecture.py`, `check_alembic_heads.py`, `openapi-check`, `postman-check`, and the dashboard's lint/typecheck/vitest/schema regen, all clean. Didn't run `tests/integration` (needs a live Gemini key) or the SDK codegen check (needs a JRE I don't have here); the schema change is an additive nullable int in the same shape as `latency_ms` right next to it, so I'd be surprised if either broke.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes #337

## Checklist

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

## Summary

- Added optional provider latency tracking for Groq and Ollama.
- Converted provider timing to milliseconds and stored it separately from canonical end-to-end latency.
- Exposed provider latency through usage APIs and the activity view.
- Added safeguards for missing, malformed, negative, non-finite, and oversized timing data.
- Preserved provider usage extras without allowing them to overwrite normalized usage fields.
- Added time-to-first-token tracking for streamed requests.
- Added tests for extraction, normalization, forwarding, persistence, and display.

This provides provider compute-time and streaming diagnostics without affecting billing or request handling.


AmirF194:
Both real. Pushed a fix covering both plus one thing I found while verifying.

`provider_latency_ms_of` is keyed by implementation, not instance name. `log_usage` now takes an optional `provider_type`, resolved from the instance at each call site that actually carries usage data (vision billing, standalone dispatch, streaming settlement). The instance name still keys usage/pricing rows as before.

`round(inf)`/`round(nan)` both raise, so added a finite check before the multiply.

The actual CI failure isn't either finding: `test_list_usage_response_shape`'s expected dict predates this PR's new column and was missing `provider_latency_ms`. Fixed in the same push.

Verified locally: 186 unit tests across the touched files plus ruff/mypy/the architecture check, all clean. Couldn't re-run `test_usage_endpoint.py` itself here, testcontainers needs a docker socket this sandbox doesn't pass through to nested containers. The fix is one missing dict key matching the CI diff exactly, so I'm confident in it without having watched it go green locally.

src/gateway/core/usage.py
```diff
@@ -116,3 +122,32 @@ def cache_tokens_in_prompt_of(usage: CompletionUsage) -> bool:
...
+    never raise or affect billing.
+    """
+    if provider is None:
+        return None
+    field = _PROVIDER_LATENCY_FIELDS.get(provider)
+    if field is None:
+        return None
+    key, to_ms = field
+    raw = (usage.model_extra or {}).get(key)
+    if isinstance(raw, bool) or not isinstance(raw, int | float) or not math.isfinite(raw):
```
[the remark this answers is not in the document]
AmirF194: Real, and the previous fix missed it: isfinite on the raw value doesn't cover overflow after scaling. Groq's total_time at 1e308 is finite but 1e308 * 1000 is inf, and round(inf) raises. Pushed 0a822cd checking the scaled value instead, plus a regression test.

alembic/versions/615fa323e931_add_usage_logs_provider_latency_ms.py
```diff
@@ -0,0 +1,30 @@
...
+"""
+
+from collections.abc import Sequence
+
+import sqlalchemy as sa
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision: str = "615fa323e931"
+down_revision: str | Sequence[str] | None = "f1c4a8e2d6b9"
```
khaledosman, line 16: Branched revision graph once this merges. `f1c4a8e2d6b9` is already claimed by `a1d4f7c2e8b3` on main, and main's head is now `c4e7a9b2d6f8`. Merging `origin/main` into this head locally gives:

```
2 Alembic heads. `alembic upgrade head` cannot resolve a target:

  615fa323e931  down_revision=f1c4a8e2d6b9
  c4e7a9b2d6f8  down_revision=a1d4f7c2e8b3
```

`make lint` runs `check-migrations`, so this fails lint, and every unit test that builds a schema errors with it (`test_pricing_provenance_schema_chain.py`, 5 errors). The green checks here predate `c4e7a9b2d6f8`; the branch is 14 commits behind.

Rebase and re-point `down_revision` at `c4e7a9b2d6f8`. The ordering changes, so re-run the migration round-trip afterwards.

src/gateway/api/routes/chat.py
```diff
@@ -189,13 +189,17 @@ def extract_stream_usage(self, chunk: ChatCompletionChunk) -> CompletionUsage |
         if not chunk.usage:
             return None
         details = chunk.usage.prompt_tokens_details
-        return GatewayUsage(
+        # Same forwarding as GatewayUsage.from_completion_usage (otari#337):
+        # provider timing on a streamed response lands in the final chunk's usage.
+        fields = dict(chunk.usage.model_extra or {})
```
khaledosman, line 194: Splatting `model_extra` into `GatewayUsage(**fields)` lets provider-supplied keys set real billing fields. The `fields.update(...)` below pins only `prompt_tokens`, `completion_tokens`, `total_tokens`, `prompt_tokens_details` and `cache_read_tokens`; `cache_write_tokens`, `cache_write_1h_tokens` and `cache_tokens_in_prompt` are left open. On this head:

```
usage extras {cache_write_tokens: 999, cache_tokens_in_prompt: False}
  -> cache_write_tokens = 999, cache_tokens_in_prompt = False, model_extra = {}
```

Both are priced by `calculate_metered_cost`, and `cache_tokens_in_prompt=False` flips an OpenAI-shaped response onto the additive convention. Before this change those keys stayed inert in `model_extra`.

Filter the extras to non-field keys:

```python
fields = {k: v for k, v in (chunk.usage.model_extra or {}).items() if k not in GatewayUsage.model_fields}
```

`from_completion_usage` is safe today (all four are explicit there), but the same filter there would make `test_from_completion_usage_explicit_fields_win_over_same_named_extra` unnecessary rather than load-bearing.

src/gateway/core/usage.py
```diff
@@ -116,3 +122,35 @@ def cache_tokens_in_prompt_of(usage: CompletionUsage) -> bool:
...
         return usage.cache_tokens_in_prompt
     return True
+
+
+# Provider-specific field any-llm leaves in ``usage.model_extra``, and the
+# multiplier that converts its unit to milliseconds (otari#337). Absent here
+# means "not reported": most providers expose no server-side timing in the
+# response body at all.
+_PROVIDER_LATENCY_FIELDS: dict[str, tuple[str, float]] = {
+    "groq": ("total_time", 1_000.0),  # seconds
```
khaledosman, line 132: These two entries are not the same measurement. Groq's `total_time` is `prompt_time + completion_time` and excludes queue time; Ollama's `total_duration` is the whole server-side request including `load_duration`, which on a cold model dominates the compute. One column documented as "compute time" then holds both, and the same number means different things per provider.

Either use `prompt_eval_duration + eval_duration` for Ollama (any-llm forwards both), or say in the docstring and in the dashboard label that Ollama's figure includes model load.

web/src/features/activity/ActivityPage.test.tsx
```diff
@@ -33,6 +33,7 @@ function entry(overrides: Partial<UsageEntry> = {}): UsageEntry {
     error_message: null,
     status_code: null,
     latency_ms: 842,
+    provider_latency_ms: null,
```
khaledosman, line 36: Fixture only. `RequestDetail` gained a field and nothing asserts it renders; frontend-standards asks for a Vitest test for any component whose behavior changes. Add one case with a populated `provider_latency_ms` asserting "Provider time" shows the formatted value.

khaledosman (changes requested):
One blocker (branched Alembic graph once this lands on current main) and one billing-integrity issue in the streaming extras forwarding; the other two are smaller.

Verified clean on this head: `ruff check`, `mypy` (590 files), `openapi-check`, `postman-check`, and the three affected unit files (88 passed).

AmirF194:
Rebased onto main and re-pointed down_revision at c4e7a9b2d6f8, checked nothing else on main claims it as a head. Single-head check and the migration round-trip test both pass now.

For the extras leak, added GatewayUsage.external_extras to drop any model_extra key that names one of our own fields before it gets used to build the object, and switched both call sites to it. Added a test that sets cache_write_tokens, cache_write_1h_tokens and cache_tokens_in_prompt through the extras on the streaming path and checks they stay at their defaults, fails on the old code.

Switched Ollama to prompt_eval_duration + eval_duration, any-llm forwards both separately, so the column is actually excluding load time now instead of just sharing a label with Groq's number. Updated the tests that hardcoded total_duration.

Added the missing Vitest case for provider_latency_ms rendering as "Provider time".

Ran ruff, mypy --strict, the alembic head check, openapi/postman checks, and the affected unit and frontend tests in a container, all green. Didn't run the full suite or the integration tests.

Thanks for the migration head catch, that would've broken lint for anyone who rebased after us.

alembic/versions/615fa323e931_add_usage_logs_provider_latency_ms.py
AmirF194: Re-pointed down_revision at b2d4f6a8c0e2, the current true head after #1015 and #1099 both merged since this PR opened. Ran check_alembic_heads.py (single head) and the full migration round-trip suite in Docker after the merge.

src/gateway/api/routes/chat.py
AmirF194: Fixed with GatewayUsage.external_extras, which drops any model_extra key naming one of our own fields before it seeds the constructor kwargs. Used at both call sites (the streaming path here and from_completion_usage, which was already safe but now does not rely on that alone). Added a regression test that fails on the pre-fix code.

src/gateway/core/usage.py
AmirF194: Switched Ollama to prompt_eval_duration + eval_duration (any-llm forwards both) and noted in the docstring that the column excludes model load time. Updated the affected tests to match.

web/src/features/activity/ActivityPage.test.tsx
AmirF194: Added a Vitest case with a populated provider_latency_ms asserting it renders as "Provider time".

src/gateway/api/routes/_pipeline.py
```diff
@@ -3029,6 +3044,7 @@ async def log_usage(
         # Which convention those cache counts were reported under, recorded rather
         # than left to be inferred from the numbers later (mozilla-ai/otari#690).
         usage_log.cache_tokens_in_prompt = cache_tokens_in_prompt_of(usage_data)
+        usage_log.provider_latency_ms = provider_latency_ms_of(usage_data, provider_type or provider)
```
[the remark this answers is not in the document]
AmirF194: Fixed, rejecting any raw duration below zero; zero stays valid. Added a regression test covering the negative case.

src/gateway/core/usage.py
```diff
@@ -116,3 +144,46 @@ def cache_tokens_in_prompt_of(usage: CompletionUsage) -> bool:
...
+        return None
+    keys, to_ms = field
+    extras = usage.model_extra or {}
+    raw_values: list[int | float] = []
+    for key in keys:
+        raw = extras.get(key)
+        if isinstance(raw, bool) or not isinstance(raw, int | float):
+            return None
+        raw_values.append(raw)
+    scaled = sum(raw_values) * to_ms
```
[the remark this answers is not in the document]
AmirF194: Fixed, catching OverflowError around the sum*to_ms multiply. Added a regression test with an oversized int (10**400).

AmirF194:
The key_suffix migration referenced above already exists (c4e7a9b2d6f8_add_api_key_suffix.py), predates this diff, and is not something this PR touches, so there's nothing to add here.

AmirF194:
All four points from the changes-requested review got fixed the same day (09-15); CI's been green since. GitHub isn't showing that because I can't re-request review from here, so the CHANGES_REQUESTED label is stale. Worth another look when you have a minute.

Issue #1113, opened 2026-09-13
Run locally defined guardrails on the request path


dpoulopoulos:
Make the request path run a locally defined guardrail in this process, falling back to the existing HTTP call. This is the step that makes the feature work.

In `services/guardrails.py:289-345`, resolve each profile locally first:

1. The profile matches an enabled local definition (a stored row or a `guardrails:` config entry) and the entry carries no `url`: run it in-process.
2. Otherwise: the current `_validate_one` HTTP path, unchanged.

`validate_input_guardrails` gains a local-resolver argument next to `default_url`, and the pipeline passes the overlay's lookup. An entry that names its own `url` always goes over HTTP: that is a deliberate remote backend, and it is what keeps organization entries with a stored `encrypted_credential` working.

## Hybrid mode

Hybrid skips `init_db` entirely (`src/gateway/main.py:352-355`), mounts no store router and runs no refresher, so the stored half of the overlay is always empty there. The resolver must treat "no store" as "empty store", never open a session, and still see the `guardrails:` config block. That block is the only way a hybrid gateway defines an in-process guardrail. Caller and policy guardrails already run in hybrid (`_pipeline.py:2422-2431` is unconditional; only the organization layer returns `[]` at `:2329`), and that does not change.

## Unknown profile

Today a profile with no URL raises at `guardrails.py:301-305` with no `public_detail`, so its message naming `OTARI_GUARDRAILS_URL` reaches the caller. After this change the same branch means "not defined locally and no remote URL". Raise with `public_detail=_unevaluated_detail(cfg.profile)` so the response names only the profile, and update `tests/unit/test_guardrails_service.py:322`, which asserts the old message.

## Caching

The store read must not add a query per request. Use the overlay pattern from `provider_store_service.py:107-186`: startup load, 30s TTL refresher, immediate post-write refresh. Not the per-request read `resolve_organization_guardrails` uses. That one reads a *mandate*, which must take effect on the next request; this reads a *definition*, which is deployment config like a provider credential. Register the load and refresher in the `else` branch of the lifespan (`main.py:352`), beside the provider and search-tool ones.

Everything else stays as it is: the three-layer merge (`_pipeline.py:2252-2293`), the `mandated` handling, the SSRF check (`guardrails.py:266-283`), the 400/403/502 mapping (`_helpers.py:264-366`) and the `X-Otari-Guardrails` header.

## Success criteria

[...]

Issue #1111, opened 2026-09-13
Store guardrail definitions in the database


dpoulopoulos:
Add a `guardrail_credentials` table and its management API, so a guardrail is defined in Otari instead of in the sidecar's YAML. Nothing on the request path reads it yet.

The row's `name` is the `profile` a caller sends. Model it on `SearchToolCredential` (`models/entities.py:560-606`): a named backend with an encrypted secret and a free-form options blob merged over a config baseline.

## Secrets are one encrypted map, not columns

Guardrails do not share a secret shape. Counting `ParameterSpec.secret=True` create-stage parameters: `bedrock_guardrails` 3, `watsonx_guardian` 2, `lakera_guard` / `openai_moderation` / `azure_content_safety` 1, `any_llm` 0. A guardrail added upstream tomorrow may need four. So non-secret constructor arguments go in a plain JSON blob, and every secret goes in a single `{name: value}` map encrypted as one string. Splitting them by `ParameterSpec.secret` needs no per-guardrail knowledge.

## Columns

| Column | Notes |
|---|---|
| `name` | PK. The `profile` a caller sends. |
| `guardrail_name` | The `any_guardrail` class, e.g. `lakera_guard`. |
| `create_kwargs` (JSON) | Non-secret constructor args. Plain. |
| `encrypted_create_secrets` (Text, nullable) | `{name: value}`, encrypted as one string. |
| `validate_kwargs` (JSON) | Per-call args. Masked on output by `redact_secret_like_values`, as the organization table does. |
| `enabled` | Stop it without losing the config. |
| `created_at` / `updated_at` | |

## Scope

- `to_public_dict()` returns `create_kwargs` as stored and `create_secrets` as `{name: "***"}`, never plaintext.
- New `services/guardrail_store_service.py` on the `search_tool_store_service.py` / `provider_store_service.py` model: `UNSET` tri-state on PATCH, encrypt and decrypt at the service edge only, a row that will not decrypt is skipped with a warning rather than fatal. It splits submitted arguments by the catalog's `secret` flag, refuses a secret typed `json` with a 400, and treats `***` as keep-stored via `restore_redacted_values`.
- Optional `guardrails:` block in `config.yml` as the read-only baseline, stored rows winning on a name collision (`provider_store_service.apply_to_config:107-119`). It must load from `GatewayConfig` alone with no database, because it is the only definition source a hybrid gateway has.
[...]

PR #1118, opened 2026-09-14
feat(guardrails): add an in-process guardrail runner

### LINKED ISSUE #1110: Add an in-process guardrail runner
Add `services/guardrail_runner.py`: build a guardrail from `any-guardrail` and run it in this process. Nothing calls it yet.

`AnyGuardrail.create(GuardrailName(name), **create_kwargs)` builds it; `guardrail.validate(input_text, **validate_kwargs)` runs it.

## Scope

- **`validate` is a plain sync `def`** with no async variant. Offload with `asyncio.to_thread`, as `api/routes/providers.py:197` and `services/pricing_refresh_service.py:150` already do. An LLM-judge holding the event loop for seconds is the failure to avoid.
- Wrap the offload in `asyncio.wait_for` with the same 30s ceiling the HTTP path uses (`services/guardrails.py:40`). A thread cannot be cancelled, so on timeout the request fails as unavailable and the thread finishes on its own. Say so in the docstring. Where a guardrail exposes a client timeout in its create-stage spec, the store's form should surface it so the vendor call gives up first.
- Cache built guardrails by `(guardrail_name, hash of merged create_kwargs)`. A model-backed one is expensive to build. Expose `evict(profile_name)` so a store write can drop a stale entry.
- Map `GuardrailOutput` (`valid`, `explanation`, `score`, plus `categories`, `spans`, `usage`, `extra`) onto the existing `GuardrailResult` (`services/guardrails.py:71`). Upstream `valid` is a plain `bool`; the tri-state is Otari's. Keep today's meaning exactly: `valid is False` is flagged, `valid is None` is inconclusive and does **not** block.
- Every failure (bad kwargs, missing extra, vendor 4xx/5xx, timeout) raises the existing `GuardrailsNotReachableError` with `public_detail` from `_unevaluated_detail`, so the fail-open / fail-closed branch at `services/guardrails.py:313-337` governs it unchanged and the response names only the profile.
- An `ImportError` from a missing extra becomes a log line naming the extra, never the vendor module path.

## Success criteria

- Unit tests cover a passing verdict, a flagged verdict, an inconclusive verdict, a timeout, and a missing extra.
- No test imports torch.

dpoulopoulos:
## Description

Today a guardrail only runs if you start a second container. Otari posts the text to it over HTTP, and the list of guardrails that container offers is fixed by a file on disk, so adding one means editing that file and restarting.

This adds the piece that lets Otari run a guardrail itself, with no second container. Nothing uses it yet, so nothing changes for anyone running Otari today. It is the engine, landed before the wiring: the place to define a guardrail comes in #1111, and the request path starts choosing it in #1113.

Two things it has to get right, both invisible until later. A guardrail backed by a local model can take minutes to load, and loading it must not freeze the server or be thrown away when one request gives up waiting, so the load happens off to the side and the result is kept for the next request. And when a guardrail cannot run at all, the caller is told only which guardrail failed, never why, because the reason can contain the operator's API key.

## How to test it locally

Nothing is reachable from the API yet, so there is no endpoint to call. The behavior is covered by tests:

```bash
uv run pytest tests/unit/test_guardrail_runner.py -v
```

28 tests cover the verdicts a guardrail can return, a timeout, missing packages, the cache, dropping a stale entry, a slow load surviving the deadline and still being reused, and that two requests never run through one local model at the same time. None of them loads a model.

Already proven, so there is nothing to eyeball:

```bash
make lint && make typecheck && make test
```

All pass. Unit 3249 passed, integration 2282 passed, 13 skipped. No route or schema changed, so no generated file moves.

Worth a reviewer's attention, since each was a deliberate choice:

- Guardrails are called through the SDK's own dispatch helper rather than directly, because the 40 of them do not share one call signature.
- A caller's own settings still win over stored ones, matching what happens with the container today. Guardrails an organization forces on a request are unaffected, because the layer above already strips the caller's settings from those.
- Nothing a vendor library says is repeated back, only the kind of failure it was. The one exception is the message naming a missing argument, which contains argument names and never their values.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Closes #1110
Unblocks #1111 and #1113

## Checklist

Notes on the last two. No user-facing docs changed: `docs/guardrails.md` describes the container, which is still the only path that runs, and it is rewritten in #1113 when that stops being true. The comment on the dependency pin did change, because it claimed this repo never builds a guardrail, which is no longer true. The API contract did not change; `make openapi-check` and `make postman-check` confirm no drift.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

Claude Code, with Claude Opus 5.

The design was reviewed and changed before any code was written. Three points moved during that review: the stored settings no longer override a caller's, since the layer above already handles the case that rule was meant to protect; the cache-filling method was made private rather than public, because nothing calls it yet; and the fix for an unhashable field was swapped for one that keeps value equality.

**NOTE:**

## Summary

Added an in-process `GuardrailRunner` that builds and evaluates guardrails without a second container.

The runner:
- Runs synchronous operations outside the event loop.
- Applies the existing timeout.
- Shares builds and serializes local evaluations.
- Supports profile-based eviction.
- Keeps timed-out work running safely.
- Maps verdicts to `GuardrailResult`.
- Converts failures to sanitized `GuardrailsNotReachableError` errors.

Updated dependency documentation and added tests for verdicts, timeouts, caching, eviction, error handling, and concurrency. Tests do not import model backends.


dpoulopoulos:
Both review comments were worth raising. One is a real bug and is fixed here; the other I've split out.

**The gate leak is fixed.** This was the more serious of the two. A local-model guardrail is serialised so two requests never run through one model object at once, but the lock was held by an `async with` in the waiting coroutine. When a deadline passed, that coroutine was cancelled and the lock came free while the worker thread was still inside the model, so the next request walked straight in. Exactly the overlap the lock exists to stop.

The lock is now taken by hand and released from the task's done callback, so its lifetime follows the thread rather than the waiter. The abandoned caller still returns at its deadline instead of waiting for an answer nobody wants.

There's a regression test for it, and I checked it earns its keep: with the old `async with` restored it fails with `assert 2 == 1`, and passes with the fix.

**The cache bound is now #1119**, under the epic. One correction on the finding though: the entry *count* is already bounded by the deployment. A key only lands in `_built` after a profile claimed it, and `_forget` drops it the moment no profile resolves to it, so there's at most one entry per defined profile. A profile that changes its constructor arguments replaces its entry rather than adding one. I've added a test pinning that down, and a comment saying so, since it wasn't obvious from reading.

What genuinely isn't bounded is the size of those entries: a local-model guardrail holds loaded weights and nothing releases one that's gone quiet. That's the real complaint and it's worth fixing, but not here. Nothing routes to the runner yet, and the right policy depends on how operators actually define guardrails, which doesn't exist until #1111. A cap guessed now and set badly is worse than none, since evicting a model under load means reloading it, which costs minutes and more memory during the overlap. An idle timeout looks closer to the mark than a count cap, but that's a decision for when there's something to measure.

Folded into the original commit rather than added on top. `make lint`, `make typecheck`, the OpenAPI and Postman checks all pass; 27 unit tests on the runner now.

dpoulopoulos:
Found a hole in the gate fix from the last round, so here's a follow-up.

A guardrail that can't take two concurrent calls is serialised by a lock on its cache entry. The problem was that callers waiting on the same build didn't reliably end up with the same entry. Each one re-read the cache once the build finished, and if the entry had been evicted in the meantime, each wrapped the one guardrail object in an entry of its own carrying a lock of its own. Two locks serialise nothing, so two requests could sit inside the same model at once.

The build now produces the entry and its gate together, and waiters return the task's own result rather than going back to the cache. One build, one entry, one gate, shared by everyone who waited on it and by the cache if it keeps it. That also removed the fallback branch entirely, which is a nice side effect.

Reaching it needed a store write evicting a local-model guardrail mid-build with more than one request queued behind it, so it was narrow, and nothing released is affected since the runner still has no caller.

The test goes through the real eviction path rather than poking internals: two concurrent calls on a cold local guardrail, `evict` while the build is running, then assert the two evaluations never overlapped. It fails with `assert 2 == 1` against the old code.

Left as its own commit rather than folded in, since there's work stacked on this branch and rewriting the base would be unkind. Happy to squash it into the runner commit if you'd rather.

`make lint`, `make typecheck` and the architecture check pass; 28 tests on the runner, 94 across the guardrail files.

src/gateway/services/guardrail_runner.py
```diff
@@ -0,0 +1,404 @@
...
+class _Entry:
+    """One built guardrail, and the gate that decides whether calls may overlap.
+
+    ``gate`` is ``None`` when calls may run concurrently. It is held for as long
+    as the worker thread runs rather than for as long as a caller waits, so see
+    :meth:`GuardrailRunner._evaluate` rather than wrapping it in ``async with``.
+    """
+
+    guardrail: Guardrail
+    gate: asyncio.Lock | None
```
javiermtorres: I'd rather use a queue with a callback (or a timed out async request). It is safer to use, and multiple consumers can be added later on. Exposing the lock itself is usually error prone.

src/gateway/services/guardrail_runner.py
```diff
@@ -0,0 +1,404 @@
...
+    def __init__(self, *, timeout_s: float = _DEFAULT_TIMEOUT_S) -> None:
+        self._timeout_s = timeout_s
+        # No capacity limit, because the deployment already is one: a key reaches
+        # `_built` only after some profile claimed it, and `_forget` drops it as
+        # soon as none does, so the cache holds at most one entry per profile the
+        # operator has defined. What that does not bound is size, since each entry
+        # may be a loaded model, and nothing here releases one that has gone quiet.
+        # Capping resident models is otari#1119.
+        self._built: dict[_CacheKey, _Entry] = {}
+        self._inflight: dict[_CacheKey, asyncio.Task[_Entry]] = {}
```
javiermtorres: I wouldn't call something holding the guardrails by key a "cache". Guardrails are not likely to be renewed. AFAICT they're not even lazily started, and IMHO they shouldn't be, being instead started in the usual otari service initialization. Maybe "container" is a more appropriate definition.

Issue #1119, opened 2026-09-14
Release guardrail models the runner is holding but not using


dpoulopoulos:
`GuardrailRunner` (`services/guardrail_runner.py`, added in #1118) keeps every guardrail it has built for the life of the process. There is no size limit, no least-recently-used policy and no idle timeout.

## What is and is not bounded

The entry *count* is already bounded by the deployment. A key only reaches `_built` after some profile claimed it, and `_forget` drops it as soon as no profile resolves to it, so the cache holds at most one entry per profile the operator has defined. A profile that changes its constructor arguments replaces its entry rather than adding one, and `evict` removes it outright.

What is not bounded is the *size* of those entries. A local-model guardrail holds loaded weights, so an operator who defines a dozen of them keeps a dozen models resident forever, including ones that were used once and never again. Nothing here releases a guardrail that has gone quiet.

## Why this was not done in #1118

Nothing routes to the runner yet, so no deployment can reach the behavior. More to the point, a sensible policy needs to know how an operator defines and uses guardrails, and the place that defines them does not exist until #1111. Picking a cap now would be a guess, and a cap set wrongly is worse than none: evicting a model under load means reloading it on the next request, which costs minutes and more memory during the overlap.

Worth settling first: whether the right lever is a count cap, an idle timeout, or a resident-bytes budget, and whether it is configurable or a fixed default. An idle timeout looks closest to the actual complaint, since the problem is models nobody is using rather than too many models.

## Where this came from

Raised in review on #1118: https://github.com/mozilla-ai/otari/pull/1118#discussion_r4002290091

PR #1120, opened 2026-09-14
feat(guardrails): store guardrail definitions in the database

### LINKED ISSUE #1111: Store guardrail definitions in the database

dpoulopoulos:
## Description

Today a guardrail lives in a YAML file inside a second container. Adding one means editing that file on disk and restarting the container. This PR adds the place a guardrail is defined instead: a row Otari owns, written from the management API, and an optional block in `config.yml` beside it.

An operator names the guardrail, picks which one to run, and pastes the arguments it needs. The credentials in those arguments are encrypted at rest and never come back in a response. A read shows which secrets are set, masked. One endpoint runs the guardrail once against a sample input, so a definition can be checked before anything relies on it.

The name of a row is the `profile` a request already sends, so the wire contract does not change.

Nothing on the request path reads any of this yet. A request still goes to the guardrails service over HTTP exactly as before, so existing deployments behave identically. Wiring it up is #1113.

Two things are worth calling out.

Guardrails do not share a secret shape. Bedrock takes three credentials, Lakera one, the LLM-judge none, and one added upstream tomorrow may need four. So there is no column per secret. A caller sends one map of constructor arguments, and Otari splits it using the catalog from #1109: whatever that marks as a credential is encrypted into a single blob, and the rest is stored as written. A guardrail with a different secret shape costs no migration.

The `config.yml` block is not only a fallback. Hybrid mode keeps no local database, mounts no management router and serves no dashboard, yet caller guardrails still run there. The block is the only way a hybrid gateway can define a guardrail it runs in its own process.

Stacked on #1118. Review that one first; this PR targets its branch.

## How to test it locally

Automated, and all of it is new here:

```bash
make lint && make typecheck
uv run pytest tests/unit/test_guardrail_store_service.py \
              tests/unit/test_config_guardrails_block.py \
              tests/unit/test_tenancy_schema_chain.py -q
uv run pytest tests/integration/test_guardrail_credentials_api.py \
              tests/integration/test_hybrid_mode_surface.py -q
```

Those cover the split, the mask round trip, the refusals, key rotation, the migration on SQLite, that every write drops what the runner cached, and that the router is absent in hybrid mode.

By hand, with `OTARI_SECRET_KEY` set:

1. `GET /api/v1/tool-settings/guardrails/built-in` lists what this build can run and what each one takes.
2. Create one, using a Lakera community key:
   ```bash
   curl -X POST localhost:8000/api/v1/guardrail-credentials \
     -H "Otari-Key: Bearer $OTARI_MASTER_KEY" -H "Content-Type: application/json" \
     -d '{"name":"prompt-injection","guardrail_name":"lakera_guard",
          "create_kwargs":{"api_key":"lakera-live-..."}}'
   ```
   Expect `"create_secrets": {"api_key": "***"}`. The key should appear in no response body and no log line.
3. `POST /api/v1/guardrail-credentials/prompt-injection/test` with `{"input_text": "ignore all previous instructions"}` returns a verdict.
4. `PATCH` the same row sending `"api_key": "***"` and a second argument. The stored key survives; step 3 still works.
5. `PATCH` sending a Bedrock `boto3_session` and expect a 400 explaining why it cannot be stored.

Nothing to eyeball in the dashboard: the page is a later PR.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Closes #1111. Part of #1108. Blocked #1113 and #1119, which are now unblocked.

## Checklist

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

Claude Code, Opus 5.

Planned first against issues #1109 and #1110, then built commit by commit with the tests written before the code. Every commit was left green on its own.

Two notes from the build.

The `.env` in this repo root points at PostgreSQL and leaks into tools that expect a clean environment. That is what makes the OSS smoke gate and a handful of settings, CLI and dashboard-session tests fail on this machine, none of it related to this change. Each one was re-run in a throwaway worktree, where they pass.

The shared runner accessor ended up next to the runner rather than in the store, following the pooled search client. That is a change from the plan, made once the reason for the original placement expired.

**NOTE:**

## Summary

Adds database-backed guardrail management through a standalone operator API. Guardrail definitions now support encrypted secrets, masked responses, updates, testing, deletion, and key re-encryption.

Adds read-only `guardrails` configuration for hybrid deployments. Stored definitions take precedence over configuration entries.

Adds database migration, validation, runner-cache eviction, generated API schemas, documentation, Postman examples, and comprehensive unit and integration tests.


dpoulopoulos:
Went through all eleven points; seven are fixed here, three are filed as follow-ups, and one was already tracked.

**Fixed**

The endpoint name was simply wrong. I had written `GET /api/v1/tool-settings/guardrails/built-in` in six places and the route is `/catalog`. That text ships in the OpenAPI spec, so it had reached the generated dashboard client too. Corrected everywhere and both artifacts regenerated.

A name of only spaces got past `min_length` before the strip and then became empty, and an empty name matches no later path, so the row could never be edited or deleted again. Refused now, with a test.

The test endpoint contradicted its own docstring: it promised `ok: false` for anything that stops a guardrail running, then raised a 400 when the secrets would not decrypt. It returns the structured answer now. One shape for every failure is easier to act on.

A guardrail whose key was lost could not be repaired. The save path decrypted the stored secrets before looking at what was submitted, so the one request that fixes such a row, sending every secret's value, was refused by the very thing it was fixing. It only decrypts when the answer is actually needed, which is when `create_kwargs` is omitted or carries a mask. A masked patch still refuses, since keeping a secret nobody can read is not something to accept quietly.

Both docs now say the management API is standalone-only.

Two test gaps were real. The config fixture cleared three environment variables by name; it clears every `OTARI_` one now, because `load_config` layers a scalar override for any field and a named list stops covering whatever gets added later. And `test_a_row_with_no_secrets_needs_no_key` ran with the autouse key still set, so it proved nothing; it drops the key before the assertion.

**Deferred, with issues**

Nested credentials in `validate_kwargs` are a real leak, and not only here: `redact_secret_like_values` masks one level, and four columns across four tables pass through it. The harder half is `restore_redacted_values`, which has to recurse in step or a masked round trip writes three asterisks over a stored credential. Filed as #1125.

Transaction ownership and the re-encryption race are both findings about a shape all three credential stores share, not about this one. Changing only the newest would leave three stores with two shapes, and the next person copies whichever they find first. Filed together as #1127.

Bounding the runner cache is already #1119, opened off the review of #1118.

Full suite is green: 5600 passed.

PR #1123, opened 2026-09-14
fix(dashboard): open a tooltip in 300ms rather than HeroUI's 1.5s

### LINKED ISSUE #1121: Table icon tooltips take 1.5s to appear
## What happens

Hovering an icon action in any dashboard table (the disable, edit, regenerate and delete glyphs in a row's Actions lane) leaves the row unlabeled for a second and a half before the tooltip appears. Long enough that the glyphs read as having no labels at all: nobody holds a cursor still that long to find out what one does.

## Why

`Tooltip` (`web/src/design-system/overlays/Tooltip.tsx`) does not set an open delay, so HeroUI falls back to the `--tooltip-delay` custom property it reads off the document root and ships at `1.5s`, which is react-aria's warmup default. Confirmed in the built bundle and on a running gateway:

```
getComputedStyle(document.documentElement).getPropertyValue('--tooltip-delay')  // "1.5s"
```

Every tooltip in the product inherits it, but the icon lane in a table is where it is felt, because `RowAction` is the one place a control's whole name lives in the tooltip.

## Expected

The label arrives promptly on hover, around 300ms: short enough to read as an answer, long enough that sweeping a pointer across a lane of four glyphs does not flash four labels on the way past.

## Notes

Related but separate: mozilla-ai/otari-ai#2092 is the Activity chart's own hover popup, which is a recharts tooltip and not this component.

jigjigjig:
## Description

Hovering an icon action in a dashboard table left it unlabeled for a second and a half before its tooltip appeared. Long enough that the glyphs in a row's Actions lane read as having no labels at all, since nobody holds a cursor still that long to find out what one does.

The cause is a default nobody had overridden: HeroUI reads its open delay from a custom property on the page root and ships it at react-aria's 1.5s warmup value, so every tooltip in the product waited that long. Row icons are where it shows, because there the tooltip is the control's whole visible name.

Tooltips now open in 300ms: short enough to read as an answer, long enough that sweeping a pointer across a lane of four glyphs does not flash four labels on the way past. Once one tooltip in a group has been seen, the next still opens instantly, so the delay is paid once per visit to a lane rather than once per glyph.

## How to test it locally

1. Build and run the dashboard against any gateway with data, for example `make dashboard` then `otari serve`.
2. Open **API keys** and hover one of the glyphs in a row's Actions column (disable, edit, regenerate).
3. The label appears almost at once. Before this change it took 1.5 seconds; the same hover on a build of `main` is the comparison.

Automated: `pnpm --dir web run lint`, `pnpm --dir web run typecheck`, and the design-system suite (358 tests) pass. The new case in `Tooltip.test.tsx` guards the delay by reading the component source, because react-aria's hover never fires under jsdom, so no test in this repo can open a tooltip; it is the same shape as the `globals.css` sweeps in `styles/foundation.test.ts`. The 300ms itself is the part a reviewer has to eyeball.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes #1121. Part of mozilla-ai/otari-ai#2086.

## Checklist

- [x] I ran the Definition of Done checks locally. The change is dashboard-only, so the checks that cover it are the `web/` ones: `pnpm --dir web run lint`, `run typecheck`, and `pnpm --dir web test`. The Python suite is untouched by this diff and was not run.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

Claude Opus 5 through Claude Code.

The root cause was found by reading the computed `--tooltip-delay` off a running gateway rather than from the source alone, and the fix was checked the same way.

## Summary

- Set the shared `Tooltip` delay to 300ms.
- Documented the delay and its purpose.
- Added regression coverage to verify the exact delay passed to HeroUI.

This makes table action labels appear sooner while keeping enough delay to avoid flashing tooltips across adjacent icons.


web/src/design-system/overlays/Tooltip.test.tsx
```diff
@@ -45,4 +48,24 @@ describe("Tooltip", () => {
...
+   */
+  it("sets its own open delay rather than inheriting HeroUI's 1.5s", () => {
+    const source = readFileSync(
+      join(process.cwd(), "src", "design-system", "overlays", "Tooltip.tsx"),
+      "utf8",
+    )
+    expect(source).toMatch(/HeroTooltip\.Root[^>]*\bdelay=\{OPEN_DELAY_MS\}/)
+    const declared = source.match(/const OPEN_DELAY_MS = (\d+)/)
+    expect(declared).not.toBeNull()
+    expect(Number(declared?.[1])).toBeLessThanOrEqual(500)
```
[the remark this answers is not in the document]
jigjigjig: Taken, in d7824075. The test that carried it is gone too: it read the component source, which asserted the diff rather than the behavior, so it now spies on the props HeroUI is handed and takes the delay as exactly 300. `design/overlays.md` states that number, so a change to it has to change the sentence as well.

web/src/design-system/overlays/Tooltip.tsx
```diff
@@ -4,6 +4,19 @@ import type { JSX, ReactElement, ReactNode } from "react"
...
+ * How long a pointer rests on a trigger before the label appears.
+ *
+ * Left unset, HeroUI reads `--tooltip-delay` off the document root and ships it
+ * at react-aria's 1.5s warmup, which is long enough that a lane of icon actions
+ * reads as having no labels at all. A third of a second answers without
+ * flashing a label at every glyph a pointer crosses on the way past. A prop
+ * rather than an override of that variable, so the behavior travels with the
+ * component rather than with the app's stylesheet; react-aria still opens the
+ * next tooltip in a group instantly once one has been seen.
+ */
```
jigjigjig, line 17: Applied in 57214b34. The first spelling of this docstring spent three paragraphs on a one-line constant and said the prop-versus-theme choice twice. It carries what the code cannot (the 1.5s default, why a prop, the group warmup) and nothing else now.

web/src/design-system/overlays/Tooltip.test.tsx
```diff
@@ -45,4 +50,24 @@ describe("Tooltip", () => {
...
+  it("opens on its own delay rather than HeroUI's 1.5s default", () => {
+    const root = vi.spyOn(HeroTooltip, "Root")
+    render(
+      <Tooltip content="Delete">
+        <span>x</span>
+      </Tooltip>,
+    )
+    // The exact value: `design/overlays.md` states it, so moving it has to move
+    // the sentence that documents it too.
+    expect(root.mock.calls[0]?.[0]?.delay).toBe(300)
```
jigjigjig, line 71: Applied in 57214b34. `root.mockRestore()` as the last statement never runs when the assertion above it throws, leaving the spy installed for the rest of the file. It is an `afterEach` with `vi.restoreAllMocks()` now, which is what the other suites here do.

jigjigjig (commented):
Self-review pass (repo review skill, then unslop-code, unslop-design and ui-review, since the diff is under web/src). Two findings, both applied in 57214b34; nothing else survived checking. Posted by Claude Opus 5 through Claude Code.

khaledosman (approved):
Verified the guard rather than taking it on trust: ran the suite against this branch (22 passed across `Tooltip.test.tsx` and `RowAction.test.tsx`) and again with `main`'s `Tooltip.tsx` swapped back in, where the new case fails with `undefined` instead of `300`. Also confirmed the premise: `@heroui/styles` does ship `--tooltip-delay: 1.5s`, and `tooltip.js` resolves `delay ?? parseCSSTime(cssDelay)`, so the prop wins.

One thing worth recording, since the customization order in frontend-standards puts a variable ahead of a prop and this went the other way: the prop loses nothing here. `CopyButton.tsx` is the only place that reaches `@heroui/react`'s `Tooltip` directly, and it is controlled via `isOpen`, so no delay applies to it; no HeroUI component renders a tooltip internally either. This wrapper is the whole hover-opening surface, so setting `--tooltip-delay` in `globals.css` would reach exactly the same tooltips while costing the design-system layer its independence from the app stylesheet.

No findings.

---
Reviewed by Claude Opus 5 via [Claude Code](https://claude.com/claude-code).

PR #1129, opened 2026-09-14
fix(gateway): mask a credential nested inside a free-form settings dict

### LINKED ISSUE #1125: redact_secret_like_values masks only top-level keys, so a nested credential is echoed back
`redact_secret_like_values` (`src/gateway/models/secret_fields.py`) decides what to mask by walking one level of a mapping. A credential one level deeper is returned in clear:

```python
redact_secret_like_values({"headers": {"api_key": "secret"}})
# -> {"headers": {"api_key": "secret"}}
```

Four free-form JSON columns pass through it on the way to an API response:

- `provider_credentials.client_args`
- `search_tool_credentials.options`
- `organization_guardrails.validate_kwargs`
- `guardrail_credentials.validate_kwargs`

All four are operator-written and none constrains its shape, so a nested object is allowed in each. A credential put inside one is echoed back to anyone who can read that row.

## What a fix involves

Recursing through mappings and lists is the easy half. The other half is `restore_redacted_values`, which turns a resubmitted mask back into the stored value. It walks the same single level, so it would have to recurse in step, or a PATCH that echoes a masked nested object back would write three asterisks into the database where a credential used to be. That is worse than the leak, so the two must change together and be tested together.

Worth settling while doing it: whether the key-name substring rule still makes sense at depth, or whether a nested match should mask the whole subtree.

## Why this was deferred

Raised in review on #1120, which adds the fourth consumer: https://github.com/mozilla-ai/otari/pull/1120#discussion_r4003504327

The gap predates that PR and belongs to the shared helper rather than to any one table. Fixing it there changes what three existing endpoints return, which is a behavior change for the dashboard and for anyone reading those rows, and it touches the mask round trip that keeps stored credentials alive across an edit. That deserves its own change with the security-review instructions applied, not a rider on a feature PR.

L4XB:
## Description

A credential written one level down in any of the four free-form settings columns was returned to the API in clear. The masking that keeps `aws_secret_access_key` out of a `GET /api/v1/provider-credentials` response only looked at the top level of the object, so this came back untouched:

```json
{"headers": {"api_key": "secret"}}
```

Anyone who could read that row got the key. All four columns are operator-written and none constrains its shape, so a nested object is allowed in each.

Both halves of the mask now walk nested objects and lists. The second half is the reason this is one change and not two: `restore_redacted_values` turns a resubmitted `***` back into the stored value, and if it had stayed at one level while the masking went deeper, the next PATCH from the dashboard would have written `***` into the database on top of a live credential. Worse than the leak it was fixing — so the round trip is asserted per shape, not the masking alone.

Three rules that are decisions rather than mechanics:

- **A matching key masks its value whole**, dict or list included. That is what a matching *top-level* key has always done to a non-scalar, so depth 0 behaves exactly as before and the key name stays the only signal. `{"credentials": {...}}` comes back as `{"credentials": "***"}`.
- **Lists are walked, but a bare element is never masked.** An element has no key name to match on, so masking it would be a guess about its value. `restore` leans on that: a `***` element came from the caller and means itself.
- **Past a depth bound the subtree is masked rather than walked.** The alternatives were a `RecursionError` turning a read into a 500, or a depth the masking never reaches. Between "unreadable" and "leaked" this picks unreadable.

A resized list is treated as a rewrite rather than paired off by index, so a stored credential is never spliced into a position that no longer means the same thing.

The issue asked whether a nested match should mask the whole subtree — the first rule above is my answer, and it is the one that needs no new concept: it is the existing behaviour, read at depth.

## How to test it locally

```bash
uv run pytest tests/unit/test_secret_fields.py
```

23 pass. 13 are new, in two classes:

`TestNestedRedaction` — the reported leak, a credential inside a list of objects, a matching key masking its subtree, the bare-`***`-in-a-list invariant, and 40 levels of nesting masked rather than walked forever.

`TestNestedRoundTrip` — the half that makes the first half safe. A nested credential survives an edit of its visible sibling; a masked subtree is restored whole; a list round-trips; a resized list is not paired off by index; and two controls, that a **real** new nested value still replaces the stored one and that a nested entry the caller dropped stays dropped. Without those last two, "restore everything" would pass every other cell while making nested credentials uneditable.

Mutation-checked, all five caught:

| mutation | caught by |
| --- | --- |
| redaction back to one level — the reported leak | 3 cells |
| restore still walks one level — writes `***` over the credential | 2 cells |
| lists not walked on redaction | 1 cell |
| bare `***` list elements unmasked on restore | 1 cell |
| depth bound leaks instead of failing closed | 1 cell |

Already proven by automation, nothing left to eyeball beyond a reviewer's judgement on the three rules above.

## PR Type

- [x] Bug Fix

## Relevant issues

Fixes #1125. The gap was raised in review on #1120 and deferred there deliberately; this is that follow-up.

## Checklist

  - `ruff check` on both changed files — clean; `scripts/check_architecture.py` — no violations.
  - `mypy src/gateway/models/secret_fields.py` — clean.
  - `pytest tests/unit` — **3157 passed, 29 failed**, and I checked those 29 against a stashed tree: **identical failures without my change**, in `test_router_aggregate`, `test_mcp_loop_responses`, `test_deployment_bootstrap`, `test_usage_cache_tokens` and friends. Two more files (`test_inline_platform_cost.py`, `test_s3_file_store.py`) fail to *collect* in my environment on a pydantic `InputTokensDetails` field. Environmental, not this diff — but I would rather show the number than claim a green run I did not get.
- [x] Documentation was updated where necessary — the module and function docstrings carry the three rules; nothing user-facing changed.
- [x] If the API contract changed, I regenerated the OpenAPI spec — not applicable, no schema change. Response *values* change for the three existing endpoints (a nested credential is now `***`), which is the point of the fix and worth calling out for the dashboard.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

**NOTE:**

---

## Summary

- Added recursive redaction for nested objects and lists in gateway settings.
- Updated `restore_redacted_values` to preserve credentials during PATCH requests.
- Added tests for nested values, depth limits, list changes, and round trips.

This prevents nested credentials from appearing in API responses or being overwritten by redacted values.

## Technical notes

- Matching keys mask their full value.
- Resized lists are treated as replacements.
- Deep list elements remain restorable at the depth boundary.


src/gateway/models/secret_fields.py
```diff
@@ -32,13 +55,46 @@ def redact_secret_like_values(values: dict[str, Any] | None) -> dict[str, Any] |
...
-    }
+    # The walker returns whatever shape it was given, and it was given a dict.
+    # The depth bound cannot fire at the root, so this is not a `dict | str`.
+    return cast(dict[str, Any], _redact_node(values, 0))
+
+
+def _restore_node(incoming: Any, stored: Any, depth: int) -> Any:
+    """Prefer the stored value wherever the caller echoed the mask back, at any depth."""
+    if depth >= _MAX_NESTING_DEPTH:
+        return incoming
```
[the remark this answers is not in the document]
L4XB: Correct, and the mechanism is narrower than "at depth 16" — worth stating precisely because it explains why no existing cell caught it.

The depth bound is the **only** thing that masks a *bare* list element. Nothing else does: `_redact_node` masks on a key name, and an element has no key. `_restore_node` pairs a mask with its stored value by key as well, so a masked element had no way back at all.

That makes the window exactly one level wide. Measured on `bfad04ab`, varying where the list sits:

```
list at depth 14 (elements at 15): round-trip preserved = True
list at depth 15 (elements at 16): round-trip preserved = False   <-- ***  over 'live-token'
list at depth 16 (elements at 17): round-trip preserved = True
```

The last row is preserved for a different reason than the first: a list *at* the bound is masked whole as its parent's value, so the dict branch's key pairing restores it. Only a list one below the bound has its elements masked individually, and those are the ones that were lost.

Fixed in 186279e1 with your suggestion. The regression test walks all three depths rather than just the failing one, so the next person to move `_MAX_NESTING_DEPTH` sees which side of the bound each case is on:

| | |
| --- | --- |
| with the fix | `tests/unit/test_secret_fields.py` 24 passed |
| guard reverted to `return incoming` | 1 failed — `test_a_list_at_the_depth_bound_is_not_overwritten_by_its_own_mask`, 23 passed |

The dict chain was never affected, and I checked that too: a masked dict value is restored by its parent's key check at any depth, so the bound never reaches it. ruff and mypy clean.

L4XB:
Ran the full `make lint`, not just Ruff — it chains `check-architecture` and `check-migrations` ahead of the Ruff pass, and both are clean on this branch:

```
uv run python scripts/check_architecture.py
✅ No architecture violations found
uv run python scripts/check_alembic_heads.py
Single head: f1c4a8e2d6b9
uv run ruff check src tests scripts
All checks passed!
```

src/gateway/models/secret_fields.py
```diff
@@ -32,13 +55,56 @@ def redact_secret_like_values(values: dict[str, Any] | None) -> dict[str, Any] |
...
+            if value == REDACTED_VALUE and key in stored_map:
+                out[key] = stored_map[key]
+            else:
+                out[key] = _restore_node(value, stored_map.get(key), depth + 1)
+        return out
+    if isinstance(incoming, list):
+        # Positional, and only when the shapes still line up: a list the caller
+        # resized is a rewrite, and pairing it off by index would splice stored
+        # values into positions that no longer mean the same thing.
+        if isinstance(stored, list) and len(stored) == len(incoming):
```
AmirF194: Ran `secret_fields.py` from this PR's head (`74fed3ae`) in a clean container. `_restore_node`'s list branch pairs same-length lists by index, and that breaks when the two objects at a position swap: the caller reorders a list of named entries in the dashboard without touching the (still-masked) credential fields, and the restore hands each position the stored value that used to live there, not the one that belongs to the entry now sitting there.

```python
stored = {"extra_headers": [{"name": "x", "token": "live-a"}, {"name": "y", "token": "live-b"}]}
echoed = redact_secret_like_values(stored)
# user reorders the two entries, doesn't touch the masked token fields
submitted = {"extra_headers": [{"name": "y", "token": "***"}, {"name": "x", "token": "***"}]}
restore_redacted_values(submitted, stored)
# {'extra_headers': [{'name': 'y', 'token': 'live-a'}, {'name': 'x', 'token': 'live-b'}]}
# 'y' now carries 'x''s token
```

The resize guard a few lines up (`len(stored) == len(incoming)`) is there for exactly this splicing risk, but it only catches a length change, not a same-length reorder, so this path is unguarded.

Not sure what the right general fix is without knowing what identifies a list element elsewhere in the codebase, but one option: when list elements are dicts, match stored entries to incoming ones by their non-secret fields before restoring the masked ones, and fall back to `list(incoming)` (no restore) if no unambiguous match exists.

L4XB: Confirmed with your snippet at `74fed3ae`: after the reorder, `y` came back with `live-a`. Fixed in 691a627, along the lines you suggested.

A list element is now paired with the stored element it *is*: the one whose masked form it equals. That's exactly what an entry the caller didn't edit looks like, wherever it moved, so your reorder restores `y → live-b`, `x → live-a`. The same rule handles resizes better than the old length guard: appending or dropping an entry keeps the credentials of the entries left untouched, where before any length change put the mask over all of them.

Anything that can't be identified keeps the mask as submitted, never a guess:
- entries that mask to the same thing but hold different values (two `{"name": "x", "token": …}` with different tokens, or the old `[{"token": …}, {"token": …}]` case);
- an edited entry when another entry moved;
- two or more edited entries, since nothing then shows whether they also moved.

I kept one narrow exception, because "change one visible field and save" is the flow this restore exists for: an edited entry keeps its credential when it is the only unmatched entry, the only unclaimed stored entry sits at the same index, and the list kept its length. Bare elements are masked only at the depth bound and have no content to match on, so they stay positional.

`TestListElementIdentity` covers the reorder, append, drop, in-place edit, edited-and-moved, two edited, and duplicates. Six of the seven fail at `74fed3ae`; the in-place edit passes there and is kept as a guard. Each condition is mutation-checked: removing the in-place exception, its same-index check, or the equal-duplicates check fails the matching test. I renamed `test_a_resized_list_is_a_rewrite_and_not_paired_off_by_index` to `test_entries_that_mask_to_the_same_thing_are_not_guessed_between`, since resizing no longer means rewrite; the body is unchanged and still passes.

`make lint`, `make typecheck` and the unit suite pass. That includes a `mypy` arg-type error the depth-bound test already had at `74fed3ae`, now silenced like the line above it.

AmirF194: Ran `691a627` in a clean container. The swap case is fixed, and so is a single in-place edit, but editing the non-secret field on two entries at once, in place with no reorder, loses both credentials instead of one:

```
stored:   [{"name":"a","token":"secretA","label":"old-a"}, {"name":"b","token":"secretB","label":"old-b"}]
incoming: [{"name":"a","token":"***","label":"new-a"},     {"name":"b","token":"***","label":"new-b"}]
restored: [{"name":"a","token":"***","label":"new-a"},     {"name":"b","token":"***","label":"new-b"}]
```

Both tokens come back `"***"`. The old positional code (`74fed3ae`) handled this case correctly, since it never needed content matching for a same-length, non-reordered list.

The cause looks like the `len(unmatched) == 1` check in `_restore_list`: two edited entries land in `unmatched` together, so neither gets the in-place-edit fallback, and both fall through to `_restore_node(item, None, depth + 1)`, which returns their masked fields as the literal string. Generalizing that check to set equality (`same_length and unclaimed == set(unmatched)`, pairing each unmatched index with itself) looks like it would cover this without weakening the ambiguous case, but I have not verified that beyond this one repro.

L4XB: I reproduced it at `691a627`: both tokens came back as `***`. Fixed in 35fcc434.

I didn't use the set-equality version, because relabeling both entries *and* swapping them in the same save meets the same condition (unmatched and unclaimed are both `{0, 1}`). Pairing by index then crosses the tokens, which is the swap this thread started from. With your example swapped, `b` gets `secretA`.

So an edited entry is now paired by the fields the caller left alone. It goes to the stored entry it shares the most unchanged non-secret fields with, and only when each is the other's single best match. In your example `name` identifies both entries, so both tokens come back, with or without a reorder. The mask stays when nothing identifies an entry:
- a tie, e.g. a field every entry shares, like `scheme: bearer`;
- an entry whose visible fields all changed;
- duplicates.

The mutual check means one stored credential never goes to two entries.

New cells in `TestListElementIdentity`:
- your in-place edit;
- the same edit with a swap;
- a shared-field tie;
- two edited entries both closest to one stored entry.

Three of them fail at `691a627`. The tie cell passes there and is kept as a guard.

Each of these five mutations fails at least one cell:
- dropping the mutual check;
- letting a tie pick the first candidate;
- letting a zero score pair;
- counting secret fields in the score;
- index pairing on set equality.

`make lint`, `make typecheck` and `tests/unit` pass (3226 passed).

PR #1136, opened 2026-09-14
fix(dashboard): link the exhausted member dialog to Members & roles

### LINKED ISSUE #1083: [FEATURE] Give the "everyone is already in this workspace" banner a link to the organization roster
### Problem or use case

When a workspace already holds every active organization member, the add-member control is replaced by this banner:

> Every active member of this organization is already in this workspace. A workspace's members are always a subset of the organization's, so add someone there first, on the Members page.

The sentence names the next step and then makes the operator find it themselves. There is no link. "The Members page" is also the least helpful possible name for it here, because the operator is standing on a page called Members: the workspace roster is at `/members`, and the page the sentence means is the organization roster at `/organization/members`. So the copy points at a destination whose name matches the page the reader is already on.

It is worse inside the dialog. On `feat/dashboard-workspaces-form-dialog` the same banner renders inside the **New workspace member** FormDialog (screenshot below), where the only exits are Cancel and the close control, and "Add member" sits there disabled. The dialog states a dead end and offers no way through it.

(Screenshot: the dialog shows the banner as its entire body, with a greyed-out "Add member" beside "Cancel".)

### Proposed solution

Give the banner a call to action that navigates to `/organization/members`.

- Link to `/organization/members`, the organization roster, which is where a person is actually added to the organization. Not `/members`, which is the workspace roster the operator is already looking at.
- Use the router's own `Link` from `@tanstack/react-router`, not HeroUI's, so it does not full-page reload and throw away the query cache. See `.github/skills/frontend-standards/components.md`, "Internal links go through the router".
- Name the destination unambiguously. "Organization members" rather than "the Members page", so the two rosters are distinguishable from inside either one.
- **In the dialog, the link has to close the dialog.** A `Link` inside an open FormDialog navigates underneath it and leaves the operator on the organization roster with a workspace dialog still over it. Whichever shape the CTA takes, closing is part of it.
- `InfoBanner` takes `children`, so this needs no new prop: the link goes in the sentence.

One fix location serves both surfaces: `web/src/features/workspaces/WorkspaceMembersPanel.tsx:63-71`. `AddWorkspaceMember` is shared by the expanded row on the Workspaces page and by the whole of the workspace Members page, and the dialog branch renders the same component.

### Does this belong in Otari?

Yes. Dashboard copy and navigation, no provider behavior involved.

### Alternatives considered

- **Leave the prose and only rename the destination.** Fixes the ambiguity and not the dead end; the operator still has to go find the page.
- **Drop the banner and disable the whole control.** Loses the explanation, which is the part of this that is right: the subset rule is not obvious and the sentence teaches it.
- **Link to `/members`.** That is the page the banner already appears on.

### Acceptance criteria

- [ ] The banner carries a link to `/organization/members`, through the TanStack Router `Link`.
- [ ] The link's text names the organization roster, not "the Members page".
- [ ] Following it from inside the New workspace member dialog closes the dialog.
- [ ] Covered by a test in `WorkspaceMembersPanel.test.tsx` asserting the link's destination when the candidate list is empty and the roster has resolved.

jigjigjig:
## Description

When a workspace already contains every active member of its organization, the "New workspace member" dialog replaces its fields with a sentence explaining why there is nobody left to add, and telling the operator to add the person to the organization first "on the Members page". That sentence named a destination and gave no way to reach it. It was also the least helpful available name for it, because the operator is standing on a page called Members: the workspace roster is at `/members`, and the page the sentence means is the organization roster, one rail away.

The sentence now carries a link to Organization > Members & roles, so the one action that resolves the state is reachable from the words that ask for it. Following it leaves no dialog behind, because the navigation replaces the page the dialog belongs to.

Not everyone gets the link. The organization rail is opened only to a caller who manages the organization, and the route behind it has no guard of its own, so the link would be the only way a workspace admin who is merely an organization member could reach a page they can read and cannot change. They are told who can act instead: "an organization owner or admin has to add them to the organization first." An organization context that has not answered yet falls into the same sentence, which is true either way, so nothing flickers between the two.

Nothing changes in the dialog's normal state, where there are still people to add.

## How to test it locally

1. `make dashboard`, then run a gateway and open the dashboard.
2. Go to Members (the workspace roster) and press "Add member" on a workspace that already holds every active organization member. On a fresh standalone deployment that is the permanent state, since there is one identity and it owns everything.
3. The dialog's body reads "Every active member of this organization is already in this workspace. A workspace's members are always a subset of the organization's, so add someone to the organization first, on Members & roles", with the last phrase linked.
4. Follow it. You land on Organization > Members & roles with no dialog over it.
5. To see the other branch, sign in as somebody who administers a workspace but is only a member of the organization. The same dialog ends "an organization owner or admin has to add them to the organization first", with no link.

Automated coverage, all in `web/src/features/workspaces/WorkspaceMembersPage.test.tsx`:

- "sends a manager from the exhausted dialog to the organization roster" asserts the link's accessible name and `href`. This is a check: reverting the banner to its previous single sentence fails it with "Unable to find an accessible element with the role link and name Members & roles", and inverting the permission gate fails it the same way.
- "leaves no dialog over the organization roster once the link is followed" covers the acceptance criterion about closing. This is a check rather than a tautology: pointing the link at `/members` instead, so navigation happens but the route component does not change, fails it with "expected <section> to be null".
- "tells a workspace admin who is only an organization member who can act" asserts the unlinked sentence. Its companion assertion that no link is present is a guard rather than a check: no single-line regression fails that half on its own, since a mis-gated link also removes the sentence the first half looks for.

Two stories render both branches side by side in the catalog, under Features/WorkspaceMembersPanel.

Checks run locally on the pushed commit: `pnpm --dir web run lint` (579 files, clean), `pnpm --dir web run typecheck` (clean; dropping the new prop from the call site fails it with TS2741, so it is load-bearing), `pnpm --dir web test` (172 files, 5908 tests, all passing), the Storybook sweep `node .storybook/smoke.mjs` (two findings, both `layout-section--settings`, byte-identical to the same run on `origin/main`, so this branch adds none), and the Playwright behavioral suite, where `parity.tenancy.spec.ts` already opens this dialog and asserts the part of the sentence both branches still share. That spec passes, as does the whole `parity` project (43 tests).

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes #1083

## Checklist

- [ ] I ran the Definition of Done checks locally (`make lint`, `make typecheck`, `make test`). This change is confined to `web/src`, so the dashboard's own checks are the ones that apply and they are listed above with their results. The Python suites were not run, because no Python file is touched.
- [ ] Documentation was updated where necessary. No documentation describes this banner.
- [ ] If the API contract changed, I regenerated the OpenAPI spec. No route, schema or docstring changed, so neither the spec nor the Postman collection is stale.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

Three deliberate departures from the issue, all worth a reviewer's eye.

The issue asks for the link text to be "Organization members". It is "Members & roles" instead, which is what the nav registry calls that destination, so the link says the words the operator will see on the row they land on. It is also deliberately different from the existing sentence lower on the same page, which links the same route as "Organization > Members & roles": two links with the same accessible name are on screen together for a manager, and the existing test that queries the page's one by name would have become ambiguous.

The issue asks for a test in `WorkspaceMembersPanel.test.tsx`. That file does not exist; the dialog's coverage lives in `WorkspaceMembersPage.test.tsx` alongside the fixtures these cases reuse, so they were added there.

The issue treats "the link has to close the dialog" as work to do. It needs none: the dialog is mounted by the page that is the route component for `/members`, so navigating away unmounts both. That is asserted rather than assumed, by the second test listed above.

Scope left alone on purpose: the dialog's normal state gets no link, and the disabled "Add member" primary stays as it is in the exhausted state, since the dialog still has a submit that is simply not reachable.

## Summary

- Added an **Organization members** link when all organization members already belong to the workspace.
- Showed guidance for organization members who cannot manage the organization.
- Closed the add-member dialog when users navigate to the organization roster.
- Kept the normal add-member flow unchanged.
- Added tests and Storybook coverage for both permission states and dialog navigation.


web/src/features/workspaces/WorkspaceMembersPanel.tsx
```diff
@@ -63,6 +65,13 @@ export function AddWorkspaceMemberDialog({
    * told is a full workspace.
    */
   rosterResolved: boolean
+  /**
+   * Whether the caller manages the organization. Only they can act on the way
+   * out of an exhausted workspace, so only they are pointed at the page that
+   * does it; everyone else is told who can. A context that has not answered
+   * counts as false, which is the sentence that is true either way.
+   */
+  canManageOrganization: boolean
```
jigjigjig, line 74: This is the decision I am least sure of, so it is worth naming. `canManageOrganization` arrives already resolved from `canManage(context.data)`, which is false while the organization context is loading and also false when it failed. So a caller whose context query errors gets the unlinked sentence.

That matches the rest of this page, which computes the same boolean and hides its own link to the same route on the same terms. It does not match the shell, where `AppShell.tsx` deliberately fails *open* (`canManage(organization.data) || organization.isError`) so a failed read does not strand three destinations behind a hidden rail. The effect is that somebody whose context read fails can still see Members & roles on the rail, and is still told here that an owner or an admin has to act.

I took the page's side rather than the shell's: the shell is deciding whether to hide a way in, where being wrong strands a user, and this is deciding which sentence to print, where being wrong sends somebody to a page they cannot use. Happy to invert it if the preference is for one rule across both.

web/src/features/workspaces/WorkspaceMembersPanel.tsx
```diff
@@ -93,8 +102,21 @@ export function AddWorkspaceMemberDialog({
...
-          add someone there first, on the Members page.
+          A workspace's members are always a subset of the organization's, so{" "}
+          {canManageOrganization ? (
+            <>
+              add someone to the organization first, on{" "}
+              {/* Underlined, not colored alone: the link ink on this banner's
+                  muted prose is under the contrast floor, so the underline is
+                  what carries it. No `font-medium`, which resolves to 400. */}
+              <Link to="/organization/members" className="text-link underline">
+                Members &amp; roles
```
jigjigjig, line 116: The issue (#1083) asks for this to read "Organization members". It says "Members & roles" instead, for two reasons.

It is what the nav registry calls the destination (`registry.ts`, the `org-people` section), so the link says the words that will be on the row the operator lands on.

And it has to differ from the link lower on this same page, which points at the same route and reads "Organization → Members & roles". For a manager both are on screen at once, and the existing case at `WorkspaceMembersPage.test.tsx` queries that one by accessible name. Two links sharing a name would have made it ambiguous, which is a test failure that would have looked like a routing bug.

web/src/features/workspaces/WorkspaceMembersPanel.tsx
```diff
@@ -93,8 +102,21 @@ export function AddWorkspaceMemberDialog({
...
+              </Link>
+              .
+            </>
+          ) : (
+            "an organization owner or admin has to add them to the organization first."
```
jigjigjig: The sentence-ending period sits inside this string and, in the other branch, after the `</Link>`, which reads like an oversight in the diff and is not one: putting it inside the link would underline it.

The two branches also deliberately split at "so" rather than being two whole sentences. The shared half is the part that teaches the subset rule, which is the half of this banner that was already right, and keeping it literally shared is what stops the two from drifting apart in a later edit.

web/src/features/workspaces/WorkspaceMembersPanel.tsx
```diff
@@ -93,8 +102,21 @@ export function AddWorkspaceMemberDialog({
...
         <InfoBanner>
           Every active member of this organization is already in this workspace.
-          A workspace's members are always a subset of the organization's, so
-          add someone there first, on the Members page.
+          A workspace's members are always a subset of the organization's, so{" "}
+          {canManageOrganization ? (
+            <>
+              add someone to the organization first, on{" "}
+              {/* Underlined, not colored alone: the link ink on this banner's
+                  muted prose is under the contrast floor, so the underline is
```
jigjigjig, line 110: `text-link underline` rather than the `font-medium text-link hover:text-link-hover` used by the link lower on this page. The underline is not decoration: this link sits on `InfoBanner`'s muted prose, and color alone is below the contrast floor there, so the underline is what carries the affordance. `font-medium` is omitted because the token resolves to 400 (`globals.css`, the weight block), so it would be a call site trying to remember something that is no longer true.

There is precedent for both spellings in the repo, which is itself worth noting: `PolicyForm.tsx` has a bare `underline` in one place and `text-link hover:underline` in another. If somebody wants to settle that into one in-app link primitive, this is the third call site and probably the moment, but I have kept it out of this PR.

jigjigjig (commented):
Self-review. Three decisions a reviewer would reasonably stop at, and one difference from the issue that is on purpose.

web/src/features/workspaces/WorkspaceMembersPanel.tsx
```diff
@@ -93,8 +102,21 @@ export function AddWorkspaceMemberDialog({
...
+                  what carries it. */}
+              <Link to="/organization/members" className="text-link underline">
```
khaledosman: No hover state. `components.md` ("Internal links go through the router") prescribes `text-link hover:text-link-hover`, and every other link under `web/src` carries a hover of some kind, including the precedent named in the thread below (`PolicyForm.tsx:1251`, `text-link hover:underline`) and the sibling link on this same page (`WorkspaceMembersPage.tsx:150`). The underline is justified and should stay; the omission that is not is the hover.

```suggestion
              <Link
                to="/organization/members"
                className="text-link underline hover:text-link-hover"
              >
```

web/src/features/workspaces/WorkspaceMembersPanel.tsx
```diff
@@ -93,8 +102,21 @@ export function AddWorkspaceMemberDialog({
...
+                Members &amp; roles
+              </Link>
+              .
+            </>
+          ) : (
+            "an organization owner or admin has to add them to the organization first."
```
khaledosman: "add them" has no antecedent. Read whole: "Every active member of this organization is already in this workspace. A workspace's members are always a subset of the organization's, so an organization owner or admin has to add them to the organization first." The only plural named so far is the members who are already in the workspace, which is the opposite of who is meant. The other branch gets this right one line up with "add someone".

```suggestion
            "an organization owner or admin has to add someone to the organization first."
```

khaledosman (approved):
Two minor findings inline, no blockers. The contrast reasoning behind the underline holds: `--color-link` #00747e against `--color-text-muted` #2c2e33 is 2.4:1, under the 3:1 that lets color alone separate a link from the prose around it, so the underline is load-bearing rather than decorative.

Gates checked and clear: no route, schema or docstring touched, so neither the OpenAPI spec nor the Postman collection is stale; `schema.ts` and `routeTree.gen.ts` are unaffected; no new page, so no screenshot entry is owed; no dependency change; conventional title; CI green.

_Reviewed by Claude Opus 5 (1M context), via Claude Code._

web/src/features/workspaces/WorkspaceMembersPanel.tsx
khaledosman: Taking the page's side is right. The two cases are not symmetrical: the shell is deciding whether to hide a way in, where being wrong strands somebody with no route to the destination at all, while this is deciding which of two sentences to print, where being wrong sends somebody to a page they can read and cannot change. A failed context costing an owner one unhelpful-but-true sentence is the cheaper error, and the rail still carries them there because the shell fails open. Leave it.

_Reviewed by Claude Opus 5 (1M context), via Claude Code._

web/src/features/workspaces/WorkspaceMembersPanel.tsx
jigjigjig: Addressed in 1fbd716b: the link now carries `hover:text-link-hover` alongside the underline.

web/src/features/workspaces/WorkspaceMembersPanel.tsx
jigjigjig: Addressed in 1fbd716b: the unlinked branch now reads "has to add someone to the organization first", matching the other branch. Its test assertion moved with it.

PR #1141, opened 2026-09-14
feat(dashboard): group the deployment's own pages under the account menu

jigjigjig:
## Description

The sidebar's organization section is where a tenant reads about their own organization: its members, its budgets, its pricing. Two destinations sitting in it were not about the organization at all. Deployment settings configures the gateway process itself (its master key, its safety toggles, its defaults), and Accounts lists every account on the box, across organizations. Both are the deployment talking about itself, and both are refused to anyone who does not operate it.

They now sit under a Deployment heading in the menu at the foot of the sidebar, which is where the deployment's own chrome already lives. Nothing about who may reach them changes: each keeps exactly the refusal it had, and the server still authorizes every request either page makes.

The organization section's General group loses its heading along with them, because Org settings is the only row left and a heading over one row names a category with one member. Org settings keeps its name, its icon and its position. Worth knowing for anyone reviewing the organization rail rather than the menu: that change is invisible in any screenshot of the menu, and it leaves Org settings sitting under a plain divider rather than a label.

Two notes for a reviewer reading the diff.

The rows did not leave the navigation registry, and that is load-bearing rather than incidental. The registry is what answers "which entry is this pathname", so a path it does not declare is a path nothing gates. Deleting the two entries would have compiled, passed every check, and quietly changed behavior in two ways: the shell would have rendered both pages on a deployment that serves neither of their surfaces instead of refusing them, and both would have moved to the workspace context, opening the wrong sidebar and filing their visit under the wrong history key. So the entries stay exactly where they are, carrying their gating, and a new field says only that the drawing happens somewhere other than a rail row. There are tests for both halves.

The caller-axis test in the shell's own suite used to ride on Accounts, which was the only row on either rail gated on a 404 style refusal. With Accounts drawn in the menu it now rides on Providers instead, which is gated on a 403 style one. The consequence is worth stating rather than leaving to be discovered: if Providers ever loses its gate, that case stops testing the axis without failing. The 404 branch is covered in the menu's own tests, where its destination now lives.

Deliberately untouched: Providers is also operator only and also arguably misplaced, but it is a daily page on a self hosted gateway and moving it would be a product change nobody asked for.

One incidental fix. A comment in the registry claimed Accounts was the only row declaring the operator gate. That was already wrong before this change (the registry's own test has asserted three of them for some time), and it is corrected here.

## How to test it locally

Automated coverage first, since most of this is proven there:

```
pnpm --dir web run lint
pnpm --dir web run typecheck
pnpm --dir web test
```

The relevant cases are in `web/src/app/nav/AccountMenu.test.tsx` (the group renders, it is hidden whole from a non-operator, the two refusal styles behave differently when the caller's standing cannot be read, a row disappears when its surface is not hosted, the heading takes no interaction) and `web/src/app/nav/registry.test.ts` (both destinations stay registered and therefore gated, and neither is drawn on a rail). Each was verified by removing the change it covers and watching it fail.

The end-to-end walk of the management pages (`web/e2e/dashboard.spec.ts`) used to click Settings in the organization rail and now opens the account menu for Settings and for Accounts, so the path a person actually takes is covered. That one is worth knowing about: it is the check that caught the move, because it exercises the rail as a person does and the component suites do not. Verified as a real check by putting the row back on the rail and watching it time out on the menu locator.

By hand, against a standalone gateway:

1. `make dashboard && make dev`, then sign in.
2. Open the account menu at the foot of the sidebar. Under Appearance there should be a DEPLOYMENT heading with Settings and Accounts beneath it, then a rule, then Data and Privacy. Both rows should navigate.
3. Narrow the window below the `md` breakpoint and open the drawer, then the same menu. Rows take the 44px touch height; Documentation appears in the menu at that width and not above it.
4. Open the organization section of the sidebar. Org settings is still there, with no General heading above it, and Settings and Accounts are gone from it.
5. Visit `/#/settings` and `/#/admin/accounts` directly. Both still open, still open the organization sidebar, and on a gateway that does not serve those surfaces both still answer with the shell's "not available here" panel rather than the page.

Measured on the built component at 1280 and 375: menu 274x324 and 274x422, heading 24px tall at offset 100 and 116, rows 36px and 44px.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [x] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

None. This came from a design review of the settings hierarchy rather than a filed issue, so there is no issue to close.

## Checklist

Two of these need a word rather than a tick.

Documentation: nothing in `docs/` or in the scoped guidance describes where these two destinations are drawn, so there was nothing to bring into step. The navigation section of `web/AGENTS.md` describes the three gating axes, which are unchanged.

OpenAPI: no route, schema or docstring changed. This is dashboard only, so the spec, the Postman collection and `web/src/client/schema.ts` are all untouched, and no route was added so `web/src/routeTree.gen.ts` is too.

On the Definition of Done box: the change is entirely under `web/`, which `make lint` does not cover, so the checks that actually exercise it are the three dashboard commands above, all green on the pushed commit. `make lint` was also run and passes. `make typecheck` and `make test` are the Python suites and no Python changed.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

Two things are worth flagging to a reviewer because they were decided against the obvious reading rather than by following it.

The first is why the registry entries survive. The natural implementation of "move these two rows" is to delete them from the registry and write two rows into the menu by hand. That is the silent regression described above, and it was found by deleting them and reading which tests went red rather than by reasoning about it. One of the five failures came from a test that predates this change, which is worth knowing: the risk of a destination falling through to the wrong sidebar was already guarded.

The second is the heading's type. The design initially specified 11px mono at 0.1em tracking, which the shell has no example of anywhere: the single token for this role resolves to 12px sans at 0.06em, with a comment recording that it deliberately moved from 11 to 12 to stop being a one-off. The token was taken, the design has since been corrected to name the token rather than restate figures that can drift from it, and the rail's own headings were confirmed on the running dashboard as 12px sans 550 at 0.72px tracking. What did come from the drawing is the geometry around the label, which is where the menu's scale genuinely differs from the rail's: a 24px box against the rail's 32, and a 10px leading pad against 12.

One thing about that heading is worth writing down, because it is about what keeps it working rather than about this change. At 12px against 13px rows, size does no work in separating the heading from the rows it names. The boundary is carried entirely by uppercase against sentence case, `--color-text-subtle` against `--color-foreground`, and weight 550 against 500. The menu's heading also sits in a 24px box with a 6px gap where the rail's has a 32px box and section gaps around it, so it leans on those three signals harder than the rail's does. Nothing to change here, but if anyone later lightens `--color-text-subtle`, drops the uppercase, or moves that token toward the row weight, this heading stops reading as a grouping noticeably before the rail's does. Contrast is fine as it stands: #8A8F98 on #0F1011 is 5.87:1.

A smaller one, in case it saves somebody the detour: the group wants `role="group"`, and Biome's `useSemanticElements` rejects that and suggests `<fieldset>`, which is for form controls and wrong for a list of links. A named `<section>` is the shell's existing answer and is what this uses.


web/src/app/nav/types.ts
```diff
@@ -91,6 +91,20 @@ interface NavItemBase {
...
+   *
+   * Not a gate, and deliberately not expressed by leaving the entry out of the
+   * registry. `NAV_ITEMS` is what `navItemForPath` answers from, so an
+   * unregistered path is an *ungated* one: dropping these two entries would
+   * stop the shell refusing them on a deployment that does not host their
+   * surface, and would move them to the workspace context, since `ORG_PATHS` is
+   * flattened from the organization sections. The entry therefore stays where
+   * it is, carrying its `surface` and `operatorOnly`, and only the drawing
+   * moves. `visibleNavSections` is the one place a rail row is produced, so
+   * that is the one place this is read.
```
jigjigjig, line 105: A single-member union rather than a boolean, and the choice is worth defending because `railHidden?: boolean` would have been shorter.

A boolean says where the row is *not* drawn, which is the half that goes stale: the day a second surface wants entries (a command palette, say) a boolean has to be read as "not the rail, so presumably the menu", and nothing in the type says which. The union names the destination, so a second value is an addition rather than a reinterpretation, and `ACCOUNT_MENU_ITEMS` filters on the value rather than on truthiness.

It also keeps the field readable as a fact about the item, which is the registry's whole idiom: every other field here says something the entry *is*, not something a consumer should skip.

web/src/app/nav/registry.ts
```diff
@@ -740,6 +743,28 @@ export function visibleNavSections(
   isVisible: (item: NavItem) => boolean,
 ): VisibleNavSection[] {
   return sections
-    .map((section) => ({ section, items: section.items.filter(isVisible) }))
+    .map((section) => ({
+      // `rendersIn` before `isVisible`, because the two answer different
+      // questions and only this one is about the rail: a row drawn in the
+      // account menu is still a destination this caller may have, and
+      // `isVisible` is the predicate that says so for the menu as well.
```
jigjigjig, line 750: The ordering here is deliberate and will read as arbitrary in a diff, so: `rendersIn` is checked before `isVisible`, not after, because the two answer different questions and only this one is about the rail.

`isVisible` answers "may this caller have this destination on this deployment", and the account menu asks it too. If the rail's skip were folded into that predicate, the menu could not reuse it and would need a second one, which is the drift this change exists to avoid. Keeping them separate is what makes "the menu and the rail cannot disagree" true by construction rather than by care.

Please do not collapse them into one predicate later without reading the menu's call site first.

web/src/app/nav/registry.ts
```diff
@@ -740,6 +743,28 @@ export function visibleNavSections(
...
+
+/**
+ * The destinations the account menu draws, in registry order.
+ *
+ * Derived rather than listed a second time: the menu's rows and the entries
+ * that gate those routes are then the same objects, so a row cannot end up
+ * offering a page whose surface this deployment does not serve. Ungated here on
+ * purpose; the menu runs them through `useNavVisibility` exactly as the rail
+ * does its own.
+ */
```
jigjigjig, line 767: Derived at module scope from `NAV_ITEMS` rather than listed, so the menu's rows and the entries that gate those routes are literally the same objects. A hand-written list here would compile forever while silently offering a page whose surface the deployment stopped serving.

It is deliberately ungated: the menu runs these through `useNavVisibility` itself. Gating here as well would put the answer in two places, and the one that is a module constant cannot see a query.

web/src/app/nav/registry.ts
```diff
@@ -408,36 +408,39 @@ const ORGANIZATION_NAV_SECTIONS = [
   },
   {
     id: "org-general",
-    label: "General",
+    // No heading. The two deployment rows below are drawn in the account menu
+    // rather than on this rail, which leaves Org settings alone here, and a
+    // heading over one row names a category with one member.
```
jigjigjig, line 413: The section keeps its `id` and loses only its `label`. That is not cosmetic: overlay contributions and label overrides key on the id, so renaming or removing it would silently drop whatever a superset build files under this section. Losing the label alone turns it into a divider-separated group, which is what the index section at the top of the workspace rail already does for the same reason.

The comment above the Accounts entry, claiming it was the only row declaring `operatorOnly`, was wrong before this change and not because of it: `registry.test.ts` has asserted three rows for some time. Fixed here since I was in the file, and called out in the PR body so it is not mistaken for part of the move.

jigjigjig (commented):
Self-review. Six notes on decisions a reader would reasonably stop at, rather than on mechanics. Two of them are places I would welcome being overruled.

web/src/app/nav/AccountMenu.tsx
```diff
@@ -342,6 +360,44 @@ export function AccountMenu({ collapsed }: { collapsed: boolean }) {
...
+              Rows and heading appear together or not at all. Each row is gated
+              on the value it declares (`/settings` 403s, `/admin/accounts`
+              404s), so a caller can be refused both, and a label over nothing
+              names a group that is not there. */}
+          {deployment.length > 0 ? (
+            // Labeled rather than left as three loose rows: the heading is what
+            // tells a sighted reader these two belong together, and a screen
+            // reader gets the same fact from the group's name instead of
+            // meeting Settings and Accounts as peers of Log out.
+            <section
```
jigjigjig, line 383: This wants `role="group"`, which is exactly the semantic: a named set of related items. Biome's `useSemanticElements` rejects it and suggests `<fieldset>`, which is for form controls and wrong for two links.

A named `<section>` is lint-clean, announces the same way, and is what `AppShell` already uses to name a group of rail rows, so it is the shell's own answer rather than a workaround I invented. Recording the detour because the next person to reach for `role="group"` in this repo will hit the same rule and may reasonably conclude a suppression is the only way out.

web/src/app/AppShell.test.tsx
```diff
@@ -733,15 +735,17 @@ describe("AppShell entitlement gating", () => {
...
+    // declares the axis at all: the two that declared it here are drawn in the
+    // account menu now. So this covers `"refused"`, and `"unlisted"` is covered
+    // where its one destination is drawn (`nav/AccountMenu.test.tsx`).
+    await renderShell(bootstrap(), { operator: true })
 
     expect(
       await within(
         screen.getByRole("navigation", { name: "Sidebar" }),
-      ).findByRole("link", { name: "Accounts" }),
+      ).findByRole("link", { name: "Providers" }),
```
jigjigjig, line 748: This is the weakest thing in the change and I would rather point at it than let it be found.

This case used to ride on Accounts, the only row on either rail gated `operatorOnly: "unlisted"`. With Accounts drawn in the menu, the only rail row left declaring the axis at all is Providers, which is `"refused"`. So it now proves the 403 branch and no longer touches the 404 one, and if Providers ever loses its gate this case stops testing the axis without failing.

The 404 branch is covered in `nav/AccountMenu.test.tsx`, where its destination now lives, and that case fails specifically when the two values are flattened into one. I verified this retarget is still a check by forcing `isDeploymentOperator` to return false and watching it fail on the link name. If a reviewer would rather see a second rail row declaring the axis, or this case moved out of the shell suite entirely, I will take either.

jigjigjig (commented):
Reposting two self-review notes that anchored a few lines off their subject.

jigjigjig:
Closing: this work moved onto #1142 as commits a88933a3, 680c74ab and d8b1aa46, so it ships as one PR rather than three. Branch left in place.

Issue #1126, opened 2026-09-14
fix: use consistent provider display names in the add-provider picker


jigjigjig:
## Problem

On `/providers`, open **Add provider** and expand the **Provider** dropdown. Some labels use brand capitalization while others appear as lowercase provider IDs.

`list_known_provider_summaries()` in `src/gateway/services/provider_metadata_service.py` enumerates the any-llm registry and looks up each provider by exact ID in the bundled genai-prices dataset. It uses the dataset's display name when available and otherwise falls back to the raw ID. Missing records or differing IDs therefore produce inconsistent labels.

## Proposed solution

Add a shared display-name resolver in Otari's provider metadata service with this precedence:

1. A small explicit Otari override map, keyed by provider ID and using official brand spellings.
2. The matching genai-prices display name.
3. The raw provider ID for unknown or newly added providers.

Populate overrides for current gaps and ID mismatches. Use the resolver for catalog summaries, selected-provider details, and configured-provider metadata so these surfaces agree. Keep catalog listing lightweight without importing provider SDKs.

Do not automatically title-case IDs: `openai` would become `Openai`. Provider IDs, saved configuration, credential resolution, and dispatch behavior must remain unchanged.

## Acceptance criteria

- Providers currently falling back to lowercase IDs have explicit, correctly spelled display names where needed.
- The picker and provider metadata use the same name-resolution rules.
- Unknown providers still receive a nonempty label via the raw-ID fallback.
- Focused unit tests cover override precedence, dataset names, and unknown-provider fallback.

## Related work

Related to #990, which covers dashboard-wide names and logos through a frontend presentation map. Coordinate the naming source with that work to avoid competing maps. This issue captures the confirmed catalog fallback behavior and proposed backend resolver; logos are outside its scope.

Issue #990, opened 2026-09-09
Show providers with their proper brand name and logo in the dashboard


jigjigjig:
## Context

Improvement request from Slack ([thread](https://mzlai.slack.com/archives/C0BNZGFQ7TJ/p1788954103699869)):

> Please make sure providers are listed with their proper name and logo. Mistral will murder us if we show them as 'mistral' with no logo.

The dashboard currently renders provider identifiers verbatim (`mistral`, `openai`, `anthropic`, ...) with no icon. Those strings are wire identifiers, not brand names, so the UI reads as unbranded and misspells vendors' own capitalization.

## What to do

- Add a single provider presentation map in `web/src` (display name plus logo) keyed by provider id, and use it everywhere a provider is shown to a user: the providers page, the models page and its model pickers, activity/usage rows, routing and fallback surfaces, and any filter combo boxes listing providers.
- Use each vendor's own capitalization (`Mistral AI`, `OpenAI`, `Anthropic`, `Google`, ...); do not title-case the id generically, since that produces wrong names (`Openai`).
- Ship logos as local assets rather than remote URLs, so the dashboard keeps working offline and behind the CSP.
- Fall back gracefully for an unknown or user-configured provider id: no broken image, and a readable label instead of a blank.
- Keep the underlying id visible wherever a user needs it to configure something (config keys, API payloads); this is a display-layer change only.

## Notes

- Provider logo assets carry vendor trademark terms of use; check each vendor's brand guidelines before adding a mark.
- Follow `.github/skills/frontend-standards/SKILL.md` for component and token conventions.

PR #1142, opened 2026-09-14
feat(dashboard): show both ends of an API key, plus fixes from a release-readiness pass

### LINKED ISSUE #957: CopyField's default arrangement cannot copy on a non-secure origin
Raised in review on #956 and deliberately left out of that PR, because it changes behavior on six surfaces the change in hand did not touch.

## What

`CopyField.copy()` (`web/src/shared/components/actions/CopyField.tsx:124`) tries `navigator.clipboard?.writeText` and nothing else:

```ts
if (navigator.clipboard?.writeText) {
  await navigator.clipboard.writeText(value)
  ...
}
// No Clipboard API (or it threw): the text is selected, so the operator can
// press Ctrl/Cmd-C. Never claim it was copied.
setSelectHint(true)
```

The async Clipboard API is gated on a secure context, and this dashboard is routinely served from a plain-HTTP LAN address, where `navigator.clipboard` is undefined. So on those origins the default arrangement never copies. It selects the value and asks the operator to press Ctrl/Cmd-C, which is honest but is not a copy.

`copyToClipboard` (`web/src/shared/helpers/clipboard.ts`) already handles this: it tries the same API and then falls back to an offscreen textarea plus `document.execCommand("copy")`, which is the only clipboard write such an origin has. #956 moved the organization email-domain field onto `CopyButton`, which goes through that helper, so that one field now copies on a non-secure origin.

## Where it still bites

Six shared call sites keep the default arrangement and the degraded path:

- `web/src/features/keys/KeysPage.tsx:200`, the one-time secret key reveal
- `web/src/features/keys/KeysPage.tsx:215` and `:216`, the curl and Python snippets
- `web/src/features/onboarding/SetupGuideCard.tsx:244`, the issued API key
- `web/src/features/onboarding/SetupGuideCard.tsx:262` and `:267`, the same two snippets

The one-time secret reveal is the one worth caring about: it is shown once and cannot be retrieved afterwards.

(The review comment said five; it is six. The master key field on Settings is a separate local component, `SettingsPage.tsx:281`, and is not affected.)

## Suggested fix

Replace the hand-rolled attempt in `copy()` with `const copied = await copyToClipboard(value)`, keeping the existing "never claim a copy it did not make" behavior for the case where both paths fail. The markup does not change, so the six call sites keep rendering identically.

## Why not in #956

#956 changed one of the seven call sites and proved the other six render byte-identically. Routing all six through a different clipboard path is a behavior change on surfaces that PR was not about, including the one-time key reveal, so it wants its own change and its own test.

## Notes for whoever picks this up

- `legacyCopy` restores the selection and focus it found on its way out (`clipboard.ts:55-59`), so any selection made *before* the attempt is undone by the fallback. `copy()` currently selects first. Whatever replaces it has to select after the attempt resolves, in the failure branch. #956 has the same ordering handled in `CopyButton` and a test that asserts the sequence rather than the end state.
- jsdom defines no `document.execCommand`, so the fallback fails there on its own. A test that asserts only the end state passes whether the selection happens before or after the attempt; assert the order.

jigjigjig:
## Description

Small fixes from a release-readiness pass, plus two changes that needed new code.

- Copying an API key works on a plain-HTTP address; it silently failed on one of three paths.
- The bulk-action bar centers on the content instead of the window. Also fixes Activity and Budgets, which render the same bar.
- A key row's action icons stop shifting column when a row is disabled.
- Routing policy rows line up, and a section can be removed in one press instead of vanishing when you delete its last row.
- A guardrail entry's controls share one width instead of six.
- Providers show their proper names, not the wire spelling.
- Gateway API keys now store and return the last four characters of the secret, so a key can be told apart from another that shares its prefix.
- Deployment settings and Accounts move out of the organization rail into a Deployment group in the account menu.

Does not do: the guardrail entry still ignores the settings lane; the per-entry Save stays; the keys table's NAME column is untouched; `/providers` is deliberately left where it is; provider logos are out, so mozilla-ai/otari-ai#2098 cannot close on this.

## How to test it locally

`make dashboard`, then run a gateway with a few keys and a provider credential.

- **Keys**: create a key and press Copy, over a LAN address rather than `localhost`. It should copy rather than only select. Tick a row and check the floating bar centers over the content, with the sidebar expanded and collapsed. With one active and one disabled key, Edit should be in the same column on both.
- **Routing**: Create policy, then add a tier-down, a fallback chain and a routed pool. Each section has a trash control that removes the whole section in one press. With a failing provider credential, the "Could not list models" sentence sits under a row rather than inside the picker.
- **Guardrails**: a mandated guardrail's selects and credential box share one width on one line. The same page serves `/tools`, `/tools/web-search` and `/tools/code-execution`, so check those render too.
- **Models**: the provider column and its filter read "OpenAI", not "openai".
- **Account menu**: Settings and Accounts are under a Deployment group there, not in the organization rail.

Automated: `pnpm --dir web run lint`, `run typecheck`, `test` (6290), `make lint`, `make typecheck`, `make test` (5559), `make openapi-check`, `make postman-check`, and the Playwright onboarding project (12/12).

The bulk bar's centering was measured on a running dashboard at four widths (1000, 1024, 1280, 1600), zero offset against the content pane at each and inside the pane at the narrowest. The collapsed rail and the drawer below `md` rest on a single measurement each.

Two notes on what the automation does and does not cover. The bulk bar's offset is CSS, so no Vitest case can see it: `AppShell.test.tsx` guards the wiring (the shell publishes `data-rail`, and dropping it silently re-centers the bar on the window), and the new Playwright case guards the geometry. Both were confirmed by removing the fix and watching them fail. And one dashboard test failed once during development, on a tree that then passed twice, and was never identified; it has not recurred in any run since.

## PR Type

- [x] New Feature
- [x] Bug Fix
- [ ] Refactor
- [x] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes #957
Refs #990, #1126, mozilla-ai/otari-ai#2098, mozilla-ai/otari-ai#2102, mozilla-ai/otari-ai#2104, mozilla-ai/otari-ai#2117

## Checklist

The dashboard's own three checks are what gate most of this; `make lint`, `make typecheck` and `make test` do not reach `web/` and were run as well. `docs/public/openapi.json` was regenerated for the new field, and `make postman-check` reports the collection already current.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

Every number claimed here was measured on a running dashboard serving this branch's own build, and the before values the same way.

**NOTE:**

## Summary

- Improved dashboard layout and accessibility for bulk actions, key rows, routing policies, guardrail controls, and model catalog hints.
- Added clipboard fallback support for non-secure origins.
- Standardized provider names across dashboard views.
- Stored and returned the final four characters of API keys, with migration and test coverage.
- Moved Settings and Accounts into a dedicated Deployment navigation group.
- Added regression, integration, unit, Storybook, and end-to-end coverage.
- Regenerated the OpenAPI client schema.

## Technical notes

- Deployment navigation now supports workspace, organization, and deployment rails.
- API key suffixes remain nullable for legacy records.
- Existing guardrail behavior, key table naming, provider routes, and provider logos remain unchanged.


web/src/app/AppShell.test.tsx
```diff
@@ -733,15 +735,17 @@ describe("AppShell entitlement gating", () => {
...
+    // declares the axis at all: the two that declared it here are drawn in the
+    // account menu now. So this covers `"refused"`, and `"unlisted"` is covered
+    // where its one destination is drawn (`nav/AccountMenu.test.tsx`).
+    await renderShell(bootstrap(), { operator: true })
 
     expect(
       await within(
         screen.getByRole("navigation", { name: "Sidebar" }),
-      ).findByRole("link", { name: "Accounts" }),
+      ).findByRole("link", { name: "Providers" }),
```
jigjigjig, line 756: This is the weakest thing in the change and I would rather point at it than let it be found.

This case used to ride on Accounts, the only row on either rail gated `operatorOnly: "unlisted"`. With Accounts drawn in the menu, the only rail row left declaring the axis at all is Providers, which is `"refused"`. So it now proves the 403 branch and no longer touches the 404 one, and if Providers ever loses its gate this case stops testing the axis without failing.

The 404 branch is covered in `nav/AccountMenu.test.tsx`, where its destination now lives, and that case fails specifically when the two values are flattened into one. I verified this retarget is still a check by forcing `isDeploymentOperator` to return false and watching it fail on the link name. If a reviewer would rather see a second rail row declaring the axis, or this case moved out of the shell suite entirely, I will take either.

jigjigjig (commented):
One self-review note, carried over from the PR this work was folded in from. It is the only one of six worth re-posting: the others explain decisions a reader can reach from the code and its comments, and this one is a caveat a reviewer should get the chance to push back on.

jigjigjig:
@coderabbitai review

web/src/design-system/actions/CopyField.tsx
```diff
@@ -268,19 +269,24 @@ export function CopyField({
       setSelectHintFor(copying)
       return
     }
+    // The same helper the concealed path above uses, rather than
+    // `navigator.clipboard` alone: the async Clipboard API is gated on a secure
+    // context and this dashboard is routinely served from a plain-HTTP LAN
+    // address, where it is undefined. The helper falls back to an offscreen
+    // textarea and `execCommand`, which is the only clipboard write such an
+    // origin has (otari#957).
+    if (await copyToClipboard(value)) {
```
jigjigjig, line 278: This is the change I would want a second opinion on, and the issue is not in this file.

Routing the visible-value path through `copyToClipboard` is right, and the ordering comment below it is careful about exactly the thing that made #957 tricky. But it means `legacyCopy` in `design-system/helpers/clipboard.ts` is now the fallback for **every** copy in this component, where before it was reached only by the concealed path. That function appends a textarea holding the plaintext to `document.body`, and its cleanup is not in a `finally`:

```js
  source.select()          // not guarded
  let copied = false
  try { copied = document.execCommand("copy") } catch { copied = false }
  source.remove()          // skipped if select() threw
```

`execCommand` is guarded; `select()` is not. If it ever throws, the node is never removed and a DOM element containing the credential persists for the life of the page. That is the thing #957's concealment exists to prevent, reached by the path added to serve #957.

I am **not** asking for it to be fixed here. `clipboard.ts` is untouched by this PR and pre-dates it, and folding an unrelated fix into a PR that has just been cut down is the wrong trade. My suggestion is a separate issue for moving that cleanup into a `finally`, linked from here so the reasoning is not lost. Raising it on this line because this change is what widens its blast radius.

Worth saying plainly that I could not make `select()` throw, so this is defense in depth on a secret-handling path rather than a reproduced bug.

web/src/shared/helpers/providers.ts
```diff
@@ -0,0 +1,83 @@
...
+ *
+ * Generic title-casing is not an option and is why this is a map: `openai`
+ * title-cases to "Openai", which is wrong in a way that reads as a bug. An id
+ * with no entry is returned unchanged rather than guessed at, so a provider
+ * added upstream tomorrow is plain rather than misspelled.
+ *
+ * Keys are any-llm's own provider ids, read off `AnyLLM.get_supported_providers()`
+ * rather than recalled, because a key that is not a real id is a row that never
+ * gets its name and nothing fails. Ids with no widely-used brand spelling of
+ * their own are deliberately absent and fall through unchanged.
```
jigjigjig: A question about the stated rule, not a defect.

I compared the map against `AnyLLM.get_supported_providers()`: all 47 keys are real ids, no misses. Six ids have no entry and fall through unchanged, which this line says is deliberate for ids with "no widely-used brand spelling of their own".

Four of the six are fair by that rule. The other two are `mzai` and `otari`, which do have brand spellings, and one of them is this product. So on a deployment using either, the dashboard renders its own provider lowercase as `otari` and `mzai` while showing every third party under its proper name, which reads as the bug this file exists to fix rather than as a deliberate omission.

Three defensible answers and I do not think it is mine to pick: add both, add neither and narrow this sentence to say which ids are out and why, or leave it and accept that the rule as written does not cover these two. Whichever it is, the sentence and the map should agree, because right now the sentence explains an absence it does not actually explain.

jigjigjig (commented):
Reviewed the clipboard path, the bulk bar, the keys action lane, the routing controls, the guardrail widths and the provider name map, as a reviewer rather than as a colleague. **Two threads below, one of them a question rather than a defect.** Nothing else surfaced, and I am not going to invent a third to show the review happened.

Two things I checked rather than read, both of which came back clean and are worth recording because neither had a detector.

**The provider map's keys are all real.** The file says they are any-llm's own ids, which nothing enforces: a wrong key is a row that stays lowercase forever and nothing fails. I loaded `any_llm` and compared. All 47 keys are in `AnyLLM.get_supported_providers()`, zero misses. Six ids have no entry, which is the intended fall-through, and two of those are the subject of the question below.

**The bar's centring reproduces, at viewports I picked rather than the one it was tuned at.** Measured on a running gateway serving this branch, Keys page, rows selected, rail expanded, against the pane rather than the table:

| viewport | bar centre | pane centre | offset |
|---|---|---|---|
| 1000 | 632 | 632 | **0** |
| 1280 | 772 | 772 | **0** |
| 1600 | 932 | 932 | **0** |

The bar also stays inside the pane at every one: at 1000 it is 704px wide starting at x=280 against a pane starting at 264, so it never reaches back over the rail, which is what the width half of the fix is for.

One correction to my own method, in case anyone repeats it: my first pass measured the table's box and reported offsets of -158 and -18. That was wrong. The table overflows its scroll container at narrow widths, so its box is wider than the viewport and its centre is not the centre of what a reader sees. The pane is the right reference and the offsets are zero.

**Not measured:** the collapsed-rail branch and below `md`. I did not exercise the collapse toggle, so `--rail-width: 4.5rem` is unconfirmed by me and rests on the author's own measurement.

web/src/design-system/actions/CopyField.test.tsx
```diff
@@ -97,6 +97,44 @@ describe("CopyField", () => {
...
+      await user.click(screen.getByRole("button", { name: "Copy" }))
+
+      // The claim is only made because something wrote: the offscreen textarea
+      // is the only clipboard write such an origin has.
+      expect(execCommand).toHaveBeenCalledWith("copy")
+      expect(
+        await screen.findByText("Copied to clipboard."),
+      ).toBeInTheDocument()
+    } finally {
+      if (clipboard) Object.defineProperty(navigator, "clipboard", clipboard)
```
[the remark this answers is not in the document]
jigjigjig: Fixed in 087c344c. The cleanup is symmetric with the `execCommand` one beside it now, so an inherited or absent `navigator.clipboard` is deleted rather than left as an own `undefined`.

Worth saying the failure mode you named is the one that would have hurt: it would not have failed this case, it would have silently removed the Clipboard API from every case after it in the file, on a runtime where the property is inherited. Same shape as a teardown that works on the machine it was written on.

web/src/design-system/actions/CopyField.tsx
```diff
@@ -270,17 +271,18 @@ export function CopyField({
     }
     ref.current?.focus()
     ref.current?.select()
```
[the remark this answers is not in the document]
jigjigjig: Fixed in 087c344c, and the reason it mattered turned out to be stronger than the one stated here.

Your argument was that a successful copy should not move the operator selection, which is true. The sharper problem is that this PR routed this path through `copyToClipboard`, and `legacyCopy` restores whatever selection and focus it found on its way out. So a selection made before the attempt is undone by the fallback, and on exactly the origin where the copy fails, the value would not have been selected and the "press Ctrl/Cmd-C" hint pointed at nothing. The fix for #957 was defeating the behavior #957 exists to protect.

The order now matches the `action` arrangement above, and a new case asserts the sequence for this arrangement too rather than the end state, which cannot tell the two orders apart in jsdom.

web/src/shared/helpers/providers.ts
```diff
@@ -0,0 +1,62 @@
...
+  mistral: "Mistral AI",
+  moonshot: "Moonshot AI",
+  nebius: "Nebius",
+  ollama: "Ollama",
+  openai: "OpenAI",
+  openrouter: "OpenRouter",
+  perplexity: "Perplexity",
+  portkey: "Portkey",
+  sambanova: "SambaNova",
+  togetherai: "Together AI",
```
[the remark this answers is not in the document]
jigjigjig: Fixed in 087c344c, and it was worse than one missing id.

`together` is right: that is the any-llm provider id and `togetherai` is not. Checking it sent me to the source, and `ai21` and `google` were not ids either. The map had been written from memory. It is rebuilt from `AnyLLM.get_supported_providers()` and reconciled both ways, so no key is a non-id and the absences are deliberate.

The failure mode is why this deserved the trip: a wrong key fails nothing at all, it just leaves that vendor lowercase forever.

web/src/styles/globals.css
```diff
@@ -3180,8 +3180,27 @@ main {
...
+  [data-rail="collapsed"] {
+    --rail-width: 4.5rem;
+  }
+}
+
 .otari-bulk-bar {
   animation: otari-bulk-bar-in 160ms ease-out;
+  /* Centred on the content, not the viewport: `left-1/2` alone put the bar
+     132px left of the table it belongs to and over the last visible rows. */
+  left: calc(50% + var(--rail-width, 0px) / 2);
```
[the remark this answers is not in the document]
jigjigjig: Fixed in 087c344c. `width` now subtracts the rail as `left` does.

Confirmed at the width you named rather than taken on the argument: at 1024 with the rail expanded the bar was 984px in a 760px pane before, and sits inside the pane after. A Playwright case in `e2e/dashboard.spec.ts` now holds both properties, and it was confirmed by deleting `data-rail`, rebuilding and watching the centering assertion report 132px.

web/src/shared/helpers/providers.ts
jigjigjig: Taken, and you are right that the stated rule does not cover these two. Fixed in db25f1d9: `mzai` is "Mozilla AI" and `otari` is "Otari".

The argument that settles it is yours: a deployment serving its own models rendered them lowercase while every third party got proper casing, which is this map's own bug pointed at us. Four ids are still absent (`cascadia`, `gmi`, `kenari`, `neosantara`) and the docstring now says four rather than leaving the count implicit, so the sentence is checkable against the map instead of being a claim about it.

Thank you for loading `any_llm` and comparing all 47 rather than reading the list. That is the detector this file did not have, and it is how the earlier version's three invented ids would have been caught before review rather than during it.

web/src/design-system/actions/CopyField.tsx
jigjigjig: Agreed, and not fixed here: filed as #1149.

You are right on both halves. The bug pre-dates this PR and the file is untouched by it, so a cleanup with its own test belongs in its own change. And the exposure is genuinely wider afterwards: `legacyCopy` was reached only by the concealed path before, and routing the default arrangement through the same helper for #957 makes it the fallback for every copy in this component. The issue says that in those terms rather than describing it as a pre-existing bug someone happened to notice, because the reason to prioritize it is the change in reach.

One detail I added while writing it up: the unguarded window is wider than `select()`. Everything from the `appendChild` at :36 to the `remove()` at :54 is outside any `try`, including `document.getSelection()` at :38, so the leaked node is reachable from three statements rather than one. The fix is the same, a `finally` around the whole body with the existing `execCommand` catch inside it.

A test that makes `select()` throw and asserts `document.body` holds no textarea afterwards is what would turn it into a check. Worth doing in that PR rather than asserting the `finally` exists.

web/src/shared/helpers/providers.ts
```diff
@@ -0,0 +1,90 @@
...
+ * third party got proper casing, which is the bug this map exists to fix,
+ * pointed at us.
+ *
+ * Names only. Vendor logos carry trademark terms of use, so which marks ship
+ * and under what terms is a decision this file does not make.
+ */
+const PROVIDER_DISPLAY_NAMES: Readonly<Record<string, string>> = {
+  anthropic: "Anthropic",
+  atlascloud: "AtlasCloud",
+  azure: "Azure OpenAI",
```
khaledosman: `azure` and `azureopenai` are two different any-llm providers and this gives them the same name.

```
azure       -> AzureProvider        "Azure Provider using the official Azure AI Inference SDK"
                                    docs: .../foundry/foundry-models/concepts/models-sold-directly-by-azure
azureopenai -> AzureopenaiProvider  "Azure OpenAI AnyLLM"
```

Two consequences. `azure` is labeled with the wrong Azure product, and on a deployment with credentials for both, `ModelsPage`'s provider filter offers two options reading "Azure OpenAI" with different values behind them, which is exactly the "tell two rows apart" problem the key suffix elsewhere in this PR exists to solve.

`azure: "Azure AI Foundry"` matches what any-llm's own doc URL points at and keeps the two distinguishable.

web/src/features/models/ModelComboBox.tsx
```diff
@@ -55,51 +109,32 @@ export function ModelComboBox({
...
     <ComboBoxField
       label={label}
       value={value}
       onChange={onChange}
       options={visible}
-      description={hint}
+      description={
+        hintPlacement === "detached" ? undefined : (hint ?? description)
+      }
+      describedBy={describedBy}
```
khaledosman: `describedBy` is passed through unconditionally, but `SectionRow` renders the span that carries the id only `{hint ? ... : null}`. With no failed provider and fewer than `MAX_VISIBLE` matches, which is the ordinary state, `hint` is `undefined`, so the input ships `aria-describedby="pool-hint-0"` (and the tier and chain equivalents) pointing at no element. That is an invalid ARIA reference, and nothing catches it: there is no axe pass in the suite.

`hint` is already in scope on the line above:

```suggestion
      describedBy={hint ? describedBy : undefined}
```

web/src/design-system/forms/ComboBoxField.tsx
```diff
@@ -117,6 +118,22 @@ export function ComboBoxField({
...
+   * The id of text elsewhere on the page that describes this field.
+   *
+   * For a field in a control row whose message is a sentence: a sentence long
+   * enough to wrap makes that field taller than its siblings and pushes its
+   * input line out of the row (`web/design/forms.md`, "Control rows"), so the
+   * caller renders it under the row instead. The `description` slot is what
+   * wires a message to the input, and text outside the field never reaches it,
+   * so this carries the association across the gap.
+   *
+   * Pairs with an absent `description`, which is the case it exists for. Set
```
khaledosman: The caveat is narrower than the behavior. `aria-describedby` on the `Input` replaces everything react-aria wired, which is the `<FieldError>` at the foot of this component as well as the `<Description>`. A caller that sets `describedBy` on a field that can go `isInvalid` loses the error announcement, silently and in the state where it matters most.

No current call site does both, so this is about what the next one will read. Worth naming `errorMessage` here alongside `description`, or guarding it in the component by dropping `describedBy` when `isInvalid` is set.

web/src/styles/globals.css
```diff
@@ -3180,8 +3180,37 @@ main {
...
 .otari-bulk-bar {
   animation: otari-bulk-bar-in 160ms ease-out;
+  /* Centered on the content, not the viewport: `left-1/2` alone put the bar
+     132px left of the table it belongs to and over the last visible rows.
+     The width goes with it, because a fixed element's `100%` is the viewport
+     too: centered but still viewport-wide, the bar reached back over the rail
+     on a narrow desktop (at 768px with the rail expanded it is 736px wide in a
+     504px pane). The call site's `max-w-3xl` still caps it. */
+  left: calc(50% + var(--rail-width, 0px) / 2);
+  width: calc(100% - var(--rail-width, 0px) - 2rem);
```
khaledosman, line 3229: This takes ownership of two utilities the call site still spells. `BulkActionBar.tsx` carries `left-1/2 w-[calc(100%-2rem)]` in its className, and an unlayered rule beats `@layer utilities`, so both are now inert: editing either one there changes nothing, with no error to say so. It is the same cascade fact the comments at lines 1254 and 1794 already lean on, applied to a file that does not know it.

Either drop `left-1/2` and `w-[calc(100%-2rem)]` from the className, keeping `max-w-3xl` (which this rule leaves alone and the comment above already credits), or say in the component that `.otari-bulk-bar` owns left and width.

khaledosman (approved):
Four findings inline, none blocking. Reviewed at db25f1d9; the `mzai`/`otari` question you raised yourself is answered by that commit, so it is not repeated below, and neither are the other two self-flagged points or CodeRabbit's four, which 087c344c already addresses.

Gates checked on the head: `generate_openapi.py --check` and `generate_postman.py --check` are both clean, the alembic chain has a single head (`c4e7a9b2d6f8` onto `a1d4f7c2e8b3`, with no competing child on `main`), and every key in the provider map is a real `AnyLLM.get_supported_providers()` id. The dashboard job's console warnings are 654 against 648 on main, all from the new cases, and the aria-label warnings are unchanged at 22.

One scope note. Nothing under `web/src` reads `key_suffix`: the only two occurrences are `tests/fixtures.ts` and `secretCaption.test.ts`, both there to satisfy the regenerated type. The body says as much, but the title does not, and the title is what git-cliff parses, so the changelog entry will read "show both ends of an API key" for a release in which no surface shows it. Worth retitling to what shipped (store and return), and leaving "show" for the PR that renders it.

---

web/src/design-system/actions/CopyField.tsx
jigjigjig: Satisfied. Filing it as #1149 rather than fixing it here is the right call, and widening the window from `select()` to the whole `appendChild`-to-`remove()` span is more correct than what I raised: the node is reachable from three statements, not one, so a `finally` around the lot is the fix rather than a guard on the call I happened to notice. Resolved; the issue carries it from here.

web/src/shared/helpers/providers.ts
jigjigjig: Re-ran the detector against the change. 49 keys now, all of them real any-llm ids, zero misses. Exactly four ids fall through (`cascadia`, `gmi`, `kenari`, `neosantara`) and the docstring says four, so the sentence and the map agree and the count is checkable rather than implicit. Naming them is what makes it re-runnable by the next person.

web/src/app/AppShell.test.tsx
jigjigjig: Accepting it as written, and I measured what happens rather than reasoning about it, because the answer is slightly better and slightly worse than the caveat says.

I removed `operatorOnly: "refused"` from the `/providers` entry and ran `src/app`. Exactly one test failed:

```
× declares the operator-only destinations and no others   (nav/registry.test.ts)
  1 failed | 250 passed
```

So your caveat is exactly right about this case: it **kept passing** with the axis undeclared, which is the silent degradation you described. But the change does not reach main silently overall, because `registry.test.ts` asserts the full set of `operatorOnly` destinations and goes red.

That tripwire is weaker than it looks, though, and in a specific way: the natural way to fix that failing test is to edit its expected list, which is exactly what someone removing the gate would do. The guard is one deliberate act away from being disarmed, and nothing at that moment points at this file.

**So: accepted, and recorded in the PR body rather than left in the thread.**

There is a better arrangement if you want it, and I would rather put it to you than take it unilaterally since the caveat is yours. One assertion inside this case that the row it rides on declares the axis, so the premise fails where the assumption lives instead of in another file:

```ts
// The case is only about the caller axis while the row it uses still declares one.
expect(navItemForPath("/providers")?.operatorOnly).toBe("refused")
```

That turns "silently stops testing the axis" into "fails in the file that depended on it". It costs a line and an import.

I have not added it, because you raised this and it stays open until you are satisfied either way. Say the word and I will, or say it is not worth the line and I will leave it accepted as written.

web/src/design-system/actions/CopyField.tsx
jigjigjig: Endpoint check: this reply is posted through the per-PR replies route, which works. My earlier 404 was a malformed path on my side, not a missing route. Correcting it here so the record is not wrong.

web/src/features/models/ModelComboBox.tsx
jigjigjig: Right, and worse than the description implies. Fixed in `0631b903` with your suggestion.

Measured rather than taken on the argument, because the number matters: on the story with a listing catalog, **five** inputs carried an `aria-describedby` pointing at an id that does not exist. Zero after. The failing-catalog story still wires all five, and the referenced element still carries the sentence, so the association survived the guard.

Worth saying how it got past me. My local env has a deliberately failing provider credential, which is what made the original alignment defect visible at all. In that state the hint always exists, so the span always renders and there is nothing dangling to find. The instrument I built to reproduce one bug was tuned away from the state this one lives in, and I probed it there and came back clean.

web/src/shared/helpers/providers.ts
jigjigjig: Correct on both halves, and I had it wrong from the wrong end. Fixed in `0631b903`.

I checked before changing it: `get_provider_class("azure")` is `AzureProvider` pointing at models-sold-directly-by-Azure, `azureopenai` is `AzureopenaiProvider`. Two products, and I had given them one name, which is the two-identical-rows problem you name.

`azure` is now "Azure AI Foundry". The test asserts the two are **distinguishable** rather than asserting either string, since what must hold is that no two ids share a label; the particular wording is a naming choice and should not be what a test pins.

This is the second finding in this map today from someone other than me, which is the useful signal: the map has no detector of its own beyond the id check, and a wrong or duplicated label fails nothing.

web/src/design-system/forms/ComboBoxField.tsx
jigjigjig: You are right that my caveat was narrower than the behavior, and I have taken the stronger of your two options: `describedBy` is dropped while the field is invalid, in `0631b903`.

Documenting it would have left a future caller one reading away from losing an error announcement silently, in the state where losing it matters most. The docstring now names what it displaces, the error slot as well as the description, rather than describing half of it.

web/src/styles/globals.css
jigjigjig: This is the one I am least comfortable about, because it is the exact failure class this PR spends a `web/AGENTS.md` note on and I wrote it into my own diff and then read past it in an end-to-end review of the whole branch.

Fixed in `0631b903`: `left-1/2` and `w-[calc(100%-2rem)]` are gone from the call site, `max-w-3xl` and the translate stay, and the component says in a comment that `.otari-bulk-bar` owns left and width and why a class there would be inert.

Verified after rather than assumed: bar centre 852 against a pane centre of 852 at 1440 with the classes removed, and the Playwright case that holds the offset at 1024 still passes.

PR #1150, opened 2026-09-14
fix(dashboard): name an unnamed budget by what it caps, not by its id

khaledosman:
## Description

Giving a budget a name is optional, so plenty of deployments have budgets with
none. Everywhere the dashboard had to name one of those, it showed the first
chunk of the budget's uuid: `04f2f38a`. In a table that is merely unhelpful. In
a picker it is worse than unhelpful, because a picker exists to let somebody
choose, and two unnamed budgets were two opaque strings with nothing to tell
them apart. Picking the wrong one silently gives every member of a workspace the
wrong spending ceiling.

An unnamed budget is now named by the thing an admin actually chooses by: what
it caps and how often. `$50.00 / month`, or `1,000,000 tokens / week`, or
`No limit`. If two unnamed budgets would still read the same, each one carries
the head of its id as a qualifier, so they stay distinguishable. A budget that
has a name keeps showing its name, unchanged.

The create and edit forms now also say what the budget will be called while the
name box is blank, so leaving it blank is a decision rather than a surprise.

This covers every place the old fallback was written: both pickers on Workspaces
and the default-budget column beside them, the member budget picker on
Organization members, the organization budgets card and its spend-ceiling
dialog, the deployment budgets page, and the overview's budget health, which was
showing the whole uuid rather than a fragment.

Two deliberate calls worth a reviewer's eye:

- The deployment budgets **table** keeps its existing muted `(unnamed)` marker
  rather than taking the derived label. That table already carries Limit, Reset
  and the id in columns of their own, so a derived label there would only repeat
  its neighbors. The organization budgets table does take it, because it has no
  id column and two unnamed budgets would otherwise be indistinguishable rows.
- The forms say the derived label in the field's description rather than as the
  placeholder, because `web/design/forms.md` reserves a placeholder for an
  example of what to type.

No API change, no migration. Existing unnamed budgets read better immediately.

## How to test it locally

1. `make dashboard` and start the gateway, or `pnpm --dir web run dev`.
2. Create two budgets with no name, both $50 with a monthly reset, and a third
   with no name at $500.
3. Go to Workspaces, edit a workspace, and open **Default member budget**. The
   two $50 ones read `$50.00 / month (04f2f38a)` and `$50.00 / month (9b71c0de)`;
   the $500 one reads `$500.00 / month` with no qualifier.
4. Organization members has the same picker under a member's ceiling, and the
   organization's Spend page has it on Add ceiling.
5. Open Create budget and type a limit with the name box empty. The hint under
   the name field says what the budget will be called if you save it that way.
6. Give a budget a name and confirm every one of those surfaces shows the name.

Automated: `pnpm --dir web test` covers the helper directly
(`web/src/features/budgets/budgetLabel.test.ts`, including the collision
qualifier and a budget the list does not carry), the Workspaces picker, both
create forms, and the ceiling picker. `pnpm --dir web run lint`,
`pnpm --dir web run typecheck` and `make lint` are green. No backend, OpenAPI or
generated-artifact change, so nothing was regenerated.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes mozilla-ai/otari-ai#2130

## Checklist

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

The issue offered three options and the choice of scope was made by a human: the
derived label plus the create-form hint, without making a name required. The
self-review before opening caught that the first draft made every picker walk
the whole budget list once per row, which is quadratic on a list capped at
100,000 rows; that became a single pass over the list.


khaledosman:
@coderabbitai review

PR #1156, opened 2026-09-14
feat(web-search): call Serply as a first-party web-search provider

googio:
## Description

Otari can already call a licensed search API in-process instead of running a SearXNG-shaped service in front of it, and ships Tavily and Brave that way. This adds Serply as a third choice: set `web_search_provider: serply` and `web_search_provider_api_key`, and `otari_web_search` searches Google's index through it. Nothing changes for a deployment that names a different provider or none.

Serply is a Google SERP API, so it slots into the same one-function-per-provider shape as the two already there. Two things about it differ from its neighbors and are handled in the function:

- It reads `num` as an approximate bound on one page of results, not a count. `num=20` answers with the same page as `num=10`, and a page can come back holding one row more than was asked for. So the request is clamped to the page size and the answer is trimmed to what the caller asked for, which a provider that returns at most what it was asked for does not need. Without the trim, `web_search_max_results` would be a number the model-facing result block is allowed to exceed.
- It spells recency as Google's own `tbs` window, so the same `time_range` vocabulary Tavily and Brave take maps onto `qdr:d` through `qdr:y` rather than onto a spelling of its own. A stored `provider_options` keeps filtering by recency when a deployment switches provider.

Like Brave, it returns snippets rather than page text, so `extracted_content` is left unset and the caller does its own fetch-and-extract pass. `provider_options` is whitelisted the same way: only `max_results` and `time_range` reach the request, so a workspace cannot set a Serply request field the deployment never chose, and the endpoint is a module constant that no option can redirect.

One thing I found while testing that you may have an opinion on. For some queries Serply folds the local pack into `results` as a row that calls itself `organic` but links to `google.com/searchviewer/...` rather than to a site, while the `places` key stays empty. The fetch-and-extract pass then spends a fetch on it and gets nothing useful. I left it alone, since the row is what the provider returned and neither of the other two functions filters what it is given, and noted it in the docstring instead. Happy to drop those rows if you would rather the function did.

This is an Otari change rather than an any-llm one: it is the `otari_web_search` tool loop, not a provider model call.

## How to test it locally

```bash
uv run pytest tests/unit/test_web_search_providers.py -v
```

Five new tests cover the mapping, the page trim, the clamp, the recency window, and a missing `results` list. The HTTP layer is mocked, so they need no key and no network.

To see it end to end, set `web_search_provider: serply` and `web_search_provider_api_key` in `config.yml`, then send an `otari_web_search` request as [docs/tools.md](docs/tools.md) describes.

I also ran the function against the live API while writing it, which is where the two quirks above come from:

```
asked 4 -> got 4
max_results=500 -> 10 hits (page clamp + trim)
time_range=day -> 4 hits, first: https://finance.yahoo.com/technology/article/openai-ceo-sam-altman-on-...
bad key -> serply search returned HTTP 401
```

That last line is the sanitization the module already does: a real 401 body reads `{"detail":"Invalid API key"}`, and none of it reaches the message.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

None. Happy to open one if you would rather discuss a third provider before the code.

## Checklist

- [x] If the API contract changed, I regenerated the OpenAPI spec (`uv run python scripts/generate_openapi.py`). No route or schema changed, and neither generated artifact names a provider, so both are unchanged. `make openapi-check` and `make postman-check` pass.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

Claude Code (Claude Opus 5)

The function and its tests were written against the existing Tavily and Brave functions in the same module. The response mapping, the page bound and the recency window were each checked against the live Serply API before the code was written, and the numbers quoted above are from those runs.

---

Disclosure: I work with Serply. Happy to adjust scope or naming, or to drop this entirely if a third provider is not a direction you want.

## Summary

Added Serply as a supported web-search provider for `otari_web_search`.

Deployments can select `web_search_provider: serply` and use `web_search_provider_api_key`. Serply results return snippets, which preserves the existing fetch-and-extract flow.

## Technical notes

- Maps `max_results` and `time_range` to Serply requests.
- Limits page size and trims returned results.
- Adds documentation, configuration support, and five unit tests.
- Existing providers remain unchanged.


googio:
Ran the full `make lint` target on the branch head (87ee883b), not just Ruff:

```
uv run python scripts/check_architecture.py
No architecture violations found
uv run python scripts/check_alembic_heads.py
Single head: a1d4f7c2e8b3
uv run ruff check src tests scripts
All checks passed!
```

The architecture check is the one worth calling out here: `web_search_providers.py` sits in the services layer and the new code only reaches for `httpx` and the existing config, so it adds no new cross-layer import. No migrations in this PR, so the head count is unchanged.

PR #1160, opened 2026-09-15
feat(guardrails): list only the guardrails reachable over a hosted API

### LINKED ISSUE #1159: Limit the built-in guardrail catalog to guardrails a hosted API can reach
### Problem or use case

`GET /api/v1/tool-settings/guardrails/catalog` publishes every guardrail any-guardrail ships. Today that is 40 rows, because `build_builtin_guardrail_catalog()` in `src/gateway/services/guardrail_catalog.py` iterates `GuardrailName` whole.

30 of those 40 cannot run here. A `local_encoder`, `local_decoder` or `library_wrapped` guardrail works by downloading model weights and holding them in the gateway process. Otari does not do that, and a deployment that wants one already has a better place to put it: the any-guardrail service `guardrails_url` points at, whose guardrails are picked through the `/profiles` catalog beside this one. So the endpoint is a picker offering 30 choices that nothing here can build.

The two fields meant to soften that do not. `runnable` is decided by `_BACKEND_PACKAGES`, a hand-maintained 40 entry table that has to chase every guardrail upstream adds. `missing_extra` names `guardrails-local`, and `pyproject.toml` declares no such extra, so the remedy it hands a caller is a `pip install` that fails.

### Proposed solution

Keep a guardrail when a hosted API can reach it:

```python
BackendType.HOSTED_API in ({metadata.backend} | metadata.alternate_backends)
```

That leaves 10 rows: Alinia, `any_llm`, Azure Content Safety, Azure Prompt Shields, Bedrock Guardrails, Lakera Guard, OpenAI Moderation, Patronus, SusFactor, watsonx Guardian.

SusFactor is there on its alternate backend rather than its default, which is `local_encoder`. Worth knowing up front: reaching its hosted path needs a `provider=` constructor object that any-guardrail's parameter registry does not publish, so the row is listed and not yet configurable for hosted use.

Then drop `runnable`, `missing_extra`, and the machinery behind them (`LOCAL_GUARDRAILS_EXTRA`, `_TRANSFORMERS`, `_BACKEND_PACKAGES`, `_installed`, `_backend_availability`). Once no listed guardrail loads a model backend, "are this host's packages installed" stops being the question the catalog exists to answer, and the field that answered it pointed at nothing installable.

The filter belongs in `build_builtin_guardrail_catalog` alone. `_specs_for_stage` and `_parameter_specs` are shared with `fetch_guardrail_catalog`, which must keep typing all 40 classes: an operator's own guardrails service can run a local model, and the `/profiles` half is how that is picked.

This narrows a published response, so `docs/public/openapi.json`, the Postman collection, and `web/src/client/schema.ts` all move. Nothing in `web/src` outside the generated client reads these types yet, so no dashboard code changes.

The endpoint stays on `require_deployment_operator`. Dropping `runnable` removes the weaker of the two reasons for that gate. The stronger one is unchanged and in fact grows: 9 of the 10 remaining rows carry a secret constructor argument, and 7 carry an endpoint or base URL.

### Does this belong in Otari?

Yes. any-guardrail's metadata is already correct; this repository is the one choosing which of it to publish, and the choice is about what Otari itself can build.

### Alternatives considered

**Keep all 40 and let `runnable` mark the difference.** This is the current design and it does not work. `runnable` answers a question about installed packages, not about whether this gateway builds guardrails of that kind at all, and the extra it names does not exist.

**Declare the missing `guardrails-local` extra so the remedy works.** That buys a working `pip install` for a capability Otari still would not have.

**Filter on `backend` alone and drop SusFactor.** Simpler to state, but it discards a guardrail any-guardrail can reach over HTTP.

dpoulopoulos:
## Description

Otari's guardrail catalog offered every guardrail any-guardrail ships, 40 of them. Most of those cannot run here. A guardrail works in one of two ways: it calls a hosted API, or it downloads model weights and holds them in the process running it. Otari does the first only. The second belongs in the any-guardrail service `guardrails_url` points at, which Otari already talks to over HTTP and whose guardrails are picked through the profiles catalog beside this one.

So an operator opening the form saw 40 choices, 30 of which nothing here could build.

The catalog now lists the 10 a hosted API can reach: Alinia, `any_llm`, Azure Content Safety, Azure Prompt Shields, Bedrock Guardrails, Lakera Guard, OpenAI Moderation, Patronus, SusFactor, watsonx Guardian. The rule is any-guardrail's own backend metadata rather than a list kept in this repository, so a hosted guardrail upstream adds arrives here on its own. It counts alternate backends beside the default, which is why SusFactor is in: it defaults to a local encoder and also answers over a hosted API, and the hosted path is the one Otari would take.

Two response fields go with the 30. `runnable` said whether the packages a guardrail's model backend needs were installed on this host, and after the filter no listed guardrail has a model backend. `missing_extra` named the extra that would fix it, and the name it gave was `guardrails-local`, which `pyproject.toml` has never declared, so the remedy it handed a caller was a `pip install` that fails. Both are removed, along with the hand-maintained 40 entry package table behind them that had to chase every guardrail upstream added.

The endpoint keeps its deployment-operator gate. That gate rested on two reasons and loses the weaker one. The stronger one is unchanged and in fact grows: 9 of the 10 remaining rows carry a secret constructor argument, and 7 carry an endpoint or base URL, so this is still the picker behind a write that stores a vendor credential for the whole deployment.

One thing reviewers should know about SusFactor: it is listed, and its hosted path is not yet configurable. Reaching that path needs a `provider=` constructor object, and any-guardrail's parameter registry does not publish one, so the row selects and its hosted mode cannot yet be filled in. It is listed rather than dropped because the metadata says the path exists, and the alternative is a rule that silently disagrees with upstream.

No dashboard code changes. `web/src/client/index.ts` does not re-export these types and no hook or component reads them yet, so the only thing moving under `web/` is the generated client.

## How to test it locally

```bash
uv run otari serve
curl -s -H "Authorization: Bearer $OTARI_MASTER_KEY" \
  localhost:8000/api/v1/tool-settings/guardrails/catalog \
| jq '{count: (.guardrails | length), names: [.guardrails[].guardrail_name]}'
```

Expect 10 rows, `susfactor` among them, and no `llama_guard`. Every row should carry `create_parameters` and `validate_parameters`, and none should carry `runnable` or `missing_extra`.

Already proven by automation:

- `tests/unit/test_guardrail_catalog.py` derives the expected set from `GUARDRAIL_METADATA` rather than retyping it, pins the two non-obvious ends (SusFactor listed, `llama_guard` absent), and asserts the two removed fields are gone from the published model.
- `tests/unit/test_tool_settings_endpoint.py` covers the route, its master-key requirement, and that it stays on the operator router.
- `make lint`, `make typecheck`, `make openapi-check`, `make postman-check`: clean. `git diff --exit-code web/src/client/schema.ts` is clean after regeneration.
- 147 integration tests across the guardrail, tool-settings, operator-gate, hybrid-surface and API-prefix suites pass.

Left to eyeball: the wording of the three docstrings that reach the published spec, and the new `docs/guardrails.md` section.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [x] Refactor
- [x] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Closes #1159

## Checklist

Note on `make test`: 14 unit tests fail on this machine and fail identically on `main`, all in the master-key, dashboard-session and CLI-default suites. A repo-root `.env` sets `OTARI_DATABASE_URL`, which those bare-default tests read. No guardrail or tool-settings test is among them.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

The filter rule, whether to keep SusFactor, and whether to remove `runnable` rather than repair the extra it named were decided by the author, not the model.

## Summary

- Limited the built-in guardrail catalog to the 10 guardrails reachable through hosted APIs.
- Retained SusFactor through its hosted alternate backend.
- Removed `runnable`, `missing_extra`, and related package-detection logic.
- Kept the deployment-operator access requirement.
- Preserved the profiles catalog for operator-managed guardrail services.
- Updated documentation, API artifacts, generated client types, and tests.

This gives users an accurate catalog and avoids listing guardrails that Otari cannot run directly.


javiermtorres (changes requested):
Apart from the HOSTED_API ones, please include deepset, duo_guard, jasper, protectai, sentinel, granite_guardian, but making sure the base_url is used, so any-guardrail does not try to start the llama/encoderfiles on its own.

javiermtorres:
@dpoulopoulos as agreed, we'll leave the llamafile/encoderfile guardrails out until we revisit the model types in any-guardrail.

javiermtorres (approved):

PR #1161, opened 2026-09-15
fix(dashboard): conceal API keys by default and align key controls

jigjigjig:
## Description

New and regenerated API keys now start concealed, showing the first eight and last four characters with bullets between them. The request examples start concealed too, and all three fields reveal or hide together.

Copy and reveal sit side by side as compact icon controls inside the key field. Copy still copies the complete key while it is hidden and confirms success. The keys table now includes the stored suffix when available, with a prefix-only fallback for older keys.

## How to test it locally

1. Start the dashboard and create an API key. Confirm the key and request examples start concealed.
2. Copy the key without revealing it; paste into a local scratch field and confirm the complete value was copied.
3. Reveal and hide the key. Confirm the request examples follow its visibility.
4. Dismiss the dialog and regenerate the key. Confirm the replacement starts concealed and the table shows its fingerprint.

Validation:
- `make lint`: passed (architecture, migration heads, Ruff).
- Dashboard lint, typecheck, production build, and Storybook build: passed.
- `pnpm --dir web test --maxWorkers=4`: 6,406 tests passed.
- Keys page tests after the final legacy-key assertion: 46 passed.
- Playwright management flow and its onboarding/seed dependencies: 18 passed. Used a temporary config with only the standalone test server because another process occupies hybrid port 8010.
- Browser preview: inspected masked/revealed controls; both have 44px targets below the desktop breakpoint.
- Catalog smoke sweep: 844/852 renders passed. The eight failures are four unchanged PublicAuthLayout stories in both themes, all missing ThemeProvider. Copy stories passed in both themes.

## PR Type

- [x] Bug Fix

## Relevant issues

Requested API-key experience improvement. No linked issue.

## Checklist

- [ ] I ran all backend Definition of Done checks locally (frontend-only change; validation listed above).

## AI Usage

**Any additional AI details you'd like to share:** Implemented and self-reviewed against the repository's frontend and review guidance.

## Summary

- API keys now appear concealed by default, showing only the first eight and last four characters.
- Users can reveal, hide, and copy keys with compact controls inside the key field.
- Copying works while a key is concealed and copies the complete value.
- The keys table now shows the stored key suffix when available and supports older keys without one.
- Tests, examples, and design documentation now reflect the updated key experience.

These changes reduce accidental exposure of API keys while keeping key management and copying simple.


tbille (approved):

web/src/features/keys/KeysPage.test.tsx
```diff
@@ -262,11 +273,21 @@ describe("KeysPage", () => {
...
     })
+    expect(within(reveal).getByLabelText("Secret key")).toHaveValue(
+      "gw-NEWSE••••••••0000",
+    )
+    expect(within(reveal).getByLabelText("curl")).not.toHaveValue(
+      expect.stringContaining(NEW_SECRET),
+    )
+    expect(
+      within(reveal).getByLabelText("Python (OpenAI SDK)"),
+    ).not.toHaveValue(expect.stringContaining(NEW_SECRET))
```
khaledosman: These assertions are vacuous and pass whatever the field holds.

jest-dom's `toHaveValue` compares through `compareAsSet`, which for non-arrays is a bare `val1 === val2`; it never calls `asymmetricMatch`. A string is never `===` an asymmetric matcher object, so `.not.toHaveValue(expect.stringContaining(...))` always passes. Verified by rendering a textarea whose value *does* contain the secret:

```jsx
render(<textarea aria-label="curl" readOnly value="Bearer gw-SECRET-1234" />)
expect(screen.getByLabelText("curl")).not.toHaveValue(
  expect.stringContaining("gw-SECRET-1234"),
)   // 1 passed
```

These are exactly the assertions guarding this PR's central claim, that the snippets start concealed, so that claim currently has no unit coverage. Same at `:644-647` for the regenerate path.

```suggestion
    expect(
      (within(reveal).getByLabelText("curl") as HTMLTextAreaElement).value,
    ).not.toContain(NEW_SECRET)
    expect(
      (
        within(reveal).getByLabelText(
          "Python (OpenAI SDK)",
        ) as HTMLTextAreaElement
      ).value,
    ).not.toContain(NEW_SECRET)
```

web/src/features/keys/KeysPage.test.tsx
```diff
@@ -621,9 +639,12 @@ describe("KeysPage", () => {
...
     })
-    // Revealed on arrival, the same as a created key: regenerate hands over the
-    // same thing and hands it over the same way.
-    expect(within(reveal).getByDisplayValue(REGEN_SECRET)).toBeInTheDocument()
+    expect(within(reveal).getByLabelText("Secret key")).toHaveValue(
+      "gw-REGEN••••••••0000",
+    )
+    expect(within(reveal).getByLabelText("curl")).not.toHaveValue(
+      expect.stringContaining(REGEN_SECRET),
+    )
```
khaledosman: Same vacuous matcher as at `:282`: `.not.toHaveValue(expect.stringContaining(...))` passes unconditionally.

```suggestion
    expect(within(reveal).getByLabelText("Secret key")).toHaveValue(
      "gw-REGEN••••••••0000",
    )
    expect(
      (within(reveal).getByLabelText("curl") as HTMLTextAreaElement).value,
    ).not.toContain(REGEN_SECRET)
```

web/src/design-system/actions/CopyField.tsx
```diff
@@ -372,6 +366,29 @@ export function CopyField({
     </Button>
   )
 
+  const inlineControls = concealed !== undefined && !multiline
+  const copyButton = (
```
khaledosman, line 399: This is a 32px touch target. `size="sm"` is 32px (`Button.tsx:22`) and `isIconOnly` zeroes the padding to keep it square (`globals.css:1797`), and nothing here adds `min-h-11` or a bleed. `motion-and-access.md:69` is explicit: 44px is the floor, and a control under it "needs the same kind of argument, in writing, or it is a bug." The PR description's "both have 44px targets below the desktop breakpoint" is not what this renders.

`CopyButton.tsx:74`'s `before:-inset-1.5` device is unavailable here: the two controls sit `gap-1` (4px) apart, so 6px bleeds would overlap and a press near the seam would hit the wrong control. Real boxes are the fix, and the padding already written fits them exactly, since `pr-24` is 96px and 44+4+44+4 = 96. Add the same className to `revealToggle` at `:353`, which has the gap today too.

```suggestion
  const inlineControls = concealed !== undefined && !multiline
  const copyButton = (
    <Button
      size="sm"
      variant="ghost"
      isIconOnly={inlineControls}
      aria-label={inlineControls ? `Copy ${label}` : undefined}
      // 32px is under the 44px floor motion-and-access.md sets, and the 4px gap
      // to the toggle leaves no room for CopyButton's `before:` bleed. The
      // field's `pr-24` already reserves the 96px two real targets need.
      className={inlineControls ? "min-h-11 min-w-11 md:min-h-0 md:min-w-0" : undefined}
      onPress={copy}
    >
```

web/src/features/keys/KeysPage.tsx
```diff
@@ -142,12 +126,9 @@ function KeySecretStep({
   // the API. Undefined when it has not said where its gateway is (otari#823).
   const baseUrl = resolveSnippetBaseUrl(useDeployment())
   const secret = result.key
-  // One reveal for the key and both snippets, because the snippets carry the
-  // same secret: revealing the key while a snippet still printed bullets, or
-  // the reverse, would be one credential in two states on one screen. Open,
-  // because this screen exists to hand the key over and there is no second
-  // chance to read it; the toggle conceals all three.
-  const [isSecretRevealed, setIsSecretRevealed] = useState(true)
+  const concealedSecret = `${secret.slice(0, 8)}••••••••${secret.slice(-4)}`
```
khaledosman: Two things here.

The 8/8/4 fingerprint is hand-built at this call site and again at `copy.stories.tsx:201`. `CONCEALED_SECRET` is exported from the design system precisely so the stand-in's shape is decided in one place; its docstring argues that the key's length must not reach the screen, which is a rule that now has to be re-derived at every call site. Put the new shape beside it as `concealedFingerprint(value)` and call that from both.

And one credential now wears two stand-ins on one screen: this field shows `gw-NEWSE••••••••0000` while both snippets at `:145-149` still show `CONCEALED_SECRET`'s 16 bullets. Pick one.

web/src/features/keys/KeysPage.tsx
```diff
@@ -108,23 +108,7 @@ const getKeyRowKey = (k: ApiKey): string => k.id
...
- * `isDismissable={false}` is the strip's "there is one control and it is the
- * acknowledgement", kept because a stray backdrop click here loses the key
- * forever.
- *
- * The snippets come with it. An operator creating their first key arrives from
- * the empty state, and "make your first call" is the next thing they need, so
- * leaving it behind on the page would have made this dialog the one place the
- * flow got worse.
- */
+/** One-time key handoff with copyable request examples. */
```
khaledosman, line 114: The replaced docstring was mostly implementation history and fair to prune, but it carried the one live reason for `isDismissable={false}` at `:707`, which is now unexplained: a stray backdrop click here loses the key forever. Keep that sentence, on the prop rather than here.

khaledosman (approved):
Five findings inline. The first is the only one I would fix before merge, and it is a test-only change: three assertions guarding the central claim of this PR pass unconditionally.

---

web/src/features/keys/KeysPage.test.tsx
jigjigjig: Addressed in feddb527: both snippet assertions now read the textarea value and use `not.toContain`, so they fail if the plaintext reaches the field. Swept the rest of `web/src` for the same shape; these two were the only asymmetric matchers inside a `.not`.

web/src/features/keys/KeysPage.test.tsx
jigjigjig: Addressed in feddb527: same fix here, asserting on the textarea value with `not.toContain(REGEN_SECRET)`.

web/src/design-system/actions/CopyField.tsx
jigjigjig: Addressed in feddb527: both in-field controls take a real 44x44 box below `md` (`min-h-11 min-w-11 md:min-h-8 md:min-w-8`), which the `pr-24` already written fits exactly at 44+44+4 gap+4 inset. Measured in the built catalog rather than claimed: 44x44 at 375px and 32x32 at 1280px for both controls. A test in `CopyField.test.tsx` holds the floor.

web/src/features/keys/KeysPage.tsx
jigjigjig: Addressed in feddb527: the shape moves beside `CONCEALED_SECRET` as `concealedFingerprint`, called from `KeysPage` and the story, and it falls back to the plain stand-in below sixteen characters where the two ends would meet. The snippets now build around that same stand-in, so the credential wears one on that screen instead of two.

web/src/features/keys/KeysPage.tsx
jigjigjig: Addressed in feddb527: the sentence is back, on `isDismissable` itself.

PR #1163, opened 2026-09-15
fix(dashboard): make KPI sparklines non-interactive

jigjigjig:
## Description

Make the compact Overview graphs static: hovering no longer shows a blue dot, and pointer, touch, and keyboard input cannot interact with them. Their data, shape, color, dimensions, and accessible labels stay unchanged. The shared sparklines on Usage receive the same fix.

## How to test it locally

1. Build and open the dashboard with usage data.
2. Hover, click, and touch the Overview sparklines. They should remain visually unchanged with no active dot.
3. Tab through the page. The sparklines should not receive focus. Repeat on Usage's compact graphs.

Passed locally: `make lint`, dashboard Biome check, `tsc -b --noEmit`, and all 6,406 Vitest tests across 185 files. Dashboard tools ran through their installed binaries because pnpm's version switch could not reach the registry.

## PR Type

- [x] Bug Fix

## Relevant issues

User-reported Overview hover behavior; no linked issue.

## Checklist

Frontend-only change; backend typecheck/tests are not applicable. No documentation or API artifacts changed.

## AI Usage

- [ ] No AI was used.

## Summary

Dashboard KPI sparklines in Overview and Usage are now static and non-interactive. Hover, pointer, touch, and keyboard input no longer activate the charts or show a blue dot. Data, appearance, dimensions, and accessible labels remain unchanged.

## Technical notes

- Disabled pointer interaction and Recharts accessibility input.
- Removed the active hover dot.
- Added tests for non-interactive behavior and accessibility attributes.

## Validation

- Linting, Biome, and TypeScript checks passed.
- 6,406 Vitest tests passed across 185 files.


web/src/design-system/metrics/charts.test.tsx
```diff
@@ -123,6 +123,11 @@ describe("Sparkline", () => {
 
     expect(screen.getByRole("img", { name: "Spend trend" })).toBeInTheDocument()
     expect(container.querySelector(".recharts-line")).not.toBeNull()
+    expect(screen.getByRole("img", { name: "Spend trend" })).toHaveClass(
+      "pointer-events-none",
+    )
+    expect(container.querySelector("svg")).toHaveAttribute("tabindex", "-1")
+    expect(screen.queryByRole("application")).not.toBeInTheDocument()
   })
 })
```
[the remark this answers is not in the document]
jigjigjig: Added a mouse-move assertion with measured chart bounds. Verified it fails when activeDot={false} is removed and passes with the fix restored.

web/src/design-system/metrics/charts.tsx
```diff
@@ -381,14 +381,15 @@ export function Sparkline({
...
+    <div
+      role="img"
+      aria-label={ariaLabel}
+      className="pointer-events-none w-full"
+    >
       <ResponsiveContainer width="100%" height={height}>
         <LineChart
           data={data}
+          accessibilityLayer={false}
           tabIndex={-1}
```
khaledosman: This drops the comment that explained why `tabIndex={-1}` exists (recharts gives its `<svg>` `tabIndex={0}`, which put three empty focus stops on Overview) and adds a prop that supersedes it. `RootSurface.js:42-50` sets `tabIndex = hasAccessibilityLayer ? 0 : undefined` and `role = hasAccessibilityLayer ? 'application' : undefined` unless an explicit numeric `tabIndex` / string `role` is passed, so `accessibilityLayer={false}` already does this line's job. `activeDot={false}` on line 402 is similarly unreachable once pointer events are off and there is no keyboard activation.

That leaves three overlapping non-interaction props with nothing saying which one is load-bearing. Either drop `tabIndex={-1}` (along with the `tabindex` assertion at `charts.test.tsx:128`, which passes on main too and so guards nothing here), or keep it and restore a line of rationale. The docstring on line 370 already says "no ticks, no tooltip" and is the natural home for it.

Worth recording there that switching the accessibility layer off is deliberate and not an a11y regression: the file header already states that the surrounding tables and captions are the accessible source of truth. Without that note, the next reader is fairly likely to turn it back on.

khaledosman (approved):
Verified the mechanism rather than the description: recharts 3.10.1 defaults `accessibilityLayer: true` for cartesian charts (`chart/CartesianChart.js:24`), `container/RootSurface.js:42-50` derives the svg's `tabIndex` and `role` from it, and `component/ActivePoints.js` returns `null` on `activeDot === false`. So the fix is aimed at the right props.

I also ran the new assertions against main's component in a scratch worktree: the `pointer-events-none` and `role="application"` checks both fail pre-fix, so they are real guards rather than assertions that would pass either way. Biome, `tsc -b --noEmit`, and the 11 tests in the file pass at head, and the dashboard CI log carries no new console errors.

Scope looks right: Overview's other graph (`SpendChart`) is a hand-rolled bar chart with per-day buttons and stays interactive, and `KpiCell` is not clickable, so the pointer-events change has no pass-through effect.

One comment inline, about the rationale the diff removed.

---
Review by [Claude Code](https://claude.com/claude-code) (Opus 5)

web/src/design-system/metrics/charts.tsx
jigjigjig: Removed the redundant tabIndex prop and updated the test to assert that the SVG has no tabindex. Added a concise note explaining that the labeled image supplies accessibility while keyboard chart navigation is deliberately disabled. Lint, typecheck, and all 11 chart tests pass.

PR #1166, opened 2026-09-15
fix(dashboard): show the API key fingerprint in the activation guide

jigjigjig:
## Description

The activation guide now shows the first eight and last four API-key characters, matching the key-creation dialog. The shared field provides the same inline reveal and copy controls. The key field and request examples share the same fingerprint until revealed. Short keys are fully concealed, and Copy still copies the complete key.

## How to test it locally

1. Build and start the dashboard with a provider configured.
2. Create and select a fresh workspace, then open Overview.
3. Confirm the activation sheet shows a partial key with inline eye and copy controls.
4. Reveal and hide it; confirm the example follows. Copy while hidden and confirm it yields the full key.

Validated with `make lint`, dashboard lint/typecheck/build, all 23 activation component tests and 27 CopyField tests, and 12 onboarding browser tests. The full dashboard suite reported 6,409 passing tests and one unrelated Playground comparison failure; all 25 Playground tests passed on an isolated rerun. Browser tests used a temporary standalone-only server config because hybrid port 8010 is occupied locally. The live activation sheet was visually checked in the local demo.

## PR Type

- [x] Bug Fix

## Relevant issues

Follow-up to #1161 for the activation flow introduced in #1101.

## Checklist

- [ ] I ran all backend Definition of Done checks locally (frontend-only change; validation above).
- [x] Documentation was updated where necessary (the existing API-key handoff guidance already describes this display).

## AI Usage

**Any additional AI details you'd like to share:** Implemented and self-reviewed against the repository frontend guidance.

## Summary

- The activation guide now shows the same API-key fingerprint as the key-creation dialog.
- Users can reveal, hide, and copy the complete API key.
- Short or invalid keys remain fully concealed.
- Request examples use the same concealed or revealed key as the key field.

## Technical notes

- Added shared fingerprint handling in `SetupSheet`.
- Added coverage for concealment, reveal and copy behavior.
- Updated browser tests to verify the minted key across the field and request examples.

## Validation

- Linting, type checking, builds, activation tests, copy-field tests, onboarding browser tests, and dashboard tests passed.
- One unrelated Playground comparison test failed in the full suite. All Playground tests passed when rerun separately.


web/src/features/onboarding/SetupSheet.tsx
```diff
@@ -163,7 +163,7 @@ export function SetupSheet({
           <CopyField
             label="Your API key"
             value={apiKey}
-            concealed={CONCEALED_SECRET}
+            concealed={`${apiKey.slice(0, 8)}••••••••${apiKey.slice(-4)}`}
```
khaledosman: Two problems here, one fix.

**This reimplements `concealedFingerprint`** (`web/src/design-system/actions/CopyField.tsx:78`) minus its guard. The helper is this exact expression plus a `value.length >= 16` fallback to `CONCEALED_SECRET`, and the `concealed` prop's own docblock names it as what a whole-value field passes here; `KeysPage.tsx:134` already calls it. Dropping the guard matters because line 148 only checks `undefined`: an empty or short `apiKey` gets overlapping slices and puts most of the value on screen, where the helper would conceal it outright.

**And the snippet at line 94 was not moved with it.** It still embeds the 16-bullet `CONCEALED_SECRET` while this field shows `gw-XXXXX••••••••XXXX`, so the same credential wears two different stand-ins on one screen. That is the arrangement `CopyFieldProps.concealed` rules out ("A snippet passes the same snippet built around whichever of those the field beside it shows... one credential does not wear two stand-ins on one screen"), and the one `KeySecretStep` avoids by deriving the field and both snippets from a single `concealedSecret` (`KeysPage.tsx:133-152`). The PR body says this matches the key-creation dialog; right now it matches half of it.

Hoist one value and feed both:

```tsx
const concealedKey =
  apiKey === undefined ? CONCEALED_SECRET : concealedFingerprint(apiKey)
```

then `apiKey: isRevealed && apiKey !== undefined ? apiKey : concealedKey` at line 94, and `concealed={concealedKey}` here.

web/e2e/dashboard.spec.ts
```diff
@@ -81,6 +81,10 @@ test.describe("dashboard core flows", () => {
     await sheet.getByRole("button", { name: "Show Your API key" }).click()
     await expect(key).toHaveValue(/^gw-/)
     await expect(curl).toContainText(/Otari-Key: gw-/)
+    await expect(key).not.toHaveValue(/•/)
+    await sheet.getByRole("button", { name: "Hide Your API key" }).click()
+    await expect(key).toHaveValue(/^gw-.{5}•{8}.{4}$/)
+    await expect(curl).not.toContainText(/gw-/)
```
khaledosman: This assertion (and the pre-existing one at line 79) only holds while the snippet keeps the full-bullet stand-in. Once it carries the same fingerprint the field shows, `not.toContainText(/gw-/)` fails, and line 83's `toContainText(/Otari-Key: gw-/)` stops discriminating revealed from concealed as well. As written the test pins the mismatch in place.

Assert against the full key instead, the way `KeysPage.test.tsx:283` does with `not.toContain(NEW_SECRET)`. The comment at line 78, "Concealed in the example too, because it is the same secret", is already only half true as of this diff.

khaledosman (changes requested):
The field half of this is right, but the snippet beside it was left on the old stand-in, so one credential now wears two of them on one screen. Two inline comments; the first is the blocking one.

web/src/features/onboarding/SetupSheet.tsx
jigjigjig: Addressed in 57a66004: the field and examples now share concealedFingerprint, including its short-key fallback. Added coverage for matching fingerprints and empty/short keys.

web/e2e/dashboard.spec.ts
jigjigjig: Addressed in 57a66004: the browser test captures the minted key and verifies its exact presence only after reveal, its absence while concealed, and the same fingerprint in the field and example.

jigjigjig (commented):
Two test findings, neither blocking: one `it.each` case pins behavior `SetupSheet.tsx:153` documents as wrong, and the browser test asserts the fingerprint's *shape* where it has the value to assert. The component change itself is clean and sweeps to the last call site (`SettingsPage.tsx:341` keeps the plain stand-in correctly, per the settled spec: no stored prefix or suffix for the master key).

Title nit for the changelog line: the repo's settled word for this is `fingerprint`, and "show partial API keys" reads as a leak rather than a fix. `fix(dashboard): show the API key fingerprint in the activation guide` is the sentence I would release.

web/src/features/onboarding/SetupGuide.test.tsx
```diff
@@ -300,13 +302,54 @@ describe("SetupGuide", () => {
...
+
+    await user.click(screen.getByRole("button", { name: "Hide Your API key" }))
+    expect(screen.getByLabelText("Your API key")).toHaveValue(
+      "gw-setup••••••••-key",
+    )
+    expect(snippet("curl")).not.toHaveTextContent(KEY)
+    expect(snippet("curl")).toHaveTextContent("gw-setup••••••••-key")
+  })
+
+  it.each(["", "short-key", "123456789012345"])(
```
jigjigjig: Drop `""` from this list. It is the one case that is not a short key but an absent one, and asserting it renders a concealed `CopyField` pins exactly the state `SetupSheet.tsx:153` says must not exist: *"Not a `CopyField` holding an empty string: every credential field here conceals, so an empty one would show the same run of bullets a real key does and invite a copy that yields nothing."* With `apiKey=""` the `apiKey === undefined` guard passes, the field shows sixteen bullets, and Copy puts an empty string on the clipboard while announcing "Copied to clipboard."

It is also unreachable: `generate_api_key` (`src/gateway/auth/models.py:12`) returns `gw-` + `token_urlsafe(48)` and validates the format before returning, so the mint never yields a short key, let alone an empty one. `"123456789012345"` is the case worth keeping, because it is the boundary `concealedFingerprint`'s `>= 16` guard turns on; `"short-key"` is the same branch a second time.

```suggestion
  it.each(["short-key", "123456789012345"])(
```

web/e2e/dashboard.spec.ts
```diff
@@ -66,7 +73,9 @@ test.describe("dashboard core flows", () => {
     // picks up the field's own "Show API key" toggle.
     const key = sheet.getByRole("textbox", { name: "Your API key" })
     await expect(key).toBeVisible()
-    await expect(key).not.toHaveValue(/^gw-/)
+    await expect(key).toHaveValue(/^gw-.{5}•{8}.{4}$/)
```
jigjigjig: This restates `concealedFingerprint`'s shape as a regex when the test is already holding the value to derive it from. `/^gw-.{5}•{8}.{4}$/` passes on any key with that shape, including one that is not `fullKey` at all, and it makes the reader count `gw-` plus five to get back to "the first eight characters."

Derive it instead, which also retires lines 77-78 (a string built from `fullKey.slice` cannot be `fullKey`) and gives line 95 the same value to re-assert after Hide:

```suggestion
    const concealedKey = `${fullKey.slice(0, 8)}${"•".repeat(8)}${fullKey.slice(-4)}`
    await expect(key).toHaveValue(concealedKey)
```

web/src/features/onboarding/SetupGuide.test.tsx
jigjigjig: Addressed in 9d08ca5e: dropped the empty-key case, leaving the two that exercise the guard, with a note on why the boundary is the one that matters.

web/e2e/dashboard.spec.ts
jigjigjig: Addressed in 9d08ca5e: the expected fingerprint is derived from the minted key, which retired the shape regex and the two lines under it. Swept the pattern rather than the file: `parity.management.spec.ts:177` carried the same regex, and since the secret is only known there after the reveal, it now pins the stand-in against it once it is.

web/src/features/onboarding/SetupGuide.test.tsx
```diff
@@ -300,13 +302,57 @@ describe("SetupGuide", () => {
...
+      "gw-setup••••••••-key",
+    )
+    expect(snippet("curl")).not.toHaveTextContent(KEY)
+    expect(snippet("curl")).toHaveTextContent("gw-setup••••••••-key")
+  })
+
+  // 15 characters is the boundary `concealedFingerprint`'s guard turns on; the
+  // mint never produces one this short, so the case is the guard's, not a state
+  // the sheet can reach.
+  it.each(["short-key", "123456789012345"])(
```
khaledosman: Two full guide renders down the identical branch. The comment above already names 15 as the boundary, so `"short-key"` adds nothing the boundary case does not, and `CopyField.test.tsx:683` already pins `concealedFingerprint("0123456789abcde")` directly.

Keep `"123456789012345"` as a plain `it` and drop the other. The one case still earns its place here: it proves `SetupSheet` reaches the guard through the helper rather than having re-inlined the slice arithmetic, which is what this PR exists to fix.

web/src/features/onboarding/SetupGuide.test.tsx
```diff
@@ -300,13 +302,57 @@ describe("SetupGuide", () => {
     const user = userEvent.setup()
     await renderGuide()
 
-    expect(await screen.findByLabelText("Your API key")).not.toHaveValue(KEY)
+    expect(await screen.findByLabelText("Your API key")).toHaveValue(
+      "gw-setup••••••••-key",
```
khaledosman: `"gw-setup••••••••-key"` is spelled six times in this file. Hoist it beside `const KEY` (line 25) as `const CONCEALED_KEY`, so changing the fixture key cannot leave five stale literals behind that still read as deliberate assertions.

web/e2e/parity.management.spec.ts
```diff
@@ -179,6 +179,12 @@ test.describe("api keys", () => {
     await reveal.getByRole("button", { name: "Show Secret key" }).click()
     const secret = await secretField.inputValue()
     expect(secret).not.toContain("•")
+    // The shape assertion above is all the concealed field can be held to
+    // before the secret is known; now that it is, the stand-in is pinned to the
+    // fingerprint of this key rather than of any key.
+    expect(masked).toBe(
```
khaledosman: The fingerprint arithmetic is now written out twice, here and at `dashboard.spec.ts:79`. Put it in `e2e/helpers.ts` as one function both specs call.

Spelling the formula rather than importing `concealedFingerprint` is the right call, since an import would make the assertion tautological against a bug in the helper itself, but the point of deriving it independently is lost once there are two derivations to keep in step.

khaledosman (approved):
The earlier findings are addressed: `SetupSheet` routes through `concealedFingerprint` rather than reimplementing it, and both specs now pin the stand-in to the key that was actually minted instead of to a shape any key would match. Three small test-hygiene points inline; none blocking.

web/src/features/onboarding/SetupGuide.test.tsx
jigjigjig: Addressed in f0c909d3: Kept a single plain test for the 15-character boundary and removed the redundant short-key case.

web/src/features/onboarding/SetupGuide.test.tsx
jigjigjig: Addressed in f0c909d3: Hoisted CONCEALED_KEY beside KEY and reused it throughout the setup-guide assertions.

web/e2e/parity.management.spec.ts
jigjigjig: Addressed in f0c909d3: Both specs now call expectedKeyFingerprint in e2e/helpers.ts, independent of the production helper.

PR #1169, opened 2026-09-15
fix(dashboard): remove activity chart tooltip lag

jigjigjig:
## Description

Hovering across the Activity graph now moves the popup directly to the inspected bucket. The shared trend chart previously used Recharts' default 400 ms position transition, making the popup trail the pointer. This also removes that lag from Usage charts using the same component.

Adds a regression test that hovers two buckets, checks their values and the absence of a movement transition, then verifies dismissal when leaving the chart. The test fails on the original behavior.

## How to test it locally

1. Run `pnpm --dir web run storybook`.
2. Open **Dashboard / Activity / ActivityTimeline / With Errors**.
3. Move the pointer between bars. The popup should follow immediately and disappear when leaving the chart.

Validated locally:

- 30 chart and ActivityTimeline tests pass.
- Dashboard Biome lint and TypeScript checks pass.
- Production dashboard build and `make lint` pass.
- Browser recording verifies hover behavior with the real ActivityTimeline component.
- Built gateway runs locally with 308 sample Activity events; the user reviewed the local environment.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes mozilla-ai/otari-ai#2092

## Checklist

Frontend-only change: relevant frontend checks and `make lint` were run; the full backend test/typecheck suites were not run. No documentation or API contract changes are needed.

## AI Usage

- [ ] No AI was used.

**Any additional AI details you'd like to share:** Implemented and tested with Codex; the user reviewed the working local demo and requested this PR.

## Summary

- Disabled the default tooltip movement animation in `TrendChart`.
- Updated Activity and usage charts to move directly between data buckets.
- Added regression coverage for tooltip values, bucket movement, and dismissal when leaving the chart.

This removes tooltip lag and provides a more responsive chart experience.


web/src/design-system/metrics/charts.tsx
```diff
@@ -304,6 +304,7 @@ export function TrendChart({
             />
           ) : null}
           <Tooltip
+            isAnimationActive={false}
```
khaledosman, line 307: Unrelated to this line, but in the file you are touching: the `ChartTooltip` docblock at `charts.tsx:70` says it is "Exported for direct branch testing since recharts hover is impractical to drive in jsdom", and the test added here drives recharts hover in jsdom and reads `$400` / `$840.5` out of `ChartTooltip` through it. Trim the clause to `Exported for direct branch testing.`; the export is still earned by the stack and zero-row branches, just not by that reason.

khaledosman (approved):
Verified the fix against recharts 3.10.1: `isAnimationActive` on `Tooltip` feeds only `resolveTransitionProperty` in `TooltipBoundingBox`, whose sole output is the `transform 400ms ease` transition, so the change is exactly scoped (no fade, no mount behavior). Also confirmed the new test is not vacuous: reverting the prop locally turns it red with `transition: transform 400ms ease` against `transition: ;`. One stale comment.

web/src/design-system/metrics/charts.tsx
jigjigjig: Addressed in d2089db6: trimmed the docblock to “Exported for direct branch testing.” The existing hover test confirms the removed jsdom limitation is stale. make lint and Biome passed.

PR #1170, opened 2026-09-15
fix(playground): improve design of playground

jigjigjig:
## Description

Improve the design of the playground.

## How to test it locally

1. Build the dashboard and start a standalone gateway with two chat models.
2. Open Playground from the top navigation (or the account menu on mobile). Confirm the old sidebar entry is absent and deployment gating still applies.
3. Send a starting prompt in Single mode. Check streaming, Copy, Regenerate, the usage readout, and the 760px composer. Hover Tools and the model picker to compare their states.
4. Switch to Compare. Model B starts empty. Enter and Send must not dispatch a request until B is selected. Choose B, send, and confirm both answers and rating controls appear.
5. Open History, filter Chats/Comparisons, search, and use View all. Resume a saved chat and confirm its model and Single mode are restored. Save a current chat and test confirmed deletion.
6. At 390px, check stacked comparison panels, fluid composer width, touch targets, and a history popover that stays inside the viewport. Use arrow keys on the mode switch and Escape on popovers.

## Validation

- Dashboard Biome and TypeScript: passed.
- Full dashboard suite: 6,479 tests passed across 193 files, and 6,478 after the review round removed the test for a prop that is gone.
- Production Vite and Storybook builds: passed.
- Storybook smoke: 850 of 858 renders passed. The same four unchanged PublicAuthLayout stories fail in both themes because `useTheme` lacks a ThemeProvider. Changed component stories passed.
- Local browser checks covered welcome, completed chat, paired streams, missing-model send guard, keyboard mode changes, history filtering/search, and 390px layout without horizontal overflow.

Frontend-only change. Backend suites were not rerun. Generated bundles and local demo data are not committed.

Two self-review rounds have been worked and every thread is resolved. The second round removed an unused `CopyButton` label mode, gave the History control a visible label so it takes the ghost edge rather than a wrapper drawing one around an icon-only button, lifted the history date heading out of its first list item, narrowed the composer's missing-model prop, and restored the reason both comparison columns start empty.

## PR Type

- [x] Bug Fix

## Relevant issues

Part of mozilla-ai/otari-ai#2086. Follow-up to #1131.

## Checklist

- [ ] I ran all backend Definition of Done checks locally (frontend-only change; validation above).

## AI Usage

## Summary

- Moved Playground from the sidebar to the top navigation. Added mobile access through the account menu.
- Redesigned the chat layout, composer, model picker, tools menu, response actions, and comparison mode.
- Added a shared history interface with search, filtering, loading, saving, and deletion support.
- Restored the saved conversation model and Single mode when loading a conversation.
- Prevented comparison sends until Model B is selected.
- Improved responsive behavior, accessibility labels, keyboard interactions, and popover behavior.
- Added shared layout and control components, including `TurnReadout` and expanded `Segmented` sizing.
- Removed the separate conversation and comparison history dialogs.
- Added or updated tests for navigation, history, comparison mode, popovers, reasoning blocks, and usage readouts.

## Technical notes

Validation covered frontend linting, TypeScript checks, tests, production builds, Storybook builds, Storybook smoke tests, and desktop/mobile browser checks. Backend suites were not rerun because the changes are frontend-only.


web/src/features/playground/PlaygroundToolbar.tsx
```diff
@@ -1,143 +1,80 @@
...
+      <Button
+        aria-label="Conversation history"
+        className="min-h-11 min-w-11"
+        onPress={onOpenHistory}
+      >
+        <FiClock aria-hidden className="size-4" />
+        <span className="hidden lg:inline">Conversation history</span>
+      </Button>
+      <Button
+        aria-label="Saved comparisons"
```
jigjigjig: The accessible name here is "Saved comparisons" while the visible label five lines down is "Comparison history". The accessible name has to contain the visible one (WCAG 2.5.3 Label in Name), or a voice-control user reading "Comparison history" off the screen cannot activate the button. Pick one string and use it for both.

web/src/features/playground/PlaygroundToolbar.tsx
```diff
@@ -1,143 +1,80 @@
...
+      >
+        <FiBarChart2 aria-hidden className="size-4" />
+        <span className="hidden lg:inline">Comparison history</span>
+      </Button>
+      {!isComparing && hasTranscript ? (
+        <Button
+          aria-label="Save conversation"
+          className="ml-auto min-h-11 min-w-11"
+          onPress={onSaveConversation}
+          isDisabled={isSaveDisabled && !isSavePending}
```
jigjigjig: `isDisabled` and `isPending` for the same condition is what `Button`'s docstring rules out: "it must not be paired with `isDisabled` for the same condition: two sources for one state is how a button ends up disabled with no spinner." The `&& !isSavePending` here only exists because `PlaygroundPage` folds `isSavePending` into `isSaveDisabled`. Drop it from that composition and pass `isDisabled={isSaveDisabled}`.

web/src/features/playground/ModelSelect.tsx
```diff
@@ -52,7 +52,7 @@ export function ModelSelect({
 
   const unavailable = new Set(unavailableKeys ?? [])
   const pinned = new Set(pinnedKeys)
-  const selectedLabel = models.find((model) => model.key === value)?.key ?? ""
+  const selectedLabel = models.find((model) => model.key === value)?.label ?? ""
```
jigjigjig: `label` is the part of the key after the colon, so two provider instances serving the same model id (`openai:gpt-4o` and a second OpenAI-compatible instance) now render an identical trigger with nothing to tell them apart. `MessageBubble` was given `splitModelKey` so it could show the instance beside the label; the picker that chose it no longer shows it at all.

web/src/features/playground/PlaygroundPage.tsx
```diff
@@ -72,40 +68,37 @@ export function PlaygroundPage() {
...
-    // pads every page, and it is auto-height, so a page cannot fill the
-    // viewport from inside it: the conversation grows the page, the shell's
-    // `<main>` scrolls it, and the composer sticks to the bottom of the
-    // viewport so it stays reachable however long the transcript gets. One
-    // scroll container, which is also what a phone wants.
-    <div className="flex flex-col">
+    <div className="flex min-h-[calc(100dvh-6rem)] flex-col md:min-h-[calc(100dvh-6.5rem)]">
+      <div className="relative">
+        <PageIntro title="Playground" />
+        <div className="absolute right-0 top-0">
```
jigjigjig: `PageIntro` has an `action` prop for exactly this, documented as "the one action that belongs beside rather than below" the title, and its header reflows to a column below `sm` so the action drops under the heading on a phone. Wrapping it in `relative` and absolutely positioning the button opts out of both. Pass the button as `action` instead.

web/src/features/playground/PlaygroundWelcome.tsx
```diff
@@ -33,27 +16,36 @@ export function PlaygroundWelcome({
...
+        <h2 className="text-display-sub">What can I help with?</h2>
+        <p className="text-body text-muted">
+          Test a prompt with your workspace’s models.
+        </p>
+      </div>
+      <PlaygroundComposer {...composerProps} />
+      {composerProps.canChat ? (
+        <div className="flex flex-col gap-2">
+          <p className="text-caption">Try a starting point</p>
+          <Toolbar className="flex-col items-stretch gap-0">
```
jigjigjig: `Toolbar` is a named place, not a flex row: `.otari-toolbar` declares `--field-height: 32px` and strips the border off every ghost button inside it. Used here as a vertical list, it strips the borders the design wants and the code puts them back with a per-item `border-b` wrapper. A plain `<ul>` of `<li>` is the container, and it also gives the list its semantics. Worth asking the same question of the composer control row and the message action row, which are also not filter rows above a table.

web/src/features/playground/PlaygroundPage.tsx
```diff
@@ -130,22 +123,8 @@ export function PlaygroundPage() {
...
-              so it keeps the shell's width and its left edge does not sit over
-              the sidebar. The negative bottom margin cancels the wrapper's
-              padding below it, so nothing shows between it and the edge. */}
-          <div className="-mb-5 sticky bottom-0 z-10 bg-background pt-2 pb-5 md:-mb-6 md:pb-6">
-            {/* Inside the sticky block, above the composer. In normal flow it
-                sat at the document's end and the composer stuck over it, so the
-                one control the comparison exists for was the one you could not
-                see. Above the composer is also where it reads: the verdict on
-                the last exchange, then the box for the next question. */}
+          <div className="-mb-5 sticky bottom-0 z-10 mx-auto mt-auto w-full max-w-[47.5rem] bg-background pt-6 pb-[max(1.25rem,env(safe-area-inset-bottom))] md:-mb-6 md:pb-6">
```
jigjigjig: `max-w-[47.5rem]` is spelled three times across this file, `PlaygroundConversation.tsx:51` and `PlaygroundWelcome.tsx:19`. It is one measure that all three have to agree on, so it should be one exported constant; three copies will drift the first time the column is retuned.

web/src/features/playground/PlaygroundToolbar.tsx
```diff
@@ -1,143 +1,80 @@
...
-            <IconButton
-              label="Save conversation"
-              onPress={onSaveConversation}
-              isDisabled={isSaveDisabled}
-              isPending={isSavePending}
-            >
-              <FiSave aria-hidden className="size-4" />
-            </IconButton>
-          </Tooltip>
+    <Toolbar className="-mx-4 min-h-15 border-y border-border px-4 py-2 md:-mx-6 md:px-6">
```
jigjigjig: `-mx-4 … md:-mx-6` hardcodes the shell content wrapper's padding so the band can bleed to the edge. Nothing ties the two together, so changing the shell's padding leaves this band inset or overhanging with no test failing. If a full-bleed band is the design, it wants a shared utility that reads the same value the wrapper does.

web/src/app/nav/TopBarActions.tsx
```diff
@@ -3,37 +3,30 @@ import { Link } from "@tanstack/react-router"
...
-// has no row for it; the menu keeps a row of its own anyway, because this
-// cluster is hidden below `md` and the guide would otherwise have no entry
-// point on a phone.
-//
-// The design also draws a balance at the end of the cluster, and this build has
-// none to draw: this gateway meters spend but holds no wallet. `WalletNavSlot`
-// is the seam a build that does hold one replaces to contribute the chip, so
-// the gap is reachable rather than something an overlay would have to edit this
-// file to fill. It renders nothing here.
+// Desktop destinations; the account menu keeps them reachable on mobile.
```
jigjigjig, line 10: This replacement drops the standing rationale, not narration. The deleted block was the only place saying what `WalletNavSlot` is and why it renders nothing here, and the seam is still in the file below. Three more in the same shape: `registry.ts:602` loses why Playground gates on `surface` rather than `capability`, which still governs `PLAYGROUND_NAV_ITEM`; `PlaygroundComposer.tsx:29` loses the documented exception for using a bare `<textarea>` over the shared `TextArea`, so the next reader will "fix" it; and `NAV_ITEMS` loses its doc entirely at the same moment it stops being a plain flatten of the three section lists. AGENTS.md asks for narration and implementation history to go, which most of this diff does correctly. The why is the part to keep.

jigjigjig (commented):
Self-review before this leaves draft. CodeRabbit skipped the PR ("Draft PR not reviewed"), so nothing external has read it yet.

Three correctness or accessibility items, four design-system items, and one about the comments the diff removes. Popover's nested-trigger fix and the `label="Copy response"` to `label="response"` correction (which was producing an accessible name of "Copy Copy response") are both right and both covered by new tests.

Reviewed by Claude Opus 5.

web/src/features/playground/ThinkingBlock.tsx
```diff
@@ -1,44 +1,11 @@
...
- * loses the one thing that explains a surprising reply. A `<details>`-shaped
- * disclosure rather than the shared `Disclosure` primitive because this sits
- * inside a message bubble and carries no heading, border or padding of its own,
- * which is most of what that component is.
- *
- * It renders while the reasoning is still streaming, so the label is present
- * from the first fragment: a control that appears halfway through a reply moves
- * the answer under the reader's eye.
- */
+/** Only renders reasoning returned by the provider. */
```
jigjigjig: The comment this replaces argued specifically against the shared `Disclosure`: "because this sits inside a message bubble and carries no heading, border or padding of its own, which is most of what that component is." The change adopts `Disclosure` anyway, which may well be right now that the transcript has visible headings and rules, but the recorded decision is reversed with nothing in its place. One line saying what changed about the surface.

jigjigjig (commented):
Self-review by Claude Opus 5. Five findings, none blocking the design direction: one dead prop on a shared component, one hand-rolled control border, one heading nested inside a list item, one over-wide union, and one rationale comment that lost the fact it existed to record.

web/src/design-system/actions/CopyButton.tsx
```diff
@@ -21,7 +22,10 @@ export function CopyButton({
   value,
   label,
   selectOnFailure,
+  showLabel = false,
```
jigjigjig: `showLabel` is dead in production. The only three uses in the tree are this file, `copy.stories.tsx`, and `CopyButton.test.tsx`: commit 4 added it for `MessageBubble`, and commit 6 reverted that call site to the icon-only form. What ships is a branch on a shared design-system component that nothing renders, carrying an `md` size, a 6.5rem fixed width, a `FiCheck` import, and a "Copied" text state.

Drop `showLabel`, `FiCheck`, the `LabeledButton` story, and the test that covers it, returning `CopyButton` to its pre-PR shape.

web/src/features/playground/PlaygroundHistory.tsx
```diff
@@ -0,0 +1,316 @@
...
+
+  return (
+    <>
+      <Popover
+        padding="none"
+        placement="bottom end"
+        isOpen={isOpen && !isAllOpen}
+        onOpenChange={onOpenChange}
+        trigger={
+          <span className="inline-flex size-11 shrink-0 border border-control-border md:size-9">
```
jigjigjig: This hand-rolls a control border outside the design system. `Button` defaults to `variant="ghost"`, which is edgeless on purpose, so wrapping it in a bordered `<span>` puts the chrome on a non-focusable element: the focus ring draws inside the border rather than around the control, and `size-full min-w-0` on the button overrides the size scale that `size` exists to set.

Ask the button for the size and the edge (or add an `.otari-*` container, which is what `web/design/actions.md` prescribes for a group of ghosts that needs edges) instead of drawing one around it.

web/src/features/playground/PlaygroundHistory.tsx
```diff
@@ -0,0 +1,316 @@
...
+          {search ? "No matching history." : "No saved history yet."}
+        </p>
+      ) : (
+        <ul>
+          {visible.map((entry, index) => (
+            <li key={`${entry.kind}:${entry.id}`}>
+              {index === 0 ||
+              dateGroup(visible[index - 1].createdAt) !==
+                dateGroup(entry.createdAt) ? (
+                <h3 className="border-t border-border px-4 py-2 text-overline">
```
jigjigjig: The date heading is nested inside the first entry's `<li>`, so it labels a group from inside one of that group's members. A screen reader reading the list hears "Today" as part of the first row rather than as the heading over the rows that follow.

Group the entries first, then render one `<h3>` followed by its own `<ul>` per group.

web/src/features/playground/PlaygroundComposer.tsx
```diff
@@ -1,13 +1,16 @@
...
-import { IconButton } from "@/design-system/actions/IconButton"
+import { Button } from "@/design-system/actions/Button"
 
 import { ActiveToolChips } from "./ActiveToolChips"
 import { ToolsMenu } from "./ToolsMenu"
 
 export interface ComposerProps {
+  modelPicker?: ReactNode
+  hasTranscript?: boolean
+  missingModel?: "A" | "B"
```
jigjigjig: `missingModel` is typed `"A" | "B"` and only ever receives `"B"`: `PlaygroundPage` passes it when `isComparing && !panelB.model`, and panel A always carries the remembered or default model. The `"A"` half is a branch no call site can reach, and the copy it produces ("Choose model A to send") has never been seen.

Narrow it to `"B"`, or make it a boolean the composer names itself.

web/src/features/playground/hooks/usePlayground.ts
```diff
@@ -393,17 +398,9 @@ export function usePlayground() {
...
-      // Both columns start empty, and this is the one place the port departs
-      // from the hosted original, which kept the first panel's transcript. It
-      // has to: the two models are sent their own panel's history, so a column
-      // carrying an earlier conversation is answering a different prompt from
-      // the one beside it. The comparison would not be a comparison, and the
-      // rating it records is a judgment over an unequal contest, which is worse
-      // than losing an on-screen transcript nobody asked to keep. A saved one
-      // is still in the history.
-      const other = models.find((model) => model.key !== panelA.model)
+      // Both models need the same empty history for a fair comparison.
```
jigjigjig: This replaced a paragraph that recorded a decision, not a description. The old comment said that clearing both panels is where this port deliberately departs from the hosted original, which kept panel A's transcript, and why: a column carrying an earlier conversation answers a different prompt from the one beside it, so the rating would be a judgment over an unequal contest. "Both models need the same empty history for a fair comparison" is the conclusion with the departure and the reasoning removed, which is the part a reader could not derive from the code.

Pruning narration is the house rule and most of the removals in this PR are right; this one took the fact with it. Restore the departure and its reason in two lines.

web/src/design-system/actions/CopyButton.tsx
jigjigjig: Addressed in 714d3f1f: `showLabel` and its FiCheck branch, story and test are gone; CopyButton is back to the icon-only control it was, still on the design-system Button.

web/src/features/playground/PlaygroundHistory.tsx
jigjigjig: Addressed in 714d3f1f: the wrapper span is gone and the control is a labeled ghost (clock plus "History"), which takes the edge natively. Chosen over an edgeless icon button so the header still draws the box Paper shows; the control now reads about 110px wide at desktop.

web/src/features/playground/PlaygroundHistory.tsx
jigjigjig: Addressed in 714d3f1f: entries are grouped first, then each group renders one h3 followed by its own ul. Same borders and padding, so nothing moved.

web/src/features/playground/PlaygroundComposer.tsx
jigjigjig: Addressed in 714d3f1f: narrowed to `"B"`, and the hint string it feeds is now computed once instead of written at both breakpoints.

web/src/features/playground/hooks/usePlayground.ts
jigjigjig: Addressed in 714d3f1f: the departure from the hosted original and its reason are back, in five lines rather than the original eight.

web/src/features/playground/PlaygroundToolbar.tsx
jigjigjig: Addressed in aa05a230, then removed entirely in 06265065: the two history buttons became one History control, so there is no second name to disagree with.

web/src/features/playground/PlaygroundToolbar.tsx
jigjigjig: Addressed in aa05a230: `PlaygroundPage` no longer folds `isSavePending` into `isSaveDisabled`, and the Save control passes `isDisabled` and `isPending` from separate conditions.

web/src/features/playground/ModelSelect.tsx
jigjigjig: Addressed in aa05a230: `ModelSelect` falls back to the full `instance:model` key when another model in the catalog carries the same label, covered by "identifies the provider when model labels are shared".

web/src/features/playground/PlaygroundPage.tsx
jigjigjig: Addressed in aa05a230: the button is passed as `PageIntro`s `action`, and the `relative`/`absolute` wrapper is gone.

web/src/features/playground/PlaygroundWelcome.tsx
jigjigjig: Addressed in aa05a230 and 06265065: the starting points are a plain `ul`/`li`, and the composer control row and message action row are `.otari-actions` rather than `Toolbar`.

web/src/features/playground/PlaygroundPage.tsx
jigjigjig: Addressed in aa05a230: the measure is `CHAT_COLUMN` in `playgroundLayout.ts`, and no literal `47.5rem` is left in `web/src`.

web/src/features/playground/PlaygroundToolbar.tsx
jigjigjig: Addressed in aa05a230 and 06265065: the band is a `Section`, which owns the wrapper padding, and the hardcoded `-mx-4 md:-mx-6` is gone.

web/src/app/nav/TopBarActions.tsx
jigjigjig: Addressed in aa05a230: all four carry their why again, in a line each. The fifth in the same shape, `toggleCompare`, was caught in the next round and restored in 714d3f1f.

web/src/features/playground/ThinkingBlock.tsx
jigjigjig: Addressed in aa05a230: the docstring now says what changed about the surface, that reasoning is a titled section in a flat transcript, which is what `Disclosure` draws.

web/src/features/playground/hooks/usePlayground.ts
```diff
@@ -359,14 +364,17 @@ export function usePlayground() {
       setLoadError(error)
       return
     }
-    setPanelA((prev) => ({
-      ...prev,
+    setIsComparing(false)
+    setPanelB(EMPTY_PANEL)
+    setPanelA({
+      ...EMPTY_PANEL,
+      model: saved.model,
```
khaledosman: `saved.model` is not checked against the catalog. Load a chat whose model has since left `/models` and `canChat` (line 464) is still true, the picker renders "Choose a model" because `models.find` misses, and Send fails at the gateway with nothing on screen saying why.

The seeding effect above already has the guard; reuse it:

```ts
model: models.some((m) => m.key === saved.model)
  ? saved.model
  : pickInitialModel(undefined, models),
```

web/src/features/playground/PlaygroundComposer.tsx
```diff
@@ -61,9 +52,15 @@ export function PlaygroundComposer({
...
+  // Narrow enough to fit beside the controls from `lg` up, and under the frame
+  // below it, so the two nodes say the same thing at one width each.
+  const hint = missingModel
+    ? `Choose model ${missingModel} to send`
+    : "Enter to send · Shift + Enter for a new line"
+
   return (
     <form onSubmit={onSubmit} className="w-full">
-      <div className="flex flex-col gap-2 rounded-3xl border border-border bg-surface p-2.5 transition-colors focus-within:border-accent">
+      <div className="flex flex-col gap-2 border border-[var(--field-border)] bg-[var(--field-background)] px-3 pt-3 pb-2 has-[textarea:focus-visible]:otari-focus-ring">
```
khaledosman: `--field-border` / `--field-background` are HeroUI's variables, not `--color-*` tokens, and this is the first call site in `web/src` to reach them through an arbitrary class. frontend-standards #2: "an alias for one of those goes into `globals.css` rather than into a class string of ours".

`border` also hardcodes 1px where `--field-border-width` is the declared token. Either use `bg-surface border-border`, or add a composer place in `globals.css` alongside the other `.otari-*` places and let it read the field variables.

web/src/features/playground/PlaygroundHistory.tsx
```diff
@@ -0,0 +1,326 @@
...
+            </ul>
+          </section>
+        ))
+      )}
+    </div>
+  )
+  const save = onSave ? (
+    <Button
+      size="sm"
+      aria-label="Save conversation"
```
khaledosman: Accessible name is "Save conversation", visible label is "Save current chat". The name does not contain the visible text, so speech input ("click Save current chat") misses the button (WCAG 2.5.3 Label in Name).

Drop the `aria-label` and let the text be the name, then update the two `getByRole("button", { name: "Save conversation" })` queries in `PlaygroundPage.test.tsx`.

web/src/design-system/overlays/Popover.tsx
```diff
@@ -28,21 +29,23 @@ export function Popover({
...
+  padding?: "default" | "none"
   isOpen?: boolean
   onOpenChange?: (isOpen: boolean) => void
 }) {
   return (
     <HeroPopover.Root isOpen={isOpen} onOpenChange={onOpenChange}>
-      <HeroPopover.Trigger className="inline-flex">
-        {trigger}
-      </HeroPopover.Trigger>
+      {trigger}
```
khaledosman, line 54: Dropping `HeroPopover.Trigger` fixes the duplicate-button problem `design/overlays.md` describes for Tooltip, but it makes the `trigger` contract implicit: `ReactNode` now has to be a react-aria pressable, and a `<span>` or a native `<button>` silently never opens the panel.

This is a published primitive with a story catalog. Narrow the prop type or state the constraint in its docstring, and record the rule in `design/overlays.md`'s Popover section, which currently documents only the padding and placement additions.

web/src/features/playground/PlaygroundHistory.tsx
```diff
@@ -0,0 +1,326 @@
...
+}
+
+function dateGroup(value: string) {
+  const date = new Date(value)
+  const today = new Date()
+  const yesterday = new Date(today)
+  yesterday.setDate(today.getDate() - 1)
+  if (date.toDateString() === today.toDateString()) return "Today"
+  if (date.toDateString() === yesterday.toDateString()) return "Yesterday"
+  return date.toLocaleDateString(undefined, {
```
khaledosman: Hand-rolled `toLocaleDateString`. `shared/helpers/format.ts` owns date formatting (frontend-standards, formatting-and-i18n.md), and both dialogs this replaces used `formatDateTime`.

`new Date()` is also read at render, so a popover left open across midnight keeps labelling yesterday's rows "Today". Put the relative grouping in `format.ts` and call it from here.

web/src/features/playground/MessageBubble.tsx
```diff
@@ -108,32 +108,31 @@ export function MessageBubble({
...
-        </div>
+        <ErrorBanner error={new Error(turn.errorMessage)} />
       ) : null}
+      <div className="flex flex-wrap items-center justify-between gap-x-4 gap-y-2 pt-1">
+        {areActionsVisible && (response || turn.errorMessage) ? (
+          <div className="otari-actions flex shrink-0 items-center gap-3">
+            {response ? <CopyButton value={response} label="response" /> : null}
+            {onRegenerate ? (
+              <Button
+                isIconOnly
```
khaledosman, line 119: `Button isIconOnly` plus a hand-written `min-h-11 min-w-11` rebuilds `IconButton`, which exists to make the accessible name a required prop rather than an optional `aria-label` and to put the 44px box on the component instead of in a class string. Same at `PlaygroundComposer.tsx:112`/`124` and `PlaygroundHistory.tsx:212`.

`IconButton` takes a `className`, so the `md:` step down is still available:

```tsx
<IconButton label="Regenerate response" size="sm" className="md:min-h-8 md:min-w-8">
```

The page's own "Scroll to latest" still uses `IconButton`, so as it stands one feature spells icon buttons two ways.

web/src/features/playground/PlaygroundHistory.tsx
```diff
@@ -0,0 +1,326 @@
...
+                      className="flex min-h-11 min-w-0 flex-1 flex-col justify-center gap-0.5 text-left"
+                      onClick={async () => {
+                        await onLoad(entry.id)
+                        setIsAllOpen(false)
+                      }}
+                    >
+                      <span className="w-full truncate text-emphasis">
+                        {entry.title}
+                      </span>
+                      <span className="w-full truncate text-mono-caption text-subtle">
```
khaledosman: `text-mono-caption` is the identifier role. This line is "gpt-4o · 2 messages" and the one at 204 carries the saved question, which is prose. Use `text-caption`, and keep mono for a bare model key if it needs it.

.gitignore
```diff
@@ -5,7 +5,9 @@ __pycache__/
 .mypy_cache/
 .pytest_cache/
 .ruff_cache/
-.venv/
+# No trailing slash: a worktree may symlink this at another checkout's
+# environment, and a directory pattern does not match a symlink.
+.venv
```
khaledosman: Unrelated to a playground redesign and unmentioned in the PR body. Split it out: a frontend reviewer has no context for a worktree/symlink change, and it makes this PR's own diff harder to read.

khaledosman (commented):
Two findings land on files this PR did not touch, so they are here rather than inline:

- `web/e2e/screenshots/authenticated.spec.ts:26` still gates the `/playground` capture on `heading: /what can i help with/i`. That greeting is now "Try a prompt." and the page grew a `PageIntro` title, so `open()` times out and the playground baseline never renders. The comment above the entry ("the one page in this matrix with no page title") is stale for the same reason; `heading: /playground/i` is now the right regex.
- `docs/dashboard.md:88` still lists Playground in the workspace sidebar ("Playground, Models, and Routing"). `web/AGENTS.md` and `.github/instructions/frontend-standards.instructions.md` were updated for the move; this one was missed.

Review written by Claude Opus 5 acting as an agent on khaledosman's behalf.

web/src/features/playground/hooks/usePlayground.ts
jigjigjig: Fixed in 1d2ed8d8. Routed through `pickInitialModel(saved.model, models)` rather than a fresh `.some` check, since that helper already is the guard (returns the key when the catalog serves it, `models[0]` otherwise). New test: "falls back to a servable model when the saved one is gone". Also removed a duplicate `setIsHistoryOpen(false)` spotted in the same function.

web/src/features/playground/PlaygroundComposer.tsx
jigjigjig: Fixed in 1d2ed8d8. Took the second option: `.otari-composer` in globals.css reads `--field-background`, `--field-border` and `--field-border-width`, so the width comes from the token too and the pixels are unchanged. Swept `web/src` for other arbitrary `var(--field-*)` classes; this was the only one.

web/src/features/playground/PlaygroundHistory.tsx
jigjigjig: Fixed in 1d2ed8d8. Dropped the `aria-label` so "Save current chat" is the name, and updated the four (not two) `getByRole` queries in PlaygroundPage.test.tsx.

web/src/design-system/overlays/Popover.tsx
jigjigjig: Fixed in 1d2ed8d8. `trigger` is now `ReactElement`, the docstring states it must be a react-aria pressable and names what fails silently, and design/overlays.md carries both that and the `label` rule. The type cannot express "pressable", so the constraint is written rather than enforced.

web/src/features/playground/PlaygroundHistory.tsx
jigjigjig: Fixed in 1d2ed8d8. Now `formatDateGroup` in `shared/helpers/format.ts`, taking `now` as a parameter. On the staleness: I did not add a timer, so a list left open across midnight with no rerender still shows the old headings. What changed is that the clock is read once per render pass instead of per row, so one list is never measured against two clocks, and a caller that wants to pin or refresh it now can. Say the word if the timer is worth it.

web/src/features/playground/MessageBubble.tsx
jigjigjig: Fixed in 1d2ed8d8 at all four sites (MessageBubble, the composer Stop and Send, the history delete). Kept `isIconOnly` on each: without it globals.css gives an `sm` ghost 11px of side padding, so the desktop controls would have gone from 32px to 38px wide. `IconButton` passes it through, so the accessible name is required and the 44px box is the component`s, with no pixel change.

web/src/features/playground/PlaygroundHistory.tsx
jigjigjig: Fixed in 1d2ed8d8: both lines are `text-caption` now. Mono stays on the model-key column in the comparison header and on TurnReadout, which are identifiers.

.gitignore
jigjigjig: Reverted in 1d2ed8d8; the file matches main again. Worth knowing why it was there: my own `git add -A` in this worktree had committed the `.venv` and `web/node_modules` symlinks, which is what failed four CI jobs at install earlier, and the trailing-slash patterns are why they were untracked rather than ignored. #1216 already fixes `node_modules` in both ignore files. `.venv` is not covered by it or by anything else, so that one still needs a home; I have not opened a PR for it.

web/src/design-system/overlays/Popover.stories.tsx
```diff
@@ -87,3 +87,23 @@ export const Controlled: Story = {
...
+  render: () => (
+    <Popover
+      padding="none"
+      placement="bottom end"
+      trigger={<Button>History</Button>}
+    >
+      <div className="w-72 max-w-[calc(100vw-2rem)]">
+        <h2 className="border-b border-border px-4 py-3 text-heading">
+          History
+        </h2>
```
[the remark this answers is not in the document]
jigjigjig: Fixed in 1d2ed8d8, with the prop required rather than optional: `label` has no sensible default and an optional one is how a panel ships unnamed, which is the failure being described. All five call sites pass it (History, Tools, the model picker, and both stories).

web/src/features/playground/ModelSelect.tsx
```diff
@@ -46,13 +47,20 @@ export function ModelSelect({
...
   const unavailable = new Set(unavailableKeys ?? [])
   const pinned = new Set(pinnedKeys)
-  const selectedLabel = models.find((model) => model.key === value)?.key ?? ""
+  const selected = models.find((model) => model.key === value)
+  const isAmbiguous =
+    selected &&
+    models.some(
+      (model) => model.key !== value && model.label === selected.label,
+    )
+  const selectedLabel = isAmbiguous ? value : (selected?.label ?? "")
```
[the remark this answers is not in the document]
jigjigjig: Not applied, because the premise is now fixed upstream of it. `loadConversation` was the only path that could put a key the catalog does not serve into `panelA.model`, and 1d2ed8d8 routes it through `pickInitialModel`, so the picker and the request agree again. Showing a raw unservable key would have made the label honest while leaving Send pointed at a model the gateway cannot answer for, which is the half worth fixing.

web/src/features/playground/PlaygroundConversation.tsx
```diff
@@ -46,60 +29,77 @@ export function PlaygroundConversation({
...
+                value={panel.model}
+                models={models}
+                pinnedKeys={pinnedKeys}
+                onTogglePin={onTogglePin}
+                unavailableKeys={other.model ? [other.model] : undefined}
+                onChange={(key) => {
+                  setPanel((prev) => ({ ...prev, model: key }))
+                  if (name === "A" && key === panelB.model)
+                    setPanelB((prev) => ({ ...prev, model: "" }))
+                }}
```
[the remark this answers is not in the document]
jigjigjig: Fixed in 1d2ed8d8, and widened slightly: a panel drops its turns whenever its own model changes, not only on the collision, since turns belong to the model that produced them either way. The collision branch clears B`s turns as suggested.

PR #1186, opened 2026-09-15
fix(clipboard): restore cleanup when legacyCopy's select() throws

### LINKED ISSUE #1149: legacyCopy can leave a textarea holding a credential in the DOM
## What

`legacyCopy` in `web/src/design-system/helpers/clipboard.ts` builds a `<textarea>` holding the plaintext it is copying, appends it to `document.body`, and removes it again several statements later. Only the `execCommand` call sits in a `try`; the setup between the append and the removal does not:

```ts
document.body.appendChild(source)          // :36  textarea now holds the plaintext
const selection = document.getSelection()  // :38
const previousFocus = ...                  // :41
source.select()                            // :45  not guarded
let copied = false
try {
  copied = document.execCommand("copy")    // :49  guarded
} catch {
  copied = false
}
source.remove()                            // :54  only reached on the happy path
```

If anything between `:36` and `:54` throws, `source.remove()` never runs and a detached-but-attached textarea containing the value stays in the DOM for the life of the page. `document.getSelection()` and `select()` are the realistic candidates; both are documented to throw in some cross-document and detached-frame situations.

## Why it matters more than it looks

The values this helper copies are credentials. `CopyField` hands out a one-time API key and the two request snippets built around it, so the leaked node holds a secret, and it is in the DOM rather than in a variable: any extension, any `document.body.innerText`, any later error report that serializes the page picks it up.

## Why now

The helper and the missing `finally` both pre-date #1142 and nothing about them changed in it. What changed is the exposure. Before, `legacyCopy` was reached only by `CopyField`'s concealed path; #1142 routed the default arrangement through the same helper so a copy works on a plain-HTTP origin (#957), which makes it the fallback for **every** copy in that component. The bug is the same age; the number of ways to reach it is not.

## What would close this

Wrap the body in `try { ... } finally { ... }` so the removal and the two restorations run whatever happens, keeping the existing `execCommand` catch inside it. A test that makes `select()` throw and asserts `document.body` holds no textarea afterwards is what turns it into a check rather than a claim.

Found in review on #1142 (https://github.com/mozilla-ai/otari/pull/1142#discussion_r4006378016), deliberately not fixed there: the file is untouched by that PR and a cleanup with its own test belongs in its own change.

AmirF194:
## Description
`legacyCopy` (the execCommand fallback used on plain-HTTP LAN origins, where `navigator.clipboard` is undefined) builds an offscreen textarea holding the copied text, often an API key since `CopyField` routes through this helper. It calls `document.getSelection()`, reads `document.activeElement`, and calls `source.select()` before entering the only try/catch in the function, which guards `execCommand` alone. `source.remove()` runs after that, on the happy path only. `select()` is documented to throw in a detached-document/cross-document situation, and if it does, the textarea is never removed: the plaintext stays in the DOM for the life of the page.

Moved the removal, and the selection/focus restore that already existed to protect the operator's tab position, into a `finally` so they run regardless of what throws once the textarea is in the DOM.

## How to test it locally
`pnpm --dir web vitest run src/design-system/helpers/clipboard.test.ts`. The new test stubs `HTMLTextAreaElement.prototype.select` to throw and asserts no `<textarea>` is left behind; it fails on `main` (1 leaked node) and passes on this branch.

## PR Type
- [x] Bug Fix

## Relevant issues
Fixes #1149

## Checklist

## AI Usage

## Summary

- Ensured `legacyCopy` always removes its temporary textarea and restores selection and focus.
- Made `select()` failures return `false` instead of escaping.
- Added a regression test for cleanup when `select()` throws.

This prevents copied credentials from remaining in the document after setup failures.


khaledosman (approved):
Approved, with one commit pushed onto the branch (2ff99e3f) rather than sent back, since `maintainerCanModify` is on.

The `finally` fixes the leak, but `legacyCopy` still throws and `copyToClipboard` does not catch it. All four call sites read only the return value and none wraps the call, so a `select()` that throws gave the operator nothing at all on the exact path this PR is about:

- `CopyButton.tsx:45` never reaches `setState("failed")`, so no "Copy failed" tooltip
- `CopyField.tsx:278` and `:298` never focus and select the field for Ctrl/Cmd-C
- `CodeBlock.tsx:74` is silent

plus an unhandled rejection out of each press handler. The commit turns the outer `try` into `try/catch/finally` returning `false`, which is the helper's stated contract ("reporting whether it actually landed there", `clipboard.ts:1-2`). With a catch covering what follows, the selection and focus capture moves back out of the `try`, dropping the three mutable bindings it needed.

On the fallback itself, since it comes up: `navigator.clipboard` is `[SecureContext]`, so it is undefined on the plain-HTTP LAN origins this dashboard is routinely served from (#957, #478). That is not a legacy-browser gap that ages out, so `execCommand` stays.

Verified on the branch: `pnpm --dir web run lint`, `pnpm --dir web run typecheck`, and the full vitest suite (194 files, 6549 tests) pass. The new test is a real regression test, not a claim: against `origin/main`'s `clipboard.ts` it fails with exactly one leaked textarea.

AmirF194:
Thanks for catching that, @khaledosman. Missing the catch around legacyCopy meant the finally fix alone left every call site with a silent failure path, and your fix closes it correctly.

Issue #1173, opened 2026-09-15
Wire core features through a static registry, so the bootstrap hook is only ever an edition's


peteski22:
### Problem

Adding a core feature means editing a manually maintained list in each of several files: the router list, the surfaces tuple, the settings view, the lifespan workers, the metrics module, plus the shared config, models and errors files. #973 touched twelve files for one feature.

#1091 avoids that by shipping budget alerts as a bootstrap plugin: `OTARI_BOOTSTRAP=otari_alerts:register`. That misuses the hook. `OTARI_BOOTSTRAP` selects an edition. It overrides ports and adds edition-specific things. Using it to switch on a core feature costs three things:

- The feature is off in a plain deployment unless the operator sets a specific environment variable.
- The hook takes one value, and an edition already uses it. The edition would have to import the package and register it again, or the hook becomes an ordered list where the last bind wins. Both are error-prone.
- A core feature gets a separate migration chain and a rule against foreign keys to core tables. Those only make sense for out-of-tree code.

### Proposal

A static registry of core features, read by core's own wiring. A feature is a domain in the shape from #1201 (a service package, a repository package, and one module each for routes, schemas, exceptions and models) plus one registry entry. The container and the bootstrap are not involved.

```python
# src/gateway/features.py
from gateway.services import alerts, budget_alert_trigger

CORE_FEATURES = (alerts.FEATURE, budget_alert_trigger.FEATURE)
```

Each entry carries a name, a dashboard surface, an `enabled(config)` check, the routers, and an optional worker.

- `register_routers` mounts each enabled feature's routers as core routes, ungated, where the other management routers mount. This is what ARCHITECTURE.md step 6 already says a core router does.
- The lifespan starts each enabled feature's worker beside its existing periodic tasks and stops it the same way, under the same bounded wait.
- The bootstrap endpoint lists a feature's surface only when the feature is enabled. The dashboard already hides a nav entry whose surface is absent and answers its route with "not available here". No new UI mechanism.
- Tables go on core's metadata and core's Alembic chain. One chain, one head.

**Code says what ships. Config says what runs.** The registry is the only thing that decides which features are in the build. Each feature reads its own settings and decides whether to mount and start. The switch is a startup-only config field, like `enable_metrics`. It is not an entitlement and not a feature flag. Config never names a module to load.

**The switch never touches the schema.** Off means the router is not mounted and the worker is not started. The tables and rows stay. On again does nothing to the database. Off, on, off, on is a no-op.

[...]

Issue #1171, opened 2026-09-15
Core cleanup: one feature, one domain


peteski22:
### Problem

Adding a feature to the core means editing a manually maintained list in several shared files. #973 touched twelve of them for one feature.

Those same files are where people collide. In the last 90 days, four files had ten or more distinct authors. Nobody owns any of them in CODEOWNERS.

### Goal

A feature lives in one domain with one owner. It is a registry entry plus the domain's modules in each layer: a service package and a repository package, and one module each for routes, schemas, exceptions and models (#1201). Adding one touches those modules, the registry entry, the nav registry, and generated artifacts. Nothing else. The domain map is #1201.

### Plan

Five phases. Each issue lands on its own and leaves the codebase working. Code moves and behavior changes are separate PRs.

1. Guardrails: required code-owner review on `main`, and the two stable paths added to CODEOWNERS.
2. Registries for the core's own lists: features, workers, surfaces, settings view, metrics.
3. Split the shared files by domain: config, models, errors, route helpers.
4. A built-in tool seam: one interface, one loop, web fetch as the proof.
5. Layer completion per domain: each service built on its own repositories and a Unit of Work, budget reservations as Try-Confirm/Cancel, the boundary enforced, then CODEOWNERS by domain.

Owners by area come last on purpose: before the domain issues land, they would change on every rename. Phases 3 to 5 are filed as the earlier phases land.

### Out of scope

Any new extension mechanism: loading code at runtime, finding packages by scanning, or hooks on the request path. Any change to a settled row of the architecture charter. Those go to the charter's Decisions table first.

### Numbers to move

Measured on `main` at `7e05eff2`, 2026-09-15.

| Measure | Today |
| --- | --- |
| Shared core files one feature PR touches (#973) | 12 |
| Files with 10 or more authors in 90 days | 4 |
| `_pipeline.py` | 4,762 lines, 24 importing modules, 31 public functions |
| `config.py` / `entities.py` / `tenancy/errors.py` | 2,632 / 1,974 / 1,294 lines |
| Flat modules at the top of `services/` | 62 |
| Underscore helper modules in `api/routes/` | 11 |
| Manually declared worker variables in the lifespan | 9 |
| CODEOWNERS entries | 3 |

### Done when

- A feature PR of #973's size touches its domain's modules, the registry entry, the nav registry and generated artifacts only.
- No core file has more than five distinct authors in a quarter.

### Issues

- [ ] #1172 Require code-owner review on main for the guardrail paths
[...]

PR #1188, opened 2026-09-15
refactor(gateway): add the core feature registry, empty

peteski22:
## Description

Today, adding a core feature to Otari means editing a hand-kept list in several files: the router list, the surfaces the dashboard shows, the background workers the lifespan starts. Shipping a feature as a bootstrap plugin instead misuses the bootstrap hook, which exists to select an edition and takes one value.

This PR adds a registry of core features: one literal tuple, edited by hand, that says which features this build ships. The app asks each listed feature once, when it is built, whether it is enabled, and router registration, the lifespan and the deployment bootstrap all use that one answer. A feature's switch is therefore a startup setting the dashboard cannot change. A listed feature mounts as core routes with no capability gate, runs its worker beside the existing refreshers, and publishes its dashboard surface when its own setting enables it. A worker stops under the refreshers' shared shutdown bound, and `CoreFeature` states what that asks of it: let cancellation through and leave no write half done. A feature is one domain-named module per layer plus its registry entry, the same shape the rest of the core uses.

The registry is empty in this PR, so nothing changes for anyone running Otari. It is the seam the first feature, budget alerts, lands in next (PR 2 of 3 for #1173).

The architecture check learns two rules so the seam stays what it is: a service or a route may not import the registry, so a feature cannot register itself (only the app wiring reads it, and the bootstrap route gets the enabled features from the app); and nothing under the gateway imports `importlib.metadata`, `importlib_metadata` or `pkg_resources`, which is how entry-point discovery gets written. A violation of the second says that in its message.

Churn on the two files this touches most, measured over 90 days as the epic asks: `main.py` goes from 47 commits to 48; the rest of the top five is unchanged.

## How to test it locally

```
make lint && make typecheck
uv run pytest tests/unit/test_features.py tests/unit/test_check_architecture.py tests/unit/test_deployment_bootstrap.py
uv run python -c "import gateway.features, gateway.api.main, gateway.main"
```

What to look for: `GET /api/v1/bootstrap` on a standalone gateway answers the same surface list as before. The new `test_features.py` stands a probe feature in for a real one and checks that its route mounts, its surface is published, and its worker starts, stops, and reports a failure when it dies, each only when the feature says it is enabled. It also checks that a switch changed after the app is built moves none of the three, that a surface is published once even when a feature repeats it, and that a worker that will not stop is abandoned at shutdown and logged as a worker. The check-script tests cover the two new rules with synthetic files.

Run locally: lint, typecheck, `make openapi-check`, `make postman-check`, and the full unit suite (3628 passed). The 16 local failures come from this machine, not the change: a gitignored `.env` sets a master key and the shell exports `OTARI_API_KEY`. The six affected files pass 145 of 145 in a clean checkout without either. The OSS smoke gate and the integration suite are left to CI.

## PR Type

- [ ] New Feature
- [ ] Bug Fix
- [x] Refactor
- [x] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Part of #1173 (PR 1 of 3). Part of #1171.

## Checklist

- [ ] I ran the Definition of Done checks locally (`make lint`, `make typecheck`, `make test`). Lint, typecheck, and `make test-unit` ran locally; `make test-integration` is left to CI.
- [x] If the API contract changed, I regenerated the OpenAPI spec (`uv run python scripts/generate_openapi.py`). Not applicable: no route or schema changed.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:** Claude Fable 5.1 and Claude Opus 5, through Claude Code.

**Any additional AI details you'd like to share:** The design and the step-by-step spec were written by Peter. Agents implemented the spec, wrote the tests, and ran the checks. The later commits apply review feedback: daavoo's comment on reading `enabled` once, and a multi-model review whose findings Peter asked to have fixed.

## Summary

- Added an explicit, currently empty core feature registry.
- Integrated enabled features with route mounting, dashboard surfaces, and worker lifecycle management.
- Added architecture rules that protect registry boundaries and prevent runtime feature discovery.
- Added tests for feature integration, hybrid deployments, worker handling, startup settings, and architecture rules.

This provides a controlled extension point for future core features without changing current deployments.

## Technical notes

- Features use `CoreFeature` definitions in `src/gateway/core/feature.py`.
- The application evaluates feature settings once during startup.
- Hybrid deployments do not mount feature routes, publish feature surfaces, or start feature workers.
- Review findings: unavailable.
- Test execution results: unavailable.


src/gateway/main.py
```diff
@@ -475,6 +477,14 @@ async def lifespan(app: FastAPI) -> AsyncGenerator[None, None]:
                         retention_sec=config.budget_reservation_retention_sec,
                     )
                 )
+            # Workers of the feature packages the registry lists. Same
+            # supervisor as the refreshers above: created here, cancelled
+            # together in ``finally`` under one shared bound.
+            package_workers = [
+                (asyncio.create_task(package.worker(config)), package.name)
```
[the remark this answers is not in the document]
peteski22: Addressed in f61c3dbb: each package worker now runs under a wrapper that logs the failure once, when it happens. It does not re-raise, since nothing awaits the task before shutdown and the supervisor would only log the same death a second time there.

src/gateway/api/routes/bootstrap.py
```diff
@@ -143,6 +144,21 @@
 )
 
 
+def hosted_surfaces(config: GatewayConfig) -> list[str]:
+    """The surfaces this deployment hosts, sorted.
+
+    The edition's fixed set plus the surface of each enabled registry feature.
+    Empty for hybrid, which hosts none.
+    """
+    if config.is_hybrid_mode:
```
daavoo, line 153: This branch is unreachable in production. `hosted_surfaces` has one caller, line 363, and that sits below the hybrid early return at line 318, which builds its own response with a literal `surfaces=[]`.

So the rule `test_a_hybrid_gateway_hosts_no_surface_whatever_the_registry_says` states is real, but it is the literal that enforces it, not this branch: the test asserts a property of a helper the hybrid path never calls, and no test covers what the hybrid endpoint answers with an enabled feature in the registry.

Suggest calling the helper from the hybrid arm of `get_bootstrap` instead of its literal `surfaces=[]`. Then one function answers which surfaces a deployment publishes in all three modes, this branch carries weight, and the test covers what ships. That also makes the name wrong: a function answering for hybrid, hosted and standalone wants a name like `published_surfaces`, especially sitting two lines from `HOSTED_SURFACES if config.is_hosted_mode`.

The alternative is to drop the branch and let the literal stand alone.

daavoo (commented):
One note on the hybrid branch in `hosted_surfaces`, inline. Nothing blocking; the rest of the seam reads well and the worker supervision is right.

Posted by Claude Opus 5 via Claude Code.

src/gateway/core/feature.py
```diff
@@ -0,0 +1,31 @@
...
+
+@dataclass(frozen=True)
+class CoreFeature:
+    """One feature the build ships.
+
+    ``enabled`` is read at startup, when routers are mounted and workers are
+    created, and again whenever the deployment bootstrap publishes the
+    surfaces it hosts. It must therefore answer from startup-only settings;
+    a runtime-settable field would let the published surface set drift from
+    what is mounted and running.
```
daavoo: Worth strengthening this later, not now. Nothing enforces it, and the cost of breaking it is more than drift.

Runtime settings mutate the live config object: `apply_override(config, key, value)` in `runtime_settings_service.py:241` writes to the same object `app.state.config` holds. Making a field settable is one line in `_SPECS` (`runtime_settings_service.py:93`), in a file somebody else is likely editing, and nothing connects that dict to `enabled`.

If a feature's switch ends up there, three things happen, none of which log anything:

1. Switched on at runtime: the nav row appears, the page 404s. Cosmetic, fixed by a restart.
2. Switched off at runtime: the routers stay mounted and keep answering. The operator sees the row vanish and believes the feature is off.
3. Same case with a worker: the worker was created once at startup and keeps running. For alerts, that means notifications still going out after an operator switched alerts off.

The fix that closes all three is to read `enabled` once and have everyone use that answer: compute the enabled set in `create_app`, put it on `app.state`, and have `register_routers`, the lifespan and the bootstrap route read the set rather than calling `enabled(config)` again. Then the three cannot disagree whatever a feature reads.

If you would rather keep it as a documented contract, name those effects here instead of "drift". "Drift" does not warn the next person that turning a feature off can leave it sending email.

daavoo (approved):
Approving. The seam is sound, the worker supervision is right, and the tests are unusually good for an empty registry.

One more note inline, on the `enabled` contract. It is a trap for later rather than a bug now, so it does not block. My earlier note on the hybrid branch in `hosted_surfaces` still stands.

Reviewed by Claude Opus 5 via Claude Code.

src/gateway/api/routes/bootstrap.py
peteski22: Addressed in aa4b81a4: the hybrid arm now calls the helper, renamed `published_surfaces`, so one function answers for all three modes. The hybrid test now asks the endpoint with an enabled feature in the registry, and it fails if the helper's hybrid branch is removed.

src/gateway/core/feature.py
peteski22: Done in c7b8be8e, the fix rather than the documented contract. `create_app` records the enabled features on `app.state` once, and `register_routers`, the lifespan and the bootstrap use that tuple, so the bootstrap route no longer imports the registry and its architecture-check exemption is gone. One correction on case 1: a restart did not fix it. Routers mount in `create_app` before the lifespan applies stored overrides, so a stored override split the three permanently (checked with a real settings PATCH and a restart). Reading once makes them agree, but a dashboard-settable switch then never takes effect, so the `CoreFeature` docstring now says that in those words rather than "drift". `test_a_switch_changed_after_the_app_is_built_moves_nothing` covers both directions.

PR #1189, opened 2026-09-15
fix(pricing): close hosted-credential gap in organization pricing overrides

tbille:
## Description

An organization's pricing override (`POST /api/v1/organizations/me/pricing`) is meant to be refused for a model the deployment, not the organization, pays the upstream bill for — otherwise an organization could set its own rate (including zero) and undercut what it's actually being charged. That refusal only recognized the deployment's own configured provider instances as "deployment-supplied." It never asked the port an overlay uses to serve a bare `provider:model` key on a deployment-owned hosted credential when the organization has no key of its own. An organization could set its own price for a model actually served on a hosted fleet, as long as it named a bare provider rather than a configured instance.

This PR extends that refusal to also consult the hosted-credential port, while first checking whether the organization holds a real BYO key of its own for the provider (so an org that already dispatches on its own key isn't wrongly refused just because a hosted fleet also exists under the same provider name). It also closes two related loopholes found while hardening this: a construction site could previously omit the port dependency and silently get the weaker check, and a provider-key row with no actual credential material could satisfy the BYO check without being able to serve a request.

Two residual gaps are documented explicitly in the code rather than fixed here: the BYO check is organization-wide rather than workspace-scoped, and the refusal is asserted at write time only (not re-checked at pricing resolution if a key is later archived) — the same shape the existing config-instance rule already had.

## How to test it locally

- `make lint` (architecture check, ruff)
- `make typecheck` (mypy)
- `uv run pytest tests/unit/test_organization_pricing_resolution.py tests/integration/test_organization_pricing_routes.py` — includes new cases for: an org's own BYO key exempting it even when a hosted fleet also serves the same provider name, a hosted-access refusal still counting as deployment-supplied, the operator exemption applying through this path too, an unrelated hosted fleet leaving BYO pricing untouched, and a credential-less key row not satisfying the BYO check.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Related to otari-ai#2095 (the rule this refusal enforces).

## Checklist

- [ ] If the API contract changed, I regenerated the OpenAPI spec (`uv run python scripts/generate_openapi.py`). (No route shape changed.)

## AI Usage

- [ ] No AI was used.

**Any additional AI details you'd like to share:** Implemented after an investigation into hosted-provider pricing manipulation, then hardened over three rounds of independent multi-LLM review (openai/anthropic/gemini via star-chamber) with fixes applied between rounds.

## Summary

Organization pricing overrides now reject models served by deployment-owned hosted credentials when no usable organization BYO credential covers the organization’s workspaces.

Organizations with usable BYO credentials remain allowed. Deployment operators remain exempt. Unrelated hosted credentials do not cause a refusal. Credential-less key rows do not satisfy the BYO requirement.

The change adds workspace-aware validation, updates service construction, adds test coverage, and updates route and client documentation.

## Technical notes

- `OrganizationPricingService` now requires a `ModelProviderPort` dependency.
- Validation checks each organization workspace and respects disabled workspace overrides.
- The refusal applies when the pricing override is written.
- No route shape changed, so OpenAPI regeneration was not required.


src/gateway/services/organization_pricing_service.py
```diff
@@ -131,12 +167,140 @@ async def raise_if_deployment_supplied(self, user: TenancyUser, model_key: str)
...
+        """
+        if is_deployment_instance_key(self.config, model_key):
+            return True
+        if self.model_provider is None:
+            return False
+        split = split_selector(model_key)
+        if split is None:
+            return False
+        provider, model = split
+        if await self._organization_has_byo_credential(organization_id, provider):
```
khaledosman, line 184: Ask the port before the BYO check. As written, every bare-key override write pays four queries (`list_live_keys`, the workspace `COUNT`, the workspace page, `candidates_for_workspaces`) even where no hosted fleet exists at all. `SelfHostedModelProviderAdapter.resolve_hosted_credential` returns `None` with zero I/O, so:

```python
credential_or_denied = ...  # port first
if not credential_or_denied:
    return False
return not await self._organization_has_byo_credential(organization_id, provider)
```

gives the identical answer and skips all of it on every OSS deployment.

The docstring's reason for the current order ("asking it first, unconditionally, would refuse an organization pricing traffic that already dispatches on its own key") does not hold: the BYO check still gates the refusal either way, it just runs second.

src/gateway/services/organization_pricing_service.py
```diff
@@ -131,12 +167,140 @@ async def raise_if_deployment_supplied(self, user: TenancyUser, model_key: str)
...
+        return True
+
+    async def _all_workspace_ids(self, organization_id: uuid.UUID) -> list[uuid.UUID]:
+        """Every workspace id this organization has, paged past `WorkspaceRepository`'s default limit."""
+        workspace_repository = WorkspaceRepository(self.db)
+        ids: list[uuid.UUID] = []
+        skip = 0
+        limit = 100
+        while True:
+            page, total = await workspace_repository.get_by_organization(organization_id, skip=skip, limit=limit)
```
khaledosman: `_all_workspace_ids` runs a `COUNT(*)` per page and hydrates full `Workspace` ORM rows only to read ids. The offset paging also orders by `created_at desc`, so a workspace created during the walk shifts rows between pages and can drop an id, and a dropped workspace here is a wrongly-granted exemption rather than just a missing row.

One statement covers it:

```python
result = await self.db.execute(
    select(col(Workspace.id)).where(col(Workspace.organization_id) == organization_id)
)
return list(result.scalars().all())
```

This file already builds its own selects in `raise_if_overlapping`, so that stays inside the layer rules.

src/gateway/api/routes/organization_pricing.py
```diff
@@ -212,9 +212,16 @@ class OrganizationModelPricingsPublic(BaseModel):
...
-    """Build the pricing service on the request's session and provider map."""
-    return OrganizationPricingService(db, config)
+    """Build the pricing service on the request's session, provider map, and hosted-credential port.
+
+    The port is what lets the deployment-supplied-model refusal see past
+    ``config.providers``: a build with no configured instances can still serve a
+    bare ``provider:model`` key on a hosted credential an overlay's adapter owns,
+    and only the port knows that.
+    """
+    return OrganizationPricingService(db, config, model_provider=model_provider)
```
khaledosman, line 218: Nothing exercises this wiring. All nine new tests construct `OrganizationPricingService` directly with `_FakeHostedModelProvider`, so the claim that the route binds a real port is the one part of the fix with no coverage, and it is the part the PR description calls out as a closed loophole.

`tests/integration/test_organization_pricing_routes.py` already has HTTP-level tests to model one on: override `get_model_provider_port` on the app and assert `POST /api/v1/organizations/me/pricing` answers 403 for a bare key the fake serves.

src/gateway/services/organization_pricing_service.py
```diff
@@ -93,10 +105,29 @@ def _describe_period(effective_from: datetime, effective_to: datetime | None) ->
...
+        model_provider: ModelProviderPort | None,
+    ):
+        """Bind the request's session, provider map, and hosted-credential port.
+
+        ``model_provider`` has no default; see the inline comment below for why.
+        """
         self.db = db
         self.config = config
         self.organizations = OrganizationService(db)
+        # Keyword-only with no default, and nullable rather than defaulted to
```
khaledosman: Nine lines to justify a keyword-only parameter that protects exactly one construction site in `src/`. Same shape at line 175, where "A write-time answer, not a standing one" documents a gap this PR does not close.

AGENTS.md: "Add a comment only where the logic is not obvious... Do not restate the code, narrate the change, or record what the code used to do: the commit message is where that belongs. Leave the comments around a change shorter than you found them." A line each would carry it.

khaledosman (approved):
The gate is right and it errs toward refusing, which is the safe direction for a money rule. Four inline findings, none blocking: one ordering change that would drop 4 queries per write on every deployment with no hosted fleet, one pagination helper worth collapsing to a single select, an untested route wiring, and comment bloat against AGENTS.md.

src/gateway/services/organization_pricing_service.py
peteski22: Kept the BYO check first. The port's contract covers only a model no BYO key serves, and dispatch asks it only after BYO fails. Port-first would also let a port error block writes for organizations that have their own key. The saved queries are on a rare admin write, not the request path. The docstring did give the wrong reason: 60abf6a8 now states the contract, and a gotcha comment keeps the order.

src/gateway/services/organization_pricing_service.py
peteski22: Addressed in 3e792816: added `WorkspaceRepository.get_ids_by_organization`, one ID-only select with no paging, and removed the helper.

src/gateway/api/routes/organization_pricing.py
peteski22: Addressed in cfe31b6c: `test_the_route_asks_the_port_this_build_bound` rebinds the port in the app container and writes as an organization admin session. The master key cannot cover this, because the operator is exempt before the port is asked. It gets 201 for a provider the fake does not serve and 403 for one it serves, and it fails with 201 if the route drops the port.

src/gateway/services/organization_pricing_service.py
peteski22: Addressed in 60abf6a8: both comments are now one or two lines, and the other docstrings this PR added are cut to contract and gotchas.

peteski22:
The refusal in this PR runs at write time only, so it does not close one gap. An override stays in force after the organization loses BYO coverage, and requests that then fall through to a hosted credential still price at the organization's rate. #1254 tracks that gap.

src/gateway/services/organization_pricing_service.py
```diff
@@ -131,12 +156,82 @@ async def raise_if_deployment_supplied(self, user: TenancyUser, model_key: str)
...
+        matching_key_ids = {key.id for key in matching_keys}
+        candidates_by_workspace = await WorkspaceProviderKeyOverrideRepository(self.db).candidates_for_workspaces(
+            organization_id=organization_id,
+            workspace_ids=workspace_ids,
+        )
+        for workspace_id in workspace_ids:
+            candidates = [
+                (key, override)
+                for key, override in candidates_by_workspace.get(workspace_id, [])
+                if provider_key(key.provider) == normalized
```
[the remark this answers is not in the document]
peteski22: Not applied. Dispatch resolves the active key over every live key and checks usability after, so an unusable default key sends the request to the hosted credential. It does not fall back to an older usable key. This check mirrors that on purpose. Filtering first would count the older key as coverage and grant the exemption for traffic the deployment pays for. 832b8c20 adds a test that fails with the suggested change.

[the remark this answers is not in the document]
peteski22: Proof in a3811e40, `test_an_unusable_default_key_does_not_fall_back_to_a_usable_one`. The organization has a usable key and an undecryptable one.

1. With the usable key as the org default, the dispatch cache gives the workspace a credential.
2. With the undecryptable key as the default, the dispatch cache gives the workspace nothing, and `credential_ladder_exhausted` is true. Dispatch does not fall back to the usable key, so the request goes to the hosted port.
3. The pricing check then refuses the override.

With the suggested filter applied, steps 1 and 2 still pass and step 3 fails with DID NOT RAISE. The filter counts the usable key as coverage for a workspace whose traffic the deployment pays for. So the check keeps resolving over every live key, the same way dispatch does.

PR #1190, opened 2026-09-15
refactor(gateway): derive the settings view from the settings

### LINKED ISSUE #1176: Settings view derived from the settings
### Problem

`src/gateway/api/routes/settings.py` keeps `_CONFIG_VIEW` and `_DELIBERATELY_OMITTED` as central tuples of field names. Every new setting edits one of them, or the coverage test fails.

### Proposal

Carry the view metadata on the settings fields themselves: shown, omitted, or secret. Derive the two tuples from it. The endpoint's output does not change.

### Acceptance

- [ ] `test_settings_endpoint.py` passes unchanged.
- [ ] A snapshot of `GET /api/v1/settings` is identical before and after.

### Depends on

Nothing. Easier after the config split, but it can land first.

Part of #1171.

peteski22:
## Description

The dashboard's Settings page shows a read-only view of the gateway's configuration, grouped by topic, with credentials and structured blocks left out. Until now, which setting went where was kept in two hand-maintained lists in the settings endpoint. Every new setting had to be added to one of them, or a test failed.

Now each setting says where it belongs, on its own field: shown in a named group, omitted, or secret. The endpoint lays the view out from those declarations. A setting with no declaration fails at import, so a new one still cannot be forgotten, and it fails earlier than a test would.

Nothing changes for anyone using Otari. The settings response is byte-for-byte identical before and after, including the order of fields. Eleven field declarations moved within the config so that declaration order matches the order the view already had.

Churn, measured over 90 days as the epic asks: `config.py` goes from 70 commits to 71 and `routes/settings.py` from 29 to 30. Authors unchanged.

## How to test it locally

```
make lint && make typecheck
uv run pytest tests/unit/test_settings_view.py tests/unit/test_settings_endpoint.py
```

What to look for: `GET /api/v1/settings` answers exactly what it did before. `test_settings_endpoint.py` is unchanged and still passes, including the check that every config field is either shown or deliberately omitted. `test_settings_view.py` covers the mechanism on small models: group order, declaration order within a group, and that a field with no declaration, or two, is refused.

Run locally: lint, typecheck, the full unit suite (3462 passed; the 14 failures are the generated-master-key and Claude Code import tests, which fail the same way on `main` in my checkout because a local `.env` sets `OTARI_MASTER_KEY` and my shell sets `OTARI_API_KEY`). The smoke gate and the integration suite are left to CI: the same `.env` sets `OTARI_PORT`, which the smoke gate's server picks up in place of its own port.

## PR Type

- [ ] New Feature
- [ ] Bug Fix
- [x] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes #1176. Part of #1171.

## Checklist

- [ ] I ran the Definition of Done checks locally (`make lint`, `make typecheck`, `make test`). Lint, typecheck, and `make test-unit` ran locally; `make test-integration` is left to CI.
- [x] If the API contract changed, I regenerated the OpenAPI spec (`uv run python scripts/generate_openapi.py`). Not applicable: the response shape is unchanged.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:** Claude Fable 5.1, through Claude Code.

**Any additional AI details you'd like to share:** The issue and the decision to keep the field order by moving declarations were Peter's. The agent implemented the change, wrote the tests, and ran the checks. Peter reviewed the result before it was pushed.

## Summary

- Added visibility metadata to configuration fields.
- Derived the settings endpoint view from this metadata.
- Reordered selected fields to preserve the existing response order.
- Added validation for missing or duplicate visibility declarations.
- Added tests for grouping, ordering, credentials, and invalid declarations.

This removes hand-maintained visibility lists and keeps future settings changes aligned with the endpoint view.

## Technical notes

The settings response remains unchanged. Test execution results are unavailable because the repository clone failed.


src/gateway/core/settings_view.py
```diff
@@ -0,0 +1,101 @@
...
+
+
+@dataclass(frozen=True)
+class Omitted:
+    """Kept out of the view: a structured block, or a scalar no page has asked for."""
+
+
+@dataclass(frozen=True)
+class Secret:
+    """A credential. Never shown."""
```
daavoo, line 54: `Secret` and `Omitted` do the same thing: both land in `hidden`. Mark `master_key` as `OMITTED` and every test still passes, so the "never carries a credential" rule in the module docstring is not one yet.

`test_no_credential_is_shown_in_the_config_view` covers five names and asserts only that they are not shown; `web_search_provider_api_key` and `web_search_backend_token` are not in it (that gap predates this PR). Now that there is a marker to check, assert `view_of(...) is SECRET` for all seven.

src/gateway/core/config.py
```diff
@@ -1233,28 +1260,27 @@ class GatewayConfig(BaseSettings):
...
             "Off by default. Only enable for unusual setups such as a private search index."
         ),
     )
-    mcp_allow_loopback: bool = Field(
+    mcp_allow_loopback: Annotated[bool, Shown(SettingsGroup.TOOLS)] = Field(
         default=True,
         description=(
-            "SSRF gate: allow MCP server URLs that resolve to loopback (useful for same-host "
-            "sidecars). On by default."
+            "SSRF gate: allow MCP server URLs that resolve to loopback (useful for same-host sidecars). On by default."
```
daavoo: Nit: this reflow, and three more at lines 1698, 1791 and 2226, are unrelated to the change. Looks like a whole-file format ran. Worth reverting so the diff stays about one thing.

daavoo (approved):
I rebuilt both versions and compared them: the derived view is exactly what `main` hard-coded, down to group order, field order within each group, and the hidden set. The eleven moved declarations are right. 

pd @peteski22 there are merge conflicts

src/gateway/core/settings_view.py
peteski22: Right, the marker was decorative. Fixed in 9073577b: a parametrized test asserts all seven credentials carry SECRET, including the two the endpoint's own test never named. Checked it is load-bearing by marking master_key as OMITTED, which now fails that test. The constant listing them sits at the top of tests/unit/test_settings_view.py.

src/gateway/core/config.py
peteski22: Correct, a whole-file format ran. Reverted all four in 9073577b, so the diff is field annotations and the eleven declaration moves only.

PR #1205, opened 2026-09-15
refactor(metrics): move each metric beside the code that increments it

### LINKED ISSUE #1177: Metrics beside their owners
### Problem

Every Prometheus counter lives in `src/gateway/metrics.py`. Every feature edits it.

### Proposal

Move each counter next to the code that increments it, registered on the existing shared `REGISTRY`. `metrics.py` keeps the registry and the cross-cutting helpers only.

### Acceptance

- [ ] `/metrics` output has the same metric names and labels before and after. A scrape diff in the test proves it.

### Depends on

Nothing.

Part of #1171.

peteski22:
## Description

Nothing changes for someone using Otari. The Prometheus scrape at `/metrics` exposes the same 14 gateway metric families, with the same names, types and labels, before and after.

What changes is where each metric is declared. Until now every counter lived in one shared module, so every feature that added or touched a metric edited that file. Each metric is now declared in the module that increments it: the rate limiter, the budget service, the usage log writer, the authentication dependencies, the hybrid attempt runner and the request pipeline. The shared module keeps the registry, the process collector and the HTTP request instrumentation only.

The first commit pins the set of exposed families in a test, before any move, so every move was checked against the same set. That test is the scrape diff the issue asks for.

## How to test it locally

```
make lint && make typecheck && make test-unit
uv run pytest tests/integration/test_metrics.py
```

The pinned family test is `tests/unit/test_gateway_metrics.py::test_scrape_exposes_the_pinned_families`. To eyeball the scrape itself, start the gateway with `enable_metrics: true`, hit `/metrics`, and compare the `# HELP` and `# TYPE` lines against `main`; the lines are the same, only their order differs, because a family now registers when its owner module imports.

Measured before and after (90-day churn on `main` at `ea1a2471` versus this branch):

| | before | after |
|---|---|---|
| lines in `src/gateway/metrics.py` | 255 | 130 |
| modules importing from `gateway.metrics` (`src/`) | 11 | 7 |
| of which import anything but `REGISTRY` | 11 | 1 (`main.py`, the middleware and endpoint) |
| commits touching `metrics.py`, last 90 days | 5 | n/a |

Run locally: `make lint`, `make typecheck`, `make test-unit`, and `tests/integration/test_metrics.py` against a Testcontainers PostgreSQL. The rest of the integration suite was left to CI.

Not run here: the OSS edition smoke gate fails on this machine on `main` too (model discovery against the mock provider never answers, so readiness times out), and 14 master key and dashboard session unit tests fail on `main` too. Neither touches metrics; CI should confirm both.

## PR Type

- [ ] New Feature
- [ ] Bug Fix
- [x] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes #1177. Part of #1171.

## Checklist

- [ ] I ran the Definition of Done checks locally (`make lint`, `make typecheck`, `make test`). Lint, typecheck and the unit suite were run; see above for what of `make test` was not.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:** Claude Code (Claude Fable 5.1)

**Any additional AI details you'd like to share:** The agent wrote the moves and the pinned test, ran lint, typecheck, the unit suite and the metrics integration tests, and confirmed the unrelated failures reproduce on `main`.

## Summary

- Moved Prometheus metric declarations into the modules that use them.
- Kept the shared registry and request instrumentation in `gateway.metrics`.
- Preserved the existing `/metrics` metric names, types, and labels.
- Added a pinned metric-family contract test and focused tests for relocated metrics.
- Re-exported Prometheus metric types from `gateway.metrics` to keep one import boundary.

This improves ownership and keeps future metric changes closer to the related code.


daavoo:
Not a blocker, and not an objection to the move: a packaging ask that this PR can satisfy for almost nothing, and that gets materially more expensive once it lands.

Putting each declaration beside its owner is right, and the 14 families come through unchanged. The side effect is that `prometheus_client` becomes a top-level import in six modules (`api/deps.py`, `api/routes/_pipeline.py`, `api/routes/_platform.py`, `rate_limit.py`, `services/budget_service.py`, `services/log_writer.py`) where today it is one. Before this PR, `gateway/metrics.py` is the single place where prometheus could be made optional; after it, that change reopens all six.

And we do want it optional. `prometheus-client` is a core dependency, imported on every startup regardless of `enable_metrics`, which gates only the middleware and the `/metrics` route ([main.py:803](https://github.com/mozilla-ai/otari/blob/main/src/gateway/main.py#L803), [main.py:837](https://github.com/mozilla-ai/otari/blob/main/src/gateway/main.py#L837)). A deployment that never scrapes still pays the import (most of it `prometheus_client.exposition` pulling in `http.server` and `wsgiref.simple_server`, which nothing else here needs), the `ProcessCollector` registration, and 14 counters incrementing that nothing can read.

**The ask:** in those six modules, import `Counter`, `Gauge` and `Histogram` from `gateway.metrics` rather than from `prometheus_client`. Same objects, since `metrics.py` would just re-export them, so nothing about this PR's behavior or its pinned-family test changes. It keeps one seam where the `try: from prometheus_client import ... / except ImportError:` fallback can live, so the follow-up is confined to `metrics.py` and `pyproject.toml`.

Two small bonuses if you do: it retires the `PrometheusCounter` alias in `_pipeline.py`, since `gateway.metrics.Counter` collides with nothing, and `metrics.py` keeps a reason to exist beyond the registry, which reads better than a module that is only a variable.

#1235 is a reference draft of the follow-up against `main`, so you can see what the seam buys: the extra, the no-op fallback, a startup refusal when `enable_metrics` is set without it, and the OSS smoke gate (which installs no extras) becoming the thing that keeps the fallback honest. It should rebase onto this once this lands. If the import source changes here, its diff to `metrics.py` shrinks to the fallback block alone.

tbille (approved):

peteski22:
Done in 3a411130. Two notes. mypy strict does not treat an imported name as exported, so `metrics.py` now lists its public names in `__all__`. That also covers both import blocks in #1235. The `PrometheusCounter` alias stays, because `_pipeline.py` also imports `collections.Counter`.

PR #1210, opened 2026-09-16
fix(gateway): forward ttft_ms in hybrid-mode usage reports

AmirF194:
## Description

Follow-up to #1099, per coderabbitai's review comment there: `ttft_ms` never reaches the platform when `platform_active` is true. `build_streaming_response`'s `_on_complete`, `_on_no_usage` and `_on_error` all branch on `platform_active` and, when true, report through `_report_platform_usage` instead of `log_usage`. `_report_platform_usage` had no `ttft_ms` parameter or payload field, so #1099's TTFT tracking only reached the standalone path.

Gave `_report_platform_usage` an optional `ttft_ms` parameter and included it in the usage-report payload when set, then passed `_ttft_ms(started_at, first_chunk_at)` from the three streaming callbacks, the same call every standalone `log_usage` site already makes.

One thing worth flagging: this assumes the platform reads a field named `ttft_ms`, matching the name used everywhere else in this codebase (`UsageLog.ttft_ms`, `log_usage`'s own parameter). I don't have visibility into the platform-side consumer, so that naming is not verified end to end, only that the gateway now sends it.

## How to test it locally

`uv run pytest tests/unit/test_pipeline_settlement.py -k ttft` covers the three callbacks. Reverting just the `_pipeline.py`/`_platform.py` hunks turns the three new tests red (`KeyError: 'ttft_ms'`) and the existing `test_platform_stream_without_usage_reports_final_success` red too (it asserts the full report payload).

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Follow-up to #1099 (review comment: https://github.com/mozilla-ai/otari/pull/1099#discussion_r3989649601).

## Checklist

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:** AI coding assistant.

Verified with the tests and local checks described above before opening this.

## Summary

- Include `ttft_ms` in hybrid-mode platform reports for completed, no-usage, and errored streams.
- Pass the request start time through all streaming fallback routes.
- Document fallback-chain TTFT timing and add regression coverage.

This preserves time-to-first-token data in platform reports after a streamed chunk is received.


src/gateway/api/routes/_pipeline.py
```diff
@@ -3645,6 +3645,7 @@ async def _on_complete(usage_data: CompletionUsage) -> SettledCost | None:
                     outcome="success",
                     usage=usage_data,
                     session_label=session_label,
+                    ttft_ms=_ttft_ms(started_at, first_chunk_at),
```
khaledosman, line 3648: This (and the same line in `_on_no_usage` and `_on_error`) evaluates to `None` on every real hybrid request, so `_report_platform_usage` skips the field and the reported bug is unfixed.

`run_streaming_with_fallback` is the only path that reaches the `platform_active` branches: `chat.py:540`, `messages.py:764` and `responses.py:637` call it inside `if ctx.hybrid_mode`, and no caller passes `platform_correlation_id` to `run_single_attempt_stream`, so that one is standalone-only. `run_streaming_with_fallback`'s `build_streaming_response(...)` call at `_pipeline.py:4278` never passes `started_at`, which therefore defaults to `None`, and `_ttft_ms` returns `None` whenever `started_at is None`.

The PR's own `test_platform_stream_without_usage_reports_final_success` pins that production shape: `_build_platform(stream())` with no `started_at`, asserting `"ttft_ms": None`.

Fix: give `run_streaming_with_fallback` a `started_at: float | None = None` parameter, forward it at `_pipeline.py:4278`, and pass `started_at=ctx.started_at` from the three route call sites. `ctx` is already in scope in each hybrid branch, and that keeps the interval identical to the standalone path's (request start, not runner entry).

src/gateway/api/routes/_platform.py
```diff
@@ -1091,6 +1092,8 @@ async def _report_platform_usage(
     normalized_label = (session_label or "").strip()
     if normalized_label:
         payload["session_label"] = normalized_label
+    if ttft_ms is not None:
+        payload["ttft_ms"] = ttft_ms
```
khaledosman, line 1096: New wire field, but `docs/hybrid-mode-protocol.md` is not updated. Its "Usage report" payload block (around line 395) enumerates every field the gateway sends and the doc is authoritative for this contract, so add `ttft_ms` to the block with its semantics: milliseconds from request start to the first streamed chunk, omitted when no chunk arrived or no start was captured, and sent on `status: "error"` reports too (non-streaming reports never carry it).

That also answers the naming question raised in the PR description: documenting the field here is what makes `ttft_ms` the name the platform is expected to read, rather than an unverified guess.

src/gateway/api/routes/_platform.py
```diff
@@ -1060,6 +1060,7 @@ async def _report_platform_usage(
     usage: CompletionUsage | None,
     error_class: str | None = None,
     session_label: str | None = None,
+    ttft_ms: int | None = None,
```
khaledosman: Landing `ttft_ms` above the `*` makes it the 7th positional parameter. Three call sites already pass six positionals (`_pipeline.py:4091`, `4218`, `4399`), so the next parameter inserted here silently binds to the wrong argument. Every site passes this one by keyword, so move it below the `*` next to `is_final_attempt`.

tests/unit/test_pipeline_settlement.py
```diff
@@ -657,6 +661,7 @@ def _build_platform(stream: AsyncIterator[ChatCompletionChunk]) -> Any:
         reservation=None,
         platform_correlation_id="corr-1",
         platform_request_id="req-1",
+        started_at=started_at,
```
khaledosman, line 724: The three new tests call `build_streaming_response` directly with an explicit `started_at`, a shape the hybrid path never constructs, which is why they pass while hybrid streaming still reports no TTFT. Add one test that goes through `run_streaming_with_fallback` so the wiring, not just the callbacks, is covered.

khaledosman (changes requested):
`ttft_ms` still never reaches the platform: `started_at` is `None` on every hybrid streaming request, so all three new call sites compute `None` and the payload field is omitted. Details inline, plus the missing protocol-doc update.

AmirF194:
Pushed 61116bc0. run_streaming_with_fallback now takes started_at and forwards it to build_streaming_response, and the three hybrid call sites (chat.py, messages.py, responses.py) pass ctx.started_at. Moved ttft_ms below the * in _report_platform_usage per your point on the 7th positional. Documented the field in hybrid-mode-protocol.md and added a test that goes through run_streaming_with_fallback itself rather than calling build_streaming_response directly.

src/gateway/api/routes/_pipeline.py
```diff
@@ -4127,6 +4130,7 @@ async def run_streaming_with_fallback(
     rate_limit_info: RateLimitInfo | None,
     tool_ctx: ToolContext,
     session_label: str | None = None,
+    started_at: float | None = None,
```
khaledosman: Make this required rather than defaulted: `started_at: float`.

All three callers pass `ctx.started_at`, which is typed `float` (not optional) at `_pipeline.py:781`, so nothing needs the `None`. The default is the same shape as the bug being fixed here: a fourth streaming route added later compiles, typechecks, and silently reports no TTFT, with no test failing. The parameter list is already keyword-only, so dropping the default costs nothing at the call sites.

docs/hybrid-mode-protocol.md
```diff
@@ -384,8 +384,12 @@ Content-Type: application/json
   "error_class": "http_401",           // optional on error; omitted when the
                                        // Otari can't classify the failure
                                        // (e.g. mid-stream errors). See below.
-  "session_label": "my-run-personas"   // optional; the caller's cost-attribution
+  "session_label": "my-run-personas",  // optional; the caller's cost-attribution
                                        // label (see below). Omitted when absent.
+  "ttft_ms": 340                       // optional; milliseconds from request start
```
khaledosman, line 389: Worth one clause on scope. `started_at` is the handler preamble reading and the report is keyed by `correlation_id` = `chosen.attempt_id`, so on a fallback chain the final attempt's `ttft_ms` includes the time burned on earlier failed attempts and is not that attempt's own TTFT. Without that said, a platform-side consumer reads an attempt-scoped record as carrying per-attempt latency.

tests/unit/test_pipeline_settlement.py
```diff
@@ -696,10 +760,89 @@ async def stream() -> AsyncIterator[ChatCompletionChunk]:
         "outcome": "success",
         "usage": None,
         "session_label": None,
+        "ttft_ms": None,
         "is_final_attempt": True,
     }
 
 
+@pytest.mark.asyncio
+async def test_platform_stream_reports_ttft_on_complete(monkeypatch: pytest.MonkeyPatch) -> None:
```
khaledosman, line 770: These three tests monkeypatch `_report_platform_usage` out, and `test_report_platform_usage_forwards_ttft_ms` exercises the payload builder in isolation. The gap between those two layers is exactly where #1099's bug lived, and nothing here asserts the field on the wire for a hybrid stream.

`tests/integration/test_hybrid_mode_chat.py` already captures report bodies on a streaming fallback (the test that asserts only `error_reports` today), so asserting `ttft_ms` on the success report there closes the loop for two lines.

khaledosman (approved):
The wiring is complete: both `build_streaming_response` call sites now thread `started_at`, `_flush_pending_usage_reports` calls positionally so the new keyword-only parameter does not disturb it, and `_on_incomplete` has no platform branch to miss. No route or schema changed, so no generated artifacts are owed.

On the open question in the description about the field name: I checked the platform-side consumer. Its usage-report model ignores unknown fields, so the report is accepted rather than 422'd, but it has no `ttft_ms` field yet and drops the value. The gateway side is right to land first; the consumer change is tracked separately. Nothing to change here.

Three non-blocking comments inline.

---
Reviewed with [Claude Code](https://claude.com/claude-code)

AmirF194:
Pushed a23963e for the three non-blocking comments: started_at is required now (all three callers already pass a float), the protocol doc notes that ttft_ms on a fallback chain includes earlier failed attempts, and the streaming-fallback integration test now asserts ttft_ms on the wire success report instead of only through the payload builder. Resolved the threads.

AmirF194:
Appreciate the careful second pass on the started_at plumbing, that was the right catch.

PR #1215, opened 2026-09-16
fix(providers): let a tenant read the provider catalog

khaledosman:
## Description

An organization owner or admin could not add a bring-your-own provider key through the dashboard. The form's Provider dropdown was always empty, saying "No provider to offer here.", so there was nothing to pick and the key could never be created.

The dropdown is filled from a small read that lists the providers Otari knows how to talk to. That read sat behind the deployment-operator gate along with the rest of the provider-configuration endpoints, but the form it feeds is a tenant surface: the people entitled to manage an organization's provider keys are its owners and admins, who are not operators of the deployment. They passed the check on the keys themselves and were then refused the list of providers to choose from.

The list names the providers the build supports. It is not a description of anything the operator configured, holds no credential, calls no provider and writes nothing, so it now sits with the other reads of that kind and any signed-in caller can read it. The endpoints that do describe or change the gateway's own provider configuration are unchanged and stay operator-only.

Second, smaller fix in the same area: when that read failed for any reason the dropdown reported it as an empty catalog, because a failed request leaves the query with no data, which looks exactly like a deployment with no providers. It now says the catalog could not be loaded, so "there are none" and "we could not find out" stop looking alike. That is what made the original bug so hard to place from the UI, and it is worth having independently of the gate.

## How to test it locally

1. `make dev`, then open the dashboard and sign in.
2. Create an organization and add a second user to it as an `owner` or `admin`, not a superuser.
3. Sign in as that user and go to Organization, then Provider keys, and press **Add provider key**.
4. Open the **Provider** dropdown. Before this change it says "No provider to offer here."; after it lists the providers, and picking one fills in the credential fields.
5. Confirm the rest is still shut: as that same non-operator, `GET /api/v1/providers` and `GET /api/v1/provider-credentials` still answer 403.

Automated coverage:

- `tests/integration/test_deployment_operator_gate.py::test_a_member_reads_the_provider_catalog_the_byo_form_offers` is the regression test. It asserts both directions from one session: the two catalog routes answer 200, and `GET /api/v1/providers` on the same router still answers 403.
- `tests/unit/test_operator_gate_declarations.py` is the structural half. The new router is listed in `_NON_OPERATOR_ROUTERS`, so it now asserts the operator gate is *off* those two routes and still on every other route in the module. It also fails if a future route is added to either router without a deliberate decision.
- `web/.../OrganizationProviderKeysPage.test.tsx` covers the frontend half by refusing the catalog read and asserting the form says the catalog could not be loaded rather than that there are no providers.

Run locally: `make lint`, `make typecheck`, `uv run pytest tests/unit` (3542 passed), `tests/integration/test_deployment_operator_gate.py` (77 passed), `pnpm --dir web run lint`, `pnpm --dir web run typecheck`, `pnpm --dir web test` (6572 passed). `make openapi-check` and `make postman-check` both report up to date: the routes keep their paths and their shapes, so the spec is byte-identical and no generated artifact moved.

The regression test was also checked in the failing direction. With the two route decorators reverted to the operator router it fails on `assert 403 == 200`, which is the reported symptom, so it is pinning the behavior rather than passing incidentally.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

None filed; reported directly from a deployment where the form could not be used.

## Checklist

## AI Usage

- [ ] No AI was used.

The gate was found by tracing an empty provider dropdown on a running deployment back through the frontend query to the router declaration.

## Summary

Organization owners and admins can now read the provider catalog when creating provider keys. Provider configuration endpoints remain restricted to deployment operators.

The provider-key form now shows a clear error when the catalog cannot load. It still shows an empty-state message when no providers are available.

Tests cover route access, backend behavior, and the frontend error state.


peteski22:
Rebased onto main and added one commit (c4762e5c). It corrects the exception list in `require_deployment_operator`'s docstring, which this change makes more out of date. That is the nit from #1234, which I closed as a duplicate of this PR.

peteski22 (approved):

PR #1167, opened 2026-09-15
fix(dashboard): restore activation feedback and center the success burst

jigjigjig:
## Description

The activation guide's copy and reveal controls moved surrounding text, a fast status check hid the working orb before it could be seen, and the success burst started at the left-hand checkmark. This follow-up to #1101 keeps copy confirmation in the button (and its screen-reader announcement), keeps snippet guidance stable when revealing a key, restores the original 2.4-second manual-check beat and 420 ms orb crossfade, and centers confetti on the settled success dialog.

The browser comparison also exposed a layering mismatch: canvas-confetti defaults to z-index 100 while HeroUI's modal is at 100,000. The burst now clears the containing overlay. It waits for the dialog entrance before measuring its center and cancels if that presentation closes before it fires. Reduced-motion users get frozen orb phases and no artificial manual-check delay; confetti retains its reduced-motion opt-out.

Compared against [the original activation flow](https://github.com/mozilla-ai/otari-ai/pull/1462). The profile survey and request-specific inspect action remain the separately tracked omissions documented in #1101 (mozilla-ai/otari-ai#2128 and #1103). Key concealment and the divided dialog layout remain the current product design.

## How to test it locally

1. Build with `make dashboard`, start a standalone gateway with a provider, and sign in.
2. Create and select a fresh workspace, then open Overview. The guide mints its key automatically.
3. Copy the key. The button acknowledges the copy without adding green text below the field. A screen reader still receives a polite confirmation.
4. Switch between Agent, cURL, Python, and TypeScript. Reveal and hide the key; the helper text below each example stays unchanged. Copies still contain the real key.
5. Press Check now on a fast local gateway. The orb dissolves into its working phase, stays there for the original 2.4-second beat, and returns to waiting. Background polling does not animate the button.
6. Send the provided request. The burst originates at the center of the success dialog after its entrance settles, and appears above the dialog and backdrop. Repeat in a fresh workspace after reloading to reset the once-per-page celebration.
7. Repeat with reduced motion enabled and at phone width. The orb stays frozen, and a completed check has no animation hold.

Validation:
- `make lint`: architecture, migration-head, and Ruff checks passed.
- Frontend Biome, `tsc -b --noEmit`, and production Vite build passed.
- Full frontend suite: 188 files, 6,426 tests passed. Focused success-dialog test and typecheck passed after extracting the portal-mounted mark.
- Browser check against a real local gateway and a no-cost mock provider: copy/reveal behavior, working-phase overlap, and the request-to-success transition verified.
[...]

PR #1216, opened 2026-09-16
fix(dashboard): polish activation modal and simplify dismissal

jigjigjig:
## Description

The activation guide now has one clear dismissal action: Skip. The X is removed, and Escape or a backdrop click leaves the guide open, avoiding the temporary dismissal that made it reappear after a reload. Skip keeps its existing workspace-wide dismissal behavior and uses a quieter, borderless ghost treatment.

The listening animation now sweeps around the modal's outer border, in the ink
the last attempt earns: a failed request turns the sweep red, which is what the
design system already specified for it. `ScanBorder` was the documented home for
that arc and has never had a call site, so its job moves onto `Dialog` as
`isScanning` and `scanTone` and the component goes, with the four places that
recorded the old arrangement updated to describe what ships. It stops when the status check fails and remains static for reduced-motion users. Shorter helper text leaves more room below the examples while preserving the explanation that copied snippets contain the real key even when it is hidden.

## How to test it locally

1. Build with `make dashboard`, run a gateway with a configured provider, and open Overview in a fresh workspace.
2. Confirm the activation modal has a scanning outer border and no X. Escape and clicking the backdrop should leave it open.
3. Switch between Agent, cURL, Python, and TypeScript. Check the shorter helper text; revealing the key should not change it.
4. Confirm Skip has no visible border, retains focus and hover feedback, and remains full-width with a 44 px touch target on a phone.
5. Press Skip, reload, and confirm the guide stays dismissed for that workspace.
6. Enable reduced motion and confirm the scanning arc is static. A failed status check should remove the scanning animation.

The footer's phone rules reach a wrapped action as a descendant now. `w-full
sm:w-auto` had stood in for them, and that utility is keyed on the class rather
than the computed width, so it also took the desktop Skip's press away.

A committed `web/node_modules` symlink is untracked, and both ignore files drop
the trailing slash that let a symlink past them.

Validation: `make lint`, frontend Biome, TypeScript, and production build passed. The full dashboard suite passed: 196 files, 6,592 tests. Browser verification covered desktop and 390 px phone layouts, borderless Skip, modal scanning, and Escape/backdrop behavior. No backend, API, dependency, or generated-client changes.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Follow-up to #1167, based on the activation-flow review and local design preview.

## Checklist

- [x] I have added or updated tests that cover my change (`tests/unit`, `tests/integration`). Dashboard tests are colocated under `web/src`.
- [ ] I ran the Definition of Done checks locally (`make lint`, `make typecheck`, `make test`). `make lint` and all dashboard checks passed; Python typecheck and test suites were not run for this frontend-only change.
- [x] If the API contract changed, I regenerated the OpenAPI spec (`uv run python scripts/generate_openapi.py`). No API contract change.

## AI Usage

- [ ] No AI was used.

**Any additional AI details you'd like to share:** Implemented from the user's approved local preview and checked against the repository's frontend and review guidance.

## Summary

- Updated the activation modal so only the server-recorded **Skip** action dismisses it.
- Removed the close button and disabled Escape and backdrop dismissal.
- Added an animated scanning border with reduced-motion support and failure-state styling.
- Shortened helper text while preserving the explanation for hidden keys.
- Preserved full-width mobile Skip behavior and added the related design documentation.
- Replaced the standalone `ScanBorder` component with `Dialog` scanning support.
- Updated tests, stories, styles, and ignore rules for symlinked `node_modules`.

## Validation

Validation covered linting, Biome, TypeScript, production builds, the dashboard test suite, and desktop/mobile browser behavior.


web/src/features/onboarding/SetupSheet.tsx
```diff
@@ -109,6 +109,8 @@ export function SetupSheet({
       }}
       size="lg"
       isAnnouncement
+      isDismissable={false}
```
khaledosman, line 113: `isDismissable={false}` runs against `web/design/overlays.md:140`: "Turning it off takes away Escape and the outside click, and the only honest reason is unsaved work that would be lost." This sheet holds no unsaved work.

The consequence is that on the happy path (a model is served, no failed attempt) Skip is the **only** exit, and Skip is irreversible and workspace-wide. `onDismiss` survives only through the Models link at `:225`, which renders when `model === undefined`, and `ListeningPanel`'s failure-hint link. It also makes the `if (!open) onDismiss()` at `:107` dead code, since `Dialog` gates `onOpenChange` on `isDismissable`.

Fix the reappear-after-reload at its source (persist `isClosed`, or let the close control skip), or amend `overlays.md` to record a second honest reason. Diverging from the rung without saying so is the option to avoid.

web/src/features/onboarding/SetupSheet.tsx
```diff
@@ -136,9 +138,15 @@ export function SetupSheet({
         </p>
       }
       actions={
-        <Button isPending={isSkipping} onPress={onSkip}>
-          Skip
-        </Button>
+        <div className="otari-setup-actions flex w-full">
+          <Button
+            className="w-full sm:w-auto"
```
khaledosman: `w-full sm:w-auto` leaves `w-full` in the class list at every width, and `globals.css:1573` keys the press suppression on the class rather than the computed width:

```css
.button:is(.w-full, .w-full\!, .button--full-width):is(:active, [data-pressed="true"]) { transform: none }
```

So the desktop Skip button, which is auto-width, loses its `scale(0.98)` press. `design/motion-and-access.md:34` reserves that for a control that is actually full width.

The wrapper div is why the utility was needed at all: `globals.css:2292` (`.otari-dialog__actions > button { width: 100% }`) stops matching once the button is nested. Drop the wrapper and keep `<Button>` a direct child of `.otari-dialog__actions`; the phone width, the 44px touch floor and the press suppression then all come back on their own.

web/src/styles/globals.css
```diff
@@ -1786,6 +1786,7 @@ main {
 .otari-bulk-bar .button--ghost,
 .otari-breakdown .button--ghost,
 .otari-rail .button--ghost,
+.otari-setup-actions .button--ghost,
```
khaledosman, line 1792: `.otari-setup-actions` is a place invented for one button. `design/actions.md:131` scopes places to "a container that holds a row of ghosts and the edges read as a grid of boxes", and the comment directly above this selector records why `place-row` was rejected: "a form row's single trailing action reads better WITH the edge." A lone Skip in a footer is that case.

If the edge genuinely reads wrong here, that is an argument about the dialog footer rather than a new place; otherwise drop this selector, and the wrapper div goes with it.

web/src/design-system/feedback/Dialog.tsx
```diff
@@ -58,6 +58,8 @@ export interface DialogProps {
    * get-started strip, a first-run panel", and a first-run sheet is it.
    */
   isAnnouncement?: boolean
+  /** Sweeps the outer border while awaiting an external event. */
+  isScanning?: boolean
```
khaledosman, line 70: New public prop on a design-system component, and `web/design/feedback.md:55` carries `Dialog`'s signature without it. This PR already updates `design/actions.md` for the new place, so the signature list is in scope too.

web/src/features/onboarding/SetupSheet.tsx
```diff
@@ -109,6 +109,8 @@ export function SetupSheet({
       }}
       size="lg"
       isAnnouncement
+      isDismissable={false}
+      isScanning={!checkFailed}
```
khaledosman, line 114: `design/motion-and-access.md:17` still describes this as "the first-run sheet's `ScanBorder` arc and the orb inside the panel it wraps", and `globals.css:3251` still says the resting hairline is "a real `border`, set by `ScanBorder`". Neither holds now: `.modal__dialog` draws no border of its own, so when `checkFailed` flips `isScanning` off the edge disappears entirely instead of resting as a hairline, which is exactly what `ScanBorder`'s contract exists to prevent. `ScanBorder` is also left with no production caller, stories only.

Update both comments, or give the dialog a resting edge so the off state reads the way the design system says it should.

web/src/features/onboarding/SetupSheet.tsx
```diff
@@ -204,8 +212,8 @@ export function SetupSheet({
           </CodeBlock>
           <p className="text-caption text-subtle">
             {tab === "agent"
-              ? "Works with Claude Code, Codex, Cursor, and any agent that can edit files and run commands. It reads the key from your environment rather than carrying it."
-              : "Prefer to have an agent wire this up? The Agent tab is a paste-ready prompt. When the key is hidden, the example shows a stand-in; copying always includes your real key."}
+              ? "Works with Claude Code, Codex, and Cursor. Reads your key from the environment."
```
khaledosman, line 216: "Works with Claude Code, Codex, and Cursor." reads as an exhaustive list where the line it replaces was explicitly not ("and any agent that can edit files and run commands"). Cheap to keep the breadth: "Works with Claude Code, Codex, Cursor, and any agent that can run commands."

khaledosman (changes requested):
Two things want a decision before this lands: the dialog's dismissal contract against `design/overlays.md`, and whether a place is the right tool for a single button. The rest are small.

Checked and deliberately not flagged: the scan ring's geometry is fine (`--radius: 0px` at `globals.css:943`, so `min(32px, var(--radius-3xl))` resolves to 0 and the square `::after` ring has no corner to miss); reduced motion is already covered by the existing block at `globals.css:3310`; `isClosed` is still reachable, so the reworded comment in `SetupGuide.tsx` is accurate; no generated artifacts are owed.

🤖 Reviewed with [Claude Code](https://claude.com/claude-code) (Claude Opus 5)

web/src/features/onboarding/SetupSheet.tsx
jigjigjig: That was on purpose to fix spacing, remove text

jigjigjig (commented):
Self-review. Verified against a real gateway at 1280x720 and 390x844: no Close control, Escape and a backdrop press leave the sheet open, the arc paints (otari-scan, 2.6s, #0098A4), Skip is edgeless with its hover fill and focus ring intact, and 358x44 on the phone sheet. Dashboard lint, typecheck, 196 files / 6592 tests, and the onboarding e2e project all pass.

Findings are about the design system's record of this arc, which this PR is the first code to actually render, and two copy/tone calls.

web/src/design-system/feedback/Dialog.tsx
```diff
@@ -58,6 +58,8 @@ export interface DialogProps {
    * get-started strip, a first-run panel", and a first-run sheet is it.
    */
   isAnnouncement?: boolean
+  /** Sweeps the outer border while awaiting an external event. */
```
jigjigjig, line 69: `web/design/` still records this arc as `ScanBorder`'s, and it is wrong in three places once this lands:

- `feedback.md:37` routes "Is the product waiting on something to arrive?" to `ScanBorder (around the panel that is doing the waiting)".
- `feedback.md:55-57` lists `Dialog`'s signature without `isScanning`.
- `motion-and-access.md:17` names "the first-run sheet's `ScanBorder` arc and the orb inside the panel it wraps". After this PR the arc is the modal's own border and wraps nothing.

Also `styles/globals.css:3252`, inside the rule this PR now consumes: "The element's own resting hairline is a real `border`, set by `ScanBorder`". `Dialog` sets the class with no border of its own. Update all four to describe what ships.

web/src/design-system/feedback/Dialog.tsx
```diff
@@ -151,7 +154,7 @@ export function Dialog({
         >
           <Modal.Dialog
             aria-describedby={description ? descriptionId : undefined}
-            className={`otari-dialog otari-dialog--${size} flex flex-col p-0`}
+            className={`otari-dialog otari-dialog--${size} relative flex flex-col p-0 ${isScanning ? "otari-scan-border" : ""}`}
```
jigjigjig: `ScanBorder` has no call site: not on `main`, not in `web/src` on any `origin/*` branch, and not since `#1101` added it. This PR implements its documented job here instead, which leaves a documented design-system component that nothing renders. Delete it (component, test, story, and its `feedback.md` / `DESIGN.md` entries) or say in the docs that it is the band-level form of the same rule and this is the frame-level one.

web/src/features/onboarding/SetupSheet.tsx
jigjigjig, line 114: `isScanning` carries no tone, so a failed request leaves a teal arc. `feedback.md:199-201` documents the opposite as the correct use (`isActive={!checkFailed} tone={failure ? "danger" : "accent"}`), and this sheet has a real failure state: `ListeningPanel` turns "Request failed" red while the modal's most prominent motion keeps saying all is well. Nothing regressed for users (the arc never rendered before), so this is a call rather than a bug.

web/src/features/onboarding/SetupSheet.tsx
```diff
@@ -136,9 +138,15 @@ export function SetupSheet({
         </p>
       }
       actions={
-        <Button isPending={isSkipping} onPress={onSkip}>
-          Skip
-        </Button>
+        <div className="otari-setup-actions flex w-full">
```
jigjigjig, line 146: The wrapper stops `.otari-dialog__actions > button` from matching, which is where the phone sheet's `width: 100%` and its press-transform suppression came from (`globals.css:2292` and `:2299`). `w-full` restores both (`.button:is(.w-full, ...)` at `:1573`), and measured on a 390px sheet Skip is still 358x44, so no pixel moves. But `w-full` is now load-bearing rather than cosmetic and reads as optional; say so in a comment, or the next person trimming a redundant utility takes the phone layout with it.

web/src/features/onboarding/SetupSheet.tsx
jigjigjig, line 216: This turns a capability into an allowlist. The line it replaces said "any agent that can edit files and run commands"; naming three products reads as the complete set, and an operator on Windsurf or Zed now has reason to think the prompt is not for them. The brevity is worth keeping, the closed set is not.

web/design/actions.md
```diff
@@ -125,7 +125,7 @@ An edge that would be too loud is suppressed by **naming the container**, so a c
 site never has to remember. Inside one of these, a ghost renders edgeless:
 
 `.otari-toolbar` · `.otari-table` · `.otari-pagination` · `.otari-bulk-bar` ·
-`.otari-breakdown` · `.otari-rail`
+`.otari-breakdown` · `.otari-rail` · `.otari-setup-actions`
```
jigjigjig, line 128: The rule two lines down is written for "a new container that holds a row of ghosts" whose "edges read as a grid of boxes". `.otari-setup-actions` holds one button, so the entry reads as an example of a rule it is not an instance of. Either say here what the single-button case is (the sheet's one way out, quieted deliberately) or the list teaches the next reader to name a container per button.

web/src/features/onboarding/SetupSheet.tsx
jigjigjig: Right, and further than I had it: `w-full` is keyed on the class, not the computed width, so `sm:w-auto` never got it back and the desktop press was gone.

Fixed at the source in fd4812e4 rather than by dropping the wrapper: the footer rules are descendants now (`.otari-dialog__actions button`), so they reach a wrapped action. Every other `actions` slot in the tree holds a bare `<Button>`, so nothing else changes. `w-full sm:w-auto` is off the button. Measured at the PR head: desktop 62x36 auto with `w-full` absent from the class list (press restored), 390px sheet 358x44, both unchanged from before.

web/src/features/onboarding/SetupSheet.tsx
jigjigjig: Taking the second half of your option: `overlays.md` now records a second honest reason rather than leaving the divergence silent (fd4812e4).

The reason this sheet has is not unsaved work, it is that its only dismissal is the one the server records. Skip retires the guide; an Escape that put it away for one page load came back on the next, which reads as a dismissal that did not take. Persisting `isClosed` locally would be a third source of truth for a thing the server already owns.

The dead `if (!open) onDismiss()` is real. Leaving it: `onOpenChange` is required by `Dialog` and the handler is correct the moment `isDismissable` is anything but a hardcode.

web/src/styles/globals.css
jigjigjig: The edge is a decision already taken (the approved preview: Skip quieted so the screen's one way out does not compete with the request the screen is asking for), so the open question was the mechanism, not whether.

Keeping the place and saying why, in both homes: `actions.md` now states the single-button case explicitly as the exception rather than leaving it to be read as an instance of the row rule, and the comment above the selector says the same in one line. Your quote about the form row's trailing action is why it is stated rather than quietly added.

web/src/design-system/feedback/Dialog.tsx
jigjigjig: Done in 97f50b86: `feedback.md` carries `isScanning` and `scanTone` in the signature, its decision table routes the wait to `Dialog` rather than `ScanBorder`, and the section is rewritten around the frame.

web/src/features/onboarding/SetupSheet.tsx
jigjigjig: Both comments updated in 97f50b86, and `ScanBorder` is gone with them: no call site on `main`, in `web/src` on any `origin/*` branch, or since #1101 added it, so its job is now `Dialog`'s.

One correction, measured at the PR head against the running gateway: the dialog does have a resting edge on the desktop, `border-top: 1px rgb(90, 95, 107)` on `.modal__dialog`, so `isScanning` off rests as a hairline. It is only the phone sheet that has none, and that is our own geometry rule zeroing it (`globals.css`, the `width <= 639px` block), because a full-screen sheet has no frame to draw. The rule's comment now says it is painted over whatever edge the element has rather than naming `ScanBorder` as the thing that sets one.

The tone was the other half of that contract and it was missing: `scanTone` now carries it, so a failed request sweeps `--color-danger`. Verified in the browser, `--scan-ink: #cb2526` and the arc `rgb(203, 37, 38)`.

web/src/design-system/feedback/Dialog.tsx
jigjigjig: Fixed in 97f50b86: the decision table, the `Dialog` signature, `motion-and-access.md:17` and the rule's own comment in `globals.css` all describe the frame-level arc now.

web/src/design-system/feedback/Dialog.tsx
jigjigjig: Deleted in 97f50b86, component, test and story, with its `feedback.md` section rewritten around `Dialog` and its `DESIGN.md` entry removed. The tone it carried moved with it as `scanTone`.

web/src/features/onboarding/SetupSheet.tsx
jigjigjig: Fede ruled: turn it red. `scanTone={failure ? "danger" : "accent"}` in 97f50b86, with the ink mapped in `Dialog`. Verified in the browser at the PR head: `--scan-ink: #cb2526`, arc `rgb(203, 37, 38)`. Covered by a `Dialog` unit test and by the guide's failed-attempt test, plus a `ScanningFailed` story.

web/src/features/onboarding/SetupSheet.tsx
jigjigjig: Went further after @khaledosman's read of the same line: `w-full` is keyed on the class rather than the computed width, so `sm:w-auto` did not take it back and the desktop press was lost. fd4812e4 makes the footer rules descendants instead, and the utility is off the button. Measured: desktop 62x36, phone 358x44, both unchanged, press restored.

web/src/features/onboarding/SetupSheet.tsx
jigjigjig: Fede ruled: keep the three names, for the space below the examples.

web/design/actions.md
jigjigjig: Said in 97f50b86: `actions.md` now names the single-button case as the exception it is, and the comment above the selector in `globals.css` carries the same line for a reader who only sees the CSS.

web/design/DESIGN.md
```diff
@@ -64,7 +64,7 @@ for a component, not for those.
 | `metrics/SeverityMark` | `SeverityMark`, and the `Severity` type |
 | `metrics/TrendChip` | `TrendChip`, `trendState`, and the `Trend*` types |
 | `metrics/charts` | `TrendChart`, `Sparkline`, `ChartLegend`, and the `SeriesDef` / `StackedPoint` types |
-| `feedback/ErrorBanner` · `/InfoBanner` · `/EmptyState` · `/EmptyMessage` · `/PageLoading` · `/PageError` · `/ConfirmDialog` · `/FormDialog` · `/ScanBorder` · `/ErrorBoundary` · `/Skeleton` · `/errorMessage` | one each |
+| `feedback/ErrorBanner` · `/InfoBanner` · `/EmptyState` · `/EmptyMessage` · `/PageLoading` · `/PageError` · `/ConfirmDialog` · `/FormDialog` · `/ErrorBoundary` · `/Skeleton` · `/errorMessage` | one each |
 | `feedback/Dialog` | `Dialog`, `DialogSection`, and the `DialogSize` type |
```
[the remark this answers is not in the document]
jigjigjig: Right, the row names what the module exports. Added in d3768bad as `DialogSize` / `ScanTone`.

web/src/features/onboarding/SetupSheet.tsx
```diff
@@ -109,6 +109,9 @@ export function SetupSheet({
       }}
       size="lg"
       isAnnouncement
+      isDismissable={false}
```
khaledosman, line 113: `isDismissable={false}` makes the `onOpenChange` handler three lines above unreachable. `Dialog` gates the callback:

```tsx
onOpenChange={(next) => {
  if (next || isDismissable) onOpenChange(next)
}}
```

and this sheet passes `isOpen` as a literal, so `next` is never `true` either. `onDismiss()` on lines 107-109 can no longer fire from here; the prop now reaches the sheet only through `ListeningPanel`'s `onLeave` and the Models link in the body.

Make it `onOpenChange={() => {}}` so a live-looking dismissal path does not sit above the prop that killed it.

web/src/styles/globals.css
```diff
@@ -2286,16 +2290,22 @@ main {
...
-  .otari-dialog__actions > button {
+   * prop, so it would take the desktop with it.
+   *
+   * A descendant rather than a child: a footer whose actions are wrapped (the
+   * first-run sheet wraps Skip to name a place) would otherwise lose the width
+   * and the suppression below, and the `w-full` utility that looks like the
+   * obvious replacement is keyed on the class rather than the computed width,
+   * so it would take the desktop's press with it. */
+  .otari-form-dialog__actions button,
+  .otari-dialog__actions button {
```
khaledosman: Widening `> button` to a descendant reaches every button under either actions slot, not only a wrapped one. It is correct today, since every other `actions` in the tree holds a bare `<Button>`, but the phone rule now also claims an icon-only or secondary button a future slot groups beside a labeled one: 100% width with its press transform suppressed, and nothing fails when that happens.

`:is(> button, > .otari-setup-actions > button)` keeps the reach to the case that needed it, and names the wrapper in the one place a reader of the CSS would look for it.

khaledosman (approved):
Two non-blocking notes inline. Gates checked: no route, schema or docstring change, so no OpenAPI or Postman artifact owed; no `schema.ts` / `routeTree.gen.ts` drift; no new page, so no screenshot entry; no em dashes or `--` separators in the added prose; `ScanBorder` has no leftover reference anywhere under `web/src`, `web/design` or `docs`. `dashboard`, `catalog`, `e2e`, `build` and `serving` are green, and the dashboard job's stderr set is unchanged, so nothing new lands in the console.

_Reviewed with [Claude Code](https://claude.com/claude-code)._

web/src/features/onboarding/SetupSheet.tsx
jigjigjig: Agreed, and I had argued the other way on the first round, which was wrong: a handler that cannot run reads as a path that can. `onOpenChange={() => {}}` in e2334dee, with one line saying why it is empty.

web/src/styles/globals.css
jigjigjig: Good catch, and it is the reason the reach was too wide rather than merely wider. Narrowed in e2334dee.

One spelling note: `:is()` does not take relative selectors (that is `:has()`), so it is written out as `.otari-dialog__actions > button, .otari-dialog__actions > .otari-setup-actions > button`, which names the wrapper in the same place your version would have. No `.otari-form-dialog__actions` twin, since the sheet is a `Dialog`. Re-measured at the PR head: desktop 62x36, phone 358x44, unchanged.

PR #1217, opened 2026-09-16
fix(provider-keys): report a credential the deployment cannot decrypt

khaledosman:
## Description

When a provider key's stored credential can no longer be decrypted, the dashboard showed it as a perfectly normal key. The organization's provider-key list displayed it with the tail of its API key and no mark of any kind, and the workspace view reported it as enabled and in use. Meanwhile the models that provider serves were quietly missing from the Models page, because the catalog does check whether a credential can be read and withholds the provider when it cannot.

The result is a deployment that looks correctly configured and serves nothing. There is no message anywhere, no failed request to notice, and the page you would go to in order to investigate is the one giving the false reassurance. The usual cause is the deployment's encryption secret changing after the credentials were stored, for instance a migration run under a different key.

Provider keys now report whether this deployment can actually read them. An unreadable key is marked in the organization's list beside the existing "default" and "archived" marks, the page explains what happened and that re-entering the API key stores it under the current secret, and the workspace view says so next to "in use", which is the combination that describes the broken state. The key is still listed rather than hidden, because replacing it is what fixes it.

This does not change which keys work. It changes what the dashboard tells you about the ones that do not.

## How to test it locally

1. `make dev`, sign in, and go to Organization, then Provider keys.
2. Add a provider key with any API key value. It lists normally.
3. Restart the gateway with a different `OTARI_SECRET_KEY` so the stored credential can no longer be decrypted.
4. Reload the page. The key now carries a red **UNREADABLE** mark, and a banner explains that it can't be decrypted and that re-entering the API key fixes it.
5. Open Workspaces, then that workspace's provider keys. The row says "Unreadable credential" alongside "In use".
6. Edit the key, paste the API key again, and save. Both the mark and the banner go.

Automated coverage:

- `tests/integration/test_org_provider_keys.py::test_a_key_that_will_not_decrypt_says_so_on_both_surfaces` creates a key, rotates the secret out from under it, and asserts `usable` is false on the organization list and on the workspace view, while `is_effective_enabled` stays true because the workspace disabled nothing.
- `OrganizationProviderKeysPage.test.tsx` covers the mark and the banner, and asserts both stay absent when every key is readable.
- `WorkspaceProviderKeys.test.tsx` pins the "in use" and "unreadable" pairing, since that is the state worth naming.

Run locally: `make lint`, `make typecheck`, `pytest tests/integration/test_org_provider_keys.py tests/integration/test_session_model_catalog_scope.py tests/unit/test_org_provider_key_service.py` (60 passed), `pnpm --dir web run lint`, `pnpm --dir web run typecheck`, `pnpm --dir web test` (6593 passed), plus `make openapi-check` and `make postman-check`. The response models changed, so `docs/public/openapi.json`, the Postman collection and `web/src/client/schema.ts` were regenerated and are in the diff.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

None filed; found while tracing an empty Models page on a hosted deployment back to seven provider keys that every page reported as healthy.

## Checklist

## AI Usage

- [ ] No AI was used.

Note for the reviewer on one deliberate choice: `to_public` now takes `usable` as a required keyword argument rather than computing it. Deriving it inside the model would mean decrypting there, which belongs to the service that owns the secret box, and making it required is what stops a future call site from reporting a key as working without having asked. mypy caught all six existing call sites, which is the intended effect.

Stacked on nothing; independent of #1215, which touches the same page for an unrelated reason.


khaledosman:
@coderabbitai review

PR #1219, opened 2026-09-16
fix(dashboard): polish model catalog navigation and remember its view

jigjigjig:
## Description

Model cards now open from anywhere inside the card, and clickable table rows show a pointer cursor. The catalog remembers the selected List or Table view when returning from a model or reloading the page. The Model pricing and Use this model actions have more space between them.

Card navigation remains a normal link, including keyboard activation and opening in another tab. If browser storage is blocked, the view switch still works for the current visit.

## How to test it locally

1. Start the dashboard with `pnpm --dir web run dev` against a configured gateway and open Models.
2. In List view, click a model's description, metadata, or card padding. Each should open the model; the whole card should show a pointer cursor.
3. Choose Table, click a non-title cell, and return using Back or All models. Table should remain selected, including after a reload. Repeat with List.
4. Open a model as an operator and check the spacing between Model pricing and Use this model.

Validation: dashboard lint, typecheck, Vitest suite, production build, and `make lint`. Browser checks covered card-body navigation, table-cell cursors, Back navigation, and the 20px action gap. The local preview uses sample data and is not part of this PR.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

User-reported dashboard interaction fixes; no linked issue.

## Checklist

Dashboard-only change: dashboard checks and `make lint` were run. No backend behavior or API contract changed; backend typecheck/integration tests and API regeneration were not needed.

## AI Usage

- [ ] No AI was used.

**Any additional AI details you'd like to share:** Implemented and tested with Codex; the user approved the local interactive preview before requesting this PR.


jigjigjig (commented):
Self-review. The navigation fix and the remembered view are right; the findings are about where the pointer cursor is declared, the storage key's shape, and one class assertion in the tests.

web/src/features/models/ModelCatalogPage.tsx
```diff
@@ -636,6 +662,7 @@ export function ModelCatalogView({
                 sortDescriptor={sortDescriptor}
                 onSortChange={onSortChange}
                 onRowAction={onOpen}
+                rowClassName={() => "cursor-pointer"}
```
jigjigjig: Two problems with declaring the row cursor here.

First, `globals.css:1425` states the repo's rule: the pointer cursor is set once, in `@layer base`, "keyed on what the element *is* rather than on a class a call site has to remember". Activity and Usage also pass `onRowAction` and their rows still show an arrow, so this fixes one instance of a class of three. A row is clickable exactly when `onRowAction` is set, and `DataTable` is where that is known.

Second, `DataTable.tsx:320` asks callers to keep `rowClassName` referentially stable: "an inline arrow for any of them rebuilds every row on each render", which is the cache added because checkbox clicks lagged by seconds. `columns={tableColumns()}` on line 664 already breaks it every render, so fixing only this one buys nothing; `tableColumns()` takes no arguments and should be a module-level constant.

web/src/features/models/ModelCatalogPage.tsx
```diff
@@ -382,6 +368,30 @@ function ModelCard({
           </>
         ) : null}
       </p>
+    </>
+  )
+  const cardClass =
+    "group flex cursor-pointer flex-col gap-2 border border-border bg-surface p-4 hover:bg-surface-alt"
```
jigjigjig: `cursor-pointer` is redundant here: the card is a `Link`/`<a href>`, and the user agent already gives any-link a pointer. `rowStyles.ts:114` relies on exactly that ("the leaves ... are `Link`s, so the `<a href>` takes it from the user agent"), and `globals.css` carries no anchor cursor reset. Drop the utility.

Separately, `hover:bg-surface-alt` arrives with no `transition-colors`. Every other use of that fill pairs one with it: `AppShell.tsx:1149` and `:1169` explicitly, `rowStyles.ts` `ROW_RESTING` and `WorkspaceSwitcher.tsx:166` through `NAV_TRANSITION`.

web/src/features/models/ModelCatalogPage.tsx
```diff
@@ -479,7 +489,23 @@ export function ModelCatalogView({
     providers: initialProvider ? [initialProvider] : [],
   })
   const [sort, setSort] = useState("newest")
-  const [view, setView] = useState("list")
+  const [view, setView] = useState(() => {
+    try {
+      return localStorage.getItem("otari.models.view") === "table"
```
jigjigjig: The key is a bare literal repeated on line 504, and it sits outside the namespace every other stored preference uses: `otari.dashboard.sidebarCollapsed`, `otari.dashboard.selectedWorkspace`, `otari.dashboard.theme`, `otari.dashboard.lastRailContext`. Hoist it to a module const named `otari.dashboard.modelsView`, read it through `window.localStorage` like the other five call sites, and put the read in a named function the way `AppShell.tsx:91` and `useTheme.tsx:37` do. The try/catch itself is the right idiom and matches those.

web/src/features/models/ModelCatalogPage.test.tsx
```diff
@@ -293,6 +352,12 @@ describe("ModelCatalogPage", () => {
     const grid = await screen.findByRole("grid", { name: "Models" })
     expect(within(grid).getByText("GLM-5.3")).toBeInTheDocument()
     expect(within(grid).getByText("from $0.50")).toBeInTheDocument()
+    const row = within(grid).getByRole("row", {
+      name: /GLM-5.3/,
+    })
+    expect(row).toHaveClass("cursor-pointer")
```
jigjigjig: This asserts a class name rather than what a user gets, and it pins the cursor to this call site: move the declaration into `DataTable` and the test fails while the behavior improves. The click assertion two lines below is the one that covers the fix. Drop this line.

web/src/features/models/ModelCatalogPage.test.tsx
```diff
@@ -190,8 +190,11 @@ function renderPage(ui: ReactElement, url = "/models") {
 }
 
 describe("ModelCatalogPage", () => {
+  beforeEach(() => localStorage.clear())
```
jigjigjig: `localStorage.clear()` runs here and again in `afterEach`. One of the two is enough; keep the `afterEach` so a test that fails mid-run still leaves nothing behind.

web/src/features/models/ModelCatalogPage.tsx
jigjigjig: 67a11188: DataTable now adds `cursor-pointer` when a row action is set, so Activity and Usage get it too. It keys off a boolean rather than `onRowAction` itself, since every caller passes an inline arrow. `tableColumns()` is now the `TABLE_COLUMNS` constant and `getRowKey` is hoisted, so the row cache can actually pay off. Two tests in `DataTable.test.tsx` cover the clickable and inert cases.

web/src/features/models/ModelCatalogPage.tsx
jigjigjig: 67a11188: `cursor-pointer` dropped from the card, and the hover fill now carries `transition-colors duration-150 ease-out motion-reduce:transition-none`, matching `ToolStatusGroup.tsx:102`.

web/src/features/models/ModelCatalogPage.tsx
jigjigjig: 67a11188: now `VIEW_STORAGE_KEY = "otari.dashboard.modelsView"`, read through `readStoredView()` with the `typeof window` guard and `window.localStorage`, beside the other preference readers.

web/src/features/models/ModelCatalogPage.test.tsx
jigjigjig: 67a11188: dropped. The cursor is covered where it is now declared, in `DataTable.test.tsx`.

web/src/features/models/ModelCatalogPage.test.tsx
jigjigjig: 67a11188: dropped the `beforeEach`; the `afterEach` clear stays.

jigjigjig:
@coderabbitai review

web/src/features/models/ModelCatalogPage.tsx
```diff
@@ -382,81 +368,122 @@ function ModelCard({
           </>
         ) : null}
       </p>
+    </>
+  )
+  const cardClass =
+    "group flex flex-col gap-2 border border-border bg-surface p-4 transition-colors duration-150 ease-out hover:bg-surface-alt motion-reduce:transition-none"
+  return (
+    <article>
```
khaledosman, line 376: The card is now one anchor over its whole body, so a drag-select that starts and ends inside it navigates on mouseup: highlighting a model's description, or `$0.50/M input tokens` to paste into a budget, leaves the page.

The table view of the same rows deliberately does not do this. `globals.css:2549` makes row text selectable (`user-select: text !important`, issue #478) and `DataTable` guards the resulting click with `hasTextSelectionIn` before firing the row action. After this change list and table disagree about what a highlight does.

A guard on the card that suppresses only a plain click keeps everything the comment deleted with the old title link was protecting:

```tsx
onClick={(e) => {
  if (e.metaKey || e.ctrlKey || e.shiftKey || e.altKey) return
  const selection = window.getSelection()
  if (selection && !selection.isCollapsed) e.preventDefault()
}}
```

Cmd/Ctrl-click and middle-click never reach it, so opening in a new tab is untouched. `hasTextSelectionIn` is module-private in `DataTable.tsx`; exporting it would keep the two paths on one definition.

web/src/features/models/ModelCatalogPage.tsx
```diff
@@ -382,81 +368,122 @@ function ModelCard({
...
   { value: "table", label: "Table" },
 ]
 
+// Hoisted beside the columns: DataTable caches its rendered rows on these two,
+// and an inline arrow would rebuild every row on each render.
+const rowKey = (row: CatalogModelSummary) => row.id
+
+const VIEW_STORAGE_KEY = "otari.dashboard.modelsView"
+
+function readStoredView(): string {
```
khaledosman: `: string` re-widens the view at exactly the point an untrusted value enters. `shared/hooks/useTheme.tsx:11-36` is the same read one rung up, and it is the precedent `typescript-and-react.md` cites for this:

```ts
export const THEME_PREFERENCES = ["system", "light", "dark"] as const
export type ThemePreference = (typeof THEME_PREFERENCES)[number]
function isPreference(value: string | null): value is ThemePreference { ... }
function readStored(): ThemePreference { ... }
```

`as const satisfies` on `VIEW_OPTIONS` plus a `CatalogView` union gives `readStoredView` and `changeView` the same shape, and turns "falls back to list for an invalid saved view" into a narrowing the compiler enforces as well as the test.

khaledosman (approved):
Two non-blocking notes inline. Gates checked: no route, schema or docstring change, so no OpenAPI or Postman artifact owed; no `schema.ts` / `routeTree.gen.ts` drift; no new page, so no screenshot entry; no em dashes or `--` separators in the added prose. `dashboard`, `catalog`, `e2e`, `build` and `serving` are green, and the dashboard job's stderr set is unchanged, so nothing new lands in the console. Moving the row cursor into `DataTable` also picks up Activity and Usage, and `renderRow` is in `Table.Body`'s `dependencies`, so the boolean invalidates the row cache correctly.

_Reviewed with [Claude Code](https://claude.com/claude-code)._

web/src/features/models/ModelCatalogPage.tsx
jigjigjig: 6616c81a: `CATALOG_VIEWS` / `CatalogView` with an `isCatalogView` guard, `as const satisfies` on `VIEW_OPTIONS`, and `changeView` narrowing on the way back from `Segmented` (whose `onChange` is `(next: string) => void`, so the widening has to be undone rather than avoided). Both paths fall to the same default.

web/src/features/models/ModelCatalogPage.tsx
jigjigjig: Measured this one before changing it, and the premise does not hold: a drag inside the card does not navigate, because it never becomes a selection in the first place. Chromium treats a drag that starts inside an anchor as a link drag.

Staged the card as plain HTML and drove it with Playwright:

| action | `getSelection()` | navigated |
| --- | --- | --- |
| plain click | `""` | yes |
| drag across the price | `""` | **no** |
| double-click the price | `""` | yes |
| click with a selection made elsewhere | n/a | yes |

So the guard would never fire: the selection inside the card is always collapsed, and `hasTextSelectionIn` would return false on every click. Adding `draggable={false}` to free the drag makes it worse, not better, and is what actually produces the reported failure: drag then selects nothing **and** navigates on mouseup.

The real cost is larger than the one described and has no fix at this shape: card text is not selectable at all, where `globals.css:2549` makes table row text selectable on purpose. A stretched-link overlay does not recover it either, since the overlay takes the mousedown. So whole-card click and selectable card text are exclusive, and this PR chose the click.

Leaving the card as it is and not adding a guard that cannot run. The price and the context are both in the table view, which stays selectable. Happy to file an issue for selection parity in list view if you want it tracked; say the word and I will.

Issue #1218, opened 2026-09-16
feat(auth): adopt newsletter opt-in with consent persistence and account preferences


jigjigjig:
## Problem

The Otari registration page has no newsletter opt-in. The current signup API accepts email, password, full name, and terms acceptance, but cannot record a newsletter preference. The existing otari-ai product already has an optional checkbox, consent fields and timestamps, and a privacy-settings control. Adopt that behavior through the appropriate OSS/hosted boundary rather than adding a checkbox that discards its value.

## Design requirements

- Add an optional, unchecked-by-default checkbox to registration, directly below the terms row when present and above Create account. Keep newsletter consent separate from required terms acceptance. Leaving it unchecked must never block registration.
- Proposed copy: “Subscribe to our newsletter for product updates (optional).” Include a clearly labeled privacy-policy link when this deployment publishes one. Do not hardcode a Mozilla marketing destination into arbitrary self-hosted deployments.
- Use the existing HeroUI/design-system Checkbox, semantic tokens, square controls, Mozilla typography, and light/dark themes. Associate the full label with the control; support keyboard Space, a visible focus ring, and a minimum 44px touch target. Links inside the label must navigate without inadvertently changing consent.
- Keep the centered auth shell, existing animated background, and compact account/Help footer. Fit the default registration state, including terms and newsletter, at 1280 × 720 and 390 × 844 with no document scrolling. Reduce redundant copy/spacing rather than text size or touch targets. On shorter viewports, zoom, validation errors, or an open keyboard, all controls must remain reachable without clipping.
- Preserve the choice through validation failures and retriable signup errors. Do not infer consent from terms acceptance, email verification, OAuth, passkey use, invitations, or a pre-existing account.
- Provide a signed-in account/privacy setting to opt in or withdraw later, with loading, saved, and recoverable error states. If OAuth/invitation onboarding offers the checkbox, use the same consent contract and defaults.
- Publish updated desktop/mobile and light/dark designs for review. The current auth design is a baseline, not an already approved newsletter design: https://app.paper.design/file/01M2MTEMV01ST3SG1YV6A1JCGN

## API and backend requirements

- Add a backward-compatible `newsletter_opt_in: bool = false` to the applicable signup request and persist explicit consent for a successfully created/claimed identity. Validate the field with the existing request-validation conventions.
[...]

PR #1220, opened 2026-09-16
fix(dashboard): center auth forms and simplify secondary actions

jigjigjig:
## Description

Login and registration could extend below the viewport because the auth panel used a fixed top offset and stacked several secondary links. Center the shared panel within the available viewport, compact the forms, and move verification recovery and the welcome guide into an accessible Help popover. Password recovery stays beside the password field, and OAuth providers share a row.

The existing animated background, auth behavior, and deployment-specific signup eligibility remain intact. Empty field-message rows are omitted when there is no description. Short windows, enlarged text, and longer error states can still scroll so controls are never clipped.

[Paper design: desktop/mobile login, registration, and Help](https://app.paper.design/file/01M2MTEMV01ST3SG1YV6A1JCGN)

## How to test it locally

1. Start a gateway with password signup and mail enabled, or run `OTARI_DEV_API=https://otari.mzai.dev pnpm --dir web run dev` to use the dev deployment's public bootstrap configuration.
2. Open the login page and `/#/signup` at 1280 × 720 and 390 × 844. Both default forms should be vertically and horizontally centered without document scrolling. Check both themes.
3. Open Help, check the verification and welcome destinations, then press Escape and verify focus returns to Help. Confirm password recovery remains available and OAuth buttons keep their full accessible names.
4. Check a short viewport and validation errors: all form controls must remain reachable.

Validated locally: `make lint` (architecture, migration graph, Ruff), dashboard Biome, TypeScript project typecheck, production build, all 139 auth tests, and design-foundation checks (3,837 assertions with the auth suite). The full dashboard suite passed 6,597 of 6,600 tests initially; the three failures were stale login-heading assertions, updated and rechecked in the app suite. Browser measurements at both dimensions confirmed zero center offset and no document overflow. Keyboard dismissal of Help returns focus to its trigger.

The installed local tool binaries were invoked directly because pnpm's automatic version switch could not verify its registry download. No dependency or lockfile changes are included.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Addresses the dev-dashboard layout feedback. Newsletter opt-in is separate follow-up work in #1218; this PR does not implement or close it.

## Checklist

Frontend-only change; no API contract, generated client, or public documentation changes are needed. The full Python test/typecheck gates are not applicable to the changed files.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:** GPT-6 in Codex; Paper for design review.

**Any additional AI details you'd like to share:** Implemented from user feedback and checked with component tests, design-system checks, a production build, and browser interaction/geometry measurements.


web/src/features/auth/Login.tsx
```diff
@@ -54,7 +55,7 @@ const ERROR_IDS: Record<CredentialField, string> = {
   masterKey: "login-master-key-error",
 }
 
-const CARD = "flex flex-col gap-6"
+const CARD = "flex flex-col gap-4"
```
jigjigjig, line 58: `CARD` and `CARD_FLAT` now hold the identical string, so the two names distinguish nothing. Collapse to one constant and use it for all three branches.

web/src/features/auth/AuthFields.tsx
```diff
@@ -59,13 +59,11 @@ export function AuthEmailField({
         autoComplete="username"
         className="read-only:bg-surface-alt read-only:text-muted"
       />
-      <FieldMessages>
-        {description ? (
-          // HeroUI's Description reaches the input as aria-describedby through
-          // the TextField's "description" slot, which a raw span does not.
+      {description ? (
+        <FieldMessages>
```
jigjigjig, line 63: The comment explaining why this is HeroUI's `Description` rather than a raw span (it reaches the input as `aria-describedby` through the TextField's `description` slot) was dropped here and in the two fields below, while the code still depends on that fact. Moving the conditional outward is the right fix for the empty row; keep the one line that stops someone swapping `Description` for a span. Same in `AuthPasswordField` and `AuthTextField`.

web/src/features/auth/AuthHelp.tsx
```diff
@@ -0,0 +1,57 @@
...
+          Open the welcome guide
+        </Link>
+        {usesPassword !== undefined ? (
+          <p className="border-t border-border pt-3 text-caption">
+            {usesPassword ? (
+              "Your password is sent once and exchanged for a session cookie. It is never stored in the browser."
+            ) : (
+              <>
+                Your{" "}
+                <a href="/welcome" className="text-link">
```
jigjigjig: This `/welcome` link has no hover state, while the `Link` twelve lines above and every `PublicAuthLink` in the popover carry `hover:text-link-hover`. Give it the same treatment, or use `Link` as the sibling does.

web/src/features/auth/LoginPageShell.tsx
```diff
@@ -16,7 +16,7 @@ export function LoginPageShell({ children }: { children: ReactNode }) {
     preference === "system" ? FiMonitor : preference === "dark" ? FiMoon : FiSun
 
   return (
-    <div className="relative isolate flex min-h-svh flex-col bg-background">
+    <div className="relative isolate flex min-h-dvh flex-col bg-background">
```
jigjigjig: `svh` to `dvh` works against the goal here. `dvh` tracks the mobile URL bar, so once a short viewport does scroll, the centered panel re-centers under the finger as the bar retracts. `svh` is the smallest viewport, which is the one "fits without scrolling" has to be true of. Revert to `min-h-svh` unless the centering was measured against a retracting bar.

jigjigjig (commented):
Self-review. Measured at 1280x720, 390x844 and two 500px-tall viewports: the panel centers exactly (equal gap above and below, zero horizontal offset), nothing overflows at full height, and a short viewport scrolls the document rather than clipping the panel, so the top stays reachable. Help fits at 390px (352px wide) and its links keep 44px rows. Dashboard lint, typecheck and all 6600 tests pass.
Four findings inline. One more that has no diff line to sit on: `AcceptInvitationPage.tsx` renders the same `LoginPageShell` and still carries the old footer shape this PR replaced on the other two public pages (`border-t ... pt-5` and "New to Otari? Open the welcome guide"), so the three pages on one shell now space and word their footers differently.

web/src/features/auth/Login.tsx
jigjigjig: Addressed in 5985a0d0: CARD_FLAT is gone and all three branches take CARD.

web/src/features/auth/AuthFields.tsx
jigjigjig: Addressed in 5985a0d0: the aria-describedby note is back above the first Description, covering all three fields.

web/src/features/auth/AuthHelp.tsx
jigjigjig: Addressed in 5985a0d0: the link now carries hover:text-link-hover like its siblings.

web/src/features/auth/LoginPageShell.tsx
jigjigjig: Addressed in 5985a0d0: back to min-h-svh. Re-measured after the change, centering is unchanged at 1280x720, 390x844 and both 500px-tall viewports.

jigjigjig:
The fifth finding from the review body (the invitation page keeping the old footer) is fixed in 5985a0d0 too: same rule spacing, same 44px hit area, same wording as the other two pages on this shell.

web/src/features/auth/Login.test.tsx
```diff
@@ -131,7 +131,7 @@ describe("Login", () => {
     expect(screen.queryByRole("link", { name: /verification link/ })).toBeNull()
```
khaledosman, line 135: This assertion no longer proves anything. `/resend-verification` moved into `AuthHelp`'s popover, which renders nothing while closed, so `queryByRole` returns null whatever `mail_ready` is. The test is named for otari#650's "hidden rather than offered and then refused with a 503" rule and no longer holds it.

Same defect in `hides recovery on an unclaimed deployment` below (lines 161 and 163): with the master-key box showing, *both* `/Forgot your password/` (now the `usesPassword === false` popover branch) and `/verification link/` sit inside the closed popover, so neither assertion tests `offersRecovery`.

Open Help first in all three places, then assert absence.

web/src/features/auth/AuthHelp.tsx
```diff
@@ -0,0 +1,57 @@
+import { Link } from "@heroui/react"
+import { Popover } from "@/design-system/overlays/Popover"
+import { PublicAuthLink } from "./PublicAuthLayout"
+
+export function AuthHelp({
+  offersRecovery,
+  usesPassword,
+}: {
+  offersRecovery: boolean
+  usesPassword?: boolean
```
khaledosman: `usesPassword?: boolean` is a three-state discriminator: `true` is the password box, `false` is the master-key box, and `undefined` means "not the sign-in page at all" (`SignupPage`'s call, which is what keeps line 22 and the credential note off that page). Nothing here says so, so a caller that reasonably passes `usesPassword={false}` from signup gets a "Forgot your password?" link on a form whose job is to set one.

`frontend-standards/typescript-and-react.md`: "A discriminated union beats a bag of optionals." Either split the prop into the two questions it is answering, or comment the three states.

web/src/features/auth/AuthHelp.tsx
```diff
@@ -0,0 +1,57 @@
...
+    <Popover
+      trigger={
+        <span className="inline-flex min-h-11 items-center px-3 text-sm font-medium text-link">
+          Help
+        </span>
+      }
+      placement="top"
+    >
+      <div className="flex max-w-xs flex-col gap-2">
+        {offersRecovery && usesPassword === false ? (
```
khaledosman: No test reaches this branch. The two `Login.test.tsx` cases that open Help run with `mail_ready` false or without `password` in `sign_in_methods`, so `offersRecovery && usesPassword === false` is never both-true anywhere in the suite.

The file also has no colocated `AuthHelp.test.tsx`; the standards ask for `Foo.tsx` → `Foo.test.tsx` on a new component.

web/src/features/auth/Login.tsx
```diff
@@ -575,22 +574,17 @@ export function Login() {
...
     <LoginPageShell>
       <div className={CARD}>
-        <div className="flex flex-col gap-1.5">
-          <h1 className={HEADING}>Otari</h1>
+        <div className="flex flex-col gap-1.5 text-center">
+          <h1 className={HEADING}>Sign in to Otari</h1>
           <p className="text-sm text-pretty text-muted">
             {usesPassword
-              ? "Sign in to browse models, set pricing, and manage settings."
+              ? "One account for every model."
```
khaledosman, line 586: "One account for every model." restates the shell footer's "One gateway. Every model." two rows below it, and drops what the screen is for, which the master-key branch on the next line still states. The dashboard is an operator tool rather than a marketing surface; keep the password branch saying what signing in gets you.

web/src/features/auth/Login.tsx
```diff
@@ -641,11 +635,18 @@ export function Login() {
                 isInvalid={errorField === "password"}
                 className="flex flex-col gap-2"
               >
-                <LabelRow
-                  label="Password"
-                  error={errorField === "password" ? error : null}
-                  errorId={ERROR_IDS.password}
-                />
+                <div className="flex flex-wrap items-center justify-between gap-x-2">
```
khaledosman, line 643: Wrapping `LabelRow` in a second `justify-between` row makes `LabelRow`'s own `justify-between` inert: it is now a content-sized flex item with no free space to distribute, so an inline password refusal sits beside the label instead of at the field's right edge where it used to. Add `flex-1` to `LabelRow` if that alignment was intended.

khaledosman (approved):
Centering and the compaction look right: `min-h-svh` on the shell keeps `justify-center` from stranding a tall panel above the scroll origin, and the OAuth row caps at two providers so `flex-1` cannot squeeze below a readable width at 390px.

Six findings inline. The test ones are the substantive pair: moving the resend-verification link behind a popover turned two `queryByRole(...).toBeNull()` assertions into assertions about the popover being shut.

---

web/src/features/auth/Login.test.tsx
jigjigjig: Addressed in 3d9e73cc: both tests open Help before asserting absence. One extra wrinkle worth recording, since it bit on the first attempt: the open popover is a dialog, so it takes the rest of the page out of the accessibility tree behind it. The "Set your password" assertion on the unclaimed test had to move ahead of the click, or it fails to find a link that is on the page.

web/src/features/auth/AuthHelp.tsx
jigjigjig: Addressed in 3d9e73cc: the prop is now `credential?: "password" | "master-key"`, so the three states are named rather than encoded in a boolean and its absence, and a signup caller has no false to pass.

web/src/features/auth/AuthHelp.tsx
jigjigjig: Addressed in 3d9e73cc: `AuthHelp.test.tsx` covers all three credential states, including the recovery-plus-master-key branch nothing reached before.

web/src/features/auth/Login.tsx
jigjigjig: Half right, and I measured before acting on it. The premise holds: `LabelRow` is content-sized in that wrapper, 77px inside 382px, so its `justify-between` has no free space. But `flex-1` does not restore the old alignment, it makes things worse. The recovery link holds 159px of the 382px row, which leaves 215px, and "Password" plus "Enter your password." needs 231px. So with `flex-1` the refusal wraps onto a second line and the row grows from 20px to 39px each time one appears, pushing the input down.

Measured at 1280px, password refusal showing:
- with `flex-1`: error at 449-603 on a second line, row 39px
- without: error at 538-692 beside the label, row 20px, no shift

Left content-sized, with a comment at `LabelRow` saying why. Fede picked the one-line version over both `flex-1` and moving the recovery link below the input.

web/src/features/auth/Login.tsx
jigjigjig: Taking this one to Fede rather than deciding it, and he kept "One account for every model." The echo of the footer is real and noted; he wants the shorter line, which is what the compaction in this PR is for. Leaving the copy as it stands.

PR #1225, opened 2026-09-16
fix(web): clarify API key actions and responsive layouts

jigjigjig:
## Description
API key rows now offer a labeled Actions menu instead of an unexplained strip of icons. The three-dot trigger keeps its intended size, and Edit opens the existing modal with focus returning to its row when closed. Regeneration uses a confirmation modal and displays failures there.

Rows remain compact as the available width shrinks: secondary columns move into the menu details, and phone layouts use a selectable list. Long names and owners truncate in the list while their full values remain available in the menu. Existing design tokens, HeroUI controls, key suffixes, and live-key deletion safeguards are preserved.

## How to test it locally
1. Start the gateway and dashboard with `OTARI_DEV_API=http://127.0.0.1:8090 pnpm --dir web dev` (adjust the backend port as needed).
2. Open API keys. Open a row's Actions menu using the mouse and keyboard; confirm labels and disabled deletion for active keys.
3. Choose Edit, change a field, and verify Save and discard confirmation. Cancel an unchanged edit and check focus returns to its Actions trigger.
4. Resize from desktop to phone width. Check row identification, selection, prefix copying, and full details in the menu.
5. Cancel regeneration, then verify regeneration and deletion with disposable keys.

Validated locally: `make lint`, dashboard Biome, `tsc -b --noEmit`, production Vite build, and all 6,603 Vitest tests across 196 files (including 47 API Keys tests). Dashboard checks used installed tool binaries because the local pnpm launcher could not resolve its pinned version. The management E2E selectors were updated; that suite was not rerun.

## PR Type
- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues
User-requested API Keys usability improvements; no linked issue.

## Checklist

Frontend-only change; dashboard checks and repository lint passed. Backend typecheck/tests were not run. No API contract or documentation changes were needed.

## AI Usage
- [ ] No AI was used.

**Any additional AI details you'd like to share:** Implemented from the supplied Paper design and user feedback, then rebased onto current main and validated locally.

## Summary

- Replaced API key row icons with an accessible Actions menu.
- Added edit, regeneration, enable/disable, and delete actions with confirmation and focus restoration.
- Added responsive layouts for desktop, compact, and phone screens.
- Moved secondary key details into menus on smaller layouts.
- Added truncation, selectable phone lists, key fingerprints, and improved touch targets.
- Updated API key tests and management E2E selectors.

## Technical notes

- Added `KeyActionsMenu`, `keyFingerprint`, and `Checkbox.hasTouchTarget`.
- Preserved deletion safeguards for active and pending keys.
- Local validation included linting, type checking, production build, and 6,603 Vitest tests.
- Management E2E, backend type checking, and backend tests were not run.


web/src/features/keys/KeysPage.tsx
```diff
@@ -1037,234 +975,168 @@ export function KeysPage() {
...
+      id: "status",
+      header: "Status",
+      cell: (k) => <StatusMark apiKey={k} />,
+    },
+    // Every key on a member's page is their own, so an Owner column there
+    // would repeat one name down the table.
+    ...(isDeploymentWide
+      ? [
+          {
+            id: "owner",
```
jigjigjig: Owner is flattened to one face. `ownerLabel` feeds every owner into `text-mono-caption text-muted`, so a member alias like `Operator` now renders in monospace. The Direction A board at 1440 keeps two faces: a member alias in the body face, a raw id and `virtual` in mono. The `title={k.user_id}` that kept the raw id recoverable from this column is also gone. Restore both.

web/src/features/keys/KeysPage.tsx
```diff
@@ -1037,234 +975,168 @@ export function KeysPage() {
...
+          </span>
+          <div className="truncate text-caption">
+            {layout === "compact" && isDeploymentWide ? (
+              <>
+                <span>{ownerLabel(k)}</span>
+                {" · "}
+              </>
+            ) : null}
+            <KeyMetaLine apiKey={k} />
+            {layout === "compact"
```
jigjigjig: The compact fold line renders in three faces. `KeyMetaLine` is `text-mono-overline` (11px mono, 0.1em tracking, uppercase), so the line reads `alice@example.com · ALL MODELS · Used 3d ago` with 13px sans on either side of an uppercase mono middle. The 1024 board is one uniform sentence-case line: `alice@example.com · All models · used 3d ago`. Give the meta line the caption face when it folds, and lowercase `Used`.

web/src/features/keys/KeyActionsMenu.tsx
```diff
@@ -0,0 +1,173 @@
...
+      >
+        <div className="flex flex-col gap-1 border-b border-border px-3 py-3">
+          <p className="break-words text-body">{name}</p>
+          {owner ? <p className="break-all text-caption">{owner}</p> : null}
+          <p className="break-all text-mono-caption text-muted">
+            {apiKey.key_prefix
+              ? `${apiKey.key_prefix}…${apiKey.key_suffix ?? ""}`
+              : "No key prefix"}
+          </p>
+          <dl className="flex flex-col gap-1 pt-2 text-caption">
```
jigjigjig: This details block is not in the approved menu. The board's header is name, owner, prefix and suffix, three lines; this adds six more (Created, Last used, Expires, Models, and conditionally Budget and User matching) on every layout, including 1440 where those same columns are on screen a few hundred pixels to the left. On a phone it is the only place those facts survive, which is the argument for keeping it there. Asking Fede whether it stays everywhere, only below `wide`, or not at all.

web/src/features/keys/KeysPage.tsx
```diff
@@ -1037,234 +975,168 @@ export function KeysPage() {
...
-        <ArmedStrip
-          confirmRef={confirmRef}
-          confirmLabel="Regenerate"
-          isPending={rotateKey.isPending}
-          message={
-            <>
-              Regenerate the secret for <strong>{label(k)}</strong>? The current
-              secret stops working immediately, with no grace period.
-            </>
+  const columns: DataTableColumn<ApiKey>[] = [
```
jigjigjig: `columns` lost its `useMemo`, so it is a new array on every render and `visibleColumns` with it. DataTable's docstring is explicit that callers must keep `columns` referentially stable or the per-row cache rebuilds every row on each render, and it names the symptom that motivated the cache: checkbox clicks lagging by whole seconds at large page sizes. This table is the selection-heavy one. Memoize on what the cells read: `layout`, `isDeploymentWide`, `memberLabels`, the two pending flags, `setActive`, `lastAction`.

web/src/features/keys/KeysPage.tsx
```diff
@@ -1037,234 +975,168 @@ export function KeysPage() {
...
-      triggerRef,
-      lastArmed,
-    ],
+  const ownerLabel = (k: ApiKey) =>
+    isVirtualUser(k.user_id)
+      ? "virtual"
+      : k.user_id
+        ? (memberLabels.get(k.user_id) ?? k.user_id)
+        : "—"
+  const renderActions = (k: ApiKey) => (
```
jigjigjig: `renderActions` and `renderPrefix` (1026) are rebuilt each render too, and both are closed over by the column cells, so memoizing `columns` alone will not hold. Wrap both in `useCallback` with the same inputs.

web/src/features/keys/KeysPage.tsx
```diff
@@ -1037,234 +975,168 @@ export function KeysPage() {
...
-  }, [])
-  const { triggerRef, confirmRef } = useConfirmationFocus(armed !== null)
+  const [pendingRegenerate, setPendingRegenerate] = useState<ApiKey>()
   const [pendingDelete, setPendingDelete] = useState<ApiKey>()
+  const [lastAction, setLastAction] = useState<string>()
+  const { triggerRef: actionTriggerRef } = useConfirmationFocus(
+    !!(pendingRegenerate || pendingDelete || editing || regenerated),
+  )
+  const tableRegion = useRef<HTMLDivElement>(null)
+  const [layout, setLayout] = useState<"wide" | "compact" | "mobile">("wide")
```
jigjigjig: `layout` starts at `"wide"`, so a phone paints the full desktop table for a frame and then swaps to the list once the ResizeObserver fires. Seed the initial state from `window.innerWidth`. Separately, `window.innerWidth` is only read inside the observer callback, so a viewport change that leaves the region's content width alone never re-evaluates it.

web/src/features/keys/KeysPage.test.tsx
```diff
@@ -603,7 +644,7 @@ describe("KeysPage", () => {
       within(row).queryByRole("button", { name: "Delete" }),
```
jigjigjig: This assertion is now tautological: there is no `Delete` button anywhere in the new UI, so it passes whatever the guard does. It was the check that a live key withholds permanent delete. Replace it with the menu check, `getByRole("menuitem", { name: /^Delete/ })` carrying `aria-disabled="true"`, which is what actually enforces it now.

web/src/features/keys/KeyActionsMenu.tsx
```diff
@@ -0,0 +1,173 @@
...
+      <Dropdown.Popover
+        placement="bottom end"
+        className="otari-key-actions-menu"
+      >
+        <div className="flex flex-col gap-1 border-b border-border px-3 py-3">
+          <p className="break-words text-body">{name}</p>
+          {owner ? <p className="break-all text-caption">{owner}</p> : null}
+          <p className="break-all text-mono-caption text-muted">
+            {apiKey.key_prefix
+              ? `${apiKey.key_prefix}…${apiKey.key_suffix ?? ""}`
```
jigjigjig: `${key_prefix}…${key_suffix ?? ""}` is spelled here and again in `KeysPage.renderPrefix`. One helper, two call sites.

web/src/features/keys/KeysPage.test.tsx
```diff
@@ -627,16 +670,8 @@ describe("KeysPage", () => {
...
-    // message and the confirm are siblings of that row rather than inside it.
-    const armed = screen
-      .getByText(/stops working immediately/)
-      .closest("tr") as HTMLTableRowElement
-    expect(armed).not.toBe(row)
-    expect(row.nextElementSibling).toBe(armed)
-    // The message names the key it is about, which is the whole point of
-    // confirming in place rather than in a dialog.
+    await chooseAction(user, row, "Regenerate")
+    const armed = await screen.findByRole("alertdialog")
```
jigjigjig: `armed` is a leftover name from the strip; it is an alertdialog now. Same at 826.

web/src/styles/globals.css
```diff
@@ -3097,42 +3097,106 @@ main {
...
+}
+.otari-keys-table .table__column {
+  white-space: nowrap;
+}
+.otari-keys-table .table__column:first-child {
+  width: 3rem;
 }
 .otari-keys-table .table__column[data-key="status"] {
-  width: 120px;
+  width: 7.5rem;
```
jigjigjig, line 3178: Five of these widths are px-to-rem conversions that compute to the same pixel (120→7.5rem, 170→10.625rem, 110→6.875rem, 120→7.5rem), which is drive-by reformatting that hides the one real change among them: `key` goes 150px→11.25rem, a 30px widening, needed for the suffix plus the CopyButton. Worth saying so in the commit rather than leaving it to be found.

jigjigjig (commented):
Self-review. The menu matches the approved Direction A board on the things it was drawn for: 268px popover, 36/44px items, 14px glyphs, 32/44px trigger, Delete disabled with its reason, and the 1024 fold to Name/Status/Key/Actions.
Three divergences from that board and one perf regression the DataTable docstring names explicitly. Details inline.

web/src/features/keys/KeysPage.tsx
jigjigjig: Addressed in 95fa8997: the Owner cell branches again, a member alias in `text-sm text-foreground` and a raw id or virtual owner in `text-mono-caption`, with `title` carrying the raw id in both. Covered by a new test that pins the wide layout.

web/src/features/keys/KeysPage.tsx
jigjigjig: Addressed in 95fa8997: `KeyMetaLine` takes a `face` prop and the folded line passes `text-caption`, so the whole line is one face; `used` is lowercase now. The mobile list takes the same treatment.

web/src/features/keys/KeyActionsMenu.tsx
jigjigjig: Fede chose: only below the wide layout. Addressed in 95fa8997 behind a `showDetails` prop, so the menu matches the board at 1440 and keeps the folded lanes readable at 1024 and on a phone. Both branches are covered (the 390 test asserts the block, the new 1440 test asserts its absence).

web/src/features/keys/KeysPage.tsx
jigjigjig: Addressed in 95fa8997: `columns` is back behind `useMemo` on layout, isDeploymentWide, memberLabels, ownerLabel and the two renderers, and `visibleColumns` is memoized on it.

web/src/features/keys/KeysPage.tsx
jigjigjig: Addressed in 95fa8997: `renderActions`, `renderPrefix` and `ownerLabel` are all `useCallback` now, which is what lets the columns memo actually hold.

web/src/features/keys/KeysPage.tsx
jigjigjig: Addressed in 95fa8997: a `layoutFor` helper seeds the state from `window.innerWidth`, and a `resize` listener covers a viewport crossing md that leaves the region width alone.

web/src/features/keys/KeysPage.test.tsx
jigjigjig: Addressed in 95fa8997: the guard is asserted on the menu item, both `aria-disabled` and the visible reason "Disable it first".

web/src/features/keys/KeyActionsMenu.tsx
jigjigjig: Addressed in 95fa8997: one `keyFingerprint` in `secretCaption.ts`, called from both. It went there rather than into `KeysPage` to avoid a cycle, since `KeysPage` imports the menu.

web/src/features/keys/KeysPage.test.tsx
jigjigjig: Addressed in 95fa8997: both renamed to `confirm`.

web/src/styles/globals.css
jigjigjig: Noted rather than reverted: unwinding four no-op conversions would churn the diff a second time for no rendered change. The real one is called out in the commit message, key 150px to 180px for the suffix plus the CopyButton.

web/src/styles/globals.css
```diff
@@ -3097,42 +3097,106 @@ main {
...
 }
 .otari-keys-table .table__column[data-key="actions"] {
-  width: 192px;
+  width: 5rem;
 }
-.otari-keys-table .otari-table .otari-detail-row > td {
-  background-color: transparent;
+.otari-keys-list label[data-rac] {
```
khaledosman: Reaches into react-aria's internal `data-rac` marker to force a 44px target on our own `Checkbox`, which takes no `className` (`design-system/forms/Checkbox.tsx:96`). Rung 2 puts a component prop above a CSS rule, and scoping it to `.otari-keys-list` fixes one selector rather than the value: every other below-`md` list keeps the 16px box.

Put the affordance on `Checkbox` itself — `CopyButton`'s `before:-inset-1.5` device (`CopyButton.tsx:77`) already solves the same 44px problem in the tree.

web/src/features/keys/KeyActionsMenu.tsx
```diff
@@ -0,0 +1,181 @@
...
+      label: "Delete…",
+      hint: apiKey.is_active ? "Disable it first" : "Cannot be undone",
+      icon: FiTrash2,
+      disabled: apiKey.is_active || isPending,
+      run: onDelete,
+    },
+  ]
+  return (
+    <Dropdown>
+      <Tooltip.Root>
```
khaledosman: `Tooltip.Root` straight from `@heroui/react` with no `delay`. `--tooltip-delay` is not declared anywhere in `globals.css`, so HeroUI's `TooltipRoot` resolves `delay ?? parseCSSTime(undefined)` → `undefined` → react-aria's 1.5s warmup.

That is exactly what `design-system/overlays/Tooltip.tsx:18` exists to fix ("a lane of icon actions reads as having no labels at all"), and this tooltip is now the only visible label the icon trigger has. Pass `delay={300}` or route through the primitive.

web/src/features/keys/KeysPage.tsx
```diff
@@ -1420,24 +1401,115 @@ export function KeysPage() {
       {/* Suppress the table (and its own empty message) while the onboarding
           panel owns the empty state, so a fresh gateway shows one call to action,
           not a panel stacked over a redundant "no rows" table. */}
-      {showOnboarding ? null : (
+      {showOnboarding ? null : layout === "mobile" && !loading ? (
```
khaledosman: `loading` falls through to the desktop table, and `visibleColumns === columns` when layout is `"mobile"`, so a phone paints the full 7-column skeleton inside `TableScrollFrame` for the whole initial load and then swaps to the list.

That is the flash the `useState` seed at line 1018 was written to prevent. Gate on `layout === "mobile"` alone and render a list-shaped loading state inside that branch.

web/src/features/keys/KeysPage.test.tsx
```diff
@@ -36,6 +36,35 @@ function user(overrides: Partial<User> = {}): User {
   }
 }
 
+// The page picks its layout from the region it is given, which jsdom reports as
+// 0 wide. Feeding the observer a width is what selects wide, compact or the list.
+function stubRegionWidth(width: number) {
```
khaledosman, line 41: The stub forwards the observer callback outside React's act boundary, so `setLayout` lands unacted. The dashboard job logs five `The current testing environment is not configured to support act(...)` plus six `An update to KeysPage inside a test was not wrapped in act(...)`, all attributed by name to the two new tests.

Wrap the forwarded invocation in `act()`.

web/src/styles/globals.css
```diff
@@ -3097,42 +3097,106 @@ main {
...
+  width: 2.75rem;
+  height: 2.75rem;
+  flex-shrink: 0;
+  color: var(--color-text-muted);
+}
+.otari-key-actions-trigger:hover {
+  color: var(--color-text);
+  background: var(--color-surface-muted);
+}
+.otari-key-actions-trigger[data-focus-visible],
```
khaledosman, line 3213: Hand-written `outline: 2px solid var(--color-focus)` forks the values `@utility otari-focus-ring` (globals.css:1473) owns, including `--focus-ring-ink`.

Use `data-[focus-visible]:otari-focus-ring` on the trigger and keep only the `aria-expanded` half here, or say why the `-2px` inset needs its own copy.

web/src/styles/globals.css
```diff
@@ -3097,42 +3097,106 @@ main {
...
+}
+.otari-key-actions-menu .menu-item {
+  display: flex;
+  align-items: center;
+  gap: 0.5rem;
+  min-height: 2.75rem;
+  padding: 0.5rem 0.75rem;
+  color: var(--color-text-muted);
+}
+.otari-key-actions-menu .menu-item .label {
```
khaledosman, line 3234: `font-size: var(--text-sm)` here and `var(--text-xs)` on `[slot="description"]` below set type from CSS rather than from a role. The description element already carries `text-caption` at `KeyActionsMenu.tsx:171`; these rules are unlayered, so they silently outrank it and drop the hint from 13px to 12px.

Drop both `font-size` declarations (keep `margin-left`/`text-align`/`white-space`) and put `text-body` on the `<Label>` at `KeyActionsMenu.tsx:169`.

web/src/features/keys/KeyActionsMenu.tsx
```diff
@@ -0,0 +1,181 @@
...
+  apiKey: ApiKey
+  triggerRef?: RefObject<HTMLButtonElement | null>
+  onAction: () => void
+  owner?: string
+  /**
+   * Created, Last used, Expires and Models are lanes on the wide table; below it
+   * they leave the row, and the menu is the only place they can still be read
+   * whole.
+   */
+  showDetails?: boolean
```
khaledosman: `showDetails`, and `disabled:` on the `actions` entries (lines 52, 60, 68, 76). Booleans read as a yes/no question (`.github/skills/frontend-standards/naming-conventions.md`, Variables) — `hasDetails`, `isDisabled`. The three `show*` props in `charts.tsx` predate the rule.

web/src/features/keys/KeysPage.tsx
```diff
@@ -1037,40 +1005,127 @@ export function KeysPage() {
...
-  const [lastArmed, setLastArmed] = useState<string | null>(null)
-  const arm = useCallback((id: string) => {
-    setLastArmed(id)
-    setArmed(id)
-  }, [])
-  const { triggerRef, confirmRef } = useConfirmationFocus(armed !== null)
+  const [pendingRegenerate, setPendingRegenerate] = useState<ApiKey>()
   const [pendingDelete, setPendingDelete] = useState<ApiKey>()
+  const [lastAction, setLastAction] = useState<string>()
+  const { triggerRef: actionTriggerRef } = useConfirmationFocus(
```
khaledosman, line 1017: `confirmRef` is destructured away and never attached, so the `armed` branch of `useConfirmationFocus` focuses null. The hook's contract is the in-place trigger↔Confirm swap, which no longer exists here.

With `trigger.current?.focus()` already running in `onAction` before the dialog opens, react-aria's own restore has the right target. Either drop the hook (and `lastAction`, and the dual-ref plumbing in `KeyActionsMenu`) or leave a line saying what it still earns.

khaledosman (approved):
Reviewed at head `95fa899`. No blockers; findings inline.

Two things that have no line to sit on:

**PR title.** `fix(web):` — `cliff.toml:30` renders the scope verbatim into the changelog, and 0 of the last 40 commits on `main` use `web` while 22 use `dashboard` (this PR's own second commit does too). Retitle to `fix(dashboard):` before squash.

**Pattern divergence.** This makes keys the only one of 14 table pages not using `RowAction`/`RowActionRow`. Fine as a trial, but say whether it is meant to spread — if so `KeyActionsMenu` belongs in `design-system/actions/` with the `ApiKey` coupling lifted out.

---

web/src/styles/globals.css
jigjigjig: Addressed in 6bfa8d3b as a `hasTouchTarget` prop on `Checkbox`, with the CSS rule gone. Taking the `CopyButton` device literally does not work here, and the prop's docstring says so: a 16px box needs 14px each way to reach 44, and in hairline-divided rows those bleeds overlap the neighbors above and below, so a press near a seam lands on the wrong row. The prop claims the space instead. Added a `TouchTarget` story, which `foundation.test.ts` requires.

web/src/features/keys/KeyActionsMenu.tsx
jigjigjig: Confirmed and addressed in 6bfa8d3b. `--tooltip-delay` appears nowhere in `globals.css`, so this was the 1.5s warmup on the trigger's only visible label. `OPEN_DELAY_MS` is exported from the primitive as `TOOLTIP_OPEN_DELAY_MS` and passed here, rather than a second 300 written at the call site.

web/src/features/keys/KeysPage.tsx
jigjigjig: Addressed in 6bfa8d3b: the branch gates on `layout === "mobile"` alone and carries its own spinner, so the list owns the whole below-md state. New test asserts no `grid` exists while "Loading…" is up, against a held keys response.

web/src/features/keys/KeysPage.test.tsx
jigjigjig: Addressed in 6bfa8d3b: the forwarded invocation is wrapped in `act`. One correction on the attribution, since it changes what the number means: these predate my commit. The run at 8acc4ebd already logged 174 `An update to KeysPage` and 619 `not configured to support act`, so the source is the stub the PR introduced rather than the two tests added on top of it, which took it to 229. The local runner swallows them either way, so the next CI run is the real check.

web/src/styles/globals.css
jigjigjig: Addressed in 6bfa8d3b: the ink comes from `--focus-ring-ink` with the same fallback the utility uses. The inset stays and now says why: the trigger is flush with the row's right edge, where the utility's outward 2px is clipped by the table's overflow.

web/src/styles/globals.css
jigjigjig: Addressed in 6bfa8d3b: both `font-size` declarations dropped, `text-body` on the `<Label>`. Confirmed the arithmetic before removing them, `--text-sm` 14px and `--text-xs` 12px against `text-caption`'s 13px, so the hint was indeed losing a point.

web/src/features/keys/KeyActionsMenu.tsx
jigjigjig: Addressed in 6bfa8d3b: `hasDetails`, and all four `isDisabled`.

web/src/features/keys/KeysPage.tsx
jigjigjig: Half addressed, half pushed back on, in 6bfa8d3b. The observation is right: `confirmRef` is never attached and the docstring describes a swap that is gone. But the hook is still load-bearing, which I measured rather than argued: handing the menu no `triggerRef` fails both "returns focus after cancelling regeneration" and "opens the edit form in a dialog". The reason is that a menu closes on choice, so the focused item is unmounted before the dialog opens and react-aria's restore has no target left; `trigger.current?.focus()` in `onAction` runs before that unmount, not after it. So the hook, `lastAction` and the dual ref all stay, and there is now a comment saying exactly that.

jigjigjig:
CI here is red on a breakage inherited from `main`, not on anything in this PR.

`main` is red at 217b6fa2: #1220 added `features/auth/AuthHelp.tsx` using `Popover` without a `label`, and #1170 made `label` required on `Popover`. Each was green on its own; together they are not. Checks run against the merge with `main`, so `dashboard`, `serving`, `build` and `e2e` all fail on the one error, `AuthHelp.tsx(24,6) TS2741`.

Merging `origin/main` into this branch locally is conflict-free and leaves that error as the only one, with the keys, design-system and foundation suites green (4,115 tests). #1232 is open and green and fixes it; this PR rebases and re-runs once that lands.

Issue #1224, opened 2026-09-16
An abandoned worker can repopulate a cache that shutdown already cleared


peteski22:
### Problem

`_stop_refreshers` bounds its wait at `_REFRESHER_STOP_TIMEOUT_SECONDS`, then returns and logs any worker that did not stop. The lifespan clears that worker's cache straight after. A worker that absorbed its cancellation can finish an in-flight tick afterwards and refill the cache it was supposed to leave empty, so shutdown does not reliably leave those caches cleared.

The bound itself is not the problem and must stay. An unbounded wait hangs shutdown behind a background refresh, which is what the comment on `_REFRESHER_STOP_TIMEOUT_SECONDS` records and what `tests/unit/test_gateway_lifespan_shutdown.py` exists to prevent.

In production this is unobservable, because the loop is torn down immediately after. It becomes reachable wherever lifespans run repeatedly in one process, which the integration suite does.

### A stale claim to settle with it

The comment on `_REFRESHER_STOP_TIMEOUT_SECONDS` justifies abandoning a stuck task with "no refresher owns state that a late tick could corrupt". `_LIFESPAN_WORKERS` now pairs each worker with the cache it keeps warm, so that is no longer what the file expresses.

### Proposal

Narrow the claim to what still holds, that the loop is torn down immediately after so a late tick cannot be observed. Or make cache writes generation-aware, so a worker belonging to a finished lifespan cannot publish into a cleared cache.

### Acceptance

- [ ] Either the comment states only what remains true, or a late tick cannot publish into a cleared cache.
- [ ] If the code changes, a test covers a worker writing to its cache after absorbing its first cancellation.

Noticed while reviewing #1174.

Issue #1223, opened 2026-09-16
A failing cache reset aborts the rest of lifespan shutdown


peteski22:
### Problem

In `_create_lifespan`'s `finally` (`src/gateway/main.py`), the worker cache resets run before `log_writer.stop()`, `close_search_client()` and `dispose_db()`. A reset that raises skips all three, leaking the writer, the pooled search client and the engine.

This is pre-existing. The same statements sat in the same `finally` before the worker registry, with the same exposure. Every reset today is a dict or attribute clear with no plausible raise, so nothing triggers it now.

### Proposal

Run the resets after the resources whose closure actually matters, since a reset touches process memory only and needs nothing of the loop. Alternatively guard each reset and log the failure the way `_log_refresher_stop` does.

Reordering shutdown is a behavior change, which is why it belongs in its own issue rather than in a refactor that promises none.

### Acceptance

- [ ] A reset that raises does not prevent the log writer, the search client and the engine from closing.
- [ ] A test covers it.

Noticed while reviewing #1174.

Issue #1222, opened 2026-09-16
Lifespan startup can orphan background workers when one fails to start


peteski22:
### Problem

`_start_lifespan_workers` (`src/gateway/main.py`) builds its list of started workers in one comprehension, and the lifespan assigns the result before entering the `try`/`finally` that cancels them. If an entry's `start` raises after earlier entries have already become tasks, the function never returns, `workers` keeps its initial empty value, and the `finally` has no handle on the tasks already created. They stay scheduled on a loop that startup is about to abandon.

This is pre-existing exposure rather than a regression. Before the worker registry each task was assigned to its own local, but creation still sat outside the same `try`, so an exception there skipped the cleanup block just the same.

### Why the trigger rate is near zero today

Every `start` either calls an `async def`, which builds a coroutine without running its body, or is `_start_reservation_sweeper`, which reads three config attributes. `asyncio.create_task` does not fail while a loop is running. Nothing on the current list has a plausible synchronous raise. The registry is what makes adding a `start` that does real work easy, which is why this is worth closing before it is reachable.

### Proposal

Either start the workers inside the lifespan's existing `try`, or have `_start_lifespan_workers` cancel what it has created before re-raising.

### Acceptance

- [ ] A `start` that raises leaves no scheduled task behind.
- [ ] A test covers it.

Noticed while reviewing #1174.

PR #1226, opened 2026-09-16
refactor(gateway): drive lifespan workers from one registry

### LINKED ISSUE #1174: Lifespan worker registry
### Problem

`src/gateway/main.py` keeps nine manually declared task variables for its periodic workers. Each is created individually in the lifespan and listed again individually at shutdown. Adding a worker means editing the lifespan in three places.

### Proposal

One list of `(name, start)` pairs, built from the modules that own the workers. The lifespan iterates it to start them and hands the same list to the existing bounded shutdown. Same cancellation bound, same logging, same order. No behavior change.

### Acceptance

- [ ] No `*_refresher: asyncio.Task | None` variables remain in `main.py`.
- [ ] The shutdown tests in `test_gateway_lifespan_shutdown.py` pass unchanged.
- [ ] A worker is added by adding one pair to the list.

### Measure

`main.py` no longer appears among the most frequently changed files over 90 days (10 authors today).

### Depends on

Nothing.

Part of #1171.

peteski22:
## Description

The gateway's startup and shutdown hook kept a separate variable for each of its eleven periodic background workers. Starting a worker, listing it for shutdown, and clearing the cache it keeps warm were three edits in three places, and the function had grown past 200 lines of mostly repetition.

This replaces those variables with one list. Each entry names a worker, says how to start it, and says which cache to clear when it stops. Startup walks the list. Shutdown hands the same list to the bounded cancel that was already there, then clears the caches. Adding a worker is now one entry.

Nothing about behavior changes. The same workers start in the same order, the shutdown grace period and its log lines are untouched, the reservation sweep is still the only worker a setting can turn off, and a hybrid gateway still starts none. Resets stay gated on the worker having started, which is what the per-variable checks did before.

Feature workers from #1188 are folded in: both the registry's workers and the enabled features' workers go into the one bounded stop, and each keeps the label #1188 gave it.

## How to test it locally

```
make lint && make typecheck && make test-unit
```

The existing shutdown tests in `tests/unit/test_gateway_lifespan_shutdown.py` pass unchanged, including the end to end one that boots the lifespan and shuts it down behind a worker that refuses to stop.

Two tests are added to that file. One pins the property this change introduces: every entry must look its worker function up when the app starts rather than hold it. The suite replaces those functions by name to stay off the network, and an entry holding the original would keep calling it with nothing to show for the substitution. The other pins that turning the reservation sweep off drops that worker and no other.

To see it by hand, boot the gateway and stop it. Startup and shutdown are unchanged, and `Application shutdown complete.` still arrives promptly.

## PR Type

- [ ] New Feature
- [ ] Bug Fix
- [x] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Closes #1174. Part of #1171.

Three pre-existing weaknesses found while reviewing this change are filed rather than fixed here, since each is a behavior change this refactor promises not to make: #1222, #1223, #1224.

A follow-up that unifies how both sources hand a started task to the supervisor is #1227.

## Checklist

No route, schema or config field changed, so there is no generated artifact to regenerate and no documentation this affects.

## Measurements

Measured against `main` at the merge base:

| | before | after |
|---|---|---|
| `_create_lifespan` | 214 lines | 114 lines |
| `src/gateway/main.py` | 876 lines | 844 lines |
| `asyncio.create_task` call sites | one per worker | one |

Over the last 90 days `src/gateway/main.py` took 48 commits from 10 authors, the fifth most changed file under `src/gateway`. The repetition that made it a merge target is what this removes.

## Verification performed

`make lint`, `make typecheck` (636 files) and the full unit suite: 3634 passed. Twelve failures are environment noise from a local `.env` that sets `OTARI_MASTER_KEY`, and they fail identically on a clean `main`. The OSS edition smoke gate passes, booting the packaged CLI and shutting it down through the new registry.

Integration tests were not run locally; CI covers them.

## AI Usage

- [ ] No AI was used.

The issue, the design of the registry and the review of the result are mine. The agent wrote the mechanical transposition and the two added tests to that design. I checked the new guard test by hand: changing an entry to hold its function instead of looking it up makes it fail in under two seconds.

## Summary

The gateway now manages periodic workers through one shared registry instead of separate startup and shutdown code for each worker.

Startup and shutdown continue to preserve worker order, timing, logging, configuration, hybrid gateway behavior, and conditional cache resets. The reservation sweep remains excluded when disabled.

Tests now verify runtime worker lookup and conditional removal of the reservation sweep. This reduces repeated lifespan code and makes future worker additions simpler.

## Technical notes

- Added `_LifespanWorker` and `_LIFESPAN_WORKERS`.
- Reused the registry for bounded shutdown and worker-specific cache resets.
- Removed individual worker task variables from `src/gateway/main.py`.
- No routes, schemas, or configuration fields changed.


tbille (approved):
Verified independently on the PR head: the registry is a faithful transposition of the eleven per-variable workers (same start order, same shutdown labels, same reset pairings and ordering, resets still gated on the worker having started), hybrid mode still starts none, and the sweep-interval gate still drops exactly one worker. The late-binding property the root conftest depends on (substituting run_* by name at gateway.main) holds through the registry's lambdas, and the new guard test fails fast if an entry ever captures the function object instead. The documented early binding of reset functions affects nothing today: no test or overlay substitutes a reset on gateway.main. No changes landed on main for either file since the merge base, so the older base merges clean. Ran the touched test file locally: 8 passed.

No findings.

Reviewed by Claude Code (Fable 5), run by the reviewer.

PR #1229, opened 2026-09-16
feat(policy-checks): add command_match, a second Agent Gates gate type

agpituk:
## Description

Adds command_match, a second Agent Gates gate type, alongside the existing changed_path. It blocks a forbidden command before it runs, the same PreToolUse-blocking mechanism changed_path already uses for edit tools, now covering Bash: "don't force-push", "use pnpm, not npm", and similar rules a repo can enforce on its own working tree.

A forbidden entry is a shell phrase ("`git push --force`", "`npm`"). Matching is token-based, not a substring check, on purpose: a phrase's own tokens must appear, in order and unbroken, within one `&&/;/|/||-`separated segment of the submitted command. A plain substring check gets two real cases wrong: "`npm`" would match inside "`pnpm`", and "--force" would match inside the deliberately safer "--force-with-lease". Both are pinned as regression tests.

otari hook's PreToolUse branch now also handles Bash tool calls, submitting the command as evidence (a tool call is either an edit or a shell command, never both, so exactly one evidence list is populated per call). This repo's own `.otari-gates.yml` now dogfoods a real command_match gate (`use-pnpm-not-npm`, enforcing web/AGENTS.md's actual pnpm-not-npm rule).

Two things this gate does not do, on purpose, documented in docs/agent-gates.md:
- It sees only the literal command text of one tool call, not what a script or program that command invokes does internally.
- It only splits a command into segments on `&&/;/|/|| `when they are whitespace-separated (cmd1&&cmd2 is not split).

## How to test it locally

1. Check out this branch and install it:
`git fetch origin agent-gates-command-match`
`git checkout agent-gates-command-match`
`uv sync`
This installs otari from this checkout in editable mode, so uv run otari hook --help already shows the command below. (Or `source .venv/bin/activate` first if you'd rather not prefix every command with uv run.)
2. Start the gateway: `OTARI_MASTER_KEY=otari-local-dev-master-key uv run otari serve`.
3. If you already have a `.claude/settings.local.json` from testing the Hook Server before, add Bash to its PreToolUse matcher. Otherwise create one:
```
{
  "hooks": {
    "PreToolUse": [
      {
        "matcher": "Edit|Write|NotebookEdit|Bash",
        "hooks": [
          {
            "type": "command",
            "command": "/abs/path/to/this/checkout/.venv/bin/otari hook --harness claude-code --api-key otari-local-dev-master-key"
          }
        ]
      }
    ]
  }
}
```
3. The path is absolute on purpose: the hook subprocess doesn't inherit an activated shell's PATH.
4. In web/, ask Claude Code to run `npm install`. The tool call is refused before it runs, citing this repo's own dogfooded use-pnpm-not-npm gate. Ask it to run pnpm install instead and it goes through normally.
5. You can also add your own rule directly in `.otari-gates.yml`, e.g.
 ```
 - id: no-force-push
    type: command_match
    enforcement: required
    forbidden: ["git push --force", "git push -f"]
    message: Force-pushing is not allowed; use --force-with-lease if you must.
```
5. and ask Claude Code to run git push --force (refused) or git push --force-with-lease (allowed; the token-based match correctly does not treat it as --force).
## PR Type

- [X] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues
None

## Checklist

- [X] I understand the code I am submitting.
- [X] I have added or updated tests that cover my change (`tests/unit`, `tests/integration`).
- [X] I ran the Definition of Done checks locally (`make lint`, `make typecheck`, `make test`).
- [X] Documentation was updated where necessary.
- [X] If the API contract changed, I regenerated the OpenAPI spec (`uv run python scripts/generate_openapi.py`).

## AI Usage

- [ ] No AI was used.
- [X] This is fully AI-generated.

**AI Model/Tool used:**
Claude Sonnet 5/Claude Code.

**NOTE:**

## Summary

Adds `command_match` gates that block forbidden shell phrases before execution. Bash command evidence now flows through the hook API and CLI.

The change supports mixed gate types, validates and deduplicates forbidden phrases, applies processing limits, and distinguishes unknown evidence from not-applicable evidence. It also updates documentation, schemas, repository policy, CLI error handling, and tests for parsing, matching, budgets, and integration behavior.

### CODE REVIEW

File: docs/agent-gates.md [hunk 1]
```diff
@@ -156,19 +185,22 @@ nothing here needs copying out of this document.
...
+what a `Bash` call's `tool_input` names; a single `PreToolUse` call is either
+an edit or a shell command, never both, so it submits whichever evidence that
+one call can produce, never both kinds. On a `Stop` event, `changed_path`
 instead submits `git status --porcelain`'s output, which does cover
-shell-written changes (anything a `Bash` call touched), at the cost of only
-catching them after the fact rather than preventing them.
+shell-written file changes (anything a `Bash` call touched), at the cost of
+only catching them after the fact rather than preventing them; `command_match`
+does not participate in the `Stop` event at all, since there is no evidence
+of which commands ran that a `Stop` payload carries or Git can reconstruct.
```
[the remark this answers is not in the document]
agpituk: Addressed in 85914776: an explicit empty `commands` list (what a Stop event or a PreToolUse edit call submits) now resolves `not_applicable` instead of `pass` in evaluate_command_match, so a required command_match gate can no longer read absent evidence as a clean check. docs/agent-gates.md updated to describe this.

src/gateway/agent_runtime/domain/evaluators.py
```diff
@@ -105,6 +115,212 @@ def _matches_any(path: str, patterns: tuple[str, ...]) -> str | None:
...
+            result.append(char)
+            index += 1
+            continue
+        if quote == "'":
+            # Nothing is special inside single quotes, not even backslash.
+            result.append(char)
+            if char == "'":
+                quote = None
+            index += 1
+            continue
```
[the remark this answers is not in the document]
agpituk: Addressed in 85914776: _strip_shell_comment now tracks Bash ANSI-C quoting ($'...') as its own state, honoring backslash escapes inside it (so \' is literal content, not the closing quote) and rewriting it into an equivalent plain-quoted form shlex tokenizes correctly. Added the suggested echo $'a\' # b' regression test plus a case in test_strip_shell_comment_cases.

src/gateway/api/routes/hooks.py
```diff
@@ -126,6 +171,11 @@ class PolicyCheckRequest(BaseModel):
         max_length=_MAX_CHANGED_PATHS,
         description="Repo-relative paths the caller observed changed (e.g. `git status --porcelain`).",
     )
+    commands: list[str] = Field(
+        default_factory=list,
```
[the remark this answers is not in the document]
agpituk: Addressed in 85914776: commands now defaults to None (not an empty list), and command_evidence returns None when it was never submitted, so an omitted commands field resolves unknown/blocking rather than a silent pass. Added integration coverage for both the omitted and explicit-empty cases.

tests/integration/test_hooks_route.py
```diff
@@ -245,3 +245,266 @@ def test_still_requires_a_token(self, hybrid_client: TestClient) -> None:
...
+    client: TestClient, master_key_header: dict[str, str]
+) -> None:
+    """Submitting `commands` evidence against a policy with no command_match
+
+    gate must not pay any tokenizing cost at all: the result is moot
+    regardless, so this must resolve quickly and successfully rather than
+    being rejected by a budget meant for command_match gates that do not
+    exist here.
+    """
+    commands = [" " * 4000 for _ in range(600)]
```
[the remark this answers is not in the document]
agpituk: Addressed in 85914776: made the 600 commands distinct (trailing index) so Request.command_evidence's dedup can't collapse the workload to one command.

src/gateway/cli.py
```diff
@@ -335,21 +340,36 @@ def hook(harness: str, config: str | None, url: str | None, api_key: str | None)
     if not gates_file.is_file():
         return
 
-    changed_paths: list[str]
+    changed_paths: list[str] = []
```
daavoo, line 350: **Blocking.** A `Bash` PreToolUse call leaves `changed_paths` as `[]` and sends it. `evaluate_changed_path` has no empty-evidence branch, so it returns `PASS` ("No forbidden paths changed") for a call it never had any path to check. Every `changed_path` gate now certifies a clean pass on every Bash tool call.

This is exactly the confusion `evaluate_command_match` was written to avoid: it returns `not_applicable` on `[]` with a comment explaining why a pass would be dishonest. Give `evaluate_changed_path` the same branch, and make `changed_paths` tri-state (`None`) like `commands` so an absent collector blocks a required gate instead of passing it.

Verified live: `{"tool_name":"Bash","tool_input":{"command":"pnpm --dir web install"}}` exits 0 with the `no-changelog-edits` gate reported as `pass`.

src/gateway/api/routes/hooks.py
```diff
@@ -111,6 +116,46 @@ async def verify_hook_caller(
 # path_count.
 _MAX_COMPARISONS = 1_000_000
 
+_MAX_COMMANDS = 10_000
+_MAX_COMMAND_LENGTH = 4096
```
daavoo, line 120: **Blocking.** A command over this limit gets a 422, `raise_for_status()` in `cli.py` raises `HTTPStatusError` (a subclass of `httpx.HTTPError`), and the handler prints `could not reach <url> ..., not blocking` and exits 0.

Two problems. The request arrived and was rejected, so "could not reach" sends whoever debugs this to the wrong place. And the fail-open is total: the `changed_path` gates in the same policy do not run either, because the whole request was refused.

4,096 characters is not a hypothetical ceiling for a Claude Code `Bash` call; a heredoc clears it routinely. Suggest dropping or truncating the oversize command in the CLI (with a stderr note) so one long command cannot take the policy down, and splitting the 4xx branch out of the "could not reach" message. No test covers this path.

.otari-gates.yml
```diff
@@ -11,3 +11,11 @@ gates:
     message: >-
       CHANGELOG.md is generated by git-cliff at release time (see AGENTS.md,
       "Generated Artifacts"). Do not hand-edit it in a PR.
+
+  - id: use-pnpm-not-npm
+    type: command_match
+    enforcement: required
+    forbidden: ["npm"]
```
daavoo: Non-blocking, but it will bite within a day of dogfooding. A one-word phrase matches `npm` as **any** token in any position, so ordinary commands that merely mention it are refused:

```
grep -rn npm web/AGENTS.md   -> blocked
```

Verified through the hook CLI against this branch. The gate has no notion of argv[0] versus an argument. Either narrow this entry (`npm install`, `npm run`, `npm ci`) or add a `position: head` option to `command_match` so a bare tool name is usable.

src/gateway/agent_runtime/domain/evaluators.py
```diff
@@ -105,6 +115,258 @@ def _matches_any(path: str, patterns: tuple[str, ...]) -> str | None:
...
+    non-empty by construction, `domain.policy` rejects one that isn't), so
+    keeping it around costs a phrase-count's worth of guaranteed-`False`
+    comparisons for nothing. A caller-controlled string built from many
+    separators and no real content, ";" * 500 against 500 forbidden
+    phrases, measured ~25,000,000 such comparisons and ~1.1s before this.
+    """
+    try:
+        tokens = shlex.split(_strip_shell_comment(command), posix=True)
+    except ValueError:
+        return [[command]]
```
daavoo: Non-blocking, but the two instances of this got opposite verdicts. The fallback is safe against a false positive, and it is a silent fail-open against a single-token phrase:

```
npm install "unterminated   -> pass   (gate forbids "npm")
```

`test_apostrophe_in_a_trailing_comment_no_longer_evades_the_gate` treats precisely this shape as a bug worth fixing; `test_unbalanced_quotes_do_not_crash_and_do_not_match` pins it as correct, and its docstring's "never equal to a real **multi-token** forbidden phrase" is where the qualifier hides the gap. Since the harder instance is already fixed, suggest returning `unknown` for a command `shlex` cannot tokenize, rather than a token that can only pass.

docs/agent-gates.md
```diff
@@ -72,10 +78,32 @@ provenance, never confused with something Otari observed directly). Renames
...
+`git push --force-with-lease`; a plain substring check would get both wrong.
+
+Two things this gate does not do, on purpose, for now:
+
+- It sees only the literal command text of one tool call. It does not, and
+  cannot, see what a script or program that command invokes does internally:
+  `./deploy.sh` is one opaque token to this gate even if the script itself
+  runs `git push --force`. This is a footgun-catcher for a cooperative agent,
+  not a sandbox against one deliberately working around it.
+- It only splits a command into segments on `&&`/`;`/`|`/`||` when they are
```
daavoo, line 106: Non-blocking, wording. "not split into two segments" reads as a risk of over-joining (a false positive). The actual consequence is a **missed detection**, which is the opposite of what a reader infers:

```
cd web;npm install   -> pass   (token is "web;npm", equal to no phrase)
```

`cd dir;cmd` is ordinary shell. Worth saying plainly here that the gate fails open on it. Cheap fix if you want it instead: pad standalone operators with spaces before `shlex.split`, reusing the quote-aware scanner already in `_strip_shell_comment`.

src/gateway/agent_runtime/domain/evaluators.py
```diff
@@ -105,6 +115,258 @@ def _matches_any(path: str, patterns: tuple[str, ...]) -> str | None:
...
+        # this gate never had anything to check. not_applicable is the
+        # non-blocking outcome that says so honestly.
+        return GateResult(
+            gate_id=gate.id,
+            enforcement=gate.enforcement,
+            outcome=Outcome.NOT_APPLICABLE,
+            message="No commands were submitted to check.",
+        )
+
+    forbidden_phrases = [tokenize_phrase(phrase) for phrase in gate.forbidden]
```
daavoo: Non-blocking. Every forbidden phrase is tokenized once per gate here, and `hooks.py` already tokenized all of them for its cost estimate and discarded the result. Bounded by policy size so it is not a real cost, but `segment_cache` exists to avoid exactly this shape for commands; phrases deserve the same cache or none of the ceremony.

src/gateway/cli.py
```diff
@@ -380,7 +400,11 @@ def hook(harness: str, config: str | None, url: str | None, api_key: str | None)
     try:
         response = httpx.post(
             f"{resolved_url.rstrip('/')}{API_ROOT}/hooks/check",
-            json={"policy_yaml": gates_file.read_text(encoding="utf-8"), "changed_paths": changed_paths},
+            json={
+                "policy_yaml": gates_file.read_text(encoding="utf-8"),
+                "changed_paths": changed_paths,
+                "commands": commands,
```
daavoo, line 427: Non-blocking, observation. This always sends a list (`[]` on `Stop` and on edit calls), never `null`, so the `unknown` state the `commands` field is carefully designed for has no producer outside the integration tests. Fine as is; noting it because it is where the wiring goes if `changed_paths` gets the same tri-state treatment.

daavoo (commented):
Review of the `command_match` gate. Findings inline, each marked blocking or non-blocking. Verified against a live `otari serve` on this branch with the real hook CLI, not read-only.

Two blocking, both fail-open: a `Bash` PreToolUse call now certifies `changed_path` gates as passing on empty evidence, and a command over 4,096 characters takes down the whole policy check with a message that blames the network.

Reviewed by Claude Opus 5 (1M context) via Claude Code.

daavoo:
**Not a blocker, on the shape rather than this PR.**

`changed_path` and `command_match` are the same gate with a different matcher: same four fields, same evaluator skeleton (`None` is unknown, empty is non-blocking, a match fails with sorted detail, else pass). The copy has already drifted at two types: empty evidence is `not_applicable` for one and `pass` for the other, `changed_paths` defaults to `[]` while `commands` defaults to `None`, and the route carries a hand-written budget block per type. Two of the inline findings above are that drift.

Three changes, each worth doing on its own, roughly in this order:

1. One shared wrapper for the `None` / `[]` / non-empty semantics, so a gate type cannot get them wrong. Small, and it fixes both blocking findings for every type at once.
2. Evidence as one extensible bag (`evidence: {changed_paths, commands, ...}`) that tolerates unknown kinds, resolving the gates that need a kind as `unknown` instead of a 422 that fails the whole check open. That is also what makes `otari hook` a pure harness adapter (event in, evidence out, no gate knowledge), which is the seam a second harness plugs into.
3. A gate registry (parser, evidence kind, cost estimator, evaluator per type) so the route stops knowing types. Best done as the first commit of `check_passed`, where a runner with deadlines and output caps does not fit the `forbidden` shape at all.

Orthogonal to all three: shipping our own opinions (`npm`, force-push, changelog) as a bundled preset a policy can `extends`, rather than as YAML copied from the docs. Gate types are the vocabulary; presets are the content.

Reviewed by Claude Fable 5.1 via Claude Code.

daavoo:
All seven findings above are addressed in 167e0401. Two of them were my own overreaches, and dogfooding this gate on this repo caught both within minutes of applying them, so correcting them here rather than leaving the inline comments standing:

**The unbalanced-quote finding.** I suggested returning `unknown` for a command `shlex` cannot tokenize. That blocked the very first `Bash` call I made afterward: a heredoc carrying Python or SQL is routinely unparseable to `shlex`, so `unknown` blocks every ordinary tool call rather than the forbidden ones. The fix shipped instead is a whitespace-split fallback, which still catches `npm install "unterminated` against a gate forbidding the tool name, and still does not match `git push "--force` against `git push --force`. Degraded on purpose, and the trade is the right way round: it costs a false positive on a command that was already malformed and buys back detection on the shape an evasion would take.

**The `changed_paths` finding.** I suggested the CLI send `null` for the evidence kind an event cannot carry. That resolves as `unknown`, `unknown` blocks a required gate, and this repo's policy has two, so it locked every `Bash`, `Edit` and `Write` call in the session. `null` should mean "this caller never collects this kind at all"; `otari hook` collects both kinds, just never both from one event, so `[]` is the true statement and resolves `not_applicable`. The real bug was only ever on the evaluator side, `evaluate_changed_path` returning `pass` where `evaluate_command_match` returns `not_applicable`, and that fix alone is sufficient. The tri-state still lands on the request model, where it matters for a third-party caller that omits the field.

Verified end to end against a live gateway: the tool name blocked, an unbalanced quote blocked, `grep -rn npm web/AGENTS.md` allowed, `pnpm --dir web install` allowed.

Reviewed by Claude Opus 5 (1M context) via Claude Code.

PR #1232, opened 2026-09-16
fix(dashboard): restore auth popover and catalog CI

jigjigjig:
## Description

Restore dashboard builds and the published component catalog after the auth-page and popover changes met on main. The Help control now opens a properly named dialog, and public auth examples render with the theme context they need. The catalog toolbar still switches between light and dark themes.

The build failure is the missing popover label in AuthHelp; its old span trigger also stopped opening the panel. The separate catalog failure is `useTheme must be used within a ThemeProvider` in all four PublicAuthLayout stories, in both themes.

## How to test it locally

- Run `pnpm --dir web run typecheck`, `pnpm --dir web run lint`, and `pnpm --dir web test --maxWorkers=4`.
- Run `pnpm --dir web run build` and `STORYBOOK_BASE_PATH=/otari/ pnpm --dir web run storybook:build`.
- Serve the catalog under `/otari/` and run `SMOKE_ORIGIN=http://localhost:6006/otari pnpm --dir web exec node .storybook/smoke.mjs`. All stories should render in both themes.
- Open Help on the sign-in page and verify its links appear. AuthHelp tests cover the named dialog and recovery options; the new catalog theme test covers toolbar changes, stored preferences, and the in-page theme control.
- Repository lint: `make lint`.

## PR Type

- [x] Bug Fix
- [x] Infrastructure / CI

## Relevant issues

Follow-up to #1220 and #1170.

## Checklist

Dashboard-only change; Python typecheck and tests are not applicable. No API contract changes.

## AI Usage

**Any additional AI details you'd like to share:** Diagnosed from GitHub Actions logs and validated in an isolated worktree.

## Summary

- Restored Storybook theme context for dashboard and public authentication stories.
- Added tests for theme switching and stored theme preferences.
- Replaced the `AuthHelp` span trigger with a functional, labeled `Button`.
- Updated auth help tests to verify that the dialog opens.
- Clarified Storybook theme documentation.

These changes restore affected builds and preserve light and dark theme switching.


web/src/features/auth/AuthHelp.tsx
```diff
@@ -22,11 +23,8 @@ export function AuthHelp({
 }) {
   return (
     <Popover
-      trigger={
-        <span className="inline-flex min-h-11 items-center px-3 text-sm font-medium text-link">
-          Help
-        </span>
-      }
+      label="Help"
+      trigger={<Button variant="ghost">Help</Button>}
```
jigjigjig, line 32: The Button is needed (a span is not a react-aria pressable, which is what stopped the panel opening), but a bare ghost changes the page. Measured in the catalog, light theme: the trigger goes from 55.55x44, teal rgb(0,116,126), no border, to 63.55x36, near-black rgb(8,9,10), inside a 1px rgb(90,95,107) box. It sits directly beside `PublicAuthLink` "Create an account", which is 44px and teal, so the row now reads as two control families and the target drops under the 44px that `PublicAuthLink`'s docstring says the phone viewport asks for.

Keep the Button and give it back the link look: link color, transparent border, `min-h-11`. `ListDetail.tsx:147` already passes `min-h-11` to a Button for the same reason.

web/.storybook/theme.test.tsx
```diff
@@ -0,0 +1,51 @@
...
+      {preference}: {resolved}
+    </button>
+  )
+}
+
+function Catalog({ theme }: { theme: string }) {
+  return withTheme(
+    ThemeConsumer as Parameters<Decorator>[0],
+    { globals: { theme } } as Parameters<Decorator>[1],
+  )
```
jigjigjig: Two `as` casts to fake a decorator's arguments. `withTheme` only reads `context.globals.theme`, so a typed local helper says that plainly instead of asserting a story function and a full `StoryContext` that are not what is passed.

Also: `vite.config.ts:171` still says "`.storybook/` is in the list for one file". This PR makes it two.

web/.storybook/theme.tsx
```diff
@@ -1,29 +1,9 @@
...
- * which duplicates it because it has to run before React exists. Wrapping the
- * real `ThemeProvider` instead would not work: it reads the operator's stored
- * preference and owns its own state, so a toolbar could not drive it. Keep these
- * three writes in step with that effect.
- *
- * The preference names come from `THEME_PREFERENCES` rather than being restated,
- * minus "system": a catalog exists to show both themes deliberately, and
- * "whatever this laptop is set to" is not a case worth a toolbar entry.
- */
+/** The catalog toolbar drives the same theme context as the dashboard. */
```
jigjigjig: Dropping the three-writes explanation is right, since the decorator no longer duplicates them. But the filter below still loses its reason: "system" is left out because a catalog exists to show both themes deliberately. Keep that clause on the line that does the filtering.

jigjigjig (commented):
Self-review. The catalog fix is right and verified: both PublicAuthLayout stories render in light and dark with no page error, and the toolbar drives data-theme, .dark and color-scheme. One visual regression on the sign-in page, two cleanups.

web/src/features/auth/AuthHelp.tsx
khaledosman, line 32: A ghost button is not edgeless: `globals.css:1735` gives `.button--ghost` `border-width: 1px` in opaque `--color-text-subtle` (6.01:1 in light, 6.13:1 in dark, per the comment above that rule). So Help goes from #1220's `text-link` span to a strongly bordered box, sitting beside a bare `PublicAuthLink` in the same `justify-between` row (`Login.tsx:860`, `SignupPage.tsx:148`).

`design/actions.md`, "Places: where a ghost drops its edge", prescribes the fix: name the container. `.otari-actions` "groups inline actions ... drops ghost edges without imposing the field density of a filter toolbar", and `PlaygroundComposer.tsx:92` is the existing precedent for exactly this shape. Add `otari-actions` to those two footer rows, or say the edge is intended here.

(Touch target is fine: the 36px `md` default is raised by the `[data-slot="button"]` floor at `globals.css:2169` below 767px.)

web/.storybook/theme.tsx
khaledosman: This line describes `withTheme`, but it sits flush above `const THEMES`, which it says nothing about. That is the stranded-docblock shape `AGENTS.md` calls out under Repository Conventions. Move it onto `withTheme` below, or separate it with a blank line so it reads as a module header.

web/.storybook/theme.tsx
```diff
@@ -42,18 +22,24 @@ export const themeGlobalType = {
...
-  // a node React does not own, and the React Compiler is free to re-run a render.
+function StoryTheme({
+  theme,
+  children,
+}: {
+  theme: "light" | "dark"
+  children: ReactNode
+}) {
+  const { setPreference } = useTheme()
   useEffect(() => {
```
khaledosman, line 47: The effect fires only when `theme` changes, so the toolbar can go inert. A story that owns a theme control (`LoginPageShell`'s appearance button) can set the preference to dark while the global still reads `light`; re-picking "Light" in the toolbar then writes the same global value, the prop does not change, and the theme is never restored.

The new test encodes the precedence (in-page control wins) but not this consequence. Nit-level for a dev-only catalog, so a line in the file is enough if it is the accepted trade-off.

khaledosman (approved):
Both fixes are the right ones: `Popover` takes a required `label` (added by #1170) and its docblock states the trigger must be a react-aria pressable, which #1220's `<span>` was not. Three findings inline, none blocking.

No route, schema or generated-artifact impact; `schema.ts` and `routeTree.gen.ts` are correctly untouched.

---

web/src/features/auth/AuthHelp.tsx
jigjigjig: Fixed in 8527701d. Kept the Button, added an `.otari-auth-actions` place class in globals.css and put it on both footer rows, the way `.otari-setup-actions` already quiets the first-run sheet's Skip. Measured in the catalog after: 44px tall in both themes, rgb(0,116,126) light / rgb(0,165,178) dark, transparent edge, no fill under the pointer, hover to link-hover. The remaining difference from what shipped is 2px of width from the transparent border the ghost box model keeps.

web/.storybook/theme.test.tsx
jigjigjig: Fixed in 8527701d. One cast that narrows `withTheme` to the contract it actually has, instead of two that assert the arguments away. `vite.config.ts` comment rewritten in the same commit.

web/.storybook/theme.tsx
jigjigjig: Fixed in 8527701d. The "system" clause is back on the line that does the filtering.

web/.storybook/theme.tsx
```diff
@@ -42,18 +22,24 @@ export const themeGlobalType = {
...
+  return children
 }
+
+export const withTheme: Decorator = (Story, context) => (
+  <ThemeProvider>
+    <StoryTheme theme={context.globals.theme === "dark" ? "dark" : "light"}>
+      <Story />
+    </StoryTheme>
+  </ThemeProvider>
+)
```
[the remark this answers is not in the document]
jigjigjig: Real, and left as is. The window is one frame in a dev catalog, before the effect commits: rendered in both themes with the toolbar against a conflicting stored value, data-theme, .dark and color-scheme all land correct and no page error is raised.

The proposed fix puts a Storybook-only `initialPreference` on the product `ThemeProvider`, which is the catalog reaching into shipped code for a flash no user sees. The decorator-local alternative, holding the story back until `preference` matches the toolbar, breaks the in-page theme control that `theme.test.tsx` now covers: an operator flipping it inside a story would blank the story.

web/src/features/auth/AuthHelp.tsx
jigjigjig: Agreed on the diagnosis, and it was already fixed in 8527701d, before this landed. Went with a new place rather than `.otari-actions`, because the edge is not the whole gap: measured beside the link, a bare ghost is also near-black rgb(8,9,10) against the link's rgb(0,116,126), and 36px against its 44. `.otari-actions` would have to gain link ink and a 44px floor to close that, which would restyle every inline action row that wants its button to read as a button.

So `.otari-auth-actions` carries all three, and 678302c8 registers it in `design/actions.md` next to the others with that reasoning. Measured after: 44px in both themes, link ink, transparent edge, no fill under the pointer.

Thank you for the touch-target note, that is right and the comment now says the `min-height` is desktop alignment rather than a floor.

web/.storybook/theme.tsx
jigjigjig: Fixed in 678302c8. The module line is on `withTheme` now, and what sits above `THEMES` is the reason "system" is filtered out.

web/.storybook/theme.tsx
jigjigjig: Real, and taken as the trade-off with a line in the file (678302c8). The alternative is an effect with no dependency list, which restores the toolbar but overwrites the in-page control on its next render, and that precedence is what the new test pins. Storybook raises no update when the held value is re-picked, so there is nothing for the decorator to hear.

PR #1234, opened 2026-09-16
fix(gateway): open the provider catalog to organization owners/admins

tbille:
## Description

Organization owners/admins could not add an organization-scoped provider key at all: the "Add provider key" dropdown on the Providers page came back empty, because its picker reads `GET /api/v1/providers/catalog` (and the per-provider detail route), and that endpoint sat on `providers.py`'s deployment-operator-only router. Creating an org-scoped key is already authorized separately for organization owners/admins (`require_active_organization_management_access` in `org_provider_key_service.py`), so a non-superuser owner/admin got a 403 on the catalog read and had no provider to select.

The fix splits those two catalog reads onto their own router gated by `verify_catalog_reader` (any authenticated session or key, not only a deployment operator), the same split `models.py` and `pricing.py` already use for their own catalog reads, and mounts it alongside the existing operator-only `providers.router`. The rest of `providers.py` (the deployment-wide `/providers` and `/provider-credentials` surface) is untouched.

## How to test it locally

1. Sign in as an organization owner or admin who is not a deployment operator/superuser.
2. Go to the organization's Providers page and click "Add provider key".
3. The provider dropdown should now list the full catalog (openai, anthropic, ollama, etc.) instead of showing nothing.

Automated coverage: `tests/integration/test_provider_credentials_api.py::test_catalog_reads_admit_an_organization_owner_who_is_not_a_deployment_operator` builds a dashboard session for a non-superuser organization owner and asserts `GET /providers/catalog` and `/providers/catalog/{id}` now return 200 for that identity, while `GET /provider-credentials` (the deployment-wide surface) still correctly returns 403.

Also ran:
- `uv run python scripts/check_architecture.py` — no violations
- `make openapi-check` — no drift (route paths/schemas/tags unchanged)
- `uv run --frozen pytest tests/integration/test_provider_credentials_api.py tests/integration/test_org_provider_keys.py tests/unit/test_org_provider_key_service.py` — all pass

## PR Type

- [x] Bug Fix

## Checklist

- [x] I ran the Definition of Done checks locally (`make lint`, `make typecheck`, `make test` — targeted suites; see above).
- [x] Documentation was updated where necessary. (none needed)
- [x] If the API contract changed, I regenerated the OpenAPI spec (`uv run python scripts/generate_openapi.py`). (no contract change; verified via `make openapi-check`)

## AI Usage

**Any additional AI details you'd like to share:** Investigated and fixed end-to-end by Claude Code from a bug report; a human reviewed the diagnosis and diff before this PR was opened.

## Summary

Organization owners and admins can now read provider catalog data without deployment-operator access. Deployment-wide provider credential endpoints remain restricted to deployment operators.

The change separates provider, catalog, and credential routes. It adds integration and access-control checks for catalog access, session authentication, and credential restrictions. Documentation updates describe the revised access rules.

## Benefit

Organization users can access provider information without gaining access to deployment-wide credentials.


src/gateway/api/routes/providers.py
```diff
@@ -169,7 +180,7 @@ def _to_known_schema(provider: KnownProvider) -> KnownProviderSchema:
     )
 
 
-@router.get("/providers/catalog")
+@catalog_router.get("/providers/catalog")
```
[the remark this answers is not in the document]
tbille: Added a server-enforced `limit` query parameter (default and max 500) on `GET /providers/catalog`, so this response can't grow unbounded even after a future any-llm bump. One nuance worth noting: today's list is any-llm's fixed, curated provider registry (52 entries, no DB, no tenant growth) rather than tenant-owned data, so it isn't the same risk class as an unbounded query — but adding the cap costs nothing and keeps the guideline satisfied either way. OpenAPI spec and Postman collection regenerated for the new param.

Also added a one-line docstring to the one genuinely undocumented new helper (`_default_organization_id`). I left the pre-existing, self-descriptively-named test functions and `_register_core_routers` (a large private function I only added one line to) undocumented, consistent with this repo's own AGENTS.md convention of docstrings only where the logic isn't obvious from the name.

docs/public/openapi.json
```diff
@@ -27743,8 +27743,24 @@
...
+        "description": "List every known provider for the add-provider picker: id and name only.\n\nLightweight by design so the picker never lags: provider ids come from the\nany-llm registry and names from the bundled genai-prices dataset, so no\nprovider SDK is imported. The autofill hints for a chosen provider come from\nGET /api/v1/providers/catalog/{provider_id}, which imports only that one SDK.\n\n``limit`` is a ceiling rather than a page size: the registry any-llm\nexposes (`AnyLLM.get_supported_providers()`) is a small, curated set fixed\nby that dependency's own pinned version (52 today), not tenant data that\ngrows unbounded, so the default comfortably covers it. It exists so a\nfuture any-llm bump cannot make this response unbounded, not to page\ntoday's list; the picker's own search box is how a caller narrows it.",
         "operationId": "providers-provider_catalog",
+        "parameters": [
```
khaledosman: This spec change owes a third generated artifact. `web/src/client/schema.ts` is generated from this file and committed with a drift check, and the `dashboard` job fails at `git diff --exit-code -- src/client/schema.ts`: the committed client still has `query?: never` for `providers-provider_catalog` and carries neither the `limit` parameter nor the 422.

Run `pnpm --dir web run client:generate` and commit, or drop the parameter (see the comment on `providers.py`).

src/gateway/api/routes/providers.py
```diff
@@ -169,19 +189,31 @@ def _to_known_schema(provider: KnownProvider) -> KnownProviderSchema:
     )
 
 
-@router.get("/providers/catalog")
-async def provider_catalog() -> list[KnownProviderSummarySchema]:
+@catalog_router.get("")
+async def provider_catalog(
+    limit: Annotated[
```
khaledosman: Suggest dropping this parameter and the commit that added it.

`list_known_provider_summaries()` builds an in-memory list sorted by display name from a registry fixed by the pinned any-llm. The closest sibling on this same gate, `GET /tools` (`tools.py:99`), carries no limit for exactly that reason; `/pricing` carries one because it reads a growing table.

And `summaries[:limit]` truncates silently. If the registry ever did pass 500, the picker would lose the alphabetical tail with no signal, which is the same "provider missing from the dropdown" failure this PR exists to fix. The cost paid today is a permanent public query parameter, a 422 on the contract, and a third generated artifact to keep in step.

tests/unit/test_operator_gate_declarations.py
```diff
@@ -67,6 +67,7 @@
     ("models", models.operator_router),
     ("pricing", pricing.operator_router),
     ("providers", providers.router),
+    ("providers.credentials", providers.credentials_router),
```
khaledosman, line 70: `credentials_router` is listed here, but `catalog_router` is still absent from `_NON_OPERATOR_ROUTERS` (line 86). That list is what asserts both directions for a reader router: it keeps its own gate, and it gains no operator gate. `catalog_router` currently passes `test_every_router_declares_a_gate_or_says_why_it_does_not` only by declaring some gate, so moving these two routes back onto an operator-gated router would go uncaught.

Add `("providers.catalog", providers.catalog_router, verify_catalog_reader)`; both names are already imported in this module.

`tests/unit/test_dashboard_session.py:412` (`test_the_cookie_still_reads_the_catalog`) enumerates the same plane as `/models`, `/pricing`, `/tools` and wants the fourth surface too.

src/gateway/api/routes/providers.py
```diff
@@ -1,23 +1,32 @@
...
+over config.yml providers. Every route in this module describes or changes the
+gateway's own configuration, so both routers are operator-gated; standalone-mode
+only (neither is mounted in hybrid).
+
+The ``/api/v1/providers/catalog`` reads are the exception, on their own
+``catalog_router`` below, the same split ``models.py`` and ``pricing.py`` use:
+they describe the any-llm registry rather than this deployment, so the
+add-provider picker on the organization-scoped provider-keys page
+(``org_provider_keys.py``, open to an organization owner/admin, not only a
+deployment operator) needs them too. See ``api/deps.verify_catalog_reader`` for
```
khaledosman, line 16: Pointing at `deps.verify_catalog_reader` while leaving its docstring describing the old plane. `deps.py:473-474` still enumerates it as `GET /api/v1/models`, `GET /api/v1/pricing` and `GET /api/v1/tools`, and `deps.py:480` still says "The three routers that serve these reads". Four now.

src/gateway/api/routes/providers.py
```diff
@@ -54,6 +63,17 @@
...
+    prefix="/providers",
+    tags=["providers"],
+    dependencies=[Depends(require_deployment_operator)],
+)
+catalog_router = APIRouter(
+    prefix="/providers/catalog",
+    tags=["providers"],
+    dependencies=[Depends(verify_catalog_reader)],
+)
+credentials_router = APIRouter(
```
khaledosman, line 75: The prefix refactor and the `credentials_router` split are correct, but they are a separate change: nine route decorators rewritten inside a bug fix, and `AGENTS.md` asks for minimal, targeted edits over broad refactors. Squash-merge also means this lands in the changelog under the `fix(gateway)` title, which does not describe it. Worth splitting into its own PR.

khaledosman (commented):
The router split is the pattern `src/gateway/AGENTS.md` prescribes for a route on a deployment-wide router that needs a different rule, and route paths are provably unchanged (`openapi-spec` is green over the prefix refactor).

The `dashboard` job is red: the spec change left `web/src/client/schema.ts` stale. Details inline, along with the gate roster that still does not list `catalog_router`.

One scope note rather than a finding: `verify_catalog_reader` admits any signed-in session, not only an owner or admin, so `role="owner"` in the new test is decorative. That matches the rest of the catalog plane and the description says so.

tests/integration/test_provider_credentials_api.py
```diff
@@ -426,6 +431,78 @@ def test_catalog_detail_requires_master_key(client: TestClient) -> None:
...
+
+    `/providers/catalog` and its detail route feed `ProviderComboBox` on the
+    organization-scoped provider-keys page (`org_provider_keys.py`), which
+    `require_active_organization_management_access` already opens to an
+    organization owner or admin. Gating the picker's own catalog reads behind
+    `require_deployment_operator` left that caller unable to populate the
+    dropdown, and therefore unable to add a key at all, even though they were
+    never asking for deployment-wide authority.
+    """
+    assert client.get(f"{API_ROOT}/organizations/me", headers=master_key_header).status_code == 200
```
khaledosman: This asserts a status code from a route unrelated to the test's subject, with no stated purpose. The default organization is seeded by the migration chain (`alembic/versions/d5e7f1a2b3c4`, restored per test by `clean_database`), so `_default_organization_id` does not need it.

If it is here to force first-boot provisioning before the `/provider-credentials` 403 is asserted, say so in a comment; otherwise drop it.

tests/integration/test_provider_credentials_api.py
```diff
@@ -426,6 +431,78 @@ def test_catalog_detail_requires_master_key(client: TestClient) -> None:
     assert client.get(f"{API_ROOT}/providers/catalog/openai").status_code in (401, 403)
 
 
+def _default_organization_id(session_factory: Callable[[], Session]) -> uuid.UUID:
+    """The bootstrap tenancy root's id, seeded by the migration chain under the "default" slug."""
+    session = session_factory()
+    try:
+        organization = session.query(Organization).filter(col(Organization.slug) == "default").one()
```
khaledosman: Use `DEFAULT_ORGANIZATION_SLUG` from `gateway.services.tenancy.provisioning_service` rather than the literal, as `tests/integration/test_workspace_member_budget_policies.py:46` does. The helper's docstring is right that the migration chain seeds it, and the constant is what the migration and the runtime both look it up by.

src/gateway/api/deps.py
```diff
@@ -471,14 +471,15 @@ async def verify_catalog_reader(
...
+    the deployment rather than act on it: ``GET /api/v1/models``, ``GET /api/v1/pricing``,
+    ``GET /api/v1/tools``, and ``GET /api/v1/providers/catalog`` (with their by-id
+    variants). The dashboard's Models, Pricing, and add-provider pages are built on
+    these, so a session has to reach them; they call no provider, write nothing, and
+    bill nothing, so reaching them deployment-wide costs a signed-in caller's own
+    organization nothing.
 
     Split out rather than left as a branch inside the other dependency so that
-    adding a route to this plane defaults to refusing the cookie. The three
+    adding a route to this plane defaults to refusing the cookie. The four
```
khaledosman, line 494: Nit, and pre-existing rather than introduced here. `require_deployment_operator` carries the same fact from the gate's side and still says it the old way at line 407: "The three modules that hold an exception (``models.py`` and ``pricing.py`` for the catalog reads, ``usage.py`` for external-event ingestion)". It was already understated on main, since `tool_settings.py` holds an operator/reader split too, and `providers.py` now puts it further off.

Worth fixing in the same pass as the one just corrected here, since the two enumerations drift together.

khaledosman (approved):
All six code findings addressed, and CI is green. Dropping the `limit` reverted the spec and Postman changes too, so the PR no longer touches `docs/public/` or `web/`; the `dashboard` and `generate` jobs are absent because those workflows filter on those paths, not because a check was lost.

One thing I checked rather than assumed: removing the `organizations/me` call left nothing to provision the bootstrap identity, but `has_administration_access` (`deployment_user_service.py:90-93`) returns `False` when the marker is absent, so the `/provider-credentials` 403 still holds for the right reason.

One nit inline, non-blocking. The prefix and `credentials_router` refactor is still riding along in this PR; still worth splitting, but not something I would hold the fix for.

docs/public/openapi.json
tbille: Fixed by dropping the `limit` parameter instead (see the reply on the `providers.py:194` thread) — regenerated `openapi.json`, the Postman collection, and `web/src/client/schema.ts` to match. No drift now (`230afc0e`).

src/gateway/api/routes/providers.py
tbille: Agreed, dropped the param and the commit that added it. `list_known_provider_summaries()` stays unbounded, matching its sibling `GET /tools`, and the OpenAPI spec / Postman collection / dashboard client are regenerated to match (`230afc0e`).

tests/unit/test_operator_gate_declarations.py
tbille: Added `("providers.catalog", providers.catalog_router, verify_catalog_reader)` to `_NON_OPERATOR_ROUTERS`, and extended `test_the_cookie_still_reads_the_catalog` in `test_dashboard_session.py` to assert the fourth surface too (`230afc0e`).

src/gateway/api/routes/providers.py
tbille: Updated `verify_catalog_reader`'s docstring to name `GET /api/v1/providers/catalog` and say four routers instead of three (`230afc0e`).

tests/integration/test_provider_credentials_api.py
tbille: Dropped the assertion and the now-unused `master_key_header` parameter. It wasn't there to force provisioning; the default organization is already seeded by the migration chain and restored per test by `clean_database` (`230afc0e`).

tests/integration/test_provider_credentials_api.py
tbille: Switched to `DEFAULT_ORGANIZATION_SLUG` from `gateway.services.tenancy.provisioning_service` (`230afc0e`).

src/gateway/api/deps.py
tbille: Good catch, thanks. Updated the enumeration to add `tool_settings.py`'s reader split and `providers.py`, and dropped the hardcoded count so this and `verify_catalog_reader`'s docstring stop drifting independently (`e817261c`).

src/gateway/api/routes/providers.py
tbille: Fair point on `AGENTS.md`'s minimal-diff guidance, but keeping the prefix/`credentials_router` split in this PR rather than splitting it out — it's a direct, small consequence of the same router this PR already touches for the auth fix, and it's covered by the same test run. Appreciate the flag.

khaledosman:
might be related https://github.com/mozilla-ai/otari/pull/1215

peteski22:
Closing as a duplicate of #1215, which fixes the same bug. #1215 has the smaller backend change, and it also fixes the provider picker: a failed catalog read no longer shows as "No provider to offer here."

The review fixes made here that still apply are in #1215: `providers.catalog_router` is listed in `_NON_OPERATOR_ROUTERS`, `verify_catalog_reader`'s docstring names the provider catalog, and the exception list in `require_deployment_operator`'s docstring is corrected. The prefix and `credentials_router` refactor did not carry over. It can come back as its own PR if we still want it.

PR #1237, opened 2026-09-16
fix(db): close the request session when a data-plane dependency is torn down

### LINKED ISSUE #1236: Pooled DB connections leak on failed or canceled inference requests until the pool is exhausted (503)
## Symptom

A production deployment on PostgreSQL (Otari 0.5.0, asyncpg) reports that connections are never returned to the pool. After enough traffic every request fails with:

```
Database connectivity check failed: QueuePool limit of size 10 overflow 20 reached, connection timed out, timeout 30.00
POST /v1/messages HTTP/1.1" 503 Service Unavailable
```

preceded by repeated:

```
The garbage collector is trying to clean up non-checked-in connection <AdaptedConnection <asyncpg.connection.Connection ...>>, which will be terminated.
Exception terminating connection ... asyncio.exceptions.CancelledError: Cancelled via cancel scope ... by <Task ... BaseHTTPMiddleware.__call__.<locals>.call_next.<locals>.coro ...>
```

## Root cause

`get_db_if_needed` in `src/gateway/api/deps.py` consumes the `get_db` dependency with a bare `async for`:

```python
async for db in get_db():
    yield db
```

An `async for` never closes the iterator it drives. When FastAPI tears the dependency down at the end of the request, the inner `get_db()` generator is abandoned with its `async with _SessionLocal()` block unfinished, so the session is never closed. The connection comes back only when the garbage collector finalizes that generator, which asyncio does by scheduling `aclose()` at an arbitrary later moment in whatever task is current. Under sustained traffic that reclaim is outrun and the pool fills. When the reclaim lands in a request task that is being cancelled (client disconnect mid-stream, which is the `BaseHTTPMiddleware` cancel scope in the trace), the close is cancelled and the connection is lost permanently. The trace in the report shows that second step, not the leak itself.

Every inference route (`/v1/messages`, `/v1/chat/completions`, `/v1/responses`) takes its session through this dependency.

## Proof

Driving the real `get_db` and the real engine against PostgreSQL 17 over asyncpg (`AsyncAdaptedQueuePool`), with no `await` between teardown and the read:

```
async-for checkedout: during=1 right_after_teardown=1 after_one_tick=1  gc_finalizer_ran_for_get_db=True
aclosing  checkedout: during=1 right_after_teardown=0 after_one_tick=0  gc_finalizer_ran_for_get_db=False
```

## Why it was not seen before

Introduced in 3e593b66 (the async DB conversion) and present in every release since, including 0.6.1 and current `main`. The code was a mechanical translation of a synchronous `for db in get_db()`, which was correct because reference counting reclaimed the sync generator immediately. Three things hide it: the SQLite default uses `NullPool`, which has no pool accounting at all; the suites run at low concurrency, so the GC reclaim keeps up; and nothing asserts pool checkout state.

`db_pool_recycle` does not help: it retires idle connections that are back in the pool, and a leaked connection is checked out. The code default is already 1800, so setting it to 1800 changes nothing.

## Follow-up (separate)

`SecurityHeadersMiddleware` is the last `BaseHTTPMiddleware` in the stack. It is what turns a delayed reclaim into a permanent loss, and it should become pure ASGI like `InFlightMiddleware` and `MetricsMiddleware` already are. That is a distinct change from this fix.

daavoo:
## Description

On PostgreSQL, an inference request (Messages, Chat Completions, Responses) that raised or was canceled left its database connection checked out after the response was sent. The connection only came back when Python's garbage collector got around to it, and under steady traffic that was not fast enough: the pool filled up and the gateway answered every request with a 503 until it was restarted.

The cause is one line in the dependency that hands a request its session. It iterated the underlying session generator with a plain `async for`, which never closes what it iterates. FastAPI tears a yield dependency down through `asynccontextmanager`: on success it calls `anext()`, which ran `get_db` to its end and closed the session, so a successful request did not leak. When the request raises or is canceled it calls `athrow()`, which raised out of the bare `async for` and left `get_db` suspended until garbage collection. So the leak hit every request that raised or was canceled: an `HTTPException` from the handler or from a later dependency such as auth or budget, a provider error, or a client disconnect mid-stream. This change wraps the iteration in `aclosing`, so the connection returns to the pool on every teardown path, in the request's own task.

This affects 0.5.0 through 0.6.1 and current `main`. It does not reproduce on the SQLite default, which has no connection pool, and `db_pool_recycle` cannot help because recycling only applies to connections that are already back in the pool.

## How to test it locally

Automated: `tests/unit/test_deps_session_teardown.py` drives the dependency the way FastAPI does (enter, then `anext()` for a success teardown or `athrow()` for a failed or canceled one) and asserts the session is closed. The two `athrow()` cases, `RuntimeError` and `CancelledError`, fail on `main` and pass here; the success case passes on both, because that path never leaked. `make lint` and `make typecheck` pass locally. I did not run the full `make test` locally; CI covers it.

Against a real pool, the reproduction (PostgreSQL 17 over asyncpg, `AsyncAdaptedQueuePool`) showed the connection staying checked out after an `athrow()` teardown under the bare `async for`, and returning immediately under `aclosing`. The earlier figures quoted here were taken from a teardown that closed the session on both versions, so they did not show what they claimed; the `athrow()` path is the one that distinguishes them.

To eyeball it on a running gateway: point `database_url` at PostgreSQL, send a few hundred `POST /v1/messages` with a client that cancels mid-stream, and watch for the `non-checked-in connection` warnings. They stop with this change.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes #1236

## Checklist

- [ ] I ran the Definition of Done checks locally (`make lint`, `make typecheck`, `make test`). Lint and typecheck yes; `make test` left to CI.
- [x] Documentation was updated where necessary. None needed; no user-facing surface changed.
- [x] If the API contract changed, I regenerated the OpenAPI spec (`uv run python scripts/generate_openapi.py`). Not applicable.

## AI Usage

- [ ] No AI was used.

**Any additional AI details you'd like to share:** Diagnosis, fix, regression test and the asyncpg reproduction were done by the agent under review by @daavoo. The first hypothesis (the `BaseHTTPMiddleware` cancel scope in the trace) was wrong and is recorded in the issue as the secondary effect it is.

## Summary

- Ensured `get_db_if_needed` closes the asynchronous database generator during request teardown.
- Added regression coverage for normal, exceptional, and canceled teardown paths.
- Prevented PostgreSQL connection leaks that could exhaust the pool and cause 503 responses.


src/gateway/api/deps.py
```diff
@@ -534,8 +535,16 @@ async def get_db_if_needed(
...
 
-    async for db in get_db():
-        yield db
+    # ``aclosing`` rather than a bare ``async for``: ``async for`` never closes
+    # the iterator it drives, so tearing this dependency down would abandon
+    # ``get_db``'s generator with its ``async with`` block unfinished. The
+    # session would then be closed only when the garbage collector finalized
+    # that generator, at an arbitrary later moment in an unrelated task, and its
+    # pooled connection stays checked out until then. Under load that outruns
+    # the pool.
```
peteski22: This narrates history, and the mechanism it gives is off. On a normal teardown FastAPI calls `anext()`, and the bare `async for` already ran `get_db` to its end. Only an error or cancellation thrown in at teardown (`athrow()`) left `get_db` open. The comment only needs the gotcha, so nobody turns this back into a bare `async for`:

```suggestion
    # A bare ``async for`` leaves ``get_db`` open when an error or cancellation is thrown in at teardown,
    # so its session would hold a pooled connection until garbage collection.
```

tests/unit/test_deps_session_teardown.py
```diff
@@ -0,0 +1,81 @@
+"""Regression coverage for the pooled connection leak in ``get_db_if_needed``.
+
+The dependency used to consume ``get_db`` with a bare ``async for``. An
+``async for`` never closes the iterator it drives, so tearing the dependency
+down abandoned ``get_db``'s generator with its ``async with`` block unfinished.
+The session was then closed only when the garbage collector finalized that
+generator, at an arbitrary later moment in an unrelated task, leaving its
+pooled connection checked out until then. Under sustained traffic that outran
+the pool and every request failed with a ``QueuePool`` timeout.
+"""
```
peteski22: Same wrong model as the comment in `deps.py`. A module docstring should also say what the tests cover, not what the code used to do:

```suggestion
"""Teardown of ``get_db_if_needed`` closes its session whether the request succeeds, fails or is canceled."""
```

tests/unit/test_deps_session_teardown.py
```diff
@@ -0,0 +1,81 @@
...
+async def test_teardown_closes_the_session_it_borrowed(monkeypatch: pytest.MonkeyPatch) -> None:
+    session = _FakeSession()
+    monkeypatch.setattr(database, "_SessionLocal", lambda: session)
+
+    dependency = get_db_if_needed(GatewayConfig(database_url="sqlite:///./test.db"))
+    yielded: object = await anext(dependency)
+    assert yielded is session
+    assert not session.closed, "the session must stay open for the life of the request"
+
+    await dependency.aclose()
```
peteski22: FastAPI never calls `aclose()` on a yield dependency; a successful request is torn down with `anext()`. This test fails on `main` only because `aclose()` throws `GeneratorExit` into the `yield`, which a real request never does, so it suggests successful requests leaked. Tear down the way FastAPI does. The test then passes on `main` too, which is correct, because that path never leaked:

```suggestion
    with pytest.raises(StopAsyncIteration):
        await anext(dependency)
```

tests/unit/test_deps_session_teardown.py
```diff
@@ -0,0 +1,81 @@
...
+    """The connection goes back even when the handler raises, which is the path a
+    cancelled or erroring request takes."""
+    session = _FakeSession()
+    monkeypatch.setattr(database, "_SessionLocal", lambda: session)
+
+    dependency = get_db_if_needed(GatewayConfig(database_url="sqlite:///./test.db"))
+    await anext(dependency)
+
+    with pytest.raises(RuntimeError):
+        await dependency.athrow(RuntimeError("handler blew up"))
```
peteski22: This is the real regression test, but it only throws `RuntimeError`. The trace in #1236 is a client disconnect, which reaches the dependency as `CancelledError`. Cover both (needs `import asyncio`). The new docstring also fixes "cancelled", which is "canceled" in prose here:

```suggestion
@pytest.mark.asyncio
@pytest.mark.parametrize(
    "error", [RuntimeError("handler blew up"), asyncio.CancelledError()], ids=["error", "canceled"]
)
async def test_teardown_closes_the_session_when_the_request_fails(
    error: BaseException, monkeypatch: pytest.MonkeyPatch
) -> None:
    """The connection goes back when the request raises or is canceled."""
    session = _FakeSession()
    monkeypatch.setattr(database, "_SessionLocal", lambda: session)

    dependency = get_db_if_needed(GatewayConfig(database_url="sqlite:///./test.db"))
    await anext(dependency)

    with pytest.raises(type(error)):
        await dependency.athrow(error)
```

peteski22 (approved):
The fix is right: `aclosing` closes `get_db` on every teardown path, so the session no longer waits for garbage collection. The comments, one test and the description explain the leak wrongly, though.

FastAPI tears a yield dependency down through `asynccontextmanager`. On success it calls `anext()`; when the request raises or is canceled it calls `athrow(exc)`. With the old bare `async for`, `anext()` ran `get_db` to its end and closed the session, so a successful request did not leak. `athrow()` raised out of the loop and left `get_db` suspended until garbage collection. So the leak hit every request that raised or was canceled (an `HTTPException` from the handler or from a later dependency such as auth or budget, a provider error, a client disconnect mid-stream), not every inference request.

Checked on FastAPI 0.135.3 with the old and new dependency behind `asynccontextmanager` and `AsyncExitStack`: on success both close the session; on `RuntimeError` and `CancelledError` the old one leaves it open and this one closes it.

The description and #1236 should say the same. "Every inference request" overstates it, and the reproduction figures (`right_after_teardown=1` for `async for`) cannot come from a success teardown, which closes the session on both versions.

Follow-up split out of #1236: #1238.

<sub>Drafted with Claude Code.</sub>

src/gateway/api/deps.py
daavoo: Applied. Comment now states only the gotcha: a bare `async for` leaves `get_db` open when an error or cancellation is thrown in at teardown.

tests/unit/test_deps_session_teardown.py
daavoo: Applied, verbatim.

tests/unit/test_deps_session_teardown.py
daavoo: Applied. Confirmed your reading: with the bare `async for` restored, this test passes and only the two `athrow()` cases fail, which is the correct split.

tests/unit/test_deps_session_teardown.py
daavoo: Applied, verbatim. Both parameters fail on the old mechanism and pass here.

Issue #1241, opened 2026-09-16
Data-plane database sessions are untested: a pooled connection leak cannot fail a test


daavoo:
### What is wrong

The test suite cannot see the health of the database connection pool on the inference routes, so a leaked connection there reaches production without failing anything. That is how mozilla-ai/otari#1236 got in: every inference request kept a pooled connection checked out until the garbage collector reclaimed it, and under sustained traffic the pool drained and every request came back as a 503.

### Why nothing caught it

Two reasons, and both have to be fixed for a test to be able to fail.

1. The integration suite hands the app a database session of its own by overriding the session dependency. The inference routes do not take that dependency directly; they take a wrapper that decides whether a local session is needed at all (hybrid mode has no local database), and that wrapper calls the session dependency as a plain function rather than through FastAPI's dependency system. The override therefore never reaches the inference routes: they run against the engine the app built at startup, while the management routes run against the test engine. Confirmed by observation, not just by reading: checkouts during a chat completion land on the startup engine.
2. Nothing anywhere asserts pool state, so even on the right engine a connection that is never handed back is invisible.

There is a third trap worth writing down, because it makes a naive test pass on the broken code. The abandoned session does get cleaned up eventually, by the event loop finalizing the orphaned generator, so a check made a moment after the request succeeds whether or not the code returns the connection. A gate has to measure the teardown in the same task that performed it.

### What we want instead

An integration-level assertion over the real pool, so a future leak on the data plane fails a test rather than a deployment. Note that the pool counter only exists on PostgreSQL: the SQLite setup uses a pool that holds nothing.

PR #1242, opened 2026-09-16
feat(observability): export database connection pool stats on /metrics

### LINKED ISSUE #1240: Connection pool exhaustion is invisible in metrics and slow to detect in the readiness probe
A production deployment exhausted its PostgreSQL connection pool and every request started answering 503. Two operational gaps made that outage invisible while it built up, and slow to detect once it arrived. They are separate problems with one shared cause: nothing in the gateway reports how much of the connection pool is in use.

## Gap 1: the connection pool is not observable

The Prometheus endpoint exports request counts, durations, tokens, cost, budget events and usage-log writer health, but nothing at all about the SQLAlchemy connection pool. There is no way to see connections being checked out and not returned, so a slow leak is invisible right up to the moment requests begin failing, and the first signal an operator gets is a wall of 503s.

What is missing, at minimum: connections currently checked out, connections idle and available, connections open beyond the base pool size, and the ceiling the pool will never hand out more than. The usage-log writer runs on a second, smaller pool of its own, so that one wants the same treatment.

## Gap 2: the readiness probe takes 30 seconds to say what it already knows

The readiness probe issues a trivial query to prove the database is reachable. When the pool is saturated that query does not fail, it queues, for the full pool timeout (30 seconds by default), and only then reports a generic "database unavailable". Three things go wrong at once:

- Kubernetes waits out the whole timeout on every probe before it can mark the pod unready, so the pod keeps taking traffic long after it can serve any.
- The probe is itself queueing for a pool slot, so it competes with real requests for the resource that has already run out.
- The reported state names the wrong problem. The database is perfectly healthy; this process has run out of ways to reach it. An operator reading "database unavailable" goes and looks at the database.

The log line from the real incident:

```
Database connectivity check failed: QueuePool limit of size 10 overflow 20 reached, connection timed out, timeout 30.00
```

## Proposed fix

One helper that reads the pool counters for the active engines, shared by both halves:

- Prometheus gauges for checked out, idle, overflow and capacity, filled at scrape time rather than from a background timer, labeled by which pool they describe.
- The readiness probe consults that helper before issuing its query. If the pool has no capacity left it answers 503 immediately, naming pool exhaustion as its own state rather than reusing "database unavailable". Reading the counters is side-effect free, so this needs no timeout and no cancellation, which matters: cancelling a database operation mid-checkout is the hazard behind the leak fixed in #1236.

The helper needs to cope with the SQLite default pool, which keeps none of these counters, and with being called before the database is initialized. Neither is an error; both mean there is no pool ceiling to report.

daavoo:
## Description

When Otari runs on PostgreSQL it serves requests from a fixed pool of database connections. If that pool runs out, every request fails at once, and until now there was no way to see it coming: nothing in `/metrics` said how many connections were in use or how many the pool would ever hand out, so the first sign of trouble was a gateway answering 503 and no way to tell from the outside whether the database or the pool was the problem.

This adds four gauges to `/metrics`, labeled by which pool they describe (`request` for request traffic, `log` for the usage-log writer): connections checked out, connections idle, connections open beyond the base pool size, and the ceiling the pool will never exceed. An operator can now alert on checked-out connections approaching that ceiling and act before requests start failing. `docs/deployment.md` says how.

Nothing about how Otari serves requests changes. The numbers are the ones SQLAlchemy already keeps, read at scrape time, so a scrape reports the live pool rather than a value a timer left behind. They do not appear on SQLite, which opens a connection per use and keeps no pool to report.

### What changed after review

An earlier version of this branch also made the readiness probe answer 503 as soon as the pool was full. That is out, on review: one instantaneous reading cannot tell a pool that is briefly full from one that is stuck, so under load every replica would have gone unready together and a traffic spike would have become an outage. A readiness check that judges the pool over time is tracked separately in #1247.

Fixes #1240

## How to test it locally

Requires PostgreSQL to see the gauges at all; on SQLite the endpoint correctly reports no pool series.

```bash
# Unit coverage for the stats reader and the collector
uv run pytest tests/unit/test_database_pool_stats.py tests/unit/test_gateway_metrics.py -q

# Against a running gateway with OTARI_ENABLE_METRICS=true and a PostgreSQL database
curl -s localhost:8000/metrics | grep gateway_db_pool
```

Already run locally and green: `make lint`, `make typecheck`, the full `uv run pytest tests/unit` (3654 passed), `make openapi-check`, `make postman-check`, and the OSS-edition smoke gate (`uv run --frozen --no-dev python scripts/oss_edition_smoke.py`). The integration suite was left to CI.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes #1240. Follow-up: #1247.

## Checklist

- [x] I ran the Definition of Done checks locally (`make lint`, `make typecheck`, `make test`). `make lint`, `make typecheck` and the full unit suite are green; the integration suite was left to CI.
- [x] Documentation was updated where necessary. `docs/deployment.md` documents the new metrics.
- [x] If the API contract changed, I regenerated the OpenAPI spec (`uv run python scripts/generate_openapi.py`). No API surface changes now that the readiness change is out, and the spec, the Postman collection and the dashboard client are back to their `main` contents; `make openapi-check` and `make postman-check` pass.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

The metric names and label set were pinned into the existing `test_scrape_exposes_the_pinned_families` test, because a scrape is an external contract that dashboards and alerts outside this repository key on.

**NOTE:**

## Summary

- Added PostgreSQL connection-pool metrics for request and usage-log pools.
- Added unit tests for pool statistics and Prometheus output.
- Updated gateway metrics coverage and deployment guidance.
- Documented pool capacity and SQLite limitations.

These changes make connection leaks easier to detect and support operational alerts before the pool is exhausted.

## Technical notes

- Metrics report checked-out, idle, overflow, and capacity connections.
- Pool capacity includes configured overflow for the request pool.
- SQLite does not expose these pooled-connection metrics.


src/gateway/api/routes/health.py
```diff
@@ -123,6 +132,22 @@ async def health_readiness(
...
+            pool.capacity,
+        )
+        raise HTTPException(
+            status_code=status.HTTP_503_SERVICE_UNAVAILABLE,
+            detail={
+                "status": "unhealthy",
+                "database": POOL_EXHAUSTED,
+                "version": __version__,
+            },
+        )
```
peteski22: Please drop this check from this PR.

`is_saturated` is one reading at one moment. A busy pool can be full for a moment while its connections come back within milliseconds. The `SELECT 1` below would get one of them and pass. Under sustained load every replica fills its pool at the same time, so every replica goes unready together, and a traffic spike becomes an outage. #1247 tracks a check that judges the pool over time.

The `logger.error` also fires on every probe from every pod while the pool is full. The 503 body and the pool gauges already show the same thing.

Dropping the check also removes `POOL_EXHAUSTED`, `request_pool_stats`, `PoolStats.is_saturated`, the new docstring lines, the readiness tests, the readiness sentence in `docs/deployment.md`, and the regenerated OpenAPI, Postman and `schema.ts` changes.

src/gateway/metrics.py
```diff
@@ -19,6 +19,7 @@
 from starlette.responses import Response
 
 from gateway.core.config import API_ROOT, API_VERSION
+from gateway.core.database import pool_stats
```
peteski22: This puts database gauges back into `metrics.py`, which #1205 just emptied. It also makes `metrics.py` import `core.database`, so `core.database` can never declare a metric of its own: that import would loop.

Suggest a collector in `core/database.py` that reads `pool_stats()` on each scrape. `metrics.py` then re-exports `GaugeMetricFamily` and `Collector`, and drops the four gauges, `refresh_db_pool_metrics` and its call in `metrics_endpoint`.

```python
class _PoolCollector(Collector):
    def collect(self) -> Iterator[GaugeMetricFamily]:
        checked_out = GaugeMetricFamily(
            "gateway_db_pool_connections_checked_out",
            "Pooled database connections currently checked out",
            labels=["pool"],
        )
        # idle, overflow and capacity the same way
        for name, stats in pool_stats().items():
            checked_out.add_metric([name], stats.checked_out)
        yield checked_out


REGISTRY.register(_PoolCollector())
```

A collector emits no series when there is no pool, so SQLite shows none and a disposed engine leaves no old values behind. Every `generate_latest(REGISTRY)` call gets current values, not only `metrics_endpoint`. The pinned family test then reads label names from the metric when the collector has none: `getattr(collector, "_labelnames", None) or getattr(metric, "_labelnames", ())`. Both work on prometheus_client 0.24.1.

src/gateway/core/database.py
```diff
@@ -196,6 +197,96 @@ async def release_session(session: AsyncSession | None) -> bool:
...
+        return None
+    # Negative until the pool has created its full complement of base
+    # connections, which would read as "overflow in use" the wrong way round.
+    overflow = max(pool.overflow(), 0)
+    return PoolStats(
+        checked_out=pool.checkedout(),
+        checked_in=pool.checkedin(),
+        overflow=overflow,
+        size=pool.size(),
+        max_overflow=max(getattr(pool, "_max_overflow", 0), 0),
```
peteski22: `_max_overflow` is a private SQLAlchemy field. If it goes away, capacity silently reads as `size`, and an alert on checked-out connections near capacity fires early. The gateway sets this value itself: `db_max_overflow` for the request pool and `0` for the log pool. Please take it from there. The docstring paragraph about the fallback then goes too.

src/gateway/core/database.py
```diff
@@ -196,6 +197,96 @@ async def release_session(session: AsyncSession | None) -> bool:
     return True
 
 
+# The name each pool reports under, in metrics and in anything that iterates
+# :func:`pool_stats`.
```
peteski22: This comment describes the code that uses these names. The constant names already say what they are, so the comment can go.

src/gateway/core/database.py
```diff
@@ -196,6 +197,96 @@ async def release_session(session: AsyncSession | None) -> bool:
...
+
+
+def _engine_pool_stats(engine: AsyncEngine | None) -> PoolStats | None:
+    """Read *engine*'s pool, or ``None`` when there is nothing to read.
+
+    Returns ``None`` for an engine that was never built and for one on
+    ``NullPool``, which is SQLite's pool here and implements none of the
+    counters below: it opens a connection per checkout and keeps no pool to
+    saturate. Callers treat ``None`` as "no pool ceiling applies" rather than
+    as an error, so this never raises.
```
peteski22: Please describe the contract, not what callers do with the result. For example: "Returns `None` when there is no pool to read: no engine yet, or a pool with no counters (`NullPool`, which SQLite uses)."

peteski22 (commented):
Drafted with Claude Code (Claude Opus 5), then checked by a human before posting.

The pool metrics cover the first gap in #1240. Four things before merge:

1. Move the pool gauges from `metrics.py` to `core/database.py`, so they follow the pattern from #1205.
2. Drop the readiness change. One reading cannot tell a busy pool from a stuck one. #1247 tracks the design.
3. Take pool capacity from config, not from a private SQLAlchemy field.
4. The PR description is for a different change. It says test-only and `Fixes #1241`, so merging would close #1241 without its integration test. The code here fixes #1240.

The older log and raise in the `SELECT 1` branch is #1246, so it stays out of this PR.

src/gateway/core/database.py
```diff
@@ -196,6 +197,96 @@ async def release_session(session: AsyncSession | None) -> bool:
...
+    defensively and falls back to ``0``, which understates capacity rather than
+    inventing it.
+    """
+    if engine is None:
+        return None
+    pool = engine.pool
+    if not isinstance(pool, _CountingPool):
+        return None
+    # Negative until the pool has created its full complement of base
+    # connections, which would read as "overflow in use" the wrong way round.
```
peteski22: Nit: "`QueuePool` reports a negative overflow until it has opened its base connections." says the same thing more plainly.

tests/unit/test_gateway_metrics.py
```diff
@@ -238,3 +243,51 @@ def test_metrics_expose_process_memory() -> None:
...
+def test_db_pool_gauges_are_left_alone_without_a_pool(monkeypatch: pytest.MonkeyPatch) -> None:
+    """SQLite runs on NullPool, and an uninitialized engine has no pool at all."""
+    from gateway.core import database
+
+    monkeypatch.setattr(database, "_engine", None)
+    monkeypatch.setattr(database, "_log_engine", None)
+
+    refresh_db_pool_metrics()
+
+    assert _sample("gateway_db_pool_capacity", {"pool": "unbuilt"}) == 0.0
```
peteski22: Nothing ever sets `pool="unbuilt"`, and `_sample` returns `0.0` for a missing sample. This test therefore passes even if the refresh writes zeros to `request`. With a collector, assert that `REGISTRY.get_sample_value("gateway_db_pool_capacity", {"pool": "request"})` is `None`.

docs/deployment.md
```diff
@@ -19,6 +19,27 @@ A durable standalone deployment should:
...
+connections, and running out of them makes every request fail at once. `/metrics`
+reports the pool directly, labeled by `pool` (`request` for request traffic,
+`log` for the usage-log writer):
+
+| Metric | Meaning |
+| --- | --- |
+| `gateway_db_pool_connections_checked_out` | Connections in use right now |
+| `gateway_db_pool_connections_idle` | Connections available to hand out |
+| `gateway_db_pool_overflow_connections` | Connections open beyond `db_pool_size` |
+| `gateway_db_pool_capacity` | Ceiling: `db_pool_size` plus `db_max_overflow` |
```
peteski22: These limits are right for the `request` pool only. The `log` pool uses `db_log_pool_size` and has no overflow.

src/gateway/api/routes/health.py
daavoo: Dropped in 34e10928, along with POOL_EXHAUSTED, request_pool_stats, PoolStats.is_saturated, the readiness docstring lines, the readiness tests and the deployment.md sentence. The OpenAPI spec, the Postman collection and web/src/client/schema.ts are back to their main contents, and openapi-check and postman-check pass. #1247 is now referenced from the PR body as the follow-up.

src/gateway/metrics.py
daavoo: Done in 34e10928: the four gauges and refresh_db_pool_metrics are gone, metrics.py no longer imports core.database, and core/database.py registers a _PoolCollector that reads pool_stats() on each scrape. metrics.py re-exports GaugeMetricFamily and Collector. The pinned-family test now falls back to the family's label names when the collector keeps none.

src/gateway/core/database.py
daavoo: Fixed in 34e10928: init_db records the max_overflow it configured (db_max_overflow for the request pool, 0 for the log pool) and capacity reads from there. No _max_overflow access left, and the fallback paragraph is gone.

src/gateway/core/database.py
daavoo: Removed in 34e10928.

src/gateway/core/database.py
daavoo: Reworded in 34e10928 to your suggestion.

src/gateway/core/database.py
daavoo: Taken in 34e10928.

tests/unit/test_gateway_metrics.py
daavoo: Replaced in 34e10928: the collector test asserts REGISTRY.get_sample_value("gateway_db_pool_capacity", {"pool": "request"}) is None with no pool, for both pool labels.

docs/deployment.md
daavoo: Fixed in 34e10928: the table row is now generic and a paragraph below it splits the two, request sized by db_pool_size plus db_max_overflow, log sized by db_log_pool_size with no overflow.

tests/unit/test_database_pool_stats.py
```diff
@@ -0,0 +1,142 @@
...
+    assert REGISTRY.get_sample_value("gateway_db_pool_connections_checked_out", labels) == 9.0
+
+
+def test_the_collector_publishes_no_series_without_a_pool() -> None:
+    """SQLite runs on NullPool, and an uninitialized engine has no pool at all."""
+    database._engine = None
+    database._log_engine = None
+
+    assert REGISTRY.get_sample_value("gateway_db_pool_capacity", {"pool": REQUEST_POOL}) is None
+    assert REGISTRY.get_sample_value("gateway_db_pool_capacity", {"pool": LOG_POOL}) is None
```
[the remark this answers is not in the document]
peteski22: The gap is real. These are unit tests of the stats reader, so a connection leak on the inference path would not fail them.

That coverage is tracked in #1241, and #1243 adds it. It sends chat completions through the real PostgreSQL pool the app builds at startup, and it checks that the connection comes back in the same task, before the runtime can clean up an abandoned session. This PR stays scoped to #1240, the pool metrics.

peteski22 (approved):

PR #1243, opened 2026-09-16
test: gate the data-plane connection pool against a leaked connection

### LINKED ISSUE #1241: Data-plane database sessions are untested: a pooled connection leak cannot fail a test

daavoo:
## Description

Otari recently shipped a fix for a connection leak on the inference path: every request kept a database connection reserved until the garbage collector got around to releasing it, and under sustained traffic the gateway ran out of connections and answered everything with a 503. Nothing in the test suite noticed, because the tests give the app a database connection of their own for the management endpoints and the inference endpoints quietly keep using the one the app opened at startup, and because no test ever looks at how many connections are in use.

This adds that missing coverage. Two tests now watch the connection pool the running gateway actually uses, so the next leak on the inference path fails a test run instead of a deployment. Nothing about how Otari behaves changes; this is test-only.

It also nails down a subtlety that makes this easy to get wrong. The abandoned connection does come back eventually, once the runtime cleans up after itself, so a test that checks a moment later passes on broken code. One of the two tests therefore checks at the exact point the request finishes with the connection, which is the only moment the difference is visible.

**This is expected to fail on `main` today, and to pass once #1237 merges.** #1237 carries the fix for the leak itself, and it should merge first. Verified locally both ways: the new pool test fails against current `main` and passes with #1237 applied on top. Nothing from #1237 is included here, and this branch does not touch the file it changes.

### Why the tests were not simply pointed at the test database instead

The cheaper option was to make the test setup hand its own connection to the inference routes too, so both halves of the app agree on which database they are talking to. That was rejected deliberately: it would tidy up the inconsistency while making this whole class of bug permanently untestable, because the connection the real code leaks would be one the test setup never lent out. The tests observe the real pool instead.

Fixes #1241

## How to test it locally

Requires Docker, or `TEST_DATABASE_URL` pointing at a PostgreSQL server, as the rest of the integration suite does. The pool counter these tests read only exists on PostgreSQL; the SQLite setup uses a pool that holds nothing.

```bash
# On this branch: one test fails, and that failure is the point.
uv run pytest tests/integration/test_data_plane_pool_release.py -q

# With the fix from #1237 on top: both pass.
git cherry-pick 1e0efe801947fc9c49d1960de8db4172155cb86b
uv run pytest tests/integration/test_data_plane_pool_release.py -q
git reset --hard HEAD~1
```

Already run locally and green: `make lint`, `make typecheck`, and the new module alongside neighboring integration modules that share the same worker database (`test_exact_budget_ledger.py`, `test_messages_endpoint.py`, `test_transaction_safety.py`, and the unit tests from #1237), to confirm the new module's setup and teardown does not disturb them. The full `make test` was not run locally; CI covers it, and the one expected failure above is the only one anticipated until #1237 lands.

## PR Type

- [ ] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [x] Infrastructure / CI

## Relevant issues

Fixes #1241

## Checklist

- [ ] I ran the Definition of Done checks locally (`make lint`, `make typecheck`, `make test`). `make lint` and `make typecheck` were run and are green. The full `make test` was not: the new module and the integration modules around it were run instead, and CI covers the rest.
- [x] Documentation was updated where necessary. No user-facing documentation changes; this is test-only.
- [x] If the API contract changed, I regenerated the OpenAPI spec (`uv run python scripts/generate_openapi.py`). No API surface changed, so no regeneration was owed.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

The claim that the inference routes bypass the suite's database override was checked by observation before any test was written, by counting connection checkouts on the startup engine during a chat completion, rather than taken from reading the code. The first attempt at a gate, checking the pool right after each request, passed on the broken code; that is what led to the in-task measurement the final tests use, and it is written down in the module so the next person does not lose the same hour.

**NOTE:**

## Summary

- Added PostgreSQL integration tests for data-plane connection cleanup.
- Verified connection release during request teardown.
- Verified chat requests leave the real startup pool idle.
- Counted pool checkouts to confirm the tests observe the engine used by data-plane routes.
- Waited for startup refresh activity to finish before measuring pool state.

These tests detect connection leaks before they reach production. They are PostgreSQL-specific because SQLite uses `NullPool`.

## Technical notes

- Test-only change.
- Test results were not provided.


tests/integration/test_data_plane_pool_release.py
```diff
@@ -0,0 +1,135 @@
...
+    """
+    key = client.post(f"{API_ROOT}/keys", json={"key_name": "pool"}, headers=master_key_header)
+    assert key.status_code == 200, key.text
+    headers = {API_KEY_HEADER: f"Bearer {key.json()['key']}"}
+
+    engine = database._engine
+    assert engine is not None
+    pool = engine.pool
+    assert isinstance(pool, AsyncAdaptedQueuePool), "checkedout() needs a queue pool, so PostgreSQL"
+    baseline = pool.checkedout()
```
khaledosman: `baseline` and the closing `checkedout()` both race the lifespan's own work on this engine. `run_selector_index_refresher` (`src/gateway/main.py:482`) does its pass *before* sleeping, and `rebuild_selector_index` → `build_merged_catalog` runs `catalog_scope`/`get_pricing_map` on a `create_session()` from the same `_engine` this test measures. If that prime is still in flight when `baseline` is read, `baseline` is 1 and the tail assert fails; if it starts just after, it fails the other way. The other refreshers (alias/policy/provider/price, 30s; sweeper, 300s) sleep first, so they cannot tick inside one test's lifespan; this one can.

Fix: quiesce before measuring. Poll `pool.checkedout() == 0` against a short deadline, then capture the window and assert `0` at the end; the selector pass is 60s from its next tick by then. The same interleaving inflates `checkouts`, though `>= 8` still holds.

khaledosman (approved):
One finding. Merge order holds as the description says: this is red until #1237 lands.

🤖 Review by [Claude Code](https://claude.com/claude-code) (Claude Opus 5, 1M context)

tests/integration/test_data_plane_pool_release.py
daavoo: Confirmed and fixed in 6d7851fd. `run_selector_index_refresher` is the only one that rebuilds before its `asyncio.sleep`, and it does that on a `create_session()` from the same engine, so a freshly started lifespan can have a checkout in flight either side of `baseline`.

The test now polls the pool to zero against a 10s deadline before opening the window (asserting it got there, so a pool that never quiesces fails loudly rather than being measured), and the tail asserts `0` outright rather than a baseline that might have been read mid-pass. `checkouts >= 8` is unchanged, as you noted it still holds.

Verified: both tests pass on this base, the first one still goes red when the `aclosing` wrapper in `get_db_if_needed` is reverted, and the module is green alongside `test_exact_budget_ledger.py`, `test_messages_endpoint.py` and `test_transaction_safety.py`. `make lint` and `make typecheck` green.

PR #1244, opened 2026-09-16
feat(guardrails): build every stored guardrail at startup and run it in this process

dpoulopoulos:
Stacked on #1211 (`feat-guardrail-credential-store`), which is its base. Review that one first; the diff here is only the eight commits on top.

## Description

#1211 gave an operator a way to write a guardrail down: pick one from the catalog, fill in its arguments, paste its API key, and Otari stores the definition. Nothing read those rows, which its own docstring said plainly.

This reads them. When the gateway starts it builds every stored guardrail and holds it for the life of the process, and a write rebuilds the one it touched. So a guardrail is ready before the first request that names it, rather than being constructed by that request. `POST /api/v1/guardrail-credentials/{name}/test` runs one against text you supply, so a key that was typed wrong is caught where it was typed rather than by a user.

Two things follow from building eagerly.

**The design gets smaller, not just earlier.** A lazy runner needed a cache key over the arguments, an in-flight table, `asyncio.shield`, reference counting and an evict-during-build case, all of it there to make "build on the request that first names this profile" correct under concurrency. Building only at startup and after a write removes every one of those problems: a check is a lookup, and a profile nobody built is simply not available, which the existing `on_unavailable` rule already governs.

**Only guardrails that are an API call run here.** The catalog listed a guardrail whose hosted path it cannot actually reach: SusFactor answers over 0DIN's hosted API, but selecting that path means handing the constructor a live provider object, which is not an argument any-guardrail publishes and not a value a JSON column can hold. What a stored SusFactor definition would build is the local encoder, weights and all. So the rule narrows from "has a hosted backend anywhere in its metadata" to "is a hosted backend", the picker goes from ten entries to nine, and no model weights ever load in a gateway process.

Measured, since it decides whether eager is affordable: building all nine costs about a second in total, and no constructor dials out. 847 ms of that is `any_llm`, whose SDK the gateway already imports elsewhere.

## How to test it locally

Automated, all green locally:

```bash
make lint && make typecheck
uv run pytest tests/unit/test_guardrail_runner.py tests/unit/test_guardrail_catalog.py \
              tests/unit/test_guardrail_credential_service.py tests/unit/test_guardrails_service.py \
              tests/unit/test_tool_settings_endpoint.py tests/unit/test_sdk_endpoint_coverage.py -q
uv run pytest tests/integration -k "guardrail or operator_gate or hybrid_mode_surface" -q
```

Full suites: 3690 unit and 2443 integration pass. Seventeen fail on my machine only (fifteen unit around master key / sqlite defaults, two in `test_config_env_loading.py`); I checked out the base commit and confirmed the same seventeen fail there identically. They are caused by a local `.env` supplying `OTARI_DATABASE_URL` and `OTARI_MASTER_KEY`, which CI does not have. The OSS-edition smoke gate passes too, though it has to be run from a directory without that `.env`.

By hand, with one real vendor key (a Lakera community key is enough) and no guardrails container running:

1. `docker compose up` with no `--profile guardrails`, then sign in.
2. `POST /api/v1/guardrail-credentials` with `name: prompt-injection`, `guardrail_name: lakera_guard`, `create_kwargs: {"api_key": "..."}`. The response masks the key as `***` and it appears in no log line.
3. `POST /api/v1/guardrail-credentials/prompt-injection/test` with an injection-shaped `input_text`: `ok: true`, `valid: false`. Benign text: `valid: true`.
4. Restart. The log reads `Built 1 of 1 stored guardrails`, and step 3 answers with no cold-start pause.
5. `PATCH` the row with `{"enabled": false}`, then check the runner no longer holds it.
6. `GET /api/v1/tool-settings/guardrails/catalog` lists nine guardrails and no `susfactor`; storing a `susfactor` definition is a 400.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Refs #1110, #1108. Both are closed; this is the eager variant of the runner they describe, and the request path is still untouched.

## Checklist

## Things reviewers should know

**The `/test` endpoint registers nothing.** An earlier draft had it install what it built, so that testing a profile also repaired one that failed at startup. That turned out to be a way to put a *disabled* definition in front of live traffic, so it now builds a throwaway and discards it. A profile becomes live through a write, never through a test. The cost is that repairing a failed build means re-saving the row.

**Two bugs a review of this branch caught, both fixed here.** The lock in the runner guarded only the two assignments, which are already atomic between awaits, so it closed no race at all; it now covers the whole build, and a delete that lands mid-pass is reconciled rather than resurrected by the swap. And the rebuild-after-write path never read `row.enabled`, so disabling a guardrail left it running until the next restart. The fix put the rule in one place, `guardrail_loader.apply_stored_guardrail`, which both the startup pass and the write path go through.

**Known gaps, stated rather than fixed.** The startup pass is sequential with a per-build deadline and no whole-pass ceiling; cutting it short would make the profiles it skipped look undefined, and it is a background task nobody awaits. A write rebuilds on the worker that served it, so siblings keep their own until they restart, which is the same cross-worker gap the provider overlay has. And `azure_content_safety` is in the catalog but needs a vendor package the published image does not carry; its build fails with a message naming it.

## AI Usage

- [ ] No AI was used.

The plan was reviewed and approved before implementation, and the finished branch was then reviewed along two axes (repo standards, and fidelity to that plan). The two bugs described above came out of that review and are fixed in the commits they belong to. The race fixes are covered by tests that were each verified to fail against the old behaviour before being kept.

## Summary

- Build stored guardrails during gateway startup and after create/update operations.
- Keep built guardrails available for the worker lifetime.
- Reconcile disabled, deleted, unavailable, and undecryptable guardrails.
- Add the operator-only `POST /api/v1/guardrail-credentials/{name}/test` endpoint.
- Limit stored guardrails to constructible hosted backends and remove `susfactor` from the catalog.
- Document worker-local guardrails, startup behavior, and Azure Content Safety packaging limits.
- Add unit and integration coverage for loading, rebuilding, failures, concurrency, deletion, and endpoint behavior.

## Technical notes

- Startup loading runs in the background and does not block gateway startup.
- Build and evaluation failures preserve existing `on_unavailable` behavior.
- The test endpoint probes a guardrail without registering it for enforcement.
- Each worker maintains its own in-process guardrails.
- Test execution results were not provided.


dpoulopoulos:
Closing with its base #1211, in favor of an organization-scoped design. See #1211 for the reasoning.

The runner is the part most worth keeping, and the replacement ports it: the failure taxonomy, the rule that a vendor exception's message never reaches a log line (it echoes the arguments it was handed, which hold the API key), and the `GuardrailOutput` to `GuardrailResult` mapping.

Three changes it needs, found while planning the replacement:

- Keyed on `(organization_id, profile)` rather than a name alone, and living under `services/tenancy/` so it passes `check_flat_modules` (bad022e4).
- The 30 second refresh becomes a diff rather than a rebuild. This cache holds constructed vendor clients, not decrypted dicts, so rebuilding all of them twice a minute per worker does real I/O for nothing. A `(row.id, row.updated_at)` fingerprint is enough, and the column already has `onupdate`.
- Its own bounded `ThreadPoolExecutor`. `asyncio.to_thread` uses the process-wide default executor, shared with `services/file_extractors.py`, so a guardrail fan-out would stall file extraction as collateral.

PR #1249, opened 2026-09-16
fix(dashboard): offer one way to add an organization member

khaledosman:
## Description

The organization Members page offered two buttons that looked like alternatives but were not: "Add member" put someone on the roster immediately with no email, and "Invite member" emailed them a link they had to follow before their membership went live. Which of the two actually works is decided by the deployment, not by the person clicking: an invitation needs a mail transport, and without one there is no email to send and nothing to accept.

So the page now shows one action, and picks it from what the deployment can do. Where mail is configured it reads "Invite member" and sends the invitation. Where it is not, it reads "Add member" and the membership is active straight away. Each dialog now explains its own case instead of telling you to go use the other button.

Nothing changes on a hosted deployment beyond the second button going away, and a self-hosted deployment with no SMTP keeps the only path that works for it.

## How to test it locally

1. `make dashboard` and run the gateway with no SMTP settings (`SMTP_HOST` and `MAIL_FROM_EMAIL` unset), which is the default.
2. Open the dashboard, go to the organization's Members page as an owner or admin. The header has one button, "Add member". Open it: the address field says nothing is emailed and the membership is active straight away. Add someone and the row appears Active.
3. Restart with `SMTP_HOST` and `MAIL_FROM_EMAIL` set. The same header now reads "Invite member" only, and its dialog says an accept link is emailed. Invite someone and the row appears as invited.

Covered automatically by `pnpm --dir web test` (`OrganizationMembersPage.test.tsx` has a case per deployment for which trigger renders and which is absent, plus the existing add, invite, guard and accept-link cases). `pnpm --dir web run lint` and `pnpm --dir web run typecheck` pass. No backend, route or schema change, so nothing generated is owed.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes mozilla-ai/otari-ai#2148

## Checklist

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

The choice between consolidating the two buttons into one dialog and letting the deployment decide was put to the repository owner; they picked the deployment-decides shape implemented here.

## Summary

The Members page now shows one member-entry action based on mail configuration:

- With mail configured, users see **“Invite member”** and receive an invitation email.
- Without mail configured, users see **“Add member”** and the membership activates immediately.

The dialogs now explain the selected behavior. Tests cover both configurations and existing member-management flows.


khaledosman:
Addressed in 61121151: the nitpick was valid, `mail_sent: true` had no test. Added one asserting the confirmation names the recipient, that neither the failure copy nor the copyable accept link renders, and that the dialog is dismissable on that branch (it is not when the send failed, since the link lives only there). The finding landed in the review summary rather than as an inline thread, so there is nothing to resolve.

PR #634, opened 2026-08-18
feat(pricing/usage): track and persist reasoning/thinking tokens (#430)

Sourav-Tripathy:
## Description
This PR implements end-to-end support for tracking and metering "reasoning tokens" (used by OpenAI's reasoning models and Anthropic's extended thinking models) within the Otari gateway. 
Changes include:
* **Database Persistence:** Added a nullable `reasoning_tokens` integer column to the `usage_logs` table via Alembic migration (`6163c9f0d2ef`).
* **Adapters & Ingestion:** Extended the `GatewayUsage` core carrier and updated the endpoint adapters (`chat.py`, `messages.py`, `responses.py`) to extract reasoning/thinking tokens from OpenAI and Anthropic response usage payloads (handling stream chunks and final responses).
* **OTLP / Ingestion:** Updated `ExternalUsageEvent` and the OTLP telemetry parser (`otlp.py`) to support extraction of reasoning tokens from standard semantic attributes.
* **Pricing Engine Integration:** Registered `reasoning_tokens` under `billing_meters` in the pricing engine (`metered_pricing.py`) with non-additive pricing (ensuring correct output token pricing without double-billing).
* **Validation & Specs:** Regenerated and validated the OpenAPI spec (`openapi.json`) and Postman collections.
* **Tests:** Added a complete unit test suite in `tests/unit/test_usage_reasoning_tokens.py` covering adapters, stream aggregation, telemetry, and pricing.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues
Fixes #430

## Checklist

- [x] I understand the code I am submitting.
- [x] I have added or updated tests that cover my change (`tests/unit`, `tests/integration`).
- [x] I ran the Definition of Done checks locally (`make lint`, `make typecheck`, `make test`).
- [x] Documentation was updated where necessary.
- [x] If the API contract changed, I regenerated the OpenAPI spec (`uv run python scripts/generate_openapi.py`).
- [ ] If this PR adds or changes a persistent table, a control-plane contract the otari.ai platform consumes or serves, a capability the platform also has, or a mode-specific surface, I appended an entry to the M4 reconciliation ledger ([otari-ai#1587](https://github.com/mozilla-ai/otari-ai/issues/1587)). See [.github/instructions/reconciliation-ledger.instructions.md](instructions/reconciliation-ledger.instructions.md).

## AI Usage

- [ ] No AI was used.
[...]

PR #1255, opened 2026-09-16
refactor(models): split entities.py into one model module per domain

### LINKED ISSUE #1192: Models into domain modules
### Problem

`models/entities.py` (1,974 lines) and `models/tenancy.py` (1,583) hold every table.

### Proposal

One model module per domain, imported by a single `models/__init__.py` list so `Base.metadata` is complete before Alembic or `create_all` runs. Tables, columns, constraints and names unchanged.

### Acceptance

- [ ] `alembic check` reports no drift.
- [ ] `alembic heads` is one head.
- [ ] The full unit suite passes.
- [ ] No migration file is added.

### Measure

`entities.py` no longer appears among the most frequently changed files over 90 days.

### Depends on

Nothing.

Part of #1171.

peteski22:
## Description

Every table the gateway stores lived in one 2,016-line models file, so nearly every feature that touched the database edited it: 52 commits from 6 authors in the last 90 days. This moves each table into the model module of the domain it belongs to (budgets, pricing, tools, and so on), and moves the shared base types into a module of their own.

Nothing changes for someone running Otari. The database schema, the API, and the dashboard behave exactly as before, and no migration is added. The only wire-visible difference is three schema descriptions in the OpenAPI spec that named the old module.

**Open PRs that add or edit a table will conflict with this.** The update is mechanical; see "Updating an open PR" below.

## Updating an open PR

- Put a table in its domain's module, `src/gateway/models/<domain>.py`. `Base`, `UtcDateTime` and the SQLModel mixins come from `gateway.models.base`.
- A new model module that declares tables joins the import list in `src/gateway/models/__init__.py`. Without it, Alembic does not see its tables.
- `from gateway.models.entities import X` becomes `from gateway.models.<domain> import X`. The billing `User` (table `users`) is now `gateway.models.users.User`. The dashboard sign-in `User` (table `user`) stays in `gateway.models.tenancy`.
- Open PRs that touch these files today: #1211, #1137, #1117 and #634 (tables), #1106 (tenancy), #1185 (the package import list).

**Downstream: otari-ai.** The enterprise overlay imports the old module in its runtime and tests, and three of its migration drift guards read the old file. A matching change is prepared and has to land with the otari pin bump to the first release that contains this PR. Nothing breaks before that bump, because the pin does not move on its own.

## Measures

| Measure | Before | After |
| --- | --- | --- |
| `entities.py` | 2,016 lines; 52 commits, 6 authors in 90 days | removed |
| Largest model module | `entities.py`, 2,016 lines | `tenancy.py`, 1,588 lines |
| Model modules that declare tables | 4 | 14 |

The issue's own measure (`entities.py` leaves the 90-day churn list) can only be read once 90 days have passed.

## How to test it locally

```bash
make lint && make typecheck && make test
uv run alembic heads    # one head
```

To confirm the schema did not move, migrate a PostgreSQL database to head and run `uv run alembic check` on `main` and on this branch. Both report the same single item, a duplicate unique constraint on `api_keys.key_hash` that already exists on `main`, and nothing else. That item is why the issue's "no drift" criterion is not literally met; fixing it needs a migration, which this PR must not add.

What was checked:

- **Every commit on its own**, not only the tip: `make lint`, `make typecheck`, `make openapi-check`, `make postman-check`, and a clean import of every model module and app entry point in a fresh interpreter.
- **Schema identical at every commit**: the generated `CREATE TABLE` and `CREATE INDEX` statements for all 48 tables on SQLite and PostgreSQL, plus every mapper, column attribute and relationship, match `main` byte for byte.
- **Moved code is verbatim**: all 137 top-level definitions in the two old files still exist with the same code. The only text changes are in comments and docstrings: a few pointed "above", "below" or at "neighbors" that now live in another module, and now state their own reason; and links to moved classes name their new module.
- **Suites on the tip**: unit 3,698 passed; integration 2,416 passed; the OSS smoke gate passes; the committed dashboard client matches a fresh generation. After the last rebase onto #1014, which changed no commit in this branch, lint, typecheck, the artifact checks, the schema comparison and the 226 integration tests for the areas #1014 touched passed again. One integration test, `test_mcp_dependency_ceiling`, fails locally because this machine's Python cannot build the throwaway virtualenv the test creates. It fails the same way on `main`.

## PR Type

- [ ] New Feature
- [ ] Bug Fix
- [x] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes #1192

Part of #1171

## Checklist

No tests were added: behavior does not change, and the existing suites cover it. Test files changed only their imports. The API contract did not change either; the spec, Postman collection and dashboard client were regenerated because three schema descriptions named the old module.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

An AI coding agent, run by the author.

The author made the layout decisions: one module per domain; a separate module for the billing `User` rather than renaming either `User`; new domain modules beside the existing ones whose names do not match a domain (`provider_keys`, `playground`, `mcp`), which keep their names for now; deleting the old module rather than keeping a re-export shim; and a shared base module. The agent did the moves and import rewrites, wrote the schema comparison run at every commit, and drafted this description.

**NOTE:**


tbille (approved):
Verified independently on the PR head: the full SQLAlchemy metadata (all 48 tables: columns, types, defaults, FKs + ondelete, constraints, indexes, mappers) is byte-identical to main; all 138 top-level definitions from the old modules exist with AST-identical code; every changed line outside models/ and docs is an import rewrite; the model-module dependency graph is acyclic (table modules depend only on base, money, secret_fields); lint, architecture check, and single-head check pass locally.

One leftover the sweep missed, in a file this PR does not touch so it cannot carry an inline comment: alembic/versions/a7d3f5b1c9e4_add_organization_model_pricing.py line 11 still points readers at gateway.models.entities.OrganizationModelPricing, which this PR deletes. Should read gateway.models.pricing.OrganizationModelPricing. Fine as a follow-up commit here or a separate one-line PR.

Reviewed by Claude Code (Fable 5), run by the reviewer.

Issue #1245, opened 2026-09-16
Add the dashboard page that defines a guardrail Otari runs itself


dpoulopoulos:
## Problem

#1211 added the store (`/api/v1/guardrail-credentials`) and #1244 made the gateway build every
stored row at startup. Both landed without a way to reach them from a browser: the only
interface today is `curl`.

The guardrails screen (`/tools/guardrails`) currently offers a `guardrails_url` field for the
old sidecar and the organization mandate card that names a `profile`. There is no way to
*create* the profile a mandate names, so the two halves do not meet.

## What to build

Change the current guardrails screen. Add an operator-only card above the mandate card, with a
dialog that asks in the order an operator actually decides:

1. **What do you want checked?** A dropdown of operations, for example prompt injection.
2. **Which guardrail?** A second dropdown, disabled until the first is answered, listing the
   guardrails that do that job.
3. **Its own fields.** The chosen guardrail's `create_parameters` and `validate_parameters`,
   rendered typed, with credentials masked.

On submit the definition is stored, and the card lists what is in the database.

Supported actions: create, list, edit, delete, enable/disable, and test against a sample input.

## The taxonomy is any-guardrail's, not ours

No category list, ordering, or copy is written down in this repository.

- **Which operations exist** is the union of `categories` over the catalog entries.
- **Which guardrails do one** is `categories` membership, not `primary_category`.
  `AnyGuardrail.group_by("category")` is upstream's own grouping and its docstring states that
  for a set-valued dimension "a guardrail appears under every value it carries".
  `primary_category` exists to place a guardrail exactly once in grouped docs navigation, which
  is not what a picker does. Using it would hide Alinia, Patronus and watsonx Guardian from
  prompt injection although all three detect it, and four of the nine operations would never
  appear at all.
- **Ordering** follows upstream: `group_by` returns its keys sorted.
- **The label** is the existing mechanical helper `parameterLabel`
  (`prompt_injection` becomes `Prompt injection`). Formatting, not authored copy.

One gap worth noting: `GuardrailCategory` carries a one-line docstring per member, but Python
drops member docstrings at import, so the gateway cannot read them and no help line is shown
under the first dropdown. A follow-up asks any-guardrail to publish runtime-readable category
descriptions.

## Scope

Frontend only. `BuiltInGuardrailSpec` already carries `categories`, and `web/src/client/schema.ts`
is already regenerated on `feat-eager-guardrail-runner`, so there is no backend change, no
[...]

PR #1256, opened 2026-09-16
feat(dashboard): add the page that defines a guardrail otari runs itself

dpoulopoulos:
> **Stacked.** This targets `feat-eager-guardrail-runner` (#1244), not `main`, because it
> consumes the store #1211 added and the runner #1244 builds. Review those first; the diff
> here is the seventeen commits on top. Two consequences: CodeRabbit skips a PR whose base is
> not `main`, so it needs a `@coderabbitai review` comment, and `protect-main` does not apply
> to a child branch, so this reports mergeable with no approval and must not land before its
> parents.

## Description

#1211 gave Otari a place to store a guardrail definition and #1244 taught it to build one at
startup. Neither could be reached from a browser: the only interface was `curl`, and the
guardrails screen still offered a URL field for a separate guardrails service and the
organization mandates over it.

**Tools → Guardrails is now a page of its own**, shaped like the providers page, because it is
the same kind of thing: a list of credentialed entries an operator adds, edits, tests and
removes. A row shows the profile a request names, the guardrail behind it and who publishes
it, every check it performs, and whether it is running.

Adding one asks in the order the decision is actually made. **What do you want checked**, in
plain words. Then **which guardrail** does that job, each named with its vendor. Then **its
own fields**, typed from the catalog, with the detection you asked for already switched on.

The old screen's two other halves are gone rather than moved. `guardrails_url` configures a
service this page no longer talks about and stays a config-file setting; organization mandates
were removed so nothing here is about a second service.

### Nothing about the taxonomy is written down

The list of operations, which guardrails do each one, and their order are read from
`GET /tool-settings/guardrails/catalog`, which publishes any-guardrail's own metadata. A
category upstream adds appears here with no change.

Two decisions the derivation does not make alone, both settled upstream. It reads `categories`
rather than `primary_category`, because `AnyGuardrail.group_by("category")` documents that a
guardrail appears under every value it carries; reading the headline would hide Alinia,
Patronus and watsonx from prompt injection although all three detect it. And it sorts, which is
what `group_by` returns.

A guardrail is offered for a job only when it can be **set up** for it. Alinia and Patronus
both declare personal data and nothing readable names the detection that does it, so they are
not offered there. watsonx Guardian is, although its metadata lists no `pii` category, because
it documents a `pii` detector: a key that is documented is better evidence than a category that
is missing.

### The JSON arguments

Eleven of a guardrail's arguments are typed `json`, and each was a textarea to fill with
braces. Alinia cannot run without one, so defining Alinia meant knowing its detection
vocabulary by heart.

Each now gets controls with a **Fields / JSON** switch: named switches, named presets, key and
value rows, or a list of text. Upstream's own `ParameterType.JSON` docstring calls it "the not
flat-form-able, use a JSON editor signal", so nothing published says whether a field is a list
of sentences or a map of switches, let alone what its keys are. That vocabulary is therefore
written down in `guardrailFieldSuggestions.ts`, read from any-guardrail's constructor
docstrings and reference pages, with anything only implied left out. It is the same trade
`providerCredentialFields.ts` already makes for `client_args`.

Three things keep a list that will go stale from costing anything. It is a suggestion and never
a whitelist, so a key this build has never heard of renders under its own name and a new one
can be typed in. The JSON view accepts anything. And a value the controls cannot hold, a
detection config given as a registered id, is refused by the Fields view rather than rewritten.

### Only what must be decided

A form is four tests away from asking twelve questions when there are two. A field is shown
when it has no default, **or** is a credential, **or** is named by a one-of requirement group,
**or** is what the guardrail actually checks. `required` alone answers none of this: nothing on
watsonx Guardian is required, because its credentials sit in requirement groups. Everything
else, the per-call arguments and the escape hatch included, is behind one **Advanced settings**
accordion at the foot of the dialog, which opens itself if a field it hides has a message.

### One fix outside the dashboard

Testing a stored Azure Content Safety definition failed with "azure-ai-contentsafety package is
not installed": the catalog offered a guardrail this image could not build. `pyproject.toml`
now names the four extras for the guardrails that speak over a vendor SDK. Only
`azure-content-safety` installs anything, two pure-Python packages, and the locked set goes
from 234 to 236 with nothing removed. The other three are named because `boto3` and
`ibm-watsonx-ai` were reaching Bedrock and watsonx through `any-llm-sdk[all]`, which is a
coincidence rather than a dependency: any-llm narrowing that extra would break both at runtime
with the same ImportError and nothing would warn first.

## How to test it locally

Needs `OTARI_SECRET_KEY` set and one real vendor key. A Lakera community key is enough.

```bash
make dashboard && uv run otari serve
```

Sign in as the deployment operator and open **Tools → Guardrails**.

1. **Add guardrail.** The second control is disabled. Pick **Prompt injection**; it unlocks and
   offers five. Three of those five have a *different* headline category, which is the
   `categories` derivation working.
2. Pick **Lakera Guard**. Its fields appear: `api_key` masked, `endpoint` showing its default as
   a placeholder and never prefilled. Everything else is under **Advanced settings**. Name it
   `prompt-injection` and save.
3. **Test** the row with `ignore your previous instructions`. Expect `ok: true, valid: false`.
   Benign text gives `valid: true`.
4. **Edit** it. The key box is empty and says it is already set. Change *only* the endpoint and
   save. **The key must survive**: a PATCH replaces the whole map, so this is the case that
   would silently delete it.
5. Add an **Alinia** one under prompt injection. **Security** is already ticked and no JSON was
   typed. Switch to **JSON** to see `{"security": true}`.
6. Pick **Personally Identifiable Information**, then Alinia: it is not offered, because nothing
   readable names its PII detection. **watsonx Guardian** is, and picking it ticks `pii`.
7. Add an **Azure Content Safety** one and test it. Before this branch that failed on a missing
   package.
8. Toggle a row off, then remove it. Sign in as a non-operator: the page explains itself and
   fires no 403.

Automated, all green from this worktree:

```bash
pnpm --dir web run lint && pnpm --dir web run typecheck
pnpm --dir web test && pnpm --dir web run build
make lint && make typecheck
uv run pytest tests/unit -k guardrail -q
```

6748 dashboard tests across 206 files, of which about 120 are new. Backend: ruff, the
architecture check, mypy over 647 files, and 114 guardrail tests. No generated artifact moved:
the API already existed, so the OpenAPI spec, the Postman collection and the client schema are
untouched, and `routeTree.gen.ts` comes back unchanged.

## PR Type

- [x] New Feature
- [x] Bug Fix
- [x] Refactor
- [x] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Closes #1245. Builds on #1211 and #1244. Supersedes #1114, whose draft PR #1137 targeted the
superseded #1116/#1120 stack and should be closed with it.

## Checklist

Two notes on that list. The change is almost entirely in `web/`, so its tests are the dashboard
suites rather than `tests/unit` and `tests/integration`; the one backend commit is a dependency
bump with no code behind it. And the API contract did not change: every endpoint consumed here
arrived with #1211 and #1244.

## Things reviewers should know

**Three deletions worth a look.** `OrganizationGuardrailsCard` and `GuardrailProfileField` are
removed rather than moved, so organization mandates have no UI on this branch.
`GuardrailParametersSection` lost its last caller when the accordion absorbed it. And
`useGuardrailProfiles` with three helpers went with them, so the dashboard no longer reads the
sidecar's profile catalog at all. The endpoints are untouched; only the unused clients are.

**AnyLlm is not selectable.** The last commit drops `general_judge` from the list of jobs on
the grounds that it is not a subject to detect, and AnyLlm declares nothing else. Deliberate,
stated in that commit, and a one-line revert.

**The OSS smoke gate fails locally**, at `otari migrate`, with an empty error message. I
checked out the base and it fails identically, so it is not from this branch, but CI runs it on
any dependency change and this branch has one. Worth knowing before reading that job's result.

**Two follow-ups, unfiled.** Feeding these profile names into a mandate picker, and asking
any-guardrail to publish enumerated keys for a `json` parameter, which would shrink the written
vocabulary and close the Alinia PII gap properly.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

Claude Opus 5 (1M context), through Claude Code.

The plan was reviewed and approved before implementation. Several decisions were the author's
rather than the model's: building fresh instead of reviving #1137, removing the sidecar and
mandate cards from this screen, reshaping it after the providers page, accepting a written
vocabulary for the JSON fields, and dropping the general-judge group.

Three things the model got wrong and had to be told: splitting required from advanced on the
`required` flag alone, which put watsonx Guardian's whole configuration behind the accordion;
passing Python docstring markup into the UI so the form rendered stray double backticks; and
offering a guardrail for a job it could not be configured for, which the reviewer found by
selecting personal data and reaching a dead end.

**NOTE:**

## Summary

- Added a dedicated `/tools/guardrails` dashboard for creating, listing, editing, testing, enabling, disabling, and deleting stored guardrails.
- Derived available operations and compatible guardrails from catalog metadata.
- Added typed parameter forms, JSON editing, credential masking, and secret preservation during edits.
- Added runtime status, publisher details, documentation links, and encryption checks.
- Removed the legacy sidecar and organization-mandate UI from this page. `guardrails_url` remains configuration-only.
- Added supporting API hooks, tests, documentation, and vendor SDK dependencies.

## Technical notes

The change is frontend-focused. It uses the existing guardrail credential APIs and catalog data without adding migrations or API schema changes.


dpoulopoulos:
@coderabbitai review

Stacked on `feat-eager-guardrail-runner` (#1244) rather than `main`, so this needs asking explicitly. Only the seventeen commits on top of that branch are this PR's.

dpoulopoulos:
Closing with its bases #1244 and #1211, in favor of an organization-scoped design. See #1211 for the reasoning.

The replacement targets the organization surface rather than the operator one, and lives at `/organization/guardrails`, where a placeholder route and a nav row already wait on a surface no build declares yet.

What carries over: the picker taxonomy, which is the most valuable part of this PR. The operation list is the union of each spec's `categories` and not `primary_category`, the ordering follows upstream's sorted `group_by` keys, and the labels come from the existing mechanical `parameterLabel`. The new work ports `guardrailOperations.ts` from here.

PR #1261, opened 2026-09-16
fix(catalog): list hosted providers for members and flag their models as deployment-supplied

### LINKED ISSUE #1260: Members do not see models served on a hosted (deployment-owned) provider credential
## What happens

On a hosted deployment, a provider the deployment serves on its own credential (an overlay-bound `ModelProviderPort` adapter, such as otari.ai's hosted backends) is listed on the Models page for deployment operators only. An organization member or owner signed in to the dashboard does not see any of its models, although a request for one of them is served.

## Why

`resolve_session_catalog_scope` (`services/tenancy/organization_model_access.py`) builds a member session's allow-list from two sources: the `config.providers` instances and the organization's own BYO keys. The hosted rung the dispatch path consults (`ModelProviderPort.resolve_hosted_credential`) is not one of them, so a bare `provider:model` selector served on the deployment's credential is filtered out. An operator session bypasses the filter entirely, which is why operators see the models.

A second, related mismatch: `deployment_managed` on a catalog entry is derived from `config.providers` alone. The pricing write path already refuses an organization override for a model a hosted credential serves (`OrganizationPricingService.raise_if_deployment_supplied`), but the catalog does not flag those models, so the dashboard offers an owner a rate-override control that the gateway then answers with a 403.

## Expected

- Every member of an organization sees the models of the providers the deployment serves them on its own credential, alongside the configured instances and their organization's BYO providers.
- Those models carry `deployment_managed: true` for that viewer, so the dashboard withholds the organization rate override the way it does for a configured instance. A provider the organization also holds a BYO key for stays the organization's to price, because dispatch uses the BYO key.
- Operators and master-key callers are unchanged.

Related: mozilla-ai/otari-ai#1969 introduced the member scoping; mozilla-ai/otari-ai#1942 is the roles matrix it implements.

tbille:
## Description

On a hosted deployment, the deployment serves some providers on its own credential. Before this change, only a deployment operator saw those models on the Models page. A member or owner saw none of them, even though a request for one is served.

The catalog now asks the bound `ModelProviderPort` which providers it hosts for the caller's organization, through the new `get_hosted_providers` method.

- A member sees a hosted provider's models where one of their workspaces has no active BYO key for that provider. A workspace's key restriction still applies where it has a key, because dispatch uses that key.
- A model is flagged `deployment_managed` when the deployment pays for it in at least one workspace. The flag uses the same rule as the organization rate-override gate (`OrgProviderKeyService.get_byo_providers`), so the dashboard offers an override only where the gateway accepts it.
- The grouped catalog labels such an offering `hosted` (`CatalogCredential.HOSTED`), not `organization`.
- An organization alias or policy may target a hosted model.
- An adapter failure fails the catalog read, the same as on the dispatch and pricing paths.

Operators, master-key callers and the core build are unchanged. The core adapter hosts no provider.

**Known gap:** an API key's listing still reads `deployment_managed: false` for a hosted model. Its organization is known and the rate-override gate applies to it. #1267 tracks the fix.

**Follow-up in otari.ai:** `HostedBackendModelProviderAdapter` must implement `get_hosted_providers` (the enabled `hosted_backend` rows whose key decrypts) before the submodule pin moves onto a release with this change. Without it, a member's catalog read fails.

## How to test it locally

1. `make dev`, sign in as the deployment operator, and store a price for `mistral:mistral-small-latest` through `POST /api/v1/pricing`.
2. Bind a hosted adapter through `OTARI_BOOTSTRAP` whose `get_hosted_providers` returns `{"mistral"}` and whose `resolve_hosted_credential` returns a credential for `mistral`.
3. Sign in as a member of an organization with no mistral key. `GET /api/v1/models` lists the model with `deployment_managed: true`, and the Models page labels it hosted with no "Set your rate" link.
4. Give the organization a mistral key used by every workspace. The model reads `deployment_managed: false`, and the owner may price it. Disable the key in one workspace, and the flag and the rate write both change back.

Automated coverage:

- `tests/integration/test_session_model_catalog_scope.py`: the flag agrees with the rate write (parametrized), hosted models are listed only where dispatch would ask the port (parametrized), the grouped catalog label, and an adapter failure.
- `tests/integration/test_org_provider_keys.py`: `get_byo_providers` edge cases (parametrized).
- `tests/integration/test_organization_routing_write_scope.py`: alias and policy targets only the port serves (parametrized).

Run with `make lint`, `make typecheck`, `make test`.

Some local failures also fail the same way on `main` in the same shell:

- 16 master-key, hook and Claude Code import unit tests, plus `test_load_config_treats_empty_otari_scalar_env_as_unset`. They read a repo-root `.env` or `OTARI_API_KEY`, and they pass when run outside the repo folder with `OTARI_API_KEY` unset.
- The OSS smoke gate. The repo-root `.env` binds the gateway to port 8000, not the port the gate waits on. It was not re-run without that file.
- `test_mcp_constraint_resolves_to_an_importable_version`, whose temporary Python aborts locally.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes #1260

## Checklist

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:** Claude Code (Claude Fable 5.1 for the first commit, Claude Opus 5 for the commits after it)

**Any additional AI details you'd like to share:** An agent drafted the first commit from the bug report. A human review then found the gaps in it: the flag disagreed with the pricing gate, the listing widened a restricted key, the grouped catalog label was wrong, and adapter failures were hidden. The review also set the design of the fix: a per-provider port answer, one BYO rule on `OrgProviderKeyService`, the names, and failures that propagate. An agent wrote the later commits under that review, test first, and ran the checks locally.

## Summary

- Organization members and owners can see models served by deployment-owned hosted providers.
- Hosted models are marked `deployment_managed` unless the organization has a usable BYO key.
- Organizations retain pricing control when their BYO key serves the model.
- Operators and master-key callers keep their existing access.
- Catalog, routing validation, and pricing checks now use consistent provider information.
- Added coverage for hosted model visibility, BYO precedence, access rules, and provider failures.

## Technical notes

- Added `get_hosted_providers` to `ModelProviderPort`.
- Added the `hosted` catalog credential type.
- Updated catalog scope resolution and organization key handling.
- Updated routing and catalog endpoints to pass provider information.


peteski22 (approved):

src/gateway/services/tenancy/organization_model_access.py
```diff
@@ -147,43 +132,56 @@ async def resolve_session_catalog_scope(
...
+            organization_id=scope.organization.id, workspace_ids=scope.workspace_ids or []
         )
+        byo_allowlist = await _get_byo_allowlist(db, active_keys)
+        # Dispatch asks the port only in a workspace with no active key that has a credential.
+        reachable_hosted = frozenset(
+            provider
+            for provider in hosted
+            if any((key := keys.get(provider)) is None or not has_credential(key) for keys in active_keys.values())
+        )
+    byo_providers = (
```
khaledosman, line 176: Once the port serves anything, `candidates_for_workspaces` runs twice per catalog read: line 166 for the caller's workspaces, then again inside `get_byo_providers` for the whole organization. The org-wide result is a superset of the caller's, so one call at the top of the branch, sliced for `scope.workspace_ids`, serves both. The owner branch pays `list_live_keys` twice the same way (159-163, then inside `get_byo_providers`).

`/api/v1/models` is a dashboard read, so this is a query per navigation rather than per deploy.

src/gateway/api/routes/catalog.py
```diff
@@ -478,16 +484,17 @@ async def _with_usage(db: AsyncSession, grouped: _Grouped, members: list[_Offeri
...
+    ``config`` refuses the reserved hosted instance name in ``providers:``,
+    so an offering carrying it came from an overlay.
     """
     if instance == HOSTED_OFFERING_INSTANCE:
-        return "hosted"
-    return "deployment" if instance in config.providers else "organization"
+        return CatalogCredential.HOSTED
+    if instance in config.providers:
+        return CatalogCredential.DEPLOYMENT
+    return CatalogCredential.HOSTED if deployment_managed else CatalogCredential.ORGANIZATION
```
khaledosman, line 497: `deployment_managed` is org-wide, so this makes `credential` stop answering the question `CatalogOffering.credential` documents at lines 172-175 ("Whose key serves it").

Concretely: organization holds an openai key, one workspace disables it, the port serves openai. That provider drops out of `byo_providers`, so `deployment_supplied_providers` contains it, and a member of a workspace that *does* hold an active key for it reads `credential: "hosted"` on a model their own organization's key dispatches.

The actions column is right either way (`web/src/features/models/ModelDetailPage.tsx:326` shows "Deployment priced", and the rate gate would refuse), so the fix is the description rather than the derivation: say the value follows who may price the offering, not who dispatches it, and regenerate the spec and the client.

src/gateway/api/routes/catalog.py
```diff
@@ -109,10 +112,12 @@
 # it as "nobody" rather than as a key that failed to verify.
 CatalogCaller = tuple[APIKey | None, bool] | None
 
-# The reserved instance a hosted edition serves deployment-owned offerings under;
-# the base gateway never configures one, so the label only ever appears where an
-# overlay contributes such an offering.
-Credential = Literal["deployment", "organization", "hosted"]
+class CatalogCredential(StrEnum):
```
khaledosman, line 115: One blank line before a top-level class; the rest of the file uses two (see line 126). Same class of nit at `src/gateway/services/tenancy/org_provider_key_service.py:840-841`, which has two blank lines inside the class body before `get_active_keys`.

Neither fails CI: ruff's `check` config here does not select E3, and the repo does not run `ruff format`.

khaledosman (approved):
Gates verified on the head commit in a clean worktree: `make lint`, `make typecheck`, `generate_openapi.py --check`, `generate_postman.py --check` (the `credential` description change stays inside a schema, so the collection is genuinely untouched), and the regenerated `web/src/client/schema.ts`.

PR #1281, opened 2026-09-17
docs(catalog): describe an offering's credential as who may price it

peteski22:
## Description
The model catalog gives each way to reach a model a `credential` value. The API description said this value names whose key serves the model. That is not always true on a hosted deployment.

Example: an organization holds an OpenAI key, one workspace turns that key off, and the deployment also serves OpenAI. The deployment now pays for OpenAI in that workspace, so the organization may not set its own OpenAI rate. Every member then sees `hosted`, even a member whose own workspace still uses the organization's key.

The value follows who may set the price, and that is what the dashboard uses it for. This PR changes the description to say so. No behavior changes. It also fixes blank-line spacing that came in with #1261.

Both points came from the post-merge review on #1261.

Not in this PR:
- The dashboard's Provider column still shows `hosted` for the member in the example.
- The review's note about repeated key queries on a catalog read. That is a separate performance change.

## How to test it locally
Nothing changes at runtime, so there is nothing to click through.

- `make openapi-check` and `make postman-check` pass. The Postman collection does not change, because the edit is inside a schema description.
- The dashboard client is regenerated with `pnpm --dir web run client:generate`, and CI's drift check covers it.
- `make lint` and `make typecheck` pass.
- `tests/integration/test_catalog_endpoint.py` and `tests/integration/test_session_model_catalog_scope.py` pass unchanged (41 tests).

## PR Type

- [ ] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [x] Documentation
- [ ] Infrastructure / CI

## Relevant issues
Follow-up to the review on #1261.

## Checklist

No tests were added, because only descriptions and whitespace change. `make lint` and `make typecheck` ran locally. The full `make test` did not, only the catalog integration tests above.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**
Claude Code (Claude Opus 5)

A teammate's review found the issues. A human decided which findings to act on and the scope of this PR. Claude Code checked the findings against `main`, drafted the change, and ran the checks.

**NOTE:**


khaledosman (approved):
No findings.

Gates verified on the head commit in a clean worktree: `generate_openapi.py --check` and `generate_postman.py --check` both pass, so the description's claim that a schema-level description does not reach the collection is right (`Whose key serves` appears zero times in `otari.postman_collection.json`).

The new wording is accurate against the code, which is the part worth checking on a description-only change:

- `get_byo_providers` (`org_provider_key_service.py:867`) returns only providers that **every** workspace calls with its own key, so `deployment_supplied_providers = hosted - byo_providers` in `organization_model_access.py:184` is exactly "the deployment pays in at least one workspace". That is what the new `hosted` sentence says, and it is what makes the example in the description happen.
- "Who may price" matches the consumer: `ModelDetailPage.tsx:326` offers "Set your rate" only on `credential === "organization"`, and the comment above it records that the gateway refuses an override on a deployment instance (otari#1164). So all three values line up on pricing authority, where only `organization` was true under the old "whose key serves it" reading.

The red `test-integration` is `test_hooks_route.py::test_many_separator_only_commands_resolve_quickly` failing its wall-clock assertion (`< 0.5`, got ~0.57s), the same flake as on #1285 and the subject of 8a5fded2. Not this PR.

PR #1283, opened 2026-09-17
chore(lint): refuse a new top-level package under src/

### LINKED ISSUE #1273: Refuse a new top-level package under src/
### Problem

`scripts/check_architecture.py` checks modules under `src/gateway/`. A new package beside it, such as `src/otari_<name>/`, is outside every rule. It can import any core internal. Wired through `OTARI_BOOTSTRAP`, it becomes a second way to add a feature to the core, beside the registry in `src/gateway/features.py`.

A feature that lives in this repository belongs under `src/gateway/` and in the registry.

### Proposal

The check fails when `src/` holds a top-level package that is not on an explicit list in the script. The list starts as `gateway`. Adding a name is a reviewed change to the check, and CODEOWNERS owns the check.

### Acceptance

- [ ] `make lint` fails on a synthetic `src/otari_probe/__init__.py`.
- [ ] The allowed list holds `gateway` only.
- [ ] `tests/unit/test_check_architecture.py` covers an allowed and a refused package.

### Depends on

Nothing.

Part of #1171.

peteski22:
## Description
The architecture check keeps the backend's layers apart, but it only looks inside the main gateway package. A new package placed next to it is outside every rule. It can reach into any part of the core, and loaded through the edition hook it becomes a second, unreviewed way to add a feature.

This makes the check fail when the source directory holds any importable package other than the gateway itself. A feature that lives in this repository belongs inside the gateway package and in its feature registry. Allowing another package is now a visible, reviewed change to the check.

### Architecture rule
This adds a rule to `scripts/check_architecture.py`: `src/` holds no top-level package but `gateway`. Why: a package beside the gateway answers to none of the layer rules, and it would be a second way to add a feature to the core.

## How to test it locally
2. Create `src/otari_probe/__init__.py` (empty) and run `make check-architecture`. It fails with "Top-level package src/otari_probe is not allowed". Delete the file and it passes again.
3. `uv run pytest tests/unit/test_check_architecture.py` covers the allowed package, a refused package, a refused namespace package, a refused top-level module, a directory with no Python source (the editable install's `gateway.egg-info`), and `main()` failing on a new package.

Also run: `make typecheck` passes. `make test-unit` passes apart from four tests in `test_hook_cli.py` and `test_claude_code_import.py` that fail only when `OTARI_API_KEY` is set in the shell; they pass with it unset and do not touch this code.

## PR Type
- [ ] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [x] Infrastructure / CI

## Relevant issues
Fixes #1273

Part of #1171.

## Checklist

## AI Usage
- [ ] No AI was used.

**AI Model/Tool used:**
Claude Opus 5 via Claude Code

**NOTE:**

## Summary

- Added an architecture check that allows only `gateway` as a top-level package under `src/`.
- Updated architecture documentation.
- Added tests for allowed packages, rejected packages and modules, ignored directories, and failure reporting.

This prevents features from bypassing the gateway package and feature registry.


scripts/check_architecture.py
```diff
@@ -14,6 +14,8 @@
    route may not import it; and nothing under gateway/ imports
    importlib.metadata, importlib_metadata or pkg_resources, so nothing is
    discovered.
+10. Top-level packages: src/ holds only the packages on an explicit list, so a
```
khaledosman, line 17: Collides with #1286, which numbers its new rule `10.` here too, appends to the same AGENTS.md sentence, and adds its own branch to the same `if` in `main()`. Whichever merges second needs `11.` and a rebase; worth deciding the order now so the renumber is one edit rather than a conflict resolved twice.

khaledosman (approved):
No findings on the rule itself. One coordination note inline.

Worth recording, because it is not obvious from the diff: `pyproject.toml:173` is already `include = ["gateway*"]`, so an arbitrarily-named sibling under `src/` never reaches a wheel. The case this check actually closes is `src/gateway_extras/`, which matches that glob and today would ship while answering to none of the layer rules. The check is narrower than the description suggests and still earns its place.

peteski22:
@coderabbitai review

peteski22:
@coderabbitai review

scripts/check_architecture.py
peteski22: Merge order: #1283, then #1286, then #1292. Each later PR rebases after the one before it merges and renumbers its rule in the docstring list (#1286 to 11, #1292 to 12), resolving the AGENTS.md sentence and the `if` in `main()` in the same rebase.

Issue #1198, opened 2026-09-15
Move search behind the seam, delete the old dispatch


peteski22:
### Problem

`SearchTool` in `services/search_backend.py` is the third built-in tool. Once it moves, the old per-tool dispatch paths have no caller and can go.

### Proposal

Dispatch search through the registry; delete the per-tool branches and the `Tool` string enum's role in dispatch (keep the enum if the wire schema still needs the names). The dialect loops then contain one dispatch path.

### Acceptance

- [ ] The dialect loops contain no tool-specific branches.
- [ ] Every existing search test passes.
- [ ] `_pipeline.py`'s importing-module count and public-function count are recorded before and after.

### Depends on

#1196, #1197.

Part of #1171.

Issue #1197, opened 2026-09-15
Move sandbox (code execution) behind the seam


peteski22:
### Problem

`SandboxBackend` in `services/sandbox_backend.py` is the second of the three built-in tools hand-wired into the pipeline and the three dialect loops. After 3.2 it is the last tool still dispatched by name alongside the registry.

### Proposal

The pipeline and the three loops dispatch the sandbox tool through the registry from 3.1; delete the hand-wiring for it. The sandbox's own wire contract to the execution backend is unchanged; only the gateway-side dispatch moves.

### Acceptance

- [ ] Every existing sandbox and code-execution test passes.
- [ ] No `sandbox`-specific branch remains in `mcp_loop.py`, `mcp_loop_messages.py` or `mcp_loop_responses.py`.
- [ ] The backend mention count for `sandbox` in `src/gateway` is recorded in the PR.

### Depends on

#1195.

Part of #1171.

peteski22:
#1366 added code-execution branches inside the dialect loops, which this issue removes: `emit_native_code_execution` flags in `services/mcp_loop_messages.py` and `services/mcp_loop_responses.py`, and a `CODE_EXECUTION_TOOL_NAME` branch in the Messages loop that mints native `server_tool_use` and `code_execution_tool_result` blocks. Counted with `grep -ci 'sandbox\|code_execution\|code_interpreter'`, before #1366 (`d311450c`) and after (`9c8acdec`): `mcp_loop_messages.py` 2 to 24, `mcp_loop_responses.py` 4 to 39. They are in this issue's scope now. Tracked in #1470.

Issue #1196, opened 2026-09-15
Move web search behind the seam


peteski22:
### Problem

`WebRetrievalBackend` in `services/web_retrieval_backend.py` is dispatched by name in the pipeline and the dialect loops, beside the registry entry `_web_search_tool.TOOL` that #1195 added. The registry landed without deleting the hand-wiring, so both paths exist. `services/web_search_backend.py` is a compatibility shim re-exporting the same names, imported by `services/web_search_budget.py` and two test modules. The `emit_native_web_search` flag carries the same per-tool decision through `api/routes/_pipeline.py`, `api/routes/chat.py` and `api/routes/messages.py` as a named parameter.

### Proposal

The pipeline and the three loops dispatch web search through the registry; delete the hand-wiring for it. `web_fetch` is a registry entry of its own (`_web_fetch_tool.TOOL`), so this issue covers web search alone and #1200 owns the web fetch proof.

### Measured at `1edd04de` (2026-09-23)

Files counted with `git grep -l -i web_search origin/main -- src/gateway | wc -l`, lines with `git grep -c -i`.

| Where | Count |
| --- | --- |
| Files in `src/gateway` mentioning `web_search` | 42 |
| Lines in `api/routes/_pipeline.py` | 143 |
| Lines in `services/mcp_loop_messages.py` | 41 |
| Lines in `services/mcp_loop_responses.py` | 19 |
| Lines in `services/mcp_loop.py` | 5 |

The 38-file figure this issue carried before was measured before #1366.

### Acceptance

- [ ] Every existing web search test passes.
- [ ] No `web_search`-specific branch remains in `mcp_loop.py`, `mcp_loop_messages.py` or `mcp_loop_responses.py`.
- [ ] `emit_native_web_search` is absorbed into the registry entry, or the pull request names why it stays a parameter.
- [ ] The file mention count for `web_search` in `src/gateway` drops, and the number is recorded in the pull request.
- [ ] The compatibility shim `services/web_search_backend.py` is deleted, or the pull request names why it stays.

### Depends on

#1195, closed 2026-09-17 by #1285 and #1309. The registry is `services/tools/_registry.py`.

Part of #1171.

PR #1285, opened 2026-09-17
refactor(gateway): list the tools the gateway runs itself in one registry

### LINKED ISSUE #1195: Define the tool interface and registry
### Problem

Three built-in tools (`WebSearchBackend`, `SandboxBackend`, `SearchTool`) are unrelated classes; `Tool` is a string enum; each of the three API dialect loops hand-wires each tool. `_pipeline.py` references web search or sandbox 239 times. A new tool touches the pipeline, three loops, pricing and settings.

### Proposal

One `BuiltinTool` protocol: declare the tool's wire schema per dialect, run a call, report its charge. A registry keyed by tool name, populated from a literal tuple in code. No tool moves yet; this PR adds the interface and wraps the three existing tools without changing any call site.

### Acceptance

- [ ] All three tools satisfy the protocol under `mypy --strict`.
- [ ] No behavior change.
- [ ] No call site edited.

### Depends on

Nothing.

Part of #1171.

peteski22:
## Description

Otari runs two tools itself when a model calls them: web search and code execution. Each one is wired by hand in several places. The request pipeline decides whether it may run, pricing checks that it has a rate, the usage page lists it as a filter, and `GET /api/v1/tools` describes it. A new tool has to repeat all of that.

This PR adds the first half of the seam that removes the repetition: one registry that lists the tools the gateway runs itself, with the facts each of those places needs. Nothing reads the registry yet, so nothing changes for anyone running Otari. The PRs that follow move each place onto it, one tool at a time.

### What exists today

- **The runtime seam already exists.** `ToolBackend` in `services/_tool_loop.py` is the protocol the three API format loops (Chat Completions, Messages, Responses) execute every gateway-run call through: `openai_tools`, `owns_tool`, `call_tool`, `purpose_hints`. `SandboxBackend`, `WebRetrievalBackend` and `MCPClientPool` implement it. A backend records each call on the request's `ToolUsageTally` under the tool's name, and settlement prices that name under `otari:<name>`. So running a call and reporting its charge already go through one interface.
- **What has no seam is everything around it.** The `Tool` enum and the per-tool extraction in `api/routes/_tools.py`, admission and workspace policy in `prepare_gateway_tools`, the one-backend-per-request ladder in dispatch, `_require_tool_pricing`, the startup pricing warning, `GATEWAY_TOOL_NAMES` in the usage routes, and the tools endpoint each name `web_search` or `code_execution` directly.
- **`SearchTool` is not a tool the model calls.** It configures the direct `POST /api/v1/search` endpoint and never enters the tool loop, so it does not fit this interface and is not listed. #1198's wording needs updating.
- **A tool's schema is not per API format.** Each tool declares one function definition in the Chat Completions shape, and the loops convert it for Messages and Responses. The only per-format, per-tool code is web search's native result items (`server_tool_use` and `web_search_call`), which code execution deliberately does not produce.

### The interface

`BuiltinTool` in `services/builtin_tool.py`, one entry per tool the model can call:

| Member | What it is | Who needs it once the call sites move |
| --- | --- | --- |
| `name` | The function name the model calls. It is also the name a call is metered and priced under, so one field carries all three. | The pricing gate, the startup pricing warning, usage filters, and the loops' web search checks |
| `definition()` | A new function definition in the Chat Completions shape on each call. A tool author never writes a per-format schema. | The tools endpoint and the backend that advertises the tool |
| `configured(config)` | Whether this deployment has a backend that can run the tool. Both entries point at a `GatewayConfig` method: `web_search_configured`, and a new `sandbox_configured` that replaces a sixth copy of the sandbox URL check. | The tools endpoint, the startup pricing warning, and admission's "not configured" 400 |

It is a frozen dataclass rather than a `Protocol`, for the reason `CoreFeature` is one: each entry is one named value, and mypy checks its callables where the entry is built.

`services/builtin_tool_registry.py` holds `BUILTIN_TOOLS`, a literal tuple edited by hand. Each tool's own module declares its entry (`services/web_search_tool.py`, `services/code_execution_tool.py`). Nothing is discovered, and nothing registers itself on import. There is no lookup by name, because the next changes walk the tuple.

### Why admission and opening a backend are not members yet

Those are the two steps that differ most between the tools, and both are resolved in the routes layer today.

- Web search folds its workspace policy into the caller's tool entry. Code execution keeps its policy results (execution timeout, session image, allowed tool kinds) out of the entry, so a caller cannot set them. One `open(request)` signature over both needs either an untyped bag or a type per tool.
- `_build_web_search_backend` lives in `api/routes/_tools.py`, and a services module may not import the API layer.

So those members are added in the PRs that move that code, where their shape can come from the code rather than from a guess.

### How the next PRs are expected to use it

- **#1196, web search.** Move web search's extraction, admission and backend construction into its tool module, and add the missing members: an admission step that returns a prepared tool, and a way to open that tool's `ToolBackend` for one request. Dispatch then walks `BUILTIN_TOOLS` instead of naming web search. The pricing gate, the pricing warning, the usage names and the tools endpoint read the registry.
- **#1197, code execution.** The same for the sandbox. It is the test of the interface, because its policy results do not live in its entry.
- **#1198.** Once both tools are behind the registry, the per-tool branches in dispatch go. Web search's native result items remain a web search concern. The issue's "search" step does not apply (see above).
- **Web fetch** (#979) then becomes one module and one entry.

### Guards added now

Tests pin today's hand-kept lists to the registry, so a tool added to one and not the other fails CI:

- every `Tool` member has an entry, and every entry has a `Tool` member;
- the usage routes' `GATEWAY_TOOL_NAMES` matches the registry;
- the tools endpoint's `available` matches `configured` across configurations, including the sandbox URL set only in the environment;
- the backend that runs each tool advertises the registry's definition and owns its name, typed as `ToolBackend` so mypy checks the backend against the protocol.

## How to test it locally

```bash
make lint && make typecheck
uv run pytest tests/unit/test_builtin_tools.py -q
make test-unit
```

No call site changed: `git diff origin/main --stat` lists the new files and one added method in `core/config.py`. The integration tests that cover web search, code execution and the tool loops pass unchanged.

## PR Type

- [ ] New Feature
- [ ] Bug Fix
- [x] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes #1195

Part of #1171.

## Checklist

- [x] I ran the Definition of Done checks locally (`make lint`, `make typecheck`, `make test`). `make lint`, `make typecheck`, `make test-unit`, the smoke gate, and the 17 integration files for web search, code execution, search and the tool loops (460 passed). Four unit tests in `test_claude_code_import.py` and `test_hook_cli.py` pick up a credential from this machine's environment and fail identically on `main`; CI runs them clean.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

The design was checked against both tools, the MCP pool, the three format loops and the pipeline on current `main` before any code was written.

**NOTE:**


src/gateway/services/code_execution_tool.py
```diff
@@ -0,0 +1,17 @@
+"""Code execution, as the tool registry lists it."""
+
+from gateway.core.config import GatewayConfig
+from gateway.core.env import otari_env
+from gateway.services.builtin_tool import BuiltinTool
+from gateway.services.sandbox_backend import CODE_EXECUTION_TOOL_NAME, code_execution_tool_definition
+
+
+def _configured(config: GatewayConfig) -> bool:
+    return bool(config.sandbox_url or otari_env("SANDBOX_URL"))
```
khaledosman: Sixth copy of this expression. It already appears at `api/routes/tools.py:80`, `api/routes/workspace_code_execution_policy.py:53`, `services/playground_service.py:240`, `services/pricing_init_service.py:101` and `api/routes/_pipeline.py:2607`.

Web search's entry, one file over, does the opposite: it points at `GatewayConfig.web_search_configured`, whose docstring says it exists so "the request path, the per-workspace page and the tool catalog" cannot disagree about it. The same reasoning applies here, and this PR is the natural place to finish the job.

Add `GatewayConfig.sandbox_configured()` beside `web_search_configured` and set `configured=GatewayConfig.sandbox_configured`. `test_configured_agrees_with_what_the_tools_endpoint_reports` then guards a fact that is true by construction rather than one that has to be re-checked.

src/gateway/services/builtin_tools.py
```diff
@@ -0,0 +1,19 @@
...
+Each entry names the ``BuiltinTool`` its own module declares, so nothing is built here.
+"""
+
+from gateway.services import code_execution_tool, web_search_tool
+from gateway.services.builtin_tool import BuiltinTool
+
+BUILTIN_TOOLS: tuple[BuiltinTool, ...] = (web_search_tool.TOOL, code_execution_tool.TOOL)
+
+
+def find_builtin_tool(name: str) -> BuiltinTool | None:
```
khaledosman: No caller in `src/`; only the test at the bottom of `test_builtin_tools.py` uses it. None of the follow-ups the description sketches needs it either, since #1196 and #1198 both walk `BUILTIN_TOOLS` rather than look one up by name.

Drop it until something wants it, so the seam lands with exactly the surface the next PR consumes. `test_find_builtin_tool_matches_the_function_name_only` goes with it; the name/type distinction it pins is already covered by `test_every_gateway_tool_type_has_exactly_one_listed_tool`.

src/gateway/services/builtin_tool_registry.py
```diff
@@ -0,0 +1,19 @@
+"""The registry of tools the gateway runs itself.
```
khaledosman, line 1: `builtin_tool.py` and `builtin_tools.py` sit in one package and differ by one character, and the two are imported together in every file that touches either. The split itself is forced (`builtin_tools` imports `web_search_tool`, which imports `builtin_tool`), so this is a naming fix, not a structural one: `builtin_tool_registry.py` for this module reads unambiguously at every import site and still matches the purpose-named service convention in AGENTS.md.

khaledosman (approved):
Verified on the head commit in a clean worktree: `tests/unit/test_builtin_tools.py` passes (16 tests), and nothing under `src/` imports the new modules, so the "nothing reads the registry yet" claim holds.

The red `test-integration` is `test_hooks_route.py::test_many_separator_only_commands_resolve_quickly` failing its wall-clock assertion (`< 0.5`, got ~0.57s). Same failure on #1281's run, and already the subject of 8a5fded2. Not this PR.

peteski22:
@coderabbitai review

src/gateway/services/code_execution_tool.py
peteski22: Addressed in abc40143: added `GatewayConfig.sandbox_configured()` beside `web_search_configured` and pointed the entry at it. The test still compares it with the tools endpoint, which keeps its own copy until that call site moves onto the registry.

src/gateway/services/builtin_tools.py
peteski22: Addressed in 180de06f: removed `find_builtin_tool` and its test.

src/gateway/services/builtin_tool_registry.py
peteski22: Addressed in d5f7e86f: renamed the module to `builtin_tool_registry.py`.

peteski22:
@coderabbitai review

Issue #1264, opened 2026-09-17
Document the modular monolith, so new code follows the target shape and not the code beside it


peteski22:
### Problem

The cleanup moves the backend to a modular monolith: one process and one deploy, with the code cut by domain. Services and repositories are one package per domain inside their layer, and routes, schemas, exceptions and models are one module per domain (#1201). The docs that tell a contributor how to write backend code do not say so, and one of them says the opposite.

`.github/skills/backend-standards/SKILL.md`, section "Layering", says:

- "Keep request/response Pydantic models near the handler". The target is `schemas/<domain>.py`.
- Services "own the DB work". The target is that a repository runs every query.
- "Service-specific exceptions live beside their service". The target is `exceptions/<domain>_exceptions.py`.

Nothing else fills the gap:

- `ARCHITECTURE.md` has one paragraph on a core feature: a module per layer plus a registry entry. That paragraph is out of date, and the document does not name the shape, list the layers, or say which layer may import which.
- No doc says how a service is built (#1262) or who commits (#1276).
- `.github/instructions/` has no layer rule, so CodeRabbit cannot flag a new route that runs a query.
- No doc says which modules are still in the old shape. Contributors and agents copy the code beside them, and most of that code is in the old shape.

### Proposal

One PR, docs only. Each fact lives in one place, and the other places link to it.

| File | Holds |
| --- | --- |
| `ARCHITECTURE.md` | A short section that names the modular monolith: one process, code cut by domain, a package per domain in `services/` and `repositories/`, one module per domain in the other layers, which layer may import which, and the feature registry as the one wiring list. It replaces the out-of-date core feature paragraph. It sits beside the "where new code goes" section from #1272, and links to `docs/domains.md` (#1201) for the map. |
| `.github/skills/backend-standards/SKILL.md` | "Layering" rewritten as the rules for each layer: its path, what it does, and what it must not do. It adds how a service is built: one service per domain with a small public API, by constructor injection, with its own repositories and the Unit of Work, never the session (#1262). It adds that only a Unit of Work block commits (#1276). It adds that divider comments which cut a module into sections mean the module should be split along them. |
| `src/gateway/AGENTS.md` | Links only. |
| `.github/instructions/` | A backend architecture file with `applyTo: "src/gateway/**/*.py"` that restates the layer rules for CodeRabbit, which does not read `ARCHITECTURE.md` or the skill. The root `AGENTS.md` lists it beside the other instruction files. |

[...]

PR #1286, opened 2026-09-17
chore(lint): keep queries out of routes and services, with a shrinking baseline

### LINKED ISSUE #1203: The layering rule in the boundary check, with a baseline of known violations
### Proposal

Extend `scripts/check_architecture.py` with the layer rules from #1201.
Each rule starts from a baseline in the script: the list of today's offenders.
A baseline may only shrink, and a PR that adds a name to one fails review by rule.
Every domain issue removes its names.

1. Nothing under `services/` imports `sqlalchemy`. A service can neither hold a session nor build a query (#1262).
2. Nothing under `api/routes/` imports `sqlalchemy`. Today the check forbids only `sqlalchemy.orm`.
3. Only the Unit of Work (#1276) calls `commit()` or `rollback()`.
4. Only `services.<domain>` and `api/deps.py` import `repositories.<domain>`.
5. Code outside a domain imports only the package root, `gateway.services.<domain>`. A module whose name starts with `_` is private to its package.
6. Domain service packages do not import each other in a cycle.

`services/tenancy/`, `services/routing/` and `services/mail/` are packages already, so their current importers start on the baseline for rule 5.

Measured on `main` at `9152d0bf`:

| Rule | Offenders today |
| --- | --- |
| 1. Service modules that import `sqlalchemy` | 48 |
| 2. Route modules that import `sqlalchemy` | 61 of 73 |
| 3. Modules outside the database core that call `commit()` or `rollback()` | 54 (30 in `services/`, 23 in `api/`, 1 in `adapters/`) |
| 5. Modules that import a submodule of `services/tenancy/`, `routing/` or `mail/` from outside it | 57, 11 and 4 |

### Acceptance

- [ ] Each rule fails on a synthetic offender.
- [ ] Each baseline is the current offender set, and the PR records the counts.
- [ ] `tests/unit/test_check_architecture.py` covers every rule.
- [ ] `make lint` runs it.

### Depends on

#1276 for rule 3. Land it before #1202, so #1202 shrinks the baselines.

Part of #1171.

peteski22:
## Description
The backend is moving to a layered shape where every database query lives in a repository, and routes and services only call repositories. Today many routes and services still build queries themselves. New code tends to copy whatever sits beside it, so the number can grow while the cleanup tries to shrink it.

This adds a lint rule. A route or a service that builds a query now fails the architecture check, unless it is one of the modules that already did when the rule landed. That list only shrinks: when a module stops building queries, the check fails until its name is removed from the list. Each domain cleanup PR therefore records its progress by deleting names, and no new module can join the list without a visible, reviewed edit to the check.

### Measured
On `main` at the parent of this commit. "Builds a query" means the module imports or references `select`, `insert`, `update`, `delete` or `text` from SQLAlchemy or SQLModel, directly or through a submodule it binds. The 39th service, `playground_dispatch.py`, arrived on `main` in #1297 after this PR opened.

| Measure | Before | After |
| --- | --- | --- |
| Service modules that build a query | 39 of 104 | 39, all on the baseline |
| Route modules that build a query | 18 of 72 | 18, all on the baseline |
| New route or service modules that may build a query | Any | None |

The domain map's earlier count (40 services, 19 routes) also required a call on the session, and was taken before several modules moved.

### Architecture rule
This adds a rule to `scripts/check_architecture.py`: a route or a service builds no query unless it is on a shrinking baseline. Why: the target layering puts every query in a repository, and without the check new code copies the modules that still build queries.

## How to test it locally
2. Add `from sqlalchemy import select` to any service that is not on the baseline, for example `src/gateway/services/tool_format.py`, and run `make check-architecture`. It fails with "builds a query with select; a query belongs in a repository". Revert it.
3. Remove every query import from a module on the baseline, or delete one baseline entry's file, and the check fails with "is on the query baseline but builds no query; remove it from the baseline".
4. `uv run pytest tests/unit/test_check_architecture.py` covers each spelling (`from sqlalchemy import`, `from sqlmodel import`, a dialect `insert`, `sa.update` through an alias, `sqlalchemy.sql.delete` through an unaliased import, raw `text`, and `sql.select` or `expression.update` through a submodule a from-import binds), in routes, services and a service subpackage. It also covers repositories staying free to query, query types and column helpers staying allowed, the baseline exemption, a stale baseline entry for both a module that stopped querying and a module that no longer exists, and `main()` failing.

The backend skill still says services own the database work. #1264 rewrites that section to match this rule.

## PR Type
- [ ] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [x] Infrastructure / CI

## Relevant issues
Fixes #1203

Part of #1171.

## Checklist

## AI Usage
- [ ] No AI was used.

**AI Model/Tool used:**
Claude Opus 5 via Claude Code

**NOTE:**

## Summary

- Added an architecture check that prevents route and service modules from building database queries.
- Added a shrinking baseline for existing violations.
- Updated documentation to direct queries to repository modules.
- Added tests for query detection, supported imports, baseline cleanup, and allowed repository usage.

## Benefits

This keeps database query logic in repository modules and prevents new layering violations without changing runtime behavior.

## Technical notes

The check covers `select`, `insert`, `update`, `delete`, and `text` from SQLAlchemy and SQLModel, including aliases and `from` imports. Test execution status is not provided.


scripts/check_architecture.py
```diff
@@ -262,6 +265,125 @@ def check_file(file_path: Path, src_root: Path) -> list[tuple[int, str, str]]:
...
+    """Return the name an attribute chain starts from, or None when it starts from an expression."""
+    while isinstance(node, ast.Attribute):
+        node = node.value
+    return node.id if isinstance(node, ast.Name) else None
+
+
+def _query_primitives(tree: ast.Module) -> list[tuple[int, str]]:
+    """Return the line and name of each query-building primitive a module imports or references."""
+    # An unaliased `import sqlalchemy.sql` binds the name `sqlalchemy`.
+    library_names = {
```
khaledosman, line 348: `library_names` is seeded only from `ast.Import`, so a submodule bound through `ImportFrom` never reaches it and the attribute path below misses the query. Verified on this branch with the baseline emptied:

```python
# gateway/services/sneaky_service.py
from sqlalchemy import sql

def q(db):
    return db.execute(sql.select(1))
```

`check_query_layering` returns `[]`. Same for `from sqlalchemy import orm` and any other submodule.

Fix: also fold `ImportFrom` aliases into the set when the module is a query library.

```python
library_names = {
    alias.asname or alias.name.split(".")[0]
    for node in ast.walk(tree)
    if isinstance(node, ast.Import)
    for alias in node.names
    if alias.name.split(".")[0] in QUERY_LIBRARIES
} | {
    alias.asname or alias.name
    for node in ast.walk(tree)
    if isinstance(node, ast.ImportFrom) and node.level == 0 and node.module is not None
    if node.module.split(".")[0] in QUERY_LIBRARIES
    for alias in node.names
    if alias.name not in QUERY_PRIMITIVES
}
```

scripts/check_architecture.py
```diff
@@ -14,6 +14,9 @@
    route may not import it; and nothing under gateway/ imports
    importlib.metadata, importlib_metadata or pkg_resources, so nothing is
    discovered.
+10. Query layering: a route or a service builds no query, because a query
```
khaledosman: Collides with #1283, which numbers its new rule `10.` here too, appends to the same AGENTS.md sentence, and adds its own branch to the same `if` in `main()`. Whichever merges second needs `11.` and a rebase; worth deciding the order now so the renumber is one edit rather than a conflict resolved twice.

khaledosman (approved):
Verified on the head commit in a clean worktree: `check_architecture.py` exits 0, and the baseline's 38 service and 18 route entries match the table in the description.

Two findings inline, neither blocking.

scripts/check_architecture.py
peteski22: Reproduced with two new cases (`from sqlalchemy import sql` then `sql.select`, and `from sqlalchemy.sql import expression as e` then `e.update`); both returned an empty result. Fixed in 6cd4936f: names a from-import of a query library binds are now followed into attribute chains. The baseline did not change for this fix; the rebase added `playground_dispatch.py`, which #1297 brought to main.

scripts/check_architecture.py
peteski22: Merge order: #1283, then #1286, then #1292. Each later PR rebases after the one before it merges and renumbers its rule in the docstring list (#1286 to 11, #1292 to 12), resolving the AGENTS.md sentence and the `if` in `main()` in the same rebase.

peteski22:
@coderabbitai review

scripts/check_architecture.py
```diff
@@ -262,6 +265,132 @@ def check_file(file_path: Path, src_root: Path) -> list[tuple[int, str, str]]:
...
+            relative_path = py_file.relative_to(src_root).as_posix()
+            try:
+                tree = ast.parse(py_file.read_text(encoding="utf-8"), filename=str(py_file))
+            except SyntaxError:
+                continue  # check_file already reports an unparseable file.
+            primitives = _query_primitives(tree)
+            if not primitives:
+                continue
+            querying.add(relative_path)
+            if relative_path not in QUERY_BASELINE:
```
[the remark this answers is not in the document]
peteski22: Not changing this. The finding is accurate that a PR can add a path to `QUERY_BASELINE` and `make lint` passes; #1203 settles that as a review rule ("a PR that adds a name to it fails review by rule"). A fixed ceiling would not enforce more: it is a constant in the same file, raised in the same one-line diff that adds the name, so review sees exactly what it sees today. A comparison against the merge base would make the lint script depend on git history and on CI fetch depth, and it would also fail a rebase onto a `main` that gained an offender before this rule landed, which happened on this PR with #1297. `scripts/check_architecture.py` is a CODEOWNERS path, so once `main` requires code-owner review, growing the baseline needs a code owner's approval, which is the gate this list relies on.

PR #1291, opened 2026-09-17
test: fix the flaky hooks timing test and clear the integration warnings

peteski22:
## Description
One integration test fails on some CI runs, on `main` and on unrelated PRs. It checks that a command made only of `;` separators is not compared against every forbidden phrase. It did that by timing the whole HTTP request against a 0.5 second limit. The bug it guards against took about 1.1 seconds, but a correct request on a busy CI runner took 0.54 to 0.65 seconds in the five failed runs. No time limit tells those two apart reliably.

The test now counts how many times a command segment is compared against a phrase, and expects zero. It then sends one real command and expects a nonzero count, so the counter cannot go quiet unnoticed. With the fix removed it counts 25,050,000 and fails.

Three more hooks tests used the same kind of timer, with a 1 second limit. They now count how often commands are tokenized instead:
- Commands over the character limit are refused before any of them is tokenized.
- A policy with no command check never tokenizes commands.
- Each command is tokenized once per request, however many gates check it.

This PR also clears the 15 warnings that every integration run prints:
- Five organization domain tests are not async, but a file-wide asyncio mark applied to them. Only the async tests carry the mark now.
- Five tests sent the old `provider/model` model name, and any-llm warns that this name is deprecated. Four of them test the old name on purpose, so they ignore that one warning. The fifth used it by accident and now uses `provider:model`.
- google-genai uses a typing name that Python 3.14 deprecates. Its newest release still does, so an upgrade does not help. The test config ignores that one warning from that one module, and says when to remove the filter.

Nothing changes for someone running Otari. Only tests and test configuration change.

## How to test it locally
1. Run `make lint` and `make typecheck`.
2. Run `uv run pytest tests/integration/test_hooks_route.py`. All tests pass.
3. To see a count catch its bug, change the last line of `_command_segments` in `src/gateway/agent_runtime/domain/evaluators.py` to `return segments`. Run `uv run pytest tests/integration/test_hooks_route.py -k separator_only`. It fails with `assert 25050000 == 0`. Undo the change.
4. Run `uv run pytest tests/integration -n auto` on Python 3.14. The warnings summary is gone. The google-genai warning shows only on Python 3.14. The other warnings show on 3.13 too.

The `test-integration` CI job runs all of this.

## PR Type
- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues
No issue. The flaky test failed `main` in run 35206085491, and failed on #1278 and #1279.

## Checklist
- [x] I ran the Definition of Done checks locally (`make lint`, `make typecheck`, `make test`). The unit and integration suites ran on Python 3.14 with `-n auto`, as CI runs them.

## AI Usage
- [ ] No AI was used.

**AI Model/Tool used:**
Claude Opus 5 via Claude Code

Peter found the failure, asked for a separate fix from `main` that also clears the warnings, set the constraints, and decided how to handle each review note. Claude Code traced the failure across recent runs, chose the fixes, and wrote them. Each new count was checked against the code with a matching bug put back in.

**NOTE:**


tests/integration/test_hooks_route.py
```diff
@@ -521,31 +522,36 @@ def test_apostrophe_in_a_trailing_comment_does_not_evade_a_required_gate(
...
-    start = time.time()
     response = client.post(
         f"{API_ROOT}/hooks/check",
         json={"policy_yaml": policy, "commands": commands},
         headers=master_key_header,
     )
-    assert time.time() - start < 0.5
     assert response.status_code == 200, response.text
     assert response.json()["blocked"] is False
+    assert comparisons == 0
```
khaledosman, line 564: This assertion is vacuously satisfiable. It only counts anything because `evaluate_command_match` resolves `_contains_subsequence` through module globals (`evaluators.py:582`). Inline that call, rename the helper, or restructure the comprehension into a different helper and the patch never fires, so the test stays green while asserting nothing.

That is a worse failure mode than the timing flake it replaces: the timer at least went red when it broke.

Prove the counter is wired in the same test. After this line, reset it and send one real command through the same patched client:

```python
    assert comparisons == 0

    # The zero above only means anything if the counter fires at all.
    response = client.post(
        f"{API_ROOT}/hooks/check",
        json={"policy_yaml": policy, "commands": ["p0"]},
        headers=master_key_header,
    )
    assert response.status_code == 200, response.text
    assert comparisons > 0
```

tests/integration/test_hooks_route.py
```diff
@@ -521,31 +522,36 @@ def test_apostrophe_in_a_trailing_comment_does_not_evade_a_required_gate(
...
+    This counts comparisons because a timed request on a loaded CI runner measures the runner.
     """
+    comparisons = 0
+
+    def counting_contains_subsequence(segment: list[str], phrase: list[str]) -> bool:
+        nonlocal comparisons
+        comparisons += 1
+        return _contains_subsequence(segment, phrase)
+
+    monkeypatch.setattr("gateway.agent_runtime.domain.evaluators._contains_subsequence", counting_contains_subsequence)
```
khaledosman, line 549: Three wall-clock guards of exactly the class this PR is removing are still in this file, at lines 440, 464 and 495, all `assert time.time() - start < 1.0`, and 8a5fded2 already had to relax a fourth to 3.0. The technique on this line retires the whole class.

Either convert them the same way, or add a line to each saying why roughly 7s of bug against a 1.0s budget has headroom that 1.1s against 0.5s did not.

pytest.ini
```diff
@@ -2,3 +2,6 @@
 timeout = 120
 markers =
     dashboard_bundle: needs the built dashboard bundle, so it skips without one. Applied by requires_dashboard_bundle (tests/unit/_dashboard_bundle_support.py) and selected with `-m dashboard_bundle` by the one CI job that builds a bundle (.github/workflows/otari-dashboard-serving.yml).
+# google-genai uses typing._UnionGenericAlias, which Python 3.14 deprecates.
```
khaledosman, line 5: The comment records what the warning is but not when the ignore goes. The durable facts are in the PR description (google-genai's newest release still does this, so an upgrade does not help), and a PR description is not where the next reader looks.

Put the removal condition and an upstream issue link here. `CONTRIBUTING.md#when-the-fix-is-upstream-but-otari-cannot-wait` is the repo's rule for the analogous case, and its reasoning carries over: a shim with no upstream issue is permanent by accident.

khaledosman (approved):
Counting comparisons instead of timing the round trip is the right call, and I confirmed the patch target actually fires: `evaluate_command_match` looks `_contains_subsequence` up through module globals at `src/gateway/agent_runtime/domain/evaluators.py:582`. Also checked that all 34 async tests in `test_organization_domains.py` picked up an explicit mark with none stranded on the 5 sync ones, that both `filterwarnings` message regexes match any-llm's actual text (`any_llm.py:305`), and that `pytest.ini` still parses as intended under iniconfig.

Three notes inline, none blocking.

---

dpoulopoulos (approved):

tests/integration/test_hooks_route.py
peteski22: Addressed in e23f7d12: after the zero, the test sends `p0` through the same patched client and asserts a nonzero count. With the evaluator calling the helper under another name, it fails with `assert 0 > 0`.

tests/integration/test_hooks_route.py
peteski22: Addressed in 191bbd6e: all three now count `_command_segments` calls through a shared fixture. The two zero-count tests then send one real command as a control, and the per-gate test expects each command exactly once. Each assertion fails against a matching break in the route or the evaluator. The 3.0s dedup guard still uses a timer.

pytest.ini
peteski22: Addressed in b3826862: the comment now has a TODO to remove the filter once the locked google-genai version no longer imports that name. Upstream tracks it in googleapis/python-genai#1640, with a fix open in googleapis/python-genai#2225. That link is in this reply and the commit message, not in the comment.

Issue #1262, opened 2026-09-17
Services receive their own repositories and a Unit of Work, never the session


peteski22:
### Problem

Most service code takes the database session as a call parameter.

A function that takes a session also takes the config, the ports and the other collaborators it needs, on every call. A new dependency becomes one more parameter on every function in the chain. `build_merged_catalog` takes eight.

Moving the session into `__init__` is not enough. A service that holds the session can still run a query, so the layer rule stays a matter of review.

### Proposal

Each domain has one service, built by constructor injection (Martin Fowler, "Inversion of Control Containers and the Dependency Injection pattern", 2004, https://martinfowler.com/articles/injection.html). It receives:

- its own domain's repositories (Repository: Edward Hieatt and Rob Mee, in Martin Fowler, *Patterns of Enterprise Application Architecture*, 2002, https://martinfowler.com/eaaCatalog/repository.html),
- the Unit of Work (#1276),
- config and ports,
- the services of other domains it needs.

It never receives the session, and never receives another domain's repository. A service with no session cannot run a query.

- The service lives in its domain's package, `services/<domain>/`. The package exports the service and the types its public methods use, and nothing else.
- Its public API is small. Each public method is one use case, and the implementation hides in the package's private modules (deep modules: John Ousterhout, *A Philosophy of Software Design*, 2018, ch. 4, "Modules Should Be Deep").
- A builder in `api/deps.py` builds it for a request. A worker calls the same builder with a Unit of Work it creates from its own session.
- A module-level function that takes a session becomes a method of its domain's service, or a repository method if all it does is run a query. A helper that needs no database stays a plain function.

Do this per domain, as a step in each domain issue from #1201, not as one sweep.

### Acceptance

- [ ] Nothing under `services/` imports `sqlalchemy` or takes an `AsyncSession`. The boundary check enforces it from a baseline that may only shrink (#1203).
- [ ] Each domain's service is built by one builder in `api/deps.py`.
- [ ] No service receives another domain's repository.

### Measured

On `main` at `9152d0bf`, 2026-09-17. The count is every top-level `def` or `async def` under `services/` with a parameter annotated `AsyncSession`.

| Measure | Today |
| --- | --- |
| Module-level service functions that take a session | 185 (151 public, 34 private) |
| Files that hold them | 40 |
| Of those functions, with five or more parameters | 29 |
| Service classes that receive the session in `__init__` | 14 |
[...]

PR #1292, opened 2026-09-17
chore(lint): stop new service functions taking a session, with a shrinking baseline

peteski22:
## Description
The cleanup is moving services to one shape: each domain has a service built with its own repositories, a Unit of Work, config and the collaborators it needs, and it never receives the database session. Today most service code instead takes the session as a parameter on every call, and every new dependency becomes one more parameter down the whole call chain. New code copies the functions beside it, so the count can grow while each domain issue tries to shrink it.

This adds a lint rule. A module-level function under `services/` that takes a database session now fails the architecture check, unless it is one of the functions that already did when the rule landed. The list names functions, not files, so a new function added to a file that is already on the list still fails. The list only shrinks: when a function stops taking a session, or goes away, the check fails until its name is removed.

This covers one part of the check in #1262 only: module-level functions. Moving each domain's functions onto its service, and checking what a service receives when it is built, stay with the domain issues and a later change, so this PR refers to #1262 rather than closing it.

### Architecture rule
This adds a rule to `scripts/check_architecture.py`: a module-level function under `services/` takes no `AsyncSession` unless it is on a shrinking baseline. Why: such a function belongs on its domain's service, which receives repositories and a Unit of Work, never a session, and without the check new code copies the functions that still take one.

### Measured
On `main` at the parent of this commit, counting every function defined at module scope under `services/`, including inside module-level control flow, with a parameter annotated `AsyncSession`, `AsyncSession | None`, `Optional`, `Union` or `Annotated` around it. A parameter that only mentions the type, such as a factory returning a session, does not count. #1262 counted 185 before #1297 added `playground_dispatch.py::resolve_dispatch_key`.

| Measure | Before | After |
| --- | --- | --- |
| Module-level service functions that take a session | 186 (152 public, 34 private), in 41 files | 186, all on the baseline |
| New module-level service functions that may take a session | Any | None |

## How to test it locally
2. Add `async def probe(db: AsyncSession) -> None: ...` at the top level of any service, for example `src/gateway/services/budget_service.py`, and run `make check-architecture`. It fails with "probe takes a session; move it onto its domain's service, which receives repositories and a Unit of Work, never a session", even though other functions in that file are on the baseline. Revert it.
3. Delete a baseline entry's function, or remove its session parameter, and the check fails with "is on the session parameter baseline but takes no session; remove it from the baseline".
4. `uv run pytest tests/unit/test_check_architecture.py` covers a plain annotation, an optional keyword-only one, a string annotation, a dotted type, and `Optional`, `Annotated` and `None | AsyncSession` wrappers. It covers a factory annotation (`Callable[[], AsyncSession]`, `async_sessionmaker[AsyncSession]`) staying clean, and a function inside a module-level `if`, `try` or `with` failing. It also covers methods, nested functions and functions with no session staying clean, repositories staying free to take a session, a function on the baseline passing while a new function in the same file fails, a stale entry for both a function that stopped taking a session and a file that no longer exists, and `main()` failing.

## PR Type
- [ ] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [x] Infrastructure / CI

## Relevant issues
Refs #1262

Part of #1171.

## Checklist

## AI Usage
- [ ] No AI was used.

**AI Model/Tool used:**
Claude Opus 5 via Claude Code

**NOTE:**

## Summary

Adds an architecture check that prevents new module-level functions under `gateway/services/` from accepting `AsyncSession`.

The check tracks existing exceptions in a shrinking baseline. It reports new violations and baseline entries that no longer need an exception.

Updates architecture guidance and adds tests for supported annotations, methods, nested functions, repositories, baseline handling, and failure reporting.

## Benefits

- Keeps session handling consistent.
- Prevents new violations.
- Supports gradual cleanup of existing services.


scripts/check_architecture.py
```diff
@@ -214,6 +218,253 @@ def _imported_modules(node: ast.Import | ast.ImportFrom, file_path: Path, src_ro
...
+    "gateway/services/workspace_scope.py::resolve_workspace_id",
+    "gateway/services/workspace_scope.py::workspace_for_key_id",
+)
+
+
+def _annotation_names_session(annotation: ast.expr | None) -> bool:
+    """Return whether a parameter annotation names the async session type, including in a string annotation."""
+    if annotation is None:
+        return False
+    for node in ast.walk(annotation):
```
khaledosman: `ast.walk` matches the type anywhere in the annotation subtree, so a parameter that only mentions it is flagged as taking a session. `factory: Callable[[], AsyncSession]` fails the check today, and so would `async_sessionmaker[AsyncSession]`. Those are shapes a function moving off a live session would plausibly use, and the baseline is shrink-only, so there is no way to record one.

Every one of the 185 baseline entries annotates a plain `AsyncSession` or `AsyncSession | None`, so the deep walk buys nothing today and only adds the false positive. Match a `Name` or `Attribute` at the top level of the annotation, unwrapping `X | None` and `Annotated[...]`, instead of walking the whole subtree.

scripts/check_architecture.py
```diff
@@ -214,6 +218,253 @@ def _imported_modules(node: ast.Import | ast.ImportFrom, file_path: Path, src_ro
...
+    """Check that no module-level service function off the baseline takes a session, and that every entry still does."""
+    violations: list[str] = []
+    taking: set[str] = set()
+    for py_file in sorted((src_root / SESSION_SCOPE).rglob("*.py")):
+        relative_path = py_file.relative_to(src_root).as_posix()
+        try:
+            tree = ast.parse(py_file.read_text(encoding="utf-8"), filename=str(py_file))
+        except SyntaxError:
+            continue  # check_file already reports an unparseable file.
+        for node in tree.body:
```
khaledosman: Only `tree.body` is scanned, so a module-level function nested in a module-level `if` or `try` escapes the rule. `if sys.version_info:` followed by an indented `async def f(db: AsyncSession) -> None: ...` in a service reports clean, though it defines `f` at module level at runtime. Recurse into the bodies of module-level `If`, `Try` and `With` nodes; the nested-function exemption the tests cover stays intact because those are inside a `FunctionDef`.

khaledosman (approved):
Verified locally at `d27e9a97`: the check passes, the baseline is 185 entries across 40 files (34 private), ruff and `ruff format --check` are clean, and `tests/unit/test_check_architecture.py` is 59 passed.

Two non-blocking findings inline, both about the matcher's edges rather than the baseline.

Not findings, for the record: an import alias (`AsyncSession as DbSession`) escapes the check, but mypy `--strict` forces the annotation and deliberate evasion is not what the rule is for; and the cardinal-rules paragraph in `ARCHITECTURE.md` does not list the new rule, but it does not list the `_repository.py` naming rule either, so that is existing precedent.

scripts/check_architecture.py
peteski22: Reproduced: `Callable[[], AsyncSession]`, `async_sessionmaker[AsyncSession]` and the string form of the first were each flagged. Fixed in eac55715: the rule matches the session type at the top of the annotation, through `X | None`, Optional, Union and Annotated (first member only). New tests pin both the clean factory cases and the wrapped cases that must still fail. The baseline did not change.

scripts/check_architecture.py
peteski22: Reproduced: a function defined inside a module-level `if`, `try`/`except` or `with` reported clean. Fixed in eac55715: the rule checks every function defined at module scope, walking module-level statement blocks, except handlers and match cases, and still skips class bodies and nested functions. New tests cover `if`, `except` and `with`. The baseline did not change.

scripts/check_architecture.py
```diff
@@ -214,6 +218,253 @@ def _imported_modules(node: ast.Import | ast.ImportFrom, file_path: Path, src_ro
...
+            if not isinstance(node, ast.FunctionDef | ast.AsyncFunctionDef) or not _takes_session(node):
+                continue
```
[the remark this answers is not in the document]
peteski22: Reproduced: a function defined inside a module-level `if`, `try`/`except` or `with` reported clean. Fixed in eac55715: the rule checks every function defined at module scope, walking module-level statement blocks, except handlers and match cases, and still skips class bodies and nested functions. New tests cover `if`, `except` and `with`. The baseline did not change.

peteski22:
@coderabbitai review

scripts/check_architecture.py
```diff
@@ -214,6 +219,286 @@ def _imported_modules(node: ast.Import | ast.ImportFrom, file_path: Path, src_ro
...
+    return None
+
+
+def _annotation_names_session(annotation: ast.expr | None) -> bool:
+    """Return whether a parameter annotation is the async session type, optionally wrapped.
+
+    A session inside another type, such as a factory that returns one, is not a session.
+    """
+    if isinstance(annotation, ast.Name | ast.Attribute):
+        return _type_name(annotation) == SESSION_TYPE
```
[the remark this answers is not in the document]
peteski22: Reproduced with `from sqlalchemy.ext.asyncio import AsyncSession as DBSession` and `db: DBSession` (plain and `| None`): both reported clean. Fixed in a89fd1a5: the names a module binds to the session type by importing it from `sqlalchemy.ext.asyncio` are matched, and a same-named alias of any other type is not (tested both ways). The baseline is unchanged at 186.

PR #1293, opened 2026-09-17
fix(migrations): keep autogenerate from dropping tables another chain owns

### LINKED ISSUE #1274: Keep autogenerate from dropping tables that Otari's migration chain does not own
### Problem

`alembic/env.py` passes Otari's metadata to autogenerate with no `include_object` filter. Run `alembic revision --autogenerate` against a database that also holds tables from another migration chain, and the generated migration drops every table that is not on Otari's metadata.

An edition that keeps its own tables in the same database is the case this breaks. The migration looks like any other, so a reviewer can miss the drops.

### Proposal

Filter autogenerate so that it proposes no change to a table outside Otari's chain. A table that Otari's own models stop declaring must still get its drop. How the filter tells the two apart is for the PR to settle and state.

### Acceptance

- [ ] Autogenerate against a database that holds a foreign table emits no operation for that table.
- [ ] Removing a model from Otari's metadata still produces a drop for its table.
- [ ] A test covers both cases.

### Depends on

Nothing. Refs #921.

Part of #1171.

peteski22:
## Description
Otari keeps its database schema up to date with a chain of migrations, and developers write new migrations by asking the migration tool to compare the models with a real database. A database can also hold tables that belong to someone else's migrations, for example an edition that keeps its own tables beside Otari's. Today the comparison proposes dropping every one of those tables, and the generated migration looks like any other, so a reviewer can miss it.

After this change, the comparison leaves a table alone when it exists only in the database and no Otari migration ever created it. A table that Otari's own models stop declaring was created by an Otari migration, so it still gets its drop.

Nothing changes when Otari upgrades a database. This only affects generating and checking migrations.

### How it tells the two cases apart
Ownership comes from the migration chain itself. Every table a revision passes to `op.create_table`, or renames a table to with `op.rename_table`, belongs to Otari's chain. The names are read from each revision's source, without running it, so a revision must name its tables with a string literal or a module-level constant. All 48 tables in the chain do that today. If a future revision names a table any other way, generating or checking a migration stops with an error naming that revision, whatever tables the database holds, rather than guessing and either dropping another chain's table or keeping one Otari retired. The migration section of the backend skill now states that rule. The reading lives in `gateway.core.migration_chain`, where ruff, mypy and direct unit tests reach it; `alembic/env.py` keeps only the filter.

## How to test it locally
1. `uv run pytest tests/integration/test_autogenerate_foreign_tables.py` (needs Docker). One test creates a table and an index that no Otari revision creates, and asserts that autogenerate proposes nothing for them. It fails on `main` with `remove_index` and `remove_table` for that table. The other test empties Otari's metadata and asserts that autogenerate still proposes `remove_table` for every one of Otari's tables. That test also fails if ownership misses a table, which I confirmed by breaking the constant lookup.
2. `uv run pytest tests/unit/test_alembic_table_ownership.py` copies the migration chain, adds a revision that names its table with a function call, upgrades a SQLite database, and asserts that `alembic check` refuses the revision by name, both with and without a foreign table present. `uv run pytest tests/unit/test_migration_chain.py` calls the reader directly: every model table is owned by the real chain, each readable way of naming a table (literal, module constant, keyword, rename) is owned, and each unreadable one is refused.
3. By hand, against any PostgreSQL: `OTARI_DATABASE_URL=... uv run alembic upgrade head`, create any table by hand, then `uv run alembic check`. On `main` it lists that table under `remove_table`; on this branch it does not.

Also run: `make lint`, `make typecheck`, `make test-unit`, and `uv run --frozen --no-dev python scripts/oss_edition_smoke.py` all pass. `make test-integration` passes apart from `test_mcp_dependency_ceiling.py::test_mcp_constraint_resolves_to_an_importable_version`, whose throwaway virtual environment could not load its Python library on this machine; it does not touch migrations.

Found on the way, not changed here: `alembic check` against a freshly migrated PostgreSQL database on `main` already reports one difference, a unique constraint on `api_keys.key_hash` that the models do not declare. It is unrelated to this change and needs its own issue.

## PR Type
- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues
Fixes #1274

Part of #1171.

## Checklist
- [x] I ran the Definition of Done checks locally (`make lint`, `make typecheck`, `make test`). One integration test failed for a local environment reason, noted above.

## AI Usage
- [ ] No AI was used.

**AI Model/Tool used:**
Claude Opus 5 via Claude Code

Written by an agent working through the core cleanup epic, with a self-review against the repo's review skill before opening. The failing test was written and confirmed first.

**NOTE:**


alembic/env.py
```diff
@@ -139,6 +144,68 @@ def _reject_foreign_history(engine: Engine) -> None:
...
+def _include_object(
+    obj: SchemaItem, name: str | None, type_: NameFilterType, reflected: bool, compare_to: SchemaItem | None
+) -> bool:
+    """Leave out of autogenerate a table that exists only in the database and that this chain never created.
+
+    A database can hold another migration chain's tables beside otari's,
+    and autogenerate must not propose dropping them.
+    A table that otari's own models stop declaring was created by this chain, so it still gets its drop.
+    """
+    if type_ == "table" and reflected and compare_to is None:
```
khaledosman, line 163: The ownership set is read only when the database already holds a table the chain never created, so the refusal this PR promises ("generating a migration stops with an error naming that revision", and the new rule in the backend skill) never fires in the ordinary case: an otari-only database.

Confirmed locally by adding a revision that names its table with a call, upgrading a SQLite database, and running `alembic check` with no foreign table present: it reports `AutogenerateDiffsDetected` and `_tables_created_by` is never reached. The only thing that catches an unreadable revision today is `test_autogenerate_still_drops_every_table_the_models_stop_declaring`, which needs Docker.

Read the set unconditionally so the failure lands where the rule is broken:

```python
owned = _tables_this_chain_creates()
if type_ == "table" and reflected and compare_to is None:
    return name in owned
return True
```

`include_object` is called for every existing table on every autogenerate run (alembic `compare/tables.py:207`), and `@cache` keeps it to one parse.

alembic/env.py
```diff
@@ -139,6 +144,68 @@ def _reject_foreign_history(engine: Engine) -> None:
...
 
 
+def _table_name_argument(call: ast.Call, position: int, keyword: str) -> ast.expr | None:
+    """Return the argument a table operation received for one parameter, by position or by keyword."""
+    if len(call.args) > position:
+        return call.args[position]
+    return next((argument.value for argument in call.keywords if argument.arg == keyword), None)
+
+
+def _tables_created_by(script: Script) -> set[str]:
```
khaledosman: 45 lines of AST logic land in the one directory CI checks nothing in: `make lint` runs `ruff check src tests scripts` and mypy has `files = ["src", "tests", "scripts"]`, and both exclude `alembic/`. `env.py` also cannot be imported outside an alembic run (module-level `context.config`), so nothing can exercise `_tables_created_by` directly; both new tests go through `command.check`.

Moving the reader into a module ruff and mypy see, and a unit test can import, buys both. `scripts/check_alembic_heads.py` is the existing precedent for chain-reading logic that CI checks.

I ran `mypy --strict` and ruff over the file by hand and they are clean today, so this is about the next edit rather than this one.

khaledosman (approved):
Two points inline: the ownership read is lazy, so the refusal this PR documents never fires on an otari-only database, and the new logic sits in the one directory neither ruff nor mypy covers. Neither blocks.

Verified while reading: the table filter does suppress the foreign table's index too (alembic's removed-table branch, `compare/tables.py:165`, guards the index comparison as well), the reader resolves all 48 declared tables including the three `_TABLE` module constants, `@cache` is per command because alembic re-execs `env.py` without `sys.modules` caching, and the new unit test passes in 2.5s.

---
Review by Claude Opus 5 via [Claude Code](https://claude.com/claude-code)

peteski22:
@coderabbitai review

alembic/env.py
peteski22: Reproduced: with a revision that names its table with a call and no foreign table present, `alembic check` reported ordinary diffs and never refused. Fixed in 006ad1f1: the filter reads ownership on every call (cached), so the refusal no longer depends on what the database holds. The SQLite unit test now runs with and without a foreign table.

alembic/env.py
peteski22: Moved in 73fd57e2 to `gateway.core.migration_chain`, beside `core/database.py`, which already drives alembic, so ruff and mypy check it and env.py imports it. `tests/unit/test_migration_chain.py` calls it directly with no database: the real chain owns every model table, each readable way of naming a table is owned, and each unreadable one is refused by revision. env.py keeps only `_include_object`.

PR #1301, opened 2026-09-17
fix(dashboard): let the user guide use the whole page

### LINKED ISSUE #1299: The user guide renders as a two-column page whose second column is empty
## What happens

Opening Documentation on a wide window shows the guide pinched into a narrow column on the left, with a vertical rule running the full height of the page beside it and nothing at all after that rule. It reads as a two-column layout whose second column failed to load, and the guide itself, tables included, is squeezed into the first one.

## Why

`web/src/features/docs/DocsPage.tsx` wraps the guide in a flex row and gives the single child a right-hand rule:

```tsx
<div className="flex flex-1 border-t border-border">
  <div className="min-w-0 border-r border-border px-4 py-8 md:px-6">
    <Markdown className="max-w-[560px]" …>
```

Nothing is ever rendered after that child, so `border-r` is a divider with nothing on its far side. No other page does this: a page is `PageIntro` followed by `Section` bands, whose rules run the width of the scroll area and whose content sits in the page column.

## Expected

The guide reads as one ordinary page: no full-height divider, no empty second column, prose in the page column at a readable measure like every other surface.

khaledosman:
## Description

Opening Documentation showed the guide squeezed into a narrow strip on the left of a wide window, with a vertical line running the full height of the page beside it and nothing after that line. It read as a two column page whose second column had failed to load, and most of the window was empty.

There was never a second column. The guide now sits in the page's own band, the way every other page does: the horizontal rule runs the full width of the window, no line is drawn beside the text, and the guide fills the page column instead of stopping at a reading width of its own. Its tables, code blocks and headings now line up with the width everything else in the dashboard uses. Nothing about the guide's content changes.

One small removal comes with it: `PageIntro` had an option to narrow a page's opening sentence, which existed only so the guide's lede stayed inside that old reading width. The guide no longer needs it and nothing else used it, so the option, its story and its test are gone.

## How to test it locally

1. `make dashboard && uv run otari serve` (or `pnpm --dir web run dev`), sign in, and open **Documentation** in the top bar.
2. On a wide window the guide should read as an ordinary page: one rule under the title running edge to edge, the text using the width of the page, no vertical line and no empty panel to the right.
3. Narrow the window to phone width: the guide keeps the page's side margins and nothing scrolls sideways.

Covered automatically by `pnpm --dir web test` (`DocsPage.test.tsx` has a new case that the guide is a bleeding band with no vertical rule and no measure of its own), plus `pnpm --dir web run lint`, `pnpm --dir web run typecheck` and the Storybook story smoke. Verified by hand in a browser at 1920px and at 390px. The `/docs` screenshot entry already exists, and baselines are not committed, so there are no PNGs to update.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes #1299

## Checklist

- [x] I ran the Definition of Done checks locally: the dashboard's own (`pnpm --dir web run lint`, `run typecheck`, `test`). The Python suites are untouched by a `web/src` change.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

Dashboard-only change. Checks run were the dashboard's own (`lint`, `typecheck`, `test`); the Python suites are untouched by it.


khaledosman:
@coderabbitai review

PR #1305, opened 2026-09-17
fix(dashboard): move Providers into organization General settings

### LINKED ISSUE #1303: Move Providers into organization General settings
## Problem

Providers is separated from organization settings: standalone deployments show it under workspace Access, while hosted deployments show it under People & access.

## Expected behavior

Move Providers into the organization navigation, under General, immediately above Org settings.

## Acceptance criteria

- Standalone Providers appears under Organization > General, above Org settings, and is absent from workspace Access.
- Hosted organization provider keys appears in the same position instead of People & access.
- Existing URLs, provider behavior, surface gates, and operator permissions are preserved.
- Navigation works on desktop and mobile, including direct links.
- Update navigation coverage and the dashboard guide.

Epic: https://github.com/mozilla-ai/otari-ai/issues/2086

jigjigjig:
## Description

Providers now appears in organization navigation under General, immediately above Org settings. Standalone deployments move it out of workspace Access; hosted deployments move organization provider keys out of People & access. Existing URLs and permission gates stay the same.

The dashboard guide and navigation tests follow the new location, including mobile drawer behavior, direct links, and sidebar telemetry.

## How to test it locally

1. Start the gateway and dashboard, then sign in as an operator.
2. Open Organization and look under General. Providers should appear immediately above Org settings.
3. Open Providers and confirm the page loads with its navigation item selected. Return to the workspace and confirm Access no longer contains Providers.
4. Repeat on a phone-sized viewport. Hosted deployments should show their organization provider keys in the same General position.

Validation: dashboard lint, typecheck, production build, repository `make lint`, Vitest, and targeted Playwright navigation/provider-management flows. The full Vitest run identified one stale overlay navigation expectation; it was corrected and the navigation and shell suites rerun.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [x] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes #1303

Epic: https://github.com/mozilla-ai/otari-ai/issues/2086

## Checklist

Dashboard-only change: Python typecheck/tests and OpenAPI regeneration are not applicable.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:** Codex (GPT-6)

**Any additional AI details you'd like to share:** Implemented and checked in an isolated worktree. Draft requested for review.

## Summary

Providers now appears under **Organization > General**, immediately above **Org settings**, in standalone and hosted deployments.

- Removed Providers from workspace Access and hosted People & access.
- Preserved existing URLs, permissions, and provider behavior.
- Updated desktop and mobile navigation coverage, direct-link flows, telemetry tests, and dashboard documentation.


web/src/app/AppShell.test.tsx
```diff
@@ -585,6 +573,7 @@ describe("AppShell surface gating", () => {
       "Email domains",
       "Spend & budgets",
       "Model pricing",
+      "Providers",
```
jigjigjig, line 593: This snapshot now races. `Providers` is `operatorOnly: "refused"`, so `useNavVisibility` only admits it once `GET /organizations/me` resolves, while the guard above awaits `Org settings`, which paints with the first render and therefore guarantees nothing about this row.

Measured, not supposed: replace the guard's `findByRole` with `getByRole` and this assertion fails on exactly `- "Providers"`. It passes today only because `findByRole` happens to flush a cycle.

Await `Providers` instead: it is the last row to arrive, so it is the one the snapshot has to wait on. That is the same guard this PR correctly removed from the workspace test above, moved to the rail the row moved to.

web/src/app/nav/registry.ts
```diff
@@ -390,15 +365,22 @@ const ORGANIZATION_NAV_SECTIONS = [
...
-    // Keeps its heading with one row in it, where the index section at the top
-    // of the workspace rail has none. That is the same rule read in different
-    // surroundings rather than an exception to it: the index is first, with
-    // nothing above it to be absorbed into, and General is last under two
-    // labelled siblings, so a row with no heading here reads as the tail of
-    // Cost & billing. A heading earns its place when the section has labelled
-    // siblings, which is also why the deployment rail's one section has none.
     label: "General",
     items: [
+      // Each deployment reports the surface for its credential store.
```
jigjigjig: This drops the one sentence that made two identically labelled rows safe: "A deployment reports one surface or the other, never both, so exactly one of the two rows renders." The rows were on different rails before, so the ambiguity could not arise; now they are adjacent in one section with the same label and the same `FiBox`, and nothing in the file or the tests says a deployment cannot report both. The new `it.each` grants one surface per case, so it does not cover it either.

Restore the invariant on the pair, and say the two are different tables behind different endpoints (`provider_credentials` keyed on the instance, versus the tenant's own set) so the duplicate label reads as a deliberate either/or rather than a copy-paste.

web/e2e/dashboard.spec.ts
```diff
@@ -117,15 +117,8 @@ test.describe("dashboard core flows", () => {
...
-    for (const [link, heading] of [
-      ["Models", "Models"],
-      ["Providers", "Providers"],
-    ]) {
-      await nav(page).getByRole("link", { name: link }).click()
-      // Exact match: the Budgets onboarding heading ("No budgets yet") would
-      // otherwise also substring-match the page title.
-      await expect(pageHeading(page, heading)).toBeVisible()
-    }
+    await nav(page).getByRole("link", { name: "Models", exact: true }).click()
```
jigjigjig, line 120: Deleting the workspace loop took its `exact` rationale with it: "the Budgets onboarding heading (\"No budgets yet\") would otherwise also substring-match the page title." That case has not gone anywhere, it is in the organization loop below, and the `exact: true` it explains is baked into `pageHeading` in `helpers.ts`, which carries no comment at all.

Move the sentence onto `pageHeading`, where the decision now lives.

web/src/app/AppShell.test.tsx
```diff
@@ -248,12 +248,8 @@ describe("AppShell responsive layout", () => {
     const user = userEvent.setup()
     await renderShell()
 
-    const overview = screen.getByRole("link", { name: "Overview" })
-    // Awaited: Providers is one of the three rows gated `operatorOnly`, so it
-    // arrives with the membership context rather than with the first paint.
+    await user.click(await screen.findByRole("link", { name: "Organization" }))
```
jigjigjig, line 264: The comment above still says this pins that `"/"` lights only on the index "rather than on every page under it", but the test no longer looks at the Overview link or at `document.title`. Both assertions went, and the contract they covered did not change: `/` is still the index and still must not light up on a page under it.

Assert Overview's `aria-current` (and the title) while still on `/`, before clicking into the organization rail. Otherwise drop the comment, because it now describes a test that is not here.

web/src/app/nav/registry.ts
```diff
@@ -252,13 +242,9 @@ const BASE_NAV_SECTIONS = [
...
  * invariant. A surface gate says exactly this and needs no test change.
  *
- * Both have a page on the *workspace* rail that looks like them and is not:
- * `/providers` is this process's credentials, and `/tools/guardrails` is
- * what this process refuses. The organization ones are a tenant-wide credential
- * set and a ceiling over every workspace, which are different tables behind
- * different endpoints. Pointing the organization rows at the workspace pages
- * would put one destination on both rails, which `navContextForPath` cannot
- * express and `registry.test.ts` forbids.
+ * Both provider destinations live under General, with deployment surfaces
```
jigjigjig: This paragraph restates the one two above it (which already explains that a hosted deployment reports `organization_providers` in place of `providers`), and its second sentence names no row in this file any more, so it reads as a note about what the diff removed. Delete it and fold the one new fact, that both rows now sit under General, into the paragraph that already owns the surface split.

web/src/app/AppShell.test.tsx
```diff
@@ -731,23 +717,8 @@ describe("AppShell entitlement gating", () => {
...
-    // Riding on one row makes this case only as good as that row's gate, and a
-    // row losing its gate is a registry edit that says nothing about this file.
-    // So the premise is asserted here rather than described: with the axis
-    // undeclared the row appears for everyone, this case passes on a shell that
-    // no longer gates anything, and nothing points back at it. Measured rather
-    // than supposed: removing the gate and running `src/app` fails one test, and
-    // it is not this one.
     expect(navItemForPath("/providers")?.operatorOnly).toBeDefined()
-    await renderShell(bootstrap(), { operator: true })
+    await renderShell(bootstrap(), { operator: true, url: "/organization" })
```
jigjigjig, line 774: Fine to prune the long note, but `expect(navItemForPath("/providers")?.operatorOnly).toBeDefined()` a few lines up now reads as a stray: nothing says why a registry fact is asserted inside a rendering test. Keep one line for it, that this case is only as good as that row's gate, so the gate is asserted rather than assumed.

jigjigjig (commented):
Self-review. The move itself is right and matches #1303: order, surfaces, gates and URLs all hold, lint/typecheck/`src/app` vitest are green here. Six findings, one of them a proven test race.

Not a finding, but worth recording: `/providers` is linked from three workspace pages (`OverviewPage`'s `useAddProviderRoute`, `ModelDetailPage`, `PlaygroundGateNotice`), so "Add a provider" now swaps the whole rail to the organization one. Inherent to the move, not fixable here.

web/src/app/AppShell.test.tsx
jigjigjig: Fixed in 376835b0: the guard now awaits Providers, with a note that it is the row the caller axis holds back.

web/src/app/nav/registry.ts
jigjigjig: Fixed in 376835b0: the either/or invariant and the label rationale are back on the pair.

web/e2e/dashboard.spec.ts
jigjigjig: Fixed in 376835b0: the sentence moved onto `pageHeading` in helpers.ts, where the `exact: true` lives.

web/src/app/AppShell.test.tsx
jigjigjig: Fixed in 376835b0: the index assertions are back, taken before the click into the organization rail.

web/src/app/nav/registry.ts
jigjigjig: Fixed in 376835b0: the paragraph is gone, and its one new fact folded into the one above that owns the surface split.

web/src/app/AppShell.test.tsx
jigjigjig: Fixed in 376835b0: one line restored for why the registry premise is asserted here.

jigjigjig:
@coderabbitai the hosted organization-rail gap was real and is covered in 619c1455. `AppShell.test.tsx` now renders `HOSTED_SURFACES`, finds the General row, asserts its `href` is `/organization/provider-keys` and that it sits immediately above Org settings in the rendered order, then clicks through to the page.

Mutation-checked rather than assumed: swapping the two rows in the registry fails it with `expected [ "Org settings", "Providers" ] to deeply equal [ "Providers", "Org settings" ]`.

web/src/app/nav/registry.ts
```diff
@@ -390,15 +362,27 @@ const ORGANIZATION_NAV_SECTIONS = [
...
     label: "General",
     items: [
+      // Two rows, one label, and exactly one of them ever renders: a
+      // deployment reports the process-global surface or the tenant-scoped one,
+      // never both. They are different tables behind different endpoints, so
+      // the shared label is the honest one rather than a duplicate. "Providers"
+      // and not "Provider credentials" either way: the page manages the
+      // credential *and* the instance it belongs to, and the rail has one line.
+      {
+        to: "/providers",
```
khaledosman, line 374: This row declares deployment authority (`operatorOnly: "refused"`) but is now only reachable through a rail gated on organization authority. Both entrances to the organization rail, the desktop footer link and the mobile drawer button, sit behind `footerHasRowAboveBand` → `managesOrganization` = `canManage(organization.data)`, and `roles.ts` is explicit that superuser status is not an organization role. So a deployment operator who is not an owner or admin of the organization keeps the grant on this row and loses every rail path to it.

The other two `operatorOnly` rows avoid exactly this by living on the deployment rail, reached from the account menu, which gates on nav visibility alone. The docblock at L397 calls `/settings` "the running process", and process-global provider credentials are the same scope.

No effect in standalone today, since the local operator owns the organization the gateway provisioned. But the docblock at L207 says that gate becomes load-bearing at otari-ai#1716, and that is the point at which this row silently disappears for a caller the server would still serve. Worth either admitting a deployment operator to the rail entry, or noting here why this row is the one exception to the deployment-rail rule.

web/src/app/nav/registry.ts
```diff
@@ -231,7 +221,8 @@ const BASE_NAV_SECTIONS = [
  * provider credentials are the other, and that one is a *choice* rather than an
  * absence: the API and the page both exist, and a hosted deployment reports
  * `organization_providers` in place of the process-global `providers`, because
- * a credential keyed on an instance name alone is served to every tenant.
+ * a credential keyed on an instance name alone is served to every tenant. Both
+ * are rows under General, beside the row they belong with.
```
khaledosman, line 225: "Both" reads as the two rows the paragraph opened with, the organization guardrail ceiling and the organization's own provider credentials, and the guardrail row is in `org-gateway`, not General. Name the pair you mean: `providers` and `organization_providers` are the two rows under General.

web/e2e/helpers.ts
```diff
@@ -33,6 +33,8 @@ export const nav = (page: Page): Locator =>
  * what `nav()` does in the other direction, and it stays right however the rail's
  * markup changes.
  */
+// Exact: the Budgets onboarding heading ("No budgets yet") would otherwise also
```
khaledosman, line 36: This lands between `pageHeading`'s docblock and the declaration it documents, which AGENTS.md rules out ("the way it stops doing so is an insertion"). Fold the sentence into the docblock above instead.

khaledosman (approved):
Placement matches the issue and the gating is preserved on the entry itself. One thing to decide on: the row's reachability now composes with the organization rail's own role gate, which is a different axis from the one the row declares. Two comment nits alongside.

web/src/app/nav/registry.ts
jigjigjig: Fixed in 8081fd65: documented as the deliberate exception, since admitting the operator to the rail itself is a reachability change worth its own call rather than folding into this pass. The row now says why `operatorOnly` lives here instead of on the rail entry point, and names otari-ai#1716 as where it stops being silent.

web/src/app/nav/registry.ts
jigjigjig: Fixed in 8081fd65: names the pair directly (`providers` and `organization_providers`) instead of "Both".

web/e2e/helpers.ts
jigjigjig: Fixed in 8081fd65: folded into pageHeading's docblock.

PR #1306, opened 2026-09-17
refactor(core): add a Unit of Work, so a transaction starts and ends in one visible place

### LINKED ISSUE #1276: Add a Unit of Work, so each transaction starts and ends in one visible place
### Problem

A transaction has no owner.
A raw `AsyncSession` travels through routes and services, and whichever code holds it commits.

| Measure | Today |
| --- | --- |
| Route files that call `commit()` | 20 |
| Module-level service functions that take a session | 185, in 40 files |
| Manual `release_session` before dispatch | `api/routes/_pipeline.py:2888` |

A read opens a transaction on its first statement, and that transaction holds a pooled connection until something ends it.
That is why the request path calls `release_session` by hand before the provider call, and why a missed end leaks a connection (#1236).

### Proposal

A Unit of Work: one object that marks where a business transaction starts, commits and rolls back.
Sources: Martin Fowler, *Patterns of Enterprise Application Architecture* (2002), https://martinfowler.com/eaaCatalog/unitOfWork.html; Harry Percival and Bob Gregory, *Architecture Patterns with Python* (2020), ch. 6, https://www.cosmicpython.com/book/chapter_06_uow.

- The type lives in `core/unit_of_work.py`, beside `core/database.py`.
- Each request, and each worker job, has one Unit of Work. It wraps that scope's one session. Every domain service built in that scope shares it.
- A business step is one `async with uow:` block. The block commits when it ends and rolls back on an error. After either, the connection goes back to the pool.
- Blocks nest. An inner block joins the outer one, and only the outermost block commits, so a step that writes to two domains is atomic.
- A repository built on a Unit of Work raises when it is called outside a block. No transaction stays open by accident.
- Only a service opens a block. Repositories flush and never commit. Routes never open a block.
- A request gets its Unit of Work from a dependency over `get_db`. A worker creates one over `create_session()` or `create_log_session()`.
- Hybrid mode has no local database and gets no Unit of Work.

This issue adds the type, the request dependency and the worker helper.
It moves no existing code onto them, so behavior does not change.
#1202 is the first domain to use it.

### Acceptance

- [ ] A block commits on exit, and rolls back and re-raises on an error.
- [ ] A nested block joins the outer one, and only the outermost exit commits.
- [ ] A repository built on a Unit of Work raises when called outside a block.
- [ ] The connection is back in the pool after a block ends.
- [ ] A request dependency and a worker helper both produce a Unit of Work.
- [ ] No existing call site changes.

### Depends on

Nothing.

Refs #1262, #1203, #1236.

Part of #1171.

peteski22:
## Description
Today a database session travels through routes and services, and whichever code holds it commits. So nothing marks where a business step's transaction ends, and a session that has read something keeps a pooled connection until something ends its transaction.

This adds a Unit of Work, the object that marks where a business transaction starts, commits and rolls back. A business step is one `async with uow:` block:
- The block commits when it ends. It rolls back and re-raises when it ends on an error, cancellation included. Either way, the connection goes back to the pool.
- Blocks nest. An inner block joins the outer one, and only the outermost block commits, so a step that writes to two domains is atomic.
- If an inner block failed, the whole step is rolled back even when the surrounding code caught the error. The outermost block then raises `UnitOfWorkRolledBackError` from the first inner failure, instead of committing half the step.
- If the rollback itself fails, for example on a lost connection, the caller still sees the error that ended the step. The rollback failure is logged.
- A Unit of Work belongs to one task.
- A repository gets the session through `session_for(uow)`, and only inside a block. Outside one it raises `OutsideUnitOfWorkError`, so no transaction stays open by accident. The Unit of Work has no public session property, so code that holds one cannot reach the session without importing `session_for`. A later boundary check can allow that import only in `repositories/`. This PR does not add that check.

A request gets its Unit of Work from `get_unit_of_work`, a dependency over `get_db`. It wraps the request's own session, which the rest of the request also writes to, so until code moves into blocks, a block's commit or rollback also covers what that code staged. A worker gets one from `create_unit_of_work` (request pool) or `create_log_unit_of_work` (metering pool). A hybrid gateway has no local database, so it has no Unit of Work.

Nothing changes for someone running Otari. No existing code moves onto the Unit of Work. The one change to existing code is additive: `BaseRepository` also accepts a Unit of Work, and a repository built on a session behaves exactly as before. The existing tenancy repositories keep their session-only signatures until their own domain moves.

Decided in review:
- **Rolling back a step whose inner failure was caught.** Kept. The rolled-back step's error now names that failure as its cause.
- **How a repository reaches the session.** Through the function `session_for(uow)`, not a public property. A renamed property such as `session_for_repositories` was also considered. `scripts/check_architecture.py` checks imports, so it can limit who imports `session_for` to `repositories/`, the way it already limits who imports the feature registry. A property is reached by attribute access on any object that holds a Unit of Work, which an import check cannot see.
- **Refusing an already-active transaction at the outermost block.** Not done. The request's authentication dependencies read on the same session first, so a transaction is already active in practically every request.

## How to test it locally
1. `env -u OTARI_API_KEY uv run pytest tests/unit/test_unit_of_work.py` (16 tests):
   - commit on exit;
   - rollback and re-raise on an error and on cancellation;
   - rollback and re-raise on a failed commit;
   - a failed rollback keeps the error of a failed block, of a failed commit, and of a caught inner failure;
   - a nested block joins the outer one, and only the outermost commits;
   - a step whose inner failure was caught is rolled back and raises, with that failure as the cause;
   - the session is unavailable outside a block;
   - a repository built on a Unit of Work raises outside a block;
   - a repository built on a session needs no block;
   - both worker helpers produce a Unit of Work;
   - the request dependency wraps the request's own session.
2. `env -u OTARI_API_KEY uv run pytest tests/integration/test_unit_of_work_pool.py` (4 tests, PostgreSQL). After a block commits or rolls back, the connection is back in the real pool: request pool, metering pool, and the request dependency while the request is still open.
3. Every test above was written first and failed for its own reason before the code it covers existed.
4. Also run: `make lint`, `make typecheck`, `env -u OTARI_API_KEY make test-unit` (4091 passed, 10 skipped), the Unit of Work, tenancy repository, tenancy, invitation, domain, provider key, pool-release and race integration suites (218 passed), and `uv run --frozen --no-dev python scripts/oss_edition_smoke.py`.

## PR Type
- [ ] New Feature
- [ ] Bug Fix
- [x] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues
Fixes #1276

Part of #1171.

## Checklist
- [x] I ran the Definition of Done checks locally (`make lint`, `make typecheck`, `make test`). Lint, typecheck, the unit suite, and the integration suites that touch repositories and the pool.

## AI Usage
- [ ] No AI was used.

**AI Model/Tool used:**
Claude Opus 5 via Claude Code

Test-first against a skeleton with the final public names and no transaction behavior.

**NOTE:**

## Summary

- Added `UnitOfWork` transaction management for requests and workers.
- Added nested block handling, rollback protection, and connection release.
- Updated repositories to support `UnitOfWork` while preserving direct session behavior.
- Added request and worker helpers, including the log-session helper.
- Added unit and PostgreSQL integration tests.
- Existing application call sites remain unchanged.

## Benefits

Services can share one transaction within a request or worker job. Failed operations cannot leave partial commits or open pooled connections.

## Technical notes

Repositories backed by `UnitOfWork` reject access outside an active `async with` block. Only the outermost nested block commits. Test execution results were not provided.


src/gateway/core/unit_of_work.py
```diff
@@ -0,0 +1,108 @@
...
+    ) -> None:
+        self._depth -= 1
+        if exc is not None:
+            self._failed = True
+        if self._depth > 0:
+            return
+
+        failed, self._failed = self._failed, False
+        if failed:
+            await self._session.rollback()
```
khaledosman: `rollback()` is unguarded here, and again at line 84 in the commit path. When it raises, that database error is what the caller sees and the block's own exception survives only as `__context__` (checked: a block raising `ValueError` on a session whose rollback fails propagates the rollback error). A `TenancyError` carrying its own `status_code` then renders as a 500. The cases where the rollback fails are the ones where this matters most: connection lost, or cancelled mid-statement, which is the path the `cancel` test covers.

`release_session` in `core/database.py` already takes the other side of this for the same reason:

```python
with contextlib.suppress(*DATABASE_ERRORS):
    await session.rollback()
```

Same treatment here when `exc is not None`, and in the `except BaseException` arm, with a `logger.warning`.

src/gateway/core/unit_of_work.py
```diff
@@ -0,0 +1,108 @@
...
+            if exc is None:
+                msg = "The step was rolled back because a step inside it failed"
+                raise UnitOfWorkRolledBackError(msg)
```
khaledosman: The inner failure is dropped entirely: a caught `ValueError("inner failed here")` produces a `UnitOfWorkRolledBackError` whose `__cause__` and `__context__` are both `None`, and whose message names neither the exception nor the block it came from. That error is the only signal the caller gets, and it points at nothing.

Keep the exception rather than the bool (`self._failure: BaseException | None`), and `raise UnitOfWorkRolledBackError(msg) from first_failure`. Costs one attribute and makes the rolled-back step debuggable.

src/gateway/core/unit_of_work.py
```diff
@@ -0,0 +1,108 @@
...
+class OutsideUnitOfWorkError(RuntimeError):
+    """The session was asked for while no block of its Unit of Work was open."""
+
+
+class UnitOfWorkRolledBackError(RuntimeError):
+    """A step was rolled back because a block inside it failed, although the error was caught."""
+
+
+class UnitOfWork:
+    """One business transaction at a time over one session.
```
khaledosman, line 29: `_depth` is per-instance, not per-task, so one Unit of Work shared by two coroutines interleaves. Under `asyncio.gather(step_a(uow), step_b(uow))` the first block to exit sees depth 2 and does not commit; the second commits both. If either fails, the other is rolled back and gets `UnitOfWorkRolledBackError` for a step that succeeded. (Reproduced against a fake session: depth goes 1, 2, then A exits at 2 and B commits at 0.)

Same single-task rule as `AsyncSession` itself, but `AsyncSession` raises `InvalidRequestError` on overlapping I/O while this just draws the boundary in the wrong place. Worth a line in this docstring saying a Unit of Work belongs to one task, since `_pipeline.py` does gather.

src/gateway/api/deps.py
```diff
@@ -560,6 +561,14 @@ async def get_db_if_needed(
             yield db
 
 
+def get_unit_of_work(db: Annotated[AsyncSession, Depends(get_db)]) -> UnitOfWork:
```
khaledosman, line 564: Worth saying in this docstring that the block is over the **request** session, which unmigrated code writes to outside any block. While adoption is partial, a block's commit covers whatever else is already staged on that session and its rollback discards it, so a step is only as atomic as the rest of the request is disciplined.

`_bump_last_used_at` in this same file goes out of its way to avoid that ("never commits the caller's session or leaves it in a dirty state"), so the asymmetry is worth a sentence for whoever moves the first service onto this.

src/gateway/repositories/base_repository.py
```diff
@@ -38,16 +40,27 @@ class BaseRepository(Generic[ModelType, CreateSchemaType, UpdateSchemaType]):
         UpdateSchemaType: The schema describing an update payload.
     """
 
-    def __init__(self, db: AsyncSession, model_class: type[ModelType]):
-        """Bind the repository to a session and the table it serves.
+    def __init__(self, db: AsyncSession | UnitOfWork, model_class: type[ModelType]):
```
khaledosman, line 43: Every concrete repository still narrows this back to `db: AsyncSession` (`tenancy/organization_repository.py:17`, `user_repository.py:33`, `workspace_repository.py:26`, `invitation_repository.py:25`, `organization_member_repository.py:31`, `organization_domain_repository.py:20`, `org_provider_key_repository.py:76`), so a Unit of Work can be passed only to `BaseRepository` itself, which is what the test does. Widening those seven one-line signatures here would make the capability reachable from real code; left as is, it ships unusable and the first adopter pays for it in a second PR.

khaledosman (approved):
Reviewed at `d0274ba`. `make lint`, `make typecheck` and `tests/unit/test_unit_of_work.py` (12/12) are green on the branch here. Nothing blocking: nothing calls this yet.

On the two decisions you asked about:

- **Rolling back a step whose inner failure was caught.** Right call. On PostgreSQL a DB-level inner failure has already aborted the transaction, so committing stores a half-step at best. The gap is that there is no escape hatch for a genuinely *optional* sub-step; `session.begin_nested()` (SAVEPOINT) is the answer when one turns up, and it can be added later without changing these semantics. What the current shape does need is for the error it raises to name its cause (inline).
- **The public `session` property.** Fine. The guard is what the type cannot give you and it is there. `uow.session.commit()` is reachable from a service exactly as `commit()` is, so if that gap is worth closing it belongs in `scripts/check_architecture.py`, not in the type.

src/gateway/core/unit_of_work.py
peteski22: Fixed in ee10196f. Every rollback now goes through one helper that logs and suppresses a database error, as `release_session` does, so the caller sees the error that ended the step. Cancellation still propagates. Three tests cover it (a block error, a failed commit, and a caught inner failure, each with a failing rollback), and each failed before the fix.

src/gateway/core/unit_of_work.py
peteski22: Fixed in cbbc1ce5. The Unit of Work keeps the first inner failure and raises `UnitOfWorkRolledBackError` from it. A test asserts `__cause__` is that failure, and it failed before the fix.

src/gateway/core/unit_of_work.py
peteski22: Addressed in 6f7ade03. The class docstring now says a Unit of Work belongs to one task, because blocks that concurrent tasks open on it would commit each other's steps.

src/gateway/api/deps.py
peteski22: Addressed in 6f7ade03. The docstring now says the rest of the request writes to the same session, so a block's commit also stores what that code staged outside a block, and its rollback discards it.

src/gateway/repositories/base_repository.py
peteski22: Not changed here. No caller passes a Unit of Work to those seven repositories yet. The first domain to adopt the Unit of Work, budgets, has no repository today and builds new ones on `BaseRepository`, typed for a Unit of Work from the start. The tenancy repositories widen in their own domain step, when a tenancy service is built with a Unit of Work, and that step also moves them into the domain's repository package. Widening them now would accept a type that nothing passes.

src/gateway/core/unit_of_work.py
```diff
@@ -0,0 +1,108 @@
...
+        failed, self._failed = self._failed, False
+        if failed:
+            await self._session.rollback()
+            if exc is None:
+                msg = "The step was rolled back because a step inside it failed"
+                raise UnitOfWorkRolledBackError(msg)
+            return
+
+        try:
+            await self._session.commit()
```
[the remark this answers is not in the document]
peteski22: Not changed. The observation is right, and 6f7ade03 documents it on `get_unit_of_work`. The proposed fix does not fit this request path. The authentication dependencies read through the same `get_db` session before a route runs, and a session begins a transaction on its first statement, so a transaction is already active in practically every request. Refusing an active transaction at the outermost entry would make the request Unit of Work unusable. A dedicated session per Unit of Work would give a request two sessions and two pooled connections, and writes outside a block could no longer share its transaction. The design is one Unit of Work per request over that request's one session. The gap closes as each domain moves its writes into blocks.

peteski22:
@coderabbitai review

PR #1310, opened 2026-09-17
chore(lint): keep sqlalchemy and sqlmodel out of services and routes

peteski22:
## Description
The backend is moving to a shape where services and routes never touch the database directly: a repository runs every query, and a service is built with its repositories and a Unit of Work, never the session. Two lint rules already guarded part of that. One caught a route or service that built a query, and one caught a service function that took a session as a parameter. Both left gaps. A service class that takes the session in its constructor passed, and each rule kept its own list of known offenders.

This replaces both with one simpler idea, applied to two layers. Nothing under `services/` imports `sqlalchemy` or `sqlmodel`, and nothing under `api/routes/` does either. Any import counts, including one used only for type checking, so the constructor case fails too. Code that already imports either library is named on a per-layer list that may only shrink: when a module stops importing, the check fails until its name is removed.

### Architecture rule
This replaces rules 11 and 12 in `scripts/check_architecture.py` with two import rules that take their numbers: rule 11, nothing under `services/` imports `sqlalchemy` or `sqlmodel`; rule 12, nothing under `api/routes/` does. Why: an import rule has one baseline per layer, covers every way code reaches the database, and closes the gap where a service class took the session in `__init__`. Rules 3 to 6 of #1203 remain.

What the rule guarantees: a service can neither build a query nor name the session type. It does not by itself stop a service reaching a session. With the Unit of Work (#1306), a repository gets its session through `session_for(uow)` from `gateway.core.unit_of_work`, which raises outside a block, and a later rule will limit importing `session_for` to `repositories/`. This PR does not add that rule.

### Why `sqlmodel` counts
SQLModel re-exports `select` and `Session`, so a rule on `sqlalchemy` alone would let `from sqlmodel import select` build a query, which the old query rule caught. Counting it costs nothing today: every module that imports `sqlmodel` also imports `sqlalchemy`, so no module joins a baseline because of it.

### Measured
On `main` at `bad022e4`, the parent of this commit.

| Measure | Before | After |
| --- | --- | --- |
| Service modules that import `sqlalchemy` or `sqlmodel` | 49 of 108 (not counting `__init__.py`) | 49, all on the service baseline |
| Route modules that import `sqlalchemy` or `sqlmodel` | 61 of 72 (not counting `__init__.py`) | 61, all on the route baseline |
| Baselines the check keeps for this concern | 2 (57 modules for queries, 186 functions for sessions) | 2 (49 service modules, 61 route modules) |

The route rule that forbids `sqlalchemy.orm` stays, so a route already on the baseline still cannot add the sync ORM.

Rule 13 (#1311) keeps its number and its baseline. The docs that described rules 11 and 12 now describe the import rules beside it: the lint sentence in `AGENTS.md`, rules 1 and 2 and "What the boundary check enforces" in `ARCHITECTURE.md`, the layering section of the backend standards skill, and the backend architecture review instructions. Rules 1 and 2 in `ARCHITECTURE.md` now name `sqlmodel` too, so the doc and the check agree.

### What this trades
A baseline entry exempts its whole module, where rules 11 and 12 were narrower. Measured on this branch against `main` at `2cc9cc0b`, and unchanged at `bad022e4`:

- **Queries.** The two new baselines hold 110 modules and include all 57 on `QUERY_BASELINE`. So 53 modules that rule 11 stopped from building a query may now build one: 43 routes and 10 services. Adding `from sqlalchemy import select` to `api/routes/_normalize.py` fails on `main` and passes here.
- **Sessions.** `SESSION_PARAMETER_BASELINE` named functions, so a new session-taking function failed even in a listed module. Here any of the 49 service modules on the baseline can add one. Adding `async def probe_new_function(db: AsyncSession)` to `services/budget_service.py` fails on `main` and passes here.

Why the trade: one baseline per concern ([#1203, decided](https://github.com/mozilla-ai/otari/issues/1203#issuecomment-5715194096)), not one per pattern. Each baseline only shrinks, and each domain PR removes its modules. New code is checked in full: a new module is on no baseline, and #1311 refuses new flat modules. The gap is additions inside old baselined modules. The backend architecture review instructions flag a new session-taking function there, and a new query anywhere, so review covers it until the module leaves its baseline.

## How to test it locally
2. Add `from sqlmodel import select` to a service that is not on the baseline, for example `src/gateway/services/tool_format.py`, and run `make check-architecture`. It fails with "imports sqlmodel; a service reaches the database through its repositories and the Unit of Work". Revert it.
3. Remove every `sqlalchemy` and `sqlmodel` import from a module on a baseline, or delete the module, and the check fails with "is on the database import baseline but imports no database library; remove it from the baseline".
4. `uv run pytest tests/unit/test_check_architecture.py` covers `from sqlalchemy import`, `from sqlmodel import`, `import sqlalchemy as sa`, `from sqlalchemy.ext.asyncio import AsyncSession`, and an import only under `TYPE_CHECKING`, in a service, a service subpackage and a route. It also covers a service class that takes an `AsyncSession` in `__init__`, a repository staying free to import, an unrelated module whose name starts with `sqlalchemy`, the per-layer baseline, a stale entry for both a module that stopped importing and one that no longer exists, and `main()` failing.

## PR Type
- [ ] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [x] Infrastructure / CI

## Relevant issues
Refs #1203

Part of #1171.

## Checklist

## AI Usage
- [ ] No AI was used.

**AI Model/Tool used:**
Claude Opus 5 via Claude Code

**NOTE:**


peteski22:
@coderabbitai review

scripts/check_architecture.py
```diff
@@ -577,36 +272,21 @@ def check_file(file_path: Path, src_root: Path) -> list[tuple[int, str, str]]:
...
-    "gateway/api/routes/scoped_budgets.py",
-    "gateway/api/routes/usage.py",
-    "gateway/api/routes/users.py",
+DATABASE_LIBRARIES = ("sqlalchemy", "sqlmodel")
+SERVICE_SCOPE = "gateway/services"
+ROUTE_SCOPE = "gateway/api/routes"
+# Modules that imported a database library when rules 11 and 12 landed. An
+# entry that stops importing one fails the check until it is removed, so each
+# list only shrinks.
+SERVICE_DATABASE_IMPORT_BASELINE = (
```
khaledosman, line 285: One baseline per module is coarser than what it replaces, so this loses enforcement the description does not account for.

**Queries.** `QUERY_BASELINE` exempted 57 modules. The two new baselines exempt 110 (49 services + 61 routes) and are a strict superset of it, so 53 modules (43 routes, 10 services) may now build a query where the check previously refused one.

**Sessions.** `SESSION_PARAMETER_BASELINE` was per function, so a *new* module-level function taking a session failed even inside an already-listed module. Per module, all 49 baselined services can now add one silently.

Both verified by appending to a baselined module and running the check on each branch:

```
# src/gateway/services/budget_service.py
async def probe(session: AsyncSession) -> None:
    await session.execute(select(1))
# main:        ❌ ... probe takes a session
# this branch: ✅ No architecture violations found

# src/gateway/api/routes/workspaces.py
from sqlalchemy import select
def probe():
    return select(1)
# main:        ❌ ... builds a query with select
# this branch: ✅ No architecture violations found
```

The import rule closes the constructor gap and is worth adding, but it does not subsume the other two. Keep `check_query_layering` and `check_session_parameters` with their baselines beside it until those baselines are empty. If they do go, say so: "covers every way code reaches the database" is true of the rule and false of its coverage, and the docstring at line 20 ("a service can neither build a query nor name the session type") holds only off the baseline.

.github/instructions/backend-architecture.instructions.md
```diff
@@ -47,8 +47,6 @@ same" as a reason.
   that receives the session or another domain's repository.
 - A builder in `api/deps.py` builds the service. Flag a new service builder
   defined anywhere else.
-- Flag a new module-level function under `services/` that takes an
```
khaledosman, line 50: This bullet is the review backstop for exactly the case the check stops catching (see the comment on `check_architecture.py`): a new module-level session-taking function inside one of the 49 baselined service modules. Automated and human coverage for it are removed in the same change. Keep the bullet whichever way that comment goes. The Services row at line 33 says "takes or holds a session", which covers the service class but not the module-level functions this named.

khaledosman (changes requested):
The import rule is a real improvement in kind, but as a *replacement* it is coarser than the two rules it retires, and the enforcement it drops is not mentioned in the description. Details inline.

Checked and clean: `make check-architecture` passes, `tests/unit/test_check_architecture.py` (75) passes, Ruff and format clean, no stale references to `QUERY_BASELINE` / `SESSION_PARAMETER_BASELINE` anywhere in the tree, docs consistent, no em dashes, conventional title, branch level with `main`.

khaledosman (approved):

.github/instructions/backend-architecture.instructions.md
peteski22: Fixed in 97214936. The bullet is back and names the baselined case: flag a new module-level function under `services/` that takes an `AsyncSession`, including one added to a module already on `SERVICE_DATABASE_IMPORT_BASELINE`, which the check does not catch. The Services row now says "takes or holds a session in a class or in a module-level function".

scripts/check_architecture.py
peteski22: Verified on this branch against main at 2cc9cc0b. The new baselines hold 110 modules and include all 57 on `QUERY_BASELINE`, so 53 more (43 routes, 10 services) may now build a query. Any of the 49 baselined service modules can add a session-taking function. Both probes fail on main and pass here.

The user decided to keep the trade: import rules only, one baseline per concern (https://github.com/mozilla-ai/otari/issues/1203#issuecomment-5715194096). New code is fully covered: a new module is on no baseline and is checked in full, and #1311 refuses new flat modules. The gap is only additions inside old baselined modules, which each domain PR moves out. The review rule from the other thread covers that gap, kept in 97214936. The PR description records the numbers under "What this trades".

PR #1319, opened 2026-09-17
test(lint): make every main() test fail only when its own rule is unwired

peteski22:
## Description
The architecture check has one test per rule that runs the whole check and expects it to fail. Two of those tests passed even with the rule they name removed from the check.

Both build a small fake repository in a temporary folder. The check's baselines list real modules, none of which exist in that folder, so every entry counted as stale and failed the check on its own. The test then saw a failure and passed, whatever its own rule did.

Each of these tests now goes through one helper that points the check at the temporary folder and empties every baseline, including any baseline added later. Each test first checks that the clean folder passes, then writes the thing it is about and expects the failure.

This changes tests only. Nothing changes for someone running Otari, and no rule changes.

## How to test it locally
1. `uv run pytest tests/unit/test_check_architecture.py` passes (85 tests).
2. To see the weakness on `main`: in `scripts/check_architecture.py`, replace `import_violations` with an empty list at the end of `main()`, and `test_main_discovers_tests_root_and_fails_on_overlay_import` still passes there. Do the same for `package_violations` and `test_main_fails_on_a_new_top_level_package` still passes.
3. On this branch, each of the four `main()` tests fails when, and only when, its own rule is removed. I checked all sixteen combinations:

| Rule removed from `main()` | overlay import | top-level package | database import | flat module |
| --- | --- | --- | --- | --- |
| imports | fails | passes | passes | passes |
| top-level packages | passes | fails | passes | passes |
| database imports | passes | passes | fails | passes |
| flat modules | passes | passes | passes | fails |

Also run: `make lint` and `make typecheck` pass.

## PR Type
- [ ] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [x] Infrastructure / CI

## Relevant issues
Part of #1171.

## Checklist
- [x] I ran the Definition of Done checks locally (`make lint`, `make typecheck`, `make test`). Lint, typecheck and the check tests; this PR touches one test file.

## AI Usage
- [ ] No AI was used.

**AI Model/Tool used:**
Claude Opus 5 via Claude Code

**NOTE:**

## Summary

- Added a shared helper for architecture-check tests.
- Updated affected tests to use temporary repositories with cleared baselines.
- Each test now confirms the clean repository passes and the relevant violation fails the check.
- This prevents stale baseline data from masking unwired rules.
- Runtime behavior and architecture rules remain unchanged.


peteski22:
@coderabbitai review

tbille (approved):

PR #1321, opened 2026-09-17
chore(lint): let only the Unit of Work end a transaction, and only a repository reach the session

peteski22:
## Description
A database transaction should start and end in one visible place. The Unit of Work is that place: a service opens a block, the block commits when it ends and rolls back on an error, and a repository reaches the session through it. Nothing stopped new code from committing on its own, or from taking the session and running a query outside a repository.

Two lint rules close that. A `commit()` or `rollback()` call anywhere under `src/gateway/` now fails the check, except in the Unit of Work itself, unless the module is one of the 56 that call one today. That list only shrinks: when a module stops calling them, the check fails until its name is removed, so each domain PR records its progress by deleting names. And an import of `session_for` outside `repositories/` fails, which needs no list because the base repository is its only importer.

### Architecture rule
This adds rules 14 and 15 to `scripts/check_architecture.py`. Rule 14: only the Unit of Work calls `commit()` or `rollback()`, from a shrinking baseline. Rule 15: only `repositories/` imports `session_for`. Why: these are rule 3 of the modular monolith's layer and import rules, which #1203 lists, and #1306 has now landed the Unit of Work they depend on. Rules 4 to 6 are not in this PR.

### Two decisions this makes
**What counts as a call.** An attribute call named `commit` or `rollback` on anything. No type inference, and no allowlist of receivers. Every one of the 259 such calls under `src/gateway/` today is on a database session (`db`, `self.db`, `session`, `self._session`, `ctx.db`), so the simple rule has no false positives here. If a non-session object ever grows a `commit` method, its module joins the baseline or the rule gains an exception, and either is a visible, reviewed edit.

**What is exempt.** Only `core/unit_of_work.py`, which is the rule's subject. `core/database.py` is on the baseline rather than exempt: its pair of calls is `release_session`, which commits before dispatch to hand the pooled connection back early, and the Unit of Work removes the need for it, so it should leave the baseline rather than sit outside the rule forever. `adapters/telemetry_storage_adapter.py` is on the baseline for the same reason: it is code that has not moved yet, not a second place where transactions may end.

### Measured
On `main` at `25ad3998`, the parent of this commit.

| Measure | Before | After |
| --- | --- | --- |
| Modules under `src/gateway/` that call `commit` or `rollback` | 57, with 259 calls | 56 on the baseline, plus the Unit of Work itself |
| Where they sit | `services/` 31, `api/` 23, `core/` 2, `adapters/` 1 | `core/` 1 after the Unit of Work is excluded |
| Modules that import `session_for` | 1 (`repositories/base_repository.py`) | 1, and no other module may |
| New modules that may do either | Any | None |

The baseline constant is named `TRANSACTION_CONTROL_BASELINE`.

## How to test it locally
2. Add `await db.commit()` to any service that is not on the baseline, for example `src/gateway/services/tool_format.py`, and run `make check-architecture`. It fails with "calls commit; only a Unit of Work block ends a transaction". Revert it.
3. Remove the `commit` and `rollback` calls from a module on the baseline, and the check fails with "is on the transaction control baseline but calls neither commit nor rollback; remove it from the baseline".
4. Add `from gateway.core.unit_of_work import session_for` to any service or route and the check fails with "Forbidden import in Repositories only". The same import in a repository passes.
5. `uv run pytest tests/unit/test_check_architecture.py` covers a commit and a rollback in a function and in a method, the Unit of Work staying free to call both, the baseline exemption, a stale entry for both a module that stopped calling them and one that no longer exists, `session_for` refused in a service, a route and a core module, allowed in a repository, the Unit of Work type itself staying importable anywhere, and `main()` passing on a clean tree then failing on a commit. Removing rule 14 from `main()` makes that last test fail, and leaves the other `main()` tests passing.

## PR Type
- [ ] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [x] Infrastructure / CI

## Relevant issues
Refs #1203

Part of #1171.

## Checklist

## AI Usage
- [ ] No AI was used.

**AI Model/Tool used:**
Claude Opus 5 via Claude Code

**NOTE:**


peteski22:
@coderabbitai review

tbille (approved):

PR #1367, opened 2026-09-18
fix(dashboard): use rem for confirmed arbitrary widths

### LINKED ISSUE #1342: Two arbitrary values are written in px where the rule is rem, and six look like it and are not
Two arbitrary values are written in `px` where the rule asks for `rem`. Six more look like it and are not, which is most of what this issue is for: the discriminator is whether the value should follow the reader's root font size, and a grep cannot tell.

## The rule

`.github/skills/frontend-standards/responsiveness.md`:

> **`rem`, not `px`, in arbitrary values.** `h-[20rem]`, not `h-[300px]`, because `rem` follows the reader's root font size and `px` overrides an accessibility setting they chose deliberately. A `1px` hairline border is the exception. A `text-[11px]` is a second exception that should not exist: that size is `text-overline` in the type scale, so use the role.

## Convert these two

| File | Line | Value | Why |
| --- | --- | --- | --- |
| `features/models/PublicCatalogPage.tsx` | 21 | `h-6 w-[26px]` | Not a unit conversion: `26px` and `1.638rem` are different lengths. The mark is 273x250, so at `h-6` (24px) its true width is `24 x 1.092 = 26.208px = 1.638rem`, which is what the twin at `features/auth/LoginPageShell.tsx:27` already carries. `26px` rounds the aspect to 1.083 and squashes the mark by about 0.8%. Check the viewBox rather than copying the string. |
| `features/budgets/BudgetsPage.tsx` | 428 | `min-w-[140px]` | A floor on a column of text, which is the case the rule names in rem: "`min-w-[11.25rem]` on a wrapping stat card is fine: it is a floor, not a fixed width". At 140px the text inside it grows and the floor does not. |

## Two worth a decision rather than a conversion

`features/overview/OverviewPage.tsx:1048` and `:1060` both set `h-[180px]`: the y-axis lane and the plot beside it, which have to stay equal or the axis labels stop lining up with the bars.

The argument for rem is that the lane holds text (`formatAxisUsd` values) and a reader who has scaled their font up gets the same 180px to fit taller labels into. The argument against is that a chart is a drawing, and a plot that grows with the root font while its bar count does not changes the shape of the thing being read.

Whichever way it goes, both sites move together. Worth answering once here rather than per site.

## Leave these four alone

They match the pattern and are not what the rule is about. Named so a later sweep does not "fix" them:

| File | Line | Value | Why it is px |
| --- | --- | --- | --- |
| `features/usage/ShareDialog.tsx` | 433 | `-left-[9999px]` | Not a measurement. It is the off-screen position the share card is rendered at to be rasterized, since `display: none` has no layout to capture. Scaling it with the root font means nothing. |
| `features/overview/OverviewPage.tsx` | 1060 | `gap-[3px]` | The gap between chart bars. A drawing detail, in the same family as the hairline border the rule already exempts. |
| `design-system/forms/Toggle.tsx` | 59 | `translate-x-[21px]` | The thumb's travel, which has to land exactly on a track `design/forms.md` specifies in pixels (44x24). In rem the travel and the track desynchronize at any non-default root size and the thumb overshoots its own track. |
| `features/usage/ShareDialog.tsx` | 316 | `sm:w-[340px]` | The preview pane, sized to a PNG whose dimensions are fixed in pixels (`CARD_SIZES`). The preview is documented as byte-for-byte what gets posted. |

## What `rem` currently buys here, which is less than this issue implies

`globals.css:1236` pins `:root` to `font-size: 16px`, unlayered, so `rem` resolves to the same
length as the `px` it replaces for every reader. The conversions above are still right, and the
reason is in #1368: while the tree is consistently in `rem`, honoring the reader's preference is
a one-line change rather than a sweep. But `responsiveness.md`'s stated reason ("`rem` follows
the reader's root font size") is not true today, and a reviewer is entitled to ask.

## Correction to the original issue

This issue was first written off a grep, and two of its claims were wrong. A third, the
"straight copy" above, was imprecise and is corrected in the table rather than here, because it
would have sent someone to copy a string when the thing to check was the mark's aspect ratio.

It said `ShareDialog.tsx:433`'s `[9999px]` "is a pill radius rather than a measurement" and that "`rounded-full` says it without a number". It is `fixed -left-[9999px]`, a position. Acting on that line would have replaced an off-screen render position with a border radius and broken the share preview.

It also presented all eight as a single pass. Six of the eight are not conversions at all, so the work here is two lines plus one decision, not a sweep.

## Found in

A standards audit of `web/`, corrected after the same grep-not-the-file method produced four false positives in #1348.

LejeuneA:
## Description

Converts the two confirmed arbitrary pixel widths to their rem equivalents so they scale with the reader's root font size and follow the frontend responsiveness standard.

The chart-height decision and the explicitly exempt pixel usages are left unchanged.

## How to test it locally

This change was made through the GitHub web editor, so I did not run the local Definition of Done commands.

Review the public catalog mark and the budget spend cell and confirm their sizing remains visually consistent while using rem-based widths. CI can provide the automated validation.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes #1342

## Checklist

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

ChatGPT

AI was used to help review the issue, verify the intended scope, and draft the small code changes. I reviewed and applied the changes myself.

**NOTE:**


khaledosman (approved):

khaledosman:
Looks good 🚀 , thank you @LejeuneA

Issue #1374, opened 2026-09-18
Region-tag API keys and gateway tokens as untrusted routing hints, with 421 fail-fast


tbille:
## Summary

The gateway side of mozilla-ai/otari-ai#1665. Hosted otari.ai is moving to a credential envelope that carries a region tag, so a key presented to the wrong regional stack fails fast with `421 Misdirected Request` naming the correct host, instead of a confusing auth error. The tag is an untrusted routing hint only; the hash lookup remains the trust boundary.

This is a wire contract the gateway consumes, so the canonical contract lives here in `docs/hybrid-mode-protocol.md` and the two issues stay in sync.

## Hosted key format

`otr_<kind>_v1_<region>_<secret><checksum>`

- `otr` is the product prefix; `tk` alone is too generic for secret scanning.
- `<kind>` is `tk` for user API keys and `gw` for gateway tokens.
- `v1` is the format version.
- `<region>` is `us` or `eu`; only `us` is live today.
- `<secret>` is 43 base62 characters (256 bits).
- `<checksum>` is a CRC32 of everything before it, as 6 base62 characters, so a fake key is rejected without a database lookup.

## Rules

- The region only routes; it never grants access. The whole key is still hashed and looked up.
- Legacy keys keep working by hash: `gw-…`, `tk_…`, `tk_live.…`, and OSS `tk-…`. A key is never rejected only because of its shape (#646 took the shape check off the verify path for this reason).
- An `otr_` key with a bad checksum, an unknown region, or the wrong kind for its header gets `401`, with no database lookup.
- An `otr_` key for another region gets `421 Misdirected Request`, with a body naming the correct host.

## Scope for this repository

- An `ApiKeyFormatPort` with a default adapter that mints the open-source format and routes every key locally, so OSS behavior does not change; every mint site and the verify path go through it.
- Hybrid mode forwards a `421` from the platform's resolve, with its detail, and the protocol doc describes it.
- The dashboard's one-time reveal is masked from the stored `key_prefix` and `key_suffix`, not a fixed slice of the plaintext.

The hosted adapter itself lands in otari-ai's overlay once a release carrying the port is pinned there.

## Out of scope

- Registering the `otr_` pattern with GitHub secret scanning.
- Deciding when legacy keys must rotate.

PR #1326, opened 2026-09-17
fix(keys): mint user API keys with the tk- prefix instead of gw-

tbille:
## Description
Creating an API key from the dashboard, the CLI bootstrap, or the API handed out a key starting `gw-`. That prefix is the one the hybrid-mode protocol reserves for the gateway's own credential to the platform (`X-Gateway-Token`); end-user keys are the `tk` family (`X-User-Token`). Every key the gateway mints now starts `tk-`. Existing `gw-` keys keep working, because authentication hashes the presented key and never inspects its prefix.

The docs and dashboard copy that display a user key now show `tk-`. Gateway-token examples keep `gw_`.

## How to test it locally
1. Start the gateway on an empty database (`otari serve` or `docker compose up`). The bootstrap key printed once at startup starts `tk-`.
2. Open the dashboard, go to Keys, create a key. The reveal dialog shows `tk-…`, and so does the curl snippet under it.
3. Regenerate that key: the new secret also starts `tk-` and still authenticates a chat completion.

Covered by automated checks: `tests/unit/test_auth_models.py` (minted prefix, validator), `tests/unit/test_gateway_bootstrap.py` (bootstrap key), `tests/integration/test_key_management.py` (create and rotate), `web/e2e/parity.management.spec.ts` (dashboard reveal), and `scripts/oss_edition_smoke.py` (end-to-end mint against the packaged CLI).

## PR Type
- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [x] Documentation
- [ ] Infrastructure / CI

## Relevant issues
None filed; reported from the hosted dashboard's key-creation dialog.

## Checklist
- [x] I understand the code I am submitting.
- [x] I have added or updated tests that cover my change (`tests/unit`, `tests/integration`).
- [x] I ran the Definition of Done checks locally (`make lint`, `make typecheck`, `make test`).
- [x] Documentation was updated where necessary.
- [x] If the API contract changed, I regenerated the OpenAPI spec (`uv run python scripts/generate_openapi.py`). (Not needed: no route or schema changed.)
- [x] If this changes a rule in `ARCHITECTURE.md` or `scripts/check_architecture.py`, the description names the rule and says why. (Not applicable.)

## AI Usage
- [ ] No AI was used.
- [x] AI was used for drafting/refactoring.
- [ ] This is fully AI-generated.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary

- Changed newly generated user API keys from `gw-` to `tk-`.
[...]

PR #1375, opened 2026-09-18
feat(keys): put the API key format behind a port

tbille:
## Description
Each edition of Otari can now decide what its API keys look like. A new `ApiKeyFormatPort` owns three things: minting a key, the fingerprint stored as `key_prefix`, and where a presented key is checked. The open-source adapter mints the `tk-` shape from #1326 and checks every key locally, so nothing changes for a standalone gateway. A hosted overlay can bind a format that carries a region and a checksum, so a key presented to the wrong regional stack is answered `421 Misdirected Request` naming the right host, and a key that claims the hosted format but fails its checksum is answered `401`, both without a database lookup. Legacy keys of every shape (`gw-`, `tk_`, `tk_live.`, OSS `tk-`) keep authenticating by hash.

In hybrid mode, a `421` from the platform's resolve is forwarded to the caller with its detail. The dashboard's one-time reveal now masks the key with the `key_prefix` and `key_suffix` the create response carries, instead of slicing eight characters off the plaintext, so it matches the Keys table whatever length a format gives its prefix.

Built on top of #1326, whose three commits appear in this diff until it merges; the test workflows only run against `main`, so the base is `main` rather than the stacked branch. This is PR 1 of the two mozilla-ai/otari-ai#1665 asks for. The otari side is tracked in #1374, filed for this work.

## How to test it locally
1. `otari serve` on an empty database: the bootstrap key printed once still starts `tk-`, and its row on the Keys page shows a ten-character fingerprint.
2. Create and regenerate a key from the dashboard: the concealed reveal reads `tk-XXXXXXX••••••••YYYY`, the same halves the table shows after you dismiss it.
3. Send a completion with a `gw-` key whose hash is on a row (or run `tests/integration/test_migrated_key_auth.py`): it still authenticates.
4. To see the port swapped, point `OTARI_BOOTSTRAP` at a module whose `register` binds `ApiKeyFormatPort` to an adapter of your own. `tests/integration/test_api_key_format_rebound.py` does exactly this with a probe adapter and walks create, rotate, the 421, the 401 and a legacy key.

Automated coverage: `tests/unit/test_api_key_format_adapter.py` (the default adapter behaves as before), `tests/unit/test_api_key_routing.py` (Misdirected and Malformed do no lookup, Local reaches it, hybrid forwards the platform's 421), `tests/unit/test_gateway_bootstrap.py` (the first-run key follows a rebound port), `tests/integration/test_api_key_format_rebound.py` and the two new cases in `tests/integration/test_workspace_activation.py` (a rebound port changes what every mint site produces), plus the existing `test_migrated_key_auth.py` and `test_verify_api_key_shape.py` for legacy keys. Dashboard: `CopyField.test.tsx`, `KeysPage.test.tsx`, `SetupGuide.test.tsx`, and the parity e2e spec.

## PR Type
- [x] New Feature
- [ ] Bug Fix
- [x] Refactor
- [x] Documentation
- [ ] Infrastructure / CI

## Relevant issues
Part of #1374 (the otari side of mozilla-ai/otari-ai#1665). Builds on #1326.

## Checklist
- [x] Documentation was updated where necessary (`docs/hybrid-mode-protocol.md`, `ARCHITECTURE.md`).
- [x] If the API contract changed, I regenerated the OpenAPI spec (`uv run python scripts/generate_openapi.py`). (No route or schema changed; the 421 is a new status on existing routes and carries the standard `detail` body.)
- [x] If this changes a rule in `ARCHITECTURE.md` or `scripts/check_architecture.py`, the description names the rule and says why. (No rule changed; the port table and diagram gain an entry.)

## AI Usage
- [ ] No AI was used.

## Summary

- Added `ApiKeyFormatPort` for API key minting, fingerprints, and routing.
- Preserved local `tk-` keys and legacy key authentication.
- Added hosted-key routing with `421 Misdirected Request` and checksum validation with `401`.
- Updated all key issuance paths to use the port.
- Updated hybrid-mode forwarding and dashboard masking to use stored key metadata.
- Updated documentation and expanded test coverage.

## Benefits

Hosted deployments can use region-aware key formats without changing core authentication. Legacy keys continue to authenticate through hash lookup. Dashboard reveals no longer depend on plaintext key slicing.


docs/hybrid-mode-protocol.md
```diff
@@ -292,7 +294,7 @@ Nothing platform-side stores or returns a revision.
 
 | Status | Behavior |
 |---|---|
-| `400`, `401`, `402`, `403`, `404`, `429` | Status code is forwarded to the client; `429`'s `Retry-After` header is preserved. The `detail` is the platform's JSON `detail` string when present, otherwise the fallback `"MCP server resolution failed"`. |
+| `400`, `401`, `402`, `403`, `404`, `421`, `429` | Status code is forwarded to the client; `429`'s `Retry-After` header is preserved. The `detail` is the platform's JSON `detail` string when present, otherwise the fallback `"MCP server resolution failed"`. |
```
peteski22, line 297: `/api/v1/mcp` does not forward this 421. `_McpRoute` sends every `HTTPException` through `_classify` (`src/gateway/api/routes/mcp.py:224`), which has no 421 case, so the caller gets a 502 `resolution_failed` and no host. That happens for a 421 from the platform's MCP resolve in hybrid mode, and in standalone mode for a `Misdirected` key from `verify_api_key_or_master_key`.

Fix: add a 421 case to `_classify` with its own code, with a test for each mode. `_classify` drops the platform's detail on purpose, so either give the MCP error body a field for the host, or say here that MCP returns the 421 without it. The sentence on line 220 ("a `421` from either is forwarded the same way") needs the same correction.

src/gateway/ports/api_key_format_port.py
```diff
@@ -0,0 +1,89 @@
...
+    """
+
+    host: str
+
+
+@dataclass(frozen=True)
+class Malformed:
+    """The key claims this adapter's format and fails it.
+
+    A bad checksum, an unknown region, or the wrong kind for the header it came
```
peteski22: `route(presented)` gets no header, so an adapter cannot tell "the wrong kind for the header it came in". A hosted adapter has to guess about `otr_gw_…`:
- `Malformed`: then a gateway token can never be routed through this port.
- `Local`: then a gateway token pasted as `Otari-Key` skips the fail-fast 401.

Fix: keep this port about API keys only. Say in the `route` docstring that it only sees keys presented as API keys, and that a key in the product format but of another kind is `Malformed`. Gateway tokens are then parsed by whatever checks them (the overlay), not through this port. Reword "the header it came in" here to match.

src/gateway/api/routes/_platform.py
```diff
@@ -584,7 +584,9 @@ async def _post_resolve(
                 detail="Authorization service unavailable",
             ) from None
 
-    if response.status_code in {400, 401, 402, 403, 404, 429}:
+    # 421 is the platform saying the user token belongs to another region; its
+    # detail names the host, and the caller needs both to go there (otari-ai#1665).
+    if response.status_code in {400, 401, 402, 403, 404, 421, 429}:
```
peteski22, line 589: `_post_resolve` sends both `X-Gateway-Token` and `X-User-Token`, and otari-ai#1665 puts a region tag on gateway tokens too. If the platform answers 421 for a gateway token from the wrong region, this line forwards it to every end user as "use another host". That is an operator config mistake the caller cannot fix. Before this change it was a 502 "Authorization service unavailable".

Fix: state in `docs/hybrid-mode-protocol.md` that a 421 from resolve only ever refers to `X-User-Token`, and that a gateway token from the wrong region gets the same answer as any other bad gateway token. Then this forward stays correct.

peteski22 (approved):
Please can you fix the few comments first before merging?

Drafted with help from Claude Code.

web/src/design-system/actions/CopyField.tsx
```diff
@@ -66,19 +66,22 @@ export function CopyableValue({
...
+ * gives its prefix. The bullet run is fixed for `CONCEALED_SECRET`'s reason, so
+ * the length of the key stays off the screen. A response with no fingerprint
+ * falls back to the plain stand-in.
  */
-export function concealedFingerprint(value: string): string {
-  return value.length >= 16
-    ? `${value.slice(0, 8)}••••••••${value.slice(-4)}`
-    : CONCEALED_SECRET
+export function concealedFingerprint(
+  keyPrefix: string | null | undefined,
```
khaledosman: `null` from the wire spreads into the design system, the last layer it belongs in. The TypeScript rung: "`undefined`, not `null`, for absent values in your own types and props. The API layer may hand back `null` … convert at the boundary rather than letting `null` spread through the component tree."

Type both params `string | undefined` (and `SetupSheet.tsx:67-68` likewise), converting at the two callers: `KeysPage.tsx:146` → `result.key_prefix ?? undefined`, `SetupGuide.tsx:206-207` → `issued?.key_prefix ?? undefined`.

ARCHITECTURE.md
```diff
@@ -119,9 +121,10 @@ flowchart LR
...
...t looks it up; the core adapter mints the open-source `tk-` shape and checks every key locally, and a hosted overlay binds a region-tagged, checksummed format behind it (otari-ai#1665). `TelemetryStoragePort` is resolved by the OTLP receiver, the telemetry read endpoints and the purge paths. `ModelProviderPort` is asked on the standalone dispatch path (`resolve_dispatch_provider` in `api/routes/_pipeline.py`) for a candidate the credential ladder could not serve, after an organization's key, a stored instance and `config.yml` have all missed; a routed multi-candidate chain does not reach it yet, because those candidates are credentialed by the routing compiler, which is synchronous and does no I/O. `IdentityProviderPort` is resolved by the OAuth sign-in route, whose core adapter answers with the base build's roster policy (sign in as an account an operator already added; never provision one). The other three are still the seam's mechanism rather than its surface, so the choices they describe (when to meter, what a deployment is entitled to) are made by the mode switch and the hand-wired dependencies described next. The remaining ports in the table above arrive as they gain callers.
```
khaledosman: The caller count is wrong in the sentence this PR rewrites. `GrowthSignalPort` has two core callers that invoke it: `api/routes/auth_signup.py:138` (`await growth.record_signup(`) and `api/routes/organization_keys.py:286` (`await growth.record_activation(`).

So "Four have core callers" should be five, and "The other three" should be the other two: `BillingPort` and `EntitlementPort`, the only ports with no core caller (`require_capability` in `deps.py:751` serves contributed routers). The parenthetical that follows, "when to meter, what a deployment is entitled to", already names exactly those two.

web/src/features/onboarding/SetupGuide.test.tsx
```diff
@@ -37,6 +37,8 @@ interface ApiOptions {
   activation?: WorkspaceActivation
   models?: string[]
   apiKey?: string
+  /** Whether the mint reports the fingerprint it stored; a row without one has nothing to show concealed. */
+  withFingerprint?: boolean
```
khaledosman: `hasFingerprint`. The naming rung covers "every boolean a reader meets, including props".

khaledosman (approved):
Reviewed the increment over #1326 (`e6312007`). The port holds up: no `generate_api_key`/`key_prefix(` caller survives, all three verify entry points funnel through `_verify_and_update_api_key`, the master-key check short-circuits before any DB read so the "no round trip" claim is real, and legacy-by-hash is preserved and covered.

Three non-blocking findings inline.

Two things I checked and am not flagging: the 421 is absent from `docs/public/openapi.json`, but the spec models auth statuses almost nowhere (two `"401"` occurrences repo-wide), so declaring it here would be the inconsistency; and `_api_key_format` resolves the port on the `get_db` session while `ApiKeyFormatPortDep` uses `get_db_if_needed`, which `deps.py:664-670` warns about, but it is harmless given the port's stated contract that every adapter's methods are pure with no I/O.

docs/hybrid-mode-protocol.md
tbille: Fixed in 34404c76. `_classify` maps 421 to a new `misdirected_request` code, keeping the status and dropping the host with the rest of the detail, as R-ERR-1 wants. Tests for both modes: the platform's 421 in hybrid (`test_mcp_execute_endpoint.py`) and a `Misdirected` key in standalone (`test_api_key_format_rebound.py`). The protocol doc and `docs/mcp.md` say MCP returns the 421 without the host, and the shared-ladder sentence now names only the Web Access resolve.

src/gateway/ports/api_key_format_port.py
tbille: Fixed in 34404c76. The port is API keys only: the class and `route` docstrings say it sees a credential offered as an API key and no other header, that gateway tokens are parsed by whatever checks them, and that a gateway token in the product's format is `Malformed`. The "header it came in" wording is gone.

src/gateway/api/routes/_platform.py
tbille: Fixed in 34404c76. The protocol doc states that a 421 from resolve only ever refers to `X-User-Token`, and that a gateway token from the wrong region is the operator's configuration, answered the way any other bad gateway token is, so the forward stays correct.

web/src/features/providers/ProvidersPage.tsx
```diff
@@ -950,7 +950,7 @@ function OnboardingPanel({
           Once a provider exists, the <strong>Overview</strong> page usually
           offers a setup guide that hands you an API key and the call to make.
           Either way, point your app at <code>/v1</code> on this gateway with an
-          API key (the one printed in the server logs starts <code>gw-…</code>).
+          API key (the one printed in the server logs starts <code>tk-…</code>).
```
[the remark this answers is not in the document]
tbille: Fixed in 34404c76: the copy now says "the one printed in the server logs on first start will do" and names no format.

web/src/design-system/actions/CopyField.tsx
tbille: Fixed in 34404c76. `concealedFingerprint` and the `SetupSheet` props take `string | undefined`; the two callers convert with `?? undefined` at the boundary.

ARCHITECTURE.md
tbille: Fixed in 34404c76: five with core callers, and the other two named as `BillingPort` and `EntitlementPort`.

web/src/features/onboarding/SetupGuide.test.tsx
tbille: Renamed to `hasFingerprint` in 34404c76.

peteski22 (approved):

PR #1387, opened 2026-09-18
feat(policy-checks): add a judge gate that evaluates via a local model

agpituk:
## Description

Adds `judge`, a fourth Agent Gates gate type whose verdict comes from a real model call instead of a mechanical pattern match: `rubric` is free text describing what to check (e.g. "does this diff add a narrative comment"), and `otari hook` shells out to `claude -p` locally, against the caller's own Claude Code subscription, to get a pass/fail. Always advisory: a model's verdict is not reproducible the way a glob match is, so it can warn but never block.

This shipped with unit test coverage, but landing it and actually running it live against a real Claude Code session surfaced problems no amount of mocked-subprocess testing would have caught, and a follow-up code review found several more in the same area. All of them are fixed here, each reproduced against the real failure first and each covered by a regression test that was itself verified to fail without the fix.

**Found live:**
- A real session's transcript tokenizes far worse than the "~4 chars/token" prose estimate the old char caps assumed (measured ~1.7 chars/token on real content), so an ordinary long session hit "Prompt is too long" reliably. The judge prompt now carries only the assistant's own text replies from the transcript (never a tool call's own payload, which dominated the bytes and carried no "why" signal anyway), with caps recalibrated against the real measurement.
- Worse: the judge's own `claude -p` call ran with `cwd` inside the repo it was judging, and since Claude Code resolves hooks by directory, its `Stop` hook was this same `otari hook` command — an unguarded run recursed into itself, producing over a dozen real billed model calls in a few minutes before the process tree was killed by hand. It now runs from a dedicated `~/.otari/judge-workdir/`, and with `--tools ""` / `--strict-mcp-config`, since the prompt embeds attacker-influenceable diff/transcript text and never needs a tool or an MCP server to do its one job.
- A slow judge gate (or several) could eat Claude Code's own ~600s hook timeout by itself, past which Claude Code kills `otari hook` and discards its output entirely — meaning *every* gate in the policy, mechanical and required ones included, goes unevaluated for that `Stop` event. `otari hook` now shares one elapsed-time budget across every judge gate and retry in a run, leaving margin for the request that still has to reach the server afterward.
- A judge gate scoped by `when_changed` showed a spurious advisory on *every* matching `PreToolUse` edit, since that event never collects a verdict and the server read the absence as "unknown" the same as a genuinely missing one. It now resolves `not_applicable` on `PreToolUse` specifically, distinct from a real gap.

**Found in review, each reproduced before being fixed:**
- A judge subprocess launch failure (a workdir that fails to create, `claude` disappearing between resolving it and calling it) or an embedded NUL byte in the diff/transcript crashed `otari hook` outright, before it ever reached the server — taking an unrelated required gate down with it. The prompt now goes over stdin (which sidesteps the NUL byte case rather than merely catching it), and the remaining failure modes are caught and reported as a judge error.
- The policy parser accepted a gate id of any length, but the server's own verdict schema caps it at 200 characters; a longer one parsed fine and then 422'd the whole request once a verdict tried to name it. Both limits now share one constant, enforced at parse time.
- A `git diff` that failed to collect fell back to an empty string indistinguishable from a real empty diff, so the judge call still ran and asked the model to judge a change it could not see — which a real call confirmed can still come back "pass". Collection failure now skips the model call and reports an error directly.
- The CLI's advisory/block message only ever printed a gate's static configured text, never the judge verdict's own model reasoning (the same field, it turns out, three other gate types use for their own specifics) — a verdict naming a specific line of code was invisible; only the generic policy sentence ever reached the terminal.

Also restores the seven pre-existing gates in `.otari-gates.yml` (an earlier commit had replaced them with just the new one instead of adding to it), so this repo's own dogfooded policy is complete again.

## How to test it locally

```
make lint
make typecheck
make openapi-check
make postman-check
make test          # tests/unit + tests/integration
pnpm --dir web run lint
pnpm --dir web run typecheck
pnpm --dir web test
```

To see the gate itself work without spending a real model call: `uv run otari hook setup --api-key <key>` against a running `uv run otari serve`, then edit a file under `src/gateway/**` with an obviously narrative comment (e.g. `# Added this to do X`) and let a `Stop` event fire; `~/.otari/judge-calls.log` gets an `invoking` line either way, and `--judge-dry-run` (already the flag `hook setup` does not add, so the default install is real) reports the estimated prompt size without spending anything. A real run costs one small Haiku call.

Every fix in this PR was verified the same way: reproduce the bug against real `git`/`subprocess` behavior (a Latin-1 tracked file, a 201-character gate id, a real recursive process tree, a real "prompt is too long" rejection) or a crafted mock matching the real server's response shape, write the regression test, then temporarily revert the fix and confirm the new test fails before restoring it. That discipline is why the diff is larger than the bug list alone would suggest — most of it is tests plus the doc updates in `docs/agent-gates.md` that keep the mechanism explanations honest about what was actually verified against a real call versus assumed.

## PR Type

- [x] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

None filed; the bugs fixed here were found in this PR's own live testing and review, not from a tracked issue.

## Checklist

The API contract changed twice: `judge_results` gained a `command_scope`-style sibling field from a parallel PR already on `main` (unrelated to this one, resolved by the rebase) and, in this PR, moved from a plain list defaulting to `[]` to an `Optional` list defaulting to `None`, so `otari hook` can tell the server "this event never runs judge gates" (`PreToolUse`) apart from "it does, and found none" (`Stop`). `docs/public/openapi.json`, the Postman collection, and the dashboard's `web/src/client/schema.ts` are all regenerated and committed. No `ARCHITECTURE.md` rule changed.

## AI Usage

- [ ] No AI was used.

Every line of code and every test in this PR was written by Claude Code. The process was not autonomous, though: the live-testing discoveries came from directed end-to-end testing against a real Claude Code session (registering the hook for real, watching it recurse, watching it hit a real "prompt is too long"), and the four review-round fixes came from the user's own independent code reviews, run outside this session and fed back in as specific findings with file/line references. Each finding was verified against the real failure before being fixed (a real Latin-1 tracked file, a real 201-character gate id, a real recursive process tree measured via `ps`, a real "prompt is too long" rejection captured with `--output-format json`), not accepted on the reviewer's word alone, and each fix's regression test was checked to actually fail when the fix was temporarily reverted.

**NOTE:**

## Summary

- Added advisory `judge` gates that evaluate free-text rubrics with a local Claude model.
- Added path scoping, model selection, dry-run mode, audit logging, retries, prompt limits, and shared hook time budgets.
- Improved hook error handling, validation, diff collection, recursive-hook prevention, and CLI reasoning output.
- Restored existing gates and added the `no-narrative-comments` gate.
- Updated API schemas, documentation, dashboard schemas, and Postman support.
- Added unit, integration, and CLI test coverage for judge gates and failure cases.

## Technical notes

Judge verdicts are submitted through `judge_results`. Judge gates remain advisory and cannot block requests. The implementation validates gate IDs and rubrics, bounds request fields, and supports `pass`, `fail`, and `error` outcomes. The hook invokes `claude -p --model <model>` in an isolated working directory.


dpoulopoulos:
Currently, we can only use Claude with this, since the CLI explicitly calls `claude --model <name> -p`. You can change the model, but not the provider, at least not easily. You could repoint Claude Code itself using environment variables, but it's cumbersome.

This seems like an ideal use case for open-weights models and small local models though, and it could be an easy win. Instead of hard-coding the command while keeping only the model name flexible, we should make both the command and the model name dynamic. That way, I could use LM Studio with Qwen locally if I want. We could still have something like Claude being the default, but we can provide the option to change this. Everything is happening locally anyway, why not have a local model act as the judge too?

Happy for this to be a follow-up issue, but if I understand the architecture here, it seems easy to implement. WDYT?

agpituk:
> Currently, we can only use Claude with this, since the CLI explicitly calls `claude --model <name> -p`. You can change the model, but not the provider, at least not easily. You could repoint Claude Code itself using environment variables, but it's cumbersome.
> 
> This seems like an ideal use case for open-weights models and small local models though, and it could be an easy win. Instead of hard-coding the command while keeping only the model name flexible, we should make both the command and the model name dynamic. That way, I could use LM Studio with Qwen locally if I want. We could still have something like Claude being the default, but we can provide the option to change this. Everything is happening locally anyway, why not have a local model act as the judge too?
> 
> Happy for this to be a follow-up issue, but if I understand the architecture here, it seems easy to implement. WDYT?

100% I'd want this to happen, I'd do it in a follow up if possible but yeah, I'd go this way :)

dpoulopoulos (approved):

dpoulopoulos (approved):

Issue #1331, opened 2026-09-18
add_member reports a duplicate membership for any integrity violation, including a workspace deleted mid-join


peteski22:
## What happens

`WorkspaceService.add_member` wraps its membership insert in one `except IntegrityError`, and translates every one of them into `WorkspaceMemberAlreadyExistsError`, whose message is:

> `User {user_id} is already a member of this workspace`

That is true for the violation it was written for, the unique index on the membership pair. It is false for any other integrity violation the same insert can raise.

The one that matters: a member joining a workspace that is being deleted concurrently. The insert fails on the foreign key, not the unique index, and the caller is told the user is already a member of a workspace that no longer exists. Right refusal, wrong reason, and a 409 where the honest answer is closer to a 404.

## Why now

Found while fixing #1327. That fix takes the workspace lock in `delete_workspace`, which makes the interleaving deterministic: the losing `add_member` now reliably fails on the foreign key instead of sometimes winning and leaving an orphaned ceiling behind.

So #1327 turns a data bug into a message bug, which is a clear improvement. It does not create this one. The same wrong message was reachable before, just less predictably.

## Why it was left alone

The message is wire-visible, so changing it is a decision rather than a fix, and #1327 was scoped to the lock and its test. Two things need deciding:

- Does the caller get a different **message** at the same 409, or a different **status** as well? A workspace that no longer exists is arguably a 404, and that is a bigger change than rewording.
- Does anything depend on the current text? The integration suites assert on error details in places, and the dashboard may surface this string.

## The codebase already has the technique

`user_service._is_email_conflict` discriminates an `IntegrityError` by the constraint the engine names rather than by "an `IntegrityError` happened", precisely so a different violation keeps its own error. Its docstring covers the awkward part: PostgreSQL names the index in the message, SQLite names the column, and SQLAlchemy's asyncpg wrapper does not expose `constraint_name`, so matching is partly structural and partly textual.

`add_member` wants the same shape: match the membership unique index, keep `WorkspaceMemberAlreadyExistsError` for that, and let anything else raise something that says what actually went wrong.

## Acceptance

- [ ] A member added to a workspace being deleted concurrently gets an error that describes that, not a duplicate-membership error.
- [ ] A genuine duplicate membership still gets `WorkspaceMemberAlreadyExistsError` with its current text and status.
- [ ] The discrimination is by constraint, following `_is_email_conflict`, and works on both engines the test suite runs against.
[...]

Issue #1202, opened 2026-09-15
First domain: budgets


peteski22:
### Proposal

The six steps from #1201, for the budgets domain:

1. Schemas out of `budgets.py`, `scoped_budgets.py`, `organization_budgets.py` and `workspace_member_budget_policies.py` into `schemas/budgets.py`.
2. Queries out of those routes and the seven budget services into `repositories/budgets/`, over `BaseRepository`.
3. `services/budgets/`: one `BudgetService` with a small public API, built with its own repositories and the Unit of Work (#1262, #1276). `budget_service.py`, `scoped_budget_service.py`, `budget_periods.py`, `budget_reservation_ledger.py`, `budget_retiming.py`, `tenancy/organization_budget_service.py` and `tenancy/workspace_budget_default_service.py` become its methods or its private modules. So does the budget alert trigger, if #1173 PR 2 lands first; it keeps calling the alerts service to deliver.
4. Commits into Unit of Work blocks.
5. Errors into `exceptions/budget_exceptions.py`, each class keeping its status.
6. The budget names removed from the baselines (#1203).

Code moves and behavior changes in separate commits.

The request path is out of scope. `reserve_budget`, `reconcile_reservation`, `refund_reservation` and `increase_reservation` keep their signatures, stay importable from `gateway.services.budgets` for `_pipeline.py`, `_passthrough.py`, `batches.py` and `search.py`, and stay on the baselines. #1277 replaces them with Try-Confirm/Cancel.

### Acceptance

- [ ] No budget route or service imports `sqlalchemy`, except the module that holds the four request-path functions.
- [ ] Every other budget query is in `repositories/budgets/`.
- [ ] Code outside the domain imports budgets only from `gateway.services.budgets`.
- [ ] The budget integration tests pass unchanged.
- [ ] The domain's counts are recorded before and after. Today: 7 service modules, 4,018 lines, 42 public functions and methods.

### Depends on

#1201, #1203 and #1276. #1191 and #1193 make it easier; not required.

Part of #1171.

peteski22:
Planned on 2026-09-17, against `main` at `d23ab736`, with the decisions in the core cleanup spec.

### Eight PRs, each landing on its own and changing nothing on the wire

1. Organizations announces membership changes through a listener, so it stops importing budgets.
2. The budget errors move into `exceptions/budget_exceptions.py`.
3. The budget request and response models move into `schemas/budgets.py`.
[...]

PR #1395, opened 2026-09-18
fix(tenancy): stop a workspace delete leaving a joining member's ceiling behind

### LINKED ISSUE #1327: Deleting a workspace can orphan a member's spending ceiling when a member is added concurrently
## What happens

Deleting a workspace sweeps the spending ceilings keyed on it and on its memberships, because `scoped_budgets.scope_id` is not a foreign key and nothing cascades. It reads the membership ids first, then deletes.

That read is not serialized against a member being added at the same moment. `WorkspaceService.add_member` locks the workspace row before it creates the membership and materializes that member's ceiling. `WorkspaceService.delete_workspace` locks only the **organization**, which is a different row, taken for a different reason (the "last workspace cannot go" count).

So two transactions can interleave like this:

1. **A** (`delete_workspace`) locks the organization, counts the remaining workspaces, and reads the membership ids for the doomed workspace.
2. **B** (`add_member`) locks the workspace, inserts a membership, materializes that member's `scoped_budgets` row, commits.
3. **A** deletes the workspace. The database cascade removes B's membership row. A's sweep never saw it, so B's ceiling is left behind.

## Why it matters

An orphaned ceiling is not inert. It holds a RESTRICT reference to its budget, so that budget can never be deleted again, and there is no page in the dashboard that lists ceilings for someone to go and find it on. It also never binds to anything, because the membership it names is gone. This is the exact state `routes/scoped_budgets._require_scope_exists` refuses to create on the way in.

## Where it is

On `main` at `832d3978`:

- `src/gateway/services/tenancy/workspace_service.py`, `delete_workspace`: the only lock taken is `await self.organizations.organizations.lock(organization.id)`.
- `src/gateway/services/tenancy/workspace_service.py`, `_delete_scoped_budgets_for`: reads `member_ids` with no workspace lock, then issues the delete.
- `src/gateway/services/tenancy/workspace_service.py`, `add_member`: does take `await self.workspaces.lock(workspace.id)`, and says why in a comment.

The asymmetry between those two paths is the bug.

## Why CI has never caught it

There is no test for this interleaving. `tests/integration/test_tenancy_races.py` has 14 race tests, including `test_concurrent_deletes_cannot_remove_the_last_workspace`, which covers delete against delete. None covers delete against add. A green pipeline means the tests pass, not that the code is free of races, and no test exercises this one.

## Suggested fix

Take the same lock `add_member` takes, immediately before reading the membership ids, so the snapshot, the sweep and the workspace delete all sit behind it in one transaction:

```python
await self.workspaces.lock(workspace_id)
member_ids = await self.members.ids_for_workspace(workspace_id)
```

A concurrent `add_member` then either lands before the snapshot and is swept, or blocks until the workspace is gone and fails, which is the correct outcome either way.

Worth pairing with a test in `tests/integration/test_tenancy_races.py` following the shape of the existing ones, since a lock with no test around it is easy to drop later.

## Notes

- Found by CodeRabbit reviewing #1325, which moves this code but does not change its locking. The race predates that PR: the same read-then-delete with no workspace lock is on `main` today.
- #1325 is a deliberate no-behavior-change refactor, so adding a lock there was out of scope.
- #1325 has merged, so the sweep now lives in `WorkspaceBudgetDefaultService.workspace_deleted` and the read is `self.members.ids_for_workspace(workspace_id)` in `delete_workspace`. The lock belongs immediately before that read.
- The budgets domain work (#1202) moves this code again in its PR 6, which rewrites the listener over repositories. The lock is orthogonal to that move and belongs on whichever shape is current, but fixing it before the move is cheaper than after.

peteski22:
## Description

Deleting a workspace could leave behind a spending ceiling belonging to a member who joined at the same moment. This closes that gap.

When a workspace is deleted, the ceilings keyed on it and on its memberships have to be swept by hand, because `scoped_budgets.scope_id` is not a foreign key and nothing cascades. The delete read the list of memberships first and swept second, and nothing stopped a member being added in between:

1. The delete reads the workspace's membership ids.
2. A member is added, gets a ceiling, and commits.
3. The delete removes the workspace. The cascade removes the new membership, but the sweep never saw it, so its ceiling survives.

An orphaned ceiling is not harmless. It holds a RESTRICT reference to its budget, so that budget can never be deleted again, and no page lists ceilings for anyone to find it on.

The cause was an asymmetry: `add_member` locks the workspace row before it writes, and `delete_workspace` locked only the organization, for an unrelated reason (the "last workspace cannot go" count). The fix takes the same workspace lock in `delete_workspace`, immediately before the membership snapshot. A concurrent join now either lands before the snapshot and is swept, or waits until the workspace is gone and is refused.

`WorkspaceRepository.lock`'s docstring said two races share that lock. It is three now, so it is updated.

**One known consequence, filed separately as #1331.** The member who loses that race is refused, which is correct, but with the wrong words. `add_member` reports every integrity violation as "User X is already a member of this workspace", and the losing insert fails on the foreign key, not the unique index. This PR does not create that message; it was reachable before, just less predictably. What this PR does is turn a data bug into a wording bug. Fixing the wording changes what callers see, so it needs its own decision and is not slipped in here.

## How to test it locally

```
uv run pytest tests/integration/test_tenancy_races.py -v
```

The new test, `test_a_join_during_a_workspace_delete_leaves_no_orphaned_ceiling`, drives a delete and a join concurrently on separate sessions and asserts no ceiling outlives its membership. It is a real pin rather than a lucky pass: without the lock it failed 5 runs out of 5, and with it it passes every time (8 of 8 when written, 5 of 5 again after rebasing onto current `main`).

It is committed on its own before the fix, so checking out the first commit shows the failure.

That file already held 14 race tests, including one for two deletes racing each other. None covered a delete racing a join, which is why this was never caught.

`make lint`, `make typecheck` and the tenancy integration suites pass.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes #1327. Follow-up in #1331.

## Checklist

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

Claude, via Claude Code.

The race was first reported by CodeRabbit on #1325. It was confirmed against `main` before anything was written, which is how it turned out to predate that PR.

## Summary

- Workspace deletion now locks the workspace before reading memberships.
- This prevents orphaned spending ceilings when a member joins during deletion.
- Added an integration race test for concurrent deletion and member addition.
- Updated `WorkspaceRepository.lock` documentation to describe this race.


src/gateway/repositories/tenancy/workspace_repository.py
```diff
@@ -47,6 +47,9 @@ async def lock(self, workspace_id: uuid.UUID) -> None:
           excepted, since a just-created workspace has no default that could
           race it) takes this same lock before its own read, so the two either
           serialize or, for ``create_workspace``, never overlap.
+        - A workspace's deletion reads its memberships before announcing them,
+          and a membership created after that read rides the delete cascade
+          while the ceiling keyed on it survives.
```
khaledosman, line 52: The bullet is right for the membership paths, but it's a fourth producer of the same orphan that takes no lock: `OrganizationBudgetService.create_ceiling` (`organization_budget_service.py:788`) and `POST /v1/scoped-budgets` (`api/routes/scoped_budgets.py:226`) both check the scope exists and then insert, and both accept `workspace` and `workspace_member` scopes. A delete committing between that check and the insert leaves exactly the ceiling this PR describes: no page lists it and its RESTRICT reference pins the budget forever.

So the join side is closed and the direct side isn't. Either take `WorkspaceRepository.lock(scope_id)` in `create_ceiling` for the two workspace-family scopes (order stays organization then workspace, so no inversion with `delete_workspace`), or file it next to #1331 and say so here, since the docstring now reads as if the lock covers the delete race in general.

tests/integration/test_tenancy_races.py
```diff
@@ -76,6 +83,11 @@
 
 _RACERS = 4
 
+# How long the delete waits for the join before giving up on it. Generous,
+# because the join blocks on the workspace lock once the delete holds it, and
+# that wait is paid in full on every passing run.
+_JOIN_WINDOW = 1.0
```
khaledosman: With the fix in place the join blocks on the workspace lock until the delete commits, so `asyncio.wait` can never be satisfied and every green run pays the full 1.0s. The window only has to cover the join's commit on the *broken* path, which is fast, so 0.25 would do the same job at a quarter of the cost.

tests/integration/test_tenancy_races.py
```diff
@@ -415,6 +427,90 @@ async def run_one(workspace_id: uuid.UUID) -> object:
...
+    service = WorkspaceService(async_db, membership_listener=WorkspaceBudgetDefaultService(async_db))
+    target = await service.create_workspace(user=owner, workspace_create=WorkspaceCreate(name="Target"))
+    await service.create_workspace(user=owner, workspace_create=WorkspaceCreate(name="Survivor"))
+    joiner = await create_member(async_db, organization, role="member", full_name="Joiner")
+    await WorkspaceBudgetDefaultService(async_db).create_default(
+        user=owner,
+        workspace_id=target.id,
+        request=WorkspaceMemberBudgetPolicyCreate(budget_id=await create_budget(async_db, max_budget=10.0)),
+    )
+    await async_db.commit()
```
khaledosman, line 476: Add a positive control here: assert one `workspace_member` ceiling exists before the delete. The final assertion is a negative over a set that is empty in the passing case, so if the default ever stopped materializing (a renamed scope string, a changed `create_default` signature) the test would go green having pinned nothing.

khaledosman (approved):
The lock closes the join-side race and the test is a real pin. One gap worth a follow-up: the same orphan is still reachable through the ceiling API, which takes no lock. Two test nits.

src/gateway/repositories/tenancy/workspace_repository.py
peteski22: Verified. Both paths check the scope, then `add`, then `commit`, and neither file takes a lock anywhere, so a delete committing between the check and the insert leaves the same orphan.

Filed as #1404 rather than fixed here, taking your second option. Both paths are about to move under #1202 (`create_ceiling` onto the budget service and its repository, the deployment route after it), and the deployment route reaches the database directly today with no `WorkspaceRepository` to hand, so fixing it now means fixing it twice. Each path also wants its own failing race test, which is more than this PR should carry. #1404 has the lock-ordering note you gave: these paths would take only the workspace lock, so no inversion with `delete_workspace`.

The docstring point stands, fixed in 325d078b: it now says this lock serializes a deletion against membership creation only, and that a ceiling created directly on a workspace or membership scope takes no lock.

tests/integration/test_tenancy_races.py
peteski22: Done in 8ffa6cc8, with one addition. Shrinking the window on its own makes CodeRabbit's comment below worse: the window also had to cover the join opening a session, loading the actor and running its pre-checks, so at 0.25 a slow runner could let the window lapse before the join ever reached the lock, and the test would pass on broken code.

So the window now opens only once the join is at the lock. The join's `WorkspaceRepository.lock` is wrapped to set an event first, and the delete waits for that event (bounded at 5s, never paid on a healthy run) before starting the 0.25s window. What is left inside the window is the insert and commit on the unlocked path, which is the fast part you described.

Measured: 2.8s per run, down from 3.5s. With the lock removed from `delete_workspace` it still fails 5 runs out of 5, on one orphaned ceiling.

tests/integration/test_tenancy_races.py
peteski22: Added in 2d33dc6e: exactly one `workspace_member` ceiling must exist before the delete.

I checked it can fail. With `create_default` taken out of the setup the test goes red at the control. Without the control, that same run passed having pinned nothing, which is the case you described.

tests/integration/test_tenancy_races.py
```diff
@@ -415,6 +427,90 @@ async def run_one(workspace_id: uuid.UUID) -> object:
...
+    async def member_joined(self, member: WorkspaceMember) -> None:
+        await self._inner.member_joined(member)
+
+    async def member_removed(self, member: WorkspaceMember) -> None:
+        await self._inner.member_removed(member)
+
+    async def workspace_deleted(self, workspace_id: uuid.UUID, member_ids: Sequence[uuid.UUID]) -> None:
+        await self._inner.workspace_deleted(workspace_id, member_ids)
+        self._swept.set()
+        await asyncio.wait({self._joining}, timeout=_JOIN_WINDOW)
```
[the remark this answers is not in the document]
peteski22: Done in 8ffa6cc8, with a checkpoint as suggested: the join's `WorkspaceRepository.lock` is wrapped to set an event before it takes the real lock, and the delete waits on that event with `asyncio.wait_for`, so a join that never gets there fails the test rather than passing quietly.

One correction to the analysis. With the fix in place the outcome is the same whether the join is at the lock or still in its pre-checks, because it cannot insert either way, so that path was never at risk. The exposure was on the unfixed path: a join slow enough to outlast the window would let the test pass on broken code. The checkpoint closes that, and it is also what makes the shorter window requested above safe.

Verified both ways: 5 of 5 pass with the lock, 5 of 5 fail without it.

Issue #1335, opened 2026-09-18
Three pages format money, tokens and dates with their own helpers, and the copies disagree on locale and precision


khaledosman:
Three of the four largest pages format money, tokens, dates and latency with helpers of their own rather than through `shared/helpers/format.ts`. The copies are not equivalent: they disagree on locale, on precision, and on what an absent or malformed value renders as.

## The rule already exists

`.github/skills/frontend-standards/formatting-and-i18n.md`:

> Every user-visible number, cost, and timestamp goes through `web/src/shared/helpers/format.ts`. That module is the only place the dashboard decides how a value reads, which is what keeps a cost formatted the same way in a table cell, a stat tile, and a share card.

and, on locale:

> Formatters are pinned to `en-US`, deliberately. The dashboard's copy is US English (see AGENTS.md), and a mixed page (US wording, locale-formatted numbers) is worse than a consistent one. It is also an operator tool where a number's shape is often copied into a ticket or a support thread, and a value that changes shape by machine makes those reports harder to compare. If localization ever arrives, it arrives here, in one module.

`web/design/DESIGN.md` puts it more bluntly:

> **Never hand-roll a formatter.** `toLocaleString()` at a call site is how two pages come to print the same number differently. Every value in a cell or a `KpiCell` goes through `@/shared/helpers/format`.

`features/usage/ShareCard.tsx` is the one place that already carries a header explaining why it formats on its own, and it is not in scope here.

## The copies

| File | Line | Local helper | Call sites |
| --- | --- | --- | --- |
| `features/activity/ActivityPage.tsx` | 72 | `usd`, locale `undefined`, 4 fraction digits | 3 |
| | 78 | `formatUSD` | 8 |
| | 82 | `formatTokens` | 11 |
| | 90 | `usdPrecise`, locale `undefined` | 1 (via `formatUnitRate:96`) |
| | 114 | `formatLatency` | 4 |
| | 120 | `absolute` | 2 |
| `features/budgets/BudgetsPage.tsx` | 70 | `usd`, locale `undefined`, 2 fraction digits | via `formatUSD` |
| | 76 | `formatUSD` | 5 |
| | 101 | `absolute` | 3 |
| `features/organization/OrganizationMembersPage.tsx` | 104 | `usd`, locale `undefined`, 2 fraction digits | 3 |
| `features/usage/usageTotals.ts` | 53 | `formatLatency` (exported) | 2 |

[...]

PR #1399, opened 2026-09-18
refactor(activity): put the activity page's derivations and parts where they can be tested

### LINKED ISSUE #1339: ActivityPage implements what it should compose: 35 helpers and 8 components before the page starts
`features/activity/ActivityPage.tsx` is 2340 lines. The page component starts at line 1147, after 35 helper functions and 8 sub-components. The standards name this file as the example of the thing not to do, so this is the issue for doing something about it.

## The rule already exists

`.github/skills/frontend-standards/component-architecture.md`, "File size is a design signal":

> The largest pages here are past two thousand lines (`features/activity/ActivityPage.tsx`), and their tests are larger still. Nothing about the domains requires that: it is what happens when every dialog, row renderer, and derived table lives in the file that renders the page.
>
> The rule is not a line count, it is this: **a page component composes; it does not also implement.** [...] Splitting an existing giant is welcome as its own change; growing one is not.

`ActivityPage.test.tsx` is 2783 lines, the largest test in the tree, which is the "their tests are larger still" half.

## What comes out, in the order it is worth doing

**Formatters first (#1335).** Lines 72 to 158 are money, token, latency and timestamp helpers that belong in `shared/helpers/format.ts`. Tracked separately because the copies disagree with the shared module rather than merely duplicating it.

**Components, each to its own file in `features/activity/`:**

| Line | Component | Roughly |
| --- | --- | --- |
| 160 | `InFlightWait` | 30 lines |
| 193 | `InFlightControl` | 85 |
| 421 | `StatusMark` | 20 |
| 599 | `TokenBar` | 60 |
| 749 | `RoutingCell` | 30 |
| 802 | `RoutingPlan` | 115 |
| 920 | `DetailField` | 35 |
| 971 | `RequestDetail` | 175 |

**Pure derivations, to a module of their own** (`activityRows.ts` or similar), because these are what a unit test should reach without rendering a page:

`sortedBreakdown:106`, `tokenComposition:494`, `toolUsage:523`, `formatToolUsage:549`, `toolCost:558`, `selectionReasonLabel:663`, `indexGroupOutcomes:697`, `attemptSentence:717`, `attemptOutcome:778`, `planOrder:791`, `pricingSelectorOf:958`, `resolveWindow:369`, `resolveExtentWindow:399`.

**Lookup tables** (`STATUS_LABELS:441`, `SOURCE_LABELS:452`, `STATUS_OPTIONS:299`, `PRICED_OPTIONS:309`, `TOOL_OPTIONS:317`, `TOKEN_SEGMENTS:575`) go with the derivations that read them.

What is left is the page: URL state, the queries, the filter bar, the table, and the composition. That is the shape `component-architecture.md` describes.

## Doing it safely

`ActivityPage.test.tsx` covers the behavior today and should keep passing untouched through the move, which is what makes this a mechanical change rather than a rewrite. New tests go beside each extracted file afterwards; the derivations are the ones that gain the most, since none has a direct case today.

Worth doing in more than one PR. The formatters, then the pure derivations, then the components, is an order where each step leaves the file smaller and nothing half-moved.

## Found in

A standards audit of `web/`. `features/usage/UsagePage.tsx` (1596), `features/keys/KeysPage.tsx` (1579), `features/providers/ProvidersPage.tsx` (1503) and `features/overview/OverviewPage.tsx` (1464) have the same shape at smaller scale, and the same treatment applies; this issue is about the one the docs already name.

khaledosman:
## Description

The Activity screen looks and behaves exactly as it did. What changes is the code behind it.

Everything the page renders used to live in one file: the status mark, the token bar, the routing cell and its plan table, the in-flight popover, the request detail panel, plus every calculation each of those needs. The page component itself did not start until most of the way down. That made the screen hard to change with any confidence, because nothing in it could be exercised on its own: the only way to check that a token split or a fallback sentence was right was to render the whole page and read the table.

Each piece now sits in its own file, and the calculations sit in one module of their own. Nothing about what the screen shows moved with them. The payoff is in the tests: the sixteen calculations behind the log (how a request's tokens split into fresh, cached and output; what a routed request's fallback chain ended up doing; what a time preset resolves to; what a row's tool calls cost) now have direct cases for the first time, and so does each of the eight parts.

Four formatters moved along with the rest so the parts could keep using them. They are deliberately left as they are rather than folded into the shared formatting module, because the two disagree on rounding and on how an absent value is spelled; #1335 tracks reconciling them, and doing it here would have been a silent change to what the screen prints.

One documentation change rides along, because this is the PR adding nine new files under `web/src`: `.github/instructions/frontend-standards.instructions.md`, which is what CodeRabbit loads as review guidance, carried no file-naming rule, so the one the skill has (a component file is named for its component, a hook and a pure module are camelCase, a route file is whatever file-based routing requires) was never enforced on a review. It says so now, with the carve-out the tree already relies on for a module whose exports share no one subject.

## How to test it locally

Nothing to eyeball that a reviewer would not already know: this is a move, and the page's own test suite is the proof. `web/src/features/activity/ActivityPage.test.tsx` (2,511 lines) and `ActivityPage.filters.test.tsx` are **unchanged by this PR**, and both pass, which is what makes the move mechanical rather than a rewrite.

```
pnpm --dir web run lint        # clean
pnpm --dir web run typecheck   # clean
pnpm --dir web test            # 217 files, 6,946 tests, all passing
```

If you want to see the screen: `make dashboard`, then open `/activity`, expand a row, and open the in-flight popover.

The ten new test files add 80 cases. Each one was checked by breaking the behavior it covers in the source and confirming it went red, then restoring: 74 deliberate breaks across the model and the eight components, all caught. No backend file is touched, so the Python checks, the OpenAPI spec and the Postman collection are not implicated, and nothing under `design-system/` or any `.stories.tsx` changed, so the Storybook smoke gate is not either.

## PR Type

- [ ] New Feature
- [ ] Bug Fix
- [x] Refactor
- [x] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Fixes #1339

## Checklist

- [x] If the API contract changed, I regenerated the OpenAPI spec (`uv run python scripts/generate_openapi.py`). (No backend file is touched, so the contract did not change.)
- [x] If this changes a rule in `ARCHITECTURE.md` or `scripts/check_architecture.py`, the description names the rule and says why. (No rule changes; this follows the existing `component-architecture.md` rule.)

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

The fourteen renames were driven through the TypeScript language service's `findRenameLocations` rather than search and replace, so no reference was moved by pattern matching. Every new test was mutation-checked as described above.

**NOTE:**


khaledosman:
@coderabbitai review

PR #1413, opened 2026-09-21
feat(policy-checks): let a repo author its own check_passed verifier

agpituk:
## Description

@daavoo opening this one as a draft to discuss your comment about the primitives — whether this is the right way to do it.

The core idea: `check_passed`, a fourth gate type where the "check" isn't a glob or a phrase Otari matches itself, but a **verifier**: a script the repo itself owns (e.g. `.otari-gates/verifiers/no-conflict-markers.sh`), that `otari hook` just executes and reads the exit code from. `0` = pass, `1` = fail, anything else = error. No allowlist of otari-shipped checks, no otari release needed to add a new one — you write the script, point a gate at it, done. That's the main thing that changed from the earlier prototype: that one hardcoded a fixed set of Python checks into `cli.py`, which didn't scale past the first check anyone wanted.

I added two examples here, both dogfooded on this repo's own `.otari-gates.yml`:
- `no-conflict-markers.sh` — fails on a leftover `<<<<<<</=======/>>>>>>>` marker.
- `no-stranded-docblocks.py` — reimplements the stranded-docblock check AGENTS.md already documents by hand.

Curious to hear your take, especially on:
- Trust model: a verifier is just as trusted as anything else already in the repo (a Makefile target, a pre-commit hook) — no sandboxing, no "must predate this diff" guard. Is that the right line, or too permissive?
- No dependency isolation: a verifier gets a `cwd`, nothing else. A Python one either stays stdlib-only or self-bootstraps via `uv run --script`; there's no venv guarantee.
- Sharing verifiers across repos is explicitly not built here — still an open question for later.

Stacked on #1387 (agent-gates-llm-judge). Once that merges, this needs a rebase onto main dropping whatever commits duplicate it.

## How to test it locally

1. `otari migrate --config <config.yml>` then `otari serve --config <config.yml>` against a throwaway DB.
2. `otari hook setup` in this repo (or POST directly to `/api/v1/hooks/check`, see `docs/agent-gates.md`).
3. Trigger a `Stop` event: e.g. append a `<<<<<<<`/`=======`/`>>>>>>>` block to a tracked file and let `otari hook` run — `no-leftover-conflict-markers` should block with the offending lines as detail. Revert and it passes.
4. Automated coverage: `tests/unit/agent_runtime/test_check_passed.py` (evaluator), `tests/unit/agent_runtime/test_policy.py` (parsing), `tests/integration/test_hooks_route.py` (wire contract), `tests/unit/test_hook_cli.py` (verifier subprocess execution, including real-script exit-code cases, not mocks).

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

None filed — follows up on review discussion on #1387.

## Checklist

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:** Claude Code (Claude Sonnet 5)

Implemented end-to-end (domain types, policy parsing, evaluator, wire contract, CLI subprocess execution, docs, tests, and both dogfooded verifier scripts) from a detailed spec agreed on beforehand. Verified against real execution rather than mocks where load-bearing: real `otari serve` + `otari hook` runs, real BSD (macOS) and GNU (`debian:trixie-slim` container) `grep`/`git`, and the real system Python outside any project venv (which caught a real `TypeError` from a 3.10-only annotation).

**NOTE:**

## Summary

Adds repository-owned `check_passed` policy gates. Repositories can run custom verifiers and use their results in policy decisions.

The change includes:

- Policy, API, evaluator, and CLI support.
- Timeout, output, path, and workload limits.
- Two example verifiers for conflict markers and stranded documentation blocks.
- Stop-event documentation and `.otari-gates.yml` examples.
- Unit and integration test coverage.

## Benefits

Repositories can define custom checks without changes to Otari. Required checks can block operations when they fail or report an error.


dpoulopoulos (approved):

dpoulopoulos:
Thanks for the review. All five points looked at, four changed, one split into "fix the wording here" and "track the real question separately". Everything is folded into the commit it belongs to; no new commits.

**`no-conflict-markers.sh`: diff3 base delimiter.** Added `|||||||` to the pattern, and the header comment now says why it belongs there. Checked against a real repo with a diff3 conflict: the verifier flags a file whose only leftover is the base delimiter, which it used to pass.

**`no-stranded-docblocks.py`: unreadable file.** `read_text` is wrapped now, printing the path and the error and returning 2. Exit 1 stays reserved for a real docblock mismatch, so an unreadable file no longer looks like one.

**Verifier output decoding.** `encoding="utf-8"` with `errors="replace"`, matching what `_hook_collect_diff` already does for the same reason. A verifier that echoes a tracked file's own bytes can emit something that is not valid UTF-8, and the strict decode would raise out of `subprocess` and take every other gate in the policy with it. Covered by a test that prints raw `\xff\xfe`.

**Timeout across the process tree.** Implemented, with one correction to the reasoning. On POSIX, `subprocess.run`'s timeout cleanup calls `wait()` after `kill()`, not `communicate()`, so the direct call did not actually block on a background child holding the pipe (measured: 0.2s, not the child's 30s). What it did do was leave that child running as an orphan with the captured pipes still open, and on Windows, where the cleanup does read those pipes, it is the hang you described. The verifier now runs under `start_new_session=True` and a timeout kills the whole group. The test asserts the background child is gone afterwards, which is the part that actually fails on the old code; an elapsed-time assertion would have passed either way and proved nothing.

**Trust before running a `check_passed` verifier.** You are right about the documentation and I have rewritten it. The old paragraph claimed `otari hook` already ran whatever the policy named, which is not true: `command_match` and `command_if_changed` inspect command text, `changed_path` matches globs, `judge` sends a prompt. `check_passed` is the first gate type that executes something the policy names, and the docs and `CheckPassedGate`'s docstring now say exactly that, and name the case the repo-is-the-boundary argument does not cover: reviewing someone else's branch, where a policy and a verifier arriving together run on the next `Stop` event without the reviewer choosing to run anything.

I have not added a trust step in this PR. Doing it properly means a mechanism (a fingerprint of the policy plus its verifiers per repo, a prompt on change, a user-level allowlist, or some mix), and that is a product decision about where the boundary sits and what first run feels like, not something to settle in the PR introducing the gate type. Tracked in #1442, with the alternatives written up and linked back to this thread.

Verified: `make lint`, `make openapi-check`, `make postman-check`, and `tests/unit/test_hook_cli.py` plus `tests/unit/agent_runtime` all pass. Both verifiers in `.otari-gates/verifiers/` still exit 0 against this repo.

Issue #1416, opened 2026-09-21
Move the reservation status names beside the column they name


peteski22:
## What

Move the reservation status names (`RESERVATION_ACTIVE`, `RESERVATION_SETTLED`, `RESERVATION_RELEASED`, `RESERVATION_EXPIRED`) from `src/gateway/services/budget_reservation_ledger.py` to `src/gateway/models/budgets.py`, beside the `BudgetReservation.status` column whose values they are.

## Why

#1202 settled where a closed vocabulary lives: in the domain's models module, beside the string column it names. `ScopeType` and `ResetAlignment` moved there, and `docs/domains.md` now says so in the Models row. The reservation statuses are the one budgets vocabulary still outside that rule. The column's own comment in `models/budgets.py` points at the service module for its values, which is a model that depends on a service for its meaning.

## Suggested shape

The same shape the other two vocabularies have: a `Literal` first, a tuple derived with `get_args`, then each constant annotated with the `Literal` so mypy proves it is a member, and a line in `tests/unit/test_budget_vocabularies.py` asserting the constants cover the `Literal`. The ledger imports the names from the models module and does not re-export them. The trailing comments on the constants today restate the names and can go.

The statuses are not published in the OpenAPI spec, so this one carries no wire risk.

## When

Cheapest alongside the step of #1202 that moves `budget_reservation_ledger.py` into `services/budgets/`, since that change already touches every importer of the module.

## Relevant issues

Follow-up from #1202.

Issue #1415, opened 2026-09-21
Give the two scoped ceiling response models a shared base


peteski22:
## What

Give `ScopedBudgetResponse` and `OrganizationScopedBudgetPublic` a shared base. Both are in `src/gateway/schemas/budgets.py`. The organization model repeats all 21 fields of the deployment one and the whole body of its `from_model`, and adds one field, `manageable`.

`UpdateScopedBudgetRequest` and `OrganizationScopedBudgetUpdate` are the same case in miniature: two fields each, identical.

## Why

The two response models describe one thing, a spend ceiling and its live counters, on two surfaces. They lived in a route module and a service module until the budget models moved into one schemas module as part of #1202, so nobody saw them side by side. A field added to one and not the other is now the likely mistake, and nothing would catch it.

## Suggested shape

A base model that holds the common fields and the common part of `from_model`, in the way `OrganizationBudgetRates` already serves `OrganizationBudgetCreate` and `OrganizationBudgetUpdate` in the same module. The two public classes keep their names and their docstrings, because both are published as OpenAPI components.

## The constraint on the change

No change on the wire. Pydantic inlines inherited fields, and an unreferenced base emits no component of its own, so the spec should not move. That has to be proven and not assumed: `make openapi-check && make postman-check`, then `pnpm --dir web run client:generate && git diff --exit-code web/src/client/schema.ts`, all with no regenerated diff. Field order inside each published schema is part of what those checks compare.

## Relevant issues

Follow-up from #1202.

Issue #1414, opened 2026-09-21
Define the provider_key_id request constraints once, not three times


peteski22:
## What

Give the `provider_key_id` request field one definition. `src/gateway/schemas/budgets.py` writes the same constraints three times, on `CreateScopedBudgetRequest`, `OrganizationScopedBudgetCreate` and `WorkspaceMemberBudgetPolicyCreate`: `min_length=1`, `max_length=255` and `pattern=r"^\S+$"`.

## Why

The constraint is a correctness rule, not a style rule. A blank `provider_key_id` matches no provider instance, so a ceiling that stored one would be created, listed and never enforced. The three copies lived in three modules until the budget models moved into one schemas module as part of #1202, which is what made the repetition visible. A fix applied to one copy and not the other two leaves the hole open on the surface that was missed.

## Suggested shape

One `Annotated[str | None, Field(...)]` alias that carries the three constraints, used by all three models. The three `description` texts differ and are published, so each model keeps its own.

## The constraint on the change

No change on the wire. The field is published in `docs/public/openapi.json`, the Postman collection and `web/src/client/schema.ts`, and each has a CI drift check. Pydantic should inline the alias exactly as it inlines the fields today, but that has to be proven and not assumed: `make openapi-check && make postman-check`, then `pnpm --dir web run client:generate && git diff --exit-code web/src/client/schema.ts`, all with no regenerated diff.

## Relevant issues

Follow-up from #1202.

Issue #1193, opened 2026-09-15
Errors into one module per domain


peteski22:
### Problem

`services/tenancy/errors.py` is 1,294 lines of every tenancy error; features append to it.

### Proposal

Each error class moves to `exceptions/<domain>_exceptions.py`, one module per domain, the cq-exchange shape; the status-code handling that reads them is unchanged. A `tenancy/errors.py` shim re-exports for one release, then goes.

### Acceptance

- [ ] The API error responses are byte-identical (existing integration tests).
- [ ] No import of a moved error breaks across `src/gateway` or `tests`.

### Depends on

Nothing.

Part of #1171.

PR #1417, opened 2026-09-21
refactor(budgets): move the budget request and response models into schemas/budgets.py

peteski22:
## Description

The budget request and response models get a home of their own, `schemas/budgets.py`, and the two budget vocabularies get one home instead of three. Nothing changes on the wire: every path, status code, field, description and error message is the same as before, and the OpenAPI spec, the Postman collection and the generated dashboard client are untouched.

Python import paths do change, and there are no shims. The 20 models are imported from `gateway.schemas.budgets` and no longer from the two route modules or the two tenancy services. `ScopeType`, `SCOPE_TYPES`, the `SCOPE_*` constants, `ResetAlignment`, `RESET_ALIGNMENTS` and the `ALIGN_*` constants are imported from `gateway.models.budgets` and no longer from `scoped_budget_service` or `budget_periods`. Every importer in this repository is repointed. The otari-ai overlay was searched for each of those names and for every import from the changed modules: it imports `WorkspaceBudgetDefaultService`, `ApplicableBudget`, `BudgetScopeRequest` and `applicable_budgets`, which are all still where they were, and none of the moved names.

**Stacked on #1394.** The first seven commits shown here belong to that PR and drop out when it merges and this branch is rebased. The twelve commits of this change start at `refactor(budgets): give the reset-alignment vocabulary one home in models/budgets.py`. It targets `main` and not the #1394 branch because CI only runs for pull requests into `main`.

Until now the 20 budget models were defined inline in two route modules and two service modules. This is the first domain to move its models out, following the "one module per domain" shape in #1193:

```
src/gateway/
  models/budgets.py     ORM tables, plus ScopeType, ResetAlignment and their constants
  schemas/__init__.py
  schemas/budgets.py    the 20 request and response models
```

- **The vocabularies live in `models/budgets.py`, beside the columns they name.** `ScopeType` and `ResetAlignment` are the values two plain string columns accept, and enforcement and window derivation read them, so they are not wire shapes. With them in the models module, routes, schemas and services all import downward. `MAX_COUNT_LIMIT` already sits there and is imported the same way. Each constant is annotated with its `Literal`, so mypy proves it is a member. They stay `Literal` and do not become enums, because Pydantic publishes an enum as a named component, which would move the spec.
- **One scope vocabulary, not three.** `organization_budget_service.py` kept its own `OrganizationScopeType`, `ORGANIZATION_SCOPE_TYPES` and five constants, and `workspace_budget_default_service.py` kept two private copies, all to avoid an import cycle that #1325 removed. They are gone, and so are the bare strings that keyed `_SCOPE_SUBJECTS` in the deployment route. `grep -rn '"org_member"\|"api_token"' src/gateway` now finds `models/budgets.py` only.
- **`schemas/budgets.py` imports pydantic and the models layer, and nothing else.** To get there, `OrganizationBudgetPublic.from_model` takes `organization_id` from the service, which holds it at all three call sites, in place of reading it off the row and raising a not-found error when it was `None`. `from_model` refuses a row whose owner is anything else, the deployment's `None` included, with a `ValueError`: such a row can only arrive through a query that lost its filter, which is a defect and not a missing resource, so it must not answer 404, and failing there keeps another tenant's budget out of the response. All three call sites pass the organization they filtered on, so the response is unchanged, and the schema no longer needs a domain error.
- **No re-exports are left behind.** `budget_periods` and `scoped_budget_service` stop exporting the vocabulary, `scoped_budget_service` stops re-exporting `period_window`, and the two services stop exporting the models. Every importer in the repository is repointed in the commit that moves the name.
- Class names, fields, class docstrings and `Field` descriptions are untouched, because OpenAPI publishes them. Inline comments are not published, and most repeated a published description or the line beside them, so they are cut: about forty lines become a handful of sentences.

One rule is added to `scripts/check_architecture.py`, so naming it here: rule 16, "Schema boundaries". A module under `gateway/schemas/` may not import the API, service, repository or exception layers, or the composition root. `gateway.core` stays importable: it sits below the models layer that schemas build on, and request validation in other domains reads constants from it. It is a deny-list like the other fifteen rules, and it leaves database libraries to rules 11 and 12, which already own them. `schemas/` is a new layer and the check had no rule for it, so nothing would have stopped the next domain's schemas module importing a service or raising a domain error, which are the two things review caught in the first version of this change. `budgets.py` already passes. It is the last commit and stands alone, so it can be dropped without touching the rest.

`docs/domains.md` gains one clause in each of two rows of the layer table. Models: a models module also holds the closed vocabulary of a string column, which is what turns the placement above into a rule the next domain can follow. Schemas: a schemas module also maps ORM rows to its response models, which is what every `from_model` here does. `.github/instructions/backend-architecture.instructions.md` restates both for CodeRabbit.

One cost to know about: `services/budget_periods.py` used to import nothing from the gateway, and it now imports `gateway.models.budgets`, so importing it loads the ORM table registry. It still loads no FastAPI, no schemas and no exceptions.

No baseline in `scripts/check_architecture.py` changes. No module started or stopped importing SQLAlchemy, and no flat module was added.

## How to test it locally

```
make lint && make typecheck
env -u OTARI_API_KEY make test-unit
uv run pytest -n auto tests/integration
pnpm --dir web run client:generate && git diff --exit-code web/src/client/schema.ts
```

All of those pass on this branch, with no diff in either generated artifact or in the dashboard client, which is the proof that nothing moved on the wire. The one exception is `tests/integration/test_mcp_dependency_ceiling.py`, which needs network egress and fails the same way on `main` without it.

Three test modules carry the weight, all in `tests/unit/`:

- `test_budget_vocabularies.py` asserts three things. The named constants of each vocabulary cover its `Literal`. `_SCOPE_SUBJECTS` has a row for every scope type: mypy cannot prove a dict literal covers a `Literal`, and without the row a new scope would validate and then raise `KeyError` on create. And both ceiling create bodies, deployment and organization, refuse a scope type outside the vocabulary. That last test replaces `test_organization_budget_scopes.py`, which compared two rosters when there were two.
- `test_shared_helpers_pure.py` gains the `OrganizationBudgetPublic.from_model` cases: it names the owner it was given, and it refuses a deployment budget and another organization's budget.
- `test_check_architecture.py` covers the new rule in both directions: each forbidden layer is flagged, and an import of the models layer is clean.

To see the dependency direction for yourself: `uv run python -c "import sys, gateway.services.budget_periods; print('fastapi' in sys.modules, 'gateway.schemas.budgets' in sys.modules)"` prints `False False`.

## PR Type

- [ ] New Feature
- [ ] Bug Fix
- [x] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Part of #1202. Second instance of the shape in #1193, after #1394. Follow-ups found by moving the models side by side: #1414 (the `provider_key_id` constraints are written three times), #1415 (two near-identical ceiling response models), #1416 (the reservation status names still sit in a service). Raised in review of this PR: #1421 (decide whether the database should enforce the two vocabularies) and #1422 (prove the organization scope resolver handles every scope type).

## Checklist

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

Claude, via Claude Code.

I directed this work and reviewed it commit by commit before anything was pushed; Claude wrote the code under that direction. I set the plan, the commit structure and the comment standard, and sent the work back several times: for batching changes into too few commits, for comments that broke the project's rules, and for a stale route reference left in a touched comment. The first version put the vocabularies in `schemas/budgets.py`. Review rejected that placement for the dependency direction described above, and the branch was rebuilt with them in `models/budgets.py`.

## Summary

- Moved budget schemas into `gateway.schemas.budgets`.
- Moved budget vocabularies into `gateway.models.budgets`.
- Moved budget exceptions into `gateway.exceptions.budget_exceptions`.
- Updated imports while preserving API behavior and generated output.
- Added schema-layer architecture checks and documentation.
- Added tests for vocabulary coverage, validation, import boundaries, and exception ownership.
- Updated `OrganizationBudgetPublic.from_model` to receive `organization_id` from callers.

## Benefits

This creates a clearer budget domain structure, removes duplicated definitions, and prevents schemas from depending on higher application layers.


peteski22:
@coderabbitai review

scripts/check_architecture.py
```diff
@@ -148,6 +151,11 @@ class LayerRule(TypedDict):
         "forbidden": ["gateway.services", "gateway.api", "gateway.adapters"],
         "description": "Repositories",
     },
+    "gateway/schemas": {
+        "allowed": ["gateway.models"],
+        "forbidden": ["gateway.api", "gateway.services", "gateway.repositories", "gateway.exceptions"],
```
khaledosman: Rule 16's `forbidden` list leaves two gaps that the rule's stated intent covers. `allowed` is documentation only (the comment above `RULES` says so, and `check_file` reads `forbidden` alone), so a schemas module may import `gateway.core` and `gateway.container` and still pass. `gateway/types`, the nearest sibling leaf layer, forbids `gateway.core` for the same reason you give here, and `gateway.container` is the composition root.

```suggestion
        "forbidden": [
            "gateway.api",
            "gateway.services",
            "gateway.repositories",
            "gateway.exceptions",
            "gateway.core",
            "gateway.container",
        ],
```

`budgets.py` needs neither today, so this costs nothing and closes the hole before the next domain's schemas module finds it.

src/gateway/api/routes/scoped_budgets.py
```diff
@@ -11,22 +11,25 @@
...
+)
 from gateway.models.tenancy import Organization, OrganizationMember, Workspace, WorkspaceMember
-
-# ``ScopeType`` comes from the service that resolves a scope, not from a copy
-# here: the ``Literal`` is what puts the allowed values in the OpenAPI schema,
-# and a second roster would eventually let a client create a scope enforcement
-# does not know.
-from gateway.services.scoped_budget_service import ScopeType, period_window
+from gateway.schemas.budgets import CreateScopedBudgetRequest, ScopedBudgetResponse, UpdateScopedBudgetRequest
+from gateway.services.budget_periods import period_window
```
khaledosman, line 32: This repoint is right, but one importer of the same incidental re-export was missed: `tests/unit/test_scoped_budget_periods.py:13` still does `from gateway.services.scoped_budget_service import period_window`. That name is not in `scoped_budget_service.__all__` and never was; it is reachable only because the module imports it. Repoint it to `gateway.services.budget_periods` so the description's "no re-exports are left behind" holds for the whole tree.

src/gateway/schemas/budgets.py
```diff
@@ -0,0 +1,559 @@
...
+    def from_model(
+        cls,
+        budget: Budget,
+        *,
+        organization_id: uuid.UUID,
+        ceiling_count: int,
+    ) -> OrganizationBudgetPublic:
+        """Build the public form of a budget that the given organization owns.
+
+        NOTE: organization_id must be the organization that owns this budget.
```
khaledosman: Worth noting the inconsistency this leaves inside one file: `OrganizationScopedBudgetPublic.from_model` (line 429) takes the same `organization_id` parameter and *verifies* it, deriving `manageable` from `budget.organization_id == organization_id`, while this one takes it on trust and publishes it as the owner.

All three call sites are correct today (`list_budgets`, `create_budget` and `update_budget` each pass the organization they filtered the query on), so this is not a live bug. But the only thing standing between a future caller and a response that names the wrong owner is a `NOTE:`, and a plain `ValueError` keeps the invariant without reintroducing the domain-error import the move was about:

```python
if budget.organization_id is not None and budget.organization_id != organization_id:
    raise ValueError("organization_id does not own this budget")
```

khaledosman (approved):
Reviewed the increment over #1394 (`cf0e84c7..2a82fbc4`). No wire change confirmed: `openapi-spec` and the dashboard client drift check are both green, and the two vocabularies now have exactly one home (`grep` for the five scope strings and the three alignment strings finds `models/budgets.py` only). Three non-blocking points inline.

scripts/check_architecture.py
peteski22: Took `gateway.container`, left `gateway.core` out.

The composition root is never a legitimate import for a schema, so that one is free and now has a test case.

`gateway.core` is different from the `gateway/types` case. Types sit below everything, so core is above them. Schemas sit above models, and models already import core (`models/money.py` reads `gateway.core.metered_pricing`), so core is below the schemas layer and the import points the legal way. It would also bite the next domain: the inline provider request validation in `routes/providers.py` reads `RESERVED_PROVIDER_INSTANCE_NAMES` and `PROVIDER_TYPE_ALIASES` from `gateway.core.config`, and that code is what moves into `schemas/providers.py`. The commit message now records why core stays importable.

src/gateway/api/routes/scoped_budgets.py
peteski22: Fixed in the scope-vocabulary commit: `tests/unit/test_scoped_budget_periods.py` imports `period_window` from `budget_periods`, and the name is gone from `scoped_budget_service.__all__`, so nothing reaches through the enforcement service for it any more.

One correction for the record: `period_window` was in `scoped_budget_service.__all__`, at the base and at the reviewed commit, so it was a declared re-export and not an incidental one. Same conclusion either way.

src/gateway/schemas/budgets.py
peteski22: Took the concern, with a stricter check than the suggested one. The suggested guard skips a row whose `organization_id` is `None`, so a deployment budget passed in by mistake would still be published as owned by the organization, which is the exact case the old code refused. It is now `if budget.organization_id != organization_id: raise ValueError(...)`, with unit tests for both cases (a deployment budget, another organization's budget). The `NOTE:` is gone, since the code enforces it now.

On the comparison: `OrganizationScopedBudgetPublic.from_model` does not verify its argument. A ceiling's budget can legitimately belong to the deployment, and `manageable` publishes that state. So the two differ on purpose: one row must be owned, the other may not be.

PR #1418, opened 2026-09-21
fix(dashboard): offer the Playground the models the catalog lists

tbille:
## Description

The Playground's model picker and the Models page answered "which models can I use" differently. The picker read the gateway's flat OpenAI-compatible listing, which lists every provider offering separately and also includes aliases and routing policies; the Models page reads the grouped catalog, which folds offerings into one entry per model and deliberately leaves aliases and policies to the Routing page. A user in a workspace could therefore see the Playground offering names the Models page never showed.

The Playground now presents exactly what the Models page presents: one entry per model (for example `moonshotai/kimi-k3`), grouped by vendor, with nothing provider related in the list. Picking a model sends the catalog's model-level selector, so the gateway chooses the offering the same way the Models page's "Use model" drawer describes; while the gateway has not yet indexed the catalog, the first offering stands in so the model stays usable. The filter that hides non-chat models (embeddings, speech, moderation) inspects only the model's own name, so a vendor whose name happens to contain a word like "guard" cannot hide its chat models.

One deliberate behavior note: pinned models are stored by picker key, and the key respelling (offering selector to model selector) means pins made before this change no longer match a row. They are cheap to redo, and the stale rows are inert; they are not auto-dropped, because favorites are replaced as a whole list and pruning against a transiently empty or still-loading catalog would erase live pins.

## How to test it locally

Automated (all green locally):

```sh
pnpm --dir web run typecheck
pnpm --dir web run lint
pnpm --dir web test   # 6936 tests, includes new coverage for folded multi-provider models, the unindexed-catalog fallback, and vendor grouping
```

Manual: `make dashboard`, run a gateway with at least one provider plus a model alias or routing policy configured, sign in, and compare the Playground's model picker with the Models page. The picker lists the same model entries the Models page shows, grouped by vendor; provider offerings, the alias, and the policy do not appear (the latter two live on Routing).

No backend files changed, so `make lint` / `make typecheck` / `make test` were left to CI.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

None filed; fixes a directly reported mismatch between the Playground picker and the Models page.

## Checklist

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

Claude Fable 5 (Claude Code)

The change went through a three-round multi-model review (star-chamber: gpt-5.6, claude-opus-5, gemini-3.6-flash) before opening; the agreed findings (name-only chat filter, doc corrections, error normalization, shared test fixture) are applied. The second commit reshapes the picker from one-row-per-offering to one-row-per-model at the repo owner's request. The full web suite was re-run after rebasing onto main and after each commit.

## Summary

- Updated the Playground to use the grouped catalog shared with the Models page.
- Added one picker entry per catalog model, grouped by vendor.
- Kept aliases and routing policies on the Routing page.
- Fixed chat-model filtering for provider names such as `guard` and `embeddings`.
- Updated model captions to show the catalog vendor.
- Added catalog fixtures and tests for multi-provider models and name filtering.

## Technical notes

- Replaced `/models` data with `CatalogResponse` from `/catalog/models`.
- Used model-level selectors, with the first offering as a fallback for unindexed models.
- Preserved catalog order and removed duplicate model entries.
- Test execution status was not provided.


web/src/features/playground/helpers/playgroundModels.ts
```diff
@@ -96,28 +113,27 @@ export function groupPlaygroundModels(params: {
     (model) => !needle || model.key.toLowerCase().includes(needle),
```
khaledosman: Search still filters on `key`, but `key` is no longer a superset of what the picker renders. The rows show `label` (the catalog id) under a `vendor` heading, and `key` is now the model-level selector, so typing what you can see can return "No models match." Two cases, both reproduced against this head:

- Vendor `Z.ai` renders the heading `Z.ai` while the id is `z-ai/glm-5.3`; searching `Z.ai` yields 0 groups. Same for `Moonshot AI` vs `moonshotai`.
- While the catalog is unindexed, a row labeled `z-ai/glm-5.3` has key `fireworks:accounts/fireworks/models/glm-5p3`; searching `glm-5.3` yields 0 groups.

Before this PR the key contained the label, so matching the key always matched the visible text. Match `label` and `vendor` as well:

```ts
const matching = params.models.filter(
  (model) =>
    !needle ||
    `${model.label} ${model.vendor} ${model.key}`.toLowerCase().includes(needle),
)
```

web/src/features/playground/helpers/playgroundModels.ts
```diff
@@ -48,21 +58,25 @@ export function splitModelKey(key: string): {
...
-  catalog: ModelListResponse | undefined,
+  catalog: CatalogResponse | undefined,
 ): PlaygroundModel[] {
-  const chatKeys = (catalog?.data ?? [])
-    .map((entry) => entry.id)
-    .filter((id) => isChatModel(id))
-  return [...new Set(chatKeys)].map((key) => ({ key, ...splitModelKey(key) }))
+  const seen = new Set<string>()
+  return (catalog?.models ?? []).flatMap((model) => {
+    const key = model.selector ?? model.selectors[0] ?? ""
```
khaledosman, line 75: This respells the key that pins are stored under. `playground_favorite_model.model_key` holds whatever the picker passed to `onTogglePin`, which until now was `openai:gpt-4o`; from here it is `gpt-4o`. So every existing pin stops matching a row: the Pinned group disappears, the stars render unpinned, and the old rows stay behind unreferenced. Remembered selections and saved conversations survive this because both go through `pickInitialModel`, but pins do not.

Probably acceptable (a pin is cheap to redo), but it should be a stated decision rather than a side effect: either say so in the PR body, or drop the unmatched keys on the next `replace_favorite_models` write so the rows do not accumulate.

web/src/features/playground/PlaygroundConversation.tsx
```diff
@@ -48,6 +48,7 @@ export function PlaygroundConversation({
         ] as const
       ).map(({ name, panel, setPanel, other }) => {
         const identity = splitModelKey(panel.model)
+        const vendor = models.find((model) => model.key === panel.model)?.vendor
```
khaledosman: `vendor` comes from the found model but the two display strings beside it are still re-derived from the key by `splitModelKey`, which now parses a catalog id rather than `instance:model`:

- the caption falls through to `"Other"` whenever the catalog knows no vendor (`identity.instance` is `""` for a key with no colon), and `Other` under a model name states nothing;
- line 104's empty state says ``${identity.label}'s answer appears here.``, which under the unindexed fallback prints `accounts/fireworks/models/glm-5p3` while the picker above it says `z-ai/glm-5.3`.

The found model already carries the right string. Take `label` off it too and use that for the empty state, and let the caption render nothing rather than `"Other"` when there is no vendor; `splitModelKey` then leaves this file.

web/src/tests/fixtures.ts
```diff
@@ -36,6 +38,47 @@ import type {
   WorkspaceWebSearchConfig,
 } from "@/client"
 
+export function catalogModelSummary(
+  overrides: Partial<CatalogModelSummary> &
+    Pick<CatalogModelSummary, "id" | "selectors">,
+): CatalogModelSummary {
+  return {
+    name: overrides.id,
```
khaledosman, line 46: `selector` is `selector?: string | null` in the generated client, so a call that omits it compiles and silently produces the unindexed catalog, not the ordinary one. `buildPlaygroundModels` then keys the model by `selectors[0]`, which is the state this PR added a fallback for rather than the normal state. Default it here (`selector: overrides.id`) so a test opts into the unindexed path by passing `selector: null`, the way `playgroundModels.test.ts` already does.

khaledosman (approved):
Catalog parity is the right call and the folding reads well. Four inline notes; the search one is the only user-visible regression.

web/src/features/playground/helpers/playgroundModels.ts
```diff
@@ -96,28 +113,27 @@ export function groupPlaygroundModels(params: {
     (model) => !needle || model.key.toLowerCase().includes(needle),
   )
   const pinnedModels = matching.filter((model) => pinned.has(model.key))
-  const byInstance = matching
+  const byVendor = matching
```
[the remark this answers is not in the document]
tbille: Addressed in 3b7acdac: the search now matches key, label, and vendor, with a test for a vendor the key does not contain.

web/src/features/playground/helpers/playgroundModels.ts
tbille: Addressed in 3b7acdac: the search matches key, label, and vendor, which covers both cases (Z.ai via the vendor field, glm-5.3 via the label under the unindexed fallback).

web/src/features/playground/helpers/playgroundModels.ts
tbille: Stated in the PR body as a deliberate decision: old pins stop matching and are cheap to redo. Not auto-dropped, because favorites are replaced as a whole list and pruning against a transiently empty or still-loading catalog would erase live pins.

web/src/features/playground/PlaygroundConversation.tsx
tbille: Addressed in e22a2b71: the caption and empty state now read the found model's vendor and label, the caption stays empty when the catalog knows no vendor, and splitModelKey left the file.

web/src/tests/fixtures.ts
tbille: Addressed in e22a2b71: the fixture defaults selector to the id, so a test opts into the unindexed path with selector: null.

PR #1423, opened 2026-09-21
fix(hybrid): accept the standalone credential headers on the platform path

tbille:
## Description

A key that works against a standalone Otari gateway stopped working the moment the same caller pointed at a hybrid deployment: hybrid mode read only the `Authorization: Bearer` header, so the `Otari-Key` header the dashboard teaches, and the `x-api-key` header Anthropic-native clients send by default, were silently ignored and answered with a misleading `401 Missing authentication token`. This was hit in practice by a hosted-onboarding user pasting the dashboard's own curl snippet.

Hybrid mode now accepts the same three credential headers as standalone mode across the generation APIs, the MCP endpoints, and the hook server, so the deployment mode no longer matters to the caller. The token is still forwarded opaquely to the platform as `X-User-Token`, and the platform remains the one verifying it. When no credential is present at all, the 401 now names the three accepted headers instead of the generic detail.

## How to test it locally

```sh
make test  # tests/integration/test_hybrid_mode_* cover the new header forms
```

Or by hand, against a gateway running in hybrid mode:

```sh
curl http://localhost:8000/api/v1/chat/completions \
  -H "Otari-Key: <otari.ai user token>" \
  -H "Content-Type: application/json" \
  -d '{"model": "openai:gpt-4o-mini", "messages": [{"role": "user", "content": "Hello"}]}'
```

Before this change that returned `401 {"detail": "Missing authentication token"}`; now it behaves exactly like the `Authorization: Bearer` form (and `x-api-key` works too). A new parametrized integration test (`test_hybrid_mode_accepts_standalone_credential_headers`) asserts each header form forwards the token to the platform resolve call as `X-User-Token`.

Already proven by automated checks: `make lint`, `make typecheck`, the full unit suite (4230 passed), and the hybrid-mode + auth-adjacent integration suites. `scripts/generate_openapi.py` produces no spec diff.

The dependency signatures survive the cleanup: `verify_api_key` keeps its unused `config` parameter because otari-ai calls it directly with three arguments (`backend/overlay/gateway_runtime/routes.py`), so the shape is API. The hybrid-only extractor wrapper is gone entirely; every mode now reads credentials through one public `extract_credential_token` helper, which also strips surrounding whitespace and answers a whitespace-only credential as missing in every mode.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

None filed; found while debugging a hosted-onboarding 401 (the dashboard's curl snippet uses `Otari-Key`, which hybrid mode did not read).

## Checklist

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

Claude Code (Claude Fable 5)

Diagnosed from a live hosted-onboarding failure, implemented, and tested in a supervised Claude Code session.

## Summary

Hybrid deployments now accept `Authorization: Bearer`, `Otari-Key`, and `x-api-key` credentials across generation APIs, MCP endpoints, and the hook server.

Credential extraction is shared across deployment modes. It trims surrounding whitespace and rejects empty credentials with a clear 401 response. Hybrid credentials are forwarded as `X-User-Token` for platform verification.

Documentation and integration tests now cover all supported header forms and endpoints.


src/gateway/api/deps.py
```diff
@@ -116,7 +116,7 @@ def get_enabled_features(request: Request) -> tuple[CoreFeature, ...]:
     return enabled
 
 
-def _extract_bearer_token(request: Request, config: GatewayConfig) -> str:
+def _extract_bearer_token(request: Request) -> str:
```
khaledosman: The name is now the bug's own cause left in place. This reads `Otari-Key`, `Authorization` and `x-api-key`, and as of this PR it is imported from another module, but it is still spelled `_extract_bearer_token`: private, and named for one of the three headers it handles. That name is a fair guess at why the hybrid path only ever looked at `Authorization`.

Rename to a public `extract_credential_token` and update the two call-site modules plus `tests/unit/test_extract_bearer_token.py`.

src/gateway/api/routes/_platform.py
```diff
@@ -466,19 +467,16 @@ def _mark_locked_in(_pos: int = attempt.position) -> None:
...
+    there).
     """
-    auth_header = request.headers.get("Authorization")
-    if not auth_header or not auth_header.startswith("Bearer "):
-        raise HTTPException(
-            status_code=status.HTTP_401_UNAUTHORIZED,
-            detail="Missing authentication token",
-        )
-    token = auth_header[7:].strip()
+    token = _extract_bearer_token(request).strip()
```
khaledosman: Two divergences survive here, both the shape this PR exists to remove.

`.strip()` is applied on the hybrid path and not in the shared helper, so `Otari-Key: "tk_abc "` authenticates against a hybrid gateway and 401s against a standalone one. Move the strip into the helper (or drop it) so the two modes read the same bytes.

And the fallback detail two lines down is still `"Missing authentication token"`, the exact string this PR replaced everywhere else. It is reachable: a whitespace-only `x-api-key`, or `Otari-Key: "Bearer "`, lands there and gets the old misleading message while a wholly absent header gets the new one. Reuse the helper's `f"Missing {API_KEY_HEADER}, Authorization, or {X_API_KEY_HEADER} header"`.

src/gateway/api/deps.py
```diff
@@ -349,14 +349,12 @@ def _api_key_format(request: Request, db: AsyncSession) -> ApiKeyFormatPort:
 async def verify_api_key(
     request: Request,
     db: Annotated[AsyncSession, Depends(get_db)],
```
khaledosman, line 371: `verify_api_key` loses its `config` parameter. Unused here so the removal is right, but it is a public dependency and this is a signature change: `Depends(verify_api_key)` is unaffected, a direct three-argument call is a `TypeError` at request time. `verify_api_key_or_master_key` is called directly in `messages.py`, so the pattern exists in this tree. Worth confirming otari-ai has no direct caller, and saying so in the description the way #1417's does.

docs/api-reference.md
```diff
@@ -33,8 +34,8 @@ organization: `/api/v1/organizations/me/usage` for usage, and
 `/api/v1/organizations/me/keys` for a member's own API keys.
 
 In hybrid mode, the generation APIs and the `/api/v1/mcp` endpoints accept an
```
khaledosman: This names the generation APIs and `/api/v1/mcp`, but the hook server changed too (`verify_hook_caller` reaches the same helper), so the doc now under-describes the fix. Add `/api/v1/hooks`.

Same file for the other half: `routes/hooks.py`'s `verify_hook_caller` docstring still says hybrid "only requires a bearer token to be present", which is one header of the three as of this change. It is outside the diff, so flagging rather than suggesting.

khaledosman (approved):
Right fix, and routing both modes through one helper is what makes the parity hold rather than be asserted. Four non-blocking points inline; three of them are residual divergences of the same kind this PR is removing.

src/gateway/api/deps.py
tbille: Renamed in 4cbf1b75: the helper is now a public `extract_credential_token`, the hybrid wrapper is deleted so every caller goes through it directly, and the test file moved to `tests/unit/test_extract_credential_token.py`.

src/gateway/api/routes/_platform.py
tbille: Both closed in 4cbf1b75. The strip moved into the shared helper, and a credential that strips to nothing is answered with the standard `Missing Otari-Key, Authorization, or x-api-key header` detail there too; "Missing authentication token" is gone from the tree. With that, the hybrid wrapper had no body left, so it was removed entirely; the divergence cannot reopen because there is no second path to diverge. New unit tests cover the padded-token and whitespace-only cases.

src/gateway/api/deps.py
tbille: Good catch: otari-ai does call it directly. `backend/overlay/gateway_runtime/routes.py` invokes `verify_api_key(request, db, config)` with three positional arguments, so the removal would have been a request-time `TypeError` at the next submodule bump. 4cbf1b75 restores the `config` parameter with a docstring note that the three-argument shape is API (same reason `verify_api_key_or_master_key` keeps it), and the PR description now says so.

docs/api-reference.md
tbille: Both fixed in 4cbf1b75: the api-reference sentence now names `/api/v1/hooks` alongside the generation APIs and `/api/v1/mcp`, and the `verify_hook_caller` docstring says "credential header" instead of "bearer token".

src/gateway/api/deps.py
```diff
@@ -116,33 +116,43 @@ def get_enabled_features(request: Request) -> tuple[CoreFeature, ...]:
...
-        return auth_header[7:]
-
-    raw_token = request.headers.get(X_API_KEY_HEADER)
-    if raw_token:
-        return raw_token
+        token = value[7:] if value.startswith("Bearer ") else value
+    else:
+        auth_header = request.headers.get("Authorization")
+        if auth_header:
+            if not auth_header.startswith("Bearer "):
```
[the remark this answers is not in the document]
tbille: Fixed in d7b762bf: leading whitespace comes off both header values before the Bearer scheme check, so a padded `Otari-Key` returns the bare token and a padded `Authorization` no longer 401s as invalid format. Added a parametrized test covering both header forms with `" Bearer tk_padded "`.

src/gateway/api/deps.py
```diff
@@ -119,33 +119,48 @@ def get_enabled_features(request: Request) -> tuple[CoreFeature, ...]:
...
+        # Leading whitespace comes off before the scheme check, so a padded
+        # value still has its Bearer prefix recognized rather than kept as
+        # part of the token.
+        value = value.lstrip()
+        token = value[7:] if value.startswith("Bearer ") else value
+    else:
+        auth_header = request.headers.get("Authorization")
+        if auth_header:
+            auth_header = auth_header.lstrip()
+            if not auth_header.startswith("Bearer "):
```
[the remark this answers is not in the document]
tbille: Fixed in the latest commit: the scheme check now applies only to a non-empty value, so a whitespace-only `Authorization` falls through to the shared missing-credentials 401 like the other two headers, and `invalid_format` is reserved for a real non-Bearer scheme. Added `{"Authorization": "   "}` to the whitespace-only parametrized test.

PR #1427, opened 2026-09-21
feat(dashboard): show providers and model makers with their own marks

jigjigjig:
## Description

Providers and model makers now appear with their own mark beside their name, instead of a bare name on its own. On the Models page the filter rail's Providers and Vendors lists each carry the company's mark, and so do the "by OpenAI" line on a model card and the maker line under a model's name in the table view. A company with no mark available gets a lettermark tile rather than a blank or a broken image, so a list stays aligned and readable either way.

This is the second half of the naming work: the display names landed earlier, and this adds the marks. It is a display change only.

The path data is fetched after the page paints rather than with it. The list of which companies have a mark is small and answers whether a row needs a slot at all; the drawings are 64 kB and only matter once something is on screen, so they load on demand into a reserved box that does not move the row when it fills. Nothing about how a provider is identified, configured or matched changes, and every place a reader needs the underlying id still shows it.

Scoped deliberately to the Models page. The presentation layer it adds is general, but the other surfaces that show a provider (the Providers page, activity and usage rows, the model pickers, the organization provider keys page) are unchanged and still read as they did. Issue 990 stays open as the tracker for those.

Two things a reviewer should know. The marks are monochrome and take the page's text ink rather than each vendor's brand color: of the brand colors the sources carry, several fall below the contrast floor on one of the two theme grounds and two are pure white. And the marks ship on the understanding that permission has been granted for each vendor's mark; the trademark position was reviewed separately and is not settled by this PR.

## How to test it locally

1. `make dashboard` (or `pnpm --dir web run build`), then run the gateway and open `/#/models`.
2. In the filter rail, expand **Providers** and **Vendors**. Every row should carry a 16px mark or a lettered tile, aligned in one column. A provider instance you named yourself tiles under its own initial.
3. Switch between **List** and **Table**. On a card, the mark sits before "by <maker>"; in the table it sits before "<maker> · N providers". Both at 14px.
4. Check both themes. Marks are ink-colored, so they invert with the theme rather than staying one color.
5. Rows with no maker read "Unknown vendor" with no mark and no tile, which is deliberate: an initial there would name a company that does not exist.

Automated coverage: `pnpm --dir web test` (the resolution tables, the tile fallback, the slot rule, and the four call sites), `pnpm --dir web run typecheck`, `pnpm --dir web run lint`. Each new case was verified by removing the behavior and watching the case fail, rather than only by watching it pass.

What is left to eyeball: how the marks read at 14px on a real catalog, since the automated checks assert that a mark renders and not that it is legible.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Part of #990. Deliberately does not close it: this ships the presentation layer and the Models page, and the issue tracks the remaining surfaces.

## Checklist

Three boxes are unticked rather than ticked, with reasons instead. No documentation changed: this adds no concept a reader of the docs has to learn, and the rules that govern the marks live in the modules' own docblocks where the next person to edit them will be. The API contract did not change, so no spec regeneration is owed; no route, schema or docstring is touched and neither generated dashboard artifact moved. And no architecture rule changed.

The dashboard checks are the ones that apply here (`pnpm --dir web run lint`, `typecheck`, `test`); `make lint` and `make test` were run as well and are unaffected, since no Python changed.

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

The mark geometry is vendored path data rather than a dependency. `@lobehub/icons` is the source for most of it (MIT, Copyright (c) 2023 LobeHub) and is credited in the module header; depending on it directly would pull in `antd-style` and peer-require `antd` and `@lobehub/ui`, a second design system this dashboard has no use for. Two marks come from Simple Icons through `react-icons`, already a dependency, because LobeHub ships no Databricks or OVH mark. One comes from the vendor's own repository.

One known smell, called out rather than hidden: the lettermark tile borrows `text-shell-monogram` for its 9px step. That role is already exactly this value and its rationale is the same (a single glyph in a small box is recognized rather than read, which no step on the type scale carries), but its name says "shell" and this is not the shell. Declaring a third off-scale exception in the foundation seemed worse than borrowing one; renaming the role is the fix if it reads wrong, and that is a separate change.

**NOTE:**

## Summary

Adds monochrome provider and vendor marks to the Models page and model detail offerings.

- Shows marks in provider and vendor filters, model cards, table rows, and offering tables.
- Uses letter tiles when no mark is available.
- Shows “Unknown vendor” without a mark when no vendor exists.
- Adds shared mark components, lookup helpers, and automated tests.

The change affects display only. Provider configuration, matching, IDs, and other provider surfaces remain unchanged.


web/src/shared/components/marks/BrandMark.tsx
```diff
@@ -0,0 +1,175 @@
...
+      // is borrowed rather than declared again at a third name: its docblock's
+      // reasoning is this tile's exactly, one glyph in a small box being
+      // recognized rather than read. If the "shell" in it reads wrong here,
+      // renaming the role is the fix.
+      //
+      // Close to `design-system/indicators/Avatar`, deliberately not it: that
+      // one is 20px or 26px with a sans semibold monogram, where these are 16px
+      // and 14px in mono, and reaching for it would mean overriding a
+      // primitive's own size from a call site.
+      className={`${box} text-shell-monogram flex shrink-0 items-center justify-center border border-control-border bg-surface-alt font-mono font-medium uppercase`}
```
jigjigjig, line 91: Known smell, flagged rather than hidden: this borrows `text-shell-monogram` for its 9px step, and the role's name says "shell" while this is not the shell.

The value and the rationale are identical (one glyph in a small box is recognized rather than read, which no step on the type scale carries), and the alternative was declaring a third off-scale exception in the foundation, which its own docblock sets a bar against. `foundation.test.ts` also rejects any `text-[<digit>` at a call site, so writing 9px here was not an option.

If the name reads wrong, renaming the role is the fix, as its own separate change.

web/src/shared/helpers/brandMarks.ts
```diff
@@ -0,0 +1,808 @@
+/**
+ * Each provider's own mark, as vendored path data.
+ *
+ * The companion to {@link providerDisplayName}: that answers what a provider is
+ * called, this answers what it looks like. Same display-layer rule, so nothing
+ * here changes an id, and every surface that needs the id keeps showing it.
+ *
+ * The geometry is vendored rather than pulled from a package. `@lobehub/icons`
```
jigjigjig, line 15: Vendoring path data rather than adding `@lobehub/icons` is the main structural decision here.

The package peer-requires `antd` and `@lobehub/ui`, a second design system, for geometry we consume as a few dozen static paths. Its `Mono` and `Color` modules import nothing but React, so a deep import would work, but it leaves two unmet peers in the lockfile for 40 files used.

The cost is that a mark upstream fixes does not reach us until somebody re-vendors it. Attribution is in this header.

web/src/shared/helpers/brandMarks.ts
```diff
@@ -0,0 +1,808 @@
...
+  viewBox: string
+  /** A wrapping transform the source needed. Absent for everything on a 24 grid. */
+  transform?: string
+  shapes: readonly ProviderMarkShape[]
+  /** Where the geometry came from, for the attribution above. */
+  source: string
+}
+
+/**
+ * `bedrock` and `sagemaker` are deliberately absent, so they tile.
```
jigjigjig, line 52: An exception to the rule the file otherwise follows, so worth stopping on.

Every other AWS-credential id would take the AWS mark. The only AWS mark in the source set is the lowercase wordmark, which is illegible at 16px and worse at 14px. Verified by rendering it: at 16px in the filter rail it is a grey smudge.

So these two tile, and `maker:amazon` is left out of `MAKER_MARKS` to match, which is what keeps a card and a rail row from disagreeing about Amazon.

web/src/shared/helpers/brandMarks.ts
```diff
@@ -0,0 +1,808 @@
...
+ * above: the geometry is already written down once in the design system, and a
+ * second copy here would be a third place to edit when the mark moves. So this
+ * names them and the component reaches for that primitive.
+ *
+ * They are in the display-name map for the reason every vendor is, and they are
+ * marked for the same reason: a deployment serving its own models would
+ * otherwise show a lettermark tile while every third party got a logo, which is
+ * the bug this pair of modules exists to fix, pointed at us.
+ */
+const PRODUCT_MARK_IDS: readonly string[] = ["mzai", "otari"]
```
jigjigjig: Why our own ids are named here instead of carrying geometry like everything else: `design-system/ProductMark` already holds that path at `currentColor`, and a second copy would be a second place to edit when the mark moves.

That split is the reason `hasProviderMark` exists next to `providerMark`. The slot decision asks "is there a mark", which is true for these two, while `providerMark` answers "what shapes", which is `undefined`. A caller using `providerMark` alone would under-count them and drop the slot on a list that then draws one.

web/src/shared/helpers/brandMarks.ts
```diff
@@ -0,0 +1,808 @@
...
+ *
+ * The vocabulary is closed: a model's vendor is voted from `_ORG_VENDORS` and
+ * `_KEY_VENDORS` in `services/model_identity.py` and from nowhere else, so a
+ * new maker is a change in this repository rather than a dataset refresh.
+ *
+ * Amazon is deliberately absent. Its mark in the source set is the aws
+ * wordmark, which is unreadable at 14px, so it tiles like the Bedrock and
+ * SageMaker rows it shares a vendor with.
+ */
+const MAKER_MARKS: Readonly<Record<string, ProviderMarkGlyph>> = {
```
jigjigjig, line 689: Two tables rather than one, which is the decision most likely to read as duplication.

The key spaces disagree on the same companies: provider `mistral` against maker slug `mistralai`, `zai` against `z-ai`, `moonshot` against `moonshotai`. A single table keyed either way silently misses all three, and the miss looks like a missing logo rather than a bug. The glyphs are shared; only the keys differ, and `brandMarks.test.ts` pins the three disagreeing pairs.

web/src/features/models/ModelCatalogPage.tsx
```diff
@@ -234,28 +251,53 @@ function FilterRail({
...
             isSelected={filters.vendors.includes(option.value)}
             onChange={(isOn) =>
               set("vendors", toggle(filters.vendors, option.value, isOn))
             }
           >
-            {option.label}
+            {vendorsMarked ? (
+              <span className="flex items-center gap-2">
+                {/* The unknown bucket is not a company, so it takes no mark and
+                    no tile: an initial for "Unknown vendor" would name a vendor
```
jigjigjig, line 285: Deliberate asymmetry, since a reviewer will notice the two unknown cases behave differently.

In the rail the unknown row keeps its 16px slot and draws neither mark nor tile: an initial would name a company called U, and dropping the slot would misalign one row of an aligned column. On the card and the table sub-line the unknown row holds no slot at all, because that line is prose next to a maker's name rather than a column, and an empty box before "by unknown vendor" reads as a mark that failed to load.

web/src/features/models/catalog.ts
```diff
@@ -341,14 +341,53 @@ export const SORT_OPTIONS: {
...
+ * The vendor slug a row's catalog id carries, or `undefined` when it carries
+ * none.
+ *
+ * The gateway builds a model's id as `vendor_slug(vendor)/slug` and as the bare
+ * slug where no vendor could be named (`model_identity.CatalogIdentity`), so the
+ * prefix already is the normalized vendor and reading it off is what keeps a
+ * second copy of those rules out of this codebase. Both halves are required:
+ * `vendor` says a prefix was added, the separator says one is there to take.
+ *
+ * The prefix cannot contradict the vendor, which is what makes reading it safe:
```
jigjigjig, line 354: This paragraph is the load-bearing part of the keying, so it is worth a reviewer's eye.

Reading the vendor off the id's prefix is only safe because the prefix cannot contradict `vendor`: `CatalogIdentity.id` is a property computed from `vendor` (`model_identity.py:301`) and the response sets both from that one object.

The alternative was mirroring `vendor_slug` in TypeScript, which would be a second implementation of rules like `Z.ai` to `z-ai` for the two to drift apart on.

jigjigjig (commented):
Self-review before anyone else reads this. No blockers found; the comments are on the decisions a reviewer would stop at rather than on the mechanics.

Checked against the repo gates: no route, schema or docstring is touched, so no OpenAPI or Postman regeneration is owed; `schema.ts` and `routeTree.gen.ts` are untouched; no dependency is added, so no lockfile or `allowBuilds` change; no page is added, so no screenshot entry; `make lint` (architecture check then Ruff) and the dashboard's own lint, typecheck and 6998 tests all pass on this commit.

Two things I measured rather than assumed. The built CSS hash came out identical to main's, which is exactly what an unemitted utility class looks like, so I confirmed all nine classes used here emit real declarations. And on the 390px phone viewport the page reports 0px horizontal overflow with a 44px rail row, which is the touch floor exactly.

Each new test case was verified by removing the behavior and watching it fail. One of them was an empty green when first written and is noted in a comment below.

web/src/shared/helpers/brandMarks.ts
```diff
@@ -0,0 +1,808 @@
...
+/**
+ * `bedrock` and `sagemaker` are deliberately absent, so they tile.
+ *
+ * Both are AWS credentials and the rule would give them the AWS mark, but the
+ * only AWS mark in the source set is the lowercase wordmark, which is not
+ * legible at 16px and less so at 14px. A lettermark "A" says as much and stays
+ * readable. `maker:amazon` is left out of {@link MAKER_MARKS} for the same
+ * reason, so the provider rows and the maker row agree.
+ */
+const PROVIDER_MARKS: Readonly<Record<string, ProviderMarkGlyph>> = {
```
khaledosman, line 609: **The Models route chunk grew 5x.** Measured from the two CI builds rather than estimated: `ModelCatalogPage-*.js` is **20.58 kB -> 103.10 kB** raw, **7.58 -> 38.38 kB gzip** (main run 35585640345 vs. this PR's 35586176184). That makes the Models page the third-largest first-party chunk, after the entry and Markdown.

`performance.md:176` calls a route chunk that *doubles* a review finding, and every visitor now pays for all 109 glyphs while a deployment shows maybe five providers. `performance.md:56` prescribes the fix: put the two tables behind a dynamic `import()` inside the mark components, so the marks fill in after paint instead of gating it.

web/src/shared/helpers/brandMarks.ts
```diff
@@ -0,0 +1,808 @@
...
+ *
+ * Matched the way {@link providerDisplayName} matches, on a trimmed
+ * lower-cased id, because an id reaches the dashboard from a catalog row, a
+ * stored credential and a model selector's prefix, and those do not agree on
+ * case.
+ */
+export function providerMark(
+  providerId: string,
+): ProviderMarkGlyph | undefined {
+  return PROVIDER_MARKS[providerId.trim().toLowerCase()]
```
khaledosman: **A prototype key crashes the page.** Indexing an object literal resolves inherited keys, so `providerMark("constructor")` returns `Object`, `hasProviderMark` then returns `true`, and `Glyph` throws `Cannot read properties of undefined (reading 'map')` on `glyph.shapes`.

Provider ids here are operator-chosen instance names, and `core/config.py:1619` rejects only `:` and `/`, so `constructor`, `toString` and `valueOf` are all legal names. `providerDisplayName` has the same hazard today but only degrades; this turns it into a white screen.

```ts
const key = providerId.trim().toLowerCase()
return Object.hasOwn(PROVIDER_MARKS, key) ? PROVIDER_MARKS[key] : undefined
```

Or declare both tables on `Object.create(null)`. Same at line 807 for `makerMark`.

web/src/shared/helpers/brandMarks.ts
```diff
@@ -0,0 +1,808 @@
...
+}
+
+/**
+ * Each model maker's mark, keyed on the vendor slug the catalog id carries.
+ *
+ * A second key space rather than a second name for the first, because the two
+ * disagree on the same companies: the maker `Mistral AI` slugs to `mistralai`
+ * where the provider id is `mistral`, `Z.ai` to `z-ai` against `zai`, and
+ * `Moonshot AI` to `moonshotai` against `moonshot`. One table would miss all
+ * three. The glyphs are shared; only the keys differ.
```
khaledosman, line 679: **"The glyphs are shared" is not what the code does.** They are copied: 28 of the 109 path entries are duplicates, 15.5 kB of the added bytes. The Azure geometry appears verbatim three times (lines 72, 89, 106), and each company present in both tables carries two copies.

Only `mistralai` and `openai` are pinned by `brandMarks.test.ts:134`, and with `toEqual` rather than identity, so editing `PROVIDER_MARKS.anthropic` and not `MAKER_MARKS.anthropic` fails nothing and ships one company drawn two ways.

Hoist each glyph to a named const and reference it from both tables. The docblock then becomes true, and finding 1 gets 15 kB cheaper.

web/src/shared/helpers/brandMarks.test.ts
```diff
@@ -0,0 +1,185 @@
...
+  it("tiles the two makers that have no usable mark", () => {
+    // Amazon's only mark in the set is the aws wordmark, unreadable at these
+    // steps, so it tiles like the Bedrock and SageMaker rows. OpenBMB has none.
+    expect(makerMark("amazon")).toBeUndefined()
+    expect(makerMark("openbmb")).toBeUndefined()
+  })
+
+  it("covers every maker the gateway can report", () => {
+    // The vocabulary is closed, so this is the whole set. A maker added to
+    // `model_identity.py` without a mark shows up here rather than as a silent
```
khaledosman: **This claims a gate that does not exist.** `MAKER_SLUGS` is hand-typed twenty lines below, so it changes only when someone remembers to change it; a maker added to `model_identity.py` shows up as a silent tile, which is the outcome the comment says it prevents.

The list is accurate today (exactly `_ORG_VENDORS` union `_KEY_VENDORS`), which is what makes the stale copy plausible later. Either drop the claim, or add the Python-side assertion that those two dicts' values slug to this list, the pattern `tests/unit/test_frontend_rule_coverage.py` already uses for the same problem.

web/src/features/models/ModelCatalogPage.tsx
```diff
@@ -408,11 +456,22 @@ const TABLE_COLUMNS: DataTableColumn<CatalogModelSummary>[] = [
...
-        <span className="text-caption">
-          {row.vendor ?? "Unknown vendor"} ·{" "}
-          {row.provider_count === 1
-            ? "1 provider"
-            : `${row.provider_count} providers`}
+        <span className="flex items-center gap-1.5 text-caption">
+          {/* No slot held when the maker is unknown: the sub-line is prose on
+              one row of a table whose other rows carry a mark, and an empty box
+              in front of "Unknown vendor" reads as a mark that failed to load. */}
+          {makerKeyOf(row) === undefined ? null : (
```
khaledosman: `makerKeyOf(row)` runs twice, and the `?? ""` on the next line re-widens a value this guard just proved defined. Give the cell a block body and one `const makerKey`, as `ModelCard` does at line 368.

web/src/shared/components/marks/BrandMark.tsx
```diff
@@ -0,0 +1,175 @@
...
+
+const BOX: Record<BrandMarkStep, string> = {
+  16: "size-4",
+  14: "size-3.5",
+}
+
+/**
+ * The geometry, drawn.
+ *
+ * Decorative, so it is hidden from the accessibility tree rather than labelled:
```
khaledosman: US spelling: "labelled" -> "labeled". AGENTS.md names this one explicitly. Same in `BrandMark.test.tsx:23`, "colour" -> "color".

web/src/shared/helpers/brandMarks.ts
```diff
@@ -0,0 +1,808 @@
+/**
```
khaledosman, line 1: Worth defining "mark" once here. The module rests on mark / wordmark / lettermark being distinct terms, which is why Amazon is excluded at line 44 (the only AWS asset in the source set is the wordmark) and why `Lettermark` is named as it is. The PR description mixes "mark" and "logo" for the same thing, so the reader arrives without the distinction. One sentence.

khaledosman (approved):
Nothing blocking. Findings 1 and 2 are the substantive ones; 3 is cheap and pays for part of 1.

_Reviewed by Claude Opus 5 via [Claude Code](https://claude.com/claude-code)._

web/src/shared/helpers/brandMarks.ts
khaledosman: if the size is due to the logo images, can't we use svgs for these where possible? we could also use webp or mozjpeg for smaller compressed images

jigjigjig (commented):
Self-review extended to the second commit, which adds the mark to the model detail page's offerings table.

That column is site 2 of the audit and was outside the original scope; it came in because the page is under `/models` and a reader following a card into the detail page would otherwise watch the marks disappear.

Gates on this commit: typecheck clean, lint clean, 6999 tests pass. The new case asserts every offering row carries an `svg`, and it was break-tested by removing the mark, since a table row could hold an `svg` for other reasons and the case would then have been an empty green.

Also verified on a real deployment rather than a fixture: the built bundle on a local env serves this commit and the offerings row renders the provider's mark against the env's own seed.

web/src/features/models/ModelDetailPage.tsx
```diff
@@ -208,15 +209,25 @@ function offeringColumns({
...
-              : ""}
-            {credentialLabel(row.credential)}
-            {row.quantization ? ` · ${row.quantization}` : ""}
+        <span className="text-body flex items-center gap-2 whitespace-nowrap">
+          {/* Keyed on the type rather than the instance, which is what the
+              instance is named after until an operator renames it. The name
+              stays the instance either way, and the type is still spelled out
+              below when the two differ. */}
+          <ProviderMark
+            providerId={row.provider_type || row.provider}
```
jigjigjig, line 218: `||` rather than `??` on purpose. `provider_type` is `string` and not nullable in `CatalogOffering`, so `??` would never fire; what this guards is the empty string, which is the shape an offering with no type recorded would arrive as. With `??` an empty type would key the lookup on `""`, miss, and tile a provider that has a perfectly good mark under its instance name.

Seen live on the deployed env: an instance whose type equals its name resolves through this unchanged.

web/src/features/models/ModelDetailPage.tsx
```diff
@@ -208,15 +209,25 @@ function offeringColumns({
...
-        <span className="text-body whitespace-nowrap">
-          {providerDisplayName(row.provider)}
-          <span className="text-caption">
-            {" · "}
-            {row.provider_type !== row.provider
-              ? `${providerDisplayName(row.provider_type)} · `
-              : ""}
-            {credentialLabel(row.credential)}
-            {row.quantization ? ` · ${row.quantization}` : ""}
+        <span className="text-body flex items-center gap-2 whitespace-nowrap">
```
jigjigjig, line 212: This cell gained a wrapper, which is worth a reviewer's eye because it is a row header.

The mark has to be a flex sibling of the text, so the text moved into its own inner `span` and the outer one became the flex row. `whitespace-nowrap` stays on the outer element, which is what the original comment above relies on: the selector is as long as the provider makes it and must not set the row height.

The cell renders the same three parts in the same order as before.

web/src/features/models/ModelDetailPage.tsx
```diff
@@ -208,15 +209,25 @@ function offeringColumns({
...
+          {/* Keyed on the type rather than the instance, which is what the
```
jigjigjig, line 213: Worth recording against this comment, because the deployed env is a live counter-example to the happy path it describes.

That env declares a provider instance named `google` whose `provider_type` is also `google`. `google` is not an any-llm provider id (any-llm's is `gemini`), so this keys on `google`, misses, and draws a lettermark tile. Correct behavior, and the fallback running on real operator data rather than on a fixture.

The list page's Providers filter has the same limitation for the same reason, and cannot currently do better: the catalog *list* response carries only the instance name, not its type. Keying that rail on the type would need a field added to the response, which is out of scope here and belongs with the remaining surfaces on the issue.

web/src/shared/helpers/brandMarks.ts
jigjigjig: Fixed in 1694578d. You were right and your number was right: 103.10 kB was the first commit, where only one route imported the marks so the art inlined into the route chunk. By the second commit two routes imported it and Rolldown had hoisted it to a shared 64 kB chunk, still statically imported, so the defect held in a different shape.

Split so only the drawing defers. `brandMarkKeys.ts` is the synchronous half (which companies have a mark, ~3 kB) and the geometry is behind `lazy()`. Shared static chunk goes 64.17 kB -> 7.99 kB raw, 27.04 -> 3.07 gzip; route chunks unchanged. Fallback is the mark's own reserved box per performance.md's above-the-fold rule, so nothing shifts.

Verified in a browser rather than only in the bundle, since a chunk that never resolves would show empty boxes and pass every test: on a local env the glyph chunk is fetched once with a 200, 19 marks render, 0 reserved boxes left unfilled.

web/src/shared/helpers/brandMarks.ts
jigjigjig: Fixed in 1694578d. Reproduced before touching it: `providerMark("constructor")` returned `[Function Object]`, `hasProviderMark` then returned true, and the renderer threw on `glyph.shapes.map`.

Both lookups now go through `Object.hasOwn`, and the synchronous key lists are `Set`s rather than object literals, so the same hazard cannot come back through the new module. Pinned with a case covering `constructor`, `toString`, `__proto__` and `valueOf`.

web/src/shared/helpers/brandMarks.ts
jigjigjig: Fixed in 1694578d, and the docblock was indeed claiming something the code did not do. Each of the 44 distinct glyphs is now one declaration referenced from both tables. 90,408 -> 74,937 bytes, 15,471 saved, within a hair of your 15.5 kB.

The test is now identity rather than `toEqual`, derives its pairs instead of listing them, and asserts the count so an empty loop cannot pass. It also pins the one pair that must not share: maker `google` is the G, provider `gemini` is the sparkle.

web/src/shared/helpers/brandMarks.test.ts
jigjigjig: Fixed in 1694578d. You are right, and it is the failure mode I had been watching for elsewhere and missed here.

The gate is now `tests/unit/test_provider_mark_coverage.py`, following the pattern you pointed at: it reads the maker keys off `brandMarks.ts` and compares them against `_ORG_VENDORS` union `_KEY_VENDORS` slugged through `vendor_slug`, total in both directions, with the two tiled makers exempted by name and reason. Break-tested by deleting the `google` mark, which fails it.

The dashboard test now states that it is not the gate and points at the one that is.

web/src/features/models/ModelCatalogPage.tsx
jigjigjig: Fixed in 1694578d: the cell has a block body with one `const makerKey`, and the `?? ""` is gone with it.

web/src/shared/components/marks/BrandMark.tsx
jigjigjig: Fixed in 1694578d, both of them.

web/src/shared/helpers/brandMarks.ts
jigjigjig: Added in 1694578d. The header now separates the three: a mark is the glyph alone, a wordmark is the name drawn as art (which is why Amazon is excluded, its only asset in the set being the lowercase "aws" wordmark), and a lettermark is the fallback tile the component draws when no mark resolves.

The PR description is updated to use "mark" throughout rather than mixing it with "logo".

Issue #1431, opened 2026-09-21
Bring the dashboard overview into the target shape, and update the domain inventory for it


peteski22:
## What

The dashboard overview (#1428) landed in a shape that differs from the target in `docs/domains.md` in three places, and it left the inventory in that document out of date. None of it is a bug. It is listed here so it gets picked up deliberately and not rediscovered.

## The three differences

1. **`api/routes/overview.py` defines its three response models inline** (`WorstAllocationResponse`, `AllocationHealthResponse`, `OverviewSummaryResponse`). The target shape puts them in `schemas/overview.py`, and the Routes row says a route must not "define schemas inline". `schemas/budgets.py` is the worked example, and rule 16 in `scripts/check_architecture.py` holds that layer's imports.
2. **`repositories/overview/overview_repository.py` takes a raw `AsyncSession` and is not a `BaseRepository`.** The Repositories row says "every query, over `BaseRepository`". The practical difference: a `BaseRepository` reaches the session through the Unit of Work and raises outside a block, so a call site that forgets the block fails loudly. This one cannot.
3. **It reads another domain's tables directly**: `Budget`, `ScopedBudget`, `APIKey`, `Workspace`, `WorkspaceMember` and `users`. The target shape has a domain ask another domain's service for its data. This matters soonest for budgets: #1202 is moving the budget queries behind budgets repositories and a `BudgetService`, and #1313 moves the per-user budget state off `users`. Each of those will have to find and move these overview queries too, and nothing but a grep for `overview_repository` will point at them.

What it got right, for the record: it passes `scope_type` through as the stored string and writes no scope name of its own, so the one home for that vocabulary (`models/budgets.py`) still holds.

## The inventory in `docs/domains.md`

The "shape today" table and the domain list were not updated. Measured on `main` at `53617904`:

| Row | Says | Is |
| --- | --- | --- |
| Route modules | 72 | 74 |
| Route modules that define Pydantic models inline | 40 | 41 |
| Repository modules | 9: a base, `users_repository.py`, and 7 under `tenancy/` | 10, with `overview/overview_repository.py` |
| Service packages per domain | 1 (`services/tools/`) | `services/overview/` is a second |
| Repository packages per domain | None | `repositories/overview/` is the first |

[...]

PR #1438, opened 2026-09-21
refactor(overview): move the overview response models into schemas/overview.py

peteski22:
## Description

Nothing changes for someone who uses Otari. The dashboard overview endpoint answers with the same fields, and the published API description is identical.

The overview feature landed in a shape that differs from the target shape of a backend domain, and it left the domain inventory out of date. This PR does the cheap parts of the cleanup:

- The three response models of the overview move out of the route module into the schemas layer, where the target shape puts them. The budgets domain did the same move first.
- The "shape today" table in the domain inventory is measured again. Seven rows were stale. Five are the rows the issue lists. Two more showed up during the measurement: the count of service modules, and the count of service modules that run queries.
- The overview gets an entry of its own in the domain list. It is a read model across three domains, and nothing imports it, so it does not belong under Shared. The entry names the two places where its code still differs from the target shape.

One number differs from the issue. Route modules are 73, not 74, because the table has never counted `__init__.py`.

The repository base class and the cross-domain reads are not part of this PR. They stay in the issue, to be settled with the later budgets work.

## How to test it locally

Nothing to see in a browser, because the wire does not change. The proof is that three generated files do not move:

1. `make openapi-check` and `make postman-check` both pass.
2. `uv run python scripts/generate_openapi.py`, then `pnpm --dir web run client:generate`, then `git status`. No file changes.
3. `make lint` passes. It runs the architecture check, which holds the import rules of the schemas layer.
4. `uv run pytest tests/integration/test_overview_endpoint.py` passes. It covers the endpoint and its response shape.

I also ran `make typecheck` and `make test-unit`. I did not run the full integration suite locally.

To check the inventory numbers, count the modules under `src/gateway/` on `cf2968c1` and leave out every `__init__.py`.

## PR Type

- [ ] New Feature
- [ ] Bug Fix
- [x] Refactor
- [x] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Refs #1431. It stays open for the repository base class and the cross-domain reads.

## Checklist

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

Claude Code (Claude Fable 5.1)

I wrote the issue, set the scope and the commit split, and reviewed each commit before the push. The agent measured the inventory, moved the models, ran the checks and drafted this description. The extra commit that names the measured commit came out of my review.

**NOTE:**

## Summary

- Moved the overview response models into `schemas/overview.py`.
- Updated the route to import those models without changing the endpoint response fields or API description.
- Refreshed the domain inventory with corrected measurements and module counts.
- Added the overview read-model domain and documented its cross-domain data sources.

## Benefits

- Improves separation between routes and schemas.
- Keeps the public overview API unchanged.
- Provides a more accurate domain baseline.


tbille (approved):

docs/domains.md
```diff
@@ -55,28 +55,28 @@ not the module beside it.
...
 | Measure | Count |
 | --- | --- |
-| Service modules | 108, of which 71 sit flat at the top of `services/` |
-| Service modules that run queries | 39, plus 2 that only call `session.get` |
-| Route modules | 72 |
+| Service modules | 111, of which 71 sit flat at the top of `services/` |
+| Service modules that run queries | 38, plus 2 that only call `session.get` |
+| Route modules | 73 |
 | Route modules that run queries | 17, plus 1 that only calls `session.get` |
 | Route modules that define Pydantic models inline | 40 |
```
khaledosman, line 69: This row owes a decrement. `## What one domain change does` step 1 is exactly what this PR does for overview, and that section says "The counts above go down for the domain, and the pull request records them." `overview.py` defined three inline models on `cf2968c1` and defines none now, so this is 39.

The table also mixes two conventions as it stands: `Modules in schemas/` five lines down already names `overview.py`, which does not exist at `cf2968c1`, while this row is still the base measurement.

docs/domains.md
```diff
@@ -281,6 +281,21 @@ lives in the domain it watches and calls the alerts service.
...
+and budgets, so it has no tables of its own and writes nothing.
+
+- Routes: `overview.py`
+- Services: `overview/`
+- Repositories: `overview/`
+- Schemas: `overview.py`
+
+`overview/overview_repository.py` reads the tables of those three domains
+directly, and it takes the session instead of extending `BaseRepository`. The
+target shape has the overview service ask each domain's service for its data.
```
khaledosman, line 297: overview is the sixteenth domain now but does not appear in `## Order of work`, which still lists nine items. alerts is legitimately absent ("No code yet"); overview has code and this entry names two concrete pieces of remaining work, so it needs a slot in that list, or a sentence here saying why it sits outside the order.

khaledosman (approved):
Verified on the head commit: `scripts/generate_openapi.py` leaves `docs/public/openapi.json` byte-identical, the architecture check and ruff pass, and the reproducible inventory rows (111/71 service modules, 73 route modules, 38 + 17 query-running modules, 19 models, 10 repositories) all match. Two inline notes on the inventory table.

---
Reviewed by Claude Opus 5 (1M context) via [Claude Code](https://claude.com/claude-code).

docs/domains.md
peteski22: The decrement is in the PR, but the net diff hides it. The row said 40 on `main` and was stale: an AST count of route modules with a `BaseModel` subclass gives 41 on `cf2968c1`. `63e504e9` corrects it to 41, and `613f001b` takes it to 40 when the overview models move. So both this row and the `schemas/` row include this PR's change, which is what "updated by each domain change since" above the table means.

docs/domains.md
peteski22: Addressed in `25c2495f`: the overview entry now says why it has no slot of its own. Its queries move with each domain it reads, starting with budgets.

PR #1439, opened 2026-09-21
refactor(budgets): define the provider_key_id request constraints once

### LINKED ISSUE #1414: Define the provider_key_id request constraints once, not three times

peteski22:
## Description

Nothing changes for someone using Otari. Three request bodies let a caller narrow a budget ceiling to one provider, and each one wrote the same rule by hand: the value must not be blank, must hold no spaces, and must be 255 characters or fewer. The rule matters, because a blank value matches no provider, so a ceiling that stored one would be listed and never enforced. With three copies, a fix to one copy could miss the other two. The rule now has one definition that all three bodies use. Each body keeps its own published description.

## How to test it locally

- `uv run pytest tests/unit/test_budget_provider_narrowing.py`: holds the accepted values, the refused values and the published schema fixed on all three bodies. It was committed first and passed before the change.
- `make openapi-check && make postman-check`, then `pnpm --dir web run client:generate && git diff --exit-code web/src/client/`: all three report no diff, which proves the wire did not move.
- The blank-value integration tests in `test_scoped_budgets.py`, `test_organization_budgets.py` and `test_workspace_member_budget_policies.py` still pass.

## PR Type

- [ ] New Feature
- [ ] Bug Fix
- [x] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Closes #1414. Follow-up from #1202.

## Checklist

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

Claude, via Claude Code.

I wrote the issue and set the constraints for this change: tests first, small commits, and no change on the wire. Claude wrote the code and the tests under that direction. I reviewed the commits before anything was pushed.

**NOTE:**

## Summary

- Centralized `provider_key_id` validation for three budget request models.
- Preserved each model’s published description and existing API behavior.
- Added coverage for valid, invalid, omitted, and boundary values.
- Verified that the shared definition keeps the published schema constraints consistent.

## Benefits

Future constraint changes now require one update instead of three, which reduces drift risk.


src/gateway/schemas/budgets.py
```diff
@@ -29,6 +30,9 @@
     "to express a calendar month. Mutually exclusive with budget_duration_sec"
 )
 
+# A blank value matches no provider instance, so a ceiling that stored one would never bind.
+_ProviderKeyId = Annotated[str | None, Field(min_length=1, max_length=255, pattern=r"^\S+$")]
```
khaledosman: The alias absorbs `| None`, so the three field declarations no longer show that the field is nullable while still passing `default=None`. Every other optional field in this module spells `X | None` at the declaration (`name`, `max_budget`, `token_limit`, and the `Update` bodies below), so this is now the one annotation in the file where you have to jump to the alias to learn the field is optional.

Keeping the constraint shared and the nullability local:

```python
_ProviderKeyId = Annotated[str, Field(min_length=1, max_length=255, pattern=r"^\S+$")]
```

with `provider_key_id: _ProviderKeyId | None = Field(default=None, description=...)` at each of the three sites. I checked the published schema for both forms: same `anyOf: [{string + constraints}, {null}]`, same `default: null`, so the wire and `test_the_published_narrowing_constraints_are_the_same_on_every_body` are unaffected either way.

tests/unit/test_budget_provider_narrowing.py
```diff
@@ -0,0 +1,51 @@
...
+) -> None:
+    assert body.model_validate({**required, "provider_key_id": accepted}).model_dump()["provider_key_id"] == accepted
+
+
+@pytest.mark.parametrize(("body", "required"), _NARROWING_BODIES)
+def test_an_omitted_narrowing_is_none(body: type[BaseModel], required: dict[str, Any]) -> None:
+    assert body.model_validate(required).model_dump()["provider_key_id"] is None
+
+
+@pytest.mark.parametrize(("body", "_required"), _NARROWING_BODIES)
```
khaledosman: `_required` is never used in this test; it is only carried because the parametrize reuses `_NARROWING_BODIES`. Parametrize over the classes alone:

```python
@pytest.mark.parametrize("body", [body for body, _ in _NARROWING_BODIES])
def test_the_published_narrowing_constraints_are_the_same_on_every_body(body: type[BaseModel]) -> None:
```

khaledosman (approved):
Verified locally at `45778cc`: `make lint` clean (architecture check included), `scripts/generate_openapi.py` leaves `docs/public/openapi.json` byte-identical, `make openapi-check` passes, and the 30 new unit tests pass. `pattern=r"^\S+$"` had exactly three occurrences on `main` and all three are folded in, so nothing was left behind. Two non-blocking nits inline.

src/gateway/schemas/budgets.py
peteski22: Done in c4766044. I checked both forms first: the published schema is byte-identical and the same values pass and fail. These three were the only `default=None` fields in the module without `| None` in the annotation.

tests/unit/test_budget_provider_narrowing.py
peteski22: Done in e1f93925, with your parametrize.

peteski22:
@coderabbitai Added `"x"` in 378c4f4d. One correction: a changed `min_length` was already caught. With `min_length=2`, `test_the_published_narrowing_constraints_are_the_same_on_every_body` fails on all three bodies. The new case covers the behavior side.

PR #1440, opened 2026-09-21
refactor(budgets): give the two scoped ceiling responses a shared base

### LINKED ISSUE #1415: Give the two scoped ceiling response models a shared base

peteski22:
## Description

Two API responses describe the same thing, a spend ceiling and its live counters: one for the whole deployment and one inside an organization. The organization response repeated all 21 fields of the other and added one, `manageable`. A field added to one and not the other was the likely mistake, and nothing would catch it. Both responses now share one base. The two update bodies were the same case in miniature and share a base too. All four published names and descriptions stay the same.

One thing moves on the wire. In the organization ceiling response, `manageable` is now the last key, and it is last in that schema's `required` list in the OpenAPI spec. JSON objects are unordered, so no client is affected. The alternative was a base that stops before the timestamps, which keeps the order and leaves two fields repeated. I chose the cleaner code.

This PR followed #1439, which is merged. The branch is rebased on `main` and holds only its own commits.

## How to test it locally

- `uv run pytest tests/unit/test_scoped_ceiling_responses.py`: holds the published field order, the values each response produces and the update body fields fixed. It was committed first, against the old code.
- `make openapi-check && make postman-check`, then `pnpm --dir web run client:generate && git diff --exit-code web/src/client/`: the regenerated spec is committed, and its whole diff is the one moved `required` entry. The Postman collection and the dashboard client regenerate with no diff.
- `test_scoped_budgets.py` and `test_organization_budgets.py` still pass.

## PR Type

- [ ] New Feature
- [ ] Bug Fix
- [x] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Closes #1415. Follow-up from #1202.

## Checklist

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

Claude, via Claude Code.

I wrote the issue and set the constraints: tests first, small commits, and stop if the spec moves. Claude wrote the code and the tests under that direction. Claude's first version kept the field order with a smaller base and flagged the trade. I chose the full base and accepted the moved key, and that commit was rebuilt. I reviewed the commits before anything was pushed.

**NOTE:**

## Summary

- Added shared bases for the deployment and organization scoped budget response models.
- Added a shared base for both scoped budget update request models.
- Preserved the four published model names, descriptions, fields, and API behavior.
- Added tests for field order, model conversion, and update request fields.
- Exported the new shared base models.
- Renamed `ScopedCeilingFigures` to `ScopedBudgetFigures` for consistency.

## Benefits

This reduces duplicated schema and conversion logic. It also helps keep deployment and organization models consistent as they change.


src/gateway/schemas/budgets.py
```diff
@@ -222,32 +218,50 @@ class ScopedBudgetResponse(BaseModel):
...
-            current_requests=budget.current_requests,
-            reserved_requests=budget.reserved_requests,
-            budget_duration_sec=limit.budget_duration_sec,
-            reset_alignment=limit.reset_alignment,
-            period_start=budget.period_start.isoformat() if budget.period_start else None,
-            period_end=budget.period_end.isoformat() if budget.period_end else None,
-            created_at=budget.created_at.isoformat(),
-            updated_at=budget.updated_at.isoformat(),
-        )
+        return cls(**cls._figures_of(budget, limit))
```
khaledosman: `_figures_of(ceiling, budget)` takes the ceiling first, but this signature calls the ceiling `budget` and the budget `limit`, so `budget` names both models three lines apart. Rename the params to `ceiling, budget` to match the helper and `OrganizationScopedBudgetPublic.from_model`; all four call sites pass positionally (`scoped_budgets.py:159,185,201,256`), so the rename is safe today, and it is the version where a later switch to keyword args cannot silently swap the two.

src/gateway/schemas/budgets.py
```diff
@@ -181,24 +182,19 @@ class CreateScopedBudgetRequest(BaseModel):
...
 
-    Unlike ``/api/v1/budgets``, the counters are the row's own: a scoped ceiling is
-    enforced against ``current_spend + reserved_spend``, so there is no rollup
-    over users to compute.
 
-    Every limit, along with ``budget_duration_sec`` and ``reset_alignment``, is
-    read off the budget rather than stored here, and carried on the wire so a
-    caller can render a ceiling without fetching every budget to resolve one id.
-    """
+class ScopedCeilingFigures(BaseModel):
```
khaledosman: Two adjacent new bases name the same entity with two different nouns: `ScopedBudgetChanges` (line 185) and `ScopedCeilingFigures`. `ScopedBudgetFigures` would match the model, both subclasses and the existing `OrganizationBudgetRates` precedent.

khaledosman (approved):
Reviewed as the increment over #1439 (`45778ccf..b6e9c7a3`). The spec regeneration is confined to the one moved `required` entry: `properties` is emitted alphabetically, so `web/src/client/schema.ts` and the four SDK codegen jobs see no change, and `openapi-spec` (both checks) is green on the head. Two naming notes inline, neither blocking.

---
Reviewed by Claude Opus 5 (1M context) via [Claude Code](https://claude.com/claude-code).

src/gateway/schemas/budgets.py
peteski22: Done in 07ff4545. I checked the callers first: all six pass both values by position, in `src` and in the tests, so no call changed.

src/gateway/schemas/budgets.py
peteski22: Done in 3b097ae9: `ScopedBudgetFigures`. The base publishes no component, so the spec did not move.

peteski22:
@coderabbitai review

PR #1441, opened 2026-09-21
refactor(budgets): gather the budget services into one package

peteski22:
## Description

The seven budget service modules move into one package, `gateway.services.budgets`. Nothing changes for someone using Otari: no route, status code, field, error message or behavior moves, and the OpenAPI spec, the Postman collection and the generated dashboard client are untouched. A syntax-tree comparison of the seven files against `main` shows no executable line changed. Only import lines, docstrings and comments differ.

Python import paths do change, and there are no shims:

| Was | Is |
| --- | --- |
| `services/budget_periods.py` | `services/budgets/_periods.py` |
| `services/budget_retiming.py` | `services/budgets/_retiming.py` |
| `services/budget_service.py` | `services/budgets/_reservations.py` |
| `services/scoped_budget_service.py` | `services/budgets/_scoped_enforcement.py` |
| `services/budget_reservation_ledger.py` | `services/budgets/_ledger.py` |
| `services/tenancy/organization_budget_service.py` | `services/budgets/_organization_surface.py` |
| `services/tenancy/workspace_budget_default_service.py` | `services/budgets/_member_policies.py` |

Callers import from the package root, `from gateway.services.budgets import ...`, which exports the 20 names they use today. The modules are private, and only white-box tests name one. Every importer and every `patch()` string in this repository is repointed. This is the third step of the budgets domain in #1202, after the errors and the schemas: the files are gathered first and unchanged, so that the later PRs, which put `BudgetService` on repositories and the Unit of Work, show as changes to code and not as moves.

**The otari-ai overlay imports some of these names.** `ReservationHandle`, `reconcile_reservation`, `ApplicableBudget`, `BudgetScopeRequest`, `applicable_budgets` and `WorkspaceBudgetDefaultService` all become `from gateway.services.budgets import ...`. otari-ai pins released versions of otari, so that change follows the first release that contains this PR.

Three baselines in `scripts/check_architecture.py` change, and no rule does:

- `SERVICE_DATABASE_IMPORT_BASELINE` renames six entries and `TRANSACTION_CONTROL_BASELINE` renames five. Those modules still import SQLAlchemy and still call `commit()` at their new paths. Neither count changes.
- `FLAT_MODULE_BASELINE` deletes five entries. A module inside `services/budgets/` is no longer flat, and a renamed entry would grandfather a file against a rule it does not break.

About thirty docstrings and comments named one of the seven modules, most of them as `budget_service.<function>`. A `git mv` does not touch prose, so each one was stale after the move. Each is cut to contract and intent and not patched by one word, which is why five of the twelve commits are `docs`. Two things in them are worth a reviewer's eye:

- The dispatch comment in `_pipeline.py` warned that `refund_reservation` is not idempotent. It has been idempotent by reservation identity since the ledger landed, and its own docstring says so, so the warning is gone.
- Five of the texts sit outside the budgets domain, in four files under `services/tenancy/`. They are in one commit, `docs(tenancy): ...`, so it can be dropped alone.

One stale path is left on purpose, in the docstring of migration `f7a2c4e6b8d1`, because a migration is a historical record.

`docs/domains.md` lists `services/budgets/` for the domain and counts it as the third domain package. The flat service module count drops from 71 to 66.

## How to test it locally

```
make lint && make typecheck
env -u OTARI_API_KEY make test-unit
uv run pytest -n auto tests/integration
pnpm --dir web run client:generate && git diff --exit-code web/src/client/schema.ts
```

All of those pass on this branch, with no diff in either generated artifact or in the dashboard client, which is the proof that nothing moved on the wire. The one exception is `tests/integration/test_mcp_dependency_ceiling.py`, which installs `mcp` into a throwaway venv and failed here on creating that venv. `AGENTS.md` lists it as environment noise.

No new test is needed for a move, and no assertion about behavior changed. Two tests carry the weight:

- `tests/unit/test_service_module_imports.py` imports the package root and each private module first in a fresh interpreter, which is the only way to see an import cycle. It also asserts that no organizations module loads anything under `gateway.services.budgets`.
- `tests/unit/test_check_architecture.py` names the package in its fixture source text, which the checker parses and never imports. One expected violation string changes with its fixture.

To see that the move is a move: `git diff -M main... --summary -- src/gateway/services` shows seven renames, and `git grep -n -E "\b(budget_service|scoped_budget_service|budget_reservation_ledger|budget_periods|budget_retiming)\b" -- src tests scripts docs` finds nothing.

Not every commit is green, and each message says which checks pass at it. The app does not import at the first commit, because callers still name the old paths. `make lint` passes from the baselines commit on.

## PR Type

- [ ] New Feature
- [ ] Bug Fix
- [x] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Relevant issues

Part of #1202. Follows #1325, #1394 and #1417. #1416 (the reservation status names still sit in `_ledger.py`) comes straight after this one, because it edits a file this PR moves.

## Checklist

## AI Usage

- [ ] No AI was used.

**AI Model/Tool used:**

Claude, via Claude Code.

This was shared work. I designed the change: the package layout, the names the root exports, how each of the three baselines moves, and the order of the steps all come from the plan I wrote for #1202, and I checked that plan against `main` before the work started. I also set the rules the work ran under: one small commit per step, and any comment or docstring the change touches is rewritten whole to the project's comment standard. Claude carried out the move under that direction, ran the checks, and drafted the rewritten texts and this description. I reviewed all twelve commits, and every added or touched comment and docstring, before anything was pushed.

## Summary

- Moved seven budget service modules into the `gateway.services.budgets` package.
- Added package-level exports for the 20 names used by callers.
- Updated application imports, tests, patch paths, architecture checks, and documentation.
- Removed compatibility shims for the old Python import paths.
- Preserved runtime behavior and external contracts.

This structure gives the budget domain a clearer boundary and makes future budget changes easier to manage.

## Technical notes

- Internal modules are now private, including `_ledger`, `_reservations`, `_periods`, `_retiming`, and `_scoped_enforcement`.
- OpenAPI, Postman, dashboard client output, routes, status codes, fields, and error messages remain unchanged.
- Test and lint results are not independently reported here.


src/gateway/api/routes/organization_budgets.py
```diff
@@ -1,31 +1,9 @@
...
-cannot name an organization at all, because the caller's identity already points
-at one, so there is no parameter that could be confused with an authorization
-decision.
-
-These sit *beside* ``/api/v1/budgets`` and ``/api/v1/scoped-budgets``, which stay the
-deployment's own surface behind ``require_deployment_operator``. The tables are
-shared; what differs is which rows a caller may reach. A budget with no
-``organization_id`` is the deployment's, and no route here lists, offers or
-repoints one.
+"""The caller's organization's spend budgets and ceilings.
```
khaledosman, line 1: The docstring no longer says which modes mount this router. The old one spelled out standalone **and hosted** precisely because the shorthand the neighbouring `/organizations/me` routers use names the wrong audience here. `scoped_budgets.py:3` and `workspace_member_budget_policies.py:1` dropped "standalone mode only" the same way. Mounting is decided in `_register_core_routers` and is not visible from the module, so put one line back in each of the three.

src/gateway/models/budgets.py
```diff
@@ -329,9 +329,7 @@ class BudgetReservation(Base):
     # ``estimate > 0`` because a zero-cost request on an enforced budget still
     # takes the hold, and the release has to match what the reserve did.
     user_reserved: Mapped[bool] = mapped_column(default=False, server_default=false())
-    # A plain string rather than a database enum, matching ``scoped_budgets.scope_type``:
-    # a new state should not need an enum migration. Values are the
-    # ``RESERVATION_*`` constants in gateway.services.budget_reservation_ledger.
+    # The status is a plain string and not a database enum, so a new state needs no enum migration.
```
khaledosman, line 332: The pointer to the `RESERVATION_*` constants was deleted rather than repointed to `services/budgets/_ledger.py`, so the column's legal values are now unreferenced from the model. Repointing a stale path is what this PR is for, and #1416 moves those constants next.

src/gateway/services/budgets/_scoped_enforcement.py
```diff
@@ -493,22 +472,9 @@ async def blocked_axis(
...
-
-    Both of the clauses :func:`admits` builds, and for its reason: an arrival is
-    refused either because the axis is already at its cap or because this hold
-    would push it past, and a helper reproducing one of them finds no axis for
-    the other shape. ``amount=None`` skips the dollar axis, which :func:`reserve`
-    likewise does not ask about. The hold therefore has to come from the caller, since the
-    row cannot say what was being asked of it. The dollar axis is called
-    "budget", so its message is unchanged; see
-    ``budget_service._blocked_axis``.
+    """Return the name of the capped axis that left this ceiling no room.
```
khaledosman, line 475: The rewrite dropped the one invariant here that is not visible from the code: `blocked_axis` has to reproduce both clauses `admits` builds (the axis is already at its cap, and this hold would push it past). A version carrying only one of them finds no axis for the other shape, which surfaces as an empty or wrong 403 reason rather than an error. Nothing else in the tree records that coupling.

src/gateway/services/budgets/_reservations.py
```diff
@@ -477,47 +464,16 @@ async def reserve_budget(
...
-    the request never gets back to is reclaimable on its own rather than only in
-    aggregate. ``record_reservation=False`` is for :func:`increase_reservation`,
-    which grows the row this request already has instead of opening a second one.
+    Raises ``HTTPException`` 404 for an unknown user and 403 when the user is blocked or a budget has no room.
+    No row lock is held across the provider call.
+    ``organization_id`` selects the rate overrides that decide whether ``model`` is free.
+    Without it the deployment price list decides.
+    ``requests`` is 1 for a request that takes its own hold and 0 for a top-up.
+    ``new_request=False`` marks a top-up, which is checked only for whether the delta fits.
+    ``scope`` adds the tenancy-scoped ceilings, and every one of them must also admit the estimate.
```
khaledosman, line 474: Two caller-visible facts went with the rewrite. Scoped ceilings are held first and released again if the per-user gate then refuses, which is what keeps a rejected request from leaving a counter behind. And a caller that passes no token estimate holds nothing on that axis, so a token ceiling binds it at settlement rather than at admission. `src/gateway/AGENTS.md` carries the total order and compensation *within* scoped reservations, not the ordering between the two mechanisms, so neither survives elsewhere.

src/gateway/services/tenancy/workspace_code_execution_policy_service.py
```diff
@@ -1,46 +1,15 @@
...
-workspace comes off the key, never off a header (``services/workspace_scope``).
-It is called from ``prepare_gateway_tools`` at admission, where the request's
-session is live, and its values land on ``ToolContext``.
+The sandbox stays an operator concern, and a policy decides who may ask for code execution and within which limits.
+A workspace row may veto and may refine, and it never grants.
+``enabled=False`` refuses ``otari_code_execution`` for the workspace.
+``max_iterations`` and ``exec_timeout_s`` can only lower what the request would otherwise get.
+``default_purpose_hint`` applies only when the request names none.
+``tools`` intersects the tool kinds the sandbox backend serves, and an empty result refuses the request.
+``image`` may name only an image on the operator's allow-list.
```
khaledosman: The allow-list lost its reason: a workspace-settable image is a supply-chain surface, which is the sentence that stops someone relaxing this column later. The #655 and #678 provenance for veto-and-refine went with it, and the same `docs(tenancy)` commit drops otari#1011 from `services/tenancy/authorization.py:1`. This file's only tie to the PR is one clause naming `workspace_budget_default_service`. The description says that commit can be dropped alone; either drop it or keep the why clauses and the issue references.

src/gateway/api/routes/mcp.py
```diff
@@ -320,22 +320,10 @@ async def _authenticate(
...
-    workspace, which would let operator credentials reach one tenant's
-    configured servers. No stored server is accessible to it, and that is what
-    the 404 says.
-
-    A blocked user is refused too, and has to be refused here. ``users.blocked``
-    is read in ``budget_service.reserve_budget``, which is the only place any
-    request plane consults it, and these routes never reach it because they
-    reserve nothing. Without this, blocking someone would stop their completions
-    and leave them driving mutating MCP tools through the gateway.
+    In hybrid mode the user token passes through to the platform resolver, which authorizes and rate-limits.
```
khaledosman: Same shape in `docs(gateway)`: the R-ADM-2 reference and "Otari preserves whatever 429 it returns" are both gone, and neither is derivable from the body below. The blocked-user and master-key reasons survive the cut well; these two do not.

khaledosman (approved):
Verified the move is behavior-preserving: an AST comparison of the seven modules against `main` shows only import paths differ, every name in the new `__all__` has a caller outside the package, the three baseline edits are renames and not additions, and `scripts/check_architecture.py` passes on the head.

The findings below are all in the docstring and comment rewrites, none in the move.

src/gateway/api/routes/organization_budgets.py
peteski22: Added back, one line in each of the three docstrings.

src/gateway/models/budgets.py
peteski22: Repointed to `services/budgets/_ledger.py`. #1416 moves the constants into `models/budgets.py` next, and the line moves with them.

src/gateway/services/budgets/_scoped_enforcement.py
peteski22: Right, and nothing else recorded it. The docstring now says the test must mirror both clauses of the admission test in `reserve`. It names `reserve` because `admits` is nested inside it.

src/gateway/services/budgets/_reservations.py
peteski22: Both facts are back, one sentence each.

src/gateway/services/tenancy/workspace_code_execution_policy_service.py
peteski22: The supply-chain reason is back on the `image` line. I left the issue numbers out on purpose: I do not put issue numbers in code or comments. I kept the commit, because dropping it leaves four files that name a module that no longer exists.

src/gateway/api/routes/mcp.py
peteski22: The 429 sentence is back. So is R-ADM-2.

PR #1446, opened 2026-09-21
feat(policy-checks): evaluate otari hook policies locally by default

agpituk:
## Summary
- `otari hook` now evaluates `.otari-gates.yml` in process by default, needing no running `otari serve` and no credential. The parse-and-evaluate logic that used to live inline in `routes/hooks.py` is now a shared, pure function (`agent_runtime.domain.check.run_policy_check`) that both the Hook Server route and the CLI call.
- `POST /api/v1/hooks/check` stays as an opt-in path via `--url`/`--api-key` (or `OTARI_URL`/`OTARI_API_KEY`), for a caller that wants a shared or hosted gateway to decide instead.
- `otari hook setup` no longer resolves or prompts for a credential by default; it only embeds one when `--api-key` is given explicitly.
- Also carries the small Codex-harness commit from `hooks-server-persistency` that had not gone through a PR yet.
- Regenerates `docs/public/openapi.json` and the Postman collection for the `check_policy` route's updated docstring.

## Test plan
- [x] `make lint` (architecture check + Ruff)
- [x] `uv run mypy` (whole project)
- [x] `tests/unit/test_hook_cli.py`, `test_hook_cli_codex.py`, `test_hook_judge_cli.py`, `test_hook_setup_cli.py`, `tests/unit/agent_runtime/` (315 tests)
- [x] `tests/integration/test_hooks_route.py` (58 tests, unchanged HTTP contract)
- [x] `make openapi-check` / `make postman-check`
- [x] Manually verified end to end: registered the hook via `otari hook setup` with no server running and no credential, then triggered real `Edit`/`Bash` tool calls that were blocked by the local evaluator before they ran.


agpituk:
Closing: this bundled the still-unreviewed Codex-harness commit from hooks-server-persistency with the local-eval work. Splitting so the Codex harness change gets its own PR first (see #<will-link>), and the local-eval PR follows once that merges.

